PR #14116, opened 2019-06-07
Do not set a cookie if it finally has not changed

jocel1:
| Questions     | Answers
| Branch?       | develop
| Description?  | Do not set a cookie which has finally not changed
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no

This modification could be useful if we are modifying several times the content of the prestashop cookie, to finally put at the end the same value than at the beginning.

Let's say for exemple we have 

$this->context->cookie->last_visited_category == null

The front ProductController sets $this->context->cookie->last_visited_category = 2;
Hence _modified inside the cookie will be set to true.
Then we put back $this->context->cookie->last_visited_category = null; through a module

The _modified flag will still be true, and hence the cookie will be written even if there's no change inside it. This PR modify this behavior to avoir writing an unchanged cookie.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14116)

### DIFF
--- a/classes/Cookie.php
+++ b/classes/Cookie.php
@@ -298,6 +298,7 @@ class CookieCore
 
             /* Get cookie checksum */
             $tmpTab = explode('¤', $content);
+            // remove the checksum which is the last element
             array_pop($tmpTab);
             $content_for_checksum = implode('¤', $tmpTab) . '¤';
             $checksum = crc32($this->_salt . $content_for_checksum);
@@ -386,10 +387,11 @@ class CookieCore
             return;
         }
 
-        $cookie = '';
+        $previousChecksum = $cookie = '';
 
         /* Serialize cookie content */
         if (isset($this->_content['checksum'])) {
+            $previousChecksum = $this->_content['checksum'];
             unset($this->_content['checksum']);
         }
         foreach ($this->_content as $key => $value) {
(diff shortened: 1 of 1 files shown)


matks:
Makes sense ✅ 

Travis failed for bad reasons, I restart it

matks (approved):

matks:
Just curious, how did you end up thinking about this 😄? Did it impact badly performances on a shop you work on ?

PierreRambaud (approved):

jocel1:
@matks Well actually I have a customer with the ps_categorytree module enabled, and it sets 
```$this->context->cookie->last_visited_category = (int)$product->id_category_default;```

In that case, because a cookie is set, reverse proxy cannot cache the page. So I wanted to add an option to always reset the last_visited_category to null in my Varnish module, but because we just check the _modified flag, the cookie was still sent :)

matks:
@PierreRambaud I think no QA is needed, is it ?

PierreRambaud:
@matks I was thinking about it, maybe just check everything still working as expected is enough?

matks:
Thank you @jocel1

mvorisek:
Instead of `crc32` `md5` or even `hash('sha256', )` should be used to prevent collisions, i.e. to prevent not refreshing the cookie.

PR #14117, opened 2019-06-08
CO: Fix getDeliveryOption for multiple instances of different Cart

kijamve:
| Questions     | Answers
| Branch?       | develop and all version of PS
| Description?  | Fix getDeliveryOption for multiple instances of different Cart in same code
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no

If the same code is present this code:

`$cart1 = new Cart(xx);`
`$d1 = $cart1->getDeliveryOption();`
`$cart2 = new Cart(yy);`

This line cause one problem:

`$d2 = $cart2->getDeliveryOption(); //<<<--- This return delivery_option of $cart1!...`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14117)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -3111,6 +3111,7 @@ class CartCore extends ObjectModel
         if (empty($delivery_option) || count($delivery_option) == 0) {
             $this->delivery_option = '';
             $this->id_carrier = 0;
+            static::$cacheDeliveryOption = array();
 
             return;
         }
@@ -3134,6 +3135,7 @@ class CartCore extends ObjectModel
         }
 
         $this->delivery_option = json_encode($delivery_option);
+        static::$cacheDeliveryOption = array();
 
         // update auto cart rules
         CartRule::autoRemoveFromCart();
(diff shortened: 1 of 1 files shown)


classes/Cart.php
```diff
@@ -3173,7 +3175,7 @@ protected function getIdCarrierFromDeliveryOption($delivery_option)
      */
     public function getDeliveryOption($default_country = null, $dontAutoSelectOptions = false, $use_cache = true)
     {
-        $cache_id = (int) (is_object($default_country) ? $default_country->id : 0) . '-' . (int) $dontAutoSelectOptions;
+        $cache_id = (int) (is_object($default_country) ? $default_country->id : 0) . '-' . (int) $dontAutoSelectOptions . '-' . (int) $this->id;
```
PierreRambaud, line 3178: I think, only this line is needed. :thinking:

kijamve: But if something is build in this order: get, set, get. The second get will not work.

PierreRambaud: So, the cacheId should be dynamicly built and use, or maybe it's time to remove static cache :+1:

matks:
Hello,

Since we had no news from you for more than a year, I'll close this pull request. It does not mean we reject this contribution. If you are willing to continue this topic, feel free to reopen or open another one if you think it's still relevant. 😉 

Thanks!

PR #14137, opened 2019-06-10
Adds missing form_rest in Shop Parameters > Preferences

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Adds missing form_rest in `PrestaShopBundle/Resources/views/Admin/Configure/ShopParameters/preferences.html.twig`
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? |  
| How to test?  | No current visible changes in front-end. When form contains fields that are not rendered manually in twig(e.g. if form is extended by module), those fields should be automatically rendered in same block as all general fields.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14137)

### DIFF
--- a/src/PrestaShopBundle/Resources/views/Admin/Configure/ShopParameters/preferences.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Configure/ShopParameters/preferences.html.twig
@@ -33,14 +33,14 @@
 {% block content %}
     {{ form_start(form, {'attr' : {'class': 'form', 'id': 'configuration_form'} }) }}
     <div class="row justify-content-center">
-        {% block preferences_form_general %}
-            <div class="col-xl-10">
-                <div class="card">
-                    <h3 class="card-header">
-                        <i class="material-icons">settings</i> {{ 'General'|trans({}, 'Admin.Global') }}
-                    </h3>
-                    <div class="card-block row">
-                        <div class="card-text">
+        <div class="col-xl-10">
+            <div class="card">
+                <h3 class="card-header">
+                    <i class="material-icons">settings</i> {{ 'General'|trans({}, 'Admin.Global') }}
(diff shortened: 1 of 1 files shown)


sarjon:
@matks can you merge it today? I need this by tomorrow. :smile:

matks (approved):

src/PrestaShopBundle/Resources/views/Admin/Configure/ShopParameters/preferences.html.twig
```diff
@@ -165,6 +165,9 @@
                                     {{ form_widget(generalForm.shop_activity) }}
                                 </div>
                             </div>
+                            {% block shop_preferences_form_rest %}
+                              {{ form_rest(form) }}
```
PierreRambaud, line 169: mixin 2 spaces & 4 spaces.
We need to do something for our templates :)

sarjon: Agree, but do you think it can be done in separate PR for all template files with some automatic tool?

PierreRambaud: Yes of cour,se it's just a reminder :)

matks:
Thanks @zuk3975

src/PrestaShopBundle/Resources/views/Admin/Configure/ShopParameters/preferences.html.twig
matks: https://github.com/PrestaShop/PrestaShop/issues/14146

PR #14138, opened 2019-06-10
Add comment in Customer registration form, add better error message

### LINKED ISSUE #13640: Better alert for customer creation and customer info edit forms
Following 1.7.5.2 patch, when an existing customer having a first name such as "John R.Hamond" wants to modify its account info an alert is raised when saving saying "Incorrect name".
This is because his/her name doesn't match the new forms' conditions.
Problem is that this name used to be good and is not anymore, we need to explain it.

Capture d’écran 2019-05-07 à 09 39 17

This alert is not explicit enough as the name was previously good and not anymore. We need to inform the customer about new conditions introduced in 1.7.5.2 .

**Todo:**
- [X] Study feasibilty
- [ ] Add new error message 

It is first necessary to study the complexity to modify this text. The message might come from the database content validation and not from the form itself. In the first case, it would require to create a specific event.

**If the message is generated at the form level:**
An alert with more context/information about fields conditions for the "First name" and "Last name" fields:
"
Invalid name.
The following characters are not accepted: / \ ` ^ *
The characters . and 。 are allowed only if they are directly followed by a space or the end of the string.
"

**If the message is generated at the db level:**
Add a message below the fisrtname and lastname field. The message should always be displayed, like in this example: 
57281825-06116580-70ac-11e9-841d-acd7482c85fc

Text: 
"
not accepted: / \ ` ^ *
A space is necessary after a "."
"

---
Should be added to the account creation form, the edit account and the form in the checkout.

Thanks,

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Add comment in Customer registration form, add better error message
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #13640
| How to test?  | Try to register a new customer, you should see comments under Firstname and Lastname fields If trying to send an invalid value `John R.Hamond` a more detailed error message is displayed

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14138)

### DIFF
--- a/classes/ValidateConstraintTranslator.php
+++ b/classes/ValidateConstraintTranslator.php
@@ -49,9 +49,27 @@ class ValidateConstraintTranslatorCore
      */
     public function translate($validator)
     {
-        if ($validator === 'isName' || $validator === 'isCustomerName') {
+        if ($validator === 'isName') {
             return $this->translator->trans(
                 'Invalid name',
+                [],
+                'Shop.Forms.Errors'
+            );
+        }
+
+        if ($validator === 'isCustomerName') {
+            return $this->translator->trans(
+                'Invalid name',
+                [],
+                'Shop.Forms.Errors'
+            ) . PHP_EOL .
+            $this->translator->trans(
+                'Invalid characters: 0-9!<>,;?=+()@#"°{}_$%/\^*`',
+                [],
+                'Shop.Forms.Errors'
+            ) . PHP_EOL .
+            $this->translator->trans(
+                'A space is required after . and 。',
                 [],
                 'Shop.Forms.Errors'
             );
--- a/themes/classic/templates/_partials/form-errors.tpl
+++ b/themes/classic/templates/_partials/form-errors.tpl
@@ -27,7 +27,7 @@
     {block name='form_errors'}
       <ul>
         {foreach $errors as $error}
-          <li class="alert alert-danger">{$error}</li>
+          <li class="alert alert-danger">{$error|nl2br nofilter}</li>
         {/foreach}
       </ul>
     {/block}


classes/ValidateConstraintTranslator.php
```diff
@@ -49,14 +49,32 @@ public function __construct(TranslatorInterface $translator)
...
                 'Shop.Forms.Errors'
             );
         }
 
+        if ($validator === 'isCustomerName') {
+            return $this->translator->trans(
+                'Invalid name',
+                [],
+                'Shop.Forms.Errors'
+            ) . "\n" .
```
PierreRambaud: use `PHP_EOL` instead

classes/form/CustomerFormatter.php
```diff
@@ -133,6 +137,10 @@ public function getFormat()
                     'Shop.Forms.Labels'
                 )
             )
+            ->addAvailableValue(
+                'comment',
+                $this->translator->trans("Not accepted: / \ ` ^ *\nA space is necessary after a \".\"", array(), 'Shop.Forms.Help')
```
PierreRambaud:
```suggestion
                $this->translator->trans("Not accepted: / \ ` ^ *\nA space is necessary after a \".\"", [], 'Shop.Forms.Help')
```

classes/ValidateConstraintTranslator.php
PierreRambaud: use `PHP_EOL` instead

themes/classic/templates/_partials/form-errors.tpl
```diff
@@ -27,7 +27,7 @@
     {block name='form_errors'}
       <ul>
         {foreach $errors as $error}
-          <li class="alert alert-danger">{$error}</li>
+          <li class="alert alert-danger">{$error|strip_tags|nl2br nofilter}</li>
```
PierreRambaud: why strip_tags?

jolelievre: to be sure no html tag is inserted (since I use `nofilter`) only the converted `\n` remain

classes/ValidateConstraintTranslator.php
jolelievre: Done

classes/form/CustomerFormatter.php
jolelievre: Done

themes/classic/templates/_partials/form-errors.tpl
jolelievre: I can remove it, it will still work but I was worried of XSS injection 😛

PierreRambaud: only if someone include xss in translations, but we should be worried if it happens ^^

jolelievre: alright I'll remove it then

themes/classic/templates/_partials/form-fields.tpl
```diff
@@ -170,7 +170,7 @@
           >
           {if isset($field.availableValues.comment)}
             <span class="form-control-comment">
-              {$field.availableValues.comment}
+              {$field.availableValues.comment|nl2br nofilter}
```
jolelievre: @PierreRambaud I also removed this `strip_tag`

themes/classic/templates/_partials/form-errors.tpl
jolelievre: Done

PierreRambaud (dismissed):
I don't like the concat but not other way from now :sob:

jolelievre:
> I don't like the concat but not other way from now 😭

Me neither, but if I return an array the error blocks are separated.. This would require to change too many templates and error management..

classes/ValidateConstraintTranslator.php
```diff
@@ -49,14 +49,32 @@ public function __construct(TranslatorInterface $translator)
...
         }
 
+        if ($validator === 'isCustomerName') {
+            return $this->translator->trans(
+                'Invalid name',
+                [],
+                'Shop.Forms.Errors'
+            ) . PHP_EOL .
+            $this->translator->trans(
+                'The following characters are not accepted: / \ ^ * `',
```
LouiseBonnard: I think we should tend towards shorter wordings, otherwise error notifications will be too present and discouraging. What about _Invalid characters: /\^*`_ instead?

classes/ValidateConstraintTranslator.php
```diff
@@ -49,14 +49,32 @@ public function __construct(TranslatorInterface $translator)
...
+                [],
+                'Shop.Forms.Errors'
+            ) . PHP_EOL .
+            $this->translator->trans(
+                'The characters . and 。 are allowed only if they are directly followed by a space or the end of the string.',
```
LouiseBonnard: I wonder about the "。" character... it seems pretty specific, is it an absolute necessity to mention it?

classes/ValidateConstraintTranslator.php
jolelievre: Seems good to me, @samuel-pires wdyt?

classes/ValidateConstraintTranslator.php
jolelievre: Well this character is indeed forbidden, but maybe you're right If someone tries to use this character it is probably in order to use the breach so maybe it's not useful to indicate it for usual users

marionf:
@jolelievre some returns, seen with @samuel-pires and @TristanLDD 

1. Delete the message under the fields

capture d'écran_1664

2. There are some other prohibited characters that are not accepted and not displayed in the message like: **0-9!<>,;?=+()@#"°{}_$%:** we should add them

3. Replace "The following characters are not accepted: " by "Invalid characters: "

4. Replace "The characters . and 。 are allowed only if they are directly followed by a space or the end of the string." by "A space is needed after "." and "。""

LouiseBonnard:
@marionf, about the wording, you can also write _A space is required after "." and "。"_ to give a sense of obligation.

jolelievre:
@marionf @LouiseBonnard @samuel-pires @TristanLDD 
All your modifications have been addressed, this should be good unless you find other issues?

classes/ValidateConstraintTranslator.php
```diff
@@ -49,14 +49,32 @@ public function __construct(TranslatorInterface $translator)
...
+                [],
+                'Shop.Forms.Errors'
+            ) . PHP_EOL .
+            $this->translator->trans(
+                'Invalid characters: 0-9!<>,;?=+()@#"°{}_$%/\^*`',
+                [],
+                'Shop.Forms.Errors'
+            ) . PHP_EOL .
+            $this->translator->trans(
+                'A space is required after . and 。',
```
LouiseBonnard, line 72: Shouldn't it be written with quotation marks? Example: _A space is required after "." and "。"_

LouiseBonnard: I don't know what is the best practice.

ghost: Tristan was for the ""

jolelievre: So you mean using `'A space is required after "." and "。"'`
(Isn't the space after 。 weird in this case?)

jolelievre: NB: the space can't be removed, it's part of the character

ghost: @matks told us it is the equivalent of "." in China and that you can use it as a dot in a URL

ghost: But yeah you are right the space is part of it so I don't really understand how it can be used for a URL 🤔

jolelievre: I think maybe some email browsers simplify them and convert them into regular `.` thus creating a clickable url
But my question was is it ok if we have this little space inside the `""` ?

LouiseBonnard: @jolelievre, as for me, it is!

jolelievre: Great, then @PierreRambaud can you re-approve this PR please? ^^

matks (approved):

classes/ValidateConstraintTranslator.php
matks: > I think maybe some email browsers simplify them and convert them into regular `.` thus creating a clickable url

I confirm, try visiting https://www.apple。com 😄

LouiseBonnard: But... this PR has been merged without the fix on the wording?

matks: > Great, then @PierreRambaud can you re-approve this PR please? ^^

@LouiseBonnard we've been tricked !

jolelievre: Oups my bad, I was too quick and thought I had already added it
Here's a PR to fix this ^^
https://github.com/PrestaShop/PrestaShop/pull/14163

PR #14147, opened 2019-06-10
Optimize PrestaShop’s image upload function

MathiasReker:
| Questions     | Answers
| Branch?       | develop
| Description?  | https://www.sunnytoo.com/42096/an-important-fix-to-optimize-prestashops-image-upload-function
| Type?         | bug fix
| Category?     | CO
| Fixed ticket? | https://www.sunnytoo.com/42096/an-important-fix-to-optimize-prestashops-image-upload-function
| How to test?  | /

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14147)

### DIFF
--- a/classes/ImageManager.php
+++ b/classes/ImageManager.php
@@ -177,6 +177,9 @@ class ImageManagerCore
             return !($error = self::ERROR_FILE_NOT_EXIST);
         }
 
+        if (preg_match('/\/img\/(p|c|m|s|st)\/(\d+\/)*\d+\.(jpg|jpeg|png|gif)$/Us', $destinationFile)) {
+            return @copy($sourceFile, $destinationFile);
+        }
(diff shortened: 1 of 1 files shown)


matks:
Hi @MathiasReker I dont understand the meaning of this PR.
The article https://www.sunnytoo.com/42096/an-important-fix-to-optimize-prestashops-image-upload-function raises an interesting issue but I do not agree with the fix they suggest ^^ this hijacks the `resize()` function behavior.

The `resize()` function should remain unchanged as it works as expected. Instead, if I understand correctly the article goal, we should rather *not call resize()* in the right places to avoid image quality loss. The fix should be removing some bad calls to `resize()` instead of preventing `resize()` to do its job.

MathiasReker:
@matks I get your point. I will close the PR for now :-)

PR #14156, opened 2019-06-11
Add missing null check in EditCmsPageHandler

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | `EditCmsPageHandler` was missing one if statement, which checks whether shop association was set. This PR adds that missing statement. Without it, calling `EditCmsPageCommand` with `shopAssociation null` value would cause an error.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? |
| How to test?  | It is not reproducible in prestashop web enviroment, because form always fills command with shop association values. After this PR `Improve > Design > Pages > edit page` should work as before.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14156)


Quetzacoalt91 (approved):

Quetzacoalt91:
Looks like we do not need of QA review on this one, thank you @zuk3975

Issue #13490, opened 2019-04-19
Issues when creating new module for PS 1.7.6


tomas862:
### Summary
Needs to be done for module release and 1.7.6.0 release
- [x] How to handle errors happening inside module (translation + display to user) https://github.com/PrestaShop/PrestaShop/issues/13490 ✅ fixed by https://github.com/PrestaShop/PrestaShop/pull/14239
- [x] Translation is not working in a module https://github.com/PrestaShop/PrestaShop/issues/13942 ✅ fixed in the module
- [x] https://github.com/PrestaShop/PrestaShop/issues/12585  ✅ fixed by https://github.com/PrestaShop/PrestaShop/pull/13875
- [x] https://github.com/PrestaShop/PrestaShop/issues/13944  🚧 ✅ fixed by https://github.com/PrestaShop/PrestaShop/pull/14008

#### Can be done after module release (as an improvement)
- [x] How to add a JS extension for your modern module https://github.com/PrestaShop/PrestaShop/issues/13943

**Description**
I recently created a sample module which uses CQRS and new hooks in its own system and I found several things which a worth to be investigated and fixed:

**Link to the module**

https://github.com/friends-of-prestashop/demo-cqrs-hooks-usage-module

**Problems**

1. Can't make the translations to work (transferred to https://github.com/PrestaShop/PrestaShop/issues/13942)

- according to this https://github.com/PrestaShop/PrestaShop/blob/develop/tests/Resources/modules/translations/src/Controller/Admin/ExempleController.php#L36
- I should be able to have working transaltions from my module controller but they just don't work here https://github.com/friends-of-prestashop/demo-cqrs-hooks-usage-module/blob/master/src/Controller/Admin/CustomerReviewController.php#L54 . But it works in **Translations** module

- for both **Translations** module and my sample module the translations does not work in main module file as well https://github.com/friends-of-prestashop/demo-cqrs-hooks-usage-module/blob/master/ps_democqrshooksusage.php#L68

2. Using JS extension with module (transferred to https://github.com/PrestaShop/PrestaShop/issues/13943)

I have added `ToggleColumn` in Customers list using the module and it works like a charm. `ToggleColumn` required javascript extension  added to the grid `new ColumnTogglingExtension()`. To my luck, Customers list has this extension but developers might face the issues when they add column but the list does not have this extension in javascript compiled.

Solution:
[...]

PR #14161, opened 2019-06-12
Catching exceptions from module

tomas862:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | in module exception is thrown and in core exception is catched and handled. See :arrow_double_down:  for sample usage within module
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/13490
| How to test?  | no testing data provided yet.

in module:

```php
throw \PrestaShop\PrestaShop\Core\Exception\ModuleException::buildFromArray(
            [
                $this->getTranslator()->trans('First error from module', [], 'Modules.Democqrshooksusage.Admin'),
                $this->getTranslator()->trans('Second error from module', [], 'Modules.Democqrshooksusage.Admin'),
            ]
        );
```

in core controller:

```php
catch (ModuleException $e) {
            $this->setModuleErrorMessages($e);
        }
```

Result:

Screenshot (2)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14161)

### DIFF
--- a/src/Core/Exception/ModuleException.php
+++ b/src/Core/Exception/ModuleException.php
@@ -27,7 +27,6 @@
 namespace PrestaShop\PrestaShop\Core\Exception;
 
 use Exception;
-use Throwable;
 
 /**
  * Exception that is thrown in modules context.
@@ -42,9 +41,9 @@ class ModuleException extends Exception
     /**
      * @param string $message
      * @param int $code
-     * @param Throwable|null $previous
+     * @param Exception|null $previous
      */
-    public function __construct($message = "", $code = 0, Throwable $previous = null)
+    public function __construct($message = "", $code = 0, $previous = null)
     {
         parent::__construct($message, $code, $previous);
 
--- a/src/Core/Exception/ModuleException.php
+++ b/src/Core/Exception/ModuleException.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to http://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Exception;
+
+use Exception;
+
+class ModuleException extends Exception
+{
+    /**
+     * @var string[]
+     */
+    private $errors;
+
+    public function __construct(array $errors, $message = "", $code = 0, $previous = null)
+    {
+        $this->setErrors($errors);
+        parent::__construct($message, $code, $previous);
+    }
+
+    /**
+     * @param string[] $errors
+     *
+     * @return ModuleException
+     */
+    public static function buildFromArray(array $errors)
+    {
+        return new self($errors);
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getErrors()
+    {
+        return $this->errors;
+    }
+
+    /**
+     * @param string[] $errors
+     */
+    private function setErrors(array $errors)
+    {
+        $this->errors = $errors;
+    }
+}
--- a/src/Core/Exception/ModuleException.php
+++ b/src/Core/Exception/ModuleException.php
@@ -27,43 +27,52 @@
 namespace PrestaShop\PrestaShop\Core\Exception;
 
 use Exception;
+use Throwable;
 
+/**
+ * Exception that is thrown in modules context.
+ */
 class ModuleException extends Exception
 {
     /**
      * @var string[]
      */
-    private $errors;
+    private $messages = [];
 
-    public function __construct(array $errors, $message = "", $code = 0, $previous = null)
+    /**
+     * @param string $message
+     * @param int $code
+     * @param Throwable|null $previous
+     */
+    public function __construct($message = "", $code = 0, Throwable $previous = null)
     {
-        $this->setErrors($errors);
         parent::__construct($message, $code, $previous);
+
+        if (!empty($message)) {
+            $this->messages[] = $message;
+        }
     }
 
     /**
-     * @param string[] $errors
+     * Build exception instance with an array of messages.
      *
-     * @return ModuleException
+     * @param string[] $messages
+     *
+     * @return self
      */
-    public static function buildFromArray(array $errors)
+    public static function buildWithMessages(array $messages)
     {
-        return new self($errors);
+        $instance = new self();
+        $instance->messages = $messages;
+
+        return $instance;
     }
 
     /**
      * @return string[]
      */
-    public function getErrors()
+    public function getMessages()
     {
-        return $this->errors;
-    }
-
-    /**
-     * @param string[] $errors
-     */
-    private function setErrors(array $errors)
-    {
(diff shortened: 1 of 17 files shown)


src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
```diff
@@ -482,4 +483,31 @@ protected function getErrorMessageForException(Exception $e, array $messages)
...
+    {
+        $moduleErrors = [];
+
+        foreach ($e->getErrors() as $error) {
+            if (is_string($error)) {
+                $moduleErrors[] = $error;
+            }
+        }
+
+        if (empty($moduleErrors)) {
```
matks:
```suggestion
        if (!empty($moduleErrors)) {
```

src/Core/Exception/ModuleException.php
```diff
@@ -0,0 +1,69 @@
...
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Exception;
+
+use Exception;
+
+class ModuleException extends Exception
```
sarjon, line 34: `ModuleException` seems like a very generic exception, how about `ModuleHookException`? @matks wdyt?

src/Core/Exception/ModuleException.php
```diff
@@ -0,0 +1,69 @@
...
+{
+    /**
+     * @var string[]
+     */
+    private $errors;
+
+    public function __construct(array $errors, $message = "", $code = 0, $previous = null)
```
sarjon: I'd name it `translatedErrors` to make it more clear.

src/Core/Exception/ModuleException.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        $this->setErrors($errors);
```
sarjon: Are you sure setter is needed here?

src/Core/Exception/ModuleException.php
```diff
@@ -0,0 +1,69 @@
...
+        parent::__construct($message, $code, $previous);
+    }
+
+    /**
+     * @param string[] $errors
+     *
+     * @return ModuleException
+     */
+    public static function buildFromArray(array $errors)
```
sarjon: Maybe `buildWithMessages` or just `withMessages`? Wdyt?

src/Core/Exception/ModuleException.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        return new self($errors);
```
sarjon: Although I'm not sure this static factory is useful, is it?

tomas862: For me it also seems quite a strange way to do it. @matks, wdyt? You have suggested this approach I think the reason was to make module developers a simplified version of how to raise an exception?

src/Core/Exception/ModuleException.php
matks: But we could use it outside of a hook 🤔if we use SF services override for example

src/Core/Exception/ModuleException.php
matks: Because I wanted to allow module developers to build a `ModuleException` with either 1 or multiple error messages for maximum flexibility (with a constructor `public function __construct($error, ...)`

I expect most module developers to throw one exception for each error which means
```
throw ModuleException($error);
```
so `buildFromArray()` was a helper function to be able to deal with an array of errors

src/Core/Exception/ModuleException.php
sarjon: Well, `ModuleException` could be the base exception for other module related features.

src/Core/Exception/ModuleException.php
matks: Also we might consider adding a `addError()` function for developers which want to build the ModuleException gradually and throw it at the end 🤔 ? or is it too much ?

Quetzacoalt91:
Hi @tomas862,

We're close to release PrestaShop 1.7.6.0, only regressions from 1.7.5.x will be merged on this branch. I advise you to switch on the `develop` branch. :)

matks:
> Hi @tomas862,
> 
> We're close to release PrestaShop 1.7.6.0, only regressions from 1.7.5.x will be merged on this branch. I advise you to switch on the `develop` branch. :)

It's a PR to allow modules on new migrated pages to handle errors properly 😅is it a big issue ?

matks:
Also the PR aims to fix https://github.com/PrestaShop/PrestaShop/issues/13490 which is on 1.7.6 roadmap 😉

kpodemski:
Oh come on... this needs to be merged into 1.7.6...

src/Core/Exception/ModuleException.php
```diff
@@ -0,0 +1,78 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Exception;
+
+use Exception;
+use Throwable;
```
sarjon: `Throwable` is part of PHP 7, thus should not be used in 1.7.6 PS.

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/SupplierController.php
```diff
@@ -37,7 +37,6 @@
 use PrestaShop\PrestaShop\Core\Domain\Supplier\Exception\SupplierConstraintException;
 use PrestaShop\PrestaShop\Core\Domain\Supplier\Exception\SupplierException;
 use PrestaShop\PrestaShop\Core\Domain\Supplier\Exception\SupplierNotFoundException;
-use PrestaShop\PrestaShop\Core\Domain\Supplier\ValueObject\SupplierId;
```
rokaszygmantas, line 40: namespace was unused

rokaszygmantas:
Updated this PR - cleaned up `ModuleException` class, and now migrated controllers are catching it in every relevant create/update action. Also introduced `ModuleHookException`, which extends `ModuleException`, which will allow us to control exception scope in the future.

Created a small module to test it - [link to module](https://github.com/rokaszygmantas/hookerrortest)
1. Install this module
2. Go to edit a customer, save the form. You should see multiple error messages: https://prnt.sc/o1crqq
3. Go to add new customer, fill mandatory fields, save the form. You should see one error message that's coming from the module: https://prnt.sc/o1ct8z

ping @matks

matks (dismissed):

src/Core/Exception/ModuleException.php
```diff
@@ -0,0 +1,77 @@
...
+     * @var string[]
+     */
+    private $messages = [];
+
+    /**
+     * @param string $message
+     * @param int $code
+     * @param Exception|null $previous
+     */
+    public function __construct($message = "", $code = 0, $previous = null)
```
PierreRambaud:
```suggestion
    public function __construct($message = '', $code = 0, $previous = null)
```

src/Core/Exception/ModuleException.php
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Build exception instance with an array of messages.
+     *
+     * @param string[] $messages
+     *
+     * @return self
+     */
+    public static function buildWithMessages(array $messages)
```
PierreRambaud, line 62: Never used

rokaszygmantas: it is meant to be used in modules :)

matks:
> Oh come on... this needs to be merged into 1.7.6...

Hi @kpodemski, I can understand your frustration but rest assured that we will not ship 1.7.6 with a proper error system for modules !

However after discussing this, @eternoendless and I agree that this PR is a new feature and consequently cannot be merged into 1.7.6.x as we are in feature freeze and really, really close to release date. The risk is way too high.

That is why we are trying to find an eligible workaround. https://github.com/PrestaShop/PrestaShop/pull/14206 should do the job (now waiting for @eternoendless review): it focuses on fixing the issue (modules error unable to be displayed correctly) and is small enough not to introduce a risk for the incoming 1.7.6.0 release.

Once the PR workaround has been merged, we'll sit back and see what we do about this PR. Because the workaround might replace it (fully or partially) or require us to update it. When this is done, we'll merge this PR in `develop` for next version 1.7.7

matks:
This matter was finally handled in https://github.com/PrestaShop/PrestaShop/pull/14239

PR #14168, opened 2019-06-12
Add missing check for CMS page category existence on create/edit actions

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | There was missing `check if cms category exists by provided id` when creating or editing the cms page. Without this check it was possible to create cms page for non existing cms category. ( To reproduce this you need to change cms category checkbox value manually within webtools, or execute the command from command line, because the form in front-end doesn't allow to select wrong value).
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Go to Improve > Design > Pages > Create new page. Inspect category checkbox and change its value to non existing cms category id (e.g. 350). Fill other required fields and save the form. You should get an error "Object cannot be loaded (or found)" and cms page will not be created in database `ps_cms` table.. Before this PR you would be able to create cms page for non-existing category and this would provoke further issues.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14168)

### DIFF
--- a/src/Adapter/CMS/Page/CommandHandler/AddCmsPageHandler.php
+++ b/src/Adapter/CMS/Page/CommandHandler/AddCmsPageHandler.php
@@ -32,7 +32,6 @@ use PrestaShop\PrestaShop\Core\Domain\CmsPage\CommandHandler\AddCmsPageHandlerIn
 use PrestaShop\PrestaShop\Core\Domain\CmsPage\Exception\CannotAddCmsPageException;
 use PrestaShop\PrestaShop\Core\Domain\CmsPage\Exception\CmsPageException;
 use PrestaShop\PrestaShop\Core\Domain\CmsPage\ValueObject\CmsPageId;
-use PrestaShop\PrestaShop\Core\Domain\CmsPageCategory\Exception\CmsPageCategoryException;
 use PrestaShopException;
 
 /**
@@ -42,9 +41,6 @@ final class AddCmsPageHandler extends AbstractCmsPageHandler implements AddCmsPa
 {
     /**
      * {@inheritdoc}
-     *
-     * @throws CmsPageException
-     * @throws PrestaShopException
      */
     public function handle(AddCmsPageCommand $command)
     {
@@ -76,9 +72,6 @@ final class AddCmsPageHandler extends AbstractCmsPageHandler implements AddCmsPa
      * @param AddCmsPageCommand $command
      *
      * @return CMS
-     *
-     * @throws PrestaShopException
-     * @throws CmsPageCategoryException
      */
     protected function createCmsFromCommand(AddCmsPageCommand $command)
     {
--- a/src/Adapter/CMS/Page/CommandHandler/AddCmsPageHandler.php
+++ b/src/Adapter/CMS/Page/CommandHandler/AddCmsPageHandler.php
@@ -32,6 +32,7 @@ use PrestaShop\PrestaShop\Core\Domain\CmsPage\CommandHandler\AddCmsPageHandlerIn
 use PrestaShop\PrestaShop\Core\Domain\CmsPage\Exception\CannotAddCmsPageException;
 use PrestaShop\PrestaShop\Core\Domain\CmsPage\Exception\CmsPageException;
 use PrestaShop\PrestaShop\Core\Domain\CmsPage\ValueObject\CmsPageId;
+use PrestaShop\PrestaShop\Core\Domain\CmsPageCategory\Exception\CmsPageCategoryException;
 use PrestaShopException;
 
 /**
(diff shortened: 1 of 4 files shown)


zuk3975:
@matks

PierreRambaud:
@zuk3975 Instead of writing a comment to ping @matks. You can simply set him as reviewer =) (or trying to Assignee it to him but don't know if you're able to do that)

src/Adapter/CMS/Page/CommandHandler/AddCmsPageHandler.php
```diff
@@ -77,13 +78,17 @@ public function handle(AddCmsPageCommand $command)
      * @return CMS
      *
      * @throws PrestaShopException
+     * @throws CmsPageCategoryException
```
PierreRambaud: In fact, this function doesn't throws exception, otherwise if you want to add all exceptions, you must add all database exception, Exception and others.

src/PrestaShopBundle/Controller/Admin/Improve/Design/CmsPageController.php
```diff
@@ -256,6 +256,11 @@ public function editAction(Request $request, $cmsPageId)
                     'cmsPageId' => $cmsPageId,
                 ]),
             ]);
+
+            /** @var EditableCmsPage $editableCmsPage */
+            $editableCmsPage = $this->getQueryBus()->handle(new GetCmsPageForEditing($cmsPageId));
+            $previewUrl = $editableCmsPage->getPreviewUrl();
```
matks: If you move this statement, then we're going to present the `$previewUrl` *before* the update. Looks like an issue 🤔. If the merchant modifies the CMS Page and what should be previewed, after update he will still see the outdated preview.

zuk3975: :thinking: I doubt it's an issue. In form you cannot change anything that would impact the `$previewUrl` itself. Url contains id_cms and id_lang (those aren't configurable in edit action). In addition - if `$previewUrl` is defined after handler it might cause troubles in case handler throws an exception (the redirect page cannot be rendered as the `$previewUrl` is not yet defined). Does it make sense, or am I missing something?

matks: > 🤔 I doubt it's an issue. In form you cannot change anything that would impact the `$previewUrl` itself. Url contains id_cms and id_lang (those aren't configurable in edit action).

That's a good point 👍 

However
>  In addition - if `$previewUrl` is defined after handler it might cause troubles in case handler throws an exception (the redirect page cannot be rendered as the `$previewUrl` is not yet defined). Does it make sense, or am I missing something?

You have raised an important issue 😉 if `$previewUrl` is used to render the page, then it cannot be in the `try/catch` block as there is a possibility we get out of the `try` without assigning the variable.

We have 2 solutions to handle it:
- how is `$previewUrl` used in twig ? can we pass a default value like an empty string ?
- if not, then we must check, just before the render, `if ($previewUrl === null)` and handle the usecase*

*By handle the usecase, I mean:
- if the previewUrl is a critical item of the page, we throw an exception
- if the previewUrl is a mandatory item of the page, we display a red error message in twig 
- if the previewUrl is n optional item of the page, we display a orange warning message in twig

zuk3975: Isn't it enough to handle it here: https://github.com/PrestaShop/PrestaShop/blob/03aa43cfb7745f1c6e2a8df86e666e4cdb83e499/src/PrestaShopBundle/Controller/Admin/Improve/Design/CmsPageController.php#L288 ?
PreviewUrl is a property of EditableCmsPage, so it could be missing only if EditableCmsPage is not found. In that case it's redirecting to index with "The object cannot be loaded (or found)" error

matks: No it could be missing if anything else goes wrong in `$this->getQueryBus()->handle(new GetCmsPageForEditing($cmsPageId));` 😄

zuk3975: https://github.com/PrestaShop/PrestaShop/pull/14168/commits/a2759bfbf639cbf1d1621f0933b1330910d76355
wdyt ? Ensures the $form and $editableCmsPage are defined if reaches return (code looks terrible with 2 tries though). Same case in CategoryController edit action.

matks (dismissed):

matks:
@zuk3975 Can you rebase before QA validation ?

sarjon:
Still needs rebase.

zuk3975:
> Still needs rebase.

:heavy_check_mark:

matks (approved):

matks:
Relaunching Travis, if it's green we go for QA ✅

PierreRambaud:
Thanks @zuk3975

PR #14169, opened 2019-06-12
Bug with select multilang objects

fest73:
When there is no data in the multilanguage entity there is no data in the multilanguage table

| Questions     | Answers
| Branch?       | develop
| Description?  | When there is no data in the multilanguage entity there is no data in the multilanguage table
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | When there is no data in the multilanguage entity there is no data in the multilanguage table. When there is no data in the multilanguage object, there is no data in the multilanguage table. By default, the query selects the "primary" value from the multilanguage table, since there it is empty, it will return null and there will be a bug

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14169)

### DIFF
--- a/classes/PrestaShopCollection.php
+++ b/classes/PrestaShopCollection.php
@@ -336,7 +336,7 @@ class PrestaShopCollectionCore implements Iterator, ArrayAccess, Countable
         $this->is_hydrated = true;
 
         $alias = $this->generateAlias();
-        $this->query->select('*, ' . $alias . '.`' . $this->definition['primary'] . '`');
+        $this->query->select($alias . '*');
         $this->query->from($this->definition['table'], $alias);
 
         // If multilang, create association to lang table
--- a/classes/PrestaShopCollection.php
+++ b/classes/PrestaShopCollection.php
@@ -336,7 +336,7 @@ class PrestaShopCollectionCore implements Iterator, ArrayAccess, Countable
         $this->is_hydrated = true;
 
         $alias = $this->generateAlias();
-        $this->query->select($alias . '*');
+        $this->query->select($alias . '*, ' . $alias . '.`' . $this->definition['primary'] . '`');
         $this->query->from($this->definition['table'], $alias);
 
         // If multilang, create association to lang table
--- a/classes/PrestaShopCollection.php
+++ b/classes/PrestaShopCollection.php
@@ -336,7 +336,7 @@ class PrestaShopCollectionCore implements Iterator, ArrayAccess, Countable
         $this->is_hydrated = true;
 
         $alias = $this->generateAlias();
-        $this->query->select('*, a0.`' . $this->definition['primary'] . '`');
+        $this->query->select('*, ' . $alias . '.`' . $this->definition['primary'] . '`');
         $this->query->from($this->definition['table'], $alias);
 
         // If multilang, create association to lang table
--- a/classes/PrestaShopCollection.php
+++ b/classes/PrestaShopCollection.php
@@ -336,7 +336,7 @@ class PrestaShopCollectionCore implements Iterator, ArrayAccess, Countable
         $this->is_hydrated = true;
 
         $alias = $this->generateAlias();
-        $this->query->select('*, a0.`'.$this->definition['primary'].'`');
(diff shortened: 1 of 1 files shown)


fest73:
@Quetzacoalt91 please, check my commit)

classes/PrestaShopCollection.php
```diff
@@ -336,7 +336,7 @@ public function getAll($display_query = false)
         $this->is_hydrated = true;
 
         $alias = $this->generateAlias();
-        //$this->query->select($alias.'.*');
+        $this->query->select('*, a0.`' . $this->definition['primary'] . '`');
```
PierreRambaud: Can you explain it. Why a0? where it comes from?

fest73: PrestashopCollection class
  protected function generateAlias($association = '')
    {
        if (!isset($this->alias[$association])) {
            $this->alias[$association] = 'a'.$this->alias_iterator++;
        }
        return $this->alias[$association];
    }
When you call a multilingual collection, the first table gets alias a0

fest73: Rather, any entity

fest73: @PierreRambaud You can check? I read that this is a critical error

PierreRambaud: This is why you must use `$alias` instead of `a0` :thinking:

fest73: @PierreRambaud "a0" the first entity table always comes, otherwise you won’t do it, and without this addition, the collection to the prestashop becomes useless

fest73: @PierreRambaud When you receive a collection of multilanguage entities, a choice is made of the type "a0. *, a1. *" If there is no data in the multilanguage, then "a1.id_table = null" overwrites "a0.id_table" and the collection is formed incorrectly

fest73: @PierreRambaud See code please, where formed collection from data from database

PierreRambaud: Can you give us more information about how to reproduce.
What do you mean by multilanguage entity and multilanguage table? 

As you can see the code if you have
```php
$alias = $this->generateAlias();
```
`$alias` => `a0`.

Otherwise you need to retrieve the real alias, not writing something like something which can be dynamic.
What if in the future we change the generateAlias and use `b.iterator` instead `a.iterator`. We will have a bug because the alias getter wasn't used :/

fest73: @pierreavizou I updated file PrestaShopCollection.php, replaced a0 on $alias

classes/PrestaShopCollection.php
```diff
@@ -336,7 +336,7 @@ public function getAll($display_query = false)
         $this->is_hydrated = true;
 
         $alias = $this->generateAlias();
-        //$this->query->select($alias.'.*');
+        $this->query->select('*, ' . $alias . '.`' . $this->definition['primary'] . '`');
```
PierreRambaud:
```suggestion
        $this->query->select($alias . '*');
```

Is enough, since `$this->definition['primary']` is already in `*` :thinking:

classes/PrestaShopCollection.php
fest73: @PierreRambaud Aaaa, noooo. You lose other fields, from table_lang

fest73: @PierreRambaud  In ObjectModel::hydrateCollection an object is going from fields, please look, you suggested a bad option, my worker, I already added in my project it works very well

PierreRambaud: You miss the `.` in your commit: https://github.com/PrestaShop/PrestaShop/pull/14169/commits/addfd67e8bb7a1f60f6cca2b7d4dca7f1c86561b#diff-9d88e764dbaf9a1eabd6fd2246342223R339

It should be `$this->query->select($alias . '.*');`

PierreRambaud:
Hi @dariusakafest ,

Thanks!

PR #14189, opened 2019-06-13
Fix manufacturer controller using wrong DomainException ns

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | By mistake manufacturer controller used Domain exception instead of ../Core/Domain/DomainException. This has serious impact, because edit/create forms wouldn't catch any exception.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Sell > Catalog > Brands & Suppliers brands create and edit actions working properly.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14189)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
@@ -26,7 +26,7 @@
 
 namespace PrestaShopBundle\Controller\Admin\Sell\Catalog;
 
-use DomainException;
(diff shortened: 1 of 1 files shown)


PierreRambaud (approved):

matks:
@zuk3975 Woah o_O it's scary we could not detect this.
We need to find a way / tool to make sure this does not happen.
Created issue https://github.com/PrestaShop/PrestaShop/issues/14194 to solve this matter

PierreRambaud:
Thanks @zuk3975

PR #14190, opened 2019-06-13
You need to be authenticated to play with addresses

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | You can access to this page http://fo.demo.prestashop.com/fr/adresse with is logically forbidden
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Try to create an address as guest, create command, do everything around addresses

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14190)

### DIFF
--- a/controllers/front/AddressController.php
+++ b/controllers/front/AddressController.php
@@ -25,7 +25,7 @@
  */
 class AddressControllerCore extends FrontController
 {
-    public $auth = false;
+    public $auth = true;
     public $guestAllowed = true;
     public $php_self = 'address';
     public $authRedirection = 'addresses';


matks (approved):

jolelievre (approved):

Quetzacoalt91 (approved):
I smelled an issue during the checkout process while managing an address, but it looks like everything's fine.

PierreRambaud:
@Quetzacoalt91 Same to me, I trie on checkout, with or without guest account. I still don't know why they keep the `$auth = false;` during many years ^^

matks:
Maybe it used to be an issue in previous 1.6 checkout process but it changed and this is a leftover

Issue #14154, opened 2019-06-11
[SF Migration] Devtool: dynamic CQRS mapping


matks:
Following @zuk3975 implementation of SF Command `prestashop:list:commands-and-queries` I think a nice dev tool for developers would be a BO dev-only page that displays an helpful mapping of CQRS.

By this I mean:
- a map that show which handlers catch which Query or Command
- with searching capabilities
- extra: links to the code (how can de wo that outside of github / an IDEA ? 🤔 )

Suggested idea of @PierreRambaud

matks:
@zuk3975 I believe you implemented `a map that show which handlers catch which Query or Command` ?

zuk3975:
> @zuk3975 I believe you implemented `a map that show which handlers catch which Query or Command` ?

this one just lists the commands with description, but doesn't show the handler:
https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Command/ListCommandsAndQueriesCommand.php
Also there was one to print more detailed info into `.md` file for docs (it includes handlers and more), but it was moved to separate bundle. You can follow this:
https://github.com/PrestaShop/PrestaShop/pull/21876

Hlavtox:
@PrestaShop/prestashop-core-developers Still relevant? Can we close?

boo-code:
On whether this can be closed: the mapping this asks for exists on `develop`, from two directions.

**The full static map.** The command bus is Symfony Messenger now (`MessengerCommandBus`), so
`bin/console debug:messenger` prints every dispatchable message with its handler:

    PrestaShop\PrestaShop\Core\Domain\CustomerService\Command\DeleteCustomerThreadCommand
        handled by prestashop.core.domain.customer_service.command_handler.delete_customer_thread_handler (when method=handle)

Measured: 513 `PrestaShop\...\Core\Domain` messages, 374 commands and 139 queries, each with its
handler. That is the same population `prestashop:list:commands-and-queries` reports (also 513), so
coverage is identical and only `debug:messenger` carries the handler.

**The dev-only view in the back office.** #14198 was merged in 2019 and is still in place. Its
profiler panel renders `Command` / `Command Handler` / `Called from` and the same three for queries,
with `file:line` for the dispatch site, and it is wired in `config_dev.yml` so it only exists in the
dev environment. That covers the "links to the code" line of the request, for whatever ran during the
request being profiled.

What is genuinely not covered is narrower than the original request: no single command gives a domain
[...]

PR #14198, opened 2019-06-13
Add Commands/Queries Profiler tab

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | Adds Profiler tab that informs developers about executed Commands and Queries.
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/14154
| How to test?  | Not needed, it's a dev tool. See screenshot below.

### Profiler tab

Screenshot from 2019-06-13 17-38-25

### Profiler page

Screenshot (3)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14198)

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -136,10 +136,3 @@ prestashop:
         api_client:
             ttl: 7200 # 2h
             # verify_ssl: ~ # Bundle CA by default, declaring "addons.api_client.verify_ssl" parameter overrides its usage
-
-tactician:
-  commandbus:
-    default:
-      middleware:
-        - prestashop.core.command_bus.middleware.command_register_middleware
-        - tactician.middleware.command_handler
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -136,3 +136,10 @@ prestashop:
         api_client:
             ttl: 7200 # 2h
             # verify_ssl: ~ # Bundle CA by default, declaring "addons.api_client.verify_ssl" parameter overrides its usage
+
+tactician:
+  commandbus:
+    default:
+      middleware:
+        - prestashop.core.command_bus.middleware.command_register_middleware
+        - tactician.middleware.command_handler
--- a/src/Core/CommandBus/Parser/CommandTypeParser.php
+++ b/src/Core/CommandBus/Parser/CommandTypeParser.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\CommandBus\Parser;
+
+final class CommandTypeParser
+{
+    /**
+     * @param string $commandName Fully-qualified class name of command
+     *
+     * @return string
+     */
+    public function parse($commandName)
+    {
+        if (strpos($commandName, '\Command\\')) {
+            return 'Command';
+        }
+
+        return 'Query';
+    }
+}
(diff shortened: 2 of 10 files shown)


sarjon:
@matks is this something you expected? Do you have any suggestions for improvements?

src/PrestaShopBundle/DataCollector/HookRegistry.php
```diff
@@ -48,6 +48,7 @@ final class HookRegistry
 
     public function __construct()
     {
+        $a = 1;
```
matks: 😄 ?

src/Core/CommandBus/Parser/CommandTypeParser.php
```diff
@@ -0,0 +1,44 @@
...
+namespace PrestaShop\PrestaShop\Core\CommandBus\Parser;
+
+final class CommandTypeParser
+{
+    /**
+     * @param string $commandName Fully-qualified class name of command
+     *
+     * @return string
+     */
+    public function parse($commandName)
```
matks, line 36: In the future, maybe we could find whether this is a query or command from the bus used `prestashop.core.command_bus` or `prestashop.core.query_bus`

matks:
> @matks is this something you expected? Do you have any suggestions for improvements?

I ❤️ it !

@PierreRambaud Does this PR addresses your need about CQRS being hard to navigate ? 😄

src/PrestaShopBundle/DataCollector/HookRegistry.php
sarjon: Yeah.... :smile:

src/Core/CommandBus/Parser/CommandTypeParser.php
sarjon: Maybe, but now we are using the same bus behind the scene for both.

PierreRambaud:
@matks Oh yeah! :heart:

app/config/config.yml
```diff
@@ -136,3 +136,10 @@ prestashop:
         api_client:
             ttl: 7200 # 2h
             # verify_ssl: ~ # Bundle CA by default, declaring "addons.api_client.verify_ssl" parameter overrides its usage
+
+tactician:
```
mickaelandrieu: if this middleware only needs to be enabled in dev mode, how about put this configuration in `config_dev.yml` file instead? This may remove the check about the application environment in the class.

mickaelandrieu:
I'm wondering who is the real user/use case of introducing this?

Don't expect PrestaShop developers to expect what is CQRS and/or to figure out what/why this information is displayed in the debug toolbar...

As we have the related hooks for each command/query, for an integrator this tab has no value (in its current state), but we could keep the application (ie the page "Command/Query) in the web profiler if it helps the core team and the Symfony developers who want to play with PrestaShop using the "CQRS way" ?

Think about it: the debug toolbar is not infinite so we need **to care a lot** about what information should be available for the PrestaShop developers.

For example, I think we should also remove the one about Guzzle as no ones really figure out what it does and no one really cares in the community.

app/config/config.yml
sarjon: Yeah, that would make sense.

sarjon:
> Don't expect PrestaShop developers to expect what is CQRS and/or to figure out what/why this information is displayed in the debug toolbar...

This is a dev tool for PS developers, sooner or later they will have to start using it I think (and from trainings I see a very big interest in CQRS from PS devs). There is already a documentation about CQRS (not so complete, but still). Either way, it should not harm anyone having this tool around.

> Think about it: the debug toolbar is not infinite so we need to care a lot about what information should be available for the PrestaShop developers.

There is already a lot of information in toolbar that is not really related to PS nor needed for PS devs. If we see that this tool is not needed (and you can only know by getting feedback from developers), then we can remove it from debug toolbar without any BC break I think.

> For example, I think we should also remove the one about Guzzle as no ones really figure out what it does and no one really cares in the community.

Agree, as mentioned before, not all information in debug toolbar is now useful and could be removed. Unfortunately, it's not part of this PR to do so.

matks:
> I'm wondering who is the real user/use case of introducing this?
> 
> Don't expect PrestaShop developers to expect what is CQRS and/or to figure out what/why this information is displayed in the debug toolbar...

We'll tell them 😄. This tool addresses a need expressed by @PierreRambaud : "I'm working on a page, I see some commands and queries are dispatched and I find it hard to trace them to the right Handler to find out where the code I need to debug/update is." This is a common issue when using events (tracing the listener from the dispatch) and even when using Symfony services. It can be solved by going through the configuration, but the aim is to make things clearer, faster, easier.

> As we have the related hooks for each command/query

🤔 I dont see what you mean. Are you sure it's there ?

> Think about it: the debug toolbar is not infinite so we need **to care a lot** about what information 
should be available for the PrestaShop developers.

Indeed. But since this is a dev-only tool any mistake we do will not critical. Also I'm thinking that maybe we can work around this matter by allowing dev to toggle the features they want to see in the debug bar.

> For example, I think we should also remove the one about Guzzle as no ones really figure out what it does and no one really cares in the community.

Yep, makes sense.

matks:
@sarjon As suggested by @mickaelandrieu, a nice improvement (maybe for a future PR) would be to add profiler capabilities to this tab. Being able to see the response time of a handler and the consumed memory could be great debug assets 👍

sarjon:
Hmm, tests are failing, could you rerun them?

matks (approved):

matks:
Thank you @sarjon ❤️

PR #14206, opened 2019-06-14
Introduce the usage of Module user-displayable exceptions to handle module errors

matks:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Add a hook that allow to manage the mapping in controllers for error messages
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/13490 last issue: "How to handle errors happening inside module (translation + display to user)"
| How to test?  | See below

# How to test

Attached is a demonstration module.
If you install it on develop, it's going to hook on the Add a Customer form and the Edit a Customer form.
- on Add form, if you save you'll see a Symfony HTTP 500 error "Uncatched error"
- on Edit form, if you save you'll see a Symfony HTTP 500 error "This is a nice display message"

If you use this branch,
- on Add form, if you save you'll see an error message "An unexpected error occurred. [RuntimeException code 0]"
- on Edit form, if you save you'll see an error message "This is a nice display message"

# PR notes

### Background

This PR solves 2 issues:
- How to display error messages from a module hooked on a Add/Edit form process (see https://github.com/PrestaShop/PrestaShop/issues/13490)
- It also handles the usecase "module does throw an uncatched exception". In the Core, we do not catch the generic `\Exception` because we consider our duty to handle all exception paths and make sure we display the right error message for each, however we cannot guarantee Modules will do the same. So we catch `\Exception` in order to avoid the "blank page" behavior.

### Example of usage

In a module that extends the Customer Form, so it's hooked on `hookActionAfterUpdateCustomerFormHandler`

If an error is found in hook `hookActionAfterUpdateCustomerFormHandler`, the module has 2 possibilities:

- throw a ModuleEloquentException that will be displayed to the merchant
`throw new \PrestaShop\PrestaShop\Core\Module\ModuleEloquentException('This error message should be displayed to the merchant')`

- or, if it uses its own Exceptions, it can use hook `actionGetEloquentExceptionsFromModule` to provide to the Core a list of Exception classes whose behavior should be the same as a ModuleEloquentException
```
    public function hookActionAfterUpdateCustomerFormHandler(array $params)
    {
        throw new \MyModule\Exception\BadBehaviorException(
            'This is a beautiful error message'
        );
    }

    public function hookActionGetEloquentExceptionsFromModule(array $params)
    {
        $whitelist = $params['whitelist'];
        $whitelist[] = \MyModule\Exception\BadBehaviorException::class;
        $params['whitelist'] = $whitelist;
    }
```
The Core will take this whitelist and use it to know whether it should return the exception error message directly.

Final result:
Capture d’écran 2019-06-14 à 10 35 19

### To be decided

1) Naming "ModuleEloquentException" might be improved
2) I put ModuleEloquentException in namespace `PrestaShop\PrestaShop\Core\Module`, is that right ?
3) If Module throws an uncatched Exception, it will be displayed like this:
Capture d’écran 2019-06-14 à 10 56 18

Because of current generic fallback error message
```
    protected function getFallbackErrorMessage($type, $code)
    {
        return $this->trans(
            'An unexpected error occurred. [%type% code %code%]',
            'Admin.Notifications.Error',
            [
                '%type%' => $type,
                '%code%' => $code,
            ]
        );
    }
```
⏩ should we update it ?

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14206)

[module_for_pr_hook_error.zip](https://github.com/PrestaShop/PrestaShop/files/3290604/module_for_pr_hook_error.zip)

### DIFF
--- a/src/Core/Module/ModuleEloquentException.php
+++ b/src/Core/Module/ModuleEloquentException.php
@@ -1,0 +1,34 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Module;
+
+/**
+ * Module Exception whose message can be displayed directly to end-user
+ */
+class ModuleEloquentException extends ModuleException
+{
+}
--- a/src/Core/Module/ModuleException.php
+++ b/src/Core/Module/ModuleException.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Module;
+
+use PrestaShop\PrestaShop\Core\Exception\CoreException;
+
+/**
+ * Base class for all module exceptions
+ */
+class ModuleException extends CoreException
+{
+}
--- a/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
+++ b/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
@@ -31,6 +31,7 @@ use Exception;
 use PrestaShop\PrestaShop\Adapter\Shop\Context;
 use PrestaShop\PrestaShop\Core\ConfigurationInterface;
 use PrestaShop\PrestaShop\Core\Grid\GridInterface;
+use PrestaShop\PrestaShop\Core\Module\ModuleEloquentException;
 use PrestaShopBundle\Security\Voter\PageVoter;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
@@ -462,6 +463,24 @@ class FrameworkBundleAdminController extends Controller
      */
     protected function getErrorMessageForException(Exception $e, array $messages)
     {
+        if ($e instanceof ModuleEloquentException) {
+            return $e->getMessage();
+        }
+
+        $whitelist = [];
+        $this->get('prestashop.core.hook.dispatcher')->dispatchWithParameters('actionGetEloquentExceptionsFromModule', [
+            'controller_name' => get_class($this),
+            'exception' => $e,
+            'whitelist' => &$whitelist,
+        ]);
+        if (is_array($whitelist)) {
+            foreach ($whitelist as $whitelistItem) {
+                if ($e instanceof $whitelistItem) {
+                    return $e->getMessage();
+                }
+            }
+        }
+
         $exceptionType = get_class($e);
         $exceptionCode = $e->getCode();
 
--- a/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
+++ b/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
@@ -475,6 +475,9 @@ class FrameworkBundleAdminController extends Controller
         ]);
         if (is_array($whitelist)) {
             foreach ($whitelist as $whitelistItem) {
+                if (!is_string($whitelistItem) && !is_object($whitelistItem)) {
+                    continue;
+                }
                 if ($e instanceof $whitelistItem) {
                     return $e->getMessage();
                 }
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
@@ -26,6 +26,7 @@
 
 namespace PrestaShopBundle\Controller\Admin\Sell\Customer;
 
+use Exception;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\BulkDeleteCustomerCommand;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\BulkDisableCustomerCommand;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\BulkEnableCustomerCommand;
@@ -149,7 +150,7 @@ class CustomerController extends AbstractAdminController
 
                 return $this->redirectToRoute('admin_customers_index');
             }
-        } catch (CustomerException $e) {
+        } catch (Exception $e) {
             $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));
         }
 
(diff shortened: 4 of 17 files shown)


src/Core/Module/ModuleEloquentException.php
```diff
@@ -0,0 +1,34 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Module;
+
+/**
+ * Module Exception whose message can be displayed directly to end-user
+ */
+class ModuleEloquentException extends ModuleException
```
matks, line 32: Naming can be improved

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
```diff
@@ -462,6 +463,24 @@ protected function getContextShopId()
      */
     protected function getErrorMessageForException(Exception $e, array $messages)
     {
+        if ($e instanceof ModuleEloquentException) {
+            return $e->getMessage();
+        }
+
+        $whitelist = [];
+        $this->get('prestashop.core.hook.dispatcher')->dispatchWithParameters('actionGetEloquentExceptionsFromModule', [
```
matks, line 471: Hook name can be improved. Once we have settled it, I'll register in SQL upgrade and hook.xml file.

src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
```diff
@@ -149,7 +150,7 @@ public function createAction(Request $request)
 
                 return $this->redirectToRoute('admin_customers_index');
             }
-        } catch (CustomerException $e) {
+        } catch (Exception $e) {
```
matks, line 153: This has to be done in every Identifiable Object forms

src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
```diff
@@ -191,7 +192,7 @@ public function editAction($customerId, Request $request)
...
```
sarjon: Shouldn't it be changed in every controller to catch `CoreException` instead?

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
sarjon, line 471: I don't think that suffix `FromModule` is necessary as hooks are meant to be used only by modules.

src/Core/Module/ModuleEloquentException.php
```diff
@@ -0,0 +1,34 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Module;
```
sarjon, line 27: Why not putting it inside `Core\Module\Exception`?

src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
sarjon: Should it catch every possible exception or only `CoreException`?

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
matks: Indeed

src/Core/Module/ModuleEloquentException.php
matks: Was not sure at the beginning 😄

src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
matks: It must be all Exceptions. @eternoendless refuses that a Module that does not handle properly exceptions breaks the page.

src/Core/Module/ModuleEloquentException.php
jolelievre: agreed, we'd better keep the exceptions at the same place when possible/logical

src/Core/Module/ModuleException.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Module;
+
+use PrestaShop\PrestaShop\Core\Exception\CoreException;
+
+/**
+ * Base class for all module exceptions
+ */
+class ModuleException extends CoreException
```
jolelievre, line 34: Nice for the `extends CoreException`

src/Core/Module/ModuleEloquentException.php
jolelievre: `ModuleDisplayedException` ? `ModuleAlertException` ?

I still think an interface could be nice for this use case, something like:

```php
interface DisplayedExceptionInterface
{
    /**
     * @return array
     */
    public function getDisplayedMessage();
}
```
or
```php
interface AlertExceptionInterface
{
    /**
     * @return array
     */
    public function getAlertMessages();
}
```
(not sure if a string or an array of string is best)

And it doesn't prevent `ModuleEloquentException` to be created and implement this interface.
I see several advantages to using an interface:
- modules can create their own exception without extending a core one
- we can use this behaviour even for core exception later, so we don't have to implement their management several times

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
jolelievre: in the same spirit as for the exception name: `actionGetDisplayedExceptionsFromModule` `actionGetAlertExceptionsFromModule`

src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
jolelievre: I agree with the fact of protecting the exceptions from modules, however I'm afraid catching `Exception` might be a bit too wide
If another exception which has nothing to do with a module is thrown it will be caught and displayed as an alert, which can be bothersome for dev and debug (like database exception, client exception, routing, Symfony exceptions, ...)

I'm wondering instead, would it be possible to add a layer in the `HookDispatcher` so that it catch exceptions and throw a `ModuleException` wrapping the first one right away. This would allow us to catch the module exceptions independently. Quick draft:

```php
class HookDispatcher {
    public function dispatch($eventName, Event $event = null)
    {
        if ($event === null) {
            $event = new HookEvent();
        }
        if (!$event instanceof HookEvent) {
            throw new \Exception('HookDispatcher must dispatch a HookEvent subclass only. ' . get_class($event) . ' given.');
        }

        try {
             return parent::dispatch($eventName, $event);
        } catch (\Exception $e) {
             throw new ModuleException($e->getMessage(), $e->getCode(), $e);
        }
    }
}
```

src/Core/Module/ModuleEloquentException.php
matks: Actually, now that you say this, I'm thinking: maybe using the interface we can get rid of the hook ?

The hook was there to tell to the core "eyh, please consider these exceptions as EloquentExceptions (or whatever name we choose) although they do not extend it"
But the interface would do it nicely. And it's a standard practice, for example people creating new Loggers use the Monolog interface.

src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
matks: I'm not a fan 😐because:
- you oblige all exceptions coming from a hook to be ModuleException, which might make module developers think "pfff, there's no use to define different exceptions in my module, it's going to be casted to ModuleException by the core anyway"
- this PR is supposed to solve the issue we have at hands in a very simpler manner as it must be a bugfix (because it's feature freeze ❄️) and this would a massive behavioral change (=> maybe in `develop` then)

src/Core/Module/ModuleEloquentException.php
jolelievre: I think it's also easier for developers, just need to add the interface and it's done No hook registering And the core code stays cleaner (no hook to call in between)

src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
jolelievre: The exception is indeed wrapped into a ModuleException but you still have access to the previous exception
But I agree it's not the best either.. I'm just concerned about catching all exceptions Then the developer (both module and core) will think "What's the point of raising an exception it will be caught as alert message anyway 😅"

matks: We could catch it and re-throw it in DEV, but keep it in PROD ?
Also we could log it

But this can come in `develop` 😉

src/Core/Module/ModuleEloquentException.php
eternoendless: What about:
- `ModuleErrorMessageException` (easy to understand)
- `ModuleErrorException` (maybe more semantic alternative)

Not a fan of "Eloquent", sorry 😛 

In most cases, developers will just throw a `ModuleErrorMessageException` (or a child class) and get done with it, no need for the hook in the end.

src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
eternoendless: Any non-identified exception goes in the fallback error message. It should be logged in prod and probably displayed better in dev. That is an improvement that can be done in develop.

In a perfect world we'd only catch RuntimeExceptions and let LogicExceptions bubble up. But that's not the world we live in 😅...

src/Core/Module/ModuleEloquentException.php
jolelievre: I prefer the shorter `ModuleErrorException` personnaly

matks:
Closed in favor of https://github.com/PrestaShop/PrestaShop/pull/14239

PR #14223, opened 2019-06-14
Fix unescaped dash in character group in autoloader regex

mvorisek:
| Questions     | Answers
| Branch?       | develop
| Description?  | PHP needs 2 backslashes to produce one backslash in the string itself. <br/>Regex needs 2 backslashes to produce one backslash in the regex itself.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Try with php7.3 and PHP7.1

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14223)

### DIFF
--- a/classes/PrestaShopAutoload.php
+++ b/classes/PrestaShopAutoload.php
@@ -286,9 +286,10 @@ class PrestaShopAutoload
                 } elseif (substr($file, -4) == '.php') {
                     $content = file_get_contents($rootDir . $path . $file);
 
-                    $namespacePattern = '[\\a-z0-9_]*[\\]';
-                    $pattern = '#\W((abstract\s+)?class|interface)\s+(?P<classname>' . basename($file, '.php') . '(?:Core)?)'
-                                . '(?:\s+extends\s+' . $namespacePattern . '[a-z][a-z0-9_]*)?(?:\s+implements\s+' . $namespacePattern . '[a-z][\\a-z0-9_]*(?:\s*,\s*' . $namespacePattern . '[a-z][\\a-z0-9_]*)*)?\s*\{#i';
+                    $namePattern = '[a-z_\x7f-\xff][a-z0-9_\x7f-\xff]*';
+                    $nameWithNsPattern = '(?:\\\\?(?:' . $namePattern . '\\\\)*' . $namePattern . ')';
+                    $pattern = '~(?<!\w)((abstract\s+)?class|interface)\s+(?P<classname>' . basename($file, '.php') . '(?:Core)?)'
+                                . '(?:\s+extends\s+' . $nameWithNsPattern . ')?(?:\s+implements\s+' . $nameWithNsPattern . '(?:\s*,\s*' . $nameWithNsPattern . ')*)?\s*\{~i';
 
                     //DONT LOAD CLASS WITH NAMESPACE - PSR4 autoloaded from composer
                     $usesNamespace = false;


mvorisek:
Resubmit https://github.com/PrestaShop/PrestaShop/pull/14224#pullrequestreview-250466315 to `develop` branch.

PierreRambaud (approved):

mvorisek:
Can you please merge so I can delete the fork?

PierreRambaud:
Regarding the tests look like your fix break the autoloader :thinking:

PierreRambaud (changes requested):
Need investigation to find why tests are failing.

mvorisek:
Can you confirm that the build is broken by exactly this PR? Try to cherry pick the commit on top of some non-failing branch.

The regexes seems to be working for me, see https://3v4l.org/hBE88

I have also added there an example why the 4 backslashes are important to properly escape 1 backslash.

PierreRambaud:
yes it's because of your PR :/
It breaks the Autoloader

mvorisek:
Oh, I think I found it. The original code seems to me to be actually wrong:
```
$namespacePattern = '[\\a-z0-9_]*[\\]';
```
parses to string `[\a-z0-9_]*[\]`, the first backslash is (currently) honored in the regex as the next char does not need an escape char and it is withing character group (where escape characters are currently handled less strictly).

But the second backslash is is escaping the following `]` bracket. And due to the usage later with additional `]` (which closes the character group) it "works/compiles".

From my point of view the "namespace pattern" is not intended to parse any `]` characters.

Is it the case?

PierreRambaud:
The best way to test if everything is ok is to update your PR :D

mvorisek:
> Oh, I think I found it. The original code seems to me to be actually wrong:

Yeah, the code was wrong and it worked by coincidences.

mvorisek:
The code is final and passing the CI tests. Can it be reviewed and merged?

classes/PrestaShopAutoload.php
```diff
@@ -286,9 +286,10 @@ protected function getClassesFromDir($path, $hostMode = false)
                 } elseif (substr($file, -4) == '.php') {
                     $content = file_get_contents($rootDir . $path . $file);
 
-                    $namespacePattern = '[\\a-z0-9_]*[\\]';
-                    $pattern = '#\W((abstract\s+)?class|interface)\s+(?P<classname>' . basename($file, '.php') . '(?:Core)?)'
-                                . '(?:\s+extends\s+' . $namespacePattern . '[a-z][a-z0-9_]*)?(?:\s+implements\s+' . $namespacePattern . '[a-z][\\a-z0-9_]*(?:\s*,\s*' . $namespacePattern . '[a-z][\\a-z0-9_]*)*)?\s*\{#i';
+                    $namePattern = '[a-z_\x7f-\xff][a-z0-9_\x7f-\xff]*';
```
PierreRambaud, line 289: Why adding \x7f and \xff, I don't find any reason to add it?

classes/PrestaShopAutoload.php
```diff
@@ -286,9 +286,10 @@ protected function getClassesFromDir($path, $hostMode = false)
...
+                    $nameWithNsPattern = '(?:\\\\?(?:' . $namePattern . '\\\\)*' . $namePattern . ')';
```
PierreRambaud, line 290: Shouldn't be this?
```suggestion
                    $nameWithNsPattern = '(?:\\\\)?(?:' . $namePattern . '\\\\)*' . $namePattern;
```

classes/PrestaShopAutoload.php
mvorisek: https://stackoverflow.com/questions/6362241/php-variable-function-class-names-using-special-characters#6362268 To match all valid names no matter if Unicode names are currently used or not.

classes/PrestaShopAutoload.php
mvorisek: No, it shouldn't. `\\\\` is properly escaped `\` char and the `(?:` + `)` outside parenthesis are there for safe use with quantifiers (imagine `ab` later used as `ab?` instead of `(?:ab)?`).

PierreRambaud: `?:` only mean `Non-capturing group`.
And `ab?` is different than `(?:ab)?)`.
But here I understand that `\\\\` is only one character, and I think my suggestion is easier to understand, (especially for people not familiar with regex ^^)

mvorisek: > `?:` only mean `Non-capturing group`.
> But here I understand that `\\\\` is only one character, and I think my suggestion is easier to understand, (especially for people not familiar with regex ^^)

Yes, I know. If you want, wrap the first `\\\\` inside another `(?:` + `)`, but it is not needed, as it is only one character.

> And `ab?` is different than `(?:ab)?)`.

Yes, that is the example of the difference, beaware that the second/last `)` is part of the text :)

PierreRambaud (approved):

classes/PrestaShopAutoload.php
PierreRambaud: > Yes, I know. If you want, wrap the first \\\\ inside another (?: + ), but it is not needed, as it is only one character.

Of course it's not needed, but I think many people are not familiar with regex and if you ask to many people they will answer you it's two `\` :D But indeed, let's merge your PR :+1:

matks:
Thank you @mvorisek

PR #14224, opened 2019-06-14
Bugfix: Fix unescaped dash in character group in regex 

mvorisek:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | PHP needs 2 backslashes to produce one backslash in the string itself. <br/>Regex needs 2 backslashes to produce one backslash in the regex itself.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Try with php7.3

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14224)

### DIFF
--- a/classes/PrestaShopAutoload.php
+++ b/classes/PrestaShopAutoload.php
@@ -286,9 +286,9 @@ class PrestaShopAutoload
                 } elseif (substr($file, -4) == '.php') {
                     $content = file_get_contents($rootDir . $path . $file);
 
-                    $namespacePattern = '[\\a-z0-9_]*[\\]';
+                    $namespacePattern = '[\\\\a-z0-9_]*[\\\\]';
                     $pattern = '#\W((abstract\s+)?class|interface)\s+(?P<classname>' . basename($file, '.php') . '(?:Core)?)'
-                                . '(?:\s+extends\s+' . $namespacePattern . '[a-z][a-z0-9_]*)?(?:\s+implements\s+' . $namespacePattern . '[a-z][\\a-z0-9_]*(?:\s*,\s*' . $namespacePattern . '[a-z][\\a-z0-9_]*)*)?\s*\{#i';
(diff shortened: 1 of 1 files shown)


classes/PrestaShopAutoload.php
```diff
@@ -286,9 +286,9 @@ protected function getClassesFromDir($path, $hostMode = false)
                 } elseif (substr($file, -4) == '.php') {
                     $content = file_get_contents($rootDir . $path . $file);
 
-                    $namespacePattern = '[\\a-z0-9_]*[\\]';
+                    $namespacePattern = '[\\\\a-z0-9_]*[\\\\]';
```
PierreRambaud, line 289: We're in a quote context.
```php
echo '\t';  // => '\t'
echo "\t"; // => '	'
```
Can you give us an example where it fails? :thinking:

mvorisek: PHP 7.3 uses newer POSIX regex engine and regexes like `~[\d-_]~` became invalid. The fixed regex may became invalid someday too, so I fixed it. - as character is safe to use only as regular range operator or when it is placed the most left or the most right in the character group.

mvorisek:
PHP 7.3 uses newer POSIX regex engine and regexes like `~[\d-_]~` became invalid. The fixed regex may became invalid someday too, so I fixed it. - as character is safe to use only as regular range operator or when it is placed the most left or the most right in the character group.

PierreRambaud:
Oh yes, you're right, didn't think about 7.3 :P

PierreRambaud (approved):
I'm in favor of applying this change. But don't know if we can keep the 1.7.6.x target.

mvorisek:
@PierreRambaud 
> I'm in favor of applying this change. But don't know if we can keep the 1.7.6.x target.

Opened a PR to `develop` branch, see https://github.com/PrestaShop/PrestaShop/pull/14223

PR #14225, opened 2019-06-14
Update facetedsearch cache

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Facetedsearch template is not created after installation due to the fact that fixtures are installed after modules.
| Type?         | bug fix
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Facetedsearch must be installed with a template. #14197 must be merged first.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14225)

### DIFF
--- a/install-dev/fixtures/fashion/install.php
+++ b/install-dev/fixtures/fashion/install.php
@@ -42,4 +42,21 @@ class InstallFixturesFashion extends XmlLoader
 
         return $this->createEntity('customer', $identifier, 'Customer', $data, $data_lang);
     }
+
+    /**
+     * @{inheritdoc}
+     */
+    public function populateFromXmlFiles()
+    {
+        parent::populateFromXmlFiles();
+
+        /**
+         * Refresh facetedsearch cache
+         */
+        /** @var Ps_Facetedsearch */
+        $facetedSearch = Module::getInstanceByName('ps_facetedsearch');
+        $facetedSearch->buildLayeredCategories();
+        $facetedSearch->rebuildLayeredCache();
+        $facetedSearch->fullPricesIndexProcess(0, false);
+    }
 }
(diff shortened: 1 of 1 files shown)


install-dev/fixtures/fashion/install.php
```diff
@@ -42,4 +42,23 @@ public function createEntityCustomer($identifier, array $data, array $data_lang)
...
+
+        /**
+         * Refresh facetedsearch cache
+         */
+        /** @var Ps_Facetedsearch */
+        $facetedSearch = Module::getInstanceByName('ps_facetedsearch');
+        if ($facetedSearch) {
+            $facetedSearch->buildLayeredCategories();
+            $facetedSearch->rebuildLayeredCache();
+            $facetedSearch->fullPricesIndexProcess(0, false);
```
Quetzacoalt91: I'm not in favor having a strong dependency between the core and a specific module. If one day one of these method disparears without taking care of the compatibility, all installation may crash.

May I suggest a hook instead?

PierreRambaud: No problem, I'll try something else :)

PierreRambaud: I tried with `to_reset` in https://github.com/PrestaShop/PrestaShop/blob/develop/themes/classic/config/theme.yml#L54 but because there's no data, nothing happened.
And there's no hook in the whole installer. And I think fixtures must be the last step of the installer :( I'm stuck :/

PierreRambaud: Should be better using the module manager :)

Quetzacoalt91 (approved):

PR #14236, opened 2019-06-16
Removes unneeded parentheses around return statement

MathiasReker:
| Questions     | Answers
| Branch?       | develop
| Description?  | Removes unneeded parentheses around return statement
| Type?         | improvement
| Category?     | TE
| Fixed ticket? | /
| How to test?  | /

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14236)

### DIFF
--- a/tests-legacy/resources/modules/ps_banner/ps_banner.php
+++ b/tests-legacy/resources/modules/ps_banner/ps_banner.php
@@ -54,11 +54,11 @@ class Ps_Banner extends Module implements WidgetInterface
 
     public function install()
     {
-        return (parent::install() &&
+        return parent::install() &&
             $this->registerHook('displayHome') &&
             $this->registerHook('actionObjectLanguageAddAfter') &&
             $this->installFixtures() &&
-            $this->disableDevice(Context::DEVICE_MOBILE));
+            $this->disableDevice(Context::DEVICE_MOBILE);
     }
 
     public function hookActionObjectLanguageAddAfter($params)


PierreRambaud:
We'll plan to remove this directoryu because of legacy context. Not sure it's needed to improve this part.
wdyt @matks ?

matks:
> We'll plan to remove this directoryu because of legacy context. Not sure it's needed to improve this part.
> wdyt @matks ?

Can't hurt 😅the removal is going to happen very slowly, so I think it's alright to put some ❤️in this directory

Quetzacoalt91:
On my side, I think modifying this modules which is only here as fixtures is unecessary. This change should be applied on the source repository in https://github.com/PrestaShop/ps_banner, or it will be lost.

By the way, I wonder if these modules should not be included in this repository as submodules, because we duplicated code and we don't really know anymore if this part should be maintained. I'm closing this PR, but this question will remain as long as we keep this code.

matks:
> By the way, I wonder if these modules should not be included in this repository as submodules, because we duplicated code and we don't really know anymore if this part should be maintained. I'm closing this PR, but this question will remain as long as we keep this code.

Maybe we should not rely on real modules in our tests 🤔only fake dummy modules ?

PierreRambaud:
I'm for in favor of dummy modules, with only the necessary, which means, if your test need two hooks, create a little module with two hooks, etc.

PR #14239, opened 2019-06-17
Introduce the usage of Module user-displayable exceptions to handle module errors (part 2)

matks:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Allow hooked modules to throw a ModuleErrorException or ModuleErrorInterface to display error messages to end-user
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/13490 last issue: "How to handle errors happening inside module (translation + display to user)"
| How to test?  | See below

# How to test

Attached is a demonstration module.
[module_for_pr_hook_error.zip](https://github.com/PrestaShop/PrestaShop/files/3300628/module_for_pr_hook_error.zip)

If you install it and test this PR,
- on Add form, if you save you'll see an error message "An unexpected error occurred. [RuntimeException code 0]"
- on Edit form, if you save you'll see an error message "This is a nice display message"

# PR notes

### Background

This PR solves 2 issues:
- How to display error messages from a module hooked on a Add/Edit form process (see https://github.com/PrestaShop/PrestaShop/issues/13490)
- It also handles the usecase "module does throw an uncatched exception". In the Core, we do not catch the generic `\Exception` because we consider our duty to handle all exception paths and make sure we display the right error message for each, however we cannot guarantee Modules will do the same. So we catch `\Exception` in order to avoid the "blank page" behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14239)

### DIFF
--- a/src/Core/Module/Exception/ModuleErrorInterface.php
+++ b/src/Core/Module/Exception/ModuleErrorInterface.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Module\Exception;
+
+/**
+ * If an exception that implements this ModuleErrorInterface
+ * is thrown, its message will be displayed to the end-user
+ */
+interface ModuleErrorInterface
+{
+}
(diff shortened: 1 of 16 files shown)


src/Core/Module/Exception/ModuleErrorInterface.php
```diff
@@ -0,0 +1,35 @@
...
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Module\Exception;
+
+/**
+ * If an exception that implements this ModuleErrorInterface
+ * is thrown, its message will be displayed to the end-user
+ */
+interface ModuleErrorInterface
```
jolelievre, line 33: I'm wondering is it ok to have an empty interface? I don't mind it's just I wasn't sure it was possible

I had thought of a single `getErrorMessage` method which would allow to "customize" the exception message a bit (the simplest implementation being to simply return the exception message). This would allow to separate the exception message and the error message which might be different:
- exception message => technical message with (potentially) sensitive info (db infos, path infos, ...)
- error message => user message more "feature oriented" and safe to display

Besides by using `ModuleErrorInterface` you restrict its use to module only, maybe we could call it `FlashErrorInterface` which would allow us to use it with core exceptions as well this could simplify our internal `getErrorMessages` methods (yet it doesn't make this PR more complicated), wdyt?

eternoendless: Ideally it should extend [Throwable](https://www.php.net/manual/en/class.throwable.php), then it wouldn't be empty. But it's not possible. It's just a marker interface so that you have more flexibility to create your own exceptions in this specific case.

From a DX standpoint, it's easier to use the exception message. Regarding `FlashErrorInterface` it could be integrated in the core later and be made a parent to this one.

eternoendless: Ideally it should extend [Throwable](https://www.php.net/manual/en/class.throwable.php), then it wouldn't be empty. But it's not possible. It's just a marker interface so that you have more flexibility to create your own exceptions in this specific case.

From a DX standpoint, it's easier to use the exception message. Regarding `FlashErrorInterface` it could be integrated in the core later and be made a parent to this one.

jolelievre: Ok let's do that for now then :)

jolelievre (approved):
Thanks @matks

matks:
I fixed the empty demo module, QA can proceed 🚀

matks:
Thank you for review and QA 🎉

PR #14248, opened 2019-06-17
Cannot use accented URL with arabic language

### LINKED ISSUE #14184: Cannot use accented URL with arabic language
**Describe the bug**
problem with language translation RTL in category (Arabic for example)

**Screenshots**
no-valid0
no-valid

**Additionnal information**
PrestaShop version: 1.7.6.0 Beta
PHP version: 7.2

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Use Accented URL feature to check if link_rewrite in category can use arabic language (and any others languages)
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14184 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14248)


src/Adapter/Feature/AccentedUrlFeature.php
```diff
@@ -0,0 +1,87 @@
...
+namespace PrestaShop\PrestaShop\Adapter\Feature;
+
+use Feature;
+use PrestaShop\PrestaShop\Adapter\Configuration;
+use PrestaShop\PrestaShop\Core\Feature\FeatureInterface;
+
+/**
+ * This class manages Accented URL feature.
+ */
+class AccentedUrlFeature implements FeatureInterface
```
sarjon: This could be `final`.

src/Adapter/Feature/AccentedUrlFeature.php
```diff
@@ -0,0 +1,87 @@
...
+{
+    /**
+     * @var Configuration
+     */
+    private $configuration;
+
+    public function __construct(Configuration $configuration)
```
sarjon: If you where to depend on interface, then service could be moved to `Core` namespace.

Quetzacoalt91 (dismissed):

src/Adapter/Feature/AccentedUrlFeature.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        return $this->isActive();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isActive()
+    {
+        return $this->configuration->get('PS_ALLOW_ACCENTED_CHARS_URL');
```
sarjon: I'm not sure what are the rules for using `FeatureInterface`. It seems that it could be used for every setting if we follow this. Wdyt?

src/Adapter/Feature/AccentedUrlFeature.php
PierreRambaud: No other class in this directory are in final :thinking:

src/Adapter/Feature/AccentedUrlFeature.php
PierreRambaud: I did exactly what other services doing :thinking:

src/Adapter/Feature/AccentedUrlFeature.php
PierreRambaud: I think too, but after checking for MultistoreFeature, and FeatureFeature, this is what they are doing too :/
I think we can improve this "directory" after the 1.7.6 release (and take in consideration what we said on Gitter about services name)

src/Adapter/Feature/AccentedUrlFeature.php
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Adapter\Feature;
+
+use Feature;
```
jolelievre: I don't see any use of `Feature` class, I think you can remove it

src/Adapter/Feature/AccentedUrlFeature.php
jolelievre: I think the other services are in `Adapter` because they depend on a legacy class `Feature`
Since you don't have this dependency here you could indeed move this to the Core, you just need to replace `Configuration` with `PrestaShop\PrestaShop\Core\ConfigurationInterface`

jolelievre (commented):
I'm wondering if you really need to create a `FeatureInterface` to manage this, the existing ones seem to use configuration settings like `PS_FEATURE_FEATURE_ACTIVE` `PS_COMBINATION_FEATURE_ACTIVE`, which seems to indicate they are toggable features.

I'm wondering if the accented url is expected to behave the same way? Then aren't all toggable settings from the configuration potentially transformable into `FeatureInterface` This is something that is worth thinking about, but if we start doing it here (during a beta freeze at that) I'm afraid we open the door to "let's put any boolean setting into the feature folder!!"

Maybe as a quick fix you could only rely the configuration in your `FormType` and not introduce this `FeatureInterface` right now?

jolelievre (commented):
I'm wondering if you really need to create a `FeatureInterface` to manage this, the existing ones seem to use configuration settings like `PS_FEATURE_FEATURE_ACTIVE` `PS_COMBINATION_FEATURE_ACTIVE`, which seems to indicate they are toggable features.

I'm wondering if the accented url is expected to behave the same way? Then aren't all toggable settings from the configuration potentially transformable into `FeatureInterface` This is something that is worth thinking about, but if we start doing it here (during a beta freeze at that) I'm afraid we open the door to "let's put any boolean setting into the feature folder!!"

Maybe as a quick fix you could only rely the configuration in your `FormType` and not introduce this `FeatureInterface` right now?

src/Adapter/Feature/AccentedUrlFeature.php
PierreRambaud: I change everything by use the Configuration class instead :)

PierreRambaud:
@jolelievre Changes applied! Thanks for advice!

src/PrestaShopBundle/Form/Admin/Catalog/Category/AbstractCategoryType.php
```diff
@@ -30,6 +30,7 @@
 use PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints\DefaultLanguage;
 use PrestaShop\PrestaShop\Core\Domain\Category\SeoSettings;
 use PrestaShop\PrestaShop\Core\Feature\FeatureInterface;
+use PrestaShop\PrestaShop\Adapter\Configuration;
```
jolelievre: Can you use the `ConfigurationInterface` from core instead please?

jolelievre (approved):
Thanks @PierreRambaud

matks:
Thank you @PierreRambaud and @jolelievre

PR #14259, opened 2019-06-17
Avoid bug when creating manually created a language

### LINKED ISSUE #14247: Error when upgrading a added language
**Describe the bug**
Error when upgrading a added language 

**To Reproduce**
1. Go to International -> localization -> language
2. Click on add a language
3. Create a language (English, en, us)
4. Go to international -> traduction 
5. Click on add/upgrade language and choose the language you just add
5. See error

**Screenshots**
https://drive.google.com/file/d/1FE_2kQA5_K_L0-kH8um8rfH5vD-0Q8q7/view

**Additionnal information**
PrestaShop version: 1.7.6.x
PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Avoid bug when creating manually a language, the locale field is automatically filled when possible
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14247
| How to test?  | ~

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14259)

### DIFF
--- a/src/Adapter/Language/CommandHandler/AddLanguageHandler.php
+++ b/src/Adapter/Language/CommandHandler/AddLanguageHandler.php
@@ -100,6 +100,9 @@ final class AddLanguageHandler extends AbstractLanguageHandler implements AddLan
         $language = new Language();
         $language->name = $command->getName();
         $language->iso_code = $command->getIsoCode()->getValue();
+        if (false !== ($languageDetails = Language::getLangDetails($command->getIsoCode()->getValue()))) {
+            $language->locale = $languageDetails['locale'];
+        }
         $language->language_code = $command->getTagIETF()->getValue();
         $language->date_format_lite = $command->getShortDateFormat();
         $language->date_format_full = $command->getFullDateFormat();
--- a/classes/Language.php
+++ b/classes/Language.php
@@ -1073,13 +1073,8 @@ class LanguageCore extends ObjectModel
         if (Language::downloadLanguagePack($iso, $version, $errors)) {
             if ($install) {
                 Language::installLanguagePack($iso, $params, $errors);
-                Language::updateMultilangTable($iso);
             } else {
-                $lang_pack = self::getLangDetails($iso);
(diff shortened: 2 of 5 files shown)


src/Adapter/Language/CommandHandler/AddLanguageHandler.php
```diff
@@ -100,6 +100,7 @@ private function createLegacyLanguageObjectFromCommand(AddLanguageCommand $comma
         $language = new Language();
         $language->name = $command->getName();
         $language->iso_code = $command->getIsoCode()->getValue();
+        $language->locale = strtolower($language->iso_code) . '-' . strtoupper($language->iso_code);
```
PierreRambaud: Hum if it generate locale from iso_code, 'en' will generate 'en-EN', which is not avalid locale 🤔

Quetzacoalt91: Looks like you can have the data in `$command->getTagIETF()->getValue()`, don't you?

jolelievre: The problem is that the validation on IETF is not correct, you can input en, en-en, ... And even if the format was correct syntaxicaly it doesn't mean the locale exists
So I think getting the locale thanks to `Language::getLangDetails` is actually the least worst solution ^^

Quetzacoalt91:
As far as I know, this PR does not fix the issue when the language already exist in the database?

jolelievre:
> As far as I know, this PR does not fix the issue when the language already exist in the database?

No because you didn't want to change the `Translate` class to avoid hiding errors ^^

Quetzacoalt91:
Rhaaaaaa :skull:

matks (approved):

PierreRambaud:
Thanks @jolelievre

PR #14263, opened 2019-06-18
Fix display voucher conversion between currencies

### LINKED ISSUE #9914: [BOOM-3688] Voucher value of a currency is converted twice in cart summary
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-3688).

- _**Reporter:**_ infisual
- _**Created at:**_ Tue, 22 Aug 2017 19:02:07 +0200

<p>I have two currencies in my store, default is US$ and other is SAR. When a customer completed a previous order in SAR currency, their loyalty points converted to a voucher with SAR value with correct amount converted from US$ to SAR. Now when the customer is placing a new order with currency in SAR then they apply the voucher in checkout, the amount deducted is correct in shopping cart; however the cart summary line for the voucher is displaying the voucher name and value multiplied again in conversion rate as if it is converted from US$ to SAR. See attached example, the loyalty points converted 500 points to $5 which is equal to 18.75SAR (5 x 3.75 = 18.75). If I keep the voucher in SAR, the shopping cart will display the value as (18.75x3.75 = 70.30SAR). As if the voucher value is always treated as default US$</p>
        
Create a voucher with a value in secondary currency. Make a new order while in second currency, then apply the voucher code. The deduction and calculation is done correctly, however, the voucher text is displaying the wrong amount.

tomlev:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fixes voucher display when discount is in other currency than current one
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/9914
| How to test?  | see ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14263)

### DIFF
--- a/src/Adapter/Presenter/Cart/CartPresenter.php
+++ b/src/Adapter/Presenter/Cart/CartPresenter.php
@@ -511,10 +511,14 @@ class CartPresenter implements PresenterInterface
                     $value = 0;
                 } else {
                     // convert to default currency
+                    $defaultCurrencyId = (int) Configuration::get('PS_CURRENCY_DEFAULT');
                     $value /= $currencyFrom->conversion_rate;
-                    // convert to destination currency
-                    $value *= $currencyTo->conversion_rate;
+                    if ($defaultCurrencyId == $currencyTo->id) {
+                        // convert to destination currency
+                        $value *= $currencyTo->conversion_rate;
+                    }
                 }
+                // following will do currency conversion to current one
                 $cartVoucher['reduction_formatted'] = $this->priceFormatter->convertAndFormat($value);
             }
 
--- a/src/Adapter/Presenter/Cart/CartPresenter.php
+++ b/src/Adapter/Presenter/Cart/CartPresenter.php
@@ -505,6 +505,16 @@ class CartPresenter implements PresenterInterface
                 $cartVoucher['reduction_formatted'] = $cartVoucher['reduction_percent'] . '%';
             } elseif (isset($cartVoucher['reduction_amount']) && $cartVoucher['reduction_amount'] > 0) {
                 $value = $this->includeTaxes() ? $cartVoucher['reduction_amount'] : $cartVoucher['value_tax_exc'];
+                $currencyFrom = new \Currency($cartVoucher['reduction_currency']);
+                $currencyTo = new \Currency($cart->id_currency);
+                if ($currencyFrom->conversion_rate == 0) {
+                    $value = 0;
+                } else {
+                    // convert to default currency
+                    $value /= $currencyFrom->conversion_rate;
+                    // convert to destination currency
+                    $value *= $currencyTo->conversion_rate;
+                }
                 $cartVoucher['reduction_formatted'] = $this->priceFormatter->convertAndFormat($value);
             }
 
--- a/src/Core/Cart/CartRuleCalculator.php
+++ b/src/Core/Cart/CartRuleCalculator.php
@@ -122,65 +122,67 @@ class CartRuleCalculator
         }
 
         // Discount (%) on the whole order
-        if ($cartRule->reduction_percent && $cartRule->reduction_product == 0) {
-            foreach ($this->cartRows as $cartRow) {
-                $product = $cartRow->getRowData();
-                if ((($cartRule->reduction_exclude_special && !$product['reduction_applies'])
-                        || !$cartRule->reduction_exclude_special)) {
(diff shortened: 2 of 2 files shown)


src/Adapter/Presenter/Cart/CartPresenter.php
```diff
@@ -505,6 +505,16 @@ private function getTemplateVarVouchers(Cart $cart)
                 $cartVoucher['reduction_formatted'] = $cartVoucher['reduction_percent'] . '%';
             } elseif (isset($cartVoucher['reduction_amount']) && $cartVoucher['reduction_amount'] > 0) {
                 $value = $this->includeTaxes() ? $cartVoucher['reduction_amount'] : $cartVoucher['value_tax_exc'];
+                $currencyFrom = new \Currency($cartVoucher['reduction_currency']);
+                $currencyTo = new \Currency($cart->id_currency);
+                if ($currencyFrom->conversion_rate == 0) {
+                    $value = 0;
+                } else {
+                    // convert to default currency
+                    $value /= $currencyFrom->conversion_rate;
```
matks, line 515: I'm worried about rounding issues when I see this 😐 . Am I too paranoid ?

src/Core/Cart/CartRuleCalculator.php
```diff
@@ -122,65 +122,67 @@ protected function applyCartRule(CartRuleData $cartRuleData, $withFreeShipping =
...
-            foreach ($this->cartRows as $cartRow) {
-                $product = $cartRow->getRowData();
-                if (((($cartRule->reduction_exclude_special && !$product['reduction_applies'])
-                        || !$cartRule->reduction_exclude_special)) && ($cartRowCheapest === null
-                        || $cartRowCheapest->getInitialUnitPrice()->getTaxIncluded() > $cartRow->getInitialUnitPrice()
-                            ->getTaxIncluded())
-                ) {
-                    $cartRowCheapest = $cartRow;
+            // Discount (%) on the cheapest product
+            if ($cartRule->reduction_product == -1) {
```
matks, line 148: Can we improve a little bit the code and replace this `-1` by something that has meaning, like either a variable or a constant ?

src/Core/Cart/CartRuleCalculator.php
```diff
@@ -122,65 +122,67 @@ protected function applyCartRule(CartRuleData $cartRuleData, $withFreeShipping =
...
-                foreach ($this->cartRows as $cartRow) {
-                    $product = $cartRow->getRowData();
-                    if (in_array($product['id_product'] . '-' . $product['id_product_attribute'], $selected_products)
-                        || in_array($product['id_product'] . '-0', $selected_products)
-                           && (($cartRule->reduction_exclude_special && !$product['reduction_applies'])
-                               || !$cartRule->reduction_exclude_special)) {
-                        $amount = $cartRow->applyPercentageDiscount($cartRule->reduction_percent);
-                        $cartRuleData->addDiscountApplied($amount);
+            // Discount (%) on the selection of products
+            if ($cartRule->reduction_product == -2) {
```
matks, line 174: Can we improve a little bit the code and replace this `-2` by something that has meaning, like either a variable or a constant ?

matks (commented):
I would like us to put all of this logic inside `private` methods 😅 but I'm not going to bother too much. Added a few questions/suggestions

src/Adapter/Presenter/Cart/CartPresenter.php
PierreRambaud: You're not

src/Core/Cart/CartRuleCalculator.php
tomlev: this part was just refactored, I'm not sure to find quickly what these values are meant to be

src/Adapter/Presenter/Cart/CartPresenter.php
tomlev: unfortunately I've no tool to handle it correctly, nor guidelines to know how to round it.
following call $this->priceFormatter->convertAndFormat($value); will round correctly if the precision is large enough, which will be the case if the conversion rate is not 1000000+ ^^

matks (dismissed):

marionf:
@tomlev I still have the same issue with this PR

capture d'écran_1699

matks (approved):

matks:
Thank you @tomlev

PR #14287, opened 2019-06-19
Behat tests for Category

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | Adds behat tests for Category commands.
| Type?         | improvement
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | n/a
| How to test?  | Build should eventually be green.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14287)

### DIFF
--- a/tests/Integration/Behaviour/Features/Context/CategoryFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/CategoryFeatureContext.php
@@ -100,6 +100,9 @@ class CategoryFeatureContext extends AbstractPrestaShopFeatureContext
         $propertiesKey = sprintf('%s_properties', $reference);
         $properties = SharedStorage::getStorage()->getWithDefault($propertiesKey, []);
 
+        // Test database is loaded with fixtures
+        // which have 3 customer groups
+        // that are defined below
         $groupsMap = [
             'Visitor' => 1,
             'Guest' => 2,
--- a/tests/Integration/Behaviour/Features/Context/CategoryFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/CategoryFeatureContext.php
@@ -1,0 +1,209 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace Tests\Integration\Behaviour\Features\Context;
+
+use Category;
+use Configuration;
+use Group;
+use RuntimeException;
+use Tests\Integration\Behaviour\Features\Context\Util\PrimitiveUtils;
+
+class CategoryFeatureContext extends AbstractPrestaShopFeatureContext
+{
+    /**
+     * @var int
+     */
+    private $defaultLanguageId;
+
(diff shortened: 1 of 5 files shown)


tests/Integration/Behaviour/Features/Context/CategoryFeatureContext.php
```diff
@@ -0,0 +1,209 @@
...
+     */
+    public function specifyGroupAccess($value, $reference)
+    {
+        $propertiesKey = sprintf('%s_properties', $reference);
+        $properties = SharedStorage::getStorage()->getWithDefault($propertiesKey, []);
+
+        $groupsMap = [
+            'Visitor' => 1,
+            'Guest' => 2,
+            'Customer' => 3,
```
sarjon, line 109: @matks is it ok to use hardcoded ids? They are part of demo data.

tests/Integration/Behaviour/Features/Scenario/Category/category_management.feature
```diff
@@ -0,0 +1,25 @@
...
+
+  Scenario: Adding new Category
+    Given I specify "name" "PC parts" for new category "category1"
+    And I specify displayed to be "enabled" for new category "category1"
+    And I specify "parent category" "Home Accessories" for new category "category1"
+    And I specify "description" "Best PC parts" for new category "category1"
+    And I specify "meta title" "PC parts meta title" for new category "category1"
+    And I specify "meta description" "PC parts meta description" for new category "category1"
+    And I specify "friendly url" "pc-parts" for new category "category1"
+    And I specify group access for "Customer,Guest,Visitor" for new category "category1"
```
sarjon, line 16: @matks instead of using table I've decided to write step for each property? The amount of lines are the same, but the context that it gives (for example compare `And I specify group access` with `And I specify "meta title"`, in case of table, they would be the same) is much clearer imo, wdyt?

tests/Integration/Behaviour/Features/Context/CategoryFeatureContext.php
PierreRambaud: Not a good idea. Functional tests must be able to be executed in prod environement. It's always better to setup and cleanup generated data.

tests/Integration/Behaviour/Features/Context/CategoryFeatureContext.php
```diff
@@ -0,0 +1,209 @@
...
+     */
+    private $defaultLanguageId;
+
+    public function __construct()
+    {
+        $this->defaultLanguageId = (int) Configuration::get('PS_LANG_DEFAULT');
+    }
+
+    /**
+     * @Given /^I specify "(.+)" "(.+)" for new category "(.+)"$/
```
PierreRambaud, line 48: The phrase is too generic.
Feel free to improve, we need to understand what kind of value we can write. Pm don't know it must be a property, value and reference.

```suggestion
     * @Given /^I specify "(.+)" as property and "(.+)" as value for new category with reference"(.+)"$/
```

sarjon: Hmmmm, how about:
`@Given /^I specify property "(.+)" with value "(.+)" for new category "(.+)"` or 
`@Given /^I specify property "(.+)" and "(.+)" as its value for new category "(.+)"`?

tests/Integration/Behaviour/Features/Context/CategoryFeatureContext.php
sarjon: These are not really functional tests and setting up something that is already there seems redundant. :/ 

In addition, in some cases setting up test data could take more steps/code than an actual tests.

tests/Integration/Behaviour/Features/Context/CategoryFeatureContext.php
PierreRambaud: Look good :+1: !

tests/Integration/Behaviour/Features/Context/CategoryFeatureContext.php
matks: Currently, the database reset function being used loads some data so we can use it IMO. But if one day we clean up the reset database processing and remove the fixtures loading (instead we only load an empty database schema), there we will have to load it by ourselves.

Can you add a comment to write this down, in order to explain this to other developers who would work on this test ?

tests/Integration/Behaviour/Features/Scenario/Category/category_management.feature
matks: It works 😄 I find tables more usable and more readable but I find them especially useful to avoid writing all these steps. If writing all these steps is not bothering you too much, all right.

matks:
Thank you @sarjon

matks:
Solves partially https://github.com/PrestaShop/PrestaShop/issues/14480

matks (dismissed):

matks:
Rebased

matks (approved):

sarjon:
@matks could you rerun build?

PierreRambaud:
Thanks @sarjon

PR #14290, opened 2019-06-19
Extend customer page

### LINKED ISSUE #14275: Client tab is not display properly in laptop
**Describe the bug**
Client tab is not display properly in laptop

Appear only with little screen.

**To Reproduce**
1. Go to customer page in BO
2. See the customer tab

**Screenshots**
Capture d’écran 2019-06-18 à 16 52 53

**Additionnal information**
PrestaShop version: 1.7.6.X
PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Add special case to allow customer page to extend in width
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14275
| How to test?  | ~

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14290)


admin-dev/themes/new-theme/scss/components/layout/_content_div.scss
```diff
@@ -38,6 +38,14 @@
   }
 }
 
+// Special case for customer page the table has too many columns to fit the screen so we allow an horizontal scroll
```
matks: I think the same issue can happen on any page, providing it has enough columns 🤔 

I tried with "languages", you can see the issue with a specific screen size

jolelievre: Yes it can, but this is a quick fix just for the customer page because it can happen on many screens
For other pages we need to think about a more permanent fix for all pages, but it will be done in 177 This is what we decided with @TristanLDD

matks: Makes sense 👍

matks: Shouldn't it go in https://github.com/PrestaShop/PrestaShop/tree/develop/admin-dev/themes/new-theme/scss/pages ?

matks (approved):

PR #14298, opened 2019-06-20
Fix english mail generation on fresh install

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | English mails were not installed on fresh install, because the default theme was not in database yet, also fix a warning in Mail::send which was annoying during CLI install (logs tried to be sent by mail causing lots of Warning lines)
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | Perform a fresh install and check that english mails use the modern theme

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14298)

### DIFF
--- a/classes/Mail.php
+++ b/classes/Mail.php
@@ -151,34 +151,38 @@ class MailCore extends ObjectModel
             $idShop = Context::getContext()->shop->id;
         }
 
-        $keepGoing = array_reduce(
-            Hook::exec(
-                'actionEmailSendBefore',
-                [
-                    'idLang' => &$idLang,
-                    'template' => &$template,
-                    'subject' => &$subject,
-                    'templateVars' => &$templateVars,
-                    'to' => &$to,
-                    'toName' => &$toName,
-                    'from' => &$from,
-                    'fromName' => &$fromName,
-                    'fileAttachment' => &$fileAttachment,
-                    'mode_smtp' => &$mode_smtp,
-                    'templatePath' => &$templatePath,
-                    'die' => &$die,
-                    'idShop' => &$idShop,
-                    'bcc' => &$bcc,
-                    'replyTo' => &$replyTo,
-                ],
-                null,
-                true
-            ),
-            function ($carry, $item) {
-                return ($item === false) ? false : $carry;
-            },
+        $keepGoing = Hook::exec(
+            'actionEmailSendBefore',
+            [
+                'idLang' => &$idLang,
+                'template' => &$template,
+                'subject' => &$subject,
+                'templateVars' => &$templateVars,
+                'to' => &$to,
+                'toName' => &$toName,
+                'from' => &$from,
+                'fromName' => &$fromName,
+                'fileAttachment' => &$fileAttachment,
+                'mode_smtp' => &$mode_smtp,
+                'templatePath' => &$templatePath,
+                'die' => &$die,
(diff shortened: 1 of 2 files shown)


jolelievre:
Close that ugly thing in favor of https://github.com/PrestaShop/PrestaShop/pull/14309

PierreRambaud (dismissed):

sarahdib:
Emails are no longer sent  @matks

classes/Mail.php
```diff
@@ -182,6 +182,8 @@ function ($carry, $item) {
                 },
                 true
             );
+        } elseif (is_array($keepGoing) && empty($keepGoing)) {
```
matks: @PierreRambaud @Quetzacoalt91 In the previous implementation of this logic, since `$keepGoing` was assigned as the result of the `array_reduce()` operation, and `true` was given as the 3rd argument of `array_reduce()`, what was happening was that **when the Hook call would return an empty array, `$keepGoing` value was true**. Then the logic would continue.

With Jonathan modification, we compte 1st the result of the Hook call in `$keepGoing` then we perform the `array_reduce()` **only if `$keepGoing` is valid** using the infamous `if ($keepGoing)`. So we got a faulty behavior when the Hook call would return an empty array because `if ($keepGoing)` would be false, then the `array_reduce()` would not be executed and consequently `$keepGoing` would not be converted to `true`, which would stop the execution of the logic.

I see 2 ways to fix it:
1) revert the modification of Jonathan to restore previous logic
2) handle the empty array usecase as I did to force `$keepGoing` to true

matks: If I was sure of what are the possible values to be returned by `Hook::exec('actionEmailSendBefore',...)` I could handle it more properly. But I'm not sure.

matks:
Additionaly to QA test, we need a Dev test to check hook 'actionEmailSendBefore' works as before ⚠️

classes/Mail.php
PierreRambaud: Can you please change by 
```php
if (empty($keepGoing)) {
    // Continue since there is no hook called
    $keepGoing = true;
} else {
    $keepGoing = array_reduce(
        $keepGoing,
        function ($carry, $item) {
            return ($item === false) ? false : $carry;
        },
        true
    );
}


if (!$keepGoing) {
    return true;
}
```
With a better comment than mine? ^^

jolelievre: actually I changed this code because during install process, the hook always returned `null` which kicked a warning about using null instead of array in `array_reduce`
My modif was clearly not ideal, my bad But I'm afraid this one changes the behaviour as well
Maybe we should simply check after the hook:
```php
if (null === $keepGoing) {
    return true;
}
```

This will keep the previous behaviour and avoid the warning

matks:
@sarahdib So we removed the improvement of Jonathan as it introduced crazy edge-cases impacts. We only check the `null` case we needed.

Please check (I already did, but better 2 checks than 1) :
- this PR fixes the initial bug
- mails are sent AND they are sent using the new design Jonathan worked on 🤔 
- install is OK

PierreRambaud (approved):

matks:
Thank you @jolelievre for the PR and @PierreRambaud for the review

PR #14308, opened 2019-06-20
Fix wrong tax when duplicating product in BO with several taxes (multishop)

### LINKED ISSUE #14302: Duplicate Product in BO with several taxes (multishop)
**To Reproduce**

1. Set a multishop with 2 shops (or more) and several taxe rule. For instance:
- shop FR with Tax rate FR 20%
- shop CH with Tax rate CH 10%
2. Create a product in context all shop
3. On each shop set a specific tax rate
4. Finally on products board, select the previous product and click on duplicate product on all shop context.

**Describe the bug**

A product duplicated from all shop context set the default shop's tax for all shops. It should set the same tax than the original product (the tax corresponding to the shop).

**Additionnal information**
PrestaShop version: 1.7.x
PHP version: all

202-ecommerce:
| Questions     | Answers
| Branch?       | develop
| Description?  | A product duplicated from all shop context set the default shop's tax for all shops. It should set the same tax than the original product (the tax corresponding to the shop).
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14302 
| How to test?  | Steps to reproduce the behavior:
1. Set a multishop with 2 shops (or more) and several taxe rule. For instance:
    - shop FR with Tax rate FR 20%
    - shop CH with Tax rate CH 10%
2. Create a product in context all shop
3. On each shop set a specific tax rate
4. Finally on products board, select the previous product and click on duplicate product on all shop context.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14308)

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -4456,8 +4456,8 @@ class ProductCore extends ObjectModel
         if (!empty($results)) {
             foreach ($results as $result) {
                 Db::getInstance()->update(
-                    'product_shop', 
-                    array('id_tax_rules_group' => (int) $result['id_tax_rules_group']), 
+                    'product_shop',
+                    array('id_tax_rules_group' => (int) $result['id_tax_rules_group']),
                     'id_product=' . (int) $id_product_new . ' AND id_shop = ' . (int) $result['id_shop']
                 );
             }
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -4443,7 +4443,7 @@ class ProductCore extends ObjectModel
 
         return Db::getInstance()->insert('product_tag', $data);
     }
-    
+
     public static function duplicateTaxes($id_product_old, $id_product_new)
     {
         $query = new DbQuery();
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -4455,11 +4455,13 @@ class ProductCore extends ObjectModel
 
         if (!empty($results)) {
             foreach ($results as $result) {
-                Db::getInstance()->update(
+                if (!Db::getInstance()->update(
                     'product_shop',
                     array('id_tax_rules_group' => (int) $result['id_tax_rules_group']),
                     'id_product=' . (int) $id_product_new . ' AND id_shop = ' . (int) $result['id_shop']
-                );
+                )) {
+                    return false;
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


classes/Product.php
```diff
@@ -4443,6 +4443,28 @@ public static function duplicateTags($id_product_old, $id_product_new)
...
+        $query = new DbQuery();
+        $query->select('id_tax_rules_group, id_shop');
+        $query->from('product_shop');
+        $query->where('`id_product` = ' . (int) $id_product_old);
+
+        $results = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS($query->build());
+
+        if (!empty($results)) {
+            foreach ($results as $result) {
+                Db::getInstance()->update(
```
PierreRambaud: We can maybe check if something went wrong and return false if a problem happened?

202-ecommerce:
@PierreRambaud 

We finally chose the third way (deleting the duplicated product if the product duplication process returns false).

What do you think about this solution ?

matks:
> @PierreRambaud
> 
> We finally chose the third way (deleting the duplicated product if the product duplication process returns false).
> 
> What do you think about this solution ?

So if `Db::getInstance()->update(///)` returns false, this will delete the product and "cancel" the duplication (note: I hope it works well and the cancel impact all shops 😛 )

What happens if the tax rate was already the good one ? Wouldn't it cancel the duplication process ?

PierreRambaud:
@matks As I understand, if the tax rate was already the good one, nothing happened, the `Db::getInstance()->update` will only returns false in case of failure, so it continues.

I think it's a good idea to delete the product if something's going wrong, but (I'm sorry), what happened if `$product->add()` failed and returns false, you'll be able to use `$product->delete()`?

PierreRambaud (approved):

202-ecommerce:
@matks It won't change anything for the update and cancel the product for all shops.
The duplicated product will be deleted after its creation only if it exists since our last commit.

Thanks to @PierreRambaud for the approval.

matks:
Thank you @202-ecommerce, another wonderful PR 😄 about multishop

PR #14267, opened 2019-06-18
Migration of products list

tomas862:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | This PR migrates product list according to our standarts - in this part it only migrates the display related things and not a single action. To access it for now a temporary url is created **admin-dev/index.php/sell/catalog/products/v2/**
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10631
| How to test?  | I'll write different test cases and what to test since this page is not going to be migrated all within single PR as it might get too huge. There might be more so we should not miss any

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14267)

PR #14320, opened 2019-06-21
Introduce NumberMinMaxFilterType and IntegerMinMaxFilterType

tomas862:
| Questions     | Answers
| Branch?       | develop
| Description?  | As for products list page I need two inputs besides each - one type is  for filtering ids while another type is for numbers such as price.
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | partialy required for https://github.com/PrestaShop/PrestaShop/pull/14267
| How to test?  | Within this PR this component will not appear anywhere. It will be used by others once it is merged so no manual testing is required

To do after release of this:

- [ ] Document it!

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14320)


rokaszygmantas:
```php
$minMaxForm = $this->createForm(MinMaxType::class, [], [
            'required' => true,
            'min_field' => IntegerType::class,
            'min_field_options' => [
                'help' => 'inner help message',
            ],
            'max_field' => NumberType::class,
        ]);
```

maybe we should introduce two min max types, one for integer and one for float values (if it's needed)? I think it's better compared to having one generic min max type and having to define types when building the form.

tomas862:
> ```
> $minMaxForm = $this->createForm(MinMaxType::class, [], [
>             'required' => true,
>             'min_field' => IntegerType::class,
>             'min_field_options' => [
>                 'help' => 'inner help message',
>             ],
>             'max_field' => NumberType::class,
>         ]);
> ```
> 
> maybe we should introduce two min max types, one for integer and one for float values (if it's needed)? I think it's better compared to having one generic min max type and having to define types when building the form.

@rokaszygmantas  we have more then just two types:

`MoneyType`
`PercentType`
`RangeType`

but Im thinking right now that you might be right here - I thought about this as well. 

Maybe we could have :

`IntegerMinMaxType`
`NumberMinMaxType`
`PercentMinMaxType`
`MoneyMinMaxType`

and we also have native type `RangeType` which is `<input type="range">` so it would be `RangeMinMaxType` which basically adds two sliders 

anyway this is closely related to my second problem which is:
how to control layout in grid since same html does not fit anymore so well.

two options:

1. Either we will have `GridIntegerMinMaxType`, `GridNumberMinMaxType` etc...
2. Either I try to apply some css magic to fit these types in Grid so the same type can be rendered in form and in grid.

tomas862:
gonna apply it for grid use case only - when someone will need it they can extend the functionality.

tomas862:
@matks this one is required for continue of product page listing development - review when you can :)

matks (approved):

matks:
Review is ✅ and I restart Travis

tomas862:
need restart Travis again - too random error to be related with this PR :D

matks:
LGTM

PR #14331, opened 2019-06-21
Handle bad theme error nicely

### LINKED ISSUE #14326: Command EnableThemeHandler
**Describe the bug**

Using a string in the php reset function.

**To Reproduce**
1. Duplicate the classic theme to create a new theme
2. Arrange for the new theme to install normally.
3. Use the classic theme.
4. Remove in the new theme theme.css (in new_theme/assets/css/theme.css).
5. Define "_PS_MODE_DEV_" on true
6. Enable the new theme.

**Screenshots**

**Additionnal information**
PrestaShop version: 1.7.6.0
PHP version: 7.1
bug_1

Value of $errors :

image

matks:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | If Theme Install fails and return a string error message, displays it nicely
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14326
| How to test?  | See ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14331)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Improve/Design/ThemeController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Improve/Design/ThemeController.php
@@ -509,7 +509,7 @@ class ThemeController extends AbstractAdminController
                     'Admin.Modules.Notification',
                     [
                         '%action%' => strtolower($this->trans('Install', 'Admin.Actions')),
-                        '%module%' => $e->getModuleName(),
+                        '%module%' => ($e instanceof FailedToEnableThemeModuleException) ? $e->getModuleName() : '',
                         '%error_details%' => $e->getMessage(),
                     ]
                 ),
--- a/src/Core/Addon/Theme/ThemeManager.php
+++ b/src/Core/Addon/Theme/ThemeManager.php
@@ -239,7 +239,7 @@ class ThemeManager implements AddonManagerInterface
      *
      * @param string $themeName The technical theme name
      *
-     * @return array|false
+     * @return array|string|bool
      */
     public function getErrors($themeName)
     {
--- a/src/Core/Domain/Theme/CommandHandler/EnableThemeHandler.php
+++ b/src/Core/Domain/Theme/CommandHandler/EnableThemeHandler.php
@@ -86,9 +86,17 @@ final class EnableThemeHandler implements EnableThemeHandlerInterface
 
         if (!$this->themeManager->enable($plainThemeName)) {
             $errors = $this->themeManager->getErrors($plainThemeName);
-            $error = is_array($errors) ? reset($errors) : '';
 
-            throw new CannotEnableThemeException(reset($error));
+            if (is_array($errors)) {
+                $error = reset($errors);
+            } elseif ($errors) {
+                $error = $errors;
+            } else {
+                // handle bad error usecases
+                $error = '';
+            }
+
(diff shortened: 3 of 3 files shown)


src/PrestaShopBundle/Controller/Admin/Improve/Design/ThemeController.php
```diff
@@ -509,7 +509,7 @@ private function handleEnableThemeException(ThemeException $e)
                     'Admin.Modules.Notification',
                     [
                         '%action%' => strtolower($this->trans('Install', 'Admin.Actions')),
-                        '%module%' => $e->getModuleName(),
+                        '%module%' => ($e instanceof FailedToEnableThemeModuleException) ? $e->getModuleName() : '',
```
matks, line 512: Additionnal bugfix (found while working on this PR) :
If `$e` was an instance of something else than `FailedToEnableThemeModuleException` (example: `CannotEnableThemeException`) this code would fail

PierreRambaud: Why do we not add unit tests to prevent this kind of situation?

matks: Because it's a Controller, so I think E2E tests are a better solution to test it, don't you think ?

PierreRambaud: I was talking about `EnableThemeHandler.php` sorry ^^

matks: All right 😄
Unfortunately, we only were able to build a proper setup / method of using Behat at the end of 1.7.6 development phase . For each new page we'll migrate in 1.7.7, we'll have every Command (and most Queries) tested with Behat.
For the pages migrated before, we'll use pauses time windows (such as waiting for code review) to complete the coverage.

However this one will probably be a nightmare to test 🤔as a Theme installation involves the database, the filesystem, file uploads...

PierreRambaud: If it's unit tests we don't care what we are using (db, fs, etc).

matks: We use Behat here because any Handler is actually part of the Adapter layer which uses extensively legacy classes a.k.a "static classes paradise" and is a nightmare to unit test 😢so our Behat tests are basically "black box testing" as isolation is veeeeery hard to achieve

PierreRambaud: Not sure file like `src/Core/Domain/Theme/CommandHandler/EnableThemeHandler.php` are hard to unit test :thinking:

matks: Oh 😶you're right. Usually all Handlers are in Adapter context but not this one (probably because we could get rid of all legacy logic) 😄

PierreRambaud (approved):

matks:
Thank you @PierreRambaud for review and @sarahdib for QA

Issue #14131, opened 2019-06-10
Improve dealing with Carts in BO (dont rely on Context)


matks:
**Is your feature request related to a problem? Please describe.**

We are migrating the Orders page. During this migration, we have discovered that *multiple processings related to carts management require to load the Cart into the common Context*.

Example: create a new Cart from BO
```
$cart = new Cart();
        $cart->id_shop_group = $order->id_shop_group;
        $cart->id_shop = $order->id_shop;
        $cart->id_customer = $order->id_customer;
        $cart->...

        Context::getContext()->cart = $cart;
        Context::getContext()->customer = new Customer($order->id_customer);

        return $cart;
```
We need to perform `Context::getContext()->cart = $cart;` in order for this Cart object to be properly configured.

This is a big issue and does not make sense from an architectural point of view.

If you decide to create multiple carts, this means you're going to call multiple times `Context::getContext()->cart = $cart;` with different carts. It has no meaning.

**Describe the solution you'd like**
We should extract all the processings related to cart management from the Context, at least for the ones required for the BO.

matks:
ping @sarjon

sarjon:
> We should extract all the processings related to cart management from the Context, at least for the ones required for the BO.

I'm afraid that other services not related to cart/order like price calculation or any other may depend on context that are not directly related to checkout, but used in checkout. :cry:

matks:
> > We should extract all the processings related to cart management from the Context, at least for the ones required for the BO.
> 
> I'm afraid that other services not related to cart/order like price calculation or any other may depend on context that are not directly related to checkout, but used in checkout. 😢

😭

Hlavtox:
Removing this legacy coupling is one of the key points to fix pricing system.

boo-code:
Spent some time mapping this across the pricing/cart path. The encouraging part: the project already has the architecture to fix this — the work is *finishing its application to carts/pricing*, not inventing anything new.

## The direction is already set — ADR 0024

[...]

PR #14352, opened 2019-06-25
Move Cart secure_key setter at the right place

jocel1:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix potentiel issue with setting the wrong secureKey to an existing cart
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | fill a cart while not logged, and verify cart is still ok after login

Let say we do the following:

```
$cart = new Cart($idCart1); // $idCart1 belongs to customer 2
$context = Context::getContext();
$context->cart = $cart;
$customer = new Customer($idCustomer);
$context->updateCustomer($idCustomer);
$cart->save();
```
before the fix, if PS_CART_FOLLOWING is set and $idCustomer have a not ordered cart, the $context->cart (and hence $cart) will have its secure key updated, but not its $id_customer, because $context->cart and $cart are copied by reference.
This could cause some nasty bugs, especially in modules handling "loginAsCustomer"

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14352)

### DIFF
--- a/classes/Context.php
+++ b/classes/Context.php
@@ -335,15 +335,16 @@ class ContextCore
 
         if (Configuration::get('PS_CART_FOLLOWING') && (empty($this->cookie->id_cart) || Cart::getNbProducts($this->cookie->id_cart) == 0) && $idCart = (int) Cart::lastNoneOrderedCart($this->customer->id)) {
             $this->cart = new Cart($idCart);
+            $this->cart->secure_key = $customer->secure_key;
         } else {
             $idCarrier = (int) $this->cart->id_carrier;
+            $this->cart->secure_key = $customer->secure_key;
             $this->cart->id_carrier = 0;
             $this->cart->setDeliveryOption(null);
             $this->cart->updateAddressId($this->cart->id_address_delivery, (int) Address::getFirstCustomerAddressId((int) ($customer->id)));
             $this->cart->id_address_delivery = (int) Address::getFirstCustomerAddressId((int) ($customer->id));
             $this->cart->id_address_invoice = (int) Address::getFirstCustomerAddressId((int) ($customer->id));
         }
-        $this->cart->secure_key = $customer->secure_key;
         $this->cart->id_customer = (int) $customer->id;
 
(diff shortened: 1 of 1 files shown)


matks:
Thanks @jocel1 !
I created issue https://github.com/PrestaShop/PrestaShop/issues/14131 to deal once and for all with all possible issues of the current Cart / Context coupling.

classes/Context.php
```diff
@@ -332,12 +332,13 @@ public function updateCustomer(Customer $customer)
         $customer->logged = 1;
         $this->cookie->email = $customer->email;
         $this->cookie->is_guest = $customer->isGuest();
-        $this->cart->secure_key = $customer->secure_key;
 
         if (Configuration::get('PS_CART_FOLLOWING') && (empty($this->cookie->id_cart) || Cart::getNbProducts($this->cookie->id_cart) == 0) && $idCart = (int) Cart::lastNoneOrderedCart($this->customer->id)) {
             $this->cart = new Cart($idCart);
+            $this->cart->secure_key = $customer->secure_key;
         } else {
```
matks, line 339: Would it make sense to add another validation layer by adding something like `if ($this->cart !== null)` ?

jocel1: I don't think so, new Cart() should always return a cart, or throw an Exception

matks: I was targeting the `else` case 😄 but the github UI makes it look like I show the `if`

PierreRambaud: Why not adding this on line 348 after 
https://github.com/PrestaShop/PrestaShop/pull/14352/files#diff-c606149f06205f3d2db241862eb4e84eR348?

jocel1: I tried that before but Travis was failing so I try with this approach, perhaps the upcoming function calls need the secure key

PierreRambaud (approved):
Let's go!

PierreRambaud:
Thanks @jocel1

PR #13903, opened 2019-05-21
Migrate Sell > Orders > Credit slips page

zuk3975:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Migrate Sell > Orders > Credit slips page to symfony framework. Waiting for clarification of how options should behave when multi-shop is enabled #14357
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #10581 
| How to test?  | Go to Sell > Orders > Credit Slips page. It must display a list of credit slips, a form for generating pdf by date range and options form with slip prefix field. Overall everything must behave as in legacy page. :warning:

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/13903)

PR #14353, opened 2019-06-25
[POC] adds attribute rendering for LinkColumn

tomas862:
| Questions     | Answers
| Branch?       | develop
| Description?  | In my specific case I needed the page to open in a new tab. `<a target="_blank"></a>` will fit my need so I improved `LinkColumn` - but I added more to it. More explanation below :arrow_double_down: 
| Type?         | improvement
| Category?     |  BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | partially required for https://github.com/PrestaShop/PrestaShop/pull/14267
| How to test?  | no need for manual testing

for me right now I only need `taget ="_blank"` attribute and property configuration for `<a></a>` tag. But what if I need more for this - anything from this huge list https://www.w3schools.com/tags/tag_a.asp. There are two options:

1. For each new attribute create new allowed type
2. (My proposal) lets add native `attr` type which already exists in symfony form types and apply it accordingly.

so the result would be:

```php

>add((new LinkColumn('name'))
                ->setName($this->trans('Name', [], 'Admin.Global'))
                ->setOptions([
                    'field' => 'name',
                    'route' => 'admin_cms_pages_index',
                    'route_param_name' => 'id_cms_category',
                    'route_param_field' => 'id_cms_category',
                    'attr' => [
                        'target' => '_blank',
                        'rel' => 'author',
                        'class' => 'btn btn-primary',
                    ],
                ])
            )

```

pros:
1. Highly customizable.
2. Symfony does this already for its form types

cons:
1. this will allow for devs to send even the css from php part ( or we can safelocks to prevent adding styles and class in attribures array :thinking:  ?)

@matks , @rokaszygmantas , @sarjon , @zuk3975, @PierreRambaud  wdyt?

P.s if confirmed this can be applied not only for LinkColumn

- [ ] update docs if any of the column is changed

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14353)

### DIFF
--- a/src/Core/Grid/Column/Type/Common/LinkColumn.php
+++ b/src/Core/Grid/Column/Type/Common/LinkColumn.php
@@ -50,6 +50,7 @@ final class LinkColumn extends AbstractColumn
         $resolver
             ->setDefaults([
                 'sortable' => true,
+                'attr' => [],
             ])
             ->setRequired([
                 'field',
@@ -62,6 +63,7 @@ final class LinkColumn extends AbstractColumn
             ->setAllowedTypes('route_param_name', 'string')
             ->setAllowedTypes('route_param_field', 'string')
             ->setAllowedTypes('sortable', 'bool')
+            ->setAllowedTypes('attr', 'array')
         ;
     }
 }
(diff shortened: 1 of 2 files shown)


zuk3975:
Associated. Needed the link to look as button here ->
https://github.com/PrestaShop/PrestaShop/pull/13903/files#diff-14f1ab735df8e9837582c7499938b47dR35 Not sure if it's a good idea though, waiting for more responses on this topic.

sarjon:
I think that creating a specific twig template would make more sense instead of adding UI related configuration. Otherwise, we'll start mixing HTML related stuff (classes, ids & etc.) in PHP code.

tomas862:
have to agree with @sarjon . Closing this issue as it is not needed in my case

Issue #10509, opened 2018-09-19
Migrate "Catalog > Attributes & Features > Add new / edit feature" page


matks:
Part of Symfony migration project
Type: page

Access `/admin-dev/index.php/sell/catalog/features/1/edit`

## Status

Todo:
- [x] Migrate form https://github.com/PrestaShop/PrestaShop/pull/14396/
- [x] Update module ps_facetedsearch as it is hooked here https://github.com/PrestaShop/ps_facetedsearch/pull/133
- [x] URL bug in faceted_search https://github.com/PrestaShop/ps_facetedsearch/pull/145
- [x] Migrate "Cancel" button and "Save" buttons
- [x] Update legacy links
- [x] Release ps_facetedsearch and update Composer to make sure PS 1.7.7 uses updated module version

#### Controller / routing
- [x] All controller actions are migrated and protected with `@AdminSecurity` annotations when eligible and `@DemoRestricted` annotations when eligible
- [x] All links targeting this page are updated: either using _legacy_link or updating the code
- [x] The new SF routes have `_legacy_link` and `_legacy_parameters` provided and explained
- [x] Controller catches Domain Exception and translate them nicely
#### Template
- [x] Twig templates must contain relevant blocks to allow partial extension by modules
- [x] Twig forms must have `form_rest()` and `form_errors()`
- [x] JS being used on the page can be used as enabled extensions and use a PageMap for selectors
- [x] Use `@PrestaShop` instead of `PrestaShopBundle` for template root path aliasing
##### Conventions
- [x] The new controller actions comply with our naming convention
- [x] The new SF routes names and paths comply with our naming convention
- [x] The new Twig template names comply actions with our naming convention
- [x] Created files (controller, JS assets, CSS assets and twig views) must be in the right folder following the convention
- [x] CSS being added must follow theme conventions

#### Core layer
- [ ] Commands and Queries only use native types but internally use VOs
- [ ] All Handlers must have an Interface
##### Conventions
- [ ] The created SF service names comply with our naming convention
- [ ] The created files must follow the folder convention (Domain, Core, Adapter...)

#### Polishing
- [ ] The legacy controller and templates files are gone
- [ ] If PR includes a new concept/mechanism, documentation must be written
- [ ] JS and CSS assets must be compiled and up-to-date
- [ ] License headers must be up-to-date
- [ ] New hooks have been added to the hooks list and the SQL updates
[...]

PR #14396, opened 2019-06-27
Migrate Catalog > Attributes&Features > Features > Add new/edit feature page (blocked by faceted_search update)

rokaszygmantas:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migrates the add new/edit feature page.
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10509
| How to test?  | Compare to legacy page.

:exclamation: new pages are not accessible via interface, but can be accessed via these URLs:
* Create feature page: `/admin-dev/index.php/sell/catalog/features/new`
* Edit feature page: `/admin-dev/index.php/sell/catalog/features/1/edit` (change the ID in URL)

Note:
You should not be able to create/edit features when features are disabled in performance settings page (https://prnt.sc/oa0lh4).

Things that are not supposed to be working in this page, while related pages (features listing, feature values pages) are not migrated:
* "Cancel" button in form.
* "Save" button redirects to the same page now, but might redirect to the listing later on.
* Legacy links are not updated.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14396)


src/PrestaShopBundle/Controller/Admin/Sell/Catalog/FeatureController.php
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * Get translated error messages for feature exceptions
+     *
+     * @return array
+     */
+    private function getErrorMessages()
+    {
+        return [
+            FeatureConstraintException::class => [
+                FeatureConstraintException::EMPTY_NAME => $this->trans('', 'Admin.Notifications.Error'),
```
LouiseBonnard: 🤔

rokaszygmantas: woops :zipper_mouth_face:

rokaszygmantas: Fixed!

LouiseBonnard: Thank you, @rokaszygmantas!

src/Adapter/Feature/CommandHandler/AddFeatureHandler.php
```diff
@@ -0,0 +1,67 @@
...
+        $feature = new Feature();
+
+        $feature->name = $command->getLocalizedNames();
+
+        if (false === $feature->validateFields(false)) {
+            throw new FeatureConstraintException('Invalid feature data');
+        }
+
+        if (false === $feature->validateFieldsLang(false)) {
+            throw new FeatureConstraintException('Invalid feature data', FeatureConstraintException::EMPTY_NAME);
```
matks: Why use `FeatureConstraintException::EMPTY_NAME` when lang fields are not valid ?

src/Core/Form/IdentifiableObject/DataHandler/FeatureFormDataHandler.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create(array $data)
+    {
+        /** @var FeatureId $featureId */
+        $featureId = $this->commandBus->handle(new AddFeatureCommand(
+            $data['name'],
+            $data['shop_association'] ?? []
```
matks, line 60: I see you ❤️ php7.0

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/FeatureController.php
```diff
@@ -0,0 +1,178 @@
...
+     * @param int $featureId
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function editAction($featureId, Request $request)
+    {
+        try {
+            $editableFeature = $this->getQueryBus()->handle(new GetFeatureForEditing((int) $featureId));
+        } catch (Exception $e) {
```
matks, line 98: What about `NotFoundException` if `$featureId` does not exist ?

Maybe add something like

```
            if ($e instanceof FeatureNotFoundException) {
                return $this->redirectToRoute(...);
            }
```

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/FeatureController.php
```diff
@@ -0,0 +1,178 @@
...
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Controller responsible for "Sell > Catalog > Attributes & Features > Features" page
+ */
+class FeatureController extends FrameworkBundleAdminController
+{
+    /**
+     * Create feature action.
+     *
```
matks, line 45: Missing security annotations on all Actions

src/PrestaShopBundle/Resources/config/services/adapter/shop_feature.yml
```diff
@@ -0,0 +1,25 @@
+services:
+  _defaults:
+    public: true
+
+  prestashop.adapter.combination_feature:
```
matks, line 5:
```suggestion
  prestashop.adapter.combination.feature:
```
?

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Features/Blocks/disabled_feature_warning.html.twig
```diff
@@ -0,0 +1,32 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ *#}
+
+<div class="alert alert-warning" role="alert">
+  <p>
+    {{ 'This feature has been disabled. You can activate it here: %url%.'|trans({
+      '%url%': '<a href="' ~ path('admin_performance') ~ '#optional_features">' ~ 'Performance'|trans({}, 'Admin.Global') ~ '</a>'
```
matks: I'm not a fan of all this concatenation happening in the twig file, looks a bit like black magic
Maybe we should do this on controller side, or create a Twig extension for it ?

tests/Integration/Behaviour/Features/Context/Domain/FeatureFeatureContext.php
```diff
@@ -0,0 +1,162 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Feature\Command\AddFeatureCommand;
+use PrestaShop\PrestaShop\Core\Domain\Feature\Command\EditFeatureCommand;
+use PrestaShop\PrestaShop\Core\Domain\Feature\Exception\FeatureConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Feature\Query\GetFeatureForEditing;
+use PrestaShop\PrestaShop\Core\Domain\Feature\QueryResult\EditableFeature;
+use PrestaShop\PrestaShop\Core\Domain\Feature\ValueObject\FeatureId;
+use RuntimeException;
+use Tests\Integration\Behaviour\Features\Context\SharedStorage;
+
+class FeatureFeatureContext extends AbstractDomainFeatureContext
```
matks, line 42: ❤️

tests/Integration/Behaviour/Features/Scenario/Feature/feature_management.feature
```diff
@@ -0,0 +1,28 @@
...
+    Given shop "shop1" with name "test_shop" exists
+
+  Scenario: Create product feature
+    When I create product feature "feature1" with specified properties:
+      | name | My feature |
+    Then product feature "feature1" name should be "My feature"
+
+  Scenario: Update product feature name
+    Given product feature with id "1" exists
+    When I update product feature with id "1" name in default language to "My great feature"
```
matks: `with id "1" name` I dont understand 😅 

I think `name` is not supposed to be here ?

tests/Integration/Behaviour/Features/Scenario/Feature/feature_management.feature
```diff
@@ -0,0 +1,28 @@
...
+    Then product feature with id "1" name in default language should be "My great feature"
+
+  Scenario: Updating product feature with empty name should not be allowed
```
matks, line 21: 👍

matks (changes requested):

src/Core/Form/IdentifiableObject/DataHandler/FeatureFormDataHandler.php
rokaszygmantas: why not though :stuck_out_tongue: php 5.6 will be dropped for 1.7.7 right? :smile:

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/FeatureController.php
rokaszygmantas: whoops totally forgot that

src/PrestaShopBundle/Resources/config/services/adapter/shop_feature.yml
rokaszygmantas: this is not a new service so it would break BC. I've just moved these services from `src/PrestaShopBundle/Resources/config/services/adapter/feature.yml` to `src/PrestaShopBundle/Resources/config/services/adapter/shop_feature.yml` because I needed `feature.yml` name :slightly_smiling_face:

tests/Integration/Behaviour/Features/Scenario/Feature/feature_management.feature
rokaszygmantas: I update the name field of feature with id 1 :)

rokaszygmantas: I'll rephrase it because it really sounds confusing now when you said it :smile:

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Features/Blocks/disabled_feature_warning.html.twig
rokaszygmantas: I don't think we can make it clean from controller side, as there is this `<a>` tag being replaced, but maybe a twig extension would work here :thinking:

tests/Integration/Behaviour/Features/Context/Domain/FeatureFeatureContext.php
rokaszygmantas: I loved the naming as well :smile:

src/Adapter/Feature/CommandHandler/AddFeatureHandler.php
rokaszygmantas: wrong constant used, will fix

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/FeatureController.php
rokaszygmantas: it needs to redirect in case of any exception, but now I've fixed it to show a specific error message if the feature was not found

src/PrestaShopBundle/Resources/config/services/adapter/shop_feature.yml
PierreRambaud: Can you create an alias, we add this discussion on Gitter few days ago with @sarjon.
It's even better to have:
`prestashop.adapter.feature.SOMETHING`
`Namespace.Context.Type.Name`
wdyt @matks ?

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Features/Blocks/disabled_feature_warning.html.twig
rokaszygmantas: after thinking about it, does it really make sense to have a twig extension which would simply render an `<a>` link? seems a bit overengineered to me :thinking:

rokaszygmantas: @matks, would this be considered less magical:
```twig
{% set href = path('admin_performance') ~ '#optional_features' %}
{% set linkText = 'Performance'|trans({}, 'Admin.Global') %}

<div class="alert alert-warning" role="alert">
  <p>
    {{ 'This feature has been disabled. You can activate it here: %url%.'|trans({
      '%url%': "<a href=\"#{href}\">#{linkText}</a>"
    }, 'Admin.Catalog.Notification')|raw }}
  </p>
</div>
```
:male_detective:

src/Core/Form/IdentifiableObject/DataHandler/FeatureFormDataHandler.php
matks: Yes, it was just a comment, not a bad thing 👍 😄

src/PrestaShopBundle/Resources/config/services/adapter/shop_feature.yml
matks: Yes, we need to avoid as much as possible BC 😞we can only have some if there is a really good reason behind. Here, keeping the alias is not an big maintainance burden so we can keep it.

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Features/Blocks/disabled_feature_warning.html.twig
matks: @rokaszygmantas Yes ! It's so much better this way 😄

src/PrestaShopBundle/Resources/config/services/adapter/shop_feature.yml
rokaszygmantas: but I didn't change the service name in this PR at all (I just changed the yaml file name), do we really need to introduce a new alias? @matks

matks: Indeed this is another issue that will be solved in another PR.
The issue is that name "prestashop.adapter.combination_feature" is wrong and we should fix the name *and* keep the previous name available as an alias to avoid BC

matks (dismissed):

sarahdib:
Hello @rokaszygmantas 
There is a design problem in create and edit attribute page : 
image

image

rokaszygmantas:
@sarahdib thanks for feedback.

Interestingly, this design problem comes from the faceted search module as both "URL" and "Meta title" inputs are coming from that module. 

I doubt that this can be fixed from the core :thinking: @matks any insights?

EDIT: related code from `ps_facetedsearch` that breaks the style: https://github.com/PrestaShop/ps_facetedsearch/blob/master/views/templates/hook/feature_form.tpl#L25

IMO the form fields in the module should be rendered using prestashop's form theme instead of custom template, that would fix the issue.

sarahdib:
@rokaszygmantas @matks
Weird because this problem wasn't in legacy page. 
The migration mustn't broke this page. and if we have this problem with factedsearch we can have it with other native module.

What do you think ?

matks:
> @rokaszygmantas @matks
> Weird because this problem wasn't in legacy page.

The faceted search module injects some fields in this page and the fields are powered by Smarty and "built" to fit in the legacy design. The page design has involved through the migration (Bootstrap 3 -> Bootstrap 4, updated design, Twig instead of Smarty) so previous design looks now weird

matks:
Answer from @eternoendless : the Faceted Search module will need a new major version which will be compatible with 1.7.7

matks:
@sarahdib Can you ignore the bad display issue, that will be solved in the module, and validate the other features ?

sarahdib:
@matks Ok I check

@rokaszygmantas  I found an error with add and edit feature.

https://drive.google.com/file/d/1QK8Q8Cdib0JXpiIkKQ5O5N2O94lnBXQN/view

- When we change the language for name field it's not apply to URL and meta
- When we change the language for URL or Meta field it's not apply to name

PierreRambaud:
> Answer from @eternoendless : the Faceted Search module will need a new major version which will be compatible with 1.7.7

I'm really not in favor of making this module only compatible with the latest version since the community use and wait for updates from this module.

Ping @colinegin @marionf

marionf:
Are there other solutions ?

PierreRambaud:
Yes we can create templates specific to this version an higher.

matks:
> I'm really not in favor of making this module only compatible with the latest version since the community use and wait for updates from this module.

I understand but this is the deal with BO migration to Symfony: we replace Smarty with Twig, Bootstrap 3 with Bootstrap 4, legacy framework by Symfony. This is obviously a breaking change that requires modules too update.

PierreRambaud:
I'm sure I can find a way to make it work with both version of Bootstrap 🤔

matks:
> I'm sure I can find a way to make it work with both version of Bootstrap 🤔

I trust you but this will be another layer of complexity and provides an example for other module developers. I dont think this is what we want to promote 😐 

@eternoendless we need your opinion here

eternoendless:
> and provides an example for other module developers.

Actually, module developers who sell on the marketplace need to account for the widest compatibility range as possible, so this could actually be a good opportunity to set up an example: let's adapt faceted search, extract two classes that implement the same behavior, one that implements it for PS < 1.7.7 and another for PS >= 1.7.7, then use one or another depending on the context. Wdyt?

matks:
> > and provides an example for other module developers.
> 
> Actually, module developers who sell on the marketplace need account for the widest compatibility range as possible, so this could actually be a good opportunity to set up an example: let's adapt faceted search, extract two classes that implement the same behavior, one that implements it for PS < 1.7.7 and another for PS >= 1.7.7, then use one or another depending on the context. Wdyt?

If you're OK with the extra work needed here, I'm OK too 👍 
I think if the code is correct, we can ask @rokaszygmantas to do the update (or @PierreRambaud wants to keep the topic 😄 ?

PierreRambaud:
> > > and provides an example for other module developers.
> > 
> > 
> > Actually, module developers who sell on the marketplace need account for the widest compatibility range as possible, so this could actually be a good opportunity to set up an example: let's adapt faceted search, extract two classes that implement the same behavior, one that implements it for PS < 1.7.7 and another for PS >= 1.7.7, then use one or another depending on the context. Wdyt?
> 
> If you're OK with the extra work needed here, I'm OK too 
> I think if the code is correct, we can ask @rokaszygmantas to do the update (or @PierreRambaud wants to keep the topic  ?

If he's bored, he can do it. Otherwise I'll be happy to help you :)

rokaszygmantas:
@PierreRambaud @matks  I'm not bored - got at least 3 PRs to work on at any given time :smile:

rokaszygmantas:
@matks so as I understand, I can rebase this PR and it will be good to go?

matks (dismissed):

matks:
@sarahdib The 2 issues you have reported are related to the ps_facetedsearch module that is hooked into this page. Do you accept we merge this PR *but* this page is not considered "fully migrated" until the ps_facetedsearch has been updated for 1.7.7 ?

sarahdib:
I'm not very confortable with merging this PR with this bug. 
I understand it's coming from the module faceted search but I prefer to have the fix from faceted search to test this PR just to be sure we don't add some regression.

matks:
🚫 This PR is blocked because faceted search module needs to be updated to comply with new BO theme

colinegin:
ping @marionf

marionf:
Guys, we need to talk about it to find a way to not maintain 2 versions of facetedsearch & avoid to have a new version of facetedsearch compatible only from 1.7.7

matks:
> Guys, we need to talk about it to find a way to not maintain 2 versions of facetedsearch & avoid to have a new version of facetedsearch compatible only from 1.7.7

There will be no issue 😄the cool thing is that *the new migrated pages use different hooks than the legacy page, and remove the old hooks*.

So if the module does register itself on both the legacy hooks,
- on versions < 1.7.7 it will use the old hooks, new hooks will be ignored as they do not exist
- on versions >= 1.7.7, it will use the new hooks, old hooks will be ignored as they are removed

So it's just a regular upgrade 😃

EDIT: there will be an issue is a module uses a hook that have different behavior, regarding whether related page is legacy or symfony. but this should never happen

matks:
@rokaszygmantas Do you feel like updating faceted_search ?

rokaszygmantas:
@matks I can update the faceted search module, but it will take some time because I will have a very limited time in September. So if it's not urgent - I could do it :slightly_smiling_face:

matks:
https://github.com/PrestaShop/ps_facetedsearch/pull/133 is waiting for QA ❤️

sarahdib:
Hello @rokaszygmantas 

Can you resolve the conflicting file ?

I need to test the pr in ps_factedsearch with this one.

Thank you

matks:
https://github.com/PrestaShop/ps_facetedsearch/pull/133 has been approved and merged ❤️

matks:
@sarahdib So do we get a "QA ✅ " now that faceted search is compatible ? I'll make sure 1.7.7 is being shipped with updated faceted search version

matks (approved):

sarahdib:
👍

sarjon:
Thanks @sarahdib for QA, @matks for review and @rokaszygmantas for hard work! :tada:

Issue #10586, opened 2018-09-21
Migrate "Shipping > Carriers > Add new / edit carrier" page


matks:
Part of Symfony migration project
Type: page

# Status

Todo:
- [ ] Create "add carrier" command https://github.com/PrestaShop/PrestaShop/pull/14551
- [ ] At each save make sure there is data modification before for saving a new carrier in the database 
- [ ] https://docs.prestashop-project.org/functional-documentation/functional-documentation/ux-ui/back-office/improve/shipping/carrier
- [ ] https://www.figma.com/file/QlgjdPJHkpTZ8NSgLa4qzr/Carrier-page?type=design&node-id=0%3A1&mode=design&t=VLVfvOOawEK9yXvJ-1 mock up with new feature ideas
- [ ] Migrate form rendering https://github.com/PrestaShop/PrestaShop/pull/14397
- [ ] Improve wording https://github.com/PrestaShop/PrestaShop/issues/10586
- [ ] Fix old bugs (see below)
- [ ] Multistore header: [specs](https://github.com/PrestaShop/prestashop-specs/blob/master/broader-topics/multistorespecialsspecs.md#list-of-pages-for-specific-case-3-the-page-can-be-displayed-only-in-all-shops-context)

Old bugs:
- [ ] #10384
- [x] https://github.com/PrestaShop/PrestaShop/issues/15214

#### Controller / routing
- [ ] All controller actions are migrated and protected with `@AdminSecurity` annotations when eligible and `@DemoRestricted` annotations when eligible
- [ ] All links targeting this page are updated: either using _legacy_link or updating the code
- [ ] The new SF routes have `_legacy_link` and `_legacy_parameters` provided and explained
- [ ] Controller catches Domain Exception and translate them nicely
- [ ] Add/Edit controllers must catch global Exception in order to handle ModuleErrorException
#### Template
- [ ] Twig templates must contain relevant blocks to allow partial extension by modules
- [ ] Twig forms must have `form_rest()` and `form_errors()`
- [ ] JS being used on the page can be used as enabled extensions and use a PageMap for selectors
- [ ] Use `@PrestaShop` instead of `PrestaShopBundle` for template root path aliasing
##### Conventions
- [ ] The new controller actions comply with our naming convention
- [ ] The new SF routes names and paths comply with our naming convention
- [ ] The new Twig template names comply actions with our naming convention
- [ ] Created files (controller, JS assets, CSS assets and twig views) must be in the right folder following the convention
[...]

PR #14397, opened 2019-06-27
[On pause] Migrate carrier create/edit form rendering

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migrate to symfony framework - Improve > Shipping > Carrier create/edit form rendering.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #10586

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14397)


matks:
@zuk3975 Ready for review ?

zuk3975:
> @zuk3975 Ready for review ?

@matks, PR is far from over, but you could guide me through if you see something going in wrong direction. I've left some todo's, that should ease the understanding where I am heading to... So yes, good luck with this one :sweat_smile: . I hope i don't need to rewrite whole js :smile:

admin-dev/themes/new-theme/js/pages/carrier/add-range-handler.js
```diff
@@ -0,0 +1,143 @@
...
+  addRangeColumn() {
+
+    for (let i = 0; i < Object.keys(this.$rows).length; i++) {
+      const $row = $(this.$rows[i]);
+
+      if ($row.hasClass('js-range-from')) {
+
+        // replace range-from index placeholder with actual value and enable input
+        const inputFrom = (this.$rangeFromTemplate.get(0).outerHTML)
+          .replace(/__RANGE_INDEX__/g, this.rangeIndex)
```
matks: I'm not a fan of this 😥 can very easily break

admin-dev/themes/new-theme/js/pages/carrier/add-range-handler.js
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+   * Append button which removes corresponding range column
+   *
+   * @param rangeIndex
+   */
+  appendRangeRemovalButton(rangeIndex) {
+    $(this.rangeRemovingBtnRow).removeClass('d-none');
+    $(this.rangeRemovingBtnRow).append(
+      `<td data-range-index="${rangeIndex}">${$(this.removeRangeBtn).get(0).outerHTML}</td>`,
```
matks: HTML in JS 😐 ?

admin-dev/themes/new-theme/js/pages/carrier/add-range-handler.js
zuk3975: No idea of how else it could be achieved. Its dynamic element, generated on click in js. That RANGE_INDEX indicates basically which column it is for later manipulation.

admin-dev/themes/new-theme/js/pages/carrier/add-range-handler.js
zuk3975: Tried to avoid as much as possible, but once again, idk how to achieve it else way. :disappointed:

ghost:
Hello @zuk3975,

It took us some time to review this page to answer your question about the "Ranges" table but it is good now 👍 

So after we discussed it with @TristanLDD , we agree with your idea in the tab n°2 to hide the price-related fields when the "Free Shipping" option is enabled.
Same about the Ranges table, we agree, if "Free Shipping" is enabled we can hide the first 2 rows (which allow setting range limits) and the right column (which allow setting range prices per zone).

So what you did is good, we take it 👍 

Though, there is just a couple of behaviors we missed in the table as it is in this PR.
I let you have a look at the specs to have the table behaviors: [here](https://github.com/PrestaShop/prestashop-specs/blob/master/core/back-office/Shipping-Carriers-add-edit.md#tab-2-shipping-locations-and-costs) (look at the Ranges item in tab 2)
I also share a  video that shows more behaviors on the current version: [here](https://drive.google.com/file/d/1L_AwIVLQMc6SWR56ibv7EygmXff7HyXf/view?usp=sharing)
This behavior is pretty useful, you can use the "All" field to set prices to the zones you have selected:
ranges_table_behavior


The table has some buggy/weird behaviors and I couldn't list them all in the specs. If you wonder if it should be fixed or kept, make the choice that will allow the user to avoid repetitive tasks and be the most effective when setting ranges 🙂 

Feel free to ping me if not sure!


------
About the other tabs and the logo block. We discussed it with @TristanLDD as well, some work is needed to refresh the page and make it easier to use, but it will be done later and not during this migration, as it needs quite some time to rethink what would be the best user journey for carrier creation.
Feel free if you have question


Thanks

matks:
Hi @zuk3975 I close this for now because we dont work on it. We will reopen it when we start working on this again 😉

Issue #14361, opened 2019-06-25
Classic theme: product with combinations: checkout shows base product name instead of chosen combination name


ProBackup-nl:
**Describe the bug**
Having a web shop with only 1 product, in several combinations, the checkout screen shows the base products description/title and the combination image.

It would be more consistent if next to the combination image, furthermore the base product name, the chosen combination product title would be shown to.

**To Reproduce**
1. Create 1 product with multiple (6 in our case) combinations
2. Give each combination a different set of images (3 in our case)
3. Purchase a combination product and go to checkout, section "Order items"
4. See screen shot

**Screenshot checkout - order items**
![Order items](https://user-images.githubusercontent.com/515451/60100588-51fe9380-975a-11e9-9edd-2d28335ed90c.png)

**Products**
![Products](https://user-images.githubusercontent.com/515451/60100796-c3d6dd00-975a-11e9-8198-9c702ff827b2.png)

**Combinations**
![Combinations](https://user-images.githubusercontent.com/515451/60100818-cc2f1800-975a-11e9-9acf-2f002c3cdb3e.png)

**Additional information**
PrestaShop version: 1.7.5.2
Theme: classic
PHP version: 7.2.18

khouloudbelguith:
Hi @ProBackup-nl,

Thanks for your report.
In the BO => Shop Parameters => Order Settings page => We need to enable the "final summary" to reproduce the issue.
I have the same issue with PS1.7.5.2 & PS1.7.6.0.
![image](https://user-images.githubusercontent.com/16067358/60106280-d3efbc00-975c-11e9-9b47-ba972a8b9121.png)

In the order confirmation is OK.
![image](https://user-images.githubusercontent.com/16067358/60106305-e4a03200-975c-11e9-97eb-46e24338da2a.png)

I’ll add this to the debug roadmap so that it’s fixed. If you have already fixed it on your end or if you think you can do it, please do send us a pull request!
Thanks!

khouloudbelguith:
Related to: https://github.com/PrestaShop/PrestaShop/issues/9644

ProBackup-nl:
@khouloudbelguith In [/classes/order/OrderDetail.php](https://github.com/PrestaShop/PrestaShop/blob/aec97ddb1551fd7d6d56f75549a992d6937caf50/classes/order/OrderDetail.php) I see that `product_name` is defined as 
```
 $product['name'] .
            ((isset($product['attributes']) && $product['attributes'] != null) ?
[...]

PR #14426, opened 2019-06-28
order-final-summary shipping carrier name remove

ProBackup-nl:
When the selected carrier delay contains text, that is most likely the widest column. On the other hand the carrier image is most often the company logo.

**Before screenshot**
Classic theme_ product with combinations_ checkout shows base product name instead of chosen combination name · Issue #14361 · PrestaShop_PrestaShop

When the company logo is displayed, is no need to display the name in the 2nd column, as it is already the alternative text of the image. This change re-uses the first column in the summary to display the selected carrier name when no carrier images is present. By that change the carrier-name column can be removed, and the selected_delivery_option.delay column with can be doubled.

**After screenshot**
order-final-summary after improvement

| Questions     | Answers
| Branch?       | develop
| Description?  | classic theme
| Type?         | improvement
| Category?     | FO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | (optional) 
| How to test?  | In the BO => Shop Parameters => Order Settings page => We need to enable the "final summary" to reproduce the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14426)

### DIFF
--- a/themes/classic/templates/checkout/_partials/order-final-summary.tpl
+++ b/themes/classic/templates/checkout/_partials/order-final-summary.tpl
@@ -70,14 +70,11 @@
               {if $selected_delivery_option.logo}
                 <img src="{$selected_delivery_option.logo}" alt="{$selected_delivery_option.name}">
               {else}
-                &nbsp;
+                <span class="carrier-name">{$selected_delivery_option.name}</span>
               {/if}
             </div>
           </div>
-          <div class="col-md-4">
-            <span class="carrier-name">{$selected_delivery_option.name}</span>
-          </div>
-          <div class="col-md-4">
+          <div class="col-md-8">
             <span class="carrier-delay">{$selected_delivery_option.delay}</span>
           </div>
           <div class="col-md-2">


matks:
Hi @ProBackup-nl !

> When the company logo is displayed, is no need to display the name in the 2nd column, as it is already the alternative text of the image.

This is probably true for most carriers but ... 😞 I think some of them have a very generic logo. Also some modules use the name to indicate a specification, like "Postal service for X zone" or "Postal service in 3 days"

@Matt75 I think you told me about such cases ?

Matt75:
Hi

There is no convention for Carrier name, you can create a default carrier (without module) with a name and a custom image, both can be unrelated. Or you can choose to do not provide a logo... Carrier can be an organization or not.

For example : Logo of a building and carrier name is Pickup at the depot. If you display only logo, its unclear for customer.

Capture d’écran 2019-07-04 à 09 47 36

Capture d’écran 2019-07-04 à 09 51 48


If we apply this change, merchant has to be aware of that and adapt their carrier. I think we are going to have a lot a complaining 😅

But this demonstrate this part can be improved 👍

matks:
> For example : Logo of a building and carrier name is Pickup at the depot. If you display only logo, its unclear for customer.

That was the example I was looking for 👍 

I'm sorry @ProBackup-nl, that is why we cant accept this PR

PR #14428, opened 2019-06-29
Introduce fuzzy search

Tanguy-SALMON:
| Questions     | Answers
| Branch?       | develop
| Description?  | Implementation of the Levenshtein algorithm in the default search provider.
| Type?         | new feature
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #14588 
| How to test?  | Just run a vanilla Presta with the updated search provider, search something. Spelling errors are welcome.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14428)


Tanguy-SALMON:
Hi,
Here, just an implementation of the Levenshtein algorithm.
https://www.php.net/manual/en/function.levenshtein.php

The goal is to find some products with the search functionality. Even if you get some miss spelling in the search input.

That works perfectly with the vanilla Prestashop with default products.

Tooltip_007

For example, you search  "ummi" instead of "hummi", search controller will be able to find the closest word, here "hummingbird", and will display all hummingbird products in the result.

Some precisions about the changes:
-------------------------------------------------
a. In case there are many close words, we take the word with the heaviest word of the list.
b. The algorithm is really fast, only 0.005 - 0.010 second to check 2000 words.
b. The algorithm could find some words, really far for the original.
c. Later in a another commit, we could improvement again the script, for example, by adding the possibility to modify the coefficients of the length of words we want to compare in the BO.
d. works with long queries with many words

Regards,
Lathanao

src/Adapter/Search/SearchProductSearchProvider.php
```diff
@@ -144,4 +163,66 @@ public function runQuery(
...
+    /**
+     * @param ProductSearchContext $context
+     * @param $queryString
+     *
+     * @return string
+     */
+    public static function findClosestWeightestWords(ProductSearchContext $context, $queryString)
+    {
+        $distance = array(); // cache levenshtein distance
+        $closestWords = '';
```
PierreRambaud:
```suggestion
        $closestWords = [];
```

src/Adapter/Search/SearchProductSearchProvider.php
```diff
@@ -144,4 +163,66 @@ public function runQuery(
...
+                }
+
+                if ($distance[$a['word']] != $distance[$b['word']]) {
+                    return $distance[$a['word']] < $distance[$b['word']] ? $a : $b;
+                } else {
+                    return $a['weight'] > $b['weight'] ? $a : $b;
+                }
+            }, array('word' => 'initial', 'weight' => '0'))['word'];
+
+            if (next($queries)) {
```
PierreRambaud: You can remove this condition and `implode(' ', $closestWords);`

src/Adapter/Search/SearchProductSearchProvider.php
```diff
@@ -144,4 +163,66 @@ public function runQuery(
...
+                    FROM `' . _DB_PREFIX_ . 'search_word` sw
+                    LEFT JOIN `' . _DB_PREFIX_ . 'search_index` si ON (sw.`id_word` = si.`id_word`)
+                    WHERE sw.`id_lang` = ' . (int) $context->getIdLang() . '
+                    AND sw.`id_shop` = ' . (int) $context->getIdShop() . '
+                    AND LENGTH(sw.`word`) > ' . $targetLenghtMin . '
+                    AND LENGTH(sw.`word`) < ' . $targetLenghtMax . '
+                    GROUP BY sw.`word`;';
+
+            $selectedWords = Db::getInstance()->executeS($sql);
+            $closestWords .= array_reduce($selectedWords, function ($a, $b) use ($query, &$distance /* Cache */) {
```
PierreRambaud:
```suggestion
            $closestWords []= array_reduce($selectedWords, function ($a, $b) use ($query, &$distance /* Cache */) {
```

src/Adapter/Search/SearchProductSearchProvider.php
```diff
@@ -144,4 +163,66 @@ public function runQuery(
...
+                    $distance[$a['word']] = levenshtein($a['word'], $query);
+                }
+
+                if (!isset($distance[$b['word']])) {
+                    $distance[$b['word']] = levenshtein($b['word'], $query);
+                }
+
+                if ($distance[$a['word']] != $distance[$b['word']]) {
+                    return $distance[$a['word']] < $distance[$b['word']] ? $a : $b;
+                } else {
```
PierreRambaud: useless else statement

src/Adapter/Search/SearchProductSearchProvider.php
```diff
@@ -144,4 +163,66 @@ public function runQuery(
...
+     * @param ProductSearchContext $context
+     * @param $queryString
+     *
+     * @return string
+     */
+    public static function findClosestWeightestWords(ProductSearchContext $context, $queryString)
+    {
+        $distance = array(); // cache levenshtein distance
+        $closestWords = '';
+        $lenghtWordCoefMin = 0.7;
```
PierreRambaud: Transform these variables into const please

src/Adapter/Search/SearchProductSearchProvider.php
```diff
@@ -144,4 +163,66 @@ public function runQuery(
...
+        $lenghtWordCoefMax = 1.5;
+        $MINWORDLEN = (int) Configuration::get('PS_SEARCH_MINWORDLEN');
```
PierreRambaud:
```suggestion
        $searchMinWordLength = (int) Configuration::get('PS_SEARCH_MINWORDLEN');
```

PierreRambaud:
Good idea, wdty @colinegin

src/Adapter/Search/SearchProductSearchProvider.php
```diff
@@ -144,4 +169,61 @@ public function runQuery(
...
+                    FROM `' . _DB_PREFIX_ . 'search_word` sw
+                    LEFT JOIN `' . _DB_PREFIX_ . 'search_index` si ON (sw.`id_word` = si.`id_word`)
+                    WHERE sw.`id_lang` = ' . (int) $context->getIdLang() . '
+                    AND sw.`id_shop` = ' . (int) $context->getIdShop() . '
+                    AND LENGTH(sw.`word`) > ' . $targetLenghtMin . '
+                    AND LENGTH(sw.`word`) < ' . $targetLenghtMax . '
+                    GROUP BY sw.`word`;';
+
+            $selectedWords = Db::getInstance()->executeS($sql);
+            $closestWords []= array_reduce($selectedWords, function ($a, $b) use ($query, &$distance /* Cache */) {
```
PierreRambaud:
```suggestion
            $closestWords [] = array_reduce($selectedWords, function ($a, $b) use ($query, &$distance /* Cache */) {
```

Tanguy-SALMON:
Hey!
Do I have any chance of it being merged?
Do you have something to do more ?
Regards,
Lathanao

src/Adapter/Search/SearchProductSearchProvider.php
```diff
@@ -144,4 +169,61 @@ public function runQuery(
...
+
+    /**
+     * @param ProductSearchContext $context
+     * @param $queryString
+     *
+     * @return string
+     */
+    public static function findClosestWeightestWords(ProductSearchContext $context, $queryString)
+    {
+        $distance = array(); // cache levenshtein distance
```
PierreRambaud:
```suggestion
        $distance = []; // cache levenshtein distance
```

PierreRambaud (dismissed):

PierreRambaud:
@Lathanao Sorry, we're still waiting for PM to get their feedback.
But I add a label for our QA team to begin tests.

matks (dismissed):
Hi, I'm really worried about the SQL cost of this PR for huge catalogs (as reported in https://github.com/PrestaShop/PrestaShop/issues/14766) :/ I'm not sure we can merge it as it is right now.

Can we give some time for @matthieu-rolland to explore the possibilities he suggested in https://github.com/PrestaShop/PrestaShop/issues/14766 to reduce the possible performance stress on small servers ?

Tanguy-SALMON:
Hey !
No problem for that, I understand.

At the beginning, the $lenghtWordCoefMin and $lenghtWordCoefMax was some variable, depending how many word the database currently have.

I made a commit yesterday. You can see the switch. It's probably not the best way, cause the code is a bit dirty.

I will be very glad if I could create functionalities as mentioned in  #14766 and merge after validation.

I check this this week if I can do something better.

LouiseBonnard:
Hey @Lathanao, did you have time to check the requested changes? ;-)

Tanguy-SALMON:
Hi @LouiseBonnard , 
Yes, I saw requested changes. I'm sorry, I was a little busy last week. I try to do something for Sunday evening.
Regards.

LouiseBonnard:
@Lathanao, no worries, we were just wondering about it, thanks for the reply. ;-)

Tanguy-SALMON:
Hi @LouiseBonnard, @PierreRambaud, @matks,

I try to implement the Levenshtein algorithm by MySQL, but when you compare Levenshtein in PHP versus Levenshtein in MySql, you can see that Levenshtein in PHP is about 100 time faster than MySql function. And I don't get enough skills to improve this MySql function.

If needed, I can provide a script to compare.
So, that's why I continued to use Levenshtein in PHP.

=---------------------------------------------------------------------------------------------------------------------

I just push a new solution. So for now:
- I didn't implement any option in admin panel, but I will.
- I don't know how to make a test file, data test, etc... with Prestashop standard. So, for now, It's full of dirty debug. I let this debug cause I think I will more easier to understand what happen when searching for a word.
- I added a lot of comment, but may be some explanations more will be needed.
- I added a system to scale the length a word we want to compare, because, we probably don't want to compare a long search word like "HUMMINGBIRD" with short words like 'Mug' or 'good' or ''Day'. At the opposite, we probably don't want to compare the search word 'Mug' with "HUMMINGBIRD". That's why variable like $coefMax, $targetlenght were introduce in the script.
- To reduce again performance stress on small servers, we can introduce a new coef witch can be modified in BO, either directly by editing MAX_WORDS_IN_TABLE or by adding a new coef to do MAX_WORDS_IN_TABLE * newcoef.

=---------------------------------------------------------------------------------------------------------------------

With a vanilla Prestashop (452 words in ps_serach_word), I suggest you to search, in one search, these strings : 
atmosphere atmas atmasph paccesssries  bird minibird sleeves ee eee eeee eeeeeeee eeeeeeeeeeeeee eeeeeeeeeeeeeeeeeeeeee zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz

On the search page (or browser dev panel -> network -> XHR -> preview), you should to get a debug report like that :

atmosphere
=-------------
Eligible Products 2 : 9 10 11

atmas
=-------------
coefMin : 0.50226
coefMax : 1.99548
targetLenghtMin : 3
targetLenghtMax : 9
Number of word for the levenshtein calculation : 191
Levenshtein search time : 0.0013840198516846
Levenshtein loop: 1
Levenshtein nb search : 1
Levenshtein found: ideas
Eligible Products 2 : 16 17 18

Etc ... every words you look for will get a debug result.

=---------------------------------------------------------------------------------------------------------------------

The database I used contain 49818 words in ps_search _word (4648 products). 
Here one of the first part of the debug :

capterr
=-------------
coefMin : 0.74909
coefMax : 1.50182
targetLenghtMin : 5
targetLenghtMax : 10
Number of word for the levenshtein calculation : 9001
Levenshtein search time : 0.099408864974976
Levenshtein loop: 1
Levenshtein nb search : 1
Levenshtein found: capteur
Eligible Products 2 : 2309 6087 6219 6226 6227 6655 6831 6846 6894 6897 6900 7011 7025 7026 7027 7032 7051 7066 7078 7080 7089 7099 7104 7105 7107 7108 7109 7112 7113 7114 7116 7117 7122 7123 7124 7125 7126 7134 7135 7139 7141 7153

etc...

We can see for comparing 9001 words with Levenshtein algorithm, it's took 0.099 second.
My computer get an old CPU (i5-6200U), my opinion is that search performances are good.

=---------------------------------------------------------------------------------------------------------------------

Depending what you would get as performances, you can play by modify constants  MAX_WORDS_IN_TABLE and COEF_MAX and see what happen.
(COEF_MIN get less interests because that will not improve so much the relevance of results).

If you decrease MAX_WORDS_IN_TABLE, the set of words selected for the levenshtein calculation will decrease. Performance stress will be lower.

If you decrease COEF_MAX, the set of words selected for the levenshtein calculation will decrease. Performance stress will be lower.

-> Of course, performance will be better. But If you decrease the set of words selected to much, results could become irrelevant.

At the opposite, if you increase MAX_WORDS_IN_TABLE and COEF_MAX, performance stress will be highter, and results could become relevant.

So for me, with the current constant values, for 49818 words in ps_search _word (4648 products), I get :
- good relevant search results
- good time execution
- good performance stress.

Regards,
Lathanao

matthieu-rolland:
Thank you very much for the improvements, and the time you put in this PR @Lathanao !
We'll watch it closely asap

matthieu-rolland:
@Lathanao 
Your code seems to work great, and the feature is validated by the product team 🚀 
It has to be throughly tested by the QA team before we can merge though.

But we are not there yet (almost !), here are the things that need to be done before we send this PR to the QA team:

**1/** Remove debugs (thank you by the way, it was really helpful to understand your code and its effects). Also make sure that the ajax search is still working after you remove your debug (the display was broken, probably because of debugs)

**2/** You can clean the part of the code that you qualified as dirty :D (don't worry there will be another code review once you will have made your changes)

**3/** We need to make this optional, with a on / off option in the back office's main shop configuration page. The default would be **on**, according to @colinegin 's suggestion in #14766 .  So when the user disables your feature, it should work **exactly** like it used to work.

**4/** We need behat tests for this feature

For the 3/ and 4/, if you can do it that's great, but if for some reason you're not so sure about how this is done, we can take it from there, it's up to you 👍

Tanguy-SALMON:
I will. I ll try to do it this week-end.

Tanguy-SALMON:
Hi @LouiseBonnard, @PierreRambaud, @matks,

I pushed some new commits. So requirements :

1/ Done
2/ Done
3/ Done
4/ ? (about  behat tests )

For testing, let the PS_SEARCH_LEVENSHTEIN_LOOP constant to 10. But for prod, put it to 2, 3 or 4 should be enough.

If you find some performance problem for huge databases (e.g. > 200000 words), I can improve again the script (e.g. first checking the weight, then calculate the Levenshtein distance (this doesn't work some small database), add minimum weight to be selected, or find another trick).

Ready for testing ?

Tanguy-SALMON:
Ping ! @LouiseBonnard, @PierreRambaud, @matks

matks:
@Lathanao as @matthieu-rolland started working on this topic with you, I'll let him continue on it rather than intruding in it 😉

classes/Search.php
```diff
@@ -944,4 +962,95 @@ public static function getSearchParamFromWord($word)
...
+            if (!self::$coefMin) {
+                //self::$coefMin && self::$coefMax depend of the number of total words in ps_search_word table, need to calculate only for every search
+                self::$coefMin = 0.5 / PS_SEARCH_MAX_WORDS_IN_TABLE * self::$totalWordInSearchWordTable + PS_SEARCH_ABSCISSA_MIN; //y = ax + b
+                self::$coefMax = -1 / PS_SEARCH_MAX_WORDS_IN_TABLE * self::$totalWordInSearchWordTable + PS_SEARCH_ABSCISSA_MAX; //y = ax + b
+            }
+            // self::$targetLenghtMin depends of the length of the $queryString, need to calculate for every word
+            self::$targetLenghtMin = (int) (strlen($queryString) * self::$coefMin);
+            self::$targetLenghtMax = (int) (strlen($queryString) * self::$coefMax);
+
+            if (self::$targetLenghtMin < $searchMinWordLength) {
```
PierreRambaud:
```suggestion
            if (self::$targetLengthMin < $searchMinWordLength) {
```
Typo use Length instead Lenght

classes/Search.php
```diff
@@ -23,7 +23,14 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
+
 define('PS_SEARCH_MAX_WORD_LENGTH', 15);
+define('PS_SEARCH_LEVENSHTEIN_LOOP', 10); /* Max time Levenshtein algo is use by search*/
```
PierreRambaud: These const must be configurable, maybe in adavanced menu. Or in a new Pull request?

Tanguy-SALMON:
@PierreRambaud 

Of course, we didn't talk about this variable again. But for testing, I left quite a few, try 10 times and necessary.

For release in prod, 2,3 or 4 times seems to be pretty as default.

Of sure, I can / I will add a field in BO. I can do it now or in a new PR.

Up to you.

PierreRambaud:
Could be awesome to do it in this PR :open_mouth:

Tanguy-SALMON:
Hi @LouiseBonnard, @PierreRambaud, @matthieu-rolland,

Ping !

Tanguy-SALMON:
Hey @colinegin 
Ping !

matthieu-rolland:
> Hi @LouiseBonnard, @PierreRambaud, @matthieu-rolland,
> 
> Ping !

Thanks for the improvements @Lathanao , we'll review it asap !
Also before we can accept the PR the wording will be checked by @LouiseBonnard 

We are aware this is taking some time, but we are looking forward to implementing this feature.

controllers/admin/AdminSearchConfController.php
```diff
@@ -181,6 +181,49 @@ public function __construct()
                             ),
                         ),
                     ),
+                    'PS_SEARCH_FUZZY' => array(
+                        'title' => $this->trans('Fuzzy Search', array(), 'Admin.Shopparameters.Feature'),
```
LouiseBonnard:
```suggestion
                        'title' => $this->trans('Fuzzy search', array(), 'Admin.Shopparameters.Feature'),
```

controllers/admin/AdminSearchConfController.php
```diff
@@ -181,6 +181,49 @@ public function __construct()
...
+                        'validation' => 'isBool',
+                        'cast' => 'intval',
+                        'type' => 'bool',
+                        'desc' => $this->trans(
+                                'By default, fuzzy search is enabled and spelling error are allowed. To search “bird”, you can to enter “burd”, “bard”, “beerd” etc...',
```
LouiseBonnard:
```suggestion
                                'By default, the fuzzy search is enabled. It means spelling errors are allowed, e.g. you can search for "bird" with words like "burd", "bard" or "beerd".',
```

controllers/admin/AdminSearchConfController.php
```diff
@@ -181,6 +181,49 @@ public function __construct()
...
+                                array(),
+                                'Admin.Shopparameters.Help'
+                            ) . '<br/>' .
+                            $this->trans(
+                                'With this option disabled, to get good result, you need get exact writting contained in the word.',
```
LouiseBonnard:
```suggestion
                                'Disabling this option will require exact spelling for the search to match result.',
```

LouiseBonnard: Is it part of the same tooltip or is it another tooltip?

controllers/admin/AdminSearchConfController.php
```diff
@@ -181,6 +181,49 @@ public function __construct()
...
+                                array(),
+                                'Admin.Shopparameters.Help'
+                            ),
+                        'hint' => array(
+                            $this->trans(
+                                'Enable spelling errors during search.',
```
LouiseBonnard: How about _Enable approximate string matching_? Is it a feature or a help text?

controllers/admin/AdminSearchConfController.php
```diff
@@ -181,6 +181,49 @@ public function __construct()
...
+                            $this->trans(
+                                'Fuzzy search will check if a close word of the search word exist in indexed words in database. This will be very resource-consuming.',
+                                array(),
+                                'Admin.Shopparameters.Help'
+                            ),
+                        ),
+                    ),
+                    'PS_SEARCH_FUZZY_MAX_LOOP' => array(
+                        'title' => $this->trans(
+                            'Maximum words with spelling error allowed by search',
```
LouiseBonnard:
```suggestion
                            'Maximum approximate strings allowed by search',
```

controllers/admin/AdminSearchConfController.php
```diff
@@ -181,6 +181,49 @@ public function __construct()
...
+                                'Admin.Shopparameters.Help'
+                            ),
+                        'hint' => array(
+                            $this->trans(
+                                'Enable spelling errors during search.',
+                                array(),
+                                'Admin.Shopparameters.Help'
+                            ),
+                            $this->trans(
+                                'Fuzzy search will check if a close word of the search word exist in indexed words in database. This will be very resource-consuming.',
```
LouiseBonnard: Do we need this wording again? It seems we are repeating ourselves since we already explained it in the first tooltip.

controllers/admin/AdminSearchConfController.php
```diff
@@ -181,6 +181,49 @@ public function __construct()
...
+                        ),
+                    ),
+                    'PS_SEARCH_FUZZY_MAX_LOOP' => array(
+                        'title' => $this->trans(
+                            'Maximum words with spelling error allowed by search',
+                            array(),
+                            'Admin.Shopparameters.Feature'
+                        ),
+                        'hint' => $this->trans(
+                            'A fuzzy search is very resource-consuming. More words you have to solve, more a search could take time.',
```
LouiseBonnard:
```suggestion
                            'Note that this option is time-consuming: the more you check, the longer it searches.',
```

LouiseBonnard: Is it time-consuming or resource-consuming? I think that there is a slight difference between the two.
🤔

controllers/admin/AdminSearchConfController.php
Tanguy-SALMON: Done

controllers/admin/AdminSearchConfController.php
Tanguy-SALMON: Done

controllers/admin/AdminSearchConfController.php
Tanguy-SALMON: Here, the problem is that we need something really explicite. The user have to understand that this 2 options are linked. Please see attachment : 
image

What about : Maximum approximate strings allowed by fuzzy search ? 
(I commit this string and, waiting for your review)

And the right expression is "approximate strings matching". If we use it, without the word "matching", I don't know if we have to write : 
[...]approximate strings [...]
or
[...]approximated strings [...]

controllers/admin/AdminSearchConfController.php
Tanguy-SALMON: Done

controllers/admin/AdminSearchConfController.php
Tanguy-SALMON: The problem here is we have to explain technical constraint. For us, it's clear. But for a newbie or a lambda user, it will be totally confusing, whatever we write... So up to you !

matthieu-rolland: @LouiseBonnard 
I would choose **resource consuming**, because the time it will take for a given query also depends on the shop's infrastructure. 
Also **resource consuming** enforces the idea that it's not only about the time you'll wait for the result to come, but also about the stability of your shop.

Tanguy-SALMON: Done !

controllers/admin/AdminSearchConfController.php
LouiseBonnard: It suits me fine, thanks @Lathanao! Let's keep it as is, with 'approximate strings' and not 'approximated strings'. ;-)

controllers/admin/AdminSearchConfController.php
LouiseBonnard: Would it be a feature or a help text?

controllers/admin/AdminSearchConfController.php
LouiseBonnard: Nice, thank you, it's perfectly clear now. What about _Note that this option is resource-consuming: the more you search, the longer it takes._ instead? I think it is even more explicit.

Tanguy-SALMON: Done

controllers/admin/AdminSearchConfController.php
Tanguy-SALMON: 1. I forgot to reply, but already done.
2. Both

classes/Search.php
```diff
@@ -23,7 +23,12 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
-define('PS_SEARCH_MAX_WORD_LENGTH', 30);
+define('PS_SEARCH_MAX_WORD_LENGTH', 15);
```
PierreRambaud: Can you make all of this constants configurable in the BO?

classes/Search.php
```diff
@@ -229,22 +240,36 @@ public static function find(
             return false;
         }
 
+        $eligible_products2 = array();
```
PierreRambaud: Prefer shortarray syntax :)
```suggestion
        $eligible_products2 = [];
```

classes/Search.php
```diff
@@ -229,22 +240,36 @@ public static function find(
...
+                $sql = 'SELECT DISTINCT si.id_product
+                            FROM ' . _DB_PREFIX_ . 'search_word sw
+                            LEFT JOIN ' . _DB_PREFIX_ . 'search_index si ON sw.id_word = si.id_word
+                            WHERE sw.id_lang = ' . (int) $id_lang . '
+                                AND sw.id_shop = ' . $context->shop->id . '
+                                AND sw.word LIKE';
+
+                while (!($result = $db->executeS($sql . "'" . $sql_param_search . "';", true, false))) {
+                    if (!Configuration::get('PS_SEARCH_FUZZY') ||
+                        $fuzzyLoop++ > Configuration::get('PS_SEARCH_FUZZY_MAX_LOOP') ||
```
PierreRambaud: Can we use a variable defined before loop for all `Configuration::get` statements?

classes/Search.php
```diff
@@ -944,4 +963,95 @@ public static function getSearchParamFromWord($word)
...
+            *  With actual contante values, we have :
+            *  Linear function for $coefMin : a = 0.5 / 100000, b = 0.5
+            *  Linear function for $coefMax : a = -1 / 100000, b = 2
+            *  Result :
+            *  500 words id DB give coefMin : 0.5025, coefMax : 1.995
+            *  20,000 words id DB give $coefMin : 0.6, $coefMax : 1.8
+            *  40,000 words id DB give $coefMin : 0.7, $coefMax : 1.6
+            *  60,000 words id DB give $coefMin : 0.8, $coefMax : 1.4
+            *  80,000 words id DB give $coefMin : 0.9, $coefMax : 1.2
+            *  100,000 words id DB give $coefMin : 1, $coefMax : 1*/
```
PierreRambaud, line 1060: I love your comments :heart:

PierreRambaud:
Really love what you did. Only for the constants defined at the top of the file, everything looks ok :+1:

classes/Search.php
Tanguy-SALMON: I love Presta !

classes/Search.php
Tanguy-SALMON: Done

classes/Search.php
Tanguy-SALMON: Done

classes/Search.php
Tanguy-SALMON: Done

PierreRambaud: I still see all `define` at the top of the file https://github.com/PrestaShop/PrestaShop/pull/14428/files#diff-4d2b8c521b121179fd17f512b9ffaebcR26 and doesn't see the Configuration values here https://github.com/PrestaShop/PrestaShop/pull/14428/files#diff-3f46cad4e139058c5f0e6f05504eab7aR143

Tanguy-SALMON: Right now, I put only PS_SEARCH_MAX_WORD_LENGTH as configurable variable in BO.

Would you like to have PS_SEARCH_MAX_WORDS_IN_TABLE, PS_SEARCH_ORDINATE_MIN, PS_SEARCH_ORDINATE_MAX, PS_SEARCH_ABSCISSA_MIN and PS_SEARCH_ABSCISSA_MAX as configurable values too?

I fact, I don't really advise to change this values.
To remember, in any case, the goal to change PS_SEARCH_ORDINATE_XXX or PS_SEARCH_ABSCISSA_XXX is to vary the target length of world we want to compare.

Changing :
PS_SEARCH_ORDINATE_XXX: could broke the auto scale. Only variations with +/- 0.1-0.2 could get sense for not braking anything. 
Because, if linear functions are to flat, the auto-scale is useless.
If linear functions are too leaning, the range could tends too quickly to coef max and coef min = 1. And the chance to find a matching word on small DB reduce.

A best way to do vary the slope of the straight lines softly, without braking anything, is to vary PS_SEARCH_MAX_WORDS_IN_TABLE. The result will be same, and safer.

For PS_SEARCH_ABSCISSA_XXX:
For small database, I think it's large enough (have to do more test about matching result on small DB). And for large DB, influence will not be so obvious.
Again there, if we want to decrease or increase the target length, vary PS_SEARCH_MAX_WORDS_IN_TABLE will be more conformable, and the result will be safer.

--------------------------------------------------------------------------------------------------------------

In my opinion:

Is making configurable PS_SEARCH_ORDINATE_XXX or PS_SEARCH_ABSCISSA_XXX a solution to manage stress server on small server ?

1. The first best way is to reduce server stress is to vary the "Maximum approximate strings allowed by fuzzy search" already implemented in BO.

2. The second one could be to adjust PS_SEARCH_MAX_WORDS_IN_TABLE in BO (not implemented yet).

--------------------------------------------------------------------------------------------------------------

What yould you like I implement ? PS_SEARCH_MAX_WORDS_IN_TABLE, PS_SEARCH_ORDINATE_MIN, PS_SEARCH_ORDINATE_MAX, PS_SEARCH_ABSCISSA_MIN and PS_SEARCH_ABSCISSA_MAX together ? Only PS_SEARCH_MAX_WORDS_IN_TABLE ?

Up to you.

PierreRambaud: If values are not going to change, you should use class constante instead of `define` and it's ok for me :wink:

Tanguy-SALMON: Done

classes/Search.php
```diff
@@ -229,22 +239,38 @@ public static function find(
...
+                            FROM ' . _DB_PREFIX_ . 'search_word sw
+                            LEFT JOIN ' . _DB_PREFIX_ . 'search_index si ON sw.id_word = si.id_word
+                            WHERE sw.id_lang = ' . (int) $id_lang . '
+                                AND sw.id_shop = ' . $context->shop->id . '
+                                AND sw.word LIKE';
+
+                while (!($result = $db->executeS($sql . "'" . $sql_param_search . "';", true, false))) {
+                    if (!$psFuzzySearch ||
+                        $fuzzyLoop++ > $fuzzyMaxLoop ||
+                        !($sql_param_search = self::findClosestWeightestWord($context, $word))) {
```
PierreRambaud: Could you use static instead of self because of legacy class :sweat_smile:

Tanguy-SALMON: I think it's ok.

Tanguy-SALMON: I didn't set any visibility for constantes. Should have ?

PierreRambaud: It's about `self::findClosestWeightestWord`, => ̀static::findClosestWeightestWord`

`self` refers to the same class in which the `new` keyword is actually written.
`static` refers to whatever class in the hierarchy you called the method on

PierreRambaud: and also every const / static calls :sweat_smile:

Tanguy-SALMON: Ok, I think I get it. 
As you can guess, I'm not a master for this side of code specifications. Thx for the training ! 
Done.

matthieu-rolland (dismissed):
Thank you very much @Lathanao for your hard work, and your patience ! This feature is very appreciated.

Thank you @PierreRambaud for your insightful feedbacks !

Robin-Fischer-PS:
Hi @colinegin @Lathanao  ! I've some questions on this PR :

- When the search finds the exact string, it does not search further for approximate string : This functionality is OK, but when an offline product matches the exact searched string, the search will still stops, and will not display anything.
Wouldn't it be better to also display the approximate results in this case?

- Random searches, such as "fzefnzjnzef", "SaveMeImBeingHeldHostageInQA", or even "これはテストです"  will almost always give results (obviously, results without link with the search, since it's totally random). Is this okay ? It doesn't penalize the merchant or the user, but some merchants could find this weird...

- When "Maximum word length (in characters)" is empty in BO, search results always show all products of the catalog.

Thanks :smile:

colinegin:
Hello @Robin-Fischer-PS ,

Thanks for seeing these issues ! 

- Issue related to offline product --> I agree with you, this can be a problem. @Lathanao do you think you can fix this issue ? 

- Issue with random searches --> i'm not sure to understand how the BO field added by this PR works. Isn't it supposed to limit results if some random searches are done ?

- Maximum word length --> i think this field should be mandatory, with a preset length. What do you think @Lathanao ?

Last issue, we found it quite hard to understand and use with @Robin-Fischer-PS , is the wording ok for you @LouiseBonnard ?

LouiseBonnard:
If this feature introduces new behaviors or new keys to understand how search options will now work on PrestaShop, I'm volunteering to write an article about it for Build. Perhaps with @Lathanao, it could be nice to have a community member presenting his work. :-)

Tanguy-SALMON:
Hey

@Robin-Fischer-PS
Thx for the review!

@colinegin 
1. Right now, the script can't check if a searched word owned by a product, has the product online or not. This is because the script is on the top of the main query (the big SQL query wich bring products depending the search query string).

Need a little more time to think about the best solution to solve this problem. But I think it's possible to find a small tricky code to solve that.

2. For now, there is no limitation for the distance (how many character differ from both words). That's explain why there are some really weird search result. Just need to add one more constant value, and add a conditional statement. I do it quickly.

3. Maximum word length : Do you mean that 'Maximum word length' should be a constant, and not a configurable value in BO
 - It's clear it'a again one setup more in BO. And not really useful to change it because 15 is a good value for Latin alphabets.
 - but 15 is a little short for Asian languages (but Search::satinization() get - hyper - bad results for Asian languages. So.. don't care so much)
 
-> For me, shop owner shouldn't change this value. IMO, the best, is to get a constant, and not a configurable value in BO. 

@LouiseBonnard 
No problem! I don't know how you want to proceed. But, e.g. if you have already the page structure, you can give me chapter titles, then, I can make the text, and explain technical stuff. Just let me know what's should I do.

colinegin:
Hello @Lathanao ,

Thanks for your quick reply !

1. This would be great :) let us know if you need any help from our dev team.

2. This would be perfect.

3. I am sorry my sentence was not really clear. What i meant was that I would recommend it to make it a mandatory field to avoid people from leaving it blank, creating weird behaviour. There should be a preset value (15 if this sounds enough for you), and if the merchant chooses to remove it, he has to add a new one. Would that be ok for you ?

Tanguy-SALMON:
3. Ok, so now, for you it's blank. Because you just make an update of the code from the branch. And the value is not already set. 

But, for new shop installation, the value will be automatically set, and the field will be set a 15 by default, so, not blank.

You can see the installation default value in install-dev/data/xml/configuration.xml 

Yes, if the merchant chooses to remove it, he has to add a new one.
If the merchant choose to let blank, have to check the behavior....

install-dev/data/xml/configuration.xml
```diff
@@ -143,6 +140,18 @@
...
+  </configuration>
+  <configuration id="PS_SEARCH_FUZZY" name="PS_SEARCH_FUZZY">
+    <value>1</value>
+  </configuration>
+  <configuration id="PS_SEARCH_FUZZY_MAX_LOOP" name="PS_SEARCH_FUZZY_MAX_LOOP">
+    <value>4</value>
+  </configuration>
+  <configuration id="PS_SEARCH_MAX_WORD_LENGTH" name="PS_SEARCH_MAX_WORD_LENGTH">
+    <value>15</value>
+  </configuration>
```
Tanguy-SALMON, line 154: Here is where the field will be set during the installation.

Tanguy-SALMON:
Hey

To remember, benchmark given is for a database with : 
ps_product : 4,648 rows
ps_search_word : 46,895 rows
ps_search_index : 607,825 rows

----------------------------------------------------------------------------------

1. Issue related to offline product --> I agree with you, this can be a problem. @Lathanao do you think you can fix this issue ?

I tried some solutions, e.g. by adding some conditions in SQL query to be sure to select only product active = 1, visibility = "search / both", and indexed = 1.
But whatever how I created queries, query executions is about time 3 time slower.
Here some results (10655 words selected): 

SQL 0 time : 0.068441867828369
SQL 1 time : 0.070348024368286
SQL 2 time : 0.21622896194458 => with 1 join more and 3 filters
SQL 3 time : 0.19993901252747 => with 1 join more and 3 filters

The problem is, to make a fuzzy search, we need to run this queries 2 time (line 254, and 1035).
These 2 queries are not exactly same, but are really similar.
And they need to be similar for the well working of the script.

So, I can't add filters (join + active = 1, visibility = "search / both", and indexed = 1) only for one.
So, we will be 3 time lower, 2 time (for both queries), during the search.
In my case, it's look like to add 0.3s for every fuzzy search... not really efficient IMO.

There is probably a better way to achieve this issue, but right now, I don't get it :(

------------------------------------------------------------------------------

Issue with random searches --> i'm not sure to understand how the BO field added by this PR works. Isn't it supposed to limit results if some random searches are done ?

I added PS_DISTANCE_MAX constante.
I did some test, so for getting a good idea of what could be an distance :

levenshtein('ad6160509eg', '6160509eg')         = 2
levenshtein('elaticty', 'latching')                         = 6
levenshtein('captar', 'capteur')                           = 2
levenshtein('sdqsdqsdqsd', 'sdqfsqsdfqsdfsdf') = 6
levenshtein('transmission30', 'transmi')             = 7
levenshtein('throughputsuppo', 'throught')         = 7
levenshtein('sdqsdqsdqsd', 'これはテストです')   = 24

So, I set to 8 the PS_DISTANCE_MAX constante, it's look to be a good value in my case.
What's your opinion about the value ?
Should I do PS_DISTANCE_MAX a configurable value in BO?

------------------------------------------------------------------------------

I change a little the logic in array_reduce(). 
Method findClosestWeightestWord() is about 5% faster (only :().

colinegin:
Hello @Lathanao ,

A few comments about your messages :

> Yes, if the merchant chooses to remove it, he has to add a new one.
If the merchant choose to let blank, have to check the behavior....
==> we actually tested by letting it blank, and this was giving us weird results. This is why I recommend to make it mandatory.

> There is probably a better way to achieve this issue, but right now, I don't get it :(
==> @matthieu-rolland do you have any idea which could help @Lathanao ?

> So, I set to 8 the PS_DISTANCE_MAX constante, it's look to be a good value in my case.
What's your opinion about the value ?
Should I do PS_DISTANCE_MAX a configurable value in BO?
==> If you put 8 as a value, does it mean that only products whose name have at least 8 letters similar to your search will be found ? 
To be honest, i'm not sure about the right answer here. Do you think of any situation where it could be interesting to change it ?

matthieu-rolland:
@Lathanao , about point **1/**

I've been looking what could be done, and as seen with @colinegin, your fix is welcome despite the loss of SQL performance, because the feature freeze is coming very soon and we don't have much time to do anything else. This feature will be presented as BETA/To be improved.

@Lathanao , **Do you think you could add the joins so that only active products are fetched?**

PS: I have an idea for a more performant solution,  but it will take some refactoring work on the **find** method, so I will make an issue about this, and we'll do it for a future patch if my idea is validated by the core team. I'll ping you in the issue so that you can give us your opinion ;)

Tanguy-SALMON:
Will do quickly, I hope this weekend.

colinegin:
Thanks a lot @Lathanao !

colinegin:
Hello @Lathanao , hope you are going well !
Did you have any time by chance to work on your PR and add @matthieu-rolland's recommendation ? 

Thanks a lot !

Tanguy-SALMON:
Hey, Sorry to be late. 

I'll do the PR today.

Quickly, as I get both queries there (simple one and with filters), a let you them there.
So, with filters, it's still 3 time slower. But i don't understand why? I don't add so many parameters more. IMO, it should not be 3 time slower. 
-> May be I wrong somewhere. 

Here simple query -> time exec : 0.113 s :
SELECT null as levenshtein, -SUM(weight) as weight, sw.`word`
FROM `ps_search_word` sw
LEFT JOIN `ps_search_index` si ON (sw.`id_word` = si.`id_word`)
WHERE sw.`id_lang` = 1
AND sw.`id_shop` =1
AND LENGTH(sw.`word`) >= 5
AND LENGTH(sw.`word`) <= 15
GROUP BY sw.`word`;

With filter -> time exec : 0.366 s :
SELECT null as levenshtein, -SUM(weight) as weight, sw.`word`
FROM `ps_search_word` sw
LEFT JOIN `ps_search_index` si ON (sw.`id_word` = si.`id_word`)
LEFT JOIN `ps_product_shop` product_shop ON (si.`id_product` = product_shop.`id_product`)
WHERE sw.`id_lang` = 1
AND sw.`id_shop` =1
AND LENGTH(sw.`word`) >= 5
AND LENGTH(sw.`word`) <= 15
AND product_shop.`active` = 1
AND product_shop.`visibility` IN ("both", "search")
AND product_shop.indexed = 1
GROUP BY sw.`word`;

Tanguy-SALMON:
IMO, a greater solution to clean result of deactivated product could be to add an hook when changing product status (visibility / status etc... ), and clean the ps_search_world from all words referencing a deactivated product. 

Pro
-> Because finally, words should not be in this table.
-> This will light a bit tables
-> We don't have to refactor Search::find()

Con
-> Have to make hook (what else?) for delete words / update words in search tables.

colinegin:
Thanks @Lathanao for your feedback !
Wdyt of this suggestion @matthieu-rolland ?

matthieu-rolland:
> Thanks @Lathanao for your feedback !
> Wdyt of this suggestion @matthieu-rolland ?

Hi @colinegin and @Lathanao 

It seems like a good idea, but the feature freeze is very soon so I suggest we send back this PR to QA as is (with @Lathanao 's last modifications), and we'll make it better in the next patch release.

One last thing, is this file needed? If not it would be nice to delete it:
`tools/leveithen.php`

I send this PR in QA, because we need to go fast, I'll approve the PR once the file mentioned above will be deleted (except if you tell me it's necessary for some reason I don't get?)

Thank you !

Tanguy-SALMON:
Hey,
I deleted the file.
Ok, algo could be improve alter.

matthieu-rolland (dismissed):
Thank you very much @Lathanao for your precious work on this pull request !

Also thank you to everyone who participated on this PR 👍

Robin-Fischer-PS:
Hi :)

I've some feedbacks with this new version of the PR : 

- One really small typo on the new Warning message : "github" should be written "GitHub"

- The deactivation of the Fuzzy Search in the BO does not seem to work fine, especially when "Maximum word length (in characters)" is set low (1 or 2 for example).

- Finally, it was decided that "Maximum word length (in characters)" field should not be empty, this has not been done.

Thanks !

matthieu-rolland:
@Robin-Fischer-PS @colinegin 

1/  I'll take care of it
2/ I think I found out why, I'm working on it
3/ I just fixed it

matthieu-rolland:
@Robin-Fischer-PS 

I fixed your 3 feedbacks, could you try again?

@colinegin 

The issue was that the "maximum word length" used to be defined in the code, at *30*.
Now people can change it for fuzzy search, but it was still used for normal search.
So I kept a default value for when we are not in fuzzy search.

I think it would be nice in the configuration page to state that this option is only for fuzzy search, now we have ` Maximum word length (in characters)`. People might think that this value will be used for search without fuzzy mode too.

EDIT:  I added this message below the max word length configuration field:

**This parameter will only be used if fuzzy search is activated, the lower the value, the more tolerant your research will be**

To be validated by @LouiseBonnard ... I'm not so sure about the formulation.

LouiseBonnard:
> No problem! I don't know how you want to proceed. But, e.g. if you have already the page structure, you can give me chapter titles, then, I can make the text, and explain technical stuff. Just let me know what's should I do.

Hi @Lathanao, can you send me an email so that we can start writing the article?

PierreRambaud:
Pull request rebased with a little refactoring :)

Hlavtox:
@PierreRambaud @matthieu-rolland Guys, please, before you finally merge it, test that it doesn't find half of the shop for every query... 😄 

In PS's current form, if you search "red car", it finds all things that are red and all cars. (Unless you go back in time a bit, as I proposed in my PR: #16023 😎)

Hlavtox:
And one more thing. @Lathanao 

Guys, can you please test this scenario?
Make a product called **Prestashop S-100**.
Does the algorithm find it, if you search **Prestashop S 100** or **Prestashop S100**?

Otherwise, this will be super awesome!!!!!!!

PierreRambaud:
@Hlavtox I'm on it to integrate your commits to this pull request.

PierreRambaud:
@Hlavtox Wasn't your original commit, but I make sure you're the author of the changes I've done to make it work :wink: 

Plus, I tried with a specific name like `S layer` or `Slayer` and everything is ok.
Watch out if you use `S-100` or `S 100` because the search system is also looking for words in description and attributes values, so you'll see result for the attribute `size 100` as example :)

It's something we can improve later, with a on/off button to search only on the title, or on everything. But this pull request must be merge :+1:

matthieu-rolland (approved):
Thank you @Hlavtox , @PierreRambaud for making this PR even better ! 👍 
And of course @Lathanao for the amazing work ! 🚀 

And last but not least @LouiseBonnard , @matks , @colinegin and the whole PM team for your participation 👍 

There will be optimizations in the near future, but this is an awesome start :) Let's pray for the QA validation now 😅

Robin-Fischer-PS:
Hi @Lathanao , @matthieu-rolland !

This time, it all seems to be good for me, so it's QA :heavy_check_mark: !

Thanks for the hard work on this PR everyone :)

rdy4ever:
I would've exepected this search to show "hummingbird"
image
.. or this to show "the adventure.."
image

But is seems common misspellings don't work

Hlavtox:
@rdy4ever @Lathanao

Good point, could this be integrated? 😎

Tanguy-SALMON:
Hey @rdy4ever @Hlavtox 

Does your Prestashop works on the right branch?
Following your comments, I was worry. I just reinstalled a fresh Prestashop [Lathanao:patch-3]

I get :
image
image

With a fresh Prestashop, no need to set anything, Fuzzy search works fine without any setup in BO.
So, for me, everything is fine.

Did you check if your branch get last commits?

Tanguy-SALMON:
Just test PrestaShop:develop
Works fine too :
image

Hlavtox:
@Lathanao I didn't test it yet, just commented on @rdy4ever :-D

If it works, awesome. I will put it on our store after public release and I will instruct employees to record any use cases that could happen and would not find what they want. Maybe there will be some room for improvement. But so far, it looks perfect. 🚀

rdy4ever:
Hi @Lathanao 
Yes, I have merged upstream/develop into my local develop, just before testing it. I didn't perform any database updates though, is there anything I need to change in the database?

rdy4ever:
@Lathanao My bad. I only put 5 at maximum word length in config.  With 30 it works great. Thanks!

PR #14444, opened 2019-07-01
Check if selected address belongs to user

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | The selected address during checkout process must be owned by the connected user. 
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Try to change the input id in your developper console.<br>image

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14444)

### DIFF
--- a/classes/checkout/CheckoutAddressesStep.php
+++ b/classes/checkout/CheckoutAddressesStep.php
@@ -157,7 +157,9 @@ class CheckoutAddressesStepCore extends AbstractCheckoutStep
                     'Shop.Notifications.Error'
                 );
             }
-        } elseif (isset($requestParams['confirm-addresses'])) {
+        }
+
+        if (isset($requestParams['confirm-addresses'])) {
             if (isset($requestParams['id_address_delivery'])) {
                 $id_address = $requestParams['id_address_delivery'];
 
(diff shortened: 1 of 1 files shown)


classes/checkout/CheckoutAddressesStep.php
```diff
@@ -71,25 +71,6 @@ public function handleRequest(array $requestParams = array())
             }
         }
 
-        if (isset($requestParams['id_address_delivery'])) {
```
Quetzacoalt91, line 74: I don't know all the whole process work, but removing these tests from here may break an existing feature don't you think?

PierreRambaud: It's not remove but move to the bottom. You will able to set a delivery adress even if you're not in confirm context.

Quetzacoalt91: How is that even possible if you're moving these checks inside a whole `} elseif (isset($requestParams['confirm-addresses'])) {`?

PierreRambaud: Because these variables exist only in `confirm-addresses` context

matks (approved):

matks:
Thank you @PierreRambaud

PR #14458, opened 2019-07-02
Avoid people to destroy their database on schema update command

mickaelandrieu:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Since 1.7.1 our foreign keys are invalid and Doctrine ORM is not fine with it. For now, let's disable Foreign Keys support so people won't break their shop when using `php bin/console doctrine:schema:update --force`
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | None **yet**, but critical ones later
| How to test?  | Simple, using the Command line interface, do "php bin/console doctrine:schema:update" before and after this contribution. Ping @jolelievre for review :)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14458)

### DIFF
--- a/src/PrestaShopBundle/Resources/config/services/bundle/services.yml
+++ b/src/PrestaShopBundle/Resources/config/services/bundle/services.yml
@@ -107,5 +107,5 @@ services:
         class: PrestaShop\PrestaShop\Core\Product\ProductCsvExporter
         arguments: ['@translator', '@prestashop.core.admin.data_provider.product_interface']
 
-    prestashop.core.ignore_foreign_keys_platform:
-        class: PrestaShopBundle\Doctrine\IgnoreForeignKeysPlatform
+    prestashop.doctrine.ignore_foreign_keys_platform:
+      class: PrestaShopBundle\Doctrine\IgnoreForeignKeysPlatform
--- a/src/PrestaShopBundle/Resources/config/services/bundle/services.yml
+++ b/src/PrestaShopBundle/Resources/config/services/bundle/services.yml
@@ -106,3 +106,6 @@ services:
     prestashop.core.product.csv_exporter:
         class: PrestaShop\PrestaShop\Core\Product\ProductCsvExporter
         arguments: ['@translator', '@prestashop.core.admin.data_provider.product_interface']
+
+    prestashop.core.ignore_foreign_keys_platform:
+        class: PrestaShopBundle\Doctrine\IgnoreForeignKeysPlatform
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -84,7 +84,7 @@ doctrine:
                 password: "%database_password%"
                 server_version: 5.1
                 charset:  UTF8
-                platform_service: prestashop.core.ignore_foreign_keys_platform
+                platform_service: prestashop.doctrine.ignore_foreign_keys_platform
                 # if using pdo_sqlite as your database driver:
                 #   1. add the path in parameters.yml
                 #     e.g. database_path: "%kernel.root_dir%/data/data.db3"
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -84,6 +84,7 @@ doctrine:
                 password: "%database_password%"
                 server_version: 5.1
                 charset:  UTF8
+                platform_service: prestashop.core.ignore_foreign_keys_platform
                 # if using pdo_sqlite as your database driver:
                 #   1. add the path in parameters.yml
                 #     e.g. database_path: "%kernel.root_dir%/data/data.db3"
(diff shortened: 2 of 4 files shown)


eternoendless:
While this could avoid the need for the `prestashop:schema:update-without-foreign` command, it would effectively prevent foreign keys on module entities. Maybe we could use some kind of annotation on our entities?

mickaelandrieu:
I'm not sure we have the power to do that with an annotation, it's been a very long time I haven't go deep into Doctrine, though...

As I told you, this can help if we want to avoid big drama with people using Doctrine when we will release the 1.7.6 but the only right way to fix it is to figure out why we can't apply these foreign keys: this may be a time consuming and difficult issue to solve, but in a long way it's for the good: everyone loves data integrity, right :laughing: ?

src/PrestaShopBundle/Resources/config/services/bundle/services.yml
```diff
@@ -106,3 +106,6 @@ services:
     prestashop.core.product.csv_exporter:
         class: PrestaShop\PrestaShop\Core\Product\ProductCsvExporter
         arguments: ['@translator', '@prestashop.core.admin.data_provider.product_interface']
+
+    prestashop.doctrine.ignore_foreign_keys_platform:
```
jolelievre, line 110: Why do you need to define the service here as well, if it's already defined in `config/services/common.yml` ?

Quetzacoalt91:
If the purpose of this PR is to disable the foreign key generation with `php bin/console doctrine:schema:update`, you should probably deprecate the home-made command and make it as an alias of the doctrine one?

src/PrestaShopBundle/Resources/config/services/bundle/services.yml
mickaelandrieu: is config/services/common.yml file loaded in the Main Symfony container?

mickaelandrieu:
> you should probably deprecate the home-made command and make it as an alias of the doctrine one?

it sounds risky to change the installation process 1 week before the stable release, I don't think we should do it "now" :)

> LegacyTests\Integration\PrestaShopBundle\Controller\Api\StockManagementControllerTest::itShouldReturnValidResponseWhenRequestingBulkStockEdition
PHPUnit_Framework_Exception: Segmentation fault (core dumped)

Any idea why this assertion is broken?

Quetzacoalt91:
Oh I didn't see this PR targets `1.7.6.x` branch, but my comment could be valid for `develop` :)

src/PrestaShopBundle/Resources/config/services/bundle/services.yml
jolelievre: To be checked, but it should be yes this is how the CLDR service is accessible in both front and admin

jolelievre: Oh no my bad it's not exactly this, the container loads `admin/service*.yml` and `front/service*.yml` depending on the context And then the common.yml is included in each folder
So to sum it up, yes it should be loaded!

jolelievre:
> > PHPUnit_Framework_Exception: Segmentation fault (core dumped)
> 
> Any idea why this assertion is broken?

Travis has lots of bug on PHP5.6 environment, recurrent segmentation fault crashes
You need to restart the job and hope for it to go through the end (I just restarted it)

About the service definition, it would appear the services present in `config/services` are only used by the "legacy" `ContainerBuilder` not in the Symfony context. Maybe this could be nice to add this support later. But for now I think you either need to define it in both environments.

There is also some service definitions which are shared/included in both environments such as `- { resource: ../../src/PrestaShopBundle/Resources/config/services/core/cldr.yml }` (see in `config/services/common.yml`)

As your definition is in the bundle it's maybe not the best place, but you could create a new `src/PrestaShopBundle/Resources/config/services/common.yml` file which would also be included in front

eternoendless:
I think it's too late to integrate this fix in 1.7.6 but it should absolutely be done in develop.

eternoendless (changes requested):
Needs to be rebased into develop

mickaelandrieu:
> I think it's too late to integrate this fix in 1.7.6 but it should absolutely be done in develop.

if it's for 1.7.7, shitstorm is already on us: let's fix these weird FK issues instead ;)

Closed

matks:
> I think it's too late to integrate this fix in 1.7.6 but it should absolutely be done in develop.

@eternoendless I'm willing to reopen this PR and finish it but can you explain what you mean by

> While this could avoid the need for the `prestashop:schema:update-without-foreign` command, it would effectively prevent foreign keys on module entities. Maybe we could use some kind of annotation on our entities?

I dont see how an annotation can solve this issue.

Issue #11631, opened 2018-12-06
Bug : Add to cart success without stock 


francoispeyret:
**Description**
1. Go to a page product with stock.
2. In BO go to this product and update the quantity to 0.
3. in the page page produit (without page refresh) add the product to cart.
4. Ouch...

The product was added but it's a success for shoppingcart but it'snot in real life.

![capture du 2018-12-06 10-37-27](https://user-images.githubusercontent.com/15829643/49575660-8851fe00-f943-11e8-8435-c6305eb3fe09.png)

khouloudbelguith:
Hi @francoispeyret,

Thanks for your report.
If you click on "Proceed too checkout" button, you will be redirected to shopping cart page with a red alert "The item The best is yet to come' Framed poster in your cart is no longer available in this quantity. You cannot proceed with your order until the quantity is adjusted." & the "Proceed too checkout" is disabled so you cannot proceed your checkout.
![image](https://user-images.githubusercontent.com/16067358/49576964-b258ef80-f946-11e8-93d9-c60738965a72.png)
Recently, we have made some changes to fix those issues: 
https://github.com/PrestaShop/PrestaShop/pull/8752
https://github.com/PrestaShop/PrestaShop/pull/8774
https://github.com/PrestaShop/PrestaShop/pull/8607
Thanks to check & feedback.

francoispeyret:
Right but the proceed was stop before that in 1.6..

![image](https://user-images.githubusercontent.com/15829643/49577784-7757bb80-f948-11e8-9956-fdfdab879d9f.png)

The modal-dialog hasn't got every information of current product in this case.

no bug :
![image](https://user-images.githubusercontent.com/15829643/49578035-15e41c80-f949-11e8-9c82-cb85267f121d.png)

buged : 
![image](https://user-images.githubusercontent.com/15829643/49575660-8851fe00-f943-11e8-8435-c6305eb3fe09.png)

khouloudbelguith:
@francoispeyret, thanks for your feedback.
Yes, I tried with PS1.6 & we are stopped before the redirection to shopping cart
https://drive.google.com/file/d/1pHl6_aO4IbVLQsMdDfm3c57_8aHM3I5S/view
But with PS1.7.5.0, we are redirected to the shopping cart & the alert will be displayed.
@marionf, what do you think?
Thanks!

marionf:
We could add a check to verify the quantity when we click on the "add to cart" button on the product page or in the quickview
[...]

PR #14489, opened 2019-07-04
Disallow ordering a product that was transformed from regular into product with combinations

### LINKED ISSUE #13497: It is possible to order product transformed from regular into product with combinations
It is possible to order perviously created simple product after its transformation to a product with combinations. (so, the simple product should be unavailable for order)

Here is an exemple:

- create a simple product
- add it to cart
- modify product and switch to product with combinations mode, save
- go to your cart and buy the product.

This product is not available in the shop anymore but it is still available in **previously created shopping carts.** 
**The main problem** is not old shoppings cards but the **"order again" button** in user account. It is also possible to order simple product instead of the product with combination.

_Additional info_ : if I increase the quantity of the simple product (that should be normally unavailable) it will add the second line with product with default combination.

PrestaShop : 1.7.4.2

202-ecommerce:
| Questions     | Answers
| Branch?       | develop
| Description?  | It is possible to order perviously created simple product after its transformation to a product with combinations. (so, the simple product should be unavailable for order)
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #13497
| How to test?  | Please see original issue #13497. Please also check "standard" checkout usecases

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14489)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -4828,7 +4828,8 @@ class CartCore extends ObjectModel
     }
 
     /**
-     * Are all products of the Cart still exist?
+     * Are all products of the Cart still available in the current state ? They might have been converted to another
+     * type of product since then
      *
      * @return bool False if not all products in the cart still exist
      */
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -4832,7 +4832,7 @@ class CartCore extends ObjectModel
      *
      * @return bool False if not all products in the cart still exist
      */
-    public function isAllProductsStillExist()
+    public function checkAllProductsAreStillAvailableInThisState()
     {
         foreach ($this->getProducts(false, false, null, false) as $product) {
             $currentProduct = new Product();
@@ -4842,6 +4842,8 @@ class CartCore extends ObjectModel
                 return false;
             }
         }
+
+        return true;
     }
 
     /**
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -4281,6 +4281,14 @@ class CartCore extends ObjectModel
         $success = true;
         $products = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS('SELECT * FROM `' . _DB_PREFIX_ . 'cart_product` WHERE `id_cart` = ' . (int) $this->id);
 
+        foreach ($products as $key => $product) {
+            $currentProduct = new Product();
+            $currentProduct->hydrate($product);
+            if ($currentProduct->hasAttributes() && $product['id_product_attribute'] === 0) {
+                unset($products[$key]);
+            }
+        }
+        
         $orderId = Order::getIdByCartId((int) $this->id);
         $product_gift = array();
         if ($orderId) {
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -4831,7 +4831,7 @@ class CartCore extends ObjectModel
      * Are all products of the Cart still available in the current state ? They might have been converted to another
      * type of product since then
      *
-     * @return bool False if not all products in the cart still exist
+     * @return bool False if one of the products from the cart has been changed into a new type of product
      */
     public function checkAllProductsAreStillAvailableInThisState()
     {
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -4281,14 +4281,6 @@ class CartCore extends ObjectModel
         $success = true;
         $products = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS('SELECT * FROM `' . _DB_PREFIX_ . 'cart_product` WHERE `id_cart` = ' . (int) $this->id);
 
-        foreach ($products as $key => $product) {
-            $currentProduct = new Product();
-            $currentProduct->hydrate($product);
-            if ($currentProduct->hasAttributes() && $product['id_product_attribute'] === 0) {
-                unset($products[$key]);
-            }
-        }
-        
         $orderId = Order::getIdByCartId((int) $this->id);
         $product_gift = array();
         if ($orderId) {
@@ -4833,6 +4825,23 @@ class CartCore extends ObjectModel
             FROM `' . _DB_PREFIX_ . 'customer` cu
             LEFT JOIN `' . _DB_PREFIX_ . 'cart` ca ON (ca.`id_customer` = cu.`id_customer`)
             WHERE ca.`id_cart` = ' . (int) $id_cart);
+    }
+
+    /**
+     * Are all products of the Cart still exist?
+     *
+     * @return bool False if not all products in the cart still exist
+     */
+    public function isAllProductsStillExist()
+    {
+        foreach ($this->getProducts(false, false, null, false) as $product) {
+            $currentProduct = new Product();
+            $currentProduct->hydrate($product);
+
+            if ($currentProduct->hasAttributes() && $product['id_product_attribute'] === '0') {
+                return false;
+            }
+        }
     }
 
     /**
--- a/classes/checkout/CheckoutPaymentStep.php
+++ b/classes/checkout/CheckoutPaymentStep.php
@@ -43,9 +43,11 @@ class CheckoutPaymentStepCore extends AbstractCheckoutStep
 
     public function handleRequest(array $requestParams = array())
     {
-        $allProductsInStock = $this->getCheckoutSession()->getCart()->isAllProductsInStock();
+        $cart = $this->getCheckoutSession()->getCart();
+        $allProductsInStock = $cart->isAllProductsInStock();
+        $allProductsExist = $cart->isAllProductsStillExist();
 
-        if ($allProductsInStock !== true) {
+        if ($allProductsInStock !== true || $allProductsExist !== true) {
             $cartShowUrl = $this->context->link->getPageLink(
                 'cart',
                 null,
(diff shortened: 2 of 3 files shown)


matks:
@202-ecommerce thanks for all the contributions, looks like you're on fire 🔥

👏 🎉

tlhuillier:
Could we add a warning on the cart page, to tell the user that a product was not available anymore ?
Since it may cause frustration, the user can think he really is ordering the same orders.

matks:
@202-ecommerce I'm sorry, can you explain what is happening in this PR ? I see that you unset the products without a combination ID, but this happens in `Cart::duplicate()`. I dont see how it can be related to the issue.

Is `Cart::duplicate()` used in "Order again" button, so you make sure that when the "Order again" button is used, the products that were changed are not copied ?

202-ecommerce:
@matks 
We find method call in AdminOrderController but it's the same issue.
Yes Cart::duplicate() is used in "Order again" button.

@tlhuillier We look for a solution together ?

classes/Cart.php
```diff
@@ -4281,6 +4281,14 @@ public function duplicate()
         $success = true;
         $products = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS('SELECT * FROM `' . _DB_PREFIX_ . 'cart_product` WHERE `id_cart` = ' . (int) $this->id);
 
+        foreach ($products as $key => $product) {
+            $currentProduct = new Product();
+            $currentProduct->hydrate($product);
+            if ($currentProduct->hasAttributes() && $product['id_product_attribute'] === 0) {
+                unset($products[$key]);
```
jolelievre: Hi @202-ecommerce 
I'm worried about this solution, because:
- it will result in potentially empty carts, I wonder if this can cause other issues later
- (considering the initial issue) it doesn't fix the problem when you have a Cart containing a former simple product, the you cal still finish your order

I think we'd rather need a validation of the cart during the checkout process, or even in the cart preview, which prevents from ordering this kind of product.

202-ecommerce: @jolelievre we think it's not a problem to continue without the unavailable product because we consider it's not good to stop reorder with one product unavailable.

In this case @tlhuillier propose to add a error notification. We suggest a red alert like #11631 ?
"You cannot reorder the product xxx because it is not available any more."

So is this solution better for you ?

jolelievre: Hi @202-ecommerce,
indeed I think it's better to notify the user that something is wrong. The error message you mention in the linked PR seems a good solution. So I see two possibilities:
- remove the unavailable product and display an error notification to notify the user that one product was removed
- leave the unavailable product but display a warning notifying that the cart can't be checkout because the user needs to select a combination or he removes this product

The second solution seems more user friendly to me because:
- the user can still access the product and select the appropriate combination (if you remove it he needs to take time finding the right product again..)
- what happens if you remove all the products from the cart (including when there's only one), you end up with an empty cart which can be confusing

jolelievre: In case we agree on leaving the unavailable products but with an error, of course the checkout button should be blocked, and the checkout tunnel inaccessible (redirect to the cart)
I realize this solution is longer to develop but I think it's a better experience for the user

202-ecommerce:
We have resolve a problem during reorder a cart with a product who not exist anymore without attribut. We have choose your suggested solution. The product is add but we can't checkout the cart. We throw a message to tell to customer to delete and add the product with attribute.

We haven't manage translation because we want a confirmation the text is all right for you. Can you confirm that plz ? By the way, we don't know how to find id_translation for the translation file 'app/Resources/translations/default/ShopNotificationsError.xlf'

classes/Cart.php
```diff
@@ -4827,6 +4827,23 @@ public static function isGuestCartByCartId($id_cart)
             WHERE ca.`id_cart` = ' . (int) $id_cart);
     }
 
+    /**
+     * Are all products of the Cart still exist?
+     *
+     * @return bool False if not all products in the cart still exist
+     */
+    public function isAllProductsStillExist()
```
matks:
```suggestion
    public function checkAllProductsAreStillAvailableInThisState()
```
If I get it right, this function checks that the products in the cart have not been transformed from a standard product to a product-with-combinations 😉so we dont check they "still exist" but rather they "still are available in this state"

classes/Cart.php
```diff
@@ -4827,6 +4827,23 @@ public static function isGuestCartByCartId($id_cart)
             WHERE ca.`id_cart` = ' . (int) $id_cart);
     }
 
+    /**
+     * Are all products of the Cart still exist?
+     *
+     * @return bool False if not all products in the cart still exist
```
matks:
```suggestion
     * @return bool False if one of the products from the cart has been changed into a new type of product
```

classes/Cart.php
```diff
@@ -4827,6 +4827,23 @@ public static function isGuestCartByCartId($id_cart)
             WHERE ca.`id_cart` = ' . (int) $id_cart);
     }
 
+    /**
+     * Are all products of the Cart still exist?
```
matks:
```suggestion
     * Are all products of the Cart still available in the current state ? They might have been converted to another
     * type of product since then
```

classes/Cart.php
```diff
@@ -4827,6 +4827,23 @@ public static function isGuestCartByCartId($id_cart)
...
+    public function isAllProductsStillExist()
+    {
+        foreach ($this->getProducts(false, false, null, false) as $product) {
+            $currentProduct = new Product();
+            $currentProduct->hydrate($product);
+
+            if ($currentProduct->hasAttributes() && $product['id_product_attribute'] === '0') {
+                return false;
+            }
+        }
```
matks, line 4845:
```suggestion
        }
        
        return true;
```

classes/checkout/CheckoutPaymentStep.php
```diff
@@ -43,9 +43,11 @@ public function __construct(
 
     public function handleRequest(array $requestParams = array())
     {
-        $allProductsInStock = $this->getCheckoutSession()->getCart()->isAllProductsInStock();
+        $cart = $this->getCheckoutSession()->getCart();
+        $allProductsInStock = $cart->isAllProductsInStock();
+        $allProductsExist = $cart->isAllProductsStillExist();
```
matks:
```suggestion
        $allProductsAreStillAvailableInThisState = $cart-> checkAllProductsAreStillAvailableInThisState();
```

controllers/front/CartController.php
```diff
@@ -597,11 +597,27 @@ protected function shouldAvailabilityErrorBeRaised($product, $qtyToCheck)
...
     {
+        $products = $this->context->cart->getProducts();
+
+        foreach ($products as $product) {
+            $currentProduct = new Product();
+            $currentProduct->hydrate($product);
+
+            if ($currentProduct->hasAttributes() && $product['id_product_attribute'] === '0') {
+                return $this->trans(
+                    'The item %product% in your cart is no longer available in this version. You cannot proceed with your order until you delete or replace your product by a new one with attribute.',
```
matks: ping @LouiseBonnard >e need a wording for the usecase where:
- someone orders a product without combinations
- then one day after the merchant converts this product into a product-with-combinations
- the user wants to reorder it, so the product has changed now, he cannot reorder it as it was

matks (changes requested):

matks:
@202-ecommerce Sorry for the late review, idea is good, I added some naming suggestions and required Content team to check the wording 😉

classes/Cart.php
jolelievre: Maybe something a bit shorter? like `checkProductsTypeChanged` ?

controllers/front/CartController.php
jolelievre: I think we can have a more simple/generic text and simply say "You cannot proceed with your order until you remove this product from your cart"
I'm afraid most customers won't understand the logic of "attributes"

jolelievre: Or maybe we should talk about "combinations" which would be more user friendly

LouiseBonnard: It seems to be the opposite actually: talking about combinations doesn't make sense for most merchants, that's why we must introduce the notion of attributes before so that the (not-so-logic) link can be made between combinations and attributes. :-)

Here, I would like to keep the same construction as the wording about adjusting quantities but turn it into a more positive tone. So I suggest: _The item %product% in your cart is now a product with attributes. Please delete it and choose one of its combinations to proceed with your order._ in _Shop.Notifications.Error_ as well.

202-ecommerce: Thank you @LouiseBonnard Commit done.

classes/checkout/CheckoutPaymentStep.php
202-ecommerce: Thank you @matks Commit done

classes/Cart.php
202-ecommerce: Thank you @matks Commit done

classes/Cart.php
202-ecommerce: commit with @matks version

controllers/front/CartController.php
202-ecommerce: @LouiseBonnard It looks like there is a conflict https://github.com/PrestaShop/PrestaShop/pull/14489/conflicts 
In fact the message is used for more generic case of available product.
What do you suggest ?

LouiseBonnard: @202-ecommerce, what do you mean by "more generic case of available product"?

202-ecommerce: @LouiseBonnard My mistake. I've change the bad string ! New commit with the right translate had been submit

matks:
Code is now perfect ✅ let's go for QA 😄

matks (approved):

matks:
Thank you @202-ecommerce !

PR #14501, opened 2019-07-05
Fix conditions in function deleteProduct

MDW-Thomas:
| Questions     | Answers
| Branch?       | develop
| Description?  | In the cart class in the function deleteProduct() we set in params default value for id_product_attribute = 0 but after in the code we test if $id_product_attribute != null.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | In the function deleteProduct of the class cart.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14501)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1712,7 +1712,7 @@ class CartCore extends ObjectModel
                 WHERE `id_cart` = ' . (int) $this->id . '
                 AND `id_product` = ' . (int) $id_product . '
                 AND `id_customization` = ' . (int) $id_customization .
-                ($id_product_attribute != null ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
+                ($id_product_attribute != 0 ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
             );
         }
 
@@ -1723,7 +1723,7 @@ class CartCore extends ObjectModel
                 SET `quantity` = ' . (int) $preservedGifts[(int) $id_product . '-' . (int) $id_product_attribute] . '
                 WHERE `id_cart` = ' . (int) $this->id . '
                 AND `id_product` = ' . (int) $id_product .
-                ($id_product_attribute != null ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
+                ($id_product_attribute != 0 ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
             );
         }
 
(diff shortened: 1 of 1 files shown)


classes/Cart.php
```diff
@@ -1712,7 +1712,7 @@ public function deleteProduct(
                 WHERE `id_cart` = ' . (int) $this->id . '
                 AND `id_product` = ' . (int) $id_product . '
                 AND `id_customization` = ' . (int) $id_customization .
-                ($id_product_attribute != null ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
```
PierreRambaud, line 1715: No need this change since `0 == null`

```php
var_dump(null == 0);  => bool(true)
```

classes/Cart.php
```diff
@@ -1723,7 +1723,7 @@ public function deleteProduct(
                 SET `quantity` = ' . (int) $preservedGifts[(int) $id_product . '-' . (int) $id_product_attribute] . '
                 WHERE `id_cart` = ' . (int) $this->id . '
                 AND `id_product` = ' . (int) $id_product .
-                ($id_product_attribute != null ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
+                ($id_product_attribute != 0 ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
```
PierreRambaud, line 1726: same here

classes/Cart.php
```diff
@@ -1732,7 +1732,7 @@ public function deleteProduct(
         DELETE FROM `' . _DB_PREFIX_ . 'cart_product`
         WHERE `id_product` = ' . (int) $id_product . '
         AND `id_customization` = ' . (int) $id_customization .
-            (null !== $id_product_attribute ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '') . '
+            (0 !== $id_product_attribute ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '') . '
```
PierreRambaud, line 1735: Wdyt @PrestaShop/prestashop-core-developers it looks legit :thinking:

classes/Cart.php
MDW-Thomas: Should we change to this ? : 

$id_product_attribute !== 0

PierreRambaud: I'm more in favor of
```suggestion
                ($id_product_attribute !== 0 ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
```

But not sure it will change something for the two affected requests

matks:
Hi @MDW-Thoma, in order to get this PR merged, we need you to handle code review feedback 😉

PierreRambaud:
Hi @MDW-Thomas ,

Thanks!

PR #14529, opened 2019-07-08
Fix position problem on a fresh install

### LINKED ISSUE #14504: Cannot update default category positions
Category data has a field `position` that is not correctly initialized. It's initialized to 0, for all new categories, including default categories.

Consequently if you try to modify categories positions from the Categories listing, the ajax request fails. The UI seems to show it worked, but actually the ajax request failed so if you refresh the page, you can see the categories positions are not modified.

However if you perform 1 successfull positions update, for example moving the very last category created, then all category data items will be successfully fixed with `positions` data being correctly filled.

**To Reproduce**
1. New fresh 1.7.6 PrestaShop install
2. In Back-Office, go on Catalog > Categories
3. Open your inspecter, on the "network" tab to see ajax requests
4. Try to drag-and-drop one category to modify its position. It seems to be working.
5. However you'll see in the inspector "network" tab that the ajax request failed and returned a HTTP 500 error page. See below the screenshot of the response.
6. Refresh the page, you can see category positions are not updated.

Capture d’écran 2019-07-05 à 14 51 23

PrestaShop version: 1.7.6 but I think the bug was there before, in all PS 1.7 versions. However it could not be seen. Now that Category page was migrated to Symfony in 1.7.6 the error is more obvious.

PierreRambaud:
| Questions     | Answers
| Branch?       | develop
| Description?  | AddPosition must add the position in category table and updatePosition should test for negative position
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14504 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14529)

### DIFF
--- a/classes/Category.php
+++ b/classes/Category.php
@@ -1766,16 +1766,19 @@ class CategoryCore extends ObjectModel
         }
         // < and > statements rather than BETWEEN operator
         // since BETWEEN is treated differently according to databases
-        $result = (Db::getInstance()->execute('
-            UPDATE `' . _DB_PREFIX_ . 'category` c ' . Shop::addSqlAssociation('category', 'c') . '
-            SET c.`position`= c.`position` ' . ($way ? '- 1' : '+ 1') . ',
-            category_shop.`position`= category_shop.`position` ' . ($way ? '- 1' : '+ 1') . ',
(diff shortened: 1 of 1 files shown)


matthieu-rolland (approved):
LGTM

PR #14532, opened 2019-07-08
Check if index.php file exists before ignoring it

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | When creating annotation driver we need to check if the file exists or the annotation scanning will fail
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | Test the `./bin/console doctrine:mapping:info` with the productcomments module installed (which contains an index.php file) AND the doctrine example module from https://github.com/friends-of-presta/doctrine (which does not contain an index.php file). The entities from both modules need to be scanned correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14532)

### DIFF
--- a/src/PrestaShopBundle/DependencyInjection/Compiler/ModulesDoctrineCompilerPass.php
+++ b/src/PrestaShopBundle/DependencyInjection/Compiler/ModulesDoctrineCompilerPass.php
@@ -111,7 +111,10 @@ class ModulesDoctrineCompilerPass implements CompilerPassInterface
     {
         $reader = new Reference('annotation_reader');
         $driverDefinition = new Definition('Doctrine\ORM\Mapping\Driver\AnnotationDriver', [$reader, [$moduleEntityDirectory]]);
-        $driverDefinition->addMethodCall('addExcludePaths', [[$moduleEntityDirectory . '/index.php']]);
+        $indexFile = $moduleEntityDirectory . '/index.php';
+        if (file_exists($indexFile)) {
+            $driverDefinition->addMethodCall('addExcludePaths', [[$moduleEntityDirectory . '/index.php']]);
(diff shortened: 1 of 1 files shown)


PierreRambaud (approved):

matks:
Restarting Travis...

matks:
Thank you @jolelievre

src/PrestaShopBundle/DependencyInjection/Compiler/ModulesDoctrineCompilerPass.php
```diff
@@ -111,7 +111,10 @@ private function createAnnotationMappingDriver($moduleNamespace, $moduleEntityDi
     {
         $reader = new Reference('annotation_reader');
         $driverDefinition = new Definition('Doctrine\ORM\Mapping\Driver\AnnotationDriver', [$reader, [$moduleEntityDirectory]]);
-        $driverDefinition->addMethodCall('addExcludePaths', [[$moduleEntityDirectory . '/index.php']]);
+        $indexFile = $moduleEntityDirectory . '/index.php';
+        if (file_exists($indexFile)) {
+            $driverDefinition->addMethodCall('addExcludePaths', [[$moduleEntityDirectory . '/index.php']]);
```
MathiasReker, line 116: Why not use `$indexFile` instead of `$moduleEntityDirectory . '/index.php'`?

jolelievre: You're right, that's what I intended to do!! ^^
Thank you

PR #14542, opened 2019-07-09
Introduces async bulk action

tomas862:
| Questions     | Answers
| Branch?       | develop
| Description?  | As in products page right now each bulk action submits partial data - 1 product at the time. By having thousands of products performing cost efficient operation this solution can be good. I added new type and grid javascript extension to handle this case in reusable way for each grid. I also added **step** property which can determine how much data to send to the server in 1 iteration. Next step would be to add loading pop-up animation due to currently it has nothing which can evaluate loading state.
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | partially required for product list
| How to test?  | this PR will only be used in product list so no need for manual tesing for this one

- [ ] docs after merged

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14542)

### DIFF
--- a/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
+++ b/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
@@ -1,0 +1,130 @@
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+const $ = window.$;
+
+export default class AsyncSubmitBulkActionExtension {
+  constructor() {
+    return {
+      extend: grid => this.extend(grid),
+    };
+  }
+
+  /**
+   * Extend grid with bulk action submitting
+   *
+   * @param {Grid} grid
+   */
+  extend(grid) {
+    grid.getContainer().on('click', '.js-bulk-action-submit-async-btn', (event) => {
+      this.submit(event, grid);
+    });
+  }
+
+  submit(event, grid) {
+    const $submitBtn = $(event.currentTarget);
+    const confirmMessage = $submitBtn.data('confirm-message');
+
+    if (typeof confirmMessage !== 'undefined' && confirmMessage.length > 0 && !confirm(confirmMessage)) {
+      return;
+    }
+
+    const $form = $(`#${grid.getId()}_filter_form`);
+    const $checkedInputs = this.getCheckedInputs($form);
+    const ids = this.getCheckedIds($checkedInputs);
+
+    const inputName = this.getInputName($checkedInputs);
+    const chunkSize = parseInt($submitBtn.data('step'), 10);
+    const chunkedIds = this.chunkArray(ids, chunkSize);
+    const submitMethod = $submitBtn.data('form-method');
+    const formUrl = $submitBtn.data('form-url');
+
+    if (!['POST', 'GET', 'DELETE', 'PUT'].includes(submitMethod)) {
+      return;
+    }
+
+    const asyncSubmit = (items, successCallback, errorCallback) => {
+      if (items.length === 0) {
+        return;
+      }
+
+      const firstIdsChunk = items.shift();
+      const data = {};
+      data[inputName] = firstIdsChunk;
+
+      $.ajax({
+        type: submitMethod,
+        url: formUrl,
+        data,
+        dataType: 'json',
+      }).then((response) => {
+        if (!response.success) {
+          errorCallback(response);
+        }
+
+        if (response.success && items.length > 0) {
+          asyncSubmit(items, successCallback, errorCallback);
+        } else {
+          successCallback(response);
+        }
+      }).catch((error) => {
+        const response = error.responseJSON;
+        errorCallback(response);
+      });
+    };
+
+    asyncSubmit(
+      chunkedIds,
+      (response) => {
+        showSuccessMessage(response.message);
+        window.location.reload();
+      },
+      (response) => {
+        showErrorMessage(response.message);
+        window.location.reload();
+      });
+  }
+
+  getCheckedInputs($form) {
+    return $form.find('.js-bulk-action-checkbox:checked');
+  }
+
+  getCheckedIds($inputs) {
+    return $inputs.map((index, el) => parseInt($(el).val(), 10));
+  }
+
+  getInputName($inputs) {
+    return $inputs.attr('name').replace('[]', '');
+  }
+
+  chunkArray(array, chunkSize) {
+    const results = [];
+    while (array.length) {
+      results.push(array.splice(0, chunkSize));
+    }
+
+    return results;
+  }
+}
--- a/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
+++ b/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
@@ -43,6 +43,11 @@ export default class AsyncSubmitBulkActionExtension {
     });
   }
 
+  /**
+   * Collects required data and submits the form.
+   * @param event
+   * @param grid
+   */
   submit(event, grid) {
     const $submitBtn = $(event.currentTarget);
     const confirmMessage = $submitBtn.data('confirm-message');
@@ -65,6 +70,12 @@ export default class AsyncSubmitBulkActionExtension {
       return;
     }
 
+    /**
+     * Async submit action which repeats itself in recursive order.
+     * @param items
+     * @param successCallback
+     * @param errorCallback
+     */
     const asyncSubmit = (items, successCallback, errorCallback) => {
       if (items.length === 0) {
         return;
@@ -107,18 +118,39 @@ export default class AsyncSubmitBulkActionExtension {
       });
   }
 
+  /**
+   * Gets the only checked inputs.
+   * @param $form
+   * @return {*}
+   */
   getCheckedInputs($form) {
     return $form.find('.js-bulk-action-checkbox:checked');
   }
 
+  /**
+   * From checked inputs it returns input values
+   * @param $inputs
+   * @return {*}
+   */
   getCheckedIds($inputs) {
     return $inputs.map((index, el) => parseInt($(el).val(), 10));
   }
 
+  /**
+   * Since bulk action has same name we can return single name from all.
+   * @param $inputs
+   * @return {*}
+   */
   getInputName($inputs) {
     return $inputs.attr('name').replace('[]', '');
   }
 
+  /**
+   * Divides results by chunk size.
+   * @param array
+   * @param chunkSize
+   * @return {Array}
+   */
   chunkArray(array, chunkSize) {
     const results = [];
     while (array.length) {
--- a/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
+++ b/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
@@ -123,7 +123,7 @@ export default class AsyncSubmitBulkActionExtension {
     const chunkSize = parseInt($submitBtn.data('chunkSize'), 10);
     const chunkedIds = this.chunkArray([...ids], chunkSize);
     const submitMethod = $submitBtn.data('formMethod');
-    const formUrl = $submitBtn.data('form-url');
+    const formUrl = $submitBtn.data('formUrl');
 
     if (!['POST', 'GET', 'DELETE', 'PUT'].includes(submitMethod)) {
       return;
--- a/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
+++ b/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
@@ -54,7 +54,6 @@ export default class AsyncSubmitBulkActionExtension {
     const $submitBtn = $(event.currentTarget);
     const confirmMessage = $submitBtn.data('confirm-message');
     const confirmTitle = $submitBtn.data('confirmTitle');
-
     const hasConfirmTitle = confirmTitle !== undefined;
     const hasConfirmationMessage = confirmMessage !== undefined && confirmMessage.length > 0;
     const hasAllConfirmationMessages = hasConfirmationMessage && hasConfirmTitle;
@@ -149,7 +148,7 @@ export default class AsyncSubmitBulkActionExtension {
       confirmButtonLabel,
       closeButtonLabel,
       confirmButtonClass,
-    }, () => this.postForm($submitBtn, grid));
+    }, () => this.submitForm($submitBtn, grid));
 
     modal.show();
   }
--- a/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
+++ b/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
@@ -132,6 +132,9 @@ export default class AsyncSubmitBulkActionExtension {
       const data = {};
       data[inputName] = firstIdsChunk;
 
+      // initialises the start of an ajax call
+      this.progressCallback({}, ids, ids);
+
       $.ajax({
         type: submitMethod,
         url: formUrl,
--- a/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
+++ b/admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
@@ -125,7 +125,7 @@ export default class AsyncSubmitBulkActionExtension {
     const submitMethod = $submitBtn.data('formMethod');
     const formUrl = $submitBtn.data('formUrl');
 
-    if (!['POST', 'GET', 'DELETE', 'PUT'].includes(submitMethod)) {
+    if (!['POST', 'GET', 'DELETE', 'PUT'].includes(submitMethod.toUpperCase())) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


tomas862:
@matks as I found out it is quite a required thing to keep in product list then I added this PR - need early opinion on this one :) I think the next step will be to add optional pop-up for each bulk action - from php type extension you can change pop-up title etc...

matks:
@tomas862 I'm sorry I dont understand the usecase 🤔 

Bulk actions on product listing are: delete, enable, disable, duplicate

Are you worried one of the actions is going to be too long for a single HTTP request to handle, and might timeout ?

tomas862:
in the response to 1st problem mentioned here https://github.com/PrestaShop/PrestaShop/pull/14267#issuecomment-508716498 it seems like I need to keep loading animation for bulk actions - imagine bulk duplicate 1000 products. The server will go down for sure. 

so I think product team wants this functionality of async calls  that remains in product list. So this component will help to solve the issue and make it highly reusable in all lists.

matks:
I understand now, however I'm not sure this is the right answer.

I think that if you have a shop running on a small server, you're supposed to have a small catalog so there is no need for this.
If you have a bigger catalog, and consequently you might need to perform very large/long bulk actions, I expect you to have a server with the right power and configuration to handle it.

Here, you're being kind and you try to avoid timeout issues for the people who have the big catalog _and_ a small server 😄. Or maybe "duplicate" should not be an action available for bulk if it creates such issues.

As this topic is highly debatable, I'll need more opinions @rokaszygmantas @eternoendless @PierreRambaud @jolelievre @sarjon @zuk3975 @Quetzacoalt91

kpodemski:
@matks 

it's not only about catalog size Mathieu, remember that every time product is:
- duplicated
- enabled
- disabled

PrestaShop can perform some actions using Hooks, those actions could be time consuming. (clearing cache in filter modules for example, this can take a lot of time, Advanced Search 4, one of the most popular filter modules creates millions of cache files...)

I think that this POC is a great step forward to make PrestaShop more friendly, both, for small, and big catalogs.

tomas862:
good points as well @kpodemski - especially product duplication takes huge amount of time.

sarjon:
@tomas862 is this PR still WIP? And since it's POC, is it expected to be merged?

tomas862:
@sarjon Im waiting to confirm this idea with someone. If confirmed then I'll continue to implement loading animation pop-ups alongside this.

matks:
@tomas862 Idea and code are ✅ !

tomas862:
ready to be battle tested! @matks

Progi1984 (dismissed):

Progi1984:
@tomas862 If not in WIP, please remove the WIP label.

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
```diff
@@ -0,0 +1,162 @@
...
+      return;
+    }
+
+    /**
+     * Async submit action which repeats itself in recursive order.
+     * @param items
+     * @param successCallback
+     * @param errorCallback
+     */
+    const asyncSubmit = (items, successCallback, errorCallback) => {
```
jolelievre, line 138: you could define this as a class function instead of a local const

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * Collects required data and submits the form.
+   * @param event
+   * @param grid
+   */
+  submit(event, grid) {
+    const $submitBtn = $(event.currentTarget);
+    const confirmMessage = $submitBtn.data('confirm-message');
+
+    if (typeof confirmMessage !== 'undefined' && confirmMessage.length > 0 && !confirm(confirmMessage)) {
```
jolelievre: Maybe you could integrate the modal confirmation like in the `SubmitBulkActionExtension` With the component created by @NeOMakinG it should be easy to integrate
You would need to add a modal option in the grid definition as well

jolelievre (changes requested):
I like the idea, the only problem is that it's a pure POC so how can we test it?
It requires a bulk ajax controller as well I guess? Or it doesn't matter what the controller returns as long as it's a 200 code response?

Last thing I'm worried, sometimes we return a redirect response which may be interpreted as a success, but some of the deletion have failed Then we inform the user of those failures through session messages. In this case those error messages will be lost in each ajax calls, and when the page reloads they won't be displayed. So either we find a mechanism to "catch" or "store" temporarily those messages, or we need a proper ajax response with errors.
An ajax dedicated controller might work even without a json response, the messages are stored in session, but since they're not displayed in the ajax response they will be in the page reload.

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
tomas862: done :)

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
tomas862: if I do that there will be one disadvantage:

I will need to pass two more variables `submitMethod` and `formUrl` via method params or I will need to define them as private properties in the class to be able to use them. None of the options looks clean for me.

Im taking here advantage of having anonymous function here. Without using it I would need to think of how to pass these two extra variables which dont participate in recursive iterations

tomas862:
> @tomas862 If not in WIP, please remove the WIP label.

@Progi1984  I dont think I have permissions to do this

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
jolelievre: I would be fine with using private fields, but I get your point of view And it's valid so nevermind what I said :)

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
```diff
@@ -0,0 +1,197 @@
...
+    const $submitBtn = $(event.currentTarget);
+    const confirmMessage = $submitBtn.data('confirm-message');
+    const confirmTitle = $submitBtn.data('confirmTitle');
+    const hasConfirmTitle = confirmTitle !== undefined;
+    const hasConfirmationMessage = confirmMessage !== undefined && confirmMessage.length > 0;
+    const hasAllConfirmationMessages = hasConfirmationMessage && hasConfirmTitle;
+
+    if (hasAllConfirmationMessages) {
+      this.showConfirmModal($submitBtn, grid, confirmMessage, confirmTitle);
+    } else if (hasConfirmationMessage && !hasAllConfirmationMessages && confirm(confirmMessage)) {
```
jolelievre: No need to check for `hasAllConfirmationMessages` since you are already in `else if`

tomas862:
> I like the idea, the only problem is that it's a pure POC so how can we test it?
> It requires a bulk ajax controller as well I guess? Or it doesn't matter what the controller returns as long as it's a 200 code response?
> 
> Last thing I'm worried, sometimes we return a redirect response which may be interpreted as a success, but some of the deletion have failed Then we inform the user of those failures through session messages. In this case those error messages will be lost in each ajax calls, and when the page reloads they won't be displayed. So either we find a mechanism to "catch" or "store" temporarily those messages, or we need a proper ajax response with errors.
> An ajax dedicated controller might work even without a json response, the messages are stored in session, but since they're not displayed in the ajax response they will be in the page reload.

@jolelievre  I did some testing but did not comitted it. This will be used in new product list which is not merged yet. Definetly will use it.

I think there are no worries regarding it due to we call a custom action which expects that the form is submitted using ajax and returns json response.

src/Core/Grid/Action/Bulk/Type/SubmitAsyncBulkAction.php
```diff
@@ -0,0 +1,51 @@
...
+        $resolver
+            ->setRequired([
+                'submit_route',
+            ])
+            ->setDefaults([
+                'confirm_message' => null,
+                'submit_method' => 'POST',
+                'modal_options' => null,
+                'route_params' => [],
+                'step' => 1,
```
jolelievre: So `step` is used to define `chunkSize` Maybe then `chunkSize` would be a more obvious parameter name? Or something more verbose like `sentIdsPerRequest` or `idsNumberPerRequest`

jolelievre (commented):
Minor improvement left but it seems like a good feature

I read the whole conversation so I understand the use case better now ^^ If it is linked to a ajax json controller then no problem with the session messages Besides I also thought this definitely needs to display the advancement in a way but I believe you planned to integrate this in another PR?

I would have a few recommendations on missing feature that could be dealt with in the next PR as well:
- you stop the loop as soon as an error is returned, but the action may fail for one product and succeed for all the following, so we should have a parameter to define if a single error should be blocking
- the success/error callback are quite blunt for now This is enough for the POC and the generic class but making them (optionally) configurable would allow more advanced use cases for this extension
- speaking about configurable callbacks I guess you could add a `progressCallback` called after each requests, this way you could leave this extension nearly unmodified in the next PR and manage all the progression display in another component

src/Core/Grid/Action/Bulk/Type/SubmitAsyncBulkAction.php
tomas862: chunkSize atleast for me sound better - will change to it :)

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  /**
+   * Collects required data and submits the form.
+   * @param event
+   * @param grid
+   */
+  submit(event, grid) {
+    const $submitBtn = $(event.currentTarget);
+    const confirmMessage = $submitBtn.data('confirm-message');
```
jolelievre, line 102: Please use camel cas here: `$submitBtn.data('confirmMessage')`

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
```diff
@@ -0,0 +1,238 @@
...
+    const $checkedInputs = this.getCheckedInputs($form);
+    const ids = this.getCheckedIds($checkedInputs);
+
+    const inputName = this.getInputName($checkedInputs);
+    const chunkSize = parseInt($submitBtn.data('chunkSize'), 10);
+    const chunkedIds = this.chunkArray([...ids], chunkSize);
+    const submitMethod = $submitBtn.data('form-method');
+    const formUrl = $submitBtn.data('form-url');
+
+    if (!['POST', 'GET', 'DELETE', 'PUT'].includes(submitMethod)) {
```
jolelievre: Convert `submitMethod` to upper case when you retrieve it from the data, so that this extension doesn't block the behaviour in case someone uses `get`, `post`, ...

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
```diff
@@ -0,0 +1,238 @@
...
+      data[inputName] = firstIdsChunk;
+
+      $.ajax({
+        type: submitMethod,
+        url: formUrl,
+        data,
+        dataType: 'json',
+      }).then((response) => {
+        this.progressCallback(response, firstIdsChunk, ids);
+        if (!response.success) {
```
jolelievre: You should use an `else if` after this statement because in case the last call has an error, the condition `response.success && items.length > 0` won't be valid, then it will go in the `else` So you will consecutively call `errorCallback` and `successCallback` which may be confusing ^^

Besides as I told you in a previous comment, I don't think you should stop the loop in case of one mistake The bulk action may fail on only one out of ten products but you should try performing it for the remaining ones

So if you manage the success/error callbacks in the `then` and `catch` method, but systematically perform the `asyncSubmit` check and call in the `always` method, then you have proper display of error/success messages and you still try to loop through all the product list

jolelievre (changes requested):
Thanks for the modif @tomas862 

There is the error management which I think should not be that blocking (eg: it should loop through all the products), but it's worth having another view point @matks what do you think? (more details in my comments but tldr: I don't think a single error should stop the following bulk action calls)

jolelievre:
@Progi1984 I'm curious about your opinion as well on my feedback ^^ Since you also reviewed this PR

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
```diff
@@ -0,0 +1,238 @@
...
+
+  submitForm($submitBtn, grid) {
+    const $form = $(`#${grid.getId()}_filter_form`);
+    const $checkedInputs = this.getCheckedInputs($form);
+    const ids = this.getCheckedIds($checkedInputs);
+
+    const inputName = this.getInputName($checkedInputs);
+    const chunkSize = parseInt($submitBtn.data('chunkSize'), 10);
+    const chunkedIds = this.chunkArray([...ids], chunkSize);
+    const submitMethod = $submitBtn.data('form-method');
```
Progi1984:
```suggestion
    const submitMethod = $submitBtn.data('formMethod');
```

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
```diff
@@ -0,0 +1,238 @@
...
+    const formUrl = $submitBtn.data('form-url');
```
Progi1984:
```suggestion
    const formUrl = $submitBtn.data('formUrl');
```

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
PierreRambaud: Why? :thinking: 
image
Both are valid.
But IMHO, using `confirm-message` is better to use, because it's the same as the html `data-confirm-message`.

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
PierreRambaud: same as https://github.com/PrestaShop/PrestaShop/pull/14542/files#r374128685

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
PierreRambaud: same as https://github.com/PrestaShop/PrestaShop/pull/14542/files#r374128685

src/Core/Grid/Action/Bulk/Type/SubmitAsyncBulkAction.php
```diff
@@ -0,0 +1,51 @@
...
+            ->setDefaults([
+                'confirm_message' => null,
+                'submit_method' => 'POST',
+                'modal_options' => null,
+                'route_params' => [],
+                'chunk_size' => 1,
+            ])
+            ->setAllowedTypes('submit_route', 'string')
+            ->setAllowedTypes('confirm_message', ['string', 'null'])
+            ->setAllowedValues('submit_method', ['POST', 'GET', 'DELETE', 'PUT'])
```
PierreRambaud:
```suggestion
            ->setAllowedValues('submit_method', [Request::METHOD_POST, Request::METHOD_GET, Request::METHOD_DELETE, Request::METHOD_PUT])
```

src/Core/Grid/Action/Bulk/Type/SubmitAsyncBulkAction.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    protected function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver
+            ->setRequired([
+                'submit_route',
+            ])
+            ->setDefaults([
+                'confirm_message' => null,
+                'submit_method' => 'POST',
```
PierreRambaud:
```suggestion
                'submit_method' => Request::METHOD_POST,
```

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
jolelievre: I know both notations work, but since we use the `data()` function which is an interface on the dataset I think it's preferable to emphasize its use (I think dataset in vanilla code requried the use of camel case)
Or if you hate camel case so much you could also use `attr('data-confirm-message')` ^^ (even though I'm aware they are not strictly equivalent)
Anyway it's not that important which case we use but at least if not consistent in the whole project (which is lame..) let's keep it consistent in the same file!
```js
const confirmMessage = $submitBtn.data('confirm-message');
const confirmTitle = $submitBtn.data('confirmTitle');
```

We should also make the same modif in the original extension by the way: https://github.com/PrestaShop/PrestaShop/blob/develop/admin-dev/themes/new-theme/js/components/grid/extension/submit-bulk-action-extension.js#L61

PierreRambaud: You are right, we should follow to RFC convention
```
- the prefix data- is removed (including the dash);
- for any dash (U+002D) followed by an ASCII lowercase letter a to z, the dash is removed and the letter is transformed into its uppercase counterpart;
- other characters (including other dashes) are left unchanged.
```

Progi1984 (commented):
Some feedback

src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Actions/Bulk/submit_async.html.twig
```diff
@@ -0,0 +1,41 @@
...
+        type="button"
+        data-form-url="{{ path(action.options.submit_route, action.options.route_params) }}"
+        data-form-method="{{ action.options.submit_method }}"
+        data-confirm-message="{{ action.options.confirm_message }}"
+        data-chunk-size="{{ action.options.chunk_size }}"
+        {% if action.options.modal_options %}
+          data-close-button-label="{{ action.options.modal_options.options.close_button_label|default('Close'|trans({}, 'Admin.Actions')) }}"
+          data-confirm-title="{{ action.options.modal_options.options.title }}"
+          data-confirm-button-class="{{ action.options.modal_options.options.confirm_button_class|default('btn-primary') }}"
+          data-confirm-button-label="{{ action.options.modal_options.options.confirm_button_label|default('Confirm'|trans({}, 'Admin.Actions')) }}"
```
jolelievre, line 37: I'm thinking maybe we should add some "lazy loading/access" in the `ModalOptions` class, so that we wouldn't need to access the options via the the options field and then we could simply use `action.options.modal_options.confirm_button_class` in the template?

admin-dev/themes/new-theme/js/components/grid/extension/async-submit-bulk-action-extension.js
```diff
@@ -0,0 +1,237 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+import ConfirmModal from '../../modal';
```
PierreRambaud, line 26:
```suggestion
import ConfirmModal from '@components/modal';
```
Maybe?

PierreRambaud:
Hi @tomas862 ,

Thanks!

Issue #12901, opened 2019-03-13
Make carrier name multilingual


VorobiovM:
**Is your feature request related to a problem? Please describe.**
Many Prestashop e-commerce's uses local carriers, that may have different names depending on the language.

**Describe the solution you'd like**
Add the ability to localize courier name. This feature was first proposed by [PSCSX-8505] back in 2016.

Here are the specs for the localization.

## Specs

Page: Shipping > Carriers > Add new / edit carrier

[US]: As a user, I want to be able to localize a carrier name so that I can adapt it to different languages 

### Solution

Add a dropdown menu to the "Carrier name" field in order to select the language.
![image](https://user-images.githubusercontent.com/47456895/57872396-e0066680-780b-11e9-9a70-9fbb01411c64.png)

khouloudbelguith:
Hi @VorobiovM,

Thanks for your suggestion.
Yes, it could be a good improvement to have a carrier name in Back Office is translatable Multilanguage 
![image](https://user-images.githubusercontent.com/16067358/54262522-a6488380-456e-11e9-9e6f-1c90862436a6.png)
@marionf, @colinegin what do you think?
Thanks!

Previous ticket on forge: http://forge.prestashop.com/browse/BOOM-4699

ghost:
Hi @VorobiovM,

Thanks for you feedback 👍 

Just to be sure, what you would like is to have is a dropdown menu on the "Carrier name" line with your different localizations? 
Like in the picture below :
 
![Capture d’écran 2019-03-13 à 15 22 44](https://user-images.githubusercontent.com/47456895/54286706-9cda0e00-45a4-11e9-89de-f58958b2cd14.png)

VorobiovM:
Hi @samuel-pires,

Yes, a dropdown will be perfect.

rdy4ever:
I'm in on this too. Sometimes, you don't want to disclose the carrier name, and you would preffer to be something like "Express delivery" - that needs to be translated.
Also - the "Pick-up in store" delivery method needs to be translated.

ghost:
Thank you both for your feedback. We'll look into it !
Do not hesitate if you have more info about this feature.

It is always good for us to know how you use a feature and in which context 👍

ghost:
@rdy4ever,  when you say "Pick-up in store" delivery method has to be translated, are you talking about the "transit time" field?    (see image)
[...]

PR #14551, opened 2019-07-09
[On pause] Create AddCarrierCommand and cover with tests

zuk3975:
| Questions     | Answers
| Branch?       | develop 
| Description?  | Creates AddCarrier command. There is no user interface currently migrated for this, so it is covered with tests within the same PR.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #10586
| How to test?  |

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14551)


src/Core/Domain/Carrier/Command/AddCarrierCommand.php
```diff
@@ -0,0 +1,203 @@
...
+
+    /**
+     * @var string
+     */
+    private $trackingUrl;
+
+    /**
+     * @var bool
+     */
+    private $includeShippingCost;
```
tomas862: `isShippingCostIncluded` maybe? Thats more standart way to tell this variable is boolean type . add prefix `is`, `does`, `has` etc...

src/Core/Domain/Carrier/Command/AddCarrierCommand.php
```diff
@@ -0,0 +1,203 @@
...
+
+    /**
+     * @var int
+     */
+    private $taxId;
+
+    /**
+     * @var bool
+     */
+    private $outOfRangeBehavior;
```
tomas862: from the legacy form I see that it currently has two values:

0 is equal to "Apply the cost of the highest defined range"
1 is equal to "Disable carrier"

Screenshot (7)

is this really the right way to treat is as a boolean? I think its enum type and you can create object of it will all available values for it.

src/Core/Domain/Carrier/Command/AddCarrierCommand.php
```diff
@@ -0,0 +1,203 @@
...
+    /**
+     * Resolves array parameters to contain valid structure
+     *
+     * @param array $params
+     *
+     * @return array
+     */
+    private function resolveRangeParams(array $params)
+    {
+        $resolver = new OptionsResolver();
```
tomas862: interesting - should we not try to avoid symfony usage within the command at all cost? For such simple thing maybe simple validations here would be enough ?

classes/Carrier.php
```diff
@@ -57,6 +57,11 @@ class CarrierCore extends ObjectModel
     /** @var string Name */
     public $name;
 
+    /**
+     * @var array of names where key represents language id
+     */
+    public $localized_name;
```
sarjon: Not sure about new property. It was the same situation in `Currency` when `name` was kept as is after it became translatable.

install-dev/data/db_structure.sql
```diff
@@ -96,7 +96,7 @@ CREATE TABLE `PREFIX_carrier` (
   `id_carrier` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `id_reference` int(10) unsigned NOT NULL,
   `id_tax_rules_group` int(10) unsigned DEFAULT '0',
-  `name` varchar(64) NOT NULL,
+  `name` varchar(64) NOT NULL, /* deprecated since 1.7.7.0 , use carrier_lang.localized_name instead */
```
sarjon: It looks strange to have deprecation in SQL, I'm not sure if anyone reads this file besides core team. :grin: You can use comments as they are visible in tools like phpmyadmin & etc, but again, not sure this deprecation is needed.

install-dev/data/db_structure.sql
```diff
@@ -128,6 +128,7 @@ CREATE TABLE `PREFIX_carrier_lang` (
   `id_shop` int(11) unsigned NOT NULL DEFAULT '1',
   `id_lang` int(10) unsigned NOT NULL,
   `delay` varchar(512) DEFAULT NULL,
+  `localized_name` varchar(64) NOT NULL, /* since 1.7.7.0 */
```
sarjon: I think you should use `install/upgrade/sql/1.7.7.0.sql` to introduce schema changes instead.

ping @matks for confirmation.

src/Adapter/Carrier/CommandHandler/AddCarrierHandler.php
```diff
@@ -0,0 +1,242 @@
...
+         * SHIPPING_METHOD_DEFAULT @deprecated 1.5.5
+         */
+        if (Carrier::SHIPPING_METHOD_DEFAULT === $shippingMethod) {
+            $shippingMethod = $this->defaultShippingMethod ?
+                ShippingMethod::SHIPPING_METHOD_WEIGHT : ShippingMethod::SHIPPING_METHOD_PRICE;
+        }
+        $carrier->shipping_method = $shippingMethod;
+        $carrier->is_free = ShippingMethod::SHIPPING_METHOD_FREE === $shippingMethod ? true : false;
+
+        $localizedNames = $command->getLocalizedCarrierNames();
```
sarjon: Does it need `Carrier` word in `getLocalizedCarrierNames`? It does not have `Carrier` in `getShippingMethod` and other methods. :thinking:

src/Adapter/Carrier/CommandHandler/AddCarrierHandler.php
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * @param Carrier $carrier
+     * @param ShippingMethod $shippingMethod
+     * @param ShippingRange[] $shippingRanges
+     *
+     * @throws CarrierException
+     * @throws PrestaShopException
+     * @throws \PrestaShopDatabaseException
+     */
+    private function handleRanges(Carrier $carrier, ShippingMethod $shippingMethod, array $shippingRanges): void
```
sarjon: How about `addShippingRanges()` or similar instead of `handleRanges`?

src/Adapter/Carrier/CommandHandler/AddCarrierHandler.php
```diff
@@ -0,0 +1,242 @@
...
+                $range = new RangePrice();
+                $this->addRange($range, $carrierId, $shippingRange);
+                $this->addDeliveryPrice($shippingRange, $shippingMethod, $carrier, (int) $range->id);
+            }
+
+            return;
+        }
+
+        if ($methodValue === ShippingMethod::SHIPPING_METHOD_WEIGHT) {
+            foreach ($shippingRanges as $shippingRange) {
```
sarjon: Seems that this `foreach` is identical to previous, can it be optimized?

src/Core/Domain/Carrier/Command/AddCarrierCommand.php
```diff
@@ -0,0 +1,360 @@
...
+    private function setMeasures(int $width, int $height, int $depth, float $weight)
+    {
+        foreach ([$width, $height, $depth, $weight] as $measure) {
+            $this->assertMeasureIsNonNegative($measure);
+        }
+
+        $this->maxPackageWidth = $width;
+        $this->maxPackageHeight = $height;
+        $this->maxPackageDepth = $depth;
+        $this->maxPackageWeight = $weight;
```
sarjon: Looks like it could be VO?

src/Core/Domain/Carrier/ValueObject/ShippingMethod.php
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Represents shipping method when the shipping price depends from total package price
+     */
+    const SHIPPING_METHOD_PRICE = 2;
+
+    /**
+     * Represents shipping method when the shipping is free of charge
+     */
+    const SHIPPING_METHOD_FREE = 3;
```
sarjon: Is `const` part of VO? It seems that it does not belong here. `assertValueIsDefinedShippingMethod` does not consider  it as valid value.

src/Core/Domain/Carrier/ValueObject/ShippingRange.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * @param array $range
+     *
+     * @return ShippingRange
+     *
+     * @throws CarrierConstraintException
+     */
+    public static function buildFromArray(array $range)
```
sarjon: Hmm, I think you should use constructor in this case, because it's the same instead developer does not know what arguments should be passed, thus making it more error prone.

classes/Carrier.php
zuk3975: Was wondering that too, but then BC break was done. This way i think its possible to avoid that.

src/Core/Domain/Carrier/Command/AddCarrierCommand.php
zuk3975: I had created 2 VO's for this case, the PackgageSizeMeasure (for int value) and PackageWeightMeasure (for float values), but these seemed to be overhead when i started using strict types, as it only asserts for non-negative value inside.

install-dev/upgrade/php/ps_1770_copy_name_from_carrier_to_carrier_lang.php
```diff
@@ -0,0 +1,34 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+function ps_1770_copy_name_from_carrier_to_carrier_lang()
+{
+    Db::getInstance()->execute(
+        "UPDATE " . _DB_PREFIX_ . " carrier_lang cl
```
rokaszygmantas: wouldn't the blank space after database prefix make the table name invalid?

src/Adapter/Carrier/CommandHandler/AddCarrierHandler.php
```diff
@@ -0,0 +1,224 @@
...
+        $this->fillLegacyCarrierWithData($carrier, $command);
+
+        try {
+            if (false === $carrier->validateFields(false) || false === $carrier->validateFieldsLang(false)) {
+                throw new CarrierException('Carrier contains invalid field values');
+            }
+
+            if (false === $carrier->add()) {
+                throw new CarrierException(
+                    sprintf('Failed to add new carrier')
```
rokaszygmantas: `sprintf` not needed

src/Adapter/Carrier/CommandHandler/AddCarrierHandler.php
```diff
@@ -0,0 +1,224 @@
...
+                );
+            }
+            $this->associateWithShops($carrier, $command->getAssociatedShopIds());
+            $carrier->setTaxRulesGroup($command->getTaxRulesGroupId());
+            $carrier->setGroups($command->getAssociatedGroupIds());
+            $this->addShippingRanges($carrier, $command->getShippingMethod(), $command->getShippingRanges());
+        } catch (PrestaShopException $e) {
+            throw new CarrierException(
+                sprintf('An error occurred when trying to add new carrier')
```
rokaszygmantas: `sprintf` not needed

src/Core/Domain/Carrier/ValueObject/CarrierName.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * @param string $value
+     *
+     * @throws CarrierConstraintException
+     */
+    private function assertIsValidLength(string $value)
+    {
+        if (0 === strlen($value) || self::MAX_LENGTH < strlen($value)) {
```
rokaszygmantas: calling `strlen()` once would be more efficient

src/Core/Domain/Carrier/ValueObject/CarrierName.php
```diff
@@ -0,0 +1,104 @@
...
+            throw new CarrierConstraintException(sprintf(
+                'Carrier name "%s" is invalid. It must be 1 - %s characters long.',
+                self::MAX_LENGTH,
```
rokaszygmantas: wrong order of arguments, and formatting looks strange :thinking:

tests/Unit/Core/Domain/Carrier/Command/TestAddCarrierCommand.php
```diff
@@ -0,0 +1,231 @@
...
+            'depth' => 20,
+            'weight' => 20.1,
+            'associated_group_ids' => [1, 2],
+            'associated_shop_ids' => [1],
+        ];
+    }
+
+    private function createCommandFromArray(array $data, bool $freeShipping)
+    {
+        $command = AddCarrierCommand::withPricedShipping(
```
rokaszygmantas: IMO it would be more efficient to put this into `else` condition instead of overwriting it later on

install-dev/data/db_structure.sql
matks: > It looks strange to have deprecation in SQL, I'm not sure if anyone reads this file besides core team. 😁 You can use comments as they are visible in tools like phpmyadmin & etc, but again, not sure this deprecation is needed.

Yes, it seems a lot of people do not trust the 1-click upgrade and performs Prestashop upgrades by themselves, copy-pasting the SQL queries

classes/Carrier.php
sarjon: Then you introduce 2 ways of setting name for Carrier.

As a developer should I set `name` or `localized_name`? How do i Know which one should I use?

src/Adapter/Carrier/CommandHandler/AbstractAddCarrierHandler.php
```diff
@@ -0,0 +1,188 @@
...
+     * @throws PrestaShopException
+     */
+    protected function addShippingRanges(Carrier $carrier, ShippingMethod $shippingMethod, array $shippingRanges): void
+    {
+        $methodValue = $shippingMethod->getValue();
+        $carrierId = (int) $carrier->id;
+        $range = new RangePrice();
+
+        if ($methodValue === ShippingMethod::SHIPPING_METHOD_WEIGHT) {
+            $range = new RangeWeight();
```
sarjon: How about 

```php
$range = $methodValue === ShippingMethod::SHIPPING_METHOD_WEIGHT ? 
    new RangePrice() : 
    new RangeWeight();
```

src/Adapter/Carrier/CommandHandler/AbstractAddCarrierHandler.php
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        foreach ($shippingRanges as $shippingRange) {
+            $this->addRange($range, $carrierId, $shippingRange);
+            $this->addDeliveryPrice($shippingRange, $shippingMethod, $carrier, (int) $range->id);
```
sarjon: Isn't `$rage->id` always `null` or `0` as you create it above?

src/Adapter/Carrier/CommandHandler/AbstractAddCarrierHandler.php
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * @param ObjectModel $range RangePrice|RangeWeight
+     * @param int $carrierId
+     * @param ShippingRange $shippingRange
+     *
+     * @throws CarrierException
+     * @throws PrestaShopException
+     */
+    protected function addRange(ObjectModel $range, int $carrierId, ShippingRange $shippingRange)
```
sarjon: It can accept any `ObjectModel`, but that would cause and error, right? Maybe you should have 2 separate methods for adding price range or weight range?

src/Adapter/Carrier/CommandHandler/AbstractAddCarrierHandler.php
```diff
@@ -0,0 +1,188 @@
...
+     * @throws CarrierException
+     */
+    protected function addDeliveryPrice(
+        ShippingRange $shippingRange,
+        ShippingMethod $shippingMethod,
+        Carrier $carrier,
+        int $rangeId
+    ) {
+        $rangePriceId = $rangeId;
+        $rangeWeightId = null;
```
sarjon: These 2 line should be in `else` block of below `if` imo.

src/Adapter/Carrier/CommandHandler/AbstractAddCarrierHandler.php
```diff
@@ -0,0 +1,188 @@
...
+
+        if ($shippingMethod->getValue() === ShippingMethod::SHIPPING_METHOD_WEIGHT) {
```
sarjon: Seems there aren't such thing as `ShippingMethod`, but `Billing` instead. My suggestion:
1. Rename `ShippingMethod` to `Billing`.
2. Add methods `Billing::accordingToTotalWeight(): bool` and `Billing::accordingToTotalPrice(): bool`, so you use it in `if`s like:

```php
if ($billing->accordingToTotalWeight()) {
    // do something ...
}
```
That would better promote rich domain model imo.

src/Adapter/Carrier/CommandHandler/AbstractAddCarrierHandler.php
```diff
@@ -0,0 +1,188 @@
...
+            $rangePriceId = null;
+            $rangeWeightId = $rangeId;
+        }
+
+        foreach ($shippingRange->getPricesByZoneId() as $zoneId => $price) {
+            $priceList[] = array(
```
sarjon: You can use short syntax array.

src/Adapter/Carrier/CommandHandler/AddCarrierHandler.php
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        return new CarrierId((int) $carrier->id);
+    }
+
+    /**
+     * @param Carrier $carrier
+     * @param AddCarrierCommand $command
+     */
+    private function fillCarrierWithData(Carrier $carrier, AddCarrierCommand $command)
```
sarjon: Maybe you could rename it to `createCarrierFromCommand`? It could return new instance with filled data.

src/Adapter/Carrier/CommandHandler/AddModuleCarrierHandler.php
```diff
@@ -0,0 +1,79 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Command\AddModuleCarrierCommand;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\CommandHandler\AddModuleCarrierHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierException;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\CarrierId;
+use PrestaShopException;
+
+/**
+ * Handles AddModuleCarrierCommand using legacy object model
+ */
+final class AddModuleCarrierHandler extends AbstractAddCarrierHandler implements AddModuleCarrierHandlerInterface
```
sarjon, line 39: Hmmm, do we have use case for adding module carrier? :thinking: If so, maybe having just one command would be enough since handlers are nearly identical.

src/Core/Domain/Carrier/Command/AbstractAddCarrierCommand.php
```diff
@@ -0,0 +1,388 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\TrackingUrl;
+use PrestaShop\PrestaShop\Core\Domain\Language\ValueObject\LanguageId;
+
+/**
+ * Provides reusable properties/methods for AddCarrierCommands
+ */
+abstract class AbstractAddCarrierCommand
+{
+    /**
+     * @var CarrierName[]
```
sarjon: Don't use VO for name.

src/Core/Domain/Carrier/Command/AbstractAddCarrierCommand.php
```diff
@@ -0,0 +1,388 @@
...
+     */
+    protected $localizedCarrierNames;
```
sarjon: I'm not sure about naming, but how many names are there? As I understand it should be in singular form as there is only one name, but in different languages.

I'd call it something like `$localizedName`.

src/Core/Domain/Carrier/Command/AbstractAddCarrierCommand.php
```diff
@@ -0,0 +1,388 @@
...
+
+    /**
+     * @var ShippingDelay[]
```
sarjon, line 50: Same here, I don't think VO is needed.

src/Core/Domain/Carrier/Command/AbstractAddCarrierCommand.php
```diff
@@ -0,0 +1,388 @@
...
+    /**
+     * @var int[]
+     */
+    protected $associatedGroupIds;
+
+    /**
+     * @var int[]
+     */
+    protected $associatedShopIds;
+    /**
```
sarjon, line 119: Need empty line.

src/Core/Domain/Carrier/Command/AbstractAddCarrierCommand.php
```diff
@@ -0,0 +1,388 @@
...
+     */
+    protected function assertRangesSequence(array $currentRange, array $previousRange)
+    {
+        if ($currentRange['from'] < $previousRange['to']) {
+            throw new CarrierConstraintException(sprintf(
+                'Shipping ranges %s - %s and %s - %s are overlapping.',
+                $previousRange['from'],
+                $previousRange['to'],
+                $currentRange['from'],
+                $currentRange['to']),
```
sarjon: Weird formatting.

src/Core/Domain/Carrier/Command/AddCarrierCommand.php
```diff
@@ -0,0 +1,156 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\OutOfRangeBehavior;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\ShippingMethod;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\SpeedGrade;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\TrackingUrl;
+
+/**
+ * Adds new carrier
+ */
+final class AddCarrierCommand extends AbstractAddCarrierCommand
```
sarjon: Remove `final`.

src/Core/Domain/Carrier/Command/AddModuleCarrierCommand.php
```diff
@@ -0,0 +1,290 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\OutOfRangeBehavior;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\ShippingMethod;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\SpeedGrade;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\TrackingUrl;
+
+/**
+ * Responsible for adding a carrier via module
+ */
+final class AddModuleCarrierCommand extends AbstractAddCarrierCommand
```
sarjon: Remove `final`.

src/Core/Domain/Carrier/Command/AddModuleCarrierCommand.php
```diff
@@ -0,0 +1,290 @@
...
+     */
+    public function getModuleName(): string
+    {
+        return $this->moduleName;
+    }
+
+    /**
+     * @return bool
+     */
+    public function doModuleCalculateShippingPrice(): bool
```
sarjon:
```diff
+ moduleCalculateShippingPrice(): bool
- doModuleCalculateShippingPrice(): bool
```

src/Core/Domain/Carrier/Command/AddModuleCarrierCommand.php
```diff
@@ -0,0 +1,290 @@
...
+    {
+        return $this->moduleCalculatesShippingPrice;
+    }
+
+    /**
+     * @return bool
+     */
+    public function doModuleNeedCoreShippingPrice(): bool
```
sarjon:
```diff
+ moduleNeedsCoreShippingPrice(): bool
- doModuleNeedCoreShippingPrice(): bool
```

src/Core/Domain/Carrier/Command/AddModuleCarrierCommand.php
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    /**
+     * @param string $name
+     *
+     * @throws CarrierConstraintException
+     */
+    private function assertModuleName(string $name)
+    {
+        $maxLength = 64;
```
sarjon: Move this into constant.

src/Core/Domain/Carrier/ValueObject/CarrierId.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierConstraintException;
+
+/**
+ * Provides valid carrier identification data
+ */
+final class CarrierId
```
sarjon: Remove `final`.

src/Core/Domain/Carrier/ValueObject/CarrierName.php
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierConstraintException;
+
+/**
+ * Provides valid carrier name value
+ */
+final class CarrierName
```
sarjon: Consider removing this VO.

src/Core/Domain/Carrier/ValueObject/ShippingDelay.php
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierConstraintException;
+
+/**
+ * Provides valid carrier shipping delay value
+ */
+final class ShippingDelay
```
sarjon: Consider removing this VO.

src/Core/Domain/Carrier/ValueObject/OutOfRangeBehavior.php
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierConstraintException;
+
+/**
+ * Provides valid out of range behavior values
+ */
+final class OutOfRangeBehavior
```
sarjon: Remove `final`.

src/Core/Domain/Carrier/ValueObject/ShippingMethod.php
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierConstraintException;
+
+/**
+ * Provides valid values for shipping method
+ */
+final class ShippingMethod
```
sarjon: As mentioned before, `ShippingMethod` is misleading, rename to `Billing` (or something similar that makes sense).

Remove `final`.

src/Core/Domain/Carrier/ValueObject/ShippingRange.php
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * @var array
+     */
+    private $pricesByZoneId;
+
+    /**
+     * @param int $from
+     * @param int $to
+     * @param array $pricesByZoneId
```
sarjon, line 54: You could typehint something more specific if possible, like `string[]`.

src/Core/Domain/Carrier/ValueObject/ShippingRange.php
```diff
@@ -0,0 +1,140 @@
...
+     *
+     * @throws CarrierConstraintException
+     */
+    public function __construct(int $from, int $to, array $pricesByZoneId)
```
sarjon: I wonder if `$from` and `$to` can be floats?

src/Core/Domain/Carrier/ValueObject/ShippingRange.php
```diff
@@ -0,0 +1,140 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Carrier\Exception\CarrierConstraintException;
+
+/**
+ * Provides shipping range and its prices by zone
+ */
+final class ShippingRange
```
sarjon: If it does not implement interface, don't make it `final`. Check other VOs as well.

src/Core/Domain/Carrier/ValueObject/SpeedGrade.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * @param $value
+     *
+     * @throws CarrierConstraintException
+     */
+    private function assertValueIsNonNegativeIntegerLessThanTen(int $value)
+    {
+        if (0 > $value || 9 < $value) {
```
sarjon: What are these values? Could you explain or maybe define as constants?

src/PrestaShopBundle/Resources/config/services/adapter/carrier.yml
```diff
@@ -0,0 +1,20 @@
+services:
+
```
sarjon, line 4: Make your services public.

src/PrestaShopBundle/Resources/config/services/adapter/carrier.yml
```diff
@@ -0,0 +1,20 @@
...
+    prestashop.adapter.currency.command_handler.abstract_add_carrier:
```
sarjon: I smell copy-paste. :grin: Check other service ids.

tests/Integration/Behaviour/Features/Context/Domain/CarrierFeatureContext.php
```diff
@@ -0,0 +1,482 @@
...
+        $carrier = SharedStorage::getStorage()->get($reference);
+        $needsPrice = 'should' === $condition;
+        $actualResult = (bool) $carrier->need_range;
+
+        if ($needsPrice !== $actualResult) {
+            throw new RuntimeException(sprintf(
+                'Carrier "%s" module %s the shipping price calculated by PrestaShop, but was expected it to %s',
+                $reference,
+                $actualResult ? 'needs' : 'does not need',
+                $condition === 'should' ? 'need' : 'not need'
```
sarjon: Could you simplify message so we can avoid having these conditions?

tests/Integration/Behaviour/Features/Context/Domain/CarrierFeatureContext.php
```diff
@@ -0,0 +1,482 @@
...
+    /**
+     * @Then /^Carrier "(.+)" (should|should not) belong to module$/
+     */
+    public function assertIsModule($reference, $condition)
+    {
+        $isModule = 'should' === $condition;
+
+        /** @var Carrier $carrier */
+        $carrier = SharedStorage::getStorage()->get($reference);
+        $actualResult = (bool) $carrier->is_module;
```
sarjon, line 282: This var seems redundant.

tests/Integration/Behaviour/Features/Context/Domain/CarrierFeatureContext.php
```diff
@@ -0,0 +1,482 @@
...
+                $carrier->localized_name[$this->defaultLangId],
+                $value
+            ));
+        }
+    }
+
+    /**
+     * @Then Carrier :reference shipping delay in default language should be :value
+     */
+    public function assertShippingDelay($reference, $value)
```
sarjon: Do you think you could merge this assertion with `assertLocalizedName()` since they do exactly the same.

tests/Integration/Behaviour/Features/Scenario/Carrier/carrier_management.feature
```diff
@@ -0,0 +1,175 @@
...
+      | ranges_to                 | 2, 3, 4                                           |
+      | zone_ids                  | 1, 2, 3                                           |
+      | prices                    | 5, 6, 7                                           |
+      | max_width                 | 10                                                |
+      | max_height                | 20                                                |
+      | max_depth                 | 15                                                |
+      | max_weight                | 5.5                                               |
+      | group_ids                 | 1                                                 |
+      | shop_ids                  | 1                                                 |
+    Then Carrier "my carrier" name in default language should be "My carrier"
```
sarjon: I think Carrier should not start with capital letter.

tests/Unit/Core/Domain/Carrier/Command/ValidDataProviderForCommandCreation.php
```diff
@@ -0,0 +1,73 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace Tests\Unit\Core\Domain\Carrier\Command;
+
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\OutOfRangeBehavior;
+use PrestaShop\PrestaShop\Core\Domain\Carrier\ValueObject\ShippingMethod;
+
+final class ValidDataProviderForCommandCreation
```
sarjon, line 32: Not sure about this one, I'd prefer to keep data it in the test class. We can see what others think.

sarjon (changes requested):

classes/Carrier.php
zuk3975: @matks, could you suggest how this should be handled? Having 2 names sounds like crap, maybe it is worth breaking bc? Something similar like with currency name?

src/Adapter/Carrier/CommandHandler/AbstractAddCarrierHandler.php
zuk3975: rage indeed :imp: You are right, I'm on it..

zuk3975: $this->addRange(..) adds the range to database inside and the legacy object model gets id value

tests/Integration/Behaviour/Features/Context/Domain/CarrierFeatureContext.php
zuk3975: The bool cast is necessary and this var is used couple times, so it seems ok to me

tests/Unit/Core/Domain/Carrier/Command/ValidDataProviderForCommandCreation.php
zuk3975: It is shared between multiple addCarrier commands

src/Adapter/Carrier/CommandHandler/AddModuleCarrierHandler.php
zuk3975: I don't. But I guess there will be no use cases unless i create carriers module.. I'm afraid it would be forgotten to migrate at all.
You mean to get rid of AddModuleCarrierCommand and instead use something like $isModule property to preset all redundant data based on it?

src/Core/Domain/Carrier/Command/AbstractAddCarrierCommand.php
zuk3975: Why? where do I put carrier specific name and length validation and constants?

src/Core/Domain/Carrier/Command/AbstractAddCarrierCommand.php
zuk3975: Renamed, but not sure though, technically it doesn't sound wrong to have it plural

src/Core/Domain/Carrier/ValueObject/ShippingRange.php
zuk3975: Its id => value pairs array, I think typehinting could be missleading, no? I could do like float[], but IMO this way makes it look like the index value is not important.

classes/Carrier.php
matks: So this is about making the carrier name translatable, right ? 🤔 

This feature *requires* a BC break, it is a BC break in behavior which means a BC break in the code 😉 .
So `name` it is.

install-dev/data/db_structure.sql
matks: Answered, we stay with `name`

matks:
@zuk3975 Can you handle https://github.com/PrestaShop/PrestaShop/pull/14551/files#r314652605 then I review ? because this is going to change noticeably the code

sarjon:
> @zuk3975 Can you handle https://github.com/PrestaShop/PrestaShop/pull/14551/files#r314652605 then I review ? because this is going to change noticeably the code

@matks as discussed here https://github.com/PrestaShop/PrestaShop/issues/12901#issuecomment-521965177 we should do a separate PR for translatable carrier name.

matks:
So we decided to not translate the name for this version. @zuk3975 I'm sorry, can you remove the code related to translation of the name ? No more `$localized_name` 😞

matks:
@zuk3975  I close this PR as it currently appears in our backlog of PRs and until I (or someone else) find some time to finish it, it won't move 😞  .

PR #14564, opened 2019-07-09
ProductId should throw a ProductException instead of an OrderException

### LINKED ISSUE #14520: ProductId throw an OrderException instead of ProductException
**Describe the bug**
I need to use a [ProductId](https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Domain/Product/ValueObject/ProductId.php) for a module (1.7.6.+) but it throw an OrderException instead of ProductException

**To Reproduce**
1. Go to https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Domain/Product/ValueObject/ProductId.php
2. Look on assertIntegerIsGreaterThanZero function
3. See Exception thrown is not a ProductException

**Additionnal information**
PrestaShop version: develop
PHP version: PHP 7.2

Matt75:
| Questions     | Answers
| Branch?       | develop
| Description?  | ProductId should throw a ProductException instead of an OrderException
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14520
| How to test?  | Run Unit Tests

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14564)

### DIFF
--- a/tests/Unit/Core/Domain/Product/ValueObject/ProductIdTest.php
+++ b/tests/Unit/Core/Domain/Product/ValueObject/ProductIdTest.php
@@ -24,50 +24,31 @@
  * International Registered Trademark & Property of PrestaShop SA
  */
 
-namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject;
+namespace Tests\Unit\Core\Domain\Product\ValueObject;
 
+use PHPUnit\Framework\TestCase;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
 
-/**
- * Product identity.
- */
-class ProductId
+class ProductIdTest extends TestCase
 {
     /**
-     * @var int
+     * @dataProvider getInvalidValue
      */
-    private $productId;
-
-    /**
-     * @param int $productId
-     */
-    public function __construct($productId)
+    public function testItThrowsExceptionWhenWrongValueIsGiven($invalidValue)
     {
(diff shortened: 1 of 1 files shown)


PierreRambaud (approved):

mickaelandrieu:
While I agree with the fix, this is a BC break :'(

We had to merge that before the 1.7.6.0, I don't think this is worth to update it before the 1.8.x...

wdyt @matks ?

matks:
> While I agree with the fix, this is a BC break :'(

I disagree.
We do not modify a function signature nor its return, we only modify the Exception being thrown. If you're worried that some external code relies on that, we can mention it in the 1.7.7 release note as [it has been done for the 1.7.6 release note](http://build.prestashop.com/news/prestashop-1-7-6-0-available/) in a "breaking or risky changes" section.

Also, it's an obvious mistake. Patch versions are allowed to break _behavior_ (not signature) if it's a bugfix to me. If a security vulnerability allows someone to go through a bad check, fixing the check is not considered a breaking change although some external code could rely on it ?

Matt75:
So what's the status on this issue ?

PierreRambaud (dismissed):

PierreRambaud:
Can you please rebase your branch?
@matks Can you please approve this change.

Matt75:
Should be ok now

PierreRambaud (approved):

Matt75:
Oh wait, looks like it was already replaced by ProductConstraintException so this PR is useless

PierreRambaud:
Haha sorry @Matt75 ^^'
Really sorry for the delay, I'm closing your PR.

PR #14288, opened 2019-06-19
Migrate Customer Service view

sarjon:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Migrates "View" action of Customer Service
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #10553 
| How to test?  | :warning: Build assets before testing :warning: Create at least 1 customer service and access new page with url `admin-dev/index.php/sell/customer-service/customer-threads/{ID}/view` where `{ID}` is Customer Service ID. 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14288)

PR #14569, opened 2019-07-10
fix: set customer thread status to closed when answering in backoffice

### LINKED ISSUE #14313: Customer service - status not changed on reply 1.7.5.2
**Describe the bug**
Status of a message thread is not being changed to "Closed" when replying from order's page.

**To Reproduce**
1. Create an account, place an order, go to contact us page, select order ref number and send a message
2. In BO view the order and reply to the message:
reply
3. Go to Customer Service - thread is still with green light - open status (should be status closed)
4. Despite the fact that the message is correctly visible in thread:
thread

The thread should be marked as handled / closed no matter if replied from order's page or from its thread.
 
**Screenshots**
https://drive.google.com/file/d/10ERrehIxwITxuc0GDK9rOAUO80akPU83/view

**Additionnal information**
PrestaShop version: 1.7.5.2
PHP version: 7.3.6

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fixes customer thread's status in customer service not being set to closed after the administrator answered.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14313
| How to test?  | The process is described in issue #14313

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14569)

### DIFF
--- a/classes/CustomerThread.php
+++ b/classes/CustomerThread.php
@@ -25,6 +25,11 @@
  */
 class CustomerThreadCore extends ObjectModel
 {
+    const openStatus = 'open';
+    const closedStatus = 'closed';
+    const pending1Status = 'pending1';
+    const pending2Status = 'pending2';
+
     public $id;
     public $id_shop;
     public $id_lang;
--- a/classes/CustomerThread.php
+++ b/classes/CustomerThread.php
@@ -25,6 +25,8 @@
  */
 class CustomerThreadCore extends ObjectModel
 {
+    // TODO: remove those constants, they will be implemented in CustomerThreadStatus class
+    // in an incoming pull request (https://github.com/PrestaShop/PrestaShop/pull/14288/files)
     const openStatus = 'open';
     const closedStatus = 'closed';
     const pending1Status = 'pending1';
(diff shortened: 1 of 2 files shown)


classes/CustomerThread.php
```diff
@@ -25,6 +25,11 @@
  */
 class CustomerThreadCore extends ObjectModel
 {
+    const openStatus = 'open';
+    const closedStatus = 'closed';
+    const pending1Status = 'pending1';
+    const pending2Status = 'pending2';
```
sarjon, line 33: These statuses are introduced here https://github.com/PrestaShop/PrestaShop/pull/14288/files#diff-7a6118dbd41aee4add4f5f04d7bd3f35R34 

I prefer to use `CustomerThreadStatus` VO instead of this, @matks wdyt?

matthieu-rolland: I think you're totally right @sarjon , I had no knowledge of this.
I will add a todo in comment above those constants and where I used it, once the other PR will have been approved, I'll use the CustomerThreadStatus class (since it doesn't exist yet  on develop branch)

Your migration of the customer service view will replace this anyway.

matthieu-rolland:
This PR has no reason to exist anymore.

PR #14574, opened 2019-07-10
Reusable regex patterns for validation

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | When validating forms we already have typedRegexValidator which encapsulates reusable regex patterns in one class and eases code reusability, but we still use plain regex when validating Domain models. It fits well when pattern is unique and it can be defined in corresponding domain VO (`e.g. carrier_name`). But it doesn't look right when the pattern is reusable e.g. `generic_name`, `catalog_name`... Maybe similar PR could be prepared to encapsulate those reusable patterns in Domain and clean the code where it is already being used as plain string.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? |
| How to test?  |

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14574)


zuk3975:
It is a very basic example to explain the idea, but it should be clear. 
@matks, @sarjon, @rokaszygmantas, @tomas862 wdyt ? 
Also any ideas how could the inverse patterns be handled ? e.g. `message` https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/ConstraintValidator/TypedRegexValidator.php#L151 ?

src/Core/Domain/Validation/RegexPattern.php
```diff
@@ -34,5 +34,7 @@ final class RegexPattern
     /**
      * Pattern for valid generic names
      */
-    const GENERIC_NAME = '/^[^<>={}]*$/u';
+    const PATTERN_GENERIC_NAME = '/^[^<>={}]*$/u';
+
+    const INVERSE_PATTERN_MESSAGE = '/[<>{}]/i';
```
rokaszygmantas: I don't understand the idea behind inverse pattern :thinking:

zuk3975: basically it contains pattern which is valid when preg_match is false. It is maybe possible to rewrite the pattern to fit normally :thinking:

rokaszygmantas: I think we could just write the inverted pattern, such as `/^[^\[<>{}\]]+$/i` in this case

src/Core/Domain/Validation/RegexPattern.php
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Provides reusable regex patterns for validation
+ */
+final class RegexPattern
+{
+    /**
+     * Pattern for valid generic names
+     */
+    const PATTERN_GENERIC_NAME = '/^[^<>={}]*$/u';
```
sarjon: You don't need to prefix it with `PATTERN_` I think.

src/Core/Domain/Validation/RegexPattern.php
```diff
@@ -0,0 +1,40 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Validation;
```
sarjon: I don't think that `Validation` should be part of `Domain`.

zuk3975: But it is for business rules validation

sarjon: As I understand, `Domain` should represent important concepts of PrestaShop (ecommerce) that could be Product, Category, Payment, Manufacturer & etc. 

Each of these subdomains have their own constraints and it happens to have similar (or identical) constraints between each other, this is where code duplication occurs. 

However, this allows different subdomains to evolve independently, without affecting each other.

In this particular case, you extracted common constraints into new subdomain called `Validation` which does not have any conceptual meaning for PrestaShop (ecommerce), at least this is how I see it.

@matks wdyt?

zuk3975: makes sense :thinking: . Do you offer just having the patterns in each Domain VO's where it's needed and don't mind the duplication as those VO's 'coincidentally' have the same business rules ?

sarjon: Not sure, maybe I'd put it in `Core/Validation`, let's see what @matks thinks.

src/Core/Domain/Validation/RegexPattern.php
matks: Indeed, as pattern is already in the class name

src/Core/Domain/Validation/RegexPattern.php
matks: Argh :/ I'm afraid there's no easy answer

Strictly speaking, Sarunas is right about keeping subdomains independent from each other.

However we could also consider this class to be some Utils class that is used in domains, but could live outside of it. For example Doctrine traits like TimestampableTrait are _usable and useful_ behaviors that are then _added_ to real domain elements (the entities). So they are not part of a subdomain but they are used in multiple of them.

So I think we can allow this class to exist, like the src/Core/Domain/Exception folder exists because it's used everywhere. It's a util class. It could be put inside a dedicated git repository and provided on packagist as "a useful repository of regex patterns to valide standard strings in the e-commerce world"

It's a very blurry line 😢

matks:
@zuk3975 Can you handle
- https://github.com/PrestaShop/PrestaShop/pull/14574#pullrequestreview-260080493
- https://github.com/PrestaShop/PrestaShop/pull/14574#pullrequestreview-261760480

Then we can merge 😄 
Regarding https://github.com/PrestaShop/PrestaShop/pull/14574#discussion_r303389827 as I said it's a very blurry line, I think we can stick to current implementation and refactor it later if needed

sarjon:
> I think we can stick to current implementation and refactor it later if needed

Not until at least next major. :confused:

zuk3975:
> Regarding [#14574 (comment)](https://github.com/PrestaShop/PrestaShop/pull/14574#discussion_r303389827) as I said it's a very blurry line, I think we can stick to current implementation and refactor it later if needed

There is already existing dir `Core/Validation`, it seems to fit in well d301b4a  :thinking: ?

src/Core/Domain/CmsPageCategory/Command/AbstractCmsPageCategoryCommand.php
```diff
@@ -47,12 +44,12 @@ abstract class AbstractCmsPageCategoryCommand
     protected function assertCategoryName(array $names)
     {
         foreach ($names as $name) {
-            if (!preg_match(self::CATEGORY_NAME_REGEX_PATTERN, $name)) {
+            if (!preg_match(RegexPattern::CATALOG_NAME, $name)) {
```
PierreRambaud, line 47: Why rename CATEGORY_NAME into CATALOG_NAME?
A catalog is different than a category. More we're in Category context.

zuk3975: It is the same regex pattern. It was validated by isCatalogName in legacy before and the category belongs to catalog context, so I thought why not :thinking: . If this is not suitable, then this PR is might be not needed at all, as all duplicate regex patterns could be considered just coincidental and each be kept in corresponding domain without trying to reuse ?

PierreRambaud: Even if it's the same, the context is different.
A catalog is a list of items, and a category describe what kind of item is it.
If in the future, we need to update this regex pattern only for one case like we did for the Customer name, we're fucked!
We must separate regex, or otherwise, use a more generic pattern, or something stupid as `CATALOG_OR_CATEGORY_NAME`

zuk3975: Ok then this PR seems to be a bad idea.

matks: Eyyyyyh 😄 dont go so fast
The PR is a good idea for generic names, not specific names

sarjon: The idea behind this PR was good imo.

zuk3975: All of currently defined patterns in RegexPattern.php are generic ones, specific are kept in corresponding domain models, e.g. IsoCode or TagIETF. Category name isn't specific, its the same catalog_name :confused: . Idk maybe its confusing in some way... Anyway this concept needs to be modified further then to be accepted by all, but idk how.

PR #14578, opened 2019-07-10
install and enable theme in CLI

### LINKED ISSUE #14577: PrestaShop Console : Theme command improvements
Starting from PrestaShop 1.7.6 we can enable theme in console. The problem is that we need to install the theme first. Using the backoffice for installation is not suitable as we want to do all steps in console. If we just unzip the theme in themes directory, some important actions will not be executed, for example, installation of the modules from theme dependencies.
So it will be great to have the symfony command to install the theme from zip path, using the ThemeManager.

After when we enable the theme in console, by default, the theme will be enabled for the first shop. If we have multishop, we can't enable theme for other shops in console. So it can be useful to add the shop argument to ThemeEnablerCommand. It can be the optional parameter, where we can set the shop id, for which we want to enable the theme.

In the same time i think to submit pull request with these suggestions.

202-ecommerce:
| Questions     | Answers
| Branch?       | develop
| Description?  | Pull request add console command for install theme in CLI and add new argument "shop ID" to theme enabler command. 
| Type?         | new feature
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14577 
| How to test?  | Activate multishop. Launch the commands `php bin/console prestashop:theme:install theme_path` <br> and `php bin/console prestashop:theme:enable theme shop_id` in cli. Check in the prestashop that theme is installed and enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14578)


src/PrestaShopBundle/Command/ThemeEnablerCommand.php
```diff
@@ -75,8 +78,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $io = new SymfonyStyle($input, $output);
         $theme = $input->getArgument('theme');
+        $shop = $input->getArgument('shop');
         $this->init($input, $output);
 
+        if ($shop !== null) {
+            if (Validate::isLoadedObject(new Shop($shop))) {
```
PierreRambaud, line 85: store `new Shop($shop)` in a variable please :)

src/PrestaShopBundle/Command/ThemeInstallCommand.php
```diff
@@ -0,0 +1,85 @@
...
+        Context::getContext()->employee = new Employee();
+    }
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('prestashop:theme:install')
+            ->setDescription('Manage your themes via command line')
```
PierreRambaud:
```suggestion
            ->setDescription('Install your theme via command line')
```

mickaelandrieu:
Hello, thanks for the contribution

Can we preserve the same API than the [module command](https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Command/ModuleCommand.php#L36)?

So you will have only one command for all the themes manipulations :+1:

src/PrestaShopBundle/Command/ThemeInstallCommand.php
```diff
@@ -0,0 +1,85 @@
...
+final class ThemeInstallCommand extends ContainerAwareCommand
+{
+    /**
+     * @var int if the activation of the theme fails, return the right code
+     */
+    const RETURN_CODE_FAILED = 1;
+    /**
+     * {@inheritdoc}
+     */
+    protected function init(InputInterface $input, OutputInterface $output)
```
mickaelandrieu: for information @PrestaShop/prestashop-core-developers => this allows people to use Legacy classes into the commands.

I see an improvement for 1.7.7: listen to an event at the init (if we listen to the right [event](https://symfony.com/doc/3.4/components/console/events.html#the-consoleevents-command-event))

202-ecommerce:
> Hello, thanks for the contribution
> 
> Can we preserve the same API than the [module command](https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Command/ModuleCommand.php#L36)?
> 
> So you will have only one command for all the themes manipulations

Hello, Thanks for your suggestion, I will modify the code to have only one command like for the modules.

202-ecommerce:
I have added new command ThemeCommand for all the themes manipulations : install, enable, export, uninstall. For uninstall command i had to add missing function isThemeUsed in ThemeManager class. 

For old commands (ExportThemeCommand, ThemeEnablerCommand) i think to add deprecation notice.

src/Core/Addon/Theme/ThemeManager.php
```diff
@@ -132,6 +132,24 @@ public function uninstall($name)
...
+     * Check if the theme is used in at least one shop
+     *
+     * @param $name string theme name
+     *
+     * @return bool
+     */
+    public function isThemeUsed($name)
+    {
+        foreach (Shop::getShops() as $shop) {
+            if ($shop['theme_name'] == $name) {
```
PierreRambaud:
```suggestion
            if ($shop['theme_name'] === $name) {
```

src/PrestaShopBundle/Command/ThemeCommand.php
```diff
@@ -0,0 +1,168 @@
...
+use Employee;
+use Shop;
+use Validate;
+
+/**
+ * Class ThemeCommand run command actions in CLI
+ */
+class ThemeCommand extends ContainerAwareCommand
+{
+    private $allowedActions = array(
```
PierreRambaud: Must be a const

src/PrestaShopBundle/Command/ThemeCommand.php
```diff
@@ -0,0 +1,168 @@
...
+
+            return self::RETURN_CODE_FAILED;
+        }
+        $this->io->success(sprintf('Your theme has been correctly exported: "%s".', $path));
+    }
+
+    protected function executeGenericThemeAction($action, $argument)
+    {
+        /**
+         * @var \PrestaShop\PrestaShop\Core\Addon\Theme\ThemeManager
```
PierreRambaud: namespace instead

202-ecommerce: hello, i don't understand what actually i need to change here? the documentation? Could you explain please?

src/PrestaShopBundle/Command/ThemeCommand.php
202-ecommerce: > Must be a const

corrigé

src/PrestaShopBundle/Command/ThemeCommand.php
PierreRambaud: at the top of the file
```php
use PrestaShop\PrestaShop\Core\Addon\Theme\ThemeManager;
```

and 
```suggestion
         * @var ThemeManager
```

202-ecommerce: thanks, it's fixed

src/PrestaShopBundle/Command/ThemeCommand.php
```diff
@@ -0,0 +1,176 @@
...
+            ->addArgument('theme', InputArgument::REQUIRED, 'Theme or path to zip on which the action will be executed')
+            ->addArgument('shop', InputArgument::OPTIONAL, 'Shop id on which the action will be executed');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function init(InputInterface $input, OutputInterface $output)
+    {
+        require $this->getContainer()->get('kernel')->getRootDir() . '/../config/config.inc.php';
```
clotairer: Since PR #15515 this require generate a "Notice Constant _PS_SSL_PORT_ already defined".
We can remove this line witch is take in charge on console.php head script.

clotaire202: Since PR #15515 this require is not nécessary and throw a notice "Constant _PS_SSL_PORT_ already defined"

I think we need to remove it.

src/PrestaShopBundle/Command/ModuleCommand.php
```diff
@@ -83,7 +83,7 @@ protected function init(InputInterface $input, OutputInterface $output)
         $this->translator = $this->getContainer()->get('translator');
         $this->input = $input;
         $this->output = $output;
-        require $this->getContainer()->get('kernel')->getRootDir() . '/../config/config.inc.php';
+        require_once $this->getContainer()->get('kernel')->getRootDir() . '/../config/config.inc.php';
```
PierreRambaud: I think this like is not needed anymore :thinking:

src/PrestaShopBundle/Command/ThemeCommand.php
```diff
@@ -0,0 +1,175 @@
...
+                $this->executeExportThemeAction($theme);
+                break;
+            case 'install':
+                if (!filter_var($theme, FILTER_VALIDATE_URL)
+                    && !file_exists($theme)) {
+                    $this->io->error('Zip file doesn\'t exist or url is not valid');
+
+                    return self::RETURN_CODE_FAILED;
+                }
+                $this->executeGenericThemeAction($action, $theme);
```
PierreRambaud:
```suggestion
```

src/PrestaShopBundle/Command/ThemeCommand.php
```diff
@@ -0,0 +1,175 @@
...
+                break;
```
PierreRambaud:
```suggestion
```

202-ecommerce: Hello, why you suggest to remove these lines? we need to execute some action here...

src/PrestaShopBundle/Command/ModuleCommand.php
202-ecommerce: removed

src/PrestaShopBundle/Command/ThemeCommand.php
PierreRambaud: Because if $action is install, it will do the zip check and executeGenericThemeAction whatever happened.

```php
switch($action) {
    case 'uninstall':
        echo 'Oh noooo!!' . PHP_EOL;
        break;
    case 'install':
        echo 'Installation in progress' . PHP_EOL;
    default:
        echo 'Default ' . $action . PHP_EOL;
}
```
With `$action = 'uninstall'`:
```
$ php test.php 
Oh noooo!!
```

With `$action 'install' `: 
```
$ php test.php 
Installation in progress
Default install
```

With `$action = 'enable'`:
``` test.php 
Default enable
```

It prevent to use the same piece of code. Like having the same code in a try & a catch, use finnaly instead :D

PierreRambaud:
Hi @202-ecommerce  and @Progi1984 ,

Thanks!

Progi1984:
I take it and I will finish it.

PierreRambaud:
Seen with you, I'm closing it @Progi1984 
Feel free to create a new PR when you'll be able to redo the work

PR #14586, opened 2019-07-11
Implement Circuit Breaker for fetching blog news

### LINKED ISSUE #12827: CR - Page Login/Dashboard - Blog RSS feed API
This call gets the a xml files containing a list of articles with the corresponding links to the blog. These articles are displayed on the Dashboard in the "PrestaShop news" block.

Capture d’écran 2019-03-26 à 12 08 32

## Current behavior

**When is the call made?**
Asynchro

**At which frequency is it made?**
Every 24h or when not in the cache

**What API does it call?**
Prestashop API

**Why is it used for?**
To display the latest PS blog articles in the Dashboard

**impact on loading time**
Difficult to evaluate.

## Expected behavior

**At which frequency is it made?**
ISO - Every 24h or when not in the cache

**How long before the cache expires ?**
ISO - 24h

**How long should we wait to get the answer (timeout)?**
3 seconds

**If timeout, should we try again? How many times ?**
3 times

**If timeout on the retries, how long do we wait before trying again ?**
24h

**What do we do while we don't get an answer?**
Nothing - Currently if no answer the block displays a link to the PS blog "Find more news", we keep that.

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | Implement Circuit Breaker for fetching blog news
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #12827
| How to test?  | * Check if news are well displayed in backoffice<br>* Activate plane mode (no wifi, no network)<br>* Check if news are well displayed in backoffice (cache is used)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14586)


classes/NewsFetcher.php
```diff
@@ -0,0 +1,154 @@
...
+                'short_desc' => Tools::truncateString(strip_tags((string) $item->description), 150),
+                'link' => (string) $article_link,
+            ];
+            --$articles_limit;
+        }
+        $data['has_errors'] = false;
+
+        return $data;
+    }
+}
```
mickaelandrieu: you can use PHP CS Fixer to add a EOF line

classes/NewsFetcher.php
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * @param string $isoCode
+     */
+    public function __construct(
+        string $isoCode
+    ) {
+        $this->isoCode = $isoCode;
+
+        // Doctrine cache used for Guzzle and CircuitBreaker storage
```
mickaelandrieu: ping @jolelievre: I see a command pattern here, maybe can you introduce a "PrestaShopCircuitBreaker" with all that stuff" inside, don't you think ?

classes/NewsFetcher.php
Progi1984: :+1:

mickaelandrieu (dismissed):

classes/NewsFetcher.php
```diff
@@ -0,0 +1,154 @@
...
+use Doctrine\Common\Cache\FilesystemCache;
+use GuzzleHttp\Message\Request;
+use GuzzleHttp\Subscriber\Cache\CacheStorage;
+use GuzzleHttp\Subscriber\Cache\CacheSubscriber;
+use PrestaShop\CircuitBreaker\AdvancedCircuitBreakerFactory;
+use PrestaShop\CircuitBreaker\Contract\FactoryInterface;
+use PrestaShop\CircuitBreaker\FactorySettings;
+use PrestaShop\CircuitBreaker\Storage\DoctrineCache;
+
+class NewsFetcherCore
```
jolelievre: We want to get rid of the object model classes (all classes in the `classes` directory) It is some legacy code that we are stuck with, so we don't have a choice for now
But let's avoid adding some in it ^^ Besides object model classes are usually linked to persisted data in the database which is not the case here
And the `ClassnameCore` convention of these objects is used to allow the infamous *override* feature in PrestaShop which we also want to get rid of (or at least not encourage for now ^^)

So in this case you'd better create a service, and you can divide it into multiple sub services:
```yaml
# circuit_breaker.yml
services:
    prestashop.core.circuit_breaker.doctrine_cache:
        class: Doctrine\Common\Cache\FilesystemCache
        arguments:
            - "@=service('prestashop.adapter.environment').getCacheDir()/circuit_breaker"

    prestashop.core.circuit_breaker.guzzle.cache_storage:
        class: GuzzleHttp\Subscriber\Cache\CacheStorage
        arguments:
             - "@prestashop.core.circuit_breaker.doctrine_cache"
             - 86400
    
    prestashop.core.circuit_breaker.guzzle.cache_subscriber:
        class: GuzzleHttp\Subscriber\Cache\CacheSubscriber

    prestashop.core.circuit_breaker.storage:
        class: PrestaShop\CircuitBreaker\Storage\DoctrineCache
        arguments:
            - "@prestashop.core.circuit_breaker.doctrine_cache"

    prestashop.core.circuit_breaker.advanced_factory:
        class: PrestaShop\CircuitBreaker\AdvancedCircuitBreakerFactory
```

These are the basic services you will need to create different instances of circuit breakers. Maybe they will we needed in legacy environment, not just Symfony so be careful where you define them (more details about service scopes here: https://devdocs.prestashop.com/1.7/modules/concepts/services/#services-in-legacy-environment

Then once you have those basics you can manage your circuit breaker settings via services as well:

```yaml
# news_services.yml
services:
    prestashop.core.news.circuit_breaker.settings:
        class: PrestaShop\CircuitBreaker\FactorySettings
        # Here I use hard coded values but they should be in a class constants
        arguments:
            - 3
            - 3
            - 0
        calls:
            - ['setThreshold', [86400]]
            - ['setStrippedFailures', [3]]
            - ['setStrippedTimeout', [3]]
            - ['setStorage', ['@prestashop.core.circuit_breaker.storage']]
            - ['setClientOptions', ['method': 'GET', 'subscribers': ['@prestashop.core.circuit_breaker.guzzle.cache_subscriber']]]

    prestashop.core.news.circuit_breaker:
        factory: ["@prestashop.core.circuit_breaker.advanced_factory", create]
        arguments:
            - "@prestashop.core.news.circuit_breaker.settings"

    prestashop.core.news.fetcher:
        class: "RightNamespace\NewsFetcher"
        arguments:
            - "@prestashop.core.news.circuit_breaker"
```

Y hasta!! Now all you need to do is create your `NewsFetcher` service class (which you already have) without caring about the circuit breaker instanciation and focusing on parsing your datas. And this will be easier to unit test as well

classes/NewsFetcher.php
jolelievre: I'm not sure what you mean, are you refering to a command/handler pattern like in CQRS?

jolelievre (changes requested):
Hi @Progi1984,
nice job but as we discussed I think you can make this first circuit breaker integration a bit sexier using Symfony and defining basic services that will be reused in other use cases.

controllers/admin/AdminDashboardController.php
```diff
@@ -448,9 +448,8 @@ public function ajaxProcessSetSimulationMode()
 
     public function ajaxProcessGetBlogRss()
     {
-        $newsFetcher = new NewsFetcher(
-            $this->context->language->iso_code
-        );
+        $newsFetcher = $this->get('prestashop.core.news.provider');
+        $newsFetcher->setIsoCode($this->context->language->iso_code);
```
jolelievre: Maybe it would we simpler to use isoCode as a function argument instead of injecting it?
`$return = $newsFetcher->getData($this->context->language->iso_code);`

src/PrestaShopBundle/Cache/Factory/CacheSubscriberFactory.php
```diff
@@ -0,0 +1,40 @@
...
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShopBundle\Cache\Factory;
+
+use GuzzleHttp\Message\Request;
+use GuzzleHttp\Subscriber\Cache\CacheStorageInterface;
+use GuzzleHttp\Subscriber\Cache\CacheSubscriber;
+
+class CacheSubscriberFactory
```
jolelievre: Make this class `final` (we have no interest for developers to override it, so it will allow us to modify/improve it more easily if needed) Also you need to add PHPDoc on it (every classes/interfaces we add must contain a PHPDoc describing its purpose).

src/PrestaShopBundle/Resources/config/services/core/circuit_breaker.yml
```diff
@@ -0,0 +1,26 @@
+# circuit_breaker.yml
+services:
+  prestashop.core.circuit_breaker.doctrine_cache:
+    class: Doctrine\Common\Cache\FilesystemCache
+    arguments:
+      - '@=service("prestashop.adapter.environment").getCacheDir() ~ "/circuit_breaker"'
```
jolelievre, line 9: I was absolutely not sure about the syntax but does this work? ^^

src/PrestaShopBundle/Resources/config/services/core/circuit_breaker.yml
```diff
@@ -0,0 +1,26 @@
...
+    arguments:
+      - "@prestashop.core.circuit_breaker.guzzle.cache_storage"
+
+  prestashop.core.circuit_breaker.storage:
+    class: PrestaShop\CircuitBreaker\Storage\DoctrineCache
+    arguments:
+      - "@prestashop.core.circuit_breaker.doctrine_cache"
+
+  prestashop.core.circuit_breaker.advanced_factory:
+    class: PrestaShop\CircuitBreaker\AdvancedCircuitBreakerFactory
```
jolelievre: Missing EOL at the end of file

src/PrestaShopBundle/Resources/config/services/core/common.yml
```diff
@@ -0,0 +1,2 @@
+imports:
+  - { resource: ./circuit_breaker.yml }
```
jolelievre: Missing EOL

src/PrestaShopBundle/Resources/config/services/core/circuit_breaker.yml
```diff
@@ -0,0 +1,26 @@
+# circuit_breaker.yml
+services:
+  prestashop.core.circuit_breaker.doctrine_cache:
```
jolelievre, line 6: You should add:
```yaml
  _defaults:
    public: true
```

It's not good practice but for now all our services are usually public

src/PrestaShopBundle/Resources/config/services/core/circuit_breaker.yml
```diff
@@ -0,0 +1,26 @@
...
+    class: Doctrine\Common\Cache\FilesystemCache
+    arguments:
+      - '@=service("prestashop.adapter.environment").getCacheDir() ~ "/circuit_breaker"'
+
+  prestashop.core.circuit_breaker.guzzle.cache_storage:
+    class: GuzzleHttp\Subscriber\Cache\CacheStorage
+    arguments:
+      - "@prestashop.core.circuit_breaker.doctrine_cache"
+      - 86400
```
jolelievre, line 15: This should probably be a const or a parameter Either from parameters.yml but if we follow "PrestaShop style" this should probably be stored in our `Configuration`
That said, fi we go further into this idea, it would require a form to edit this value. So maybe it's overkill for now 🤔

src/PrestaShopBundle/Resources/config/services/core/news.yml
```diff
@@ -0,0 +1,33 @@
+services:
+  _defaults:
+    public: true
+
+  prestashop.core.news.circuit_breaker.settings:
+    class: PrestaShop\CircuitBreaker\FactorySettings
+    # Here I use hard coded values but they should be in a class constants
```
jolelievre: You can remove the comment now 😉
Nice for the const. By the way threshold is the third argument of the constructor so you can use it directly instead of injecting it via `setThreshold` (my bad I didn't notice this mistake in my code)

src/PrestaShopBundle/Resources/config/services/core/news.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  prestashop.core.news.provider:
+    class: PrestaShopBundle\Service\DataProvider\Admin\NewsProvider
+    arguments:
+      - "@prestashop.core.news.circuit_breaker"
+      - "@prestashop.adapter.data_provider.country"
+      - "@prestashop.adapter.tools"
+      - "@prestashop.adapter.legacy.configuration"
+      - "@prestashop.adapter.validate"
+      - "@=service('prestashop.adapter.legacy.context').getContext().mode"
```
jolelievre: Missing EOL

src/PrestaShopBundle/Service/DataProvider/Admin/NewsProvider.php
```diff
@@ -24,79 +24,87 @@
...
+     */
+    private $circuitBreaker;
+
+    /**
+     * @var Configuration
+     */
+    private $configuration;
+
+    /**
+     * @var Validate
```
jolelievre: `@var int`

classes/NewsFetcher.php
Progi1984: Done

src/Adapter/Tools.php
```diff
@@ -244,4 +244,46 @@ public function getAdminImageUrl($fileName, $escapeHtmlEntities = false)
...
+    {
+        return LegacyTools::displayDate($date, $id_lang, $full, $separator);
+    }
+
+    /**
+     * @see LegacyTools::htmlentitiesUTF8()
+     *
+     * @return string
+     */
+    public function htmlentitiesUTF8($string, $type = ENT_QUOTES)
```
jolelievre: We need to avoid making this class bigger and bigger, or else it will simply end up as a copy of the legacy Tools class ^^ Besides this class is merely a "bandage" to avoid using the Tools class directly, but it makes way too many things and will need to be refacto some day

So try to keep the functions in this service to the minimum, for example `htmlentitiesUTF8` and `strtoupper` you can probably use the native PHP functions (based on what the Tools methods do). Although for `displayDate` and `truncateString` you might not have much choice..

src/PrestaShopBundle/Service/DataProvider/Admin/NewsProvider.php
```diff
@@ -24,79 +24,87 @@
...
-use GuzzleHttp\Message\Request;
-use GuzzleHttp\Subscriber\Cache\CacheStorage;
-use GuzzleHttp\Subscriber\Cache\CacheSubscriber;
-use PrestaShop\CircuitBreaker\AdvancedCircuitBreakerFactory;
-use PrestaShop\CircuitBreaker\Contract\FactoryInterface;
-use PrestaShop\CircuitBreaker\FactorySettings;
-use PrestaShop\CircuitBreaker\Storage\DoctrineCache;
-
-class NewsFetcherCore
+namespace PrestaShopBundle\Service\DataProvider\Admin;
```
jolelievre: I think it should rather be in the Adapter namespace (see https://devdocs.prestashop.com/1.7/development/architecture/understanding-src-folder/)
Because it's a business logic for PrestaShop but it still has a legacy/adapter dependencies, so we can't have it in the Core namespace
And PrestaShopBundle is used for framework specific services/classes/...

Also if you move it, logically you will have to move its service definition as well, if the class is not in the core namespace its service definition should be in adapter service definitions
We accept service definitions with adapter injections, but only when it matches a Core interface

controllers/admin/AdminDashboardController.php
```diff
@@ -448,62 +448,14 @@ public function ajaxProcessSetSimulationMode()
...
-        }
-        die(json_encode($return));
+        $newsFetcher = $this->get('prestashop.core.news.provider');
+        $newsFetcher->setIsoCode($this->context->language->iso_code);
+        $return = $newsFetcher->getData();
+
+        // Response
+        header('Content-Type: application/json');
+        $this->ajaxRender(json_encode($return));
+        exit();
```
jolelievre: You should **never** use exit in any code, this is a cancer that has spread in legacy code and causes us problems ^^ Let's not help it grow 😉

jolelievre (commented):
This is getting better! Noce job
A few feedback to address again ^^

And I think you can try adding unit tests for NewsProvider (you just need to mock the circuit breaker, that's the advantage of injecting it this way)

controllers/admin/AdminDashboardController.php
Progi1984: :ok:

src/PrestaShopBundle/Cache/Factory/CacheSubscriberFactory.php
Progi1984: :ok:

src/PrestaShopBundle/Resources/config/services/core/circuit_breaker.yml
Progi1984: Yes, it's the only solution for concatenate a method call and a string

src/PrestaShopBundle/Resources/config/services/core/circuit_breaker.yml
Progi1984: :ok:

src/PrestaShopBundle/Resources/config/services/core/common.yml
Progi1984: :ok:

src/PrestaShopBundle/Resources/config/services/core/circuit_breaker.yml
Progi1984: :ok:

src/PrestaShopBundle/Resources/config/services/core/news.yml
Progi1984: Sorry :pray: for the comment.... 

:ok:

src/PrestaShopBundle/Resources/config/services/core/news.yml
Progi1984: :ok:

src/PrestaShopBundle/Service/DataProvider/Admin/NewsProvider.php
Progi1984: :ok:

src/Adapter/Tools.php
Progi1984: :ok:

src/PrestaShopBundle/Resources/config/services/core/circuit_breaker.yml
Progi1984: I consider it as overkill and don't change it :)

src/PrestaShopBundle/Service/DataProvider/Admin/NewsProvider.php
Progi1984: Done

controllers/admin/AdminDashboardController.php
Progi1984: I correct it.

Progi1984: Fixed

controllers/admin/AdminDashboardController.php
```diff
@@ -446,64 +446,19 @@ public function ajaxProcessSetSimulationMode()
         die('k' . Configuration::get('PS_DASHBOARD_SIMULATION') . 'k');
     }
 
-    public function ajaxProcessGetBlogRss()
+    /**
+     * Returns last news from the blog
+     *
+     * @throws PrestaShopException
+     */
+    public function displayAjaxGetBlogRss()
```
jolelievre, line 454: 👍

src/Adapter/News/NewsDataProvider.php
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Adapter\News;
+
+use ContextCore;
```
jolelievre:
```suggestion
use Context;
```
Prefer using classes without the Core suffix, or the override system won't work. Even though we don't want to promote the use of override we should remain consistent and use the real class names instead of the *Core ones.

src/Adapter/News/NewsDataProvider.php
```diff
@@ -0,0 +1,173 @@
...
+
+        foreach ($rss->channel->item as $item) {
+            if ($articles_limit == 0
+                || !$this->validate->isCleanHtml((string) $item->title)
+                || !$this->validate->isCleanHtml((string) $item->description)
+                || !isset($item->link, $item->title)) {
+                break;
+            }
+            $analytics_params['utm_content'] = 'download';
+            if (in_array($this->contextMode, array(ContextCore::MODE_HOST, ContextCore::MODE_HOST_CONTRIB))) {
```
jolelievre: here as well of course replace `ContextCore` by `Context`

src/Adapter/News/NewsDataProvider.php
```diff
@@ -0,0 +1,173 @@
...
+            'utm_medium' => 'rss',
+            'utm_campaign' => 'back-office-' . $shop_default_iso_country,
+        ];
+
+        foreach ($rss->channel->item as $item) {
+            if ($articles_limit == 0
+                || !$this->validate->isCleanHtml((string) $item->title)
+                || !$this->validate->isCleanHtml((string) $item->description)
+                || !isset($item->link, $item->title)) {
+                break;
```
jolelievre, line 143: Careful, you use break for each conditions. The `break` makes sense when you reach the limit, but maybe in the other cases you should use `continue` or else any invalid item will stop the loop and you can't reach the expected limit.

jolelievre: also, maybe you could use `empty($item->link, $item->title)` instead of `!isset`

src/Adapter/News/NewsDataProvider.php
```diff
@@ -0,0 +1,173 @@
...
+            $article_link = (string) $item->link . '?' . http_build_query($analytics_params);
+
+            $url_query = parse_url($item->link, PHP_URL_QUERY);
+            parse_str($url_query, $link_query_params);
+            if ($link_query_params) {
+                $full_url_params = array_merge($link_query_params, $analytics_params);
+                $base_url = explode('?', (string) $item->link);
+                $base_url = (string) $base_url[0];
+                $article_link = $base_url . '?' . http_build_query($full_url_params);
+            }
```
jolelievre: manage the `else` case to integrate `$analytics_params` even if the link has no parameters

src/Adapter/News/NewsDataProvider.php
```diff
@@ -0,0 +1,173 @@
...
+            $analytics_params['utm_content'] = 'download';
+            if (in_array($this->contextMode, array(ContextCore::MODE_HOST, ContextCore::MODE_HOST_CONTRIB))) {
+                $analytics_params['utm_content'] = 'cloud';
+            }
+            $article_link = (string) $item->link . '?' . http_build_query($analytics_params);
+
+            $url_query = parse_url($item->link, PHP_URL_QUERY);
+            parse_str($url_query, $link_query_params);
+            if ($link_query_params) {
```
jolelievre:
```suggestion
            if (!empty($link_query_params)) {
```

tests/Unit/Adapter/News/NewsDataProviderTest.php
```diff
@@ -0,0 +1,196 @@
...
+            ContextCore::MODE_STD
+        );
+        $expectedJson = $generator->getData($expectedLocale);
+
+        $this->assertInternalType('array', $expectedJson);
+        $this->assertArrayHasKey('has_errors', $expectedJson);
+        $this->assertFalse($expectedJson['has_errors']);
+        $this->assertArrayHasKey('rss', $expectedJson);
+        $this->assertInternalType('array', $expectedJson['rss']);
+        $this->assertCount(NewsDataProvider::NUM_ARTICLES, $expectedJson['rss']);
```
jolelievre, line 114: you should test the format of the items you return, and maybe check the url parameters, since the most complex thing in the provider is building the url with campaign parameters (just checking they are here should be enough)

jolelievre (commented):
Nice job @Progi1984 
Some little feedback left, but we're almost there!

src/Adapter/News/NewsDataProvider.php
Progi1984: :+1:

src/Adapter/News/NewsDataProvider.php
Progi1984: Done

src/Adapter/News/NewsDataProvider.php
Progi1984: :+1:

src/Adapter/News/NewsDataProvider.php
Progi1984: :+1:

src/Adapter/News/NewsDataProvider.php
Progi1984: @jolelievre So stupid I was. Thanks.

tests/Unit/Adapter/News/NewsDataProviderTest.php
Progi1984: :ok:

src/Adapter/News/NewsDataProvider.php
```diff
@@ -140,26 +140,23 @@ public function getData($isoCode)
             if ($articles_limit == 0
                 || !$this->validate->isCleanHtml((string) $item->title)
                 || !$this->validate->isCleanHtml((string) $item->description)
-                || !isset($item->link, $item->title)) {
-                break;
+                || empty($item->link)
+                || empty($item->title)) {
+                continue;
```
jolelievre, line 149: Special case for `$articles_limit == 0` you can `break`, it will be more efficient (although the result will be the same)

jolelievre (commented):
Haha one last change :D

Matt75:
This block stay inside Core or should go to MBO ? 🤔

Progi1984:
Ping @colinegin for answering to @Matt75

src/Adapter/News/NewsDataProvider.php
Progi1984: Done

jolelievre:
@Matt75 I think this will go into the core for now, so that we have the basic circuit breaker services available. But about the specific management of news, this could be migrated to mbo later I guess.

jolelievre (dismissed):

sarahdib:
@Progi1984 when the wifi is disable the block is display but not like when I have wifi.

image

Progi1984:
@sarahdib Sorry, I have bad read a specification. It's corrected : the block must be displayed with no news inside.

sarahdib:
@Progi1984 ok just waiting for the review :) before the test

src/Adapter/News/NewsDataProvider.php
```diff
@@ -0,0 +1,175 @@
...
+     */
+    public function getData($isoCode)
+    {
+        $data = ['has_errors' => true, 'rss' => []];
+        $apiUrl = $this->configuration->get('_PS_API_URL_');
+
+        $blogXMLResponse = $this->circuitBreaker->call($apiUrl . '/rss/blog/blog-' . $isoCode . '.xml');
+
+        if (empty($blogXMLResponse)) {
+            $data['has_errors'] = false;
```
jolelievre, line 121: Hi, this last modification seems weird. If no answer is received (nor cached) you should return `has_errors = true` no?
I get that when the connection or the service is not available the block must not be displayed. But then it should be managed in the front when an error is detected. Or the `has_errors` field is useless since it's always set to false.

Progi1984: @jolelievre I just implement the rule :
> **What do we do while we don't get an answer?**
> Nothing - Currently if no answer the block displays a link to the PS blog "Find more news", we keep that.

For that, `$data['has_errors']` must be at `false` in the case where I don't get an answer (so `$blogXMLResponse = ''`). If there are XML errors, then `$data['has_errors'] = true`.

jolelievre: Alright, perfect ;)

jolelievre (approved):
Thanks @Progi1984

khouloudbelguith:
Hi @Progi1984,
when the wifi is disabled, The `PRESTASHOP NEWS` block is OK.
image
But when we enable the wifi again, there is no news displayed.
image

I tried when the cache is enabled & cache disabled => Same behavior.

We need to clear cache from the BO => Performance page => to show the exact links
https://drive.google.com/file/d/14ngxoJTQlSPP44JOQmVCvTyxMnbQ1KFc/view

Thanks!

khouloudbelguith:
Hi @Progi1984,

As discussed with @jolelievre, this a normal behavior
>If timeout on the retries, how long do we wait before trying again?
24h

I checked with `OPEN_THRESHOLD_SECONDS = 120seconds` => OK.

Thanks!

jolelievre:
Thanks @Progi1984

PR #13733, opened 2019-05-09
Migrate login page

rokaszygmantas:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Login page migration to symfony.
| Type?         | improvement
| Category?     | BO
| BC breaks?    | yes (legacy security system removed), `actionAdminLoginControllerSetMedia` hook removed
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/12953
| How to test?  | Compare to legacy login page. List with edge cases is below.

Edge cases to test:

1.  This PR includes login form, together with forgot password and reset password forms. After submitting "forgot password" form you should receive an email with a link to "reset password" form. 
2. Password can be reset only every X amount of minutes, where X is defined by a setting in Advanced params->Team->Employees (form below the employees list) https://prnt.sc/o4bkec
You should get an error message if you reset too often. However you can receive "forgot password" emails without a limit (same as it was in legacy page).
3. If you access any BO page without being logged in, you should be redirected to login page. Then, after logging in, you should be redirected back to that previously requested page.
4. Having enabled SSL and going to back office with **http://** will show a warning. It will show a different warning if you are visiting from a localhost IP or from one of maintenance IP addresses.
5. Having not deleted the `install/` directory from PrestaShop root will show an error and hide the login form.
6. Having not renamed the `admin/` folder after PS installation will automatically rename it to a random name upon visiting login page. If something fails during renaming you will see an error message with a hint to rename it manually. You cannot see the login form while the admin folder is not renamed.
7. "Stay logged in" will keep you logged in for 1 year. Without it, you will be logged out after 15 minutes of inactivity. 
8. Please test the authentication part very carefully as it was changed from legacy to symfony authentication. :slightly_smiling_face: 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/13733)

PR #14593, opened 2019-07-11
First POM implementation

SimonGrn:
| Questions     | Answers
| Branch?       | develop
| Description?  | First implementation of the Page object pattern with puppeteer.
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | `node_modules/mocha/bin/mocha test/smoke/main`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14593)

### DIFF
--- a/tests/puppeteer/package.json
+++ b/tests/puppeteer/package.json
@@ -6,7 +6,7 @@
     "test": "test"
   },
   "scripts": {
-    "test": "node test/linkchecker.js",
+    "linkchecker": "node test/linkchecker.js",
     "lint": "eslint  --fix --ignore-path .gitignore .",
     "specific-test": "node_modules/mocha/bin/mocha"
   },
--- a/tests/puppeteer/package.json
+++ b/tests/puppeteer/package.json
@@ -28,6 +28,7 @@
     "chai-string": "^1.5.0",
     "eslint-config-prestashop": "0.0.2",
     "mocha": "^6.1.4",
+    "mochawesome": "^4.0.1",
     "puppeteer": "^1.18.1"
   },
   "devDependencies": {
--- a/tests/puppeteer/package.json
+++ b/tests/puppeteer/package.json
@@ -6,8 +6,9 @@
     "test": "test"
   },
   "scripts": {
-    "linkchecker": "node test/linkchecker.js",
-    "lint": "eslint  --fix --ignore-path .gitignore ."
+    "test": "node test/linkchecker.js",
+    "lint": "eslint  --fix --ignore-path .gitignore .",
+    "specific-test": "node_modules/mocha/bin/mocha"
   },
   "repository": {
     "type": "git",
@@ -23,7 +24,10 @@
   },
   "homepage": "https://github.com/PrestaShop/PrestaShop#readme",
   "dependencies": {
+    "chai": "^4.2.0",
+    "chai-string": "^1.5.0",
     "eslint-config-prestashop": "0.0.2",
+    "mocha": "^6.1.4",
     "puppeteer": "^1.18.1"
   },
   "devDependencies": {
--- a/tests/puppeteer/package.json
+++ b/tests/puppeteer/package.json
@@ -22,7 +22,7 @@
   "bugs": {
     "url": "https://github.com/PrestaShop/PrestaShop/issues"
   },
-  "homepage": "https://github.com/PrestaShop/PrestaShop#readme",
+  "homepage": "https://github.com/PrestaShop/PrestaShop/tree/develop/tests/puppeteer#readme",
   "dependencies": {
     "chai": "^4.2.0",
     "chai-string": "^1.5.0",
--- a/tests/puppeteer/pages/BO/BO_login.js
+++ b/tests/puppeteer/pages/BO/BO_login.js
@@ -1,0 +1,36 @@
+const Page = require('../page');
+
+module.exports = class BO_login extends Page {
+
+  constructor() {
+    super();
+
+    this.title = 'Dashboard • PrestaShop';
+
+    this.BO_login_email_input = "#email";
+    this.BO_login_password_input = "#passwd";
+    this.BO_login_submitLogin_button = "#submit_login";
+    this.BO_login_stayLoggedIn_checkbox = "#stay_logged_in";
+
+    this.BO_login_forgotPassword_link = "#a.show-forgot-password";
+    this.BO_login_emailForgot_input = "#email_forgot";
+    this.BO_login_sendResetLink_button = "#button[name=submitLogin].btn-default";
+  }
+
+  /*
+  Methods
+   */
+
+  async login(email, passwd) {
+    await global.page.type(this.BO_login_email_input, email);
+    await global.page.type(this.BO_login_password_input, passwd);
+    await global.page.click(this.BO_login_submitLogin_button);
+    await global.page.waitForNavigation();
+  }
+
+  async checkPageTitle() {
+    let currentTitle = await global.page.title();
+    await expect(currentTitle).to.equal(this.title);
+  }
+
+};
--- a/tests/puppeteer/pages/BO/BO_login.js
+++ b/tests/puppeteer/pages/BO/BO_login.js
@@ -1,6 +1,6 @@
-const Page = require('../page');
+const CommonPage = require('../commonPage');
 
-module.exports = class BO_login extends Page {
+module.exports = class BO_login extends CommonPage {
 
   constructor(page) {
     super(page);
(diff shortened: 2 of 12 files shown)


tests/puppeteer/test/utils/helpers.js
```diff
@@ -0,0 +1,31 @@
...
+
+const puppeteer = require('puppeteer');
+
+global.test = (name, instructions) => it(name, () => {
+  return instructions().catch();
+});
+
+global.scenario = (name, tests, close = false) =>
+  describe(name, async () => {
+    before(async () => {
```
boubkerbribri: Maybe add an await to before function

tests/puppeteer/test/utils/helpers.js
```diff
@@ -0,0 +1,31 @@
...
+      global.browser = await puppeteer.launch({
+        headless: global.HEADLESS,
+        //slowMo: 100,
```
boubkerbribri: Can you delete slowMo comment here please ?

tests/puppeteer/test/utils/globals.js
```diff
@@ -0,0 +1,7 @@
+global.URL_FO = process.env.URL_FO || 'http://localhost/prestashop/';
+global.URL_BO = process.env.URL_BO || URL_FO + 'admin-dev/';
+global.EMAIL = process.env.LOGIN || 'demo@prestashop.com';
+global.PASSWD = process.env.PASSWD || 'prestashop_demo';
+global.HEADLESS = process.env.HEADLESS || true;
+global.browser = null;
+global.page = null;
```
boubkerbribri: I think that Global args could be used in linkchecker.js to.

SimonGrn: Are we sure we're gonna keep this folder architecture ? Because I need to require the `globals.js` file and it might move...

boubkerbribri: I'm not so sure, We can wait until defining architecture is finished.

tests/puppeteer/test/utils/helpers.js
SimonGrn: Je viens de tester, mocha n'aime pas ça du tout et me met `0 passing` donc on oublie :-D

tests/puppeteer/test/utils/helpers.js
SimonGrn: Done !

tests/puppeteer/test/utils/globals.js
SimonGrn: I added it anyway, that's a good idea. We can always change it later, it's a relative path

tests/puppeteer/test/utils/helpers.js
boubkerbribri: OK

tests/puppeteer/package.json
```diff
@@ -23,7 +24,10 @@
   },
   "homepage": "https://github.com/PrestaShop/PrestaShop#readme",
```
boubkerbribri: Maybe change the homePage to https://github.com/PrestaShop/PrestaShop/tree/develop/tests/puppeteer#readme

SimonGrn: Done, thanks !

tests/puppeteer/pages/page.js
```diff
@@ -0,0 +1,14 @@
+module.exports = class Page {
+
+  constructor(page) {
+  }
+
+  async getPageTitle() {
+    return global.page.title();
```
PierreRambaud: Using global is a bad practice, use dependency injection instead.
```js
module.exports = class Page {

  constructor(page) {
      this.page = page;
  }

  async getPageTitle() {
    return this.page.title();
   }

  async goTo(URL) {
    await this.page.goto(URL);
  }
};
```

It will be easier to maintain

tests/puppeteer/pages/BO/BO_login.js
```diff
@@ -0,0 +1,31 @@
+const Page = require('../page');
+
+module.exports = class BO_login extends Page {
+
+  constructor() {
+    super();
+
+    this.BO_login_pageTitle = 'Dashboard • PrestaShop';
```
PierreRambaud: You're already in `BO_Login` context you can remove each occurrences of this

SimonGrn: I wanted to, but PHPStorm is *not* happy about this at all. It makes me do this : 
image

tests/puppeteer/pages/BO/BO_login.js
```diff
@@ -0,0 +1,31 @@
...
+    this.BO_login_emailForgot_input = "#email_forgot";
+    this.BO_login_sendResetLink_button = "#button[name=submitLogin].btn-default";
+  }
+
+  /*
+  Methods
+   */
+
+  async login(email, passwd) {
+    await global.page.type(this.BO_login_email_input, email);
```
matks: Can we avoid `global` variables and use Dependency Injection instead ?

Global variables can lead to crazy buggy situations when one part of the code modifies it and the other one depends on it _and you have no clue who did this_ 😛 and also are a bit of code smell as it promotes lazy variables sharing.

Dependency Injection will protect you from that 😉

tests/puppeteer/pages/BO/BO_login.js
PierreRambaud: I mean:

```js
this.pageTitle = 'Dasho..';
this.emailInput = ...
```
And if you can be consistent and use camelcase or snake case only :)

tests/puppeteer/pages/page.js
SimonGrn: We kinda have a problem with this implementation. I'd love to use DI of course, but our `page` object is created outside, in `helpers.js` which is automatically pulled by Mocha. AFAIK we can't get it from there in our main script.

tests/puppeteer/pages/BO/BO_login.js
SimonGrn: Oh right, the naming. In our test script, we will have a *lot* of page objects and selectors, so it seemed like a good idea to name all selectors with the page they're from. Even if they're prefixed with the page object's object.

tests/puppeteer/pages/page.js
SimonGrn: We changed the way we handle the `page` object. Probably better now.

tests/puppeteer/pages/BO/BO_login.js
SimonGrn: We changed the way we name selectors and used camelCase everywhere.

tests/puppeteer/pages/BO/BO_login.js
SimonGrn: Resolved, thanks !

tests/puppeteer/pages/BO/BO_login.js
```diff
@@ -0,0 +1,31 @@
...
+
+  constructor(page) {
+    super(page);
+
+    this.pageTitle = 'Dashboard • PrestaShop';
+
+    this.emailInput = "#email";
+    this.passwordInput = "#passwd";
+    this.submitLoginButton = "#submit_login";
+    this.stayLoggedInCheckbox = "#stay_logged_in";
```
rokaszygmantas: Will need to update the selectors when the login page is migrated, because some of them will be changed

PR for refrence: https://github.com/PrestaShop/PrestaShop/pull/13733

boubkerbribri: @rokaszygmantas can you please tell us what selectors will be changed and what's the new values of it ?
Thanks !

SimonGrn: @rokaszygmantas Are you sure these selectors need to be changed ? We need an ID to target inputs efficiently, and every selector change after 1760 will be labelled at critical bug and will need to be fixed ASAP.

rokaszygmantas: @SimonGrn it's not mandatory to change them during the login page migration, so if that brings any problems we can stick to the old ones :slightly_smiling_face:

SimonGrn: @rokaszygmantas if you can, that would be really nice for us and for maintenance to add IDs to every interesting element in the page: inputs, buttons, selectors, etc. That will make our job really easy and make the tests better.
Thanks !

matks (approved):
A few things to fix, that can be fixed in future PRs:
- naming tests/puppeteer/test is not great (duplicate test word)
- still a few global items

Maybe you can create issues not to forget them 😉?

PR #14601, opened 2019-07-11
Revert removed templates for mail dynamic contents, order mails mainly, create MailPartialTemplateRenderer to manage this feature

jolelievre:
| Questions     | Answers
| Branch?       | develop
| Description?  | Revert removed templates for mail dynamic contents, order mails mainly, create MailPartialTemplateRenderer to manage this feature. Also fix the HookDispatcher which ignored events with any uppercase
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | Test an order and check that the email is correctly sent (with the product list). You can also go to Design > Email Theme and check the `order_conf` preview

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14601)

### DIFF
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -36,6 +36,9 @@ abstract class PaymentModuleCore extends Module
 
     const DEBUG_MODE = false;
 
+    /** @var MailPartialTemplateRenderer */
+    protected $partialRenderer;
+
     public function install()
     {
         if (!parent::install()) {
@@ -879,6 +882,18 @@ abstract class PaymentModuleCore extends Module
     }
 
     /**
+     * @return MailPartialTemplateRenderer
+     */
+    protected function getPartialRenderer()
+    {
+        if (!$this->partialRenderer) {
+            $this->partialRenderer = new MailPartialTemplateRenderer($this->context->smarty);
+        }
+
+        return $this->partialRenderer;
+    }
+
+    /**
      * Fetch the content of $template_name inside the folder
      * current_theme/mails/current_iso_lang/ if found, otherwise in
      * mails/current_iso_lang.
@@ -896,9 +911,7 @@ abstract class PaymentModuleCore extends Module
             return '';
         }
 
-        $partialRenderer = new MailPartialTemplateRenderer($this->context);
-
-        return $partialRenderer->render($template_name, $var);
+        return $this->getPartialRenderer()->render($template_name, $this->context->language, $var);
     }
 
     protected function createOrderFromCart(
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -24,6 +24,7 @@
  * International Registered Trademark & Property of PrestaShop SA
  */
 use PrestaShop\PrestaShop\Adapter\StockManager;
+use PrestaShop\PrestaShop\Adapter\MailTemplate\MailPartialTemplateRenderer;
 
 abstract class PaymentModuleCore extends Module
 {
@@ -895,22 +896,9 @@ abstract class PaymentModuleCore extends Module
             return '';
         }
 
-        $pathToFindEmail = array(
-            _PS_THEME_DIR_ . 'mails' . DIRECTORY_SEPARATOR . $this->context->language->iso_code . DIRECTORY_SEPARATOR . $template_name,
-            _PS_THEME_DIR_ . 'mails' . DIRECTORY_SEPARATOR . 'en' . DIRECTORY_SEPARATOR . $template_name,
-            _PS_MAIL_DIR_ . $this->context->language->iso_code . DIRECTORY_SEPARATOR . $template_name,
-            _PS_MAIL_DIR_ . 'en' . DIRECTORY_SEPARATOR . $template_name,
-        );
+        $partialRenderer = new MailPartialTemplateRenderer($this->context);
 
-        foreach ($pathToFindEmail as $path) {
-            if (Tools::file_exists_cache($path)) {
-                $this->context->smarty->assign('list', $var);
-
-                return $this->context->smarty->fetch($path);
-            }
-        }
-
-        return '';
+        return $partialRenderer->render($template_name, $var);
     }
 
     protected function createOrderFromCart(
--- a/mails/_partials/order_conf_product_list.tpl
+++ b/mails/_partials/order_conf_product_list.tpl
@@ -1,0 +1,136 @@
+{**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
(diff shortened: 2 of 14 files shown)


classes/PaymentModule.php
```diff
@@ -895,22 +896,9 @@ protected function getEmailTemplateContent($template_name, $mail_type, $var)
             return '';
         }
 
-        $pathToFindEmail = array(
-            _PS_THEME_DIR_ . 'mails' . DIRECTORY_SEPARATOR . $this->context->language->iso_code . DIRECTORY_SEPARATOR . $template_name,
-            _PS_THEME_DIR_ . 'mails' . DIRECTORY_SEPARATOR . 'en' . DIRECTORY_SEPARATOR . $template_name,
-            _PS_MAIL_DIR_ . $this->context->language->iso_code . DIRECTORY_SEPARATOR . $template_name,
-            _PS_MAIL_DIR_ . 'en' . DIRECTORY_SEPARATOR . $template_name,
-        );
+        $partialRenderer = new MailPartialTemplateRenderer($this->context);
```
PierreRambaud: Can we use dependency injection instead? Use an accessor at least?

mails/_partials/order_conf_product_list.tpl
```diff
@@ -0,0 +1,136 @@
...
+  			</table>
+  		</td>
+  		<td style="border:1px solid #D6D4D4;">
+  			<table class="table">
+  				<tr>
+  					<td width="5">&nbsp;</td>
+  					<td align="right">
+  						<font size="2" face="Open-sans, sans-serif" color="#555454">
+  							{if count($product['customization']) > 1}
+  								{$customization['customization_quantity']}
```
PierreRambaud, line 124: :+1:

mails/_partials/order_conf_product_list.txt
```diff
@@ -0,0 +1,20 @@
+{foreach $list as $product}
+						{$product['reference']}
```
PierreRambaud, line 2: really weird indent

src/Core/MailTemplate/Transformation/MailVariablesTransformation.php
```diff
@@ -26,22 +26,48 @@
...
+     */
+    private $replaceVariables;
+
+    /**
+     * @param string $type
+     * @param array $replaceVariables
+     *
+     * @throws InvalidArgumentException
+     */
+    public function __construct($type, array $replaceVariables = array())
```
PierreRambaud, line 51:
```suggestion
    public function __construct(string $type, array $replaceVariables = [])
```

src/Adapter/MailTemplate/MailPartialTemplateRenderer.php
```diff
@@ -0,0 +1,79 @@
...
+     * Fetch the content of $partialTemplateName inside the folder
+     * current_theme/mails/current_iso_lang/ if found, otherwise in
+     * mails/current_iso_lang.
+     *
+     * @param string $partialTemplateName template name with extension
+     * @param array $variables sent to smarty as 'list'
+     *
+     * @return string
+     */
+    public function render($partialTemplateName, array $variables)
```
PierreRambaud: Something you want to render partial template without variables
```suggestion
    public function render(string $partialTemplateName, array $variables = [])
```

src/Core/MailTemplate/Transformation/MailVariablesTransformation.php
```diff
@@ -26,22 +26,48 @@
...
+
     /**
      * {@inheritdoc}
      */
     public function apply($templateContent, array $templateVariables)
     {
+        $replaceVariables = $this->replaceVariables;
         if (!empty($templateVariables['templateVars'])) {
-            $templateContent = strtr($templateContent, $templateVariables['templateVars']);
+            $replaceVariables = array_merge(
```
PierreRambaud:
```suggestion
            $replaceVariables = array_merge_recursive(
```

If you have
```php
$array = [
  'data' => [
    'a' => 1,
  ]
];
$array2 = [
  'data' => [
    'b' => 2,
  ]
];

var_dump(array_merge($array, $array2));
var_dump(array_merge_recursive($array, $array2));
// array(1) {
//   ["data"]=>
//   array(1) {
//     ["b"]=>
//     int(2)
//   }
// }
// array(1) {
//   ["data"]=>
//   array(2) {
//     ["a"]=>
//     int(1)
//     ["b"]=>
//     int(2)
//   }
// }

```

PierreRambaud (changes requested):
See comments

mails/_partials/order_conf_product_list.txt
jolelievre: I agree, but it's a txt template from the legacy so I prefer to keep it as is

src/Core/MailTemplate/Transformation/MailVariablesTransformation.php
jolelievre: does it work on every PHP versions? or is it fine since we dropped 5.6?

src/Adapter/MailTemplate/MailPartialTemplateRenderer.php
jolelievre: Not in this case, there are always data necessary to rendre the product list
But I'll add this in case this component is used for other templates in the future

src/Core/MailTemplate/Transformation/MailVariablesTransformation.php
jolelievre: great! thanks

src/Core/MailTemplate/Transformation/MailVariablesTransformation.php
jolelievre: oh ok it's since `PHP 7.0.0` so we're good to go!

jolelievre: oh but since it already existed in 176 is it a BC?

jolelievre:
@PierreRambaud I think it's alright!

mails/_partials/order_conf_product_list.txt
PierreRambaud: :(

src/Core/MailTemplate/Transformation/MailVariablesTransformation.php
PierreRambaud: When we'll drop the 5.6 support, it'll not be a BC :)

PierreRambaud (approved):

PierreRambaud:
Thanks @jolelievre

PR #14612, opened 2019-07-12
Update docker for puppeteer

boubkerbribri:
| Questions     | Answers
| Branch?       | develop
| Description?  | refactoring dockerFiles and docker-compose file
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | See the readMe file to know how test with docker-compose

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14612)

### DIFF
--- a/tests/puppeteer/.docker/Dockerfile
+++ b/tests/puppeteer/.docker/Dockerfile
@@ -3,23 +3,23 @@
 # Based upon:
 # https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker
 
-FROM node:8-slim
+FROM node:10-slim
 
-RUN  apt-get update \
-     # See https://crbug.com/795759
-     && apt-get install -yq libgconf-2-4 \
-     # Install latest chrome dev package, which installs the necessary libs to
-     # make the bundled version of Chromium that Puppeteer installs work.
-     && apt-get install -y wget --no-install-recommends \
-     && wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+# Install latest chrome dev package and fonts
+RUN  wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
      && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
      && apt-get update \
-     && apt-get install -y google-chrome-unstable --no-install-recommends \
+     && apt-get install -y google-chrome-unstable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst \
+     ttf-freefont --no-install-recommends \
      && rm -rf /var/lib/apt/lists/*
 
-# Install Puppeteer under /node_modules so it's available system-wide
-COPY ["./package.json", "./test/linkchecker.js","./"]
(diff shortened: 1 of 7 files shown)


tests/puppeteer/.docker/Dockerfile
```diff
@@ -3,23 +3,23 @@
...
-     # Install latest chrome dev package, which installs the necessary libs to
-     # make the bundled version of Chromium that Puppeteer installs work.
-     && apt-get install -y wget --no-install-recommends \
-     && wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+# Install latest chrome dev package and fonts
+RUN  wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
      && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
      && apt-get update \
-     && apt-get install -y google-chrome-unstable --no-install-recommends \
+     && apt-get install -y google-chrome-unstable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst \
```
SimonGrn: Are you sure wee need all of theses exotic fonts ? Seems unnecessary to me

PierreRambaud (dismissed):

tests/puppeteer/test/linkchecker.js
```diff
@@ -128,6 +128,11 @@ const run = async () => {
   const curDate = new Date();
 
   const filename = `report_${curDate.toJSON().slice(0, 10)}_${curDate.getHours()}${curDate.getMinutes()}${curDate.getSeconds()}.json`;
+  // Create folder reports in not exist
```
SimonGrn: s/in/if

tests/puppeteer/.docker/Dockerfile
boubkerbribri: From puppeteer github : 
> This fonts are used to support major charsets (Chinese, Japanese, Arabic, Hebrew, Thai and a few others)

I think maybe we will need that for installation tests

tests/puppeteer/test/linkchecker.js
boubkerbribri: Done, thanks !

tests/puppeteer/.docker/Dockerfile
PierreRambaud: Just thinking, is it a good idea to use `google-chrome-unstable`?
Should be `google-chrome-stable`

boubkerbribri: `google-chrome-unstable` is used by developers of puppeteer. why? I don't really know.
Maybe to test sooner the beta version of chrome with puppeteer.
Should we do the same thing ? Not necessarily but I prefer to have my DockerFile based on the one in the official page.

PierreRambaud: Watch out, reading their FAQ, it's only because they contribute to Chromium and also specific to puppeteer development cycle.

If you do this, you must update `puppeteer` as often as they do.
Each version of puppeteer target a specifc version of chromium: https://github.com/GoogleChrome/puppeteer/blob/master/package.json#L11

boubkerbribri: ok, changes done.
thanks

PierreRambaud (approved):

PR #14613, opened 2019-07-12
Fix broken translations in module list page

### LINKED ISSUE #14589: Wording seems to be off translation
**Describe the bug**
Currently, there are a few wordings in the back office that are not available in the Crowdin translation files. It means that translations are missing in some parts. It seems that the "Browse" button is related to the Symfony migration. See screenshots.

**To Reproduce**
1. Go to the Categories, Import, Modules and Theme pages
2. Change the back office language for a non-English
3. See missing translations

**Screenshots**
Capture d’écran 2019-07-11 à 11 27 58
Capture d’écran 2019-07-10 à 16 17 06
Capture d’écran 2019-07-10 à 16 16 35

**Additional information**
PrestaShop version: 1.7.6
PHP version: N/A

matthieu-rolland:
- link module actions names to related translation domain
- fix module action display name generation
- use correct translation domain

| Questions     | Answers
| Branch?       | develop
| Description?  | Translation broken in module listing page, in the action dropdown of each item.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14589
| How to test?  | See description in the issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14613)

### DIFF
--- a/src/Adapter/Module/AdminModuleDataProvider.php
+++ b/src/Adapter/Module/AdminModuleDataProvider.php
@@ -53,14 +53,9 @@ class AdminModuleDataProvider implements ModuleInterface
     const _DAY_IN_SECONDS_ = 86400; /* Cache for One Day */
 
     /**
-     * @var array of defined and callable module actions
+     * @const array giving a translation domain key for each module action
      */
-    protected $moduleActions = array('install', 'uninstall', 'enable', 'disable', 'enable_mobile', 'disable_mobile', 'reset', 'upgrade');
-
-    /**
-     * @var array giving a translation domain key for each module action
-     */
-    protected $actionsTranslationDomains = array(
+    const _ACTIONS_TRANSLATION_DOMAINS_ = array(
         'install' => 'Admin.Actions',
         'uninstall' => 'Admin.Actions',
         'enable' => 'Admin.Actions',
@@ -71,6 +66,11 @@ class AdminModuleDataProvider implements ModuleInterface
         'upgrade' => 'Admin.Actions',
         'configure' => 'Admin.Actions',
     );
+
+    /**
+     * @var array of defined and callable module actions
+     */
+    protected $moduleActions = array('install', 'uninstall', 'enable', 'disable', 'enable_mobile', 'disable_mobile', 'reset', 'upgrade');
 
     /**
      * @var int
@@ -346,7 +346,7 @@ class AdminModuleDataProvider implements ModuleInterface
 
             $urls = $this->filterAllowedActions($urls, $addon->attributes->get('name'));
             $addon->attributes->set('urls', $urls);
-            $addon->attributes->set('actionTranslationDomains', $this->actionsTranslationDomains);
+            $addon->attributes->set('actionTranslationDomains', self::_ACTIONS_TRANSLATION_DOMAINS_);
             if ($specific_action && array_key_exists($specific_action, $urls)) {
                 $addon->attributes->set('url_active', $specific_action);
             } elseif ($url_active === 'buy' || array_key_exists($url_active, $urls)) {
--- a/src/Adapter/Module/AdminModuleDataProvider.php
+++ b/src/Adapter/Module/AdminModuleDataProvider.php
@@ -58,6 +58,21 @@ class AdminModuleDataProvider implements ModuleInterface
     protected $moduleActions = array('install', 'uninstall', 'enable', 'disable', 'enable_mobile', 'disable_mobile', 'reset', 'upgrade');
 
     /**
+     * @var array giving a translation domain key for each module action
+     */
+    protected $actionsTranslationDomains = array(
+        'install' => 'Admin.Actions',
+        'uninstall' => 'Admin.Actions',
+        'enable' => 'Admin.Actions',
+        'disable' => 'Admin.Actions',
+        'enable_mobile' => 'Admin.Modules.Feature',
+        'disable_mobile' => 'Admin.Modules.Feature',
+        'reset' => 'Admin.Actions',
+        'upgrade' => 'Admin.Actions',
+        'configure' => 'Admin.Actions',
+    );
+
+    /**
      * @var int
      */
     private $languageISO;
@@ -331,6 +346,7 @@ class AdminModuleDataProvider implements ModuleInterface
 
             $urls = $this->filterAllowedActions($urls, $addon->attributes->get('name'));
             $addon->attributes->set('urls', $urls);
+            $addon->attributes->set('actionTranslationDomains', $this->actionsTranslationDomains);
             if ($specific_action && array_key_exists($specific_action, $urls)) {
                 $addon->attributes->set('url_active', $specific_action);
             } elseif ($url_active === 'buy' || array_key_exists($url_active, $urls)) {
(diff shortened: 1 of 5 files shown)


src/Adapter/Module/AdminModuleDataProvider.php
```diff
@@ -57,6 +57,21 @@ class AdminModuleDataProvider implements ModuleInterface
...
+        'uninstall' => 'Admin.Actions',
+        'enable' => 'Admin.Actions',
+        'disable' => 'Admin.Actions',
+        'enable_mobile' => 'Admin.Modules.Feature',
+        'disable_mobile' => 'Admin.Modules.Feature',
+        'reset' => 'Admin.Actions',
+        'upgrade' => 'Admin.Actions',
+        'configure' => 'Admin.Actions',
+    );
+
```
matthieu-rolland, line 69: I'm not so sure I'm doing this at the right place, if anyone as a better suggestion I'll be willing to discuss and implement it !

PS: I couldn't reuse the $moduleActions array because it lacked an action (configure) that also needs to be linked to its translation domain.

src/PrestaShopBundle/Resources/views/Admin/Module/Includes/action_button.html.twig
```diff
@@ -23,7 +23,7 @@
  * International Registered Trademark & Property of PrestaShop SA
  *#}
 
-{% set displayAction = action|capitalize|replace({'_': " "})|trans({}, 'Admin.Actions') %}
+{% set displayAction = action|title|replace({'_': " "})|trans({}, transDomain[action]) %}
```
matthieu-rolland, line 27: the **capitalize** filter would only capitalize the first character, here we need a **ucwords** kind of filter. Happily the **title** filter does just that.

matthieu-rolland:
@LouiseBonnard 
We need a translation for the word "browse" for when a button opens the file explorer.
(that would be "parcourir" in french)

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
```diff
@@ -701,6 +701,11 @@
 {% endblock birthday_widget %}
 
 {% block file_widget %}
+    <style>
```
PierreRambaud: two spaces for indent please

src/Adapter/Module/AdminModuleDataProvider.php
PierreRambaud: If you don't change its value, use const instead

matthieu-rolland: done !

LouiseBonnard:
Hello @breizoreol, thanks a lot for the PR! What translation do you need? Actually, the word _Browse_ (localized in _Admin.Actions_) is okay, we just have to make it translatable - this way, it will be part of the 1.7.7 catalog on [Crowdin](https://crowdin.com/project/prestashop-official) and the community will be able to translate it in all different languages. Feel free to ask me more information if needed, perhaps my issue was not that clear. ;-)

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
```diff
@@ -701,6 +701,11 @@
 {% endblock birthday_widget %}
 
 {% block file_widget %}
+<style>
```
PierreRambaud: Still wrong indent :smile:

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
```diff
@@ -701,6 +701,11 @@
...
+  .custom-file-label:after {
+    content: "{{ "browse"|trans({}, 'Admin.Actions')  }}";
```
LouiseBonnard: Please write it _Browse_ instead, with a capital letter at the beginning, thank you.

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
matthieu-rolland: done !

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
matthieu-rolland: done !

LouiseBonnard:
One question @breizoreol, how about the _Disable mobile_ entry, would it be fixed in another PR or in this one?

src/PrestaShopBundle/Resources/views/Admin/Module/Includes/action_button.html.twig
jolelievre: Shouldn't the `trans` filter be called before using the `title` one? Or else the key might not be recognized?

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
```diff
@@ -701,6 +701,11 @@
 {% endblock birthday_widget %}
 
 {% block file_widget %}
+  <style>
```
PierreRambaud: Can you add the style tag in the `div class="custom-file">` and try this
```suggestion
  <style scoped>
```

PierreRambaud (approved):

matthieu-rolland:
> One question @breizoreol, how about the _Disable mobile_ entry, would it be fixed in another PR or in this one?

@LouiseBonnard yes, the 'Disable mobile' was not translated due to a mistake in the code, that is fixed in this pr.

src/PrestaShopBundle/Resources/views/Admin/Module/Includes/action_button.html.twig
matthieu-rolland: Good question @jolelievre, initially the **capitalize** filter was used to build the translation key. But it failed, because capitalize makes a **ucfirst**, when what was needed to build a proper translation key was a **ucwords** (in this case anyway). The title filter does the needed **ucwords**

jolelievre: Ok! then you need top check if the keys are present in the catalogue. Because the catalog is built by scanning the PrestaShop code and fetching all the hard coded strings. It is not able to compute variables. So unless these keys are used explicitly somewhere else in the code they won't be present in the catalogue and won't be translated.

matks:
Thank you @matthieu-rolland 🎉

src/PrestaShopBundle/Resources/views/Admin/Module/Includes/action_button.html.twig
matthieu-rolland: Yes, they exist in the catalog, from what I've seen and understood, those values are used as constants in the code (php side), and made to be transformed into translation keys easily (twig side).

Issue #14591, opened 2019-07-11
Error upgrading to 1.7.6 with php 5.6


davidglezz:
Upgrading from 1.6.0 to 1.7.6 (php 5.6):
**PHP Fatal error: Call to a member function get() on null in /var/www/html/install/upgrade/php/ps_1760_copy_data_from_currency_to_currency_lang.php on line 55**

Sorry, I did not have time to try it in a fresh prestashop.
I was able to update up to 1.7.5.2, but not up to 1.7.6.

1. From old PrestaShop version (1.6.0.x by example), with php 5.6
2. Do manual upgrade https://devdocs.prestashop.com/1.7/basics/keeping-up-to-date/upgrade/#manual-upgrade-process-details
4. See error above

**Additional information**
PrestaShop version: 1.7.6
PHP version: 5.6
MySQL : 5.5

khouloudbelguith:
Hi @davidglezz,

Thank you for your report.
We'll first try to reproduce it and we'll come back to you if we need more information.

Thanks!

khouloudbelguith:
@davidglezz, could you please provide me the exact steps (step by step) that you followed to make the manual upgrade to PS1.7.6.0.

Thanks!

davidglezz:
Steps to reproduce
```
docker network create prestashop-test-net
docker run -ti --name prestashop-test-mysql --network prestashop-test-net -e MYSQL_ROOT_PASSWORD=admin -p 3307:3306 -d mysql:5.7
#docker run -ti --name prestashop-test --network prestashop-test-net -e DB_SERVER=prestashop-test-mysql -p 8888:80 -d prestashop/prestashop:1.6-5.6-apache
docker run -ti --name prestashop-test --network prestashop-test-net -e DB_SERVER=prestashop-test-mysql -p 8888:80 -d prestashop/prestashop:1.5-5.6-apache

# Install http://localhost:8888/install/index.php

docker exec -it prestashop-test bash

# Update prestashop : script https://devdocs.prestashop.com/1.7/basics/keeping-up-to-date/upgrade/

mkdir /prestashop-upgrade
wget -q -O /prestashop-upgrade/prestashop-upgrade.zip https://github.com/PrestaShop/PrestaShop/releases/download/1.7.6.0/prestashop_1.7.6.0.zip
[...]

Issue #14617, opened 2019-07-12
1054 Unknown column 't0.filter id' in 'field list when i upgrade 1.7.5.2 TO 1.7.6


yoss55:
**Describe the bug**

**To Reproduce**

**Screenshots**
If applicable, add screenshots or screenrecords to help explain your problem.

**Additional information**
PrestaShop version: N/A
PHP version: N/A

yoss55:
HI,
When I upgrade v1.7.5.2 to 1.7.6 I have this error: 1054 Unknown column 't0.filter id' in 'field list.
And I can't access to the product and employee.
Please help me

khouloudbelguith:
Hi @yoss55,
Could you please try to clear cache manually by deleting all the folder into var/cache (prod & dev).
Can you please provide us the log of your upgrade? It can be found in the log.txt file, available in <admin folder>/autoupgrade/tmp/log.txt
Thanks!

yoss55:
hi, 
this is the log : 
[2019-07-12 08:09:31] Migrating old setting file...
[2019-07-12 08:09:31] parameters file already exists!
[2019-07-12 08:09:31] Finished...
[2019-07-12 10:09:31] <div class="upgradeDbOk">[OK] SQL 1.7.6.0 SET SESSION sql_mode = ''</div>
[2019-07-12 10:09:31] <div class="upgradeDbOk">[OK] SQL 1.7.6.0 SET NAMES 'utf8'</div>
[2019-07-12 10:09:31] <div class="upgradeDbOk">[OK] SQL 1.7.6.0 ALTER TABLE `ps_currency` ADD `numeric_iso_code` varchar(3) NOT NULL DEFAULT '0' AFTER `iso_code`</div>
[2019-07-12 10:09:31] <div class="upgradeDbOk">[OK] SQL 1.7.6.0 ALTER TABLE `ps_currency` ADD `precision` int(2) NOT NULL DEFAULT 6 AFTER `numeric_iso_code`</div>
[2019-07-12 10:09:32] <div class="upgradeDbOk">[OK] SQL 1.7.6.0 ALTER TABLE `ps_currency` ADD KEY `currency_iso_code` (`iso_code`)</div>
[2019-07-12 10:09:32] <div class="upgradeDbOk">[DROP] SQL `ps_currency_lang` table has been dropped.</div>
[2019-07-12 10:09:32] <div class="upgradeDbOk">[OK] SQL 1.7.6.0 /* Localized currency information */
[...]

PR #14620, opened 2019-07-12
Init properly the Kernel on major *manual* upgrades

Quetzacoalt91:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Upgrades to PS 1.7.6 require the kernel to run. Unfornately, the migration of the settings file when upgrading from PS < 1.7 was done AFTER the checks allowing the kernel to be enabled. This PR orders the functions calls accordingly.
| Type?         | bug fix
| Category?     | IN
| BC breaks?    | Nope
| Deprecations? | Nope
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/14591 & https://github.com/PrestaShop/PrestaShop/issues/14617
| How to test?  | Follow https://github.com/PrestaShop/PrestaShop/issues/14591#issuecomment-510841509

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14620)

### DIFF
--- a/install-dev/upgrade/upgrade.php
+++ b/install-dev/upgrade/upgrade.php
@@ -45,8 +45,11 @@ if (isset($_GET['adminDir']) && $_GET['adminDir'] && !defined('_PS_ADMIN_DIR_'))
     define('_PS_ADMIN_DIR_', base64_decode($_GET['adminDir']));
 }
 
-require_once dirname(__FILE__).'/../init.php';
+require_once __DIR__ . '/../../vendor/autoload.php';
+require_once __DIR__ . '/../../config/defines.inc.php';
+require_once __DIR__ . '/../../config/autoload.php';
 Upgrade::migrateSettingsFile();
+require_once dirname(__FILE__).'/../init.php';
 require_once _PS_CONFIG_DIR_.'bootstrap.php';
 
 $logDir = _PS_ROOT_DIR_.'/var/logs/'.(_PS_MODE_DEV_ ? 'dev' : 'prod').'/';


install-dev/upgrade/upgrade.php
```diff
@@ -45,8 +45,11 @@
     define('_PS_ADMIN_DIR_', base64_decode($_GET['adminDir']));
 }
 
-require_once dirname(__FILE__).'/../init.php';
+require_once __DIR__ . '/../../vendor/autoload.php';
```
jolelievre, line 48: should you check for its existence like in the module?

eternoendless: I don't think so, because without that autoloader PrestaShop won't work.

install-dev/upgrade/upgrade.php
```diff
@@ -45,8 +45,11 @@
...
+require_once __DIR__ . '/../../config/defines.inc.php';
+require_once __DIR__ . '/../../config/autoload.php';
```
eternoendless, line 50: This is already done in `init.php` I think

install-dev/upgrade/upgrade.php
jolelievre: Unless you forgot to run `composer install` But then nothing would work, so yeah you're right

install-dev/upgrade/upgrade.php
jolelievre: Yes but `Upgrade::migrateSettingsFile()` needs to be called before the init so that the kernel is properly instanciated And I guess this method has dependency on const which are defined in these other files

Quetzacoalt91: Precisely, thanks @jolelievre

PierreRambaud (approved):

PierreRambaud:
Milestone must be created!

Quetzacoalt91:
Done

jolelievre (approved):

PierreRambaud:
Thanks @Quetzacoalt91

PR #14645, opened 2019-07-15
Fix modules listing in their correct categories, not in "Other"

### LINKED ISSUE #15298: Our Payment modules are being considered as non-compatible in Prestashop 1.7+
Hello,

I'm a contributor on the Prestashop addons store with few modules. And there are 2 very popular payment modules on mine which are considered as not compatible from Prestashop 1.7+ version. There are 2 main issues:
1. These are listed under **Other** category instead of **payment_gateways**. And doesn't show on the Payment->Payment Methods. But module is showing under Payment->Preferences just fine.

2. Another issue is that the module gets auto deleted from the FTP while upgrading Prestashop with statement of non-compatibility.

Before I came to post this issue here, I got the answer from the Prestashop addons validating team that these bugs are due to the core.
My payment modules are:
https://addons.prestashop.com/en/product.php?id_product=17856
https://addons.prestashop.com/en/product.php?id_product=19407

Other then this, there is no issue present in the modules related with any PS 1.7+ non-compatibility.
There are lots clients who reported these issues but I can no more ignore this issue now.

Best Regards,
Dev

debuss:
Fixes the issue on Module Manager page where modules were not in their categories but in "Other".

| Questions     | Answers
| Branch?       | develop
| Description?  | In the Module Manager, most of the modules were listed in "Other" instead of the their correct categories (e.g.: "Trafic & Marketplaces", "Administration", etc). This pull request will fix that, see screenshot below.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | No
| Deprecations? | No
| Fixed ticket? | Fixes #15298
| How to test?  | Install the new PrestaShop 1.7.6.0, install a few modules, you'll see they are in "Other" category instead of the one set in **ModuleCore::$tab**.

Before :
image

After :
image

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14645)

### DIFF
--- a/src/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProvider.php
+++ b/src/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProvider.php
@@ -187,6 +187,8 @@ class CategoriesProvider
             foreach ($parentCategory->categories as $childCategory) {
                 if ($childCategory->name === $categoryName) {
                     return $parentCategory->name;
+                } elseif (isset($childCategory->tab) && $childCategory->tab === $categoryName) {
+                    return $parentCategory->name;
                 }
             }
         }
@@ -235,10 +237,13 @@ class CategoriesProvider
      *
      * @param ApiModule $installedProduct Installed product
      * @param array $categories Available categories
+     * @return string
      */
     private function findModuleCategory(ApiModule $installedProduct, array $categories)
     {
-        $moduleCategoryParent = $installedProduct->attributes->get('categoryParentEnglishName');
+        $moduleCategory = $installedProduct->attributes->get('tab');
+        $moduleCategoryParent = $this->getParentCategory($moduleCategory);
+
         if (!isset($categories['categories']->subMenu[$moduleCategoryParent])) {
             if (in_array($installedProduct->attributes->get('name'), $this->modulesTheme)) {
                 $moduleCategoryParent = self::CATEGORY_THEME;


src/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProvider.php
```diff
@@ -187,6 +187,8 @@ public function getParentCategory($categoryName)
             foreach ($parentCategory->categories as $childCategory) {
                 if ($childCategory->name === $categoryName) {
                     return $parentCategory->name;
+                } elseif (isset($childCategory->tab) && $childCategory->tab === $categoryName) {
```
eternoendless, line 190: This `if` could be merged with the previous one:

```php
if ($childCategory->name === $categoryName
    || (isset($childCategory->tab) && $childCategory->tab === $categoryName)
) {
    return $parentCategory->name;
}

eternoendless (commented):
Hi @debuss and thanks for this contribution!

Need feedback from @PierreRambaud regarding the fix and @PrestaShop/prestashop-product-team regarding the target branch.

PierreRambaud:
@eternoendless If we take this PR, we must remove everything about API calls which become useless.
This is something under discussion with @colinegin, we also want to have this behavior (using module tab only instead depending on an API)
The branch must be develop.
I think the problem came from `config/xml/*modules*.xml` which are not cleared when using `Clear cache` button.

debuss:
`config/xml/*modules*.xml` only concerns PrestaShop modules and affiliates, not sure the problem of categorization comes from those xml files.
For now I suppose that is the only fix as it depends on an API.

PierreRambaud:
In this section, there're requests to get information about modules. These requests save xml in this directory, and this category is waiting for a match with the english category name.

colinegin:
Hello @debuss ,

Thanks for your feedback. 
I've just made a test by uploading the Amazon module from Common Services, and it is correctly displayed in Traffic & Marketplaces category. 
I've also tried by installing a module directly from the Module catalog page in the BO, and it also works.

Could you tell me when your issue occurs ?

debuss:
@colinegin 

Just install PrestaShop 1.7.6.0, and see in Module Manager :

image

Everything is listed in **Autre**...

If you need a marketplace example, create your own marketplace module, one that is not listed among PrestaShop Partnair/Official modules (or Spartoo Marketplace for instance).
Install it, then check where it is listed on Module Manager page.

If needed : [the module](https://www.dropbox.com/s/7ecl4ydk5n9wwaz/nullmarketplace.zip?dl=0).
Result :
image

There is absolutely no module in the corresponding category, they all are in "Autre" (=Other).

debuss:
@colinegin 
Concerning Amazon module, set your back-office in french language, it wont be listed in "Traffic & Marketplaces".

colinegin:
I'm really sorry but even with what you say I cannot reproduce it. Could you help us by sending us a video or all the steps needed to reproduce your issue ?

Thanks !

colinegin:
Hello @debuss ,

since we did not get any news for more than a month, I am closing this issue. Do not hesitate to reopen it and comment it if you have any more information.

Have a nice day !

Quetzacoalt91:
Hi there!

Too bad this issue was closed for inactivity. FYI we receive similiar requests for bought modules manually downloaded for the marketplace. Relying by default on the `tab` property would help to not store all modules in the `others` category.

PierreRambaud:
Agreed with @Quetzacoalt91, this is something we discussed weeks ago. We should remove the marketplace dependencies if possible. Or at least, move to the right category depending on the tab when the module is not on the marketplace.

nexustotalsolutions:
This issue still exists and I open it here:
https://github.com/PrestaShop/PrestaShop/issues/15298

colinegin:
I'm all for fixing it guys ! The issue is that we did not manage to reproduce the issue, so it's hard to know how to fix it...

nexustotalsolutions:
The API moves the different modules in the proper category only if the merchant is connected to its Marketplace account in the backoffice.
In the case he downloaded it from the marketplace (let's say from the URL given in the first link), then upload these modules manually in its Backoffice, the module will be stored in the "Others" category.

Actually we'd deal with a fallback here. If the marketplace has no information to provide to a module, we could use its tab property instead of the "Others" category by default.

colinegin:
Hello @PrestaShop/prestashop-core-developers , what should we do with this PR ? Is it useful for what we want to do ? 

I've added #15298 to the PrestaShop 1.7.7 kanban.

colinegin:
Hello @debuss , thanks again for reporting this bug. 
We have been working on a solution on our side, the PR (#16658) is currently being reviewed. We hope to have it fixed for 1.7.7 ! I close this PR so that everyone can follow the new one.

Have a nice day,

Coline

PR #14647, opened 2019-07-15
Add filter_id column in upgrade script

Quetzacoalt91:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Bring back column creation, as it seems Doctrine won't handle it. Partially reverts https://github.com/PrestaShop/PrestaShop/pull/13540.
| Type?         | bug fix
| Category?     | IN
| BC breaks?    | Nope
| Deprecations? | Nope
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/14617
| How to test?  | Upgrade and check the languages page.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14647)

### DIFF
--- a/install-dev/upgrade/sql/1.7.6.1.sql
+++ b/install-dev/upgrade/sql/1.7.6.1.sql
@@ -1,0 +1,5 @@
+SET SESSION sql_mode = '';
+SET NAMES 'utf8';
+
+ALTER TABLE `PREFIX_admin_filter`
+	ADD `filter_id` VARCHAR (255) DEFAULT '' NOT NULL AFTER `shop`;


PierreRambaud (approved):

jolelievre:
@Quetzacoalt91 Does it not handle it at all? Or does it handle it too late? (after the upgrade scripts are launched?)

Quetzacoalt91:
Well, it is properly handled by doctrine.

After reading again the original issue, it appears the error by an incomplete upgrade. It failed because on an uninitialized Symfony kernel on CLDR upgrade. This fix is not needed.

PR #13485, opened 2019-04-19
Migrate "Catalog > Brand & Suppliers > Suppliers create/edit action

zuk3975:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Sell > Catalog > Brand & Suppliers > Suppliers create and edit action.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #10541 
| How to test?  | Use `admin-dev/index.php/sell/catalog/suppliers` link to open suppliers list, then click on `edit` supplier action (or follow the link `admin-dev/index.php/sell/catalog/suppliers/{supplierId}/edit`) to see the form for editing . For create form - click on `Add new supplier` (or follow the link `admin-dev/index.php/sell/catalog/suppliers/new`) . Forms should be handled same as in legacy page.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/13485)

PR #14656, opened 2019-07-15
Fix editing shop association in Category and Manufacturer

zuk3975:
…egory edit handlers

| Questions     | Answers
| Branch?       | develop
| Description?  | `associateWithShops()` method of AbstractObjectModelHandler didn't work when used before object->update(). So it caused issues, when updating shop association (nr. 1@ https://github.com/PrestaShop/PrestaShop/pull/13485#issuecomment-510792137). This PR moves that method usage to be after update() and take effect in EditManufacturerHandler and EditCategoryHandler .
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Go to Manufacturers or Category page while having multiple shops, edit it -> change shop association. See the change is made after submit. (previously the change didn't take any effect)Screenshot from 2019-07-15 14-49-58 Screenshot from 2019-07-15 14-51-26

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14656)

### DIFF
--- a/src/Adapter/Category/CommandHandler/EditCategoryHandler.php
+++ b/src/Adapter/Category/CommandHandler/EditCategoryHandler.php
@@ -105,10 +105,6 @@ final class EditCategoryHandler extends AbstractObjectModelHandler implements Ed
             $category->groupBox = $command->getAssociatedGroupIds();
         }
 
-        if ($command->getAssociatedShopIds()) {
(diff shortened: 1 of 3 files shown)


matks (approved):

PierreRambaud:
Thanks  @zuk3975

Issue #10545, opened 2018-09-20
Migrate "Catalog > Brand & Suppliers > Suppliers" page


matks:
Type : page
Part of Symfony migration project

Todo:
- [x] Migrate listing https://github.com/PrestaShop/PrestaShop/pull/11153
- [x] Add empty state https://github.com/PrestaShop/PrestaShop/pull/12498

Access: `/admin-dev/index.php/sell/catalog/suppliers/`

BilalDr:
This page needs to have an Empty States.
refer to this issue : #10434

Thanks

matks:
@colinegin Bilal said this page needs an Empty State, but when I look at https://github.com/PrestaShop/PrestaShop/issues/10434 I cannot see the related Empty State. Is it a mistake ?

colinegin:
Sorry, I forgot the empty state part ! It should be fine now, I've added details in the same issue. I will update the URL document tomorrow so that you can have the correct user doc urls.

Just let me know if you need anything else !

PR #14665, opened 2019-07-15
Allow to display/hide manufacturers and suppliers independently

### LINKED ISSUE #11332: Disable Supplier in front office
Hello,

I have the same request very regularly from E-merchants.

Most of them do not want suppliers with back office information to be visible in the front office.

Brands, of course, but not their suppliers.

Also, I think it should be disabled by default, and that an option could enable front-office visibility for those who want it.

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Ability to display or hide suppliers and brands pages independantly
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | fixes #11332
| How to test?  | Everything is detailed in the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14665)


matthieu-rolland:
Needed wording on the general configuration page

- Display suppliers
- Enable suppliers page on your front office even when its module is disabled.
- Display brands
- Enable brands page on your front office even when its module is disabled.

Needed wording on the brands / suppliers admin page

- The display of your brands is disabled/enabled on your store. Go to Shop Parameters > General to edit settings.
- The display of your suppliers is disabled/enabled on your store. Go to Shop Parameters > General to edit settings.

controllers/admin/AdminSuppliersController.php
```diff
@@ -565,4 +567,23 @@ protected function afterUpdate($object)
 
         return true;
     }
+
+    /**
+     * supplierInfoMessage: builds a message related to the display suppliers configuration status
```
PierreRambaud:
```suggestion
     * Builds a message related to the display suppliers configuration status
```

classes/controller/AdminController.php
```diff
@@ -4882,4 +4882,19 @@ private function prepareNumberSpecifications(Context $context)
...
+     *
+     * @see admin-dev/themes/default/template/content.tpl
+     *
+     * @param string $message
+     *
+     * @return void
+     */
+    protected function setTopAlertInfoMessage($message)
+    {
+        $this->context->smarty->assign('alertInfo', $message);
```
PierreRambaud: Isn't too overkill?
Should it be more generic?
```php
protected function addTopAlertMessage($message, $namespace = 'info') // Maybe use const here
{
    $messages = $this->context->smarty->getVariable('alertMessages')->value;
    if ($messages === null) {
        $messages = [];
    }

    if (!isset($messages[$namespace])) {
        $messages[$namespace] = [];
    }

    $messages[$namespace][] = $messages;

    $this->context->smarty->assign('alertMessages', $messages);
}
```

matthieu-rolland: I don't think its overkill, the place where I was asked to put this info message text seems pretty generic to all legacy admin templates...  So it seems wrong to implement something specific coming only from AdminSuppliersController at this place.

But your suggestion is indeed more generic, I didn't think about all the kinds of info messages, I'll implement this !

matthieu-rolland: done !

classes/controller/AdminController.php
```diff
@@ -4882,4 +4894,30 @@ private function prepareNumberSpecifications(Context $context)
...
+    /**
+     * Using this method will display the given message as an "alert" message at the top of the current
+     * administration page.
+     *
+     * @see admin-dev/themes/default/template/content.tpl
+     *
+     * @param string $message The message to be displayed
+     * @param string $namespace Type of alert, use FRONT_ALERT_XXX constants
+     */
+    protected function addTopAlertMessage($message, $namespace = false)
```
PierreRambaud:
```suggestion
    protected function addTopAlertMessage($message, $namespace = self::FRONT_ALERT_INFO)
```
and remove the `if (!$namespace) {` condition :)

controllers/admin/AdminPreferencesController.php
```diff
@@ -194,6 +194,13 @@ public function __construct()
                     'cast' => 'intval',
                     'type' => 'bool',
                 ),
+                'PS_DISPLAY_MANUFACTURERS' => array(
+                    'title' => $this->trans('Display brands', array(), 'Admin.Shopparameters.Feature'),
+                    'desc' => $this->trans('Enable brands page on your front office even when its module are disabled.', array(), 'Admin.Shopparameters.Help'),
```
LouiseBonnard:
```suggestion
                    'desc' => $this->trans('Enable brands page on your front office even when its module is disabled.', array(), 'Admin.Shopparameters.Help'),
```

controllers/admin/AdminPreferencesController.php
```diff
@@ -194,6 +194,13 @@ public function __construct()
...
+                    'validation' => 'isBool',
+                    'cast' => 'intval',
+                    'type' => 'bool',
+                ),
                 'PS_DISPLAY_BEST_SELLERS' => array(
                     'title' => $this->trans('Display best sellers', array(), 'Admin.Shopparameters.Feature'),
                     'desc' => $this->trans('Enable best sellers page on your front office even when its respective module is disabled.', array(), 'Admin.Shopparameters.Help'),
```
LouiseBonnard: Wouldn't it be the occasion to harmonize the help wording here? We could write _Enable best sellers page on your front office even when its module is disabled._ like the previous one.

controllers/admin/AdminSuppliersController.php
```diff
@@ -565,4 +567,23 @@ protected function afterUpdate($object)
...
+     * @return string
+     */
+    protected function supplierInfoMessage()
+    {
+        $message = sprintf(
+            'The display of your suppliers is %s on your store. Go to Shop Parameters > General to edit settings.',
+            Configuration::get('PS_DISPLAY_SUPPLIERS') ? 'enabled' : 'disabled'
+        );
+
+        return $this->trans(
```
LouiseBonnard: Is it okay if the wording does not follow the `$this->trans` here?

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
```diff
@@ -784,4 +785,14 @@ private function getAddressFormHandler()
...
     }
+
+    private function getAlertInfoMessage()
+    {
+        $message = sprintf(
+            'The display of your brands is %s on your store. Go to Shop Parameters > General to edit settings.',
+            $this->configuration->get('PS_DISPLAY_MANUFACTURERS') ? 'enabled' : 'disabled'
+        );
+
+        return $this->trans($message, 'Admin.Catalog.Notification');
```
LouiseBonnard: Same question here. ;-)

controllers/admin/AdminPreferencesController.php
LouiseBonnard: We write this element twice? cf. https://github.com/PrestaShop/PrestaShop/pull/14665/files#diff-bc4c4faa966c237b7346ad189fdda6de

classes/controller/AdminController.php
matthieu-rolland: Good tip ! I was so persuaded I couldn't use a constant as a default parameter...

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
matthieu-rolland: Because the wording is dynamic, I build it first, inside the $message variable, then I use it inside $this->trans.
(is it what you meant ?)

controllers/admin/AdminSuppliersController.php
matthieu-rolland: Because the wording is dynamic, I build it first, inside the $message variable, then I use it inside $this->trans.
(is it what you meant ?)

LouiseBonnard: Alright, perfect, I don't know much about code so I wanted to ask just to be sure. ;-)

LouiseBonnard:
Two related suggestions from the community, about how to display manufacturers on a PrestaShop store:

http://feedback.prestashop.com/forums/387864/suggestions/36422122
http://feedback.prestashop.com/forums/387864/suggestions/36421930

@colinegin, @marionf, @samuel-pires

controllers/admin/AdminPreferencesController.php
matthieu-rolland: true, it's done !

controllers/admin/AdminPreferencesController.php
```diff
@@ -188,15 +188,22 @@ public function __construct()
                     'class' => 'fixed-width-xxl',
                 ),
                 'PS_DISPLAY_SUPPLIERS' => array(
-                    'title' => $this->trans('Display brands and suppliers', array(), 'Admin.Shopparameters.Feature'),
-                    'desc' => $this->trans('Enable brands and suppliers pages on your front office even when their respective modules are disabled.', array(), 'Admin.Shopparameters.Help'),
+                    'title' => $this->trans('Display suppliers', array(), 'Admin.Shopparameters.Feature'),
+                    'desc' => $this->trans('Enable suppliers page on your front office even when its module are disabled.', array(), 'Admin.Shopparameters.Help'),
```
LouiseBonnard:
```suggestion
                    'desc' => $this->trans('Enable suppliers page on your front office even when its module is disabled.', array(), 'Admin.Shopparameters.Help'),
```

controllers/admin/AdminPreferencesController.php
matthieu-rolland: yes, but it will not be displayed twice on the page, I just followed the existing logic for other fields in the same page and gave the translation key in this place as well.

classes/controller/AdminController.php
jolelievre: AS Pierre suggested you can use the constant directly as a default value

controllers/admin/AdminSuppliersController.php
jolelievre: Actually the `TranslationTool` won't be able to scan this message, because it can't `eval` the content of `$message` variable, si you'd better use an `if` condition and write two calls of `$this->trans` with hard coded strings

src/Adapter/Preferences/PreferencesConfiguration.php
```diff
@@ -80,8 +81,10 @@ public function updateConfiguration(array $configuration)
             $this->configuration->set('PS_ROUND_TYPE', $configuration['price_round_type']);
             $this->configuration->set('PS_PRICE_DISPLAY_PRECISION', $configuration['price_display_precision']);
             $this->configuration->set('PS_DISPLAY_SUPPLIERS', $configuration['display_suppliers']);
+            $this->configuration->set('PS_DISPLAY_MANUFACTURERS', $configuration['display_manufacturers']);
             $this->configuration->set('PS_DISPLAY_BEST_SELLERS', $configuration['display_best_sellers']);
             $this->configuration->set('PS_MULTISHOP_FEATURE_ACTIVE', $configuration['multishop_feature_active']);
+            $this->configuration->set('PS_MULTISHOP_FEATURE_ACTIVE', $configuration['multishop_feature_active']);
```
jolelievre: You have a duplicated line here

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
jolelievre: Same thing here, yo need to use an `if` condition and two calls to `$this->trans` with hard coded strings.

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
```diff
@@ -100,6 +100,7 @@ public function indexAction(
             'help_link' => $this->generateSidebarLink($request->attributes->get('_legacy_controller')),
             'manufacturerGrid' => $this->presentGrid($manufacturerGrid),
             'manufacturerAddressGrid' => $this->presentGrid($manufacturerAddressGrid),
+            'alertInfoMessage' => $this->getAlertInfoMessage(),
```
jolelievre: I'm wondering if you couldn't use the flash messages for this, instead of adding a special variable and change the template.

controllers/admin/AdminSuppliersController.php
```diff
@@ -67,6 +67,8 @@ public function __construct()
             'products' => array('title' => $this->trans('Number of products', array(), 'Admin.Catalog.Feature'), 'align' => 'right', 'filter_type' => 'int', 'tmpTableFilter' => true),
             'active' => array('title' => $this->trans('Enabled', array(), 'Admin.Global'), 'align' => 'center', 'active' => 'status', 'type' => 'bool', 'orderby' => false, 'class' => 'fixed-width-xs'),
         );
+
+        $this->addTopAlertMessage($this->supplierInfoMessage());
```
jolelievre: This message is added on a legacy page, but maybe it's gonna be migrated during 177 development. We should update the migration issue so that this feature is not forgotten.

controllers/admin/AdminSuppliersController.php
jolelievre: I think it would be a good idea to integrate a link in the message, so that user can go directly on the appropriate page

jolelievre: Also, with this the message will always be displayed, is there a way of closing this message?

jolelievre (changes requested):
Nice job @matthieu-rolland 
I have a few technical feedback, and some about the feature because I remember discussing a lot about this issue ^^

controllers/admin/AdminSuppliersController.php
jolelievre: Oh actually the listing page has already been migrated but is hidden for now: https://github.com/PrestaShop/PrestaShop/issues/10545
You can access it by typing the url directly, so you should modify the migrated page rather than the legacy one

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
matthieu-rolland: I thought about it... here are the rasons why I did it this way, don't hesitate to tell me if you disagree:

1/ This is not really a "flash" message, because it is to be displayed permanently, so it shouldn't depend on session's flashbag. It is not supposed to be closed/removed either.

2/ There is currently no type of flash message for this "info" type of message, I used the already existing **infotip** helper in twig :
```twig
{{ ps.infotip(alertInfoMessage) }}
```
3/ I think the problem here is the name I gave to this variable and its method, "alertInfoMessage" is misleading, I should give it a more sensical name, like "permanentInfoMessage" and getPermanentInfoMessage.

What do you think?

controllers/admin/AdminSuppliersController.php
matthieu-rolland: yep, done !

controllers/admin/AdminSuppliersController.php
matthieu-rolland: Ok, I had no idea the translation messages were scanned without executing the code...
I made the fix.

matthieu-rolland: About the link, yes I will add it ;)

jolelievre: > Ok, I had no idea the translation messages were scanned without executing the code...

Indeed, of course I'm not talking about the `Translator` service, but `TranslatorTool` which we use to generate the XLF catalogs, it scan (almost) all the files of the projects (PHP, Twig, ...) looking code match `trans('my text')` or `'my text'|trans` and can then extract the translation keys. But of course it can not deduce the context that would generate a key if you integrated variables inside.

jolelievre: About the link, you should use a key looking like this `To manage all your modules go to the [1]Installed module page[/1]` this prevents adding html code inside them

controllers/admin/AdminPreferencesController.php
```diff
@@ -188,15 +188,22 @@ public function __construct()
                     'class' => 'fixed-width-xxl',
                 ),
                 'PS_DISPLAY_SUPPLIERS' => array(
-                    'title' => $this->trans('Display brands and suppliers', array(), 'Admin.Shopparameters.Feature'),
-                    'desc' => $this->trans('Enable brands and suppliers pages on your front office even when their respective modules are disabled.', array(), 'Admin.Shopparameters.Help'),
+                    'title' => $this->trans('Display suppliers', array(), 'Admin.Shopparameters.Feature'),
```
jolelievre: Do you really need to change this legacy controller? I thought this page was already migrated to `/configure/shop/preferences/preferences`

controllers/admin/AdminSuppliersController.php
jolelievre: Here's an example https://github.com/PrestaShop/PrestaShop/blob/e762a65d41ff00588bcdccb1ee196bc6c835d4cc/src/PrestaShopBundle/Resources/views/Admin/Product/ProductPage/product.html.twig#L197

Although I think you can inject the parameters directly to `trans` instead of using `|replace`

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
jolelievre: Alright that makes sense! Although I find it weird that you can never hide this message. We introduced in recent versions some help cards (or showcase) in some pages (which can be hidden thanks to cookies), maybe the same feature could have been used (although it's not up to you to decide)

jolelievre: About the name indeed it might be more understandable if you rename it, maybe `settingsInfoMessage` or `settingsTipMessage`

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/SupplierController.php
```diff
@@ -500,4 +501,19 @@ private function getExceptionMessageByExceptionCode(Exception $exception)
 
         return $this->trans('Unexpected error occurred.', 'Admin.Notifications.Error');
     }
+
+    protected function getAlertInfoMessage()
```
jolelievre: same renaming here as discussed on manufacturer page

controllers/admin/AdminSuppliersController.php
matthieu-rolland: Argh, the page didn't refresh with your last comment, I used another way (using %s in trans(), it seems to work, the translator actually does a sprintf)

controllers/admin/AdminPreferencesController.php
matthieu-rolland: I added it here fearing that not doing so might break something. If this controller is fully deprecated then yes, there's not need to modify it.

classes/controller/AdminController.php
matthieu-rolland: done !

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
matthieu-rolland: done !

src/Adapter/Preferences/PreferencesConfiguration.php
matthieu-rolland: fixed !

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
matthieu-rolland: done ! I gave it this name

controllers/admin/AdminPreferencesController.php
jolelievre: Now that it's done I think you can leave, but were you able to test it? Because if the routing is properly set with `_legacy_link` you shouldn't be able to access the legacy page any more

controllers/admin/AdminSuppliersController.php
jolelievre: yes it's better to use the trans parameter than replace, it's weird for the `[1]` we have a lot of them in our trads

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Manufacturer/index.html.twig
```diff
@@ -25,22 +25,11 @@
...
-    icon: 'add_circle_outline',
-  },
-  add_manufacturer_address: {
-    href: path('admin_manufacturer_addresses_create'),
-    desc: 'Add new brand address'|trans({}, 'Admin.Catalog.Feature'),
-    icon: 'add_circle_outline',
-  },
-}
-%}
+
```
jolelievre: Why did you remove this part?

controllers/admin/AdminSuppliersController.php
jolelievre: You didn't manage the link in this legacy controller, this will result in two translation keys (one with the link placeholders, the other one without them)

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Manufacturer/index.html.twig
matthieu-rolland: Damn, that's the result of a wrong manipulation on this file on my locale...  This is fixed now.

controllers/admin/AdminSuppliersController.php
matthieu-rolland: This legacy controller is no longer used so I just removed what I did here.

controllers/admin/AdminPreferencesController.php
matthieu-rolland: You're right, I couldn't test a page that is not linked. I just knew that the 'Display brands' translation had to be reworded, and I was not sure about the consequences if I left the old wordings untouched at this place.

matthieu-rolland:
@jolelievre @matks  Addressed all review comments !

PierreRambaud:
Sorry, you need to rebase your branch

matthieu-rolland:
@PierreRambaud damn...yes 
I just did it

controllers/admin/AdminPreferencesController.php
```diff
@@ -0,0 +1,253 @@
+<?php
```
jolelievre: Seems like this file was removed from the branch, and you're re-adding it You can remove it from your branch as well

jolelievre (changes requested):
One file to remove and we're good to go

jolelievre (dismissed):

install-dev/upgrade/sql/1.7.7.0.sql
```diff
@@ -2,3 +2,4 @@ SET SESSION sql_mode='';
 SET NAMES 'utf8';
 
 ALTER TABLE `PREFIX_search_word` MODIFY `word` VARCHAR(30) NOT NULL;
+INSERT INTO `PREFIX_configuration` (`name`, `value`, `date_add`, `date_upd`) VALUES('PS_DISPLAY_MANUFACTURERS', '1', NOW(), NOW());
```
PierreRambaud:
```suggestion
INSERT INTO `PREFIX_configuration` (`name`, `value`, `date_add`, `date_upd`) VALUES ('PS_DISPLAY_MANUFACTURERS', '1', NOW(), NOW());
```

matthieu-rolland:
@PierreRambaud thanks for your input, I implemented your suggestion

src/PrestaShopBundle/Resources/views/Admin/macros.html.twig
```diff
@@ -39,7 +39,7 @@
 {% macro infotip(text)%}
 <div class="alert alert-info" role="alert">
   <p class="alert-text">
-    {{ text }}
+    {{ text|raw }}
```
PierreRambaud: Any reason to use raw here?

matthieu-rolland: It is so that html is interpreted, otherwise when our translated message contains a link, that is to say a html tag (<a href ...) it will be displayed without being interpreted.

matthieu-rolland: @PierreRambaud would you prefer something like this?

`{{ text|striptags('<a>')|raw}}`

So that only `<a>` tags are accepted ?

matthieu-rolland: @PierreRambaud  up :)

matks: I think he would prefer no `raw` at all 😄

Do you know which usecases require to display such links ? It would be better to treat them as edge-cases with dedicated process. Something like:

```
{% if message.is_regular %}
    {{ text }}
{% elseif message.has.link %}
    {{ display_message_with_link(message) }}
{%  endif %}
```

PierreRambaud: If you change the behaviour by adding `raw`, you must be sure all variables using this macro are escaped.
Good idea from @matks :+1:

matthieu-rolland: The idea of `{{ text|striptags('<a>')|raw}}`was that striptags escapes everything except `<a>` tags, and the raw makes sure the non escaped `<a>` tag is interpretated. 

So all variables using this macro would be escaped (except their `<a>` tags).

But I like @matks 's solution too, I will do that.  👍

matks: 🤔I dont think scripttags is reliable
Here what we want to avoid are XSS vulnerabilities, and you can encode tags in a looooot of different ways. I dont think scripttags removes them all

matks: See https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet

matthieu-rolland: @matks @PierreRambaud 
Ok I made a modification: instead of detecting if the link has a `<a>`tag or not, I made it possible to pass allowed tags as a parameter in the infotip macro function, if no allowed tag is specified then raw is not used at all, so that there are no breaking changes.

What do you think?

matthieu-rolland: damn I just saw your previous message @matks, so how do you suggest I display a variable having a `<a>` tag in it, without using raw ? 🤔

PierreRambaud: Striptags can remove unwanted characters

```
Hi, <:) we found it
```
will display
```
Hi,
```

Do we really need to add a link in the infotip?
Or do we want to add create a special infotip which is not escape? Maybe a `infotipraw`

matthieu-rolland: @PierreRambaud 

Well the link was not asked at the beginning, but suggested in this PR during the code review.
I do think it is useful for the end user though.

Do you mind if I add a raw parameter to the infotip function, whose default value is false, so that we don't duplicate the html inside the macro ?

matthieu-rolland: (that's what I did)

src/PrestaShopBundle/Resources/views/Admin/macros.html.twig
```diff
@@ -36,10 +36,14 @@
   </span>
 {% endmacro %}
 
-{% macro infotip(text)%}
+{% macro infotip(text, raw = false)%}
```
PierreRambaud:
```suggestion
{% macro infotip(text, use_raw = false)%}
```

To be sure there is zero conflict between `|raw` and the variable.

src/PrestaShopBundle/Resources/views/Admin/macros.html.twig
PierreRambaud: If the link was suggested, I'm more in favor to remove it and keep the original behavior.
Otherwise if it's mandatory, adding a second parameter like you said if a good idea :).
At least it will be use only if necessary

src/PrestaShopBundle/Resources/views/Admin/macros.html.twig
matthieu-rolland: @PierreRambaud done !

matks:
@matthieu-rolland  it seems you wait for @PierreRambaud  validation of the changes you added following his code review, is that correct ?

matthieu-rolland:
@matks 
Yep ! (up) 

@PierreRambaud 
I think you resolved my change, but didn't approve the PR as a whole ?

PierreRambaud:
@matks @matthieu-rolland Still one issue unresolved. You checked it too fast :wink:

matthieu-rolland:
> @matks @matthieu-rolland Still one issue unresolved. You checked it too fast 😉

@PierreRambaud  Indeed I didn't see the spacing fix suggestion, I made a commit out of it !

PierreRambaud (approved):

Issue #14480, opened 2019-07-04
[SF Migration] Cover all CommandHandlers and QueryHandlers by behat tests


matks:
Type: tech-issue

We need to cover all CommandHandlers and QueryHandlers by behavior (behat) tests.

This is important because **this is actually the legacy behavior that is being tested**. So providing tests for them will
- increase the legacy code coverage
- allow us later to remove the legacy code with a safety net to make sure we do not introduce regressions

Here's the list, as of the 4th of July 2019:

### Manufacturers
- [x] Domain/Manufacturer/CommandHandler/BulkToggleManufacturerStatusHandler
- [x] Domain/Manufacturer/CommandHandler/BulkDeleteManufacturerHandler
- [x] Domain/Manufacturer/CommandHandler/DeleteManufacturerHandler
- [x] Domain/Manufacturer/CommandHandler/EditManufacturerHandler
- [x] Domain/Manufacturer/CommandHandler/ToggleManufacturerStatusHandler
- [x] Domain/Manufacturer/CommandHandler/AddManufacturerHandler
- [x] Domain/Manufacturer/QueryHandler/GetManufacturerForEditingHandler
- [x] Domain/Manufacturer/QueryHandler/GetManufacturerForViewingHandler

### Orders
- [x] Domain/Order/CommandHandler/DuplicateOrderCartHandler
- [x] Domain/Order/CommandHandler/AddOrderFromBackOfficeHandler

### Misc
- [x] Domain/Configuration/CommandHandler/SwitchDebugModeHandler
- [x] Domain/ShowcaseCard/CommandHandler/CloseShowcaseCardHandler
- [x] Domain/ShowcaseCard/QueryHandler/GetShowcaseCardIsClosedHandler
- [x] Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler

### Contacts
- [x] Domain/Contact/CommandHandler/AddContactHandler
- [x] Domain/Contact/CommandHandler/EditContactHandler
- [x] Domain/Contact/QueryHandler/GetContactForEditingHandler

### Categories
- [x] Domain/Category/CommandHandler/EditCategoryHandler
- [x] Domain/Category/CommandHandler/DeleteCategoryCoverImageHandler
- [x] Domain/Category/CommandHandler/DeleteCategoryHandler
- [x] Domain/Category/CommandHandler/BulkDeleteCategoriesHandler
[...]

PR #14670, opened 2019-07-16
Add Behat tests for Customer commands

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | Adds Behat tests for Setting private note about customer & setting required fields.
| Type?         | improvement
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Partially https://github.com/PrestaShop/PrestaShop/issues/14480
| How to test?  | Build should be green.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14670)

### DIFF
--- a/tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
@@ -32,6 +32,7 @@ use Country;
 use Customer;
 use Exception;
 use PrestaShop\PrestaShop\Adapter\Validate;
+use RuntimeException;
 
 class CustomerFeatureContext extends AbstractPrestaShopFeatureContext
 {
@@ -100,6 +101,40 @@ class CustomerFeatureContext extends AbstractPrestaShopFeatureContext
     {
         $this->checkCustomerWithNameExists($customerName);
         Context::getContext()->updateCustomer($this->customers[$customerName]);
+    }
+
+    /**
+     * @Given private note is not set about customer :reference
+     */
+    public function assertPrivateNoteIsNotSetAboutCustomer($reference)
+    {
+        /** @var Customer $customer */
+        $customer = SharedStorage::getStorage()->get($reference);
+
+        if ($customer->note) {
+            throw new RuntimeException(sprintf(
+                'It was expected that customer "%s" should not have private note.',
+                $reference
+            ));
+        }
+    }
+
+    /**
+     * @Then customer :reference private note should be :privateNote
+     */
+    public function assertPrivateNoteAboutCustomer($reference, $privateNote)
+    {
+        /** @var Customer $customer */
+        $customer = SharedStorage::getStorage()->get($reference);
+
+        if ($customer->note !== $privateNote) {
+            throw new RuntimeException(sprintf(
+                'It was expected that customer "%s" private note should be "%s", but actually is "%s".',
+                $reference,
+                $privateNote,
+                $customer->note
+            ));
+        }
     }
 
     /**
--- a/tests/Integration/Behaviour/Features/Context/Domain/CustomerFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/Domain/CustomerFeatureContext.php
@@ -26,6 +26,8 @@
 
 namespace Tests\Integration\Behaviour\Features\Context\Domain;
 
+use Customer;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Command\SetPrivateNoteAboutCustomerCommand;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\SetRequiredFieldsForCustomerCommand;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Query\GetRequiredFieldsForCustomer;
 use RuntimeException;
@@ -73,6 +75,19 @@ class CustomerFeatureContext extends AbstractDomainFeatureContext
         $requiredFields = $this->getSharedStorage()->get('customer_required_fields');
 
         $this->getCommandBus()->handle(new SetRequiredFieldsForCustomerCommand($requiredFields));
+    }
+
+    /**
+     * @When I set :privateNote private note about customer :reference
+     */
+    public function setPrivateNoteAboutCustomer($privateNote, $reference)
+    {
+        /** @var Customer $customer */
+        $customer = $this->getSharedStorage()->get($reference);
+
+        $this->getCommandBus()->handle(new SetPrivateNoteAboutCustomerCommand((int) $customer->id, $privateNote));
+
+        $this->getSharedStorage()->set($reference, new Customer($customer->id));
     }
 
     /**
--- a/tests/Integration/Behaviour/Features/Context/Domain/CustomerFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/Domain/CustomerFeatureContext.php
@@ -1,0 +1,91 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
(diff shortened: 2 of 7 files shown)


sarjon:
@matks a few tests, if build is green I think you could merge it if no code changes are needed. :)

tests/Integration/Behaviour/Features/Context/Domain/CustomerFeatureContext.php
```diff
@@ -0,0 +1,106 @@
...
+use Customer;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Command\SetPrivateNoteAboutCustomerCommand;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Command\SetRequiredFieldsForCustomerCommand;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Query\GetRequiredFieldsForCustomer;
+use RuntimeException;
+
+class CustomerFeatureContext extends AbstractDomainFeatureContext
+{
+    /**
+     * @Given /^"(Partner offers)" is "(required|not required)"$/
```
matks, line 38: The value `Partner offers` is hardcoded, which makes this step not reusable for other fields 🤔 ?

tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
```diff
@@ -102,6 +103,40 @@ public function setCurrentCustomer($customerName)
         Context::getContext()->updateCustomer($this->customers[$customerName]);
     }
 
+    /**
+     * @Given private note is not set about customer :reference
```
matks, line 107:
```suggestion
     * @Given there is no private note registered for customer :reference
```

tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
```diff
@@ -102,6 +103,40 @@ public function setCurrentCustomer($customerName)
...
+        if ($customer->note) {
+            throw new RuntimeException(sprintf(
+                'It was expected that customer "%s" should not have private note.',
+                $reference
+            ));
+        }
+    }
+
+    /**
+     * @Then customer :reference private note should be :privateNote
```
matks, line 123:
```suggestion
     * @Then customer :reference private note content should be :privateNote
```
Maybe we should use a [Bheat pystring](https://docs.behat.org/en/v2.5/guides/1.gherkin.html#pystrings) here in order to allow more complex notes to be specified ?

tests/Integration/Behaviour/Features/Context/Domain/CustomerFeatureContext.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public function saveSpecifiedRequiredFields()
+    {
+        $requiredFields = $this->getSharedStorage()->get('customer_required_fields');
+
+        $this->getCommandBus()->handle(new SetRequiredFieldsForCustomerCommand($requiredFields));
+    }
+
+    /**
+     * @When I set :privateNote private note about customer :reference
```
matks, line 81:
```suggestion
     * @When I write a private note for customer :reference with the content :content
```
Let's aim for scenarios that sound, as much as possible, feature-oriented 😉 

I know this is hard because the BO is basically a big CRUD. But we must try to avoid using generic CRUD-related words like "save", "set", "edit", "modify".

Example: when we migrate Customer Service page, instead of writing "I save a message for customer X" we can use "I answer to customer X message" 😉 this sentence carries the meaning, the goal, rather than the action.

tests/Integration/Behaviour/Features/Scenario/Customer/customer_private_note.feature
```diff
@@ -0,0 +1,12 @@
+# ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml -s customer
+@reset-database-before-feature
```
matks, line 2: I'm not sure we should write 1 dedicated file for it 🤔 are we not going to create _a lot_ of folders then ?

I understand you want to split the scenarios, but if a file contains only 1 scenario, it looks like too much ?

matks (changes requested):

tests/Integration/Behaviour/Features/Context/Domain/CustomerFeatureContext.php
sarjon: Not really, it's the only field that is available for customers, there might be more in future, so I left it like so.

tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
sarjon: I'm afraid your wording suggestion is a bit wrong, there is not command for "register private note for customer" nor it is written like so in BO, it's more like "setting private note about customer" instead. :thinking:

tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
sarjon: Not sure, note is a simple string.

Regarding your wording, do we really need to point out "content"? There is nothing else (header, footer & etc) about it but note. :thinking:

tests/Integration/Behaviour/Features/Context/Domain/CustomerFeatureContext.php
sarjon: > words like "save", "set", "edit", "modify".

I did use "set". :smile: 

Actually, I've had discussion quite some time ago with Pablo about this private note and we settled with "set private note **about** customer". Using "save/set private note **for** customer" would be wrong, because it is not visible for customer itself, only for BO admins, that's why "**about** customer" was kept, as it is private note. This is very feature-oriented imo. :)

tests/Integration/Behaviour/Features/Scenario/Customer/customer_private_note.feature
sarjon: Yeah, I agree, but these are "feature" files for a reason, we can't put everything into customer management (which is basically CRUD feature) or some other feature file, do we only have 1 feature about Customers?

How can we improve that? :thinking: 

I see that Sylius, which heavily depends on Behat are splitting their features into a lot of different feature-files. One of examples https://github.com/Sylius/Sylius/tree/master/features/payment/managing_payments Some of them have 1 scenario, others more.

sarjon:
Tests needs to be rerun.

PierreRambaud:
@sarjon Done, waiting for CI.

tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
matks: I'm just trying to remove the "set" word that is very generic, in order to find something more relatable or more pictureable 😅 

Any other idea ?

tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
matks: > Not sure, note is a simple string.
Then we can use a single-column-single-row Behat `TableNode`

```
* @Then customer :reference private note content should be :
| Hey this is a not I wrote for Sarunas ! |
```

> Regarding your wording, do we really need to point out "content"? There is nothing else (header, footer & etc) about it but note. 🤔

Maybe it was overthinking 😺 thanks for pointing this out

tests/Integration/Behaviour/Features/Context/Domain/CustomerFeatureContext.php
matks: OK for the "about", what about the "write" then 😄 ?

tests/Integration/Behaviour/Features/Scenario/Customer/customer_private_note.feature
matks: Mmmm

All right ☺️ 

Even if later we find we have written too many files, we can merge them because tests do not fall under the SemVer "no BC !" rule

tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
sarjon: Yes, "set", like "add", "edit" and other words is "generic", but is it bad? Even command is called `SetPrivateNoteAboutCustomerCommand`. :thinking:

tests/Integration/Behaviour/Features/Context/Domain/CustomerFeatureContext.php
sarjon: Hmm, the use case is to set private note, the command says "set", why add new wording. :thinking:

tests/Integration/Behaviour/Features/Scenario/Customer/customer_private_note.feature
sarjon: Exactly. :+1:

matks (approved):

matks:
Thank you @sarjon

Issue #12640, opened 2019-02-20
[SF Migration] Modern Tabs with symfony routes


jolelievre:
**Is your feature request related to a problem? Please describe.**
The current Tab can only point to a legacy controller using the `class_name` property. So to point to a Symfony route you need to have a `_legacy_link` and use it in the Tab.

Another problem is with the automatic Tabs registering, which uses the `$tabs` property of modules. It only works if the Tab you are creating has a `class_name` which matches with an existing file in your module, if the file does not exist the Tab is not created so you need to create it manually with a `new Tab()`.

**Describe the solution you'd like**
We should add a `route_name` field in Tab so that the menu can use it when rendering the menu: if `class_name` is present use `getAdminLink`, if `route_name` is present use the Symfony router.
We need to update `ModuleTabRegister::registerTabs` so that it can manage new Tabs with `route_name`

jolelievre:
Fixed by https://github.com/PrestaShop/PrestaShop/pull/14679

PR #14679, opened 2019-07-16
Introduce automatic Tab registration with symfony routes, menu now handles disabled modules

### LINKED ISSUE #11578: Manage enable/disable in ModuleTabManagementSubscriber
The class `ModuleTabManagementSubscriber` manages the installation and uninstallation of modules. It automatically register/unregister tabs based on the `getTabs` method and the controller files present in the module folder.
The problem is that it does not manage enable/disable so when you disable a module, if it does not manage it itself, the tab is still present.

The class `ModuleTabManagementSubscriber` needs to listen to `ModuleManagementEvent::ENABLE` and `ModuleManagementEvent::DISABLE` events. It can't simply unregister/register the tabs because they are removed and if the tab was created manually by the module you lose any reference to it.

So we need to "hide" these tabs, we can't rely on the `active` field because some tabs are volontarily created inactive for module controllers and must not be shown. So we would reactivate them when we re-enable the module.

So in order to manage this behavior we need to add a new field like `visible` in order to manage this hide/show feature. And of course the side menu needs to react appropriately.

**Additional context**
Version: 1.7.6

jolelievre:
| Questions     | Answers
| Branch?       | develop
| Description?  | Manage automatic tab registration with symfony tabs (even when controller is missing), the menu is now able to use this new `route_name` parameter to generate the urls. The enable/disable feature now hides the tabs linked to the module automatically
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #11578 and #12640
| How to test?  | You can use this branch of the ps_linklist module https://github.com/PrestaShop/ps_linklist/pull/70 It uses the new automatic tab creation and symfony route, just check that after installation its tab is correctly displayed (right url)<br/>Then you can test the enable/disable feature on this module, when the module is disabled the link in the menu disappears

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14679)

### DIFF
--- a/classes/Tab.php
+++ b/classes/Tab.php
@@ -35,6 +35,9 @@ class TabCore extends ObjectModel
     /** @var string Class and file name */
     public $class_name;
 
+    /** @var string Route name for Symfony */
+    public $route_name;
+
     public $module;
 
     /** @var int parent ID */
@@ -66,6 +69,7 @@ class TabCore extends ObjectModel
             'position' => array('type' => self::TYPE_INT, 'validate' => 'isUnsignedInt'),
             'module' => array('type' => self::TYPE_STRING, 'validate' => 'isTabName', 'size' => 64),
             'class_name' => array('type' => self::TYPE_STRING, 'required' => true, 'size' => 64),
+            'route_name' => array('type' => self::TYPE_STRING, 'required' => false, 'size' => 256),
             'active' => array('type' => self::TYPE_BOOL, 'validate' => 'isBool'),
             'hide_host_mode' => array('type' => self::TYPE_BOOL, 'validate' => 'isBool'),
             'icon' => array('type' => self::TYPE_STRING, 'size' => 64),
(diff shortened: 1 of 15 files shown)


classes/Tab.php
```diff
@@ -137,11 +145,14 @@ public static function initAccess($idTab, Context $context = null)
         $slug = 'ROLE_MOD_TAB_' . strtoupper(self::getClassNameById($idTab));
 
         foreach (array('CREATE', 'READ', 'UPDATE', 'DELETE') as $action) {
-            Db::getInstance()->execute('INSERT INTO `' . _DB_PREFIX_ . 'authorization_role` (`slug`) VALUES ("' . $slug . '_' . $action . '")');
+            //Check if authorization role does not exist (this can happen if you want to create several tabs with the same class_name or route_name)
```
PierreRambaud: Maybe more readable?
```php
/**
 * Check if authorization role does not exist.
 * This can happen if you want to create several tabs with
 * the same class_name or route_name.
 */
$slug = pSQL($slug . '_' . $action);
$authorizationRoleExist = Db::getInstance()->numRows(
    'SELECT 1 FROM `' . _DB_PREFIX_ . 'authorization_role` ' .
    'WHERE `slug` = "' . $slug . '"'
);

if ($authorizationRoleExist === 0) {
    Db::getInstance()->execute(
        'INSERT INTO `' . _DB_PREFIX_ . 'authorization_role` (`slug`) VALUES ("' . $slug . '")'
    );
}

```

install-dev/upgrade/sql/1.7.7.0.sql
```diff
@@ -1,4 +1,6 @@
-SET SESSION sql_mode='';
+SET SESSION sql_mode = '';
 SET NAMES 'utf8';
 
 ALTER TABLE `PREFIX_search_word` MODIFY `word` VARCHAR(30) NOT NULL;
+
+/* PHP:ps_1770_preset_tab_enabled(); */;
```
matthieu-rolland, line 6: Shouldn't this method **ps_1770_preset_tab_enabled** be called outside of a sql file?  (or is it somehow parsed and executed?)

PierreRambaud: It's parsed and executed https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Install/Upgrade.php#L456

mickaelandrieu (approved):

PierreRambaud (approved):

sarahdib:
@jolelievre 

All backoffice menu disappear after applying this PR 

image

sarahdib:
it's ok with a fresh install :) 
Thank you @jolelievre

PierreRambaud:
Thanks @jolelievre

matks:
@jolelievre Does it fully handles the need expressed in https://github.com/PrestaShop/PrestaShop/issues/12640 ? Can we close the ticket https://github.com/PrestaShop/PrestaShop/issues/12640 😄?

jolelievre:
@matks yes this PR deals with both issues, so I'm going to close the #12640

PR #14682, opened 2019-07-16
Add preview extension for Grids

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | New orders list will require "Preview" feature of products, this PR introduces it
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/14732
| How to test?  | :warning: Build assets before testing :warning: When hover over row in new orders list you should see "drop-down" icon next to id, which after clicking loads order preview. List is accessible from `/admin-dev/index.php/sell/orders/orders` URL.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14682)


sarjon:
@matks here you go. :) This is main component for Preview, we can improve UI later on.

sarjon:
What the ..., another random build failure. :open_hands:

matks:
@sarjon Nice !
Now comes the question: how do we handle the content shown in the preview 🤔.

The most simple idea seems to create a dedicated controller action, triggered by an ajax call, that returns some HTML. But it's resource-consuming as each click on "preview" will perform an ajax call...
Another good point for this idea: very easy to modify/extend the behavior.

sarjon:
I think that having endpoint like `/orders/1/preview` which returns JSON would be the best. It's totally decoupled from list and can be used elsewhere if needed. 

> But it's resource-consuming as each click on "preview" will perform an ajax call...

Hmm, I really don't think it's a problem, it's just an ajax, like any other. :man_shrugging:

matks:
It's supposed to be a "quick preview" and it's useful only if it's faster to use than loading the whole page 😅 so it needs to be quick enough to provide added value

Why return JSON ? Why not html to put in the modal ?

PierreRambaud:
>Why return JSON ? Why not html to put in the modal ?

Because it's harder to customize. What if you want to return more than one html content?
It's really easier to do 

```json
{
  "res_products": "<div ....",
  "res_categories": "<div ..."
}
```

Than a single HTML block.

matks:
> > Why return JSON ? Why not html to put in the modal ?
> 
> Because it's harder to customize. What if you want to return more than one html content?
> It's really easier to do
> 
> ```json
> {
>   "res_products": "<div ....",
>   "res_categories": "<div ..."
> }
> ```
> 
> Than a single HTML block.

But then you must have some JS-side rendering 🤔 is'nt it a bit overcomplicated for our usecase ?

sarjon:
> Why return JSON ? Why not html to put in the modal ?

Because it's not a modal in the first place. :smile: 

We wouldn't load whole order information, only data that is needed for preview. I've seen poorly written modules do that, and they don't have any performance issues.

PierreRambaud:
> > > Why return JSON ? Why not html to put in the modal ?
> > 
> > 
> > Because it's harder to customize. What if you want to return more than one html content?
> > It's really easier to do
> > ```json
> > {
> >   "res_products": "<div ....",
> >   "res_categories": "<div ..."
> > }
> > ```
> > 
> > 
> > Than a single HTML block.
> 
> But then you must have some JS-side rendering  is'nt it a bit overcomplicated for our usecase ?

Not really, many modules can add a hook to add data in the returns value, and populate whatever they want. It's really easier than doing another ajax request.
I'm just thinking about modules whose want to add modal, zoom, or many other information in the preview block (or outside this one).

admin-dev/themes/new-theme/js/pages/order/index.js
```diff
@@ -34,11 +34,30 @@ import SubmitBulkExtension from '../../components/grid/extension/submit-bulk-act
...
 
 const $ = window.$;
 
 $(() => {
+  function previewRenderer(id) {
+    return new Promise((resolve, reject) => {
+      $.ajax({
+        url: this.previewDataUrl,
+        method: 'GET',
+        dataType: 'html',
```
PierreRambaud: :scream: 
No possibility to get a json with exploitable data?

sarjon: What do you mean? Instead of HTML we should get JSON data and render template from JS?

PierreRambaud: Yes, and because if a module add something in its template and json, he can automatically add data (maybe we should also trigger js events here)

sarjon: Yeah, here are 2 ways I can see:

1. Get HTML from ajax, in that case we could dispatch hook to allow modules to add custom content to final template.

2. Get JSON data from ajax, but in that case, we would need to render template in JS. It could be a bit more difficult for both us and module developers, because 1. We would need to manage translations in JS 2. Complex template rendering in JS would be harder than in Twig 3. It's harder for module developers to deal with JS event and adding custom content in JS than to execute hook when rendering from backend.

What do you think @PierreRambaud?

PierreRambaud: 3. Get JSON data from ajax with `{"htmlContent": "blablabl", "myCompanyHook": "Blabalbal"}` and add a `prestashop.emit('onOrderPreviewDataUrlSomething', {eventType: 'blabla', resp: resp}`. More because you maybe wanted to execute or render something elsewhere the BO is rendering the `htmlContent`. So you can keep the rendering with PHP (if this one is too complicated to render with vuejs or jquery), and also trigger a custom event for modules.

admin-dev/themes/new-theme/js/pages/order/index.js
```diff
@@ -34,11 +34,34 @@ import SubmitBulkExtension from '../../components/grid/extension/submit-bulk-act
...
+import PreviewExtension from '../../components/grid/extension/preview-extension';
 
 const $ = window.$;
 
 $(() => {
+  // @todo: I think previewRenderer() should accept `id` and some kind of `params` argument,
+  // so grid can pass other data (like URL) to preview renderer
+  // in case it uses ajax to load content.
+  // @todo: Extract into separate file.
+  function previewRenderer(id) {
```
sarjon: Seems we don't need ID anymore?

admin-dev/themes/new-theme/js/pages/order/index.js
```diff
@@ -34,11 +34,34 @@ import SubmitBulkExtension from '../../components/grid/extension/submit-bulk-act
...
+    console.log($('.preview-toggle').data('preview-data-url'));
+    return new Promise((resolve, reject) => {
+      $.ajax({
+        url: $('.preview-toggle').data('preview-data-url'),
```
sarjon: URL should be passed as function argument, otherwise, how do you know it takes correct URL?

src/Core/Domain/Order/Query/GetOrderPreview.php
```diff
@@ -0,0 +1,53 @@
...
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Order\Query;
+
+use PrestaShop\PrestaShop\Core\Domain\Order\ValueObject\OrderId;
+
+class GetOrderPreview
```
sarjon, line 31: Maybe `GetOrderForPreviewing`? So it follows same naming as other queries.

src/Core/Domain/Order/QueryHandler/GetOrderPreviewHandler.php
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * @param Connection $connection
+     * @param string $dbPrefix
+     * @param int $contextLangId
+     */
+    public function __construct(
+        Connection $connection,
+        string $dbPrefix,
+        int $contextLangId
```
sarjon: I'd pass `$langId` in query instead, @matks wdyt? Let's stop injecting context, it will be future refactoring easier.

src/Core/Domain/Order/QueryHandler/GetOrderPreviewHandler.php
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(GetOrderPreview $query): OrderPreview
+    {
+        $orderId = $query->getOrderId()->getValue();
+        $qb = $this->getCommonQueryBuilder($orderId);
+
+        return new OrderPreview(
+            $this->getInvoiceDetails($qb)
```
sarjon: I suppose there will be more data?

src/Core/Grid/Definition/Factory/OrderGridDefinitionFactory.php
```diff
@@ -144,6 +145,18 @@ protected function getColumns()
                     'field' => 'id_order',
                 ])
             )
+            ->add(
+                (new PreviewColumn('preview'))
+                    ->setOptions([
+                        'text_type' => 'static',
+                        'static_text' => '<',
+                        'identifier_field' => 'id_order',
+                        'preview_data_route' => 'admin_orders_get_preview_data',
```
sarjon: How about `preview_route`?

src/Core/Grid/Definition/Factory/OrderGridDefinitionFactory.php
```diff
@@ -144,6 +145,18 @@ protected function getColumns()
...
+                        'preview_record_params' => [
```
sarjon: How about `preview_route_params`?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -215,6 +217,17 @@ public function exportAction(OrderFilters $filters)
             ->setFileName('order_' . date('Y-m-d_His') . '.csv');
     }
 
+    public function getPreviewDataAction(Request $request): Response
```
sarjon: You could inject `$orderId` as method argument.

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -215,6 +217,17 @@ public function exportAction(OrderFilters $filters)
...
+    {
+        $orderId = $request->query->getInt('id');
+        /** @var OrderPreview $orderPreview */
+        $orderPreview = $this->getQueryBus()->handle(new GetOrderPreview($orderId));
+
+        return $this->render('@PrestaShop/Admin/Sell/Order/Order/preview.html.twig', [
+            'invoiceDetails' => $orderPreview->getInvoiceDetails(),
```
sarjon: Pass whole `$orderPreview` to template instead.

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/orders.yml
```diff
@@ -43,3 +43,10 @@ admin_orders_export:
   defaults:
     _controller: PrestaShopBundle:Admin/Sell/Order/Order:export
     _legacy_controller: AdminOrders
+
+admin_orders_get_preview_data:
+  path: /preview-data
```
sarjon: It should be `/{orderId}/preview` instead imo.

src/Core/Grid/Definition/Factory/OrderGridDefinitionFactory.php
zuk3975: wanted to point out that values are taken from records. In case somebody needs to "hardcode" the parameter directly in gridDefinitionFactory?

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
```diff
@@ -0,0 +1,128 @@
...
+    this.expandIcon = '.js-expand';
+    this.collapseIcon = '.js-collapse';
+    this.previewOpen = 'preview-open';
+    this.previewToggle = '.preview-toggle';
+  }
+
+  extend(grid) {
+    grid.getContainer().find('tbody tr').on('mouseover mouseleave', (event) => {
+      this._handleIconHovering(event);
+    });
```
tomas862: less code is better code - you can write `(event) => this._handleIconHovering(event)`

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
```diff
@@ -0,0 +1,128 @@
...
+      $columnRow.next('.preview-row').remove();
+      $columnRow.removeClass('preview-open');
+      this._showExpandIcon($columnRow);
+      this._hideCollapseIcon($columnRow);
+
+      return;
+    }
+    const dataUrl = $(event.currentTarget).data('preview-data-url');
+
+    Promise.resolve(this.renderer(dataUrl).then((result) => {
```
tomas862: what happens when error appears - seems like its not handled at the moment

admin-dev/themes/new-theme/scss/pages/_orders.scss
```diff
@@ -3,4 +3,18 @@
...
+      border: none;
+      padding-top: 3.2px;
+      padding-bottom: 3.2px;
+    }
+    .details td, .product td {
+      padding-top: 0;
+      padding-bottom: 0;
+    }
+    .product th {
+      background-color: #eaebec;
```
tomas862: maybe introduce new color #eaebec in `/var/www/html/PrestaShop/admin-dev/themes/default/sass/modules/_variables.sass` ?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -215,12 +216,32 @@ public function exportAction(OrderFilters $filters)
...
+        try {
+            /** @var OrderPreview $orderPreview */
+            $orderPreview = $this->getQueryBus()->handle(new GetOrderPreview($orderId));
+
+            return $this->json([
+                'preview' => $this->renderView('@PrestaShop/Admin/Sell/Order/Order/preview.html.twig', [
+                    'orderPreview' => $orderPreview,
+                ]),
+            ]);
+        } catch (Exception $e) {
```
tomas862, line 473: there are million reasons to catch unintended Exception here sometimes hiding true errors which should be displayed in any case. Can you catch here more predictable and specific exception or multiple exceptions if such occur? :thinking:  @matks is it related with modules throwing their errors? Here's only query handler being initialized

tomas862:
todo: 

1. double click loads two previews
2. loading animation somewhere?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
```diff
@@ -0,0 +1,121 @@
...
+        </tr>
+      </tbody>
+    </table>
+    <div class="offset-2 col-6 ">
+      <div class="table-responsive">
+        <table class="table product table-borderless border-bottom-0">
+          <thead>
+          <tr>
+            {% if orderPreview.taxIncluded %}
+              {% set taxInclusion = 'Tax included'|trans({}, 'Admin.Catalog.Feature') %}
```
LouiseBonnard:
```suggestion
              {% set taxInclusion = 'Tax included'|trans({}, 'Admin.Global') %}
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
```diff
@@ -0,0 +1,121 @@
...
+            {% else %}
+              {% set taxInclusion = 'Tax excluded'|trans({}, 'Admin.Catalog.Feature') %}
```
LouiseBonnard:
```suggestion
              {% set taxInclusion = 'Tax excluded'|trans({}, 'Admin.Global') %}
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
```diff
@@ -0,0 +1,121 @@
...
+              <td class="p-1">{{ productDetail.totalPrice }}</td>
+            </tr>
+          {% endfor %}
+          </tbody>
+        </table>
+      </div>
+
+      <div class="text-right">
+        <a href="{{ path('admin_orders_view', {'orderId': orderId}) }}" class="btn btn-primary btn-sm">
+          {{ 'Open order'|trans({}, 'Admin.Orderscustomers.Feature') }}
```
LouiseBonnard: Open order? Does it mean 'see order'?

sarjon: Yeah, it opens order view page. I followed new UI design which had this wording.

LouiseBonnard: Alright, it looks weird since it is not a wording often employed in PrestaShop and, also, it does not open a popup but the order page. Here are my suggestions for this button:

_View order_ in _Admin.Orderscustomers.Feature_
_View full details_ in _Admin.Actions_
_See more_ in _Admin.Actions_

cc @MatShir

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
```diff
@@ -0,0 +1,196 @@
...
+    this._lock(dataUrl);
+
+    $.ajax({
+      url: dataUrl,
+      method: 'GET',
+      dataType: 'json',
+    }).then((response) => {
+      this._renderPreviewContent($columnRow, response.preview);
+
+      this._unlock(dataUrl);
```
PierreRambaud: Move this line in `complete` to don't duplicate it in catch method.

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
```diff
@@ -0,0 +1,196 @@
...
+
+      return;
+    }
+    const dataUrl = $(event.currentTarget).data('preview-data-url');
+
+    if (this._isLocked(dataUrl)) {
+      return;
+    }
+
+    // Prevents loading preview multiple times.
```
matks, line 100: 👍

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
```diff
@@ -0,0 +1,196 @@
...
+   *
+   * @param $columnRow
+   * @param content
+   *
+   * @private
+   */
+  _renderPreviewContent($columnRow, content) {
+    const rowColumnCount = $columnRow.find('td').length;
+
+    const previewTemplate = `
```
matks, line 129: HTML in JS... are you a [JSX](https://www.w3schools.com/react/react_jsx.asp) fan ? :trollface: 

I guess here it's hard to not use it, but I dont feel good about it 😐 
Any suggestions @eternoendless @PierreRambaud ?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
matks: Yes it's related to modules throwing errors.

Let's leave this like it for now, as this issue is global and has been raised by @sarjon in https://github.com/PrestaShop/PrestaShop/issues/15732. Following the solution we'll choose for https://github.com/PrestaShop/PrestaShop/issues/15732 we'll apply it everywhere

matks (commented):
Looks good to me 👍 please address other reviewers comment and then we're good !

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
sarjon: I don't think it's a big deal, I wanted to use `Promise` API, but I can use jQuery's `complete` as well.

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
sarjon: I kind of like JSX. :D But I'm open for suggestions if you have any. :)

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
PierreRambaud: If you want to use Promise, use `finally` :wink:

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
sarjon: So which wording should I use @LouiseBonnard ? :)

src/Core/Domain/Order/Query/GetOrderPreview.php
sarjon: @matks wdyt?

matks: Both names look good to me 👍 Preview or Previewing
I dont think we need to have all query naming uniform, Queries and Commands are wrappers for Actions

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
PierreRambaud: I'm not a big fan to have this kind of template in JS, but, afaik there is no proper solution by using jquery / vanilla js (`<template>` is not supported everywhere). If only we were on VueJs here :sob:

tomas862: > I'm not a big fan to have this kind of template in JS, but, afaik there is no proper solution by using jquery / vanilla js (`<template>` is not supported everywhere). If only we were on VueJs here

web components maybe? With polyfills it also supports IE11 I think.

PierreRambaud: Otherwise if you want to do it more simple, pass this template as PreviewExtension's parameter, and this template is stored in a twig template :thinking: 
At least, people will be able to improve/edit it.

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
LouiseBonnard: I have even better! I suggest _Order details_ in _Admin.Orderscustomers.Feature_, it will be consistent with the order wordings on this page.
💡

sarjon: Done!

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
sarjon: I'll use old school `complete` since jQuery does not support `finally()` :man_shrugging:

admin-dev/themes/new-theme/js/components/grid/extension/preview-extension.js
sarjon: We would end up with something like this:
```twig
{% block preview_template %}
<tr class="preview-row">
      <td></td> {# Maybe even without <td> #}
</tr>
{% endblock %}
```

It would be so easy to break things up (we need to remember to add copy/paste template every time, it's easy to mess it up), developers are free implement their own previews if they wish to.

Does it really makes sense to move it away from JS and add dependency?

matks: No it would be overkill indeed

matks (dismissed):

Robin-Fischer-PS:
Hi @sarjon :)

Can you resolve the conflicting files before testing ?

Thanks !

matks:
@Robin-Fischer-PS Conflicts solved 👍

matks (dismissed):

Robin-Fischer-PS:
Hi @colinegin @MatShir ! This PR seems to differ from the spec (https://github.com/PrestaShop/PrestaShop/issues/14732). Could you confirm that the PR is consistent with expected behavior or, if not, make your returns ?

Thanks :)

sarjon:
@LouiseBonnard I think this PR needs wording update once again. :(

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
```diff
@@ -0,0 +1,154 @@
...
+                <strong>{{ 'Carrier'|trans({}, 'Admin.Global') }}:</strong>
+                {% if not orderPreview.isVirtual %}
+                  {{ orderPreview.shippingDetails.carrierName }}
+                {% else %}
+                  -
+                {% endif %}
+              </p>
+
+              <p class="mb-0">
+                <strong>{{ 'Tracking ID'|trans({}, 'Admin.Global') }}:</strong>
```
LouiseBonnard:
```suggestion
                <strong>{{ 'Tracking number'|trans({}, 'Admin.Shipping.Feature') }}:</strong>
```

LouiseBonnard: We should stay consistent and keep mentioning 'tracking number' since it is already used throughout the software - on the Shipping section of the back office and in the Order Details page. ;-)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
```diff
@@ -0,0 +1,154 @@
...
+                {% if not orderPreview.isVirtual and orderPreview.shippingDetails.trackingNumber is not empty %}
+                  {{ orderPreview.shippingDetails.trackingNumber }}
+                {% else %}
+                  -
+                {% endif %}
+              </p>
+
+              <p class="mb-2">
+                <strong>{{ 'Shipping details'|trans({}, 'Admin.Global') }}:</strong>
```
LouiseBonnard:
```suggestion
                <strong>{{ 'Shipping details'|trans({}, 'Admin.Orderscustomers.Feature') }}:</strong>
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
```diff
@@ -0,0 +1,154 @@
...
+        <div class="col">
+          <div class="row">
+            <div class="col-1">
+              <p class="mb-0">
+                <i class="material-icons pr-1">receipt</i>
+              </p>
+            </div>
+            <div class="col">
+              <p class="mb-1">
+                <strong>{{ 'Invoice details'|trans({}, 'Admin.Global') }}:</strong>
```
LouiseBonnard:
```suggestion
                <strong>{{ 'Invoice details'|trans({}, 'Admin.Orderscustomers.Feature') }}:</strong>
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
```diff
@@ -0,0 +1,154 @@
...
+          <tr>
+            {% if orderPreview.taxIncluded %}
+              {% set taxInclusion = 'Tax included'|trans({}, 'Admin.Global') %}
+            {% else %}
+              {% set taxInclusion = 'Tax excluded'|trans({}, 'Admin.Global') %}
+            {% endif %}
+
+            <th>{{ 'Product'|trans({}, 'Admin.Global') }}</th>
+            <th>{{ 'Reference'|trans({}, 'Admin.Global') }}</th>
+            <th>{{ 'Stock location'|trans({}, 'Admin.Global') }}</th>
```
LouiseBonnard:
```suggestion
            <th>{{ 'Stock location'|trans({}, 'Admin.Orderscustomers.Feature') }}</th>
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
```diff
@@ -0,0 +1,154 @@
...
+            {% endif %}
+          {% endfor %}
+          </tbody>
+        </table>
+      </div>
+
+      <div class="text-right">
+        <a href="{{ path('admin_orders_view', {'orderId': orderId}) }}" class="btn btn-primary btn-sm">
+          {% if orderPreview.productDetails|length > productsPreviewLimit %}
+            {{ 'View the other %count% products'|trans({'%count%': orderPreview.productDetails|length - productsPreviewLimit}, 'Admin.Orderscustomers.Feature') }}
```
LouiseBonnard: Do we agree that this will be a small call-to-action to list the other product(s) of the order inside the preview line? Or does it redirect to the order details page?

sarjon: It will redirect to Orders View page.

LouiseBonnard: In my opinion, it is pretty obvious that we cannot list all the products in a cart in the preview, so there is no need to emphasize it with a wording. We could suggest it with a design effect and therefore keep the number of products in the cart as in the mockup and keep one call-to-action at the end 'Order details'. I'll be at the stand-up tomorrow so we can discuss it all together. :-)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
```diff
@@ -0,0 +1,154 @@
...
+        <table class="table product table-borderless border-bottom-0">
+          <thead>
+          <tr>
+            {% if orderPreview.taxIncluded %}
+              {% set taxInclusion = 'Tax included'|trans({}, 'Admin.Global') %}
+            {% else %}
+              {% set taxInclusion = 'Tax excluded'|trans({}, 'Admin.Global') %}
+            {% endif %}
+
+            <th>{{ 'Product'|trans({}, 'Admin.Global') }}</th>
```
LouiseBonnard: Could we also display the number of products ordered like in the mock-up [here](https://projects.invisionapp.com/d/main?origin=v7#/console/18147703/376656277/preview): _Products (%count%)_?

sarjon:
@LouiseBonnard could you make Github suggestions? That would save sooo much time by jumping back and forth between PRs. :)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/preview.html.twig
sarjon: For me, anythings works, we already changed this text following feedback from Product team, but it seems it was not enough. 

I followed this wording that was feedback from product team https://github.com/PrestaShop/PrestaShop/issues/14732#issuecomment-541561339

sarjon:
It's ready once again. :rocket:

matks (dismissed):

Robin-Fischer-PS:
Hi @sarjon !

I've noticed two defects on this PR :

1/ The Shipping and Invoice Adress are missing some fields (postal code, city, VAT Number, and Company name).
See screenshot : 
PR14682 missing address info


It should be as in the order page : 
PR14682 missing address info - Legacy 1

2/ On product list, there should be an indication of total number of product (see the invision screen) : 
PR14682 missing products quantity Legacy 1

Otherwise, it appears good to me ! You should have soon a mail from the UX for some minors bugs on this page.

:smile:

sarjon:
Hi @Robin-Fischer-PS,

I've added missing fields. :)

matks (approved):

Robin-Fischer-PS:
Thanks @sarjon ! It's now QA Approved :smile:

matks:
As Louise has already reviewed this PR (she has not yet given the seal of approval because she has not validated the changes yet) and in order to avoid this PR to need rebases again, I merge it but we're listening to feedbacks if there is 😉and we'll handle them in another PR

PR #14684, opened 2019-07-16
Allow to edit order in valid scenarios

kpodemski:
| Questions     | Answers
| Branch?       | develop
| Description?  | Even tho order has Waiting for bank wire state we couldn't edit it
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14684)

### DIFF
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -514,19 +514,19 @@ class OrderCore extends ObjectModel
         $paid = false;
         $shipped = false;
         if ($filters > 0) {
-            if ($filters & OrderState::FLAG_NO_HIDDEN) {
+            if ($filters && OrderState::FLAG_NO_HIDDEN) {
                 $no_hidden = true;
             }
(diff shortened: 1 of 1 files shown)


PierreRambaud:
Could you please provide us more information?
Because `$filters` is already greater than 0 your change is equal to:
```php
        if ($filters > 0) {
            $no_hidden = true;
            $delivery = true;
            $logable = true;
            $paid = true;
            $shipped = true;
        }
```
Which I think, isn't a good idea :/

kpodemski:
@PierreRambaud you're right, this is my mistake, i have it edited in other way locally, i've tried to do changes using online code editor and do few mistakes, i'll close this one and open another PR with proper changes

overall, if you have order with pay by bank wire state you're not able to edit it, even tho you should

Issue #14686, opened 2019-07-16
Button erase all doesn't work for logs & email pages


Stephane1874:
**Describe the bug**

Hello the team,

Go to the advanced settings page> logs, tried to delete the logs, the button does not work.

**To Reproduce**
1. Go to advanced settings > logs
2. Click on Delete all
3. See error

**Screenshots**
![supprimer](https://user-images.githubusercontent.com/46202635/61301979-4cd6b680-a7e5-11e9-937c-df6853e94b9f.PNG)

**Additional information**
PrestaShop version: 1.7.6
PHP version: 7.1.3

khouloudbelguith:
Hi @Stephane1874,

Thanks for your report.
I manage to reproduce the issue with PS1.7.6.0 fresh install & after upgrade from PS1.7.5.
https://drive.google.com/file/d/1kRxIzVAU3WIKNYQ5dxY-wz-sCtMn64yF/view
In the PS1.7.5.1 => OK
I’ll add this to the debug roadmap so that it’s fixed. If you have already fixed it on your end or if you think you can do it, please do send us a pull request!
Thanks!

marionf:
Same issue on email page, erase all button doesn't work
But if you select all and delete with bulk action it's working

matks:
Fixed by https://github.com/PrestaShop/PrestaShop/pull/14701

arafatx:
This issue came back in 1.7.8.0. I will open a new issue.

PR #11095, opened 2018-10-20
Update grid design regarding grid gear icon position and dropdown toggle icon modification

tomas862:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | As mentioned in https://github.com/PrestaShop/PrestaShop/pull/10774#issuecomment-431363447 some design changes are needed on all lists
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #11080 #11077 
| How to test?  | Check the design updates according to https://github.com/PrestaShop/PrestaShop/pull/10774#issuecomment-431363447 and also test if the list actions which are now located in list header works

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/11095)

PR #14701, opened 2019-07-17
Fix log delete all action on Logs page

matks:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Fix log delete all action on Logs page
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/14686
| How to test?  | See ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14701)


PierreRambaud (approved):

admin-dev/themes/new-theme/js/components/grid/extension/submit-grid-action-extension.js
```diff
@@ -36,7 +36,7 @@ export default class SubmitGridActionExtension {
   }
 
   extend(grid) {
-    grid.getContainer().on('click', '.js-grid-action-submit-btn', (event) => {
+    grid.getHeaderContainer().on('click', '.js-grid-action-submit-btn', (event) => {
```
mickaelandrieu, line 39: is it a BC compared to 1.7.5?

How about existing grids in the community modules? /c @sarjon

sarjon: > is it a BC compared to 1.7.5?

Even if it is, it's part of 1.7.6 now.

I see that in 1.7.5 grid actions used to be inside container (retrieved by `getContainer()`) see https://prnt.sc/og9h0z, but in 1.7.6 actions where move outside container, see https://prnt.sc/og9hbv.

It seems wrong to have grid actions outside container. :confused:

matks: > I see that in 1.7.5 grid actions used to be inside container (retrieved by `getContainer()`) see https://prnt.sc/og9h0z, but in 1.7.6 actions where move outside container, see https://prnt.sc/og9hbv.

Indeed that's the reason behind this fix, and I agree with

> It seems wrong to have grid actions outside container. 😕

But it's now released in 1.7.6.0 😥

sarjon: Hmmm, I wonder what if we increase the scope of container to the element which is above to cover grid actions? In theory, everything should still be working as original container is still inside?

matks: The change happened in https://github.com/PrestaShop/PrestaShop/pull/11095

If existing grids built in 1.7.5 have been using
```
        {% include '@PrestaShop/Admin/Common/Grid/grid_panel.html.twig' with { 'grid': grid } %}
```
then actions have been moved correctly, so using the container header is fine
Thenn

PierreRambaud: It's not a BC since it select only the parent container. It's also working like before if the grid is inside the gridContainer

Issue #10547, opened 2018-09-20
Migrate "Catalog > Discounts > Cart rules > Add new / edit cart rule" page


matks:
Part of Symfony migration project

## Status

Todo:
- [x] Migrate of "add cart rule" action #14707 (CQRS)
- [x] Mgrate edit cart rule action (CQRS)
- [x] Migrate creation form (https://github.com/PrestaShop/PrestaShop/pull/31904). _Edited by @zuk3975_ )
- [ ] Migrate edition form (https://github.com/PrestaShop/PrestaShop/pull/31904). _Edited by @zuk3975_ )
- [ ] Fix "red borders" on cart rule tabs and redirect to tab is missing when form has errors (styling probably should be moved from product_page.scss to some global place like form.scss (or prestakit)?.). _Edited by @zuk3975_
- [ ] Migrate restrictions CQRS . _Edited by @zuk3975_ (left to finish https://github.com/PrestaShop/PrestaShop/pull/33076, https://github.com/PrestaShop/PrestaShop/pull/33050)
- [ ] Migrate restrictions form part. _Edited by @zuk3975_
- [ ] Add integration tests for CartRuleController and units for CartRuleFormDataProvider. _Edited by @zuk3975_
- [ ] Multistore header: [Specs](https://github.com/PrestaShop/prestashop-specs/blob/7e820f30347fe69ccd3db5a99ab02ef6f7bbdd54/back-office/multistoregeneralspecs.md#multistore-header-only-from-178). Multistore header might need to be removed (need some input from UX managers). As discussed with @jolelievre all the pages that only has shop associations, but not different values between shops, should be without multistore header, as it doesn't have any impact in the data anyway and just creates confusion. In such pages shop association is only handled in edition form. _Edited by @zuk3975_
- [ ] Don't forget to handle discount application type "selected products". Its not yet handled in creation PR. It should be allowed to select that type only when product restrictions are selected. _Edited by @zuk3975_
- [ ] Double check if ProductSearchType can be reused in more places by adding some options etc. (there are usages mostly in product page, which looks promises) _Edited by @zuk3975_
- [ ] Homogenize responsive .js-arrow in tabs to be reused (now it only works for product page as its js/css related code was only designed for product. See admin-dev/themes/new-theme/js/pages/product/nav-tabs.ts) _Edited by @zuk3975_
[...]

PR #14707, opened 2019-07-17
Migration of "add cart rule" action

rokaszygmantas:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migrates the logic of adding new cart rule. UI is not migrated, only CQRS command. Covered with unit and behat tests.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | A small part of this ticket: https://github.com/PrestaShop/PrestaShop/issues/10547
| How to test?  | Travis should pass, nothing for QA.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14707)

### DIFF
--- a/src/Adapter/CartRule/CommandHandler/AddCartRuleHandler.php
+++ b/src/Adapter/CartRule/CommandHandler/AddCartRuleHandler.php
@@ -205,7 +205,7 @@ final class AddCartRuleHandler implements AddCartRuleHandlerInterface
             case DiscountApplicationType::SPECIFIC_PRODUCT:
                 if (!$hasPercentageDiscount && !$hasAmountDiscount) {
                     throw new CartRuleConstraintException(
-                        'Cart rule, which is applied to a specific product, '.
+                        'Cart rule, which is applied to a specific product, ' .
                         'must have percentage or amount application type.',
                         CartRuleConstraintException::INCOMPATIBLE_CART_RULE_ACTIONS
                     );
@@ -225,7 +225,7 @@ final class AddCartRuleHandler implements AddCartRuleHandlerInterface
             case DiscountApplicationType::ORDER_WITHOUT_SHIPPING:
                 if (!$hasAmountDiscount && !$hasPercentageDiscount) {
                     throw new CartRuleConstraintException(
-                        'Cart rule, which is applied to whole order without shipping, '.
+                        'Cart rule, which is applied to whole order without shipping, ' .
                         'must have percentage or amount application type.',
                         CartRuleConstraintException::INCOMPATIBLE_CART_RULE_ACTIONS
                     );
--- a/src/Adapter/CartRule/CommandHandler/AddCartRuleHandler.php
+++ b/src/Adapter/CartRule/CommandHandler/AddCartRuleHandler.php
@@ -34,7 +34,6 @@ use PrestaShop\PrestaShop\Core\Domain\CartRule\Exception\CartRuleException;
 use PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject\CartRuleAction\CartRuleActionInterface;
 use PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject\CartRuleId;
 use PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject\DiscountApplicationType;
-use PrestaShopDatabaseException;
 use PrestaShopException;
 
 /**
@@ -104,8 +103,8 @@ final class AddCartRuleHandler implements AddCartRuleHandlerInterface
         $minimumAmount = $command->getMinimumAmount();
         $cartRule->minimum_amount = $minimumAmount->getMoneyAmount()->getAmount();
         $cartRule->minimum_amount_currency = $minimumAmount->getMoneyAmount()->getCurrencyId()->getValue();
-        $cartRule->minimum_amount_shipping = $minimumAmount->isShippingExcluded();
-        $cartRule->minimum_amount_tax = $minimumAmount->isTaxExcluded();
+        $cartRule->minimum_amount_shipping = !$minimumAmount->isShippingExcluded();
+        $cartRule->minimum_amount_tax = !$minimumAmount->isTaxExcluded();
 
         $cartRule->quantity = $command->getTotalQuantity();
         $cartRule->quantity_per_user = $command->getQuantityPerUser();
--- a/src/Adapter/CartRule/CommandHandler/AddCartRuleHandler.php
+++ b/src/Adapter/CartRule/CommandHandler/AddCartRuleHandler.php
@@ -1,0 +1,240 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Adapter\CartRule\CommandHandler;
+
+use CartRule;
+use PrestaShop\PrestaShop\Core\Domain\CartRule\Command\AddCartRuleCommand;
+use PrestaShop\PrestaShop\Core\Domain\CartRule\CommandHandler\AddCartRuleHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\CartRule\Exception\CartRuleConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\CartRule\Exception\CartRuleException;
+use PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject\CartRuleAction\CartRuleActionInterface;
+use PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject\CartRuleId;
+use PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject\DiscountApplicationType;
+use PrestaShopDatabaseException;
+use PrestaShopException;
+
+/**
+ * Handles adding new cart rule using legacy logic.
+ */
+final class AddCartRuleHandler implements AddCartRuleHandlerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(AddCartRuleCommand $command): CartRuleId
+    {
+        try {
+            $cartRule = $this->buildCartRuleFromCommandData($command);
+
+            if (false === $cartRule->validateFields(false) || false === $cartRule->validateFieldsLang(false)) {
+                throw new CartRuleConstraintException('Cart rule contains invalid field values');
+            }
+            if (false === $cartRule->add()) {
+                throw new CartRuleException('Failed to add new cart rule');
+            }
+        } catch (PrestaShopException $e) {
+            throw new CartRuleException('An error occurred when trying to add new cart rule');
+        }
+
+        return new CartRuleId((int) $cartRule->id);
+    }
+
+    /**
+     * @param AddCartRuleCommand $command
+     *
+     * @return CartRule
+     *
+     * @throws PrestaShopException
+     */
+    private function buildCartRuleFromCommandData(AddCartRuleCommand $command): CartRule
+    {
+        $cartRule = new CartRule();
+
+        $cartRule->name = $command->getLocalizedNames();
+        $cartRule->description = $command->getDescription();
+        $cartRule->code = $command->getCode();
+        $cartRule->highlight = $command->isHighlightInCart();
+        $cartRule->partial_use = $command->isAllowPartialUse();
+        $cartRule->priority = $command->getPriority();
+        $cartRule->active = $command->isActive();
+
+        $this->fillCartRuleConditionsFromCommandData($cartRule, $command);
+        $this->fillCartRuleActionsFromCommandData($cartRule, $command);
+
+        return $cartRule;
+    }
+
+    /**
+     * Fills cart rule with conditions data from command.
+     *
+     * @param CartRule $cartRule
+     * @param AddCartRuleCommand $command
+     */
+    private function fillCartRuleConditionsFromCommandData(CartRule $cartRule, AddCartRuleCommand $command): void
+    {
+        $cartRule->id_customer = null !== $command->getCustomerId() ? $command->getCustomerId()->getValue() : null;
+
+        $cartRule->date_from = $command->getValidFrom()->format('Y-m-d H:i:s');
+        $cartRule->date_to = $command->getValidTo()->format('Y-m-d H:i:s');
+
+        $minimumAmount = $command->getMinimumAmount();
+        $cartRule->minimum_amount = $minimumAmount->getMoneyAmount()->getAmount();
+        $cartRule->minimum_amount_currency = $minimumAmount->getMoneyAmount()->getCurrencyId()->getValue();
+        $cartRule->minimum_amount_shipping = $minimumAmount->isShippingExcluded();
+        $cartRule->minimum_amount_tax = $minimumAmount->isTaxExcluded();
+
+        $cartRule->quantity = $command->getTotalQuantity();
+        $cartRule->quantity_per_user = $command->getQuantityPerUser();
+
+        $cartRule->country_restriction = $command->hasCountryRestriction();
+        $cartRule->carrier_restriction = $command->hasCarrierRestriction();
+        $cartRule->group_restriction = $command->hasGroupRestriction();
+        $cartRule->cart_rule_restriction = $command->hasCartRuleRestriction();
+        $cartRule->product_restriction = $command->hasProductRestriction();
+        $cartRule->shop_restriction = $command->hasShopRestriction();
+    }
+
+    /**
+     * Fills cart rule with actions data from command.
+     *
+     * @param CartRule $cartRule
+     * @param AddCartRuleCommand $command
+     */
+    private function fillCartRuleActionsFromCommandData(CartRule $cartRule, AddCartRuleCommand $command): void
+    {
+        $cartRuleAction = $command->getCartRuleAction();
+        $amountDiscount = $cartRuleAction->getAmountDiscount();
+        $percentageDiscount = $cartRuleAction->getPercentageDiscount();
+        $giftProduct = $cartRuleAction->getGiftProduct();
+        $cartRule->free_shipping = $cartRuleAction->isFreeShipping();
+
+        $cartRule->gift_product = null !== $giftProduct ? $giftProduct->getProductId()->getValue() : null;
+        $cartRule->gift_product_attribute = null !== $giftProduct ? $giftProduct->getProductAttributeId() : null;
+        $cartRule->reduction_amount = null !== $amountDiscount ?
+            $amountDiscount->getMoneyAmount()->getAmount() :
+            null;
+        $cartRule->reduction_currency = null !== $amountDiscount ?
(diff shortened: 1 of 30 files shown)


tests/Integration/Behaviour/Features/Transform/CurrencyTransform.php
```diff
@@ -0,0 +1,50 @@
...
+use InvalidArgumentException;
+
+trait CurrencyTransform
+{
+    /**
+     * Transforms currency iso code to instance
+     *
+     * @Transform /^currency "([^"]+)"$/
+     */
+    public function transformIsoToCurrency(string $currencyIso)
```
rokaszygmantas, line 40: @matks I've introduced behat `Transform` feature using traits, would like your opinion on it. I know you don't like traits, but this seems very reusable in tests. :slightly_smiling_face:

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
```diff
@@ -0,0 +1,635 @@
...
+     * @Given /^(it) gives free shipping$/
+     */
+    public function itGivesFreeShipping(array $properties)
+    {
+        $properties['free_shipping'] = true;
+        SharedStorage::getStorage()->setLatestResource($properties);
+    }
+
+    /**
+     * @Given /^(it) gives a reduction amount of "([^"]+)" in (currency "[^"]+") which is tax (included|excluded) and (applies to (order without shipping|specific product|cheapest product|selected products))$/
```
rokaszygmantas, line 186: can we do something if PrettyCI will be mad about the length of these lines?
I don't think we can split it into multiple lines...

rokaszygmantas: nevermind, PrettyCI is green

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
```diff
@@ -242,12 +242,14 @@ public function iSaveIt(array $properties)
             $properties['discount_product_id'] ?? null
         );
 
-        ObjectModel::resetStaticCache();
+        ObjectModel::disableCache();
```
rokaszygmantas: @matks  I had to disable object model cache because it was conflicting with other scenarios that create cart rules :thinking: another cart rule was loaded out of cache with the same ID.

Do you think this is a good solution? 

I'm also thinking if we should disable the object model cache globally before all test suites, wdyt?

src/Adapter/CartRule/CommandHandler/AddCartRuleHandler.php
```diff
@@ -0,0 +1,239 @@
...
+                    throw new CartRuleConstraintException(
+                        'Cart rule, which is applied to selected products, must have percent discount type.',
+                        CartRuleConstraintException::INCOMPATIBLE_CART_RULE_ACTIONS
+                    );
+                }
+
+                $cartRule->reduction_product = -2;
+
+                break;
+
```
sarjon: Why is there empty lines after `break`?

src/Core/Domain/CartRule/Command/AddCartRuleCommand.php
```diff
@@ -0,0 +1,591 @@
...
+        $this->isActive = $isActive;
+        $this->validFrom = $validFrom;
+        $this->validTo = $validTo;
+        $this->cartRuleAction = $cartRuleAction;
+    }
+
+    /**
+     * @return DiscountApplicationType|null
+     */
+    public function getDiscountApplicationType(): ?DiscountApplicationType
```
sarjon, line 218: Can this be `null`?

src/Core/Domain/CartRule/Command/AddCartRuleCommand.php
```diff
@@ -0,0 +1,591 @@
...
+     */
+    public function getCode(): string
+    {
+        return $this->code;
+    }
+
+    /**
+     * @return MoneyAmountCondition
+     */
+    public function getMinimumAmount(): MoneyAmountCondition
```
sarjon: Seems that naming is a little wrong, I'd expect `public function getMinimumAmount(): MinimumAmount` instead.

src/Core/Domain/CartRule/Command/AddCartRuleCommand.php
```diff
@@ -0,0 +1,591 @@
...
+        return $this->cartRuleAction;
+    }
+
+    /**
+     * @param string $description
+     */
+    public function setDescription(string $description): void
+    {
+        $this->description = $description;
+    }
```
sarjon, line 427: Some setters return `$this` and some don't, it is intended?

src/Core/Domain/CartRule/ValueObject/CartRuleAction/AmountDiscountAction.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        return $this->isFreeShipping;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPercentageDiscount(): ?PercentageDiscount
+    {
+        return null;
```
sarjon, line 84: You should return something here.

src/Core/Domain/CartRule/ValueObject/CartRuleAction/CartRuleActionBuilderInterface.php
```diff
@@ -0,0 +1,80 @@
...
+namespace PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject\CartRuleAction;
+
+use PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject\GiftProduct;
+use PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject\MoneyAmountCondition;
+use PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject\PercentageDiscount;
+
+/**
+ * Describes a builder which builds cart rule actions.
+ */
+interface CartRuleActionBuilderInterface
```
sarjon, line 36: I'm thinking if interface is needed for builder. :thinking:

src/Core/Domain/CartRule/ValueObject/DiscountApplicationType.php
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\CartRule\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\CartRule\Exception\CartRuleConstraintException;
+
+/**
+ * Discount application type
+ */
+class DiscountApplicationType
```
sarjon, line 34: I'm not sure about naming, I could be wrong, but how about `DiscountTarget`? So discount target can be `cheapest_product` or `specific_product`?

It seems that `CartRuleActionInterface` should be `CartRuleTypeInterface` instead, as it describes what type of cart rule is, whether it's percent type, amount type & etc. 

But again, I could be wrong here.

src/Core/Domain/CartRule/ValueObject/MoneyAmountCondition.php
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * @param float $amount
+     * @param int $currencyId
+     * @param bool $taxExcluded
+     * @param bool|null $shippingExcluded
+     *
+     * @throws DomainConstraintException
+     */
+    public function __construct(float $amount, int $currencyId, bool $taxExcluded, bool $shippingExcluded = null)
```
sarjon: You could require `MoneyAmount` instead of `float $amount, int $currencyId`

Would it make sense to keep `$shippingExcluded` as `bool`? With default `true` or `false`?

src/Core/Domain/ValueObject/MoneyAmount.php
```diff
@@ -0,0 +1,79 @@
...
+
+namespace PrestaShop\PrestaShop\Core\Domain\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Currency\ValueObject\CurrencyId;
+use PrestaShop\PrestaShop\Core\Domain\Exception\DomainConstraintException;
+
+/**
+ * An amount of money with currency
+ */
+class MoneyAmount
```
sarjon: I was wondering when this VO will appear in PS codebase. :grin: 

Maybe name `Money` would be enough?

src/Core/Domain/ValueObject/MoneyAmount.php
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * @var CurrencyId
+     */
+    private $currencyId;
+
+    /**
+     * @param float $amount
+     * @param int $currencyId
+     */
+    public function __construct(float $amount, int $currencyId)
```
sarjon: We could use `Decimal` here as it is related to money can could be used in future to calculate prices. 

Maybe it could also accept `CurrencyId $currencyId` instead?

src/PrestaShopBundle/Resources/config/services/core/domain/cart_rule.yml
```diff
@@ -0,0 +1,6 @@
+services:
+  _defaults:
+    public: true
+
+  prestashop.core.domain.cart_rule.value_object.cart_rule_action.cart_rule_action_builder:
```
sarjon: Is it ever used?

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
```diff
@@ -0,0 +1,640 @@
...
+            'selected products' => DiscountApplicationType::SELECTED_PRODUCTS,
+        ];
+
+        return $availableTypesMap[$type] ?? $type;
+    }
+
+    /**
+     * @When /^I want to create a cart rule named "([^"]+)"$/
+     */
+    public function iCreateCartRuleWithName(string $name)
```
sarjon: The method is a little misleading imo, it should be `specifyCartRuleName` or something similar as you don't create it here.

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
```diff
@@ -0,0 +1,640 @@
...
+        SharedStorage::getStorage()->set($key, [
+            'name' => $name,
+        ]);
+    }
+
+    /**
+     * @Given /^I specify (its) "([^"]+)" as "([^"]+)"$/
+     * @Given /^I specify that (its) "([^"]+)" is "([^"]+)"$/
+     */
+    public function iSpecifyItsAs(array $properties, string $property, string $value)
```
sarjon: I think `i` prefix should be removed as you should be able to call it from code as well. Maybe `specifyCartRuleProperty()` instead.

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
```diff
@@ -0,0 +1,640 @@
...
+    {
+        $properties[$property] = $value;
+        SharedStorage::getStorage()->setLatestResource($properties);
```
sarjon: Does it make sense to set latest resource? Shouldn't latest resource refer to something that was last set with `set()` method? Otherwise, this may not be latest.

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
sarjon: The downside I see is it would require to hit database every time resulting in longer builds. But that may not be a problem.

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
```diff
@@ -0,0 +1,640 @@
...
+     * Asserts that discount application type is valid for given cart rule
+     *
+     * @param CartRule $cartRule
+     * @param string $discountApplicationType
+     */
+    private function assertDiscountApplicationTypeIsValid(CartRule $cartRule, string $discountApplicationType)
+    {
+        switch ($discountApplicationType) {
+            case DiscountApplicationType::ORDER_WITHOUT_SHIPPING:
+                $hasError = 0 !== (int) $cartRule->reduction_product;
```
sarjon: I see these hardcoded values for the seconds time, would it make sense to add  constants?

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
```diff
@@ -0,0 +1,640 @@
...
+            $minimumAmountCurrencyId,
+            !$minimumAmountTaxIncluded,
+            !$minimumAmountShippingIncluded
+        );
+
+        null !== $code ? $command->setCode($code) : null;
+        null !== $description ? $command->setDescription($description) : null;
+        null !== $customerId ? $command->setCustomerId($customerId) : null;
+        null !== $discountApplicationType ? $command->setDiscountApplicationType($discountApplicationType) : null;
+        null !== $discountProductId ? $command->setDiscountProductId($discountProductId) : null;
```
sarjon: This one is funny. :smile: Even though it seems shorter and cooler, it takes more brain power to understand compared to simple if. :smile:

tests/Integration/Behaviour/Features/Scenario/CartRule/add_cart_rule.feature
```diff
@@ -0,0 +1,81 @@
...
+    And I specify that partial use is disabled for it
+    And I specify its status as enabled
+    And I specify that it should not be highlighted in cart
+    And I specify its "code" as "PROMO_2019"
+    And its minimum purchase amount in currency "CHF" is "10"
+    And its minimum purchase amount is tax excluded
+    And its minimum purchase amount is shipping included
+    And it gives free shipping
+    And it gives a reduction amount of "15" in currency "USD" which is tax included and applies to order without shipping
+    When I successfully save it
```
sarjon: Can you "unsuccessfully" save it? :grin: I think "successfully" is not needed here.

src/Core/Domain/CartRule/Command/AddCartRuleCommand.php
rokaszygmantas: yes, it's applicable only when cart rule has percentage or amount discount. In case of gift product or free shipping cart rules - this is `null`.

src/Core/Domain/CartRule/Command/AddCartRuleCommand.php
rokaszygmantas: probably not, will fix

src/Core/Domain/CartRule/ValueObject/CartRuleAction/AmountDiscountAction.php
rokaszygmantas: amount discount action doesn't have percentage discount, that's why it's returning null

src/Core/Domain/CartRule/ValueObject/DiscountApplicationType.php
rokaszygmantas: I'm not sure as well, I followed the naming in the interface, where it's called cart rule "Actions". @matks wdyt?

src/Core/Domain/ValueObject/MoneyAmount.php
rokaszygmantas: yes!

src/Core/Domain/CartRule/ValueObject/CartRuleAction/CartRuleActionBuilderInterface.php
rokaszygmantas: why not have it though :thinking: allows to extend the code easily

src/Core/Domain/CartRule/ValueObject/MoneyAmountCondition.php
rokaszygmantas: >Would it make sense to keep $shippingExcluded as bool? With default true or false?

In some cases `$shippingExcluded` is not applicable, that's why it can be `null`.

src/PrestaShopBundle/Resources/config/services/core/domain/cart_rule.yml
rokaszygmantas: it's not, because the cart rule creation workflow is not fully migrated yet :thinking: I'll remove it - we can add it when needed.

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
rokaszygmantas: you don't create it here, but you _want_ to start creating a cart rule, so it specifies the beginning of creation process :smile:

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
matks: > I'm also thinking if we should disable the object model cache globally before all test suites, wdyt?

This is the solution ✅ 

@sarjon If we can use more hardware to deal with it, it's not an issue 😄 but 👍 it's good you thought about it

For any integration tests, we can reset as many mechanisms as needed : database, cache, config, filesystem

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
rokaszygmantas: I had this concern as well. The problem is that we are using arrays to pass the properties through scenario methods. We should be using objects, so we wouldn't need to `set()` anything as we'd be referring an object that's already in the storage. In this case calling `set()` needs to specify the resource name, which is a bit redundant. But I guess there's no better way at the moment :man_shrugging:

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
rokaszygmantas: maybe, somewhere in the `Adapter` namespace as it's coming from legacy

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
rokaszygmantas: -2 lines per condition :older_man:

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
rokaszygmantas: fixed anyway

tests/Integration/Behaviour/Features/Context/Domain/CartRuleFeatureContext.php
rokaszygmantas: should I refactor it? :smile:

sarjon: Not sure if it's worth. :man_shrugging:  But I would avoid using it, image having these conditions all over code base.

rokaszygmantas: fixed

src/Core/Domain/CartRule/Command/AddCartRuleCommand.php
```diff
@@ -0,0 +1,645 @@
...
+        DateTime $validFrom,
+        DateTime $validTo,
+        int $totalQuantity,
+        int $quantityPerUser,
+        CartRuleActionInterface $cartRuleAction,
+        float $minimumAmount,
+        int $minimumAmountCurrencyId,
+        bool $isMinimumAmountTaxExcluded,
+        bool $isMinimumAmountShippingExcluded
+    ) {
```
matks, line 196: This is one big Command 😄

src/Core/Domain/CartRule/ValueObject/DiscountApplicationType.php
matks: Mmmmm ... this is a hard question.
I think @sarjon suggestion is right from a semantic point of view, but @rokaszygmantas did the right choice by sticking to legacy name as it will ease people to understand what it means.

I think we should stick to Actions for this reason 👍Action is "good enough" and helps people acquiring this new vocabulary we are building in Domain folder

tests/Integration/Behaviour/Features/Transform/CurrencyTransform.php
matks: I ❤️ it

I actually like traits, it's Pablo and Pierre who dislike them 😄
I think that, used in the right scenario, they do wonder !

matks:
@rokaszygmantas Sorry for the very late review, but the good news is thanks to @sarjon review I find no issue in this PR !

If you rebase it, we can merge it 😉

matks (dismissed):

rokaszygmantas:
@matks rebased, just need to wait for build :slightly_smiling_face:

matks (dismissed):

rokaszygmantas:
@matks had to rebase once again - feel free to reapprove and merge :slightly_smiling_face:

matks (approved):

matks:
Thank you @rokaszygmantas

PR #14711, opened 2019-07-17
Update Decimal library to 1.2.0

### LINKED ISSUE #14694: An InvalidArgumentException could be thrown in some circumstances
**Describe the bug**

PHP automatically expresses very large numbers, or numbers with more than 4 decimal places in scientific notation, as can be seen here: https://3v4l.org/AbYbl

This means that after certain operations, code could attempt to instantiate a `Decimal\Number` using scientific notation numbers (unsupported by the lib right now), resulting in an exception as described in issue #14644:

```
Uncaught InvalidArgumentException: 
	"1.0E+15" cannot be interpreted as a number
	in /vendor/prestashop/decimal/src/Number.php:466

Stack trace:
#0 /vendor/prestashop/decimal/src/Number.php(72):
	PrestaShop\Decimal\Number->initFromString('1.0E+15')
#1 /src/Core/Localization/Number/Formatter.php(149):
	PrestaShop\Decimal\Number->__construct('1.0E+15')
#2 /src/Core/Localization/Number/Formatter.php(106):
	PrestaShop\PrestaShop\Core\Localization\Number\Formatter->prepareNumber(1.0E+15)
#3 /src/Core/Localization/Locale.php(155): 
	PrestaShop\PrestaShop\Core\Localization\Number\Formatter->format(1.0E+15, Object(PrestaShop\PrestaShop\Core\Localization\Specification\Price))
#4 /classes/Tools.php(776):
	PrestaShop\PrestaShop\Core\Localization\Lo...'
```

This must be fixed in https://github.com/PrestaShop/decimal and updated in PS.

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Update Decimal library to 1.2.0 to avoid errors when using it with numbers expressed in scientific notation
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14694
| How to test?  |

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14711)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3325,16 +3325,16 @@
         },
         {
             "name": "prestashop/decimal",
-            "version": "1.1.0",
+            "version": "1.2.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/PrestaShop/decimal.git",
-                "reference": "42e9bc8c2ec533dc2a247f9321b8f3ed55c5341f"
+                "reference": "11e0f87b101644614e61cfbea175548bdfb389f1"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/PrestaShop/decimal/zipball/42e9bc8c2ec533dc2a247f9321b8f3ed55c5341f",
-                "reference": "42e9bc8c2ec533dc2a247f9321b8f3ed55c5341f",
+                "url": "https://api.github.com/repos/PrestaShop/decimal/zipball/11e0f87b101644614e61cfbea175548bdfb389f1",
+                "reference": "11e0f87b101644614e61cfbea175548bdfb389f1",
                 "shasum": ""
             },
             "require": {
@@ -3369,7 +3369,7 @@
                 "precision",
                 "prestashop"
             ],
-            "time": "2017-08-22T14:23:19+00:00"
+            "time": "2019-07-17T13:14:07+00:00"
         },
         {
             "name": "prestashop/graphnvd3",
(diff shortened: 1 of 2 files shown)


matks (dismissed):

PierreRambaud:
Can you update the composer.json too?

PierreRambaud (changes requested):
Update composer.json too

jolelievre (approved):

jolelievre:
> Update composer.json too

Why update the composer.json? `"prestashop/decimal": "^1.0.0"` allows for the `1.2.0` update

PierreRambaud:
If you don't update composer.json, this mean your app is working weel without the 1.2.0 and have the same behavior with 1.1.0.

jolelievre:
> If you don't update composer.json, this mean your app is working weel without the 1.2.0 and have the same behavior with 1.1.0.

Oh so you mean we should update the minimum required version then? `"prestashop/decimal": "^1.2.0"` ? That makes sens

jolelievre (dismissed):
Update minimum version `"prestashop/decimal": "^1.2.0"`

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "778c2ba69a45162c2024867ab93f50bb",
+    "content-hash": "270c65f46ea48c6629d755ad2a1e8384",
```
eternoendless, line 7: No idea why this hash was updated while the rest of the file didn't change.

PierreRambaud: because of this: 
https://github.com/composer/composer/blob/1.5.2/src/Composer/Package/Locker.php#L65-L102
It depends of the composer.json content.

PierreRambaud (approved):

matks (approved):

Issue #11038, opened 2018-10-16
Configurable debug mode


sarjon:
**Is your feature request related to a problem? Please describe.**
Its a common need to debug issue in production shop. Since a lot of merchant dont have multiple environments (and if they do, sometimes you still have to debug on live shop), developers providing support (and not only) have to be really careful when debugging problems on live shop.

It would be a nice feature to be able to enable Debug mode for different IP addresses.

**Describe the solution you'd like**
Be able to configure IP addresses on which debug mode would be enabled from Back Office.

**Describe alternatives you've considered**
Hardcode IP check directly in defines.inc.php file to define `_PS_MODE_DEV_`.

**Additional context**
Its something similar as configuring Maintenance IP in Back Office.

matks:
Makes sense 👍

ghost:
Brilliant @sarjon ! So simple!

sarjon:
@matks i would be willing to give it a shot, wdyt? however, i'd like to see some mockup how it should look like. :)

sarjon:
@colinegin wdyt?

colinegin:
Hello @sarjon , sorry I missed your comment !
Yes, it could be a great idea for developers ! 
I guess we could have the same feature as for maintenance mode, but let's wait for @TristanLDD 's mockups :)

ghost:
Any news about this FR?

sarjon:
hi @PrestaShark,

as you can see, it's waiting for UI mockups. :)

sarjon:
@TristanLDD any news on this?

Gipielle:
Upvote !

Gipielle:
It is possible to use the same operation that uses the site under maintenance but on the contrary, you could integrate it very quickly. It would be a great thing!

Thank you

Gipielle:
I have found this:
https://www.matthieu-jalbert.fr/activer-le-mode-debug-par-adresse-ip-sur-prestashop/

but make a flag in prestashop backoffice will be a lot better

jf-viguier:
I've posted a pr to support ip V6 et ip subnet masks in maintenance ip. This feathure should also support it : https://github.com/PrestaShop/PrestaShop/pull/16876

madsoliver:
I was just about to add a feature request for this myself but found this little gem ;-)
It would be great to have this.

[...]

PR #14716, opened 2019-07-17
[POC] Debug mode based on IP address

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | This PR provides POC for issue.
| Type?         | new feature
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/11038
| How to test?  | This is POC for feature implementation, it's not ready for QA testing.

### How it looks:
debug_mode

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14716)


config/defines.inc.php
```diff
@@ -24,9 +24,15 @@
  * International Registered Trademark & Property of PrestaShop SA
  */
 
+$debugIps = require __DIR__.DIRECTORY_SEPARATOR.'debug_ip.php';
+
 /* Debug only */
 if (!defined('_PS_MODE_DEV_')) {
-    define('_PS_MODE_DEV_', true);
+    if (in_array($_SERVER['REMOTE_ADDR'], $debugIps)) {
```
sarjon, line 31: Not the best way to get client IP address.

config/defines.inc.php
```diff
@@ -24,9 +24,15 @@
...
+        define($debugConfigName = '_PS_MODE_DEV_', true);
```
sarjon, line 32: When switching debug mode on/off, it looks for regex pattern in file to rewrite it's value, using variable does the trick to avoid switching value on this line.

src/Adapter/Debug/DebugModeConfiguration.php
```diff
@@ -153,4 +156,26 @@ private function updateDebugMode($enableStatus)
...
+    {
+        if (!empty($debugIps)) {
+            $ips = explode(',', $debugIps);
+
+            $code = '<?php return ' . var_export($ips, true) . ';';
+        } else {
+            $code = '<?php return [];';
+        }
+
+        file_put_contents(__DIR__.'/../../../config/debug_ip.php', $code);
```
sarjon, line 171: This could be improved I guess.

config/debug_ip.php
```diff
@@ -0,0 +1,3 @@
+<?php return array (
+  0 => '127.0.0.1',
+);
```
sarjon, line 3: Autogenerated file, by default it could be empty and should be .gitignored.

sarjon:
@matks @PierreRambaud wdyt? Keep in mind it's a POC. :smiling_imp:

src/Adapter/Debug/DebugModeConfiguration.php
PierreRambaud: Better to generate a yaml or json file instead of php

config/debug_ip.php
```diff
@@ -0,0 +1,3 @@
+<?php return array (
+  0 => '127.0.0.1',
```
PierreRambaud, line 2: Add the capability to allow ranges

config/defines.inc.php
```diff
@@ -24,9 +24,15 @@
  * International Registered Trademark & Property of PrestaShop SA
  */
 
+$debugIps = require __DIR__.DIRECTORY_SEPARATOR.'debug_ip.php';
```
matks, line 27: If files does not exist, this might throw a warning, isn't it ?

sarjon: Yeah, there is a lot of stuff to be improved, I'm just wondering if this approach for enabling debug based on IP is good and we I can continue it.

matks:
@sarjon 
If I get it right, you use a file `config/debug_ip.php` to contain the "debug IPs" right ? As Pierre said, why not use JSON or XML to store it ?

So when we use the feature for the 1st time, it creates the `config/debug_ip.php` file but when do you delete it in order to "cancel" ?
I dont like the idea of files being written or deleted to store shop configuration 😅 it might introduce lot of dangerous behaviors (files deleted by mistake, server config refuses to allow files writing or files deleting ...)

So what about this:
- either *always* have a YML/XML file on the server (so not create, no delete, but modify its content
- or store the configuration settings in something else (database ? but it needs to be easily retrievable ; cookie ? needs to be very strongly protected and encrypted ; environment variable ?)

sarjon:
> As Pierre said, why not use JSON or XML to store it

:+1: I used php file for POC as it was easier to implement.

 > either always have a YML/XML file on the server (so not create, no delete, but modify its content

I'd go for this. I would like to store it in db, but this file is loaded very early in request, so PrestaShop autoloader is not loaded yet.

PierreRambaud:
Hey, just thinking but, it should really be better to use a system like:
image
And you can generate a file like we did for .htaccess
Will be easier for merchant to customize

sarjon:
> Hey, just thinking but, it should really be better to use a system like:

What did you mean @PierreRambaud ? :thinking: Maybe you missed screenshot in description? :grin:

PierreRambaud:
image

Haha, I have the screen here :thinking:

PierreRambaud:
Oh you're right :roll_eyes:

sarjon:
@matks what do we do with this PR?

jf-viguier:
Very usefull PR, thanks.
Don't forget to support ip v6 filtering with prefix because ip v6 address has an every day changing end.
Htaccess filtering can be done with this syntax : 2a11:e0d:237:eec0:0000:0000:0000:0000/64
Maybe we could support this syntax or something like this : 2a01:e0a:257:eec0:* or just compare the begining with 2a01:e0a:257:eec0

config/defines.inc.php
jf-viguier: Maintenance mode uses Tools::getRemoteAddr()

sarjon:
> Don't forget to support ip v6

Good idea. :+1: Unfortunately, I don't know if this PR will see day light any time soon. :(

Gipielle:
Hello, any news ?

config/debug_ip.php
jf-viguier: L've done ip V6 and subnet masks support with Symfony here :
https://github.com/PrestaShop/PrestaShop/pull/16876

jf-viguier:
This PR could be very usefull for module devs but it seems abandoned. @matks @PierreRambaud I can improve it but tell me before what you've decided.

Gipielle:
> This PR could be very usefull for module devs but it seems abandoned. @matks @PierreRambaud I can improve it but tell me before what you've decided.

Yes please ! Not only for module devs but specially for small changes in live shops

jf-viguier:
@matks @PierreRambaud I'm still waiting for your point of view on this pr ? 
@sarjon Have you planned to go further here ?

Gipielle:
> @matks @PierreRambaud I'm still waiting for your point of view on this pr ?
> @sarjon Have you planned to go further here ?

Hopefully, this feature is a must

PierreRambaud:
@jf-viguier @Gipielle We will work on it as soon as possible. We also love this feature!

Gipielle:
Thank you !

config/defines.inc.php
PierreRambaud: It can't be used because the autoloader is not loaded yet :thinking:

matks:
Hi everybody I'm afraid we finally could not find time to finish this. @sarjon I dont know if you're willing to continue it 😅  else maybe @Gipielle ?

I close this pull request because we clean old stale PRs. If you are willing to continue this topic, feel free to reopen or open another one if you think it's still relevant. 😉

Thanks!

PR #14721, opened 2019-07-17
Manage edit/view link for customers in HelperList

### LINKED ISSUE #14672: Backoffice quick search - email / customer view redirect to customer page
I noticed that when using the BO search for a customers name or on email adres. After finding the result and clicking on the found customer. When clicking on view, it doesn't open the page with the customer, but redirects to the customer page with all customers where you have to look once again for the customer your searching for. This was not the case in previous versions

PrestaShop version:1.7.6
PHP version: 7.2

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Manage edit/view link for customers in HelperList
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14672
| How to test?  | See issue, test customer links Perform other researches (on products, orders, ...) to be sure nothing is broken.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14721)

### DIFF
--- a/classes/helper/HelperList.php
+++ b/classes/helper/HelperList.php
@@ -269,6 +269,12 @@ class HelperListCore extends Helper
                 }
             }
 
+            if (in_array('view', $this->actions)) {
+                $this->_list[$index]['onclick'] = $this->getViewLink($this->token, $id);
+            } else {
+                $this->_list[$index]['onclick'] = $this->getEditLink($this->token, $id);
+            }
+
             // @todo skip action for bulk actions
             // $this->_list[$index]['has_bulk_actions'] = true;
             foreach ($this->fields_list as $key => $params) {
@@ -473,8 +479,9 @@ class HelperListCore extends Helper
             self::$cache_lang['View'] = Context::getContext()->getTranslator()->trans('View', array(), 'Admin.Actions');
         }
 
+        $href = $this->getViewLink($token, $id);
         $tpl->assign(array(
-            'href' => $this->currentIndex . '&' . $this->identifier . '=' . $id . '&view' . $this->table . '&token=' . ($token != null ? $token : $this->token),
+            'href' => $href,
             'action' => self::$cache_lang['View'],
         ));
 
@@ -491,22 +498,7 @@ class HelperListCore extends Helper
             self::$cache_lang['Edit'] = Context::getContext()->getTranslator()->trans('Edit', array(), 'Admin.Actions');
         }
 
-        $href = $this->currentIndex . '&' . $this->identifier . '=' . $id . '&update' . $this->table . ($this->page && $this->page > 1 ? '&page=' . (int) $this->page : '') . '&token=' . ($token != null ? $token : $this->token);
-
-        switch ($this->currentIndex) {
-            case 'index.php?controller=AdminProducts':
-            case 'index.php?tab=AdminProducts':
-                // New architecture modification: temporary behavior to switch between old and new controllers.
-                $pagePreference = SymfonyContainer::getInstance()->get('prestashop.core.admin.page_preference_interface');
-                $redirectLegacy = $pagePreference->getTemporaryShouldUseLegacyPage('product');
-                if (!$redirectLegacy && $this->identifier == 'id_product') {
-                    $href = Context::getContext()->link->getAdminLink('AdminProducts', true, ['id_product' => $id, 'updateproduct' => 1]);
-                }
-
-                break;
-            default:
-        }
-
+        $href = $this->getEditLink($token, $id);
         $tpl->assign(array(
             'href' => $href,
             'action' => self::$cache_lang['Edit'],
@@ -825,5 +817,65 @@ class HelperListCore extends Helper
         )));
 
         return $this->footer_tpl->fetch();
+    }
+
+    /**
+     * @param string $token
+     * @param int $id
+     *
+     * @return string
+     *
+     * @throws PrestaShopException
+     */
+    protected function getViewLink($token, $id)
+    {
+        switch ($this->table) {
+            case 'customer':
+                $updateAction = 'view' . $this->table;
+                $href = Context::getContext()->link->getAdminLink('AdminCustomers', true, ['id_customer' => $id, $updateAction => 1]);
+                break;
+            default:
+                $href = $this->currentIndex . '&' . $this->identifier . '=' . $id . '&view' . $this->table . '&token=' . ($token != null ? $token : $this->token);
+                break;
+        }
+
+        return $href;
+    }
+
+    /**
+     * @param string $token
+     * @param int $id
+     *
+     * @return string
+     *
+     * @throws PrestaShopException
+     */
+    protected function getEditLink($token, $id)
+    {
+        switch ($this->table) {
+            case 'customer':
+                $updateAction = 'update' . $this->table;
+                $href = Context::getContext()->link->getAdminLink('AdminCustomers', true, ['id_customer' => $id, $updateAction => 1]);
+                break;
+            default:
+                $href = $this->currentIndex . '&' . $this->identifier . '=' . $id . '&update' . $this->table . ($this->page && $this->page > 1 ? '&page=' . (int) $this->page : '') . '&token=' . ($token != null ? $token : $this->token);
+
+                switch ($this->currentIndex) {
+                    case 'index.php?controller=AdminProducts':
+                    case 'index.php?tab=AdminProducts':
+                        // New architecture modification: temporary behavior to switch between old and new controllers.
+                        $pagePreference = SymfonyContainer::getInstance()->get('prestashop.core.admin.page_preference_interface');
+                        $redirectLegacy = $pagePreference->getTemporaryShouldUseLegacyPage('product');
+                        if (!$redirectLegacy && $this->identifier == 'id_product') {
+                            $href = Context::getContext()->link->getAdminLink('AdminProducts', true, ['id_product' => $id, 'updateproduct' => 1]);
+                        }
+
+                        break;
+                    default:
+                }
+                break;
+        }
+
+        return $href;
     }
 }
--- a/classes/helper/HelperList.php
+++ b/classes/helper/HelperList.php
@@ -269,11 +269,7 @@ class HelperListCore extends Helper
                 }
             }
 
-            if (in_array('view', $this->actions)) {
-                $this->_list[$index]['onclick'] = $this->getViewLink($this->token, $id);
-            } else {
-                $this->_list[$index]['onclick'] = $this->getEditLink($this->token, $id);
-            }
+            $this->_list[$index]['onclick'] = in_array('view', $this->actions) ? $this->getViewLink($this->token, $id) : $this->getEditLink($this->token, $id);
 
             // @todo skip action for bulk actions
             // $this->_list[$index]['has_bulk_actions'] = true;
--- a/classes/helper/HelperList.php
+++ b/classes/helper/HelperList.php
@@ -285,7 +285,7 @@ class HelperListCore extends Helper
                 }
             }
 
-            $this->_list[$index]['onclick'] = in_array('view', $this->actions) ? $this->getViewLink($this->token, $id) : $this->getEditLink($this->token, $id);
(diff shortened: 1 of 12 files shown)


classes/helper/HelperList.php
```diff
@@ -826,4 +818,64 @@ public function displayListFooter()
...
+     * @param int $id
+     *
+     * @return string
+     *
+     * @throws PrestaShopException
+     */
+    protected function getViewLink($token, $id)
+    {
+        switch ($this->table) {
+            case 'customer':
```
jolelievre: Ok clearly it's a quick fix, the thing is do I extend this behavior to other migrated entities as well?

matks: 😅
Yeah, but this violates the O of SOLID.

jolelievre: do you really want me to change this? ^^
I could try and perform a generic change but I'm afraid of the side effects..

matks: Since the issue is a trivial one, I'm not sure the cost is worth it :s

classes/helper/HelperList.php
```diff
@@ -269,6 +269,12 @@ public function displayListContent()
                 }
             }
 
+            if (in_array('view', $this->actions)) {
+                $this->_list[$index]['onclick'] = $this->getViewLink($this->token, $id);
+            } else {
+                $this->_list[$index]['onclick'] = $this->getEditLink($this->token, $id);
```
PierreRambaud:
```php
 $this->_list[$index]['onclick'] = in_array('view', $this->actions) ? $this->getViewLink($this->token, $id) : $this->getEditLink($this->token, $id);
```

easier to read

classes/helper/HelperList.php
jolelievre: sadly I agree.. it's just that we will have to add other switch cases as we continue migrating pages
by the way do you think there are other pages that would need to be managed? based on the contents managed in the quick search:
product, customer, order, invoices, cart, modules
There might be a bug on the module links, but I don't know what was the expected result?

classes/helper/HelperList.php
jolelievre: Done

classes/helper/HelperList.php
matks: The more we dig, the worse it gets 😄

matthieu-rolland (dismissed):
LGTM

marionf:
Is it ready for QA @jolelievre ?

PierreRambaud:
@matks Wasn't in favor of making this change :/

jolelievre:
@PierreRambaud I know when you read his comment, but I actually wasn't sure if it was a troll or not ^^
The thing is this helper is used in various places (including modules), juste in the search page it is used for 7 different entities/tables, but the way it generates the links is not good it concatenates various strings to build the url but it should use `getAdminLink`

Now to minimize the side effects of my modification, I only managed the proper generation for customer table, and I left the other tables unchanged. So I have no doubts that we will have other issues with other tables as they get migrated (we may even have some already) and we will have to update this class later.

That's why I asked him if it was worth changing it for all other cases which he responded with `Since the issue is a trivial one, I'm not sure the cost is worth it :s` So, I might misunderstand him but I think we're good to go.

Maybe what we can do @marionf is at least check if all the results form the search page have correct links (products, customers, orders, invoices, carts, IPs and modules)?

marionf:
> Maybe what we can do @marionf is at least check if all the results form the search page have correct links (products, customers, orders, invoices, carts, IPs and modules) ?

Yes, we can do that, so let's go for QA @khouloudbelguith

khouloudbelguith:
Hi @jolelievre,

It is OK
https://drive.google.com/file/d/1frA2HnzMKwe9xh_bSsWHqs0NaXzytyws/view

Thanks!

PierreRambaud:
> @PierreRambaud I know when you read his comment, but I actually wasn't sure if it was a troll or not ^^

He wasn't ^^'

matks:
> That's why I asked him if it was worth changing it for all other cases which he responded with `Since the issue is a trivial one, I'm not sure the cost is worth it :s` So, I might misunderstand him but I think we're good to go.

Hi there ! Sorry, I can see that indeed my comment was ambiguous 😓.

What I meant was this "[the issue that this PR fixes](https://github.com/PrestaShop/PrestaShop/issues/14672) has been labeled 'minor' and this PR uses a very dirty fix that is going to decrease the quality/reliability/maintainability of the codebase if we merge it*, so is this cost (the decrease in quality/reliability/maintainability) not too expensive to fix a minor issue ?"

I would rather be in favor of leaving it un-fixed in 1.7.6.1 as I think it is not used a lot, and perform an heavier refactoring work on the HelperList class in 1.7.7.0 in order to allow it to use `getAdminLink()` for all BO links, which will make it fully compliant with BO URL migration.

Doing this refactoring now, for 1.7.6.1 looks overkill to me because it might introduce massive regressions in a patch version. But for 1.7.7.0 it makes sense, and we could use unit and integration tests to cover ourselves.

And only merging this dirty fix for 1.7.6.1 does not look worth it to me because:
- it's a quick dirty fix, so merging it will bring all the bad consequences we all know
- as @jolelievre [stated](https://github.com/PrestaShop/PrestaShop/pull/14721#discussion_r305249178), similar issues might be already there or will pop because this can happen for any BO link built by HelperList and targeting a migrated BO page

But since I'm supposed to be off right now I will let you decide whether or not to merge this PR in 1.7.6.x branch 😉

*For only a bit, yeah, but bit by bit this is how a codebase rots

jolelievre:
I totally agree with the refacto of this class for 177! And about the fact that the fix is ugly 😅
However if we create the issue for the refacto right now and make sure to do it for 177, I don't see why we could not merge this PR for 1761, I prefer this (knowing that we will do better rather soon) than leaving a know issue on purpose.. The ufgly fiw won't stay for that long in the codebase anyway What do you think?

classes/helper/HelperList.php
```diff
@@ -269,6 +269,8 @@ public function displayListContent()
                 }
             }
 
+            $this->_list[$index]['onclick'] = in_array('view', $this->actions) ? $this->getViewLink($this->token, $id) : $this->getEditLink($this->token, $id);
```
eternoendless: Shouldn't this be `link` instead of `onclick`?

classes/helper/HelperList.php
eternoendless: This is hard to fix for several reasons:

- If we use getAdminLink() then it's a breaking change (in this class' contract, links are meant to be built following a specific schema)
- If we don't, then the component will be more and more broken as we migrate pages
- We cannot inject dependencies because it would be a BC

Basically, it's broken by design, but we cannot let it stay broken.

My suggestion:

1. Extract these two methods into an interface.
2. Create at least two implementations, one for the previous link generation, one for the new one.
3. Use a factory to build the correct implementation according to parameters.
4. It needs to be covered by tests
3. It must support all the migrated CRUD pages.

classes/helper/HelperList.php
jolelievre: This is linked to the modif I made in the template where the "onclick" url was generated.., I wanted to keep in mind the use of this variable

https://github.com/PrestaShop/PrestaShop/pull/14721/files#diff-a3eeb66715b0f2f29851fb427bf72e14R52

But I can rename it with `link` and adapt the template accordingly if you prefer?

jolelievre:
@eternoendless I created the interface we talked about with two implementations, one based on `Link::getAdminLink` and the other one on concatenation as in the "legacy way"
There is room for improvement of course, but I thought for a patch it's enough Let me know if you think otherwise.

@matks your feedback is more than welcome ;)

classes/helper/HelperList.php
eternoendless: I think it would be best since it's semantically wrong. Its value is not something that you'd put in onclick directly, it's an URL.

src/Adapter/Routing/AdminLinkBuilder.php
```diff
@@ -0,0 +1,101 @@
...
+{
+    /** @var Link */
+    private $link;
+
+    /** @var array */
+    private $entitiesControllers;
+
+    /**
+     * @param Link $link
+     * @param array $entitiesControllers
```
eternoendless: Please add more details regarding the schema for this parameter. I should be able to learn how use any method without needing to look up usage examples elsewhere.

src/Adapter/Routing/AdminLinkBuilder.php
```diff
@@ -0,0 +1,101 @@
...
+ * Class AdminLinkBuilder is able to build entity links based on the Link::getAdminLink
+ * method (which indirectly allows it to build symfony url as well).
+ */
+class AdminLinkBuilder implements EntityLinkBuilderInterface
+{
+    /** @var Link */
+    private $link;
+
+    /** @var array */
+    private $entitiesControllers;
```
eternoendless: I think the proper term would be `entityControllers`

src/Core/Routing/EntityLinkBuilderInterface.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+interface EntityLinkBuilderInterface
+{
+    /**
+     * @param string $entity
+     * @param array $parameters
+     *
+     * @return string
+     */
+    public function buildViewLink($entity, array $parameters);
```
eternoendless: How about changing the names to `get<something>Link` instead?

src/Adapter/Routing/AdminLinkBuilder.php
```diff
@@ -0,0 +1,101 @@
...
+            [$entityId => $parameters[$entityId], $editAction => 1]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function canBuild($entity)
+    {
+        return array_key_exists($entity, $this->entitiesControllers);
```
eternoendless: You can just use `isset` which is a language construct
```suggestion
        return isset($this->entitiesControllers[$entity]);
```

src/Core/Routing/EntityLinkBuilderFactory.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * @param string $entity
+     *
+     * @return EntityLinkBuilderInterface
+     *
+     * @throws BuilderNotFoundException
+     */
+    public function getBuilder($entity)
```
eternoendless: I think it would be more idiomatic if this was renamed to `getBuilderFor`

eternoendless (changes requested):
Changes requested

src/Adapter/Routing/AdminLinkBuilder.php
jolelievre: You mean details in the PHPDoc or strong typed objects?

classes/helper/HelperList.php
```diff
@@ -124,11 +128,23 @@ class HelperListCore extends Helper
...
 
+    /** @var EntityLinkBuilderFactory */
+    private $linkBuilderFactory;
+
     public function __construct()
     {
         $this->base_folder = 'helpers/list/';
         $this->base_tpl = 'list.tpl';
 
+        $adminLinkBuilder = new AdminLinkBuilder(Context::getContext()->link, [
```
jolelievre, line 139: @eternoendless I addressed your feedback
I was wondering about this initialization, maybe it would be handy to have this info about migrated controllers somewhere in a configuration? To centralize this information which might be used somewhere else, or in a future service definition

src/Adapter/Routing/AdminLinkBuilder.php
eternoendless: I meant a detailed PHPdoc, thanks

classes/helper/HelperList.php
eternoendless: It could be a class constant.

jolelievre: Yes, but I wonder which would be the best candidate to store this const? Should we use `AdminLinkBuilder` for now? or do you think of a more generic/transversal class?

eternoendless: I think we can leave it as is for now.

eternoendless (approved):

eternoendless:
Thank you @jolelievre

PR #13804, opened 2019-05-15
Fix shared smarty compilation ID across shops / themes

202-ecommerce:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | In all version of PS 1.7, if you have several shops with different theme, they share the same compile ID. Consequently first theme in cache are used for all shops.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9763
| How to test?  | Install two shops with two different theme with cache active and force compile off. Each shop should be load its right theme. You can although try with a child theme and modify for instance only one template (for instance footer.tpl).

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/13804)

PR #14727, opened 2019-07-18
Fix compile id {render} override issue on theme

### LINKED ISSUE #9763: [BOOM-4673] Shopping cart multi-store
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-4673).

- _**Reporter:**_ alexandre@weggs.fr
- _**Created at:**_ Mon, 22 Jan 2018 12:00:45 +0100

<p>Hi !</p>

<p>I have activate multistore with two different themes.</p>

<p>When i add a product in cart in the second shop, the html element .blockcart is well updated but it update also the .blockcart element on the other shop.</p>

<p>If compilation is set on force compilation for each call, it doesn't append. Seems to be a file name cahce problem ?</p>
        
N/A

202-ecommerce:
| Questions     | Answers
| Branch?       | develop
| Description?   |  In case of multishop with multiple active theme theme, the smarty tag: `{render file='module:mymodule/views/templates/front/_partials/form.tpl' ui=$form}` is not compile properly. It's the same bug in an other case than #9763 and in fact partially fixed by #13804
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9763
| How to test?  | You need to have multishop and a different theme activated. you need include a sub template from a module with the {render} smarty tag. The option on backoffice is set to "never recompile templates". You'll see that you will show only one version of the compiled template.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14727)

### DIFF
--- a/classes/form/AbstractForm.php
+++ b/classes/form/AbstractForm.php
@@ -110,12 +110,16 @@ abstract class AbstractFormCore implements FormInterface
             $this->smarty
         );
 
+        $context = Context::getContext();
+        $theme = $context->shop->theme->getName();
+
         $scope->assign($extraVariables);
         $scope->assign($this->getTemplateVariables());
 
         $tpl = $this->smarty->createTemplate(
             $this->getTemplate(),
-            $scope
+            $scope,
+            $theme
         );
 
         return $tpl->fetch();


PierreRambaud (approved):

flegars:
# How to reproduce 

Here's a [.zip](https://github.com/PrestaShop/PrestaShop/files/3480683/themes.zip) which contains the theme folders with needed themes. You should extract it at the root of your project.

## First part

1. Create a multistore with 2 stores. 
2. The first store will use the "Classic" theme and the second a child theme from the "Classic".

## Second part

1. Go on the login form with the two stores.
2. The option <code>Template compilation</code> should be on <code>force compilation</code>, and you'll need to clear all caches.
3. Now, turn the option <code>Template compilation</code> on <code>Never recompile template files</code> and clear all caches, again.
4. Go on two different login form and you'll see the content of the first compiled template on both stores.

clotairer:
I hope this zip will be helpfull to reproduce and confirm the fix of the bug. 
#QAValidation

matks:
Thank you @clotaire202 @Florian-lg @202-ecommerce !

PR #14739, opened 2019-07-18
Keep Composer.lock file in release

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Composer.lock file must be kept in the release to avoid warning from Symfony requirements
| Type?         | bug fix
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | This recommended message about composer must'nt be displayed<br>image

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14739)

### DIFF
--- a/tools/build/Library/ReleaseCreator.php
+++ b/tools/build/Library/ReleaseCreator.php
@@ -76,7 +76,6 @@ class ReleaseCreator
         '.gitignore',
         '.gitmodules',
         '.travis.yml',
-        'composer.lock',
         'package-lock.json',
         '.babelrc',
         'postcss.config.js',


PierreRambaud (approved):

matks:
I'm not sure this is right. Actually I wish we could keep both composer.json and composer.lock in zip file to help prestashop developers use Composer on prestashop.

But adding the file to hide the warning seems like a trick to me. Isn't the issue more "why does SF checks that" ?

jolelievre:
> I'm not sure this is right. Actually I wish we could keep both composer.json and composer.lock in zip file to help prestashop developers use Composer on prestashop.
> 
> But adding the file to hide the warning seems like a trick to me. Isn't the issue more "why does SF checks that" ?

Actually Symfony uses the composer.lock to detect which Symfony version is used, and then know which PHP min version it needs... not that good.
We could remove or adapt this check from symfonyRequirements.php, we're not supposed to do that (it's explained in the file) because it is updated when we update Symfony (which doesn't occur that often)

jolelievre:
Thanks @PierreRambaud @sarahdib

thepixelabs:
@PierreRambaud @jolelievre If users keep this file at the root of their site after install, for example /var/www/html, won't that pose a security risk for attackers to see the contents of the file? Pentest tools fire an alarm on this file which i didn't notice existed till then.
There is no documentation of this file to be removed after install

matks:
> @PierreRambaud @jolelievre If users keep this file at the root of their site after install, for example /var/www/html, won't that pose a security risk for attackers to see the contents of the file? Pentest tools fire an alarm on this file which i didn't notice existed till then.
> There is no documentation of this file to be removed after install

Yes you are right 😄  this has been marked as a security advisory https://github.com/PrestaShop/PrestaShop/security/advisories/GHSA-492w-2pp5-xhvg and has been fixed in PS 1.7.6.6

thepixelabs:
@matks I started my shop installation at 1.7.6.5 and upgraded to 1.7.6.6 and after to 1.7.6.7.

If somebody upgrades this file DOESNT get deleted

matks:
> @matks I started my shop installation at 1.7.6.5 and upgraded to 1.7.6.6 and after to 1.7.6.7.
> 
> If somebody upgrades this file DOESNT get deleted

composer.lock file is useful for some prestashop users so it was not deleted. It is being protected by .htaccess https://github.com/PrestaShop/PrestaShop/commit/35ef7e9d892287c302df1fc5aa05ecfc6f15bc76

thepixelabs:
@matks Will not be protected by default for non Apache users ;) Thanks for clarification, I moved it anyhow.

PR #14767, opened 2019-07-19
Behat tests for Seo & Urls page commands - some code improvements

tomas862:
| Questions     | Answers
| Branch?       | develop
| Description?  | It adds behat tests plus also modifies one unit test due to configuration value is being cached in behat test context. Also, 
| Type?         | improvement
| Category?     | BO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | partially https://github.com/PrestaShop/PrestaShop/issues/14480
| How to test?  | no need.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14767)

### DIFF
--- a/src/Adapter/Meta/MetaDataProvider.php
+++ b/src/Adapter/Meta/MetaDataProvider.php
@@ -57,6 +57,14 @@ class MetaDataProvider implements MetaDataProviderInterface
     }
 
     /**
+     * @return array
+     */
+    public function getAllPages()
+    {
+        return Meta::getPages();
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function getDefaultMetaPageNameById($metaId)
--- a/src/Adapter/Meta/MetaDataProvider.php
+++ b/src/Adapter/Meta/MetaDataProvider.php
@@ -59,9 +59,9 @@ class MetaDataProvider implements MetaDataProviderInterface
     /**
      * @return array
      */
-    public function getAllPages()
+    public function getAvailablePages()
     {
-        return Meta::getPages();
+        return Meta::getPages(true);
     }
 
     /**
--- a/src/Core/ConstraintValidator/IsUrlRewriteValidator.php
+++ b/src/Core/ConstraintValidator/IsUrlRewriteValidator.php
@@ -39,16 +39,18 @@ use Symfony\Component\Validator\ConstraintValidator;
 class IsUrlRewriteValidator extends ConstraintValidator
 {
     /**
-     * @var ConfigurationInterface
+     * @var ConfigurationInterface|bool
      */
-    private $configuration;
+    private $accentedCharsConfiguration;
 
     /**
-     * @param ConfigurationInterface $configuration
+     * @param ConfigurationInterface|bool $accentedCharsConfiguration - this variable can accept boolean value
+     * of already predefined accented chars allowance configuration to not introduce BC break. The recommended approach
+     * is to pass PrestaShop\PrestaShop\Adapter\Configuration as a service instead to avoid keeping cached scalar value.
      */
-    public function __construct(ConfigurationInterface $configuration)
+    public function __construct($accentedCharsConfiguration)
     {
-        $this->configuration = $configuration;
+        $this->accentedCharsConfiguration = $accentedCharsConfiguration;
     }
 
     /**
@@ -78,7 +80,7 @@ class IsUrlRewriteValidator extends ConstraintValidator
     }
 
     /**
-     * Validates url rewrite according the patterns which vary based on ascended chars allowed setting.
+     * Validates url rewrite according the patterns which vary based on ascented chars allowed setting.
      *
      * @param string $urlRewrite
      *
@@ -88,10 +90,24 @@ class IsUrlRewriteValidator extends ConstraintValidator
     {
         $pattern = '/^[_a-zA-Z0-9\-]+$/';
 
-        if ($this->configuration->get('PS_ALLOW_ACCENTED_CHARS_URL')) {
+        if ($this->getAllowAccentedCharsSetting()) {
             $pattern = '/^[_a-zA-Z0-9\pL\pS-]+$/u';
         }
 
         return preg_match($pattern, $urlRewrite);
+    }
+
+    /**
+     * Gets the accented chars url setting.
+     *
+     * @return bool
+     */
+    private function getAllowAccentedCharsSetting()
+    {
+        if ($this->accentedCharsConfiguration instanceof ConfigurationInterface) {
+            return $this->accentedCharsConfiguration->get('PS_ALLOW_ACCENTED_CHARS_URL');
+        }
+
+        return $this->accentedCharsConfiguration;
     }
 }
--- a/src/Core/ConstraintValidator/IsUrlRewriteValidator.php
+++ b/src/Core/ConstraintValidator/IsUrlRewriteValidator.php
@@ -44,9 +44,11 @@ class IsUrlRewriteValidator extends ConstraintValidator
     private $accentedCharsConfiguration;
 
     /**
-     * @param ConfigurationInterface|bool $accentedCharsConfiguration - this variable can accept boolean value
-     * of already predefined accented chars allowance configuration to not introduce BC break. The recommended approach
-     * is to pass PrestaShop\PrestaShop\Adapter\Configuration as a service instead to avoid keeping cached scalar value.
+     * this constructor can accept boolean value of already predefined accented chars allowance configuration to not
+     * introduce BC break. The recommended approach is to pass
+     * PrestaShop\PrestaShop\Adapter\Configuration as a service instead to avoid keeping cached scalar value.
+     *
+     * @param ConfigurationInterface|bool $accentedCharsConfiguration
      */
     public function __construct($accentedCharsConfiguration)
     {
--- a/src/Core/ConstraintValidator/IsUrlRewriteValidator.php
+++ b/src/Core/ConstraintValidator/IsUrlRewriteValidator.php
@@ -26,6 +26,7 @@
 
 namespace PrestaShop\PrestaShop\Core\ConstraintValidator;
 
+use PrestaShop\PrestaShop\Core\ConfigurationInterface;
 use PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints\IsUrlRewrite;
 use Symfony\Component\Validator\Exception\UnexpectedTypeException;
 use Symfony\Component\Validator\Constraint;
@@ -38,16 +39,16 @@ use Symfony\Component\Validator\ConstraintValidator;
 class IsUrlRewriteValidator extends ConstraintValidator
 {
     /**
-     * @var bool
+     * @var ConfigurationInterface
      */
-    private $isAscendedCharsAllowed;
+    private $configuration;
 
     /**
-     * @param bool $isAscendedCharsAllowed
+     * @param ConfigurationInterface $configuration
      */
-    public function __construct($isAscendedCharsAllowed)
+    public function __construct(ConfigurationInterface $configuration)
     {
-        $this->isAscendedCharsAllowed = $isAscendedCharsAllowed;
+        $this->configuration = $configuration;
     }
 
     /**
@@ -87,7 +88,7 @@ class IsUrlRewriteValidator extends ConstraintValidator
     {
         $pattern = '/^[_a-zA-Z0-9\-]+$/';
 
-        if ($this->isAscendedCharsAllowed) {
+        if ($this->configuration->get('PS_ALLOW_ACCENTED_CHARS_URL')) {
             $pattern = '/^[_a-zA-Z0-9\pL\pS-]+$/u';
         }
 
--- a/src/PrestaShopBundle/Resources/config/services/adapter/meta.yml
+++ b/src/PrestaShopBundle/Resources/config/services/adapter/meta.yml
@@ -25,6 +25,7 @@ services:
         class: 'PrestaShop\PrestaShop\Adapter\Meta\CommandHandler\EditMetaHandler'
         arguments:
           - '@validator'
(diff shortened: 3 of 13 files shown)


tests/Integration/Behaviour/Features/Scenario/Meta/meta_management.feature
```diff
@@ -0,0 +1,150 @@
+# ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml -s meta
+@reset-database-before-feature
+Feature: Meta management (Traffic & Seo)
+  PrestaShop allows BO users to manage page metadata - title, description, keywords, url_rewrite etc...
+  As a BO user I must be able to create, edit, delete and update meta data.
```
zuk3975, line 5: edit and update?
And I think this behat is missing delete operations

tomas862: Yeah it does not have delete operation implemented with CQRS.  but on the other hand it would not be hard to write it but its different PR's responsability to do that IMO

tests/Integration/Behaviour/Features/Context/Domain/MetaFeatureContext.php
```diff
@@ -0,0 +1,300 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Meta\ValueObject\MetaId;
+use RuntimeException;
+use Tests\Integration\Behaviour\Features\Context\SharedStorage;
+
+class MetaFeatureContext extends AbstractDomainFeatureContext
+{
+    /**
+     * @Given /^I specify following properties for new meta "([^"]*)":$/
+     */
+    public function specifyFollowingPropertiesForNewMeta($reference, TableNode $node)
```
matks: 😮 
Euh ... I'm a bit surprised 😅 why do you do this ? Why first put the data in the shared storage, then use them to create a metadata item in `addMetaWithSpecifiedProperties()` ?

matks: I mean we could do it in one single step, right ?
If it's in a single step, it is easier to reuse, avoid using the shared storage for that and consequently is a bit more stateless

tests/Integration/Behaviour/Features/Context/Domain/MetaFeatureContext.php
```diff
@@ -0,0 +1,300 @@
...
+     */
+    public function updateMetaWithSpecifiedProperties($reference)
+    {
+        $propertiesKey = sprintf('%s_properties', $reference);
+
+        $data = SharedStorage::getStorage()->get($propertiesKey);
+        $data = $this->getWithDefaultLanguage($data);
+
+        $command = (new EditMetaCommand((int) $data['meta_id']))
+            ->setLocalisedPageTitles(isset($data['localized_page_title']) ? $data['localized_page_title'] : [])
```
matks: If we use the right property names, then we can use `DataTransfer::transferAttributesFromArrayToObject($data, $command);`

See https://github.com/PrestaShop/PrestaShop/blob/develop/tests/Integration/Behaviour/Features/Context/CustomerManagerFeatureContext.php#L146

tests/Integration/Behaviour/Features/Context/Domain/MetaFeatureContext.php
```diff
@@ -0,0 +1,300 @@
...
+        $this->assertLastErrorCodeIs(MetaConstraintException::INVALID_URL_REWRITE);
+    }
+
+    /**
+     * @Then /^I should get error that page name value is incorrect$/
+     */
+    public function assertItShouldGetErrorThatPageNameValueIsIncorrect()
+    {
+        $this->assertLastErrorIs(MetaConstraintException::class);
+        $this->assertLastErrorCodeIs(MetaConstraintException::INVALID_PAGE_NAME);
```
matks, line 253: Good idea to check the error code

tests/Integration/Behaviour/Features/Scenario/Meta/meta_management.feature
```diff
@@ -0,0 +1,150 @@
...
+    Then I should get error that page name value is incorrect
+
+  Scenario: Updating metadata with already existing page name
+    Given I specify following properties for new meta "meta12":
+      | meta_id                            | 4                                |
+      | page_name                          |  	pagenotfound                  |
+    When I update meta "meta11" with specified properties
+    Then I should get error that page name value is incorrect
+
+  Scenario: Get non existing meta for editing it raises an error
```
matks:
```suggestion
  Scenario: Fail to edit a meta item that does not exist
```

tests/Integration/Behaviour/Features/Scenario/Meta/meta_management.feature
```diff
@@ -0,0 +1,150 @@
...
+    And meta "meta8" field "url_rewrite" for default language should be ""
+
+  Scenario: Updating metadata without default language for url rewrite should not be allowed
+    Given I specify following properties for new meta "meta9":
+      | meta_id                            | 1                                        |
+      | localized_rewrite_urls             | rewrite-url                              |
+    When I update meta "meta9" with specified properties without default language
+    Then I should get error that url rewrite value is incorrect
+
+  Scenario: Updating metadata with invalid url rewrite should not be allowed when ascended chars setting is turned off
```
matks, line 81: Maybe we can merge this one with `Scenario: Updating metadata with invalid url rewrite should be allowed when ascended chars setting is turned on` (1st you try without the setting, it fails, then you turn the config on and it works)

So it would become a scenario `Update a metadata item with a bad URL rewrite, fails when ascended chars setting is turned on but succeeds else`

tests/Integration/Behaviour/Features/Scenario/Meta/meta_management.feature
```diff
@@ -0,0 +1,150 @@
...
+    Given I specify following properties for new meta "meta13":
+      | meta_id                          | 4                                     |
+      | page_name                        | index                                 |
+      | localized_page_title             | page title in default language        |
+      | localized_meta_description       | meta description in default language  |
+      | localized_meta_keywords          | meta keywords in default language     |
+      | localized_rewrite_urls           | rewrite-url-default                   |
+    When I update meta "meta13" with specified properties
+    When I get meta "meta13" with specified properties
+    Then meta "meta13" editable form field "page" should be equal to "index"
```
matks, line 122: I'm wondering if we should try to make these steps "the command saved everything successfully' automatic, so we can apply this to every add/edit command.

A step "everything was saved successfully" with a bit of magic in it

Maybe a bad idea

I dont know

Tell me 😅

matks (changes requested):

matks:
@tomas862 PrettyCI is not happy 😄

src/Adapter/Meta/MetaDataProvider.php
```diff
@@ -56,6 +56,14 @@ public function getIdByPage($pageName)
         return $idMeta;
     }
 
+    /**
+     * @return array
+     */
+    public function getAvailablePages()
+    {
+        return Meta::getPages(true);
```
sarjon, line 64: Do you think query would make sense here instead? :thinking:

src/PrestaShopBundle/Resources/config/services/adapter/meta.yml
```diff
@@ -16,6 +16,7 @@ services:
             - '@prestashop.core.hook.dispatcher'
             - '@validator'
             - '@=service("prestashop.adapter.legacy.configuration").get("PS_LANG_DEFAULT")'
+            - '@=service("prestashop.adapter.meta.data_provider").getAvailablePages()'
```
sarjon: Consider injecting service instead, data will be cached if you only inject result which may cause unwanted side effects.

src/PrestaShopBundle/Resources/config/services/adapter/meta.yml
```diff
@@ -24,6 +25,7 @@ services:
         class: 'PrestaShop\PrestaShop\Adapter\Meta\CommandHandler\EditMetaHandler'
         arguments:
           - '@validator'
+          - '@=service("prestashop.adapter.meta.data_provider").getAvailablePages()'
```
sarjon: Same here.

tests/Integration/Behaviour/Features/Context/Domain/MetaFeatureContext.php
```diff
@@ -0,0 +1,300 @@
+<?php
+
```
sarjon, line 26: Missing license.

tests/Integration/Behaviour/Features/Context/LanguageFeatureContext.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
```
sarjon, line 26: Same here. Check all files.

tests/Integration/Behaviour/Features/Context/Domain/MetaFeatureContext.php
tomas862: yep , agree with that - removed this part :+1:

tests/Integration/Behaviour/Features/Context/Domain/MetaFeatureContext.php
tomas862: nice idea!

tests/Integration/Behaviour/Features/Scenario/Meta/meta_management.feature
tomas862: the command and the object has different fields associated - Im afraid not unless we specify some kind of mapping but it would be too much for such task I guess :smile:

src/Adapter/Meta/MetaDataProvider.php
tomas862: ahh dont think so - if you check inside that function you will see a lot of stuff going on there ;D

tests/Integration/Behaviour/Features/Context/Domain/MetaFeatureContext.php
```diff
@@ -0,0 +1,287 @@
...
+    public function getMetaWithSpecifiedProperties($reference, TableNode $node)
+    {
+        $data = $node->getRowsHash();
+        $queryCommand = new GetMetaForEditing((int) $data['metaId']);
+
+        try {
+            /** @var EditableMeta $editableMeta */
+            $editableMeta = $this->getQueryBus()->handle($queryCommand);
+
+            SharedStorage::getStorage()->set("editable_{$reference}", $editableMeta);
```
matks, line 81: Why prefix with `editable` ? Why not using `$reference` directly ?

src/Core/ConstraintValidator/IsUrlRewriteValidator.php
```diff
@@ -38,16 +39,16 @@
...
      */
-    private $isAscendedCharsAllowed;
+    private $configuration;
 
     /**
-     * @param bool $isAscendedCharsAllowed
+     * @param ConfigurationInterface $configuration
      */
-    public function __construct($isAscendedCharsAllowed)
+    public function __construct(ConfigurationInterface $configuration)
```
matks: I'm afraid this is a BC break 😞 
https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/src/Core/ConstraintValidator/IsUrlRewriteValidator.php#L48

What we can do is
- make the 2 params optional but check that either `$isAscendedCharsAllowed` or `$configuration` is provided
- add a `@todo` to do it in next major
- create a new file but I feel like it would overdoing

tests/Integration/Behaviour/Features/Context/Domain/MetaFeatureContext.php
tomas862: in behat scenarios I use same references "meta1" , "meta2" and "meta3" which actually represent data which is stored in objectModel "Meta". Hovewer this reference is retrieved from Query `GetMetaForEditing` and it is stored as object `EditableMeta` so I prefix it differently because it holds kind of different object.

src/Core/ConstraintValidator/IsUrlRewriteValidator.php
tomas862: never heard that someone would use validator class directly.

You call `\Symfony\Component\Validator\Validator\ValidatorInterface` like this:

1 sample:
```php
$this->validator->validate("my url rewrite string", new \PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints\IsUrlRewrite())
```

even surfing in symfony docs I can't find samples where it is used like this:

2 sample:
```php
$validator = new \PrestaShop\PrestaShop\Core\ConstraintValidator\IsUrlRewriteValidator($myConfigurationVariable) 
```

although maybe some people use it in wrong way - Im not even sure if it would work like that. Its complex to set up all dependencies in order to start validating like in second sample ;D 

We should even cover the code even if it is being used incorrectly?

matks: What if I do this:
```
class MyCustomIsUrlRewriteValidator extends IsUrlRewriteValidator
```
so I have created my own URL validator because I have a shop with complex URL rules, and when I upgrade it's broken because my service `myshop.app.validator.my_custom_url_validator` cant be instantiated by the container anymore, as it the parent now requires a new constructor parameter ?

This is espescially important as we are going to push services override as the new mechanism to replace legacy overrides.

tomas862: yes you are totally correct here - indeed some people might have added this behavior already. 

But its very important to keep this test working so instead of adding TODO I'll choose the first option which is to introduce both bool type and ConfigurationInterface via constructor.

tomas862:
@matks please check again the PR - I eliminated the BC break risk and also kept very important test functional :)

matks (approved):

matks:
That's a 👍 for me LGTM ✅

PR #14786, opened 2019-07-19
Sprintf function broken in 1.7.6 for custom module templates

### LINKED ISSUE #14706: sprintf function broken in 1.7.6 for custom module templates.
**Describe the bug**
Using the sprintf function in non-native module templates causes an error depending on how the string is contructed.

Multiple bugs here.
1.75 templates using this function no longer work, 
Using this format

{l s='%s available' mod='roja45productrental' sprintf=['$model.availability]}

resultant string is corrupted in 1.7.6 whereas it is correctly displayed in 1.7.5.

image

Using this format, as templates in 1.7.6 seem to use:
{l s='%qty% available' mod='roja45productrental' sprintf=['%qty%' => $model.availability]}

This string template contruction works:
{l s='%qty% available' mod='roja45productrental' sprintf=['%qty%' => $model.availability]}

This string template does not,

{l s='Collection After: %collect%' mod='roja45productrental' sprintf=['%collect%'=>$rental_product->collection_time]}

The position of the %xxx% is causing the error.  One is matched by the regex used in checkAndReplaceArgs in Translate.php, the other is not, so they follow different string replacement paths in the code.

Update the template to use 
**To Reproduce**
As described

**Screenshots**

**Additional information**
PrestaShop version: N/A
PHP version: N/A

roja45:
| Questions     | Answers
| Branch?       | develop
| Description?  | When using the sprintf function in a front office module template, the translation function is called twice on the input string to be translated.  As far as I can see in the code, the line I have removed is not used as the variable being set is immediately overwritten.  It does however translate the input string, which in then passed into the translation a second time, this corrupts the output.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14706
| How to test?  | There is a test module attached to the original ticket.  If installed in 1.7.6 clean, you will see the error.  Apply this change and the corrupted string is fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14786)

### DIFF
--- a/classes/Translate.php
+++ b/classes/Translate.php
@@ -230,8 +230,6 @@ class TranslateCore
                 ) {
                     $string = Translate::checkAndReplaceArgs($string, $sprintf);
                 }
-
-                $ret = str_replace('"', '&quot;', $string);
             }
 
             $currentKey = strtolower('<{' . $name . '}' . _THEME_NAME_ . '>' . $source) . '_' . $key;
--- a/classes/Translate.php
+++ b/classes/Translate.php
@@ -222,16 +222,6 @@ class TranslateCore
         if (isset($langCache[$cacheKey])) {
             $ret = $langCache[$cacheKey];
         } else {
-            if ($_MODULES == null) {
-                if (
-                    $sprintf !== null &&
-                    (!is_array($sprintf) || !empty($sprintf)) &&
-                    !(count($sprintf) === 1 && isset($sprintf['legacy']))
-                ) {
-                    $string = Translate::checkAndReplaceArgs($string, $sprintf);
-                }
-            }
-
             $currentKey = strtolower('<{' . $name . '}' . _THEME_NAME_ . '>' . $source) . '_' . $key;
             $defaultKey = strtolower('<{' . $name . '}prestashop>' . $source) . '_' . $key;
 


PierreRambaud:
Not sure this is the right way to fix the original bug :thinking: 
@eternoendless @matthieu-rolland  Wdyt?

marionf:
It's a regression of 1.7.6.0, so should be fixed on 1.7.6.x branch

roja45:
@PierreRambaud yup, that's quite likely, the translation code is not somewhere I would usually find myself.  All I can say is that;

this line: **$ret = str_replace('"', '&quot;', $string);** is not doing anything, that I can see.

and it's this line: **$string = Translate::checkAndReplaceArgs($string, $sprintf);** being called twice that is causing the problem.

        ```
    if ($_MODULES == null) {
                if (
                    $sprintf !== null &&
                    (!is_array($sprintf) || !empty($sprintf)) &&
                    !(count($sprintf) === 1 && isset($sprintf['legacy']))
                ) {
                    $string = Translate::checkAndReplaceArgs($string, $sprintf);
                }
                $ret = str_replace('"', '&quot;', $string);
            }
```

Let me know if I need to do something more.

Cheers

PierreRambaud (dismissed):

PierreRambaud (dismissed):

PierreRambaud:
We investigate and we think your solution is the best one :+1: 
(Because of https://github.com/PrestaShop/PrestaShop/blob/b786f476238728e3356b51b65459e4471a6c992a/classes/Translate.php#L249 which is a duplicate :D)

PierreRambaud:
PR rebased for CI.

PierreRambaud (approved):

PierreRambaud:
Thanks a lot @roja45  :smile_cat:

Issue #14731, opened 2019-07-18
Multistore behavior improvement in Back Office


sarjon:
This issue should help prepare for _Multistore_ behavior  improvements in Back Office (BO).

### Main pain points with current _Multistore_ behavior in BO:

1. Since PS 1.7 all new _Configuration forms_ are missing _Multistore_ behavior
2. Persisting configuration behavior has changed since 1.6 (see behavior explanation below).
3. All _Identifiable Object forms_ have manually added & rendered _Shop association_ inputs.
4. When updating configuration in "Single shop" context in PS 1.6, it only submits those values that are overridden, but not all form as it is the case in 1.7

### Layers of application that should be improved/changed:

1. **UI layer**
Rendering of forms and inputs with Multistore behavior, ideally it should be done by form theme automatically according to configuration that was done in layer below. 
2. **Shop Context & Form Building (aka Controller) layer**
This is where the translation between UI & Core happens. This layer is responsible for:
    * Form configuration (be it Configuration or Identifiable Object form) with _Multistore_ behavior
    * Prepare submitted _Configuration form_ data & shop context data and pass it to command for saving/updating
    * Prepare submitted _Identifiable Object form_ data and pass it to Identifiable Object form data handler (this should not change, since handlers already take care of _Shop association_)
3. **Command/Query layer** (not applicable for Identifiable Objects)
This layer deals with retrieving configuration & deciding when to delete/update it based on shop context data.
4. **Persistence layer** (not applicable for Identifiable Objects)
Dealing with `Configuration` & `ConfigurationInterface`, basically low level interaction with configuration. I guess interface needs to be updated to accept language and shop data.

#### In 1.6
When I enable "PS_GIFT_WRAPPING" configuration in "All shops" context, then it is saved as:

| id_shop  | id_shop_group | value |
| ------------- | ------------- | --- |
| NULL  | NULL  | 1 |

When I disable "PS_GIFT_WRAPPING" in "Single shop" context, then it is saved as (creates new configuration as it overrides "All shops" value):

| id_shop  | id_shop_group | value |
| ------------- | ------------- | --- |
| NULL  | NULL  | 1 |
| 1  | 1  | 0 |

When I enable "PS_GIFT_WRAPPING" in "Single shop" context, then it is saved as (it deletes specific value for "Single shop" as it matches "All Shops" context value):

| id_shop | id_shop_group | name|
|------------|--------------------|--------|
| NULL     | NULL               | 1       |

#### In 1.7
[...]

Issue #12540, opened 2019-02-12
Multi-store feature usage within the forms


tomas862:
### Description
On legacy pages we had feature which allowed us to restrict certain settings update to certain shops.

### Explanation of the functionality

In short, when a multistore feature is enabled and when you are in specific shop context you can see a form which looks like this:

![multi_store_all_disabled](https://user-images.githubusercontent.com/17051250/52632683-e31b4f00-2eca-11e9-93a3-4f91b2b317fe.png)

the disabled fields and checkboxes near the form elements means that **the form inherits all the settings which can only be changed in all shops context**

If I start ticking the checkboxes then I choose a **specific configuration only for the current shop context I am in**

#### Displaying checkboxes and switch

It seems like we need a way to modify each form type - to decorate it with extra checkboxes and the `YesNoSwtich` selector. Talking in very abstract way, I'm thinking of having a Decorator class which will append extra form types:

1. Use it in your form type like:
```php
public function buildForm(FormBuilderInterface $builder, array $options)
    {
        $builder
            ->add('theme_to_adapt', ChoiceType::class, [
                'choices' => $this->themeChoices,
            ])
            ->add('generate_rtl_css', SwitchType::class, [
                'required' => false,
                'data' => false,
            ])
        ;

        $this->multiStoreFormDecorator->decorate($builder);
    }
```

also , maybe we don't want that all fields would have multi-store checkbox relation:

```php
public function buildForm(FormBuilderInterface $builder, array $options)
    {
        $builder
            ->add('theme_to_adapt', ChoiceType::class, [
                'choices' => $this->themeChoices,
            ])
            ->add('generate_rtl_css', SwitchType::class, [
                'required' => false,
                'data' => false,
            ])
        ;

        $this->multiStoreFormDecorator->decorateFor($builder, ['theme_to_adapt']);
    }
```
2. Inside the decorator I'll iterate over available form types and add another type which will be a checkbox and it will be named like `theme_to_adapt_is_restricted_to_shop` or something :thinking: (Basic idea is to append some text to the original for type for easier usage withing javasscript etc..)
Also a new switch type will be appended as well  - **not sure if it should be optional or not**

With this usage we have full control ir our twig template to put these checkboxes in appropriate place. **Maybe in form_group_row macro we can attach it automatically ?**

#### Submitting the data

[...]

PR #12331, opened 2019-01-28
Migration of Theme & Logo page with style and ux improvements

tomas862:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | This PR changes styles and adds UX improvements
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10576
| How to test?  | Check Zeplin how it looks like - should be similar. Don't forget to rebuild assets by running command `npm run build`. Some useful information  when testing - the file inputs still has old design but the PR in https://github.com/PrestaShop/PrestaShop/pull/12436 will fix them

Warning: there is a known issue if you upload a bad theme file using a zip archive. It's fixed by https://github.com/PrestaShop/PrestaShop/pull/13105

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/12331)

PR #14802, opened 2019-07-22
[POC] Multistore behavior for forms

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | This PR introduces POC of how multistore behavior could be handled in forms.
| Type?         | new feature
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #14731 
| How to test?  | Not needed.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14802)

### DIFF
--- a/src/PrestaShopBundle/Form/Admin/Configure/ShopParameters/ProductPreferences/ProductType.php
+++ b/src/PrestaShopBundle/Form/Admin/Configure/ShopParameters/ProductPreferences/ProductType.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShopBundle\Form\Admin\Configure\ShopParameters\ProductPreferences;
+
+use PrestaShopBundle\Form\Admin\Type\ConfigurationType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+// @todo: TO BE REMOVED
+class ProductType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('my_field', TextType::class)
+        ;
+    }
+
+    public function getParent()
+    {
+        return ConfigurationType::class;
+    }
+}
--- a/src/PrestaShopBundle/Form/Admin/Extension/ConfigurationTypeExtension.php
+++ b/src/PrestaShopBundle/Form/Admin/Extension/ConfigurationTypeExtension.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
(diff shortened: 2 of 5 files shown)


src/PrestaShopBundle/Form/Admin/Configure/ShopParameters/ProductPreferences/ProductType.php
```diff
@@ -0,0 +1,48 @@
...
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('my_field', TextType::class)
+        ;
+    }
+
+    public function getParent()
+    {
+        return ConfigurationType::class;
```
sarjon, line 46: All you have to do is make `ConfigurationType` to be parent.

src/PrestaShopBundle/Form/Admin/Extension/ConfigurationTypeExtension.php
```diff
@@ -0,0 +1,100 @@
...
+            foreach ($form->all() as $child) {
+                $configFields[$child->getName()] = $child->getName();
+            }
+
+            $form->add('multistore_config_switch', SwitchType::class, [
+                'required' => false,
+                'data' => false,
+            ]);
+
+            $form->add('multistore_config', ChoiceType::class, [
```
sarjon, line 83: Not sure if `ChoiceType` would be the best here, since we might need to render each multistore checkbox separately.

src/PrestaShopBundle/Form/Admin/Extension/ConfigurationTypeExtension.php
```diff
@@ -0,0 +1,100 @@
...
+            ]);
+        });
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getExtendedType()
+    {
+        return ConfigurationType::class;
```
sarjon, line 98: Extends all `ConfigurationType` with multistore behavior.

src/PrestaShopBundle/Form/Admin/Type/ConfigurationType.php
```diff
@@ -0,0 +1,33 @@
...
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShopBundle\Form\Admin\Type;
+
+use Symfony\Component\Form\AbstractType;
+
+class ConfigurationType extends AbstractType
```
sarjon, line 31: I think this could be removed in favor of multistore `option` on form type.

matks:
@eternoendless You want to work with us on this 😉

src/PrestaShopBundle/Form/Admin/Extension/ConfigurationTypeExtension.php
```diff
@@ -0,0 +1,100 @@
...
+use PrestaShopBundle\Form\Admin\Type\SwitchType;
+use Symfony\Component\Form\AbstractTypeExtension;
+use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\Extension\Core\Type\CollectionType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
+
+class ConfigurationTypeExtension extends AbstractTypeExtension
```
matks, line 41: It's a very nice idea to use `AbstractTypeExtension` I did not even know about it 👏

matks:
So, like @tomas862 did for Theme&Logo page, you plan to use form event listeners to dynamically add the multishop checkboxes ?

If I understand correctly, once the system is done, when a developer wants to create a multistore-compatible form, he will have to:
- add getParent() { return ConfigurationType::class; } to make it multishop-aware
- indicate which fields is multistore-eligible (so we know which fields gets a checkbox)
- where will be retrieved the checkbox data ? the multistore_config will not be removed from the form, so in the FormHandler we'll have a multistore_config input we can extract the data from ?

matks:
By the way, here is the 1st POC built by @tomas862 :
https://github.com/PrestaShop/PrestaShop/pull/12331/files#diff-c761f50618701a6d1ec1fea9dcd817cc

sarjon:
> add getParent() { return ConfigurationType::class; } to make it multishop-aware

Yes, but I'm also thinking about another solution, instead of extending configuration form, it would be a global option, for example:
```php
$this->createForm(MyFormType::class, [], ['is_multistore_aware' => true]);
```

> indicate which fields is multistore-eligible (so we know which fields gets a checkbox)

By default, all fields inside form should be multistore-eligible, however, developer should be able to exclude some manually (probably via option, but this is not implemented)

> where will be retrieved the checkbox data ? the multistore_config will not be removed from the form, so in the FormHandler we'll have a multistore_config input we can extract the data from ?

Yes, wherever form is handled you should be able to access this data. Unfortunately, current implementations for persisting configuration data don't care about multistore, so I think there should be a separate PR to implement that.

matks:
Found old issue about the topic, interesting as well 👍 
https://github.com/PrestaShop/PrestaShop/issues/12540

sarjon:
@matks what do we do with this PR?

matks:
> @matks what do we do with this PR?

This work is on pause right now so we can focus on Orders pages. Once we get our hands out of water, we can come back on this. @eternoendless wants to collaborate on this topic as it is critical we build a multistore system for Symfony a lot more reliable than legacy one

matks:
I believe @matthieu-rolland used this as a POC, thanks @sarjon 😊  . I close it now that we're close to achieving it.

Issue #10632, opened 2018-09-21
Improve "Catalog > Products > Add new / edit product" page & "Products List" page


matks:
Part of Symfony migration project - this page needs rework to be built the same way other migrated pages are done
Type: page

# Product Page Multistore migration

Use the opportunity of product page migration to provide the best product page for merchants that is compatible with multistore.

Many merchants have difficulties about the multistore that is not quite fully working. The multistore actions are unclear on the product page. The merchant doesn't understand if the actions will impact all the shops or not.

## Status

Todo:
- [x] Find all technical obstacles and fix them, see https://github.com/PrestaShop/PrestaShop/issues/14405
- [x] Prepare the new architecture, see https://github.com/PrestaShop/PrestaShop/issues/14492
- [x] Multistore header: [Specs](https://github.com/PrestaShop/prestashop-specs/blob/7e820f30347fe69ccd3db5a99ab02ef6f7bbdd54/back-office/multistoregeneralspecs.md#multistore-header-only-from-178)
- [x] Multistore - Update message when a group is selected https://github.com/PrestaShop/PrestaShop/issues/19459
- [x] Enable product page as stable one https://github.com/PrestaShop/PrestaShop/issues/31086
- [x] #31373
### Left todo
**Product List**

- [x] Button Preview https://github.com/PrestaShop/PrestaShop/pull/29039
- [x] Duplicate 
- [x] Delete 
- [x] Enable/Disable 

**Multishop**

//Product List
- [x] Modal to select shop on creation when listing is in All shops
- [x] The multishop selector in product page only shows shops associated to current product & Handle « add product » for multishop (assignation) https://github.com/PrestaShop/PrestaShop/pull/27712
- [x] #29986
- [x] #29987
- [x] Enable/Disable from all shop / one shop https://github.com/PrestaShop/PrestaShop/pull/28245

//Product Add/Edit
- [x] #30119
- [x] #30120
- [x] #30121
- [x] Handle BasicInformation for multishop
- [x] Handle prices data for multishop https://github.com/PrestaShop/PrestaShop/pull/25464
- [x] Handle seo data for multishop
- [x] Handle options data for multishop
- [x] Handle stock data for multishop
- [x] #29343
- [ ] #30124 (postponed to V9 for BC reasons)
- [x] Handle customization fields data for multishop
https://github.com/PrestaShop/PrestaShop/issues/27579
[...]

PR #14824, opened 2019-07-23
Migration of product command parameters and value objects only

tomas862:
| Questions     | Answers
| Branch?       | develop
| Description?  | Huge command is needed to save all fields of products - despite that, different type of products have different fields which are needed. By following this logic each product type will have their own command - the abstract command holds all the common fields which can be used in any type of the product. Product commands will not have all values scalar due to reason that data structure is super complex to put it in array.
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10632
| How to test?  | This PR will be covered with Integration tests when the handlers will be working - right now it only has unit tests. No manual testing will be needed for this.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14824)


src/Core/Domain/Product/ValueObject/NewConfigurableImage.php
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * @param UploadedFile $file
+     * @param int $position
+     * @param bool $isCover
+     * @param array $localizedCaptions
+     *
+     * @throws ProductConstraintException
+     */
+    public function __construct(
+        UploadedFile $file,
```
sarjon: It should accept path to image instead.

src/Core/Domain/Product/ValueObject/ResponseCode.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Response code for product unavailable case.
+ */
+class ResponseCode
```
sarjon: Not sure about this VO, response code is very technical. :/

tomas862: but it does exist in Domain - even visible in the form. If it only would hold id then I would agree with you but this VO also validates available types

tomas862:
Hi @eternoendless got a few questions regarding products form commands structure.

1 question:

I know our goal is to make the commands as simple as possible to use which one part is:
**use scalar values**

but the products commands -  add, edit are very complex and has huge amount of data needed to be collected in order to submit it fully. 

To understand it better here's the first sample https://github.com/PrestaShop/PrestaShop/pull/14824/files#diff-41a11dc481ef2d1f7e79a0d2844c8976R242 which is `RetailPrice` VO object which accepts 3 arguments via constructor : price, taxRuleId, displayOnSaleFlag.  I don't think it would be better to have an array with all those 3 parameters. Its just a single sample from huge command. More complex example is https://github.com/PrestaShop/PrestaShop/pull/14824/files#diff-41a11dc481ef2d1f7e79a0d2844c8976R202 `ImageCollection` - it does not only have nested data structure inside but for Image upload, update and delete actions I want to apply different ValueObjects - using special case design pattern. If its gonna be array all the way then there will be a lot of conditions to construct such array in the first place.


2 question:

if you see nothing wrong with using VO's instead of complex arrays then my second idea is combine them form types - it means that when we create a form type and it is submitted we use Symfony data transformer to transform form values to the required structure containing product value objects. What do you think about this idea?

eternoendless:
For argument's sake, let's say we agree on using multiple VOs for the command. What would the HTTP POST request look like, and how would you map it to your command?

tomas862:
1. If I understand you correctly we agree that using scalar values such as array for complex data structures would be not wise in this case?


2. Regarding the:

> For argument's sake, let's say we agree on using multiple VOs for the command. What would the HTTP POST request look like, and how would you map it to your command?

example $_POST:
```php
[
    'retail_price' => [
        'price' => 5.5,
        'tax_rule_id' => 1,
        'display_on_sale_flag' => true,
    ]
];
```

so using the https://symfony.com/doc/3.4/form/data_transformers.html#creating-the-transformer with `RetailPriceType` 

```php
class MyTransformer {
    public function reverseTransform($value)
    {
        return new RetailPrice(
            $value['price'],
            $value['tax_rule_id'],
            'display_on_sale_flag',
        );
    }
}
```

so after this action which is applied automatically on each POST request instead of having array form value `retail_price` we will have an object instead.

So benefits are that we can test each type of transformation and have very clean solution when setting it in form handler. We would only need to use something like this:

```php

(new AddProductCommand())->setRetailPrice($dataFromPost['retail_price']);

```

eternoendless:
I'm divided regarding this subject. On one hand it would simplify the interface, but on the other hand it would leak domain objects.

Maybe we're approaching the issue from the wrong end. For example, you said:

> Its just a single sample from huge command

This is a red flag. A lot of parameters often point to a design problem due to excessive complexity. The command is huge as it is. But _does it have to?_

Looking at `ps_product`, most properties don't seem to be mandatory. Maybe the `CreateProduct` could be very, very simple. If most properties are optional, they could be handled via one or many update processes after creation.

tomas862:
Hi @eternoendless - reduced the size of the product command. Now it has **20** properties from which only **1** is required. It is still a lot.

By just looking here https://github.com/PrestaShop/PrestaShop/pull/14824/files#diff-41a11dc481ef2d1f7e79a0d2844c8976R52 its obvious that I can move out price information, reference information, meta fields information in separate commands but at this point I have a question:

If we start thinking about legacy object `Product` when instead of calling 1 time `save` method we will do that multiple times. We might brake some compatibility for module developers.

but I assume we should ignore that and create the command which has more sense for new system and ignore legacy problems, right? :)

src/Core/Domain/Product/Attachment/Command/UpdateAttachmentsCommand.php
```diff
@@ -0,0 +1,75 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Attachment\Command;
```
sarjon: I know @RaimondasSapola is working on `Attachment` and he put it under `Domain/Attachment` instead. Is attachment Product only specific thing?

src/Core/Domain/Product/Attachment/Command/UpdateAttachmentsCommand.php
```diff
@@ -0,0 +1,75 @@
...
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Attachment\Exception\ProductAttachmentException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Attachment\ValueObject\Attachment;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+
+/**
+ * Updates product attachments.
+ */
+class UpdateAttachmentsCommand
```
sarjon: Maybe `UpdateProductAttachmentsCommand`, not sure if it's necessary, but maybe it's more clear.

src/Core/Domain/Product/Attachment/ValueObject/Attachment.php
```diff
@@ -0,0 +1,141 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\Attachment\Exception\ProductAttachmentException;
+use SplFileObject;
+use function strlen;
+
+/**
+ * Add files that customers can download directly on the product page (instructions, manual, recipe, etc.).
+ */
+final class Attachment
+{
+    public const MAX_SIZE = 32;
```
sarjon: What's max size?  Is it file size? What unit is 32?

src/Core/Domain/Product/Attachment/ValueObject/Attachment.php
```diff
@@ -0,0 +1,141 @@
...
+     */
+    public function __construct(
+        string $filePath,
+        array $localizedTitles,
+        array $localizedDescriptions
+    ) {
+        $this->assertIsTitleLengthValid($localizedTitles);
+
+        $file = new SplFileObject($filePath);
+        $this->assertIsFile($file);
```
sarjon: I think `new SplFileObject` throws exception if file does not exist.

src/Core/Domain/Product/Category/Command/UpdateCategoriesCommand.php
```diff
@@ -0,0 +1,89 @@
...
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Category\Command;
+
+use PrestaShop\PrestaShop\Core\Domain\Category\Exception\CategoryException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Category\ValueObject\Category;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+
+/**
+ * Updates product categories.
+ */
+class UpdateCategoriesCommand
```
sarjon: How about `UpdateProductCategoriesCommand`?

src/Core/Domain/Product/Category/Command/UpdateCategoriesCommand.php
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * @param array $categories
+     * @throws CategoryException
+     */
+    private function setCategories(array $categories)
+    {
+        foreach ($categories as $category) {
+            $this->categories[] = new Category(
+                $category['id'],
+                $category['is_main_category']
```
sarjon: Shouldn't you check if there is only 1 main category?

src/Core/Domain/Product/Category/ValueObject/Category.php
```diff
@@ -0,0 +1,74 @@
...
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Category\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Category\Exception\CategoryException;
+use PrestaShop\PrestaShop\Core\Domain\Category\ValueObject\CategoryId;
+
+/**
+ * Holds information about product category.
+ */
+final class Category
```
sarjon: How about `ProductCategory`?

src/Core/Domain/Product/Command/AbstractProductCommand.php
```diff
@@ -0,0 +1,568 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\Visibility;
+
+/**
+ * Holds the abstraction of common product data which does not depend from product type.
+ */
+abstract class AbstractProductCommand
+{
+    /**
+     * @todo: I need defaultLanguage validation in handler
+     * @var ProductName[]
```
sarjon: Maybe it's a bit overhead to have `ProductName` VO? We could keep it simple by having `string[] $localizedName`.

src/Core/Domain/Product/Command/AbstractProductCommand.php
```diff
@@ -0,0 +1,568 @@
...
+     */
+    private $localizedDescription;
+
+    /**
+     * @var ManufacturerId
+     */
+    private $manufacturerId;
+
+    /**
+     * @var MetaTitle[]
```
sarjon: Same for meta VOs, are they really necessary?

src/Core/Domain/Product/Command/AbstractProductCommand.php
```diff
@@ -0,0 +1,568 @@
...
+    {
+        return $this->retailPrice;
+    }
+
+    /**
+     * @param RetailPrice $retailPrice
+     *
+     * @return self
+     */
+    public function setRetailPrice(RetailPrice $retailPrice): self
```
sarjon: Shouldn't it accept only scalar values?

src/Core/Domain/Product/Command/AbstractProductCommand.php
```diff
@@ -0,0 +1,568 @@
...
+    {
+        return $this->unitPrice;
+    }
+
+    /**
+     * @param UnitPrice $unitPrice
+     *
+     * @return self
+     */
+    public function setUnitPrice(UnitPrice $unitPrice): self
```
sarjon: Same here, I guess it's the same for all setters.

src/Core/Domain/Product/Command/AbstractProductCommand.php
```diff
@@ -0,0 +1,568 @@
...
+    {
+        return $this->localizedSummary;
+    }
+
+    /**
+     * @param array|string[] $localizedSummary
+     *
+     * @return self
+     */
+    public function setLocalizedSummary($localizedSummary): self
```
sarjon: Typehint `array`.

src/Core/Domain/Product/Command/AbstractProductCommand.php
```diff
@@ -0,0 +1,568 @@
...
+    {
+        return $this->localizedDescription;
+    }
+
+    /**
+     * @param array|string[] $localizedDescription
+     *
+     * @return self
+     */
+    public function setLocalizedDescription($localizedDescription): self
```
sarjon: Typehint `array`.

src/Core/Domain/Product/Command/UpdateRelatedProductsCommand.php
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Command;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+
+/**
+ * Update related products.
+ */
+class UpdateRelatedProductsCommand
```
sarjon: How about `UpdateProductRelatedProductsCommand`?

src/Core/Domain/Product/CustomizationField/Command/UpdateCustomizationFieldsCommand.php
```diff
@@ -0,0 +1,73 @@
...
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\CustomizationField\Command;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\CustomizationField\ValueObject\CustomizationFieldInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+
+/**
+ * Update product customization fields.
+ */
+class UpdateCustomizationFieldsCommand
```
sarjon: How about `UpdateProductCustomizationFieldsCommand`?

src/Core/Domain/Product/CustomizationField/ValueObject/CustomizationFieldInterface.php
```diff
@@ -0,0 +1,47 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\CustomizationField\ValueObject;
+
+/**
+ * Defines contract for customization fields.
+ */
+interface CustomizationFieldInterface
```
sarjon: This interface seems redundant for me.

src/Core/Domain/Product/CustomizationField/ValueObject/FileCustomizationField.php
```diff
@@ -0,0 +1,77 @@
...
+     * @param array $localizedLabels
+     * @param bool $isRequired
+     *
+     * ProductConstraintException
+     */
+    public function __construct(array $localizedLabels, bool $isRequired)
+    {
+        $this->localizedLabels = array_map(
+            static function (string $label) {
+                return new Label($label);
```
sarjon: Having `Label` VO seems too much imo. :sweat_smile:

src/Core/Domain/Product/CustomizationField/ValueObject/TextCustomizationField.php
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\CustomizationField\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+
+/**
+ * Text type customization field.
+ */
+final class TextCustomizationField implements CustomizationFieldInterface
```
sarjon: I think it's better to have `CustomizationField` VO with 3 properties: name, isRequired and type.

src/Core/Domain/Product/Feature/Command/UpdateFeaturesCommand.php
```diff
@@ -0,0 +1,68 @@
...
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Feature\Command;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Feature\DTO\FeatureCollection;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+
+/**
+ * Update product features.
+ */
+class UpdateFeaturesCommand
```
sarjon: How about `UpdateProductFeaturesCommand`?

src/Core/Domain/Product/Exception/ProductConstraintException.php
```diff
@@ -0,0 +1,50 @@
...
+    public const INVALID_META_DESCRIPTION = 13;
+    public const META_DESCRIPTION_NAME_TOO_LONG = 14;
+    public const FRIENDLY_URL_TOO_LONG = 15;
+    public const INVALID_RESPONSE_CODE = 16;
+    public const INVALID_CONDITION_TYPE = 17;
+    public const INVALID_VISIBILITY_TYPE = 18;
+    public const INVALID_REFERENCE = 19;
+    public const INVALID_ISBN_REFERENCE = 20;
+    public const INVALID_EAN13_REFERENCE = 21;
+    public const INVALID_UPC_REFERENCE = 22;
```
sarjon: Not sure if last 3 consts should have `_REFERENCE` suffixes. :thinking:

src/Core/Domain/Product/Feature/Command/UpdateFeaturesCommand.php
```diff
@@ -0,0 +1,68 @@
...
+     * @var ProductId
+     */
+    private $productId;
+
+    /**
+     * @var FeatureCollection
+     */
+    private $features;
+
+    public function __construct(int $productId, FeatureCollection $features)
```
sarjon: Shouldn't it accept `array` instead of `FeatureCollection`?

src/Core/Domain/Product/Feature/DTO/FeatureCollection.php
```diff
@@ -0,0 +1,82 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Feature\DTO;
```
sarjon: `DTO` namespace seems weird, although I don't have a better alternative right now. :/

src/Core/Domain/Product/Feature/DTO/FeatureCollection.php
```diff
@@ -0,0 +1,82 @@
...
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Feature\ValueObject\CustomizableFeatureValue;
+use PrestaShop\PrestaShop\Core\Domain\Product\Feature\ValueObject\FeatureValue;
+
+/**
+ * Gets a collection from all possible feature values that can be applied for product.
+ */
+class FeatureCollection
```
sarjon: How about calling it `ProductFeatures` or `ProductFeaturesCollection`?

src/Core/Domain/Product/Feature/DTO/FeatureCollection.php
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * @var FeatureValue[]
+     */
+    private $existingFeatureValues;
+
+    /**
+     * @var CustomizableFeatureValue[]
+     */
+    private $customizableFeatureValues;
+
```
sarjon: Isn't there any constructor? :/

src/Core/Domain/Product/Feature/DTO/FeatureCollection.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        return $this->existingFeatureValues;
+    }
+
+    /**
+     * @param FeatureValue $existingFeatureValue
+     *
+     * @return void
+     */
+    public function setExistingFeatureValue(FeatureValue $existingFeatureValue): void
```
sarjon: You don't `set`, you `add` instead.

src/Core/Domain/Product/Feature/DTO/FeatureCollection.php
```diff
@@ -0,0 +1,82 @@
...
+{
+    /**
+     * @var FeatureValue[]
+     */
+    private $existingFeatureValues;
+
+    /**
+     * @var CustomizableFeatureValue[]
+     */
+    private $customizableFeatureValues;
```
sarjon: Hmm, do we need 2 arrays for product features? 

How about some kind of `ProductFeature` object which could be created either with `ProductFeature::withPredefinedValue($featureId, $featureValueId)` or `ProductFeature::withCustomizedValue($featureId, 'custom value')`?

src/Core/Domain/Product/Feature/ValueObject/CustomizableFeatureValue.php
```diff
@@ -0,0 +1,111 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Feature\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Feature\Exception\ProductFeatureConstraintException;
+use function strlen;
+
+final class CustomizableFeatureValue
```
sarjon: As suggested above, maybe we could avoid having this VO at all?

src/Core/Domain/Product/Feature/ValueObject/FeatureValue.php
```diff
@@ -0,0 +1,108 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Feature\ValueObject;
+
+/**
+ * Existing Feature value.
+ */
+final class FeatureValue
```
sarjon: Same for this VO, maybe it's not needed if you change design a little.

src/Core/Domain/Product/Image/Command/UpdateImagesCommand.php
```diff
@@ -0,0 +1,72 @@
...
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Image\Command;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Image\DTO\ImageCollection;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+
+/**
+ * Updates product images.
+ */
+class UpdateImagesCommand
```
sarjon: How about `UpdateProductImagesCommand`?

I'm also started to think what is image update? Shouldn't we have "upload images" instead? If so, maybe it's not Domain responsibility to do so in the first place?

src/Core/Domain/Product/Image/DTO/ImageCollection.php
```diff
@@ -0,0 +1,106 @@
...
+class ImageCollection
+{
+    /** @var NewConfigurableImage[] */
+    private $newImages;
+
+    /** @var ExistingConfigurableImage[] */
+    private $existingImages;
+
+    /** @var DeletedImage[] */
+    private $deletesImages;
```
sarjon: How do you track deleted images? What is considered a deleted image? I think that user is able to delete image one by one.

src/Core/Domain/Product/Image/ValueObject/ConfigurableImageInterface.php
```diff
@@ -0,0 +1,54 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Image\ValueObject;
+
+/**
+ * Defines contract for image which can be configured.
+ */
+interface ConfigurableImageInterface
```
sarjon: I'm not sure about this abstraction, it seems redundant for me.

src/Core/Domain/Product/Image/ValueObject/DeletedImage.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Image\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Image\Exception\ProductImageConstraintException;
+
+/**
+ * Image which is about to be deleted.
+ */
+final class DeletedImage
```
sarjon: You should double check if having this kind of VO makes sense in our domain.

src/Core/Domain/Product/Image/ValueObject/ExistingConfigurableImage.php
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Image\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Image\Exception\ProductImageConstraintException;
+
+/**
+ * Existing image
+ */
+final class ExistingConfigurableImage implements ConfigurableImageInterface
```
sarjon: Shouldn't it just be `ProductImage`? I don't think there is concept as `ExistingConfigurableImage`.

src/Core/Domain/Product/Image/ValueObject/ImageId.php
```diff
@@ -0,0 +1,59 @@
...
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Image\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Image\Exception\ProductImageConstraintException;
+
+final class ImageId
```
sarjon: Some VOs are `final` and some aren't, what is the difference?

src/Core/Domain/Product/Image/ValueObject/ImageId.php
```diff
@@ -0,0 +1,59 @@
...
+    private $imageId;
+
+    /**
+     * @param int $imageId
+     *
+     * @throws ProductImageConstraintException
+     */
+    public function __construct(int $imageId)
+    {
+        if (0 > $imageId) {
```
sarjon: Does it mean that `0` is valid image id?

src/Core/Domain/Product/Image/ValueObject/NewConfigurableImage.php
```diff
@@ -0,0 +1,151 @@
...
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Image\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Image\Exception\ProductImageConstraintException;
+use SplFileObject;
+use function in_array;
+
+/**
+ * New not created image.
+ */
+final class NewConfigurableImage implements ConfigurableImageInterface
```
sarjon: Same here, I feel like this needs redesign.

src/Core/Domain/Product/Supplier/ValueObject/Reference.php
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Supplier\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Supplier\Exception\ProductSupplierConstraintException;
+
+/**
+ * Products supplier reference
+ */
+final class Reference
```
sarjon: Can you reuse same `Reference` object from `Product`? It seems that it's the same reference, except supplier specific, right?

src/Core/Domain/Product/Supplier/ValueObject/Supplier.php
```diff
@@ -0,0 +1,135 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\Reference\Reference;
+use PrestaShop\PrestaShop\Core\Domain\Supplier\Exception\SupplierException;
+use PrestaShop\PrestaShop\Core\Domain\Supplier\ValueObject\SupplierId;
+use PrestaShop\PrestaShop\Core\Domain\ValueObject\Price;
+
+/**
+ * Holds supplier and product related information.
+ */
+final class Supplier
```
sarjon: How about `ProductSupplier`?

There are multiple objects in VO namepsace, but can they really be considered VOs? For me, it seems that a lot of them serves as DTOs.

src/Core/Domain/Product/ValueObject/Condition.php
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Holds product condition either its new , used etc...
+ */
+final class Condition
+{
+    public const IS_NEW = 'new';
+
+    public const IS_USED = 'used';
+
+    public const IS_REFURBISHED = 'refurbished';
```
sarjon: `IS_` prefix does not seem to be necessary here.

src/Core/Domain/Product/ValueObject/Condition.php
```diff
@@ -0,0 +1,97 @@
...
+     */
+    private $displayedOnProductPage;
+
+    /**
+     * @param string $condition
+     * @param bool $displayedOnProductPage
+     *
+     * @throws ProductConstraintException
+     */
+    public function __construct(string $condition, bool $displayedOnProductPage)
```
sarjon: Is `$displayedOnProductPage` really part of condition?

src/Core/Domain/Product/ValueObject/MetaData/FriendlyUrl.php
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\MetaData;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+
+/**
+ * An url which appears in the browser for product.
+ */
+final class FriendlyUrl
```
sarjon: I'd go for plain old string instead of VO.

src/Core/Domain/Product/ValueObject/Price/CostPrice.php
```diff
@@ -0,0 +1,71 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\Price;
```
sarjon: Not sure if need need to split `ValueObject` into multiple namespaces.

src/Core/Domain/Product/ValueObject/Price/CostPrice.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @param float $price
+     *
+     * @throws ProductConstraintException
+     */
+    public function __construct(float $price)
+    {
+        try {
+            $numberPrice = (new Price($price))->getValue();
```
sarjon: Why not just use `Price` instead of `CostPrice`? This VO does not add anything new.

src/Core/Domain/Product/ValueObject/Price/RetailPrice.php
```diff
@@ -0,0 +1,104 @@
...
+     * @param int $taxRuleId
+     * @param bool $displayOnSaleFlag
+     *
+     * @throws TaxRuleConstraintException
+     * @throws ProductConstraintException
+     */
+    public function __construct(float $priceWithoutTax, int $taxRuleId, bool $displayOnSaleFlag)
+    {
+        try {
+            $this->priceWithoutTax = (new Price($priceWithoutTax))->getValue();
```
sarjon: Maybe keep `$this->priceWithoutTax` as `Price` VO?

src/Core/Domain/Product/ValueObject/Price/RetailPrice.php
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * This is the net sales price for your customers.
+ * The retail price will automatically be calculated using the applied tax rate.
+ */
+final class RetailPrice
+{
+    /**
+     * @var bool
+     */
+    private $displayOnSaleFlag;
```
sarjon: Does it make sense to have this flag as part of `RetailPrice`?

src/Core/Domain/Product/ValueObject/Price/UnitPrice.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * @param float $price
+     * @param string $unit
+     *
+     * @throws ProductConstraintException
+     */
+    public function __construct(float $price, string $unit)
+    {
+        try {
+            $this->price = (new Price($price))->getValue();
```
sarjon: Same here, how about keeping `$this->price` as `Price`?

src/Core/Domain/Product/ValueObject/RedirectionPage/CategoryRedirectionPage.php
```diff
@@ -0,0 +1,86 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Category\Exception\CategoryException;
+use PrestaShop\PrestaShop\Core\Domain\Category\ValueObject\CategoryId;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\RedirectionPage\TypedRedirectionPageInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\RedirectionPage\ResponseCode;
+
+/**
+ * Category redirection page
+ */
+final class CategoryRedirectionPage implements TypedRedirectionPageInterface
```
sarjon: For me it seems more complex than it should be. :confused: 

What if we had VO like `OfflineRedirection` (name can be improved):

```php
class OfflineRedirection {
    /** @param string */
    private $redirectType; // value could be "temporary_to_product" or "permanent_to_product"

    /** @param int|null */
    private $resourceId; // Product or category id, or null if redirect type is "none"
}
```

src/Core/Domain/Product/ValueObject/RedirectionPage/RedirectionPage.php
```diff
@@ -0,0 +1,60 @@
...
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\RedirectionPage;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\RedirectionPage\RedirectionPageInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\RedirectionPage\ResponseCode;
+
+/**
+ * Response code for product unavailable case.
+ */
+final class RedirectionPage implements RedirectionPageInterface
```
sarjon: Same here, seems redundant.

src/Core/Domain/Product/ValueObject/RedirectionPage/ResponseCode.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\RedirectionPage;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Response code for product unavailable case.
+ */
+final class ResponseCode
```
sarjon: We should not deal with response codes in domain imo. As suggested above, we could have "redirect type" with value like `temporary` and that string could be translated into http code at controller layer.

src/Core/Domain/Product/ValueObject/Reference/Ean13.php
```diff
@@ -0,0 +1,83 @@
...
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\Reference;
```
sarjon: I don't think we need to put it inside `Reference` namespace.

src/Core/Domain/Product/ValueObject/Reference/Ean13.php
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * This type of product code is specific to Europe and Japan, but is widely used internationally.
+ * It is a superset of the UPC code: all products marked with an EAN will be accepted in North America.
+ */
+final class Ean13
+{
+    /**
+     * @var string
+     */
+    private $reference;
```
sarjon: It's not a reference, it's a number.

src/Core/Domain/Product/ValueObject/Reference/Isbn.php
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * The International Standard Book Number (ISBN) is used to identify books and other publications.
+ */
+final class Isbn
+{
+    /**
+     * @var string
+     */
+    private $reference;
```
sarjon: Same here, not reference, it's a number.

src/Core/Domain/Product/ValueObject/Reference/Upc.php
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Upc barcode. This type of product code is widely used in the United States, Canada, the United Kingdom, Australia,
+ * New Zealand and in other countries.
+ */
+final class Upc
+{
+    /**
+     * @var string
+     */
+    private $reference;
```
sarjon: I think it suppose to be a code, not a reference.

src/Core/Domain/Product/ValueObject/Visibility.php
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * @var bool
+     */
+    private $isAvailableForOrder;
+
+    /**
+     * @var bool
+     */
+    private $isWebOnly;
```
sarjon: Are `$isAvailableForOrder`  and `$isWebOnly` really part of `Visibility` VO?

src/Core/Domain/Product/ValueObject/Visibility.php
```diff
@@ -0,0 +1,112 @@
+<?php
+
```
sarjon, line 26: Missing license, check other files as well.

sarjon (commented):
This will definitely need Pablo review.

src/Core/Domain/Product/Attachment/Command/UpdateAttachmentsCommand.php
tomas862: changes so it does not conflict with Raimondas work

src/Core/Domain/Product/Attachment/Command/UpdateAttachmentsCommand.php
tomas862: this one is Removed since it will be introduced in Raimondas PR

src/Core/Domain/Product/Attachment/ValueObject/Attachment.php
tomas862: this part is gone

src/Core/Domain/Product/Category/Command/UpdateCategoriesCommand.php
tomas862: a namespace tells that it belongs to product - or its hard to read from it?

src/Core/Domain/Product/Attachment/ValueObject/Attachment.php
tomas862: you suggest to ignore custom file validations here since it throws exception anyway?

src/Core/Domain/Product/Command/AbstractProductCommand.php
tomas862: the only useful purpose I see that it servers as length validator - do you think we don't need to do such validations?

src/Core/Domain/Product/Command/AbstractProductCommand.php
tomas862: same https://github.com/PrestaShop/PrestaShop/pull/14824#discussion_r308784470

src/Core/Domain/Product/CustomizationField/ValueObject/FileCustomizationField.php
tomas862: so you suggest putting length validations here in or not validate at all?

src/Core/Domain/Product/Exception/ProductConstraintException.php
tomas862: they all mean the same - they are references to a product after all. You can find quicker the constants which are related with reference

src/Core/Domain/Product/Feature/Command/UpdateFeaturesCommand.php
tomas862: I have two objects here: `\PrestaShop\PrestaShop\Core\Domain\Product\Feature\ValueObject\FeatureValue` and `\PrestaShop\PrestaShop\Core\Domain\Product\Feature\ValueObject\CustomizableFeatureValue`. Both have different structure. If we use scalar array here Im afraid we have a lot of mixed values and I would need some if conditions here to collect proper objects.

src/Core/Domain/Product/Feature/DTO/FeatureCollection.php
tomas862: and in once case` ProductFeature->getValue()` will return `id` and in another case `string` ?

src/Core/Domain/Product/Feature/ValueObject/CustomizableFeatureValue.php
tomas862: lets finish discussion with this https://github.com/PrestaShop/PrestaShop/pull/14824#discussion_r308793356 and then we will decide :)

src/Core/Domain/Product/Image/Command/UpdateImagesCommand.php
tomas862: you mean it is not the domain responsibility to upload any kind of file or image? It also assigns image id with product id.

src/Core/Domain/Product/Image/DTO/ImageCollection.php
tomas862: since we avoiding "ghost" products , then in create action we will create a bunch of product images which will not be submitted instantly with an ajax due to we not yet have assigned product. But you consider that in the update action we will instantly start updating and deleting product images?

src/Core/Domain/Product/Image/ValueObject/DeletedImage.php
tomas862: replied here https://github.com/PrestaShop/PrestaShop/pull/14824#discussion_r308797406

src/Core/Domain/Product/Image/ValueObject/NewConfigurableImage.php
tomas862: well I don't think that way - to change naming is always welcome. Maybe interface is not needed as well?  Probably yes,  but having such deeply separated value objects will be beneficial due to:

imagine we have `ProductImage` VO - it always brings filePath information and other information with it. But for example when we update image caption for same image we don't need anything related with file. In your case you suggest having same value object for both add and update cases? Will it actually be value object in such case?

src/Core/Domain/Product/ValueObject/Condition.php
tomas862: there's a tiny checkbox in the form  to mark such thing - either it is visible in the product page or not.

src/Core/Domain/Product/ValueObject/Price/RetailPrice.php
tomas862: how do you suggest to pass it differently? Its related with RetailPrice

src/Core/Domain/Product/ValueObject/RedirectionPage/CategoryRedirectionPage.php
tomas862: im not sure about this now - my suggestion avoids null values . Tried to achieve that by applying special case design pattern usage here.

src/Core/Domain/Product/ValueObject/Visibility.php
tomas862: they are related  - how do you suggest to move these options since we will need to identify it anyway

src/Core/Domain/Product/Category/ValueObject/Category.php
eternoendless: A class should not be final unless it implements an interface

src/Core/Domain/Product/Command/AbstractProductCommand.php
eternoendless: I also think it's overkill. If we had to check for special characters and whatnot then maybe it could be justified, but fundamentally it's just a string.

src/Core/Domain/Product/Attachment/ValueObject/Attachment.php
sarjon: No, you can wrap native exception with domain exception.

src/Core/Domain/Product/Exception/ProductConstraintException.php
sarjon: But all of them except for reference are not even references, they're numbers, codes.

src/Core/Domain/Product/Feature/DTO/FeatureCollection.php
sarjon: No, it could have interface which is something like this:
```php
class ProductFeature {
    public function isPredefinedValue(): bool { ... }

    public function getPredefinedValue(): ?FeatureValueId { ... }

    public function getCustomizedValue(): ?string { ... }
}
```

src/Core/Domain/Product/Image/Command/UpdateImagesCommand.php
sarjon: Assigning image to product is part of domain, but dealing with filesystem is not, although it's done in other handlers as far as I remember.

src/Core/Domain/Product/Image/DTO/ImageCollection.php
sarjon: > But you consider that in the update action we will instantly start updating and deleting product images?

When you edit product, you just click delete and it should perform delete imo, you shouldn't need to click "Save product", since there is separate button on image for deleting. I guess same for upload when editing.

src/Core/Domain/Product/Image/ValueObject/NewConfigurableImage.php
sarjon: > But for example when we update image caption for same image we don't need anything related with file.

So maybe we should think differently? For me it seems as a totally different command, something like `UpdateImageCaptionCommand` or something like that.

src/Core/Domain/Product/ValueObject/Condition.php
sarjon: Yeah, but why it is related to condition? :thinking:

src/Core/Domain/Product/ValueObject/Price/RetailPrice.php
sarjon: For me it seems like a separate Product attribute that does not have anything to do with Retail Price. :thinking:

src/Core/Domain/Product/ValueObject/RedirectionPage/CategoryRedirectionPage.php
sarjon: Yeah, but that approach seems to introduce a lot of complexity imo.

src/Core/Domain/Product/ValueObject/Visibility.php
sarjon: It's just an attribute of Product. In this case, I don't think they are part of `Visibility` VO.

src/Core/Domain/Product/Command/AbstractProductCommand.php
```diff
@@ -0,0 +1,568 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\Reference\Ean13;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\Reference\Isbn;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\Reference\Reference;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\Reference\Upc;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\Visibility;
+
+/**
+ * Holds the abstraction of common product data.
+ */
+abstract class AbstractProductCommand
```
sarjon: Do you extend this class anywhere?

src/Core/Domain/Product/Feature/DTO/FeatureCollection.php
tomas862: im really not a fan of having interface with two possible outcomes - in my opinion its better to have two objects and avoid if conditions thus reducing code complexity.

src/Core/Domain/Product/Image/DTO/ImageCollection.php
tomas862: but remember when we have a new product we dont have an id of it. So you suggest when we upload new image for non existing product we will have to be forced to enter required product name before since we need an id_product?

src/Core/Domain/Product/Supplier/ValueObject/Supplier.php
tomas862: they are:
1. Once created immutable
2. Are valid

so they define what is called VO. why do you think they are not VO's ? They hold too much different data with them?

src/Core/Domain/Product/ValueObject/Condition.php
tomas862: do you have suggestions how I can move out such things? They both are related in the form - I dont want have million setters here - there are currently 20 setters already in the command which targers Product ObjectModel. Do you suggest to put another separate parameter `isConditionDisplayedOnProductPage` or something? Maybe combine these optional display booleans in some kind of DTO?

src/Core/Domain/Product/ValueObject/Price/RetailPrice.php
tomas862: same question https://github.com/PrestaShop/PrestaShop/pull/14824#discussion_r310132011

eternoendless:
I had a hard time figuring out what was the best approach here. On one hand, we want to avoid making Commands too big because they become hard to use. On the other hand, we want to follow DDD principles, that state that our design should be inspired from our business. And then, we want to leave place for things to be extensible.

Following my previous comment, here's my proposal. Given that a Product has a lot of moving parts, and given editing a Product is much more frequent than creating one, I think we should:

Set up an `AddProductCommand` that only receives mandatory parameters in its constructor. Then issue an `EditProductCommand` with ALL the other properties, which are handled via setters.

eg.

```php
$productId = $commandBus->handle(new AddProductCommand(
   // mandatory parameters only
));

$editProduct = (new EditProductCommand())
  ->setSupplier($supplier)
  ->setManufacturer($manufacturer)
  ->setPrice($price)
  // ... more setters
;
```

Grouping several properties in a single setter is tricky. Consider this case: if there were a `setDimensions($width, $height, $depth)`, how do I do if I only want to update the `$depth`, but keep the others as they were?

So for the grouping issue, the golden rule to me is: only group properties that cannot be modified if the others aren't modified as well.

eternoendless:
Regarding existing hooks, I think it should be handled on a case-by-case basis.

src/Core/Domain/Product/ValueObject/Visibility.php
tomas862: same problem - do you suggest to set them separately in the command?

src/Core/Domain/Product/Command/AbstractProductCommand.php
tomas862: currently no but I expect it will be AddProductCommand and UpdateProductCommand which will extend this.

src/Core/Domain/Product/Supplier/ValueObject/Supplier.php
sarjon: Yeah, but that can be applied to DTOs as well, especially int his case. For me it seems that we have more objects that main purpose is to transfer data.

We could ask for more opinions here.

src/Core/Domain/Product/ValueObject/Condition.php
sarjon: For me, separate property would be the best. See Pablo's answer about using setters, this would work imo.

src/Core/Domain/Product/ValueObject/Visibility.php
sarjon: Yeah.

src/Core/Domain/Product/ValueObject/Reference/Ean13.php
tomas862: just that it accepts value such as '05212-32' which is not a number actually - I would call it more like a reference.

sarjon: FIY https://en.wikipedia.org/wiki/International_Article_Number

tomas862:
@sarjon please check again the PR - most of the parts I made as requested.

the different parts:
* https://github.com/tomas862/PrestaShop/blob/debb706e89b772ffe8b4af3b491315d6943ecaa5/src/Core/Domain/Product/Feature/Command/UpdateProductFeaturesCommand.php#L37 instead of having single object I have two objects `CustomizableFeatureValue` and `FeatureValue` instead of suggested in here https://github.com/PrestaShop/PrestaShop/pull/14824#discussion_r309055527 in order to avoid null checking.

also this one is unresolved as well - https://github.com/PrestaShop/PrestaShop/pull/14824#discussion_r310148962 . For me it doesn't actually matter either we put it under `ValueObject` or `DTO` folders some objects such as `ProductSupplier` etc...

src/Core/Domain/Product/Attachment/Command/AssociateProductAttachmentsCommand.php
```diff
@@ -0,0 +1,89 @@
...
+ * Associates products with attachments.
+ */
+class AssociateProductAttachmentsCommand
+{
+    /**
+     * @var ProductId
+     */
+    private $productId;
+
+    private $attachmentIds;
```
sarjon: Missing typehint.

src/Core/Domain/Product/Attachment/Command/AssociateProductAttachmentsCommand.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public function getProductId(): ProductId
+    {
+        return $this->productId;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getAttachmentIds()
```
sarjon: You can typehint.

src/Core/Domain/Product/Category/Command/UpdateProductCategoriesCommand.php
```diff
@@ -0,0 +1,107 @@
...
+
+use PrestaShop\PrestaShop\Core\Domain\Category\Exception\CategoryException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Category\Exception\ProductCategoryConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Category\ValueObject\ProductCategory;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+
+/**
+ * Updates product categories.
+ */
+class UpdateProductCategoriesCommand
```
sarjon: I have one more suggestion regarding this command. `UpdateProductCategoriesCommand` gives impression that you can update categories data (name, description, status & etc), but in fact you can update association only. Maybe we could further improve naming by `UpdateProductCategoriesAssociationCommand` or similar? Wdyt?

src/Core/Domain/Product/Command/AddProductCommand.php
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Holds mandatory product data.
+ */
+class AddProductCommand
+{
+    /**
+     * @todo: I need defaultLanguage validation in handler
+     * @var string[]
+     */
+    private $localizedProductNames;
```
sarjon: We have only one name, but in different languages, would it make sense to rename to `$localizedName` also removing `Product` as it seems redundant. Wdyt?

src/Core/Domain/Product/Command/EditProductCommand.php
```diff
@@ -0,0 +1,745 @@
...
+
+    /**
+     * @var string[]
+     */
+    private $metaDescription;
+
+    /**
+     * @var string[]
+     */
+    private $metaKeywords;
```
sarjon: Meta data should be localized.

src/Core/Domain/Product/Image/Command/UpdateProductImagesCommand.php
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Image\Command;
+
+;
```
sarjon: Woops. :grin:

src/Core/Domain/Product/Image/ValueObject/ProductImage.php
```diff
@@ -0,0 +1,88 @@
...
+     * @param int $position
+     * @param bool $isCover
+     * @param array $localizedCaptions
+     */
+    public function __construct(
+        int $position,
+        bool $isCover,
+        array $localizedCaptions
+    ){
+
```
sarjon: Check code formatting for this line and line above.

src/Core/Domain/Product/Supplier/Command/UpdateProductSuppliersCommand.php
```diff
@@ -0,0 +1,103 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Exception\DomainConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Supplier\ValueObject\ProductSupplier;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+use PrestaShop\PrestaShop\Core\Domain\Supplier\Exception\SupplierException;
+
+/**
+ * Updates supplier data related with specific product.
+ */
+class UpdateProductSuppliersCommand
```
sarjon: Same for this command, you don't update suppliers, but association instead? :thinking: But again, I'm not sure we need to change it. :confused:

src/Core/Domain/Product/ValueObject/Condition.php
```diff
@@ -0,0 +1,83 @@
...
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+
+/**
+ * Holds product condition either its new , used etc...
+ */
+class Condition
+{
+    public const NEW = 'new';
```
sarjon: Isn't NEW a reserved php keyword? :open_mouth:

src/Core/Domain/Product/ValueObject/Ean13.php
```diff
@@ -0,0 +1,83 @@
...
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject;
+
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
```
sarjon, line 29: Too many empty lines above.

src/Core/Domain/Product/ValueObject/Isbn.php
```diff
@@ -0,0 +1,82 @@
...
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject;
+
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
```
sarjon, line 29: Same here, too many empty lines.

src/Core/Domain/ValueObject/Price.php
```diff
@@ -0,0 +1,81 @@
...
+     * @var Number
+     */
+    private $price;
+
+    /**
+     * @param float $price
+     *
+     * @throws DomainConstraintException
+     */
+    public function __construct(float $price)
```
sarjon, line 47: Maybe it would be nice to have static factory `Price::fromString()`, wdyt?

sarjon:
I hope that one day this PR will get @eternoendless review. :sun_with_face:

src/Core/Domain/Product/Attachment/Command/AssociateProductAttachmentsCommand.php
```diff
@@ -0,0 +1,89 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Attachment\Command;
```
eternoendless: We should probably avoid nesting commands too deeply. Eg. if a ProductAttachement cannot exist without a Product, maybe their commands belong to Product.

src/Core/Domain/Product/Attachment/Command/AssociateProductAttachmentsCommand.php
```diff
@@ -0,0 +1,89 @@
...
+
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+
+/**
+ * Associates products with attachments.
+ */
+class AssociateProductAttachmentsCommand
```
eternoendless: This represents which user interaction?

tomas862: This will be used when file attachment in product form is being added or removed. In another Pull request there are
 already these commands implemented and the only difference in product form that it "associates" already created attachments with the product id.

src/Core/Domain/Product/Attachment/Command/AssociateProductAttachmentsCommand.php
tomas862: this type is waiting AttachmentId value object to be merged from another PR

src/Core/Domain/Product/Category/Command/UpdateProductCategoriesCommand.php
tomas862: yeah nice naming idea :+1:

src/Core/Domain/Product/Image/Command/UpdateProductImagesCommand.php
tomas862: :o

src/Core/Domain/Product/ValueObject/Condition.php
tomas862: it works but the editor for like 1 second thinks you typed reserved keyword - I'll better change it

src/Core/Domain/ValueObject/Price.php
tomas862: since we don't have strict type checking then this is same as the static factory

`$price = new Price('0.554');`

src/Core/Domain/Product/ValueObject/Condition.php
tomas862: im not good in naming things - maybe `public const NEW_CONDITION = 'new'` ? :D

matks:
@jolelievre and @zuk3975 have continued the great work started here 😉  no need for this PR anymore.

Thank you for your work.

PR #14833, opened 2019-07-23
Bad currency format for Czech language

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Bad currency display when symbol change in latin (for example use Czech language). Following change in https://github.com/PrestaShop/ps_facetedsearch/pull/115
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Try to change specific price.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14833)


admin-dev/themes/new-theme/js/app/cldr/number-formatter.js
```diff
@@ -199,16 +201,29 @@ class NumberFormatter {
...
+    map[PLUS_SIGN_PLACEHOLDER] = symbols.getPlusSign();
+
+    return this.strtr(number, map);
   }
 
+  /**
+   * strtr() for JavaScript
+   * Translate characters or replace substrings
+   */
+  strtr(str, pairs) {
```
matthieu-rolland, line 226: I'm note sure if jsdocs are as mandatory as phpdocs ?

If that's the case, there should be **@param string str**  etc... comment above this function.

PierreRambaud: :+1: 
Because it's a doc block, I don't regenerate assets.

matthieu-rolland (approved):
LGTM

PR #14838, opened 2019-07-24
Update invalid characters in Customer Creation & Update error messages and tips

### LINKED ISSUE #14157: Customer creation error messages and tips to be updated with new invalid characters
Following the addition of new invalid characters in customer creation forms in 1.7.5.2, we have to update the **error messages** and the **tips** in the **customer creation page** in the BO.

Now, when using an invalid character listed in the tip:
Capture d’écran 2019-06-11 à 17 26 09

Now, when using one of the new invalid characters added:
Capture d’écran 2019-06-11 à 17 22 22

[UPDATED 07/17/2109]
1/ The list of invalid characters triggering the error message should be updated. New list: 
0-9!<>,;?=+()@#"°{}_$%:/\\`^*
If a dot ( 。or . ) is not followed by a space, it is considered invalid as well.

2/ The tips under the firstname and lastname fields should be replaced by: 
> Only letters and the dot (. or 。) character are allowed.

3/ New error Message when using an invalid character or a dot with no space:

> The <First name || Last name> field is invalid.
Invalid characters: 0-9!<>,;?=+()@#"°{}_$%:/\\`^*
A space is needed after "." and "。".

(First name or Last name depending on the field with an error)

**Additional context**
For similar issue see [#13640](https://github.com/PrestaShop/PrestaShop/issues/13640)

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | Update invalid characters in Customer Creation & Update error messages and tips |
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14157
| How to test?  | * **In BO**, create a new customer<br>* Fill first/last name with an invalid character (@ for example)<br>* The error message must be displayed under the element and not in top of the form<br><br>* **In FO**, register<br>* Fill first/last name with an invalid character (@ for example)<br>* The error message must be updated

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14838)

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -25,7 +25,7 @@
  */
 use Composer\CaBundle\CaBundle;
 use PrestaShop\PrestaShop\Adapter\ContainerFinder;
-use PrestaShop\PrestaShop\Adapter\ContainerNotFoundException;
+use PrestaShop\PrestaShop\Core\Exception\ContainerNotFoundException;
 use PrestaShop\PrestaShop\Core\Localization\Exception\LocalizationException;
 use PrestaShop\PrestaShop\Core\Localization\Locale\Repository as LocaleRepository;
 use PHPSQLParser\PHPSQLParser;
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -3795,12 +3795,12 @@ exit;
     {
         $context = Context::getContext();
         $containerFinder = new ContainerFinder($context);
-        $container = $containerFinder->getContainer();
-        if (null === $container) {
+        try {
+            $container = $containerFinder->getContainer();
+            $characterCleaner = $container->get('prestashop.core.string.character_cleaner');
+        } catch (Exception $e) {
             // Used in CLI where no container is instancied
             $characterCleaner = new CharacterCleaner();
-        } else {
-            $characterCleaner = $container->get('prestashop.core.string.character_cleaner');
         }
 
         return $characterCleaner->cleanNonUnicodeSupport($pattern);
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -25,6 +25,7 @@
  */
 use Composer\CaBundle\CaBundle;
 use PrestaShop\PrestaShop\Adapter\ContainerFinder;
+use PrestaShop\PrestaShop\Adapter\ContainerNotFoundException;
 use PrestaShop\PrestaShop\Core\Localization\Exception\LocalizationException;
 use PrestaShop\PrestaShop\Core\Localization\Locale\Repository as LocaleRepository;
 use PHPSQLParser\PHPSQLParser;
@@ -3798,7 +3799,7 @@ exit;
         try {
             $container = $containerFinder->getContainer();
             $characterCleaner = $container->get('prestashop.core.string.character_cleaner');
-        } catch (Exception $e) {
+        } catch (ContainerNotFoundException $e) {
             // Used when the container is not generated
             $characterCleaner = new CharacterCleaner();
         }
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -3799,7 +3799,7 @@ exit;
             $container = $containerFinder->getContainer();
             $characterCleaner = $container->get('prestashop.core.string.character_cleaner');
         } catch (Exception $e) {
-            // Used in CLI where no container is instancied
+            // Used when the container is not generated
             $characterCleaner = new CharacterCleaner();
         }
 
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -3782,17 +3782,24 @@ exit;
     /**
      * Delete unicode class from regular expression patterns.
      *
+     * @deprecated Use PrestaShop\PrestaShop\Core\String\CharacterCleaner::cleanNonUnicodeSupport() instead
+     *
      * @param string $pattern
      *
      * @return string pattern
+     *
+     * @throws Exception
      */
     public static function cleanNonUnicodeSupport($pattern)
     {
-        if (!defined('PREG_BAD_UTF8_OFFSET')) {
-            return $pattern;
+        $context = Context::getContext();
+        $container = isset($context->controller) ? $context->controller->getContainer() : null;
+        if (null === $container) {
+            $container = SymfonyContainer::getInstance();
         }
+        $characterCleaner = $container->get('prestashop.core.string.character_cleaner');
 
-        return preg_replace('/\\\[px]\{[a-z]{1,2}\}|(\/[a-z]*)u([a-z]*)$/i', '$1$2', $pattern);
+        return $characterCleaner->cleanNonUnicodeSupport($pattern);
     }
 
     protected static $is_addons_up = true;
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -3793,9 +3793,10 @@ exit;
     public static function cleanNonUnicodeSupport($pattern)
     {
         $context = Context::getContext();
-        $container = isset($context->controller) ? $context->controller->getContainer() : null;
-        if (null === $container) {
-            $container = SymfonyContainer::getInstance();
+        $containerFinder = new ContainerFinder($context);
+        $container = $containerFinder->getContainer();
+        if (null === $context->container) {
+            $context->container = $container;
         }
         $characterCleaner = $container->get('prestashop.core.string.character_cleaner');
 
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -32,6 +32,7 @@ use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\HttpFoundation\Request;
 use PrestaShop\PrestaShop\Core\Foundation\Filesystem\FileSystem as PsFileSystem;
 use PrestaShop\PrestaShop\Core\Localization\Locale;
+use PrestaShop\PrestaShop\Core\String\CharacterCleaner;
 
 class ToolsCore
 {
@@ -3795,10 +3796,12 @@ exit;
         $context = Context::getContext();
         $containerFinder = new ContainerFinder($context);
         $container = $containerFinder->getContainer();
-        if (null === $context->container) {
-            $context->container = $container;
+        if (null === $container) {
+            // Used in CLI where no container is instancied
+            $characterCleaner = new CharacterCleaner();
+        } else {
+            $characterCleaner = $container->get('prestashop.core.string.character_cleaner');
         }
-        $characterCleaner = $container->get('prestashop.core.string.character_cleaner');
 
         return $characterCleaner->cleanNonUnicodeSupport($pattern);
     }
--- a/classes/Validate.php
+++ b/classes/Validate.php
@@ -23,9 +23,9 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
-use PrestaShop\PrestaShop\Adapter\Tools as AdapterTools;
 use PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints\CustomerName;
 use PrestaShop\PrestaShop\Core\ConstraintValidator\Factory\CustomerNameValidatorFactory;
+use PrestaShop\PrestaShop\Core\String\CharacterCleaner;
 use Symfony\Component\Validator\Validation;
 
 class ValidateCore
@@ -172,7 +172,7 @@ class ValidateCore
     {
         $validatorBuilder = Validation::createValidatorBuilder();
         $validatorBuilder->setConstraintValidatorFactory(
-            new CustomerNameValidatorFactory(new AdapterTools())
+            new CustomerNameValidatorFactory(new CharacterCleaner())
         );
         $validator = $validatorBuilder->getValidator();
         $violations = $validator->validate($name, [
--- a/classes/Validate.php
+++ b/classes/Validate.php
@@ -23,6 +23,11 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
+use PrestaShop\PrestaShop\Adapter\Tools as AdapterTools;
+use PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints\CustomerName;
+use PrestaShop\PrestaShop\Core\ConstraintValidator\Factory\CustomerNameValidatorFactory;
+use Symfony\Component\Validator\Validation;
+
 class ValidateCore
 {
     const ADMIN_PASSWORD_LENGTH = 8;
@@ -165,11 +170,16 @@ class ValidateCore
      */
     public static function isCustomerName($name)
     {
-        $validityPattern = Tools::cleanNonUnicodeSupport(
-            '/^(?:[^0-9!<>,;?=+()\/\\@#"°*`{}_^$%:¤\[\]|\.。]|[\.。](?:\s|$))*$/u'
+        $validatorBuilder = Validation::createValidatorBuilder();
+        $validatorBuilder->setConstraintValidatorFactory(
+            new CustomerNameValidatorFactory(new AdapterTools())
         );
+        $validator = $validatorBuilder->getValidator();
+        $violations = $validator->validate($name, [
+            new CustomerName(),
+        ]);
 
-        return preg_match($validityPattern, $name);
+        return (count($violations) !== 0) ? 0 : 1;
     }
 
     /**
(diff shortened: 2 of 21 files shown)


classes/Validate.php
```diff
@@ -166,10 +168,15 @@ public static function isImageSize($size)
     public static function isCustomerName($name)
     {
         $validityPattern = Tools::cleanNonUnicodeSupport(
-            '/^(?:[^0-9!<>,;?=+()\/\\@#"°*`{}_^$%:¤\[\]|\.。]|[\.。](?:\s|$))*$/u'
+            CustomerNameValidator::PATTERN_NAME
         );
 
-        return preg_match($validityPattern, $name);
+        $valid = preg_match($validityPattern, $name);
```
PierreRambaud: Use `CustomerNameValidator::validate` instead :thinking:

src/Core/ConstraintValidator/CustomerNameValidator.php
```diff
@@ -0,0 +1,91 @@
...
+     * @return bool
+     */
+    private function isPointSpacedValid($name)
+    {
+        if (strpos($name, '.') === false && strpos($name, '。') === false) {
+            return true;
+        }
+
+        return (strpos($name, '. ') !== false || strpos($name, '。 ') !== false)
+            && (strpos($name, '.  ') === false && strpos($name, '。  ') === false);
```
PierreRambaud:
```suggestion
            && strpos($name, '.  ') === false && strpos($name, '。  ') === false;
```

src/Core/ConstraintValidator/CustomerNameValidator.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Validates url rewrite according a specific pattern.
+     *
+     * @param string $name
+     *
+     * @return false|int
+     */
+    private function isNameValid($name)
+    {
+        $pattern = preg_replace('/\\\[px]\{[a-z]{1,2}\}|(\/[a-z]*)u([a-z]*)$/i', '$1$2', self::PATTERN_NAME);
```
PierreRambaud: Not a good idea to duplicate cleanNonUnicodeSupport function here

tests/Unit/Core/ConstraintValidator/CustomerNameValidatorTest.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * @dataProvider getInvalidCharacters
+     *
+     * @param string $invalidChar
+     */
+    public function testIfFailsWhenBadCharactersAreGiven($invalidChar)
+    {
+        $input = 'AZE' . $invalidChar . 'RTY';
```
PierreRambaud, line 81: :+1:

src/Core/ConstraintValidator/Constraints/CustomerName.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints;
+
+use PrestaShop\PrestaShop\Core\ConstraintValidator\CustomerNameValidator;
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Class CustomerName is responsible of validating customer name according to several patterns.
+ *
+ * @Annotation
```
PierreRambaud: Is it really an Annotation? I check the code, and we only have one Annotation for IsUrlRewrite, et don't think it's used :/

PierreRambaud (changes requested):
See comments

src/Core/ConstraintValidator/CustomerNameValidator.php
Progi1984: Done

tests/Unit/Core/ConstraintValidator/CustomerNameValidatorTest.php
Progi1984: :smile:

src/Core/ConstraintValidator/Constraints/CustomerName.php
Progi1984: Removed

src/Core/ConstraintValidator/CustomerNameValidator.php
Progi1984: @PierreRambaud Did I created an adapter in PrestaShop\PrestaShop\Adapter\Tools for this method ?

Progi1984: Actually, TypedRegexValidator has ever duplicated the code in cleanNonUnicodeSupport method :(

PierreRambaud: :scream: 
Using an Adapter should be better. Duplicate code is never a good idea :smile_cat: 
ping @matks

Progi1984: So I create an Adapter, and update TypedRegexValidator

classes/Validate.php
Progi1984: Done :blush:

src/Core/ConstraintValidator/CustomerNameValidator.php
Progi1984: Fixed

src/Core/ConstraintValidator/CustomerNameValidator.php
```diff
@@ -0,0 +1,107 @@
...
+     * Check if there is not more one space after point
+     *
+     * @param string $name
+     *
+     * @return bool
+     */
+    private function isPointSpacedValid($name)
+    {
+        $lastCharPos = strlen($name) - 1;
+        if ((strpos($name, '.') === false || strpos($name, '.') === $lastCharPos)
```
PierreRambaud: substr is a little faster and $lastCharPos is now useless
```suggestion
        if ((strpos($name, '.') === false || substr($name, -1) === '.')
```

tests/Unit/Core/ConstraintValidator/CustomerNameValidatorTest.php
```diff
@@ -0,0 +1,111 @@
...
```
PierreRambaud, line 81: What about adding a dot at the end of the string (for testing)?

src/Core/ConstraintValidator/CustomerNameValidator.php
Progi1984: Done

tests/Unit/Core/ConstraintValidator/CustomerNameValidatorTest.php
Progi1984: Done

PierreRambaud (dismissed):

khouloudbelguith:
Hi @Progi1984,

When I try to navigate to the FO
I have this error
image

In the branch 1.7.6.x => OK
https://drive.google.com/file/d/1KUGbm9xktvL-xdjje4f6B0F50hlqv7pW/view

Thanks!

Progi1984:
@khouloudbelguith I just test in local and I haven't got the problem. Have you cleaned your cache ?

khouloudbelguith:
Hi @Progi1984,

Yes, it is a cache issue.
In the BO => ok
But in the FO => I have these errors
image

Thanks!

Progi1984:
@khouloudbelguith It's normal, you need to have a space after your dot character.
image

src/Core/ConstraintValidator/CustomerNameValidator.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    private function isPointSpacedValid($name)
+    {
+        if ((strpos($name, '.') === false || substr($name, -1) === '.')
+            && (strpos($name, '。') === false || substr($name, -1) === '。')) {
+            return true;
+        }
+
+        return (strpos($name, '. ') !== false || strpos($name, '。 ') !== false)
+            && strpos($name, '.  ') === false && strpos($name, '。  ') === false;
```
eternoendless: This is verified by the following regex: `/[\.。][^\s|$]/` (a dot followed by anything other than a space or end-of-string). If it matches, then the string is invalid.

src/Core/ConstraintValidator/TypedRegexValidator.php
```diff
@@ -77,11 +91,11 @@ public function validate($value, Constraint $constraint)
     private function getPattern($type)
     {
         $typePatterns = [
-            'name' => $this->cleanNonUnicodeSupport('/^[^0-9!<>,;?=+()@#"°{}_$%:¤|]*$/u'),
-            'catalog_name' => $this->cleanNonUnicodeSupport('/^[^<>;=#{}]*$/u'),
-            'generic_name' => $this->cleanNonUnicodeSupport('/^[^<>={}]*$/u'),
-            'city_name' => $this->cleanNonUnicodeSupport('/^[^!<>;?=+@#"°{}_$%]*$/u'),
-            'address' => $this->cleanNonUnicodeSupport('/^[^!<>?=+@{}_$%]*$/u'),
+            'name' => $this->tools->cleanNonUnicodeSupport('/^[^0-9!<>,;?=+()@#"°{}_$%:¤|]*$/u'),
```
eternoendless: Consider extracting this method into a class and injecting it instead of adding a dependency on a generic tool class.

khouloudbelguith:
Hi @Progi1984,

About the field names are incorrect
image
image

Thanks!

src/Core/ConstraintValidator/CustomerNameValidator.php
Progi1984: @eternoendless Be careful ! $ between [] is detected as a $ and not end of line

Progi1984: @eternoendless Found the fixed regex : `[\.。](\s{1}([^\ ]|$))`

Progi1984: :ok_hand: Unit tests

PierreRambaud: should be `[\.。]([^\s]|$)` 🤔

LouiseBonnard:
Hi there, just to be sure we all go in the same direction, here are the final wordings to apply:

- first information level is the tip (grey text right under the field): **Only letters and the dot (.) character, followed by a space, are allowed.** to localize in _Admin.Orderscustomers.Help_ when applied to the BO / in _Shop.Forms.Help_ when applied to the FO

- second information level is the error notif (when the input is invalid despite the tip): **The %s field is invalid.** to localize in _Admin.Notifications.Error_ when applied to the BO / **Invalid format.** to localize in _Shop.Forms.Errors_ when applied to the FO

@Progi1984, @samuel-pires, is it okay for you guys?

Progi1984:
@LouiseBonnard Wording fixed in last commit

src/Core/ConstraintValidator/TypedRegexValidator.php
Progi1984: @eternoendless Done

LouiseBonnard:
@Progi1984, thanks! It only regards the FO here, right?

Progi1984:
@LouiseBonnard Because the BO was ever good :)

classes/Tools.php
```diff
@@ -3782,17 +3782,25 @@ public static function arrayUnique($array)
...
+     * @throws Exception
      */
     public static function cleanNonUnicodeSupport($pattern)
     {
-        if (!defined('PREG_BAD_UTF8_OFFSET')) {
-            return $pattern;
+        $context = Context::getContext();
+        $containerFinder = new ContainerFinder($context);
+        $container = $containerFinder->getContainer();
+        if (null === $context->container) {
```
jolelievre: is it useful to assign `$context->container` here? there's no need for it in this method
if the aim is to optimize other calls of ContainerFinder, then maybe it should be made in ContainerFinder, although I don't think ti would bring much gain

classes/Validate.php
```diff
@@ -165,11 +170,16 @@ public static function isImageSize($size)
...
+        $validatorBuilder->setConstraintValidatorFactory(
+            new CustomerNameValidatorFactory(new CharacterCleaner())
         );
+        $validator = $validatorBuilder->getValidator();
+        $violations = $validator->validate($name, [
+            new CustomerName(),
+        ]);
 
-        return preg_match($validityPattern, $name);
+        return (count($violations) !== 0) ? 0 : 1;
```
PierreRambaud, line 182: `is` methods must return boolean.
```suggestion
        return count($violations) !== 0;
```

Progi1984: Yeap, but it is in the legacy and the previous behavior was 0 or 1 and false or true. So I don't change this behavior.

PierreRambaud: It sucks! :sweat_smile:

Progi1984: It's the legacy :)

classes/Tools.php
```diff
@@ -3782,17 +3783,27 @@ public static function arrayUnique($array)
...
     public static function cleanNonUnicodeSupport($pattern)
     {
-        if (!defined('PREG_BAD_UTF8_OFFSET')) {
-            return $pattern;
+        $context = Context::getContext();
+        $containerFinder = new ContainerFinder($context);
+        try {
+            $container = $containerFinder->getContainer();
+            $characterCleaner = $container->get('prestashop.core.string.character_cleaner');
+        } catch (Exception $e) {
```
PierreRambaud: can we catch something more explicit?

PierreRambaud: Like a ServiceNotFoundException? dunno if Symfony raise something like this

src/Core/ConstraintValidator/CustomerNameValidator.php
```diff
@@ -0,0 +1,106 @@
...
+            ;
+        }
+    }
+
+    /**
+     * Validates url rewrite according a specific pattern.
+     *
+     * @param string $name
+     *
+     * @return false|int
```
PierreRambaud: Since it's not a legacy code, can we return bool for these functions?
If not, the second one has a bad `@return` docblock

Progi1984: :ok_hand:

classes/Tools.php
Progi1984: :ok_hand:

PierreRambaud (approved):

matks:
Thank you @Progi1984

thbl:
Is it possible to add this fix in 1.7.6.1 manually some how ?
as this is a big issue for us

thbl:
also should "/" not be a valid character ? 
it was valid before in our case we have a lot of customers using this

ghost:
Hi @thbl,
I would recommend waiting for the 1.7.7 to be released to get this update on your shop.
Until then you can see our recommendations on how to limit spamming attacks on our tech blog: http://build.prestashop.com/news/fighting-against-spamming-again/

About the "/" character, in which situation is it used by your customers? can you tell me more, please?

thbl:
@samuel-pires some (many in our case) use their company name as first / last name.
As multiple people use the same account.
In Denmark a lot of companies has "A/S" in their name, this definds the company type and are a part of the name

ghost:
> @samuel-pires some (many in our case) use their company name as first / last name.
> As multiple people use the same account.
> In Denmark a lot of companies has "A/S" in their name, this definds the company type and are a part of the name

Thanks for the feedback @thbl!
Do you have only B2B clients or individuals as well?
Do you use the B2B mode? it adds a "company" field in the customer creation form (see image) which accepts the "/" character.
Capture d’écran 2019-09-11 à 12 41 23


@LouiseBonnard wdyt?

thbl:
@samuel-pires  Have both and yes have B2B enable.
I know about the company field but customers tendens to use first / last name any ways.

thbl:
@LouiseBonnard  @samuel-pires  any plans for this ?

LouiseBonnard:
Hi @thbl, thanks for the feedback! In my opinion, nothing needs to be changed here: S/A regards the company name and the "/" character is valid for this field. :-)

cc @samuel-pires

thbl:
@LouiseBonnard  as said many customers uses first / last name for company name tho there is a field for it. There for "/" should also be allowed in name as it was in 1.7.5

LouiseBonnard:
@thbl, thanks for your reply. My point was that they are not supposed to do so, it's a misuse... so let's offer them the chance to use the right field!

demperador:
Hello!

I am having also this issue under version 1.7.7.2 and PHP 7.2.34
I do not have any clue where the issue is coming from. In fact, I cannot even update any of my existing clients.  I have queried within the database for any invalid characters but I have not seen anything at all. Can anyone give me a clue?

Thanks and regards.
Captura de pantalla 2021-02-22 a las 17 26 44

demperador:
I have also tried creating a new account on Front Office while debugging and the following exception is thrown:
Captura de pantalla 2021-02-22 a las 17 44 30

PR #14845, opened 2019-07-24
Fix use of two translators in the back office

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Read below
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | TBD
| How to test?  | TBD

This PR requires this change to be applied and imported: https://github.com/PrestaShop/TranslationToolsBundle/pull/64

This change fixes three problems:

1. The test module `translation_test` has been renamed to `translationtest`, since translations cannot work on modules with underscores in their name. Indeed, the domain name for `translation_test` would be `Modules.Translationtest`, which can never be reversed back to `translation_test` to find the original module.
2. The third component of a translation domain, in the case of modules, is the original file name. Many file names contain underscore, but it wasn't supported because the translation interface in BO used Inflector::tableize to separate the domain parts (yes domains are stored without dots in the translator dictionary, but that's another story). This has been fixed in [this commit](https://github.com/PrestaShop/PrestaShop/commit/f02b16767d8689e0aba093becef516f9b7c1c41c), so domains with underscores in the third part are now allowed, and therefore wordings placed in files whose name contains underscores are now translatable.
3. There are two translators used a the same time in the BO, and that causes problems.

Regarding item no. 3:

- One translator is created by `Context->getTranslatorFromLocale()`: `PrestaShopBundle\Translation\TranslatorComponent`
- The other is instantiated by Symfony's FrameworkBundle: `PrestaShopBundle\Translation\Translator`, via the `translator.class` setting in `app/config/config.yml` and `PrestaShopBundle\DependencyInjection\Compiler\OverrideTranslatorServiceCompilerPass` which enforces it.

Without [this change](https://github.com/PrestaShop/PrestaShop/pull/14845/commits/5d43be3d34c57adf2a5f35816f1731d635a7f8a2), when using the translationtest module (tests/Resources/modules/translationtest), translations customized through the BO translation interface would be ignored. 

Why? Because twig templates use the translator instantiated by FrameworkBundle (that is, `PrestaShopBundle\Translation\Translator`). This component was set up so that it adds a db resource for any loaded resource, so that custom translations saved in the DB can be used. But what domains are loaded? FrameworkBundle will automatically look for xliff files in default and configured locations and will load them as resources (one file = one domain). We have many of them in `app/Resources/translations/`, including files for native modules. But why doesn't it work for third party modules? You guessed it, because **there are no xliff files for third party modules there** (nor anywhere for that matter, but that's a different subject). So no xliff file, no resource loaded, no customization possible.

So this change replaces the FrameworkBundle translator with the one created by Context. One side effect of this is that the translation element in the Symfony profiler no longer appears. My guess is that this happens because its DataCollector (which we used to replace using `OverrideTranslatorServiceCompilerPass`) is instantiated by FrameworkBundle using the original translator, and not our replacement created by Context. Since the original translator is never used (because we replace its reference with the one created by Context), no translation is tracked and therefore it no longer appears in the profiler. I tried to replace this DataCollector using decoration as well but for the time being I haven't been able to make it work.

I'm not terribly happy with this solution (making Symfony depend on Context), but I can't find any other way. Another side effect is that the Context translator isn't warmable whereas the symfony one is, so we are losing that as well.

## After checking out this change

Make sure to clear your local cache using:

```
bin/console c:c --no-warmup
```

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14845)


kpodemski:
i'm literally lost with all these changes to translation system, i just hope that one day everything will work just fine... 😆 

i'm waiting for docs to test everything 👍 

thank you for work tho, it's really important to make PrestaShop translation system stable!

eternoendless:
Docs are coming! I actually found this problem while writing documentation about it. In the end, there should be one and only one translation system, with only one translator...

eternoendless:
This PR is going to be split in two, one with the test resources and the other that handles two parallel translators.

Issue #14179, opened 2019-06-12
[SF Migration] Update and apply bulk actions strategy for failures


matks:
### Question

When performing one bulk action, when one of the actions fail, should we
**1-)** stop there and display the error message to user
or
**2-)** skip it, attempt to perform the action on the remaining items, and at the end display a list of the failed usecases ?

So far we have followed the strategy 1) because one error might be a hint of something wrong happening, so it's good to wait for user to re-try to make sure the bulk action was not triggered by mistake.

But @marionf thinks otherwise 😉(https://github.com/PrestaShop/PrestaShop/pull/13988#discussion_r292365506)

We should review each list with bulk actions, ask Product Team whether the strategy is suitable and apply it when the answer is "yes"

marionf:
@matks How it was in legacy ?

PR #14864, opened 2019-07-26
Migrate Catalog > Files page

RaimondasSapola:
| Questions     | Answers
| Branch?       | develop
| Description?  | Catalog > Files and Catalog > Files
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10550
| How to test?  | Go to /admin-dev/index.php/sell/attachments. The page should work as previously.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14864)


src/Core/Domain/Attachment/Command/BulkDeleteAttachmentsCommand.php
```diff
@@ -0,0 +1,71 @@
...
+     * @var AttachmentId[]
+     */
+    private $attachmentIds;
+
+    /**
+     * @param int[] $attachmentIds
+     *
+     * @throws AttachmentConstraintException
+     */
+    public function __construct(array $attachmentIds)
```
rokaszygmantas: we can use return type hints :slightly_smiling_face:

src/Core/Domain/Attachment/Command/BulkDeleteAttachmentsCommand.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        $this->setAttachmentIds($attachmentIds);
+    }
+
+    /**
+     * @return AttachmentId[]
+     */
+    public function getAttachmentIds()
```
rokaszygmantas: we can use return type hints :slightly_smiling_face:

src/Core/Domain/Attachment/Command/DeleteAttachmentCommand.php
```diff
@@ -0,0 +1,59 @@
...
+     * @var AttachmentId
+     */
+    private $attachmentId;
+
+    /**
+     * @param int $attachmentId
+     *
+     * @throws AttachmentConstraintException
+     */
+    public function __construct($attachmentId)
```
rokaszygmantas: scalar type hints can be used everywhere in method signatures

src/Core/Domain/Attachment/Query/GetAttachment.php
```diff
@@ -0,0 +1,58 @@
...
+{
+    /**
+     * @var AttachmentId
+     */
+    private $attachmentId;
+
+    /**
+     * @param $attachmentId
+     *
+     * @throws \PrestaShop\PrestaShop\Core\Domain\Attachment\Exception\AttachmentConstraintException
```
rokaszygmantas: class should be imported

src/Core/Domain/Attachment/QueryHandler/GetAttachmentPathHandlerInterface.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace PrestaShop\PrestaShop\Core\Domain\Attachment\QueryHandler;
+
+use PrestaShop\PrestaShop\Core\Domain\Attachment\Query\GetAttachment;
+use PrestaShop\PrestaShop\Core\Domain\Attachment\QueryResult\Attachment;
+
+/**
+ * Defines contract for GetAttachmentPathHandler
+ */
+interface GetAttachmentPathHandlerInterface
```
rokaszygmantas: since you are handling `GetAttachment` query, maybe the handler should be named `GetAttachmentHandler` as well, for consistency

src/Core/Grid/Action/Row/Type/SubmitRowAction.php
```diff
@@ -35,6 +35,9 @@
  */
 final class SubmitRowAction extends AbstractRowAction
 {
+    const MESSAGE_TYPE_STATIC = 'static';
```
rokaszygmantas: some comments would be helpful, explaining what these constants are used for

src/Core/Grid/Data/Factory/AttachmentGridDataFactoryDecorator.php
```diff
@@ -0,0 +1,158 @@
...
+     * @return RecordCollection
+     */
+    private function applyModifications(RecordCollectionInterface $attachments)
+    {
+        $modifiedAttachments = [];
+
+        foreach ($attachments as $attachment) {
+            if ((int) $attachment['products'] > 0) {
+                $productNamesArray = $this->getProductNames($attachment['id_attachment']);
+                $productNames = isset($productNamesArray['product_names']) ? $productNamesArray['product_names'] : '';
```
rokaszygmantas: you can use null coalescing operator for cleaner conditions:
```suggestion
                $productNames = $productNamesArray['product_names'] ?? ''; 
```

src/Core/Grid/Definition/Factory/AttachmentGridDefinitionFactory.php
```diff
@@ -0,0 +1,300 @@
...
+use PrestaShop\PrestaShop\Core\Grid\Column\Type\DataColumn;
+use PrestaShop\PrestaShop\Core\Grid\Filter\Filter;
+use PrestaShop\PrestaShop\Core\Grid\Filter\FilterCollection;
+use PrestaShop\PrestaShop\Core\Hook\HookDispatcherInterface;
+use PrestaShop\PrestaShop\Core\Multistore\MultistoreContextCheckerInterface;
+use PrestaShopBundle\Form\Admin\Type\SearchAndResetType;
+use Symfony\Component\Form\Extension\Core\Type\NumberType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+
+final class AttachmentGridDefinitionFactory extends AbstractGridDefinitionFactory
```
rokaszygmantas, line 50: missing docblock

src/Core/Grid/Query/AttachmentQueryBuilder.php
```diff
@@ -0,0 +1,174 @@
...
+     * @param Connection $connection
+     * @param $dbPrefix
+     * @param DoctrineSearchCriteriaApplicatorInterface $searchCriteriaApplicator
+     * @param $employeeIdLang
+     */
+    public function __construct(
+        Connection $connection,
+        $dbPrefix,
+        DoctrineSearchCriteriaApplicatorInterface $searchCriteriaApplicator,
+        $employeeIdLang
```
rokaszygmantas: missing type hint

src/Core/Grid/Query/AttachmentQueryBuilder.php
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getSearchQueryBuilder(SearchCriteriaInterface $searchCriteria)
+    {
+        $qb = $this->getQueryBuilder($searchCriteria->getFilters());
+
+        $qb
+            ->select('a.`id_attachment`, al.`name`, a.`file`, a.`file_size`')
+            ->addSelect('CONCAT(COALESCE(virtual_product_attachment.`product_count`, 0), " product(s)") AS products')
```
rokaszygmantas: `product(s)` - this part is not translatable :thinking: was it like this in the legacy code?

src/Core/Util/File/FileSizeConverter.php
```diff
@@ -0,0 +1,63 @@
...
+
+        if ($bytes >= 1024) {
+            return number_format($bytes / 1024, 2) . 'k';
+        }
+
+        if ($bytes > 1) {
+            return $bytes . 'b';
+        }
+
+        if ($bytes == 1) {
```
rokaszygmantas: this condition and the one above can be joined together

src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Actions/Row/submit.html.twig
```diff
@@ -29,10 +29,16 @@
   {% set class = class ~ ' ' ~ attributes.class %}
 {% endif %}
 
+{% set confirmation_message = action.options.confirm_message %}
+{% if action.options.confirm_message_type == 'dynamic' and record[action.options.dynamic_message_field] is defined %}
+  {% set dynamic_field = record[action.options.dynamic_message_field] %}
+  {% set confirmation_message = dynamic_field is defined ? dynamic_field : action.options.confirm_message %}
```
rokaszygmantas:
```suggestion
  {% set confirmation_message = dynamic_field|default(action.options.confirm_message) %}
```

src/Adapter/Attachment/CommandHandler/BulkDeleteAttachmentsHandler.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(BulkDeleteAttachmentsCommand $command)
+    {
+        foreach ($command->getAttachmentIds() as $attachmentId) {
+            $attachment = $this->getAttachment($attachmentId);
+
+            if (!$this->deleteAttachment($attachment)) {
+                throw new DeleteAttachmentException(
```
PierreRambaud: What if only the last one failed?
The response will return an error while everything instead of one id weren't deleted? :thinking:

src/Core/Domain/Attachment/Query/GetAttachment.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function __construct($attachmentId)
+    {
+        $this->attachmentId = new AttachmentId($attachmentId);
+    }
+
+    /**
+     * @return AttachmentId
+     */
+    public function getAttachmentId()
```
PierreRambaud:
```suggestion
    public function getAttachmentId(): AttachmentId
```:

File: src/Core/Domain/Attachment/ValueObject/AttachmentId.php [hunk 15]
```diff
@@ -0,0 +1,76 @@
...
+    public function __construct($id)
+    {
+        $this->assertIsIntegerGreaterThanZero($id);
+        $this->id = $id;
+    }
+
+    /**
+     * @return int
+     */
+    public function getValue()
```
PierreRambaud:
```suggestion
    public function getValue(): int
```

src/Core/Grid/Data/Factory/AttachmentGridDataFactoryDecorator.php
```diff
@@ -0,0 +1,158 @@
...
+                $attachment['dynamic_message'] = $this->trans(
+                    "This file is associated with the following products, do you really want to  delete it?\r\n\r\n%products%",
```
PierreRambaud: `\r\n\r\n` :scream:

src/Core/Util/File/FileSizeConverter.php
PierreRambaud: This one look useless yes.

src/Core/Util/File/FileSizeConverter.php
```diff
@@ -0,0 +1,63 @@
...
+     * @return string
+     */
+    public function convert(int $bytes)
+    {
+        if ($bytes >= 1073741824) {
+            return number_format($bytes / 1073741824, 2) . 'GB';
+        }
+
+        if ($bytes >= 1048576) {
+            return number_format($bytes / 1048576, 2) . 'M';
```
PierreRambaud:
```suggestion
            return number_format($bytes / 1048576, 2) . 'MB';
```

src/Core/Util/File/FileSizeConverter.php
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        if ($bytes >= 1024) {
+            return number_format($bytes / 1024, 2) . 'k';
```
PierreRambaud:
```suggestion
            return number_format($bytes / 1024, 2) . 'kB';
```

src/Core/Util/File/FileSizeConverter.php
```diff
@@ -0,0 +1,63 @@
...
+ * Class FileSizeConverter converts int value to formatted file size value
+ */
+final class FileSizeConverter
+{
+    /**
+     * @param int $bytes
+     *
+     * @return string
+     */
+    public function convert(int $bytes)
```
PierreRambaud: You can do something easier than using many integer:
```php
public function convert(int $bytes, int $precision = 2): string
{ 
    $units = array('B', 'KB', 'MB', 'GB', 'TB'); 
    $bytes = max($bytes, 0); 
    $pow = floor(($bytes ? log($bytes) : 0) / log(1024)); 
    $pow = min($pow, count($units) - 1); 

    return round($bytes, $precision) . ' ' . $units[$pow]; 
} 
```

src/Core/Util/File/FileSizeConverter.php
```diff
@@ -0,0 +1,63 @@
...
+        if ($bytes >= 1048576) {
+            return number_format($bytes / 1048576, 2) . 'M';
+        }
+
+        if ($bytes >= 1024) {
+            return number_format($bytes / 1024, 2) . 'k';
+        }
+
+        if ($bytes > 1) {
+            return $bytes . 'b';
```
PierreRambaud:
```suggestion
            return $bytes . 'B';
```

src/PrestaShopBundle/Resources/config/routing/admin/sell/catalog/attachment.yml
```diff
@@ -0,0 +1,69 @@
+admin_attachments_index:
+  path: /
+  methods: GET
+  defaults:
+    _controller: PrestaShopBundle:Admin/Sell/Catalog/Attachment:index
+    _legacy_controller: AdminAttachments
+#    _legacy_link: AdminAttachments
```
PierreRambaud, line 7: Commented?

tests/Unit/Core/Util/File/FileSizeConverterTest.php
```diff
@@ -0,0 +1,67 @@
...
+
+        $this->assertEquals($expectedResult, $result);
+    }
+
+    public function sizeDataProvider()
+    {
+        return [
+            [0, '0b'],
+            [1, '1b'],
+            [1024, '1.00k'],
```
PierreRambaud: Test with many other values please.

```
-0
-123456789123
-15521
1651361681351684
21321635843254
2132112
5954
```

src/Core/Grid/Data/Factory/AttachmentGridDataFactoryDecorator.php
LouiseBonnard: Hey there, do you mean:

_This file is associated with the following %products%, do you really want to delete it?_ in _Admin.Catalog.Notification_

We could also write _This file is associated with the following %products%. Are you sure you want to delete it?_, up to you. ;-)

src/Core/Grid/Definition/Factory/AttachmentGridDefinitionFactory.php
```diff
@@ -0,0 +1,300 @@
...
+    {
+        return 'attachment';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getName()
+    {
+        return $this->trans('Files', [], 'Admin.Navigation.Feature');
```
LouiseBonnard: I think it is _Admin.Navigation.Menu_.

src/PrestaShopBundle/Resources/config/routing/admin/sell/catalog/attachment.yml
rokaszygmantas: probably because the page is not visible yet, so the legacy links are not updated.

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/AttachmentController.php
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * @return array
+     */
+    private function getAttachmentToolbarButtons()
+    {
+        $toolbarButtons = [];
+
+        $toolbarButtons['add'] = [
+            'href' => $this->generateUrl('admin_attachment_create'),
+            'desc' => $this->trans('Add new Attachment', 'Admin.Catalog.Feature'),
```
LouiseBonnard: It should rather be _Add new file_ instead, shouldn't it?

src/Core/Grid/Data/Factory/AttachmentGridDataFactoryDecorator.php
```diff
@@ -0,0 +1,160 @@
...
+    private function applyModifications(RecordCollectionInterface $attachments)
+    {
+        $modifiedAttachments = [];
+
+        foreach ($attachments as $attachment) {
+            if ((int) $attachment['products'] > 0) {
+                $productNamesArray = $this->getProductNames($attachment['id_attachment']);
+                $productNames = $productNamesArray['product_names'] ?? '';
+                $attachment['dynamic_message'] = $this->trans(
+                    'This file is associated with the following products, do you really want to  delete it?',
```
LouiseBonnard: It seems that there is a useless space in the original string, but I'm not fond of letting it as-is because it is a mistake. In order to bypass the typo, I would suggest writing the following new wording _This file is associated with the following products. Are you sure you want to delete it?_, what do you think?

RaimondasSapola: Sound good to me

src/Adapter/Attachment/CommandHandler/BulkDeleteAttachmentsHandler.php
RaimondasSapola: I see there is an unresolved discussion about this topic https://github.com/PrestaShop/PrestaShop/issues/14179

src/Adapter/Attachment/AbstractAttachmentHandler.php
```diff
@@ -0,0 +1,86 @@
...
+abstract class AbstractAttachmentHandler
+{
+    /**
+     * @param AttachmentId $attachmentId
+     *
+     * @return Attachment
+     *
+     * @throws AttachmentNotFoundException
+     */
+    protected function getAttachment(AttachmentId $attachmentId): Attachment
```
matks, line 47: I ❤️ the php7 style

src/Adapter/Attachment/CommandHandler/BulkDeleteAttachmentsHandler.php
matks: Indeed 😉 nice finding it !

So we should follow strategy `skip it, attempt to perform the action on the remaining items, and at the end display a list of the failed usecases` unless we think it's not relevant, in this case we ask Product Team (@marionf for example)

src/Adapter/Attachment/QueryHandler/GetAttachmentHandler.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(GetAttachment $query): Attachment
+    {
+        $attachment = $this->getAttachment($query->getAttachmentId());
+
+        return new Attachment(
+            _PS_DOWNLOAD_DIR_ . $attachment->file,
```
matks: Can we add a `file_exists()` check here ?

I think it's useless to return a `Attachment` object if the filepath is wrong 🤔 although I'm not sure. What do you think ?

Or maybe we can check this at the controller level, and display a warning ? Wdyt @sarjon ?

src/Core/Grid/Data/Factory/AttachmentGridDataFactoryDecorator.php
```diff
@@ -0,0 +1,160 @@
...
+        foreach ($attachments as $attachment) {
+            if ((int) $attachment['products'] > 0) {
+                $productNamesArray = $this->getProductNames($attachment['id_attachment']);
+                $productNames = $productNamesArray['product_names'] ?? '';
+                $attachment['dynamic_message'] = $this->trans(
+                    'This file is associated with the following products. Are you sure you want to delete it?',
+                    [],
+                    'Admin.Catalog.Notification'
+                );
+                $attachment['dynamic_message'] .= "\n\n" . $productNames;
```
matks:
```suggestion
                $attachment['dynamic_message'] .= PHP_EOL . PHP_EOL . $productNames;
```
Dont forget Windows users 😄

src/Core/Grid/Definition/Factory/AttachmentGridDefinitionFactory.php
```diff
@@ -0,0 +1,303 @@
...
+                        ],
+                        'required' => false,
+                    ])
+                    ->setAssociatedColumn('file_size')
+            )
+            ->add(
+                (new Filter('products', NumberType::class))
+                    ->setTypeOptions([
+                        'attr' => [
+                            'placeholder' => $this->trans('Search file size', [], 'Admin.Actions'),
```
matks: Copy-paste 😉 this is not `Search file size`

src/Core/Grid/Query/AttachmentQueryBuilder.php
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Grid\Query;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Query\QueryBuilder;
+use PrestaShop\PrestaShop\Core\Grid\Search\SearchCriteriaInterface;
+
+/**
+ * Attachment query builder builds search & count queries for attachment grid.
```
matks:
```suggestion
 * Attachment query builder builds search & count data for attachment grid.
```

src/Core/Util/File/FileSizeConverter.php
```diff
@@ -0,0 +1,54 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Util\File;
+
+/**
+ * Converts int value to formatted file size value
+ */
+final class FileSizeConverter
```
matks: Why `final` ? We could create other Converters with a different output 😃

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/AttachmentController.php
```diff
@@ -0,0 +1,250 @@
...
+     *
+     * @return Response
+     */
+    public function viewAction(int $attachmentId): Response
+    {
+        try {
+            /** @var Attachment $attachment */
+            $attachment = $this->getCommandBus()->handle(new GetAttachment((int) $attachmentId));
+
+            return $this->file($attachment->getPath(), $attachment->getName());
```
matks, line 128: If we dont add a `file_exists()` is the QueryHandler, then let's check here. We cannot return `return $this->file` if filepath is bad.

It could even be used for a [LFI](https://www.owasp.org/index.php/Testing_for_Local_File_Inclusion) 😉 

@PierreRambaud how far, in your opinion, should we check that this file is in a valid area (to avoid using this controller to display `/etc/passwd` for example) ? Can we consider than "if file is contained inside the root folder of the prestashop app" is enough* ? 

*although that means if someones customizes the folder structure of the shop and extracts user-content outside of the root folder, this will block him, he'll need to customize this behavior

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/AttachmentController.php
```diff
@@ -0,0 +1,250 @@
...
+        return $this->redirectToRoute('admin_attachments_index');
+    }
+
+    /**
+     * Deletes attachment
+     *
+     * @AdminSecurity("is_granted('delete', request.get('_legacy_controller'))", redirectRoute="admin_attachments_index")
+     * @DemoRestricted(redirectRoute="admin_attachments_index")
+     *
+     * @param $attachmentId
```
matks:
```suggestion
     * @param int $attachmentId
```

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/AttachmentController.php
```diff
@@ -0,0 +1,250 @@
...
+     */
+    private function getBulkAttachmentsFromRequest(Request $request): array
+    {
+        $attachmentIds = $request->request->get('attachment_files_bulk');
+
+        if (!is_array($attachmentIds)) {
+            return [];
+        }
+
+        foreach ($attachmentIds as $i => $attachmentId) {
```
matks, line 237: you can do it with an `array_map()` 😉 
but that will not save you a lot of space, so it's fine like this too 😄

tests/Unit/Core/Util/File/FileSizeConverterTest.php
```diff
@@ -0,0 +1,74 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace Tests\Unit\Core\Util\File;
+
+use PHPUnit\Framework\TestCase;
+use PrestaShop\PrestaShop\Core\Util\File\FileSizeConverter;
+
+class FileSizeConverterTest extends TestCase
```
matks, line 32: 👍

src/Core/Grid/Data/Factory/AttachmentGridDataFactoryDecorator.php
matks: OK now I understand why you introduce the `dynamic_message_field` property for columns ... 🤔 

It works 👍 however we'll need to document this behavior when this PR gets merged

matks (changes requested):
@RaimondasSapola Nice job ! 😄 I've got a few feedbacks to handle but the work is overall very good ✅

RaimondasSapola:
Fixed CR issues including bulk delete behavior and file_exist check. @matks could you take a second look ? :)

src/Adapter/Attachment/QueryHandler/GetAttachmentHandler.php
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @throws AttachmentNotFoundException
+     */
+    public function handle(GetAttachment $query): Attachment
+    {
+        $attachment = $this->getAttachment($query->getAttachmentId());
+        $path = _PS_DOWNLOAD_DIR_ . $attachment->file;
+
+        if (!file_exists($path)) {
+            throw new AttachmentNotFoundException('Attachment file was not found');
```
matks:
```suggestion
            throw new AttachmentNotFoundException(sprintf('Attachment file was not found at %s', $path));
```

Exceptions are meant to be read by developers and consequently it's best to provide as much as possible info to them. If file does not exists, developer will need the `$path` to understand where it's supposed to be. For example if `_PS_DOWNLOAD_DIR_` is missing a slash, he'll see it immediately this way 😉

matks:
@RaimondasSapola sorry for the late answer, only 1 thing to improve and we're done !
=> https://github.com/PrestaShop/PrestaShop/pull/14864#discussion_r323245492

src/Adapter/Attachment/CommandHandler/BulkDeleteAttachmentsHandler.php
```diff
@@ -0,0 +1,65 @@
...
+                if (!$this->deleteAttachment($attachment)) {
+                    $errors[] = $attachment->id;
+                }
+            } catch (AttachmentException $e) {
+                $errors[] = $attachmentId->getValue();
+            }
+        }
+
+        if (!empty($errors)) {
+            throw new BulkDeleteAttachmentsException(implode(', ', $errors));
```
PierreRambaud: no message here?

src/Adapter/Attachment/QueryHandler/GetAttachmentHandler.php
```diff
@@ -0,0 +1,59 @@
...
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @throws AttachmentNotFoundException
+     */
+    public function handle(GetAttachment $query): Attachment
+    {
+        $attachment = $this->getAttachment($query->getAttachmentId());
+        $path = _PS_DOWNLOAD_DIR_ . $attachment->file;
```
PierreRambaud: Because one day const will be removed, can we inject the path in this class?

src/Core/Grid/Data/Factory/AttachmentGridDataFactoryDecorator.php
```diff
@@ -0,0 +1,160 @@
...
+     * @param RecordCollectionInterface $attachments
+     *
+     * @return RecordCollection
+     */
+    private function applyModifications(RecordCollectionInterface $attachments): RecordCollection
+    {
+        $modifiedAttachments = [];
+
+        foreach ($attachments as $attachment) {
+            if ((int) $attachment['products'] > 0) {
```
PierreRambaud, line 117: Count missing?
Are we returning products, or productIds?

PierreRambaud: (Because of `products_names` below

src/Core/Grid/Data/Factory/AttachmentGridDataFactoryDecorator.php
```diff
@@ -0,0 +1,160 @@
...
+                $this->dbPrefix . 'product_lang',
+                'pl',
+                'pa.`id_product` = pl.`id_product` AND pl.`id_lang` = :langId'
+            )
+            ->where('pa.`id_attachment` = :attachmentId')
+            ->setParameter('attachmentId', $attachmentId)
+            ->setParameter('langId', $this->employeeIdLang)
+            ->setMaxResults(1);
+
+        return $qb->execute()->fetch();
```
PierreRambaud:
```suggestion
        return $qb->getSingleResult();
```

ANd you can remove the setMaxResults.

But the function name is awkward, you return a string containing all names.
Why not returning an array[] of names instead :thinking: ?

RaimondasSapola: There is no getSingleResult method in QueryBuilder but i agree about returning array so $qb->execute()->fetchAll(PDO::FETCH_COLUMN); will suffice i guess :)

src/Core/Grid/Data/Factory/AttachmentGridDataFactoryDecorator.php
RaimondasSapola: products in this case is integer value of products count returned from query builder and not the result of product_names query :)

matks (dismissed):

matks (dismissed):

matks (dismissed):

Robin-Fischer-PS:
Hi @RaimondasSapola , thanks for the PR. It's all good for me :)

matks (dismissed):

src/Adapter/Attachment/CommandHandler/BulkDeleteAttachmentsHandler.php
PierreRambaud: What was decided for this behavior?

src/Core/Util/File/FileSizeConverter.php
```diff
@@ -0,0 +1,54 @@
...
+    public function convert(int $bytes, int $precision = 2): string
+    {
+        $units = ['B', 'kB', 'MB', 'GB', 'TB'];
+        $bytes = max($bytes, 0);
+        $pow = floor(($bytes ? log($bytes) : 0) / log(1024));
+        $pow = min($pow, count($units) - 1);
+
+        if ($bytes >= 1024) {
+            $bytes /= pow(1024, $pow);
+            $bytes = number_format(round($bytes, $precision), 2, '.', '');
```
PierreRambaud, line 49: The number should be formatting depending on cldr implementation? ping @matks

matks: This is a FileSizeConverter 🤔 do you think file size formats differ from one country to another ?

PierreRambaud: The file format no, but the number format yes. It's the same for liter, meters, kilograms, etc. :) Maybe we should take in consideration this for later. Let's see with @eternoendless

matks (approved):

PR #14899, opened 2019-07-29
Make the container available in WebService environment

### LINKED ISSUE #14842: Dosen't send email on state change using WS since 1.7.6
After updating to 1.7.6...
When changing ordre "current_state" using webservice, emails is not sent tho ordre state changes.

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | Initialize container in WebService environment
| Type?         | bug fix
| Category?     | WS
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | Fixes #14842 
| How to test?  | * Try the script given in issue #14842#issuecomment-514593911

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14899)

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -792,8 +792,11 @@ class ToolsCore
             return $locale;
         }
 
-        $containerFinder = new ContainerFinder();
-        $container = $containerFinder->getContainer($context);
+        $containerFinder = new ContainerFinder($context);
+        $container = $containerFinder->getContainer();
+        if (null === $context->container) {
+            $context->container = $container;
+        }
 
         /** @var LocaleRepository $localeRepository */
         $localeRepository = $container->get(self::SERVICE_LOCALE_REPOSITORY);
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -24,7 +24,7 @@
  * International Registered Trademark & Property of PrestaShop SA
  */
 use Composer\CaBundle\CaBundle;
-use PrestaShop\PrestaShop\Adapter\SymfonyContainer;
+use PrestaShop\PrestaShop\Adapter\ContainerFinder;
 use PrestaShop\PrestaShop\Core\Localization\Exception\LocalizationException;
 use PrestaShop\PrestaShop\Core\Localization\Locale\Repository as LocaleRepository;
 use PHPSQLParser\PHPSQLParser;
@@ -792,10 +792,8 @@ class ToolsCore
             return $locale;
         }
 
-        $container = isset($context->controller) ? $context->controller->getContainer() : null;
-        if (null === $container) {
-            $container = SymfonyContainer::getInstance();
-        }
+        $containerFinder = new ContainerFinder();
+        $container = $containerFinder->getContainer($context);
 
         /** @var LocaleRepository $localeRepository */
         $localeRepository = $container->get(self::SERVICE_LOCALE_REPOSITORY);
--- a/src/Adapter/ContainerFinder.php
+++ b/src/Adapter/ContainerFinder.php
@@ -48,7 +48,7 @@ class ContainerFinder
         }
         if (isset($context->controller)
             && $context->controller instanceof Controller
-            && $container = $context->controller->getContainer()
+            && ($container = $context->controller->getContainer())
             && null !== $container) {
             return $container;
         }
--- a/src/Adapter/ContainerFinder.php
+++ b/src/Adapter/ContainerFinder.php
@@ -43,12 +43,13 @@ class ContainerFinder
      */
     public function getContainer(Context $context)
     {
-        $container = isset($context->container) ? $context->container : null;
-        if (null !== $container) {
-            return $container;
+        if (isset($context->container)) {
+            return $context->container;
         }
-        $container = isset($context->controller) && $context->controller instanceof Controller ? $context->controller->getContainer() : null;
-        if (null !== $container) {
+        if (isset($context->controller)
+            && $context->controller instanceof Controller
+            && $container = $context->controller->getContainer()
+            && null !== $container) {
             return $container;
         }
 
--- a/src/Adapter/ContainerFinder.php
+++ b/src/Adapter/ContainerFinder.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Adapter;
+
+use Context;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Find the container
+ */
+class ContainerFinder
+{
+    /**
+     * @param Context $context
+     *
+     * @return ContainerBuilder|ContainerInterface|null
+     */
+    public function getContainer(Context $context)
+    {
+        $container = isset($context->container) ? $context->container : null;
+        if (!is_null($container)) {
+            return $container;
+        }
+        $container = isset($context->controller) ? $context->controller->getContainer() : null;
+        if (!is_null($container)) {
+            return $container;
+        }
+
+        return SymfonyContainer::getInstance();
+    }
+}
--- a/src/Adapter/ContainerFinder.php
+++ b/src/Adapter/ContainerFinder.php
@@ -49,7 +49,8 @@ class ContainerFinder
         if (isset($context->controller)
             && $context->controller instanceof Controller
             && ($container = $context->controller->getContainer())
-            && null !== $container) {
+            && null !== $container
+        ) {
             return $container;
         }
 
--- a/src/Adapter/ContainerFinder.php
+++ b/src/Adapter/ContainerFinder.php
@@ -27,6 +27,7 @@
 namespace PrestaShop\PrestaShop\Adapter;
 
 use Context;
+use Controller;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
@@ -43,11 +44,11 @@ class ContainerFinder
     public function getContainer(Context $context)
     {
         $container = isset($context->container) ? $context->container : null;
-        if (!is_null($container)) {
+        if (null !== $container) {
             return $container;
         }
-        $container = isset($context->controller) ? $context->controller->getContainer() : null;
-        if (!is_null($container)) {
+        $container = isset($context->controller) && $context->controller instanceof Controller ? $context->controller->getContainer() : null;
+        if (null !== $container) {
             return $container;
         }
 
--- a/src/Adapter/ContainerFinder.php
+++ b/src/Adapter/ContainerFinder.php
@@ -28,6 +28,7 @@ namespace PrestaShop\PrestaShop\Adapter;
 
 use Context;
 use Controller;
+use Exception;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
@@ -40,6 +41,8 @@ class ContainerFinder
      * @param Context $context
      *
      * @return ContainerBuilder|ContainerInterface|null
+     *
+     * @throws Exception
      */
     public function getContainer(Context $context)
     {
@@ -53,7 +56,11 @@ class ContainerFinder
         ) {
             return $container;
         }
+        $container = SymfonyContainer::getInstance();
+        if (null !== $container) {
+            return $container;
+        }
 
-        return SymfonyContainer::getInstance();
+        throw new Exception('Kernel Container is not available');
     }
 }
--- a/src/Adapter/ContainerFinder.php
+++ b/src/Adapter/ContainerFinder.php
@@ -44,6 +44,7 @@ class ContainerFinder
 
     /**
      * ContainerFinder constructor.
+     *
      * @param Context $context
      */
     public function __construct(Context $context)
--- a/src/Adapter/ContainerFinder.php
+++ b/src/Adapter/ContainerFinder.php
@@ -38,20 +38,32 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 class ContainerFinder
 {
     /**
+     * @var Context
+     */
+    protected $context;
+
+    /**
+     * ContainerFinder constructor.
      * @param Context $context
-     *
+     */
+    public function __construct(Context $context)
+    {
+        $this->context = $context;
+    }
+
+    /**
      * @return ContainerBuilder|ContainerInterface|null
      *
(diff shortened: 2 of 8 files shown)


src/Adapter/ContainerFinder.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    /**
+     * @param Context $context
+     *
+     * @return ContainerBuilder|ContainerInterface|null
+     */
+    public function getContainer(Context $context)
+    {
+        $container = isset($context->container) ? $context->container : null;
+        if (!is_null($container)) {
```
PierreRambaud: Always prefer this annotation because of performance
```suggestion
        if (null !== $container)) {
```

src/Adapter/ContainerFinder.php
```diff
@@ -0,0 +1,56 @@
...
+            return $container;
+        }
+        $container = isset($context->controller) ? $context->controller->getContainer() : null;
```
PierreRambaud: You need to test if controller is an instance of `Controller`

src/Adapter/ContainerFinder.php
```diff
@@ -0,0 +1,56 @@
...
+        if (!is_null($container)) {
```
PierreRambaud: Same as previous

PierreRambaud (changes requested):
See comments

src/Adapter/ContainerFinder.php
Progi1984: :ok_hand:

src/Adapter/ContainerFinder.php
Progi1984: :ok_hand:

src/Adapter/ContainerFinder.php
Progi1984: :ok_hand:

PierreRambaud (dismissed):

src/Adapter/ContainerFinder.php
```diff
@@ -0,0 +1,57 @@
...
+     * @param Context $context
+     *
+     * @return ContainerBuilder|ContainerInterface|null
+     */
+    public function getContainer(Context $context)
+    {
+        $container = isset($context->container) ? $context->container : null;
+        if (null !== $container) {
+            return $container;
+        }
```
eternoendless, line 71: What about this:

```
if (isset($context->container)) {
   return $context->container;
}
```

if it's `isset` it means that it's not null, no need to assign & verify.

src/Adapter/ContainerFinder.php
```diff
@@ -0,0 +1,57 @@
...
+        $container = isset($context->controller) && $context->controller instanceof Controller ? $context->controller->getContainer() : null;
+        if (null !== $container) {
+            return $container;
+        }
```
eternoendless, line 75: I think this could be simplified without the ternary:

```
if (isset($context->controller)
    && $context->controller instanceof Controller
    && $container = $context->controller->getContainer()
    && null !== $container
) {
    return $container;
}

File: tests/Unit/Adapter/ContainerFinderTest.php [hunk 6]
```diff
@@ -0,0 +1,98 @@
...
+class ContainerFinderTest extends TestCase
+{
+    public function testGetContainerFromAttribute()
+    {
+        $contextMock = $this->getMockContext();
+        $contextMock->container = $this->getMockContainerInterface();
+
+        $containerFinder = new ContainerFinder();
+        $this->assertInternalType('object', $containerFinder->getContainer($contextMock));
+        $this->assertInstanceOf('Symfony\Component\DependencyInjection\ContainerInterface', $containerFinder->getContainer($contextMock));
```
eternoendless: Prefer use of ::class
```suggestion
        $this->assertInstanceOf(Symfony\Component\DependencyInjection\ContainerInterface::class, $containerFinder->getContainer($contextMock));
```

tests/Unit/Adapter/ContainerFinderTest.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public function testGetContainerFromController()
+    {
+        $contextMock = $this->getMockContext();
+        $contextMock->controller = $this->getMockController();
+
+        $containerFinder = new ContainerFinder();
+        $this->assertInternalType('object', $containerFinder->getContainer($contextMock));
+        $this->assertInstanceOf('Symfony\Component\DependencyInjection\ContainerInterface', $containerFinder->getContainer($contextMock));
```
eternoendless: Same here:
```suggestion
        $this->assertInstanceOf(Symfony\Component\DependencyInjection\ContainerInterface::class, $containerFinder->getContainer($contextMock));
```

tests/Unit/Adapter/ContainerFinderTest.php
```diff
@@ -0,0 +1,98 @@
...
+
+class ContainerFinderTest extends TestCase
+{
+    public function testGetContainerFromAttribute()
+    {
+        $contextMock = $this->getMockContext();
+        $contextMock->container = $this->getMockContainerInterface();
+
+        $containerFinder = new ContainerFinder();
+        $this->assertInternalType('object', $containerFinder->getContainer($contextMock));
```
eternoendless: Maybe this assertion is redundant with the next one?

tests/Unit/Adapter/ContainerFinderTest.php
```diff
@@ -0,0 +1,98 @@
...
+        $this->assertInstanceOf('Symfony\Component\DependencyInjection\ContainerInterface', $containerFinder->getContainer($contextMock));
+    }
+
+    public function testGetContainerFromController()
+    {
+        $contextMock = $this->getMockContext();
+        $contextMock->controller = $this->getMockController();
+
+        $containerFinder = new ContainerFinder();
+        $this->assertInternalType('object', $containerFinder->getContainer($contextMock));
```
eternoendless: Maybe redundant?

src/Adapter/ContainerFinder.php
```diff
@@ -0,0 +1,57 @@
...
+class ContainerFinder
+{
+    /**
+     * @param Context $context
+     *
+     * @return ContainerBuilder|ContainerInterface|null
+     */
+    public function getContainer(Context $context)
+    {
+        $container = isset($context->container) ? $context->container : null;
```
eternoendless: The attribute `container` does not exist in `Context`

src/Adapter/ContainerFinder.php
```diff
@@ -0,0 +1,57 @@
...
+        if (null !== $container) {
+            return $container;
+        }
+        $container = isset($context->controller) && $context->controller instanceof Controller ? $context->controller->getContainer() : null;
+        if (null !== $container) {
+            return $container;
+        }
+
+        return SymfonyContainer::getInstance();
```
eternoendless: I think it would be a good idea to throw an exception if this is null

src/Adapter/ContainerFinder.php
```diff
@@ -0,0 +1,57 @@
...
+ * Find the container
+ */
+class ContainerFinder
+{
+    /**
+     * @param Context $context
+     *
+     * @return ContainerBuilder|ContainerInterface|null
+     */
+    public function getContainer(Context $context)
```
eternoendless: This parameter should be injected in the constructor don't you think?

eternoendless (changes requested):

src/Adapter/ContainerFinder.php
mickaelandrieu: why?

src/Adapter/ContainerFinder.php
mickaelandrieu: This is the trick to avoid to create a fake controller et so on avoid side effects.

src/Adapter/ContainerFinder.php
mickaelandrieu: Since @quetzacoalt91 pr, Symfony Container::getInstance() can't be null

mickaelandrieu (dismissed):
good job 👍

src/Adapter/ContainerFinder.php
Progi1984: :ok_hand:

tests/Unit/Adapter/ContainerFinderTest.php
Progi1984: :ok_hand:

tests/Unit/Adapter/ContainerFinderTest.php
Progi1984: :ok_hand:

tests/Unit/Adapter/ContainerFinderTest.php
Progi1984: @eternoendless I prefer check if the return is an object and not null before checking its class. No ?

classes/Tools.php
```diff
@@ -792,10 +792,8 @@ protected static function getContextLocale(Context $context)
             return $locale;
         }
 
-        $container = isset($context->controller) ? $context->controller->getContainer() : null;
-        if (null === $container) {
-            $container = SymfonyContainer::getInstance();
-        }
+        $containerFinder = new ContainerFinder();
```
jolelievre: Hi, I don't understand the need for this `ContainerFinder` it seems to be hiding some magic about how to get the container, and can return:
- the container from the Context property (which isn't even defined)
- container from the controller (if you are luck to have one in the context)
- the Symfony container as a fallback

I know this will work but in the end you don't even know which container you will have.
I would suggest simply using the `ContainerBuilder` as you use it (which is the right approach I think)

```php
$container = ContainerBuilder::getContainer('webservice', _PS_MODE_DEV_);
```

This way you are sure to have the container defined for the webservice environment, and don't worry it won't be instanciated several times because `ContainerBuilder` stores the instanciated containers by environment name

jolelievre (changes requested):
Nice approach for the `ContainerBuilder` but the `ContainerFinder` seems a bit weird

classes/Tools.php
jolelievre: Oh my bad I didn't notice all this is made in the Tools class..

Progi1984: @jolelievre But in Tools, you don't know of you need webservice/front/admin container.

tests/Unit/Adapter/ContainerFinderTest.php
jolelievre: if it's not an object it won't match the expected class either, so indeed it seems redundant

classes/Tools.php
jolelievre: Yep I realized it a bit later ^^

Progi1984: @jolelievre Finally Changes or Approved ? :)

src/Adapter/ContainerFinder.php
Progi1984: :ok_hand:

classes/Tools.php
jolelievre: Not a big fan that we need to do this.. but I don't see another solution for now So I'll approve
I believe @eternoendless has a few feedback not answered yet

jolelievre (dismissed):

classes/Tools.php
Progi1984: I understand. But after discussion with @mickaelandrieu, it's the cleanest solution to offer.

src/Adapter/ContainerFinder.php
Progi1984: @eternoendless Isn't it unsafe to replace a behavior where there are no exceptions to a case where there is ?

src/Adapter/ContainerFinder.php
eternoendless: Because the behavior is tightly coupled to Context, so it's a dependency. Otherwise, the method name could be made static and changed to `getContainerFromContext`, but it's not all that it does, and even then it would feel like the Context is being micro-managed by an external element.

src/Adapter/ContainerFinder.php
eternoendless: It can if the kernel is not available and if the request does not go though a controller :trollface:

src/Adapter/ContainerFinder.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        if (isset($context->container)) {
+            return $context->container;
+        }
+        if (isset($context->controller)
+            && $context->controller instanceof Controller
+            && ($container = $context->controller->getContainer())
+            && null !== $container) {
```
eternoendless: For readability's sake:

```suggestion
            && null !== $container
        ) {
```

src/Adapter/ContainerFinder.php
eternoendless: > Isn't it unsafe to replace a behavior where there are no exceptions to a case where there is ?

In general, I believe that class should do what it's asked to, and throw an Exception if it's unable to do it. This technique is also beneficial for identifying unexpected errors that could go by unnoticed otherwise.

webservice/dispatcher.php
```diff
@@ -24,12 +24,15 @@
...
 
+use PrestaShop\PrestaShop\Adapter\ContainerBuilder;
+
 ob_start();
 
 require_once dirname(__FILE__).'/../config/config.inc.php';
 
 // Cart is needed for some requests
 Context::getContext()->cart = new Cart();
+Context::getContext()->container = ContainerBuilder::getContainer('webservice', _PS_MODE_DEV_);
```
eternoendless, line 35: This property must be declared in Context

classes/Tools.php
```diff
@@ -792,10 +792,8 @@ protected static function getContextLocale(Context $context)
...
+        $container = $containerFinder->getContainer($context);
```
eternoendless: You could also attach the container here to `$this->container` if not set already.

eternoendless (changes requested):

src/Adapter/ContainerFinder.php
Progi1984: :ok_hand:

webservice/dispatcher.php
Progi1984: :ok_hand:

classes/Tools.php
Progi1984: @eternoendless But there is no property container in the class Tools. May be you want to attach to the property container of the class Context. Isn't it ?

src/Adapter/ContainerFinder.php
Progi1984: :ok_hand:

classes/Tools.php
eternoendless: Whoops, yes, that's what I meant 😅

src/Adapter/ContainerFinder.php
Progi1984: Done

classes/Tools.php
Progi1984: Done ;)

PierreRambaud (approved):

eternoendless (approved):

eternoendless:
Looking good!

khouloudbelguith:
Hi @Progi1984,

It is OK.

Thanks!

marionf:
Question guys, seems the issue is a regression, this PR shouldn't be on 1.7.6.x branch ?

PierreRambaud:
Thanks @Progi1984

marionf:
@Progi1984 @PierreRambaud 
This PR should have been done on 176x branch

matthieu-rolland:
Thanx @Progi1984  !

I think @marionf is right, I needed this on 1.7.6.x  for the double translation service issue :D

Maybe I should implement the same logic on 176x in my translation fixing branch ?

Renaud6322 (commented):
Dears Sirs, on my 1.7.6.0 as I had the bug "get on null" on Tools:801 when create an order via webservice (with $xml->order->current_state) I applied all the changes described in this page, but now my webservice returns : Class 'PrestaShop\PrestaShop\Adapter\ContainerFinder' not found in /homepages/21/d473697229/htdocs/mysite/classes/Tools.php:800
Could you help me ? Thanks a lot in advance.

jolelievre:
Ih @Renaud6322,

the `ContainerFinder` class was introduced in 1.7.6.1 version, which is why you can't use the content of this PR as is like a patch It is part of the development of a full version (1.7.7) not some patches you can apply independently.

Sometimes it will work, most times you don't know if you have all the dependencies so I would **strongly** advise against picking pieces of code like this. You can try integrating the `ContainerFinder` class manually in your 1.7.6 shop but it's risky, the better solution would be to wait for the 1.7.7 release and upgrade your shop, or at least update to the latest version 1.7.6.5 which includes this class.

Alternatively you can try the current beta version of 1.7.7 to see if it meets your needs.

Renaud6322:
Thanks a lot, @jolelievre I will try to upgrade to 1.7.6.5. However I don't understand why it's impossible to create an order by the webservice with the 1.7.6.0 !!!

PR #14902, opened 2019-07-29
Check the Front Office test with puppeteer - Smoke tests campaign

nesrineabdmouleh:
| Questions     | Answers
| Branch?       | develop
| Description?  | Check the Front office test with puppeteer in Smoke tests campaign
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | npm run specific-test smoke/crawlingFO

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14902)

### DIFF
--- a/tests/puppeteer/pages/FO/FO_cart.js
+++ b/tests/puppeteer/pages/FO/FO_cart.js
@@ -1,0 +1,35 @@
+const CommonPage = require('../commonPage');
+
+module.exports = class FO_CART extends CommonPage {
+  constructor(page) {
+    super(page);
+
+    // Selectors for cart page
+    this.productName = '#main li:nth-of-type(%NUMBER) div.product-line-info > a';
+    this.productPrice = '#main li:nth-of-type(%NUMBER) div.current-price > span';
+    this.productQuantity = '#main li:nth-of-type(%NUMBER) div.input-group input.js-cart-line-product-quantity';
+    this.proceedToCheckoutButton = '#main div.checkout a';
+
+    // Selectors for checkout page
+    this.checkoutStepOneTitle = '#checkout-personal-information-step > h1';
+  }
+
+  /**
+   * To check the cart details (product name, price, quantity)
+   * @param cartData, cart data to check
+   * @param productID, product id to check
+   */
+  async checkCartDetails(cartData, productID) {
+    await this.checkTextValue(this.productName.replace('%NUMBER', productID), cartData.name_fr);
+    await this.checkTextValue(this.productPrice.replace('%NUMBER', productID), cartData.price);
+    await this.checkAttributeValue(this.productQuantity.replace('%NUMBER', productID), 'value', cartData.quantity);
+  }
+
+  /**
+   * Click on Proceed to checkout button
+   */
+  async clickOnProceedToCheckout() {
+    await this.waitForSelectorAndClick(this.proceedToCheckoutButton);
+    await this.page.waitForSelector(this.checkoutStepOneTitle);
+  }
+};
--- a/tests/puppeteer/pages/FO/FO_home.js
+++ b/tests/puppeteer/pages/FO/FO_home.js
@@ -1,0 +1,27 @@
+const CommonPage = require('../commonPage');
+
+module.exports = class FO_HOME extends CommonPage {
+  constructor(page) {
+    super(page);
+
+    // Selectors for home page
+    this.logoHomePage = '#_desktop_logo';
+    this.productImg = '#content article:nth-child(%NUMBER) img';
+  }
+
+  /**
+   * Check home page
+   */
+  async checkHomePage() {
+    await this.page.waitForSelector(this.logoHomePage, {visible: true});
+  }
+
+  /**
+   * Go the product page
+   * @param id, product id
+   */
+  async goToProductPage(id) {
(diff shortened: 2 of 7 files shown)


tests/puppeteer/pages/FO/FO_cart.js
```diff
@@ -0,0 +1,35 @@
+const CommonPage = require('../commonPage');
```
PierreRambaud, line 1: why `FO/FO_` it's redundant.
Rename file please.

tests/puppeteer/pages/FO/FO_cart.js
```diff
@@ -0,0 +1,35 @@
...
+
+module.exports = class FO_CART extends CommonPage {
```
PierreRambaud, line 3:
```suggestion
module.exports = class Cart extends CommonPage {
```

tests/puppeteer/pages/FO/FO_home.js
```diff
@@ -0,0 +1,27 @@
+const CommonPage = require('../commonPage');
+
+module.exports = class FO_HOME extends CommonPage {
```
PierreRambaud, line 3:
```suggestion
module.exports = class Home extends CommonPage {
```

tests/puppeteer/pages/FO/FO_product.js
```diff
@@ -0,0 +1,34 @@
+const CommonPage = require('../commonPage');
+
+module.exports = class FO_PRODUCT extends CommonPage {
```
PierreRambaud, line 3:
```suggestion
module.exports = class Product extends CommonPage {
```

tests/puppeteer/test/data/FO/cart.js
```diff
@@ -0,0 +1,9 @@
+
```
PierreRambaud:
```suggestion
```

tests/puppeteer/pages/FO/FO_cart.js
SimonGrn: These modifications will be done in a separate PR.

tests/puppeteer/pages/FO/FO_cart.js
SimonGrn: These modifications will be done in a separate PR.

tests/puppeteer/pages/FO/FO_cart.js
```diff
@@ -0,0 +1,35 @@
...
+  constructor(page) {
+    super(page);
+
+    // Selectors for cart page
+    this.productName = '#main li:nth-of-type(%NUMBER) div.product-line-info > a';
```
jolelievre, line 8: In our recent javascript PRs (mainly on back office), we were asked to separate the selector and put them in a separate file Which seems like a very good idea to ease the reusability in E2E tests.
So I'm wondering if these selectors should be in a separate file as well?

tests/puppeteer/pages/FO/FO_home.js
```diff
@@ -0,0 +1,27 @@
...
+   */
+  async checkHomePage() {
+    await this.page.waitForSelector(this.logoHomePage, {visible: true});
+  }
+
+  /**
+   * Go the product page
+   * @param id, product id
+   */
+  async goToProductPage(id) {
```
jolelievre, line 23: This seems like a method that will be needed in a lot of tests SHould this be in the `CommonPage` ?

tests/puppeteer/pages/commonPage.js
```diff
@@ -57,4 +57,52 @@ module.exports = class CommonPage {
...
+  async checkTextValue(selector, textToCheckWith, parameter = 'equal') {
+    await this.page.waitForSelector(selector);
+    switch (parameter) {
+      case 'equal':
+        await this.page.$eval(selector, el => el.innerText)
+          .then(text => expect(text.replace(/\s+/g, ' ').trim()).to.equal(textToCheckWith));
+        break;
+      case 'contain':
+        await this.page.$eval(selector, el => el.innerText)
+          .then(text => expect(text).to.contain(textToCheckWith));
```
jolelievre, line 88: Don't you need the same treatment as with `equal` and remove extra spaces? `text.replace(/\s+/g, ' ').trim()`

tests/puppeteer/pages/FO/FO_cart.js
SimonGrn: That's a problem we've been struggling with. At the moment your implementation is incomplete and we need to go on. I think there is a way to get selectors from your files and inject them into our logic, but even if that was easily feasible we still will have the problem of the names. So when our implementation and your implementation are complete we'll have to rework everything anyway.

tests/puppeteer/pages/FO/FO_home.js
SimonGrn: Like I understand it, it's a method to choose the nth product on the homepage. So it needs to be binded to the homepage.

tests/puppeteer/pages/commonPage.js
nesrineabdmouleh: No, the contain function check if the textToCheckWith contain text, then we don't need to remove extra spaces

matthieu-rolland (approved):
lgtm

PR #14925, opened 2019-07-31
Copy images from SCSS directory to the public directory when building the new theme

### LINKED ISSUE #14757: Some images are being loaded from the SCSS directory
When we tried removing the SCSS directory from the BO new theme in the production zip release, we realized that some images were being loaded from there, and therefore weren't available anymore.

This means that either the image extraction webpack plugin is not working properly for some image, or that some of them are being referenced directly in the code.

See [this comment](https://github.com/PrestaShop/PrestaShop/pull/14579#issuecomment-510134341) for more information.

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | When building the ZIP production, some images called to SCSS directory from the them
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14757 
| How to test?  | After building the ZIP (`php tools/build/CreateRelease.php`), images doesn't have to be called in SCSS directory

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14925)


admin-dev/themes/new-theme/.webpack/common.js
```diff
@@ -185,5 +186,14 @@ module.exports = {
       $: 'jquery', // needed for jquery-ui
       jQuery: 'jquery',
     }),
+    new CopyPlugin([
```
PierreRambaud, line 214:
```js
    new CopyPlugin([
      { from: 'scss/img' },
    ])
```

src/PrestaShopBundle/Resources/views/Admin/Configure/AdvancedParameters/Employee/Blocks/showcase_card.html.twig
```diff
@@ -26,7 +26,7 @@
 {% if not isShowcaseCardClosed %}
   <div id="employeesShowcaseCard" class="showcase-card card">
     <div class="helper-card__left helper-card__team shape-one">
-      <img src="{{ asset('themes/new-theme/scss/img/helper-card/team@3x.png') }}" class="img-fluid">
+      <img src="{{ asset('themes/new-theme/public/team@3x.png') }}" class="img-fluid">
```
PierreRambaud: And it will be 
```suggestion
      <img src="{{ asset('themes/new-theme/public/helper-card/team@3x.png') }}" class="img-fluid">
```

admin-dev/themes/new-theme/.webpack/common.js
Progi1984: In fact, no, because some images are ever copied with file-loader because they are used in scss files.

PierreRambaud: oh, you're right :D

PierreRambaud (dismissed):

jolelievre (dismissed):

admin-dev/themes/new-theme/.webpack/common.js
```diff
@@ -186,5 +187,8 @@ module.exports = {
...
+      { from: 'static' },
```
eternoendless, line 215: Is it me or this directory doesn't exist and you just moved everything to public?

eternoendless (changes requested):

admin-dev/themes/new-theme/.webpack/common.js
eternoendless: Oh my mistake, you moved AND copied them.

eternoendless (dismissed):

eternoendless:
Needs a rebase

jolelievre (approved):
Thanks @PierreRambaud @Progi1984

Issue #14785, opened 2019-07-19
Error after filtering date in error logs


ChineseNorris:
**Describe the bug**
After filtering the error log in the back office I get the following error (debug mode enabled)

An exception occurred while executing 'SELECT l.*, e.email, CONCAT(e.firstname, ' ', e.lastname) as employee FROM ps_log l LEFT JOIN ps_employee e ON l.id_employee = e.id_employee WHERE (object_type LIKE :object_type) AND (l.date_add BETWEEN :date_from AND :date_to) ORDER BY id_log desc LIMIT 20 OFFSET 0' with params ["2019-07-19", "2019-07-19"]:

Value for :object_type not found in params array. Params array key should be "object_type"

**To Reproduce**
1. Go to BO/Advanced parameters/error logs
2. Click on Date range and click filter
3. See error

**Additional information**
PrestaShop version: 1.7.5.1
PHP version: 7.2

khouloudbelguith:
Hi @ChineseNorris,

Thanks for your report.
I manage to reproduce the issue with PS1.7.6.0 & PS1.7.5.2 when debug mode enabled.
I tried to make a filter by two fields => **Date** + another field => the exception will be displayed.

Filtering by only one field => OK
![image](https://user-images.githubusercontent.com/16067358/61616999-c517fe80-ac61-11e9-923b-c021bcbe8f67.png)

I’ll add this to the debug roadmap so that it’s fixed. If you have already fixed it on your end or if you think you can do it, please do send us a pull request!
Thanks!

ChineseNorris:
Hi @khouloudbelguith ,

thank you for reproducing the issue, I'm trying to clear at least the filter but without success, I've tried using another account to get the method for resetting the filter but I get error 405 "Method not allowed".

I need to reset the filter first and then I'll try to find a solution for the bug.

Regards

matks:
We need to fix this bug _and_ add Behat tests to make sure it does not come back 😄

01park:
I applied the fixed files and the error is gone, but when I want to filter an employee on employee number or name, I get no results for a selected date.

PR #14926, opened 2019-07-31
Fix logs filtering

sarjon:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Fixes log filtering by setting query parameters separately since using `setPramaters()` resets all previously set parameters into query builder.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/14785
| How to test?  | See ticket for how to test.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14926)


Progi1984 (approved):

PierreRambaud (approved):

PierreRambaud:
Thanks @sarjon

01park:
Hi,

I changed the files and the error is now gone, but it seem the filter isn't working good.

When I select a date and search on employee id, no results. The same when I search on name.

When I search on product it, it works

PR #14938, opened 2019-07-31
Fix error when loading theme that includes a translations directory but no translations

### LINKED ISSUE #14960: Importing a theme that has a translations folder but no files in it will throw an exception
**Describe the bug**
When loading a theme, ThemeManager will try to load its translations if the translations directory is present. But if it's present AND empty it would throw an exception.

**To Reproduce**
1. Create a theme that is not named classic, make sure there's a translations directory in it with nothing inside.
2. Zip it
3. Try to import it into PS.

**Additional information**
PrestaShop version: 1.7.6.0
PHP version: 7.1

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | When loading a theme, ThemeManager will try to load its translations if the `translations` directory is present. But if it's present AND empty it would throw an exception.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | Yes, `TranslationFinderTrait` has been extracted into a class of its own, `TranslationFinder`.
| Fixed ticket? | Fixes #14960
| How to test?  | Create a theme that is not named `classic`, make sure there's a `translations` directory in it with nothing inside. Zip it and try to import it into PS.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14938)

### DIFF
--- a/src/PrestaShopBundle/Translation/Provider/TranslationFinder.php
+++ b/src/PrestaShopBundle/Translation/Provider/TranslationFinder.php
@@ -39,9 +39,11 @@ use Symfony\Component\Translation\MessageCatalogueInterface;
  */
 class TranslationFinder
 {
+    const ERR_NO_FILES_IN_DIRECTORY = 1;
+    const ERR_DIRECTORY_NOT_FOUND = 2;
 
     /**
-     * @param array $paths a list of paths when we can look for translations
+     * @param string|array $paths a list of paths when we can look for translations
      * @param string $locale the Symfony (not the PrestaShop one) locale
      * @param string|null $pattern a regular expression
      *
@@ -67,13 +69,13 @@ class TranslationFinder
                     'Could not crawl for translation files: %s',
                     $e->getMessage()
                 ),
-                0,
+                self::ERR_DIRECTORY_NOT_FOUND,
                 $e
             );
         }
 
         if (count($translationFiles) === 0) {
-            throw new FileNotFoundException('There is no translation file available.');
+            throw new FileNotFoundException('There are no translation file available.', self::ERR_NO_FILES_IN_DIRECTORY);
         }
 
         /** @var SplFileInfo $file */
--- a/src/PrestaShopBundle/Translation/Provider/TranslationFinder.php
+++ b/src/PrestaShopBundle/Translation/Provider/TranslationFinder.php
@@ -53,48 +53,9 @@ class TranslationFinder
      */
     public function getCatalogueFromPaths($paths, $locale, $pattern = null)
     {
-        $messageCatalogue = new MessageCatalogue($locale);
-        $xliffFileLoader = new XliffFileLoader();
-        $finder = new Finder();
+        $translationFiles = $this->getTranslationFilesFromPath($paths, $pattern);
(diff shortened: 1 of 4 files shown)


src/PrestaShopBundle/Translation/Provider/TranslationFinder.php
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        /** @var SplFileInfo $file */
+        foreach ($translationFiles as $file) {
+            if ('xlf' === $file->getExtension()) {
+                if (strpos($file->getBasename('.xlf'), $locale) !== false) {
+                    $domain = $file->getBasename('.xlf');
+                } else {
+                    $domain = $file->getBasename('.xlf') . '.' . $locale;
+                }
```
matthieu-rolland: you could probably avoid the **else** here:

```php
$domain = $file->getBasename('.xlf');

if (strpos($file->getBasename('.xlf'), $locale) === false) {
  $domain .= '.' . $locale;
}
```

src/PrestaShopBundle/Translation/Provider/TranslationFinder.php
```diff
@@ -0,0 +1,122 @@
...
+     * @param string|array $paths a list of paths when we can look for translations
+     * @param string $locale the Symfony (not the PrestaShop one) locale
+     * @param string|null $pattern a regular expression
+     *
+     * @return MessageCatalogue
+     *
+     * @throws FileNotFoundException
+     */
+    public function getCatalogueFromPaths($paths, $locale, $pattern = null)
+    {
```
matthieu-rolland, line 55: Maybe some parts of this method could have been extracted in two different methods, that would be called in it ?

**getTranslationFilesFromPaths**() and a method **getFileCatalogueFromTranslationFile**() that would be called inside the foreach;

It could make this method easier to read, but this is not mandatory, maybe my suggestion is overkill... your call.

matthieu-rolland: example:

```php

private function getTranslationFilesFromPaths($paths)
    {
        try {
            $translationFiles = $finder->files()->notName('index.php')->in($paths);
        } catch (\InvalidArgumentException $e) {
            throw new FileNotFoundException(
                sprintf(
                    'Could not crawl for translation files: %s',
                    $e->getMessage()
                ),
                self::ERR_DIRECTORY_NOT_FOUND,
                $e
            );
        }
        if (count($translationFiles) === 0) {
            throw new FileNotFoundException('There are no translation file available.', self::ERR_NO_FILES_IN_DIRECTORY);
        }
        
        return $translationFiles;
    }

```

eternoendless:
Done!

matthieu-rolland (approved):
LGTM !

khouloudbelguith:
Hi @eternoendless,

It is OK.
checked with this theme
[childtheme-example.zip](https://github.com/PrestaShop/PrestaShop/files/3472534/childtheme-example.zip)

Thanks!

eternoendless:
Thanks everyone!

Issue #12844, opened 2019-03-08
Webservice is letting create customers with same email


mowcixo:
**Describe the bug**

When you create two customers with same email from webservice, it doesn't throw any error. Then when you try to login with one of two customers, the login form fails.

**To Reproduce**
1. Create a customer with foo@bar.com email using webservice endpoint.
2. Create another customer with foo@bar.com email using webservice endpoint.
3. Go to login form and try to login with one of them.
4. You will receive an error.

**Additionnal information**
PrestaShop version: 1.7.5.1
PHP version: N/A

I would be glad to fix it if some of mantainers give me some clue about what files should be involved.

khouloudbelguith:
Hi @mowcixo,

Yes I manage to reproduce the issue with PS1.7.5.1 & PS1.7.4.4.
I used this script

```
<?php
/**
 * Created by PhpStorm.
 * User: khouloud.belguith
 * Date: 08/13/19
 * Time: 14:44
 */

ini_set('display_errors','on');
// Here we define constants /!\ You need to replace this parameters
define('DEBUG', true); // Debug mode
//                      !! BACHA, nezapomenout přenastavit i DTB u obrázků dole !!!
define('PS_SHOP_PATH', 'http://shop/'); // Root path of your PrestaShop store
define('PS_WS_AUTH_KEY', 'Key'); // Auth key (Get it in your Back Office)
require_once('./PSWebServiceLibrary.php');

$psXML = <<<XML
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<prestashop xmlns:xlink="http://www.w3.org/1999/xlink">
 <customer>
   <email>test@test.com</email>
   <firstname>khouloud</firstname>
   <lastname>belguith</lastname>
   <passwd>abc123</passwd>
 </customer>
</prestashop>
XML;
$webService = new PrestaShopWebservice(PS_SHOP_PATH, PS_WS_AUTH_KEY, DEBUG);
$xml = new SimpleXMLElement($psXML);
var_dump($xml->asXML());
$opt = array( 'resource' => 'customers' );
$opt['postXml'] = $xml->asXML();
//$opt['id_shop'] = 'all';
for ($i = 1; $i <= 2; $i++) {
    $xml = $webService->add( $opt );
}
```
Which I can create two customers with same informations successfully.
But, in Fo, I have an error in FO.
[...]

PR #14952, opened 2019-08-01
Webservice is letting create customers with same email #12844

Mch0:
…l to validate this email already exist or not

| Questions     | Answers
| Branch?       | develop
| Description?  | We can create several times the same user with api
| Type?         | bug fix 
| Category?     | BO
| Fixed ticket? |  #12844
| How to test?  | Send several same xml api request with the same email

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14952)

### DIFF
--- a/classes/Customer.php
+++ b/classes/Customer.php
@@ -151,7 +151,7 @@ class CustomerCore extends ObjectModel
 
     protected $webserviceParameters = array(
         'objectMethods' => array(
-            'add' => 'addWs'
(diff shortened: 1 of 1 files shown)


mowcixo:
I think it should return an HTTP Conflict/Bad Request/Forbidden response, instead of updating the existent user.

matks:
Fixed by https://github.com/PrestaShop/PrestaShop/pull/16553

PR #13088, opened 2019-03-28
Ensure translations are always saved with the selected theme in back office

mickaelandrieu:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | In some cases, the theme was not selected when saving translations: now we have introduced back the concept of translation per theme, we couldn't manage to retrieve the translations.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #13084
| How to test?  | I summon the great @marionf here, she knows what she has to do :-) 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/13088)

Issue #13870, opened 2019-05-20
Allow the Back office translator to discover new core/custom wordings


eternoendless:
**Is your feature request related to a problem? Please describe.**

When working with core translations, the Back Office translation page only shows wordings which are currently present in the default catalogue (xlf files in `app/Resources/translations/default`), which is extracted manually from the code at build time.

If the core code is modified and a new wording is added, it won't be appear in the BO translations page unless it's manually added to the default catalogue.

**Describe the solution you'd like**

A button in the page would trigger discovery of new wordings and add them to the default catalogue (or maybe some sort of intermediate cache -- that would require changing the translation system by adding a loader, but it would be cleaner and better for upgrading).

**Describe alternatives you've considered**

I think that adding the translation directly in the database (table `ps_translation`) would work but I'm not sure.

Hlavtox:
That would be a nice feature to have. In 1.6, you just added a string anywhere and it instantly popped up in the translation catalog.

Regarding this, 1.7 is a huge step back...

eternoendless:
Core files shouldn't be modified in the first place, but if needed, here's a workaround that _should_ work.

1. Alter the code by adding the wording that you want.
2. If you chose an existing translation domain, add the wording to the corresponding .xlf file in `app/Resources/translations/default`, or create a new file following the [domain naming conventions](https://devdocs.prestashop.com/1.7/development/internationalization/translation/translation-domains/).
3. Clear PS cache

After that, the new wording should appear in the translation BO.

matks:
ping @jf-viguier I think you mentioned the need for this recently 😉 (but it was for modules I think)

jf-viguier:
@matks yes, it's a big issue. I use legacy trads in my modules.

boo-code:
Checked against `develop`. The premise has moved since 2019: discovery itself already works, and what is left is a path and filename mismatch.

`bin/console translation:extract` finds wordings that exist only in code. Adding `{{ "Zzz probe wording 13870"|trans({}, "Admin.Actions") }}` to an admin template and running `translation:extract en-US --dump-messages --domain=Admin.Actions` lists the new wording. A full run over the core is 6 seconds for 14163 messages, so it is viable on demand though not per request.

[...]

Issue #13863, opened 2019-05-20
[EPIC] - Translations


eternoendless:
This issue tracks problems with translations in PrestaShop

## BUGS
- [x] #9791
- [x] #9802
- [x] #9804
- [x] #9805
- [x] #10844
- [ ] #13392
- [x] #15268
- [x] #17011
- [x] #22531
- [x] #23149
- [x] #20072
- [x] #16168
- [x] #16421
- [x] #15072
- [ ] #16132
- [x] #15346
- [ ] #19007
- [x] #21305
- [x] #23156
- [ ] #23737
- [x] #25455
- [x] #26750
- [x] #26570
- [x] #28211
- [x] #29518 
- [x] #29316
- [ ] #30822
- [x] #30741
- [x] #31730
- [x] #30716
- [ ] #32175
- [x] #32452
- [ ] #32976 
- [ ] https://github.com/PrestaShop/PrestaShop/issues/35923
- [ ] #30241

## FEATURES
- [ ] #11348
- [ ] #13870
- [x] #13888
- [ ] #15870
- [x] #19242
- [ ] #23765
- [x] #23767
- [ ] #14968
- [ ] #13749

## Fixed
- [x] ~https://github.com/PrestaShop/docs/issues/284~
### Bugs
- [x] #9762
- [x] #9729
- [x] #9818
- [x] #9885
- [x] #9893
- [x] #9973
- [x] #9816
- [x] #15075
- [x] #17844
- [x] #14570
- [x] #12330
- [x] #23233
- [x] #23437
- [x] #23836
- [x] #13024
- [x] #28944
- [x] #27351
- [x] #29164
### Features
- [x] #9561
- [x] #12873
- [x] #23761
- [x] #23762
- [x] #23763
- [x] #23764
- [x] #23766

khouloudbelguith:
https://github.com/PrestaShop/PrestaShop/issues/17844: Can't use translations on module overrides

muyncky:
#17011: Translation not working when deploying with Deployer

matdu58:
@eternoendless 
Impossible to translate product title and description into another language when this language is disabled
https://github.com/PrestaShop/PrestaShop/issues/26570

PR #14962, opened 2019-08-01
Fix error preventing to translate backoffice wordings when using a theme other than classic

### LINKED ISSUE #14733: It is not possible to edit the "Back office translations" in PrestaShop 1.7.6.0
Hello,

I use the PrestaShop 1.7.6.0 and custom theme. In previous version 1.7.5.2 the "Back office translations" was editing fine. Now i can't edit the "Back office translations":

http://prntscr.com/ogrpk5 

http://prntscr.com/ogrpt7

The text of error: 

{"error":"Could not crawl for translation files: The \u0022\/home\/b\/brain7yk\/prestathemes.ru\/public_html\/1.7\/footwear\/var\/cache\/prod\/themes\/footwear\/translations\u0022 directory does not exist."}

Also you can look video with error here:
https://yadi.sk/i/lz5TPSRyiAfepA

I'm trying to edit the "Back office translations", why is the request for the "themes/footwear/ translations /" folder sent? Also this folder exists.

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | This PR fixes an issue making it impossible to translate BO wordings if the current theme is anything other than "classic". Read below for more.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14733
| How to test?  | See issue + test translating at least a wording in BO, FO, classic theme and custom theme.

## More information about this bug

Before this fix, back office translations would be pulled from the current FO theme as well, and any saved translation would also be attached to that theme, even if the wording didn't belong to the FO. This issue would render the translation of backoffice wordings impossible when using a theme other than "classic" (which has a custom behavior compared to any other theme since its wordings are included in the core). 

The problem is related to how the theme translation section works when dealing with themes: it analyzes the theme's code looking for wordings, extracts them into a xliff catalogue in cache, then reads it to build the form. This extraction ONLY happens when working with theme translations. So what happened here was that when requested to load the backoffice translations, it would try to load the theme translations for some reason, and since its xliff catalogue hadn't been extracted, it raised an exception.

This PR reverts changes in #13088 that introduced the forced theme.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14962)

### DIFF
--- a/src/PrestaShopBundle/Translation/Loader/DatabaseTranslationLoader.php
+++ b/src/PrestaShopBundle/Translation/Loader/DatabaseTranslationLoader.php
@@ -55,6 +55,13 @@ class DatabaseTranslationLoader implements LoaderInterface
     public function load($resource, $locale, $domain = 'messages', $theme = null)
     {
         static $langs = array();
+        $catalogue = new MessageCatalogue($locale);
+
+        // do not try and load translations for a locale that cannot be saved to DB anyway
+        if ($locale === 'default') {
+            return $catalogue;
+        }
+
         if (!array_key_exists($locale, $langs)) {
             $langs[$locale] = $this->entityManager
                 ->getRepository('PrestaShopBundle:Lang')
@@ -75,8 +82,6 @@ class DatabaseTranslationLoader implements LoaderInterface
         $translations = $queryBuilder
             ->getQuery()
             ->getResult();
-
-        $catalogue = new MessageCatalogue($locale);
 
         /** @var Translation $translation */
         foreach ($translations as $translation) {
--- a/src/PrestaShopBundle/Translation/Loader/DatabaseTranslationLoader.php
+++ b/src/PrestaShopBundle/Translation/Loader/DatabaseTranslationLoader.php
@@ -28,6 +28,9 @@
 namespace PrestaShopBundle\Translation\Loader;
 
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\QueryBuilder;
+use PrestaShopBundle\Entity\Lang;
+use PrestaShopBundle\Entity\Translation;
 use Symfony\Component\Translation\Loader\LoaderInterface;
 use Symfony\Component\Translation\MessageCatalogue;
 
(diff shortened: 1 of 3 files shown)


kpodemski:
Hello @eternoendless 

i've run some tests, just to clarify current state of translation system in 1.7

1. first scenario, your version from branch from this PR, edited `controllers/admin/AdminAddressesController.php`, changed "First Name" to "Customer first name"

both, with changed domain to `Admin.Custom` and `Admin.Global`, new wording never found in translation system.

i've moved this file to override folder, again, two domains. Result? new wording never found in translation system. This worked well in 1.6.

2. second scenario, wording from front-office, we want to change some strings related to products sorting wich are present in `core files`,  for example `controllers/front/listing/SearchController.php`, this line:
`$this->getTranslator()->trans('Search results', array(), 'Shop.Theme.Catalog')`

i'm going to Translations -> Theme -> My theme -> my language -> search "Search results", and what? no results, let's try with Classic theme!

i'm going to Translations -> Theme -> Classic -> my language -> search for "Search results" and... voila! it's working
image

In my opinion this is very important issue, we can't fully translate our custom theme.

But hey, let's try to insert translation directly to `ps_translation` table, this worked well in 1.7.5!
Result? Yes, it's working fine, still not a best solution for non-tech person using PrestaShop.

Maybe this is similar issue to what you've worked in this PR? I think that `Translator` should be aware of every new strings in core files and template files.

Another important issue which i have with new Translation system, is that we can't translate modules in theme context, so we can't have two themes, one module, and two translations for it, because module translations are stored in `/modules/modulename/translation/iso.php` instead of what we have in 1.6: `/themes/modules/modulename/translation/iso.php`, and i'm not sure if we have any workaround for this issue?

eternoendless:
Hi @kpodemski 

I think the best place to discuss general issues with translations would be in [this dedicated epic](https://github.com/PrestaShop/PrestaShop/issues/13863), or better, within the issues tracked there. Most of the issues you are talking about are there.

Regarding the first scenario, that's because backoffice translations are extracted from the core at build time (before we release a minor version), then put in XLIFF files which are bundled with the release or downloaded as language packs. The BO translation interface is based solely on these files when it comes to back office or front office translations, currently there's no way to make it pull new strings from the code. This is tracked by this issue: #13870

The second scenario happens in part because the theme translation system is different for classic and for other themes. The classic theme pulls its wordings directly form the XLIFF catalogue, while other themes must be analyzed to find new wordings. We need to make some tricky decisions regarding the expected behavior there (eg. should all wordings be displayed in the BO interface or just those that don't exist in the front office wordings? If we display them all, then what's the role of front office wordings?).

There are many issues with translations, and I'm meaning to untangle the bigger part of that mess in 1.7.7. For the time being, this PR only fixes the issue mentioned in the PR description, and only that.

src/PrestaShopBundle/Translation/Loader/DatabaseTranslationLoader.php
```diff
@@ -52,6 +55,13 @@ public function __construct(EntityManagerInterface $entityManager)
     public function load($resource, $locale, $domain = 'messages', $theme = null)
     {
         static $langs = array();
+        $catalogue = new MessageCatalogue($locale);
+
+        // do not try and load translations for a locale that cannot be saved to DB anyway
+        if ($locale === 'default') {
```
jolelievre, line 61: 👍

jolelievre (approved):

khouloudbelguith:
Hi @eternoendless,

Seems that the translation is not updated in the Symfony page
Steps to reproduce the issue:
1. Go to BO => International => Translations => Select 
1.1. Type of translation => Back office
1.2  Select your language => English
1.3 Click on Modify 
2. Search for the expression "View my shop"
3. Try to translate it
4. click Save => OK
5. Go to the Dashboard page => the new expression well displayed
6. Go to any page migrated to Symfony => old expression is displayed

https://drive.google.com/file/d/1n50LmEhI01xRq55Y1X16WVFR6znmet_O/view
Issue created here: https://github.com/PrestaShop/PrestaShop/issues/15052 (reproduced with PS1.7.5.1 & PS1.7.6.0 so not related to this PR)

Except for this issue => OK => PR validated
Thanks!

eternoendless:
Thanks everyone!

fafa0659:
hello

Someone can he help me and let me wich files must modifoed in this issue ? 

thanks you

ElgorTR:
Hello,

I changed the code in the specified files, but the same problem persists. do I need to make another change?

@eternoendless

direwald:
Even after applying the changes from this PR, the issue still presists when trying to translate installed modules. Any chance of a follow up to this?

eternoendless:
@direwald This PR fixes that: https://github.com/PrestaShop/PrestaShop/pull/15139

anthony3794:
I have a same problem with 1.7.6.1.
You have solution ?

PR #13059, opened 2019-03-26
Fallback on l() function in modules when calling trans() on legacy domains

mickaelandrieu:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | See #13039
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #13039 but probably most of the issues related to the translations management with #11321
| How to test?  | Look at the tests, and the provided module "translations" module.

Already tested so far:

#### Results

Got an infinite loop as l() also fallback on trans(), see in [Translate:getModuleTranslation](https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Translate.php#L280).

#### Results

Got 4 problems so far:

1. The Symfony Component expects files to respect this form: `<Domain>.<locale>.<extension>`, but legacy files are using this form: `<prestaShopLocale>.php`, I'll probably need to override low-level behavior of the component, still investigating on how to do it (almost no docs!);

2. There is already an existing Translation Loader that deals with the PHP extension which is conflicting with my tests to reuse the Legacy File Loader;

3. The way I use the loader for my previous needs is slightly different from the current use: for instance, the loader needs the complete file path when - in my previous case - I only send the translations folder path;

4. We have 2 translators instances in the application! I couldn't figure out why, why, WHY I retrieve some entries from my modules translations files but still not the translated content: see [the PrestaShop one](https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Translation/Translator.php) and the [Legacy? one](https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Context.php#L377). My "guess" is that I have to do the job twice while managing to pass the right "locale|PrestaShopLocale" to the right functions...

#### More thoughts on 1)

Another issue with the support of legacy file system is that we try to implement a Domain non-deterministic system while the Symfony Component is Domain deterministic. In a previous pull request, I had to cheat and determine the domain from the module name (this also means that PrestaShop developers can't choose their own domain name, weird...).

I've also had to break the existing architecture based on 2 constraints:

[...]

PR #14966, opened 2019-08-02
Keep BO from using two different translators in parallel 

### LINKED ISSUE #14932: Database translations not taken into account by modules
**Description**
Modules translations stored in database are not taken into account. So users can't modify their module's translations.

**To Reproduce**
1. Go to international > translations > installed modules translations
2. choose a module, change its translations
3. go to the modified module's configuration and check that wording has changed

**Additional information**
PrestaShop version: 1.7.6.x

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Fix translation issues, read below
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14932
| How to test?  | Check that translation is still working, especially translations coming from database in modules.

First, you should read this PR from @eternoendless 
https://github.com/PrestaShop/PrestaShop/pull/14845

His PR renames translation test modules, allows underscores in translation domain name, and adds useful comments to understand the two translators.

From there, this PR adds a few things, the main goal being that wether we are in FO or BO, we always use **only one translator**, either the translator component or the one from frameworkbundle. Also make it so that we still have the translation part in symfony's debug toolbar, and that the database translations are still working in modules.

- make frameworkbundle's translator load SQL translations with yaml service configuration
- fix timing issue in controllerCore (service container was called after the first call to the translator) . (edit: removed due to issue with php 5.6)  
- Fetch the right translator depending on the availability or not of the symfony service container
- Make it possible to specify which translator you need

I carefully checked that the component translator is never called in back office, and that the framework bundle translator is never called in front office.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14966)

### DIFF
--- a/classes/Context.php
+++ b/classes/Context.php
@@ -23,6 +23,7 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
+use PrestaShop\PrestaShop\Adapter\SymfonyContainer;
 use PrestaShop\PrestaShop\Core\Localization\Locale;
 use PrestaShopBundle\Translation\Loader\SqlTranslationLoader;
 use PrestaShopBundle\Translation\TranslatorComponent as Translator;
@@ -370,10 +371,16 @@ class ContextCore
             return $this->translator;
         }
 
-        $translator = $this->getTranslatorFromLocale($this->language->locale);
-        $this->translator = $translator;
+        $sfContainer = SymfonyContainer::getInstance();
 
-        return $translator;
+        if (null === $sfContainer) {
+            // symfony's container isn't available in front office, so we load and configure the translator component
+            $this->translator = $this->getTranslatorFromLocale($this->language->locale);
+        } else {
+            $this->translator = $sfContainer->get('translator');
+        }
+
+        return $this->translator;
     }
 
     /**
--- a/classes/Context.php
+++ b/classes/Context.php
@@ -365,7 +365,7 @@ class ContextCore
     /**
      * @return Translator
      */
-    public function getTranslator()
+    public function getTranslator($needComponentTranslator = false)
     {
         if (null !== $this->translator) {
             return $this->translator;
@@ -373,11 +373,12 @@ class ContextCore
 
         $sfContainer = SymfonyContainer::getInstance();
 
-        if (null === $sfContainer) {
+        if ($needComponentTranslator || null === $sfContainer) {
             // symfony's container isn't available in front office, so we load and configure the translator component
             $this->translator = $this->getTranslatorFromLocale($this->language->locale);
         } else {
             $this->translator = $sfContainer->get('translator');
+            $this->translator->setLocale($this->language->locale);
         }
 
         return $this->translator;
--- a/classes/Context.php
+++ b/classes/Context.php
@@ -363,9 +363,14 @@ class ContextCore
     }
 
     /**
+     * Returns a translator depending on service container availability and if the method
+     * is called by the installer or not.
+     *
+     * @param bool $isInstaller Set to true if the method is called by the installer
+     *
      * @return Translator
      */
-    public function getTranslator($needComponentTranslator = false)
+    public function getTranslator($isInstaller = false)
     {
         if (null !== $this->translator) {
             return $this->translator;
@@ -373,11 +378,13 @@ class ContextCore
 
         $sfContainer = SymfonyContainer::getInstance();
 
-        if ($needComponentTranslator || null === $sfContainer) {
+        if ($isInstaller || null === $sfContainer) {
             // symfony's container isn't available in front office, so we load and configure the translator component
             $this->translator = $this->getTranslatorFromLocale($this->language->locale);
         } else {
             $this->translator = $sfContainer->get('translator');
+            // We need to set the locale here because in legacy BO pages, the translator is used
+            // before the TranslatorListener does its job of setting the locale according to the Request object
             $this->translator->setLocale($this->language->locale);
         }
 
--- a/classes/controller/Controller.php
+++ b/classes/controller/Controller.php
@@ -180,6 +180,10 @@ abstract class ControllerCore
             define('_PS_BASE_URL_SSL_', Tools::getShopDomainSsl(true));
         }
 
+        if (null === $this->getContainer()) {
+            $this->container = $this->buildContainer();
+        }
+
         $localeRepo = $this->get(self::SERVICE_LOCALE_REPOSITORY);
         $this->context->currentLocale = $localeRepo->getLocale(
             $this->context->language->getLocale()
@@ -225,9 +229,6 @@ abstract class ControllerCore
         }
         if (null === $this->display_footer) {
             $this->display_footer = true;
-        }
-        if (null === $this->getContainer()) {
-            $this->container = $this->buildContainer();
         }
         $this->context = Context::getContext();
         $this->context->controller = $this;
--- a/classes/controller/Controller.php
+++ b/classes/controller/Controller.php
@@ -180,9 +180,6 @@ abstract class ControllerCore
             define('_PS_BASE_URL_SSL_', Tools::getShopDomainSsl(true));
         }
 
-        if (null === $this->getContainer()) {
-            $this->container = $this->buildContainer();
-        }
         $localeRepo = $this->get(self::SERVICE_LOCALE_REPOSITORY);
         $this->context->currentLocale = $localeRepo->getLocale(
             $this->context->language->getLocale()
@@ -228,6 +225,9 @@ abstract class ControllerCore
         }
         if (null === $this->display_footer) {
             $this->display_footer = true;
+        }
+        if (null === $this->getContainer()) {
+            $this->container = $this->buildContainer();
         }
         $this->context = Context::getContext();
         $this->context->controller = $this;
--- a/install-dev/classes/controllerHttp.php
+++ b/install-dev/classes/controllerHttp.php
@@ -143,7 +143,7 @@ class InstallControllerHttp
             $this->session->lang ?: false
         );
 
-        $this->translator = Context::getContext()->getTranslator();
+        $this->translator = Context::getContext()->getTranslator(true);
 
         if (isset($this->session->lang)) {
             $lang = $this->session->lang;
--- a/src/PrestaShopBundle/Resources/config/services/bundle/translation.yml
+++ b/src/PrestaShopBundle/Resources/config/services/bundle/translation.yml
@@ -108,6 +108,11 @@ services:
         tags:
             - {name: translation.loader, alias: db}
 
+    prestashop.translation.sql_loader:
+        class: PrestaShopBundle\Translation\Loader\SqlTranslationLoader
+        tags:
+            - {name: translation.loader, alias: db}
+
     prestashop.translation.legacy_file_reader:
         class: PrestaShopBundle\Translation\Loader\LegacyFileReader
         arguments:


classes/controller/Controller.php
```diff
@@ -180,9 +180,6 @@ public function init()
             define('_PS_BASE_URL_SSL_', Tools::getShopDomainSsl(true));
         }
 
-        if (null === $this->getContainer()) {
-            $this->container = $this->buildContainer();
-        }
         $localeRepo = $this->get(self::SERVICE_LOCALE_REPOSITORY);
```
mickaelandrieu, line 187: you call the container before its build, isn't it ?

src/PrestaShopBundle/Translation/Translator.php
```diff
@@ -29,6 +29,9 @@
 
 use Symfony\Bundle\FrameworkBundle\Translation\Translator as BaseTranslator;
 
+/**
+ * Replaces the original Symfony FrameworkBundle translator
```
mickaelandrieu: I'm not sure this PHPDoc is true, it's the compiler pass that do the replacement.

tests/Integration/PrestaShopBundle/Translation/Provider/ExternalModuleLegacySystemProviderTest.php
```diff
@@ -37,7 +37,7 @@
  */
 class ExternalModuleLegacySystemProviderTest extends KernelTestCase
 {
-    const MODULE_NAME = 'translation_test';
+    const MODULE_NAME = 'translationtest';
```
mickaelandrieu: underscore are valid in module name: why do we need to change that ?

tests/Resources/modules/translationtest/views/templates/admin/index.html.twig
```diff
@@ -4,7 +4,7 @@
 {% block content %}
     <div class="row">
         <div class="col-lg-8">
-            <h2>{{ 'Hello World'|trans({}, 'Modules.Translationtest.Translationtest') }}</h2>
+            <h2>{{ 'Hello World'|trans({}, 'Modules.Translationtest.Translation_test') }}</h2>
```
mickaelandrieu: why do we have an underscore here? What is/are the rules to guess which domain name we should use in this case ?

classes/controller/Controller.php
matthieu-rolland: yes, I call the container builder so that its built. That's what was done, I just moved this process from the init() method to the constructor.

src/PrestaShopBundle/Translation/View/TreeBuilder.php
```diff
@@ -137,11 +137,12 @@ public function makeTranslationsTree($catalogue)
 
         foreach ($catalogue as $domain => $messages) {
             $tableisedDomain = Inflector::tableize($domain);
-            $parts = array_reverse(explode('_', $tableisedDomain));
+            // the third component of the domain may have underscores, so we need to limit pieces to 3
```
mickaelandrieu: can we export this business logic into a specific class/function to figure out what is done here ?

as a basic rule, everytime I want to comment a block of code, I ask myself: why not write a new function instead with a meaningful name and unit tests ?

This is not naive here, I've already contributed this code and this patch before (but the PR [was not merged](https://github.com/PrestaShop/PrestaShop/pull/13059/files#diff-f22a4655c95d20014e673956d4511214R141)).

mickaelandrieu (commented):
some minor comments

tests/Integration/PrestaShopBundle/Translation/Provider/ExternalModuleLegacySystemProviderTest.php
matthieu-rolland: As stated in the PR's description, you should really read @eternoendless 's PR . ;)

https://github.com/PrestaShop/PrestaShop/pull/14845

classes/Context.php
```diff
@@ -370,16 +371,22 @@ public function getTranslator()
             return $this->translator;
         }
 
-        $translator = $this->getTranslatorFromLocale($this->language->locale);
-        $this->translator = $translator;
+        $sfContainer = SymfonyContainer::getInstance();
```
matthieu-rolland, line 379: You'll notice that I didn't use the **ContainerFinder**, the reason is that in its **getContainer** method, the symfony container comes last, so I ended up having another service container that doesn't have the translator service, even though the translator service is actually available via the Symfony container.

tests/Resources/modules/translationtest/src/Controller/Admin/ExempleController.php
```diff
@@ -24,15 +24,15 @@
...
 
 use PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController;
 
 class ExempleController extends FrameworkBundleAdminController
 {
     public function indexAction()
     {
-        return $this->render('@Modules/translations/views/templates/admin/index.html.twig', [
+        return $this->render('@Modules/translationtest/views/templates/admin/index.html.twig', [
             'layoutTitle' => $this->trans('Modern controller', 'Modules.Translationtest.Admin'),
```
LouiseBonnard: What is it?

tests/Resources/modules/translationtest/views/templates/admin/index.html.twig
LouiseBonnard: Yep, same question.

tests/Resources/modules/translationtest/views/templates/admin/someFile.with-things.tpl
```diff
@@ -0,0 +1 @@
+<h1>{l s='Smarty template' mod='translationtest'}</h1>
```
LouiseBonnard: Is it just for test purposes?

tests/Resources/modules/translationtest/src/Controller/Admin/ExempleController.php
PierreRambaud: These, are tests resources, you can ignore everything in tests directory @LouiseBonnard  =)

tests/Resources/modules/translationtest/views/templates/admin/someFile.with-things.tpl
matthieu-rolland: yes, this is the "translationtest" module, it's only for tests @LouiseBonnard

LouiseBonnard: Okay, so it won't be part of the Crowdin catalog? I just want to be sure. :-)

matthieu-rolland: Nope it shouldn't...  👍

classes/Context.php
eternoendless: That's weird, there shouldn't be more than one container in each environment. Maybe that's pointing to another bug somewhere else?

eternoendless: Oh you meant in the front office? Then maybe we should declare it as a FO service, don't you think?

src/PrestaShopBundle/Translation/Translator.php
eternoendless: I didn't think it could be interpreted that way. Here's a better wording:

```suggestion
 * Replacement for the original Symfony FrameworkBundle translator
```

src/PrestaShopBundle/Translation/TranslatorComponent.php
```diff
@@ -29,6 +29,9 @@
 
 use Symfony\Component\Translation\Translator as BaseTranslatorComponent;
 
+/**
+ * Translator used by Context
```
eternoendless: If I'm not mistaken now it's only used in FO right?

```suggestion
 * Translator used in FO
```

src/PrestaShopBundle/Translation/View/TreeBuilder.php
eternoendless: I think that's a different subject. This fix exists solely to allow underscores in the third part of a translation domain.

Other improvements can be done, for sure, although on a different PR.

tests/Integration/PrestaShopBundle/Translation/Provider/ExternalModuleLegacySystemProviderTest.php
eternoendless: Underscores are *NOT* valid and *WON'T* work.

tests/Resources/modules/translationtest/views/templates/admin/index.html.twig
eternoendless: @LouiseBonnard You can ignore this, they are test resources.

@mickaelandrieu It belongs to the legacy fallback system, where you declare a domain and use already existing translations savec in the legacy format. In order for the "use existing translations" part to work, the domain name must follow some very specific rules, but basically: `Modules.<Nameofthemodule>.<Name_of_the_file_without_extension>`. This is so, specially the underscore part, because translations are stored that way in the legacy files.

Rest assured, this is all being documented and will be published soon.

tests/Integration/PrestaShopBundle/Translation/Provider/ExternalModuleLegacySystemProviderTest.php
eternoendless: I mean, you can have underscores in your module name but you won't be able to translate it. Here's why:

Given a module `my_module`, its domain name would have to be either `MyModule` or `Mymodule`, because underscores are not allowed (at least not in that part). 

`MyModule` wouldn't be valid because internally domains are stored without dots, meaning that `MyModule` would be broken down into a subdomain like `My.Module`. `Mymodule` won't work either because we have no way of transforming `Mymodule` back into `my_module` to find the module directory and read its translation files.

classes/Context.php
matthieu-rolland: No I mean in back office, the ContainerFinder gives me an instance of **AdminContainer** instead of giving me a **SymfonyContainer**...

matthieu-rolland: An issue has been created regarding this service container issue:

https://github.com/PrestaShop/PrestaShop/issues/14995

matthieu-rolland:
I removed the part where I changed the place where the container was called for the first time, in
the **classes/controller/Controller.php**

The reason is that it creates a bug, only on PHP 5.6 when opcache is disabled.

Capture d’écran 2019-08-06 à 15 15 20

I know, it's weird, it looks a lot like this bug:

https://bugs.php.net/bug.php?id=66773

> opcache allows importing a class from another namespace with a name that is already in use in the current namespace

Why only in php 5.6? I dont know...

eternoendless:
> [2017-09-05 18:19 UTC] nikic@php.net
> This issue has been fixed in PHP 7 only.

It seems that they fixed it only in PHP 7 😅

classes/Context.php
```diff
@@ -370,10 +371,16 @@ public function getTranslator()
...
 
-        return $translator;
+        if (null === $sfContainer) {
+            // symfony's container isn't available in front office, so we load and configure the translator component
+            $this->translator = $this->getTranslatorFromLocale($this->language->locale);
```
eternoendless, line 383: Can you confirm this method is _NEVER_ called in the BO context? (either legacy controller or symfony controller)

matthieu-rolland: Yes, I carefully checked that we never enter this method in BO context, furthermore, I checked that the **TranslatorComponent** class is never instantiated in BO context.

The reason it works like this is that the SymfonyContainer is always available in BO context, in legacy and migrated pages alike.

matthieu-rolland: Maybe it would be nice to improve the condition and check for BO context? I'm not sure about the proper way to do that though... ❓ 
I the mean time I like this condition, because internally, it is really the availability or not of the **SymfonyContainer** (and the symfony kernel) that implies which Translator will be used.

eternoendless: Okay!

eternoendless (dismissed):
LGTM, QA label can be added once tests are green

khouloudbelguith:
Hi @matthieu-rolland,

I tried with different modules for example `productcomments` & `ps_mainmenu`.
The shop installed with an English language/France country & The employee language => French.
I would like to translate some expression's module from the back office in French => not translated.
In the FO => ok.
https://drive.google.com/file/d/16haO3ZAwdtaG0iP1PhzyRyiHG41s21Hr/view
When the employee's language is English => Translations => OK.
Thanks!

matthieu-rolland:
My mistake @khouloudbelguith , I should have waited before sending this PR to the QA...
Why? Because the BO module translation problem will be fixed in this PR from @eternoendless :
https://github.com/PrestaShop/PrestaShop/pull/14962

So I will wait until @eternoendless 's PR is merged, and then I will send this PR back to QA.

Sorry about this 🤦‍♂

classes/Context.php
```diff
@@ -364,16 +365,23 @@ public function updateCustomer(Customer $customer)
     /**
      * @return Translator
      */
-    public function getTranslator()
+    public function getTranslator($needComponentTranslator = false)
```
matthieu-rolland: I made it possible to tell via a parameter which translator you need, because during installation it would get the framework bundle translator, which is configured to init the db sql loader, but the database is not ready yet, so it fails...

classes/Context.php
```diff
@@ -364,16 +365,23 @@ public function updateCustomer(Customer $customer)
...
-        $this->translator = $translator;
+        $sfContainer = SymfonyContainer::getInstance();
 
-        return $translator;
+        if ($needComponentTranslator || null === $sfContainer) {
+            // symfony's container isn't available in front office, so we load and configure the translator component
+            $this->translator = $this->getTranslatorFromLocale($this->language->locale);
+        } else {
+            $this->translator = $sfContainer->get('translator');
+            $this->translator->setLocale($this->language->locale);
```
matthieu-rolland, line 388: On legacy back office pages, the translator is called before listeners do their job of changing the locale.

For example the LegacyContext is called in the constructor of UserLocaleListener, and it instantiates AdminController who himself has translations done in its constructor. 

This is why we need to set the locale when initiating the translator, this is only necessary for legacy back office pages...

classes/Context.php
eternoendless: I don't think this is a good idea, because

1. TranslatorComponent is meant to go away eventually
2. It exposes internal functionality
3. It's impossible tu understand if you don't know the underlying complexities 🗡 

If this is a special case just for the installer, let's just own it up:

```suggestion
    public function getTranslator($isInstaller = false)
```

or alternatively:

```suggestion
    public function getTranslator($withDatabase = true)
```

In any case, this parameter must be declared and explained in Phpdoc

classes/Context.php
eternoendless: It's better to add this comment in the code.

eternoendless (changes requested):

classes/Context.php
matthieu-rolland: done !

classes/Context.php
matthieu-rolland: done !

eternoendless (approved):
Looks good now, thanks!

matks:
Thank you @matthieu-rolland, @mickaelandrieu and @eternoendless

Issue #10552, opened 2018-09-20
Migrate "Catalog > Monitoring" page


matks:
Part of Symfony migration project
Type: page

## Status

Todo:
- [x] Migrate all grids of the page https://github.com/PrestaShop/PrestaShop/pull/13529
- [x] Implement product actions usable in grid (see `@todo` in https://github.com/PrestaShop/PrestaShop/pull/13529/files) https://github.com/PrestaShop/PrestaShop/pull/16267
- [x] Fix monitoring grids id configuration https://github.com/PrestaShop/PrestaShop/pull/15757
- [x] Add Helper card (see below) https://github.com/PrestaShop/PrestaShop/pull/16267
- [x] Fix old issue (see below): remove a button
- [x] Unhide page https://github.com/PrestaShop/PrestaShop/pull/16588

Old issue: https://github.com/PrestaShop/PrestaShop/issues/11482

#### Controller / routing
- [ ] All controller actions are migrated and protected with `@AdminSecurity` annotations when eligible and `@DemoRestricted` annotations when eligible
- [ ] All links targeting this page are updated: either using _legacy_link or updating the code
- [ ] The new SF routes have `_legacy_link` and `_legacy_parameters` provided and explained
- [ ] Controller catches Domain Exception and translate them nicely
#### Template
- [ ] Twig templates must contain relevant blocks to allow partial extension by modules
- [ ] Twig forms must have `form_rest()` and `form_errors()`
- [ ] JS being used on the page can be used as enabled extensions and use a PageMap for selectors
- [ ] Use `@PrestaShop` instead of `PrestaShopBundle` for template root path aliasing
##### Conventions
- [ ] The new controller actions comply with our naming convention
- [ ] The new SF routes names and paths comply with our naming convention
- [ ] The new Twig template names comply actions with our naming convention
- [ ] Created files (controller, JS assets, CSS assets and twig views) must be in the right folder following the convention
- [ ] CSS being added must follow theme conventions

#### Core layer
- [ ] Commands and Queries only use native types but internally use VOs
- [ ] All Handlers must have an Interface
##### Conventions
- [ ] The created SF service names comply with our naming convention
- [ ] The created files must follow the folder convention (Domain, Core, Adapter...)

#### Polishing
- [ ] The legacy controller and templates files are gone
- [ ] If PR includes a new concept/mechanism, documentation must be written
- [ ] JS and CSS assets must be compiled and up-to-date
- [ ] License headers must be up-to-date
[...]

PR #14991, opened 2019-08-05
Fix Monitoring grids data for multishop

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | When multiple shops exists, but context is single shop - most of migrated Monitoring lists where not considering the shop context id when getting data.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #10552
| How to test?  | Use `/admin-dev/index.php/sell/catalog/monitoring` to see migrated monitoring page. When multiple shops are available, select one of them and make sure that all monitoring grids are showing data just for that shop. When All shops are enabled, grids should contain data for default shop. :warning: Actions like edit/delete are not migrated within this PR, because it is highly coupled with product page

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14991)

### DIFF
--- a/src/Core/Grid/Query/Monitoring/AbstractProductQueryBuilder.php
+++ b/src/Core/Grid/Query/Monitoring/AbstractProductQueryBuilder.php
@@ -127,6 +127,10 @@ abstract class AbstractProductQueryBuilder extends AbstractDoctrineQueryBuilder
                 'p.id_product = ps.id_product AND ps.id_shop = p.id_shop_default'
         );
 
+        if ($isSingleShopContext) {
+            $qb->andWhere('ps.id_shop = :context_shop_id');
+        }
+
         $this->applyFilters($qb, $searchCriteria->getFilters());
 
         return $qb;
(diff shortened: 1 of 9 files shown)


matks (dismissed):

sarahdib:
Hello @zuk3975 

- In migrated page the product appear  in the second shop but I created this one only for the first shop

image

image

---------------------------------------------------------------------------------------------------

- When I search by id there is no result if the id isn't the same. For example if I search 3, 23 need to be in the list 

image

image

---------------------------------------------------------------------------------------------------

- Delete and edit button not working but I think it's normal. Those action aren't migrated yet ?

zuk3975:
Thanks @sarahdib,

- I've fixed the first issue with product shown in wrong shop.

- `When I search by id there is no result if the id isn't the same. For example if I search 3, 23 need to be in the list` - it is expected behavior, migrated lists filters id by exact value.

- `Delete and edit button not working but I think it's normal. Those action aren't migrated yet` - yes, not migrated, so it is normal :)

sarahdib:
Thanks @zuk3975 I'm waiting for a review and I will test it :)

zuk3975:
> Thanks @zuk3975 I'm waiting for a review and I will test it :)

@matks, could you check and also restart travis, please? :)

matks (approved):

matks:
Here we go ! 😄

matks:
PrettyCI is lost 🙄

matks:
PrettyCI is now happy 😊thank you @mnapoli

matks:
Thank you @zuk3975

PR #15008, opened 2019-08-06
Inline styles should be loaded from project root

yisraeldov:
Right now inline styles are not loaded because they can not be found. 

They are trying to load from the file system root not the presta shop root. 

eg

```
 file_get_contents(/themes/mytheme/assets/css/Inline.css): failed to open stream: No such file or directory in /var/www/html/classes/assets/StylesheetManager.php on line 134
```

| Questions     | Answers
| Branch?       | develop / 1.7.6.x / 1.6.1.x
| Category?     | FO / BO / CO / IN / WS / TE
| BC breaks?    | Does it break backward compatibility? yes/no
| Deprecations? | Does it deprecate an existing feature? yes/no

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15008)

### DIFF
--- a/classes/assets/StylesheetManager.php
+++ b/classes/assets/StylesheetManager.php
@@ -137,7 +137,7 @@ class StylesheetManagerCore extends AbstractAssetManager
         foreach ($this->list['inline'] as &$item) {
             $item['content'] =
                 '/* ---- ' . $item['id'] . ' @ ' . $item['path'] . ' ---- */' . "\r\n" .
-                file_get_contents($item['path']);
+                file_get_contents("." . $item['path']);
         }
     }
 }


classes/assets/StylesheetManager.php
```diff
@@ -137,7 +137,7 @@ private function addInlinedStyleContent()
         foreach ($this->list['inline'] as &$item) {
             $item['content'] =
                 '/* ---- ' . $item['id'] . ' @ ' . $item['path'] . ' ---- */' . "\r\n" .
-                file_get_contents($item['path']);
+                file_get_contents("." . $item['path']);
```
MathiasReker, line 140:
```suggestion
                file_get_contents('.' . $item['path']);
```

PierreRambaud:
Hi and thanks for your contribution. Could you please run cs-fixer and complete the table description?

matks:
As author is not answering, I guess this PR got stale.

Feel free to reopen if you want to discuss this PR further (and get it merged) 😉

PR #15016, opened 2019-08-07
Remove unnecessary code in AdminController

Shagshag:
| Questions     | Answers
| Branch?       | develop
| Description?  | Remove unnecessary code
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | N/A

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15016)

### DIFF
--- a/classes/controller/AdminController.php
+++ b/classes/controller/AdminController.php
@@ -3340,7 +3340,7 @@ class AdminControllerCore extends Controller
      */
     protected function getFromClause()
     {
-        return "\n" . 'FROM `' . _DB_PREFIX_ . bqSQL($this->table) . '` a ';
+        return PHP_EOL . 'FROM `' . _DB_PREFIX_ . bqSQL($this->table) . '` a ';
     }
 
     /**
--- a/classes/controller/AdminController.php
+++ b/classes/controller/AdminController.php
@@ -3340,9 +3340,7 @@ class AdminControllerCore extends Controller
      */
     protected function getFromClause()
     {
-        $sql_table = $this->table == 'order' ? 'orders' : $this->table;
-
-        return "\n" . 'FROM `' . _DB_PREFIX_ . $sql_table . '` a ';
+        return "\n" . 'FROM `' . _DB_PREFIX_ . bqSQL($this->table) . '` a ';
     }
 
     /**


classes/controller/AdminController.php
```diff
@@ -3340,9 +3340,7 @@ public function getList(
      */
     protected function getFromClause()
     {
-        $sql_table = $this->table == 'order' ? 'orders' : $this->table;
-
-        return "\n" . 'FROM `' . _DB_PREFIX_ . $sql_table . '` a ';
+        return "\n" . 'FROM `' . _DB_PREFIX_ . bqSQL($this->table) . '` a ';
```
MathiasReker:
```suggestion
        return PHP_EOL . 'FROM `' . _DB_PREFIX_ . bqSQL($this->table) . '` a ';
```

matks (dismissed):

matks:
Thanks for the help @MathiasReker 😉 
@Shagshag can you explain why this code is unnecessary ? is it related to old PS versions ?

Shagshag:
@matks This code is used to correct a typing error in the table name: order instead of order**s**.
I don't think the code should fix developer errors. 
Moreover this correction appears only in AdminController, not in FrontController, Db... The developer who made the mistake in AdminController will not be able to do it elsewhere. 
Finally orders is not the only table with a "s"

matks:
🤔 I understand your point of view, and this is indeed a good thing to clean this flawed logic however I'm afraid we cannot remove it now.

We are trying to follow as much as possible SemVer, which means we are not supposed to create BC breaks. If a BO module currently creates a custom Controller and relies on this mechanism (so the module developer spelled incorrectly "orders" to "order") to work, removing this line of code will make his module break on the next minor release.

This PR would be nice if we were building the next major PrestaShop version, PS 8.0.0, but right now I'm afraid we have to keep this errored logic for now in order to provide continous behavior stability to PS ecosystem.

What we can do for now is, when "order" is used instead of "orders", to add an error message like this (wording needs improvement 😛) :
```
        @trigger_error(
            'you have attempted to perform a FROM SQL statement on table order which does not exist, the right table name is orders instead',
            E_USER_DEPRECATED
        );
```

But I cannot merge this PR right now, sorry 😞

matks:
Closing for now as it seems discussion ended, feel free to reopen if you want to discuss it further 😉

Issue #13845, opened 2019-05-17
The steps missing in welcome module


ntiepresta:
**Describe the bug**
The welcome module isn't complete in the BO => Payments =>Payment Methods, , the step  4/6( Get your shop ready for payments) the steps 1/2 and 2/2 missing 

**To Reproduce**

1. BO => International => Locations => Countries Tab => Edit an enabled country (France)
2. Add the Customer fields in the Address format 
1. Go to  BO => Modules => Module Manager ( Reset welcome module)
2. Click on yes, reset it
3. refresh your shop and click on resume in configure tab
4. click on start and start the steps.

**Screenshots**
Before 
![avant1](https://user-images.githubusercontent.com/36889855/57942166-122dcc00-78d1-11e9-9333-8f85375c357d.png)

![avant2](https://user-images.githubusercontent.com/36889855/57942196-25d93280-78d1-11e9-8f60-88fc967f8407.png)

After
![now1](https://user-images.githubusercontent.com/36889855/57942174-178b1680-78d1-11e9-9c51-353c270bc87f.png)

![now2](https://user-images.githubusercontent.com/36889855/57942206-2bcf1380-78d1-11e9-9c5a-b1e92f373966.png)

**Additionnal information**
PrestaShop version: N/A
PHP version: N/A

marionf:
The issue is the following:
At the first step of the welcome module you are at 2/6 instead of 1/6

![capture d'écran_1568](https://user-images.githubusercontent.com/13449658/58036730-1d306880-7b2c-11e9-9e71-d365f9bd2087.png)

But this issue is also on 1.7.5.2 so it's not a regression

marionf:
@ntiepresta 

I correctly have steps 2 on payment method and carrier pages.

![capture d'écran_1569](https://user-images.githubusercontent.com/13449658/58037242-443b6a00-7b2d-11e9-8e2d-40aa61ae19ba.png)

These parts are managed by MBO, so please verify you have MBO module installed
Furthermore, you need to wait a few seconds before to see them displayed

If MBO isn't installed, it's normal if you don't see them

ntiepresta:
Thank your @marionf 

You are right for this issue, if the ps_MBO is installed we don't have this issue. In addition, the ps_mbo is not installed by default on  fresh install, and ps_bestsellers for example. The ps_mbo should installed by default.
Thanks,

marionf:
Because modules have not been updated yet for 1.7.6

PR #15017, opened 2019-08-07
Improve translation test module

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Read below
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | n/a
| How to test?  | Read below

## What's this about

This PR replaces #13845 and solves the two first problems noted previously:

1. The test module `translation_test` has been renamed to `translationtest`, since translations cannot work on modules with underscores in their name. Indeed, the domain name for `translation_test` would be `Modules.Translationtest`, which can never be reversed back to `translation_test` to find the original module. [This required updating TranslationToolsBundle](https://github.com/PrestaShop/TranslationToolsBundle/pull/64).
2. The third component of a translation domain, in the case of modules, is the original file name. Many file names contain underscore, but it wasn't supported because the translation interface in BO used Inflector::tableize to separate the domain parts (yes domains are stored without dots in the translator dictionary, but that's another story). This has been fixed in [this commit](https://github.com/PrestaShop/PrestaShop/commit/9f1c2fe9f070a8952eff1bea8eb6ebbc229954d5)

In addition, this change also allows Module AdminControllers to reference admin templates (ie. templates located in `mymodule/views/templates/admin/`) without using the full route to the template ([see change](https://github.com/PrestaShop/PrestaShop/commit/504bbd558f895d6fb9a8d362afde5ac4f3a7112c)).

## How to test

1. Copy the `translationtest` module located in `tests/Resources/modules` over to the `/modules` folder. 
1. Go to modules catalog, search "translation" and install it.
1. Configure the module, you should see a simple page containing the wording "This wording belongs to the module file".
1. Go to `/admin-dev/index.php/modules/translations`. If your shop is in French you should get a screen with two wordings "Contrôleur moderne" and "Bonjour le monde"
1. Go to `/admin-dev/index.php?controller=AdminTranslationtestFoo`. If your shop is in French you should get a page containing the wording "Le template Smarty"
1. Go to `index.php?fc=module&module=translationtest&controller=bar`. If your shop is in French you should get a blank page containing only the wording "Ceci est un contrôleur frontal"

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15017)


src/PrestaShopBundle/Translation/Translator.php
```diff
@@ -29,6 +29,9 @@
 
 use Symfony\Bundle\FrameworkBundle\Translation\Translator as BaseTranslator;
 
+/**
+ * Replaces the original Symfony FrameworkBundle translator
+ */
```
matthieu-rolland, line 34: I think you wanted to write "Replacement for the original Symfony FrameworkBundle translator" instead:

https://github.com/PrestaShop/PrestaShop/pull/14966#discussion_r310576557

tests/Resources/modules/translationtest/vendor/composer/ClassLoader.php
```diff
@@ -279,7 +279,7 @@ public function isClassMapAuthoritative()
      */
     public function setApcuPrefix($apcuPrefix)
     {
-        $this->apcuPrefix = function_exists('apcu_fetch') && ini_get('apc.enabled') ? $apcuPrefix : null;
+        $this->apcuPrefix = function_exists('apcu_fetch') && filter_var(ini_get('apc.enabled'), FILTER_VALIDATE_BOOLEAN) ? $apcuPrefix : null;
```
matthieu-rolland, line 282: Interesting way of checking a php config value (**FILTER_VALIDATE_BOOLEAN**), never saw it before 👍

eternoendless: That was composer, not me, but interesting nonetheless!

src/PrestaShopBundle/Translation/Translator.php
eternoendless: Right!

matthieu-rolland (approved):
LGTM

jolelievre:
Thanks @eternoendless

PR #15030, opened 2019-08-08
Get all categories in category tree form (not only enabled ones)

### LINKED ISSUE #14790: 1.7.6.0 BO: A disabled category is not displayed in the "Parent category" section
Hi,
I'm in PS 1.7.6.0.
I try to update a category with meta description and simplified URL and the message "Une erreur inattendue s'est produite. [PrestaShop\PrestaShop\Core\Domain\Category\Exception\CategoryConstraintException code 6]".
I tried both on google chrome and microsoft edge.
Can you help me please?
Thanks!
>[Edited by Khouloud]
A disabled category is not displayed in the "Parent category" section, which is not the same behavior as PS1.7.5.2.
To reproduce, go to Categories -> Home Edit and set "Displayed" to "No". Then try to add a new category.
If the Home category is disabled, we cannot add or edit categories.

Arnaud

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Include disabled categories in category tree.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | fixes #14790
| How to test?  | see #14790

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15030)

### DIFF
--- a/src/Adapter/Form/ChoiceProvider/CategoryTreeChoiceProvider.php
+++ b/src/Adapter/Form/ChoiceProvider/CategoryTreeChoiceProvider.php
@@ -59,7 +59,7 @@ final class CategoryTreeChoiceProvider implements FormChoiceProviderInterface
      */
     public function getChoices()
     {
-        $categories = $this->categoryDataProvider->getNestedCategories($this->contextShopRootCategoryId);
+        $categories = $this->categoryDataProvider->getNestedCategories($this->contextShopRootCategoryId, false, false);
         $choices = [];
 
         foreach ($categories as $category) {
--- a/src/Adapter/Form/ChoiceProvider/CategoryTreeChoiceProvider.php
+++ b/src/Adapter/Form/ChoiceProvider/CategoryTreeChoiceProvider.php
@@ -45,14 +45,14 @@ final class CategoryTreeChoiceProvider implements FormChoiceProviderInterface
     private $contextShopRootCategoryId;
 
     /**
-     * @var boolean
+     * @var bool
      */
     private $enabledCategoriesOnly;
 
     /**
      * @param CategoryDataProvider $categoryDataProvider
-     * @param $contextShopRootCategoryId
-     * @param boolean $enabledCategoriesOnly
+     * @param int $contextShopRootCategoryId
+     * @param bool $enabledCategoriesOnly
      */
     public function __construct(CategoryDataProvider $categoryDataProvider, $contextShopRootCategoryId, $enabledCategoriesOnly = false)
     {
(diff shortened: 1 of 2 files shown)


matthieu-rolland:
I could make this simple modification because this CategoryTreeChoiceProvider is only used by one service, which is on only used in the category form.

In the future if we need to have another category tree that doesn't take disabled categories into account, then we'll have to make that optional via the constructor, and configure it in 
**src/PrestaShopBundle/Resources/config/services/adapter/form.yml**

It just felt like premature optimization at this point, don't hesitate to tell me if you think otherwise.

eternoendless:
I bet this will become a problem once we try and use the same provider elsewhere. I think we should add the option in the constructor. Remember that other people may be using this class for purposes unknown to us.

eternoendless (dismissed):
Please add the option in the constructor

matthieu-rolland:
done !

src/Adapter/Form/ChoiceProvider/CategoryTreeChoiceProvider.php
```diff
@@ -44,22 +44,29 @@ final class CategoryTreeChoiceProvider implements FormChoiceProviderInterface
...
+     */
+    private $enabledCategoriesOnly;
+
     /**
      * @param CategoryDataProvider $categoryDataProvider
      * @param $contextShopRootCategoryId
+     * @param boolean $enabledCategoriesOnly
      */
-    public function __construct(CategoryDataProvider $categoryDataProvider, $contextShopRootCategoryId)
+    public function __construct(CategoryDataProvider $categoryDataProvider, $contextShopRootCategoryId, $enabledCategoriesOnly = false)
```
jolelievre, line 57: I think to avoid any misunderstanding if someone uses this again, you should not have a default value
`$enabledCategoriesOnly` should be required, and you then have to update the service definition, this way if someone copies it he can see all the arguments

jolelievre (dismissed):

src/Adapter/Form/ChoiceProvider/CategoryTreeChoiceProvider.php
eternoendless: >  you should not have a default value

While I agree in principle, adding a new parameter without default value would be a breaking change.

> you then have to update the service definition, this way if someone copies it he can see all the arguments 

+1 to that, we should update the service definition, even if it's using the same value as the default.

matthieu-rolland: done !

matks (dismissed):

matks:
@matthieu-rolland PR logic is ✅but codestyle is broken according to PrettyCI
Can you fix and update the PR ?

matks (approved):

eternoendless:
Thank you @matthieu-rolland

PR #15035, opened 2019-08-08
Fix for new product feature indexation

### LINKED ISSUE #15018: Product feature indexation
**Describe the bug**
Custom product feature is updarted in the search index only after second saving.

**To Reproduce**
1. Open the product in the backoffice (Catalog > Products)
2. Add feature: http://bit.ly/2KBfO9R and enter a custom value: http://bit.ly/2KymFk9
3. Save product
4. In database open ps_search_word, order by id_word, scroll to the bottom and you can see that the value is not saved: http://bit.ly/2KwUYIz
5. Return to the product edit page and save it second time. Repeat query in database and the value appears this time: http://bit.ly/2KwV8j9 Also only after the second saving the value is added to the ps_search_index too: http://bit.ly/2KA5r6h

The same behaviour if the value is deleted. You remove the feature, save product, the ps_search_index table still includes record with id_word value for this feature. You save the product again and only then the record is removed.

**Additional information**
PrestaShop version: 1.7.6.0 and older
PHP version: 7.1

reanim8ed:
| Questions     | Answers
| Branch?       | develop
| Description?  | getFeatures method is run before new custom values are saved in classes/Search.php @ 658 so these values aren't added to the the search index on the first time save. Added indexation method call straight after the feature is recorded to the DB.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15018 
| How to test?  | Add feature to the product and enter a custom value: http://bit.ly/2KymFk9 <br> Save product <br> In database open ps_search_word, order by id_word, scroll to the bottom and should see saved value

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15035)

### DIFF
--- a/controllers/admin/AdminProductsController.php
+++ b/controllers/admin/AdminProductsController.php
@@ -889,6 +889,7 @@ class AdminProductsControllerCore extends AdminController
                         }
                     }
                 }
+	            Search::indexation(false, $id_product);
             }
         } else {
             $this->errors[] = $this->trans('A product must be created before adding features.', array(), 'Admin.Catalog.Notification');
--- a/controllers/admin/AdminProductsController.php
+++ b/controllers/admin/AdminProductsController.php
@@ -889,6 +889,7 @@ class AdminProductsControllerCore extends AdminController
                         }
                     }
                 }
+                // additional indexation fixes issue where Search::getFeatures is run before new values are saved
 	            Search::indexation(false, $id_product);
             }
         } else {
--- a/controllers/admin/AdminProductsController.php
+++ b/controllers/admin/AdminProductsController.php
@@ -890,7 +890,7 @@ class AdminProductsControllerCore extends AdminController
                     }
                 }
                 // additional indexation fixes issue where Search::getFeatures is run before new values are saved
-	            Search::indexation(false, $id_product);
+                Search::indexation(false, $id_product);
             }
         } else {
             $this->errors[] = $this->trans('A product must be created before adding features.', array(), 'Admin.Catalog.Notification');


controllers/admin/AdminProductsController.php
```diff
@@ -889,6 +889,7 @@ public function processFeatures($id_product = null)
                         }
                     }
                 }
+	            Search::indexation(false, $id_product);
```
PierreRambaud: Could you please add a little comment to explain this line of code?

reanim8ed: Added additional indexation method straight after the new feature is recorded to the DB. It fixes the issue with Search::getFeatures method where it is run before new feature values are saved.

PierreRambaud: Sorry, I mean in the code :p as a phpdoc =)

Progi1984:
Hello @reanim8ed, thanks for your contribution, could you fix Pretty CI, please? Thanks

colinegin:
What should we do with this PR ? Can we handle it on our own if the author does not reply ?

Progi1984 (approved):

PierreRambaud (approved):

SD1982:
Hi @Progi1984  , @colinegin  It seems that this PR is no longer relevant because the "features values" are not stored in the "ps_search_word" anymore but in "ps_feature_value_lang" and so the issue #15018 seems to have been fixed with this database table changement .

colinegin:
I have no clue why we made this change. Any idea @Progi1984 ?
@SD1982 you confirm you can no longer reproduce the #15018 bug ?

SD1982:
Hi @colinegin YesI confirm that I have not been able to reproduce the issue and that this is certainly due to the change of schema of the database

NeOMakinG:
Hey @reanim8ed, regarding comments of @SD1982 and @colinegin, I'm closing the PR as it looks fixed due to another change, if you manage to reproduce it again, don't hesitate to open the PR again :)

Thanks for your contribution anyway, have a good day !

Issue #14846, opened 2019-07-24
#13440 creates a small BC break


rdy4ever:
**Describe the bug**
One little BC problem introduced in #13440 : when using a theme that doesn't include the button added in [personal-information.tpl](https://github.com/PrestaShop/PrestaShop/pull/13440/files#diff-936ab2713e0bb9cd8ac42716e0d33a37)  (from add-ons or other custom theme, or even Classic whithout the 1.7.6 update), if the user clicks on step one from another step (after step 1 is completed), then there's no way for him to go back (or forward, te be exact) to the next step, as this button is missing and he cannot click on the other steps. So he gets stuck and needs to refresh the page.

**To Reproduce**
Remove the button code from `themes/classic/templates/checkout/_partials/steps/personal-information.tpl`, so the file is as before #13440 and use the check-out. After completing step 1 as a logged-in user (create account, don't use the guest check-out), click on it to go back (as if you'd like to change personal info), but without the new `Continue` button you'll get stuck, no longer being able to continue to the next steps.

**Note:** Use the check-out as logged-in user, not as guest.

**Screenshots**
If applicable, add screenshots or screenrecords to help explain your problem.

**Additional information**
PrestaShop version: 1.7.6
PHP version: N/A

rdy4ever:
Ping @jolelievre

matks:
Hi @rdy4ever, thanks for the feedback ! @jolelievre is on vacations but we'll review this as BC breaks in the FO theme are a very important issue

marionf:
@PrestaShop/prestashop-core-developers should we fix it in 1.7.6.1 ?

jolelievre:
Hi @marionf yes I guess we should look for a way to avoid this BC break.

Issue #13126, opened 2019-04-01
Change checkout behavior


marionf:
**Describe the bug**
#12456 was not done completely
Here is what remains to be done:

The main difference vs current behaviour is that the user will always have to click on "Continue" to go from Step X to Step X+1, no matter if :

- he/she goes in the different steps for the first time
- he/she went to the step X and decided to go back to X-1 or X-2 to check or modify the information stored.

This means that :

- The next steps after the current step are no longer clickable (only Continue button allows to go to the next step)
- There should be a "Continue" button in the Personal information step (in case the logged in user goes back to the 1st step)
- Green checks and "Modify" CTAs should be removed in the steps following the user current one.

marionf:
The main issue is if you try to change your address without clicking on Continue button, it's not saved and y
and you're not delivered where you wanted

marionf:
@colinegin FYI

kpodemski:
@marionf 

> The main issue is if you try to change your address without clicking on Continue button, it's not saved and y and you're not delivered where you wanted

this is so true, we have this feedback a lot of times with a new checkout page

PR #15042, opened 2019-08-08
Only disable following steps in the checkout process when the current step has a continue button

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | In the checkout process only disable following steps when the current step has a continue button This will prevent to block the workflow in themes which don't have the continue button in the "Personal information" step (when logged in)
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/14846
| How to test?  | First you need to be logged in with a PrestaShop frontend user, then in the checkout process when you are (at least) at the address step you can go backward by clicking on the "Personal information" title<br /><br />Now this bug only occurs if you have a theme which doesn't include the continue button, so to simulate this you need to edit the template `themes/classic/templates/checkout/_partials/steps/personal-information.tpl` and remove continue button (the one with class `continue`)<br /><br />When you refresh and retry the previous actions on checkout page you will see that when you click on "Personal Information" title, the next steps are clickable but as soon as you go to the next step you need to click on Continue<br />So this fix shouldn't break the required change from https://github.com/PrestaShop/PrestaShop/issues/13126

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15042)


matthieu-rolland (dismissed):
lgtm

khouloudbelguith:
Hi @jolelievre,

In this case:
1. you are just logged in with a PrestaShop frontend user
2. Add some products to the cart
3. go to the checkout process
4. You are in "Addresses Step"
5. Click on the "Personal Information" step
6. The step "Addresses" is not disabled => OK but the edit button is not displayed
image

7. but, you cannot click on it => Error
8. if you reload the page => you are in the "Addresses Step" again
9. Click on Continue button => you are on the "Shipping Method" step
10. Click on the "Personal Information" step
11. The step "Addresses" is not disabled => OK & The Edit button is displayed
12. So you can click on the "Adress" Step
image

Thanks!

jolelievre:
@khouloudbelguith 
thanks, seems like I didn't test enough the use case when address is not set yet. I'm gonna try it!

jolelievre:
@khouloudbelguith 
Ok I fixed the bug, you can now click on Addresses when no "continue" button is present in the "Personal informations"

About the other steps you mentioned, when a step has already been completed and you click on continue it is logical that you move to the latest step, unless your modification requires for a change in another step (for example if you change your address you might need to redefine your shipping method)

khouloudbelguith:
@jolelievre, thanks!
It is OK now.
Waiting for review

themes/_core/js/checkout.js
```diff
@@ -51,9 +54,17 @@ function handleCheckoutStepChange() {
...
+          //If the step has no continue button, the previously completed steps are clickable
+          const $nextSteps = $clickedStep.nextAll('.checkout-step.-clickable');
+          $nextSteps.removeClass('-unreachable').addClass('-complete');
+          $('.step-title', $nextSteps).removeClass('not-allowed');
+        } else {
+          //If the step has a continue button, all next steps are disabled in order to force the user to click on continue
+          const $nextSteps = $clickedStep.nextAll();
+          $nextSteps.addClass('-unreachable').removeClass('-complete');
+          $('.step-title', $nextSteps).addClass('not-allowed');
+        }
```
eternoendless, line 67: Consider extracting functions and objects, this code is very complicated to understand.

Picture this:

```js
if (!clickedStep.isUnreachable()) {
	clickedStep.makeCurrent();
	if (clickedStep.hasContinueButton()) {
		steps.disableAllAfter(clickedStep)
	} else {
        steps.enableAllBefore(clickedStep)
    }
}
```

Looks better isn't it?

jolelievre: Yep much better, but then I need to extend jQuery functions? I'm not familiar with it but I can try
Or could this be enough:
```javascript
if (!isUnreachable(clickedStep)) {
	makeCurrent(clickedStep);
	if (hasContinueButton(clickedStep)) {
		disableAllAfter(steps, clickedStep)
	} else {
        	enableAllBefore(steps, clickedStep)
    	}
}

```

matks (approved):

matks:
Requested change from @eternoendless will be handled for 1.7.7
Current PR will be merged for 1.7.6.x as it has been validated ✅by @khouloudbelguith

Issue #15066, opened 2019-08-11
Logo is saved with wrong MIME type


MathiasReker:
Go to Design -> Images -> Set use PNG -> Save.
![1](https://user-images.githubusercontent.com/26626066/62837373-fbe4a180-bc6e-11e9-9463-7e0ad49b624d.png)

Go to Design -> Theme and logo -> Upload logo in PNG -> Save.

Add to your .htaccess file:
```
<IfModule mod_headers.c>
    Header always set X-Content-Type-Options "nosniff"
</IfModule>
```

Open browser Internet Explorer 11

Logo is not visible! -> Check console of IE11: `DOM7009: Unable to decode image at URL`.

So what is being done is that a PNG file is being saved with JPG extention.

Go to: `/src/Core/ShopLogoUploader.php`

Line 107:
Change:
            `$fileExtension = ($fieldName == 'PS_STORES_ICON') ? '.gif' : '.jpg';`
To:
            `$fileExtension = ($fieldName == 'PS_STORES_ICON') ? '.gif' : '.png';`

After this change clear browser cache and try again. Now everything is good. But this is not the right fix, it is just for demonstration.

**Additional information**
PrestaShop version: 1.7.6.0
PHP version: 7.2

MathiasReker:
I made a PR: https://github.com/PrestaShop/PrestaShop/pull/15067

Issue #14693, opened 2019-07-16
Unable to upload png files into Product categories pages


mickaelandrieu:
**Describe the bug**
I wanted to use a png file as product category image but I can't, the file is always converted to jpg file at best.

**To Reproduce**
1. Go to the Back Office, in Sell > Categories
2. Edit a category, upload a png file
3. Save and the message of success
4. Then check the file: well, it's a jpg and the transparency is lost.

![oupsy](https://user-images.githubusercontent.com/1247388/61325045-22502200-a814-11e9-8d24-362d4d927189.png)

**Additional information**
PrestaShop version: 1.7.6.0
PHP version: 7.2

khouloudbelguith:
Hi @mickaelandrieu,

I have the same issue with PS1.7.5.1 & PS1.7.5.2.
![image](https://user-images.githubusercontent.com/16067358/61358680-c7014c80-a872-11e9-9309-2ef6132c44a8.png)
I’ll add this to the debug roadmap.
Thanks!

mickaelandrieu:
Fair enough !

Thanks @khouloudbelguith :)

marionf:
@mickaelandrieu 

Can you tell us what is selected in your BO > Design > Image > Image format ?

mickaelandrieu:
Sorry for your time @marionf

PR #15067, opened 2019-08-11
Logo is saved with wrong MIME type

MathiasReker:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | https://github.com/PrestaShop/PrestaShop/issues/15066
| Type?         | bug fix
| Category?     | FO
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/15066
| How to test?  | https://github.com/PrestaShop/PrestaShop/issues/15066

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15067)

### DIFF
--- a/src/Core/Shop/LogoUploader.php
+++ b/src/Core/Shop/LogoUploader.php
@@ -104,7 +104,7 @@ class LogoUploader
                 throw new PrestaShopException(sprintf('%Upload of temporary file to %s has failed.', $tmpName));
             }
 
-            $fileExtension = ($fieldName == 'PS_STORES_ICON') ? '.gif' : '.jpg';
+            $fileExtension = ($fieldName == 'PS_STORES_ICON') ? '.gif' : '.' . Configuration::get('PS_IMAGE_QUALITY');
             $logoName = $this->getLogoName($logoPrefix, $fileExtension);
 
             if ($fieldName == 'PS_STORES_ICON') {


MathiasReker:
@eternoendless I think this should target 1.7.6.1 because wrong MIME type can be critical.

matks:
Hi @MathiasReker thanks for the PR !

I have a question 🤔here you modify the extension in the name, so if in the BO "png" has been configured for config `PS_IMAGE_QUALITY`, then the logo filename will be `mylogo.png` instead of `mylogo.jpg` => this is what your PR does.

But this is only the filename. The image is either a JPG or a PNG. What happens if someone 1) configures the BO to PNG but 2) uploads a JPG ? Will it save it with `.png` even the file is a JPG ?

mickaelandrieu:
Hello @matks & @MathiasReker,

I guess this can be linked to this issue: https://github.com/PrestaShop/PrestaShop/issues/14693

I have the feeling that the core "service/class" used to manage the upload is buggy. This is not a regression from 1.7.5 afaik, so on I'm not that sure this should be "quick-fixed" in 1.7.6.1.

Cheers

MathiasReker:
@matks you are right! What I am doing is not going to work in any case. I assumed that users uploade the file with same extension as selected in `Configuration::get('PS_IMAGE_QUALITY')`.
But that might not be the case. What I want to do is to get the extension of the uploaded image.

MathiasReker:
@mickaelandrieu Please check this URL with IE11: https://iqit-commerce.com/ps17/demo5/en/
image

The logo is not displayed in some browsers. I think that is critical. Therefor I attempted to fix it for 1.7.6.1 :-)

MathiasReker:
See this screenshot. A `JPG` image with transparent background. :trollface: 

jpg

As it is in 1.7.6.0 all logo files is forced with `JPG` even if they are `PNG`. In some cases the browser cannot decode the image because of wrong MIME type.

matks:
> As it is in 1.7.6.0 all logo files is forced with `JPG` even if they are `PNG`. In some cases the browser cannot decode the image because of wrong MIME type.

That's what I was afraid of 😱
So the issue is even worse: the filename has wrong extension AND the filename extension is unrelated to the **real** extension

I think we need a better (meaning: more complex) management of the topic

matks:
I think we need to do the following:
- only accept valid image formats to be uploaded, other formats must be rejected (need to define what is valid 😬)
- save the uploaded files with the right extension

MathiasReker:
> 
> 
> I think we need to do the following:
> 
>     * only accept valid image formats to be uploaded, other formats must be rejected (need to define what is valid 😬)
> 
>     * save the uploaded files with the right extension

Yes, agree on that. I am not the right person do that, so I don't update my PR. Should I close this PR or should I let it stay open until a new one is created?

eternoendless:
It's best to close it for now. Thanks @MathiasReker for pointing out the problem.

PR #15071, opened 2019-08-12
Migration of /international/locations/states edit and create forms

RaimondasSapola:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migration of /international/locations/states edit and create forms with custom iso code unique validation
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10607
| How to test?  | /international/locations/states edit and create forms should remain unchanged

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15071)


src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
```diff
@@ -58,10 +69,174 @@ public function getStatesAction(Request $request)
...
+                    'Admin.Notifications.Error',
+                    [sprintf('"%s"', $this->trans('ISO code', 'Admin.International.Feature'))]
+                ),
+                StateConstraintException::INVALID_STATE => $this->trans(
+                    'An error occurred when attempting to update the required fields.',
+                    'Admin.Notifications.Error'
+                ),
+            ],
+            CannotUpdateStateException::class => $this->trans(
+                'An error occurred on saving.',
```
LouiseBonnard: I find this wording quite weird when reading it like this... what do you think about writing _An error occurred while attempting to save._ instead?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/State/Blocks/form.html.twig
```diff
@@ -0,0 +1,112 @@
...
+        <div class="col-sm">
+          {{ form_widget(stateForm.name) }}
+          {{ form_errors(stateForm.name) }}
+        </div>
+      </div>
+
+      <div class="form-group row">
+        {{ ps.label_with_help(
+          'ISO code'|trans({}, 'Admin.International.Feature'),
+          ('1 to 4 letter ISO code.'|trans({}, 'Admin.International.Help')) ~ ' ' ~ ('You can prefix it with the country ISO code if needed.'|trans({}, 'Admin.International.Help')),
```
LouiseBonnard, line 56: I have a doubt... shouldn't it be _1 to 4 letter(s) ISO code._ instead?

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
```diff
@@ -58,10 +69,174 @@ public function getStatesAction(Request $request)
...
+     */
+    private function getErrorMessages(): array
+    {
+        return [
+            StateException::class => $this->trans(
+                'Unexpected error occurred.',
+                'Admin.Notifications.Error'
+            ),
+            StateNotFoundException::class => $this->trans(
+                'The object cannot be loaded (or found)',
```
LouiseBonnard, line 204: If I understand well, the two following error notifications here apply to the "States" tab of the "Locations" section of the back office? Why are we talking about an object to load, it seems to be pretty much adding a state manually or editing one without any loading action?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/State/Blocks/form.html.twig
RaimondasSapola: I left it as it was in pre migration state :) i can add the change :)

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
RaimondasSapola: For editing an object it needs to be loaded. I don't think that this will happen in practice but StateNotFoundException could be thrown in editing flow so the message should be here.

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/State/Blocks/form.html.twig
LouiseBonnard: Alright, @RaimondasSapola, just leave it as is, forget about my comment. :-)

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
LouiseBonnard: Okay, so when you create/edit a location in the International part of the back office, it is considered as a loaded object?

RaimondasSapola: Only on editing but yes

sarjon:
@matks we have 3 new closely related domains: `State`, `Zone` and `Country`. Currently they are placed at the root of `Domain` namespace. Would it make sense to group them somehow? Maybe into `Location`. :thinking:

src/Adapter/State/QueryHandler/GetStateForEditingHandler.php
```diff
@@ -0,0 +1,70 @@
...
+        $stateId = $query->getStateId();
+        $state = $this->getState($stateId);
+
+        $countryId = new CountryId($state->id_country);
+        $zoneId = new ZoneId($state->id_zone);
+
+        return new EditableState(
+            $stateId,
+            $countryId,
+            $zoneId,
```
sarjon: You can do `new ZoneId` right here to avoid using variables.

src/Adapter/State/QueryHandler/GetStateWithIsoCodeExistsHandler.php
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Handles query for determining if state with given iso code exists
+ */
+final class GetStateWithIsoCodeExistsHandler implements IsUniqueStateIsoCodeHandlerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(IsUniqueStateIsoCode $query): bool
```
sarjon: Query does not match you handler name. Query should be `GetStateWithIsoCodeExists`.

src/Adapter/State/QueryHandler/GetStateWithIsoCodeExistsHandler.php
```diff
@@ -0,0 +1,56 @@
...
+namespace PrestaShop\PrestaShop\Adapter\State\QueryHandler;
+
+use PrestaShop\PrestaShop\Core\Domain\State\Query\isUniqueStateIsoCode;
+use PrestaShop\PrestaShop\Core\Domain\State\QueryHandler\IsUniqueStateIsoCodeHandlerInterface;
+use State;
+
+/**
+ * Handles query for determining if state with given iso code exists
+ */
+final class GetStateWithIsoCodeExistsHandler implements IsUniqueStateIsoCodeHandlerInterface
```
sarjon: Interface name does not match implementation name.

src/Core/ConstraintValidator/TypedRegexValidator.php
```diff
@@ -87,6 +87,7 @@ private function getPattern($type)
             'message' => '/[<>{}]/i',
             'language_iso_code' => IsoCode::PATTERN,
             'language_code' => '/^[a-zA-Z]{2}(-[a-zA-Z]{2})?$/',
+            'state_iso_code' => '/^[a-zA-Z0-9]{1,4}((-)[a-zA-Z0-9]{1,4})?$/',
```
sarjon: I feel like there should be `StateIsoCode` VO and it could define this pattern inside as well, wdyt? @matks wdyt?

src/Core/ConstraintValidator/Constraints/UniqueStateIsoCode.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints;
+
+use PrestaShop\PrestaShop\Core\ConstraintValidator\UniqueStateIsoCodeValidator;
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Unique state iso code validator constraint
+ */
+class UniqueStateIsoCode extends Constraint
```
sarjon, line 35: Don't you need any configuration for this constraint? Like exclude state id?

src/Core/Domain/State/Exception/CannotUpdateStateException.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\State\Exception;
+
+use PrestaShop\PrestaShop\Core\Domain\Exception\DomainException;
+
+/**
+ * Thrown on failure to update state
+ */
+class CannotUpdateStateException extends DomainException
```
sarjon: Shouldn't it extend `StateException`?

src/Core/Domain/State/Exception/StateConstraintException.php
```diff
@@ -0,0 +1,37 @@
...
+namespace PrestaShop\PrestaShop\Core\Domain\State\Exception;
+
+/**
+ * Exception is thrown when State constraint is violated
+ */
+class StateConstraintException extends StateException
+{
+    public const INVALID_ID = 1;
+
+    public const INVALID_FIELD_VALUES = 2;
```
sarjon, line 36: I don't know about this one, it's so generic, it does not tell developer which field is invalid, do we need such code? @matks wdyt?

src/Core/Form/ChoiceProvider/CountryByIdConfigurableChoiceProvider.php
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Form\ChoiceProvider;
+
+use PrestaShop\PrestaShop\Adapter\Country\CountryDataProvider;
```
sarjon: It's better not to use `Adapter` services inside `Core`.

src/Core/Form/IdentifiableObject/DataHandler/StateFormDataHandler.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public function create(array $data)
+    {
+        /** @var StateId $stateId */
+        $stateId = $this->commandBus->handle(new AddStateCommand(
+            $data['id_country'],
+            $data['id_zone'],
+            $data['name'],
+            $data['iso_code'],
+            $data['active'] ?? false
```
sarjon, line 69: Why you need to check `active` here, but you don't check it in update method?

src/PrestaShopBundle/Form/Admin/Improve/International/Locations/State/StateType.php
```diff
@@ -0,0 +1,108 @@
...
+        $builder
+            ->add('name', TextType::class, [
+                'required' => true,
+                'constraints' => [
+                    new NotBlank(),
+                    new TypedRegex([
+                        'type' => 'generic_name',
+                    ]),
+                    new Length([
+                        'max' => 32,
```
sarjon: I think it should be moved to const.

src/PrestaShopBundle/Form/Admin/Improve/International/Locations/State/StateType.php
```diff
@@ -0,0 +1,108 @@
...
+                    ]),
+                    new CleanHtml(),
+                ],
+            ])
+            ->add('iso_code', TextType::class, [
+                'required' => true,
+                'constraints' => [
+                    new NotBlank(),
+                    new Length([
+                        'max' => 7,
```
sarjon: I think it should be moved to const.

src/PrestaShopBundle/Resources/config/services/adapter/state.yml
```diff
@@ -0,0 +1,36 @@
...
+    public: true
+
+  prestashop.adapter.state.command_handler.abstract_state_handler:
+    abstract:  true
+    class: 'PrestaShop\PrestaShop\Adapter\State\AbstractStateHandler'
+
+  prestashop.adapter.state.command_handler.create_state_handler:
+    public: true
+    class: 'PrestaShop\PrestaShop\Adapter\State\CommandHandler\AddStateHandler'
+    parent: 'prestashop.adapter.state.command_handler.abstract_state_handler'
```
sarjon: I don't think you need to configure parent. It's useful when abstract class has dependencies, in your case it doesn't.

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/State/edit.html.twig
```diff
@@ -0,0 +1,37 @@
...
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ *#}
+
+{% set enableSidebar = true %}
+{% set layoutTitle = 'Edit: %value%'|trans({'%value%': stateInformation.name}, 'Admin.Actions') %}
```
sarjon: Would be better to set these values in controller. :)

sarjon (changes requested):

src/Adapter/Country/QueryHandler/GetCountriesHandler.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public function __construct(CountryDataProvider $countryDataProvider)
+    {
+        $this->countryDataProvider = $countryDataProvider;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(GetCountries $query): array
```
sarjon: I think it's better not to return array but object instead.

src/Core/Domain/Country/Query/GetCountries.php
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * @var bool
+     */
+    private $containsStates = false;
+
+    /**
+     * @var bool
+     */
+    private $listStates = false;
```
sarjon: I think name of this param should be improved, it's not clear what's the purpose of it imo.

src/Core/Domain/Country/Query/GetCountries.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        return $this->active;
+    }
+
+    /**
+     * @param bool $active
+     *
+     * @return GetCountries
+     */
+    public function setActive(bool $active): GetCountries
```
sarjon: What if we drop setters here and instead use something like this:
```php
$query = (new GetCountries($langId))
    ->onlyActive()
    ->thatContainsStates()
;
```

@matks wdyt?

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
```diff
@@ -58,10 +70,173 @@ public function getStatesAction(Request $request)
             ]);
         } catch (Exception $e) {
             return $this->json([
-                    'message' => $this->getErrorMessageForException($e, []),
-                ],
+                'message' => $this->getErrorMessageForException($e, []),
```
sarjon: You don't define any custom messages, is that intentional?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/State/edit.html.twig
```diff
@@ -0,0 +1,37 @@
...
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ *#}
+
+{% set enableSidebar = true %}
+{% set layoutTitle = 'Edit: %value%'|trans({'%value%': stateName}, 'Admin.Actions') %}
```
sarjon: Pass these vars from controller instead.

src/Core/Domain/Country/Query/GetCountries.php
RaimondasSapola: Well get countries options are passed from form configuration so i think it is more clear to just pass those options to setters instead of writing additional conditions to call methods

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
RaimondasSapola: Well i only update code style here but i guess it was intended for some reason.

src/Core/Domain/Country/QueryResult/Countries.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $this->countries = $countries;
+    }
+
+    /**
+     * @return array
+     */
+    public function getCountries(): array
+    {
+        return $this->countries;
```
sarjon, line 55: Isn't it legacy array, but wrapped into `Countries`? I wonder if we can do better. :wink: 

@matks wdyt?

src/Core/Domain/Country/QueryResult/Countries.php
```diff
@@ -0,0 +1,54 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Country\QueryResult;
+
+/**
+ * Holds list of countries found by query configuration
+ */
+class Countries
```
sarjon, line 35: I feel like this is too generic, @matks wdyt?

src/Core/Form/ChoiceProvider/CountryByIdConfigurableChoiceProvider.php
```diff
@@ -0,0 +1,95 @@
...
+        $query = (new GetCountries($this->langId))
+            ->setActive($active)
+            ->setContainsStates($containsStates)
+            ->setIncludeStatesList($listStates);
+
+        /** @var Countries $countries */
+        $countries = $this->queryBud->handle($query);
+        $choices = [];
+
+        foreach ($countries->getCountries() as $country) {
```
sarjon, line 91: I'm wondering if you really need a query here, because in the end you're still dealing with legacy array. :confused:

src/PrestaShopBundle/Form/Admin/Improve/International/Locations/State/StateType.php
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Defines form for state create/edit actions (Improve > International > Locations > State)
+ */
+class StateType extends AbstractType
+{
+    /** @var int maximum number of characters for name */
+    public const MAX_NAME_LENGTH = 32;
+
+    /** @var int maximum number of characters for iso code */
+    public const MAX_ISO_CODE_LENGTH = 7;
```
sarjon: Wouldn't it make sense to move these consts into `Domain`? :thinking:

RaimondasSapola: Should i create data transfer object for that?

src/Core/Domain/Country/QueryResult/Countries.php
matks: Mmmm 🤔 I'm afraid it's just ... countries. Nothing more. Cant think of something better 😅

src/Core/Domain/Country/QueryResult/Countries.php
matks: Yes, we should create a VO Country and this would be a collection of Country VO

src/Core/Domain/State/CommandHandler/AddStateHandlerInterface.php
```diff
@@ -0,0 +1,43 @@
...
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\State\CommandHandler;
+
+use PrestaShop\PrestaShop\Core\Domain\State\Command\AddStateCommand;
+use PrestaShop\PrestaShop\Core\Domain\State\ValueObject\StateId;
+
+/**
+ * Defines contract for add state handler
```
matks: Does not add value 😉 I think we can remove it

src/Core/Domain/State/Exception/StateConstraintException.php
matks: Maybe not indeed 🤔

src/Core/Form/ChoiceProvider/CountryByIdConfigurableChoiceProvider.php
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Class provides configurable country choices with ID values.
+ */
+final class CountryByIdConfigurableChoiceProvider implements ConfigurableFormChoiceProviderInterface
+{
+    /**
+     * @var CommandBusInterface
+     */
+    private $queryBud;
```
matks:
```suggestion
    private $queryBus;
```
Query Buddy ? 😄

src/Core/Form/ChoiceProvider/CountryByIdConfigurableChoiceProvider.php
matks: Indeed so far the Providers  we have implemented are Adapters using legacy logic 🤔 . Let's stick to this strategy for now although it raises a true question: is this part relevant for CQRS usage ?

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
```diff
@@ -57,11 +69,177 @@ public function getStatesAction(Request $request)
...
+            $result = $stateFormHandler->handleFor((int) $stateId, $stateForm);
+            if ($result->isSubmitted() && $result->isValid()) {
+                $this->addFlash('success', $this->trans('Successful update.', 'Admin.Notifications.Success'));
+
+                return $this->redirectToRoute('admin_states_index');
+            }
+        } catch (Exception $e) {
+            $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages()));
+
+            return $this->redirectToRoute('admin_states_index');
```
matks, line 180: Please only redirect for listing page if `$e` is an instance of `StateNotFoundException` 😉 
```
            if ($e instanceof StateNotFoundException) {
                return $this->redirectToRoute('admin_states_index');
            }
```

src/PrestaShopBundle/Form/Admin/Type/ConfigurableCountryChoiceType.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        $this->countriesChoiceProvider = $countriesChoiceProvider;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        // Set normalizer enables to use closure for choice generation with options
```
matks: Yeah 😅 but ... we also want to keep the close as simple as possible following the [KISS](https://en.wikipedia.org/wiki/KISS_principle) principle. So if we can use a basic `foreach` here 😉 it would be better as we are onboarding the PrestaShop community towards Symfony so let's use it in the most simple way to make the learning curve as smooth as possible

src/PrestaShopBundle/Form/Admin/Type/ZoneChoiceType.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $this->zonesChoiceProvider = $zonesChoiceProvider;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        // Set normalizer enables to use closure for choice generation with options
```
matks: Same about KISS

src/PrestaShopBundle/Resources/config/routing/admin/improve/international/state.yml
```diff
@@ -1,6 +1,36 @@
...
   path: /
   methods: [GET]
+  #condition needed to differentiate from admin_states_index path
+  condition: "request.isXmlHttpRequest()"
   defaults:
     _controller: 'PrestaShopBundle:Admin\Improve\International\State:getStates'
     _legacy_controller: AdminStates
+
+admin_states_index:
+  path: /
```
matks, line 10: Let's avoid same URLs 😉 it is better according to KISS principle and avoids complex debugging issues

matks (changes requested):
Sorry for very late review 😢

src/Core/Domain/Country/QueryResult/Countries.php
sarjon: Actually there is a huge difference. Image you have 2 different use cases:
1. For X case, were we need to get countries with id, name, state and addresses that belongs to it.
2. For Y case, need to get countries with id, states, customers that belongs to it.

As a result we have 2 different queries:

```php
interface GetCountriesForXHandlerInterface {
    public function handle(GetCountriesForX $query): CountriesForX
}

interface GetCountriesForYHandlerInterface {
    public function handle(GetCountriesForY $query): CountriesForY
}
```

We even have examples in current implementation, both `GetCustomerForEditing` and `GetCustomerForViewing` returns the same customer, but with totally different data.

This is what builds PrestaShop's API, not a generic `GetCountries` query which returns everything, but a specialized rich domain model queries.

At least this is how I see it. :stuck_out_tongue:

matks: I think we should then define
- a Country VO with Country dedicated data
- another object (we need to find a suitable name) to describe the data needed for X, that will probably be a data structure that contains some country VOs
- another object (we need to find a suitable name) to describe the data needed for Y, that will probably be a data structure that contains some country VOs

sarjon: Not really, I think it's enough to have `Country/QueryResult/CountryForX` and `Country/QueryResult/CountryForY`, it's just a DTO, we don't have any business rules to protect on read side.

Regarding `Country` VO, I don't think it makes sense at this point, in the end, it's not a VO, bu entity and as far as I understand entities will be introduced in future major version (I guess v100.0.0 or later :man_shrugging: ). :)

src/Core/Domain/Country/QueryResult/Countries.php
tdavidsonas88: created a VO Country and refactored this to be a collection of Country VO

src/Core/Domain/Country/QueryResult/Countries.php
tdavidsonas88: In the current implementation - only CountryByIdConfigurableChoiceProvider is using GetCountries Query. It requires only country id and name. So created Country VO with id and name. Also ArrayCollection of these VO's.

src/Core/Form/ChoiceProvider/CountryByIdConfigurableChoiceProvider.php
tdavidsonas88: Refactored legacy object to be mapped to Country VO inside the handler.

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
tdavidsonas88: refactored like this: 

```
        } catch (StateNotFoundException $e) {
            $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages()));

            return $this->redirectToRoute('admin_states_index');
        } catch (Exception $e) {
            $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages()));
        }
```

tdavidsonas88:
@RaimondasSapola While testing found out that the current link to add new state is: /improve/international/states/new and not the /improve/international/locations/states/new . Not sure which one should be.

src/Core/Form/ChoiceProvider/CountryByIdConfigurableChoiceProvider.php
tdavidsonas88: Query Buddy was so awesome - made me laugh :slightly_smiling_face:

src/Core/Domain/State/Exception/StateConstraintException.php
tdavidsonas88: Well, there's value in code navigation to find all usages of INVALID_FIELD_VALUES in PHPStorm. It helped me find all the places where this exception is thrown quite fast. Refactored to throw more specific exception - the one which object model throws. 

Searching by StateConstraintException gives too many values.

src/PrestaShopBundle/Form/Admin/Type/ConfigurableCountryChoiceType.php
tdavidsonas88: example where we have it with foreach would make me do it faster. Searching/thinking about solution.

src/Adapter/State/CommandHandler/AddStateHandler.php
```diff
@@ -39,7 +38,7 @@
 /**
  * Handles creation of state
  */
-final class AddStateHandler extends AbstractStateHandler implements AddStateHandlerInterface
```
RaimondasSapola: Do not remove this. It will be used in later stages in development where adapters will be removed

tdavidsonas88: Ok, will revert it back. Have you seen review comment above about AddStateHandlerInterface?

RaimondasSapola: The review comment refers to a php comment above the class

tdavidsonas88: done Raimondas - we have it again.

src/PrestaShopBundle/Form/Admin/Type/ConfigurableCountryChoiceType.php
tdavidsonas88: Actually the normalizer gives a nice way to use Options $options values from FormType; Not sure how to do it simpler?!

matks: My mistake, I reviewed too fast. Normalizer is used for a good reason here 👍

tdavidsonas88:
looks like not related failures of .js tests occur in /home/runner/work/PrestaShop/PrestaShop/admin-dev/themes/new-theme/js/pages/order/view/order-view-page.js: 

/home/runner/work/PrestaShop/PrestaShop/admin-dev/themes/new-theme/js/pages/order/view/order-view-page.js
##[error]   33:1    error  `@components/router` import should occur before import of `./order-invoices-refresher`  import/order
##[error]  158:13   error  'modal' is assigned a value but never used                                              no-unused-vars
##[error]  161:25   error  Expected parentheses around arrow function argument                                     arrow-parens
##[error]  166:1    error  This line has a length of 136. Maximum allowed is 120                                   max-len
##[error]  166:120  error  Unnecessarily quoted property 'id' found                                                quote-props

Any guidance how to solve these would be helpful. Tried it on my own configuring linter but then received different errors;

src/PrestaShopBundle/Resources/config/routing/admin/improve/international/state.yml
tdavidsonas88: changed to  path: /list

tdavidsonas88:
forced pushed after rebase from develop hoping to not related .js failures to go away.

src/Adapter/Country/QueryHandler/GetCountriesHandler.php
```diff
@@ -0,0 +1,73 @@
...
+
+use Doctrine\Common\Collections\ArrayCollection;
+use PrestaShop\PrestaShop\Adapter\Country\CountryDataProvider;
+use PrestaShop\PrestaShop\Adapter\Country\ValueObject\Country;
+use PrestaShop\PrestaShop\Core\Domain\Country\Query\GetCountries;
+use PrestaShop\PrestaShop\Core\Domain\Country\QueryHandler\GetCountriesHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Country\QueryResult\Countries;
+
+/**
+ * Handles get countries query and returns countries array list
```
zuk3975: countries array list?
As I see it returns Countries object.

src/Adapter/Country/ValueObject/Country.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace PrestaShop\PrestaShop\Adapter\Country\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryConstraintException;
+
+class Country
+{
+    /** @var int */
+    private $id;
```
zuk3975, line 12: missing space.
also ps uses 
block type for props (not one line)
```
/**
* @var int
*/
```

src/Adapter/Country/ValueObject/Country.php
```diff
@@ -0,0 +1,56 @@
...
+    /** @var string */
+    private $name;
+
+    /**
+     * Country constructor.
```
zuk3975: rm this 'Country constructor.' I suggest you configuring your storm file tpl to remove this, as its useless

src/Adapter/Country/ValueObject/Country.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Validates that the value is integer and is greater than zero
+     *
+     * @param $value
+     *
+     * @throws CountryConstraintException
+     */
+    private function assertIsIntegerGreaterThanZero($value)
```
zuk3975: type value as int and as we have types, we dont need to assert if its int anymore, just 0.

src/Adapter/State/AbstractStateHandler.php
```diff
@@ -0,0 +1,64 @@
...
+{
+    /**
+     * @param StateId $stateId
+     *
+     * @return State
+     *
+     * @throws StateNotFoundException
+     * @throws StateException
+     */
+    protected function getState(StateId $stateId)
```
zuk3975: missing return type

src/Adapter/State/QueryHandler/GetStateWithIsoCodeExistsHandler.php
```diff
@@ -0,0 +1,56 @@
...
+namespace PrestaShop\PrestaShop\Adapter\State\QueryHandler;
+
+use PrestaShop\PrestaShop\Core\Domain\State\Query\GetStateWithIsoCodeExists;
+use PrestaShop\PrestaShop\Core\Domain\State\QueryHandler\GetStateWithIsoCodeExistsHandlerInterface;
+use State;
+
+/**
+ * Handles query for determining if state with given iso code exists
+ */
+final class GetStateWithIsoCodeExistsHandler implements GetStateWithIsoCodeExistsHandlerInterface
```
zuk3975: Not sure about this, maybe you've discussed it with @matks? I think we have used some sort of providers in adapters. Ping just to make sure. 
Also, @matks, should we prefix all queries with get, even if its returning bool. or maybe there is any other convention? Cuz now its hard to read the name.

src/Core/Domain/State/Config/StateValidationConfiguration.php
```diff
@@ -0,0 +1,55 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\State\Config;
```
zuk3975, line 27: just a note, we have used similar object in Addresses, but haven't put it in Config dir.
@matks, should we introduce this Config as some kind of convention?

src/Core/Domain/State/Exception/StateConstraintException.php
zuk3975: I guess it came from a generic legacy fields validation in handler, by which we don't know which field is invalid. Anyway I think its better than no code at all. Even more precise option would be to list every invalid-field as exception code, but that would be a huge overhead

src/Core/Domain/State/QueryResult/EditableState.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * @var int
+     */
+    private $zoneId;
+
+    /**
+     * @var string
+     */
+    private $name;
```
zuk3975, line 56: If im right, EditableObjects should have all properties optional except the identification (in this case is the stateId).
The point is if you want to update only one property, you don't need to fill the whole object. And in the handler it checks whether the property needs to be change or no (if its null or has a value).

src/Core/Domain/State/QueryResult/EditableState.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * @param StateId $stateId
+     * @param CountryId $countryId
+     * @param ZoneId $zoneId
+     * @param string $name
+     * @param string $isoCode
+     * @param bool $active
+     */
+    public function __construct(
```
zuk3975: Following the comment above, EditableObjects should be allowed to set only the property which you want to update, so we shouldn't have a constructor which requires to fill all fields.

src/Core/Form/ChoiceProvider/CountryByIdConfigurableChoiceProvider.php
```diff
@@ -0,0 +1,97 @@
...
+     * @var int
+     */
+    private $langId;
+
+    /**
+     * @param int $langId
+     * @param CommandBusInterface $queryBud
+     */
+    public function __construct(
+        $langId,
```
zuk3975: missing type

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
```diff
@@ -57,11 +69,179 @@ public function getStatesAction(Request $request)
...
+        } catch (Exception $e) {
+            $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages()));
+        }
+
+        return $this->render('@PrestaShop/Admin/Improve/International/Locations/State/edit.html.twig', [
+            'enableSidebar' => true,
+            'layoutTitle' => $this->trans('Edit: %value%', 'Admin.Actions', ['%value%' => $editableState->getName()]),
+            'stateForm' => $stateForm->createView(),
```
zuk3975, line 188: I think static php analyzer would tell you that $stateForm might not have been defined.
you could check it just to make sure, and redirect to index page if it isn't.

src/PrestaShopBundle/Form/Admin/Improve/International/Locations/State/StateType.php
```diff
@@ -0,0 +1,153 @@
...
+
+        $builder
+            ->add('name', TextType::class, [
+                'required' => true,
+                'constraints' => [
+                    new NotBlank(),
+                    new TypedRegex([
+                        'type' => 'generic_name',
+                    ]),
+                    new Length(
```
zuk3975: I think cs fixer doesn't mind if we write like that:
```
                    new Length([
                        'max' => StateValidationConfiguration::MAX_NAME_LENGTH,
                        'maxMessage' => $this->trans(
                            'This field cannot be longer than %limit% characters',
                            ['%limit%' => StateValidationConfiguration::MAX_NAME_LENGTH],
                            'Admin.Notifications.Error'
                        ),
                    ]),
```

src/PrestaShopBundle/Resources/config/routing/admin/improve/international/state.yml
```diff
@@ -1,6 +1,36 @@
 admin_country_states:
   path: /
   methods: [GET]
+  #condition needed to differentiate from admin_states_index path
+  condition: "request.isXmlHttpRequest()"
```
zuk3975: i doubt condition is needed as I see its a different path in index.
```
admin_states_index:
  path: /list
```

But anyway I would suggest having index as `/` and smth like `/states` for ajax

src/PrestaShopBundle/Resources/config/services/adapter/country.yml
```diff
@@ -20,3 +20,9 @@ services:
     class: 'PrestaShop\PrestaShop\Adapter\Country\CountryRequiredFieldsProvider'
     arguments:
       - '@prestashop.adapter.data_provider.country'
+  prestashop.adapter.country.query_handler.get_countries:
```
zuk3975:
```suggestion
  prestashop.adapter.country.query_handler.get_countries_handler:
```
Just to follow namespace

src/Core/Domain/State/Exception/StateConstraintException.php
tdavidsonas88: Yes, I like this const if we throw exception from object model which tells which field and what was wrong. Not wrapping up with general message.

src/Adapter/Country/ValueObject/Country.php
tdavidsonas88: Really? :question:  how the block is better than one liner? One liner seems more readable... what am I missing in my understanding? If we use block and elsewhere in the code for variables I would see it as a downgrade :disappointed:

src/Adapter/Country/ValueObject/Country.php
tdavidsonas88: nice :slightly_smiling_face:  Do you know where exactly possible to do that is storm settings. Searched for that but not so easy to find how..

src/Core/Domain/State/QueryResult/EditableState.php
tdavidsonas88: Looked how the other editable objects are made(EditableEmployee, EditableCustomer for example). See the same as here with not optional fields. Waiting for confirmation from core team how it should be.

src/Core/Form/ChoiceProvider/CountryByIdConfigurableChoiceProvider.php
tdavidsonas88: sometimes types are missing to avoid BC breaks as I remember from core team reviews. Not sure how about this case?

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
tdavidsonas88: what do you mean by static page analyzer?

src/PrestaShopBundle/Resources/config/routing/admin/improve/international/state.yml
tdavidsonas88: imlemented Julius suggestion

src/Adapter/Country/ValueObject/Country.php
tdavidsonas88: changed it to PS standard as you guys have more experience with PS even though one liner looks better for me

src/Adapter/Country/ValueObject/Country.php
tdavidsonas88: changed it to PS standard as you guys have more experience with PS even though one liner looks better for me

src/Core/Domain/State/QueryResult/EditableState.php
zuk3975: oh sorry, my bad, I had in mind Command for editing. Everything good here.

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
zuk3975: phpstan for example.

src/Adapter/Country/ValueObject/Country.php
tdavidsonas88: Thanks Julius for guiding to find: File and Code Templates -> PHP Constructor

src/Core/Form/ChoiceProvider/CountryByIdConfigurableChoiceProvider.php
zuk3975: It seems to be a new class(Or isn't it?), so I don't see how can it break smth, as you control how it works until its merged.

src/PrestaShopBundle/Controller/Admin/Improve/International/StateController.php
tdavidsonas88: well, $stateForm is inside try catch block; would be nice if phpstan also could adjust to that. I would prefer to use try catch instead of ifs..

src/Core/Form/ChoiceProvider/CountryByIdConfigurableChoiceProvider.php
tdavidsonas88: looks like the new one, added the missing type :slightly_smiling_face:  thanks!

src/Adapter/Country/ValueObject/Country.php
matks: I think there is no logical background for that. "It was done like this for years, so we continue doing it" (for the sake of having a uniform codebase)

There are probably pros and cons between these 2 styles but I dont know 😅 I'll just say that using props block types is more common, more standard for this project, so changing that would surprise / create confusion

src/Adapter/State/QueryHandler/GetStateWithIsoCodeExistsHandler.php
matks: Ah, crap, again this debate 😅 

We brainstormed a lot with @RaimondasSapola Raimondas on this one but we finally figured it out:

We should use Queries and Commands to model interactions between the User and the Domain.
In a CRUD add/edit form, the Commands AddObjectCommand and UpdateObjectCommand actually carry both the data and the intention of the user to the Domain.

Here, the GetStateWithIsoCodeExistsHandler is used by a constraint to validate the form. There is no user in this interaction, so CQRS is invalid.

Indeed what we should use instead here is a Read Model (=> an Adapter DataProvider) which is another concept linked to CQRS.

Commands and Queries are user-interactions modeled into data. If the user is interested to know if a state with ISO code "TAH" exists, then we can create such Query. If it's not a user need but rather a system need, like here, then no Query.

I'm sorry this is quite abstract 😢 please tell me if this needs further explanations.

zuk3975: All in all, it should be some kind of data provider instead, as its not a real user use-case, but rather a system required check.

rokaszygmantas: very informative explanation!

tdavidsonas88: So instead of GetStateWithIsoCodeExistsHandler will create CountryStateByIsoCodeProvider

tdavidsonas88: done :building_construction:

src/Adapter/Country/ValueObject/Country.php
tdavidsonas88: so let's keep props block then to avoid confusion for properties.

tdavidsonas88:
to start simplifying the forms https://github.com/PrestaShop/PrestaShop/pull/16938 merged to develop is needed.

PierreRambaud:
I close this PR, related to this comment: https://github.com/PrestaShop/PrestaShop/pull/16938#issuecomment-664188653

PR #12127, opened 2019-01-11
[WIP] Introduce CheckServiceDefinitionsCommand

matks:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Introduce CheckServiceDefinitionsCommand that attempts to build every available Symfony service for PrestaShop. The goal is to check each of them is buildable. Maybe also add some checks (for example the service definition is valid, to detect [this kind of errors](https://github.com/PrestaShop/PrestaShop/pull/12125) ). It should be called by Travis in the end.
| Type?         | new feature
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | No QA needed as this is a new code safety check

**This PR is in work-in-progress: it will be finished when the Command outputs no issue** (so I have to fix all found issues before the PR can be merged).

To run the Command: `bin/console prestashop:debug:services-container`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/12127)

PR #15081, opened 2019-08-12
Fixed the support of Legacy classes using Console commands

### LINKED ISSUE #14173: Impossible to load Legacy classes in Console commands

mickaelandrieu:
| Questions     | Answers
| Branch?       | develop (but should be fixed asap)
| Description?  | Creating Symfony Commands in PrestaShop is insanely error-prone! Let's try to add tests, and complete the Console enough to validate 90% of use cases.
| Type?         | bugfix
| Category?     | CO
| BC breaks?    | nope!
| Deprecations? | yes
| Fixed ticket? | Fixes #14173
| How to test?  | Tests are done, the build should be green.

```php
<?php

namespace Fop\LatestProductsModule\Command;

use Product;
use PrestaShop\PrestaShop\Core\Console\Command;
use Symfony\Component\Console\Input\InputArgument;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;
use Symfony\Component\Console\Style\SymfonyStyle;

/**
 * This command display common information about a shop.
 */
final class LatestProductsCommand extends Command
{
    /**
     * {@inheritdoc}
     */
    protected function configure()
    {
        $this
            ->setName('fop:latest-products')
            ->setDescription('Displays Shop status')
            ->setHelp('This command allows you to display status of a shop')
        ;
    }

    /**
     * {@inheritdoc}
     */
    protected function execute(InputInterface $input, OutputInterface $output)
    {
        $products = Product::getNewProducts(1);

        $io = new SymfonyStyle($input, $output);
        $io->title('Legacy Latest Products listing');

        $io->table(
            ['ID', 'Name', 'Quantity', 'Price', 'Activated?'],
            $this->formatProductInformation($products)
        );
    }

    /**
     * @param array $products the list of the products
     *
     * @return array
     */
    private function formatProductInformation(array $products)
    {
        $productsInformation = [];
        /** @var Product $product */
        foreach ($products as $product) {
            $productsInformation[] = [
                $product['id_product'],
                $product['name'],
                $product['quantity'],
                $product['price'],
                $product['active'] ? '✔' : '✘',
            ];
        }

        return $productsInformation;
    }
}
```

output

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15081)


src/Core/Console/Command.php
```diff
@@ -0,0 +1,78 @@
...
+        require_once $container->get('kernel')->getRootDir() . '/../config/config.inc.php';
+
+
+        /** @var LegacyContext $legacyContext */
+        $legacyContext = $container->get('prestashop.adapter.legacy.context');
+
+        /** @var ShopContext $shopContext */
+        $shopContext = $container->get('prestashop.adapter.shop.context');
+
+        // Set an employee
```
mickaelandrieu:
```diff
- Set an employee
+ Set required information
```

src/Core/Console/Command.php
```diff
@@ -0,0 +1,78 @@
...
+        $employeeId = $input->getOption('employee');
+        $shopId = $input->getOption('shop');
+        $shopGroupId = $input->getOption('shop_group');
+
+        if ($shopId && $shopGroupId) {
+            throw new LogicException('Do not specify an ID shop and an ID group shop at the same time.');
+        }
+
+        $legacyContext->getContext()->controller = new \stdClass();
```
mickaelandrieu: let's wrap all that "Console Context stuff" into one object.

src/PrestaShopBundle/EventListener/MultishopCommandListener.php
```diff
@@ -31,6 +31,9 @@
 use Symfony\Component\Console\Exception\LogicException;
 use Symfony\Component\Console\Input\InputOption;
 
+/**
+ * @deprecated since 1.7.7, to be removed in the next major version.
```
mickaelandrieu, line 35: can be kept, but I'm wondering if someone is using it as it's not documented

src/Core/Console/Application.php
```diff
@@ -0,0 +1,55 @@
...
+    public function __construct(KernelInterface $kernel)
+    {
+        parent::__construct($kernel);
+
+        $this->setName('PrestaShop');
+        $this->setVersion(AppKernel::VERSION);
+
+        $inputDefinition = $this->getDefinition();
+        $inputDefinition->addOption(new InputOption('employee', '-em', InputOption::VALUE_REQUIRED, 'Specify employee context (id).', null));
+        $inputDefinition->addOption(new InputOption('shop', '-s', InputOption::VALUE_REQUIRED, 'Specify shop context (id).', null));
```
mickaelandrieu: I need to reuse the previous options for BC

src/PrestaShopBundle/Resources/config/services/bundle/event_listener.yml
```diff
@@ -32,14 +32,6 @@ services:
         tags:
         - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest, priority: 40 }
 
-    prestashop.multishop_command_listener:
```
mickaelandrieu: I need tests to check if this doesn't conflict with the PrestaShop application: this probably works as a side effect.

src/Core/Console/Command.php
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Console;
+
+use Employee;
```
mickaelandrieu: all legacy classes use needs to be extracted into the adapter namespace.

mickaelandrieu:
Hello @PrestaShop/prestashop-core-developers,

any chance to get this merged in the next patch version? :pray:

When do you plan to release 1.7.6.1?

matks:
> any chance to get this merged in the next patch version? 🙏

I guess this is @eternoendless decision but if we follow our rules, this is neither a regression nor a critical bug 🤔 so it's not something we need to patch (=> would be shipped in a patch version), it's rather a nice improvement for next minor version

> When do you plan to release 1.7.6.1?

We have only 1 issue in the 1.7.6 kanban (https://github.com/PrestaShop/PrestaShop/projects/4) so once the Kanban is empty (= every issue is in "done" column), we can process for the QA validation and then the release.
I guess it'll be easier in the end of August when QA team has a little more folks (right now a lot of them are in vacations) 😄

src/Adapter/Console/ContextLoader.php
```diff
@@ -0,0 +1,74 @@
...
+{
+    private $legacyContext;
+    private $shopContext;
+
+    public function __construct(LegacyContext $legacyContext, ShopContext $shopContext, $rootDir)
+    {
+        $this->legacyContext = $legacyContext;
+        $this->shopContext = $shopContext;
+
+        require_once $rootDir . '/../config/config.inc.php';
```
matks: Could be used for a LFI, please add a validation here

src/Adapter/Console/ContextLoader.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function loadConsoleContext(InputInterface $input)
```
matks, line 55: Is `InputInterface` the right input arg for this command ? It looks like using `HttpRequest` as an input to me 🤔 maybe we should replace it with
```suggestion
    public function loadConsoleContext($employeeId, $shopId, $shopGroupdId)
```

src/Core/Console/Application.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        parent::__construct($kernel);
+
+        $this->setName('PrestaShop');
+        $this->setVersion(AppKernel::VERSION);
+
+        $inputDefinition = $this->getDefinition();
+        $inputDefinition->addOption(new InputOption('employee', '-em', InputOption::VALUE_REQUIRED, 'Specify employee context (id).', null));
+
+        // @see MultiShopCommandListener BC compatible fix to display the options in the Console
```
matks, line 49: It's nice you're trying to solve all issues at once 😄 but I think it's a big challenge
Maybe we should aim to fix the different contexts in the console with different PRs ? I suggest handling the multishop in a separate PR, I'm sure it's a little bomb 💣 that will require a lot of efforts to tackle

src/Core/Console/Application.php
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * The PrestaShop Console application
+ */
+final class Application extends FrameworkConsoleApplication
+{
+    public function __construct(KernelInterface $kernel)
+    {
+        parent::__construct($kernel);
+
+        $this->setName('PrestaShop');
```
matks, line 43:
```suggestion
        $this->setName('PrestaShop CLI');
```
😄 ?

src/Core/Console/Application.php
```diff
@@ -0,0 +1,58 @@
...
+        $this->setVersion(AppKernel::VERSION);
+
+        $inputDefinition = $this->getDefinition();
+        $inputDefinition->addOption(new InputOption('employee', '-em', InputOption::VALUE_REQUIRED, 'Specify employee context (id).', null));
```
matks, line 47: This is only needed for BO, what if the CLI is used for a FO-related processing ?

matks: I'm thinking that maybe we should aim for BO-oriented CLI console and then a FO-oriented CLI console (separate concerns => separate envs) 🤔

src/Core/Console/Command.php
```diff
@@ -0,0 +1,47 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+
+/**
+ * Symfony Command class for PrestaShop allowed to rely on legacy classes
+ *
+ * @author Mickaël Andrieu <mickael.andrieu@prestashop.com>
+ */
+abstract class Command extends ContainerAwareCommand
```
matks, line 38: Could we improve the naming ? `Command` is a bit shy to express the fact that this is a "Symfony Command class for PrestaShop allowed to rely on legacy classes"

src/PrestaShopBundle/EventListener/MultishopCommandListener.php
matks: Must be kept then 😢

matks:
@mickaelandrieu did a small review, it's great to see there's not so much to be done to fix the issue but I think it'll need extensive QA exploration for complex usecases like multishop

Nice job !

I did something a lot of time ago, dont know if it's related: I tried to make sure every SF Service we create is console-compliant
See https://github.com/PrestaShop/PrestaShop/pull/12127
but I surrendered when I saw the amount of work needed (https://github.com/PrestaShop/PrestaShop/pull/12127#issuecomment-453866363)

src/Adapter/Console/ContextLoader.php
mickaelandrieu: If I need to add a new option I will have to break compatibility, and if I use scalar values I will have to validate data in this function when it's not its role (these options should be validated in the "interact" function of the commands.

src/Core/Console/Application.php
mickaelandrieu: it doesn't support multishop feature per see, it only displays the available options for each command :)

it's up to every command to handle or not these options ('id_shop' and 'id_shop_group') and it's probably a little :bomb: I do agree

src/Adapter/Console/ContextLoader.php
mickaelandrieu: you mean, validate that $rootDir will be a folder? in all cases, this file path will be dynamic.

src/Core/Console/Command.php
mickaelandrieu: I formerly named it "PrestaShopCommand", but as this class is part of PrestaShop only; I don't see the value to prefix the name of my class with PrestaShop.

I'm open to suggestions, what do you suggest?

src/Core/Console/Application.php
mickaelandrieu: except if I'm wrong you always have an employee set up, even in the Front Office :)

What kind of commands I can create that requires services available in FO?
Also, don't be fooled by the REQUIRED stuff: a CLI option is always optional 🤣 

We need to cover more use cases, and I've already created 3 different commands in my module.

mickaelandrieu:
> I did something a lot of time ago, dont know if it's related: I tried to make sure every SF Service we create is console-compliant

What we can do is add an extra validation pass to forbid every class that is not "namespaced" as an argument of a Symfony service: we should only rely on adapters or core classes.

Also, clean all core/prestashop classes to drop every "direct" call to object models: it's not that easy to do but in the long run, this will help you to remove object models (if you intent to ?).

Second extra rule, don't rely directly on the global state like "context" or "request".

This patch sounds complete, but I'm pretty sure such hacky use cases are not covered: I want to write more and more commands to try to break something 🗡 💥

mickaelandrieu:
ping @matks: looking at your PR I've contributed my own [command](https://github.com/mickaelandrieu/commands/blob/master/src/CheckServiceContainer.php) to challenge the Console Context.

I've found some misses that I've fixed for now but I'll do it a better way soon.

check_container

As this won't be merged in 1.7.6.1 I'll ~~probably close this pull request soon~~ reduce the scope of this pull request to keep [this fix](https://github.com/PrestaShop/PrestaShop/pull/15081/commits/f280e7d6889f0c15654cd206dcdc914ac4e3b8e3) so I can work in a community module instead. This way, so we get a chance to have more feedbacks then the core team will be free to backport the module into the Core and/or include the community module.

Thanks for your review, literally food for my brain :+1:

src/Adapter/Console/ContextLoader.php
PierreRambaud: Maybe better to change this line https://github.com/PrestaShop/PrestaShop/blob/develop/autoload.php#L31 in config.inc.php to load the legacy context because of this https://github.com/PrestaShop/PrestaShop/blob/develop/bin/console#L16

classes/controller/ConsoleController.php
```diff
@@ -0,0 +1,128 @@
...
+        parent::__construct();
+
+        $this->id = 0;
+        $this->controller_type = 'console';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function checkAccess()
```
PierreRambaud, line 44: Since we are targeting 7.1+ you can type all your functions and parameters

mickaelandrieu: Sadly, doing this will break the API as the former controller I extend doesn't type functions or arguments

mickaelandrieu: Sadly, doing this will break the API as the former controller I extend doesn't type functions or arguments

src/Adapter/Console/ContextLoader.php
mickaelandrieu: Hello @PierreRambaud, I see what you mean but IMHO the comment on the related file ("Allow call of Legacy classes from classes in /src and /tests") doesn't work as expected.

PierreRambaud: This is why it could be interesting to change the line and make it works

mickaelandrieu: I'm not sure of the impacts as this file is also used in all contexts, I'd razer not rely on it and focus on what we need when we are in Console context.

How about a new file named autoload.php in console folder instead ?

PierreRambaud: This file is only loaded in `admin-dev/index.php` and `./bin/console`.
No need to create a new one to do the job its must do

mickaelandrieu: that the thing: I don't think we should use this file at all in Console context: I will remove this call and see what happens :)

mickaelandrieu:
Tests are here, I can add a lot of tests commands from my own module (that allow people to enjoy this feature starting 1.7.4 or 5).

WDYT?

classes/controller/ConsoleController.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        return '';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function buildContainer()
+    {
+        return SymfonyContainer::getInstance();
```
mickaelandrieu, line 126: this function is not used atm, but as the parent function is supposed to return a Container...

mickaelandrieu:
Tests have been added 👍 , it's not in WIP anymore!

bin/console
```diff
@@ -12,8 +12,8 @@ use Symfony\Component\Debug\Debug;
 
 set_time_limit(0);
 
-require __DIR__.'/../vendor/autoload.php';
-require_once __DIR__.'/../autoload.php';
+require_once __DIR__.'/../vendor/autoload.php';
```
PierreRambaud:
```suggestion
```
Already loaded in config.inc.php

src/PrestaShopBundle/Resources/config/services/adapter/cms_page_category.yml
```diff
@@ -6,7 +6,7 @@ services:
 
     prestashop.adapter.cms_page_category.command_handler.abstract_cms_page_category_handler:
         class: PrestaShop\PrestaShop\Adapter\CMS\PageCategory\CommandHandler\AbstractCmsPageCategoryHandler
-        abstract: true
+        abstract:  true
```
PierreRambaud:
```suggestion
        abstract: true
```

mickaelandrieu: this file is not supposed to be there , i'll remove it

bin/console
mickaelandrieu: I know but without it, I have the feeling it won't works.

Let me remove it and let's see if this file is required or not :)

mickaelandrieu:
Hi @PierreRambaud,

comments addressed.

Thanks for review


> And tests pass :tada:

mickaelandrieu:
:/ I had to restore this branch as #14173 is not fixed, but I wont have time to update/maintain it.

PR #15083, opened 2019-08-12
Use aria-label instead of hidden sr-only span in variant links

ghost:
| Questions     | Answers
| Branch?       | develop
| Description?  | Small tweak for variant-links.tpl
| Type?         | refacto
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Use screen reader on the site

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15083)

### DIFF
--- a/themes/classic/templates/catalog/_partials/variant-links.tpl
+++ b/themes/classic/templates/catalog/_partials/variant-links.tpl
@@ -3,14 +3,10 @@
     <a href="{$variant.url}"
        class="{$variant.type}"
        title="{$variant.name}"
-       {*
-          TODO:
-            put color in a data attribute for use with attr() as soon as browsers support it,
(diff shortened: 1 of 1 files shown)


eternoendless (approved):

sarahdib:
Hello @pojebunny 
Thanks for your contribution.

Can you explain more how to test ? Like what is the resultat we expect ?

Thank you

ghost:
No real change is expected.
Before screen readers were expected to read the hidden span that could potentially disrupt pages flow / styling.
By replacing the hidden span with aria-label we avoid such problems and gain consistency.

If you want to 'test' that, you can use a screen reader on both versions of the file and see if they are read the same. There is no styling, nor scrypting attached to the hidden span that is removed.

matks:
Thank you @pojebunny

Issue #14709, opened 2019-07-17
[CLDR] Manage save and interpretation of currency customization (3/3)


colinegin:
After saving the currency new display, the saved information will override information from the CLDR files and create a new CLDR format to be interpreted. 

Blocked by 1.7.6.2 and 1/3 & 2/3

colinegin:
Hello @jolelievre , is this issue fixed by #14972 ?

jolelievre:
As explained in this other comment https://github.com/PrestaShop/PrestaShop/issues/14750#issuecomment-549368083
No the PR  #14972 is only the first one, two other PRs will be coming once it is merged.

PR #15084, opened 2019-08-12
Remove price precision in admin shop preferences 

### LINKED ISSUE #12223: [CLDR] Remove Number of decimals in Shop Parameter > General
**SPECIFICATION**
 
In Shop Parameter > General, we need to remove Number of decimals input. 
 
**WHY**
 
Since the CDLR handle every currency decimal, a general decimal parameter doesn't make sense. 

**FAILING CASE**

Need to study if there will be a technical impact.

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Remove the possibility to choose price display precision in back office, see detailed explanation below
| Type?         |improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | fixes #12223
| How to test?  | Check that it is not possible to change the price display precision in back office, and that prices display are not broken everywhere backend and frontend.

So what this PR does, more precisely:

- Remove display price precision from shop preferences form in back office
- Deprecate the PS_PRICE_DISPLAY_PRECISION constant
- Deprecates the _PS_PRICE_COMPUTE_PRECISION_ constant, set its value to 6 (instead of giving it the same value as PS_PRICE_DISPLAY_PRECISION)
- Add a service that calculates the price computing precision according to a currency's max digit number.
- Replace usages of PS_PRICE_DISPLAY_PRECISION with calculated price computing value using the formerly mentioned service. (the display is dealt with by formatting method)

- minor improvements

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15084)


config/config.inc.php
```diff
@@ -126,13 +125,20 @@
...
-define('_PS_PRICE_COMPUTE_PRECISION_', _PS_PRICE_DISPLAY_PRECISION_);
+/**
+ * @deprecated since 1.7.7
+ */
+define('_PS_PRICE_DISPLAY_PRECISION_');
+
+/**
+ * @deprecated since 1.7.7
+ */
+define('_PS_PRICE_COMPUTE_PRECISION_', 6);
```
matthieu-rolland, line 142: This is the only part that changed in this file, the rest is php-cs-fixer doing his thing.

src/Core/Localization/CLDR/ComputingPrecision.php
```diff
@@ -0,0 +1,50 @@
...
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Localization\CLDR;
+
```
matthieu-rolland, line 29: I was not so sure about where to put this service, do you think I put it at the right place? 
ComputingPrecision is very tightly related to CLDR, but one could argue that it is not CLDR per se.

classes/Context.php
```diff
@@ -99,6 +100,9 @@ class ContextCore
     /** @var Translator */
     protected $translator = null;
 
+    /** @var int */
+    protected $priceComputingPrecision;
```
PierreRambaud:
```suggestion
    protected $priceComputingPrecision = null;
```

classes/Context.php
```diff
@@ -453,4 +457,19 @@ protected function getTranslationResourcesDirectories()
...
     }
+
+    /**
+     * Returns the computing precision according to the current currency
+     *
+     * @return int
+     */
+    public function getComputingPrecision()
+    {
+        if (empty($this->priceComputingPrecision)) {
```
PierreRambaud:
```suggestion
        if ($this->priceComputingPrecision === null) {
```
otherwise we can't have 0 as precision

matthieu-rolland: Damn, you're right, (and welcome back!)

PierreRambaud: Still on vacation 🙄

matthieu-rolland: @PierreRambaud I fixed it, but I think I will revert to what I did initially,  because the value of priceComputingPrecision should never be zero, but at a minimum of two, so if for some weird reason something sets the value to zero, then we should consider it as if it was null and calculate its value again.

PierreRambaud: Weird to disallow the zero precision :thinking:

matthieu-rolland: The idea is that the computing precision should always be higher than the display precision.
(The rule we chose for now is computing precision = display precision * 3, with a minimum of 2)

behatoutput.txt
```diff
@@ -0,0 +1,3711 @@
+@reset-database-before-feature
```
PierreRambaud: :roll_eyes: Never use `git add -A` or `git commit -a`

install-dev/upgrade/sql/1.6.0.11.sql
```diff
@@ -24,8 +24,6 @@ ALTER TABLE `PREFIX_product_attribute_shop` MODIFY `unit_price_impact` DECIMAL(2
 
 /* PHP:p16011_media_server(); */;
 
-INSERT INTO `PREFIX_configuration` (`name`, `value`, `date_add`, `date_upd`) VALUES ('PS_PRICE_DISPLAY_PRECISION', 2, NOW(), NOW());
```
PierreRambaud: Do not remove upgrade script :thinking: 
Add the 
```sql
DELETE `PREFIX_configuration` WHERE name = 'PS_PRICE_DISPLAY_PRECISION';
```

in the 1.7.7.0.sql instead

src/Core/Localization/CLDR/ComputingPrecision.php
```diff
@@ -0,0 +1,52 @@
...
+     *
+     * @var int
+     *
+     * @return int
+     */
+    public function getPrecision(int $displayPrecision)
+    {
+        $computingPrecision = $displayPrecision * $this->multiplier;
+
+        return ($computingPrecision < 2) ? 2 : $computingPrecision;
```
PierreRambaud:
```suggestion
        return ($computingPrecision < static::MINIMAL_PRECISION) ? static::MINIMAL_PRECISION : $computingPrecision;
```

and add the const to the ComputingPrecision + can you add an interface for this class? and make it final?

src/Core/Localization/CLDR/ComputingPrecision.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace PrestaShop\PrestaShop\Core\Localization\CLDR;
+
+/**
+ *  Will calculate the computing precision (fraction digits number used for computations) that should
+ * be used for a given display precision.
+ */
+class ComputingPrecision
+{
+    private $multiplier = 3;
```
PierreRambaud: Since this value doesn't change, can you use a const?
```suggestion
    const MULTIPLIER = 3;
```

install-dev/upgrade/sql/1.6.0.11.sql
matthieu-rolland: oh ok, I thought that update scripts where all played successively at install, my mistake.

src/Core/Localization/CLDR/ComputingPrecision.php
matthieu-rolland: done !

behatoutput.txt
matthieu-rolland: 🤦‍♂ (fixed)

a quick `git add .` was the culprit here

src/Core/Localization/CLDR/ComputingPrecision.php
matthieu-rolland: done !

install-dev/upgrade/sql/1.6.0.11.sql
matthieu-rolland: done !

src/Core/Localization/CLDR/ComputingPrecision.php
matthieu-rolland: done

matthieu-rolland:
@PierreRambaud I treated all your feedbacks

classes/Context.php
```diff
@@ -453,4 +457,19 @@ protected function getTranslationResourcesDirectories()
...
+    /**
+     * Returns the computing precision according to the current currency
+     *
+     * @return int
+     */
+    public function getComputingPrecision()
+    {
+        if ($this->priceComputingPrecision === null) {
+            $computingPrecision = new ComputingPrecision();
+            $this->priceComputingPrecision = $computingPrecision->getPrecision($this->currency->precision);
```
jolelievre, line 485: What if `$this->currency` is empty?
This might happen in install context, or Symfony command where the Context is not always fully filled (like the employee is sometimes missing)

classes/Product.php
```diff
@@ -4846,7 +4846,7 @@ public static function getProductProperties($id_lang, $row, Context $context = n
                     true,
                     $quantity
                 ),
-                (int) Configuration::get('PS_PRICE_DISPLAY_PRECISION')
+                Context::getContext()->getComputingPrecision()
```
jolelievre, line 4896: Does this mean the display precision is now the same as the computing one? Or was it falsely used here?

install-dev/data/xml/configuration.xml
```diff
@@ -793,9 +793,6 @@ Country</value>
   <configuration id="PS_ROUND_TYPE" name="PS_ROUND_TYPE">
     <value>2</value>
   </configuration>
-  <configuration id="PS_PRICE_DISPLAY_PRECISION" name="PS_PRICE_DISPLAY_PRECISION">
```
jolelievre, line 799: I wonder if we should remove it or keep it for backward compatibility? 🤔

src/PrestaShopBundle/Form/Admin/Product/ProductCombinationBulk.php
```diff
@@ -58,7 +57,6 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $is_stock_management = $this->configuration->get('PS_STOCK_MANAGEMENT');
         $this->isoCode = $options['iso_code'];
-        $this->priceDisplayPrecision = $options['price_display_precision'];
```
jolelievre, line 61: Seems like it was useless anyway ^^
Maybe you should remove the default `price_display_precision` option as well?

tests-legacy/Integration/classes/CartGetOrderTotalTest.php
```diff
@@ -453,8 +453,8 @@ public function testBasicRoundTypeLine()
         $cart->updateQty(1, $product_a->id);
         $cart->updateQty(1, $product_b->id);
 
-        $this->assertEquals(3.59, $cart->getOrderTotal(false, Cart::ONLY_PRODUCTS));
-        $this->assertEquals(4.29, $cart->getOrderTotal(true, Cart::ONLY_PRODUCTS));
+        $this->assertEquals(3.581, $cart->getOrderTotal(false, Cart::ONLY_PRODUCTS));
```
jolelievre, line 456: Since the precision will likely be higher (3 * 2 quite often) Maybe you should increase the number of decimals to 6 digits? (At least in one product)

tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/Currency/currency.feature
```diff
@@ -90,8 +90,8 @@ Feature: Cart calculation with currencies
     Given currency "currency2" is the current one
     Given there is a product in the catalog named "product1" with a price of 19.812 and 1000 items in stock
     When I add 3 items of product "product1" in my cart
-    Then my cart total should be 83.05 tax included
-    Then my cart total using previous calculation method should be 83.05 tax included
+    Then my cart total should be 83 tax included
```
jolelievre, line 93: No more decimal?

tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/RoundingType/round_line.feature
```diff
@@ -34,5 +34,5 @@ Feature: Cart calculation with rounding type LINE
     When I add 2 items of product "product2" in my cart
     When I add 3 items of product "product1" in my cart
     When I add 1 items of product "product3" in my cart
-    Then my cart total should be precisely 162.41 tax included
-    Then my cart total using previous calculation method should be precisely 162.41 tax included
+    Then my cart total should be precisely 162.4 tax included
```
jolelievre, line 37: Maybe use `162.40` to avoid confusion (it would seem the decimals is rounded to one digit)

tests/Integration/Behaviour/Features/Scenario/Cart/CartToOrder/cart_to_order.feature
```diff
@@ -28,10 +28,10 @@ Feature: Check cart to order data copy
     When I select address "address1" in my cart
     When I select carrier "carrier1" in my cart
     When I validate my cart using payment module fake
-    Then current cart order total for products should be 20.6 tax included
-    Then current cart order total for products should be 19.81 tax excluded
-    Then current cart order total discount should be 10.3 tax included
-    Then current cart order total discount should be 9.91 tax excluded
+    Then current cart order total for products should be 20.604480 tax included
```
jolelievre, line 31: How come in some cases you have 6 digits and in other only 1 (or 2 or 3) Is it because the computation is exactly one digit or are there other rounding rules? If it's an exact value maybe add a pending 0 to avoid confusion?

jolelievre (commented):
Great work @matthieu-rolland 
I'm a bit worried about two things however:

You replaced `PS_PRICE_DISPLAY_PRECISION` with the computing precision which will be higher than the former display precision(2 digits in most cases) So I'm worried it will change the expected display, besides if we're talking about display then the CLDR should be in charge of the expected display.

Second why did you remove all the legacy tests for cart? Are they irrelevant now? and all validated by behat?

tests/Integration/Behaviour/Features/Scenario/Cart/CartToOrder/cart_to_order.feature
matthieu-rolland: Since we increased the precision, in this case we can have up to 6 digits, but that doesn't mean we will always have 6 digits, the end result of a calculation could still only have one digit, I believe this is normal 🤔

tests-legacy/Integration/classes/CartGetOrderTotalTest.php
matthieu-rolland: You mean having a test where the end result has 6 digits? That's a good idea, I'll do that 👍

classes/Product.php
matthieu-rolland: Yes it was falsely used... the problem is that display precision and computing precision values were always the same, and in the code sometimes they were used indistinctly without any consequences.

The only parts that really use some kind of display precision are the formatting methods, that are used for prices that will be displayed to an end user. In all other cases, we want to round prices according to the computing precision.

install-dev/data/xml/configuration.xml
matthieu-rolland: I think one of the main goals of this PR was to remove this field in the back office.  And as you stated below it has become useless ;)

src/PrestaShopBundle/Form/Admin/Product/ProductCombinationBulk.php
matthieu-rolland: Nice catch! Done.

matthieu-rolland:
> Great work @matthieu-rolland
> I'm a bit worried about two things however:
> 
> You replaced `PS_PRICE_DISPLAY_PRECISION` with the computing precision which will be higher than the former display precision(2 digits in most cases) So I'm worried it will change the expected display, besides if we're talking about display then the CLDR should be in charge of the expected display.
> 
> Second why did you remove all the legacy tests for cart? Are they irrelevant now? and all validated by behat?

@jolelievre I understand your concerns, here are some more things to keep in mind, don't hesitate to tell me if you think we missed something (that's quite possible).

1/ In some places `PS_PRICE_DISPLAY_PRECISION` was used indistinctly in place of compute price precision, and since both always had the same values it would not cause any misbehavior.
From now on, price computing precision is used for all calculations, the idea of a display price precision is only relevant in formatting methods, for values that are supposed to be displayed to the end user.

2/ Yes I've been asked to remove those specific tests because they have been replicated on behat.

jolelievre (dismissed):
Ok @matthieu-rolland Seems fine for me ;)

jolelievre:
@matthieu-rolland  But travis and PrettyCI disagree with me ^^

matthieu-rolland:
> @matthieu-rolland But travis and PrettyCI disagree with me ^^

yes, I'm investigating this ! thanx

classes/Context.php
jolelievre: Maybe have a fallback in case Context::currency is null?

jolelievre (dismissed):
Nice work @matthieu-rolland

tests/Integration/Behaviour/Features/Context/CurrencyFeatureContext.php
```diff
@@ -74,13 +74,15 @@ public function thereIsACurrency($currencyName, $currencyIsoCode, $changeRate)
         if (!$currencyId) {
             $currency = new Currency();
             $currency->name = $currencyIsoCode;
+            $currency->precision = 2;
```
PierreRambaud, line 77: Any reason to set 2 here and having 4 in behat tests?

PierreRambaud: Or even sometimes 3 :thinking:

jolelievre: Probably because for now the CLDR uses the pattern to determine the number of decimals Not sure about it though because https://github.com/PrestaShop/PrestaShop/pull/15755 seems to use the currency precision Anyway this issue https://github.com/PrestaShop/PrestaShop/issues/14709 will deal with this problem more in details with appropriate Behat tests

But in the end I think the tests you are talking about manage the computing, so they are based on ComputingPrecision which by default is `3 * currency.precision => 6` That's why we have more decimals in the tests now The CLDR only applies on display so it's not the same value as for the 3 or 4 number of digits I guess it's only about the value

@matthieu-rolland correct me if I'm wrong on this one

jolelievre (approved):

tests/Integration/Behaviour/Features/Context/CurrencyFeatureContext.php
PierreRambaud: Better with explanations :D

matthieu-rolland: exactly @jolelievre 👍

PierreRambaud:
Thanks @matthieu-rolland

PR #15092, opened 2019-08-13
Update some outdated license headers for 1.7.6.x

matks:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Update some outdated license headers for 1.7.6.x
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15092)

### DIFF
--- a/admin-dev/themes/new-theme/tests/cldr/number-symbol.spec.js
+++ b/admin-dev/themes/new-theme/tests/cldr/number-symbol.spec.js
@@ -1,3 +1,27 @@
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
 import {expect} from 'chai';
 import NumberSymbol from '../../js/app/cldr/number-symbol';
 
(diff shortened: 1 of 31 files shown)


admin-dev/themes/new-theme/tests/cldr/number-symbol.spec.js
```diff
@@ -1,3 +1,27 @@
...
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
```
matks, line 24: I dont know if we should put license headers on JS tests files 🤔 we do it for php files, but the goal of license headers is to be shipped in the Zip Archive, right ? so these tests are fine

By the way @eternoendless we should consequently ignore these files in the Release Builder, right ?

matks:
Done using SF command `./bin/console prestashop:licenses:update`

admin-dev/themes/new-theme/tests/cldr/number-symbol.spec.js
eternoendless: > I dont know if we should put license headers on JS tests files

People can still pick up our files from Github so I think we should put the header on ALL files.

> we should consequently ignore these files in the Release Builder, right ?

Yes we should

src/PrestaShopBundle/Translation/Provider/TranslationFinder.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * 2007-2019 PrestaShop SA and Contributors
+ * 2007-2019 PrestaShop and Contributors
```
eternoendless, line 4: It's "PrestaShop SA and Contributors" ([see devdocs](https://devdocs.prestashop.com/1.7/development/coding-standards/#license-information))

eternoendless (changes requested):

src/PrestaShopBundle/Translation/Provider/TranslationFinder.php
matks: This is not what is being computed by the UpdateLicensesCommand
https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Command/UpdateLicensesCommand.php#L40

Are you sure the devdocs is right and the code is wrong ? If yes, we need to update all license headers for both 1.7.6.x and develop 😞

eternoendless: Yep I'm positive

matks: Issue created https://github.com/PrestaShop/PrestaShop/issues/15116

eternoendless:
Replaced by #15125

PR #15093, opened 2019-08-13
Fix duplicate products with attributes and suppliers

202-ecommerce:
| Questions     | Answers
| Branch?       | develop
| Description?  | In some case of products with attributes and suppliers the duplication of suppliers for attribute return an SQL error of duplicate entry.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | 1. Install a PrestaShop with default catalog on develop branch<br>2. Create a supplier and activate it.<br>3. Edit the first product - it sould be with attributes (for instance) and add a supplier on options tab. Save it.<br>4. On the list of product, clic on verticlal "..." at the end of the line and select "Duplicate".<br>5. You sould see on debug mode following errors<br /><br />Duplicate entry '22-41-1' for key 'id_product'<br />INSERT INTO `ps_product_supplier` (`id_product`, `id_product_attribute`, `id_supplier`, `product_supplier_reference`, `product_supplier_price_te`, `id_currency`) VALUES ('22', '41', '1', '', '0.000000', '1')

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15093)

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -4342,7 +4342,7 @@ class ProductCore extends ObjectModel
                 unset($row3['id_product_supplier']);
                 $row3['id_product'] = $id_product_new;
                 $row3['id_product_attribute'] = $id_product_attribute_new;
-                $return &= Db::getInstance()->insert('product_supplier', $row3);
+                $return &= Db::getInstance()->insert('product_supplier', $row3, false, true, Db::INSERT_IGNORE);
             }
         }
 


matks:
Hi @202-ecommerce I dont reproduce the issue 🤔 maybe an extra step is needed in the "How to test" to see it ?

When I check the code, nothing seems wrong in it so I'm not in favor of using `Db::INSERT_IGNORE` if we can avoid it as it is hidden a true issue. If a duplicate SQL key error happens, we should aim to fix the code, not hide the error message 😉 .

202-ecommerce:
OK we have check again with last develop prestashop version and the bug disappear.

I close this PR.

Thank you.

Issue #15088, opened 2019-08-12
Logo is missing on invoice


MathiasReker:
**Describe the bug**
Logo is missing on invoice if SSL is enabled.

**To Reproduce**
See this screenrecord: http://reker.dk/ssl.webm
Tested with 100% clean install of 1.7.6.0 on Windows with open SSL.

**Screenshots**
If applicable, add screenshots or screenrecords to help explain your problem.
![image](https://user-images.githubusercontent.com/26626066/62892105-b8aa3180-bd47-11e9-9662-7783c50bdabc.png)

**Additional information**
PrestaShop version: 1.7.6.0
PHP version: 7.2

MathiasReker:
Fixed here: https://github.com/PrestaShop/PrestaShop/pull/15089

mariem-abid:
Hi @MathiasReker,

Thanks for your fix.
I manage to reproduce the issue with the last version PS 1.7.6.0
I’ll add this to the debug roadmap

MathiasReker:
New PR: https://github.com/PrestaShop/PrestaShop/pull/15109

PR #15109, opened 2019-08-14
Fix missing invoice logo

MathiasReker:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Logo on invoice was missing if SSL was activated in BO. Use URI instead of URL fix the issue.
| Type?         | bug fix
| Category?     | FO
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/15088
| How to test?  | Enable SSL in BO. Then generate an invoice. Without this PR the logo is missing. With this PR the logo is showed.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15089)

### DIFF
--- a/classes/pdf/HTMLTemplate.php
+++ b/classes/pdf/HTMLTemplate.php
@@ -136,7 +136,7 @@ abstract class HTMLTemplateCore
             $width *= $ratio;
         }
 
-        if (_MEDIA_SERVER_1_ || _MEDIA_SERVER_2_ || _MEDIA_SERVER_3_ !== '') {
+        if ('' !== _MEDIA_SERVER_1_ || _MEDIA_SERVER_2_ || _MEDIA_SERVER_3_) {
             $logo_path = Tools::getShopProtocol() . Tools::getMediaServer(_PS_IMG_) . _PS_IMG_ . $logo;
         } else {
             $logo_path = _PS_IMG_ . $logo;
--- a/classes/pdf/HTMLTemplate.php
+++ b/classes/pdf/HTMLTemplate.php
@@ -136,8 +136,14 @@ abstract class HTMLTemplateCore
             $width *= $ratio;
         }
 
+        if (_MEDIA_SERVER_1_ || _MEDIA_SERVER_2_ || _MEDIA_SERVER_3_ !== '') {
+            $logo_path = Tools::getShopProtocol() . Tools::getMediaServer(_PS_IMG_) . _PS_IMG_ . $logo;
+        } else {
+            $logo_path = _PS_IMG_ . $logo;
+        }
+
         $this->smarty->assign(array(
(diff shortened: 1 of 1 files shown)


MathiasReker:
@matthieu-rolland I am not familiar using the CDN server, will you have a look on this way to fix it?

classes/pdf/HTMLTemplate.php
```diff
@@ -136,8 +136,14 @@ public function assignCommonHeaderData()
...
 
+        if ('' !== _MEDIA_SERVER_1_ || _MEDIA_SERVER_2_ || _MEDIA_SERVER_3_) {
+            $logo_path = Tools::getShopProtocol() . Tools::getMediaServer(_PS_IMG_) . _PS_IMG_ . $logo;
+        } else {
+            $logo_path = _PS_IMG_ . $logo;
+        }
+
         $this->smarty->assign(array(
-            'logo_path' => Tools::getShopProtocol() . Tools::getMediaServer(_PS_IMG_) . _PS_IMG_ . $logo,
+            'logo_path' => $logo_path,
```
PierreRambaud, line 146: Because of `img_ps_dir` you only need to do this:
```suggestion
            'logo_path' => _PS_IMG_ . $logo,
```

PierreRambaud (changes requested):
Follow my comment.

PierreRambaud:
Nice catch @MathiasReker :wink:

MathiasReker:
> 
> 
> Nice catch @MathiasReker 😉

I though so, but @matthieu-rolland said that it does not work with CDN

matthieu-rolland:
yes @MathiasReker , I was afraid removing the getMediaServer server method would keep users from using their cdn for this logo, it was safer to ask a review from someone who knew better about this. ;)

PierreRambaud:
I will check another time and come back to you. I need to test one more thing :)

PierreRambaud:
After checking, I confirm there is no problem.
You're probably using a selfsigned certificate, and the TCPDF can't retrieve your image over the network because of verify_peer parameter.
Without ssl the logo_path is `http://ps-develop.localhost/img/logo.png`, and with `https://ps-develop.localhost/img/logo.png`, without the certificate verification, I'm not able to see the image. With, everything is ok.

PierreRambaud:
If you want to try with https, you can change the `logo_path` with 
`https://www.prestashop.com/sites/all/themes/prestashop/images/logo_ps_second.svg` for example, you will see everything is ok. I also tried with a proper https configuration.

MathiasReker:
So this PR can be closed, since there is not problem in production?

PierreRambaud:
Closed because of self signed certificate (Not working with TCPDF).

PR #14809, opened 2019-07-22
Allow modifying email body contents

jolelievre:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Add translation provider to allow modifying body contents, add a form in Email Theme page to access this edition
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15474
| How to test?  | Go to `Design > Email Theme` page in the "Translate Emails" page select a language and start editing some texts. Check that the modifications are update in the mails previews, then re-generate emails and tests the modifications in real condition.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14809)

PR #15139, opened 2019-08-16
Make module and email translations work again when using a theme other than classic

### LINKED ISSUE #15133: Cannot translate modules when using a theme other than classic
**Describe the bug**
When using a theme other than classic, you cannot translate modules

**To Reproduce**
1. Activate a theme other than classic
2. Try to translate some installed module (a native one for example)
3. You get a domain tree, then the error message "Could not crawl for translation files: The "/path/to/prestashop/var/cache/prod/themes/TEMPLATE/translations" directory does not exist."
4. Translations don't appear afterwards

**Screenshots**
https://drive.google.com/file/d/1OvlF41Oo-6BhzIx6e9ULAKIk2CI7VTSE/view
Screenshot 2019-08-16 at 16 27 10

**Additional information**
PrestaShop version: 1.7.6.x
PHP version: 7.1

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Fixes a bug where it would be impossible to translate modules and email subjects when the current theme was anything other than "classic"
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15133
| How to test?  | Install a different theme than classic. Test all the translation types (except Email body which is handled differently and not affected by this change).

## More information

There was much confusion between the "selected" element (a module name when translating modules, a theme name when translating themes) and the selected theme. Up until now, it was half-managed by request parameters, half-forced to the current theme.

This change clarifies things a little:

- Module translations are no longer attached to the currently selected theme (you'll have to edit the theme translations if you want a different translation for each theme; this requires the specific wording to be used in the theme's templates or else it won't appear in the translation interface)
- Only theme translations are attached to a theme. All other translations are now global.
- Email translations no longer are attached to an invalid theme called "subject"

Regarding #14809, we'll have to discuss changes in how Mail translations are managed so that email translations work like modules (where `type` would be 'mails' and `selected` would be either `subject` or `body`). But this will be problematic, because I think that in difference to email subjects, which are global, mail bodies are theme-dependent.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15139)


src/PrestaShopBundle/Resources/views/Admin/Translations/overview.html.twig
```diff
@@ -35,7 +35,9 @@
         var data = {
             baseUrl: '{{ app.request.getBasePath() }}',
             locale: '{{ app.request.query.get('locale') }}',
-            selected: '{{ 'modules' == app.request.query.get('type') ? theme.name : app.request.query.get('selected') }}',
+            type: '{{ app.request.query.get('type') }}',
+            currentTheme: {{ theme.name|json_encode|raw }},
```
PierreRambaud, line 39: watchout with raw.
```suggestion
            currentTheme: '{{ theme.name|escape('js') }}',
```

mickaelandrieu:
> Module translations are no longer attached to the currently selected theme (you'll have to edit the theme translations if you want a different translation for each theme; this requires the specific wording to be used in the theme's templates or else it won't appear in the translation interface)

If I've well understood, this introduces is a regression to fix a bug, right?

admin-dev/themes/new-theme/js/app/pages/translations/components/principal/index.vue
```diff
@@ -129,21 +129,22 @@
...
-        if (
-          this.currentDomainTotalMissingTranslations
-          && this.currentDomainTotalMissingTranslations === 1
-        ) {
-          totalMissingTranslationsString = this.trans('label_missing_singular');
-        } else if (this.currentDomainTotalMissingTranslations) {
-          totalMissingTranslationsString = this.trans('label_missing').replace('%d', this.currentDomainTotalMissingTranslations);
+        if (this.currentDomainTotalMissingTranslations) {
+          if (this.currentDomainTotalMissingTranslations === 1) {
+            totalMissingTranslationsString = this.trans('label_missing_singular');
```
matks, line 144: Some extra polishing 😛

src/PrestaShopBundle/Controller/Api/TranslationController.php
```diff
@@ -140,10 +140,24 @@ public function listTreeAction(Request $request)
...
+                    $tree = $this->getModulesTree($lang, $selected, $search);
+                    break;
+
+                case 'mails':
+                    // when emails body will be implemented, it should be a different type
+                    // because domain routes only support "type" & "selected/theme" as parameters
+                    $tree = $this->getMailsSubjectTree($lang, $search);
+                    break;
+
+                default:
```
matks, line 158: What else can be translated ?
If we expect nothing else to be translated, maybe we should throw an exception in the `default` ?

src/PrestaShopBundle/Controller/Api/TranslationController.php
```diff
@@ -307,65 +321,79 @@ protected function guardAgainstInvalidTranslationResetRequest($content)
...
      * @param $lang
      * @param $type
-     * @param $selected
+     * @param string $theme Selected theme name
      * @param null $search
      *
      * @return array
      */
-    private function getNormalTree($lang, $type, $selected, $search = null)
+    private function getNormalTree($lang, $type, $theme, $search = null)
```
matks, line 329: If `getNormalTree` is supposed to only be called for Themes, I suggest we strongly validate the input by checking `$theme` actually exists and refers to an installed theme, or throw an exception.

src/PrestaShopBundle/Controller/Api/TranslationController.php
```diff
@@ -307,65 +321,79 @@ protected function guardAgainstInvalidTranslationResetRequest($content)
...
-     * @param $selected
-     * @param null $search
+     * @param string $lang Two-letter iso code
+     * @param string $selectedModuleName Selected module name
+     * @param string|null $search
      *
      * @return array
      */
-    private function getModulesTree($lang, $type, $selected, $search = null)
+    private function getModulesTree($lang, $selectedModuleName, $search = null)
```
matks, line 344: I suggest same defensive logic: if `$selectedModuleName` is illogical, we crash. Translations are a mess right now, this would clear the path and avoid ambiguous successfull behaviors

matks: (maybe not in a patch version though 😛)

src/PrestaShopBundle/Controller/Api/TranslationController.php
matks: (maybe not in a patch version though 😛)

src/PrestaShopBundle/Controller/Api/TranslationController.php
matks: (maybe not in a patch version though 😛)

matks (commented):
OK for me, same feedback than @PierreRambaud

src/PrestaShopBundle/Controller/Api/TranslationController.php
jolelievre: Yes there are other types, like `back` `others`

src/PrestaShopBundle/Controller/Api/TranslationController.php
jolelievre: Since we planned a refacto maybe it should be done in the refacto?

src/PrestaShopBundle/Controller/Api/TranslationController.php
```diff
@@ -307,65 +321,79 @@ protected function guardAgainstInvalidTranslationResetRequest($content)
...
     {
+        $theme = null;
         $locale = $this->translationService->langToLocale($lang);
+
         $moduleProvider = $this->container->get('prestashop.translation.external_module_provider');
-        $moduleProvider->setModuleName($selected);
+        $moduleProvider->setModuleName($selectedModuleName);
 
-        $treeBuilder = new TreeBuilder($locale, $this->getSelectedTheme());
+        $treeBuilder = new TreeBuilder($locale, $theme);
```
jolelievre, line 352: So `$theme` is always `null` ?

src/PrestaShopBundle/Controller/Api/TranslationController.php
```diff
@@ -307,65 +321,79 @@ protected function guardAgainstInvalidTranslationResetRequest($content)
...
     }
 
     /**
-     * @return \PrestaShop\PrestaShop\Core\Addon\Theme\Theme
-     *
-     * @todo: When the theme will be selectable in the form, this function should be updated accordingly.
+     * @return string
      */
-    private function getSelectedTheme()
+    private function getCurrentThemeName()
```
jolelievre: So it seems this method is no longer used?

matks:
Restarting Travis again

matks:
And again

matks:
And again

I could _almost_ believe the test fails for a good reason

PierreRambaud:
And again!

Always the same random reason :/

src/PrestaShopBundle/Resources/views/Admin/Translations/overview.html.twig
eternoendless: That won't work with null, functions are not json encodable, and html would be encapulated in a string. I'm I missing something?

admin-dev/themes/new-theme/js/app/pages/translations/components/principal/index.vue
eternoendless: A little bit of the ol' boy scout rule

src/PrestaShopBundle/Controller/Api/TranslationController.php
eternoendless: ^ What jo answered

src/PrestaShopBundle/Controller/Api/TranslationController.php
eternoendless: This is called on line 159 with a null theme (`BO` translations and `other`)

src/PrestaShopBundle/Controller/Api/TranslationController.php
eternoendless: I think it will crash but deeper in the call stack. But we agree that this whole thing needs to be refactored.

src/PrestaShopBundle/Controller/Api/TranslationController.php
eternoendless: Yes when translating modules. Otherwise you'd have to translate your modules again for every theme (including if you switch themes). With this change only the wordings used in module templates overridden by a theme will be allowed to have a specific translation for that theme. This could be improved later by adding an option that brings ALL wordings, including core and modules, and that allows translating them for a specific theme.

src/PrestaShopBundle/Controller/Api/TranslationController.php
eternoendless: Right, I'll delete it

eternoendless:
I removed the unused method and rebased.

> If I've well understood, this introduces is a regression to fix a bug, right?

Yes and no. It made no sense, it was counterintuitive, and in the end, it didn't work. It's a regression in the same way as fixing a bug is a regression: "But I liked that bug, put it back!" 😛

The system has to be rethought, and in most importantly... **specified**.

src/PrestaShopBundle/Resources/views/Admin/Translations/overview.html.twig
PierreRambaud: If it's not working with null.
```suggestion
            currentTheme: {% if theme.name === null %}null{% else %}'{{ theme.name|escape('js') }}'{% endif %},
```

src/PrestaShopBundle/Controller/Api/TranslationController.php
jolelievre: then you don't need this `$theme` variable you can simply call `$treeBuilder = new TreeBuilder($locale, null);`

PierreRambaud (approved):

matks:
Thank you @eternoendless

src/PrestaShopBundle/Resources/views/Admin/Translations/overview.html.twig
eternoendless: Looks overcomplicated when json_encode is build exactly for that purpose: it outputs the right Javascript syntax for whatever value you throw at it.

src/PrestaShopBundle/Controller/Api/TranslationController.php
eternoendless: I did it so that the parameter role is evident, but it could be removed as well later.

sathishkumar16:
When we will get this as an offical update in the prestashop v1.7.6.X.

eternoendless:
@sathishkumar16 Soon.

sathishkumar16:
@eternoendless  How many days will it take to get the updates officially, because we are planning to go production within a month for us the translation module is very important before going to prod. 
Else is there any alternate/temporary solution to fix it?

eternoendless:
@sathishkumar16 The QA process is looking good so it won't be long, probably within the week. You can follow the discussion on [Gitter](https://gitter.im/PrestaShop/General) or wait for the announcement on [Build](http://build.prestashop.com/).

PR #15151, opened 2019-08-19
All BO classes should inherit from BObasePage

boubkerbribri:
| Questions     | Answers
| Branch?       | develop
| Description?  | All BO classes should inherit from BObasePage
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | npm run smoke-tests

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15151)

### DIFF
--- a/tests/puppeteer/.eslintrc.js
+++ b/tests/puppeteer/.eslintrc.js
@@ -17,6 +17,6 @@ module.exports = {
   },
   rules: {
     'no-plusplus': [2, { allowForLoopAfterthoughts: true }],
-    "max-len": [2, {"code": 120}]
+    'max-len': [2, {code: 120}]
   },
 };
--- a/tests/puppeteer/.eslintrc.js
+++ b/tests/puppeteer/.eslintrc.js
@@ -16,6 +16,7 @@ module.exports = {
     sourceType: 'module',
   },
   rules: {
-    'no-plusplus': [2, { allowForLoopAfterthoughts: true }]
+    'no-plusplus': [2, { allowForLoopAfterthoughts: true }],
+    "max-len": [2, {"code": 120}]
   },
 };
--- a/tests/puppeteer/pages/commonPage.js
+++ b/tests/puppeteer/pages/commonPage.js
@@ -81,11 +81,11 @@ module.exports = class CommonPage {
     switch (parameter) {
       case 'equal':
         await this.page.$eval(selector, el => el.innerText)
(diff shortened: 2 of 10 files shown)


SimonGrn (dismissed):
Thanks ! I don't know why they were still inheriting from CommonPage, that's much better this way

matthieu-rolland (dismissed):
LGTM

tests/puppeteer/.eslintrc.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
     sourceType: 'module',
   },
   rules: {
-    'no-plusplus': [2, { allowForLoopAfterthoughts: true }]
+    'no-plusplus': [2, { allowForLoopAfterthoughts: true }],
+    "max-len": [2, {"code": 120}]
```
PierreRambaud:
```suggestion
    'max-len': [2, {code: 120}]
```

tests/puppeteer/pages/commonPage.js
```diff
@@ -81,11 +81,11 @@ module.exports = class CommonPage {
     switch (parameter) {
       case 'equal':
         await this.page.$eval(selector, el => el.innerText)
-          .then(text => expect(text.replace(/\s+/g, ' ').trim()).to.equal(textToCheckWith));
+          .then(text => global.expect(text.replace(/\s+/g, ' ').trim()).to.equal(textToCheckWith));
```
PierreRambaud, line 84: why no import expect at the top of the file?
```
import {expect} from 'chai';
```

boubkerbribri: if so, we had to import it in each file, so we choose to do it in helper as a global var.

tests/puppeteer/.eslintrc.js
boubkerbribri: Eslint don't detect errors in eslint config file.
didn't see it, thank you

tests/puppeteer/pages/commonPage.js
PierreRambaud: Bad programming practice :(

boubkerbribri: Ok, good to know, I will do it in an other dedicated PR (to be very simple to review).
Can you tell why it's a bad practice ?

PierreRambaud (approved):

Issue #14384, opened 2019-06-26
Multiple failures from Travis on build PRESTASHOP_TEST_TYPE=e2e EXTRA_TESTS=functional


matks:
We have a insane number of CI builds failing because of a random occurrence of a test failure:

```
Guest Checkout
  1) "before all" hook
26 passing (27.50s)
1 failing
1) "before all" hook:
element (#blockcart-modal) still not visible after 2500ms
running chrome on Windows 7
Error: element (#blockcart-modal) still not visible after 2500ms
    at Object.<anonymous> (node_modules/webdriverio/lib/commands/waitForVisible.js:56:19)
    at Object.safeExecute (node_modules/webdriverio/lib/utils/safeExecute.js:19:24)
    at Object.resolve (node_modules/webdriverio/lib/webdriverio.js:87:29)
    at node_modules/webdriverio/lib/webdriverio.js:227:32
    at _rejected (node_modules/webdriverio/node_modules/q/q.js:830:24)
    at node_modules/webdriverio/node_modules/q/q.js:856:30
    at Promise.when (node_modules/webdriverio/node_modules/q/q.js:1108:31)
    at Promise.promise.promiseDispatch (node_modules/webdriverio/node_modules/q/q.js:774:41)
    at node_modules/webdriverio/node_modules/q/q.js:590:44
    at runSingle (node_modules/webdriverio/node_modules/q/q.js:123:13)
npm ERR! Test failed.  See above for more details.
```

@Quetzacoalt91 already checked in localhost, test pass. And re-running the build often fixes the issue.

We need to fix it.

Quetzacoalt91:
Here are additional details:

We often see that the selenium tests fail for the same reason. The modal after clicking on "Add to cart" is never displayed. Here is the screenshot taken on Travis when the error occurs:

![image](https://user-images.githubusercontent.com/6768917/60189017-2d70ed00-9828-11e9-94e3-cc4e212ec7bc.png)

This is not caused by an exception as we may think, but a product with customizable data. It has been altered by the script `prepare-shop.php`: https://github.com/PrestaShop/PrestaShop/blob/430dd959e92c6adb0f77bdc65b846796b7789cfe/tests/Selenium/prepare-shop.php#L110-L133

However, it appears that during the test, we sometime don't select the proper product. Although tyhe selector is correct. Maybe the products are filtered during a few milliseconds which makes the click incorrect?

matks:
[...]

Issue #14497, opened 2019-07-04
Rework of the E2E test suite run by Travis


matks:
Travis runs some E2E tests using Selenium. We've had multiple issues over time with this test suite: false positives, random failures, bad test settings.

Recently we had to disable one scenario because it was failing randomly: see issue #14384
and the related PR https://github.com/PrestaShop/PrestaShop/pull/14491

We need to rebuild this E2E test suite to make it more reliable.

matks:
ping @SimonGrn 

Here's the issue to be done for 1.7.7 about Travis web acceptance tests

SimonGrn:
We will only have smoke and sanity tests before the 1.7.7 is shipped, no E2E tests and I don't think we'll ever have since we want to have functional/E2E tests only for the nightly. Is that a problem ?

matks:
> We will only have smoke and sanity tests before the 1.7.7 is shipped, no E2E tests and I don't think we'll ever have since we want to have functional/E2E tests only for the nightly. Is that a problem ?

Smoke & sanity will do 👍 we talk about E2E tests here but we mean "web acceptance tests, powered by either Selenium or Puppeeter". Remember that so far, only you has a different meaning for E2E 😉 most of us think that E2E tests = "tests that start a browser and interact with the HTML and JS interface".

For example, if you pick https://engineering.wingify.com/posts/e2e-testing-with-webdriverjs-jasmine/ as an example:
"e2e or end-to-end or UI testing is a methodology used to test whether the flow of an application is performing as designed from start to finish."
=> so far so good 😄 
"In simple words, it is testing of your application from the user endpoint where the whole system is a blackbox with only the UI exposed to the user."
=> :trollface:

SimonGrn:
"From start to finish" is the only relevant piece of info here. Smoke and sanity don't test from start to finish, neither unit tests. I don't really understand why they put UI testing and E2E testing together as it was never mixed up when it was introduced to me.
Anyway, we got the spirit and that's the most important !

matks:
> I don't really understand why they put UI testing and E2E testing together as it was never mixed up when it was introduced to me.

Because most developers are introduced to web acceptance testing (meaning: using a browser to interact with the app) by tutorials and conference talks, not professionals. So we mix things up.

matks:
Fixed by the wonderful QA team in 1.7.7

Congratulations @boubkerbribri @nesrineabdmouleh @SimonGrn

PR #15161, opened 2019-08-20
Remove another faulty e2e test from Travis

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Disable another Selenium Guest scenario as it fails randomly
| Type?         | bug fix
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/14384
| How to test?  | Travis is green

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15161)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,14 +27,11 @@ env:
   matrix:
     - PRESTASHOP_TEST_TYPE=unit
     - PRESTASHOP_TEST_TYPE=integration
-    - PRESTASHOP_TEST_TYPE=e2e
 
 matrix:
   include:
     - php: 7.2
       env: EXTRA_DEPS=phpHigh  PRESTASHOP_TEST_TYPE=unit
-    - php: 7.2
-      env: PRESTASHOP_TEST_TYPE=e2e EXTRA_TESTS=functional
     - stage: deploy
       php: 7.2
       before_install: skip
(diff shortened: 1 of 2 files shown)


PierreRambaud:
no more scenario \o/

matks:
> no more scenario \o/

plz hit the "approve" button

PierreRambaud (approved):

PierreRambaud:
But if we don't have any more scenario, should be remove travis jobs?

PierreRambaud (changes requested):
Just a question

matks:
> Just a question

Yes and no 😐 

Yes:
- these builds are useless now
- the code is dead and would be better removed

No:
- we should replace the broken tests ASAP so the builds should be useful again soon
- if we keep the builds around, it will "remind" us not to forget about updating them

PierreRambaud:
It currently add useless jobs, that are consuming resources.
It's like keeping a commented code to be sure we will not forget to updates it.

We will be able to retrieve travis configuration in the git history :wink:

matks:
Removed Travis build too.

Will be reinserted when https://github.com/PrestaShop/PrestaShop/issues/14497 is implemented

PierreRambaud (approved):

matks:
OK I think Travis lost my build now

Travis 🙄 srsly ? You dont want to be disabled so badly, so you hide the build result ?

matks:
It's done

Issue #15144, opened 2019-08-17
Bulk invoice generating -  Price specification not found for currency - fatal error -  1.7.6


01park:
When generating invoices in bulk. I get a fatal error when the currency is different than EUR.

![Screenshot 2019-08-17 at 20 23 49](https://user-images.githubusercontent.com/46492443/63216438-14503280-c12d-11e9-9024-e3cc58bc606b.png)

I tried turning on-off currency. Checked the ps_products_supplier in DB ( read about this in a different post, but this is empty ) I also already changed the number of decimals to 2.

Not sure if this has anything do with the currency problems but I noticed a difference in format when checking the log in debug mode "PLN" and "PLN"" : see image

![Screenshot 2019-08-17 at 20 44 49](https://user-images.githubusercontent.com/46492443/63216645-3d25f700-c130-11e9-96a3-75c87da3a22c.png)

khouloudbelguith:
Hi @01park,

>I tried turning on-off currency. Checked the ps_products_supplier in DB

This issue occurs only when we import products.
>Not sure if this has anything do with the currency problems but I noticed a difference in format when checking the log in debug mode "PLN" and "PLN"" : see image

You have a double quote in the same sentence 
![image](https://user-images.githubusercontent.com/16067358/63259131-7d42c200-c275-11e9-8b83-556325815ca4.png)

>When generating invoices in bulk. I get a fatal error when the currency is different than EUR.

How did you generate the invoices? from the invoices page? By order status?

Thanks!

01park:
Hi @khouloudbelguith 

When I generate an invoice from the order page, it works fine ( 1 by 1)

But when I do BULK, I do it on this page.
<img width="1702" alt="Screenshot 2019-08-19 at 12 16 04" src="https://user-images.githubusercontent.com/46492443/63261337-2809af00-c27b-11e9-8be5-032c991b0971.png">

Does this double quote need to be removed?

khouloudbelguith:
@01park,

>Does this double quote need to be removed?

No, it is a debug message.

This is a fresh installation PS1.7.6.0 or after an upgrade from a previous version?

Thanks!

01park:
@khouloudbelguith 

Its an update from 1.7.5.2 ( multistore ). 

Not sure if it makes a difference, but I did not upgrade the classic theme files. This option was disabled.

khouloudbelguith:
@01park, thanks for your feedback.
[...]

PR #15173, opened 2019-08-20
CLDR has access to all currencies regardless of the current shop

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | The CLDR used to get the currencies list related to the current shop, which caused errors in some specific contexts where orders are shared between shops, the CLDR could not display an unknown currency So from now on it can display any currency installed regardless of the shop context BUT this required to add a new method in the `CurrencyDataProviderInterface` which is a BC break as we change the interface
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/14595 and https://github.com/PrestaShop/PrestaShop/issues/15144
| How to test?  | See the issues which describes precisely at the end how to reproduce the bug (two shops with their own currencies), then try to access an order from Shop B in Shop A context, the order will be correctly displayed When you try to export orders it will also work BUT we need to check that in the front office we only display the expected currencies (one per shop)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15173)

### DIFF
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -399,14 +399,20 @@ class CurrencyCore extends ObjectModel
     /**
      * Retrieve all currencies data from the database.
      *
+     * @param bool $active If true only active are returned
+     * @param bool $groupBy Group by id_currency
+     * @param bool $currentShopOnly If true returns only currencies associated to current shop
+     *
      * @return array Currency data from database
+     *
+     * @throws PrestaShopDatabaseException
      */
-    public static function findAll($active = true, $groupBy = false, $filterShop = true)
+    public static function findAll($active = true, $groupBy = false, $currentShopOnly = true)
     {
         $currencies = Db::getInstance()->executeS('
             SELECT *
             FROM `' . _DB_PREFIX_ . 'currency` c
-            ' . ($filterShop ? Shop::addSqlAssociation('currency', 'c') : '') . '
+            ' . ($currentShopOnly ? Shop::addSqlAssociation('currency', 'c') : '') . '
                 WHERE `deleted` = 0' .
                 ($active ? ' AND c.`active` = 1' : '') .
                 ($groupBy ? ' GROUP BY c.`id_currency`' : '') .
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -189,9 +189,9 @@ class CurrencyCore extends ObjectModel
             }
 
             if (is_array($this->name)) {
-                $this->name = ucfirst($this->name[$idLang]);
+                $this->name = Tools::ucfirst($this->name[$idLang]);
             } else {
-                $this->name = ucfirst($this->name);
+                $this->name = Tools::ucfirst($this->name);
             }
 
             $this->iso_code_num = $this->numeric_iso_code;
@@ -356,6 +356,27 @@ class CurrencyCore extends ObjectModel
             ->where('c.`active` = 1');
 
         return (bool) Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue($sql);
+    }
+
+    /**
+     * Returns the name of the currency (using the translated name base on the id_lang
+     * provided on creation). This method is useful when $this->name contains an array
+     * but you still need to get its name as a string.
+     *
+     * @return string
+     */
+    public function getName()
+    {
+        if (is_string($this->name)) {
+            return $this->name;
+        }
+
+        $id_lang = $this->id_lang;
+        if ($id_lang === null) {
+            $id_lang = Configuration::get('PS_LANG_DEFAULT');
+        }
+
+        return Tools::ucfirst($this->name[$id_lang]);
     }
 
     /**
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -401,12 +401,12 @@ class CurrencyCore extends ObjectModel
      *
      * @return array Currency data from database
      */
-    public static function findAll($active = true, $groupBy = false)
+    public static function findAll($active = true, $groupBy = false, $filterShop = true)
     {
         $currencies = Db::getInstance()->executeS('
             SELECT *
             FROM `' . _DB_PREFIX_ . 'currency` c
-            ' . Shop::addSqlAssociation('currency', 'c') . '
+            ' . ($filterShop ? Shop::addSqlAssociation('currency', 'c') : '') . '
                 WHERE `deleted` = 0' .
                 ($active ? ' AND c.`active` = 1' : '') .
                 ($groupBy ? ' GROUP BY c.`id_currency`' : '') .
--- a/controllers/admin/AdminOrdersController.php
+++ b/controllers/admin/AdminOrdersController.php
@@ -1705,11 +1705,6 @@ class AdminOrdersControllerCore extends AdminController
         if (!Validate::isLoadedObject($order)) {
             $this->errors[] = $this->trans('The order cannot be found within your database.', array(), 'Admin.Orderscustomers.Notification');
         }
-        if (!in_array($order->id_shop, Shop::getContextListShopID())) {
-            $this->errors[] = $this->trans('The order is not accessible in the current shop context.', array(), 'Admin.Orderscustomers.Notification');
-
-            return '';
-        }
 
         $customer = new Customer($order->id_customer);
         $carrier = new Carrier($order->id_carrier);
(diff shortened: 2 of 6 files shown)


matthieu-rolland (dismissed):
LGTM

classes/Currency.php
```diff
@@ -189,9 +189,9 @@ public function __construct($id = null, $idLang = null, $idShop = null)
             }
 
             if (is_array($this->name)) {
-                $this->name = ucfirst($this->name[$idLang]);
+                $this->name = Tools::ucfirst($this->name[$idLang]);
```
PierreRambaud, line 192: :+1:

classes/Currency.php
```diff
@@ -358,6 +358,27 @@ public function isInstalled($currencyId = null, $shopId = null)
...
+        if (is_string($this->name)) {
+            return $this->name;
+        }
+
+        $id_lang = $this->id_lang;
+        if ($id_lang === null) {
+            $id_lang = Configuration::get('PS_LANG_DEFAULT');
+        }
+
+        return Tools::ucfirst($this->name[$id_lang]);
```
PierreRambaud, line 379: any reason to use ucfirst here and not above?

controllers/admin/AdminOrdersController.php
```diff
@@ -1705,6 +1705,11 @@ public function renderView()
         if (!Validate::isLoadedObject($order)) {
             $this->errors[] = $this->trans('The order cannot be found within your database.', array(), 'Admin.Orderscustomers.Notification');
```
PierreRambaud: no return here?

classes/Currency.php
matthieu-rolland: It seems like **$this->name** has already been set in the constructor using Tools:Ucfirst().  I'm not sure if applying the transformation again in a getter context is overkill or not though...  It seems fine to me as it is.

mickaelandrieu:
> LGTM

Hmmm nope, as I said privately in multistore Store A and B can share orders: this is a feature we don't want to break.

Did someone checked what was the behavior in 1.7.5 in the same conditions?

controllers/admin/AdminOrdersController.php
jolelievre: it would seem logical, but I don't want to change the current (expected?) behavior

classes/Currency.php
jolelievre: > > LGTM
> 
> Hmmm nope, as I said privately in multistore Store A and B can share orders: this is a feature we don't want to break.
> 
> Did someone checked what was the behavior in 1.7.5 in the same conditions?

The behavior in 175 is that you can see an order from another shop if you have the url, although when you list them you don't see it in the list which doesn't seem logical If you switch to a specific shop you should only see this shop's orders
Even though it worked in previous versions it doesn't mean it's the right behavior, this is the kind of bug that became a feature 😅

But the weird thing is that when you export invoices they are shared between the shops... So the two behaviors are contradictory

So we decided not to change the behavior for now, and CLDR is now able to display any installed currency regardless of the shop context But the orders behavior will be discussed later with the multi shop refacto

jolelievre:
> > LGTM
> 
> Hmmm nope, as I said privately in multistore Store A and B can share orders: this is a feature we don't want to break.
> 
> Did someone checked what was the behavior in 1.7.5 in the same conditions?

@mickaelandrieu  The behavior in 175 is that you can see an order from another shop if you have the url, although when you list them you don't see it in the list which doesn't seem logical If you switch to a specific shop you should only see this shop's orders
Even though it worked in previous versions it doesn't mean it's the right behavior, this is the kind of bug that became a feature 😅

But the weird thing is that when you export invoices they are shared between the shops... So the two behaviors are contradictory

So we decided not to change the behavior for now, and CLDR is now able to display any installed currency regardless of the shop context But the orders behavior will be discussed later with the multi shop refacto

controllers/admin/AdminOrdersController.php
PierreRambaud: seems legit. But not sure was the expected behavior :roll_eyes:

classes/Currency.php
jolelievre: if `$this->name` is a string then it means it already has been formatted in the constructor, so no need for an additional ucfirst
However in `EditCurrencyHandler` the `$currency->name` field is filled with an array, which is logical to allow updating the translated fields, but it caused an error when `DefaultCurrencyInMultiShopException` was thrown because the expected string field was then an array, hence this new getter

controllers/admin/AdminOrdersController.php
jolelievre: Probably not.. but I don't think there is a reported issue for now

jolelievre: And anyway this page is going to be migrated

classes/Currency.php
eternoendless: This method is useful but not a good idea because of the hidden context (aka the current language being inferred magically).

The proper getter would be `getName($idLang)` where `$idLang` cannot be null.

classes/Currency.php
```diff
@@ -380,12 +401,12 @@ public static function getCurrencies($object = false, $active = true, $groupBy =
      *
      * @return array Currency data from database
      */
-    public static function findAll($active = true, $groupBy = false)
+    public static function findAll($active = true, $groupBy = false, $filterShop = true)
```
eternoendless: I think it one of these would be easier to understand:

- `$allShops = false`
- `$currentShopOnly = true`

eternoendless: Also, missing Phpdoc

classes/Currency.php
jolelievre: Yes except that if `$this->name` doesn't contain the array the method may return a string that doesn't match the asked language, which can be confusing..
Unless we store the localized names in another field `$this->names` in the constructor, similar to what I had to do in another PR:
https://github.com/PrestaShop/PrestaShop/blob/06e1ad1711cfaacd91a94ea6cddf1939b8f762d1/classes/Currency.php#L223

This would allow to get the name in any language regardless of the context

LouiseBonnard:
Hey @jolelievre, no more wording to check here?

jolelievre:
> Hey @jolelievre, no more wording to check here?

Nope no wording in this PR

src/Core/Currency/CurrencyDataProviderInterface.php
```diff
@@ -47,12 +47,19 @@ interface CurrencyDataProviderInterface
...
+     * Return raw currencies data from the database related to the current shop.
      *
      * @return array Installed currencies
      */
     public function findAll();
 
+    /**
+     * Return raw currencies data from the database regardless of the current shop.
+     *
+     * @return array
```
eternoendless:
```suggestion
     * @return array[]
```

src/Core/Currency/CurrencyDataProviderInterface.php
```diff
@@ -47,12 +47,19 @@ interface CurrencyDataProviderInterface
     public function getCurrencies($object = false, $active = true, $group_by = false);
 
     /**
-     * Return raw currencies data from the database reated to the current shop.
+     * Return raw currencies data from the database related to the current shop.
      *
      * @return array Installed currencies
```
eternoendless:
```suggestion
     * @return array[] Installed currencies
```

src/Core/Currency/CurrencyDataProviderInterface.php
```diff
@@ -47,12 +47,19 @@ interface CurrencyDataProviderInterface
...
      */
     public function findAll();
 
+    /**
+     * Return raw currencies data from the database regardless of the current shop.
+     *
+     * @return array
+     */
+    public function findAllAvailable();
```
eternoendless: This could be misunderstood as "all languages available to install", which we use in other places. I think it would be better just to add an optional parameter to `findAll()` instead.

eternoendless (changes requested):

jolelievre:
@eternoendless thanks, all your feedbacks have been addressed

classes/Currency.php
eternoendless: Argh, you're right. This sucks.

eternoendless (approved):

eternoendless:
Thank you @jolelievre

PR #15180, opened 2019-08-21
Migration /international/locations/countries Add / Edit

RaimondasSapola:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migrated /international/locations/countries edit and create forms page 
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10605
| How to test?  | International > locations > countries edit/create page should be unchanged

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15180)


RaimondasSapola:
Waiting for design guidelines for address format field

matks:
> Waiting for design guidelines for address format field

@TristanLDD 👀

RaimondasSapola:
Screen Shot 2019-08-21 at 3 55 10 PM
Updated styling like so for now

matks:
> Screen Shot 2019-08-21 at 3 55 10 PM
> Updated styling like so for now

It does the job 👍

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
```diff
@@ -0,0 +1,182 @@
...
+        <div class="col-sm">
+          {{ form_widget(countryForm.country) }}
+          {{ form_errors(countryForm.country) }}
+        </div>
+      </div>
+
+      <div class="form-group row">
+        {{ ps.label_with_help(
+          'ISO code'|trans({}, 'Admin.International.Feature'),
+          'Two -- or three -- letter ISO code (e.g. "us" for United States).'|trans({}, 'Admin.International.Help'),
```
LouiseBonnard: Wouldn't that be more helpful to indicate which ISO code to write?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
```diff
@@ -0,0 +1,182 @@
...
+
+      {{ ps.form_group_row(countryForm.zip_code_format, {}, {
+        'label': 'Zip/postal code format'|trans({}, 'Admin.International.Feature'),
+        'help' : 'Indicate the format of the postal code: use L for a letter, N for a number, and C for the country\'s ISO 3166-1 alpha-2 code. For example, NNNNN for the United States, France, Poland and many other; LNNNNLLL for Argentina, etc. If you do not want PrestaShop to verify the postal code for this country, leave it blank.'|trans({}, 'Admin.International.Help')
+      }) }}
+
+      <div class="form-group row">
+        <label class="form-control-label">
+          <span class="text-danger">*</span>
+          {{ 'Address format'|trans({}, 'Admin.Global')}}
```
LouiseBonnard: Please localize it in _Admin.International.Feature_, thank you.

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
```diff
@@ -0,0 +1,182 @@
...
+            'defaultLayout': encodingDefaultLayout,
+            'encodingAddressLayout': encodingAddressLayout
+          }
+        ) }}
+      </div>
+
+      <div class="form-group row">
+        {{ ps.label_with_help(
+          'Active'|trans({}, 'Admin.Global'),
+          'Display this country to your customers (the selected country will always be displayed in the Back Office).'|trans({}, 'Admin.International.Help'),
```
LouiseBonnard:
```suggestion
          'Display this country to your customers (the selected country will always be displayed in the back office).'|trans({}, 'Admin.International.Help'),
```

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/format_actions.html.twig
```diff
@@ -0,0 +1,47 @@
+{% set confirmation = 'Are you sure you want to restore the default address format for this country?'|trans({},'Admin.International.Feature') %}
```
LouiseBonnard:
```suggestion
{% set confirmation = 'Are you sure you want to restore the default address format for this country?'|trans({},'Admin.International.Notification') %}
```

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/format_actions.html.twig
```diff
@@ -0,0 +1,47 @@
...
+        <a id="useDefaultLayoutSystem"
+           href="javascript:void(0)"
+           data-confirmation="{{ confirmation }}"
+           data-default-layout="{{ defaultLayout }}"
+           class="btn btn-outline-secondary reset-address-layout">
+          {{ 'Use the default format'|trans({}, 'Admin.International.Feature') }}
+        </a>
+  </span>
+  <span title="" data-toggle="tooltip" class="label-tooltip"
+        data-original-title="{{ 'This will restore your current address format.'|trans({}, 'Admin.International.Help') }}"
```
LouiseBonnard, line 33: Well, I'm not sure about the wording here, it seems weird for me to restore something that is already in place.

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
RaimondasSapola: I cant see any translation key with lower case 'back office'. is it ok to change it?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
RaimondasSapola: Well doesn't 'e.g. "us"' fulfill this purpose?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/format_actions.html.twig
RaimondasSapola: After manually modifying address format value it will be saved and this action will fill text field with saved data after other modifications with given actions were done if any. I agree that it is a bit misleading but it kind of says what it does correctly. I cant think of a valid replacement. Do you have any suggestions?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
LouiseBonnard: Yes, we will lose the translation (at least, at the beginning) but we will gain consistency and harmonization. ;-)

RaimondasSapola: 👍

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/format_actions.html.twig
LouiseBonnard: But what I don't get is the difference with the first button _This will restore your last registered address format._, it looks the same to me.
🤔

RaimondasSapola: First button takes value from backend and it is the last value submitted. The button in question restores value locally which is saved in browser while editing Country

LouiseBonnard: Alright, I think I get it. @marionf has suggested adding a tooltip here to let the users know where to set mandatory fields for customers, here is what I suggest: _Want to set required fields in the customer's registration form? Go to Customers > Addresses and select the information you need._ localized in _Admin.International.Help_.

marionf: @LouiseBonnard wdyt about: 
_Want to set required fields in the customer's registration form? Go to Customers > Addresses and **set your required fields** ?_

With a link to the page

LouiseBonnard: So, we agreed on: _Want to set required fields in the customer's registration form? Go to [1]Customers > Addresses[/1] to configure it._, still in _Admin.International.Help_ and with a redirecting link to bring users to the proper page directly. Would that be possible, @RaimondasSapola?

RaimondasSapola: There is a problem with tooltip solution. It is only shown on button hover and it disappears if hovered over itself and not he button. It is a global tooltip behavior and i am not sure it is i good idea to change it only for this one use case. I suggest adding help text under field actions. Like in the field above:
Screen Shot 2019-08-22 at 11 53 05 AM

LouiseBonnard: It suits me fine!

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
LouiseBonnard: Oops, sorry Raimondas, I forgot about this comment... Handling ISO codes can be pretty confusing for users, there are many standards depending on the nature of the object designated by the ISO code. So my opinion is that being precise is essential in order to avoid mistakes at this point. Why not write directly _3166-1 ISO code_ here?

Also, I remember I have been told that PrestaShop cannot handle three-letter ISO codes. Can you confirm, @eternoendless? If yes, I guess we should not mention it in the tooltip.

RaimondasSapola: Well i have seen multiple instances of three letter ISO codes in BO. Should '3166-1 ISO code' be in 'Admin.International.Help' domain?

LouiseBonnard: It depends on whether we decide to include it in the field name (which is _Feature_) or in the tooltip (which is _Help_) that comes with it. About the three-letter ISO codes support or not, let's wait for @eternoendless's feedback before, I would like to be sure about it.
🙂

RaimondasSapola:
Hey @matks. There is a small bug in Address form regarding this PR. In country form zip code format must be added. The code can consist N - number L - letter and C - iso code. Country form zip code field has validation for maximum 12 characters. Address form zip code field has the same constraint but if 12 symbols including at least one "C" are submitted in country form there is no way address form can be valid at this point with that country. Now address form requires zip code field to have  given format which is 13 or more symbols and at the same time doesn't allow more than 12 symbols. I understand is quite an edge case but is there something that should be done about this?

admin-dev/themes/new-theme/js/pages/country/CountryForm.js
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Class responsible for javascript actions in country add/edit page.
+ */
+export default class CountryForm {
+  constructor() {
+    this.customAddressFields = countryFormMap.customAddressFields;
+    this.patternSelect = countryFormMap.patternSelect;
+    this.formatInput = countryFormMap.formatInput;
+    this.eraseCurrentLayout = countryFormMap.eraseCurrentLayout;
```
sarjon: I don't think these variables are necessary, you just duplicate countryFormMap.

admin-dev/themes/new-theme/js/pages/country/CountryForm.js
```diff
@@ -0,0 +1,128 @@
...
+    $(this.formatInput).on('keyup', (event) => this.saveLastModified(event));
+    $(this.eraseCurrentLayout).on('click', (event) => this.resetLayout(event));
+  }
+
+  /**
+   * Handles pattern link click for address format text input
+   *
+   * @param event
+   */
+  handlePatternClick(event) {
```
sarjon, line 57: I think there is convention to start methods with `_` if they are not suppose to be called from outside aka "private".

admin-dev/themes/new-theme/js/pages/country/CountryForm.js
```diff
@@ -0,0 +1,128 @@
...
+    $(event.target).tab('show');
+  }
+
+  /**
+   * Adds pattern to a cursor position in format text box
+   *
+   * @param pattern
+   */
+  addFieldsToCursorPosition(pattern) {
+    const el = $(this.formatInput).get(0);
```
sarjon: Is that a jQuery object? If so, would be nice to prefix with `$`, it's also a convention to for jQuery objects.

admin-dev/themes/new-theme/js/pages/country/CountryForm.js
```diff
@@ -0,0 +1,128 @@
...
+    $(this.formatInput).val(content.substr(0, pos) + pattern + ' ' + content.substr(pos));
+  }
+
+  /**
+   * Resets address format text box value to default layout or last modified
+   *
+   * @param event
+   */
+  resetLayout(event) {
+    const confirmation = $(event.target).data('confirmation');
```
sarjon: I'd call custom attribute `data-confirmation-message`, to be more clear.

admin-dev/themes/new-theme/js/pages/country/country-form-map.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+/**
+ * Defines all selectors that are used in country add/edit form.
+ */
+export default {
+  customAddressFields: '#custom-address-fields',
+  patternSelect: '.add-pattern',
+  formatInput: '#country_address_format',
+  eraseCurrentLayout: '.reset-address-layout',
```
sarjon: I think it should be named `resetCurrentLayout`.

Btw, name is not clear imo, is it a button or what? If it's a button, I'd call it `resetCurrentLayoutBtn` or something similar. :)

Same for other properties.

admin-dev/themes/new-theme/scss/pages/_country_form.scss
```diff
@@ -0,0 +1,42 @@
...
+  border: 1px solid #d3d8db;
+  padding-bottom: 1px;
+}
+
+.address-format-tabs-content {
+  border: solid #d3d8db;
+  border-width: 0 1px 1px 1px;
+}
+
+.address-format-tabs a.nav-link {
```
sarjon: You could nest these, like so:

```css
.address-format-tabs {
    a.nav-link {
        // a.nav-link css goes here
    }
}
```

Wdyt?

admin-dev/themes/new-theme/scss/pages/_country_form.scss
```diff
@@ -0,0 +1,42 @@
...
+  font-size: 12px;
+  border: 1px solid;
+  margin: 1px 0;
+  padding: 1px 5px;
+  border-radius: 5px;
+}
+
+.address-format-tabs-content a.add-pattern:hover {
+  border-radius: 4px;
+  color: #1383d4
```
sarjon: Missing `;` at the end. :wink:

src/Adapter/Address/QueryHandler/IsAddressFormatValidHandler.php
```diff
@@ -0,0 +1,62 @@
...
+use AddressFormat;
+use PrestaShop\PrestaShop\Core\Domain\Address\Exception\AddressException;
+use PrestaShop\PrestaShop\Core\Domain\Address\Query\IsValidAddressFormat;
+use PrestaShop\PrestaShop\Core\Domain\Address\QueryHandler\IsAddressFormatValidHandlerInterface;
+use PrestaShopException;
+
+/**
+ * Handles validation of address format field value
+ */
+class IsAddressFormatValidHandler implements IsAddressFormatValidHandlerInterface
```
sarjon: Should be `final`. Maybe rename to `GetAddressFormatIsValid`? Wdyt?

But I don't know how I feel about this handler, but I don't have anything better to suggest. :/ @matks wdyt?

src/Adapter/Address/QueryHandler/IsAddressFormatValidHandler.php
```diff
@@ -0,0 +1,62 @@
...
+        try {
+            $addressFormat = new AddressFormat();
+        } catch (PrestaShopException $e) {
+            throw new AddressException('Failed to create AddressFormat object for validation');
+        }
+
+        $addressFormat->format = $command->getFormat();
+        $isValid = $addressFormat->checkFormatFields();
+
+        unset($addressFormat);
```
sarjon: Why do you need to unset it? :thinking:

src/Adapter/Country/AbstractCountryHandler.php
```diff
@@ -0,0 +1,68 @@
...
+
+use Country;
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Country\ValueObject\CountryId;
+use PrestaShopException;
+
+/**
+ * Abstract country handler provides common methods for country handlers
+ */
+class AbstractCountryHandler
```
sarjon: Make it actually `abstract`. :stuck_out_tongue:

src/Adapter/Country/CommandHandler/AddCountryHandler.php
```diff
@@ -0,0 +1,114 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CannotAddCountryException;
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryException;
+use PrestaShop\PrestaShop\Core\Domain\Country\ValueObject\CountryId;
+use PrestaShopException;
+
+/**
+ * Handles creation of country
+ */
+class AddCountryHandler implements AddCountryHandlerInterface
```
sarjon: `final`.

src/Adapter/Country/CommandHandler/AddCountryHandler.php
```diff
@@ -0,0 +1,114 @@
...
+     * @throws CannotAddCountryException
+     * @throws CountryConstraintException
+     * @throws CountryException
+     */
+    public function handle(AddCountryCommand $command): CountryId
+    {
+        try {
+            $country = new Country();
+
+            $country->name = $command->getLocalisedNames();
```
sarjon: How about `Localized` instead of `Localised`. :smile:

src/Adapter/Country/CommandHandler/AddCountryHandler.php
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            if (null !== $command->getZone()) {
+                $country->id_zone = $command->getZone();
+            }
+
+            if (!$country->validateFields(false) || !$country->validateFieldsLang(false)) {
+                throw new CountryConstraintException(
+                    'Country contains invalid field values',
+                    CountryConstraintException::INVALID_FIELDS
```
sarjon: This code is very generic, it does not tell more than exception itself imo.

src/Adapter/Country/CommandHandler/EditCountryHandler.php
```diff
@@ -0,0 +1,121 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryException;
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\UpdateCountryException;
+use PrestaShopException;
+
+/**
+ * Handles editing of country and country address format
+ */
+class EditCountryHandler extends AbstractCountryHandler implements EditCountryHandlerInterface
```
sarjon: Should be `final`.

src/Adapter/Country/CommandHandler/EditCountryHandler.php
```diff
@@ -0,0 +1,121 @@
...
+                throw new CountryConstraintException(
+                    'Country contains invalid field values',
+                    CountryConstraintException::INVALID_FIELDS
+                );
+            }
+
+            $addressFormat = new AddressFormat($country->id);
+            $addressFormat->format = $command->getAddressFormat();
+
+            $isInvalidAddressFormat = !$addressFormat->checkFormatFields() ||
```
sarjon: I see you used same logic here as in `Add` action, maybe it would make sense to extract it into `abstract` handler? :thinking:

src/Adapter/Country/QueryHandler/GetAddressLayoutFieldsHandler.php
```diff
@@ -0,0 +1,120 @@
...
+use AddressFormat;
+use PrestaShop\PrestaShop\Core\Domain\Country\Query\GetAddressLayoutFields;
+use PrestaShop\PrestaShop\Core\Domain\Country\QueryHandler\GetAddressLayoutFieldsHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Country\QueryResult\AddressLayoutFields;
+use Tools;
+
+/**
+ * Provides legacy address layout modification data
+ */
+class GetAddressLayoutFieldsHandler implements GetAddressLayoutFieldsHandlerInterface
```
sarjon: `final`.

I don't understand this query. :confused: It could be that name is a little misleading.

src/Adapter/Country/QueryHandler/GetAddressLayoutFieldsHandler.php
```diff
@@ -0,0 +1,120 @@
...
+            $this->getDefaultLayout()
+        );
+    }
+
+    /**
+     * Gets legacy valid fields in array
+     *
+     * @return array
+     */
+    protected function getValidFields(): array
```
sarjon: Why not `private`?

src/Adapter/Country/QueryHandler/GetAddressLayoutFieldsHandler.php
```diff
@@ -0,0 +1,120 @@
...
+     * Gets current country legacy address format layout
+     *
+     * @param int $countryId
+     *
+     * @return string
+     */
+    protected function getAddressLayout(int $countryId): string
+    {
+        $addressLayout = AddressFormat::getAddressCountryFormat($countryId);
+        if ($value = Tools::getValue('address_layout')) {
```
sarjon: Would it be possible to avoid using `Tools::getValue`?

src/Adapter/Country/QueryHandler/GetCountryForEditingHandler.php
```diff
@@ -0,0 +1,72 @@
...
+use PrestaShop\PrestaShop\Adapter\Country\AbstractCountryHandler;
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Country\Query\GetCountryForEditing;
+use PrestaShop\PrestaShop\Core\Domain\Country\QueryHandler\GetCountryForEditingHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Country\QueryResult\EditableCountry;
+
+/**
+ * Handles editable country query
+ */
+class GetCountryForEditingHandler extends AbstractCountryHandler implements GetCountryForEditingHandlerInterface
```
sarjon: `final`. :stuck_out_tongue:

src/Adapter/Zone/ZoneDataProvider.php
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * This class will provide data from DB / ORM about Zone
+ */
+final class ZoneDataProvider implements ZoneDataProviderInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getZones($active = false, $activeFirst = false): array
```
sarjon: For me, this is a code design smell. You should avoid having "flag argument" as much as possible, don't be influenced by legacy interfaces. :wink: 

See https://martinfowler.com/bliki/FlagArgument.html

src/Adapter/Zone/ZoneDataProvider.php
```diff
@@ -0,0 +1,44 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Adapter\Zone;
+
+use Zone;
+use PrestaShop\PrestaShop\Core\Domain\Zone\DataProvider\ZoneDataProviderInterface;
```
sarjon: Not sure if that's the right namespace, @matks wdyt?

src/Core/ConstraintValidator/AddressFormatValidator.php
```diff
@@ -0,0 +1,65 @@
...
+
+use PrestaShop\PrestaShop\Core\CommandBus\CommandBusInterface;
+use PrestaShop\PrestaShop\Core\Domain\Address\Query\IsValidAddressFormat;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * Validates address format field value
+ */
+class AddressFormatValidator extends ConstraintValidator
```
sarjon: I guess it _could_ be final.

src/Core/Domain/Address/Exception/CannotAddAddressException.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Thrown on failure to create address
+ */
+class CannotAddAddressException extends AddressException
+{
+    /**
+     * When address format creation failed
+     */
+    const ADDRESS_FORMAT = 1;
```
sarjon: Maybe a separate exception for it?

src/Core/Domain/Address/Exception/CannotUpdateAddressException.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Thrown on failure to update address
+ */
+class CannotUpdateAddressException extends AddressException
+{
+    /**
+     * When failed updating address format
+     */
+    const ADDRESS_FORMAT = 1;
```
sarjon: Same for this one?

src/Core/Domain/Address/Query/IsValidAddressFormat.php
```diff
@@ -0,0 +1,54 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Address\Query;
+
+/**
+ * Determines if address format field value is valid
+ */
+class IsValidAddressFormat
```
sarjon: Maybe `GetAddressFormatIsValid`? @matks wdyt about this query?

src/Core/Domain/Country/Command/AddCountryCommand.php
```diff
@@ -0,0 +1,328 @@
...
+    {
+        return $this->zipCodeFormat;
+    }
+
+    /**
+     * @param string $zipCodeFormat
+     *
+     * @return AddCountryCommand
+     */
+    public function setZipCodeFormat(string $zipCodeFormat): AddCountryCommand
```
sarjon, line 161:
```suggestion
    public function setZipCodeFormat(string $zipCodeFormat): self
```

src/Core/Domain/Country/Command/AddCountryCommand.php
```diff
@@ -0,0 +1,328 @@
...
+    {
+        return $this->defaultCurrency;
+    }
+
+    /**
+     * @param int|null $defaultCurrency
+     *
+     * @return AddCountryCommand
+     */
+    public function setDefaultCurrency(?int $defaultCurrency): AddCountryCommand
```
sarjon:
```suggestion
    public function setDefaultCurrency(?int $defaultCurrency): self
```

src/Core/Domain/Country/CommandHandler/EditCountryHandlerInterface.php
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Defines contract for edit country handler
+ */
+interface EditCountryHandlerInterface
+{
+    /**
+     * @param EditCountryCommand $command
+     */
+    public function handle(EditCountryCommand $command);
```
sarjon: You could add `void` as return type to make it explicit. :wink:

src/Core/Domain/Country/Config/CountryConstraintConfiguration.php
```diff
@@ -0,0 +1,55 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Country\Config;
+
+/**
+ * Stores country form constraints configuration values
+ */
+final class CountryConstraintConfiguration
```
sarjon, line 32: My suggestion would be to move this class one level up and rename to `CountryConstraint`, wdyt? :)

src/Core/Domain/Country/Config/CountryConstraintConfiguration.php
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Maximum length for call prefix
+     */
+    const MAX_CALL_PREFIX_LENGTH = 3;
+
+    /**
+     * Zip code field value regexp validation pattern
+     */
+    const ZIP_CODE_PATTERN = '/^[NLCnlc 0-9-]+$/';
```
sarjon: Maybe it would make sense to have ZipCode VO, @matks wdyt?

src/Core/Domain/Country/Exception/UpdateCountryException.php
```diff
@@ -0,0 +1,34 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Country\Exception;
+
+/**
+ * Is thrown on failure to update country
+ */
+class UpdateCountryException extends CountryException
```
sarjon: Shouldn't it be `CannotUpdateCountryException`?

src/Core/Domain/Country/Query/GetAddressLayoutFields.php
```diff
@@ -0,0 +1,59 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Country\ValueObject\CountryId;
+
+/**
+ * Provides data for address layout modification fields
+ */
+class GetAddressLayoutFields
+{
+    /**
+     * @var CountryId|null
```
sarjon: Can it be `null`?

src/Core/Domain/Country/Query/GetAddressLayoutFields.php
```diff
@@ -0,0 +1,59 @@
...
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Country\Query;
+
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Country\ValueObject\CountryId;
+
+/**
+ * Provides data for address layout modification fields
```
sarjon: I don't think it explains the query well. :confused:

src/Core/Domain/Country/QueryResult/AddressLayoutFields.php
```diff
@@ -0,0 +1,84 @@
...
+     */
+    private $addressLayout;
+
+    /**
+     * @var string
+     */
+    private $defaultLayout;
+
+    /**
+     * @param array $validFields
```
sarjon: Could you document what's inside array? Maybe `string[]` if that's the case.

src/Core/Domain/Country/QueryResult/EditableCountry.php
```diff
@@ -0,0 +1,265 @@
...
+     * @param int $defaultCurrency
+     * @param int $zone
+     * @param bool $needZipCode
+     * @param string $zipCodeFormat
+     * @param string $addressFormat
+     * @param bool $enabled
+     * @param bool $containsStates
+     * @param bool $needIdNumber
+     * @param bool $displayTaxLabel
+     * @param array $shopAssociation
```
sarjon: Specify what's inside, I guess `int[]`?

src/Core/Domain/Zone/DataProvider/ZoneDataProviderInterface.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    /**
+     * Return available zones.
+     *
+     * @param bool $active only active zones
+     * @param bool $activeFirst order by active DESC
+     *
+     * @return array
+     */
+    public function getZones($active = false, $activeFirst = false): array;
```
sarjon: As mentioned before, I'm not in favor of adding this interface. :)

src/Core/Form/IdentifiableObject/DataHandler/CountryFormDataHandler.php
```diff
@@ -0,0 +1,144 @@
...
+            $data['zone'],
+            $data['need_zip_code'],
+            $data['address_format'],
+            $data['is_enabled'],
+            $data['contains_states'],
+            $data['need_identification_number'],
+            $data['display_tax_label']
+        );
+
+        if (null !== $data['default_currency']) {
```
sarjon: I don't think currency is optional.

src/Core/Form/IdentifiableObject/DataProvider/CountryFormDataProvider.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * @var bool
+     */
+    private $isMultistoreFeatureActive;
+
+    /**
+     * @param CommandBusInterface $queryBus
+     * @param bool $isMultistoreFeatureActive
+     */
+    public function __construct(CommandBusInterface $queryBus, bool $isMultistoreFeatureActive)
```
sarjon: I'd like `$isMultistoreFeatureActive` to be service instead of result.

src/PrestaShopBundle/Controller/Admin/Improve/International/CountryController.php
```diff
@@ -0,0 +1,227 @@
...
+    {
+        $indexLink = $this->getAdminLink('AdminCountries', []);
+
+        return $this->redirect($indexLink);
+    }
+
+    /**
+     * @return array
+     */
+    private function getErrorMessages(): array
```
sarjon: Move `private` methods to the bottom.

src/PrestaShopBundle/Controller/Admin/Improve/International/CountryController.php
```diff
@@ -0,0 +1,227 @@
...
+            return $this->redirectToRoute('admin_countries_index');
+        }
+
+        return $this->render('@PrestaShop/Admin/Improve/International/Locations/Country/add.html.twig', [
+            'enableSidebar' => true,
+            'countryForm' => $countryForm->createView(),
+            'help_link' => $this->generateSidebarLink($request->attributes->get('_legacy_controller')),
+            'addressLayout' => $addressLayout->getAddressLayout(),
+            'encodingAddressLayout' => urlencode($addressLayout->getAddressLayout()),
+            'encodingDefaultLayout' => urlencode($addressLayout->getDefaultLayout()),
```
sarjon: Not sure if you need to prefix it with `encoding`. :thinking:

src/PrestaShopBundle/Controller/Admin/Improve/International/CountryController.php
```diff
@@ -0,0 +1,227 @@
...
+            'displayValidFields' => $addressLayout->getValidFields(),
```
sarjon: It feels like it should be `bool`, but it's not?

src/PrestaShopBundle/Form/Admin/Improve/International/Locations/Country/CountryType.php
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * @param TranslatorInterface $translator
+     * @param array $currencyChoices
+     * @param bool $isMultistoreFeatureActive
+     */
+    public function __construct(
+        TranslatorInterface $translator,
+        array $currencyChoices,
+        bool $isMultistoreFeatureActive
```
sarjon: Same for this, maybe it's better to inject services?

src/PrestaShopBundle/Form/Admin/Improve/International/Locations/Country/CountryType.php
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * Gets boolean value for edit form fields with default value "true" instead hardcoded data value
+     *
+     * @param array $data
+     * @param string $index
+     *
+     * @return bool
+     */
+    private function getDataValue(array $data, string $index): bool
```
sarjon: Maybe `getBoolValue` insead?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Macros/address_format_macros.html.twig
```diff
@@ -0,0 +1,37 @@
+{% macro tabContent(data = []) %}
```
sarjon: Hmm, maybe simple template with `include` would work as well?

sarjon (changes requested):

src/Adapter/Zone/ZoneDataProvider.php
RaimondasSapola: Using these flags enabled ability to configure choices straight from the form type. For this particular case ActiveZoneDataProvider could be used to get only active zones. I guess separate interfaces could be used for every other case but new choice provider will be needed for them all so if ZoneChoiceType to be configured, four separate services will have to be injected or no configuration will be available and separate ChoiceType will be needed for every case. So is it ok to lose configurability for a little bit of cleaner code in this case?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
LouiseBonnard: @PrestaShop/prestashop-core-developers, any recommendation?

src/Adapter/Country/AbstractCountryHandler.php
```diff
@@ -0,0 +1,122 @@
...
+
+            $addressFormat->format = $format;
+            $addressFormat->id_country = $addressFormat->id_country ?? $countryId;
+
+            $isInvalidAddressFormat = !$addressFormat->checkFormatFields() ||
+                strlen($addressFormat->format) <= 0 ||
+                !$addressFormat->validateFields(false);
+        } catch (PrestaShopException $e) {
+            throw new AddressConstraintException(
+                sprintf('Address format: "%s" is invalid', $format),
```
matthieu-rolland: AddressFormat's **checkFormatFields** method generates several error messages for different kind of errors, it would be nice to display them instead of having a single generic error message, it has a public method for that:

`$addressFormat->getErrorList()`

Besides being better for the end user, it will ensure that future error cases treated by AddressFormat will have their error messages displayed in the frontend without having to modify anything else.

matthieu-rolland (changes requested):

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
LouiseBonnard: 🆙

matks: @eternoendless 😛

eternoendless: PrestaShop only supports two-letter country iso codes AFAIK, and it also mixes up language codes and country codes (yay!).

eternoendless: I think we're talking about "ISO 3166-1 alpha-2" codes here.

LouiseBonnard: @matks, @eternoendless, thanks a lot for your feedback! In order to avoid confusion here, I recommend writing: 

_ISO 3166-1 code_ in _Admin.International.Feature_
_Two-letter ISO 3166-1 alpha-2 code (e.g. "US" for the United States)_ in _Admin.International.Help_

Also, one nice idea to avoid misunderstandings would be to authorize only two characters for this field.

admin-dev/themes/new-theme/js/pages/country/CountryForm.js
tdavidsonas88: in eslint _ is no longer supported as well as in PSR-2 not supported

tdavidsonas88:
thinking about simplifying the form with prestashop_ui_kit.html and form_widget. However * Address format field seems not so simple and holds me back from starting it.
image

tdavidsonas88:
To finish simplify the form https://github.com/PrestaShop/PrestaShop/pull/16938 is needed. It has 'help' and 'required' fields rendering standards implemented already.

admin-dev/themes/new-theme/js/pages/country/CountryForm.js
```diff
@@ -0,0 +1,124 @@
...
+  addFieldsToCursorPosition(pattern) {
+    const $el = $(countryFormMap.formatTextAreaField).get(0);
+    let pos = 0;
+
+    if ('selectionStart' in $el) {
+      pos = $el.selectionStart;
+    } else if ('selection' in document) {
+      $el.focus();
+      const sel = document.selection.createRange();
+      const selLength = document.selection.createRange().text.length;
```
Progi1984:
```suggestion
      const selLength = sel.text.length;
```

admin-dev/themes/new-theme/js/pages/country/CountryForm.js
```diff
@@ -0,0 +1,124 @@
...
+    $(countryFormMap.formatTextAreaField).val(`${content.substr(0, pos)} ${pattern} ${content.substr(pos)}`);
+  }
+
+  /**
+   * Resets address format text box value to default layout or last modified
+   *
+   * @param event
+   */
+  modifyLayout(event) {
+    const confirmation = $(event.target).data('confirmation-message');
```
Progi1984:
```suggestion
    const confirmation = $(event.target).data('confirmationMessage');
```

admin-dev/themes/new-theme/js/pages/country/CountryForm.js
```diff
@@ -0,0 +1,124 @@
...
+    let defaultLayout = $(event.target).data('default-layout');
```
Progi1984:
```suggestion
    let defaultLayout = $(event.target).data('defaultLayout');
```

admin-dev/themes/new-theme/scss/pages/_country_form.scss
```diff
@@ -0,0 +1,44 @@
...
+    font-size: 14px;
+    padding: 5px 10px;
+  }
+}
+
+.address-format-actions > span > a {
+  margin: 1px 0;
+}
+
+.address-format-tabs-content a.add-pattern {
```
Progi1984: Please use nesting and move `a.add-pattern` to its parent class :up:

admin-dev/themes/new-theme/scss/pages/_country_form.scss
```diff
@@ -0,0 +1,44 @@
...
+  font-size: 12px;
+  border: 1px solid;
+  margin: 1px 0;
+  padding: 1px 5px;
+  border-radius: 5px;
+}
+
+.address-format-tabs-content a.add-pattern:hover {
```
Progi1984: Please use nesting and move a.add-pattern to its parent class :arrow_up:

admin-dev/themes/new-theme/scss/pages/_country_form.scss
```diff
@@ -0,0 +1,44 @@
...
+  border-radius: 4px;
+  color: #1383d4;
+}
+
+.address-format-tabs a.nav-link {
+  &::before {
```
Progi1984: Please use nesting and move a.add-pattern to its parent class :arrow_up:

src/Adapter/Address/QueryHandler/GetAddressFormatIsValidHandler.php
```diff
@@ -0,0 +1,60 @@
...
+        try {
+            $addressFormat = new AddressFormat();
+        } catch (PrestaShopException $e) {
+            throw new AddressException('Failed to create AddressFormat object for validation');
+        }
+
+        $addressFormat->format = $command->getFormat();
+        $isValid = $addressFormat->checkFormatFields();
+
+        return $isValid;
```
Progi1984:
```suggestion
        return $addressFormat->checkFormatFields();
```

src/Adapter/Country/QueryHandler/GetAddressFormatDataHandler.php
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * @param int $countryId
+     *
+     * @return string
+     */
+    private function getAddressFormat(int $countryId): string
+    {
+        $addressLayout = AddressFormat::getAddressCountryFormat($countryId);
+
+        return $addressLayout;
```
Progi1984:
```suggestion
        return AddressFormat::getAddressCountryFormat($countryId);
```

src/Core/Domain/Country/Command/AddCountryCommand.php
```diff
@@ -0,0 +1,333 @@
...
+    {
+        return $this->defaultCurrency;
+    }
+
+    /**
+     * @param int $defaultCurrency
+     *
+     * @return AddCountryCommand
+     */
+    public function setDefaultCurrency(int $defaultCurrency): AddCountryCommand
```
Progi1984:
```suggestion
    public function setDefaultCurrency(int $defaultCurrency): self
```

src/Core/Domain/Country/Command/AddCountryCommand.php
```diff
@@ -0,0 +1,333 @@
...
+    {
+        return $this->zone;
+    }
+
+    /**
+     * @param int $zone
+     *
+     * @return AddCountryCommand
+     */
+    public function setZone(int $zone): AddCountryCommand
```
Progi1984:
```suggestion
    public function setZone(int $zone): self
```

src/Core/Domain/Country/Command/AddCountryCommand.php
```diff
@@ -0,0 +1,333 @@
...
+    {
+        return $this->needZipCode;
+    }
+
+    /**
+     * @param bool $needZipCode
+     *
+     * @return AddCountryCommand
+     */
+    public function setNeedZipCode(bool $needZipCode): AddCountryCommand
```
Progi1984:
```suggestion
    public function setNeedZipCode(bool $needZipCode): self
```

src/Core/Domain/Country/Command/AddCountryCommand.php
```diff
@@ -0,0 +1,333 @@
...
+    {
+        return $this->enabled;
+    }
+
+    /**
+     * @param bool $enabled
+     *
+     * @return AddCountryCommand
+     */
+    public function setEnabled(bool $enabled): AddCountryCommand
```
Progi1984:
```suggestion
    public function setEnabled(bool $enabled): self
```

src/Core/Domain/Country/Command/AddCountryCommand.php
```diff
@@ -0,0 +1,333 @@
...
+    {
+        return $this->containsStates;
+    }
+
+    /**
+     * @param bool $containsStates
+     *
+     * @return AddCountryCommand
+     */
+    public function setContainsStates(bool $containsStates): AddCountryCommand
```
Progi1984:
```suggestion
    public function setContainsStates(bool $containsStates): self
```

src/Core/Domain/Country/Command/AddCountryCommand.php
```diff
@@ -0,0 +1,333 @@
...
+    {
+        return $this->needIdNumber;
+    }
+
+    /**
+     * @param bool $needIdNumber
+     *
+     * @return AddCountryCommand
+     */
+    public function setNeedIdNumber(bool $needIdNumber): AddCountryCommand
```
Progi1984:
```suggestion
    public function setNeedIdNumber(bool $needIdNumber): self
```

src/Core/Domain/Country/Command/AddCountryCommand.php
```diff
@@ -0,0 +1,333 @@
...
+    {
+        return $this->displayTaxLabel;
+    }
+
+    /**
+     * @param bool $displayTaxLabel
+     *
+     * @return AddCountryCommand
+     */
+    public function setDisplayTaxLabel(bool $displayTaxLabel): AddCountryCommand
```
Progi1984:
```suggestion
    public function setDisplayTaxLabel(bool $displayTaxLabel): self
```

src/Core/Domain/Country/Command/AddCountryCommand.php
```diff
@@ -0,0 +1,333 @@
...
+    {
+        return $this->shopAssociation;
+    }
+
+    /**
+     * @param array $shopAssociation
+     *
+     * @return AddCountryCommand
+     */
+    public function setShopAssociation(array $shopAssociation): AddCountryCommand
```
Progi1984:
```suggestion
    public function setShopAssociation(array $shopAssociation): self
```

src/Core/Domain/Country/Command/EditCountryCommand.php
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    /**
+     * @param string $zipCodeFormat
+     *
+     * @return EditCountryCommand
+     *
+     * @throws CountryConstraintException
+     */
+    public function setZipCodeFormat(string $zipCodeFormat): EditCountryCommand
```
Progi1984:
```suggestion
    public function setZipCodeFormat(string $zipCodeFormat): self
```

src/Core/Domain/Country/Command/EditCountryCommand.php
```diff
@@ -0,0 +1,289 @@
...
+
+        return $this;
+    }
+
+    /**
+     * @param int[] $shopAssociation
+     *
+     * @return EditCountryCommand
+     */
+    public function setShopAssociation(array $shopAssociation): EditCountryCommand
```
Progi1984:
```suggestion
    public function setShopAssociation(array $shopAssociation): self
```

src/Core/Domain/Country/Query/GetCountryForEditing.php
```diff
@@ -0,0 +1,59 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Country\Exception\CountryConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Country\ValueObject\CountryId;
+
+/**
+ * Gets country information for editing.
+ */
+class GetCountryForEditing
+{
+    /**
+     * @var CountryId
```
Progi1984:
```suggestion
     * @var CountryId|null
```

src/Core/Domain/Country/Query/GetCountryForEditing.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public function __construct(int $countryId)
+    {
+        $this->countryId = new CountryId($countryId);
+    }
+
+    /**
+     * @return CountryId
+     */
+    public function getCountryId(): CountryId
```
Progi1984:
```suggestion
    public function getCountryId(): ?CountryId
```

src/Core/Domain/Country/ValueObject/CountryZipCodeFormat.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * @param string $zipCodeFormat
+     *
+     * @throws CountryConstraintException
+     */
+    private function assertIsValidZipCodeFormat(string $zipCodeFormat): void
+    {
+        if (!preg_match('/^[NLCnlc 0-9-]+$/', $zipCodeFormat)) {
```
Progi1984: Please use const

Progi1984: And check if the string is not used anywhere

src/Core/Form/IdentifiableObject/DataHandler/CountryFormDataHandler.php
```diff
@@ -0,0 +1,141 @@
...
+
+        if (null !== $data['zip_code_format']) {
+            $command->setZipCodeFormat($data['zip_code_format']);
+        }
+
+        if (null !== $data['shop_association']) {
+            $command->setShopAssociation($data['shop_association']);
+        }
+
+        $this->commandBus->handle($command);
```
Progi1984, line 141: No return ?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/available_fields_content.html.twig
```diff
@@ -0,0 +1,32 @@
+{% set tabActive = 'active show' %}
+<ul class="nav nav-tabs address-format-tabs" id="custom-address-fields">
+  {% for key, rows in availableFields %}
+    {% if loop.index != 1 %}
+      {% set tabActive = '' %}
+    {% endif %}
+    <li class="nav-item">
+      <a href="#availableListFieldsFor_{{ key }}" class="nav-link  {{ tabActive }}">
```
Progi1984:
```suggestion
      <a href="#availableListFieldsFor_{{ key }}" class="nav-link  {{ loop.index == 1 ? 'active show' : '' }}">
```

Progi1984 (commented):
Some feedback :)

src/Core/Form/IdentifiableObject/DataHandler/CountryFormDataHandler.php
tdavidsonas88: If you check all the others handlers firing **Edit**NameCommand - they all do not return anything

src/Core/Domain/Country/Query/GetCountryForEditing.php
tdavidsonas88: implemented this but not sure why null is also expected

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/available_fields_content.html.twig
tdavidsonas88: very nice review comments :slightly_frowning_face:  had only two situations where I was not sure If should implement the suggestions.

tdavidsonas88:
> Some feedback :)

very nice review comments :slightly_smiling_face: :  had only two situations where I was not sure If should implement the suggestions.

admin-dev/themes/new-theme/js/pages/country/CountryForm.js
```diff
@@ -0,0 +1,124 @@
...
+    $(event.target).tab('show');
+  }
+
+  /**
+   * Adds pattern to a cursor position in format text box
+   *
+   * @param pattern
+   */
+  addFieldsToCursorPosition(pattern) {
+    const $el = $(countryFormMap.formatTextAreaField).get(0);
```
PierreRambaud, line 87: Why not using `document.querySelector` :thinking: 
It's a waste of memory

admin-dev/themes/new-theme/js/pages/country/CountryForm.js
```diff
@@ -0,0 +1,124 @@
...
+   */
+  modifyLayout(event) {
+    const confirmation = $(event.target).data('confirmationMessage');
+    let defaultLayout = $(event.target).data('defaultLayout');
+
+    if (typeof defaultLayout === 'undefined') {
+      defaultLayout = this.lastLayoutModified;
+    }
+
+    if (defaultLayout !== null) {
```
PierreRambaud, line 118:
```suggestion
    if (defaultLayout !== null && confirm(confirmation)) {
```

admin-dev/themes/new-theme/js/pages/country/form.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+import ChoiceTree from '@components/form/choice-tree';
+import TranslatableInput from '@components/translatable-input';
+import CountryForm from './CountryForm';
```
PierreRambaud, line 28:
```suggestion
import CountryForm from './country-form';
```

PierreRambaud:
@tdavidsonas88 FYI, you must remove generated assets because of develop branch :wink:

admin-dev/themes/new-theme/scss/pages/_country_form.scss
```diff
@@ -0,0 +1,37 @@
...
+ * Country form CSS
+ */
+.address-format-tabs {
+  border: 1px solid #d3d8db;
+  padding-bottom: 1px;
+  a.nav-link {
+    font-size: 14px;
+    padding: 5px 10px;
+  }
+  a.nav-link {
```
Progi1984, line 11: Remove line 10 & 11

admin-dev/themes/new-theme/scss/pages/_country_form.scss
```diff
@@ -0,0 +1,37 @@
...
+.address-format-tabs-content {
+  border: solid #d3d8db;
+  border-width: 0 1px 1px 1px;
+  a.add-pattern {
+    font-size: 12px;
+    border: 1px solid;
+    margin: 1px 0;
+    padding: 1px 5px;
+    border-radius: 5px;
+  }
```
Progi1984, line 28: Remove line 28
Replace line 29 with `&:hover {`

Progi1984 (changes requested):
A small change
Don't forget to `rebase`

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
```diff
@@ -0,0 +1,131 @@
...
+
+      {{ form_errors(countryForm) }}
+
+      {{ ps.form_group_row(countryForm.country, {}, {
+        'label': 'Country'|trans({}, 'Admin.Global'),
+        'help' : 'Country name'|trans({}, 'Admin.International.Feature') ~ ' - ' ~ 'Invalid characters:'|trans({}, 'Admin.Global') ~ ' <>;=#{} '
+      }) }}
+
+      {{ ps.form_group_row(countryForm.iso_code, {}, {
+        'label': 'ISO code'|trans({}, 'Admin.International.Feature'),
```
LouiseBonnard, line 47:
```suggestion
        'label': 'ISO 3166-1 code'|trans({}, 'Admin.International.Feature'),
```

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
```diff
@@ -0,0 +1,131 @@
...
+        'help' : 'Two -- or three -- letter ISO code (e.g. "us" for United States).'|trans({}, 'Admin.International.Help')
```
LouiseBonnard, line 48:
```suggestion
        'help' : 'Two-letter ISO 3166-1 alpha-2 code (e.g. "US" for the United States).'|trans({}, 'Admin.International.Help')
```

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/format_actions.html.twig
```diff
@@ -0,0 +1,63 @@
...
+              <i class="material-icons">delete_sweep</i>  {{ 'Clear format'|trans({}, 'Admin.International.Feature') }}
+            </a>
+  </span>
+  <small class="form-text">
+    {{ 'Want to set required fields in the customer\'s registration form? Go to [1]Customers > Addresses[/1] to configure it.'|trans(
+      {
+        '[1]': '<a href="' ~ getAdminLink('AdminAddresses', true) ~ '">',
+        '[/1]': '</a>'
+      },
+      'Admin.International.Help')|raw
```
Junebyun, line 60:
```suggestion
    {{ 'Want to set required fields in the customer's registration form? Go to [1]Customers > Addresses[/1] to configure it.'|trans(
      {
        '[1]': '<a href="' ~ getAdminLink('AdminAddresses', true) ~ '">',
        '[/1]': '</a>'
      },
      'Admin.International.Help')|raw
```

Junebyun (approved):

src/PrestaShopBundle/Controller/Admin/Improve/International/CountryController.php
```diff
@@ -0,0 +1,227 @@
...
+            $countryFormHandler = $this->get(
+                'prestashop.core.form.identifiable_object.handler.country_form_handler'
+            );
+
+            $countryForm = $countryFormBuilder->getFormFor($countryId);
+            $countryForm->handleRequest($request);
+            $result = $countryFormHandler->handleFor($countryId, $countryForm);
+
+            if ($result->isSubmitted() && $result->isValid()) {
+                $this->addFlash('success', $this->trans('Successful update.', 'Admin.Notifications.Success'));
```
LouiseBonnard, line 161: Decision has been made to use _Update successful_ in _Admin.Notifications.Success_ rather than _Successful update._, cf. https://github.com/PrestaShop/PrestaShop/pull/17819#discussion_r384527574

src/PrestaShopBundle/Controller/Admin/Improve/International/CountryController.php
```diff
@@ -0,0 +1,227 @@
...
+        $countryForm->handleRequest($request);
+
+        try {
+            /** @var AddressFormatData $addressFormat */
+            $addressFormat = $this->getQueryBus()->handle(new GetAddressFormatData());
+
+            $handlerResult = $countryFormHandler->handle($countryForm);
+
+            if ($handlerResult->isSubmitted() && $handlerResult->isValid()) {
+                $this->addFlash('success', $this->trans('Successful creation.', 'Admin.Notifications.Success'));
```
LouiseBonnard, line 97: Perhaps this should be harmonized?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Locations/Country/Blocks/form.html.twig
```diff
@@ -0,0 +1,131 @@
...
+      }) }}
+
+      {{ ps.form_group_row(countryForm.call_prefix, {}, {
+        'label': 'Call prefix'|trans({}, 'Admin.International.Feature'),
+        'help' : 'International call prefix, (e.g. 1 for United States).'|trans({}, 'Admin.International.Help')
```
LouiseBonnard, line 53:
```suggestion
        'help' : 'International call prefix, (e.g. 1 for the United States).'|trans({}, 'Admin.International.Help')
```

matks:
@RaimondasSapola  I close this PR as it currently appears in our backlog of PRs and until I (or someone else) find some time to finish it, it won't move 😞  .

zuk3975:
related: https://github.com/PrestaShop/PrestaShop/issues/17647

PR #15206, opened 2019-08-21
Smoke tests - Fix the product selector in the FO

nesrineabdmouleh:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix the product selector in the FO home page
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | URL_FO=YOUR_SHOP_URL TEST_PATH="smoke/crawlingFO" npm run specific-test 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15206)

### DIFF
--- a/tests/puppeteer/pages/FO/home.js
+++ b/tests/puppeteer/pages/FO/home.js
@@ -6,7 +6,7 @@ module.exports = class Home extends CommonPage {
 
     // Selectors for home page
     this.logoHomePage = '#_desktop_logo';
-    this.productImg = '#content div:nth-child(%NUMBER) article img';
+    this.productImg = '#content .products div:nth-child(%NUMBER) article img';
     this.userInfoLink = '#_desktop_user_info';
     this.contactLink = '#contact-link';
   }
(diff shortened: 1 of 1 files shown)


tests/puppeteer/pages/FO/home.js
```diff
@@ -6,7 +6,7 @@ module.exports = class Home extends CommonPage {
 
     // Selectors for home page
     this.logoHomePage = '#_desktop_logo';
-    this.productImg = '#content article:nth-child(%NUMBER) img';
+    this.productImg = '#content div:nth-child(%NUMBER) article img';
```
boubkerbribri:
```suggestion
    this.productImg = '.products div:nth-child(%NUMBER) article img';
```
Maybe starting from list of products is better than content which contain carousel and some other things ?

SimonGrn: I think it's better too

nesrineabdmouleh: Yes, it's true, but I think that it's better to use an id than a class name in a selector
I propose to add an id for the products bloc, what do you think?

boubkerbribri:
```suggestion
    this.productImg = '#content .products div:nth-child(%NUMBER) article img';
```

boubkerbribri (approved):
Looks good
Thanks @nesrineabdmouleh

matthieu-rolland (approved):
lgtm

PR #15216, opened 2019-08-22
Allow users to edit address alias during checkout

### LINKED ISSUE #9913: [BOOM-3692] FO - Alias changed when editing adresse during checkout
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-3692).

- _**Reporter:**_ prasanth
- _**Created at:**_ Wed, 23 Aug 2017 10:44:07 +0200

<p>When we create a customer address there is an alias field. <br/>
In FO it's an optional field and in BO it's required.<br/>
That field is for your customer's benefit. The alias allows them to easily remember which address is which on their account, for example <em>Home</em> or <em>Office</em>.<br/>
During checkout process, this field isn't display and If I edit an address and go back to my account settings alias field is changed to the default name "<em>My address</em>".</p>

        
Create an address
<br/>
Edit it during checkout process and check alias in account settings

pauloffb:
Registered users can edit adress alias during checkout. With current behaviour if they add several addresses right from the checkout they get a bunch of "My address" which can confuse users.There is an issue open for this #9913.

| Questions     | Answers
| Branch?       | develop
| Description?  | When registered users add adresses from the checkout steps, they don't see the field for editing the alias for the address, and they accumulate "My address" which is the default alias. With this fix they see the alias field as an optional that they can edit or not.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | Don't think so.
| Deprecations? | Don't think so... sorry.
| Fixed ticket? | Fixes #9913
| How to test?  | Try to add addresses from the checkout phase bot as a guest and you should't see the alias field and as a registered user and you should see the alias field as an option.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15216)

### DIFF
--- a/themes/classic/templates/checkout/_partials/address-form.tpl
+++ b/themes/classic/templates/checkout/_partials/address-form.tpl
@@ -1,8 +1,8 @@
 {extends file='customer/_partials/address-form.tpl'}
 
 {block name='form_field'}
-  {if $field.name eq "alias"}
-    {* we don't ask for alias here *}
+  {if $field.name eq "alias" and $customer.is_guest}
+    {* we don't ask for alias here if customer is not registered *}
   {else}
     {$smarty.block.parent}
   {/if}


matthieu-rolland:
Thank you for your contribution @pauloffb !

@colinegin and @marionf 

This PR adds the **alias** field when editing an address during checkout. 
I ping you to know if this is really what you want, regarding this issue #9913

Capture d’écran 2019-08-22 à 17 11 31

themes/classic/templates/checkout/_partials/address-form.tpl
```diff
@@ -1,8 +1,8 @@
 {extends file='customer/_partials/address-form.tpl'}
 
 {block name='form_field'}
-  {if $field.name eq "alias"}
-    {* we don't ask for alias here *}
+  {if $field.name eq "alias" and $customer.is_guest}
+    {* we don't ask for alias here if customer is not registered *}
```
matthieu-rolland, line 5: This looks good to me, codewise, but I'll wait for the PM team's opinion before approving your PR. Thanks !

colinegin:
Hello @matthieu-rolland and @pauloffb ,

First thanks for your contribution !
This behaviour makes sense to me but for more consistency, I think we should also add the field to the account creation form during the checkout process.
Would that be easy for you to do it @pauloffb ?

Thanks !

pauloffb:
Hi @colinegin and @matthieu-rolland,

I see this during checkout with account creation, do you see it differently or are we talking about different things? :smile:

RTJDQcCyNY

 Or do you mean that the field should also appear as a guest, because Prestashop also gives guest addresses the default "My Address" alias and the guest could change his mind after checking out and completes the account creation form?

colinegin:
My bad, I thought your PR only added the alias field when editing the address and fixed the issue detailed in #9913.
What you actually do with this PR is :
- fixing the issue related to the change in the alias name when editing the address in the checkout process
- adding the alias field when editing the address form in the checkout
- adding the alias field in the form during account creation in the checkout

Am I correct ? If yes, this sounds perfect to me !

pauloffb:
I believe it would "fix" all those 3 things... at least making it consistent.

However one could argue **if the alias field should really be required** (in BO it is really required, in FO it is not, but it automatically assigns the default "My address" which is even worse).

colinegin:
Maybe this field could be pre-filled with "My address" but letting the user change it if needed.
What do you think ?

matthieu-rolland (approved):
LGTM;

Thank you for your contribution @pauloffb  👍

pauloffb:
Hi @colinegin and @matthieu-rolland,

Just seen @colinegin comment...

> Maybe this field could be pre-filled with "My address" but letting the user change it if needed.
> What do you think ?

I think that makes sense, should I do it? Can I just make another commit to this PR?

colinegin:
@pauloffb what is the behaviour you chose to apply in this situation ?

pauloffb:
@colinegin it is explained before. The field is empty but is shown to the user, if he choose to put someting in there it will store with that alias, if the user choose not to edit, it will store with the default alias (by language: My Address, Meu endereço...)

I agree that maybe the correct behavior should be to prefill the field with the default alias. The correction can't be made on the theme templates as before, we'll have to get it from here (I guess):
https://github.com/PrestaShop/PrestaShop/blob/3024a1693eaa2a5519f1bd3baef43da12e23e7b8/classes/form/CustomerAddressForm.php#L184-L219

colinegin:
@pauloffb ,

no worries let's leave it like it and merge this PR :)

matthieu-rolland:
Thank you @pauloffb !

PR #15229, opened 2019-08-23
Migration of Sell > Customers > Addresses page

RaimondasSapola:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migrated  Sell > Customers > Addresses page grid page, filters action and required fields form
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10557
| How to test?  | go to Sell > Customers > Addresses. Page should be as it was before the PR (actually since the link is not enabled yet go to `/sell/addresses`)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15229)


src/Core/Grid/Definition/Factory/AddressGridDefinitionFactory.php
```diff
@@ -0,0 +1,306 @@
...
+                        ],
+                        'required' => false,
+                    ])
+                    ->setAssociatedColumn('id_address')
+            )
+            ->add(
+                (new Filter('firstname', TextType::class))
+                    ->setTypeOptions([
+                        'attr' => [
+                            'placeholder' => $this->trans('Search first nam', [], 'Admin.Actions'),
```
LouiseBonnard:
```suggestion
                            'placeholder' => $this->trans('Search first name', [], 'Admin.Actions'),
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/required_fields.html.twig
```diff
@@ -0,0 +1,48 @@
...
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ *#}
+
+<div class="collapse" id="addressRequiredFieldsContainer">
+  {{ form_start(requiredFieldsForm, {'action': path('admin_addresses_save_required_fields')}) }}
+  <div class="card" >
+    <h3 class="card-header">
+      {{ 'Required Fields'|trans }}
```
LouiseBonnard: Please localize it in _Admin.Orderscustomers.Feature_, thank you.

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/required_fields.html.twig
```diff
@@ -0,0 +1,48 @@
...
+    </h3>
+    <div class="card-body">
+      <div class="alert alert-info" role="alert">
+        <div class="alert-text">
+          <p>{{ 'Select the fields you would like to be required for this section.'|trans }}</p>
```
LouiseBonnard: Do we want to keep this notification? It does not bring any further information apart from repeating it three times:

Capture d’écran 2019-08-26 à 10 36 11

@colinegin, @marionf, @samuel-pires, what do you think?

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/index.html.twig
```diff
@@ -0,0 +1,66 @@
...
+        <p>
+          <button class="btn btn-primary"
+                  type="button"
+                  data-toggle="collapse"
+                  data-target="#addressRequiredFieldsContainer"
+                  aria-expanded="false"
+                  aria-controls="addressRequiredFieldsContainer"
+          >
+            <i class="material-icons">add_circle</i>
+            {{ 'Set required fields for this section'|trans }}
```
LouiseBonnard: Same here, it should be localized in _Admin.Orderscustomers.Feature_. :-)

LouiseBonnard:
Hi there, one more thing about this page... when you want to add a new address, the very first field of the form appears with the wrong UI. Can this be corrected in this PR?

Capture d’écran 2019-08-26 à 10 42 18

RaimondasSapola:
Wrong UI will be fixed in Sell > Customers > Addresses Create/Edit form PR

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/required_fields.html.twig
marionf: Maybe we can add an information message / tooltip here, to say that before putting a mandatory field, you need to check if this field is well in the address format in International > Locations > Countries
Like we did here: https://github.com/PrestaShop/PrestaShop/pull/15180#discussion_r316551995
@LouiseBonnard wdyt ?

LouiseBonnard: Great idea! What about _Required fields apply to the customer's registration form, you should check the address formats in [1]International > Locations > Countries[/1] before._ localized in _Admin.Orderscustomers.Help_?

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/required_fields.html.twig
```diff
@@ -27,12 +27,20 @@
...
       <div class="alert alert-info" role="alert">
         <div class="alert-text">
-          <p>{{ 'Select the fields you would like to be required for this section.'|trans }}</p>
+          <p>
+            {{ 'Required fields apply to the customer\'s registration form, you should check the address formats in [1]International > Locations > Countries[/1] before.'|trans(
+              {
+                '[1]': '<a href="' ~ getAdminLink('AdminCountries', true) ~ '">',
+                '[/1]': '</a>'
+              },
+              'Admin.International.Help')|raw
```
LouiseBonnard: _Admin.Orderscustomers.Help_, please. ;-)

src/Core/Domain/Address/ValueObject/RequiredFields.php
```diff
@@ -0,0 +1,47 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Address\ValueObject;
+
+/**
+ * Stores required field values for address.
```
sarjon: Description is not really correct, see below comment. :)

src/Core/Domain/Address/ValueObject/RequiredFields.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+class RequiredFields
+{
+    /**
+     * Stores all allowed required fields
```
sarjon: These are not really "required" fields, but fields that can be configured as required. :)

src/Core/Form/ChoiceProvider/AddressRequiredFieldsChoiceProvider.php
```diff
@@ -0,0 +1,44 @@
...
+ * Provides choices for configuring required fields for address
+ */
+final class AddressRequiredFieldsChoiceProvider implements FormChoiceProviderInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getChoices()
+    {
+        return array_combine(RequiredFields::ALLOWED_REQUIRED_FIELDS, RequiredFields::ALLOWED_REQUIRED_FIELDS);
```
sarjon, line 42: Looks interesting, but I guess it works. :man_shrugging:

src/Core/Grid/Definition/Factory/AddressGridDefinitionFactory.php
```diff
@@ -0,0 +1,306 @@
...
+ * Class is responsible for defining 'Sell > Customer > Addresses' grid.
+ */
+final class AddressGridDefinitionFactory extends AbstractGridDefinitionFactory
+{
+    public const GRID_ID = 'address';
+
+    /**
+     * @param HookDispatcherInterface $hookDispatcher
+     */
+    public function __construct(HookDispatcherInterface $hookDispatcher)
```
sarjon: I think you don't need to redeclare constructor in this case. :)

src/Core/Grid/Query/AddressQueryBuilder.php
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCountQueryBuilder(SearchCriteriaInterface $searchCriteria): QueryBuilder
+    {
+        $qb = $this->getQueryBuilder($searchCriteria->getFilters())
+            ->select('COUNT(DISTINCT a.`id_address`)')
+            ->where('a.`id_customer` != 0')
```
sarjon: Duplicate condition, see `$this->getQueryBuilder`

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -0,0 +1,292 @@
...
+
+            try {
+                $this->getCommandBus()->handle(new SetRequiredFieldsForAddressCommand($data['required_fields']));
+            } catch (AddressException $e) {
+                $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));
+
+                return $this->redirectToRoute('admin_addresses_index');
+            }
+
+            $this->addFlash('success', $this->trans('Update successful', 'Admin.Notifications.Success'));
```
sarjon: You could move this statement to `try` block and remove `return` from `catch`, thus optimizing code a little. :)

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -0,0 +1,292 @@
...
+
+            return $this->redirectToRoute('admin_addresses_index');
```
sarjon: This could also be removed.

src/PrestaShopBundle/Form/Admin/Sell/Address/RequiredFieldsAddressType.php
```diff
@@ -0,0 +1,63 @@
...
+{
+    /**
+     * @var array
+     */
+    private $addressRequiredFieldsChoices;
+
+    /**
+     * @param array $addressRequiredFieldsChoices
+     */
+    public function __construct(array $addressRequiredFieldsChoices)
```
sarjon: If you can pass service instead to get choices instead of array, that would be better imo.

sarjon (commented):
Some minor comments. :)

src/Core/Domain/Address/Command/SetRequiredFieldsForAddressCommand.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        if (empty($requiredFields)) {
+            return;
+        }
+
+        if (!empty(array_diff($requiredFields, RequiredFields::ALLOWED_REQUIRED_FIELDS))) {
+            throw new InvalidAddressRequiredFieldsException(
+                sprintf(
+                    'Invalid address required fields provided. Allowed fields are: %s',
+                    implode(',', RequiredFields::ALLOWED_REQUIRED_FIELDS)
```
matks: @PierreRambaud I know you've played with this part recently 😛 do you think this is a bad thing in the context of shop customization ? This prevents a module developer to add custom fields and make them required, is that an issue ?

src/Core/Domain/Address/ValueObject/RequiredFields.php
```diff
@@ -0,0 +1,47 @@
...
+ * Stores configurable required field values for address.
+ */
+class RequiredFields
+{
+    /**
+     * Stores all allowed required fields to be configured for address
+     */
+    const ALLOWED_REQUIRED_FIELDS = [
+        'company',
+        'address2',
```
matks, line 39: No `address1` ?

matks (commented):
@RaimondasSapola sorry for late review 😢 

Thanks to @sarjon 1st review I have nothing to say about the code, it's clean ✅ 
I just have 2 questions, see below 😉

src/Core/Domain/Address/ValueObject/RequiredFields.php
RaimondasSapola: If i remember correctly address1 is always required and this list only stores fields that are configurable to be required or not

src/Core/Domain/Address/Command/SetRequiredFieldsForAddressCommand.php
PierreRambaud: Yes it's an issue, you'll not be able to add custom field that are only manage by hooks :/

RaimondasSapola: There was no explicit hook for required fields so how should we implement this in migration? And is this an issue for this PR @PierreRambaud @matks ?

matks: @RaimondasSapola At controller level, modules can add new required fields (using existing hooks). This looks like a relevant usecase so I think we need to address it 🤔

RaimondasSapola: Ok i am a bit confused. So the problem here is that module developers cannot add new required fields after migration and we should look into implementing hooks in for requiredFields migration or is it that this validation only allows predefined required fields to be added and module developers might want to add completely new fields?

matks: @RaimondasSapola Both 😄

> So the problem here is that module developers cannot add new required fields after migration

Yes because `this validation only allows predefined required fields to be added`

> we should look into implementing hooks in for requiredFields migration

Since this is controlled on controller level and module can override controllers using Symfony re-routing I think it's already OK but I'll check. Leave this to me 👍 

> module developers might want to add completely new fields?

Yes, this is a 100% relevant usecase. Module developers want to add new fields in the database and want to be able to:
- read them
- list them in grids
- modify them in add/edit forms
- make them required fields

RaimondasSapola: So in this case validation should be removed altogether?

admin-dev/themes/new-theme/.webpack/common.js
```diff
@@ -90,6 +90,7 @@ module.exports = {
     attachment: './js/pages/attachment',
     attribute: './js/pages/attribute',
     attribute_group: './js/pages/attribute-group',
+    address: './js/pages/address',
```
PierreRambaud, line 97: Could you sort this hash please?
At least you'll never have a conflict again (or less).

src/Adapter/Address/CommandHandler/BulkDeleteAddressHandler.php
```diff
@@ -38,18 +39,27 @@ final class BulkDeleteAddressHandler extends AbstractAddressHandler implements B
...
+                if (!$this->deleteAddress($address)) {
+                    $errors[] = $address->id;
+                }
+            } catch (AddressException $e) {
+                $errors[] = $addressId->getValue();
             }
         }
+
+        if (!empty($errors)) {
+            throw new BulkDeleteAddressException($errors, 'Failed to delete all of selected addresses');
```
PierreRambaud, line 62: :+1:

src/Adapter/Address/CommandHandler/SetRequiredFieldsForAddressHandler.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * {@inheritdoc}
+     *
+     * @throws CannotSetRequiredFieldsForAddressException
+     */
+    public function handle(SetRequiredFieldsForAddressCommand $command)
+    {
+        $address = new CustomerAddress();
+
+        try {
```
PierreRambaud, line 51:
```php
try {
    if ($address->addFieldsRequiredDatabase($command->getRequiredFields())) {
        return;
    }
} catch (PrestaShopDatabaseException $e) {
}
throw new CannotSetRequiredFieldsForAddressException(
    sprintf('Cannot set "%s" required fields for customer', implode(',', $command->getRequiredFields()))
);
```

PierreRambaud: The throw is redundant

RaimondasSapola:
@Robin-Fischer-PS Rebased

jolelievre (dismissed):

Robin-Fischer-PS:
Hi @RaimondasSapola !

I've an issue on this PR : 

When an address is soft deleted, it still appears on the list.

See screen record : Addresses ID 1 and 2 are hard deleted (since no order linked to these addresses), while ID 5 is soft deleted. After deletion, address with ID 5 is still shown on the list.

https://drive.google.com/open?id=1RpUboPhaeVUbaD_BQ29FUzuafKrx9lCT

Thanks :smile:

RaimondasSapola:
@Robin-Fischer-PS added fix :)

src/Core/Domain/Address/Command/SetRequiredFieldsForAddressCommand.php
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param string[] $requiredFields
+     */
+    private function assertContainsOnlyAllowedFields(array $requiredFields)
+    {
+        if (empty($requiredFields)) {
+            return;
+        }
+
+        //TODO add dynamic field validation
```
jolelievre, line 68: This can't be done now? I guess it is related to`RequiredFields::ALLOWED_REQUIRED_FIELDS`?

src/Core/Grid/Definition/Factory/AddressGridDefinitionFactory.php
```diff
@@ -0,0 +1,297 @@
...
+                    ->setTypeOptions([
+                        'required' => false,
+                    ])
+                    ->setAssociatedColumn('country_name')
+            )
+            ->add(
+                (new Filter('actions', SearchAndResetType::class))
+                    ->setAssociatedColumn('actions')
+                    ->setTypeOptions([
+                        'reset_route' => 'admin_common_reset_search',
```
jolelievre: Shouldn't you use the filterId feature rather, the filters matched by controller/action are a bit outdated since they have limitations (even though it works fine for One grid pages)

src/Core/Domain/Address/Command/SetRequiredFieldsForAddressCommand.php
RaimondasSapola: It is decided to do in a separate PR

jolelievre: Ok, I was gonna propose to add this in the issue but it's already done 😉

src/Core/Grid/Definition/Factory/AddressGridDefinitionFactory.php
RaimondasSapola: Yeh it's done. Sometimes it becomes difficult to follow witch method is the most recent and the one to use

jolelievre (approved):
Thank you @RaimondasSapola

src/Core/Grid/Definition/Factory/AddressGridDefinitionFactory.php
jolelievre: Sure, there are many things to remember And we lack documentation as well I planned to add som about this filtering feature as soon as I have time ^^

Robin-Fischer-PS:
Thanks @RaimondasSapola ! It's :heavy_check_mark: for me :)

jolelievre:
Thank you @RaimondasSapola

Issue #13829, opened 2019-05-17
[EPIC] Upgrade to jQuery latest, use jQuery migrate for retro compatibility


eternoendless:
We are currently using old or very old versions of jQuery. These versions have never been upgraded because of breaking changes introduced in later versions. Using [jQuery Migrate](https://github.com/jquery/jquery-migrate) we could upgrade them to the latest version without breaking backwards compatibility.

## Improvements

- [x] #14929 - Upgrade to jQuery latest, use jQuery migrate for retro compatibility 1/4 - CORE THEME & CLASSIC THEME
- [x] #14930 - Upgrade to jQuery latest, use jQuery migrate for retro compatibility 2/4 - DEFAULT THEME BO
- [x] #14931 - Upgrade to jQuery latest, use jQuery migrate for retro compatibility 3/4 - UI KIT
- [x] #15508 - Update jquery to latest version in Installation process 4/4

FredBrockstedt:
There are three JS security vulnerabilities in 1.7.5.2 (and lower versions):
2 - JQuery2.2.4 - Medium
1 - JQuery UI1.10.3 - High
They are fixed in the current versions of JQuery/UI.

ghost:
To cut in 4 parts:

- [x] Core theme & Classic Theme #14929 
- [x] Default Theme BO #14930 
- [x] UI kit #14931 
- [x] Installation process #15508

matthieu-rolland:
Here is the official migration guide:

https://jquery.com/upgrade-guide/3.0/#overview

For example we already use a jquery-migrate (version 1.2.1) in the back office, and it can't be used simultaneously with jquery-migrate 3.x.

So in a nutshell, the process should be as follow:

**1/** Change jquery version from 1.11 to 1.12 (minimal version supported by the migration script)

**2/** Check that nothing is broken, fix when it's broken

**3/** Once bugs have been fixed, change jquery version to 3.0 and add the jquery-migrate v3 `jquery-migrate-3.1.0.js `

**4/** Check that there is no bug, update jquery libraries if needed.

matthieu-rolland:
I added an issue to this epic: #15508

ghost:
Hi @matthieu-rolland 
Are we completely done with this EPIC or are there still some related tasks to add to the EPIC?

PR #15245, opened 2019-08-23
Update jQuery to 3.4.1 in the legacy Back Office pages

### LINKED ISSUE #14930: Upgrade to jQuery latest, use jQuery migrate for retro compatibility 2/4 - DEFAULT THEME BO
EPIC : #13829

**Please update the EPIC after completion**

We are currently using old or very old versions of jQuery. These versions have never been upgraded because of breaking changes introduced in later versions. Using jQuery Migrate we could upgrade them to the latest version without breaking backward compatibility.

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migrate jquery in back office (from 1.11 to 3.4.1)
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | fixes #14930
| How to test?  | Check that nothing is broken in back office

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15245)


matthieu-rolland:
Note that I didn't delete our jquery 1.11.0 file because it is still used for the installation process.

classes/Media.php
```diff
@@ -330,8 +330,9 @@ public static function getJqueryPath($version = null, $folder = null, $minifier
         }
 
         // added jQuery migrate for compatibility with new version of jQuery
-        // will be removed when using latest version of jQuery
-        $return[] = Media::getJSPath(_PS_JS_DIR_ . 'jquery/jquery-migrate-1.2.1.min.js');
+        if ($version === _PS_JQUERY_VERSION_) {
```
PierreRambaud: Should we really need to add jquery-migrate, only when the selected version is equal to latest?

matthieu-rolland: Yes, this jquery migrate is only compatible with jquery 3, the way it works is that jquery-migrate makes sure the legacy jquery code that we wrote before will still work with jquery 3.

In fact, this method is always called using the latest version of jquery, but if for some reason in the future someone passes another version of jquery as a parameter, we don't want to call this jquery-migrate.  (it was wrong to include jquery migrate all the time before)

matthieu-rolland: Maybe we could parse the name of the version passed as a parameter and include jquery migrate if it is a version 3.x ?  It seemed overkilled but I could do it. (I just don't see why we would have several versions of jquery 3)

PierreRambaud: If a user use `addJquery('3.2.1')` he maybe doesn't want to use jquery-migrate.
But, what if he changed the `_PS_JQUERY_VERSION_` to jQuery 2? Will it break his back office because of jquery migrate (which will be added).
Should we store the '3.4.1' elsewhere?

matthieu-rolland: This method is supposed to be generic, but it's only used for back office... so I see two ways to go about this:

- Refactor the method, maybe rename it and make it clear that this is for back office purpose, and force a version (except if we want to be able to have different versions of jquery on different bo pages?)

- Store jquery version / Jquery migrate relationships somewhere, and use it in this method.

PierreRambaud: ping @eternoendless @matks @Quetzacoalt91  wdyt?

eternoendless: Right now, `addJquery` has no effect in `FrontControllers`, `ModuleFrontControllers` and symfony controllers.

This leaves us only with `AdminControllers` (legacy BO controllers) and `ModuleAdminControllers`.

I think the `addJquery` method should be deprecated in those as well, basically because inserting a different version of jQuery will most probably break the BO.

eternoendless (changes requested):
I think the old jQuery and jQuery migrate should be removed as well shouldn't it?

matthieu-rolland:
@eternoendless I was going to delete it but the old jquery (1.11) is still used during the installation process...

classes/Media.php
matthieu-rolland: So you suggest to deprecate the addJquery method, and to call the last version of jquery and it's migration file directly?

mickaelandrieu:
is it possible to also use jQuery 3 in the Front Office?

Progi1984:
@mickaelandrieu Hi, jQuery 3 will be updated on BO, Core/Classic Theme and UI Kit (EPIC #13829)

classes/controller/AdminController.php
```diff
@@ -2701,7 +2701,10 @@ public function setMedia($isNewTheme = false)
             $this->addCSS(__PS_BASE_URI__ . $this->admin_webpath . '/themes/' . $this->bo_theme . '/css/vendor/titatoggle-min.css', 'all', 0);
             $this->addCSS(__PS_BASE_URI__ . $this->admin_webpath . '/themes/' . $this->bo_theme . '/public/theme.css', 'all', 0);
 
-            $this->addJquery();
+            // add Jquery 3 and its migration script
+            $this->addJs(_PS_JS_DIR_ . 'jquery/jquery-3.4.1.min.js');
+            $this->addJs(_PS_JS_DIR_ . 'jquery/jquery-migrate-3.1.0.min.js');
```
matthieu-rolland, line 2707: @eternoendless @PierreRambaud 

So I deprecated the addJquery method and called jquery 3 directly in AdminController, with other javascript assets. This is only executed in legacy BO pages.

I just wonder if I should still put the jquery version in a constant, like it used to be? It would make sense if for some reason we wanted back office and front office to alway share the same jquery version, I don't think it's the case, is it?

PierreRambaud: Not sure it's useful since jquery migrate requires a jquery versions range :thinking: 
More, just asking, but the jquery and jquery.migrate dependencies shouldn't be in package.json?

matthieu-rolland: I just cheked and there is no mention of jquery in admin default theme's package.json.

PierreRambaud (dismissed):

classes/Media.php
eternoendless: Yeah

classes/Media.php
```diff
@@ -291,7 +291,9 @@ public static function getMediaPath($mediaUri, $cssMediaType = null)
      *
      * @param mixed $version
      *
-     * @return string
+     * @return array
+     *
+     * @deprecated 1.7.7
```
eternoendless: Please specify why this is deprecated ("jQuery is always included, this method should no longer be used) and add a deprecation error (see https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Tools.php#L721-L743 for an example)

classes/Media.php
```diff
@@ -771,6 +773,13 @@ public static function addJsDefL($params, $content, $smarty = null, &$repeat = f
         }
     }
 
+    /**
+     * @param $output
+     *
+     * @return string|string[]|null
+     *
+     * @deprecated 1.7.7
```
eternoendless: Same comment here. 

Also please describe the deprecations in the pull request description (I already added the label).

classes/controller/Controller.php
```diff
@@ -536,6 +536,8 @@ public function removeJS($js_uri, $check_path = true)
      * @param string|null $version jQuery library version
      * @param string|null $folder jQuery file folder
      * @param bool $minifier if set tot true, a minified version will be included
+     *
+     * @deprecated 1.7.7
```
eternoendless: Same comment about deprecations

eternoendless (dismissed):

eternoendless:
> I was going to delete it but the old jquery (1.11) is still used during the installation process...

In that case the install process must be updated as well (please add an issue to the epic) so that it can be removed.

classes/Media.php
matthieu-rolland: done !

classes/Media.php
matthieu-rolland: done !

classes/controller/Controller.php
matthieu-rolland: done !

matthieu-rolland:
> > I was going to delete it but the old jquery (1.11) is still used during the installation process...
> 
> In that case the install process must be updated as well (please add an issue to the epic) so that it can be removed.

Yep I created an issue and added it to the epic #13829

classes/controller/AdminController.php
```diff
@@ -4869,4 +4872,22 @@ private function prepareNumberSpecifications(Context $context)
...
+    /**
+     * Adds jQuery library file to queued JS file list.
+     *
+     * @param string|null $version jQuery library version
+     * @param string|null $folder jQuery file folder
+     * @param bool $minifier if set tot true, a minified version will be included
+     *
+     * @deprecated 1.7.7 jQuery is always included, this method should no longer be used
+     */
+    public function addJquery($version = null, $folder = null, $minifier = true)
```
jolelievre: I understand it is deprecated, however it is only added when `setMedia` is called, but are we sure it is always called? What if a module controller only calls `addJquery` but not `setMedia`?
Maybe you should still add the required js files here:
```javascript
$this->addJs(_PS_JS_DIR_ . 'jquery/jquery-3.4.1.min.js');
$this->addJs(_PS_JS_DIR_ . 'jquery/jquery-migrate-3.1.0.min.js');
```

Since `addJs` checks the already added script it won't duplicate it any way.

Unless `setMedia` is always called by the controller class anyway, in which case forget my comment ;)

classes/Media.php
```diff
@@ -291,10 +291,16 @@ public static function getMediaPath($mediaUri, $cssMediaType = null)
...
+     * @return array
+     *
+     * @deprecated 1.7.7 jQuery is always included, this method should no longer be used
      */
     public static function getJqueryPath($version = null, $folder = null, $minifier = true)
     {
+        @trigger_error(
+            'Media::getJqueryPath() is deprecated since version 1.7.7.0, jquery is always included',
+            E_USER_DEPRECATED
+        );
```
jolelievre, line 303: Deprecated ok, but then maybe you should hard code the path to jquery and jquery-migrate just like in `setMedia` ?

classes/controller/Controller.php
```diff
@@ -536,9 +536,15 @@ public function removeJS($js_uri, $check_path = true)
      * @param string|null $version jQuery library version
      * @param string|null $folder jQuery file folder
      * @param bool $minifier if set tot true, a minified version will be included
+     *
+     * @deprecated 1.7.7 jQuery is always included, this method should no longer be used
      */
     public function addJquery($version = null, $folder = null, $minifier = true)
     {
+        @trigger_error(
```
jolelievre, line 544: Mmm, it is always included in `AdminController` but what about front controller? Nothing is sure in this case, and it might be difficult to detect if it was added since now it is bundled in the core js (cf your other PR https://github.com/PrestaShop/PrestaShop/issues/14929)
Maybe you should modify the `FrontController::addJquery` method in the other PR by the way?

classes/controller/AdminController.php
jolelievre: Maybe you can use npm to install these files instead of hard copy them?

PierreRambaud: If we use package.json without compilation, this means we have the node_modules directory. So we can't add these dependencies right now :/

classes/controller/Controller.php
matthieu-rolland: @jolelievre  The addJquery method exists in the FrontController, but it is empty, a comment specifies that jquery is already added by the core. So this should not be an issue.

classes/controller/AdminController.php
matthieu-rolland: @jolelievre I confirm that the setMedia method is always called from the controller in `ControllerCore::Run()`.

classes/controller/Controller.php
jolelievre: Yes but it is added in the core.js no? What if someone uses a theme without including core.js but uses this addJquery method instead?
The real question actually is I'm wondering what's the use of this method if jquery is always added by the core (one way or another) ?

matthieu-rolland: This method is useless and should not be used anymore, if someone did not include core.js and called this method, as of today he doesn't have jquery, because the method is empty, it does nothing...

jolelievre: ok, so we take upon ourselves that this method is now useless?
but this is weird then, because `addJquery` is empty in admin and front controller, but not in the parent Controller class..
Well anyway it doesn't prevent your PR from working, it just doesn't seem logic 😅
We can leave it as is and avoid a too long discussion

matks:
@matthieu-rolland  it seems you wait for @jolelievre validation of the changes you added following his code review, is that correct ?

matthieu-rolland:
@matks true ! (up)

@jolelievre I deleted the empty addJquery methods in the controllers extending ControllerCore, the one in controllerCore has the deprecation notice.

classes/controller/Controller.php
matthieu-rolland: @jolelievre So I removed the empty addJquery methods in BO and FO controller, and instead emptied the addJquery method in the parent controller. 👍

PierreRambaud (approved):

zuk3975:
@matthieu-rolland, @matks  - after this pr legacy order create page is not working well.

- Cannot choose customer;
- Carts/orders are shown before even choosing customer. 

Screenshot from 2019-10-09 14-18-47
Checked locally, that undoing this pr changes brings back the expected behavior.

matthieu-rolland:
@zuk3975 The goal of this PR is to keep legacy jquery running smoothly, but for some reason jquery-migrate doesn't fix all the holes 🤔  (especially features that were already deprecated before our former version of jquery). 

So from now on we should use jQuery's latest API only.

The **live** method doesn't exist anymore, one should use the **on** method instead.

Here is the **on** method:
https://api.jquery.com/on/

Here is how to use the **on** method to achieve the same thing as with the **live** method:
`$( document ).on( events, selector, data, handler );` 

The equivalent with the **live** method would be:
`$( selector ).live( events, data, handler );`

The main difference is that you put a parent selector in `$()`, and the target selector as the second parameter of the **on** method.  If you just put "document" in `$()`it will be costlier that if you use a closer parent selector.

matthieu-rolland:
@zuk3975 I just realised you were talking about the LEGACY order page, and not the new one you are working on !

I will make an issue and fix it in legacy code. 👍

PR #15271, opened 2019-08-27
Use aria-label instead of .sr-only hidden span

davidglezz:
| Questions     | Answers
| Branch?       | develop
| Description?  | Its better use aria-label attribute instead of hidden spans, like https://github.com/PrestaShop/PrestaShop/pull/15083 . More info https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Use screen reader on the site

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15271)


PierreRambaud (dismissed):

PierreRambaud (approved):

PierreRambaud:
Thanks @davidglezz

PR #15286, opened 2019-08-28
After modifying a product by webservice, the product position in the categories changes, even if we will use current position.

BlackKerio:
| Questions     | Answers
| Branch?       | develop
| Description?  | After modifying a product by webservice, the product position in the categories changes, even if we will use current position.<br>the function getWsPositionInCategory does not return correct values, bad syntax.<br>The function setWsPositionInCategory makes a change even if the wrong parameter is entered, missing return.<br>When creating a new product, the position in the category is never 0, but always starts from 1.<br>Unfortunately, the array of values from the sql function is indexing from 0, so I add an item to the top of the list to move the index to 1
| Type?         | bug fix 
| Category?     | WS
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | you have to make a modification a product with the webservice, <position_in_category notFilterable="true"><![CDATA[1]]></position_in_category>

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15286)

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -6103,10 +6103,11 @@ class ProductCore extends ObjectModel
      */
     public function getWsPositionInCategory()
     {
-        $result = Db::getInstance()->executeS('SELECT position
+        $result = Db::getInstance()->executeS(
+            'SELECT `position`
             FROM `' . _DB_PREFIX_ . 'category_product`
-            WHERE id_category = ' . (int) $this->id_category_default . '
-            AND id_product = ' . (int) $this->id);
+            WHERE `id_category` = ' . (int) $this->id_category_default . '
+            AND `id_product` = ' . (int) $this->id);
         if (count($result) > 0) {
             return $result[0]['position'];
         }
@@ -6121,22 +6122,46 @@ class ProductCore extends ObjectModel
      */
     public function setWsPositionInCategory($position)
     {
-        if ($position < 0) {
-            WebserviceRequest::getInstance()->setError(500, $this->trans('You cannot set a negative position, the minimum for a position is 0.', array(), 'Admin.Catalog.Notification'), 134);
-        }
-        $result = Db::getInstance()->executeS('
-            SELECT `id_product`
-            FROM `' . _DB_PREFIX_ . 'category_product`
-            WHERE `id_category` = ' . (int) $this->id_category_default . '
-            ORDER BY `position`
-        ');
-        if (($position > 0) && ($position + 1 > count($result))) {
-            WebserviceRequest::getInstance()->setError(500, $this->trans('You cannot set a position greater than the total number of products in the category, minus 1 (position numbering starts at 0).', array(), 'Admin.Catalog.Notification'), 135);
+        if ($position <= 0) {
+            WebserviceRequest::getInstance()->setError(
+                500,
+                $this->trans(
+                    'You cannot set 0 or a negative position, the minimum is 1.',
+                    array(),
+                    'Admin.Catalog.Notification'
+                ),
+                134
+            );
+
+            return false;
         }
 
+        $result = Db::getInstance()->executeS(
+            'SELECT `id_product` ' .
+            'FROM `' . _DB_PREFIX_ . 'category_product` ' .
+            'WHERE `id_category` = ' . (int) $this->id_category_default . '  ' .
+            'ORDER BY `position`'
+        );
+
+        if ($position > count($result)) {
+            WebserviceRequest::getInstance()->setError(
+                500,
+                $this->trans(
+                    'You cannot set a position greater than the total number of products in the category, starting at 1.',
+                    array(),
+                    'Admin.Catalog.Notification'
+                ),
+                135
+            );
+
+            return false;
+        }
+
+        array_unshift($result, null);
         foreach ($result as &$value) {
             $value = $value['id_product'];
         }
+
         $current_position = $this->getWsPositionInCategory();
 
         if ($current_position && isset($result[$current_position])) {
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -6157,6 +6157,7 @@ class ProductCore extends ObjectModel
             return false;
         }
 
+        // result is indexed by recordset order and not position. positions start at index 1 so we need an empty element
         array_unshift($result, null);
         foreach ($result as &$value) {
             $value = $value['id_product'];


classes/Product.php
```diff
@@ -6088,18 +6089,22 @@ public function getWsPositionInCategory()
      */
     public function setWsPositionInCategory($position)
     {
-        if ($position < 0) {
-            WebserviceRequest::getInstance()->setError(500, $this->trans('You cannot set a negative position, the minimum for a position is 0.', array(), 'Admin.Catalog.Notification'), 134);
+        if ($position <= 0) {
+            WebserviceRequest::getInstance()->setError(500, $this->trans('You cannot set a negative position or 0, the minimum for a position is 1.', array(), 'Admin.Catalog.Notification'), 134);
```
LouiseBonnard: I would suggest _You cannot set 0 or a negative position, the minimum is 1._ to shorten the wording.

classes/Product.php
```diff
@@ -6088,18 +6089,22 @@ public function getWsPositionInCategory()
...
         $result = Db::getInstance()->executeS('
             SELECT `id_product`
             FROM `' . _DB_PREFIX_ . 'category_product`
             WHERE `id_category` = ' . (int) $this->id_category_default . '
             ORDER BY `position`
         ');
-        if (($position > 0) && ($position + 1 > count($result))) {
-            WebserviceRequest::getInstance()->setError(500, $this->trans('You cannot set a position greater than the total number of products in the category, minus 1 (position numbering starts at 0).', array(), 'Admin.Catalog.Notification'), 135);
+        if (($position > 0) && ($position > count($result))) {
+            WebserviceRequest::getInstance()->setError(500, $this->trans('You cannot set a position greater than the total number of products in the category, (position numbering starts at 1).', array(), 'Admin.Catalog.Notification'), 135);
```
LouiseBonnard: What about _You cannot set a position greater than the total number of products in the category, starting at 1._ instead?

classes/Product.php
BlackKerio: thank you

BlackKerio:
I'm sorry, but I don't know what code format is required to pass through PrettyCI

PierreRambaud:
@BlackKerio You must run php-cs-fixer command, or manually remove all unwanted spaces.

BlackKerio:
@PierreRambaud on peut de faire le pull stp ?

PierreRambaud (dismissed):

PierreRambaud:
PR rebased and squashed. I also fix the lint and a useless double check :wink:

matks:
🎉

classes/Product.php
```diff
@@ -6121,22 +6122,46 @@ public function getWsPositionInCategory()
...
+                    array(),
+                    'Admin.Catalog.Notification'
+                ),
+                135
+            );
+
+            return false;
         }
 
+        array_unshift($result, null);
```
eternoendless, line 6161:
```suggestion
        // result is indexed by recordset order and not position. positions start at index 1 so we need an empty element
        array_unshift($result, null);
```

eternoendless (approved):

khouloudbelguith:
Hi @BlackKerio,

I used this script to update a product via webservice
```
<?php
/**
 * Created by PhpStorm.
 * User: khouloud.belguith
 * Date: 29/11/19
 * Time: 10:58
 */
define('DEBUG', true);
ini_set('display_errors','on');
define('PS_SHOP_PATH', 'http://shop/');
define('PS_WS_AUTH_KEY', 'Key');
require_once('PSWebServiceLibrary.php');
$webService= new PrestaShopWebservice(PS_SHOP_PATH, PS_WS_AUTH_KEY, DEBUG);
$id =  24; 
$category_id = 2;
$xml = $webService->get(array('resource' => 'products', 'id' => $id));
unset($xml->product->manufacturer_name);
unset($xml->product->quantity);
$product = $xml->children()->children();
$product->name->language[0][0] = "Produit webservice edit";
$product->description->language[0][0] = "Description produit webservice";
$product->price = 12;
$product->position_in_category = 23; //first position
//I don't update XML, i just resend it here
$opt['putXml'] = $xml->asXML();                  
$opt['id'] = $id;
$opt['category_id'] = $category_id;
$opt['resource'] = 'products';
$xml = $webService->edit($opt);
echo "DONE";
```
But the position of the product is changed
https://drive.google.com/file/d/1IBlyKNYr-pgztZMimc2UxLV9xpN1Xd1m/view
So, the problem is not solved.
Thanks!

BlackKerio:
@khouloudbelguith 
If we make a change to the product before correcting the error, the table unfortunately contains bad data. The index starts from 0. The first change, therefore, corrects the table, the index starts from 1 and other changes are already working properly. Changing the order is possible only in the default category, parameter id_category_default. In your test program there are lines 
$category_id = 2;
$opt ['category_id'] = '$ category_id';
without meaning and I would remove them.
I recommend clearing the cache and previewing the table.

khouloudbelguith:
Hi @BlackKerio,

I tried again a fresh install with your patch.
I edited the default product with id = 17 using this script.
```php
<?php

define('DEBUG', true);
ini_set('display_errors','on');
define('PS_SHOP_PATH', 'http://shop.com//');
define('PS_WS_AUTH_KEY', 'key');
require_once('PSWebServiceLibrary.php');
$webService= new PrestaShopWebservice(PS_SHOP_PATH, PS_WS_AUTH_KEY, DEBUG);
$id =  17;
$xml = $webService->get(array('resource' => 'products', 'id' => $id));
unset($xml->product->manufacturer_name);
unset($xml->product->quantity);

$product = $xml->children()->children();
$product->name->language[0][0] = "Brown bear notebook webservice edit";

$product->description->language[0][0] = "Description produit webservice";

$product->price = 12;
$product->position_in_category = 2;
//I don't update XML, i just resend it here
$opt['putXml'] = $xml->asXML();
$opt['id'] = $id;
$opt['resource'] = 'products';
$xml = $webService->edit($opt);
echo "DONE";
```
The position in NOK after the update.

https://drive.google.com/file/d/1g862iTW0T9yC4NZ_aVZzTQD4EBVCZeIU/view

Thanks!

BlackKerio:
Well, I found the cause of the different result. I tested with unset ($ xml-> product-> associations); Unfortunately, if you use associations, when you add, remove, or just leave categories, another setWsCategories function will delete and then reinsert the product in all of the listed categories. Update [16697](https://github.com/PrestaShop/PrestaShop/pull/16697)

khouloudbelguith:
@BlackKerio, so, we need to close this PR? since it is replaced by this: https://github.com/PrestaShop/PrestaShop/pull/16697

Thanks!

PierreRambaud:
Closed in favor of #16697

BlackKerio:
@khouloudbelguith, @PierreRambaud  Ok, we need both fixes, each fixes one function

PierreRambaud:
@BlackKerio We should fix the position problem in one pull request if possible :)

PR #15300, opened 2019-08-29
Migration for Sell > Customers > Addresses create/edit forms

RaimondasSapola:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migrated for Sell > Customers > Addresses create/edit forms. Added new validators for customer email and country states and zip code format
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10558
| How to test?  | You should see the migrated pages in Sell > Customers > Addresses and [this link](https://github.com/PrestaShop/PrestaShop/issues/10558#issuecomment-427882685) should be using the migrated page as well.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15300)


RaimondasSapola:
Waiting for https://github.com/PrestaShop/PrestaShop/pull/15229 to be merged for address required fields provider

RaimondasSapola:
Also question for @matks remains about if something should be done about https://github.com/PrestaShop/PrestaShop/pull/15180#issuecomment-525726316 issue

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,484 @@
...
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults(
+            [
+                'validation_groups' => function () {
+                    $groups = ['Default'];
+
+                    //TODO add required fields as validation groups
```
RaimondasSapola: @matks i have a question about adding new dataProvider for this case. In controller i use queryBus to fetch address required fields. It is a simple array of strings with names of required fields. For this form i need this array to add validation groups and required flags based on the same array.

Would it be a good idea to create a service `RequiredFieldsProvider` with interface witch runs the same query and this service here and at the Controller level to get required fields?

Or get rid of validation groups and pass required values as form options and use if statements to determine if fields are required?

What do you think?

matks: So, we should try to avoid dispatching too many queries **for the same HTTP request**.

CQRS ideal usecase is: "you receive one HTTP request, you dispatch 1 query and/or 1 command". That's the ideal usecase. One usecase = 1 query and/or 1 command. So if it's 1 query, then we're supposed to ask, in the query, for *all the data we need to display* and get it all together.

That is why I suggest the following, which is based on your idea:
- create this RequiredFieldsProvider
- use this RequiredFieldsProvider inside GetCustomerAddressForEditingHandler and return the required fields data into the EditableCustomerAddress so that AddressFormDataProvider can retrieve it and pass this data to the form so it can be validated correctly
- you can also use this RequiredFieldsProvider for the 2nd query (the ones you use in controller to get the data)

Is this alright for you ? If we perform a GetCustomerAddressForEditing query, it means we're asking the Domain for "all the data I need to perform an editing operation for customer address". If the "required fields" are needed to perform an editing operation they should be part of the Query response.

RaimondasSapola: Required fields are also needed for create operation so returning them with EditableCustomerAddress only solves half the problem. Or maybe for edit form i could get all required fields with EditableCustomerAddress and for create i can pass required fields as data to form?

src/Core/ConstraintValidator/CustomerAddressCountryRequiredFieldsValidator.php
```diff
@@ -0,0 +1,92 @@
...
+     * @throws CountryConstraintException
+     */
+    public function validate($value, Constraint $constraint)
+    {
+        $countryId = (int) $value['id_country'];
+        $stateId = isset($value['id_state']) ? $value['id_state'] : null;
+        $dni = isset($value['dni']) ? $value['dni'] : null;
+
+        /** @var CountryRequiredFields $requiredFields */
+        $requiredFields = $this->queryBus->handle(new GetCountryRequiredFields($countryId));
```
RaimondasSapola: So @matks we talked about required fields provider service and we rejected it because for create and edit cases data could be passed in data providers but i totally forgot abut validation case. So here i see two solutions. Use required fields provider service or pass required fields from form as constraint option. Wdyt?

matks: What about this:
- edit case: we retrieve the required fields from Query
- create case: we retrieve the required fields from Query too
- validation: we use a Provider

And internally, the QueryHandlers use the Provider that has been injected into them. Does it make sense ?

I say this because I think so far, all of the usecases where we needed data to perform validation was retrieved through an Adapter 🤔 I hope I'm not wrong ?

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
```diff
@@ -0,0 +1,139 @@
...
+            'class': 'js-manufacturer-country-select',
+            'data-customer-information-url': path('admin_customer_for_address_information') }}, {
+            'label': 'Customer email'|trans({}, 'Admin.Orderscustomers.Feature'),
+          }) }}
+        {% endif %}
+      {% endblock %}
+
+      {{ ps.form_group_row(addressForm.dni, {}, {
+        'label': 'Identification number'|trans({}, 'Admin.Orderscustomers.Feature'),
+        'help' : 'The national ID card number of this person, or a unique tax identification number.'|trans({}, 'Admin.Orderscustomers.Help')
```
LouiseBonnard, line 67: You're right to localize it under the _help_ label here. It is localized under the _feature_ label on Crowdin so it will duplicate the string but I think it is necessary in this case. :-)

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
```diff
@@ -0,0 +1,139 @@
...
+      }) }}
+
+      {{ ps.form_group_row(addressForm.alias, {}, {
+        'label': 'Address alias'|trans({}, 'Admin.Orderscustomers.Feature'),
+        'help' : 'Invalid characters:'|trans({}, 'Admin.Notifications.Info') ~ ' <>;=#{}'
+      }) }}
+
+      {{ ps.form_group_row(addressForm.first_name, {}, {
+        'label': 'First Name'|trans({}, 'Admin.Global'),
```
LouiseBonnard:
```suggestion
        'label': 'First name'|trans({}, 'Admin.Global'),
```

LouiseBonnard: We must stay consistent with the other migrated pages, let's avoid unnecessary upper cases.

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
```diff
@@ -0,0 +1,139 @@
...
+        'help' : 'Invalid characters:'|trans({}, 'Admin.Notifications.Info') ~ ' 0-9!<>,;?=+()@#"�{}_$%:'
+      }) }}
+
+      {{ ps.form_group_row(addressForm.last_name, {}, {
+        'label': 'Last Name'|trans({}, 'Admin.Global'),
```
LouiseBonnard:
```suggestion
        'label': 'Last name'|trans({}, 'Admin.Global'),
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
```diff
@@ -0,0 +1,139 @@
...
+      {{ ps.form_group_row(addressForm.vat_number, {}, {
+        'label': 'VAT number'|trans({}, 'Admin.Orderscustomers.Feature'),
+      }) }}
+
+      {{ ps.form_group_row(addressForm.address1, {}, {
+        'label': 'Address'|trans({}, 'Admin.Global'),
+      }) }}
+
+      {{ ps.form_group_row(addressForm.address2, {}, {
+        'label': 'Address'|trans({}, 'Admin.Global') ~ ' (2)',
```
LouiseBonnard: Shouldn't it be _Address (2)_ here?

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
```diff
@@ -0,0 +1,139 @@
...
+      {% endif %}
+
+      {{ ps.form_group_row(addressForm.id_state, {}, {
+        'label': 'State'|trans({}, 'Admin.Global'),
+        'class': 'js-address-state-select',
+        'row_attr': rowAttr|default({})
+      }) }}
+
+      {{ ps.form_group_row(addressForm.phone, {}, {
+        'label': 'Home phone'|trans({}, 'Admin.Global')
```
LouiseBonnard: It feels weird to read home phone... what about _Phone number_ instead?

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
```diff
@@ -0,0 +1,139 @@
...
+      }) }}
+
+      {% if addressForm.phone_mobile is defined %}
+        {{ ps.form_group_row(addressForm.phone_mobile, {}, {
+          'label': 'Mobile phone'|trans({}, 'Admin.Global')
+        }) }}
+      {% endif %}
+
+      {{ ps.form_group_row(addressForm.other, {}, {
+        'label': 'Other'|trans({}, 'Admin.Notifications.Info'),
```
LouiseBonnard:
```suggestion
        'label': 'Other'|trans({}, 'Admin.Global'),
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
```diff
@@ -0,0 +1,139 @@
...
+        'row_attr': rowAttr|default({})
+      }) }}
+
+      {{ ps.form_group_row(addressForm.phone, {}, {
+        'label': 'Home phone'|trans({}, 'Admin.Global')
+      }) }}
+
+      {% if addressForm.phone_mobile is defined %}
+        {{ ps.form_group_row(addressForm.phone_mobile, {}, {
+          'label': 'Mobile phone'|trans({}, 'Admin.Global')
```
LouiseBonnard, line 131: I wonder if we need two different fields. And if we do, how to introduce them so that it does not sound too outdated...
🤔

src/Core/ConstraintValidator/ExistingCustomerEmailValidator.php
```diff
@@ -0,0 +1,76 @@
...
+     * {@inheritdoc}
+     */
+    public function validate($value, Constraint $constraint)
+    {
+        if (!$constraint instanceof ExistingCustomerEmail) {
+            throw new UnexpectedTypeException($constraint, ExistingCustomerEmail::class);
+        }
+
+        try {
+            $this->queryBus->handle(new GetCustomerForAddressCreation($value));
```
RaimondasSapola: @matks another situation with queryBus in validator :( To have a core interface for this one case seems a bit useless. Maybe queryBus is not so bad here?

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
RaimondasSapola: Legacy had Home phone. I can make it Phone number. Leave it in Admin.Global domain?

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
RaimondasSapola: Yah we need those 😞 As for how to call it i can only suggest one of the names from the list of stationary phone synonyms https://www.powerthesaurus.org/stationary_phone/synonyms 😄

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
LouiseBonnard: Better, _Phone_ exists in the _Admin.Global_ domain. Let's pick this one instead!

LouiseBonnard:
```suggestion
        'label': 'Phone'|trans({}, 'Admin.Global')
```

src/Adapter/Country/CountryRequiredFieldsProvider.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    private function getCountry(CountryId $countryId): Country
+    {
+        $countryIdValue = $countryId->getValue();
+
+        try {
+            $country = new Country($countryIdValue);
+        } catch (PrestaShopException $e) {
+            throw new CountryNotFoundException(
+                sprintf('Country with id "%s" was not found.', $countryIdValue)
```
PierreRambaud: Could be another exception, the message must be more generic or use the `$e->getMessage()`. 
Otherwise you can do this to prevent duplicate messages.

```php
try {
    $country = new Country($countryIdValue);
} catch (PrestaShopException $e) {
}

if (empty($country->id) || $country->id !== $countryIdValue) {
    throw new CountryNotFoundException(
        sprintf('Country with id "%s" was not found.', $countryIdValue)
    );
}
```

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+        );
+    }
+
+    /**
+     * @param string $field
+     * @param array $requiredFields
+     *
+     * @return NotBlank|null
+     */
+    private function getNotBlackOrNull(string $field, array $requiredFields): ?NotBlank
```
matks:
```suggestion
    private function getNotBlanckOrNull(string $field, array $requiredFields): ?NotBlank
```

tests/Unit/Core/Domain/Country/ZipCodePatternResolverTest.php
```diff
@@ -0,0 +1,155 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace Tests\Unit\Core\Util\Country;
+
+use PHPUnit\Framework\TestCase;
+use PrestaShop\PrestaShop\Core\Domain\Country\ZipCodePatternResolver;
+
+class ZipCodePatternResolverTest extends TestCase
```
matks, line 32: ❤️

matks:
OK @RaimondasSapola I've discussed this issue with @eternoendless

### Summary

In this PR, there is 3 custom validators used to validate the Symfony form:
- `CustomerAddressCountryRequiredFieldsValidator` which uses an adapter CountryRequiredFieldsProviderInterface to get the data it needs to validate the input data it is given
- `CustomerAddressZipCodeValidator` which uses an adapter CountryZipCodeRequirementsProviderInterface to get the data it needs to validate the input data it is given
- `ExistingCustomerEmailValidator` which (currently) uses a dispatched `GetCustomerForAddressCreation` Query to get the data it needs to validate the input data it is given

### About the usage of CQRS in the app layer

I confirmed with @eternoendless that Commands and Queries must be used in order to address user-oriented usecases. This means "user needs to see a dashboard with data A, B and C" => it's a Query. "User needs to modify data X state" => it's a Command.

However here the user does not want to validate the form (when the form is submitted). The system needs to validate the form prior to executing the true intent of the user which is to persist the data provided by the user.

So the answer is: **we will not use Queries nor Commands** to retrieve the data needed by the validators, but instead some classes, we can label them as **data providers** in charge of providing this data. These data providers are read-only.

### About this PR and the implementation

For this PR, it means:
- CustomerAddressCountryRequiredFieldsValidator => it's good as it relies on an interface with an adapter implementation (the adapter will later be removed and replaced by a doctrine repository implementation)
- CustomerAddressZipCodeValidator => it's good as it relies on an interface with an adapter implementation (the adapter will later be removed and replaced by a doctrine repository implementation)
- ExistingCustomerEmailValidator => we need to remove the use of the QueryBus here and replace it with a call to a data provider

The data provider must be an interface. See examples `PrestaShop\PrestaShop\Core\Localization\Currency\RepositoryInterface`, `PrestaShop\PrestaShop\Core\Localization\Currency\DataSourceInterface`.

And the interface can be implemented by either
- an adapter class that rely on ObjectModel (this solution must be chosen if 1) there is already an ObjectModel that can provide this data with an existing function or 2) using the ObjectModel allows us to implement the function with very few lines of code)
- a Core class that rely on Doctrine DBAL to perform SQL queries (this solution must be chosen if previous usecase does not apply)

matks:
@RaimondasSapola Apart from that, I also reviewed the PR and I found only 1 issue (see above): `getNotBlackOrNull` => `getNotBlanckOrNull`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
rokaszygmantas: `getNotBlankOrNull` is the correct wording :slightly_smiling_face:

matks: Indeed 😅

admin-dev/themes/new-theme/js/pages/address/CustomerAddressForm.js
```diff
@@ -0,0 +1,157 @@
...
+    $(addressFormMap.countrySelect).on('change', () => this._handleCountryChange());
+    $(addressFormMap.customerEmail).on('blur', (event) => this._handleEmailChange(event));
+  }
+
+  /**
+   * Hide state select if country doesnt have states, show it otherwise and fill with data.
+   *
+   * @private
+   */
+  _handleCountryChange() {
```
jolelievre: Actually this whole feature was already implemented for manufacturer form adresses, so I think you should be able to adapt it for customer address forms. https://github.com/PrestaShop/PrestaShop/blob/d8d0a7b6e7368615662641c5e4ec6073bfbac77e/admin-dev/themes/new-theme/js/pages/manufacturer/manufacturer_address_form.js#L33

In manufacturer form we also manage `CountryDniRequiredToggler` for countries with dni, so I think you will have the same issue for customers.

admin-dev/themes/new-theme/js/pages/address/CustomerAddressForm.js
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  /**
+   * Handles email change event to get customer data for customer fields
+   *
+   * @param event
+   *
+   * @private
+   */
+  _handleEmailChange(event) {
```
jolelievre: Since you can remove the country management from this class, its only purpose will be to deal with the email So you can transform this class into an extension `CheckEmailExtension` and inject it the check url, so it could be used for Employye form for example just by switching the check url
And then you init this new extension (along with `CountryStateSelectionToggler` and `CountryDniRequiredToggler`) in `admin-dev/themes/new-theme/js/pages/address/form.js`

src/Adapter/Address/AbstractAddressHandler.php
```diff
@@ -38,13 +40,14 @@
...
-     *
      * @param AddressId $addressId
      *
      * @return Address
+     *
+     * @throws AddressException
+     * @throws AddressNotFoundException
      */
-    protected function getAddress(AddressId $addressId)
+    protected function getAddress(AddressId $addressId): Address
```
jolelievre: Since this class was introduced in 176 you shouldn't change the type here, or this will cause a BC for classes that would have inherited it without the strong typing
We only use strong typing with new classes/methods

src/Adapter/Address/AbstractAddressHandler.php
```diff
@@ -81,4 +84,30 @@ protected function deleteAddress(Address $address)
...
+
+    /**
+     * @return string[]
+     *
+     * @throws AddressException
+     */
+    protected function getRequiredFields(): array
+    {
+        try {
+            $requiredFields = (new CustomerAddress())->getFieldsRequiredDatabase();
```
jolelievre: You can't use `CustomerAddress` here since this abstract class is dedicated to any Address object not just the customer ones. So either this method needs to be moved in each class that need it, or create a  `AbstractCustomerAddressHandler` class that extends this one

Or maybe you can simply use `$requiredFields = (new Address())->getFieldsRequiredDatabase();` since `CustomerAddress` extends `Address` anyway, but I don't know if some people use the children classes to add specific rules.. That might be the case so you would need to use specific classes for each case.

src/Adapter/Address/CommandHandler/AddCustomerAddressHandler.php
```diff
@@ -0,0 +1,128 @@
...
+
+        if (null !== $command->getPostCode()) {
+            $address->postcode = $command->getPostCode();
+        }
+
+        if (null !== $command->getAddress2()) {
+            $address->address2 = $command->getAddress2();
+        }
+
+        if (null !== $command->getIdNumber()) {
```
jolelievre: It seems weird that the getter here is named `getIdNumber()` because in other places (ManufacturerAddress) we use `getDni()`
Your proposal is probably more correct but I'm worried the inconsistency might be confusing
Although since this field seems to be the ID number of person I wonder why it was originally called `dni` I don't understand the acronym..

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * @param EditCustomerAddressCommand $command
+     *
+     * @return Address
+     *
+     * @throws AddressException
+     * @throws AddressNotFoundException
+     */
+    private function getAddressFromCommand(EditCustomerAddressCommand $command): Address
```
jolelievre, line 80: It's more `createAddressFromCommand `

src/Adapter/Address/AbstractAddressHandler.php
jolelievre: Or maybe we can keep this one, but then we should modify the ones for ManufacturerAddress (and any other adresses..), meaning add `getIdNumber/setIdNumber` and declare `getDni/SetDni` deprecated

src/Adapter/Address/QueryHandler/GetCustomerAddressForEditingHandler.php
```diff
@@ -0,0 +1,129 @@
...
+            $address->alias,
+            $address->firstname,
+            $address->lastname,
+            $address->address1,
+            $address->city,
+            new CountryId((int) $address->id_country),
+            $this->getRequiredFields()
+        );
+
+        if (null !== $address->postcode) {
```
jolelievre: I'm not sure this check is useful:
- if `$address->postcode` is null then you can simply use `$editableCustomerAddress->setPostCode($address->postcode);` without the if, the command value will still be null anyway
- the if would have meaning with a `if (!empty($address->postcode))` in which case empty string in the object is equivalent to null in your command, it depends if you want to propagate the empty string or not, or if null has specific meaning

src/Adapter/Address/QueryHandler/GetCustomerAddressForEditingHandler.php
```diff
@@ -0,0 +1,129 @@
...
+            $editableCustomerAddress->setPostCode($address->postcode);
+        }
+
+        if (null !== $address->dni) {
```
jolelievre: Same comment for all those fields

src/Core/ConstraintValidator/CustomerAddressCountryRequiredFieldsValidator.php
```diff
@@ -0,0 +1,90 @@
...
+
+        if (!$this->countryRequiredFieldsProvider->isStatesRequired($countryIdVo) && null !== $stateId) {
+            $this->context->buildViolation($constraint->countryStateMessage)
+                ->atPath('[id_state]')
+                ->setTranslationDomain('Admin.Orderscustomers.Notification')
+                ->addViolation()
+            ;
+        }
+
+        if ($this->countryRequiredFieldsProvider->isDniRequired($countryIdVo) && null !== $dni) {
```
jolelievre: Your condition is not logical here, either dni it should be one of the two:
- `if ($this->countryRequiredFieldsProvider->isDniRequired($countryIdVo) && null === $dni)`
- `if (!$this->countryRequiredFieldsProvider->isDniRequired($countryIdVo) && null !== $dni)`

Besides I think you should use `empty()` rather than strictly checking the null value

src/Adapter/Country/CountryRequiredFieldsProvider.php
```diff
@@ -0,0 +1,83 @@
...
+class CountryRequiredFieldsProvider implements CountryRequiredFieldsProviderInterface
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @throws CountryNotFoundException
+     */
+    public function isStatesRequired(CountryId $countryId): bool
+    {
+        return $this->getCountry($countryId)->contains_states;
```
jolelievre: Add a cast here `(bool)` as ObjectModel are not really strict with their fields ^^

src/Adapter/Country/CountryRequiredFieldsProvider.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @throws CountryNotFoundException
+     */
+    public function isDniRequired(CountryId $countryId): bool
+    {
+        return $this->getCountry($countryId)->need_identification_number;
```
jolelievre: `(bool)` here as well

src/Adapter/Country/CountryRequiredFieldsProvider.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        return $this->getCountry($countryId)->contains_states;
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @throws CountryNotFoundException
+     */
+    public function isDniRequired(CountryId $countryId): bool
```
jolelievre: Maybe you can reuse some of the modification from this PR https://github.com/PrestaShop/PrestaShop/pull/16153/files
Especially `CountryDataProvider::getCountriesIdWhichNeedDni` which a bit more optimized as it doesn't instanciate any Country object but gets the data you need directly from the database
Maybe a similar function `CountryDataProvider::getCountriesIdWhichNeedState` could be added

src/Adapter/Customer/CustomerDataSource.php
```diff
@@ -0,0 +1,41 @@
...
+
+use Customer;
+use PrestaShop\PrestaShop\Core\Customer\CustomerDataSourceInterface;
+
+final class CustomerDataSource implements CustomerDataSourceInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function hasCustomerByEmail(string $email): bool
```
jolelievre: Maybe `hasCustomerWithEmail` would be more adapted It's just that the `BySomething` suffix feels more adapted to search function like `getByEmail` `findByCountry` and so on
But for a check function using `With` implies a bit more the check and unicity of the result, what do you think?

src/Core/Form/IdentifiableObject/DataHandler/AddressFormDataHandler.php
```diff
@@ -0,0 +1,195 @@
...
+     * {@inheritdoc]
+     *
+     * @throws CountryConstraintException
+     * @throws CustomerException
+     * @throws StateConstraintException
+     */
+    public function create(array $data)
+    {
+        /** @var AddressCreationCustomerInformation $addressCustomer */
+        $addressCustomer = $this->commandBus->handle(new GetCustomerForAddressCreation($data['customer_email']));
```
jolelievre: It seems you only use `$addressCustomer->getCustomerId()->getValue()` here, all the other data come from the form I saw that `AddressCreationCustomerInformation` is used somewhere else to fetch the customer data, but in this case maybe a provider or a repository would be enough `CustomerDataProvider::getIdByEmail` ?

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -262,14 +303,89 @@ public function createAction(): Response
...
      *
      * @return Response
      */
-    public function editAction(int $addressId): Response
+    public function editAction(int $addressId, Request $request): Response
     {
-        $link = $this->getAdminLink('AdminAddresses', ['id_address' => $addressId, 'updateaddress' => 1]);
+        try {
+            /** @var EditableCustomerAddress $editableAddress */
+            $editableAddress = $this->getQueryBus()->handle(new GetCustomerAddressForEditing((int) $addressId));
```
jolelievre, line 344: I think this call is redundant because the form handler already calls `$editableAddress = $this->queryBus->handle(new GetCustomerAddressForEditing((int) $addressId));`
So you can catch `AddressException` in the same block that handles the form (in case the ID is not valid), as form the `$customerInfo` construction maybe you can do it based on the Form data?

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -262,14 +303,89 @@ public function createAction(): Response
...
+    }
+
+    /**
+     * Provides customer information for address creation in json format
+     *
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function getCustomerInformationAction(Request $request): Response
```
jolelievre: Missing `@AdminSecurity` annotation, which makes me wonder if this should linked to `AddressController` or `CustomerController` Logically it should be customer, then maybe this action should be moved in another controller, but the query is really specific as it's `GetCustomerForAddressCreation` (maybe too specific?)

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -262,14 +303,89 @@ public function createAction(): Response
...
+    {
+        try {
+            $email = $request->query->get('email');
+
+            /** @var AddressCreationCustomerInformation $customerInformation */
+            $customerInformation = $this->getQueryBus()->handle(new GetCustomerForAddressCreation($email));
+
+            $serializer = $this->get('prestashop.bundle.snake_case_serializer_json');
```
jolelievre: We decided in other PRs not to use this snake case serializer, but rather use `$this->json` which uses camel case convention, so you might have to change you JS code to adapt to the new return

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                            ),
+                        ]
+                    ),
+                ],
+            ]);
+        }
+
+        $builder
+            ->add('dni', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_DNI, $requiredFields),
```
jolelievre, line 144: You need to add `'empty_data' => ''` on ALL the fields that are not required, I had to fix this bug in https://github.com/PrestaShop/PrestaShop/pull/16153/files#diff-6f21bb46d5d2e6916d3a7e4c5938b447R172

The problem is that by default when submit form the empty fields are not submitted (e.g. they are not present in the request), so the handler then interprets this missing field as null, and since in the handler we apply the modification for non null values it becomes impossible to update with empty values. The result is that once the value is set you can never remove it.

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                                'This field cannot be longer than %limit% characters',
+                                ['%limit%' => AddressConstraint::MAX_LAST_NAME_LENGTH],
+                                'Admin.Notifications.Error'
+                            ),
+                        ]
+                    ),
+                ],
+            ])
+            ->add('company', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_COMPANY, $requiredFields),
```
jolelievre, line 240: Add `'empty_data' => ''`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                                'This field cannot be longer than %limit% characters',
+                                ['%limit%' => AddressConstraint::MAX_COMPANY_LENGTH],
+                                'Admin.Notifications.Error'
+                            ),
+                        ]
+                    ),
+                ],
+            ])
+            ->add('vat_number', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_VAT_NUMBER, $requiredFields),
```
jolelievre, line 266: Add `'empty_data' => ''`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                                'This field cannot be longer than %limit% characters',
+                                ['%limit%' => AddressConstraint::MAX_ADDRESS_LENGTH],
+                                'Admin.Notifications.Error'
+                            ),
+                        ]
+                    ),
+                ],
+            ])
+            ->add('address2', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_ADDRESS_2, $requiredFields),
```
jolelievre, line 316: Add `'empty_data' => ''`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                                'This field cannot be longer than %limit% characters',
+                                ['%limit%' => AddressConstraint::MAX_CITY_LENGTH],
+                                'Admin.Notifications.Error'
+                            ),
+                        ]
+                    ),
+                ],
+            ])
+            ->add('postcode', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_POST_CODE, $requiredFields),
```
jolelievre, line 366: Add `'empty_data' => ''`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                'constraints' => [
+                    new NotBlank([
+                        'message' => $this->translator->trans(
+                            'This field cannot be empty', [], 'Admin.Notifications.Error'
+                        ),
+                    ]),
+                ],
+            ])
+            ->add('phone', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_PHONE, $requiredFields),
```
jolelievre, line 411: Add `'empty_data' => ''`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                                'This field cannot be longer than %limit% characters',
+                                ['%limit%' => AddressConstraint::MAX_PHONE_LENGTH],
+                                'Admin.Notifications.Error'
+                            ),
+                        ]
+                    ),
+                ],
+            ])
+            ->add('other', TextareaType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_OTHER, $requiredFields),
```
jolelievre, line 437: Add `'empty_data' => ''`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                'required' => true,
+                'constraints' => [
+                    new NotBlank([
+                        'message' => $this->translator->trans(
+                            'This field cannot be empty', [], 'Admin.Notifications.Error'
+                        ),
+                    ]),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'city_name',
```
jolelievre: Use `TypedRegex::TYPE_CITY_NAME`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                'required' => true,
+                'constraints' => [
+                    new NotBlank([
+                        'message' => $this->translator->trans(
+                            'This field cannot be empty', [], 'Admin.Notifications.Error'
+                        ),
+                    ]),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'phone_number',
```
jolelievre: Use `TypedRegex::TYPE_PHONE_NUMBER`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+        }
+
+        $builder
+            ->add('dni', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_DNI, $requiredFields),
+                'constraints' => [
+                    $this->getNotBlankOrNull(RequiredFields::REQUIRED_FIELD_DNI, $requiredFields),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'dni_lite',
```
jolelievre: Use `TypedRegex:: TYPE_DNI_LITE`
And also could you please update `ManufacturerAddressType` and use the constants as well? Thanks

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                'required' => true,
+                'constraints' => [
+                    new NotBlank([
+                        'message' => $this->translator->trans(
+                            'This field cannot be empty', [], 'Admin.Notifications.Error'
+                        ),
+                    ]),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'generic_name',
```
jolelievre: Use `TypedRegex::TYPE_GENERIC_NAME`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                'required' => true,
+                'constraints' => [
+                    new NotBlank([
+                        'message' => $this->translator->trans(
+                            'This field cannot be empty', [], 'Admin.Notifications.Error'
+                        ),
+                    ]),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'name',
```
jolelievre: Use `TypedRegex::TYPE_NAME`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
```
jolelievre: Use `TypedRegex::TYPE_NAME`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                    ),
+                ],
+            ])
+            ->add('company', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_COMPANY, $requiredFields),
+                'constraints' => [
+                    $this->getNotBlankOrNull(RequiredFields::REQUIRED_FIELD_COMPANY, $requiredFields),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'generic_name',
```
jolelievre: Use `TypedRegex::TYPE_GENERIC_NAME`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                    ),
+                ],
+            ])
+            ->add('vat_number', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_VAT_NUMBER, $requiredFields),
+                'constraints' => [
+                    $this->getNotBlankOrNull(RequiredFields::REQUIRED_FIELD_VAT_NUMBER, $requiredFields),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'generic_name',
```
jolelievre: Use `TypedRegex::TYPE_GENERIC_NAME`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                'required' => true,
+                'constraints' => [
+                    new NotBlank([
+                        'message' => $this->translator->trans(
+                            'This field cannot be empty', [], 'Admin.Notifications.Error'
+                        ),
+                    ]),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'address',
```
jolelievre: Use `TypedRegex::TYPE_ADDRESS`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                    ),
+                ],
+            ])
+            ->add('address2', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_ADDRESS_2, $requiredFields),
+                'constraints' => [
+                    $this->getNotBlankOrNull(RequiredFields::REQUIRED_FIELD_ADDRESS_2, $requiredFields),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'address',
```
jolelievre: Use `TypedRegex::TYPE_ADDRESS`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                    ),
+                ],
+            ])
+            ->add('postcode', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_POST_CODE, $requiredFields),
+                'constraints' => [
+                    $this->getNotBlankOrNull(RequiredFields::REQUIRED_FIELD_POST_CODE, $requiredFields),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'post_code',
```
jolelievre: Use `TypedRegex::TYPE_POST_CODE`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                    ]),
+                ],
+            ])
+            ->add('phone', TextType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_PHONE, $requiredFields),
+                'constraints' => [
+                    $this->getNotBlankOrNull(RequiredFields::REQUIRED_FIELD_PHONE, $requiredFields),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'phone_number',
```
jolelievre: Use `TypedRegex::TYPE_PHONE_NUMBER`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                    ),
+                ],
+            ])
+            ->add('other', TextareaType::class, [
+                'required' => $this->isRequired(RequiredFields::REQUIRED_FIELD_OTHER, $requiredFields),
+                'constraints' => [
+                    $this->getNotBlankOrNull(RequiredFields::REQUIRED_FIELD_OTHER, $requiredFields),
+                    new CleanHtml(),
+                    new TypedRegex([
+                        'type' => 'message',
```
jolelievre: Use `TypedRegex::TYPE_MESSAGE`

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -0,0 +1,448 @@
...
+                                'This field cannot be longer than %limit% characters',
+                                ['%limit%' => AddressConstraint::MAX_OTHER_LENGTH],
+                                'Admin.Notifications.Error'
+                            ),
+                        ]
+                    ),
+                ],
+            ]);
+
+        $builder->addEventSubscriber(new CustomerAddressFormSubscriber($this->stateChoiceProvider));
```
jolelievre: You use events to declare the state field? This seems a bit overkill since you already inject the `stateChoiceProvider`
Maybe you could inspire yourself from this FormType https://github.com/jolelievre/PrestaShop/blob/develop/src/PrestaShopBundle/Form/Admin/Sell/Address/ManufacturerAddressType.php

They are both pretty similar (maybe they should have a common abstract type), but this makes me realize that in ManufacturerAddressType the required field for state is always false.. I guess it's because it is managed in javascript but still it could be nice to initialise it correctly Once you simplify this event can you backport the modification to ManufacturerAddressType please?

src/PrestaShopBundle/Form/EventSubscriber/CustomerAddressFormSubscriber.php
```diff
@@ -0,0 +1,120 @@
...
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
+use Symfony\Component\Validator\Constraints\NotBlank;
+
+/**
+ * Subscriber resolves dynamical customer address form fields options and validation
+ */
+class CustomerAddressFormSubscriber implements EventSubscriberInterface
```
jolelievre: As explained in another comment, is this event really necessary Can't you do this in the FormType directly?

jolelievre (dismissed):
Quite a lot of feedbacks (sorry ^^) I like many of those modifications though, I'm sure most of them could be useful for manufacturer and supplier addresses (any adresses actually) This would probably be worth it to have a generic `AddressFormType` with a consistent management around it
To avoid having to duplicate the code and I'm sure that with this separated implementation when one of the form is fixed the other ones are not.. that's too bad

matks:
@atomiix please tell us when we can review again ?

atomiix:
> @atomiix please tell us when we can review again ?

@matks It's ready to be reviewed! The only part that may be useless when create order will be merged is [that file](https://github.com/PrestaShop/PrestaShop/pull/15300/files#diff-92648d120206042fc80caca9ab6ee5a2R64).

admin-dev/themes/new-theme/js/components/country-postcode-required-toggler.js
```diff
@@ -0,0 +1,78 @@
...
+    this.countryInputSelectedSelector = `${countryInputSelector}>option:selected`;
+    this.countryPostcodeInputLabelDangerSelector = `${countryPostcodeInputLabel}>span.text-danger`;
+
+    // If field is required regardless of the country
+    // keep it required
+    if (this.$countryPostcodeInput.attr('required')) {
+      return;
+    }
+
+    this.$countryInput.on('change', () => this._toggle());
```
Progi1984:
```suggestion
    this.$countryInput.on('change', () => this.toggle());
```

admin-dev/themes/new-theme/js/components/country-postcode-required-toggler.js
```diff
@@ -0,0 +1,78 @@
...
+
+    // toggle on page load
+    this._toggle();
```
Progi1984:
```suggestion
    this.toggle();
```

admin-dev/themes/new-theme/js/components/country-postcode-required-toggler.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /**
+   * Toggles Postcode input required
+   *
+   * @private
+   */
+  _toggle() {
```
Progi1984:
```suggestion
  toggle() {
```

admin-dev/themes/new-theme/js/components/country-postcode-required-toggler.js
```diff
@@ -0,0 +1,78 @@
...
+    const $countrySelectedOption = $(this.countryInputSelectedSelector);
```
Progi1984: Remove

admin-dev/themes/new-theme/js/components/country-postcode-required-toggler.js
```diff
@@ -0,0 +1,78 @@
...
+    $(this.countryPostcodeInputLabelDangerSelector).remove();
+    this.$countryPostcodeInput.attr('required', false);
+    if (1 === parseInt($countrySelectedOption.attr('need_postcode'), 10)) {
```
Progi1984:
```suggestion
    if (1 === parseInt($(this.countryInputSelectedSelector).attr('need_postcode'), 10)) {
```

admin-dev/themes/new-theme/js/components/country-postcode-required-toggler.js
```diff
@@ -0,0 +1,78 @@
...
+      this.$countryPostcodeInput.attr('required', true);
```
Progi1984:
```suggestion
      this.$countryPostcodeInput.prop('required', true);
```

admin-dev/themes/new-theme/js/components/country-state-selection-toggler.js
```diff
@@ -52,47 +52,52 @@ export default class CountryStateSelectionToggler {
     this.$countryStateSelector = $(countryStateSelector);
     this.$countryInput = $(countryInputSelector);
 
-    this.$countryInput.on('change', () => this._toggle());
+    this.$countryInput.on('change', () => this._change());
```
Progi1984:
```suggestion
    this.$countryInput.on('change', () => this.change());
```

admin-dev/themes/new-theme/js/components/country-state-selection-toggler.js
```diff
@@ -52,47 +52,52 @@ export default class CountryStateSelectionToggler {
...
 
     // toggle on page load
-    this._toggle(true);
+    this._toggle();
```
Progi1984:
```suggestion
    this.toggle();
```

admin-dev/themes/new-theme/js/components/country-state-selection-toggler.js
```diff
@@ -52,47 +52,52 @@ export default class CountryStateSelectionToggler {
...
   }
 
   /**
-   * Toggles State selection
+   * Change State selection
    *
    * @private
    */
-  _toggle(isFirstToggle = false) {
+  _change() {
```
Progi1984:
```suggestion
  change() {
```

admin-dev/themes/new-theme/js/components/country-state-selection-toggler.js
```diff
@@ -52,47 +52,52 @@ export default class CountryStateSelectionToggler {
...
+    const countryId = this.$countryInput.val();
+    if (countryId === '') {
+      return;
+    }
     $.ajax({
```
Progi1984:
```suggestion
    $.get({
```
And remove `method: 'GET'`

admin-dev/themes/new-theme/js/components/country-state-selection-toggler.js
```diff
@@ -52,47 +52,52 @@ export default class CountryStateSelectionToggler {
...
     }).then((response) => {
-      if (response.states.length === 0) {
-        this.$stateSelectionBlock.fadeOut();
+      this.$countryStateSelector.empty();
 
-        return;
-      }
+      Object.keys(response.states).forEach((value) => {
+        this.$countryStateSelector.append($('<option></option>').attr('value', response.states[value]).text(value));
+      });
```
Progi1984, line 84:
```suggestion
      Object.entries(response.states).forEach((key, value) => {
      }
        this.$countryStateSelector.append($('<option></option>').attr('value', key).text(value));
      });
```

admin-dev/themes/new-theme/js/components/country-state-selection-toggler.js
```diff
@@ -52,47 +52,52 @@ export default class CountryStateSelectionToggler {
...
-      this.$stateSelectionBlock.fadeIn();
-
-      if (isFirstToggle === false) {
-        this.$countryStateSelector.empty();
-        var _this = this;
-        $.each(response.states, function (index, value) {
-          _this.$countryStateSelector.append($('<option></option>').attr('value', value).text(index));
-        })
-      }
+      this._toggle();
```
Progi1984:
```suggestion
      this.toggle();
```

admin-dev/themes/new-theme/js/components/country-state-selection-toggler.js
```diff
@@ -52,47 +52,52 @@ export default class CountryStateSelectionToggler {
...
     }).catch((response) => {
       if (typeof response.responseJSON !== 'undefined') {
         showErrorMessage(response.responseJSON.message);
       }
     });
   }
+
+  _toggle() {
```
Progi1984:
```suggestion
  toggle() {
```

admin-dev/themes/new-theme/js/components/form/autocomplete-with-email.js
```diff
@@ -0,0 +1,55 @@
...
+const $ = window.$;
+
+export default class AutocompleteWithEmail {
+  constructor(emailInputSelector, map = []) {
+    this.map = map;
+    this.$emailInput = $(emailInputSelector);
+    this.$emailInput.on('change', () => this._change());
+  }
+
+  _change() {
```
Progi1984:
```suggestion
  change() {
```

admin-dev/themes/new-theme/js/components/form/autocomplete-with-email.js
```diff
@@ -0,0 +1,55 @@
...
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+const $ = window.$;
+
+export default class AutocompleteWithEmail {
+  constructor(emailInputSelector, map = []) {
+    this.map = map;
+    this.$emailInput = $(emailInputSelector);
+    this.$emailInput.on('change', () => this._change());
```
Progi1984:
```suggestion
    this.$emailInput.on('change', () => this.change());
```

admin-dev/themes/new-theme/js/components/form/autocomplete-with-email.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  _change() {
+    $.ajax({
```
Progi1984:
```suggestion
    $.get({
```

Remove `method: 'GET'`

admin-dev/themes/new-theme/js/components/form/autocomplete-with-email.js
```diff
@@ -0,0 +1,55 @@
...
+      dataType: 'json',
+      data: {
+        email: this.$emailInput.val(),
+      },
+    }).then((response) => {
+      Object.keys(this.map).forEach((key) => {
+        if (response[key] !== undefined) {
+          $(this.map[key]).val(response[key]);
+        }
+      });
```
Progi1984, line 47:
```
      Object.entries(this.map).forEach((key, value) => {
        if (response[key] !== undefined) {
          $(value).val(response[key]);
        }
      });
```

admin-dev/themes/new-theme/js/pages/address/form.js
```diff
@@ -0,0 +1,58 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+import AutocompleteWithEmail from '../../components/form/autocomplete-with-email';
```
Progi1984:
```suggestion
import AutocompleteWithEmail from '@components/form/autocomplete-with-email';
import CountryStateSelectionToggler from '@components/country-state-selection-toggler';
import CountryDniRequiredToggler from '@components/country-dni-required-toggler';
import CountryPostcodeRequiredToggler from '@components/country-postcode-required-toggler';
```

admin-dev/themes/new-theme/js/pages/supplier/supplier-form.js
```diff
@@ -24,6 +24,7 @@
  */
 
 import CountryStateSelectionToggler from '../../components/country-state-selection-toggler';
+import CountryDniRequiredToggler from '../../components/country-dni-required-toggler';
```
Progi1984:
```suggestion
import CountryDniRequiredToggler from '@components/country-dni-required-toggler';
```

src/Adapter/Address/AbstractAddressHandler.php
```diff
@@ -38,11 +38,12 @@
...
-     * Gets legacy address
-     *
      * @param AddressId $addressId
      *
      * @return Address
+     *
+     * @throws AddressException
+     * @throws AddressNotFoundException
      */
     protected function getAddress(AddressId $addressId)
```
Progi1984, line 48:
```suggestion
    protected function getAddress(AddressId $addressId): Address
```

atomiix: See [https://github.com/PrestaShop/PrestaShop/pull/15300#discussion_r355383609](https://github.com/PrestaShop/PrestaShop/pull/15300#discussion_r355383609) 🙂

admin-dev/themes/new-theme/js/pages/supplier/supplier-form.js
Progi1984:
```suggestion
import CountryDniRequiredToggler from '@components/country-dni-required-toggler';
```

admin-dev/themes/new-theme/js/pages/address/form.js
```diff
@@ -0,0 +1,58 @@
...
+import CountryStateSelectionToggler from '../../components/country-state-selection-toggler';
+import CountryDniRequiredToggler from '../../components/country-dni-required-toggler';
+import CountryPostcodeRequiredToggler from '../../components/country-postcode-required-toggler';
```
Progi1984: ../../components => @components

admin-dev/themes/new-theme/js/components/country-state-selection-toggler.js
atomiix: This won't work because in the data I get, the key is the state name and the value is the state id, so if I use `key` I'll have the index of the array instead of the state id.

atomiix: I'm not sur that's the best solution, `key` will be the index of the array, not the actual array key that I need. Also, `value` will be an array and I'll have to use `value[0]` to get the actual value and `value[1]` to get the array key. I think we should keep the initial proposition, wdyt?

admin-dev/themes/new-theme/js/components/form/autocomplete-with-email.js
atomiix: Same here, I need the key of the object, not the index

matks:
I checked, all review feedbacks have been adresses excluding https://github.com/PrestaShop/PrestaShop/pull/15300#pullrequestreview-334553820

As https://github.com/PrestaShop/PrestaShop/pull/15300#pullrequestreview-334553820 is in discussion, I think we can move on.

matks (approved):

PR #15312, opened 2019-08-29
Order BO tests with puppeteer - Sanity tests campaign

nesrineabdmouleh:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add sanity tests campaign, the first tests in OrdersBO are filter and edit an order in the back office
| Type?         | refacto
| Category?     |  TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | TEST_PATH="sanity/ordersBO" URL_FO=SHOP_URL npm run specific-test

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15312)


tests/puppeteer/campaigns/sanity/ordersBO/01_filterOrder.js
```diff
@@ -0,0 +1,65 @@
...
+  orderPage = await (new OrderPage(page));
+};
+
+/*
+  Connect to the BO
+  Filter the Orders table
+  Logout from the BO
+ */
+global.scenario('Filter the Orders table by ID, REFERENCE, STATUS', async () => {
+  test('should login into BO', async () => {
```
SimonGrn: We will soon have a problem with this test, since it will be reused everywhere. We need a way to have reusable scenarios.

nesrineabdmouleh: Yes, I think that the solution is to create a file named commonScenario contains all common tests used in all campaigns

boubkerbribri: Maybe regroup it in common_scenarios, but it most be done for all tests, so not in this PR I think

matthieu-rolland (approved):
lgtm

PR #15314, opened 2019-08-29
Fix not being able to remove automatically applied cart rule

### LINKED ISSUE #12608: Cannot delete cart rule with no code
**Describe the bug**
User cannot remove a cart rule with *no code* from his cart.

**To Reproduce**
1. Add a cart rule with no code
2. On the Front Office, visit the cart page and try to remove the cart rule from the cart.
3. Cart rules stays !

**Additionnal information**
Inside the templates, we can't see if a cart rule has a voucher code or not,
so we can't hide the delete button for this kind of cart rule.

damiandominella:
Avoid to show trash icon if cart rule is automatically applied (voucher code is empty)

| Questions     | Answers
| -------------| -------------------------------------------------------
| Branch?        | develop 
| Description? | Fix 
| Type?            | bug fix
| Category?     | FO
| BC breaks?   | no
| Deprecations? | no
| Fixed ticket? | Fixes #12608
| How to test?  | Create a cart rule with empty code (like described in the issue #12608)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15314)

### DIFF
--- a/themes/classic/templates/checkout/_partials/cart-voucher.tpl
+++ b/themes/classic/templates/checkout/_partials/cart-voucher.tpl
@@ -34,7 +34,9 @@
                   <span class="label">{$voucher.name}</span>
                   <div class="float-xs-right">
                     <span>{$voucher.reduction_formatted}</span>
-                    <a href="{$voucher.delete_url}" data-link-action="remove-voucher"><i class="material-icons">&#xE872;</i></a>
+                      {if !empty($voucher.code)}
+                        <a href="{$voucher.delete_url}" data-link-action="remove-voucher"><i class="material-icons">&#xE872;</i></a>
+                      {/if}
                   </div>
                 </li>
               {/foreach}
--- a/themes/classic/templates/checkout/_partials/cart-voucher.tpl
+++ b/themes/classic/templates/checkout/_partials/cart-voucher.tpl
@@ -34,7 +34,7 @@
                   <span class="label">{$voucher.name}</span>
                   <div class="float-xs-right">
                     <span>{$voucher.reduction_formatted}</span>
-                      {if !empty($voucher.code)}
+                      {if !empty($voucher.code) && $voucher.code != '0'}
                         <a href="{$voucher.delete_url}" data-link-action="remove-voucher"><i class="material-icons">&#xE872;</i></a>
                       {/if}
                   </div>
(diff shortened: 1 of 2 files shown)


matks:
Hi @damiandominella ! Thanks for the contribution ! Code looks OK ✅ but I am wondering: is there a possible usecase where the voucher code would be empty **AND** the cart rule would be deletable ? Are we sure **ALL** vouchers follow the rule "if there is no code, then the cart rule cannot be deleted" ?

damiandominella:
> Hi @damiandominella ! Thanks for the contribution ! Code looks OK ✅ but I am wondering: is there a possible usecase where the voucher code would be empty **AND** the cart rule would be deletable ? Are we sure **ALL** vouchers follow the rule "if there is no code, then the cart rule cannot be deleted" ?

I don't think so. If a voucher code is empty, it means that customers can't add it manually, so they don't have to be able to delete it. 
The logic is: empty voucher code ---> automatically applied cart rule (not deletable)

PierreRambaud:
I think we need to check for a no null value and empty string. `0` can be a voucher value. Even if it's stupid it's a possible value.

damiandominella:
> I think we need to check for a no null value and empty string. `0` can be a voucher value. Even if it's stupid it's a possible value.

The empty() function already checks for null and empty strings values. The only case of possible error is when the value is "0". I just added the control

damiandominella:
> I think we need to check for a no null value and empty string. `0` can be a voucher value. Even if it's stupid it's a possible value.

The empty() function already checks for null and empty string. The only case of possible error is when the value is "0". I just added the control

themes/classic/templates/checkout/_partials/cart-voucher.tpl
```diff
@@ -34,7 +34,9 @@
                   <span class="label">{$voucher.name}</span>
                   <div class="float-xs-right">
                     <span>{$voucher.reduction_formatted}</span>
-                    <a href="{$voucher.delete_url}" data-link-action="remove-voucher"><i class="material-icons">&#xE872;</i></a>
+                      {if !empty($voucher.code) || $voucher.code == '0'}
```
PierreRambaud: Still wrong because ` false == '0'
```suggestion
                      {if isset($voucher.code) && $voucher.code !== ''}
```

PierreRambaud (approved):

PierreRambaud:
Thanks @damiandominella

PR #15323, opened 2019-08-30
Update Cart.php

psandmore:
Default value for $id_product_attribute in deleteProduct method was changed from null to 0. Strict comparison $id_product_attribute !== null is not working with zero. Changed comparison with null to (int).

| Questions     | Answers
| Branch?       |  develop
| Description?  | any configuration
| Type?         | bug fix
| Category?     | FO

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15323)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1712,7 +1712,7 @@ class CartCore extends ObjectModel
                 WHERE `id_cart` = ' . (int) $this->id . '
                 AND `id_product` = ' . (int) $id_product . '
                 AND `id_customization` = ' . (int) $id_customization .
-                (int) $id_product_attribute ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
+                ((int) $id_product_attribute ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
             );
         }
 
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1712,7 +1712,7 @@ class CartCore extends ObjectModel
                 WHERE `id_cart` = ' . (int) $this->id . '
                 AND `id_product` = ' . (int) $id_product . '
                 AND `id_customization` = ' . (int) $id_customization .
-                (Int) $id_product_attribute ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
+                (int) $id_product_attribute ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
             );
         }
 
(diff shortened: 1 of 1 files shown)


classes/Cart.php
```diff
@@ -1712,7 +1712,7 @@ public function deleteProduct(
                 WHERE `id_cart` = ' . (int) $this->id . '
                 AND `id_product` = ' . (int) $id_product . '
                 AND `id_customization` = ' . (int) $id_customization .
-                ($id_product_attribute != null ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
+                (Int) $id_product_attribute ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
```
matthieu-rolland: thank you for your contribution @psandmore 

Could you please put the I of your "Int" in minuscule? 
There might also be unit tests to change due to your modification.

 thanx !

matthieu-rolland (changes requested):

classes/Cart.php
psandmore: updated as requested

classes/Cart.php
```diff
@@ -1712,7 +1712,7 @@ public function deleteProduct(
                 WHERE `id_cart` = ' . (int) $this->id . '
                 AND `id_product` = ' . (int) $id_product . '
                 AND `id_customization` = ' . (int) $id_customization .
-                ($id_product_attribute != null ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
+                (int) $id_product_attribute ? ' AND `id_product_attribute` = ' . (int) $id_product_attribute : '')
```
matthieu-rolland: there's a syntax error here, the last parenthesis line 1715 shouldn't be there maybe?
Dit it work for you?

matthieu-rolland (changes requested):
fix syntax error

PierreRambaud:
Hi and thanks for your contribution.
Can you please create an issue to explain what exactly happened et how we can reproduce.
Can you also complete the pull request description and title?

Kind regards

psandmore:
I do not know how to reproduce. I saw in the code that the signature of the deleteProduct method changed. The $id_product_attribute default value was previously null and now it is zero. The original code used a strict comparison $id_product_attribute !== null. When the default value was null this was correct. Now the default value is zero and it is not working with zero. So I made a fix.

matks:
Hi, [following our review guidelines](https://devdocs.prestashop.com/1.7/project/maintainers-guide/reviewing-pull-requests/) "Pull Requests may be closed after 30 days of inactivity following a request for modifications".

Since we did not get answer from author I close this PR.

Unfortunately, without a scenario "How to reproduce" we cannot test and validate this PR 😢 

Thank you for the contribution

PR #15341, opened 2019-08-31
BO: Fix admin controller's render list identifier notice issue

### LINKED ISSUE #15333: Getting issue on module admin controller's render list in Prestashop V1.7.6.1 with existing data
Hello,

Currently I am getting issue on my module's admin controller render list page in Prestashop V1.7.6.1 while my controller's render list has some data.
After research, I found that I am getting this issue because I am not using $this->identifier name similar to $this->table name on module's admin controller page.
For example, my table name is 'product_review' and here I am using primary key as 'id_review' So in this case, I will define $this->identifier = 'id_review' on module admin controller.

In previous prestashop version (below V1.7.6.1), I don't have any issue like this but on current prestashop V1.7.6.1, I am getting a error -
"Notice: Undefined index: id_product_review"

Check this screenshot -  https://prnt.sc/ozmiwc

In Prestashop V1.7.6.1, you are always getting entity like -   'id_' . $table_name
So have you restrict it in current version 1.7.6.1 because you didn't mention anything like this before.

1. Create a table in module i.e. table name 'product_review' and set primary key in this table as 'id_review'
2. Create a module's admin controller and set 
   $this->identifier = 'id_review'
   $this->table = 'product_review'

Now while you will add any data in your table and when you see render list of your admin controller then you will get issue like "Notice: Undefined index: id_product_review"
https://prnt.sc/ozmiwc

Here if I used my primary key as 'id_product_review' then I will get no error. So It looks like It is mandatory to set table primary key same table name.

Screenshots -

error

PrestaShop version: 1.7.6.1

neerajwebkul:
| Questions     | Answers
| Branch?       | develop
| Description?  | Getting notice on admin controller's render list in with edit/view when URL have identifier, when debug mode enabled.
| Type?         | bug fix 
| Category?     |BO
| BC breaks?    | Does it break backward compatibility? yes
| Fixed ticket? | Fixes #15333
| How to test?  | Check the issue page. its described there.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15341)

### DIFF
--- a/classes/helper/HelperList.php
+++ b/classes/helper/HelperList.php
@@ -875,6 +875,7 @@ class HelperListCore extends Helper
     {
         $parameters = [
             $this->identifier => $id,
+            'id_'.$this->table => $id,
             'current_index' => $this->currentIndex,
             'token' => $token != null ? $token : $this->token,
         ];


jolelievre:
Hello @neerajwebkul 

I'm sorry I didn't see your PR before I proposed mine here https://github.com/PrestaShop/PrestaShop/pull/15382

Thanks for your fix, but as I introduced this new class `LegacyHelperLinkBuilder` I thought of a more proper way to fix the bug. Although your fix totally works it fixes the symptom and not the reason of the bug (we shouldn't try to access this field in the first place)

Feel free to comment here or on my PR if you think your approach is better ;)

levyn (approved):

marionf:
@jolelievre Can we close this PR as the issue #15333 is fixed by #15382 ?

matks:
@neerajwebkul thank you or your contribution ! unfortunately issue got fixed by https://github.com/PrestaShop/PrestaShop/pull/15382
so we will not merge this one

PR #15352, opened 2019-09-01
Fix quicksearch results missing the category name

### LINKED ISSUE #9767: [BOOM-4639] Quicksearch dropdown is missing the category name
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-4639).

- _**Reporter:**_ maluma
- _**Created at:**_ Tue, 16 Jan 2018 16:55:04 +0100

<p>When using the quick search bar at the top in the frontoffice, the dropdown would usually have results in the following format: <em><span class="error">&#91;category-name&#93;</span></em> <em>> <span class="error">&#91;product-name&#93;</span></em></p>

<p>Now, the category name is missing so only the arrow and the product name are shown. This is caused by the missing property product.category_name from the result of the ajax call to /index.php?controller=search</p>

<p>In presta_root/modules/ps_searchbar/ps_searchbar.js the field is still tried to be processed so I guess it should be added back to the controller result.</p>

<p>There also used to be a lot more fields like product.cover which would be nice if they got readded because I would like to use them.</p>

<p>I updated to 1.7.2.4 from (I think) 1.7.2.1 so I'm not sure in which version the field disappeared.</p>

<p> </p>
        
&nbsp;* Install a fresh version 1.7.2.4
<br/>
&nbsp;* Go to the frontoffice homepage
<br/>
&nbsp;* Type &quot;dress&quot; into the searchbar in the top right corner
<br/>
&nbsp;* You only see a list of &gt; product an no category names

JulianEberius:
| Questions     | Answers
| Branch?       | develop
| Description?  | This should fix https://github.com/PrestaShop/PrestaShop/issues/9767 . The reason the search box does not work correctly (even in the live demo at http://demo.prestashop.com !) is that the category_name attribute is filtered in ProductListingFrontController using the SearchResultProductFilter. 
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9767 
| How to test?  | Use the default search box, see the categories are missing (all items starting with ">"). Apply fix, they should appear again.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15352)

### DIFF
--- a/src/Core/Filter/FrontEndObject/SearchResultProductFilter.php
+++ b/src/Core/Filter/FrontEndObject/SearchResultProductFilter.php
@@ -40,6 +40,7 @@ class SearchResultProductFilter extends HashMapWhitelistFilter
             'active',
             'add_to_cart_url',
             'canonical_url',
+            'category_name',
             'cover',
             'description_short',
             'discount_amount',


matthieu-rolland (approved):
Thank you @JulianEberius !

 From what I've checked in the code, your modification looks good to me.

PierreRambaud (approved):

matks:
Thank you @JulianEberius

PR #15355, opened 2019-09-02
Adding sanity tests products BO

boubkerbribri:
| Questions     | Answers
| Branch?       | develop
| Description?  | Adding sanity tests products BO
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ,o
| How to test?  | TEST_PATH="sanity/products/*" URL_FO=shop_url npm run specific-test

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15355)


tests/puppeteer/campaigns/sanity/products/02_CRUD_simple_product_in_BO
```diff
@@ -0,0 +1,83 @@
+// Using chai
```
SimonGrn: Aren't you missing the extension for this file ?

tests/puppeteer/campaigns/data/faker/product.js
```diff
@@ -0,0 +1,15 @@
+let faker = require('faker');
+
+module.exports = class Product {
+
+  constructor(productType = 'Simple product', productHasCombinations = false, wantedQuantity = '1') {
+    this.name = faker.commerce.productName().toUpperCase();
+    this.description = faker.lorem.sentence();
+    this.reference  = faker.random.alphaNumeric(7);
+    this.quantity  = faker.random.number({'min': 1, 'max': 9}).toString();
```
SimonGrn: We might want to pass this as a parameter ? For example, with `quantity`as a parameter:
```suggestion
    this.quantity  = quantity || faker.random.number({'min': 1, 'max': 9}).toString();
```

tests/puppeteer/campaigns/data/faker/product.js
```diff
@@ -0,0 +1,15 @@
...
+    this.quantity_wanted = wantedQuantity;
+    this.price  = faker.random.number({'min': 10, 'max': 20}).toString();
```
SimonGrn: Same here, might want to pass this as a parameter:
```suggestion
    this.price  = price || faker.random.number({'min': 10, 'max': 20}).toString();
```

tests/puppeteer/campaigns/sanity/products/01_Filter_products.js
```diff
@@ -0,0 +1,78 @@
...
+const {expect} = require('chai');
+
+// importing pages
+const BO_LOGIN_PAGE = require('../../../pages/BO/login');
+const BO_DASHBOARD_PAGE = require('../../../pages/BO/dashboard');
+const BOBasePage = require('../../../pages/BO/BObasePage');
+const ProductPage = require('../../../pages/BO/product');
+
+let page;
+let BO_LOGIN;
```
SimonGrn: Is there a reason for using all caps ?

tests/puppeteer/campaigns/sanity/products/01_Filter_products.js
```diff
@@ -0,0 +1,78 @@
...
+let BO_DASHBOARD;
+let boBasePage;
+let productPage;
+let numberOfProducts;
+
+// creating pages objects in a function
+const init = async () => {
+  page = await global.browser.newPage();
+  BO_LOGIN = await (new BO_LOGIN_PAGE(page));
```
SimonGrn: Same here, is it normal to not have the same caps ?

tests/puppeteer/campaigns/sanity/products/02_CRUD_simple_product_in_BO
```diff
@@ -0,0 +1,83 @@
...
+  productPage = await (new ProductPage(page));
+  addProductPage = await (new AddProductPage(page));
+  foProductPage = await (new FOProductPage(page));
+  productData = await (new ProductFaker());
+  editedProductData = await (new ProductFaker());
+};
+
+
+// Create, read, update and delete simple product in BO
+global.scenario('Filter in Products Page', async () => {
```
SimonGrn:
```suggestion
global.scenario('Create, read, update and delete a product', async () => {
```

tests/puppeteer/campaigns/utils/helpers.js
```diff
@@ -12,8 +12,8 @@ global.scenario = (name, tests, init, close = false) => describe(name, async ()
       slowMo: 25,
       args: ['--start-maximized', '--no-sandbox', '--lang=fr-FR'],
       defaultViewport: {
-        width: 1270,
-        height: 899,
+        width: 1680,
+        height: 900,
```
SimonGrn, line 16: 👍

tests/puppeteer/pages/BO/addProduct
```diff
@@ -0,0 +1,91 @@
+const BOBasePage = require('../BO/BObasePage');
```
SimonGrn: Pretty sure you forgot to add the `.js` extension in this one too 😝

tests/puppeteer/pages/BO/product.js
```diff
@@ -1,13 +1,116 @@
...
    */
+
+  /**
+   * Filter products from inputs : Name, reference and Category
+   * @param filterBy
+   * @param value
+   * @return {Promise<void>}
+   */
+  async filterProducts(filterBy, value = '') {
+    await this.page.type(this.productFilterInput.replace('%FilterBy', filterBy), value);
```
SimonGrn:
```suggestion
    await this.page.type(this.productFilterInput.replace('%FILTERBY', filterBy), value);
```
I personally find all caps parameters to be more visible (and IIRC @nesrineabdmouleh used an all caps one ?)
What do you think ?

tests/puppeteer/pages/BO/product.js
```diff
@@ -1,13 +1,116 @@
...
+      this.page.click(this.filterSearchButton),
+    ]);
+  }
+
+  /**
+   * Get number of products displayed in list
+   * @return integer
+   */
+  async getNumberOfProductsFromList() {
+    return (await this.page.$$(this.productRow)).length;
```
SimonGrn: Isn't this limited by pagination ? Maybe we can get the total number from the page itself (IIRC it's written somewhere) ?

tests/puppeteer/pages/commonPage.js
```diff
@@ -116,4 +116,19 @@ module.exports = class CommonPage {
   async reloadPage() {
     await this.page.reload({waitUntil: 'networkidle0'});
   }
+
+  /**
+   * Close actual tab and goto another tab if wanted
+   * @param tabId
+   * @return {Promise<void>}
+   */
+  async closePage(tabId = -1) {
```
SimonGrn, line 151: 👍

SimonGrn (commented):
Fantastic work, thanks a lot @boubkerbribri

tests/puppeteer/campaigns/sanity/products/02_CRUD_simple_product_in_BO
boubkerbribri: done :+1:

tests/puppeteer/pages/BO/product.js
boubkerbribri: yes, it's more visible.
done, thanks :+1:

tests/puppeteer/pages/BO/addProduct
boubkerbribri: yes, I'm sure too :smile: 
done thanks

tests/puppeteer/pages/BO/product.js
boubkerbribri: we have 19 demo products and 20 items/page showed so if we destroyed each product created, we will not have to navigate between pages

SimonGrn: At the moment, yes. But I prefer we not rely on the fixtures as there are today because we don't know how they might change. If the dev team modify the fixtures to add 25 demo products, this will fail.

tests/puppeteer/campaigns/sanity/products/01_Filter_products.js
boubkerbribri: done thanks

tests/puppeteer/campaigns/sanity/products/01_Filter_products.js
boubkerbribri: done thanks

tests/puppeteer/campaigns/data/faker/product.js
boubkerbribri: done, thanks

tests/puppeteer/campaigns/data/faker/product.js
boubkerbribri: Done, thanks

tests/puppeteer/pages/BO/product.js
boubkerbribri: The only way I see to do it is this 
```js
    async getNumberOfProductsFromList() {
    const numberOfProducts = (await this.page.$$(this.productRow)).length;
    const found = await this.elementVisible(this.paginationNextLink, 1000);
    if (!found) return numberOfProducts;

    await Promise.all([
      this.page.waitForNavigation({waitUntil: 'networkidle0'}),
      this.page.click(this.paginationNextLink),
    ]);
    return numberOfProducts + (await this.getNumberOfProductsFromList());
  }
```
What do you think ?

SimonGrn: The thing is, if there's more than 20 articles, you'll see this block at the bottom : 
image
It gives you the number of items. Maybe we can try looking for it to see if there's more than 20 items ?

boubkerbribri: that's why I click paginationNextLink (if exists).
This recursive function will return the real number of products from all list of products pages.

SimonGrn: Your method will crawl every page, and might take a while if there's a lot of pages. 
- if there's not a pagination block, we count the number of rows
- if there's a pagination block, we extract the number of lines from the text inside with a regexp (like `/out of ([0-9]+)/` )
Both methods don't require any navigation.

What do you think ?

boubkerbribri: It's better to not have navigation for sure.
It would be better if we didn't have to do regex to get it.
I changed it like suggested with regex, and looking forward to get it with a better selector.

tests/puppeteer/pages/BO/BObasePage.js
```diff
@@ -95,4 +101,15 @@ module.exports = class BOBasePage extends CommonPage {
...
     return FOPage;
   }
+
+  /**
+   * Set value on tinyMce textareas
+   * @param iFrameSelector
+   * @param value
+   * @return {Promise<void>}
+   */
+  async setValueOnTinymceInput(iFrameSelector, value) {
```
nesrineabdmouleh, line 111: We can use this function to set a value on a simple input not empty, so in this case the function name must be generic.
What do you think?

boubkerbribri: A simple `page.type(selector,value)` or `setValue` from your #15312  could do that, and no need for `page.keyboard.type`.
Thing is, I want a specific function that set value on this type of texteareas to make clear (for everyone who want to use it) to pass the iFrame selector as a param (and not input selector in the iFrame).
Also, the function could be improved (and not be used for all inputs)

tests/puppeteer/campaigns/data/faker/product.js
```diff
@@ -0,0 +1,19 @@
+const faker = require('faker');
+
+module.exports = class Product {
+  constructor(productType = 'Standard product', quantity = false, wantedQuantity = '1', price = false,
+    productHasCombinations = false, combinations = false) {
+    this.name = faker.commerce.productName().toUpperCase();
```
SimonGrn, line 5: faker is super clean I love it ❤️

tests/puppeteer/campaigns/sanity/productsBO/03_CRUDStandardProductWithCombinationsInBO.js
```diff
@@ -0,0 +1,84 @@
...
+// creating pages objects in a function
+const init = async () => {
+  page = await global.browser.newPage();
+  loginPage = await (new LoginPage(page));
+  dashboardPage = await (new DashboardPage(page));
+  boBasePage = await (new BOBasePage(page));
+  productPage = await (new ProductPage(page));
+  addProductPage = await (new AddProductPage(page));
+  foProductPage = await (new FOProductPage(page));
+  productWithCombinations = await (new ProductFaker('Standard product', false, '1', false, true));
```
SimonGrn: It's not very easy to know what are all these arguments:
```false, '1', false, true```
means we need to go to the function declaration to know what is what. Maybe you could create an object to pass instead of a few parameters ? Also it's dangerous if we want to change these in the future (ex: change parameter order, remove one, etc).
What do you think ?
```suggestion
  let productToCreate = {
  name : 'Standard product',
  quantity: false,
  wantedQuantity: '1',
  price: false,
  productHasCombinations : true
  }
  productWithCombinations = await (new ProductFaker(productToCreate));
```

tests/puppeteer/pages/BO/addProduct.js
```diff
@@ -0,0 +1,191 @@
...
+   * @return {Promise<void>}
+   */
+  async setCombinationsInProduct(productData) {
+    // GOTO Combination tab : id = 3
+    await this.goToFormStep(3);
+    // Delete All combinations if exists
+    await this.deleteAllCombinations();
+    // set Combinations
+    const keys = Object.keys(productData.combinations);
+    /*eslint-disable*/
```
SimonGrn, line 97: 👍

tests/puppeteer/pages/BO/product.js
```diff
@@ -1,13 +1,126 @@
+// Using chai
```
SimonGrn: The name of the file is `product.js` and it's for the `Products` listing page which means we might have a conflict in naming when we'll want to create a file for a single product... what do you think ?

SimonGrn (commented):
Thanks @boubkerbribri

tests/puppeteer/pages/BO/addProduct.js
boubkerbribri: yeah eslint don't like awaits in loop, and we need actions to be done sequentially (no way to use `Array.ForEach()`)

tests/puppeteer/pages/BO/product.js
boubkerbribri: done thanks

tests/puppeteer/campaigns/sanity/productsBO/03_CRUDStandardProductWithCombinationsInBO.js
boubkerbribri: I'm thinking more this :
for Simple product
```js 
  let productToCreate = {
    type : 'Standard product',
    productHasCombinations : false,
  };
  productData = await (new ProductFaker(productToCreate));
  editedProductData = await (new ProductFaker(productToCreate));
```
And for product with combinations : 
```js 
  let productToCreate = {
    type : 'Standard product',
    productHasCombinations : true,
  };
  productData = await (new ProductFaker(productToCreate));
  editedProductData = await (new ProductFaker(productToCreate));
```
more simple to add only attributes needed in `productToCreate`

boubkerbribri: done thanks

matthieu-rolland (approved):
lgtm

PR #15360, opened 2019-09-02
[WIP] Localize carrier name

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add carrier_lang name column in db structure, update Carrier object model, to use localized name. Fix possible array to string conversion related to this change.
| Type?         | improvement
| Category?     | FO / BO
| BC breaks?    | Yes. Carrier->name becomes array instead of string.
| Deprecations? | yes. `PREFIX_carrier.name` deprecates after this pr. use `PREFIX_carrier_lang.name` instead.
| Fixed ticket? | #12901
| How to test?  | 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15360)

### DIFF
--- a/admin-dev/themes/default/template/controllers/orders/_new_product.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/_new_product.tpl
@@ -107,7 +107,7 @@
 			<div class="form-group">
 				<label class="control-label col-lg-3">{l s='Carrier' d='Admin.Shipping.Feature'}</label>
 				<div class="col-lg-9">
-					<p class="form-control-static"><strong>{$carrier->name}</strong></p>
+					<p class="form-control-static"><strong>{$carrier->name[$current_id_lang]}</strong></p>
 				</div>
 			</div>
 			<div class="form-group">
--- a/classes/Carrier.php
+++ b/classes/Carrier.php
@@ -54,7 +54,7 @@ class CarrierCore extends ObjectModel
     /** @var int common id for carrier historization */
     public $id_reference;
 
-    /** @var string Name */
+    /** @var array of localized names where key represents language id */
     public $name;
 
     /** @var string URL with a '@' for */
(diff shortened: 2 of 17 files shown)


sarjon:
@matks take a look at this PR, it could potentially break a lot of shipping (and other) modules.

matks:
> @matks take a look at this PR, it could potentially break a lot of shipping (and other) modules.

Indeed. I thought about the upgrade issue but not about this one.

Mmmm 😥this is a hard choice, because the breaking change can be huge but the benefits can be huge too.

@colinegin @samuel-pires do you want to reconsider the feature ?
If we make the carrier name translatable, obviously **any module that modifies carrier names needs to upgrade their logic as you cannot handle a text field and a translatable text field in the same way.

zuk3975:
> Mmmm :disappointed_relieved: this is a hard choice, because the breaking change can be huge but the benefits can be huge too.

Just to add on... This PR is fresh, but you can already see that the changes splits all over ps including both BO/FO - order, price rule, product pages... :smile_cat:

ghost:
Hi guys !

Hope you are doing well.

To me there are 2 things: 
1. the impact on shipping modules 
2. the impact on PS

Regarding 1. 
Shipping modules are key for merchants so if we have such a breaking change we should take time to inform modules sellers long enough in advance so they can be be ready. And when we see what happenned to payment modules... it should be really long enough in advance😉 

Regarding 2.
Are the breaks easy to anticipate/localize in the code and to fix?
Is it better to do this change now or when more pages will be migrated? is there a difference?

This is a really interesting feature that could save merchants a lot of times but considering 1. and depending on your answer to 2. I think we should finish this PR later (I think 1.7.8 and we start notifying modules sellers now about it). Wdyt @colinegin @marionf?

zuk3975:
@samuel-pires, @colinegin, @marionf, just want to make sure You have decided already, so I could modify related PR's (like #14551). 
So the name stays NOT translatable for now ?

ghost:
Hi @zuk3975, 
yes, the name stays NOT translatable for now.

Thanks

PR #15364, opened 2019-09-02
[WIP] Fixed the displayed discount when adding product (linked to a coupon) to an existing order in BO

### LINKED ISSUE #13880: Bug in displayed discount when adding product (linked to a coupon) to an existing order from BO
**Describe the bug**

**To Reproduce**

**Screenshots**
1!UNITO-UNDERSCORE!part1
1!UNITO-UNDERSCORE!part2
1!UNITO-UNDERSCORE!part3
2!UNITO-UNDERSCORE!part1
2!UNITO-UNDERSCORE!part2

**Additionnal information**
PrestaShop version: 1.7.5.0
PHP version: 7.2

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | When a coupon was linked to a product, and when we add this product to an existing order, the discount calculation was bad.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #13880 
| How to test?  | [Test Scenario from @khouloudbelguith](https://github.com/PrestaShop/PrestaShop/issues/13880#issuecomment-494353896)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15364)

### DIFF
--- a/controllers/admin/AdminOrdersController.php
+++ b/controllers/admin/AdminOrdersController.php
@@ -2078,7 +2078,15 @@ class AdminOrdersControllerCore extends AdminController
         }
     }
 
-    public function ajaxProcessAddProductOnOrder()
+    /**
+     * Add product on an order
+     *
+     * @throws PrestaShopDatabaseException
+     * @throws PrestaShopException
+     * @throws SmartyException
+     * @throws \PrestaShop\PrestaShop\Core\Localization\Exception\LocalizationException
+     */
+    public function displayAjaxAddProductOnOrder()
     {
         // Load object
         $order = new Order((int) Tools::getValue('id_order'));
@@ -2093,7 +2101,9 @@ class AdminOrdersControllerCore extends AdminController
         $invoice_informations = $_POST['add_invoice'] ?? [];
 
         $result = $this->addProductToOrder($order, $product_informations, $invoice_informations, Tools::getValue('add_product_warehouse'));
-        die(json_encode(array(
+
+        header('Content-Type: application/json');
+        $this->ajaxRender(json_encode([
             'result' => true,
             'view' => $this->createTemplate('_product_line.tpl')->fetch(),
             'can_edit' => $this->access('add'),
@@ -2103,7 +2113,7 @@ class AdminOrdersControllerCore extends AdminController
             'shipping_html' => $this->createTemplate('_shipping.tpl')->fetch(),
             'discount_form_html' => $this->createTemplate('_discount_form.tpl')->fetch(),
             'refresh' => $result['refresh'],
-        )));
+        ]));
     }
 
     /**
--- a/controllers/admin/AdminOrdersController.php
+++ b/controllers/admin/AdminOrdersController.php
@@ -2080,11 +2080,6 @@ class AdminOrdersControllerCore extends AdminController
 
     /**
      * Add product on an order
-     *
-     * @throws PrestaShopDatabaseException
-     * @throws PrestaShopException
-     * @throws SmartyException
-     * @throws \PrestaShop\PrestaShop\Core\Localization\Exception\LocalizationException
      */
     public function displayAjaxAddProductOnOrder()
     {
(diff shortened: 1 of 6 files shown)


tests/Integration/Behaviour/Features/Context/OrderFeatureContext.php
```diff
@@ -82,6 +116,382 @@ public function validateCartWithPaymentModule($paymentModuleName)
...
+     * Duplicate from AdminOrderController::addProductToOrder
+     *
+     * @param Order $order
+     * @param array $product_informations
+     * @param array $invoice_informations
+     * @param bool $warehouseId
+     *
+     * @return array
+     */
+    protected function addProductToOrder(Order $order, array $product_informations, array $invoice_informations = [], $warehouseId = false)
```
PierreRambaud: What a huge function, any possibility to split it?

Progi1984: Actually it's a duplicate from the method in controllers/admin/AdminOrdersController.php but as this method is protected, we can't use it. So @matks has adviced me to duplicate it.

controllers/admin/AdminOrdersController.php
```diff
@@ -2088,21 +2088,45 @@ public function ajaxProcessAddProductOnOrder()
             )));
         }
 
-        $old_cart_rules = Context::getContext()->cart->getCartRules();
+        $product_informations = $_POST['add_product'];
+        $invoice_informations = $_POST['add_invoice'] ?? [];
+
+        $result = $this->addProductToOrder($order, $product_informations, $invoice_informations, Tools::getValue('add_product_warehouse'));
+        die(json_encode(array(
```
jolelievre: Could you use `ajaxRender` instead please?

tests/Integration/Behaviour/Features/Context/OrderFeatureContext.php
jolelievre: Mmmm, the problem is that if we need to fix this again we will have to maintain both the controller AND the behat context.. So it would be better if this can be extracted in a separate class/service

jolelievre (changes requested):
I'm really worried about this huge copied function, this needs to be extracted in a dedicated class/service if we want to be able to maintain it in the future. Even though it's dirty for now at least having a class dedicated to this will allow us to improve/refacto it more easily in the future (hopefully ^^)
What do you think?

jolelievre:
@Progi1984 did you forget this one? Or it was swallowed by blockreassurance? :P

controllers/admin/AdminOrdersController.php
Progi1984: :ok_hand:

controllers/admin/AdminOrdersController.php
```diff
@@ -2078,7 +2078,15 @@ public function ajaxProcessSendMailValidateOrder()
         }
     }
 
-    public function ajaxProcessAddProductOnOrder()
+    /**
+     * Add product on an order
+     *
+     * @throws PrestaShopDatabaseException
```
PierreRambaud: This method doesn't throws anything :thinking: 
If you want to add all possibilities, it will be really really hard to maintain

controllers/admin/AdminOrdersController.php
```diff
@@ -2089,21 +2097,47 @@ public function ajaxProcessAddProductOnOrder()
...
+
+    /**
+     * @param Order $order
+     * @param array $product_informations
+     * @param array $invoice_informations
+     * @param bool $warehouseId
+     *
+     * @return array
+     *
+     * @throws PrestaShopDatabaseException
```
PierreRambaud: same here

controllers/admin/AdminOrdersController.php
Progi1984: Fixed

controllers/admin/AdminOrdersController.php
Progi1984: Fixed

matks:
@Progi1984 still in WIP or shall we review again ?

Progi1984:
@matks Always in WIP. I wait for Invertus merge its implementation of adding Product in Order for checking if the problem is always here.

colinegin:
@Progi1984 i think you can work again on this as all PRs for the order page have been merged !

PR #15382, opened 2019-09-03
Remove useless and buggy identifier computation

### LINKED ISSUE #15318: Notice when configure some modules
**Describe the bug**

**To Reproduce**
1. Go to module
2. Click on configure ps_emailsubscription
3. See error

**Screenshots**

image

**Additional information**
PrestaShop version: 1.7.6.1
PHP version: 7.1

### LINKED ISSUE #15333: Getting issue on module admin controller's render list in Prestashop V1.7.6.1 with existing data

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | The new `LegacyHelperLinkBuilder` (introduced in https://github.com/PrestaShop/PrestaShop/pull/14721) causes an error warning when the expected parameter was not present, besides this computation was useless as it expected the parameter to be present
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15333 and fixes #15318
| How to test?  | You can check the fix by using the product comments module or ps_emailsubscription module (known bugs, but probably other modules had troubles). You also need to check that the quick search is still valid

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15382)

### DIFF
--- a/src/Adapter/Routing/AdminLinkBuilder.php
+++ b/src/Adapter/Routing/AdminLinkBuilder.php
@@ -90,12 +90,13 @@ class AdminLinkBuilder implements EntityLinkBuilderInterface
      */
     private function buildActionParameters($action, $entity, array $parameters)
     {
+        unset($parameters['current_index']);
+        unset($parameters['token']);
         $editAction = $action . $entity;
-        $entityId = 'id_' . $entity;
 
         return array_merge(
             $parameters,
-            [$entityId => $parameters[$entityId], $editAction => 1]
+            [$editAction => 1]
         );
     }
 
--- a/src/Adapter/Routing/LegacyHelperLinkBuilder.php
+++ b/src/Adapter/Routing/LegacyHelperLinkBuilder.php
@@ -26,6 +26,7 @@
 
 namespace PrestaShop\PrestaShop\Adapter\Routing;
 
+use PrestaShop\PrestaShop\Core\Exception\InvalidArgumentException;
 use PrestaShop\PrestaShop\Core\Routing\EntityLinkBuilderInterface;
 
 /**
@@ -36,10 +37,19 @@ use PrestaShop\PrestaShop\Core\Routing\EntityLinkBuilderInterface;
 class LegacyHelperLinkBuilder implements EntityLinkBuilderInterface
 {
     /**
-     * {@inheritdoc}
+     * @param string $entity
+     * @param array $parameters
+     *
+     * @return string
+     *
+     * @throws InvalidArgumentException
      */
     public function getViewLink($entity, array $parameters)
     {
+        if (!isset($parameters['current_index'])) {
+            throw new InvalidArgumentException('Missing parameter current_index to build legacy link');
+        }
+
         $currentIndex = $parameters['current_index'];
         $parameters = $this->buildActionParameters('view', $entity, $parameters);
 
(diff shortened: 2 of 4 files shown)


src/Adapter/Routing/AdminLinkBuilder.php
```diff
@@ -90,12 +90,13 @@ public function getEditLink($entity, array $parameters)
      */
     private function buildActionParameters($action, $entity, array $parameters)
     {
+        unset($parameters['current_index']);
```
PierreRambaud, line 93: always check if key exist before unset.

PierreRambaud: At least for token

MathiasReker:
Milestone should be 1.7.6.2 instead of 1.7.6.1?

src/Adapter/Routing/AdminLinkBuilder.php
jolelievre: I think `unset` doesn't care if the key is present, if it's not here it simply does nothing
But at least the code is simpler

PierreRambaud: Oh you're right :+1:

jolelievre: Well, then is it approved? ^^

PierreRambaud (approved):

matks:
Thank you @jolelievre

Mohammed1990maroc:
It's a minor problem of PrestaShop 1.7.6.1, to fix it you need to modify the /src/Adapter/Routing/LegacyHelperLinkBuilder.php file, change the line

[$entityId => $parameters[$entityId], $viewAction => 1]
to
[$entityId => isset($parameters[$entityId])?:null, $viewAction => 1]

jolelievre:
Hi @Mohammed1990maroc 

actually this is not even necessary (see this PR content for the fix). Since anyway this array is merged with `$parameters` there's no need to check for one of its field and merge it back again, this is redundant, useless and cause bugs, ...

This bug fix has been merged in the 1.7.6.2 version which was released yesterday 😉

Mohammed1990maroc:
hit @jolelievre 
prestashop has fix this bug in the 1.7.6.2 ?
This version has not yet been released

jolelievre:
Yes it's included in the 1.7.6.2 version which was released yesterday
https://github.com/PrestaShop/PrestaShop/releases/tag/1.7.6.2

PR #15417, opened 2019-09-05
Order payment duplicated depending on order state configuration

### LINKED ISSUE #12588: Order payment duplicated depending on order state configuration
**Describe the bug**
If a state has these 3 options enabled, when you call validateOrder with this state the order payment is duplicated:

- Consider the associated order as validated.
- Set the order as paid.
- Show delivery PDF.

**To Reproduce**
1. Configure "Awaiting bank wire payment" order state with the 3 options enabled.
2. Create an order paid with bankwire
3. Check order payments

It doesn't happen when you change an order into an state with these 3 options, it only happens when it's called from ValdiateOrder.

**Screenshots**
image
image

**Additionnal information**
PrestaShop version: 1.7.5.0
PHP version: N/A

Prestaworks:
Only add a new payment if it is a real invoice and missing payment info.

| Questions     | Answers
| Branch?       | develop
| Description?  | Bug fix for #12588
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? |Fixes #12588
| How to test?  | See details under the ticket.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15322)

### DIFF
--- a/classes/order/OrderHistory.php
+++ b/classes/order/OrderHistory.php
@@ -372,34 +372,35 @@ class OrderHistoryCore extends ObjectModel
 
             foreach ($invoices as $invoice) {
                 /** @var OrderInvoice $invoice */
-                if ((int) $invoice->number > 0) {
-                    $rest_paid = $invoice->getRestPaid();
-                    if ($rest_paid > 0) {
-                        $payment = new OrderPayment();
-                        $payment->order_reference = Tools::substr($order->reference, 0, 9);
-                        $payment->id_currency = $order->id_currency;
-                        $payment->amount = $rest_paid;
+                if (0 == (int) $invoice->number) {
+                    continue;
+                }
+                $rest_paid = $invoice->getRestPaid();
+                if ($rest_paid > 0) {
+                    $payment = new OrderPayment();
+                    $payment->order_reference = Tools::substr($order->reference, 0, 9);
+                    $payment->id_currency = $order->id_currency;
+                    $payment->amount = $rest_paid;
 
-                        if ($order->total_paid != 0) {
-                            $payment->payment_method = $payment_method->displayName;
-                        } else {
-                            $payment->payment_method = null;
-                        }
-
-                        // Update total_paid_real value for backward compatibility reasons
-                        if ($payment->id_currency == $order->id_currency) {
-                            $order->total_paid_real += $payment->amount;
-                        } else {
-                            $order->total_paid_real += Tools::ps_round(Tools::convertPrice($payment->amount, $payment->id_currency, false), 2);
-                        }
-                        $order->save();
-
-                        $payment->conversion_rate = ($order ? $order->conversion_rate : 1);
-                        $payment->save();
-                        Db::getInstance()->execute('
-                        INSERT INTO `' . _DB_PREFIX_ . 'order_invoice_payment` (`id_order_invoice`, `id_order_payment`, `id_order`)
-                        VALUES(' . (int) $invoice->id . ', ' . (int) $payment->id . ', ' . (int) $order->id . ')');
+                    if ($order->total_paid != 0) {
+                        $payment->payment_method = $payment_method->displayName;
+                    } else {
+                        $payment->payment_method = null;
                     }
+
+                    // Update total_paid_real value for backward compatibility reasons
+                    if ($payment->id_currency == $order->id_currency) {
+                        $order->total_paid_real += $payment->amount;
+                    } else {
+                        $order->total_paid_real += Tools::ps_round(Tools::convertPrice($payment->amount, $payment->id_currency, false), 2);
+                    }
+                    $order->save();
+
+                    $payment->conversion_rate = ($order ? $order->conversion_rate : 1);
+                    $payment->save();
+                    Db::getInstance()->execute('
+                    INSERT INTO `' . _DB_PREFIX_ . 'order_invoice_payment` (`id_order_invoice`, `id_order_payment`, `id_order`)
+                    VALUES(' . (int) $invoice->id . ', ' . (int) $payment->id . ', ' . (int) $order->id . ')');
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


Progi1984:
@Prestaworks Thanks for the new PR. Could you fix the [Pretty CI errors](https://github.com/PrestaShop/PrestaShop/pull/15417/checks?check_run_id=212986171) ? Thanks

Prestaworks:
Hi, I added the  missing space.

classes/order/OrderHistory.php
```diff
@@ -372,32 +372,34 @@ public function changeIdOrderState($new_order_state, $id_order, $use_existing_pa
...
-                    $payment->order_reference = Tools::substr($order->reference, 0, 9);
-                    $payment->id_currency = $order->id_currency;
-                    $payment->amount = $rest_paid;
-
-                    if ($order->total_paid != 0) {
-                        $payment->payment_method = $payment_method->displayName;
-                    } else {
-                        $payment->payment_method = null;
-                    }
+                if ((int) $invoice->number > 0) {
```
Progi1984: @Prestaworks For avoiding to complexify the loop, you can use `continue` for skipping the process and continue to process the list of invoices.
```suggestion
                if ((int) $invoice->number == 0) {
                    continue;
                }
```

Progi1984 (changes requested):

classes/order/OrderHistory.php
Prestaworks: should it be
if (0 == (int) $invoice->number) {
                    continue;
                }
for following standards?

Progi1984: Exactly ;) YodaStyle ;)

Prestaworks: It's modified.

Prestaworks:
So Is anyone looking at this? it's been about a month now.

Progi1984 (dismissed):

levyn (dismissed):
Thanks!
Works well.

sarahdib:
Hello @Prestaworks 

There is a problem with the delivery slip

Capture d’écran 2019-10-03 à 09 24 03

The payment method and the amount is missing 

It should be 
Capture d’écran 2019-10-03 à 09 26 17

Thank you :)

Prestaworks:
Hi Sarah,

That sounds lika another issue with the delivery slipp not getting the payment information from the correct place?

sarahdib:
Hi @Prestaworks 
 
I don't think this is another issue. The delivery slip was correct before the PR. It's a regression created by the PR

Thank you

Prestaworks:
I think your order is missing an orderPayment, the delivery slipp is looking for an orderpayment to display that information and that do not exist. If you create an invoice to your order, will the payment show up then?

sarahdib:
@Prestaworks 

Without your PR : https://drive.google.com/file/d/1FPC5noGGvKpGcouVWBZoAF79qaN-mGhg/view 
Everything it's ok 

WIth your PR : https://drive.google.com/file/d/1DmekHJ-rFkgoJ3HOCX1TzJa3rSev9Viw/view
I did the same thing and like you see in the video the PR create a regression

Prestaworks:
Hi Sarah,

But that is another issue, if you want a payment the way the system use payments now, you need to create an invoice first and then the payment is attached to the invoice. In this case, there are no invoice so there should not be any invoice payment registered.
So it is correct behaviour. 
If you expect the delivery slip to display payment information without an invoice payment, then the delivery slip payment information needs to be fetched in another way. 
So the entire deliveryslip/invoice handling seems to be faulty in that case..

sarahdib:
@Prestaworks 
Since the PR create a regression I can't validate this PR

Prestaworks:
Hi Sarah,

So what do you expect to happen? that the orderpayment is created without an invoice or that the payment information is not displayed at all on the delivery slip? (if you look at your order, you will see that there is no payment info in the payment box).

Prestaworks:
Hi Sarah, 

Can you please get back to me with an answer so we can change the PR?

ghost:
ping @MatShir, related to Orders

MatShir:
Hi @Prestaworks,

The regression concerned with the order status that has the following parameters:
> * Consider the associated order as validated.
> * Set the order as paid.
> * Show delivery PDF.

So the order should have settled price on the order and a selected payment method even if the order is created through the BO or FO. 
Many merchants use this system for B2B business with the delivery order so If we can deliver as much information as we can on the delivery slip why not. 
The price on the delivery slip could be also an alert for the logistics team if the price is set to 0 accidentally.

> that the order payment is created without an invoice

The invoice could be generated through another system. 

So, any features regressions will penalize the other merchants. 
We really appreciate your contribution but could you find a way to display as it was in the delivery slip? Did you have a technical problem to have to bypass this information to fixe the issue?

Prestaworks:
I made a new change, this gives me the expected behaviour when creating orders and changing status.

fabio-zetaweb:
@Prestaworks 

I have the same issue on v1.7.6.0 but in my case invoice generation system is turned off. So can I still apply your PR, or maybe it could cause other issues? Thanks!

Prestaworks:
@fabio-zetaweb 
I did a quick test with the invoices turned off and got the same result, so it seems to work well even with invoices turned off.

fabio-zetaweb:
@Prestaworks 
Thanks for the support! But I need to know for sure.

Prestaworks:
How is the status for this? it's been about 2 months now :)

Progi1984:
@Prestaworks Thanks for pinging team. I have not seen updates.

@PrestaShop/prestashop-core-developers Is it a BC to change the signature of a public method ?

Prestaworks:
> 
> 
> @Prestaworks Thanks for pinging team. I have not seen updates.
> 
> @PrestaShop/prestashop-core-developers Is it a BC to change the signature of a public method ?

Yes, $use_existing_payment = false is changed to $use_existing_payment = true

eternoendless:
This is a breaking change, method calls should be modified instead of the method itself.

PR #15428, opened 2019-09-05
Add check for required fields in BO > Country > Address format

### LINKED ISSUE #9812: [BOOM-4291] 500 Error adding address at checkout or via my account
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-4291).

- _**Reporter:**_ elpatron
- _**Created at:**_ Sun, 12 Nov 2017 21:13:01 +0100

<p>Hi on new client upgrade we had issue where new address caused 500 error..either at checkout or via 'my account' 'add address'.<br/>
We solved issue from this stackoverflow thread<br/>
<a href="https://stackoverflow.com/questions/42413535/prestashop-1-7-unable-to-save-order-address" class="external-link" rel="nofollow">https://stackoverflow.com/questions/42413535/prestashop-1-7-unable-to-save-order-address</a></p>

<p>This solved the issue.</p>

<p>original error info<br/>
account creation, add address 500 errors.</p>

<p>to replicate, create new account then add new address. I originally had this issue when trying to check out and adding address...<br/>
<span class="error">&#91;PrestaShopException&#93;</span></p>

<p>Property Address->id_country is empty. <br/>
at line 944 in file classes/ObjectModel.php <br/>
939. } <br/>
940. <br/>
941. $message = $this->validateField($field, $this->$field); <br/>
942. if ($message !== true) { <br/>
943. if ($die) </p>
{ 
944. throw new PrestaShopException($message); 
945. }
<p> <br/>
946. return $error_return ? $message : false; <br/>
947. } <br/>
948. } <br/>
949. </p>

<p>ObjectModelCore->validateFields - <span class="error">&#91;line 272 - classes/ObjectModel.php&#93;</span><br/>
ObjectModelCore->getFields - <span class="error">&#91;line 683 - classes/ObjectModel.php&#93;</span><br/>
ObjectModelCore->update - <span class="error">&#91;line 202 - classes/Address.php&#93;</span> - <span class="error">&#91;1 Arguments&#93;</span><br/>
AddressCore->update - <span class="error">&#91;line 219 - classes/Address.php&#93;</span><br/>
AddressCore->delete - <span class="error">&#91;line 73 - classes/form/CustomerAddressPersister.php&#93;</span><br/>
CustomerAddressPersisterCore->save - <span class="error">&#91;line 162 - classes/form/CustomerAddressForm.php&#93;</span> - <span class="error">&#91;2 Arguments&#93;</span><br/>
CustomerAddressFormCore->submit - <span class="error">&#91;line 111 - classes/checkout/CheckoutAddressesStep.php&#93;</span><br/>
CheckoutAddressesStepCore->handleRequest - <span class="error">&#91;line 57 - classes/checkout/CheckoutProcess.php&#93;</span> - <span class="error">&#91;1 Arguments&#93;</span><br/>
CheckoutProcessCore->handleRequest - <span class="error">&#91;line 201 - controllers/front/OrderController.php&#93;</span> - <span class="error">&#91;1 Arguments&#93;</span><br/>
OrderControllerCore->initContent - <span class="error">&#91;line 205 - classes/controller/Controller.php&#93;</span><br/>
ControllerCore->run - <span class="error">&#91;line 379 - classes/Dispatcher.php&#93;</span><br/>
DispatcherCore->dispatch - <span class="error">&#91;line 28 - index.php&#93;</span></p>
        
add address via checkout or my account

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | When editing the address format for a country, check that all required address fields are present. This prevents the address form in the frontend from throwing an exception because of missing required fields. 
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9812
| How to test?  | In BO > International > Zones geo > Country, change address format, remove a required field (for ex: firstname, lastname, or Country:name), you should see an error message stating that this field is required.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15428)

### DIFF
--- a/admin-dev/themes/default/template/controllers/countries/helpers/form/form.tpl
+++ b/admin-dev/themes/default/template/controllers/countries/helpers/form/form.tpl
@@ -32,7 +32,7 @@
 					<textarea id="ordered_fields" name="address_layout" style="height:150px;">{$input.address_layout}</textarea>
 				</div>
 				<div class="col-lg-8">
-					{l s='Required fields for the address (click for more details):' d='Admin.International.Feature'}
+					{l s='Some countries require different elements than others. Click on the button below to get the valid default address format for this country.' d='Admin.International.Feature'}
 					{$input.display_valid_fields}
 				</div>
 			</div>
--- a/classes/AddressFormat.php
+++ b/classes/AddressFormat.php
@@ -23,6 +23,7 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
+use PrestaShop\PrestaShop\Core\Domain\Address\Exception\AddressException;
 
 /**
  * Class AddressFormatCore.
@@ -263,11 +264,13 @@ class AddressFormatCore extends ObjectModel
     /**
      * Given a field name, get the name of the tab in which the field name can be found.
      * For ex: Country:name => the tab is 'Country'.
-     * There should be only one separator in the string, otherwise return false.
+     * There should be only one separator in the string, otherwise throw an exception.
      *
      * @param string $field
      *
      * @return bool|string
+     *
+     * @throws AddressException
      */
     private function getFieldTabName($field)
     {
@@ -282,7 +285,7 @@ class AddressFormatCore extends ObjectModel
             return $fieldTab[0];
         }
 
-        return false;
+        throw new AddressException('Address format field is not valid');
     }
 
     /**
--- a/classes/AddressFormat.php
+++ b/classes/AddressFormat.php
@@ -253,7 +253,7 @@ class AddressFormatCore extends ObjectModel
         foreach (self::getFieldsRequired() as $requiredField) {
             if (!in_array($requiredField, $fieldList)) {
                 $this->_errorFormatList[] = $this->trans(
-                    'The field %s is required (in the "%s" tab).',
+                    'The field %s is required (in the %s tab).',
                     array($requiredField, $this->getFieldTabName($requiredField)),
                     'Admin.Notifications.Error');
             }
--- a/classes/AddressFormat.php
+++ b/classes/AddressFormat.php
@@ -253,7 +253,7 @@ class AddressFormatCore extends ObjectModel
         foreach (self::getFieldsRequired() as $requiredField) {
             if (!in_array($requiredField, $fieldList)) {
                 $this->_errorFormatList[] = $this->trans(
-                    'The field %s is required (in the %s tab).',
+                    'The %s field (in tab %s) is required.',
                     array($requiredField, $this->getFieldTabName($requiredField)),
                     'Admin.Notifications.Error');
             }
--- a/classes/AddressFormat.php
+++ b/classes/AddressFormat.php
@@ -252,9 +252,37 @@ class AddressFormatCore extends ObjectModel
     {
         foreach (self::getFieldsRequired() as $requiredField) {
             if (!in_array($requiredField, $fieldList)) {
-                $this->_errorFormatList[] = $this->trans('The field %s is required.', array($requiredField), 'Admin.Notifications.Error');
+                $this->_errorFormatList[] = $this->trans(
+                    'The field %s is required (in the "%s" tab).',
+                    array($requiredField, $this->getFieldTabName($requiredField)),
+                    'Admin.Notifications.Error');
             }
         }
+    }
+
+    /**
+     * Given a field name, get the name of the tab in which the field name can be found.
+     * For ex: Country:name => the tab is 'Country'.
+     * There should be only one separator in the string, otherwise return false.
+     *
+     * @param string $field
+     *
+     * @return bool|string
+     */
+    private function getFieldTabName($field)
+    {
+        if (strpos($field, ':') === false) {
+            // When there is no ':' separator, the field is in the Address tab
+            return 'Address';
+        }
+
+        $fieldTab = explode(':', $field);
+        if (count($fieldTab) === 2) {
+            // The part preceding the ':' separator is the name of the tab in which there is the required field
+            return $fieldTab[0];
+        }
+
+        return false;
     }
 
     /**
--- a/classes/AddressFormat.php
+++ b/classes/AddressFormat.php
@@ -236,9 +236,25 @@ class AddressFormatCore extends ObjectModel
                     }
                 }
             }
+            $this->checkRequiredFields($usedKeyList);
         }
 
         return (count($this->_errorFormatList)) ? false : true;
+    }
+
+    /**
+     * Checks that all required fields exist in a given fields list.
+     * Fills _errorFormatList array in case of absence of a required field.
+     *
+     * @param array $fieldList
+     */
+    protected function checkRequiredFields($fieldList)
+    {
+        foreach (self::getFieldsRequired() as $requiredField) {
+            if (!in_array($requiredField, $fieldList)) {
+                $this->_errorFormatList[] = $this->trans('The field %s is required.', array($requiredField), 'Admin.Notifications.Error');
+            }
+        }
     }
 
     /**


matthieu-rolland:
@LouiseBonnard , when removing a field that is required from the Address Format form, it will display this error message: **The field %s is required.**

This is the form I'm talking about:

Capture d’écran 2019-09-05 à 17 07 08

I used this message because it already exists and I didn't find anything more suitable, would you have any idea for a better message ?

jolelievre (dismissed):
Nice @matthieu-rolland

classes/AddressFormat.php
```diff
@@ -236,11 +236,27 @@ public function checkFormatFields()
...
         return (count($this->_errorFormatList)) ? false : true;
     }
 
+    /**
+     * Checks that all required fields exist in a given fields list.
+     * Fills _errorFormatList array in case of absence of a required field.
+     *
+     * @param array $fieldList
+     */
+    protected function _checkRequiredFields($fieldList)
```
PierreRambaud: No reason for underscore in the function name.

```suggestion
    protected function checkRequiredFields($fieldList)
```

PierreRambaud (changes requested):
Little lint issue

classes/AddressFormat.php
matthieu-rolland: done !

PierreRambaud (dismissed):

sarahdib:
Hello @matthieu-rolland 
There is a little problem 
When you delete for exemple firstname save it's Ok
When you add by clicking on firstname and save the message is still appear:

https://drive.google.com/file/d/1dtZ4dZk1sLztqo2FvfeaJ8ANdJ_SXh79/view

matthieu-rolland:
> Hello @matthieu-rolland
> There is a little problem
> When you delete for exemple firstname save it's Ok
> When you add by clicking on firstname and save the message is still appear:
> 
> https://drive.google.com/file/d/1dtZ4dZk1sLztqo2FvfeaJ8ANdJ_SXh79/view

That's normal behaviour, because **customer:lastname** is not the same as **lastname**, you can add lastname from the Address tab.

As seen with @sarahdib , that's not really clear for the end user. 

Ping @colinegin @marionf  , I suggest we add a little text above this form, stating the required fields from the Address tab, so that the user doesn't try to add the required lastname from another tab for example. What do you think ?

marionf:
@matthieu-rolland 

> I suggest we add a little text above this form, stating the required fields from the Address tab

Why not, but we should also handle the case where the user add **customer:lastname** and also display an error if it's required

matthieu-rolland:
> @matthieu-rolland
> 
> > I suggest we add a little text above this form, stating the required fields from the Address tab
> 
> Why not, but we should also handle the case where the user add **customer:lastname** and also display an error if it's required

@marionf 
**customer:lastname** is not in the list of required fields, it's just that since it's in the first tab, after seeing the error message saying that the address is required, people would click on it thinking that this is the required field.  (this is what induced @sarahdib 's misunderstanding and I'm afraid the end users will have the same misunderstanding).  

Ok, I'll do that !

matthieu-rolland:
As seen with @colinegin , we will precise the tab of the required field in the error message, and change the text above the address format field.

admin-dev/themes/default/template/controllers/countries/helpers/form/form.tpl
```diff
@@ -32,7 +32,7 @@
 					<textarea id="ordered_fields" name="address_layout" style="height:150px;">{$input.address_layout}</textarea>
 				</div>
 				<div class="col-lg-8">
-					{l s='Required fields for the address (click for more details):' d='Admin.International.Feature'}
+					{l s='Depending on your country, some elements can be required, you can click on "Use the default format" to get valid address format for this country.' d='Admin.International.Feature'}
```
matthieu-rolland: Hi @LouiseBonnard, as seen with @colinegin we are rewording this part, don't hesitate if you have ideas for a better formulation, the two informations that we want to pass are:

1/ Depending on the country the required fields might differ.
2/ There's a button just below to get a default address format for this country.

classes/AddressFormat.php
```diff
@@ -236,11 +236,55 @@ public function checkFormatFields()
...
+     * @param array $fieldList
+     */
+    protected function checkRequiredFields($fieldList)
+    {
+        foreach (self::getFieldsRequired() as $requiredField) {
+            if (!in_array($requiredField, $fieldList)) {
+                $this->_errorFormatList[] = $this->trans(
+                    'The field %s is required (in the %s tab).',
+                    array($requiredField, $this->getFieldTabName($requiredField)),
+                    'Admin.Notifications.Error');
```
matthieu-rolland, line 259: @LouiseBonnard  We are making this error message more specific, the name of the tab where the required field can be found will be displayed in it.

LouiseBonnard: Hi @matthieu-rolland, why are we adding the tab name?

classes/AddressFormat.php
```diff
@@ -236,11 +236,55 @@ public function checkFormatFields()
...
+            return 'Address';
+        }
+
+        $fieldTab = explode(':', $field);
+        if (count($fieldTab) === 2) {
+            // The part preceding the ':' separator is the name of the tab in which there is the required field
+            return $fieldTab[0];
+        }
+
+        return false;
```
matthieu-rolland: @jolelievre @PierreRambaud 

I have doubts on how to deal with this (when this function returns false).

"Normally" it can never happen, but I wanted the method to be abnormal-proof. Do you think I should throw some kind of exception, in the legacy?

admin-dev/themes/default/template/controllers/countries/helpers/form/form.tpl
LouiseBonnard: Here is my suggestion, feel free to challenge it: _Some countries require different elements than others. Click on the button below to get the valid default address format for this country._ localized in the same domain. :-)

matthieu-rolland: 👍 awesome, thank you !

classes/AddressFormat.php
matthieu-rolland: Because, if let's say lastname is missing, there are several "lastname" elements in the form, one in the Customer tab, another in the Address tab etc... We have several elements having the same name in different tabs like this. Otherwise it's not clear for the end user to know from where he can add the missing field.

Capture d’écran 2019-09-30 à 16 47 20

LouiseBonnard: Alright then, thanks for the explanation. I would write: _The %s field (in tab %s) is required._

matthieu-rolland: @LouiseBonnard  thanks ! I made the modification

classes/AddressFormat.php
eternoendless: As a general rule, methods should do what they are expected to do, or throw an exception otherwise. Return values should only reflect "correct" and "expected" states in normal execution, all other situations should be considered exceptions.

matthieu-rolland: Thanks for your input @eternoendless  👍

PierreRambaud (approved):

Robin-Fischer-PS:
Hi,

This PR is working fine, but I'm not sure about the wording and/or position of this message : 
"Some countries require different elements than others. Click on the button below to get the valid default address format for this country. "
Since there are no less than 4 buttons under this message, + the block to add parts of adress, I wonder if the end user will easily understand which button this message is pointing to...

pr15428 not clear 1

What do you think @TristanLDD @LouiseBonnard @colinegin ?

Thanks :)

LouiseBonnard:
Hi @Robin-Fischer-PS, you're right and it is supposed to be corrected with the migration, cf. [this PR](https://github.com/PrestaShop/PrestaShop/pull/15180#discussion_r316568610).

admin-dev/themes/default/template/controllers/countries/helpers/form/form.tpl
```diff
@@ -32,7 +32,7 @@
 					<textarea id="ordered_fields" name="address_layout" style="height:150px;">{$input.address_layout}</textarea>
 				</div>
 				<div class="col-lg-8">
-					{l s='Required fields for the address (click for more details):' d='Admin.International.Feature'}
+					{l s='Some countries require different elements than others. Click on the button below to get the valid default address format for this country.' d='Admin.International.Feature'}
```
LouiseBonnard, line 35:
```suggestion
					{l s='Some countries require different elements than others. Click the button below to get the valid default address format for this country.' d='Admin.International.Feature'}
```

PR #15431, opened 2019-09-05
Add Money unit test

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add unit test
| Type?         | improvement
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Travis is green

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15431)

### DIFF
--- a/tests/Unit/Core/Domain/ValueObject/MoneyTest.php
+++ b/tests/Unit/Core/Domain/ValueObject/MoneyTest.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace Tests\Unit\Core\Domain\ValueObject;
+
+use PHPUnit\Framework\TestCase;
+use PrestaShop\Decimal\Number;
+use PrestaShop\PrestaShop\Core\Domain\Currency\ValueObject\CurrencyId;
+use PrestaShop\PrestaShop\Core\Domain\Exception\DomainConstraintException;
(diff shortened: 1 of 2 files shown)


tests/Unit/Core/Domain/ValueObject/MoneyTest.php
```diff
@@ -0,0 +1,51 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Exception\DomainConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\ValueObject\Money;
+
+class MoneyTest extends TestCase
+{
+    public function testCreateWithValidValue()
+    {
+        $money = new Money(new Number('100'), new CurrencyId(10));
+
+        $this->assertEquals(100, $money->getAmount());
```
PierreRambaud: getAmount returns a Number, not a Integer ;)
```suggestion
        $this->assertEquals('100', (string) $money->getAmount());
```

matks: I knew it 😆 I thought to myself "do I need to run such a simple test before the PR ? naaaaah I'm fine"

I got punished by fate 😉

matks:
@PierreRambaud fixed

By the way I'm still unsure whether a `Money` object should contain a `$currencyId` property or a `$isoCode` property 🤔 

The currencyId is needed anyways in order to be able to perform database-related operations but could be retrieved later. Using `$isoCode` would make the object simpler to create. Wdyt ?

PierreRambaud:
> @PierreRambaud fixed
> 
> By the way I'm still unsure whether a `Money` object should contain a `$currencyId` property or a `$isoCode` property 
> 
> The currencyId is needed anyways in order to be able to perform database-related operations but could be retrieved later. Using `$isoCode` would make the object simpler to create. Wdyt ?

I'm not the best one to answer you :sweat_smile: 
But I think to declare a `Money` you need a `Currency` and also an `IsoCode` because money are based on a currency and where it's used :thinking: 
PIng @eternoendless

tests/Unit/Core/Domain/ValueObject/MoneyTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+        $this->assertEquals('100', (string) $money->getAmount());
+    }
+
+    public function testCreateWithBadValue()
+    {
+        $this->expectException(DomainConstraintException::class);
+        $this->expectExceptionCode(DomainConstraintException::INVALID_MONEY_AMOUNT);
+
+        $money = new Money(new Number('-1'), new CurrencyId(10));
```
eternoendless: O_o, why would a negative money amount be invalid?

matks: That's why this test checks an exception is thrown 😉

eternoendless: The problem is that a negative money amount is conceptually valid (eg. a discount, a reimbursement)

matks: Depends on you consider it. You can consider a discount has the informations: "amount", "10", "action": "deduce". The VO Money has been introduced to handle CartRule informations, where the amount and he action are too separate things. But we can remove this validation rule very easily.

Progi1984 (dismissed):

Progi1984 (commented):
@matks Some PrettyCI to fix and a second review for this technical PR.

matks:
@Progi1984 fixed 😄

Progi1984 (approved):

matks:
Thanks for review !

PR #14089, opened 2019-06-05
Fix shop URL with port in header HOST

202-ecommerce:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | When a server send a request with an host with port in HTTPS, Prestashop don't found the shop and send a redirection to HTTP url. 
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | yes, potential BC with shop not on standart port 80 and/or 443
| Deprecations? | no
| Fixed ticket? | Fixes  #14085
| How to test?  | Le CLI curl -I -H "Host: www.my-prestashop.tld:443" https://www.my-prestashop.tld/ need return 200 status code not 301

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14089)

PR #15446, opened 2019-09-06
Re-enable port in shop domains

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Since https://github.com/PrestaShop/PrestaShop/pull/14089 was merged, you cannot use `mysite:myportnumber` (with a not-default port number) as a domain name although this can be useful. This PR re-enabled this as a fallback. The fact it is used as a fallback should make sure https://github.com/PrestaShop/PrestaShop/pull/14089 bugfix is still valid.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Install a shop using a not-standard port such as 8888 (default MAMP port) (so your domain should be `localhost:8888` and try to install and reach the FO.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15446)

### DIFF
--- a/classes/shop/Shop.php
+++ b/classes/shop/Shop.php
@@ -334,15 +334,13 @@ class ShopCore extends ObjectModel
             $host = Tools::getHttpHost(false, false, true);
             $request_uri = rawurldecode($_SERVER['REQUEST_URI']);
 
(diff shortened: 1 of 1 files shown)


matks:
Closing as I did not use my fork 💔

PR #15448, opened 2019-09-06
Replace uses of deprecated Tools::displayPrice in favor of Locale::formatPrice()

### LINKED ISSUE #15427: Stop using Tools::displayPrice as it is deprecated
`Tools::displayPrice()` is deprecated from 1.7.6.0

We need to update all usages of this function in the project and use `Locale::formatPrice()` instead

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | Refactoring for a deprecated method
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15427
| How to test?  |

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15448)

### DIFF
--- a/admin-dev/themes/default/template/controllers/orders/_customized_data.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/_customized_data.tpl
@@ -221,7 +221,7 @@
 								<input onchange="checkPartialRefundProductAmount(this)" type="text" name="partialRefundProduct[{$product['id_order_detail']|intval}]" />
 								{if !$currencySymbolBeforeAmount}<div class="input-group-addon">{$currency->sign}</div>{/if}
 							</div>
-							<p class="help-block"><i class="icon-warning-sign"></i> {l s='(Max %amount_refundable% %tax_method%)' sprintf=[ '%amount_refundable%' => Tools::displayPrice(Tools::ps_round($amount_refundable, 2), $currency->id), '%tax_method%' => $smarty.capture.TaxMethod] d='Admin.Orderscustomers.Help'}</p>
+							<p class="help-block"><i class="icon-warning-sign"></i> {l s='(Max %amount_refundable% %tax_method%)' sprintf=[ '%amount_refundable%' => Context::getContext()->getCurrentLocale()->formatPrice(Tools::ps_round($amount_refundable, 2), Currency::getCurrencyInstance($currency->id)->iso_code), '%tax_method%' => $smarty.capture.TaxMethod] d='Admin.Orderscustomers.Help'}</p>
 						</div>
 					</div>
 					{/if}
--- a/admin-dev/themes/default/template/controllers/orders/_customized_data.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/_customized_data.tpl
@@ -221,7 +221,7 @@
 								<input onchange="checkPartialRefundProductAmount(this)" type="text" name="partialRefundProduct[{$product['id_order_detail']|intval}]" />
 								{if !$currencySymbolBeforeAmount}<div class="input-group-addon">{$currency->sign}</div>{/if}
 							</div>
-							<p class="help-block"><i class="icon-warning-sign"></i> {l s='(Max %amount_refundable% %tax_method%)' sprintf=[ '%amount_refundable%' => Context::getContext()->getCurrentLocale()->formatPrice(Tools::ps_round($amount_refundable, 2), Currency::getCurrencyInstance($currency->id)->iso_code), '%tax_method%' => $smarty.capture.TaxMethod] d='Admin.Orderscustomers.Help'}</p>
+							<p class="help-block"><i class="icon-warning-sign"></i> {l s='(Max %amount_refundable% %tax_method%)' sprintf=[ '%amount_refundable%' => Context::getContext()->getCurrentLocale()->formatPrice($amount_refundable, Currency::getCurrencyInstance($currency->id)->iso_code), '%tax_method%' => $smarty.capture.TaxMethod] d='Admin.Orderscustomers.Help'}</p>
 						</div>
 					</div>
 					{/if}
--- a/admin-dev/themes/default/template/controllers/orders/_customized_data.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/_customized_data.tpl
@@ -221,7 +221,7 @@
 								<input onchange="checkPartialRefundProductAmount(this)" type="text" name="partialRefundProduct[{$product['id_order_detail']|intval}]" />
 								{if !$currencySymbolBeforeAmount}<div class="input-group-addon">{$currency->sign}</div>{/if}
 							</div>
-							<p class="help-block"><i class="icon-warning-sign"></i> {l s='(Max %amount_refundable% %tax_method%)' sprintf=[ '%amount_refundable%' => Context::getContext()->getCurrentLocale()->formatPrice($amount_refundable, Currency::getCurrencyInstance($currency->id)->iso_code), '%tax_method%' => $smarty.capture.TaxMethod] d='Admin.Orderscustomers.Help'}</p>
+							<p class="help-block"><i class="icon-warning-sign"></i> {l s='(Max %amount_refundable% %tax_method%)' sprintf=[ '%amount_refundable%' => Context::getContext()->getCurrentLocale()->formatPrice($amount_refundable, Currency::getIsoCodeById((int)$currency->id)), '%tax_method%' => $smarty.capture.TaxMethod] d='Admin.Orderscustomers.Help'}</p>
 						</div>
 					</div>
 					{/if}
--- a/admin-dev/themes/default/template/controllers/orders/_product_line.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/_product_line.tpl
@@ -189,7 +189,10 @@
 					<input onchange="checkPartialRefundProductAmount(this)" type="text" name="partialRefundProduct[{$product['id_order_detail']}]" />
 					{if !$currencySymbolBeforeAmount}<div class="input-group-addon">{$currency->sign}</div>{/if}
 				</div>
-        <p class="help-block"><i class="icon-warning-sign"></i> {l s='(Max %amount_refundable% %tax_method%)' sprintf=[ '%amount_refundable%' => Tools::displayPrice(Tools::ps_round($amount_refundable, 2), $currency->id), '%tax_method%' => $smarty.capture.TaxMethod] d='Admin.Orderscustomers.Help'}</p>
+        <p class="help-block"><i class="icon-warning-sign"></i> {l
+            s='(Max %amount_refundable% %tax_method%)'
+            sprintf=['%amount_refundable%' => Context::getContext()->getCurrentLocale()->formatPrice(Tools::ps_round($amount_refundable, 2), Currency::getCurrencyInstance($currency->id)->iso_code), '%tax_method%' => $smarty.capture.TaxMethod]
+            d='Admin.Orderscustomers.Help'}</p>
 			</div>
 		</div>
 		{/if}
--- a/admin-dev/themes/default/template/controllers/orders/_product_line.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/_product_line.tpl
@@ -189,10 +189,10 @@
 					<input onchange="checkPartialRefundProductAmount(this)" type="text" name="partialRefundProduct[{$product['id_order_detail']}]" />
 					{if !$currencySymbolBeforeAmount}<div class="input-group-addon">{$currency->sign}</div>{/if}
 				</div>
-        <p class="help-block"><i class="icon-warning-sign"></i> {l
+        <p class="help-block"><i class="icon-warning-sign"></i> {if !is_null($amount_refundable)}{l
             s='(Max %amount_refundable% %tax_method%)'
-            sprintf=['%amount_refundable%' => Context::getContext()->getCurrentLocale()->formatPrice(Tools::ps_round($amount_refundable, 2), Currency::getCurrencyInstance($currency->id)->iso_code), '%tax_method%' => $smarty.capture.TaxMethod]
-            d='Admin.Orderscustomers.Help'}</p>
+            sprintf=['%amount_refundable%' => Context::getContext()->getCurrentLocale()->formatPrice($amount_refundable, Currency::getIsoCodeById((int) $currency->id)), '%tax_method%' => $smarty.capture.TaxMethod]
+            d='Admin.Orderscustomers.Help'}{/if}</p>
 			</div>
 		</div>
 		{/if}
--- a/admin-dev/themes/default/template/controllers/orders/helpers/view/view.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/helpers/view/view.tpl
@@ -1111,7 +1111,7 @@
                         </div>
                         <p class="help-block"><i class="icon-warning-sign"></i> {l
                             s='(Max %s %s)'
-                            sprintf=[Tools::displayPrice(Tools::ps_round($shipping_refundable, 2), $currency->id) , $smarty.capture.TaxMethod]
+                            sprintf=[Context::getContext()->getCurrentLocale()->formatPrice(Tools::ps_round($shipping_refundable, 2), Currency::getCurrencyInstance($currency->id)->iso_code) , $smarty.capture.TaxMethod]
                             d='Admin.Orderscustomers.Feature'
                             }
                         </p>
--- a/admin-dev/themes/default/template/controllers/orders/helpers/view/view.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/helpers/view/view.tpl
@@ -1111,7 +1111,7 @@
                         </div>
                         <p class="help-block"><i class="icon-warning-sign"></i> {l
                             s='(Max %s %s)'
-                            sprintf=[Context::getContext()->getCurrentLocale()->formatPrice(Tools::ps_round($shipping_refundable, 2), Currency::getCurrencyInstance($currency->id)->iso_code) , $smarty.capture.TaxMethod]
+                            sprintf=[Context::getContext()->getCurrentLocale()->formatPrice($shipping_refundable, Currency::getIsoCodeById((int) $currency->id)) , $smarty.capture.TaxMethod]
                             d='Admin.Orderscustomers.Feature'
                             }
                         </p>
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -613,6 +613,27 @@ class CurrencyCore extends ObjectModel
     }
 
     /**
+     * Get Currency ISO Code by ID
+     *
+     * @param int $id
+     * @param bool $forceRefreshCache
+     *
+     * @return string
+     */
+    public static function getIsoCodeById(int $id, bool $forceRefreshCache = false)
+    {
+        $cacheId = 'Currency::getIsoCodeById' . pSQL($id);
+        if ($forceRefreshCache || !Cache::isStored($cacheId)) {
+            $result = Currency::getCurrencyInstance($id);
+            Cache::store($cacheId, $result->iso_code);
+
+            return $result->iso_code;
+        }
+
+        return Cache::retrieve($cacheId);
+    }
+
+    /**
      * Get Currency ID query.
      *
      * @param int $idShop Shop ID
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -82,6 +82,8 @@ class CurrencyCore extends ObjectModel
     /**
      * Currency's symbol
      *
+     * @deprecated Use $symbol
+     *
      * @var string
      */
     public $sign;
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -624,10 +624,10 @@ class CurrencyCore extends ObjectModel
     {
         $cacheId = 'Currency::getIsoCodeById' . pSQL($id);
         if ($forceRefreshCache || !Cache::isStored($cacheId)) {
-            $result = Currency::getCurrencyInstance($id);
-            Cache::store($cacheId, $result->iso_code);
+            $resultIsoCode = Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue('SELECT `iso_code` FROM ' . _DB_PREFIX_ . 'currency WHERE `id_currency` = ' . (int) $id);
+            Cache::store($cacheId, $resultIsoCode);
 
-            return $result->iso_code;
+            return $resultIsoCode;
         }
 
         return Cache::retrieve($cacheId);
--- a/classes/controller/AdminController.php
+++ b/classes/controller/AdminController.php
@@ -416,9 +416,6 @@ class AdminControllerCore extends Controller
         if (strpos($this->controller_name, 'ControllerOverride')) {
             $this->controller_name = substr($this->controller_name, 0, -18);
         }
-        if (strpos($this->controller_name, 'ControllerCore')) {
-            $this->controller_name = substr($this->controller_name, 0, -14);
-        }
         if (strpos($this->controller_name, 'Controller')) {
             $this->controller_name = substr($this->controller_name, 0, -10);
         }
--- a/classes/controller/AdminController.php
+++ b/classes/controller/AdminController.php
@@ -416,6 +416,9 @@ class AdminControllerCore extends Controller
         if (strpos($this->controller_name, 'ControllerOverride')) {
             $this->controller_name = substr($this->controller_name, 0, -18);
         }
+        if (strpos($this->controller_name, 'ControllerCore')) {
+            $this->controller_name = substr($this->controller_name, 0, -14);
+        }
         if (strpos($this->controller_name, 'Controller')) {
             $this->controller_name = substr($this->controller_name, 0, -10);
         }
--- a/classes/order/OrderHistory.php
+++ b/classes/order/OrderHistory.php
@@ -520,7 +520,7 @@ class OrderHistoryCore extends ObjectModel
             }
 
             $context = Context::getContext();
-            $data['{total_paid}'] = Tools::getContextLocale($context)->formatPrice((float) $order->total_paid, (new Currency((int) $order->id_currency))->iso_code);
+            $data['{total_paid}'] = Tools::getContextLocale($context)->formatPrice((float) $order->total_paid, Currency::getIsoCodeById((int) $order->id_currency));
 
             if (Validate::isLoadedObject($order)) {
                 // Attach invoice and / or delivery-slip if they exists and status is set to attach them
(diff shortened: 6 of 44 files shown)


matks:
@Progi1984 When a PR modifies several folders, we go for CO 😉(instead of `Category? | FO / BO / TE`)
The changelog requires we choose only one category

LedCloud:
Why not to rewite only Tools::displayPrice?

old function still might be used by some external modules.

jolelievre:
@LedCloud `Tools::displayPrice` has been deprecated because it doesn't respect the expected architecture any more. A deprecated function is meant to be removed in the next major version.

So although we did modify the function so that it calls the new expected way to display price (so that it still works for now), this method should not be used in any module nor in the core. That's why we must, at the minimum, show the example and remove its uses from the core code.

PierreRambaud:
@Progi1984 Don't forget to squash your commit to prevent:
image

composer.json
```diff
@@ -198,7 +198,7 @@
             "@php -d date.timezone=UTC ./vendor/bin/phpunit -c tests-legacy/old-controllers.xml"
         ],
         "integration-behaviour-tests": [
-            "@php -d date.timezone=UTC ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml"
+            "@php -d date.timezone=UTC ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml -vv"
```
PierreRambaud: Not sure it's a good idea to add the verbose mode since Travis have a limited amount of logs :thinking:

Progi1984: I will remove it before squashing commits

src/Adapter/Product/PriceFormatter.php
```diff
@@ -52,7 +80,9 @@ public function convertAmount($price, $currency = null)
      */
     public function format($price, $currency = null)
     {
-        return Tools::displayPrice($price, $currency);
+        $currency = $currency ?: $this->context->currency;
+        $currency = is_array($currency) ? Currency::getCurrencyInstance($currency['id_currency']) : $currency;
```
PierreRambaud:
```suggestion
        $currency = isset($currency['id_currency']) ? Currency::getCurrencyInstance($currency['id_currency']) : $currency;
```

Progi1984: :+1:

jolelievre:
@Progi1984 some conflicts to fix on this one ^^

Progi1984:
@jolelievre Fixed. Thanks

admin-dev/themes/default/template/controllers/orders/_customized_data.tpl
```diff
@@ -221,7 +221,7 @@
 								<input onchange="checkPartialRefundProductAmount(this)" type="text" name="partialRefundProduct[{$product['id_order_detail']|intval}]" />
 								{if !$currencySymbolBeforeAmount}<div class="input-group-addon">{$currency->sign}</div>{/if}
 							</div>
-							<p class="help-block"><i class="icon-warning-sign"></i> {l s='(Max %amount_refundable% %tax_method%)' sprintf=[ '%amount_refundable%' => Tools::displayPrice(Tools::ps_round($amount_refundable, 2), $currency->id), '%tax_method%' => $smarty.capture.TaxMethod] d='Admin.Orderscustomers.Help'}</p>
+							<p class="help-block"><i class="icon-warning-sign"></i> {l s='(Max %amount_refundable% %tax_method%)' sprintf=[ '%amount_refundable%' => Context::getContext()->getCurrentLocale()->formatPrice(Tools::ps_round($amount_refundable, 2), Currency::getCurrencyInstance($currency->id)->iso_code), '%tax_method%' => $smarty.capture.TaxMethod] d='Admin.Orderscustomers.Help'}</p>
```
jolelievre: I'm worried about this hard-coded 2, if it's display/compute precision then it should be modified (cf what @matthieu-rolland has done on this subject)
And I'm not even sure it should be rounded here since anyway the CLDR manages the precision when displaying a price now Maybe you could remove this `Tools::ps_round` altogether

controllers/admin/AdminCartsController.php
```diff
@@ -675,8 +675,8 @@ protected function getCartSummary()
             foreach ($summary['products'] as &$product) {
                 $product['numeric_price'] = $product['price'];
                 $product['numeric_total'] = $product['total'];
-                $product['price'] = str_replace($currency->sign, '', Tools::displayPrice($product['price'], $currency));
-                $product['total'] = str_replace($currency->sign, '', Tools::displayPrice($product['total'], $currency));
+                $product['price'] = str_replace($currency->sign, '', $this->context->getCurrentLocale()->formatPrice($product['price'], $currency->iso_code));
```
jolelievre: `$currency->sign` has been deprecated prefer using `$currency->symbol`

controllers/admin/AdminOrdersController.php
```diff
@@ -2341,7 +2342,7 @@ public function ajaxProcessAddProductOnOrder()
         $resume = OrderSlip::getProductSlipResume((int) $product['id_order_detail']);
         $product['quantity_refundable'] = $product['product_quantity'] - $resume['product_quantity'];
         $product['amount_refundable'] = $product['total_price_tax_excl'] - $resume['amount_tax_excl'];
-        $product['amount_refund'] = Tools::displayPrice($resume['amount_tax_incl']);
+        $product['amount_refund'] = is_null($resume['amount_tax_incl']) ? $resume['amount_tax_incl'] : $this->context->getCurrentLocale()->formatPrice($resume['amount_tax_incl'], $this->context->currency->iso_code);
```
jolelievre:
```suggestion
        $product['amount_refund'] = !is_null($resume['amount_tax_incl']) ? $this->context->getCurrentLocale()->formatPrice($resume['amount_tax_incl'], $this->context->currency->iso_code) : null;
```

controllers/admin/AdminOrdersController.php
```diff
@@ -2600,7 +2601,7 @@ public function ajaxProcessEditProductOnOrder()
         $resume = OrderSlip::getProductSlipResume($order_detail->id);
         $product['quantity_refundable'] = $product['product_quantity'] - $resume['product_quantity'];
         $product['amount_refundable'] = $product['total_price_tax_excl'] - $resume['amount_tax_excl'];
-        $product['amount_refund'] = Tools::displayPrice($resume['amount_tax_incl']);
+        $product['amount_refund'] = is_null($resume['amount_tax_incl']) ? $resume['amount_tax_incl'] : $this->context->getCurrentLocale()->formatPrice($resume['amount_tax_incl'], $this->context->currency->iso_code);
```
jolelievre:
```suggestion
        $product['amount_refund'] = !is_null($resume['amount_tax_incl']) ? $this->context->getCurrentLocale()->formatPrice($resume['amount_tax_incl'], $this->context->currency->iso_code) : null;
```

src/Adapter/Cart/QueryHandler/GetCartForViewingHandler.php
```diff
@@ -35,31 +35,48 @@
...
+
     /**
      * @param ImageManager $imageManager
+     * @param LocaleRepository $repository
      */
-    public function __construct(ImageManager $imageManager)
+    public function __construct(ImageManager $imageManager, LocaleRepository $repository)
     {
         $this->imageManager = $imageManager;
+        $this->context = new LegacyContext();
```
jolelievre: I think you'd better inject the locale via `'@=service("prestashop.adapter.legacy.context").getContext().language.getLocale()'` this avoid cerating a dependency with LegacyContext in this class (better to have it in the service definition)

src/Adapter/Customer/QueryHandler/GetCustomerForViewingHandler.php
```diff
@@ -90,19 +98,30 @@ final class GetCustomerForViewingHandler implements GetCustomerForViewingHandler
...
+     * @param LocaleRepository $repository
      */
     public function __construct(
         TranslatorInterface $translator,
         $contextLangId,
-        Link $link
+        Link $link,
+        LocaleRepository $repository
     ) {
+        $this->context = new LegacyContext();
```
jolelievre, line 117: Same comment as above, avoid injecting LegacyContext when you can

src/Adapter/Kpi/ShoppingCartTotalKpi.php
```diff
@@ -31,18 +31,43 @@
...
+
+    /**
+     * @var array
+     */
     private $options;
 
+    /**
+     * @param LocaleRepository $repository
+     */
+    public function __construct(LocaleRepository $repository)
```
jolelievre: Maybe it would better to inject a Locale object directly
Then you probably will need a new method in LocaleRepository, getContextLocale or something (there's probably a better solution)

jolelievre: Having such entry in a service would allow you to simplify all the classes you modified here by adding the repo and the LegacyContext And I'm sure we will need this kind of feature in the future

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -307,7 +320,8 @@ private function getOrderProducts(Order $order): OrderProductsForViewing
             $product['quantity_refundable'] = $product['product_quantity'] - $resume['product_quantity'];
             $product['amount_refundable'] = $product['total_price_tax_excl'] - $resume['amount_tax_excl'];
             $product['amount_refundable_tax_incl'] = $product['total_price_tax_incl'] - $resume['amount_tax_incl'];
-            $product['amount_refund'] = $order->getTaxCalculationMethod() ? Tools::displayPrice($resume['amount_tax_excl'], $currency) : Tools::displayPrice($resume['amount_tax_incl'], $currency);
+            $resumeAmount = $order->getTaxCalculationMethod() ? 'amount_tax_excl' : 'amount_tax_incl';
+            $product['amount_refund'] = is_null($resume[$resumeAmount]) ? $resume[$resumeAmount] : $this->locale->formatPrice($resume[$resumeAmount], $currency->iso_code);
```
jolelievre:
```suggestion
            $product['amount_refund'] = !is_null($resume[$resumeAmount]) ? $this->locale->formatPrice($resume[$resumeAmount], $currency->iso_code) : null;
```

src/Adapter/Presenter/Order/OrderDetailLazyArray.php
```diff
@@ -55,9 +68,31 @@ public function __construct(Order $order)
...
     }
 
+    /**
+     * @return LocaleRepository
+     *
+     * @throws \Exception
+     */
+    protected function getCldrLocaleRepository()
+    {
+        $context = Context::getContext();
```
jolelievre: You should use `ContainerFinder` here

jolelievre (changes requested):
Sorry I read the whole PR and found out some things that could be improved ^^
Especially you inject `prestashop.core.localization.locale.repository` in many services whereas what you actually need is a `Locale` so I think it would be very handy to have a repository or a service able to return the `Locale` object linked to the context so you can inject it directly in services when you need it

admin-dev/themes/default/template/controllers/orders/_customized_data.tpl
matthieu-rolland: You can get the display precision of a currency from the context or even directyle from the currency if you have it:

`$currency->precision`

But @jolelievre might be right about CLDR displaying the price, are you talking about a specific method ?

matthieu-rolland: ok so the formatPrice already deals with the rounding

Progi1984: :ok_hand:

controllers/admin/AdminCartsController.php
Progi1984: :ok_hand:

controllers/admin/AdminOrdersController.php
Progi1984: :ok_hand:

controllers/admin/AdminOrdersController.php
Progi1984: :ok_hand:

src/Adapter/Cart/QueryHandler/GetCartForViewingHandler.php
Progi1984: :heavy_check_mark:

src/Adapter/Customer/QueryHandler/GetCustomerForViewingHandler.php
Progi1984: :heavy_check_mark:

src/Adapter/Kpi/ShoppingCartTotalKpi.php
Progi1984: :heavy_check_mark:

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
Progi1984: :heavy_check_mark:

Progi1984:
@jolelievre I fixed your feedback. Could you check ?

admin-dev/themes/default/template/controllers/orders/_product_line.tpl
```diff
@@ -189,7 +189,10 @@
 					<input onchange="checkPartialRefundProductAmount(this)" type="text" name="partialRefundProduct[{$product['id_order_detail']}]" />
 					{if !$currencySymbolBeforeAmount}<div class="input-group-addon">{$currency->sign}</div>{/if}
 				</div>
-        <p class="help-block"><i class="icon-warning-sign"></i> {l s='(Max %amount_refundable% %tax_method%)' sprintf=[ '%amount_refundable%' => Tools::displayPrice(Tools::ps_round($amount_refundable, 2), $currency->id), '%tax_method%' => $smarty.capture.TaxMethod] d='Admin.Orderscustomers.Help'}</p>
+        <p class="help-block"><i class="icon-warning-sign"></i> {l
+            s='(Max %amount_refundable% %tax_method%)'
+            sprintf=['%amount_refundable%' => Context::getContext()->getCurrentLocale()->formatPrice(Tools::ps_round($amount_refundable, 2), Currency::getCurrencyInstance($currency->id)->iso_code), '%tax_method%' => $smarty.capture.TaxMethod]
```
jolelievre: There's still a useless round here

admin-dev/themes/default/template/controllers/orders/helpers/view/view.tpl
```diff
@@ -1111,7 +1111,7 @@
                         </div>
                         <p class="help-block"><i class="icon-warning-sign"></i> {l
                             s='(Max %s %s)'
-                            sprintf=[Tools::displayPrice(Tools::ps_round($shipping_refundable, 2), $currency->id) , $smarty.capture.TaxMethod]
+                            sprintf=[Context::getContext()->getCurrentLocale()->formatPrice(Tools::ps_round($shipping_refundable, 2), Currency::getCurrencyInstance($currency->id)->iso_code) , $smarty.capture.TaxMethod]
```
jolelievre: Useless round

classes/controller/AdminController.php
```diff
@@ -416,6 +416,9 @@ public function __construct($forceControllerName = '', $default_theme_name = 'de
         if (strpos($this->controller_name, 'ControllerOverride')) {
             $this->controller_name = substr($this->controller_name, 0, -18);
         }
+        if (strpos($this->controller_name, 'ControllerCore')) {
```
jolelievre: Mmm weird, why do you need this? Not that I don't understand the purpose but then it should have caused a few bugs long before right?

classes/order/OrderHistory.php
```diff
@@ -510,12 +510,12 @@ public function sendEmail($order, $template_vars = false)
                 $data = array_merge($data, $template_vars);
             }
 
-            $data['{total_paid}'] = Tools::displayPrice((float) $order->total_paid, new Currency((int) $order->id_currency), false);
+            $context = Context::getContext();
+            $data['{total_paid}'] = Tools::getContextLocale($context)->formatPrice((float) $order->total_paid, (new Currency((int) $order->id_currency))->iso_code);
```
jolelievre: Sometimes you use `new Currency`, sometimes `Currency::getCurrencyInstance` this lacks consistency
If anything I'd say `Currency::getCurrencyInstance` is lest worse because it stores the result and avoid creating too many instances, but in the end even this is not that great because it sill instanciate useless Currency objects
I think the real better way would be to add a `Currency::getIsoCodeById` method.. this would be more performant and even the could would look cleaner:

Instead of `(new Currency((int) $order->id_currency))->iso_code` -> `Currency::getIsoCodeById($order->id_currency)`

`Currency::getIsoCodeById` would just make a Db query, and maybe cache the result by id (on success) to avoid too many requests, and it musn't check for active or deleted status or this will cause bugs

controllers/admin/AdminOrdersController.php
```diff
@@ -194,7 +194,7 @@ public static function setOrderCurrency($echo, $tr)
             $idCurrency = (int) $order->id_currency;
         }
 
-        return Tools::displayPrice($echo, $idCurrency);
+        return (new self())->context->getCurrentLocale()->formatPrice($echo, Currency::getCurrencyInstance($idCurrency)->iso_code);
```
jolelievre: Mmm, `new self` seems like a bad idea, no need to instanciate a controller to access the context, especially since controller may change the context in their init method (reminds me of a "bug" on shop context because `LegacyContext` juste made a `new AdminController()`)
If you are in a static context then it's better (actually less worse..) to call `Context::getContext()`

jolelievre (commented):
Hi, sorry a new idea of improvement to avoid creating too many Currency instances everwhere (see in the comment about adding a `Currency::getIsoCodeById` method)

About the injection issue, I saw your modif but the thing is you still inject the repository plus a string, whereas just the Local object should enough I don't think a service to get this object exist but I thought maybe we can use symfony services and factory to our advantage, I'm not sure if the syntax is correct but something like:

```yaml
prestashop.core.localization.locale.context_locale:
        factory: ['@prestashop.core.localization.locale.repository', 'getLocale']
        class: PrestaShop\PrestaShop\Core\Localization\Locale
        arguments: ["@=service("prestashop.adapter.legacy.context").getContext().language.getLocale()"]
```

This way you could directly inject `@prestashop.core.localization.locale.context_locale` in the handlers What do you think?
This way we keep the handler constructor cleaner (only a Locale to inject), no need to create a new service, and we can also easily change the "service" in the future to get rid of the Context

jolelievre:
Oh and I didn't comment them all but there are a few useless `Tools::ps_round` still present, you can easily find them searching them in the "Files changed" section

admin-dev/themes/default/template/controllers/orders/helpers/view/view.tpl
Progi1984: Fixed

admin-dev/themes/default/template/controllers/orders/_product_line.tpl
Progi1984: Fixed

classes/order/OrderHistory.php
Progi1984: Fixed. Thanks :)

controllers/admin/AdminOrdersController.php
Progi1984: Fixed

classes/controller/AdminController.php
Progi1984: Reverted

classes/Currency.php
```diff
@@ -610,6 +612,27 @@ public static function getIdByIsoCode($isoCode, $idShop = 0, $forceRefreshCache
...
+     * @param int $id
+     * @param bool $forceRefreshCache
+     *
+     * @return string
+     */
+    public static function getIsoCodeById(int $id, bool $forceRefreshCache = false)
+    {
+        $cacheId = 'Currency::getIsoCodeById' . pSQL($id);
+        if ($forceRefreshCache || !Cache::isStored($cacheId)) {
+            $result = Currency::getCurrencyInstance($id);
```
jolelievre: Could you please make a query directly to the datatase, so that we don't instanciate a Currency object (even though `getCurrencyInstance` stores the objects and there is a cache on the result There's too much magic in the object model instanciation. So instead use Db and DbQuery object, this will minimize the data fetched as well.

src/Adapter/CombinationDataProvider.php
```diff
@@ -29,6 +29,7 @@
 use Combination;
 use PrestaShop\Decimal\Number;
 use PrestaShop\PrestaShop\Adapter\Product\ProductDataProvider;
+use PrestaShop\PrestaShop\Core\Localization\Locale;
```
jolelievre, line 32: With the new improvement you could simplify this class and inject the Locale directly Since it's an internal class I think we can change the constructor safely

src/Adapter/Customer/QueryHandler/GetCustomerForViewingHandler.php
```diff
@@ -273,7 +289,7 @@ private function getCustomerOrders(Customer $customer)
         }
 
         return new OrdersInformation(
-            Tools::displayPrice($totalSpent),
+            $this->locale->formatPrice($totalSpent, $this->context->getContext()->currency->iso_code),
```
jolelievre, line 292: I'm wondering if we shouldn't use the order currency I know it changes the initial behaviour, but maybe it was forgotten?

src/Adapter/Kpi/ShoppingCartTotalKpi.php
```diff
@@ -57,9 +72,9 @@ public function render()
         $helper->color = 'color1';
         $helper->title = $translator->trans('Total Cart', [], 'Admin.Orderscustomers.Feature');
         $helper->subtitle = $translator->trans('Cart #%ID%', ['%ID%' => $cart->id], 'Admin.Orderscustomers.Feature');
-        $helper->value = Tools::displayPrice(
+        $helper->value = $this->locale->formatPrice(
             $this->getCartTotalPrice($cart),
-            (int) $cart->id_currency
+            \Currency::getIsoCodeById((int) $cart->id_currency)
```
jolelievre: Please add a `use Currency;` instead of the `\Currency` notation

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -107,15 +113,18 @@ final class GetOrderForViewingHandler implements GetOrderForViewingHandlerInterf
      * @param ImageTagSourceParserInterface $imageTagSourceParser
      * @param TranslatorInterface $translator
      * @param Context $context
+     * @param Locale $locale
      */
     public function __construct(
         ImageTagSourceParserInterface $imageTagSourceParser,
         TranslatorInterface $translator,
-        Context $context
+        Context $context,
```
jolelievre: Since context was only used to get `$this->context->language->id` I think you can remove it

src/Adapter/Product/AdminProductWrapper.php
```diff
@@ -76,11 +84,15 @@ class AdminProductWrapper
      * Constructor : Inject Symfony\Component\Translation Translator.
      *
      * @param object $translator
+     * @param LegacyContext $legacyContext
+     * @param LocaleRepository $localeRepository
+     * @param string $locale
      */
-    public function __construct($translator, $legacyContext)
+    public function __construct($translator, $legacyContext, LocaleRepository $localeRepository, string $locale)
```
jolelievre: You can only inject the locale here. Ideally we could even remove the context and inject the employee shop list but let's calm down a bit 😅

src/PrestaShopBundle/Resources/config/services/adapter/product.yml
```diff
@@ -21,7 +21,11 @@ services:
 
     prestashop.adapter.admin.wrapper.product:
         class: PrestaShop\PrestaShop\Adapter\Product\AdminProductWrapper
-        arguments: ["@translator", "@prestashop.adapter.legacy.context"]
+        arguments:
+            - "@translator"
+            - "@prestashop.adapter.legacy.context"
+            - "@prestashop.core.localization.locale.repository"
```
jolelievre: Use `@prestashop.core.localization.locale.context_locale` instead

jolelievre (commented):
Almost there 😅

src/Adapter/CombinationDataProvider.php
Progi1984: Fixed

src/Adapter/Kpi/ShoppingCartTotalKpi.php
Progi1984: Fixed

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
Progi1984: Fixed

src/Adapter/Product/AdminProductWrapper.php
Progi1984: Double Fixed

src/PrestaShopBundle/Resources/config/services/adapter/product.yml
Progi1984: Fixed

src/Adapter/Customer/QueryHandler/GetCustomerForViewingHandler.php
Progi1984: I don't think it's a good idea. It is completely out of the scope of this PR.

Progi1984: May be you could create a dedicated issue ?

classes/Currency.php
Progi1984: I forgot it.

Progi1984: Fixed

src/Adapter/Customer/QueryHandler/GetCustomerForViewingHandler.php
matthieu-rolland: I agree with @Progi1984 , if the QA team finds an issue related to this the whole PR could be blocked for a while...

By the way are there cases when the order currency is different than the context currency ?

jolelievre: Yes, I know it's not in this PR scope But it makes me wonder it would be worth investigating on the expected result here
And yes I think in some cases the currency can be different between the order and the context, especially in an Order history If it's the order you are purchasing then it should be the same logically

jolelievre (approved):
Congratulations!!! 🎉
Nice work!

matthieu-rolland (approved):
LGTM, well done

Robin-Fischer-PS:
Hello guys ! Nice try avoiding the QA :stuck_out_tongue: 

https://github.com/PrestaShop/PrestaShop/issues/16068

(if it's not from this PR, blame @matks , he's the one who told me that)

PR #12878, opened 2019-03-12
Fix no carrier available in checkout with free shipping cart rule

djbuch:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Correction of https://github.com/PrestaShop/PrestaShop/issues/11172
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #11172
| How to test?  | 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/12878)

Issue #11172, opened 2018-10-29
No carrier available in checkout with free shipping cart rule


e-perrodon:
**EDIT**: to get the reproductible scenario, jump to https://github.com/PrestaShop/PrestaShop/issues/11172#issuecomment-502165356 ◀️ 

**Describe the bug**
When I apply a free shipping cart rule whith a code on a specific product, no carriers are available in check out, only on my production environment

**To Reproduce**
1. Update Prestashop from 1.7.4.1 to 1.7.4.2 and 1.7.4.3 with 1 click upgrade module on DEV environment.
2. Push new sources to PROD environment with git
3. Apply upgrade.php script to update database
4. Create cart rule with code and free shipping on specific country (France) and specific product
5. Order the specific product and choose a delivery address to France
6. A carrier is proposed. Apply code. Voucher is ok but no carrier are available after updating.

Steps to reproduce the issue [Khouloud]

1. Create a cart rule "Free shipping" with/without code
2. Create any cart rule without code & disable it
3. Go to the FO => add some product to the cart => complete the checkout process => in the SHIPPING METHOD step => add the code for the first cart rule (free shipping)
4. Go to BO => Enable the second cart rule
5. In step 3 => reload the page => "Unfortunately, there are no carriers available for your delivery address." is displayed

https://drive.google.com/file/d/1JWZ_iX-EpXKvwepfj9B4usDZYBBUQCBz/view
**Additionnal information**
PrestaShop version: 1.7.4.3
I haven't problem in DEV environment. If i get a dump from PROD and set in DEV, I reproduce problem.
Something must be wrong in database but I can't find what is it...

e-perrodon:
I managed to work around the problem forcing flush when calling $this->context->cart->getDeliveryOptionList in getDeliveryOptions function in classes/checkout/DeliveryOptionsFinder.php
But it's not really a solution...

khouloudbelguith:
Hi @e-perrodon,

I did not manage to reproduce the issue after upgrade from 1.7.4.2 to 1.7.4.3 using the 1-click upgrade 1.7.4.3.
I attached a video record.
https://drive.google.com/file/d/1mSCzwvkDUs7BBiXoPqzLkWDHMRhxPUwX/view
![image](https://user-images.githubusercontent.com/16067358/47641519-98b5d280-db66-11e8-8b09-483c99379770.png)
[...]

PR #14381, opened 2019-06-26
WIP Available carrier behat test

tomlev:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Available carrier list was empty under certain conditions
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/11172
| How to test?  | see ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14381)

PR #15452, opened 2019-09-06
Add failure scenario to cover PR 12878

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add a Behat test that follows bug scenario https://github.com/PrestaShop/PrestaShop/issues/11172 in order to validate potential fix https://github.com/PrestaShop/PrestaShop/pull/12878/
| Type?         | improvement
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15452)

### DIFF
--- a/tests/Integration/Behaviour/Features/Context/CartFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/CartFeatureContext.php
@@ -40,14 +40,30 @@ class CartFeatureContext extends AbstractPrestaShopFeatureContext
     protected $cart;
 
     /**
+     * @Then there is no delivery options available for my cart
+     */
+    public function noDeliveryOptions()
+    {
+        if ($this->getCurrentCart() === null) {
+            throw new \RuntimeException('No current cart, cannot check available delivery options');
+        }
+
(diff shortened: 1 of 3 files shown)


djbuch:
Tell me if I’m wrong but in this scenario you test the bug not the absence of bug ?
Be cause if the bug was not there we would have some delivery option.

Or else I didn’t understand the way we should write scenarios.

matks:
> Tell me if I’m wrong but in this scenario you test the bug not the absence of bug ?
> Be cause if the bug was not there we would have some delivery option.

I confirm ! 😄Indeed this scenario checks the behavior we know about, the one we want to fix. So we wont merge this PR like this, we'll merge it together with your PR, once updated 😉

PierreRambaud:
Could you please add the valid scenario instead based on https://github.com/PrestaShop/PrestaShop/pull/14381/files ?

matks:
Found why the test of @tomlev and my test have different outputs: function `getDeliveryOptionList()` has static cache built-in. It needs to be disabled using `getDeliveryOptionList(null, true);`

PR #15472, opened 2019-09-09
[WIP] Disable cache to fix bad delivery options computation

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | It seems https://github.com/PrestaShop/PrestaShop/issues/11172 was due to bad static cache computation https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Cart.php#L2617
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/11172
| How to test?  | Please check ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15472)

### DIFF
--- a/classes/checkout/DeliveryOptionsFinder.php
+++ b/classes/checkout/DeliveryOptionsFinder.php
@@ -72,7 +72,7 @@ class DeliveryOptionsFinderCore
 
     public function getDeliveryOptions()
     {
-        $delivery_option_list = $this->context->cart->getDeliveryOptionList();
+        $delivery_option_list = $this->context->cart->getDeliveryOptionList(null, true);
         $include_taxes = !Product::getTaxCalculationMethod((int) $this->context->cart->id_customer) && (int) Configuration::get('PS_TAX');
         $display_taxes_label = (Configuration::get('PS_TAX') && !Configuration::get('AEUC_LABEL_TAX_INC_EXC'));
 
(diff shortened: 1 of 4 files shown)


matks:
It seems to solve https://github.com/PrestaShop/PrestaShop/issues/11172 but I'm afraid of the change I introduce 🤔because it might lead to over-computations and resources allocation issues.

matks:
I added the behat scenario from https://github.com/PrestaShop/PrestaShop/pull/15452 to validate the change

matks:
> It seems to solve #11172 but I'm afraid of the change I introduce 🤔because it might lead to over-computations and resources allocation issues.

I guess it would be better to properly invalidate the cache when a a Cart Rule is modified

matks:
It seems my new behat scenario breaks the `1 product in cart, 1 cart rule` scenario of feature `Check cart to order data copy` (cart_to_order.feature) when launched before
=> something is not reset between the 2 scenarios

@tomlev you can laugh now

matks:
I confirm this PR works although it is not the best solution: if we follow the scenario #11172 then on the last HTTP request, the function `Cart::getDeliveryOptionList()` will be called multiple times and the result will be stored inside its static cache `static::$cacheDeliveryOptionList`.

However the multiple calls to `Cart::getDeliveryOptionList()` **do not call the Cart in the same state** resulting in a bad result being stored in cache and wrongfully returned on following calls. That's why not using the cache works.

Ideal solution: fix the cache strategy for this function so it does not return the same result to 2 different calls.
=> very complex as the result of this function is tied to a huge number of parameters, it would require to trace them all

The current PR forces the computation (it disables the cache) for `DeliveryOptionsFinder::getDeliveryOptions()` so the cache does not create bad results. But it might introduce performance issues on FO

PR #15517, opened 2019-09-12
Do not allow to put item in cart if available quantity is 0

### LINKED ISSUE #13374: Out of stock product is still added to the cart
**Describe the bug**
Add to cart an out of stock product (order refused) return an error but the product is still added to the cart

**To Reproduce**
1. Go to a product page
2. Change quantity to zero without reloading 
3. Add the product to the cart
4. See errror in the chrome inspector

**Screenshots**
see screenshot
http://creabilis.com/tmp/addtocartbug.jpg

**Additionnal information**
PrestaShop version: 1.7.6 develop branch
PHP version: 7.0

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | If available quantity is 0, do not add cart to product
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/13374
| How to test?  | See ticket, be careful about different "out of stock" behaviors (see screenshot below)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15517)

### DIFF
--- a/controllers/front/CartController.php
+++ b/controllers/front/CartController.php
@@ -587,7 +587,7 @@ class CartControllerCore extends FrontController
             $this->customization_id
         );
 
-        return $productQuantity < 0;
+        return $productQuantity <= 0;
     }
 
     /**


matks:
Capture d’écran 2019-09-12 à 23 14 17

jolelievre (approved):
Seems too easy ^^
Nowhere else in the code could we have this kind of checks?

matks:
> Seems too easy ^^

I agree with you 🤔that's why I think it was intentional.
In a e-commerce, you can always choose between overbook and underbook: when someone orders "the last item of product X" you have 2 choices: refuse to add it to the cart or add it in the cart *then* tell the customer it's not available anymore, either in the cart page or at the checkout

The idea is that, if you have 3 customers wanting to get all the last item, you prefer to overbook because you dont want customer 1 to change his mind but customer 2 could not add to the cart the product so in the end no one buys.

So I think this is not a bugfix but rather a change of the selling strategy

PierreRambaud (changes requested):
Things are weird. We have `$qtyToCheck` and it's unused
The method name check if an error should be raised and not if the availability is valid.
By changing this, we'll have some trouble.

jolelievre:
@matks I think the feature you are talking about is managed via `Product::isAvailableWhenOutOfStock` so it is covered which allows to order an out of stock, although when you check its behaviour it's weird it checks if `$product->out_of_stock == 2` ...
As for the use case where several customers are trying to purchase the same product I'm not sure but I think the quantity of the stock is only modified when the order is completed no?

And I agree with @PierreRambaud that `$qtyToCheck` should be used, it is used for product with attribute in `Attribute::checkAttributeQty`

So in a nutshell I think your modification should be something like `return $qtyToCheck <= $productQuantity`

matks:
Comment ` // product quantity is the available quantity after decreasing products in cart` on line 581 seems to be false 🤔

PierreRambaud:
@matks We need to be sure on what we are trying to do here.
I'm afraid about confusion in function names, or even in what the function does :sweat_smile:

jolelievre:
@matks oohh ncie catch, actually it IS correct! since the cart is passed as an argument to `Product::getQuantity` the method automatically computes the substraction between the available stock and the number of products in the cart, that's why `$qtyToCheck` is not used

So I think you can update the comment to clarify this behavior, and the modification you made is correct `$productQuantity <= 0` Either leave it that way or you can remove the cart argument and compute it yourself

By the way I think using `$this->id_product_attribute` as an argument is not much useful either, since it's already checked above

jolelievre:
@matks So I think the only missing thing is a more detailed comment and then @PierreRambaud can approve the PR

matks:
I want to check something in there.
This fixes the usecase "I display the Product page, then someones in the BO modifies the quantity to 0, then I click on 'add to cart'" but I want to be sure I did not break the classic usecase "someone wants to order the last item"

matks:
Sadly, I confirm my fear : with my modification you cannot order "the last item" in the Cart page.

If product X has 3 items
And you add 1 item of product X in your Cart
Then go to Cart page
Then increase the quantity to 3
Then an error is displayed although you should be able to order 3. You should not be able to order 4, but 3 is OK:
Capture d’écran 2019-09-19 à 17 28 43

matks:
Closed, continued in https://github.com/PrestaShop/PrestaShop/pull/15623

PR #15530, opened 2019-09-13
Disable old selenium tests

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Disable old selenium tests that are always failing and will be removed anyway
| Type?         | bug fix
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | n/a
| How to test?  | Tests should be green

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15530)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,6 @@ matrix:
   include:
     - php: 7.2
       env: EXTRA_DEPS=phpHigh  PRESTASHOP_TEST_TYPE=unit
-    - php: 7.2
-      env: PRESTASHOP_TEST_TYPE=e2e EXTRA_TESTS=functional
     - stage: deploy
       php: 7.2
       before_install: skip
@@ -56,7 +54,7 @@ matrix:
         on:
           all_branches: true
   exclude:
-    - php: 7.2 # Replaced with additional tests
+    - php: 5.6 # Only run e2e on php 7.2
       env: PRESTASHOP_TEST_TYPE=e2e
   allow_failures:
     - php: 7.2
--- a/tests/check_e2e.sh
+++ b/tests/check_e2e.sh
@@ -1,13 +1,9 @@
 #!/bin/bash
 
-bash travis-scripts/run-selenium-tests;
-SELENIUM=$?
-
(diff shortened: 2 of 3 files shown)


matks (dismissed):

PierreRambaud (dismissed):

.travis.yml
```diff
@@ -56,7 +54,7 @@ matrix:
         on:
           all_branches: true
   exclude:
-    - php: 7.2 # Replaced with additional tests
+    - php: 5.6 # Only run e2e on php 7.2
```
Quetzacoalt91, line 57:
```suggestion
    - php: 5.6 # Only run e2e with functional tests on php 7.2
```

eternoendless: From what I understood e2e were the old Selenium tests and Functional were the new e2e. Since the old test will be deleted I suppose it's better just to call them e2e, don't you think?

Quetzacoalt91: I don't disagree with that, but the comment you updated does not make any sense. By looking at the diff I see "Run e2e tests on PHP 5.6" while the comment says the opposite.

matks:
Tried to check why Travis fails in localhost but go travis env issues

I added https://github.com/PrestaShop/PrestaShop/pull/15530/commits/5404a781f8d8b6e7c0dce32401613c76db90c104 to see if we get more infos

matks:
Closing and re-opening to re-trigger Travis ..?

.travis.yml
eternoendless: > By looking at the diff I see "Run e2e tests on PHP 5.6"

Where? O_o

Quetzacoalt91: I mean that's what I understand :laughing:

matks:
@eternoendless Please have a look at https://github.com/PrestaShop/PrestaShop/pull/15589, I obtained the expected results in an other way

matks:
@eternoendless I think we can close this PR ?

eternoendless:
Done in #15590

PR #8315, opened 2017-09-07
Prevent an added product to an order from showing in the FO cart as new order

hibatallahAouadni:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.6.1.x
| Description?  | add a condition id_guest not equal 0 to the WHERE clause, to differenciate who add the product to the order/cart, if it's an admin then the id_guest=0 so we don't display the added product.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | http://forge.prestashop.com/browse/PSCSX-8554 Fixes #15521
| How to test?  | 1- Add a new product to an existing order from admin (BO). </br>2- Sign in in the FO and open cart. </br>3- The added product appears in your cart as a new order which makes confusion to the customer.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/8315)

PR #15533, opened 2019-09-13
Disallow showing bad cart to FO user

### LINKED ISSUE #15521: Products are displayed in the shopping cart after editing an order
**Describe the bug**
When we enable the option "Re-display cart at login" in the BO > shop Parameters > Customer Settings>  Customers page, then we add a product to an existing order through admin, PS creates a new cart for this product. Then when customer logs in to frontend, they can see that product sitting in their cart, even though it has already been added to their order through admin.

I would consider this either a bug or a confusing functionality that needs addressing.

**To Reproduce**
1- Add a new product to an existing order from admin (BO).
2- Sign in in the FO and open cart
3- The added product appears in your cart as a new order which makes confusion to the customer.

**Screenshots**
https://drive.google.com/file/d/1ntEhFrhdSgcJIgnOjCYtvxpNQxaguOC5/view

**Additional information**
PrestaShop version: 1.7.6.1
PHP version: 7.2

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | When "add a product to order" is used from the BO (on order page), a fake Cart is generated, probably because of https://github.com/PrestaShop/PrestaShop/issues/14131. This PR avoids the FO user to be mistakenly showed this cart when "Re-display cart at login" option is enabled. It's a port of 1.6 PR https://github.com/PrestaShop/PrestaShop/pull/8315
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/15521
| How to test?  | See ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15533)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -3999,6 +3999,7 @@ class CartCore extends ObjectModel
                 WHERE NOT EXISTS (SELECT 1 FROM ' . _DB_PREFIX_ . 'orders o WHERE o.`id_cart` = c.`id_cart`
                                     AND o.`id_customer` = ' . (int) $id_customer . ')
                 AND c.`id_customer` = ' . (int) $id_customer . '
+                AND c.`id_guest` != 0
                     ' . Shop::addSqlRestriction(Shop::SHARE_ORDER, 'c') . '
                 ORDER BY c.`date_upd` DESC';
 


matks:
I am not a fan of this solution. It uses that BO-generated carts have `id_guest` set to 0 to filter them and exclude them. But the real issue is that fake carts are generated BO-side to perform the "add product to order" action. The real issue is then https://github.com/PrestaShop/PrestaShop/issues/14131

However:
- this PR avoids bad consequences on FO
- it seems the fake cart cannot even be deleted by FO user, so it's worse than I thought 😱 

That's why I think it's worth it to merge this PR although it hides the issue, it does not solve it.

classes/Cart.php
```diff
@@ -3999,6 +3999,7 @@ public static function lastNoneOrderedCart($id_customer)
                 WHERE NOT EXISTS (SELECT 1 FROM ' . _DB_PREFIX_ . 'orders o WHERE o.`id_cart` = c.`id_cart`
                                     AND o.`id_customer` = ' . (int) $id_customer . ')
                 AND c.`id_customer` = ' . (int) $id_customer . '
+                AND c.`id_guest` <> 0
```
PierreRambaud: Can we don't mix between `!=` and `<>`. We often use `!=`

matks: Applied and commit ammended

PierreRambaud (approved):

PierreRambaud:
Thanks @matks

LorenzoMinciotti:
This solution doesn't work as expected if you disable "statsdata" module. With "statsdata" module disabled, no one record is inserted in ps_guest table and id_guest property of cookie is 0. When in FO user add product to cart, the cart record in ps_cart has id_guest = 0 and the user cannot view the old cart if logged out.

daresh:
I confirm having a problem with loading carts after login even with the statsdata module enabled. So this pul request seems to have some bad side effects.

PR #15537, opened 2019-09-13
Trim unnecessary white spaces from form fields

rdy4ever:
| Questions     | Answers
| Branch?       | develop
| Description?  | When creating a customer account, or adding a new address, some customers add unnecessary white spaces before or after the value in form fields. This makes all related e-mails look a bit unprofessional: if you have a greeting like "Hi Name, thank you for...", it would look like "Hi Name , thank you for..." (notice the white space after Name and before the comma). This PR just gets rid of these unwanted white spaces, before submitting the form (actually imediately after the input loses focus).
| Type?         | improvement
| Category?     | FO
| How to test?  | Try to create an account and add lots of white spaces before and after first name and last name. With this PR they should be trimmed on blur.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15537)


matthieu-rolland (commented):
Thanks for your contribution @rdy4ever !

I think that this trim should be applied in backend as well (even though it is nice to have it in frontend too), what do you think?

rdy4ever:
> Thanks for your contribution @rdy4ever !
> 
> I think that this trim should be applied in backend as well (even though it is nice to have it in frontend too), what do you think?

Hi @matthieu-rolland 
Yes, it could also be usefull in the BO too, but as customers are more prone to this kind of errors then shop admins I'm inclined to only do this in the FO for now, and just client-side so the change is kept as small as possible.

rdy4ever:
I think we should trim unnecessary white spaces from all text fields, like address and such.

matthieu-rolland:
> > Thanks for your contribution @rdy4ever !
> > I think that this trim should be applied in backend as well (even though it is nice to have it in frontend too), what do you think?
> 
> Hi @matthieu-rolland
> Yes, it could also be usefull in the BO too, but as customers are more prone to this kind of errors then shop admins I'm inclined to only do this in the FO for now, and just client-side so the change is kept as small as possible.

Hello @rdy4ever

I was talking about the backend (the PHP if you will), not the back office ;)

As a general rule in web development: the best practice when it comes to cleaning user input is to always do it in the backend (php side):

- If for some reasons the javascript is broken on this page, or someone sends data without doing it from the form, then the clean will not be applied. Doing it php side will ensure the trim is always applied.

- People who don't use the classic theme will not have your improvement, if you do it php side, it will benefit every shops, not only those who use the classic theme.

Don't hesitate to ask if you need help for doing it php side 👍

rdy4ever:
> > > Thanks for your contribution @rdy4ever !
> > > I think that this trim should be applied in backend as well (even though it is nice to have it in frontend too), what do you think?
> > 
> > 
> > Hi @matthieu-rolland
> > Yes, it could also be usefull in the BO too, but as customers are more prone to this kind of errors then shop admins I'm inclined to only do this in the FO for now, and just client-side so the change is kept as small as possible.
> 
> Hello @rdy4ever
> 
> I was talking about the backend (the PHP if you will), not the back office ;)
> 
> As a general rule in web development: the best practice when it comes to cleaning user input is to always do it in the backend (php side):
> 
> * If for some reasons the javascript is broken on this page, or someone sends data without doing it from the form, then the clean will not be applied. Doing it php side will ensure the trim is always applied.
> * People who don't use the classic theme will not have your improvement, if you do it php side, it will benefit every shops, not only those who use the classic theme.
> 
> Don't hesitate to ask if you need help for doing it php side 👍

I understand what you're saying, but with js disabled the checkout on Prestashop is pretty much... dead, you can't even add items to your cart, to begin with :) 
So I would'n go that far just for a little visual fix - I think a simple js helper to remove these spaces silently is enough.

ttoine:
As Matthieu already told, this fix would only apply to merchants using the classic theme, or a theme child based on it.
other themes (and there are like thousands) will not have this fix. plus, there are many other ways to send this kind of information the PrestaShop engine (the backend)

that's why it would be better to do that in the PHP. (in the engine/backeng

PR #15544, opened 2019-09-15
Apply free delivery voucher on Delievery step

### LINKED ISSUE #15458: The free delivery voucher is neither shown nor applied on Delivery step
### Describe the bug
The free delivery voucher is neither shown nor applied on Delivery step

### Expected behavoir
the selected carrier should be shown as free and the discound should be applied to the cart.

### Steps to reproduce
1. BO > Create two or more carriers. 
2. Create a voucher with free shipping option and select one of those carriers in the voucher settings. So this voucher should affect only one of those carriers. 
3. FO > Create an order in FO, 
4. proccess it to the step Address, select Address, 
5. go to the next step. 
6. You expect to see that selected in voucher settings carrier is free and expect to appear the discount in the cart. 
7. It will happend if select other carrier (click) and select the desired carrier agian or to proccess to the next step.

I created a pull request https://github.com/PrestaShop/PrestaShop/pull/15457

### Addtional infos
- PS version: 1.7.6.1
- PHP version: 7.2

LedCloud:
| Questions     | Answers
| Branch?       | develop 
| Description?  | Free delivery voucher is applied only after CheckoutDeliveryStep or after customer click other but default carrier on this step.improvement
| Type?         |  improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15458 
| How to test?  | Create two or more carriers. Create a voucher with free shipping option and select one of those carriers in the voucher settings. So this voucher should affect only one of those carriers. Create an order in FO, proccess it to the step Address, select Address, go to the next step. You expect to see that selected in voucher settings carrier is free and expect to appear the discount in the cart. It will happend if select other carrier and select the desired carrier agian or to proccess to the next step.

This is besause the carrier ID is not yet set. So I set the default carrier ID to cart right after the address step is complete.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15544)

### DIFF
--- a/classes/checkout/CheckoutAddressesStep.php
+++ b/classes/checkout/CheckoutAddressesStep.php
@@ -302,4 +302,19 @@ class CheckoutAddressesStepCore extends AbstractCheckoutStep
             $this->getTemplateParameters()
         );
     }
+    
+    public function setComplete($step_is_complete)
+    {
+        if ($step_is_complete && $this->context->cart->id_carrier == 0) {
+            $deliveryOptionSelected = $this->getCheckoutSession()->getSelectedDeliveryOption();
+            $id_address = $this->context->cart->id_address_delivery;
+
+            if ($deliveryOptionSelected) {
+                $this->getCheckoutSession()->setDeliveryOption(array(
+                    $id_address => $deliveryOptionSelected,
+                ));
+            }
+        }
+        return parent::setComplete($step_is_complete);
+    }
 }
--- a/classes/checkout/CheckoutAddressesStep.php
+++ b/classes/checkout/CheckoutAddressesStep.php
@@ -308,15 +308,17 @@ class CheckoutAddressesStepCore extends AbstractCheckoutStep
     public function setComplete($step_is_complete)
     {
         if ($step_is_complete &&
-                $this->context->cart->id_carrier == 0 &&
-                ($deliveryOptionSelected =
-                    $this->getCheckoutSession()->getSelectedDeliveryOption())) {
-            $id_address = $this->context->cart->id_address_delivery;
-            $this->getCheckoutSession()->setDeliveryOption(
-                array(
-                    $id_address => $deliveryOptionSelected
-                )
-            );
+                $this->context->cart->id_carrier == 0) {
+            $deliveryOptionSelected =
+                    $this->getCheckoutSession()->getSelectedDeliveryOption();
+            if ($deliveryOptionSelected) {
+                $id_address = $this->context->cart->id_address_delivery;
+                $this->getCheckoutSession()->setDeliveryOption(
+                    array(
+                        $id_address => $deliveryOptionSelected
+                    )
+                );
+            }
         }
         
         return parent::setComplete($step_is_complete);
--- a/classes/checkout/CheckoutAddressesStep.php
+++ b/classes/checkout/CheckoutAddressesStep.php
@@ -81,10 +81,12 @@ class CheckoutAddressesStepCore extends AbstractCheckoutStep
         }
 
         // Can't really hurt to set the firstname and lastname.
-        $this->addressForm->fillWith(array(
+        $this->addressForm->fillWith(
+            array(
             'firstname' => $this->getCheckoutSession()->getCustomer()->firstname,
             'lastname' => $this->getCheckoutSession()->getCustomer()->lastname,
-        ));
+            )
+        );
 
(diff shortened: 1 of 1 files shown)


classes/checkout/CheckoutAddressesStep.php
```diff
@@ -302,4 +302,20 @@ public function render(array $extraParams = array())
             $this->getTemplateParameters()
         );
     }
+    
+    public function setComplete($step_is_complete)
+    {
+        if ($step_is_complete && $this->context->cart->id_carrier == 0) {
+            $deliveryOptionSelected = $this->getCheckoutSession()->getSelectedDeliveryOption();
+            $id_address = $this->context->cart->id_address_delivery;
```
PierreRambaud: Could you move the `$id_address` variable in the second if?

LedCloud: It is too complicated to read now, isn't it?

PierreRambaud: Yes it is. Sorry I misspoke.
I mean 

```
if ($step_is_complete && ....) {
  $id_address = ...;
  $this->getCheckoutSession()->setDeliveryOption(
  ...
}
```

LedCloud: You wrote clearly and I get you right. But after moving the assigment of $id_address it became obvios that one if could be used. :)
I'm redoing.

classes/checkout/CheckoutAddressesStep.php
```diff
@@ -81,10 +81,12 @@ public function handleRequest(array $requestParams = array())
         }
 
         // Can't really hurt to set the firstname and lastname.
-        $this->addressForm->fillWith(array(
+        $this->addressForm->fillWith(
+            array(
```
PierreRambaud, line 85: indentation is broken, could you run php-cs-fixer please?

classes/checkout/CheckoutAddressesStep.php
```diff
@@ -302,4 +304,23 @@ public function render(array $extraParams = array())
...
         );
     }
+    
+    public function setComplete($step_is_complete)
+    {
+        if ($step_is_complete &&
+                $this->context->cart->id_carrier == 0) {
+            $deliveryOptionSelected =
+                    $this->getCheckoutSession()->getSelectedDeliveryOption();
+            if ($deliveryOptionSelected) {
```
jolelievre, line 314: You set the carrier in the address step, but how comes you already have the delivery option available?
Is it because you applied the voucher first? and it sets the `$this->getCheckoutSession()->getSelectedDeliveryOption()` option?

LedCloud: I set the default address if it is available. The same method is being invoked on the next step, so I just speed up the process.

classes/checkout/CheckoutAddressesStep.php
LedCloud: @PierreRambaud it wasn't me :) it was php-cs-fixer.
I'll redo later.

matks:
Hi, I think something wrong happened to this PR 🤔  because @jolelievre mentions items I dont see in the diff.

Anyway, I am not sure this is a good thing to do 🤔  if I understand correctly, this PR adds some logic into `setComplete()` which is a [mutator](https://en.wikipedia.org/wiki/Mutator_method). Is that the right place to do this ? Most people expect mutators to ... just control the change of a property. Maybe format it a little. But this extra logic, I dont expect it to happen here.

If the idea of this PR is to perform a series of steps to "complete" the Adresses step, may I suggest rather a new function `complete()` or `acknowledgeCompletion()` that do multiple things including the `setComplete()` ?

⬇️ 

```
public function acknowledgeCompletion(...)
{
    // do the logic about carrier ID
    // do other stuff

    $this->setComplete($complete);
}
```

LouiseBonnard:
@LedCloud, any news on this?

LedCloud:
@LouiseBonnard  I haven't heard any.

LouiseBonnard:
Hi @LedCloud, thanks for you reply! What is the status of this PR? I can see that there are some comments left, is there any work left to do? Once resolved, maintainers will be able to review your contribution. :-)

matks:
Hello, it looks like this PR is not moving at all and we get no answer from author.

I close the PR as stale. If however you wish to continue working on it please tell us ;) we can reopen

PR #15549, opened 2019-09-16
Fix calculation price in group reductions when in tax excluded mode

### LINKED ISSUE #9929: [BOOM-3580] Quantity discount wrong calculation when price display method is tax excluded
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-3580).

- _**Reporter:**_ coralgol
- _**Created at:**_ Wed, 2 Aug 2017 10:14:02 +0200

<p>When price display method is set to net, qunatity discounts are not properly calculated. The calculation method mixes amounts with tax and without tax.</p>
        
1) Set &quot;price display method&quot; to net for a particular group of users (e.g. all Visitors)
<br/>
2) Create product with net price 10 EUR and VAT 23%
<br/>
3) For the product create two specific prices: 
<br/>
&nbsp;&nbsp;&nbsp;&nbsp;a) for the first: set quantity = 10 and net price 8 EUR
<br/>
&nbsp;&nbsp;&nbsp;&nbsp;a) for the second: set quantity = 20 and net price 7 EUR
<br/>
4) So we should have the following rule:
<br/>
from 1 to 9 items = 10 EUR netto, 12.30 including tax
<br/>
from 10 to 19 items = 8 EUR netto, 9.84 including tax
<br/>
from 20 items = 7 EUR netto, 8.61 including tax
<br/>
5) save product and go to product view 
<br/>
6) in the product discount table you will see
<br/>
$quantity_discount.quantity|$quantity_discount.discount|$quantity_discount.save
<br/>
10|0.16|1.60
<br/>
20|1.39|13.90
<br/>
Which is WRONG, because the correct values are:
<br/>
10|-2|-20
<br/>
20|-3|-30
<br/>
the problem is in calculation - it mixes values with and without tax. As you can see 0.16 = 10-9.84 which is (base net price) - (discounted price with tax)
<br/>

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix wrong price calculation in grouped buy reduction mode, when taxes are excluded.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9929
| How to test?  | In back office, create a product, with a price of 10 euros HT. Then create a special price of 8 euros, starting at 10 products. And another special price of 7 euros, starting at 20 products. Then set the price display to HT (tax excluded). In front office, the product should display correct price reductions.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15549)


matks:
@matthieu-rolland do you think it would be possible to add a Behat test to cover this usecase ?

matthieu-rolland:
> @matthieu-rolland do you think it would be possible to add a Behat test to cover this usecase ?

Sure, great idea, I'll do that !

controllers/front/ProductController.php
```diff
@@ -890,7 +890,8 @@ protected function formatQuantityDiscounts($specific_prices, $price, $tax_rate,
                 // The price may be directly set
 
                 /** @var float $currentPriceDefaultCurrency current price with taxes in default currency */
-                $currentPriceDefaultCurrency = (!$row['reduction_tax'] ? $row['price'] : $row['price'] * (1 + $tax_rate / 100)) + (float) $ecotax_amount;
+                $currentPriceDefaultCurrency = $this->calculatePriceDefaultCurrency($row, $tax_rate, $ecotax_amount);
```
Progi1984: @matthieu-rolland You should send values in parameters and not array. It's not the responsibility of the function to check is the value is in the array.

matthieu-rolland: done !

controllers/front/ProductController.php
```diff
@@ -890,7 +890,8 @@ protected function formatQuantityDiscounts($specific_prices, $price, $tax_rate,
                 // The price may be directly set
 
                 /** @var float $currentPriceDefaultCurrency current price with taxes in default currency */
-                $currentPriceDefaultCurrency = (!$row['reduction_tax'] ? $row['price'] : $row['price'] * (1 + $tax_rate / 100)) + (float) $ecotax_amount;
+                $currentPriceDefaultCurrency = $this->calculatePriceDefaultCurrency($row['price'], $row['reduction_tax'], $tax_rate, $ecotax_amount);
```
PierreRambaud:
```suggestion
                $currentPriceDefaultCurrency = $this->calculatePriceDefaultCurrency(
                    $row['price'], 
                    $row['reduction_tax'], 
                    $tax_rate, $ecotax_amount
                );
```

controllers/front/ProductController.php
```diff
@@ -954,6 +955,25 @@ protected function formatQuantityDiscounts($specific_prices, $price, $tax_rate,
...
     }
 
+    /**
+     * @param float $specificPriceValue
+     * @param mixed $specificPriceReductionTax
+     * @param float $ecoTaxAmount
+     *
+     * @return mixed
+     */
+    private function calculatePriceDefaultCurrency($specificPriceValue, $specificPriceReductionTax, $taxRate, $ecoTaxAmount)
```
PierreRambaud:
```suggestion
    private function calculatePriceDefaultCurrency(
        $specificPriceValue, 
        $specificPriceReductionTax, 
        $taxRate, 
        $ecoTaxAmount
    ) {
```

controllers/front/ProductController.php
```diff
@@ -954,6 +955,25 @@ protected function formatQuantityDiscounts($specific_prices, $price, $tax_rate,
...
+    {
+        $priceDefaultCurrency = $specificPriceValue;
+
+        // only apply tax rate when tax calculation method is set to PS_TAX_INC
+        if (Product::$_taxCalculationMethod == PS_TAX_INC && $specificPriceReductionTax) {
```
PierreRambaud:
```suggestion
        if (Product::$_taxCalculationMethod === PS_TAX_INC && $specificPriceReductionTax) {
```

matthieu-rolland: This is legacy code that I extracted in a function...  (I mean this is how this condition is checked in this controller) I really don't know if Product::$_taxCalculationMethod will always be an integer (and not, let's say, a string).

It felt safer to keep the legacy behavior in this condition. I'll check that out.

PierreRambaud: This value can only be an integer (PS_TAX_EXC and PS_TAX_INC) / null (intiliazed as property):
Feel free to decide :sweat_smile:

matthieu-rolland: Ok, from what I checked in the codebase, it seems that this variable is always assigned an integer. Done.

controllers/front/ProductController.php
```diff
@@ -954,6 +960,29 @@ protected function formatQuantityDiscounts($specific_prices, $price, $tax_rate,
         return $specific_prices;
     }
 
+    /**
+     * @param float $specificPriceValue
+     * @param bool $specificPriceReductionTax
+     * @param float $ecoTaxAmount
```
PierreRambaud: `$taxRate` is missing.

controllers/front/ProductController.php
```diff
@@ -954,6 +960,29 @@ protected function formatQuantityDiscounts($specific_prices, $price, $tax_rate,
...
+     *
+     * @return mixed
+     */
+    private function calculatePriceDefaultCurrency(
```
PierreRambaud:
```suggestion
    private function calculatePriceDefaultCurrencyAmount(
```

You're not calculating a currency. :thinking: 
Wdyt @matks?

matthieu-rolland: It only calculates a price, in default currency because we take data from a specific price without applying another currency.

I gave this name according to the variable it fills in the parent function, but you are right, this function has no knowledge of any currency matters. I think I should change that. Let me think of a better name.

matthieu-rolland: I called it calculateSpecificPrice, because in the end this is what it really does, for a specific price basis, it calculates its final value (with or without tax rate and ecotax)

controllers/front/ProductController.php
matthieu-rolland: yep, done

controllers/front/ProductController.php
```diff
@@ -954,6 +960,30 @@ protected function formatQuantityDiscounts($specific_prices, $price, $tax_rate,
...
+     *
+     * @return mixed
+     */
+    private function calculateSpecificPrice(
+        $specificPriceValue,
+        $specificPriceReductionTax,
+        $taxRate,
+        $ecoTaxAmount
+    ) {
+        $priceDefaultCurrency = $specificPriceValue;
```
PierreRambaud: You forget to change the variable name :trollface:

matthieu-rolland: 👍 :trollface:

src/Core/SpecificPrice/SpecificPriceFormatter.php
```diff
@@ -0,0 +1,182 @@
+<?php
```
matthieu-rolland: This class is an extraction of the legacy function **formatQuantityDiscounts**, it is still legacy code, I extracted it so I could pass it some values has dependency, especially the **isTaxIncluded** boolean, so it can be tested with and without tax.

tests-legacy/Unit/Core/SpecificPrice/SpecificPriceFormatterTest.php
```diff
@@ -24,39 +24,29 @@
  * International Registered Trademark & Property of PrestaShop SA
  */
```
matthieu-rolland: This is a legacy test that I moved in a new test file (since the code tested is in another class).
I just added some testing without taxes included.

matthieu-rolland:
> @matthieu-rolland do you think it would be possible to add a Behat test to cover this usecase ?

So I updated the unit test of the method calculating the discount value. It now both tests tax included and tax excluded.

About the behat test, I started doing it, but I realised it was testing only this function, which is exactly what the unit test is already doing.
A proper behat test would imply starting from the product, its specific prices, testing the currencies, the language locale, the context... and many other things. Which I think is way out of the scope of this PR.

src/Core/SpecificPrice/SpecificPriceFormatter.php
```diff
@@ -0,0 +1,182 @@
...
+    private $context;
+
+    /**
+     * SpecificPriceFormatter constructor.
+     *
+     * @param array $specificPrice
+     * @param bool $isTaxIncluded
+     * @param Context $context
+     */
+    public function __construct(array $specificPrice, bool $isTaxIncluded, $context)
```
Progi1984: @matthieu-rolland It lacks
```suggestion
    public function __construct(array $specificPrice, bool $isTaxIncluded, Context $context)
```

src/Core/SpecificPrice/SpecificPriceFormatter.php
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * This is legacy code extracted from ProductController and it should be refactored
+     *
+     * @param float $initialPrice
+     * @param float $tax_rate
+     * @param float $ecotax_amount
+     *
+     * @return array
+     */
+    public function formatSpecificPrice($initialPrice, $tax_rate, $ecotax_amount)
```
Progi1984: @matthieu-rolland Isn't it possible to divide this complex method into small private methods ?

tests-legacy/Unit/Core/SpecificPrice/SpecificPriceFormatterTest.php
```diff
@@ -0,0 +1,244 @@
...
+
+        $language = new Language();
+        $language->iso_code = 'EN';
+        $language->locale = 'en-US';
+        $context->language = $language;
+
+        $specificPriceFormatter = new SpecificPriceFormatter($specificPrices[0], $isTaxIncluded, $context);
+        $formattedSpecificPrice = $specificPriceFormatter->formatSpecificPrice($price, $taxRate, $ecotaxAmount);
+
+        $priceFormatter = new PriceFormatter();
```
jolelievre: I'm not sure I understand this test You actually compare both behaviours and test if they match?
But then it's not really a unit test

src/Core/SpecificPrice/SpecificPriceFormatter.php
matthieu-rolland: @Progi1984 
As I wrote earlier, this is a legacy method that I put in a dedicated class in order to be able to inject the "taxincluded" boolean as a dependancy, making it possible to fix the unit test too.

The goal of this PR is not to refactor this part, I tried though, but it's too complex and way out of the scope of this PR. I only refactored the small part that was causing the bug.

tests-legacy/Unit/Core/SpecificPrice/SpecificPriceFormatterTest.php
matthieu-rolland: @jolelievre  There is only one method tested, it's a legacy method that returns an array, I'm checking that the **discount** and **save** values are each matching their respective expectations coming from the data provider.

Behaviors are specified in the provider (if we can call that a behavior). This is a legacy test that I had to move because I moved the method that is tested, the logic has not changed, I just added cases in the data provider, and changed the way the method is called, because it has moved in its own class.

src/Core/SpecificPrice/SpecificPriceFormatter.php
```diff
@@ -0,0 +1,182 @@
...
+        $taxRate,
+        $ecoTaxAmount
+    ) {
+        $specificPrice = $specificPriceValue;
+
+        // only apply tax rate when tax calculation method is set to PS_TAX_INC
+        if ($this->isTaxIncluded && $specificPriceReductionTax) {
+            $specificPrice = $specificPrice * (1 + $taxRate / 100) + (float) $ecoTaxAmount;
+        }
+
```
matthieu-rolland: This is the actual fix

src/Core/SpecificPrice/SpecificPriceFormatter.php
jolelievre: @matthieu-rolland Indeed just this modification left and we're good to go ;)

matthieu-rolland: done !

src/Core/SpecificPrice/SpecificPriceFormatter.php
```diff
@@ -0,0 +1,182 @@
...
+    private $context;
+
+    /**
+     * SpecificPriceFormatter constructor.
+     *
+     * @param array $specificPrice
+     * @param bool $isTaxIncluded
+     * @param Context $context
+     */
+    public function __construct(array $specificPrice, bool $isTaxIncluded, \Context $context)
```
PierreRambaud: Use namespace instead, better for refactoring :)

```suggestion
    public function __construct(array $specificPrice, bool $isTaxIncluded, Context $context)
```

src/Core/SpecificPrice/SpecificPriceFormatter.php
```diff
@@ -0,0 +1,182 @@
...
+            if ($this->specificPrice['reduction_type'] == 'amount') {
+                $currentPriceCurrentCurrency -= ($this->specificPrice['reduction_tax'] ? $this->specificPrice['reduction'] : $this->specificPrice['reduction'] / (1 + $tax_rate / 100));
+                $this->specificPrice['reduction_with_tax'] = $this->specificPrice['reduction_tax'] ? $this->specificPrice['reduction'] : $this->specificPrice['reduction'] / (1 + $tax_rate / 100);
+            } else {
+                $currentPriceCurrentCurrency *= 1 - $this->specificPrice['reduction'];
+            }
+            $this->specificPrice['real_value'] = $initialPrice > 0 ? $initialPrice - $currentPriceCurrentCurrency : $currentPriceCurrentCurrency;
+            $discountPrice = $initialPrice - $this->specificPrice['real_value'];
+
+            if (\Configuration::get('PS_DISPLAY_DISCOUNT_PRICE')) {
```
PierreRambaud: Same for Configuration

src/Core/SpecificPrice/SpecificPriceFormatter.php
matthieu-rolland: @PierreRambaud  Context has no namespace... but:

Your comment makes me realize that this class, that contains legacy code, in which I only refactored the part where my fix is applied, should not be in core, but in the legacy code. I'll move that (and change the calls accordingly)

PierreRambaud: When I said, use namespace, I mean `use Context;` at the top to import them as global classes.

matthieu-rolland: done !

src/Core/SpecificPrice/SpecificPriceFormatter.php
matthieu-rolland: done !

matks:
@matthieu-rolland  it seems you wait for @PierreRambaud  validation of the changes you added following his code review, is that correct ?

matthieu-rolland:
> @matthieu-rolland it seems you wait for @PierreRambaud validation of the changes you added following his code review, is that correct ?

Not really @matks , after talking with other devs from the core team, we concluded that I need @eternoendless 's opinion about this matter:

I extracted a complex legacy function in a class (for easy unit testing purpose), that I put in the core. But the class still contains spaghetti legacy code (only the part I fixed is refactored). 

I was not sure if this class should go in the legacy or stay in the core (knowing it has legacy in it) ?

@eternoendless your opinion would be appreciated 🤔

src/Core/SpecificPrice/SpecificPriceFormatter.php
```diff
@@ -0,0 +1,184 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\SpecificPrice;
+
+use Configuration;
+use Context;
```
eternoendless: Core code SHOULD NOT depend on legacy classes. If you have legacy dependencies, then you MUST declare an interface and implement it as an Adapter. Also, try to keep legacy dependencies out of the interface.

matthieu-rolland: Thank you for your input @eternoendless I'll do that 👍

matthieu-rolland: So as we said, there is no adapter, a legacy class was created (it will never be used by the core, and no new code was added, it's just a legacy method being moved in a new legacy class due to unit testing issues).

classes/product/SpecificPriceFormatter.php
```diff
@@ -0,0 +1,179 @@
...
+            if ($this->specificPrice['reduction_type'] == 'amount') {
+                $currentPriceCurrentCurrency -= ($this->specificPrice['reduction_tax'] ? $this->specificPrice['reduction'] : $this->specificPrice['reduction'] / (1 + $tax_rate / 100));
+                $this->specificPrice['reduction_with_tax'] = $this->specificPrice['reduction_tax'] ? $this->specificPrice['reduction'] : $this->specificPrice['reduction'] / (1 + $tax_rate / 100);
+            } else {
+                $currentPriceCurrentCurrency *= 1 - $this->specificPrice['reduction'];
+            }
+            $this->specificPrice['real_value'] = $initialPrice > 0 ? $initialPrice - $currentPriceCurrentCurrency : $currentPriceCurrentCurrency;
+            $discountPrice = $initialPrice - $this->specificPrice['real_value'];
+
+            if (Configuration::get('PS_DISPLAY_DISCOUNT_PRICE')) {
```
jolelievre: Maybe you can inject this configuration in the constructor as well?
So that we limit as much as possible the references to other legacy classes

classes/product/SpecificPriceFormatter.php
```diff
@@ -0,0 +1,179 @@
...
+                $this->specificPrice['price'],
+                (bool) $this->specificPrice['reduction_tax'],
+                $tax_rate,
+                $ecotax_amount
+            );
+
+            // Since this price is set in default currency,
+            // we need to convert it into current currency
+            $this->specificPrice['id_currency'];
+            $currentPriceCurrentCurrency = \Tools::convertPrice($currentPriceDefaultCurrency, $this->context->currency, true, $this->context);
```
jolelievre: Seems like you only need the context to use this method
But if you look at the `Tools::convertPrice` it only needs the context as a fallback if the currency parameter is not provided
So instead of injecting the context you could simply inject the currency or use it as a parameter of your method, this way you could call `\Tools::convertPrice($currentPriceDefaultCurrency, $this->currency, true);` with the same result but then you can remove the dependency to Context

jolelievre (changes requested):

classes/product/SpecificPriceFormatter.php
matthieu-rolland: @jolelievre  done !

classes/product/SpecificPriceFormatter.php
matthieu-rolland: @jolelievre  done !

jolelievre (approved):
Nice work @matthieu-rolland

PierreRambaud:
Thanks @matthieu-rolland

PR #14894, opened 2019-07-29
Use utf8mb4 charset to support emoji

PierreRambaud:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Use utf8mb4 to allow emoji. Also add the right capability to manage environment (dev, test, and prod)
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Ticket? | Fixes #11083
| How to test?  | You're able to upgrade, install and add emoji in text. :warning: The index size for utf8mb4 fields will changes to 191 due to a limitations with old MariaDb / Mysql instances.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14894)

PR #15556, opened 2019-09-16
Add Sanity tests on travis build

boubkerbribri:
| Questions     | Answers
| Branch?       | develop
| Description?  | Sanity tests (on puppeteer) will be added on travis build to validate (or not) each PR created (and to replace old/new selenium tests)
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | see travis build result

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15556)


tests/puppeteer/pages/BO/addProduct.js
```diff
@@ -128,7 +128,15 @@ module.exports = class AddProduct extends BOBasePage {
    * @return page opened
    */
   async previewProduct() {
-    return this.openLinkWithTargetBlank(this.page, this.previewProductLink);
+    this.page = await this.openLinkWithTargetBlank(this.page, this.previewProductLink);
```
SimonGrn, line 131: 👍

tests/puppeteer/pages/Install/install.js
```diff
@@ -125,7 +125,7 @@ module.exports = class Install extends CommonPage {
     await this.page.click(this.dbLoginInput, {clickCount: 3});
     await this.page.type(this.dbLoginInput, global.db_user);
     await this.page.click(this.dbPasswordInput, {clickCount: 3});
-    await this.page.type(this.dbPasswordInput, global.db_passwd);
+    await this.page.type(this.dbPasswordInput, global.db_passwd.replace('empty_string', ''));
```
SimonGrn: What for ?

tests/puppeteer/pages/commonPage.js
```diff
@@ -52,13 +52,14 @@ module.exports = class CommonPage {
...
-    const newTarget = await global.browser.waitForTarget(target => target.opener() === pageTarget);
-    this.page = await newTarget.page();
-    await this.page.waitForSelector('body');
-    return this.page;
+  async openLinkWithTargetBlank(currentPage, selector, waitForNavigation = true) {
+    const [newPage] = await Promise.all([
+      new Promise(resolve => this.page.once('popup', resolve)),
+      currentPage.click(selector),
+    ]);
+    if (waitForNavigation) await newPage.waitForNavigation({waitUntil: 'networkidle0'});
```
SimonGrn, line 60: Is there some cases we *don't* want to wait for navigation ?

travis-scripts/run-sanity-tests
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+PROJECT_PATH=$(cd "$( dirname "$0" )/../" && pwd)
+
+/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_10.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :10 -ac -screen 0 1600x1200x16
```
SimonGrn, line 5: Is the resolution OK for our tests ?

SimonGrn (commented):
Thanks @boubkerbribri

tests/puppeteer/pages/commonPage.js
boubkerbribri: Yes, after install when it click on go to FO, browser don't detect navigation.

travis-scripts/run-sanity-tests
boubkerbribri: Yes, our tests passes with this size, and we define our browser size inside helper.
https://travis-ci.org/PrestaShop/PrestaShop/builds/585502141

tests/puppeteer/pages/Install/install.js
boubkerbribri: travis has empty value for db_passwd, if we passe `DB_PASSWD=""`, he will take admin from global
```js
   global.db_passwd = process.env.DB_PASSWD || 'admin';
```
So I pass a value, and replace it with an empty string in test

SimonGrn: Nice workaround.

tests/puppeteer/pages/BO/BObasePage.js
```diff
@@ -46,6 +46,10 @@ module.exports = class BOBasePage extends CommonPage {
     this.modalDialog = '#confirmation_modal.show .modal-dialog';
     this.modalDialogYesButton = `${this.modalDialog} button.continue`;
     this.modalDialogNoButton = `${this.modalDialog} button.cancel`;
+
+    // Symfony Toolbar
+    this.sfToolbarMainContentDiv = 'div[id*=\'sfToolbarMainContent\']';
```
matks, line 51: @boubkerbribri @SimonGrn We can disable the SF toolbar from the SF config, in `app/config/config_test.yml` (this requires Symfony is run in TEST environment), would you like that ?

Also it should be disabled in PROD mode so it means your sanity tests run in DEV mode (for PrestaShop/Symfony) ?

matks (dismissed):

tests/puppeteer/pages/BO/BObasePage.js
boubkerbribri: > @boubkerbribri @SimonGrn We can disable the SF toolbar from the SF config, in `app/config/config_test.yml` (this requires Symfony is run in TEST environment), would you like that ?

We would like that, yes.

> Also it should be disabled in PROD mode so it means your sanity tests run in DEV mode (for PrestaShop/Symfony) ?

If we install Prestashop in PROD mode (for travis tests), unit and integration tests will failed, maybe 2 different installations ?

matks: Unit and integration tests should be run by different Travis builds. And in fact we should run all our tests in TEST environment, not DEV, and not PROD 😄

.travis.yml
```diff
@@ -89,6 +90,10 @@ script:
         bash tests/check_integration.sh;
     fi
 
+  - if [ "$PRESTASHOP_TEST_TYPE" = "sanity" ]; then
+    bash tests/check_sanity.sh;
```
eternoendless: A little indentation issue here

```suggestion
        bash tests/check_sanity.sh;
```

boubkerbribri: done thanks

SimonGrn (dismissed):
Thanks @boubkerbribri

tests/check_sanity.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+bash travis-scripts/run-sanity-tests;
+SANITY=$?
+
+
+if [[ "$SANITY" == "0" ]]; then
+  echo -e "\e[92mPUPPETEER SANITY TESTS OK"
```
PierreRambaud:
```suggestion
  echo -e "\e[92mPUPPETEER SANITY TESTS OK\e[0m"
```

tests/check_sanity.sh
```diff
@@ -0,0 +1,13 @@
...
+  exit 0;
+else
+  echo -e "\e[91mPUPPETEER SANITY TESTS FAILED"
```
PierreRambaud:
```suggestion
  echo -e "\e[91mPUPPETEER SANITY TESTS FAILED\e[0m"
```

tests/check_sanity.sh
```diff
@@ -0,0 +1,13 @@
...
+  exit 255;
+fi
```
PierreRambaud, line 13: You also can do:
```bash
if [[ $SANITY -ne 0 ]]; then
  echo -e "\e[91mPUPPETEER SANITY TESTS FAILED"
  exit 255;
fi

echo -e "\e[92mPUPPETEER SANITY TESTS OK"
exit 0;
```

tests/puppeteer/pages/BO/BObasePage.js
PierreRambaud: @matks I'm not totally agreed here.
We must simulate a PROD context in a TEST environment. config_test needs to be updated in order to disable the debug toolbar.
But as far as I know, there are only two env available in prestashop right now, Dev and Prod. We do something here to include the "Test" env here: https://github.com/PrestaShop/PrestaShop/pull/14894

tests/check_sanity.sh
SimonGrn: I know it's a more logical approach, but I always thought it was a little confusing to read, having this last `exit` not conditioned.

tests/puppeteer/pages/Install/install.js
PierreRambaud: So this code only works on Travis? :thinking: 
We're able to set up DB_PASSWD global variable especially for Travis.
I'm not in favor of this durty fix :/

PierreRambaud: Maybe you can do
```
global.db_passwd = process.env.DB_PASSWD || '';
```

as default, because many instances (or many devs) have no password for database connection.
Because i'm fuc**d if I have `empty_string` in my password :trollface:

tests/check_sanity.sh
boubkerbribri: Commit first suggestions instead if this one
thanks @PierreRambaud

tests/puppeteer/pages/Install/install.js
boubkerbribri: Admin was set as db_password for docker 
so I change it @PierreRambaud 
https://github.com/PrestaShop/PrestaShop/pull/15556/commits/f43d49ae73c2971d0ce8da237950330c2c7b9fef

PierreRambaud:
Thanks a lot @boubkerbribri for your awesome work!

PierreRambaud (approved):

PR #15576, opened 2019-09-17
Update php docs in class PaymentModule

FlorianBergeron:
| Questions     | Answers
| Branch?       | develop
| Description?  | Update php docs for in class PaymentModule.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | Use php stan level 5.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15576)

### DIFF
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -193,7 +193,7 @@ abstract class PaymentModuleCore extends Module
      * Function called from a payment module.
      *
      * @param int $id_cart
-     * @param int $id_order_state
+     * @param string $id_order_state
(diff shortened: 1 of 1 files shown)


PierreRambaud (changes requested):
Hum, it's an id, it must be an int.
It's cast in the code in case it's a string.
If you want to pass phpstan, cast it before using it as parameter.

matks:
Closing for now, please open a new PR following @PierreRambaud feedback 😉

Issue #15582, opened 2019-09-17
Travis distrib is outdated


matks:
**Describe the bug**
We use trusty, an outdated distrib version for Travis builds. We should upgrade to a more recent version.

https://stackoverflow.com/questions/57903415/travis-ci-chrome-62-instead-of-77

matks:
Fixed by https://github.com/PrestaShop/PrestaShop/pull/18214/files

PR #15583, opened 2019-09-17
Use Xenial instead of old Trusty distrib for Travis

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Use Xenial instead of old Trusty distrib for Travis. This should prevent outdated package errors
| Type?         | bug fix
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/15582

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15583)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ cache:
     - $HOME/.composer/cache
 
 sudo: required
-dist: trusty
+dist: xenial
 
 php:
   - 7.1


jolelievre (approved):

PierreRambaud (approved):

matks:
According to Travis output, it looks like it'll need some work :trollface:

PR #15587, opened 2019-09-17
Avoid 301 redirect in the hreflang - add combination

cedricfontaine:
| Questions     | Answers
| Branch?       | develop
| Description?  | We want to avoid to have redirect 301 in hfreflang. If we add the combination in the URL, we will fix this issue.It's best for SEO
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Check the URL in the hreflang in the header

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15587)

### DIFF
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -1193,7 +1193,7 @@ class LinkCore
             } else {
                 $id_product_attribute = 0;
             }
-            $url = $this->getProductLink((int) $params['id_product'], null, null, null, (int) $idLangnull, (int)$id_product_attribute);
+            $url = $this->getProductLink((int) $params['id_product'], null, null, null, (int) $idLang, null, (int)$id_product_attribute);
             $url = explode('#', $url);
             return $url[0];
         } elseif ($controller == 'category' && isset($params['id_category'])) {
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -1188,7 +1188,14 @@ class LinkCore
         }
 
         if ($controller == 'product' && isset($params['id_product'])) {
-            return $this->getProductLink((int) $params['id_product'], null, null, null, (int) $idLang);
+            if (array_key_exists('id_product_attribute', $params)) {
+                $id_product_attribute = $params['id_product_attribute'];
+            } else {
+                $id_product_attribute = 0;
+            }
+            $url = $this->getProductLink((int) $params['id_product'], null, null, null, (int) $idLangnull, (int)$id_product_attribute);
+            $url = explode('#', $url);
+            return $url[0];
         } elseif ($controller == 'category' && isset($params['id_category'])) {
             return $this->getCategoryLink((int) $params['id_category'], null, (int) $idLang);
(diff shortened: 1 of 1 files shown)


classes/Link.php
```diff
@@ -1188,7 +1188,14 @@ public function getLanguageLink($idLang, Context $context = null)
         }
 
         if ($controller == 'product' && isset($params['id_product'])) {
-            return $this->getProductLink((int) $params['id_product'], null, null, null, (int) $idLang);
+            if (array_key_exists('id_product_attribute', $params)) {
+                $id_product_attribute = $params['id_product_attribute'];
+            } else {
+                $id_product_attribute = 0;
+            }
+            $url = $this->getProductLink((int) $params['id_product'], null, null, null, (int) $idLangnull, (int)$id_product_attribute);
```
matks:
```suggestion
            $url = $this->getProductLink((int) $params['id_product'], null, null, null, (int) $idLang, null, (int)$id_product_attribute);
```

matks:
Hi @cedricfontaine from your PR, I'd say this is about SEO behavior for product with combinations 🤔can you tell us more about what you want to do ?

SEO is a bit a dark area where it's not easy to find the truth, sometimes we ask 10 SEO experts on such a topic and get wet 10 different answers 😅

PierreRambaud:
Hi and thanks for your contribution

But your pull request description seems to be incomplete or malformed:

Please follow our Bot instructions 😉 

Thank you!

cedricfontaine:
Hello,
I updated the description.
The main goal is too avoid Redirect 301 in alternate lang. It will reduce server load and also could help SEO.

colinegin:
Hello @cedricfontaine, thanks for this PR.
I have the same comment as for sitemap module : https://github.com/PrestaShop/gsitemap/pull/120
Isn't it better to send to the canonical URL ? Since 1.7.6, the canonical URL for products with combinations is available, preventing an automatic 301 redirection.

cedricfontaine:
Hello @colinegin,
Yes, it's better to use the same canonical everywhere, if we also have combination in canonical, I'd rather use it in here and in gsitemap.

colinegin:
Hello @cedricfontaine ,

Sorry i was not really clear. The canonical url stays : mystore.com/1-product like it was before, but when reaching this url, there will no longer be an automatic redirection to the default combination url. The canonical url will actually display the elements of the default combination product. This has been done to avoid 301 redirects. All specifications related to SEO are listed here if you want more information : https://github.com/PrestaShop/prestashop-specs/blob/master/seo-rules-and-behaviours.md.

I guess we can close this PR then, right ?

cedricfontaine:
Yes we can close.
Are you in charge of this SEO changes ? I've another issue I'd like to exchange with you.

colinegin:
Sure, i will be glad to study it !
Yes I'm working on SEO improvements, there's an EPIC about it by the way if you want to have a look at it : https://github.com/PrestaShop/PrestaShop/issues/18402

Thanks !

Issue #9592, opened 2018-08-22
[BOOM-5955] Strange behavior after editing a carrier


prestonBot:
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-5955).

- _**Reporter:**_ khouloud.belguith
- _**Created at:**_ Wed, 11 Jul 2018 18:04:33 +0200

<p>Hi,</p>

<p>After a fresh install PS1.7.4.0, navigate to the BO => IMPROVE => Shipping => Carriers, by default, we have two addresses, for the first one which id=2.<br/>
Try to edit this carrier, then click on save without changing any files, the id of this carrier will be changed.<br/>
In fact, after editing any carrier, a new carrier is created in the Database.</p>

<p>Best regards, Khouloud</p>
        
* How to reproduce the issue ?

1-Fresh Install 1.7.4.0
<br/>
2-Navigate to BO =&gt;  IMPROVE =&gt; Shipping =&gt; Carriers
<br/>
3-Edit a carrier

s1nn3r:
The same for me on 1.7.3.4
After editing a carrier the carrier is copied to a new ID, the old one is set to deleted in the database. Every now and then the ranges are not copied correctly and I receive an "overlapping ranges" error for the new carrier and it's not editable anymore.

matks:
I confirm the bug still exists in 1.7.4.2 and on `develop` branch: when I edit a carrier, it does not edit, it actually clones the carrier with updated informations. The original is not deleted in the database but does not appear in the listing.

khouloudbelguith:
Hi,

The issue is added to our bug roadmap, but not fixed yet.

Best regards, Khouloud

ANSB:
Not sure it's a bug. In order to duplicate you need to get the data and perform an INSERT which syntax is really different from an UPDATE. Also, the fact there is a delete flag in the table demonstrate the use of the flag.
I think the idea was to save old value. Eg a customer ask for a product on day 1, so the shipping cost and all other values are computed using the carrier data of day 1. On day 2 you change the carrier in the backoffice. When the user came back on day 3 he must get back carrier's data of day 1 and not the updated of day 2.
What the system must do is:
If update of carrier data check if the carrier has been used.
- If yes, create a copy in order to save old value and set the delete flag to 1 in order to be able to use the data in statistics but not for new orders.
- If no, simply perform an update
[...]

PR #8816, opened 2018-03-02
CO: Fixed old record deletion in addDeliveryPrice

gfilippakis:
Fixed id_shop_group condition in DELETE statement
Added code for old record removal

| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.6.1.x
| Description?  | Records from deleted carrier IDs remain in ps_delivery table
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9592
| How to test?  | Create a carrier or edit an existing one and check the records of ps_delivery table.

#### Important guidelines

* Make sure [your local branch is up to date](https://help.github.com/articles/syncing-a-fork/) before commiting your changes!
* Your code MUST respect [our Coding Standards](http://doc.prestashop.com/display/PS16/Coding+Standards) (for code written in PHP, JavaScript, HTML/CSS/Smarty/Twig, SQL)!
* Your commit name MUST respect our [naming convention](http://doc.prestashop.com/display/PS16/How+to+write+a+commit+message)!

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/8816)

PR #15615, opened 2019-09-19
Fix bad shop group condition and delete old delivery records

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix bad shop group condition and delete old delivery records. Port of https://github.com/PrestaShop/PrestaShop/pull/8816
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Partially https://github.com/PrestaShop/PrestaShop/issues/9592. See last comment. 
| How to test?  | 😅 

So this PR does 2 things:

- it fixes a bad SQL condition where it's obvious id_shop should be id_shop_group
- it deletes old delivery SQL table records when new delivery records are added for the new carrier
What do I mean by "new carrier" ? Actually when carrier is edited, it is, a new carrier is created with new informations and an old record, unmodified, stays hidden in database.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15615)

### DIFF
--- a/classes/Carrier.php
+++ b/classes/Carrier.php
@@ -929,6 +929,7 @@ class CarrierCore extends ObjectModel
      * Add new delivery prices.
      *
      * @param array $priceList Prices list in multiple arrays (changed to array since 1.5.0)
+     * @param bool $delete
      *
      * @return bool Insertion result
      */
@@ -946,6 +947,25 @@ class CarrierCore extends ObjectModel
             $keys[] = 'id_shop_group';
         }
 
+        $old_carrier_record_ids = array();
+        if ($delete) {
+            // When carrier is edited, it is duplicated
+            // The new carrier is updated with new informations.
+            // The old, unmodified record stays hidden in database.
+
+            $old_carrier_records = Db::getInstance()->executeS('
+                SELECT `id_carrier` 
+                FROM `' . _DB_PREFIX_ . 'carrier` 
+                WHERE `id_reference` = ' . (int) $this->id_reference
+            );
+            if (is_array($old_carrier_records) && count($old_carrier_records) > 0) {
+                $old_carrier_record_ids = array_map(
+                    function ($record) {return $record['id_carrier']; },
+                    $old_carrier_records
+                );
+            }
+        }
+
         $sql = 'INSERT INTO `' . _DB_PREFIX_ . 'delivery` (' . implode(', ', $keys) . ') VALUES ';
         foreach ($price_list as $values) {
             if (!isset($values['id_shop'])) {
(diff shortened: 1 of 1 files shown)


classes/Carrier.php
```diff
@@ -959,8 +979,9 @@ public function addDeliveryPrice($price_list, $delete = false)
                 Db::getInstance()->execute(
                     'DELETE FROM `' . _DB_PREFIX_ . 'delivery`
                     WHERE ' . (null === $values['id_shop'] ? 'ISNULL(`id_shop`) ' : 'id_shop = ' . (int) $values['id_shop']) . '
-                    AND ' . (null === $values['id_shop_group'] ? 'ISNULL(`id_shop`) ' : 'id_shop_group=' . (int) $values['id_shop_group']) . '
-                    AND id_carrier=' . (int) $values['id_carrier'] .
+                    AND ' . (null === $values['id_shop_group'] ? 'ISNULL(`id_shop_group`) ' : 'id_shop_group=' . (int) $values['id_shop_group']) .
+                    // delete delivery records related to old carrier records
+                    'AND id_carrier IN (' . implode('","', $old_carrier_record_ids) .
```
matthieu-rolland, line 984: ` 'AND id_carrier IN (` 
I can't find the closing parenthesis (maybe I missed it?)

classes/Carrier.php
```diff
@@ -946,6 +947,25 @@ public function addDeliveryPrice($price_list, $delete = false)
             $keys[] = 'id_shop_group';
         }
 
+        $old_carrier_record_ids = array();
```
PierreRambaud, line 950:
```suggestion
        $old_carrier_record_ids = [];
```

classes/Carrier.php
```diff
@@ -946,6 +947,25 @@ public function addDeliveryPrice($price_list, $delete = false)
...
+            // The old, unmodified record stays hidden in database.
+
+            $old_carrier_records = Db::getInstance()->executeS('
+                SELECT `id_carrier` 
+                FROM `' . _DB_PREFIX_ . 'carrier` 
+                WHERE `id_reference` = ' . (int) $this->id_reference
+            );
+            if (is_array($old_carrier_records) && count($old_carrier_records) > 0) {
+                $old_carrier_record_ids = array_map(
+                    function ($record) {return $record['id_carrier']; },
```
PierreRambaud, line 963:
```suggestion
                    function ($record) { return $record['id_carrier']; },
```

classes/Carrier.php
S4hosting: which one have to use array() or [] ?

matks: `[]` is the good and modern way 👍

Issue #15274, opened 2019-08-27
When you are in all shop context all categories are not displayed in product page


sitte:
**Describe the bug**
I have multisore, each store has it's own category branch, all stores has the same products.
When editing product in "all shops" context, category assigment block shows categories only from one store. In Prestashop 1.5.4.1 I see all category branches from all shops.

After saving it overrides main category in all shops. 

**To Reproduce**
1. Activate multistore, make 2 stores A and B
2. Make 2 categories CatA and CatB
3. Assign CatA to Store A and Cat B to Store B
4. Ad one product to CatA and CatB
5. Try to edit this product in "all shops context" and check what categories You see.
6. Try to save the product and check what is stored in ps_product_shop.id_category_default for this product

**Additional information**
PrestaShop version: 1.7.6
PHP version: 7.2

khouloudbelguith:
Hi @sitte,

Thanks for your report.
Following these steps:
1. Activate multistore, make 2 stores A and B
2. Make 2 categories CatA and CatB
3. Assign CatA to Store A and Cat B to Store B
4. All shops context is selected => Add new product => CatB is not displayed in the list of categories
=> It is not the same behavior as PS1.7.5.2

PS1.7.6.0/PS1.7.6.1:

![image](https://user-images.githubusercontent.com/16067358/63768023-0a65c680-c8c7-11e9-8baf-a7f61df0c35b.png)
PS1.7.5:
![image](https://user-images.githubusercontent.com/16067358/63768089-341eed80-c8c7-11e9-92d2-5af2dac7603b.png)

This is a regression.

>Try to save the product and check what is stored in ps_product_shop.id_category_default for this product

I have the same behavior as PS1.7.5 & PS1.7.6
Because when All shops context is selected => We have only a default category is selected.
I’ll add this to the debug roadmap so that it’s fixed. If you have already fixed it on your end or if you think you can do it, please do send us a pull request!
Thanks!

sitte:
I see 2 problems:
1. in "all shops" context we should see all category branches from all stores, not from default store. 
2. in "all shops" context there should be main category per shop not per multistore

I can try to fix 1. Can You point me in direction where are functions that load category structure, it would speed up the process?

I think that 2 needs not only change of loaded data but in addition change of interface - not one radio button but one for each store.

khouloudbelguith:
@sitte, we should ask developers: @PrestaShop/prestashop-core-developers

matks:
[...]

PR #11876, opened 2018-12-20
Fix lang table multi-store data issue

Dan-Orchard:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | With a multi-store set up, adding a product or category to an individual store will put a line for each of the stores into the object_lang table.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9606
| How to test?  | Create a multi-store - Create a category/product in an individual store - only lines with associated stores will be added to object_lang table rather than for all stores.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/11876)

PR #15621, opened 2019-09-19
Revert "Fix lang table multi-store data issue"

matks:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Reverts PrestaShop/PrestaShop#11876 as it seems to have introduced major regression https://github.com/PrestaShop/PrestaShop/issues/15274
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/15274
| How to test?  | See ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15621)

### DIFF
--- a/install-dev/upgrade/sql/1.7.6.2.sql
+++ b/install-dev/upgrade/sql/1.7.6.2.sql
@@ -1,0 +1,50 @@
+SET SESSION sql_mode = '';
+SET NAMES 'utf8';
+
+INSERT INTO `PREFIX_category_lang` SELECT
+`entity`.`id_category`, 1, `entity`.`id_lang`, `entity`.`name`, `entity`.`description`, `entity`.`link_rewrite`, `entity`.`meta_title`, `entity`.`meta_keywords`, `entity`.`meta_description`
+FROM `PREFIX_category_lang` entity
+LEFT JOIN `PREFIX_category_lang` entity2 ON `entity2`.`id_shop` = 1 AND `entity`.`id_category` = `entity2`.`id_category`
+WHERE `entity2`.`id_shop` IS NULL;
+
+INSERT INTO `PREFIX_cms_category_lang` SELECT
+`entity`.`id_cms_category`, `entity`.`id_lang`, 1, `entity`.`name`, `entity`.`description`, `entity`.`link_rewrite`, `entity`.`meta_title`, `entity`.`meta_keywords`, `entity`.`meta_description`
+FROM `PREFIX_cms_category_lang` entity
+LEFT JOIN `PREFIX_cms_category_lang` entity2 ON `entity2`.`id_shop` = 1 AND `entity`.`id_cms_category` = `entity2`.`id_cms_category`
+WHERE `entity2`.`id_shop` IS NULL;
+
+INSERT INTO `PREFIX_cms_lang` SELECT
+`entity`.`id_cms`, `entity`.`id_lang`, 1, `entity`.`meta_title`, `entity`.`head_seo_title`, `entity`.`meta_description`, `entity`.`meta_keywords`, `entity`.`content`, `entity`.`link_rewrite`
+FROM `PREFIX_cms_lang` entity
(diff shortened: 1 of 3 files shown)


PierreRambaud (dismissed):
If we revert this change, we must add an upgrade script to insert `_lang` values

install-dev/upgrade/sql/1.7.6.2.sql
```diff
@@ -0,0 +1,50 @@
+SET SESSION sql_mode = '';
+SET NAMES 'utf8';
+
+INSERT INTO `PREFIX_category_lang` SELECT
+`entity`.`id_category`, 1, `entity`.`id_lang`, `entity`.`name`, `entity`.`description`, `entity`.`link_rewrite`, `entity`.`meta_title`, `entity`.`meta_keywords`, `entity`.`meta_description`
+FROM `PREFIX_category_lang` entity
+LEFT JOIN `PREFIX_category_lang` entity2 ON `entity2`.`id_shop` = 1 AND `entity`.`id_category` = `entity2`.`id_category`
```
jolelievre, line 7: I don't get it, you force the `id_shop` to `1`, but select values from `entity2.id_shop = 2 AND null`
So no rows is inserted with id_shop values different from 1, and even in the select how can id_shop be both equal to 1 and null

Sure, I'm not a champion in SQL ^^ But I'm not the only one so maybe you could add a comment to explain what the query does?

src/Adapter/Category/QueryHandler/GetCategoryForEditingHandler.php
```diff
@@ -66,7 +67,7 @@ public function handle(GetCategoryForEditing $query)
     {
         $category = new Category($query->getCategoryId()->getValue());
 
-        if (!$category->id) {
+        if (!$category->id || (!$category->isAssociatedToShop() && Shop::getContext() == Shop::CONTEXT_SHOP)) {
```
jolelievre, line 70: The condition is a bit different, so maybe a dedicated exception should be thrown?
Like `InvalidShopContextException` ? or a very explicit `CategoryNotFoundInContextException`

jolelievre (commented):
I don't understand everything in your query :D

src/Adapter/Category/QueryHandler/GetCategoryForEditingHandler.php
PierreRambaud: In fact it's the original behaviour for `CategoryNotFoundException` 😅

install-dev/upgrade/sql/1.7.6.2.sql
PierreRambaud: I'm looking for rows that haven't data with id_shop = 1.
The `LEFT JOIN` combined with a `WHERE NULL` check for nonexistent rows.
This means `Give me id_category that have id_shop != 1 and don't have a row with the same id_category and id_shop 1`.

This kind of SQL request are often used for finding duplicates.

jolelievre: Alright I finally understood it ^^ Thanks for learning a new SQL technique

src/Adapter/Category/QueryHandler/GetCategoryForEditingHandler.php
jolelievre: Ok, I won't question this logic then 😛

Progi1984 (approved):

marionf:
@matks should we reopen the issue fixed by #11876 ?

PierreRambaud:
@marionf I don't think because it was the legacy behaviour and afaik it's mandatory to have all language saved in database :/

jolelievre (approved):
Thanks @PierreRambaud

PierreRambaud:
Thanks @matks

matks:
> @marionf I don't think because it was the legacy behaviour and afaik it's mandatory to have all language saved in database :/

So you think that https://github.com/PrestaShop/PrestaShop/pull/11876 is harmless and actually expected ?

PierreRambaud:
@matks After reading how the ObjectModel works with language, yes it was the expected behavior..

Issue #15624, opened 2019-09-20
Removal of a customization makes impossible to remove a product from the cart


jocel1:
**Describe the bug**
Add a customization to a product, remove the customization, put the product into your cart. You'll not be able to remove the product from the cart because will still have an id_customization associated, but no more customized_data

**To Reproduce**
1. Create a product with a customized field
2. In the product page, add a customization, click save.
3. Click once again on save without any customization text, this will delete the customization
4. Add the product into your cart
5. You will not be able anymore to remove the product from your cart

Tested on 1.7.6.1

khouloudbelguith:
Hi @jocel1,

I manage to reproduce the issue with PS1.7.6.1 & PS1.7.5.2 & PS1.7.6.0.
The Customization should not be required.
https://drive.google.com/file/d/1FL1lEjSAhlPImIUr8PPWof_uQa9MoRGb/view
I’ll add this to the debug roadmap.
Thanks!

matks:
Fixed by https://github.com/PrestaShop/PrestaShop/pull/15625

PR #15625, opened 2019-09-20
Properly cleanup the customization table if no more customized_data

jocel1:
| Questions     | Answers
| Branch?       | develop
| Description?  | Properly cleanup the customization table when customized_data is deleted
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #15624 
| How to test?  | Add a customization to a product, remove the customization, put the product into your cart. Without the fix, you'll not be able to remove the product from the cart because will still have an id_customization associated, but no more customized_data

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15625)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -4179,14 +4179,14 @@ class CartCore extends ObjectModel
         );
 
         $hasRemainingCustomData = Db::getInstance()->getValue(
-            'SELECT 1 FROM `'._DB_PREFIX_.'customized_data`
-            WHERE `id_customization` = '.(int)$cust_data['id_customization']
+            'SELECT 1 FROM `' . _DB_PREFIX_ . 'customized_data`
+            WHERE `id_customization` = ' . (int) $cust_data['id_customization']
         );
 
         if (!$hasRemainingCustomData) {
             $result &= Db::getInstance()->execute(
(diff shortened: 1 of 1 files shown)


classes/Cart.php
```diff
@@ -4178,6 +4178,18 @@ public function deleteCustomizationToProduct($id_product, $index)
             AND `index` = ' . (int) $index
         );
 
+        $hasRemainingCustomData = Db::getInstance()->getValue(
+            'SELECT 1 FROM `' . _DB_PREFIX_ . 'customized_data`
+            WHERE `id_customization` = ' . (int) $cust_data['id_customization']
+        );
+
+        if (!$hasRemainingCustomData) {
```
PierreRambaud, line 4186: Is it really needed to check for existence before removing rows?

jocel1: One customization can have several customized_data (we can put several custom fields on a product). So if there are several custom fields and only one is removed, we have to keep the customization entry

PierreRambaud: Oh yes, missreading the `!` :sweat_smile:

PierreRambaud (approved):

matks:
Thank you @jocel1 😉

Issue #15641, opened 2019-09-22
Breadcrumb is not properly computed for home category


jocel1:
**Describe the bug**

**To Reproduce**
1. Go to a home category
2. Check the source code. 
3. You can see the breadcrumb contains two elements instead of one, and the second is null.

For example on https://demo-prestashop.prestarocket.com/2-accueil
"breadcrumb":{"links":[{"title":"Accueil","url":"https:\/\/demo-prestashop.prestarocket.com\/"},null],"count":2}

With themes trying to displaying breadcrumb for home it could be annoying, and it also "corrupt" google rich snippets

PrestaShop version: 1.7.6.1

JBWModules:
Duplicate to https://github.com/PrestaShop/PrestaShop/issues/15302

khouloudbelguith:
Hi @jocel1,

Thanks for your report.
I manage to reproduce the issue with PS1.7.6.1 & PS1.7.5.2 & the develop branch.
@JBWModules, your PR is merged in develop
![image](https://user-images.githubusercontent.com/16067358/65413567-9f39e200-dde9-11e9-93b6-50bbb3318d59.png)
But, You can see the breadcrumb contains two elements instead of one, and the second is null.
![image](https://user-images.githubusercontent.com/16067358/65413716-ef18a900-dde9-11e9-8741-ad2f9ebc0788.png)
I’ll add this to the debug roadmap.
Thanks!

JBWModules:
True, it's still wrong in the JS code but with the PRs in development at least it's fixed in the visible part and meta tags relevant for SEO.

jocel1:
@jbwmodules yes but only if you're using the 'standard' theme

SNor42:
Hi!
I don't know what the solution would be, but I would prefer this one:
![exp1](https://user-images.githubusercontent.com/1334351/73102492-3b880880-3ef2-11ea-9046-550616b42a0c.jpg)
Also there is the quite same problem when we try to put the category_id 2 on the top menu. The main link goes to the home, and to the all products page:
![exp2](https://user-images.githubusercontent.com/1334351/73102696-b6e9ba00-3ef2-11ea-8146-0751ce794172.jpg)
Is it possible to fix this too? Thanks

PR #15642, opened 2019-09-22
Fix wrong breadcrumb on home category

jocel1:
| Questions     | Answers
| Branch?       | develop
| Description?  | Breadcrumb computation is wrong on home category
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #15641 
| How to test?  | Go on the home category, and check breadcrumb is the source. It should only contains 1 entry + "count":1. Without the fix it generates invalid breadcrumb, which could be annoying when using google rich snippets

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15642)

### DIFF
--- a/controllers/front/listing/CategoryController.php
+++ b/controllers/front/listing/CategoryController.php
@@ -233,7 +233,7 @@ class CategoryControllerCore extends ProductListingFrontController
             }
         }
 
-        if (!$this->category->is_root_category) {
+        if ($this->category->id_parent != 0 && !$this->category->is_root_category) {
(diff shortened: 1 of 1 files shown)


PierreRambaud (approved):

theAllard:
HI, when you go to the admin , catalog , categories page and then edit the home category, you will see in the seo preview domain_name/**undefined-2**. I think this is related to this problem.

PR #15643, opened 2019-09-23
Allow CLDR to display deleted/inactive currencies

### LINKED ISSUE #15376: Error in BO order page after deleting a currency used in an order
**Describe the bug**
Issue created from this comment: https://github.com/PrestaShop/PrestaShop/issues/14595#issuecomment-527248135 reported by @presciak & @Kis-shoes 
image

Steps: reproduce the issue:

1. Fresh install PS1.7.6 with EN language & FR country
1. Import the United states=> Dollar currency created / Import the Switzerland pack => CHF currency created
1. create some order with those currencies
1. Try to add a cart with amount reduction
1. Set the status of your orders to Delivered
1. Set the **CHF** currency by default
1. Delete the Euro & Dollar currencies
1. Go back to the BO => Orders page => an exception displayed
image
**Additional information**
PrestaShop version: 1.7.6.1

### LINKED ISSUE #15486: Error in suppliers page after deleting a currency - Regression
I would like to refer to this issue (https://github.com/PrestaShop/PrestaShop/issues/14908). This issue appears to be resolved in my other shops, but not on morganic.mu. My guess is that the alterations that I made on the database to rectify the issue prior to 1.7.6.1 update may have corrupted the database (see issue 14908 which explains the solutions I tried). 

I am hoping that we can rectify the issue or at least avoid worst case scenario: fresh reinstall of prestashop.

Exception discovered when viewing suppliers: 64470979-9d6eee00-d15c-11e9-8bb0-23b225d096d2

Exception discovered when creating a new currency:

[exception.pdf](https://github.com/PrestaShop/PrestaShop/files/3595397/exception.pdf)

64470999-d9a24e80-d15c-11e9-963c-14d67529b1bd

Image of the database:

64471155-72859980-d15e-11e9-9373-78599caa2b39
64471156-787b7a80-d15e-11e9-8958-4f14a481ca48

**Additional information**
PrestaShop version: 1.7.6.1
PHP version: 7.2

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Allow CLDR to display delete currencies by fetching any currency in database regardless of its `deleted` or `active` status, and add a few behat tests for CLDR and currencies This required to add a few methods to existing interfaces thus creating a BC break But since these interfaces should mainly be used by the core for now it is relatively safe
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | Fixes #15376<br>Fixes #15486
| How to test?  | See issues' contents for details, but mainly create orders and suppliers with prices in EUR the change the default currency and delete the EUR currency Then you should have an exception in Orders page and Suppliers page

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)

### DIFF
--- a/app/config/set_parameters.php
+++ b/app/config/set_parameters.php
@@ -62,7 +62,7 @@ if ($container instanceof \Symfony\Component\DependencyInjection\Container) {
 
     // Parameter used only in dev and test env
     $envParameter = getenv('DISABLE_DEBUG_TOOLBAR');
-    if (!isset($parameters['use_debug_toolbar']) || false !== $envParameter) {
+    if (!isset($parameters['parameters']['use_debug_toolbar']) || false !== $envParameter) {
         $container->setParameter('use_debug_toolbar', !$envParameter);
     }
 }
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -402,18 +402,20 @@ class CurrencyCore extends ObjectModel
      * @param bool $active If true only active are returned
      * @param bool $groupBy Group by id_currency
      * @param bool $currentShopOnly If true returns only currencies associated to current shop
+     * @param bool $filterDeleted It true filter deleted currencies
      *
      * @return array Currency data from database
      *
      * @throws PrestaShopDatabaseException
      */
-    public static function findAll($active = true, $groupBy = false, $currentShopOnly = true)
+    public static function findAll($active = true, $groupBy = false, $currentShopOnly = true, $filterDeleted = true)
     {
         $currencies = Db::getInstance()->executeS('
             SELECT *
             FROM `' . _DB_PREFIX_ . 'currency` c
             ' . ($currentShopOnly ? Shop::addSqlAssociation('currency', 'c') : '') . '
-                WHERE `deleted` = 0' .
+                WHERE 1' .
+                ($filterDeleted ? ' AND c.`deleted` = 0' : '') .
                 ($active ? ' AND c.`active` = 1' : '') .
                 ($groupBy ? ' GROUP BY c.`id_currency`' : '') .
                 ' ORDER BY `iso_code` ASC');
@@ -800,7 +802,7 @@ class CurrencyCore extends ObjectModel
         $symbolsByLang = $namesByLang = [];
         foreach ($languages as $languageData) {
             $language = new Language($languageData['id_lang']);
-            if ($language->locale === '') {
+            if (empty($language->locale)) {
                 // Language doesn't have locale we can't install this language
                 continue;
             }
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -428,9 +428,9 @@ class CurrencyCore extends ObjectModel
      *
      * @throws PrestaShopDatabaseException
      */
-    public static function findAllInDatabase()
+    public static function findAllInstalled()
     {
-        $currencies = Db::getInstance()->executeS(
+        $currencies = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS(
             'SELECT * FROM `' . _DB_PREFIX_ . 'currency` c ORDER BY `iso_code` ASC'
         );
 
@@ -590,14 +590,14 @@ class CurrencyCore extends ObjectModel
      *
      * @param string $isoCode ISO code
      * @param int $idShop Shop ID
-     * @param bool $noCache
+     * @param bool $forceRefreshCache [default=false] Set to TRUE to forcefully refresh any currently cached results
      *
      * @return int Currency ID
      */
-    public static function getIdByIsoCode($isoCode, $idShop = 0, $noCache = false)
+    public static function getIdByIsoCode($isoCode, $idShop = 0, $forceRefreshCache = false)
     {
         $cacheId = 'Currency::getIdByIsoCode_' . pSQL($isoCode) . '-' . (int) $idShop;
-        if ($noCache || !Cache::isStored($cacheId)) {
+        if ($forceRefreshCache || !Cache::isStored($cacheId)) {
             $query = Currency::getIdByQuery($idShop);
             $query->where('iso_code = \'' . pSQL($isoCode) . '\'');
 
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -402,23 +402,37 @@ class CurrencyCore extends ObjectModel
      * @param bool $active If true only active are returned
      * @param bool $groupBy Group by id_currency
      * @param bool $currentShopOnly If true returns only currencies associated to current shop
-     * @param bool $filterDeleted It true filter deleted currencies
      *
      * @return array Currency data from database
      *
      * @throws PrestaShopDatabaseException
      */
-    public static function findAll($active = true, $groupBy = false, $currentShopOnly = true, $filterDeleted = true)
+    public static function findAll($active = true, $groupBy = false, $currentShopOnly = true)
     {
         $currencies = Db::getInstance()->executeS('
             SELECT *
             FROM `' . _DB_PREFIX_ . 'currency` c
             ' . ($currentShopOnly ? Shop::addSqlAssociation('currency', 'c') : '') . '
-                WHERE 1' .
-                ($filterDeleted ? ' AND c.`deleted` = 0' : '') .
+                WHERE c.`deleted` = 0' .
                 ($active ? ' AND c.`active` = 1' : '') .
                 ($groupBy ? ' GROUP BY c.`id_currency`' : '') .
                 ' ORDER BY `iso_code` ASC');
+
+        return $currencies;
+    }
+
+    /**
+     * Retrieve all currencies data from the database.
+     *
+     * @return array Currency data from database
+     *
+     * @throws PrestaShopDatabaseException
+     */
+    public static function findAllInDatabase()
+    {
+        $currencies = Db::getInstance()->executeS(
+            'SELECT * FROM `' . _DB_PREFIX_ . 'currency` c ORDER BY `iso_code` ASC'
+        );
 
         return $currencies;
     }
@@ -576,13 +590,14 @@ class CurrencyCore extends ObjectModel
      *
      * @param string $isoCode ISO code
      * @param int $idShop Shop ID
+     * @param bool $noCache
      *
      * @return int Currency ID
      */
-    public static function getIdByIsoCode($isoCode, $idShop = 0)
+    public static function getIdByIsoCode($isoCode, $idShop = 0, $noCache = false)
     {
         $cacheId = 'Currency::getIdByIsoCode_' . pSQL($isoCode) . '-' . (int) $idShop;
-        if (!Cache::isStored($cacheId)) {
+        if ($noCache || !Cache::isStored($cacheId)) {
             $query = Currency::getIdByQuery($idShop);
             $query->where('iso_code = \'' . pSQL($isoCode) . '\'');
 
--- a/src/Adapter/Currency/CommandHandler/AddCurrencyHandler.php
+++ b/src/Adapter/Currency/CommandHandler/AddCurrencyHandler.php
@@ -86,6 +86,7 @@ final class AddCurrencyHandler extends AbstractCurrencyHandler implements AddCur
             if (!empty($cldrCurrency)) {
                 // The currency may not be declared in the locale, eg with custom iso code
                 $entity->precision = (int) $cldrCurrency->getDecimalDigits();
+                $entity->numeric_iso_code = (int) $cldrCurrency->getNumericIsoCode();
             }
 
             $entity->refreshLocalizedCurrencyData(Language::getLanguages(), $this->localeRepoCLDR);
--- a/src/Adapter/Currency/CommandHandler/AddCurrencyHandler.php
+++ b/src/Adapter/Currency/CommandHandler/AddCurrencyHandler.php
@@ -86,7 +86,7 @@ final class AddCurrencyHandler extends AbstractCurrencyHandler implements AddCur
             if (!empty($cldrCurrency)) {
                 // The currency may not be declared in the locale, eg with custom iso code
                 $entity->precision = (int) $cldrCurrency->getDecimalDigits();
-                $entity->numeric_iso_code = (int) $cldrCurrency->getNumericIsoCode();
+                $entity->numeric_iso_code = $cldrCurrency->getNumericIsoCode();
             }
 
             $entity->refreshLocalizedCurrencyData(Language::getLanguages(), $this->localeRepoCLDR);
--- a/src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php
+++ b/src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php
@@ -268,7 +268,7 @@ final class EditCurrencyHandler extends AbstractCurrencyHandler implements EditC
                 $langIds = Language::getLanguages(true, false, true);
                 $entity->name = [];
                 $entity->symbol = [];
-                $entity->numeric_iso_code = $currency->getNumericIsoCode();
+                $entity->numeric_iso_code = (int) $currency->getNumericIsoCode();
 
                 foreach ($langIds as $langId) {
                     $entity->name[$langId] = $currency->getDisplayName();
--- a/src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php
+++ b/src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php
@@ -268,7 +268,7 @@ final class EditCurrencyHandler extends AbstractCurrencyHandler implements EditC
                 $langIds = Language::getLanguages(true, false, true);
                 $entity->name = [];
                 $entity->symbol = [];
-                $entity->numeric_iso_code = (int) $currency->getNumericIsoCode();
+                $entity->numeric_iso_code = $currency->getNumericIsoCode();
 
                 foreach ($langIds as $langId) {
                     $entity->name[$langId] = $currency->getDisplayName();
--- a/src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php
+++ b/src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php
@@ -268,6 +268,7 @@ final class EditCurrencyHandler extends AbstractCurrencyHandler implements EditC
                 $langIds = Language::getLanguages(true, false, true);
                 $entity->name = [];
                 $entity->symbol = [];
+                $entity->numeric_iso_code = $currency->getNumericIsoCode();
 
                 foreach ($langIds as $langId) {
                     $entity->name[$langId] = $currency->getDisplayName();
--- a/src/Core/Currency/CurrencyDataProviderInterface.php
+++ b/src/Core/Currency/CurrencyDataProviderInterface.php
@@ -56,11 +56,11 @@ interface CurrencyDataProviderInterface
     public function findAll($currentShopOnly = true);
 
     /**
-     * Return raw currencies data from the database (regardless of their active or deleted status).
+     * Return raw data of all installed currencies in the database (regardless of their active or soft deleted status).
      *
-     * @return array[] Currencies in database
+     * @return array[] Currencies installed in database
      */
-    public function findAllInDatabase();
+    public function findAllInstalled();
 
     /**
      * Get a Currency entity instance by ISO code.
--- a/src/Core/Currency/CurrencyDataProviderInterface.php
+++ b/src/Core/Currency/CurrencyDataProviderInterface.php
@@ -50,10 +50,11 @@ interface CurrencyDataProviderInterface
      * Return raw currencies data from the database.
      *
      * @param bool $currentShopOnly If true returns only currencies associated to current shop
+     * @param bool $filterDeleted If true deleted currencies are filtered
      *
      * @return array[] Installed currencies
      */
-    public function findAll($currentShopOnly = true);
+    public function findAll($currentShopOnly = true, $filterDeleted = true);
 
     /**
      * Get a Currency entity instance by ISO code.
--- a/src/Core/Currency/CurrencyDataProviderInterface.php
+++ b/src/Core/Currency/CurrencyDataProviderInterface.php
@@ -47,14 +47,20 @@ interface CurrencyDataProviderInterface
     public function getCurrencies($object = false, $active = true, $group_by = false);
 
     /**
-     * Return raw currencies data from the database.
+     * Return raw currencies data from the database (not deleted + active currencies).
      *
      * @param bool $currentShopOnly If true returns only currencies associated to current shop
-     * @param bool $filterDeleted If true deleted currencies are filtered
      *
-     * @return array[] Installed currencies
+     * @return array[] Available currencies
      */
-    public function findAll($currentShopOnly = true, $filterDeleted = true);
+    public function findAll($currentShopOnly = true);
+
+    /**
+     * Return raw currencies data from the database (regardless of their active or deleted status).
+     *
+     * @return array[] Currencies in database
+     */
+    public function findAllInDatabase();
 
     /**
      * Get a Currency entity instance by ISO code.
--- a/src/Core/Localization/Currency/CurrencyDataSource.php
+++ b/src/Core/Localization/Currency/CurrencyDataSource.php
@@ -92,9 +92,9 @@ class CurrencyDataSource implements DataSourceInterface
     /**
      * {@inheritdoc}
      */
-    public function getAllCurrenciesDataInDatabase($localeCode)
+    public function getAllInstalledCurrenciesData($localeCode)
     {
-        return $this->formatCurrenciesData($this->installedDataLayer->getAllCurrencyCodesInDatabase(), $localeCode);
+        return $this->formatCurrenciesData($this->installedDataLayer->getAllInstalledCurrencyIsoCodes(), $localeCode);
     }
 
     /**
--- a/src/Core/Localization/Currency/CurrencyDataSource.php
+++ b/src/Core/Localization/Currency/CurrencyDataSource.php
@@ -92,9 +92,9 @@ class CurrencyDataSource implements DataSourceInterface
     /**
      * {@inheritdoc}
      */
-    public function getAllCurrenciesData($localeCode)
+    public function getAllCurrenciesDataInDatabase($localeCode)
     {
-        return $this->formatCurrenciesData($this->installedDataLayer->getAvailableCurrencyCodes(), $localeCode);
+        return $this->formatCurrenciesData($this->installedDataLayer->getAllCurrencyCodesInDatabase(), $localeCode);
     }
 
     /**
@@ -103,7 +103,7 @@ class CurrencyDataSource implements DataSourceInterface
      *
      * @return array
      */
-    protected function formatCurrenciesData(array $currencyCodes, $localeCode)
+    private function formatCurrenciesData(array $currencyCodes, $localeCode)
     {
         $currenciesData = [];
         foreach ($currencyCodes as $currencyCode) {
--- a/src/Core/Localization/Currency/CurrencyDataSource.php
+++ b/src/Core/Localization/Currency/CurrencyDataSource.php
@@ -60,13 +60,7 @@ class CurrencyDataSource implements DataSourceInterface
     }
 
     /**
-     * Get complete currency data by currency code, in a given language.
-     *
-     * @param LocalizedCurrencyId $localizedCurrencyId
-     *                                                 The currency data identifier (currency code + locale code)
-     *
-     * @return CurrencyData
-     *                      The currency data
+     * {@inheritdoc}
      */
     public function getLocalizedCurrencyData(LocalizedCurrencyId $localizedCurrencyId)
     {
@@ -88,17 +82,29 @@ class CurrencyDataSource implements DataSourceInterface
     }
 
     /**
-     * Get all the available (installed + active) currencies' data.
-     *
-     * @param string $localeCode
-     *                           IETF tag. Data will be translated in this language
-     *
-     * @return CurrencyData[]
-     *                        The available currencies' data
+     * {@inheritdoc}
      */
     public function getAvailableCurrenciesData($localeCode)
     {
-        $currencyCodes = $this->installedDataLayer->getAvailableCurrencyCodes();
+        return $this->formatCurrenciesData($this->installedDataLayer->getAvailableCurrencyCodes(), $localeCode);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getAllCurrenciesData($localeCode)
+    {
+        return $this->formatCurrenciesData($this->installedDataLayer->getAvailableCurrencyCodes(), $localeCode);
+    }
+
+    /**
+     * @param array $currencyCodes
+     * @param string $localeCode
+     *
+     * @return array
+     */
+    protected function formatCurrenciesData(array $currencyCodes, $localeCode)
+    {
         $currenciesData = [];
         foreach ($currencyCodes as $currencyCode) {
             $currenciesData[] = $this->getLocalizedCurrencyData(new LocalizedCurrencyId($currencyCode, $localeCode));
--- a/src/Core/Localization/Currency/DataLayer/CurrencyDatabase.php
+++ b/src/Core/Localization/Currency/DataLayer/CurrencyDatabase.php
@@ -121,6 +121,9 @@ class CurrencyDatabase extends AbstractDataLayer implements CurrencyDataLayerInt
      */
     protected function doWrite($currencyDataId, $currencyData)
     {
-        // Nothing.
+        // We should not save anything in this layer. The CLDR or its Repository nor any of its layers
+        // should modify the database. This could override customization added by the user with default
+        // CLDR values. Any changes on the database must be managed through the backoffice and the appropriate
+        // commands/handlers
     }
 }
--- a/src/Core/Localization/Currency/DataLayer/CurrencyDatabase.php
+++ b/src/Core/Localization/Currency/DataLayer/CurrencyDatabase.php
@@ -27,7 +27,6 @@
 
 namespace PrestaShop\PrestaShop\Core\Localization\Currency\DataLayer;
 
-use Exception;
 use PrestaShop\PrestaShop\Core\Currency\CurrencyDataProviderInterface;
 use PrestaShop\PrestaShop\Core\Data\Layer\AbstractDataLayer;
 use PrestaShop\PrestaShop\Core\Data\Layer\DataLayerException;
@@ -122,23 +121,6 @@ class CurrencyDatabase extends AbstractDataLayer implements CurrencyDataLayerInt
      */
     protected function doWrite($currencyDataId, $currencyData)
     {
-        if (!$currencyDataId instanceof LocalizedCurrencyId) {
-            throw new LocalizationException('First parameter must be an instance of ' . LocalizedCurrencyId::class);
-        }
-
-        $currencyCode = $currencyDataId->getCurrencyCode();
-        $currencyEntity = $this->dataProvider->getCurrencyByIsoCodeOrCreate($currencyCode, $currencyDataId->getLocaleCode());
-
-        $currencyEntity->iso_code = $currencyData->getIsoCode();
-        $currencyEntity->name = $currencyData->getNames()[$currencyDataId->getLocaleCode()];
-        $currencyEntity->numeric_iso_code = $currencyData->getNumericIsoCode();
-        $currencyEntity->symbol = $currencyData->getSymbols()[$currencyDataId->getLocaleCode()];
-        $currencyEntity->precision = $currencyData->getPrecision();
-
-        try {
-            $this->dataProvider->saveCurrency($currencyEntity);
-        } catch (Exception $e) {
-            throw new DataLayerException('Unable to persist data in DB data layer', 0, $e);
-        }
+        // Nothing.
     }
 }
--- a/src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
+++ b/src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
@@ -77,9 +77,9 @@ class CurrencyInstalled
     public function getAvailableCurrencyCodes()
     {
         $currencies = $this->dataProvider->findAll(false, true);
-        $currencyIds = array_column($currencies, 'iso_code');
+        $currencyIsoCodes = array_column($currencies, 'iso_code');
 
-        return $currencyIds;
+        return $currencyIsoCodes;
     }
 
     /**
@@ -90,8 +90,8 @@ class CurrencyInstalled
     public function getAllCurrencyCodes()
     {
         $currencies = $this->dataProvider->findAll(false, false);
-        $currencyIds = array_column($currencies, 'iso_code');
+        $currencyIsoCodes = array_column($currencies, 'iso_code');
 
-        return $currencyIds;
+        return $currencyIsoCodes;
     }
 }
--- a/src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
+++ b/src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
@@ -76,7 +76,20 @@ class CurrencyInstalled
      */
     public function getAvailableCurrencyCodes()
     {
-        $currencies = $this->dataProvider->findAll(false);
+        $currencies = $this->dataProvider->findAll(false, true);
+        $currencyIds = array_column($currencies, 'iso_code');
+
+        return $currencyIds;
+    }
+
+    /**
+     * Get all the available currencies' ISO codes (installed in database no matter if it's deleted or active).
+     *
+     * @return string[]
+     */
+    public function getAllCurrencyCodes()
+    {
+        $currencies = $this->dataProvider->findAll(false, false);
         $currencyIds = array_column($currencies, 'iso_code');
 
         return $currencyIds;
--- a/src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
+++ b/src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
@@ -87,9 +87,9 @@ class CurrencyInstalled
      *
      * @return string[]
      */
-    public function getAllCurrencyCodesInDatabase()
+    public function getAllInstalledCurrencyIsoCodes()
     {
-        $currencies = $this->dataProvider->findAllInDatabase();
+        $currencies = $this->dataProvider->findAllInstalled();
         $currencyIsoCodes = array_column($currencies, 'iso_code');
 
         return $currencyIsoCodes;
--- a/src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
+++ b/src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
@@ -76,20 +76,20 @@ class CurrencyInstalled
      */
     public function getAvailableCurrencyCodes()
     {
-        $currencies = $this->dataProvider->findAll(false, true);
+        $currencies = $this->dataProvider->findAll();
         $currencyIsoCodes = array_column($currencies, 'iso_code');
 
         return $currencyIsoCodes;
     }
 
     /**
-     * Get all the available currencies' ISO codes (installed in database no matter if it's deleted or active).
+     * Get all the available currencies' ISO codes (present in database no matter if it's deleted or active).
      *
      * @return string[]
      */
-    public function getAllCurrencyCodes()
+    public function getAllCurrencyCodesInDatabase()
     {
-        $currencies = $this->dataProvider->findAll(false, false);
+        $currencies = $this->dataProvider->findAllInDatabase();
         $currencyIsoCodes = array_column($currencies, 'iso_code');
 
         return $currencyIsoCodes;
--- a/src/Core/Localization/Currency/DataSourceInterface.php
+++ b/src/Core/Localization/Currency/DataSourceInterface.php
@@ -65,4 +65,15 @@ interface DataSourceInterface
      *                        The available currencies' data
      */
     public function getAvailableCurrenciesData($localeCode);
+
+    /**
+     * Get all the available currencies' data (installed in database no matter if it's deleted or active).
+     *
+     * @param string $localeCode
+     *                           Data will be translated in this language
+     *
+     * @return CurrencyData[]
+     *                        The available currencies' data
+     */
+    public function getAllCurrenciesData($localeCode);
 }
--- a/src/Core/Localization/Currency/DataSourceInterface.php
+++ b/src/Core/Localization/Currency/DataSourceInterface.php
@@ -67,13 +67,13 @@ interface DataSourceInterface
     public function getAvailableCurrenciesData($localeCode);
 
     /**
-     * Get all the available currencies' data (installed in database no matter if it's deleted or active).
+     * Get all the database currencies' data (present in database no matter if it's deleted or active).
      *
      * @param string $localeCode
      *                           Data will be translated in this language
      *
      * @return CurrencyData[]
-     *                        The available currencies' data
+     *                        The database currencies' data
      */
-    public function getAllCurrenciesData($localeCode);
+    public function getAllCurrenciesDataInDatabase($localeCode);
 }
--- a/src/Core/Localization/Currency/DataSourceInterface.php
+++ b/src/Core/Localization/Currency/DataSourceInterface.php
@@ -67,13 +67,13 @@ interface DataSourceInterface
     public function getAvailableCurrenciesData($localeCode);
 
     /**
-     * Get all the database currencies' data (present in database no matter if it's deleted or active).
+     * Get all installed currencies' data in database (regardless of their active or soft deleted status).
      *
      * @param string $localeCode
      *                           Data will be translated in this language
      *
      * @return CurrencyData[]
-     *                        The database currencies' data
+     *                        The installed currencies' database data
      */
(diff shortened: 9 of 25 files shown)


jolelievre:
@eternoendless unlike what we discussed I ended up adding new methods in the interfaces for few reasons:

- function names were confusing as they already were named `*AvailableCurrencies` even becoming contradictory with `isAvailableCurrency`
- even though the CLDR needs all currencies, these interfaces might be used by other services later which will really need *only* available currencies
- the modified interfaces are very recent and should only be used by internal core services for now (probably)
- although we wanted to avoid BC break by adding methods I didn't have a choice for the `CurrencyDataProviderInterface` anyway but to add a new parameter in the prototype which is a BC break already, so I figured if we're forced to add a BC we might as well add it for clean valid reasons

src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php
```diff
@@ -268,6 +268,7 @@ private function updateNameAndSymbol(Currency $entity, $newIsoCode)
                 $langIds = Language::getLanguages(true, false, true);
                 $entity->name = [];
                 $entity->symbol = [];
+                $entity->numeric_iso_code = $currency->getNumericIsoCode();
```
sarjon, line 271: You cast it to `(int)` in `AddCurrencyHandler`, why not casting it here as well?

src/Adapter/Currency/CommandHandler/AddCurrencyHandler.php
```diff
@@ -86,6 +86,7 @@ public function handle(AddCurrencyCommand $command)
             if (!empty($cldrCurrency)) {
                 // The currency may not be declared in the locale, eg with custom iso code
                 $entity->precision = (int) $cldrCurrency->getDecimalDigits();
+                $entity->numeric_iso_code = (int) $cldrCurrency->getNumericIsoCode();
```
sarjon: Shouldn't `getNumericIsoCode()` return `int` instead?

classes/Currency.php
```diff
@@ -402,18 +402,20 @@ public static function getCurrencies($object = false, $active = true, $groupBy =
...
-    public static function findAll($active = true, $groupBy = false, $currentShopOnly = true)
+    public static function findAll($active = true, $groupBy = false, $currentShopOnly = true, $filterDeleted = true)
     {
         $currencies = Db::getInstance()->executeS('
             SELECT *
             FROM `' . _DB_PREFIX_ . 'currency` c
             ' . ($currentShopOnly ? Shop::addSqlAssociation('currency', 'c') : '') . '
-                WHERE `deleted` = 0' .
+                WHERE 1' .
+                ($filterDeleted ? ' AND c.`deleted` = 0' : '') .
```
sarjon: Maybe to would be nice to refactor this SQL to be built using `DbQuery` class?

src/Core/Currency/CurrencyDataProviderInterface.php
```diff
@@ -50,10 +50,11 @@ public function getCurrencies($object = false, $active = true, $group_by = false
      * Return raw currencies data from the database.
      *
      * @param bool $currentShopOnly If true returns only currencies associated to current shop
+     * @param bool $filterDeleted If true deleted currencies are filtered
      *
      * @return array[] Installed currencies
      */
-    public function findAll($currentShopOnly = true);
+    public function findAll($currentShopOnly = true, $filterDeleted = true);
```
sarjon: I don't think it's a good idea to add yet another flag argument. :/

src/Core/Localization/Currency/CurrencyDataSource.php
```diff
@@ -88,17 +82,29 @@ public function isCurrencyAvailable($currencyCode)
...
+     * @param array $currencyCodes
      * @param string $localeCode
-     *                           IETF tag. Data will be translated in this language
      *
-     * @return CurrencyData[]
-     *                        The available currencies' data
+     * @return array
      */
-    public function getAvailableCurrenciesData($localeCode)
+    protected function formatCurrenciesData(array $currencyCodes, $localeCode)
```
sarjon: Since class is not `final`, I'd rather make this method `private` (unless there is a good reason not to), otherwise it's part of public interface.

src/Core/Localization/Currency/DataLayer/CurrencyDatabase.php
```diff
@@ -122,23 +121,6 @@ protected function doRead($currencyDataId)
...
-        $currencyEntity->numeric_iso_code = $currencyData->getNumericIsoCode();
-        $currencyEntity->symbol = $currencyData->getSymbols()[$currencyDataId->getLocaleCode()];
-        $currencyEntity->precision = $currencyData->getPrecision();
-
-        try {
-            $this->dataProvider->saveCurrency($currencyEntity);
-        } catch (Exception $e) {
-            throw new DataLayerException('Unable to persist data in DB data layer', 0, $e);
-        }
+        // Nothing.
```
sarjon: It would be nice to have a better explanation why there is `Nothing.`. :stuck_out_tongue:

src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
```diff
@@ -76,7 +76,20 @@ public function isAvailable($currencyCode)
      */
     public function getAvailableCurrencyCodes()
     {
-        $currencies = $this->dataProvider->findAll(false);
+        $currencies = $this->dataProvider->findAll(false, true);
```
sarjon: These flag arguments does not look good at all. :/

src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
```diff
@@ -76,7 +76,20 @@ public function isAvailable($currencyCode)
...
+        $currencyIds = array_column($currencies, 'iso_code');
```
sarjon: Shouldn't it be `$currencyIsoCodes` instead?

src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
```diff
@@ -76,7 +76,20 @@ public function isAvailable($currencyCode)
...
+
+    /**
+     * Get all the available currencies' ISO codes (installed in database no matter if it's deleted or active).
+     *
+     * @return string[]
+     */
+    public function getAllCurrencyCodes()
+    {
+        $currencies = $this->dataProvider->findAll(false, false);
         $currencyIds = array_column($currencies, 'iso_code');
```
sarjon: Same here, is variable name correct?

src/Core/Localization/Currency/DataSourceInterface.php
```diff
@@ -65,4 +65,15 @@ public function isCurrencyAvailable($currencyCode);
...
+    /**
+     * Get all the available currencies' data (installed in database no matter if it's deleted or active).
+     *
+     * @param string $localeCode
+     *                           Data will be translated in this language
+     *
+     * @return CurrencyData[]
+     *                        The available currencies' data
+     */
+    public function getAllCurrenciesData($localeCode);
```
sarjon: It is kind of BC break to add new method to interface. Is it acceptable?

src/Core/Localization/Currency/RepositoryInterface.php
```diff
@@ -59,4 +59,15 @@ public function getCurrency($currencyCode, $localeCode);
...
+    /**
+     * Get all the available currencies (installed in database no matter if it's deleted or active).
+     *
+     * @param string $localeCode
+     *                           IETF tag. Data will be translated in this language
+     *
+     * @return CurrencyCollection
+     *                            The available currencies
+     */
+    public function getAllCurrencies($localeCode);
```
sarjon: Same here, is it acceptable to add new method to interface?

tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature
```diff
@@ -0,0 +1,58 @@
...
+  Scenario: Display EUR
+    When I add a new language "language1" with following properties:
+      | name             | English (US) |
+      | iso_code         | en           |
+      | language_code    | en-us        |
+      | locale           | en-US        |
+      | date_format_lite | m/d/Y        |
+      | date_format_full | m/d/Y H:i:s  |
+      | is_rtl           | 0            |
+    And I add a new language "language2" with following properties:
```
sarjon: It feels like you are testing **CR**UD operations here, which should be tested somewhere else. What if we had some shortuct `Given`s here? For example:

`Given There is "English" language "lang1"`
`And There is "French" language "lang2"`

Or something similar, so this Behat scenario is focused in testing price display and not CRUD actions, wdyt?

Same for currency.

src/Adapter/Currency/CommandHandler/AddCurrencyHandler.php
Progi1984: In src/Core/Localization/Currency.php, the attribute `$numericIsoCode` is in string, because it stored in 3 characters : https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes#Current_ISO_3166_country_codes

jolelievre: What he said 😅

src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php
jolelievre: Nice catch!

src/Core/Currency/CurrencyDataProviderInterface.php
jolelievre: I agree with you, but initially @eternoendless asked me to add an additional parameter in order to limite the BC break. The first one was introduced in this PR https://github.com/PrestaShop/PrestaShop/pull/15173

In this PR I though that since this is a BC break anyway on our interfaces we might as well "break it properly", that's why I decided to add new clearer functions in the interfaces
But I need @eternoendless approval to validate this change, and if we take upon ourselves to make this change should we clean this interface a bit as well?

src/Core/Localization/Currency/CurrencyDataSource.php
jolelievre: Mmm, I don't see the link with the `final` That's precisely because it is **not** final that a protected function is worth something
But that said I did hesitate about this method to make it private, I thought this class might be extended some day to change the way the data is formatted so I let it protected But for now I agree it has no use and could/should very well be private.. That's a dilemma I can't decide ^^

src/Core/Localization/Currency/DataLayer/CurrencyDatabase.php
jolelievre: Yep, you're right ;)

src/Core/Localization/Currency/DataLayer/CurrencyInstalled.php
jolelievre: As I explained above this was requested, but if we agree on breaking/extending the interfaces then it could be made better

src/Core/Localization/Currency/DataSourceInterface.php
jolelievre: Totally it is, however the alternative we thought of at first with Pablo was to change the behaviour of the `getAvailableCurrencies*` methods and make them return all the currencies in database (wether they are active or deleted) But it bothers me because it is contradictory to the function name and can lead to misunderstanding and unexpected behaviour later

tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature
jolelievre: Yes I added these statements in behat contexts to allow adding a full Language or Currency, I'm not fully used to using behat yet so I'm not sure I understand what you mean by using a `Given` shortcut Do you mean matching the "English" word to adding a Language with hard coded values matching the english language?

What I had in mind when I used/added these statements (from develop branch) was that I didn't want to multiply the entry points to add these fixtures for several reasons:
- to avoid duplicate/multiply the way we add fixtures for the tests
- we have many bug on the CLDR linked to the way the currency are added in database, so I wanted the behat context to be the closest to the real use case
- in 177 some additional tests will need to be added to test customization so these statements will be useful to test this behaviour

That said maybe it is contradictory to the behat way of adding fixtures, I totally understand your aim to focus on the display testing but how could we do it and still meet the other constraints I mentioned?

classes/Currency.php
jolelievre: Yep great idea, I tried to make it quickly But there's a problem because of `Shop::addSqlAssociation` which simply returns the SQL condition, so it would require refacto this part as well
But I'm not sure it's worth it considering this PR scope, and it would risk adding some bugs as well This should be done in a dedicated PR with appropriate tests

matks: Agreed for keeping this outside of this PR scope

tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature
matks: CRUD operations are meaningful in a BO, because when we click on this "submit" button there is a modification of the system and the domain. Even if it's a basic modification (it sets the data) it's worth being tested if in the past it has proven to be unstable 👍

matks:
@jolelievre I see you have used several classes from `develop` into this PR. So when we merge 1.7.6.x into develop, we can expect some nice conflicts 😭 😄

matks:
Code is ✅ for me . If @eternoendless can confirm once more than the BC break is acceptable, then I think we can move on the QA stage.

src/Core/Currency/CurrencyDataProviderInterface.php
```diff
@@ -47,14 +47,21 @@ interface CurrencyDataProviderInterface
...
+     * @return array[] Available currencies
      */
     public function findAll($currentShopOnly = true);
 
+    /**
+     * Return raw currencies data from the database (regardless of their active or deleted status).
+     *
+     * @return array[] Currencies in database
+     */
+    public function findAllInDatabase();
```
PierreRambaud: isn't a BC if the interface has already been implemented?

src/Core/Localization/Currency/CurrencyDataSource.php
```diff
@@ -88,17 +82,29 @@ public function isCurrencyAvailable($currencyCode)
...
+     * @param array $currencyCodes
      * @param string $localeCode
-     *                           IETF tag. Data will be translated in this language
      *
-     * @return CurrencyData[]
-     *                        The available currencies' data
+     * @return array
      */
-    public function getAvailableCurrenciesData($localeCode)
+    private function formatCurrenciesData(array $currencyCodes, $localeCode)
```
PierreRambaud, line 106: It's a BC, do we accept this?

src/Core/Localization/Currency/RepositoryInterface.php
```diff
@@ -59,4 +59,15 @@ public function getCurrency($currencyCode, $localeCode);
...
+    /**
+     * Get all the available currencies (installed in database no matter if it's deleted or active).
+     *
+     * @param string $localeCode
+     *                           IETF tag. Data will be translated in this language
+     *
+     * @return CurrencyCollection
+     *                            The available currencies
+     */
+    public function getAllCurrenciesInDatabase($localeCode);
```
PierreRambaud: Same here

tests/Integration/Behaviour/Features/Context/CurrencyFeatureContext.php
```diff
@@ -63,12 +66,43 @@ public function cleanCurrencyFixtures()
         $this->currencies = [];
     }
 
+    /**
+     * @Given currency :reference with isoCode :isoCode exists
+     */
+    public function createCurrencyWithIsoCode($reference, $isoCode)
+    {
+        //Currency::getIdByIsoCode only returns not deleted currency so we check the storage to avoid duplicate contents
```
PierreRambaud:
```suggestion
        /**
         * Currency::getIdByIsoCode only returns not deleted
         * currency so we check the storage to avoid duplicate contents
         */
```

tests/Integration/Behaviour/Features/Context/CurrencyFeatureContext.php
```diff
@@ -63,12 +66,43 @@ public function cleanCurrencyFixtures()
...
+        if (SharedStorage::getStorage()->exists($reference)) {
+            /** @var Currency $currency */
+            $currency = SharedStorage::getStorage()->get($reference);
+            if ($currency->iso_code == $isoCode) {
+                return;
```
PierreRambaud, line 82: Add a comment to tell the iso currency is already defined.

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
```diff
@@ -0,0 +1,98 @@
...
+     */
+    protected $lastException;
+
+    /**
+     * @BeforeSuite
+     */
+    public static function prepare(BeforeSuiteScope $scope)
+    {
+        // Disable legacy object model cache to prevent conflicts between scenarios.
+        ObjectModel::disableCache();
```
PierreRambaud, line 50: You fear nothing!

tests/Integration/Behaviour/Features/Context/Domain/CurrencyFeatureContext.php
```diff
@@ -0,0 +1,334 @@
...
+                'Currency "%s" has "%s" iso code, but "%s" was expected.',
+                $reference,
+                $currency->iso_code,
+                $isoCode
+            ));
+        }
+    }
+
+    /**
+     * @Then /^currency "(.*)" should have status (enabled|disabled)$/
```
PierreRambaud, line 164: For non greedy.
```suggestion
     * @Then /^currency "(.*?)" should have status (enabled|disabled)$/
```

tests/Integration/Behaviour/Features/Context/Domain/CurrencyFeatureContext.php
```diff
@@ -0,0 +1,334 @@
...
+        $query = new DbQuery();
+        $query->select('c.id_currency');
+        $query->from('currency', 'c');
+        $query->where('active = 0');
+        $query->where('iso_code = \'' . pSQL($isoCode) . '\'');
+
+        $currencyId = (int) Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue($query->build());
+
+        if (!$currencyId) {
+            throw new RuntimeException(sprintf('Currency with ISO Code "%s" should be deleted in database', $isoCode));
```
PierreRambaud: Not sure it's the right message.

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
matks: It actually make sense to disable cache capabilities for such tests.
The funny thing is, I'm sure that it will not disable 100% of the cache 😛 I bet there is some cannot-be-disabled cache behaviors

eternoendless (changes requested):
> Reviewed 2 of 22 files at r1, 11 of 16 files at r2.
> Dismissed @PierreRambaud and @sarjon from 12 discussions.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)* status: 13 of 23 files reviewed, 12 unresolved discussions (waiting on @jolelievre, @PierreRambaud, and @sarjon)

___
*[classes/Currency.php, line 433 at r2](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqeb5QU6KbH2_8onbhm:-Lqeb5QU6KbH2_8onbhn:bbxedss) ([raw file](https://github.com/prestashop/prestashop/blob/8e58b7ca72b4d593ba2b1de4ea21752dd42a58de/classes/Currency.php#L433)):*
> ```PHP
>     public static function findAllInDatabase()
>     {
>         $currencies = Db::getInstance()->executeS(
> ```

Please use `Db::getInstance(_PS_USE_SQL_SLAVE_)` for reads
___
*[classes/Currency.php, line 593 at r2](https://reviewable.io/reviews/prestashop/prestashop/15643#-LqebqfK8mxCf5uONf5w:-LqebqfL8rEzDtlchKqv:beyzjzr) ([raw file](https://github.com/prestashop/prestashop/blob/8e58b7ca72b4d593ba2b1de4ea21752dd42a58de/classes/Currency.php#L593)):*
> ```PHP
>      * @param string $isoCode ISO code
>      * @param int $idShop Shop ID
>      * @param bool $noCache
> ```

Consider changing this parameter to `$forceRefreshCache`

```
     * @param bool $forceRefreshCache [default=false] Se to TRUE to forcefully refresh any currently cached results
```
___
*[src/Adapter/Currency/CommandHandler/AddCurrencyHandler.php, line 89 at r1](https://reviewable.io/reviews/prestashop/prestashop/15643#-LpRrXuPbDv94xNmK6o--r1-89:-LqecfAM-qgBbSuFLp7N:b-46ypdv) ([raw file](https://github.com/prestashop/prestashop/blob/938637199304854214bf50b28ad53682d482adb3/src/Adapter/Currency/CommandHandler/AddCurrencyHandler.php#L89)):*
<details><summary><i>Previously, jolelievre wrote…</i></summary><blockquote>

What he said 😅
</blockquote></details>

This value is a string, not an int, should DEFINITELY NOT be casted to int
___
*[src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php, line 271 at r1](https://reviewable.io/reviews/prestashop/prestashop/15643#-LpRrXuPbDv94xNmK6o0-r1-271:-LqecswCCYjSY9yj6MqB:b-7v5089) ([raw file](https://github.com/prestashop/prestashop/blob/938637199304854214bf50b28ad53682d482adb3/src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php#L271)):*
<details><summary><i>Previously, jolelievre wrote…</i></summary><blockquote>

Nice catch!
</blockquote></details>

It shouldn't be casted to int
___
*[src/Core/Currency/CurrencyDataProviderInterface.php, line 59 at r2](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqediv_6lX65_NX533U:-Lqediv_6lX65_NX533V:bilhjpx) ([raw file](https://github.com/prestashop/prestashop/blob/8e58b7ca72b4d593ba2b1de4ea21752dd42a58de/src/Core/Currency/CurrencyDataProviderInterface.php#L59)):*
> ```
> regardless of their active or deleted status
> ```

"regardless of their active or soft deleted status"
___
*[src/Core/Currency/CurrencyDataProviderInterface.php, line 63 at r2](https://reviewable.io/reviews/prestashop/prestashop/15643#-LpRrXuPbDv94xNmK6o2-r2-63:-Lqedmz05JVWPu-dEeKC:b-r1ii7) ([raw file](https://github.com/prestashop/prestashop/blob/8e58b7ca72b4d593ba2b1de4ea21752dd42a58de/src/Core/Currency/CurrencyDataProviderInterface.php#L63)):*
<details><summary><i>Previously, PierreRambaud (GoT) wrote…</i></summary><blockquote>

isn't a BC if the interface has already been implemented?
</blockquote></details>

Unfortunately it's necessary. It's unlikely it's been reimplemented, but it has to be duly noted in the release notes.
___
*[src/Core/Localization/Currency/DataSourceInterface.php, line 78 at r1](https://reviewable.io/reviews/prestashop/prestashop/15643#-LpRrXuPbDv94xNmK6o6-r1-78:-Lqeenfb42MNzHcRnOf9:b-wuaxoo) ([raw file](https://github.com/prestashop/prestashop/blob/938637199304854214bf50b28ad53682d482adb3/src/Core/Localization/Currency/DataSourceInterface.php#L78)):*
<details><summary><i>Previously, jolelievre wrote…</i></summary><blockquote>

Totally it is, however the alternative we thought of at first with Pablo was to change the behaviour of the `getAvailableCurrencies*` methods and make them return all the currencies in database (wether they are active or deleted) But it bothers me because it is contradictory to the function name and can lead to misunderstanding and unexpected behaviour later
</blockquote></details>

I'm uncomfortable with the `InDatabase` part here as this interface is used for any kind of resource. Maybe `getAllCurrenciesDataIncludingDeleted($localeCode)`?

src/Core/Currency/CurrencyDataProviderInterface.php
jolelievre: Sadly yes, we discussed it with @eternoendless and couldn't find a clean way to fix this without BC break We could have kept the initial interface and modify their behaviour (first thought) but then the interfaces didn't make sense any more
Although this interface should be internal, there are very few chances of it being used outside of the core so the BC is "contained"

src/Core/Localization/Currency/CurrencyDataSource.php
jolelievre: The BC here is the added method, same problem as the previous comment

src/Core/Localization/Currency/RepositoryInterface.php
jolelievre: Same answer ^^

tests/Integration/Behaviour/Features/Context/CurrencyFeatureContext.php
jolelievre: Done

tests/Integration/Behaviour/Features/Context/CurrencyFeatureContext.php
jolelievre: Added in the previous comment

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
jolelievre: I actually copied this code from the develop branch without even reading it 😅
But I can guarantee you @matks that it is indeed not enough.. for example I had to call `Currency::getIdByIsoCode` with a `noCache` parameter to avoid unwanted behaviour Although in that case the `Cache` class was used

That said although disabling the cache like this is handy, it may give us false positive tests, because everything will work fine in the test but when cache is enabled in real context then unexpected behaviour will occur I don't really know what the ObjectModel cache is supposed to do so I can't imagine the consequences, but maybe it would be safer to disable it case by case when we really need it than disabling it altogether?

tests/Integration/Behaviour/Features/Context/Domain/CurrencyFeatureContext.php
jolelievre: I don't think it's really needed here thanks to the enclosing `""` Besides all conditions are written that way for now

tests/Integration/Behaviour/Features/Context/Domain/CurrencyFeatureContext.php
jolelievre: True! Dumb copy/paste ^^

jolelievre:
> @jolelievre I see you have used several classes from `develop` into this PR. So when we merge 1.7.6.x into develop, we can expect some nice conflicts 😭 😄

Indeed I needed them to add behat tests in this branch Although some of them are exact copy so there won't be much conflict But for the other one fear not! I will take care of the merge and conflict fixing 😉
I need these modifications to start a pending issue anyway so as son as this one is merged I will merge 176x back to develop

jolelievre (commented):
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)* status: 13 of 23 files reviewed, 14 unresolved discussions (waiting on @eternoendless, @PierreRambaud, and @sarjon)

___
*[classes/Currency.php, line 433 at r2](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqeb5QU6KbH2_8onbhm:-Lqeul8G4lD06kacJgf0:b-crrrfo) ([raw file](https://github.com/prestashop/prestashop/blob/8e58b7ca72b4d593ba2b1de4ea21752dd42a58de/classes/Currency.php#L433)):*
<details><summary><i>Previously, eternoendless (Pablo Borowicz) wrote…</i></summary><blockquote>

Please use `Db::getInstance(_PS_USE_SQL_SLAVE_)` for reads
</blockquote></details>

I didn't know about this. But I don't understand it does really make sense.. I mean the const `_PS_USE_SQL_SLAVE_` implies that we use slave instead of master, but the  `Db::getInstance` has `$master = true` as a parameter So logically to use the slave server accordingly with the const we should call `Db::getInstance(!_PS_USE_SQL_SLAVE_)` or am I missing something? Because this call is present everywhere in the codebase
___
*[classes/Currency.php, line 593 at r2](https://reviewable.io/reviews/prestashop/prestashop/15643#-LqebqfK8mxCf5uONf5w:-LqevtB53RFE2M3XrdjY:b-896fix) ([raw file](https://github.com/prestashop/prestashop/blob/8e58b7ca72b4d593ba2b1de4ea21752dd42a58de/classes/Currency.php#L593)):*
<details><summary><i>Previously, eternoendless (Pablo Borowicz) wrote…</i></summary><blockquote>

Consider changing this parameter to `$forceRefreshCache`

```
     * @param bool $forceRefreshCache [default=false] Se to TRUE to forcefully refresh any currently cached results
```
</blockquote></details>

Done.
___
*[src/Adapter/Currency/CommandHandler/AddCurrencyHandler.php, line 89 at r1](https://reviewable.io/reviews/prestashop/prestashop/15643#-LpRrXuPbDv94xNmK6o--r1-89:-LqewJ8z8HDMonhKSy_A:bnllohu) ([raw file](https://github.com/prestashop/prestashop/blob/938637199304854214bf50b28ad53682d482adb3/src/Adapter/Currency/CommandHandler/AddCurrencyHandler.php#L89)):*
<details><summary><i>Previously, eternoendless (Pablo Borowicz) wrote…</i></summary><blockquote>

This value is a string, not an int, should DEFINITELY NOT be casted to int
</blockquote></details>

My bad you're right! Done
___
*[src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php, line 271 at r1](https://reviewable.io/reviews/prestashop/prestashop/15643#-LpRrXuPbDv94xNmK6o0-r1-271:-LqewPAp6W6EurIKpyYz:b-896fix) ([raw file](https://github.com/prestashop/prestashop/blob/938637199304854214bf50b28ad53682d482adb3/src/Adapter/Currency/CommandHandler/EditCurrencyHandler.php#L271)):*
<details><summary><i>Previously, eternoendless (Pablo Borowicz) wrote…</i></summary><blockquote>

It shouldn't be casted to int
</blockquote></details>

Done.
___
*[src/Core/Currency/CurrencyDataProviderInterface.php, line 59 at r2](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqediv_6lX65_NX533U:-LqewWiaAJTo73L4oV19:b-896fix) ([raw file](https://github.com/prestashop/prestashop/blob/8e58b7ca72b4d593ba2b1de4ea21752dd42a58de/src/Core/Currency/CurrencyDataProviderInterface.php#L59)):*
<details><summary><i>Previously, eternoendless (Pablo Borowicz) wrote…</i></summary><blockquote>

> ```
> regardless of their active or deleted status
> ```

"regardless of their active or soft deleted status"
</blockquote></details>

Done.
___
*[src/Core/Localization/Currency/DataSourceInterface.php, line 78 at r1](https://reviewable.io/reviews/prestashop/prestashop/15643#-LpRrXuPbDv94xNmK6o6-r1-78:-LqewuXa7EIHXyJOpdKX:b-45s0u6) ([raw file](https://github.com/prestashop/prestashop/blob/938637199304854214bf50b28ad53682d482adb3/src/Core/Localization/Currency/DataSourceInterface.php#L78)):*
> getAllCurrenciesDataIncludingDeleted
I get your point of view, but this starts to be a really long method name.. I wasn't fan of `getAllCurrenciesDataInDatabase` either I'd like to call it `getAllCurrenciesData` but as you said this could be interpreted as all the existing currencies in the world..
So maybe `getInstalledCurrenciesData` or `getAllInstalledCurrenciesData` could be nice? With the right PHPDoc on top of it I think it leaves out the confusion but remains relatively short
___
*[tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature, line 17 at r1](https://reviewable.io/reviews/prestashop/prestashop/15643#-LpRrXuQCZ-PR9Zh-nwS-r1-17:-Lqey0BpAdBmAcfNJoZu:b-wq5zky) ([raw file](https://github.com/prestashop/prestashop/blob/938637199304854214bf50b28ad53682d482adb3/tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature#L17)):*
<details><summary><i>Previously, matks (Mathieu Ferment) wrote…</i></summary><blockquote>

CRUD operations are meaningful in a BO, because when we click on this "submit" button there is a modification of the system and the domain. Even if it's a basic modification (it sets the data) it's worth being tested if in the past it has proven to be unstable 👍 
</blockquote></details>

I actually ended up siding with @sarjon proposal in order to have cleaner tests in this cldr part But in the coming issues for 177 linked to currency and their customization I intend to add a few CLDR display tests inside the currency management scenarii So this will validate the full chain nicely

eternoendless (changes requested):
> Reviewed 1 of 16 files at r2, 6 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)* status: 15 of 24 files reviewed, 10 unresolved discussions (waiting on @jolelievre, @PierreRambaud, and @sarjon)

___
*[classes/Currency.php, line 433 at r2](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqeb5QU6KbH2_8onbhm:-LqfFFkkCub_v3cEoYMg:b-9fg04i) ([raw file](https://github.com/prestashop/prestashop/blob/8e58b7ca72b4d593ba2b1de4ea21752dd42a58de/classes/Currency.php#L433)):*
<details><summary><i>Previously, jolelievre wrote…</i></summary><blockquote>

I didn't know about this. But I don't understand it does really make sense.. I mean the const `_PS_USE_SQL_SLAVE_` implies that we use slave instead of master, but the  `Db::getInstance` has `$master = true` as a parameter So logically to use the slave server accordingly with the const we should call `Db::getInstance(!_PS_USE_SQL_SLAVE_)` or am I missing something? Because this call is present everywhere in the codebase
</blockquote></details>

I agree this is counterintuitive, but it looks like this constant does the opposite of what its name would indicate: https://github.com/PrestaShop/PrestaShop/blob/develop/config/defines.inc.php#L179
___
*[src/Core/Localization/Currency/DataSourceInterface.php, line 78 at r1](https://reviewable.io/reviews/prestashop/prestashop/15643#-LpRrXuPbDv94xNmK6o6-r1-78:-LqfGlW544FCZCtHq5fk:be6ckh3) ([raw file](https://github.com/prestashop/prestashop/blob/938637199304854214bf50b28ad53682d482adb3/src/Core/Localization/Currency/DataSourceInterface.php#L78)):*
<details><summary><i>Previously, jolelievre wrote…</i></summary><blockquote>

> getAllCurrenciesDataIncludingDeleted
I get your point of view, but this starts to be a really long method name.. I wasn't fan of `getAllCurrenciesDataInDatabase` either I'd like to call it `getAllCurrenciesData` but as you said this could be interpreted as all the existing currencies in the world..
So maybe `getInstalledCurrenciesData` or `getAllInstalledCurrenciesData` could be nice? With the right PHPDoc on top of it I think it leaves out the confusion but remains relatively short
</blockquote></details>

Okay that sounds good

jolelievre (commented):
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)* status: 15 of 24 files reviewed, 10 unresolved discussions (waiting on @eternoendless, @PierreRambaud, and @sarjon)

___
*[classes/Currency.php, line 433 at r2](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqeb5QU6KbH2_8onbhm:-Lqf_vgH7ENwPTsUga0z:bcsob1x) ([raw file](https://github.com/prestashop/prestashop/blob/8e58b7ca72b4d593ba2b1de4ea21752dd42a58de/classes/Currency.php#L433)):*
<details><summary><i>Previously, eternoendless (Pablo Borowicz) wrote…</i></summary><blockquote>

I agree this is counterintuitive, but it looks like this constant does the opposite of what its name would indicate: https://github.com/PrestaShop/PrestaShop/blob/develop/config/defines.inc.php#L179
</blockquote></details>

Yeah I saw the define but it's still counter-intuitive ^^ So, considering this default definition, actually the slave connection is used by default? Maybe a better comment would be useful here?
___
*[src/Core/Localization/Currency/DataSourceInterface.php, line 78 at r1](https://reviewable.io/reviews/prestashop/prestashop/15643#-LpRrXuPbDv94xNmK6o6-r1-78:-LqfaGUfAsh-B9qKeJ5q:b42q08p) ([raw file](https://github.com/prestashop/prestashop/blob/938637199304854214bf50b28ad53682d482adb3/src/Core/Localization/Currency/DataSourceInterface.php#L78)):*
> getAllInstalledCurrenciesData

Alright I'll go with `getAllInstalledCurrenciesData` then

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
PierreRambaud: Disabling the cache means it's not the same behavior as the front :thinking: 
Hope everything's going well!

tests/Integration/Behaviour/Features/Context/Domain/CurrencyFeatureContext.php
PierreRambaud: No prob, hope we don't add a `@Then` statement in the future which will come into conflict :)

matks (dismissed):

tests/Integration/Behaviour/Features/Context/Domain/CurrencyFeatureContext.php
jolelievre: Sure, we can always change the regex when that occurs

app/config/set_parameters.php
```diff
@@ -59,5 +59,8 @@
     }
 
     $container->setParameter('cache.driver', extension_loaded('apc') ? 'apc': 'array');
+
+    // Parameter used only in dev and test env
+    $container->setParameter('use_debug_toolbar', false !== getenv('DISABLE_DEBUG_TOOLBAR') ? ! (bool) getenv('DISABLE_DEBUG_TOOLBAR') : true);
```
PierreRambaud:
```suggestion
    $container->setParameter('use_debug_toolbar', !getenv('DISABLE_DEBUG_TOOLBAR'));
```

jolelievre: Right 😅

eternoendless (commented):
I think the "disable debug toolbar in test environment" feature should be an a different PR

> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643#-:-Lqp2O7w0RI9QMWyWn7M:b-zcaei)* status: 6 of 28 files reviewed, 11 unresolved discussions (waiting on @eternoendless, @jolelievre, @matks, @PierreRambaud, and @sarjon)

eternoendless (changes requested):
> Reviewed 1 of 22 files at r1, 1 of 16 files at r2, 9 of 9 files at r4, 1 of 1 files at r6.
> Dismissed @PierreRambaud and @sarjon from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)* status: 17 of 24 files reviewed, 10 unresolved discussions (waiting on @jolelievre, @matks, @PierreRambaud, and @sarjon)

___
*[classes/Currency.php, line 433 at r2](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqeb5QU6KbH2_8onbhm:-Lqp2_qX4TW9uomP9-et:b9ldg4) ([raw file](https://github.com/prestashop/prestashop/blob/8e58b7ca72b4d593ba2b1de4ea21752dd42a58de/classes/Currency.php#L433)):*
<details><summary><i>Previously, jolelievre wrote…</i></summary><blockquote>

Yeah I saw the define but it's still counter-intuitive ^^ So, considering this default definition, actually the slave connection is used by default? Maybe a better comment would be useful here?
</blockquote></details>

If you pass the setting, then yes, by default it will use the slave server (as long as it's configured).
___
*[tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature, line 11 at r6](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqpa-6RBMNJIpgESpMh:-Lqpa-6RBMNJIpgESpMi:b-jfsdy4) ([raw file](https://github.com/prestashop/prestashop/blob/bfbeff567e426ccd53a40ca2a7bfb006ca7c3584/tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature#L11)):*
> ```Gherkin
>     Given language "language1" with locale "en-US" exists
>     Given language "language2" with locale "fr-FR" exists
>     Given currency "currency1" with isoCode "USD" exists
> ```

Be consistent: "ISO code" and not "isoCode"
___
*[tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature, line 14 at r6](https://reviewable.io/reviews/prestashop/prestashop/15643#-LqpZOwO6yjA7QUHBf_b:-LqpZOwO6yjA7QUHBf_c:b-gyuq67) ([raw file](https://github.com/prestashop/prestashop/blob/bfbeff567e426ccd53a40ca2a7bfb006ca7c3584/tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature#L14)):*
> ```Gherkin
>     Given currency "currency2" with isoCode "EUR" exists
>     Given currency "currency3" with isoCode "AUD" exists
>     And I delete currency "currency2"
> ```

Any action should be a When

If it's a state then it should be a Given:`Given the currency "currency2" has been deleted`

Also, it's confusing that currencies are sometimes identified using an identifier like "currency2" and sometimes using ISO code.
___
*[tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature, line 18 at r6](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqp_IzD-dUNiSnuipYi:-Lqp_IzD-dUNiSnuipYj:b9av3q5) ([raw file](https://github.com/prestashop/prestashop/blob/bfbeff567e426ccd53a40ca2a7bfb006ca7c3584/tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature#L18)):*
> ```Gherkin
> 
>   Scenario: Display USD
>     Then display a price of 14789.5426 "USD" with locale "en-US" should look like "$14,789.54"
> ```

I think this wording can be improved:

```suggestion
    Then a price of 14789.5426 using "USD" in locale "en-US" should look like "$14,789.54"
```
___
*[tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature, line 22 at r6](https://reviewable.io/reviews/prestashop/prestashop/15643#-LqpZjHo5I-OKniN0Eil:-LqpZjHo5I-OKniN0Eim:b-iixsg5) ([raw file](https://github.com/prestashop/prestashop/blob/bfbeff567e426ccd53a40ca2a7bfb006ca7c3584/tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature#L22)):*
> ```Gherkin
> 
>   Scenario: Display a deleted EUR currency
>     And there should be 1 currencies of "EUR"
> ```

What is this, a given? a when? -> Not a good idea to start a scenario with "and" because it's ambiguous.

Also, I think the wording can be clarified. 

- Where is that instance? (database/installed/active?)
- How could we have more than one instance of the same ISO currency at the same time?

jolelievre (commented):
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)* status: 16 of 24 files reviewed, 10 unresolved discussions (waiting on @eternoendless, @matks, @PierreRambaud, and @sarjon)

___
*[tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature, line 11 at r6](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqpa-6RBMNJIpgESpMh:-Lqq1SI1DkMM0dN6u-4S:b-896fix) ([raw file](https://github.com/prestashop/prestashop/blob/bfbeff567e426ccd53a40ca2a7bfb006ca7c3584/tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature#L11)):*
<details><summary><i>Previously, eternoendless (Pablo Borowicz) wrote…</i></summary><blockquote>

Be consistent: "ISO code" and not "isoCode"
</blockquote></details>

Done.
___
*[tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature, line 14 at r6](https://reviewable.io/reviews/prestashop/prestashop/15643#-LqpZOwO6yjA7QUHBf_b:-Lqq1gMmBKgZohBeRTnU:b-w4wmud) ([raw file](https://github.com/prestashop/prestashop/blob/bfbeff567e426ccd53a40ca2a7bfb006ca7c3584/tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature#L14)):*
<details><summary><i>Previously, eternoendless (Pablo Borowicz) wrote…</i></summary><blockquote>

Any action should be a When

If it's a state then it should be a Given:`Given the currency "currency2" has been deleted`

Also, it's confusing that currencies are sometimes identified using an identifier like "currency2" and sometimes using ISO code.
</blockquote></details>

There is a "has been deleted" statement but it is used to check that the currency is indeed deleted, whereas this "I delete" statement is used as an action to delete the currency
And the identifier are used to access the objects save in the `SharedStorage` whereas the iso code used further is just a parameter of the Locale::formatPrice function, or it is used to check the data present in the database
Which is why both notations exist and are useful for different things
(I just understood what you meant so I'll use a When instead ^^)
___
*[tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature, line 18 at r6](https://reviewable.io/reviews/prestashop/prestashop/15643#-Lqp_IzD-dUNiSnuipYi:-Lqq51TFEkf854Bu_iTd:b-896fix) ([raw file](https://github.com/prestashop/prestashop/blob/bfbeff567e426ccd53a40ca2a7bfb006ca7c3584/tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature#L18)):*
<details><summary><i>Previously, eternoendless (Pablo Borowicz) wrote…</i></summary><blockquote>

I think this wording can be improved:

```suggestion
    Then a price of 14789.5426 using "USD" in locale "en-US" should look like "$14,789.54"
```
</blockquote></details>

Done.
___
*[tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature, line 22 at r6](https://reviewable.io/reviews/prestashop/prestashop/15643#-LqpZjHo5I-OKniN0Eil:-Lqq57Kl7t6q0AVHkpXt:b-ualrxb) ([raw file](https://github.com/prestashop/prestashop/blob/bfbeff567e426ccd53a40ca2a7bfb006ca7c3584/tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature#L22)):*
<details><summary><i>Previously, eternoendless (Pablo Borowicz) wrote…</i></summary><blockquote>

What is this, a given? a when? -> Not a good idea to start a scenario with "and" because it's ambiguous.

Also, I think the wording can be clarified. 

- Where is that instance? (database/installed/active?)
- How could we have more than one instance of the same ISO currency at the same time?
</blockquote></details>

It's a test on the database content, so it's a state I replaced it with a `Given` here, and used a new sentence `Given database contains 1 rows of currency "EUR"`

About your question, it is actually possible to have several rows with the same ISO code, if you remove a currency and then install it again you will have two rows (one deleted the other not) Although it's not a great behaviour that's how PrestaShop works for now (this could be improved in another issue)

That said there was a weird bug I talked to you about, when you delete a currency but then just try to display it, with this new fix that allows to display deleted currencies, then the CLDR service automatically reinserted a new currency in the database This was because of the `CurrencyDatabase` layer which performed write operations when he shouldn't have I removed this behaviour and this test is here to assess that we never (hopefully) do the same mistake

PierreRambaud (commented):
> Reviewed 2 of 5 files at r7.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)* status: 17 of 25 files reviewed, 6 unresolved discussions (waiting on @eternoendless, @matks, and @sarjon)

PierreRambaud (commented):
> Reviewed 3 of 22 files at r1, 1 of 16 files at r2, 1 of 8 files at r3, 1 of 5 files at r7.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)* status: 23 of 25 files reviewed, 5 unresolved discussions (waiting on @eternoendless and @sarjon)

eternoendless (approved):
> Reviewed 2 of 5 files at r7.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)* status: all files reviewed, 1 unresolved discussion (waiting on @sarjon)

___
*[tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature, line 22 at r6](https://reviewable.io/reviews/prestashop/prestashop/15643#-LqpZjHo5I-OKniN0Eil:-Lr8geKOCZY0Pn025qld:b3lff8h) ([raw file](https://github.com/prestashop/prestashop/blob/bfbeff567e426ccd53a40ca2a7bfb006ca7c3584/tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature#L22)):*
> if you remove a currency and then install it again you will have two rows (one deleted the other not) Although it's not a great behaviour that's how PrestaShop works for now (this could be improved in another issue)

I agree, this should be fixed in another PR on develop. Could you please create the issue?

eternoendless (approved):
> Dismissed @sarjon from a discussion.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15643)* status: :shipit: complete! all files reviewed, all discussions resolved

jolelievre:
@eternoendless I created the issue about duplicated rows: https://github.com/PrestaShop/PrestaShop/issues/15934

PierreRambaud:
Thanks @jolelievre

PR #15651, opened 2019-09-23
Add common FO page - Puppeteer

nesrineabdmouleh:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add FObasePage file that contains the common functions used in FO
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | npm run sanity-tests

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15651)

### DIFF
--- a/tests/puppeteer/pages/FO/FObasePage.js
+++ b/tests/puppeteer/pages/FO/FObasePage.js
@@ -21,19 +21,19 @@ module.exports = class Home extends CommonPage {
   }
 
   /**
-   * Filter by category
+   * Go to category
    * @param categoryID, category id from the BO
    */
-  async filterByCategory(categoryID) {
+  async goToCategory(categoryID) {
     await this.waitForSelectorAndClick(this.categoryMenu.replace('%ID', categoryID));
   }
 
   /**
-   * Filter by subcategory
+   * Go to subcategory
    * @param categoryID, category id from the BO
    * @param subCategoryID, subcategory id from the BO
    */
-  async filterSubCategory(categoryID, subCategoryID) {
+  async goToSubCategory(categoryID, subCategoryID) {
     await this.page.hover(this.categoryMenu.replace('%ID', categoryID));
     await this.waitForSelectorAndClick(this.categoryMenu.replace('%ID', subCategoryID));
   }
--- a/tests/puppeteer/pages/FO/FObasePage.js
+++ b/tests/puppeteer/pages/FO/FObasePage.js
@@ -1,0 +1,59 @@
+const CommonPage = require('../commonPage');
+
+module.exports = class Home extends CommonPage {
+  constructor(page) {
+    super(page);
+
+    // Selectors for home page
+    this.desktopLogo = '#_desktop_logo';
+    this.cartProductsCount = '#_desktop_cart span.cart-products-count';
+    this.userInfoLink = '#_desktop_user_info';
+    this.logoutLink = `${this.userInfoLink} .user-info a.logout`;
+    this.contactLink = '#contact-link';
+    this.categoryMenu = '#category-%ID > a';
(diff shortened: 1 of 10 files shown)


tests/puppeteer/pages/FO/FObasePage.js
```diff
@@ -0,0 +1,59 @@
...
+   */
+  async goToHomePage() {
+    await this.waitForSelectorAndClick(this.desktopLogo);
+  }
+
+  /**
+   * Filter by category
+   * @param categoryID, category id from the BO
+   */
+  async filterByCategory(categoryID) {
```
SimonGrn: This is a method to filter the main page, right ? So it shouldn't be here, it should be in homePage, not in the BasePage. BasePage is for stuff we can access for *every* FO page.

SimonGrn (commented):
Thanks @nesrineabdmouleh

tests/puppeteer/pages/FO/FObasePage.js
nesrineabdmouleh: We can filter categories from every FO page not only from home page.

SimonGrn: Not from CMS pages, from the Contact Us page, from the account page, login page... :smile:

SimonGrn: What we can do is create a CatalogPage which can contain these methods. Then you can have a HomePage which extends the CatalogPage ?

tests/puppeteer/pages/FO/home.js
```diff
@@ -29,63 +23,18 @@ module.exports = class Home extends CommonPage {
...
   }
 
   /**
    * Go to the product page
    * @param id, product id
    */
   async goToProductPage(id) {
-    await this.page.waitForSelector(this.logoHomePage, {visible: true});
+    await this.page.waitForSelector(this.productImg.replace('%NUMBER', id));
     await this.waitForSelectorAndClick(this.productImg.replace('%NUMBER', id), 5000);
```
boubkerbribri, line 35: 2 times waiting for the same selector `this.productImg`

tests/puppeteer/pages/FO/FObasePage.js
SimonGrn: Oops my bad you're right for these filters, even if they don't make really a lot of sense in CMS pages or Login page...
My only concern right now is for the name, since it's not really a filter but more of a link to a Category/Subcategory: `goToCategory` or `goToSubCategory` maybe ?

tests/puppeteer/pages/FO/home.js
```diff
@@ -29,63 +23,18 @@ module.exports = class Home extends CommonPage {
...
-    await this.page.waitForSelector(this.logoHomePage, {visible: true});
-  }
-
-  /**
-   * go to the home page
-   */
-  async goToHomePage() {
-    await this.waitForSelectorAndClick(this.logoHomePage);
-    this.page.waitForNavigation({waitUntil: 'networkidle0'});
+    await this.page.waitForSelector(this.productImg.replace('%NUMBER', '1'), {visible: true});
```
boubkerbribri: Maybe waiting for something that represent the home Page 
example : `body#index` or `section#content.page-home`

tests/puppeteer/pages/FO/FObasePage.js
nesrineabdmouleh: Yes, it makes more sense to change the names of the functions ;)

matthieu-rolland (approved):
lgtm, thank ou @nesrineabdmouleh

Issue #15559, opened 2019-09-16
Main menu module: a new link is created when editing a previous link


01park:
In the main menu module, when editing a previous link, after saving a new link is created and the old one has not changed. 

Work around; remove previous link and replace it with new details.

PS: 1.7.6.1
PHP 7.2
Module version: up to date.

mariem-abid:
Hi @01park,

Could you please try to clear cache from your project and your navigator and feedback.

01park:
@mariem-abid 

This I already tried, but does not helped

Maybe its a multi-store issue.

mariem-abid:
@01park , I manage to reproduce the issue with the last version PS:1.7.6.1:

**To Reproduce**
Steps to reproduce the behavior:
1. Add a new link.
2. Click on Edit button.
3. Edit the label.
=> The update button is not existing in the PS version 1.7.6.1 (PS:1.7.6.0 is OK)
I attached a video record.
https://drive.google.com/file/d/10RGsHOtebxjVkICoY2vCs2BLOe2seNp4/view

I’ll add this to the debug roadmap so that it’s fixed. If you have already fixed it on your end or if you think you can do it, please do send us a pull request!
Thanks!

mariem-abid:
related to #15564

PR #15658, opened 2019-09-23
Use empty legacy action parameter

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | When building legacy urls the action needs to remain empty. The previous used value (1 or `true`) didn't work because in some modules check that the parameter is present AND that it's empty
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #15559
| How to test?  | See the video in the issue, try to edit a link in `ps_mainmenu` module without this PR the module shows you a form to edit but can't handle it

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15658)

### DIFF
--- a/src/Adapter/Routing/LegacyHelperLinkBuilder.php
+++ b/src/Adapter/Routing/LegacyHelperLinkBuilder.php
@@ -87,12 +87,15 @@ class LegacyHelperLinkBuilder implements EntityLinkBuilderInterface
     {
         unset($parameters['current_index']);
         $actionParameter = $action . $entity;
+
+        /**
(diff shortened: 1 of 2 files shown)


PierreRambaud (dismissed):

PierreRambaud (approved):

sarahdib:
Thanks @jolelievre  it's Ok for me : https://drive.google.com/file/d/1OythNpeK8zQXaqHoUcl5oQ0fG1dhvgmO/view

PierreRambaud:
Thanks @jolelievre

PR #15685, opened 2019-09-25
Unable to create new supplier when VAT identification number is mandatory

### LINKED ISSUE #15622: Unable to create new supplier when VAT identification number is mandatory
If we have the VAT identification number required in countries, a new provider cannot be created.

**To Reproduce**
Go to International -> Locations -> Countries 
Edit Spain (España) 
Switch "Do you need a VAT identification number?" to yes.
Go to  Catalog -> Brands & Suppliers  -> Suppliers 
Add new supplier

image

**Additional information**
PrestaShop version: 1.7.6.1
PHP version: 7.2.2

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Some countries such as Spain or Argentina require to have a Documento Nacional de Identidad (National Identity Document also named DNI). Add the capability to check if the country needs a DNI during the supplier process.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15622 
| How to test?  | You should be able to add and edit Supplier with or without DNI field. (Spain has DNI).

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15685)

### DIFF
--- a/admin-dev/themes/default/template/helpers/form/form.tpl
+++ b/admin-dev/themes/default/template/helpers/form/form.tpl
@@ -948,6 +948,11 @@
 				}
 			{/if}
 
+			dniRequired();
+			$('#id_country').change(function() {
+				dniRequired();
+			});
+
 			if ($(".datepicker").length > 0)
 				$(".datepicker").datepicker({
 					prevText: '',
@@ -977,7 +982,8 @@
 			$(".textarea-autosize").autosize();
 			{/if}
 		});
-	state_token = '{getAdminToken tab='AdminStates'}';
+		state_token = '{getAdminToken tab='AdminStates'}';
+		address_token = '{getAdminToken tab='AdminAddresses'}';
 	{block name="script"}{/block}
 	</script>
 {/if}
--- a/admin-dev/themes/default/template/helpers/form/form.tpl
+++ b/admin-dev/themes/default/template/helpers/form/form.tpl
@@ -949,9 +949,7 @@
 			{/if}
 
 			dniRequired();
-			$('#id_country').change(function() {
-				dniRequired();
-			});
+			$('#id_country').change(dniRequired);
 
 			if ($(".datepicker").length > 0)
 				$(".datepicker").datepicker({
--- a/admin-dev/themes/default/template/helpers/form/form.tpl
+++ b/admin-dev/themes/default/template/helpers/form/form.tpl
@@ -67,7 +67,7 @@
 					<div class="form-wrapper">
 					{foreach $field as $input}
 						{block name="input_row"}
-						<div class="form-group{if isset($input.form_group_class)} {$input.form_group_class}{/if}{if $input.type == 'hidden'} hide{/if}"{if $input.name == 'id_state'} id="contains_states"{if !$contains_states} style="display:none;"{/if}{/if}{if isset($tabs) && isset($input.tab)} data-tab-id="{$input.tab}"{/if}>
+						<div class="form-group{if isset($input.form_group_class)} {$input.form_group_class}{/if}{if $input.type == 'hidden'} hide{/if}"{if $input.name == 'id_state'} id="contains_states"{if !$contains_states} style="display:none;"{/if}{/if}{if $input.name == 'dni'} id="contains_dni"{if !$contains_dni} style="display:none;"{/if}{/if}{if isset($tabs) && isset($input.tab)} data-tab-id="{$input.tab}"{/if}>
 						{if $input.type == 'hidden'}
 							<input type="hidden" name="{$input.name}" id="{$input.name}" value="{$fields_value[$input.name]|escape:'html':'UTF-8'}" />
 						{else}
--- a/admin-dev/themes/default/template/helpers/form/form.tpl
+++ b/admin-dev/themes/default/template/helpers/form/form.tpl
@@ -67,7 +67,7 @@
 					<div class="form-wrapper">
 					{foreach $field as $input}
 						{block name="input_row"}
-						<div class="form-group{if isset($input.form_group_class)} {$input.form_group_class}{/if}{if $input.type == 'hidden'} hide{/if}"{if $input.name == 'id_state'} id="contains_states"{if !$contains_states} style="display:none;"{/if}{/if}{if $input.name == 'dni'} id="contains_dni"{if !$contains_dni} style="display:none;"{/if}{/if}{if isset($tabs) && isset($input.tab)} data-tab-id="{$input.tab}"{/if}>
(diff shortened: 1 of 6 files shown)


controllers/admin/AdminAddressesController.php
```diff
@@ -517,6 +517,11 @@ public function ajaxProcess()
                 echo json_encode(array('infos' => pSQL($customer['firstname']) . '_' . pSQL($customer['lastname']) . '_' . pSQL($customer['company'])));
             }
         }
+
+        if (Tools::isSubmit('contains_dni')) {
```
PierreRambaud: based on what is done on `states`

js/admin.js
```diff
@@ -1381,6 +1381,20 @@ function ajaxStates(id_state_selected)
...
   });
 
+  $.ajax({
+    url: 'index.php',
+    dataType: 'json',
+    cache: false,
+    data: 'token=' + state_token + '&ajax=1&contains_dni=1&tab=AdminAddresses&id_country=' + $('#id_country').val(),
+    success: function(resp) {
+      if (resp && resp.contains_dni) {
+        $("#contains_dni").fadeIn();
```
Progi1984: @PierreRambaud Why do you use fadeIn/fadeOut while show/hide are used in bottom ?

PierreRambaud: Same effect as States:
https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/js/admin.js#L1372

Progi1984: :ok_hand: for me

Progi1984 (dismissed):

controllers/admin/AdminSuppliersController.php
```diff
@@ -231,6 +231,14 @@ public function renderForm()
                         'name' => 'name',
                     ),
                 ),
+                array(
+                    'type' => 'text',
+                    'label' => $this->trans('DNI', array(), 'Admin.Global'),
```
LouiseBonnard, line 236: Hey, are the NIF and NIE mandatory as well? I can see that there was a string _DNI / NIF / NIE_ in 1.6.

PierreRambaud: No, it's just the DNI

classes/Address.php
```diff
@@ -347,12 +347,7 @@ public function validateField($field, $value, $id_lang = null, $skip = array(),
         }
 
         // Special validation for dni, check if the country needs it
-        $result = (bool) Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue('
-			SELECT c.`need_identification_number`
-			FROM `' . _DB_PREFIX_ . 'country` c
-			WHERE c.`id_country` = ' . (int) $this->id_country);
-
-        if ($result && Tools::isEmpty($value)) {
+        if (static::containsDni((int) $this->id_country) && Tools::isEmpty($value)) {
```
jolelievre: I think `needDni` would reflect more the behavior, it can contain DNI but not need it no?

controllers/admin/AdminAddressesController.php
jolelievre: that's weird that the two json returns are not encoded and displayed togeter..
what happens if both email and contains_dni is submitted, the returned json will be invalid no?

controllers/admin/AdminSuppliersController.php
```diff
@@ -231,6 +231,14 @@ public function renderForm()
...
+                    'name' => 'dni',
+                    'maxlength' => 16,
+                    'col' => 4,
+                    'required' => true, // Only required in case of specifics countries
```
jolelievre, line 240: If it's only required for specific countries, the shouldn't the default behaviour `required = false` and manage a specific check? or else you can never leave it empty

admin-dev/themes/default/template/helpers/form/form.tpl
```diff
@@ -67,7 +67,7 @@
 					<div class="form-wrapper">
 					{foreach $field as $input}
 						{block name="input_row"}
-						<div class="form-group{if isset($input.form_group_class)} {$input.form_group_class}{/if}{if $input.type == 'hidden'} hide{/if}"{if $input.name == 'id_state'} id="contains_states"{if !$contains_states} style="display:none;"{/if}{/if}{if isset($tabs) && isset($input.tab)} data-tab-id="{$input.tab}"{/if}>
+						<div class="form-group{if isset($input.form_group_class)} {$input.form_group_class}{/if}{if $input.type == 'hidden'} hide{/if}"{if $input.name == 'id_state'} id="contains_states"{if !$contains_states} style="display:none;"{/if}{/if}{if $input.name == 'dni'} id="contains_dni"{if !$contains_dni} style="display:none;"{/if}{/if}{if isset($tabs) && isset($input.tab)} data-tab-id="{$input.tab}"{/if}>
```
jolelievre: Not sure I get this part So the field is in the form but it is hidden? It's not a problem if you can't edit it?
Then why is it in the address format in the first place?

jolelievre: Ok I got it, you toggle it with javascript

controllers/admin/AdminAddressesController.php
PierreRambaud: It should never happened. It's the legacy logic.. Don't ask me why it was like that :/

controllers/admin/AdminSuppliersController.php
PierreRambaud: Adding required here only add the required statement for HTML.

admin-dev/themes/default/template/helpers/form/form.tpl
PierreRambaud: Yep

controllers/admin/AdminAddressesController.php
PierreRambaud: FYI, it's separated behaviour, it's because of ajaxProcess, which process all ajax requests.
I change only this part to limit changes because it's for a patch version and the supplier page will be migrate in 1.7.7.

sarahdib:
@PierreRambaud 
The DNI field didn't appear  : https://drive.google.com/file/d/1P-TUVHC9tJ5iKJyllzrQHffCXMTCSGe5/view

admin-dev/themes/default/template/helpers/form/form.tpl
```diff
@@ -948,6 +948,11 @@
 				}
 			{/if}
 
+			dniRequired();
+			$('#id_country').change(function() {
```
Progi1984: @PierreRambaud Est ce que cela ne fonctionnerait pas ?
```suggestion
			$('#id_country').change(dniRequired)
```

Progi1984 (approved):

matks:
Restarting Travis as the 2 failing builds are not code mistake but build mistakes

matks:
Thank you @PierreRambaud

PR #6509, opened 2016-09-27
Separate gifts from product manually added

thierrymarianne:
| Questions | Answers |
| --- | --- |
| Branch? | develop |
| Description? | Quantity of manually added products should not be merged with gifts quantity |
| Type? | bug fix |
| Category? | FO |
| BC breaks? | no |
| Deprecations? | no |
| Fixed ticket? | http://forge.prestashop.com/browse/BOOM-1542 |
| How to test? | Add vouchers with gift to a cart and check there are multiple lines when same product is manually added. Check discount line has been removed from right panel. |

PR #15691, opened 2019-09-25
Cart rules mustn't be auto added automatically in some cases

### LINKED ISSUE #15574: Infinite loop and order blocked with two specific cart rules
**Describe the bug**
Infinite loop and order blocked with two specific cart rules.

**To Reproduce**
With a fresh Prestashop 1.7.6.1 installation in debug mode. Follow this exactly in this order :
1. Go to BO/cart rules
2. Add a first cart rule without voucher code with 10€ HT gift
3. Add a product condition on "Home" Category and save
4. Add another cart rule without voucher code with free delivery
5. Go to front and add a product to the cart : You'll have 500 error on the ajax call
6. Reload page : you'll get the infinite loop with this error "Maximum function nesting level of '256' reached, aborting!"
7. Go to BO / Carts : I'll get the infinite loop too

If xdebug is enabled, You'll get the error "Maximum function nesting level of '256' reached, aborting!" If not, php will crash and you'll only have a 500 error.

This bug is not present in 1.6, It's a regression.

It seems that cart rules are recursively evaluated and product list is reconstructed each call of cart computation.

As already disscussed with @matks, this PR should fix the bug https://github.com/PrestaShop/PrestaShop/pull/12878

**Screenshots**
cartrule1
cartrule2

**Additional information**
PrestaShop version: 1.7.6.1
PHP version: 7.0

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Cart rules mustn't be auto added automatically during getTotalCalculationCartRules. The calculation here make the function recursive and going to an infinite loop :man_facepalming: 
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15574 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15691)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -2124,9 +2124,9 @@ class CartCore extends ObjectModel
         if ($withShipping || $type == Cart::ONLY_DISCOUNTS) {
             $cartRules = $this->getCartRules(CartRule::FILTER_ACTION_ALL);
         } else {
-            $cartRules = $this->getCartRules(CartRule::FILTER_ACTION_REDUCTION);
+            $cartRules = $this->getCartRules(CartRule::FILTER_ACTION_REDUCTION, false);
             // Cart Rules array are merged manually in order to avoid doubles
-            foreach ($this->getCartRules(CartRule::FILTER_ACTION_GIFT) as $cartRuleCandidate) {
+            foreach ($this->getCartRules(CartRule::FILTER_ACTION_GIFT, false) as $cartRuleCandidate) {
                 $alreadyAddedCartRule = false;
                 foreach ($cartRules as $cartRule) {
                     if ($cartRuleCandidate['id_cart_rule'] == $cartRule['id_cart_rule']) {
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1951,7 +1951,7 @@ class CartCore extends ObjectModel
 
         // CART CALCULATION
         $cartRules = array();
-        if (in_array($type, [Cart::BOTH, Cart::ONLY_DISCOUNTS])) {
+        if (in_array($type, [Cart::BOTH, Cart::BOTH_WITHOUT_SHIPPING, Cart::ONLY_DISCOUNTS])) {
             $cartRules = $this->getTotalCalculationCartRules($type, $type == Cart::BOTH);
         }
 
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1951,9 +1951,10 @@ class CartCore extends ObjectModel
 
         // CART CALCULATION
         $cartRules = array();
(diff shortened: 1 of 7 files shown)


matks:
I have **exactly** the same fears than for https://github.com/PrestaShop/PrestaShop/pull/12878

And consequently I suggest the same strategy: https://github.com/PrestaShop/PrestaShop/pull/12878#issuecomment-526716746

> We could create a new function, used ONLY in a single situation. Our situation. So it means: creating a duplicate of Product:: getProducts(), and this duplicate function would be called ONLY for the situation we want to fix. This way, we make sure fixing this usecase cannot modify the behavior for another usecase and thus we highly reduce the possibility of introducing unexpected side effects.

matks:
And Travis is not happy either

PierreRambaud:
> I have **exactly** the same fears than for #12878
> 
> And consequently I suggest the same strategy: [#12878 (comment)](https://github.com/PrestaShop/PrestaShop/pull/12878#issuecomment-526716746)
> 
> > We could create a new function, used ONLY in a single situation. Our situation. So it means: creating a duplicate of Product:: getProducts(), and this duplicate function would be called ONLY for the situation we want to fix. This way, we make sure fixing this usecase cannot modify the behavior for another usecase and thus we highly reduce the possibility of introducing unexpected side effects.

This is exactly what I'm thinking. I really don't know why it works, and don't know what I break :sweat_smile: 
But I'm still not in favor of creating a new function, and because we want to fix how cart rules are calculated :/

matks:
Last commit that impacted this area of code is https://github.com/PrestaShop/PrestaShop/pull/6509

It's 3 years ago. It does not make sense if this is a regression of 1.7.6 😐

PierreRambaud:
> Last commit that impacted this area of code is #6509
> 
> It's 3 years ago. It does not make sense if this is a regression of 1.7.6

Nothing make sens in the legacy code :man_facepalming:

matks (dismissed):

matks:
Travis indicates this breaks some integration tests related to cart rules calculations 😭 

On a side note, I have checked and it seems that this fix, combined with https://github.com/PrestaShop/PrestaShop/pull/12878, results in https://github.com/PrestaShop/PrestaShop/issues/11172 being fixed, and this is validated by the behat scenario https://github.com/PrestaShop/PrestaShop/pull/15452

So if we merge this fix, then it seems https://github.com/PrestaShop/PrestaShop/pull/12878 can maybe be merged.

matks (approved):

matks:
Thank you @PierreRambaud

jf-viguier:
Thank you very very very much @PierreRambaud and @matks

PR #15698, opened 2019-09-25
You can order even when no carrier is available for any of the products in the cart.

PululuK:
| Questions     | Answers
| Branch?       | develop
| Description?  | Ability to place an order despite the unavailability of carriers for certain products in the cart.
| Type?         | bug fix 
| Category?     | FO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ----.
| How to test?  | Create two products, and associate one of the products with a disabled carrier for example (or a carrier not available in the country where you want to be delivered) and the other product you associate with an available carrier. On the front, add both products to the basket. In the order process in the delivery step, you will find that despite the effect that carriers lack for one of the products you can still place your order. On the other hand, in the back office, at the level of the invoice you will miss information concerning the delivery of the products in question.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15698)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -2415,6 +2415,9 @@ class CartCore extends ObjectModel
                 }
             }
 
+            if(empty($product['carrier_list']))
+                return array();
+
             if (!isset($grouped_by_warehouse[$product['id_address_delivery']]['in_stock'][$id_warehouse])) {
                 $grouped_by_warehouse[$product['id_address_delivery']]['in_stock'][$id_warehouse] = array();
                 $grouped_by_warehouse[$product['id_address_delivery']]['out_of_stock'][$id_warehouse] = array();
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -2415,8 +2415,9 @@ class CartCore extends ObjectModel
                 }
             }
 
-            if(empty($product['carrier_list']))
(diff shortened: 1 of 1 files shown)


classes/Cart.php
```diff
@@ -2415,6 +2415,9 @@ public function getPackageList($flush = false)
                 }
             }
 
+            if(empty($product['carrier_list']))
```
PululuK: It is sufficient that one product does not have the carriers available to block the ordering process. This will display the usual message: "No carrier available for this delivery address."

classes/Cart.php
```diff
@@ -2415,6 +2415,10 @@ public function getPackageList($flush = false)
                 }
             }
 
+            if (!$product['is_virtual'] && empty($product['carrier_list'])) {
+                return array();
+            }
```
PululuK, line 2420: It is sufficient that one product does not have the carriers available to block the ordering process. This will display the usual message: ```No carrier available for this delivery address```. Predict the possibility of sending a message for this case like: ```No carrier found for the PRODUCT X, please update your cart``` ?!

PierreRambaud:
Hi and sorry for the delay, can you create an issue to be sure we can reproduce the problem?

matks:
ping @PululuK friendly reminder 😉

PierreRambaud:
Hi @PululuK ,

Thanks!

PR #15710, opened 2019-09-26
Prevent rage clicking on checkout steps submit button

### LINKED ISSUE #15703: rapid fire bug in checkout address tab give error "no carriers available" and create multiple addresses in DB
Hello, That's my first issue open here and I hope to do everything right. Please forgive me if I do something wrong!

There is a problem that is driving me crazy: in prestashop 1.7.5.1 checkout, when I create a new address, if I click more than once on "Continue" button, i've got the error "no carriers for the selected address in the carrier's tab.

Here is how to reproduce the issue:

1- Place a product in the cart.
2- Go to checkout and register, or place the order as guest.
3- insert a new address
4- click on the button "continue" in address tab multiple times, and here you got the error.

Plus, if you go back to your address, you can see that you have created a new address for every click you did in "Continue button" in address tab... that's bad!

The only way to fix this (via frontend) will be to empty and fill the cart again, but I really need a real fix for this, customers keep calling and I'm losing a lot of orders!

Thanks for anyone will help me!

PS version: 1.7.5.1

rdy4ever:
| Questions     | Answers
| Branch?       | develop
| Description?  | This PR fixes fixes #15703. Issue reproduced in [this comment](https://github.com/PrestaShop/PrestaShop/issues/15703#issuecomment-535394768). Coluld've done this in classic, but done it in core so other themes may benefit from the fix
| Type?         | bug fix
| Category?     | FO
| Fixed ticket? | Fixes #15703
| How to test?  | Go to checkout and click like a crazy person on the continue button on each step. With this PR, the button will be disabled after first click.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15710)


themes/_core/js/checkout.js
```diff
@@ -71,6 +72,16 @@ function handleCheckoutStepChange() {
   );
 }
 
+function handleSubmitButton() {
+  // prevents rage clicking on submit button and related issues
+  $(document).on('click', '.js-current-step button[type="submit"]', function () {
+    setTimeout( () => $(this).prop('disabled',true) , 0); //defer so the form still submits
```
PierreRambaud: I don't see the reason to use a setTimeout here.
You can add the `disabled` property without defering.

rdy4ever: @PierreRambaud If I don't, the form won't submit anymore. I already tried it without timeout. In order to work without timeout I think we need to do it separately for each form.

matks: I'm worried about side-effects on themes that are based on Classic Theme 🤔 

Am I too paranoid ?

rdy4ever: > I'm worried about side-effects on themes that are based on Classic Theme 🤔
> 
> Am I too paranoid ?

I don't think this would affect Classic based or any third party themes. We just disable a button via a proprety, it's pretty basic. Do you have a use case in mind when this would interfeer?

matks: If someone already implement this feature 😉 it could try to disable it twice

rdy4ever: > If someone already implement this feature 😉 it could try to disable it twice

Yes, that's true but it will still work. I just tried it: added the same function in `calssic/assets/js/custom.js` and I have no error.

matks: 🎉

PierreRambaud: If you added a setTimeout to fix a weird behaviour, the fix must be done elsewhere :/

rdy4ever: > If you added a setTimeout to fix a weird behaviour, the fix must be done elsewhere :/

You're right, but I don't know if it really is that weird. I think it's normal that you cannot submit a form if you disable its submit button first. Using the setTimeout(0) I'm just sending the execution of `$(this).prop('disabled',true)` to the end of the queue, so it’s executed after submitting the form.

rdy4ever: Hi @PierreRambaud 
I've altered this PR to use bootstrap `disabled` class instead (that uses `pointer-events: none` to disable links and buttons). It has [pretty good browser compatibility](https://caniuse.com/#feat=pointer-events) and we get rid of any setTimeout. WDYT?

rdy4ever: > Hi @PierreRambaud
> I've altered this PR to use bootstrap `disabled` class instead (that uses `pointer-events: none` to disable links and buttons). It has [pretty good browser compatibility](https://caniuse.com/#feat=pointer-events) and we get rid of any setTimeout. WDYT?

On further testing it doesn't work so well in FF, IE or Opera. It just disables the button visually (this could discourage pepole to click it again), but it's still clickable, so this still is not a fix.

PierreRambaud: Much better :+1:

PierreRambaud:
I squash your commits, LGTM.

PierreRambaud (approved):

matks:
Thank you @rdy4ever

khouloudbelguith:
Hi @rdy4ever,

with develop branch, I can quickly click two times on the continue button & the error "Unfortunately, there are no carriers available for your delivery address." => not fixed.
https://drive.google.com/file/d/1WHLw2vrn2c9KTYxmMuH5BQpGERNjWFzE/view
Thanks!

rdy4ever:
Hi @khouloudbelguith 
Yes, because this fix relies on the browser to actually implement `pointer-events: none` in a way that won't allow the user to click the button. In Chrome it worked at the time I tested it, Firefox disabled it just visually but you could still click it. Other browsers may have different implementations.
Initially, I tried to resolve this by javascript too, but it wasn't apporoved because I needed to use `setTimeout` and couldn't make it work without deffering (see discussion).
> **PierreRambaud on 26 Sep** 
> I don't see the reason to use a setTimeout here.
> You can add the `disabled` property without defering.
> ...

In the end, we just use BS `disabled` class, but the implementation is... not that good.

fredo85:
hi, why not use a data attribute (https://caniuse.com/#search=data) like this https://github.com/solarmosaic/jquery-prevent-double-submit/blob/master/jquery-prevent-double-submit.js ? in addition, to javascript. it should also add a check in the php code (a token?)

PR #15748, opened 2019-09-30
Use cover photo in listing instead of default combination

### LINKED ISSUE #12241: Product with combination
Let me explain the concern:
If I have One product A and Which I want to show with image 1.jpg (this one is set as cover image for refence of product)
Product A has 2 combination a and b with image 2.jpg and 3.jpg respectively.
Now when I see the product on the front-office. It should be shown with 1.jpg image (Cover Image) but in prestashop first combination (combination a with image 2.jpg) get selected automatically and shows in product list instead of cover image of product.

I think there should be display Product A with cover image which is only for reference image of product (This should not be allow for order) and once customer select any of attribute then related combination image should be shown and only combination can be allow for order.

## SPECS
Diplay the cover image in the product listing (category)
Display the images selected for the combination in the product page

PierreRambaud:
| Questions     | Answers
| Branch?       | develop
| Description?  | Use cover photo in listing instead of default combination
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #12241 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15748)

### DIFF
--- a/themes/classic/templates/catalog/_partials/miniatures/product.tpl
+++ b/themes/classic/templates/catalog/_partials/miniatures/product.tpl
@@ -29,12 +29,14 @@
     <div class="thumbnail-container">
       {block name='product_thumbnail'}
         {if $product.cover}
-          <a href="{$product.url}" class="thumbnail product-thumbnail">
+          {assign var='coverImage' value=Product::getCover($product->id)}
+          {assign var='coverImageId' value="{$product->id}-{$coverImage.id_image}"}
+          <a href="{$product.canonical_url}" class="thumbnail product-thumbnail">
             <img
-              src="{$product.cover.bySize.home_default.url}"
+              src="{$link->getImageLink($product.link_rewrite, $coverImageId)}"
               alt="{if !empty($product.cover.legend)}{$product.cover.legend}{else}{$product.name|truncate:30:'...'}{/if}"
               data-full-size-image-url="{$product.cover.large.url}"
-            />
+              />
           </a>
         {else}
           <a href="{$product.url}" class="thumbnail product-thumbnail">


matks (approved):

matks:
We should credit @dineshbadrukhiya 🤔for his code

Progi1984 (approved):

dineshbadrukhiya:
> We should credit @dineshbadrukhiya 🤔for his code

Thanks :)

sarahdib:
It's Ok :) Thank you @PierreRambaud 

Capture d’écran 2019-09-30 à 15 00 17

Frankyet:
Sorry but this is not working for miniatures in the search bar and other modules like seen products, this is not a theme issue but a core Prestashop, it should be working like 1.6 did.

marionf:
@Frankyet 

It's ok with classic for viewed product, search and new products

capture d'écran_2028

capture d'écran_2029

capture d'écran_2030

Frankyet:
I am talking about cover image for combination see 
[https://github.com/PrestaShop/PrestaShop/issues/12241](url)

dineshbadrukhiya:
> I am talking about cover image for combination see
> [https://github.com/PrestaShop/PrestaShop/issues/12241](url)

Sorry, But can you please share screenshot for exactly what you are talking about?

kpodemski:
I have two issues with this change:
- `$link` shouldn't be used directly in a template, this was pointed as a bad habbit from the beginning of a classic, this is why we have `url` helper and no links from the context in entire classic
- you have additional query in every single miniature called from... template, shouldn't we have this logic in ProductPresenter? or even deeper, in getProductProperties

WooXu:
The fix you did link here does not actually work that well, since it doenst load any resized picture, just the "original" picture that was uploaded. So basicly if you are using 300x300 pictures in product list it will load f.eks. 800x800 pictures for all products coz you force it to load whatever picture you did upload instead of the rezised pictures.  Can you guys come up with a proper solution for this issue?

I wonder if the issue actually happens coz theres no picture size in default named Cover.

dineshbadrukhiya:
> The fix you did link here does not actually work that well, since it doenst load any resized picture, just the "original" picture that was uploaded. So basicly if you are using 300x300 pictures in product list it will load f.eks. 800x800 pictures for all products coz you force it to load whatever picture you did upload instead of the rezised pictures. Can you guys come up with a proper solution for this issue?
> 
> I wonder if the issue actually happens coz theres no picture size in default named Cover.

You can try with below:
`src = "{$link->getImageLink($product.link_rewrite, $coverImageId, home_default)}"`

sam-pires:
Hello everyone,

We slightly modified the behavior from this issue following the feedback we received during the 1.7.7 beta.
We keep the cover image as the default image in product listing, but still allow developers to override to display combinations images, important on some themes and for some modules (faceted search for example).
Issue here: #19029 
PR: #19437 
[Specs](https://github.com/PrestaShop/prestashop-specs/blob/master/front-office/Product-images.md)

@jolelievre added some doc about it in the devdoc:
[Part on product cover in the devdoc FAQ](https://devdocs.prestashop.com/1.7/faq/product/#product-cover)
[Part on product cover in the 1.7.7 updates in the devdoc](https://devdocs.prestashop.com/1.7/modules/core-updates/1.7.7/#product-cover)

web-plus:
Hello, this fix loads "cover" image in original size, with default theme / installation it's 1000x1000 px for each image, so it's performance killer and in fact it can't be used.

matks:
@web-plus Hi! thanks for reaching out, could you open a github issue for that? In a merged PR people are not likely to notice such an important warning

PR #15755, opened 2019-09-30
Make CLDR Specification\Number immutable

### LINKED ISSUE #13168: CLDR Specification\Number should be immutable
Current CLDR implementation uses Specification\Number::setMaxFractionDigits() to override cldr specification precision with customized currency one

Specification\Number::setMaxFractionDigits() should be deleted (already marked as deprecated) in favor of class constructor to get maxFractionDigit override

**Additional context**

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | Current CLDR implementation uses Specification\Number::setMaxFractionDigits() to override cldr specification precision with customized currency one
| Type?         | improvement
| Category?     | CO
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | Fixes #13168
| How to test?  | 

**BC Break :** Class PrestaShop\PrestaShop\Core\Localization\Specification\Number is now final

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15755)

### DIFF
--- a/src/Core/Localization/Specification/Factory.php
+++ b/src/Core/Localization/Specification/Factory.php
@@ -100,7 +100,8 @@ class Factory
         CldrLocaleInterface $cldrLocale,
         Currency $currency,
         $numberGroupingUsed,
-        $currencyDisplayType
+        $currencyDisplayType,
+        ?int $maxFractionDigits = null
     ) {
         $currencyPattern = $cldrLocale->getCurrencyPattern();
         $numbersSymbols = $cldrLocale->getAllNumberSymbols();
@@ -111,7 +112,7 @@ class Factory
             $positivePattern,
             $this->getNegativePattern($currencyPattern),
             $this->computeNumberSymbolLists($numbersSymbols),
-            $this->getMaxFractionDigits($positivePattern),
+            $maxFractionDigits ?? $this->getMaxFractionDigits($positivePattern),
             $this->getMinFractionDigits($positivePattern),
             $numberGroupingUsed && $this->getPrimaryGroupSize($positivePattern) > 1,
             $this->getPrimaryGroupSize($positivePattern),
--- a/src/Core/Localization/Specification/Number.php
+++ b/src/Core/Localization/Specification/Number.php
@@ -325,13 +325,15 @@ class Number implements NumberInterface
     }
 
     /**
-     * @deprecated https://github.com/PrestaShop/PrestaShop/issues/13168
+     * @deprecated
      *
      * @param int $maxFractionDigits
+     *
+     * @throws LocalizationException
      */
     public function setMaxFractionDigits($maxFractionDigits)
     {
-        $this->maxFractionDigits = $maxFractionDigits;
+        throw new LocalizationException('Class Number is immutable. Please define maxFractionDigits from the constructor.');
     }
 
     /**
--- a/src/Core/Localization/Specification/Number.php
+++ b/src/Core/Localization/Specification/Number.php
@@ -325,18 +325,6 @@ class Number implements NumberInterface
     }
 
     /**
-     * @deprecated
-     *
-     * @param int $maxFractionDigits
-     *
-     * @throws LocalizationException
-     */
-    public function setMaxFractionDigits($maxFractionDigits)
-    {
-        throw new LocalizationException('Class Number is immutable. Please define maxFractionDigits from the constructor.');
-    }
-
-    /**
      * To array function
      *
      * @return array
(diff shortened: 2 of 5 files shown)


matks:
@Progi1984 the PR title and the PR description seem to describe 2 different things 🤔or I miss something ?

PierreRambaud (changes requested):
Method `Specification\Number::setMaxFractionDigits` must be removed or must do nothing in order to make the `Number` class immutable

Progi1984:
@matks @PierreRambaud After discussion, I don't remove the method but I raise an exception when someone try to use this method, and I mark it as deprecated. So Number class is now immutable.

PS : Need review :)

PierreRambaud:
@Progi1984 What is the reason about the exception? Since we are in PHP7.1+, we can use the `Error` base class :thinking:

Progi1984:
@PierreRambaud The exception is for avoiding the breaking change with a remove of the method. (@eternoendless)

src/Core/Localization/Specification/Factory.php
```diff
@@ -100,7 +100,8 @@ public function buildPriceSpecification(
         CldrLocaleInterface $cldrLocale,
         Currency $currency,
         $numberGroupingUsed,
-        $currencyDisplayType
+        $currencyDisplayType,
```
jolelievre, line 103: Rather than using an optional argument, maybe you can force it?
Although this represent a BC break this factory shouldn't be extended anyway and thus it should have be `final` as well

src/Core/Localization/Specification/Number.php
```diff
@@ -325,13 +325,15 @@ protected function validateData()
...
+     * @deprecated
      *
      * @param int $maxFractionDigits
+     *
+     * @throws LocalizationException
      */
     public function setMaxFractionDigits($maxFractionDigits)
     {
-        $this->maxFractionDigits = $maxFractionDigits;
+        throw new LocalizationException('Class Number is immutable. Please define maxFractionDigits from the constructor.');
```
jolelievre: Since this setter now throws an exception this could be considered as a BC break anyway, so my first comment seems realistic
Maybe you could even drop this setter?

jolelievre:
I kind of agree with @PierreRambaud as I explained in my review

This class should be reserved for internal use only and be labeled as `final` to avoid it from being extending, besides when you see how it is used it's not even a service but an object instanciated when needed so I don't think this will cause a BC
Making it final will allow us to clean to unnecessary code (useless setter and optional argument can now be explicit)

src/Core/Localization/Specification/Factory.php
Progi1984: No, I can't force it. By default, it used the MaxFractionDigits from the currency of the CldrLocale object. But I need specifically to overwrite this default value.

Progi1984:
@jolelievre After thinking, you and @PierreRambaud have right. As this class is for internal use, I mark it as `final` and remove the method.

Progi1984:
@jolelievre Finally, no. Even if this class is for internal use, the class Price inherits from Number so it can't be `final`. The solution is so an exception.

jolelievre:
@Progi1984 damn you're right.. that said the Price class is also internal I wonder if we could add an `@internal` flag and still remove this method for both classes?

PierreRambaud:
In fact, adding an exception is still useless, since we are in php 7.1+ we also can catch not implemented function :thinking:

Progi1984:
@jolelievre @PierreRambaud So two solutions : 
* I keep the method with an exception and a clear message (`'Class Number is immutable. Please define maxFractionDigits from the constructor.'`)
* I remove the method and a [BadMethodCallException](https://www.php.net/manual/en/class.badmethodcallexception.php) will be thrown.

I have a preference for first solution. And you ?

Progi1984:
@jolelievre @PierreRambaud After discussion, I take the second solution since the fatal error (Method not found) is catchable in PHP 7.1+

src/Core/Localization/Specification/Number.php
Progi1984: Method removed.

PierreRambaud (approved):

PierreRambaud:
One more review needed

src/Core/Localization/Specification/Factory.php
```diff
@@ -100,7 +100,8 @@ public function buildPriceSpecification(
...
+        ?int $maxFractionDigits = null
```
jolelievre, line 104: You should remove this optional parameter and make it mandatory

jolelievre (changes requested):

src/Core/Localization/Specification/Factory.php
jolelievre: On second thought, it will be useful if we want to rely on pattern rather than currency values!

jolelievre (approved):

PR #15757, opened 2019-09-30
Fix monitoring grids id configuration

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | In `grid_data_factory.yaml` monitoring page grid definitions where configured with wrong id's. Also empty_category grid had one wrong argument injected, which was leading to error.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | partially https://github.com/PrestaShop/PrestaShop/issues/10552
| How to test?  | open `/admin-dev/index.php/sell/catalog/monitoring`, ensure that the page contains 7 lists. Each list should be filterable, sortable, paginatable independently from each other. The only action currently working is categories status toggling (other actions are not migrated).

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15757)

### DIFF
--- a/src/Core/Grid/Data/Factory/EmptyCategoryGridDataFactory.php
+++ b/src/Core/Grid/Data/Factory/EmptyCategoryGridDataFactory.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Grid\Data\Factory;
+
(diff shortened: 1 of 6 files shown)


matks (dismissed):

PierreRambaud (dismissed):

Ashek46 (dismissed):

Robin-Fischer-PS:
Hi @zuk3975 , thanks for the PR :)

I've noticed some problems : 

1. On all lists : When doing a partial search on the "ID" column (search "1" to find all ID containing 1), it doesn't work.

Here on Legacy, I have a result : 
screenshot Monitoring with combi 1 - Legacy

On Symfony, no result : 
screenshot Monitoring with combi 1 - Migration

2. On "List of empty categories", there is a missing info on top of the list : 
"An empty category is a category that has no product directly associated to it. An empty category may however contain products through its subcategories."
(see screenshot after 4.)

3. On "List of empty categories", there is a sort missing, on "Description".
(see screenshot after 4.)

4. On "List of empty categories", in the description field, HTML tag appear in Symfony

Legacy : 
screenshot Monitoring Category 1 - Legacy

Symfony : 
screenshot Monitoring Category 1 -  Migration

Thanks :)

zuk3975:
@Robin-Fischer-PS nice catch :smile: . Filtering by id is not partial anymore(it searches for exact match) in all migrated lists, so it stays.
Ill check another bugs, thanks :)

zuk3975:
@Robin-Fischer-PS, fixed.
One thing to notice though. The info block is placed as in legacy (inside grid card), but i've noticed for example in brands page, the info block is on top of the list (not in its card). Not sure which is the good way

src/Core/Grid/Data/Factory/EmptyCategoryGridDataFactory.php
```diff
@@ -0,0 +1,82 @@
...
+     * Modify empty category records.
+     *
+     * @param array $records
+     *
+     * @return array
+     */
+    private function modifyRecords(array $records)
+    {
+        foreach ($records as $key => $record) {
+            $records[$key]['description'] = strip_tags(stripslashes($record['description']));
```
PierreRambaud, line 77: What is the reason of doing a chained strip_tags and stripslashes?

zuk3975: Not really sure of the main reason, I just kept it the same as in categories page :thinking: . Guess its for escaping backslashes

matks (approved):

Robin-Fischer-PS:
Hi @zuk3975 ! Thanks for the fix, it's good for me :)

For the info block, since on this page it's only useful for the "Empty Categories", the actual design looks good and logical to me.

PierreRambaud:
Thanks @zuk3975

PR #15773, opened 2019-10-01
CO : Performance optimisation

djbuch:
| Questions     | Answers
| Branch?       | develop
| Description?  | Optimisation of performance : We will only use the first row of this query, so limit to 1 in order to avoid querying multiple rows, because the join with stock_available will add as many rows than there are product attributes.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | This is really just good sense

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15773)

### DIFF
--- a/classes/ProductAssembler.php
+++ b/classes/ProductAssembler.php
@@ -85,7 +85,7 @@ class ProductAssemblerCore
                 LEFT JOIN {$prefix}stock_available sa
 			        ON sa.id_product = p.id_product 
 			        AND sa.id_shop = $idShop
-			    WHERE p.id_product = $idProduct";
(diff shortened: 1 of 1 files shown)


Progi1984:
@djbuch Thank you for your contribution. Could you create an issue for presenting your problem ? Thanks

matks (approved):

Progi1984 (approved):

PierreRambaud (approved):

PierreRambaud (approved):

PR #15761, opened 2019-09-30
Fix order view page wordings

matks:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Fix some wording following Content Manager feeback
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | No QA needed

Fixing some of the feedback of Content Manager @LouiseBonnard in https://github.com/PrestaShop/PrestaShop/pull/15388

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15761)

Issue #15732, opened 2019-09-27
Error handler feature


sarjon:
Hello community! :wave: 

I'd like to start a discussion about more robust error handling in PrestaShop's migrated pages.

## Problem

Currently we dispatch commands/queries from controllers and handle _all_ possible exceptions by wrapping it with `try/catch`.

However, there are a few problems with it that I can see:
1. As a developer I want to see all errors during development (e.g. when debug mode is ON) it is very useful to see full exception with trace stack of what went wrong, currently it shows configured error message or `An unexpected error occurred.`, in case it's an unknown error, in form of flash message.
2. It is common to get non `Domain` or `Core` exception during runtime, lets say Symfony exception, which often means it's a bug in code and should not be silenced with `An unexpected error occurred.` message in form of flash message, but instead show delegated error page (which PS has, see https://github.com/PrestaShop/PrestaShop/pull/10780) so user can understand that this is something worth reporting.
3. There is code duplication in every controller method that uses CQRS (as mentions before, commands/queries are wrapped with `try/catch` all the time), that increased size of controller both in lines and complexity.

## Proposal

Since developers have to configure/map exception to human readable error message, it often ends up as a huge static array in every controller, a few examples:
https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php#L748
https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php#L742

My initial idea is to have some kind of error handler, maybe by using Symfony `kernel.exception` event and filter only configured exceptions and let others bubble up to UI.

It should also be possible to extract error messages mapping into separate class (but it should stay at the controller/ui/presentation/bundle/whatever_you_call_it level) and, in addition to that, we should not forget about modules to allow them to configure custom exception/messages.

It's a very rough idea that can be refined or discarded.

@PrestaShop/prestashop-core-developers

matks:
> Hello community! 👋
> 
> I'd like to start a discussion about more robust error handling in PrestaShop's migrated pages.
> 
> ## Problem
> Currently we dispatch commands/queries from controllers and handle _all_ possible exceptions by wrapping it with `try/catch`.
> 
> However, there are a few problems with it that I can see:
> 
[...]

PR #15388, opened 2019-09-03
Migration of Orders View page

sarjon:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | This PR migrates legacy Orders View page to Symfony.
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | n/a
| How to test?  | Not ready.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15388)

PR #15778, opened 2019-10-01
Migrate Payment block of Order view page

### LINKED ISSUE #15827: [ORDER PAGE] Payments Tab - 10/11

**User Story**

As a merchant, when I check an order, I need to be able to see all the payment interaction.

**Detailed specification**
[Payment Tab specs
](https://github.com/PrestaShop/prestashop-specs/blob/cc476f1fc0c71c63bd3ba4879131f5163fca05e7/Order%20Page%20View.md#v-tabs-panel)

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | Payment block of Order view page should be fully functional.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/15827
| How to test?  | :warning: Build assets before testing :warning: Access new page under `admin-dev/index.php/sell/orders/orders/{orderId}/view`. "Payment" block (https://prnt.sc/pdf5kl) should be fully functional as it was in legacy.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15778)


admin-dev/themes/new-theme/js/pages/order/view.js
```diff
@@ -0,0 +1,45 @@
...
+
+const $ = window.$;
+
+$(() => {
+  handlePaymentDetailsToggle();
+
+  function handlePaymentDetailsToggle() {
+    $(OrderViewPageMap.orderPaymentDetailsBtn).on('click', (event) => {
+      const $paymentDetailRow = $(event.currentTarget).closest('tr').next(':first');
+      console.log($paymentDetailRow);
```
Progi1984: Console.log detected.

sarjon: Good catch!

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -344,6 +350,7 @@ public function addPaymentAction(int $orderId, Request $request): RedirectRespon
 
                 $this->addFlash('success', $this->trans('Successful update.', 'Admin.Notifications.Success'));
             } catch (Exception $e) {
+                throw $e;
```
matks: Debug only

src/PrestaShopBundle/Form/Admin/Sell/Order/ChangeOrderCurrencyType.php
```diff
@@ -0,0 +1,96 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
```
matks: Update your license header generator (in your iDE) 😉 
```suggestion
 * 2007-2019 PrestaShop SA and Contributors
```

matks (changes requested):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
sarjon: Wooops, that's why I raised #15732 issue.

LouiseBonnard:
Hello there! Change of wordings are not applied here, do I need to comment where it needs to be modified?

sarjon:
Sure @LouiseBonnard, but make sure it's not duplicate of https://github.com/PrestaShop/PrestaShop/pull/15761 Maybe I could rebase PR first. :)

sarjon:
@LouiseBonnard feel free check wording now. :)

@matks review comments addressed.

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
matks: :trollface: are you a poker player ?

matks (dismissed):

LouiseBonnard:
Hi, thanks @sarjon, two comments raised yesterday with @MatShir :

- please turn 'Card Brand' into 'Card type' at line 81 cf. [here](https://github.com/PrestaShop/PrestaShop/pull/15388#discussion_r330108977)

- also, are the card details necessarily displayed? I think about the check payment, for instance

sarjon:
> also, are the card details necessarily displayed? I think about the check payment, for instance

What do you mean? User has to click "Detail" button to see payment details.

LouiseBonnard:
@sarjon, I mean the payment is not always made by card, it can be made by bank wire or with cash at delivery. Yet, the details here only concern card payments with the number, the name, the type of card, etc. So to comply with all payment methods, we should:

- adapt the wording to cover every payment method
- or do not display this details line for non-card payments

sarjon:
@LouiseBonnard payment details are displayed for every payment regardless of payment method (cash, bank wire, paypal or etc.). See https://prnt.sc/peomqz

sarjon:
@sarahdib I've found one missing feature for this block, so I'm removing "waiting for QA" for now. :)

sarjon:
@matks could you review https://github.com/PrestaShop/PrestaShop/pull/15778/commits/c4e0ba67987a0528381b2404fbf917ba9f8c5755 commit?

matks (dismissed):

LouiseBonnard:
@sarjon, thank you, in this case, I'd prefer not to display the details if the payment is not made by card.

cc @MatShir, @TristanLDD

sarjon:
> @sarjon, thank you, in this case, I'd prefer not to display the details if the payment is not made by card.

I'm not sure if we can know that, unless we can check every field if it has value, but it might have only one value of all available.

TristanLDD:
I agree with Louise, it would be better not to display the details button when details informations are missing. @sarjon @MatShir Could you discuss it?  If we can't integrate this in the 1.7.7 we should keep it it mind for further improvement.

sarjon:
>  If we can't integrate this in the 1.7.7

We can, but lets not do it in this PR, we could improve it separate PR if needed, but I want to merge this one asap. :)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/payments.html.twig
```diff
@@ -78,7 +78,7 @@
               {% endif %}
             </p>
             <p class="mb-0">
-              <strong>{{ 'Card Brand'|trans({}, 'Admin.Orderscustomers.Feature') }}</strong>&nbsp;
+              <strong>{{ 'Card Type'|trans({}, 'Admin.Orderscustomers.Feature') }}</strong>&nbsp;
```
LouiseBonnard: _Card type_ please, no upper letter on the second word. ;-)

admin-dev/themes/new-theme/js/pages/order/view.js
```diff
@@ -0,0 +1,44 @@
...
+const $ = window.$;
+
+$(() => {
+  handlePaymentDetailsToggle();
+
+  function handlePaymentDetailsToggle() {
+    $(OrderViewPageMap.orderPaymentDetailsBtn).on('click', (event) => {
+      const $paymentDetailRow = $(event.currentTarget).closest('tr').next(':first');
+
+      if ($paymentDetailRow.hasClass('d-none')) {
```
PierreRambaud: You can use `toggleClass` instead

src/Adapter/Form/ChoiceProvider/InstalledPaymentModulesChoiceProvider.php
```diff
@@ -0,0 +1,52 @@
...
+use Validate;
+
+final class InstalledPaymentModulesChoiceProvider implements FormChoiceProviderInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getChoices(): array
+    {
+        $paymentModules = [];
```
PierreRambaud: Store $paymentModules in property, to be sure the foreach will be not executed twice if we have two `InstalledPaymentModulesChoiceProvider` in the same page.

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/payments.html.twig
```diff
@@ -62,12 +62,12 @@
...
+            <button class="btn btn-sm btn-outline-secondary js-payment-details-btn">
               {{ 'Details'|trans({}, 'Admin.Global') }}
             </button>
           </td>
         </tr>
-        <tr>
+        <tr class="d-none">
           <td colspan="6">
             <p class="mb-0">
               <strong>{{ 'Card number'|trans({}, 'Admin.Global') }}</strong>
```
LouiseBonnard, line 73: Is this information really displayed in the back office? Is it the whole card number ou just a few numbers with the rest hidden?

src/Core/Domain/Order/Payment/Command/AddPaymentCommand.php
```diff
@@ -89,10 +88,9 @@ public function __construct(
         $paymentAmount,
         $paymentCurrencyId,
         $orderInvoiceId = null,
-        $transactionId = ''
+        $transactionId = null
```
PierreRambaud, line 91: Should be an int like $orderInvoice :thinking: 
```suggestion
        ?int $transactionId = null
```

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/orders.yml
```diff
@@ -79,3 +79,11 @@ admin_orders_add_payment:
     _controller: PrestaShopBundle:Admin/Sell/Order/Order:addPayment
   requirements:
     orderId: \d+
+
+admin_orders_change_currency:
+  path: /{orderId}/change-currency
```
PierreRambaud:
```suggestion
  path: /{orderId}/currency/change
```

admin-dev/themes/new-theme/js/pages/order/view.js
sarjon: I don't like `toggleClass` because it's not predictable, it could reduce lines of code, but that wouldn't make code easier to understand imo.

src/Adapter/Form/ChoiceProvider/InstalledPaymentModulesChoiceProvider.php
sarjon: Hmm, wouldn't that make service stateful? 

Or image we have this kind of flow:
1. Call InstalledPaymentModulesChoiceProvider
2. Install payment method
3. Call InstalledPaymentModulesChoiceProvider

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/payments.html.twig
sarjon: I guess it could be both, it depends on what merchant/module sets for this value.

src/Core/Domain/Order/Payment/Command/AddPaymentCommand.php
sarjon: This transaction id is not related to PS object model, it can be any transation id from bank (e.g. UUID).

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/orders.yml
sarjon: Why so? :thinking:

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/payments.html.twig
LouiseBonnard: It is a bit annoying not to be able to know because it is a very bad practice regarding security, even if it is a setting enabled by the merchant himself or the payment module he is using... we should not allow the display of this information, or at least make sure it hides a part of it. I got in touch with the modules teams to know more about it, I keep you posted!

LouiseBonnard: It seems that our major payment modules do not even use this bloc, this information is not retrieved. And it is better because we absolutely need to prevent the back office from retrieving it. I will try various modules to be sure that this bloc is deprecated and ask for its deletion if it is the case.

sarjon: I don't think we can delete it, we had client that uses this Payments block.

LouiseBonnard: Okay, it's interesting! Can you send me an email to let me know what payment module(s) are they using? 

PS: I'm not talking about the bloc but the details information

admin-dev/themes/new-theme/js/pages/order/view.js
PierreRambaud: The `toggleClass` do the exact same condition as you did:
https://github.com/jquery/jquery/blob/cf84696fd1d7fe314a11492606529b5a658ee9e3/src/attributes/classes.js#L138

It's totally predictable :thinking:

src/Adapter/Form/ChoiceProvider/InstalledPaymentModulesChoiceProvider.php
PierreRambaud: Another possible scenario.
You have a `CollectionType` with a `InstalledPaymentModulesChoiceProvider`, you have 10 rows. This means, you call the same code 10 times, in the same form :thinking: 

If you have a form and you install a payment method in GET method, it's a bad idea.
If you have a form and generate the form before saving payment method, and after regenerate the code, it's a bad idea too.

Normal scenario:
1. GET: Generate form with `InstalledPaymentModulesChoiceProvider`
2. POST. install payment method
3. REDIRECT
4. GET: Generate form with `InstalledPaymentModulesChoiceProvider`

src/Core/Domain/Order/Payment/Command/AddPaymentCommand.php
PierreRambaud: Could you improve the comment in the php doc block please? :)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/payments.html.twig
sarjon: It was a customized solution, but this feature was needed.

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/orders.yml
PierreRambaud: Because of URI scopes.
`/admin` => You're in admin page
`/admin/order/` => You're in order context
`/admin/order/2` => You select a specific order
`/admin/order/2/currency` => You select the currency for this order
`/admin/order/2/currency/change` => You change the currency for this order.

This is exactly what you do for Advanced parameters
You didn't do `/import-data`, `/import-file-upload` `/import-file-delete`, you split it :)

It's always better to split URIs when you can.
More because you can have ten or twenty things after `/admin/order/2/*` in the future.

sarjon: Perfect. :ok_hand:

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/payments.html.twig
LouiseBonnard: So, to keep you posted, it has been decided to delete this details line. @colinegin has suggested adding a hook at this place, I'm okay if it can be useful, I just need to check the impact before.

admin-dev/themes/new-theme/js/pages/order/view.js
sarjon: I think it's similar as with CQRS commands, `ToggleSomethingCommand` is not predictable, you don't know if it will end up one way or another. So I think having `if` is more explicit and easier to debug, but I'll follow your suggestion and use `toggleClass` as I don't think it's a big issue. :wink:

src/Adapter/Form/ChoiceProvider/InstalledPaymentModulesChoiceProvider.php
sarjon: There can be many scenarios, but I'll store it in static attribute. :)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/payments.html.twig
sarjon: Can we first merge this one and perform behavior modifications in dedicated PR? It would be way better imo.

LouiseBonnard: If you prefer to do so, it suits me fine. :-)

matks (dismissed):

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/payments.html.twig
matks: Issue created here https://github.com/PrestaShop/PrestaShop/issues/15893 for this topic

matks:
> @sarjon, thank you, in this case, I'd prefer not to display the details if the payment is not made by card.
> 
> cc @MatShir, @TristanLDD

Issue created here https://github.com/PrestaShop/PrestaShop/issues/15893 for this topic

sarahdib:
Hello @sarjon 

There is a problem with the dropdown in the bill column. If there is no bill the dropdown shouldn't appear
Capture d’écran 2019-10-11 à 09 15 30

In the date field it should be a calendar not a dropdown and arrow 
Capture d’écran 2019-10-11 à 09 18 42

sarjon:
@sarahdib done! :)

matks (approved):

matks:
Thank you @sarjon

PR #15840, opened 2019-10-04
Do not apply htmlentities before insterting message in db

### LINKED ISSUE #9726: [BOOM-4929] Order message UTF8 problem
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-4929).

- _**Reporter:**_ RichardSon
- _**Created at:**_ Sat, 24 Feb 2018 20:32:59 +0100

<p>I found problem at older and new prestashop 1.7 versions also.</p>

<p>Order message which is added by customer on delivery step not working correctly with UTF8.</p>

<p> </p>

<p>When i want use for example this message :</p>

<p>"áno mám sa dobre" Then in ps_message is inserted this "áno mám sa dobre" so when i will return from payment step to delivery step then prestashop return that wrong message.</p>
        
Create new order and use my message as order message.

problem

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fixes display error in some cases
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9726
| How to test?  | See ticket, here is an example of text containing problematic characters: "toto & tata c'est, áno mám sa dobre"

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15840)

### DIFF
--- a/classes/checkout/CheckoutSession.php
+++ b/classes/checkout/CheckoutSession.php
@@ -86,7 +86,7 @@ class CheckoutSessionCore
 
     public function setMessage($message)
     {
-        $this->_updateMessage(Tools::safeOutput($message));
+        $this->_updateMessage($message);
 
         return $this;
     }


PierreRambaud (approved):

Robin-Fischer-PS:
Thanks for the PR @matthieu-rolland :heart:

PierreRambaud:
Thanks @matthieu-rolland

sotirisa:
Doesn't work for 1.7.6.7.

Any solution?

PierreRambaud:
> Doesn't work for 1.7.6.7.
> 
> Any solution?

This fix will be available for the next 1.7.7.0, be patient ;)

scwall:
https://github.com/PrestaShop/PrestaShop/pull/15840/commits/c1d2d539eae39868182abc1aac35ce111b465381 :
  ```
  public function setMessage($message)
    {
        $this->_updateMessage(Tools::safeOutput($message));
        $this->_updateMessage($message);

        return $this;
    }

```
If you remove :safeOutput you remove the SQL injection protection, it's not the best idea, is it? It would be better not to process the display than to save directly in the DB

**Comment by scwall** [comment]
modify in **console/themes/default/template/controllers/customer_threads/helpers/view/message.tpl**
and modify escape to unescape in message body and working too

```
<div class="message-body">
********************

		<p class="message-item-text">{$message.message|unescape:'html':'UTF-8'|nl2br}</p>
	</div>
```

HartLarsson:
i hope this patch will be not applied will remove any Injection control

PR #15855, opened 2019-10-07
Allow anonymous mode for AdminController

### LINKED ISSUE #15725: Rebuild search index URL and cron
**Describe the bug**
Rebuild search index URL can not be called from a cron, as it requires a login

**To Reproduce**
1. Open rebuild search index URL that appears at `Shop parameters - Search` in a private tab (without login) or in a cron
2. Error. It requires BO login. Index is not rebuilt.

**Additional information**
PrestaShop version: 1.7.6.1

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Add `allowAnonymous` property in `AdminController` in order to execute cron.  
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15725 
| How to test?  | The cron script can be executable without having an employee connected.

Example:
```php
    /**
     * {@inheritdoc}
     */
    public function init()
    {
        if ($this->isCronTask()
            && substr(
                _COOKIE_KEY_,
                static::TOKEN_CHECK_START_POS,
                static::TOKEN_CHECK_LENGTH
            ) === Tools::getValue('token')
        ) {
            $this->isAnonymousAllowed(true);
        }

        parent::init();
    }

    /**
     * Check if a task is a cron task
     *
     * @return bool
     */
    protected function isCronTask()
    {
        return Tools::isSubmit('action') && 'searchCron' === Tools::getValue('action');
    }

```

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15855)

### DIFF
--- a/classes/controller/AdminController.php
+++ b/classes/controller/AdminController.php
@@ -2887,6 +2887,11 @@ class AdminControllerCore extends Controller
      */
     public function initShopContext()
     {
+        // Do not initialize context when the shop is not installed
+        if (defined('PS_INSTALLATION_IN_PROGRESS')) {
+            return;
+        }
+
         // Change shop context ?
         if (Shop::isFeatureActive() && Tools::getValue('setShopContext') !== false) {
             $this->context->cookie->shopContext = Tools::getValue('setShopContext');
--- a/classes/controller/AdminController.php
+++ b/classes/controller/AdminController.php
@@ -247,6 +247,9 @@ class AdminControllerCore extends Controller
     /** @var HelperList */
     protected $helper;
 
+    /** @var bool */
+    private $allowAnonymous = false;
+
     /** @var int DELETE access level */
     const LEVEL_DELETE = 4;
 
@@ -770,7 +773,7 @@ class AdminControllerCore extends Controller
      */
     public function checkToken()
     {
-        if (TokenInUrls::isDisabled()) {
+        if (TokenInUrls::isDisabled() || $this->isAnonymousAllowed()) {
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


classes/controller/AdminController.php
```diff
@@ -2794,7 +2797,17 @@ public function init()
                 $this->context->cookie->last_activity = time();
             }
         }
-        if ($this->controller_name != 'AdminLogin' && (!isset($this->context->employee) || !$this->context->employee->isLoggedBack())) {
+
+        if (
+            !$this->allowAnonymous
```
Progi1984: May be a getter for this attribute could be useful for allow anonymous under conditions.

matks:
@PierreRambaud So security is 100% provided by the token ?

classes/controller/AdminController.php
PierreRambaud: Done

PierreRambaud:
@matks Yes it is, and only when running in anonymous mode. But the token is not the same as the one used for csrf protection.

Progi1984 (dismissed):

classes/controller/AdminController.php
Progi1984: :ok_hand: :+1:

Progi1984 (dismissed):

matks (approved):

matks:
Thank you @PierreRambaud

matks:
There might a BC break in this PR 🤔see the diff

eternoendless:
There's no breaking change actually, the removed function exists in the parent class.

PR #15859, opened 2019-10-07
Add clickable row option for grid actions

### LINKED ISSUE #15831: Direct access to customer's view missing from Customers list row
**Describe the bug**
In Customers > Customers

Missing behavior following the "Customers" page migration.
In 1.7.5.2. when clicking on a row, it opens the customer's view page.
In 17.6.1 it doesn't work anymore.

Workaround:
It is still possible to access the customer's view via the actions dropdown on the right side of the rows. But it takes 2 clicks instead of one #productivity. 

**To Reproduce**
1. Go to Customers > Customers
2. Click on a customer row 
4. See error -> Nothing happens

**Screenshots**
https://drive.google.com/file/d/1CTavM3hvYwZCDtWWHiWqamU14kvJVkUL/view?usp=sharing

[UPDATED OCT 8 19]
**Specs for all grids**
https://github.com/PrestaShop/prestashop-specs/pull/71/files

**Additional information**
PrestaShop version: 1.7.6.1

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | To improve the UX of grids and allow the user to click on the row directly we added an option to grid definitions, and more precisely the row LinkAction so that a click on the row triggers it
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15831
| How to test?  | We need to test all migrated pages with new grids, check that the row is now clickable to view/edit the row (pay close attention to the Category page where the action changes depending on the number of sub children). Pages affected by this PR: `Backup, Category, CmsPageCategory, CmsPage, Contact, Currency, Customer, Employee, Language, ManufacturerAddress, Manufacturer, Meta, Profile, RequestSql, Supplier, Tax, WebserviceKey`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15859)


admin-dev/themes/new-theme/js/components/grid/extension/link-row-action-extension.js
```diff
@@ -35,6 +35,23 @@ export default class LinkRowActionExtension {
...
+        const $parentCell = $rowLink.closest('td');
+        const clickableCells = $('td.data-type, td.identifier-type, td.badge-type', $parentRow)
+          .not($parentCell)
+        ;
+
+        clickableCells.addClass('cursor-pointer').click(() => {
+          document.location = $rowLink.attr('href');
+        });
+      });
+    });
```
sarjon, line 86: It would be nice to wrap it into more meaningful function.

sarjon:
Shouldn't improvement go to next minor? :thinking:

admin-dev/themes/new-theme/js/components/grid/extension/link-row-action-extension.js
jolelievre: Yep, done

jolelievre:
> Shouldn't improvement go to next minor? 🤔

Actually it's considered a regression because the feature was available in 1.6 (and all legacy pages in 1.7), that's why it's integrated in a patch version This feature had been forgotten on all migrated pages
But although it is a regression I had trouble labeling it as a bug fix 😅 But I can change type if needed @samuel-pires @eternoendless wdyt?

src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Actions/Row/link.html.twig
```diff
@@ -37,6 +37,7 @@
     data-placement="top"
     data-original-title="{{ action.name }}"
   {% endif %}
+  {% if action.options.clickable_row|default(false) %}data-clickable-row="1"{% endif %}
```
PierreRambaud: Why not? 
```suggestion
 data-clickable-row=" {% if action.options.clickable_row|default(false) %}"
```

JQuery can works with boolean values.

samuel-pires:
> Shouldn't improvement go to next minor? 🤔

Hi @sarjon, 
We decided to label the corresponding issue as a Regression and thus decided to add it in the next 
patch, the condition is that the technical solution doesn't take too much time and is pertinent.

admin-dev/themes/new-theme/js/components/grid/extension/link-row-action-extension.js
```diff
@@ -35,6 +35,23 @@ export default class LinkRowActionExtension {
...
+
+      $('.js-link-row-action[data-clickable-row=1]:first', $parentRow).each(function () {
+        const $rowLink = $(this);
+
+        const $parentCell = $rowLink.closest('td');
+        const clickableCells = $('td.data-type, td.identifier-type, td.badge-type', $parentRow)
+          .not($parentCell)
+        ;
+
+        clickableCells.addClass('cursor-pointer').click(() => {
```
PierreRambaud, line 78: What if you click on a checkbox, or a link in the cell?

jolelievre: Which is why I only apply the click on some of the cells: `td.data-type, td.identifier-type, td.badge-type` This way toggle cells, checkboxes or actions still work correctly

PierreRambaud: Can you add a little comment to say that?
And are you sure with theses type you're not able to add link? :trollface: 

Other solution could be to get the clicked element and be sure it's a `td` or a `text`.

src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Actions/Row/link.html.twig
jolelievre: Done

admin-dev/themes/new-theme/js/components/grid/extension/link-row-action-extension.js
jolelievre: Yes link are added using a LinkColumn to the definition which includes a `link-type` cell
So if you choose not to use the correct type, and override an existing template to include a link where there shouldn't be one... then you probably deserve it!!

I tried to check the event at first but the behaviour was "random" sometimes there's a div, sometimes it's the td which trigger the event, when it's a button you have other use cases Anyway it resulted in overly complex checks that still would allow false positive whereas this method is simpler to understand and (I think) guaranties the expected behaviour (we can always add additional types if need be)

jolelievre: Comment added

ghost:
> But although it is a regression I had trouble labeling it as a bug fix 😅 But I can change type if needed @samuel-pires @eternoendless wdyt?

Well I see why you don't want to label it as bug, though the behavior differs from the one defined in the specs, thus, according to the @eternoendless is is considered a bug :)

Progi1984 (dismissed):

sarahdib:
@jolelievre 

This is not working for : 
- CmsPage
- CmsPageCategory -> only working on the ID column

jolelievre:
@sarahdib Ok t's fixed, I had forgotten to activate the feature for CMS pages

About the CMS category page, actually only non-interactive columns are clickable (to avoid breaking the behaviour on checkboxes, or toggle buttons). So only ID and description were clickable, but I added the position columns since it's just an information and no interaction, this should be more intuitive now.

Progi1984 (approved):

matks:
Thank you @jolelievre

ghost:
Thanks @jolelievre 🎉

PR #15871, opened 2019-10-08
Avoid error when trying to translate an invalid module

PululuK:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Check to avoid the error: : ```Fatal Error: Call to a member function isUsingNewTranslationSystem() on null```  when translating your modules ! This can happen if you have modules that have been removed, installed, or disabled incorrectly..
| Type?         | bug fix 
| Category?     |  CO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | N/A
| How to test?  | https://drive.google.com/file/d/1-QJ-4eJg1BiiNuSddXZ0U4AmvBC3jAJ4/view

### How to reproduce ?
https://drive.google.com/file/d/1-QJ-4eJg1BiiNuSddXZ0U4AmvBC3jAJ4/view

### How to test the PR ?
Check that you no longer have the fatal error, but a cleaner error message indicating that the module concerned is invalid

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15871)

### DIFF
--- a/src/Adapter/Translations/TranslationRouteFinder.php
+++ b/src/Adapter/Translations/TranslationRouteFinder.php
@@ -32,6 +32,8 @@ use PrestaShopBundle\Service\TranslationService;
 use Symfony\Component\HttpFoundation\ParameterBag;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Module;
+use Exception;
+use Symfony\Component\Translation\TranslatorInterface;
 
 /**
  * Class TranslationRouteFinder finds the correct route for translations.
@@ -72,20 +74,28 @@ class TranslationRouteFinder
      * @var ModuleRepositoryInterface
      */
     private $moduleRepository;
+    
+    /**
+     * @var TranslatorInterface
+     */
+    private $translator;
 
     /**
      * @param TranslationService $translationService
      * @param Link $link
      * @param ModuleRepositoryInterface $moduleRepository
+     * @param TranslatorInterface $translator
      */
     public function __construct(
         TranslationService $translationService,
         Link $link,
-        ModuleRepositoryInterface $moduleRepository
+        ModuleRepositoryInterface $moduleRepository,
+        TranslatorInterface $translator
     ) {
         $this->translationService = $translationService;
         $this->link = $link;
         $this->moduleRepository = $moduleRepository;
+        $this->translator = $translator;
     }
 
     /**
@@ -204,10 +214,10 @@ class TranslationRouteFinder
     private function isModuleUsingNewTranslationSystem($moduleName)
     {
         $module = $this->moduleRepository->getInstanceByName($moduleName);
-        if ($module instanceof Module) {
-            return $module->isUsingNewTranslationSystem();
+        if (!($module instanceof Module)) {
+            throw new Exception($this->translator->trans('Invalid module : %s', array($moduleName),'Admin.International.Notification'));
         }
 
-        return false;
+        return $module->isUsingNewTranslationSystem();
     }
 }
--- a/src/Adapter/Translations/TranslationRouteFinder.php
+++ b/src/Adapter/Translations/TranslationRouteFinder.php
@@ -215,7 +215,7 @@ class TranslationRouteFinder
     {
         $module = $this->moduleRepository->getInstanceByName($moduleName);
         if (!($module instanceof Module)) {
-            throw new Exception($this->translator->trans('Invalid module : %s', array($moduleName),'Admin.International.Notification'));
+            throw new Exception($this->translator->trans('Invalid module : %s', array($moduleName), 'Admin.International.Notification'));
         }
 
         return $module->isUsingNewTranslationSystem();
--- a/src/Adapter/Translations/TranslationRouteFinder.php
+++ b/src/Adapter/Translations/TranslationRouteFinder.php
@@ -31,9 +31,9 @@ use PrestaShop\PrestaShop\Core\Addon\Module\ModuleRepositoryInterface;
 use PrestaShopBundle\Service\TranslationService;
 use Symfony\Component\HttpFoundation\ParameterBag;
 use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\Translation\TranslatorInterface;
 use Module;
 use Exception;
-use Symfony\Component\Translation\TranslatorInterface;
 
 /**
  * Class TranslationRouteFinder finds the correct route for translations.
@@ -214,6 +214,7 @@ class TranslationRouteFinder
     private function isModuleUsingNewTranslationSystem($moduleName)
     {
         $module = $this->moduleRepository->getInstanceByName($moduleName);
+        
         if (!($module instanceof Module)) {
             throw new Exception($this->translator->trans('Invalid module : %s', array($moduleName), 'Admin.International.Notification'));
         }
--- a/src/Adapter/Translations/TranslationRouteFinder.php
+++ b/src/Adapter/Translations/TranslationRouteFinder.php
@@ -74,7 +74,7 @@ class TranslationRouteFinder
      * @var ModuleRepositoryInterface
      */
     private $moduleRepository;
-    
+
     /**
      * @var TranslatorInterface
      */
@@ -214,7 +214,7 @@ class TranslationRouteFinder
     private function isModuleUsingNewTranslationSystem($moduleName)
     {
         $module = $this->moduleRepository->getInstanceByName($moduleName);
-        
+
         if (!($module instanceof Module)) {
             throw new Exception($this->translator->trans('Invalid module : %s', array($moduleName), 'Admin.International.Notification'));
         }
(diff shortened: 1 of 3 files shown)


Progi1984:
@PululuK Hi, thank you for your contribution. Please could you give us a scenario for testing this PR ? Thanks

matthieu-rolland (dismissed):
EDIT: actually, before validating I would like the opinion of another core dev

@PierreRambaud @jolelievre @Progi1984:

What do you think should happen in this method if the module is absent because it was badly removed (or for another weird reason). Do you guys think it should throw an exception ?

PierreRambaud:
Seems legit for me :+1:

src/Adapter/Translations/TranslationRouteFinder.php
```diff
@@ -203,7 +204,10 @@ public function findRouteParameters(ParameterBag $query)
     private function isModuleUsingNewTranslationSystem($moduleName)
     {
         $module = $this->moduleRepository->getInstanceByName($moduleName);
-
-        return $module->isUsingNewTranslationSystem();
+        if (Validate::isLoadedObject($module)) {
+            return $module->isUsingNewTranslationSystem();
+        } else {
```
PierreRambaud:
```suggestion
        }
```

src/Adapter/Translations/TranslationRouteFinder.php
```diff
@@ -203,7 +204,10 @@ public function findRouteParameters(ParameterBag $query)
...
+            return false;
```
PierreRambaud:
```suggestion

        return false;
```

src/Adapter/Translations/TranslationRouteFinder.php
```diff
@@ -203,7 +204,10 @@ public function findRouteParameters(ParameterBag $query)
...
+        }
```
PierreRambaud, line 211:
```suggestion
```

PululuK:
@PierreRambaud  Done ... Thank you for your suggestions!

PierreRambaud (dismissed):

Robin-Fischer-PS:
Hi @PululuK , thanks for the PR. Could you give us a scenario for testing this PR ? Thanks ;)

PululuK:
> Hi @PululuK , thanks for the PR. Could you give us a scenario for testing this PR ? Thanks ;)

Hello @Robin-Fischer-PS  @PierreRambaud ,

We had serious problems with a site, which I unfortunately can not expose here!
I will give a very basic example to reproduce this based on the principle **[Do not trust user input](https://www.enisa.europa.eu/publications/info-notes/the-dangers-of-trusting-user-input)**
 :

As you can see, the module name is retrieved from the URL parameters after the form is submitted via the GET method. So we can send anything.

Thanks for taking a look here please:

https://drive.google.com/file/d/1-QJ-4eJg1BiiNuSddXZ0U4AmvBC3jAJ4/view

src/Adapter/Translations/TranslationRouteFinder.php
```diff
@@ -203,7 +204,10 @@ public function findRouteParameters(ParameterBag $query)
     private function isModuleUsingNewTranslationSystem($moduleName)
     {
         $module = $this->moduleRepository->getInstanceByName($moduleName);
+        if (Validate::isLoadedObject($module)) {
+            return $module->isUsingNewTranslationSystem();
+        }
 
-        return $module->isUsingNewTranslationSystem();
+        return false;
```
eternoendless: This should throw an Exception if the module is not valid, not silently return false.

There should also be a try/catch block in TranslationsController::modifyTranslationsAction with a flash error message and a redirection to the overview page.

src/Adapter/Translations/TranslationRouteFinder.php
```diff
@@ -31,6 +31,7 @@
 use PrestaShopBundle\Service\TranslationService;
 use Symfony\Component\HttpFoundation\ParameterBag;
 use Symfony\Component\PropertyAccess\PropertyAccess;
+use Validate;
```
eternoendless:
```suggestion
use Module;
```

src/Adapter/Translations/TranslationRouteFinder.php
```diff
@@ -203,7 +204,10 @@ public function findRouteParameters(ParameterBag $query)
     private function isModuleUsingNewTranslationSystem($moduleName)
     {
         $module = $this->moduleRepository->getInstanceByName($moduleName);
+        if (Validate::isLoadedObject($module)) {
```
eternoendless: I think this check would be more adequate:

```suggestion
        if ($module instanceof Module) {
```

eternoendless (changes requested):
Hi @PululuK 

Thanks for this contribution! Some adjustments need to be done for this use case to be properly managed.

src/Adapter/Translations/TranslationRouteFinder.php
PululuK: Hello @eternoendless Thanks !

The legacy controller is still present for modules not used by the new translation system. We agree that `AdminTranslationsController` (legacy) will also be impacted?

I do not think that modifs only at this level `TranslationsController :: modifyTranslationsAction` can cover all impacted zones.

What do you think ?

PululuK:
Hi @PierreRambaud @eternoendless @matthieu-rolland , patches have been applied.
Can you check please ? Thanks !

PierreRambaud (dismissed):

PierreRambaud:
@PululuK PrettyCi needs to be run :)

PululuK:
Hi @PierreRambaud 

Should I do something?

Thank you

PululuK:
> @PululuK PrettyCi needs to be run :)

Hi @PierreRambaud

Should I do something?

Thank you

PierreRambaud:
You can check https://github.com/PrestaShop/PrestaShop/pull/15871/checks?check_run_id=287417626
You have unwanted spaces :)

PululuK:
>  You have unwanted spaces :)

Done ...
This is not the first time that **PrettyCI** has been trying to show me her muscles :)
Thank you @PierreRambaud

PierreRambaud (dismissed):

src/Adapter/Translations/TranslationRouteFinder.php
```diff
@@ -72,19 +75,27 @@ class TranslationRouteFinder
...
      * @param Link $link
      * @param ModuleRepositoryInterface $moduleRepository
+     * @param TranslatorInterface $translator
      */
     public function __construct(
         TranslationService $translationService,
         Link $link,
-        ModuleRepositoryInterface $moduleRepository
+        ModuleRepositoryInterface $moduleRepository,
+        TranslatorInterface $translator
```
eternoendless: I'm not sure about injecting the translator here. In addition, it would be considered a breaking change.

src/Adapter/Translations/TranslationRouteFinder.php
```diff
@@ -204,6 +215,10 @@ private function isModuleUsingNewTranslationSystem($moduleName)
     {
         $module = $this->moduleRepository->getInstanceByName($moduleName);
 
+        if (!($module instanceof Module)) {
+            throw new Exception($this->translator->trans('Invalid module : %s', array($moduleName), 'Admin.International.Notification'));
```
eternoendless: I think it would be best to throw an untranslated `PrestaShopBundle\Exception\InvalidModuleException`, which would have to be catched in `TranslationsController::modifyTranslationsAction()` and transformed into a proper error message.

For this you'd need to [add a flash message](https://symfony.com/doc/3.4/controller.html#flash-messages) and redirect back to the `admin_international_translations_show_settings` route.

And then, I think a better error message would be: "An error has occurred, this module does not exist: %s" which already already exists in `Admin.International.Notification`

eternoendless (dismissed):

src/Adapter/Translations/TranslationRouteFinder.php
PululuK: Thank you @eternoendless  ! Done

src/Adapter/Translations/TranslationRouteFinder.php
PululuK: I completely agree ! 
Why make it complicated when it can be simple ? 😀

Done

PululuK:
Hi @LouiseBonnard @khouloudbelguith 
Can you remove **waiting for author** label please ?
ping @eternoendless @PierreRambaud @matthieu-rolland 
Thank you

atomiix (dismissed):

Progi1984 (dismissed):

marionf:
Guys, what is the status on this PR ? Does it needs changes ? Or is it ready for QA ?

eternoendless:
This PR needs to be rebased into 1.7.7.x

PululuK:
> This PR needs to be rebased into 1.7.7.x

Can you help me to rebase please ?

eternoendless (approved):

eternoendless:
I did it myself, don't worry @PululuK :)

eternoendless:
By the way, the video linked in the "how to test" section is no longer available. Would you please re-upload it?

atomiix (approved):

PululuK:
> By the way, the video linked in the "how to test" section is no longer available. Would you please re-upload it?

Done ! (I am not proud of this video)

matks (approved):

PierreRambaud (approved):

SD1982:
LGTM Thanks @PululuK

PierreRambaud:
Thanks @PululuK

PR #15872, opened 2019-10-08
Check mimum quantity for products in the basket order

### LINKED ISSUE #14909: "Minimum quantity for sale" issue
This issue occurs in all PrestaShop versions (Checked in all versions above 1.7.4.x):

The user is able to order less quantity than "Minimum quantity for sale" setting.

To reproduce the issue:

1. Initially, the mimimum quantity is set 1

image

2. Added the product to cart with 1 quantity

image

3. Change product minimum quantity for sale to 10 (opr any greater value than previous value.)

image

4. Refresh cart page. The quantity will show updated but prices will not be updated anywhere

image

5. Make the order, it will be successful with old incorrect quantity.

image

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fixes bug occuring when someone adds a product on a single, someone in the BO changes the minimal quantity, and later the FO user refreshes his basket page.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14909
| How to test?  | See the referenced issue, but to sum up: 

1/ In FO, add a product in your basket (quantity: one)
2/ In BO, increase the minimum quantity for this product (ex: 10)
3/ Refresh the basket page
Expected result: 
- You get an error message warning you about the minimum quantity
- If you don't refresh but click on "order" (commander) directly, it will redirect you back to the basket page, with the same error message

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15872)

### DIFF
--- a/controllers/front/CartController.php
+++ b/controllers/front/CartController.php
@@ -98,7 +98,8 @@ class CartControllerCore extends FrontController
             Tools::redirect('index.php');
         }
 
-        // check that minimal quantity conditions are respected for each product in the cart
+        /* check that minimal quantity conditions are respected for each product in the cart
+         (this is to be applied only on page load, not for ajax calls) */
         if (!Tools::getValue('ajax')) {
             $this->checkCartProductsMinimalQuantities();
         }
--- a/controllers/front/CartController.php
+++ b/controllers/front/CartController.php
@@ -492,7 +492,17 @@ class CartControllerCore extends FrontController
                     false,
                     true
                 );
-                $update_quantity = $this->updateCartProductQuantity((int) $this->qty, (int) $this->id_product);
+                $update_quantity = $this->context->cart->updateQty(
+                    $this->qty,
+                    $this->id_product,
+                    $this->id_product_attribute,
+                    $this->customization_id,
+                    Tools::getValue('op', 'up'),
+                    $this->id_address_delivery,
+                    null,
+                    true,
+                    true
+                );
                 if ($update_quantity < 0) {
                     // If product has attribute, minimal quantity is set with minimal quantity of attribute
                     $minimal_quantity = ($this->id_product_attribute)
@@ -659,35 +669,7 @@ class CartControllerCore extends FrontController
                         'Shop.Notifications.Error',
                     )
                 );
-                // increase cart's product quantity up to minimal quantity
-                $addedQuantity = (int) $product['minimal_quantity'] - (int) $product['cart_quantity'];
-                $this->updateCartProductQuantity($addedQuantity, (int) $product['id_product']);
             }
         }
-    }
-
-    /**
-     * Wrapper for the cart's updateQty method, not to repeat the long list of parameters
-     * at each use of the method.
-     * It increases the quantity for a product in the cart.
-     *
-     * @param int $addedQuantity
-     * @param int $productId
-     *
-     * @return bool Whether the quantity has been successfully updated
-     */
-    private function updateCartProductQuantity(int $addedQuantity, int $productId)
-    {
-        return $this->context->cart->updateQty(
-            $addedQuantity,
-            $productId,
-            $this->id_product_attribute,
-            $this->customization_id,
-            Tools::getValue('op', 'up'),
-            $this->id_address_delivery,
-            null,
-            true,
-            true
-        );
     }
 }
--- a/controllers/front/CartController.php
+++ b/controllers/front/CartController.php
@@ -98,6 +98,8 @@ class CartControllerCore extends FrontController
             Tools::redirect('index.php');
         }
 
+        // check that minimal quantity conditions are respected for each product in the cart
+        $this->checkCartProductsMinimalQuantities();
         $presenter = new CartPresenter();
         $presented_cart = $presenter->present($this->context->cart, $shouldSeparateGifts = true);
 
@@ -490,17 +492,7 @@ class CartControllerCore extends FrontController
                     false,
                     true
                 );
-                $update_quantity = $this->context->cart->updateQty(
-                    $this->qty,
-                    $this->id_product,
-                    $this->id_product_attribute,
-                    $this->customization_id,
-                    Tools::getValue('op', 'up'),
-                    $this->id_address_delivery,
-                    null,
-                    true,
-                    true
-                );
+                $update_quantity = $this->updateCartProductQuantity((int) $this->qty, (int) $this->id_product);
                 if ($update_quantity < 0) {
                     // If product has attribute, minimal quantity is set with minimal quantity of attribute
                     $minimal_quantity = ($this->id_product_attribute)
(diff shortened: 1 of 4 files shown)


controllers/front/CartController.php
```diff
@@ -98,6 +98,10 @@ public function initContent()
             Tools::redirect('index.php');
         }
 
+        // check that minimal quantity conditions are respected for each product in the cart
+        if (!Tools::getValue('ajax')) {
```
PierreRambaud, line 105: Any reason to don't check it in ajax context?

themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl
```diff
@@ -129,7 +129,6 @@
                 type="number"
                 value="{$product.quantity}"
                 name="product-quantity-spin"
-                min="{$product.minimal_quantity}"
```
PierreRambaud, line 132: You should keep this one maybe? :thinking:

controllers/front/CartController.php
matthieu-rolland: Ajax calls on the cart page also go through this method, but does its own checks later and sends back a raw html template to update the page.

Here we want to do this check when the page is loading (or being refreshed), just in case someone comes back to a cart he created earlier, and minimal quantities required for a product changed since then.

(if I don't add this check ajax calls would display two error messages)

themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl
matthieu-rolland: This html 5 min attribute was useful because it automatically fixed the value according to minimal quantity at page load. 

So for example if a user had one item of a product in his cart, and in BO the minimal quantity for this product is changed to 10, then when the user refreshes his cart, the html 5 forces the value to match the minimal quantity.

But as seen with @colinegin we don't want this behavior anymore, we want to keep the old value (even if it's below the minimal quantity required), and it's the user's job to change the value before continuing his order. So that he doesn't order more products that he intented to do in the first place. Instead we display an error message (and he can't click on 'order' because of the error message).

Each change in this input field triggers an ajax call anyway, and the field goes back to its preview value if the backend doesn't validate it (no more product or the minimal quantity is not met)

I checked that this template is only used for the cart.

PierreRambaud: So be careful with https://github.com/PrestaShop/PrestaShop/blob/develop/themes/_core/js/cart.js#L94 and https://github.com/PrestaShop/PrestaShop/blob/develop/themes/_core/js/product.js#L156
and others :roll_eyes:

matthieu-rolland: You're right, I'll check it out 🤔

controllers/front/CartController.php
jolelievre: Then you should add a comment to explain this behaviour

matthieu-rolland: good idea, just did it 👍

PierreRambaud (changes requested):
> Reviewed 1 of 4 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15872)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @matthieu-rolland and @PierreRambaud)

___
*[controllers/front/CartController.php, line 101 at r2](https://reviewable.io/reviews/prestashop/prestashop/15872#-LrDefy73BvPmeP9zGDg:-LrDefy73BvPmeP9zGDh:bkhnv09) ([raw file](https://github.com/prestashop/prestashop/blob/ff8149b772a8db72ade7b974931f79e6bf1a49e2/controllers/front/CartController.php#L101)):*
> ```PHP
>         }
> 
>         /* check that minimal quantity conditions are respected for each product in the cart
> ```

 ```
        /** 
          * Check that minimal quantity conditions are respected for each product in the cart
          * (this is to be applied only on page load, not for ajax calls) 
          */
```

Watch out with comments indent, if you use multi lines you need to use this syntax

**Review by PierreRambaud — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15872)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @matthieu-rolland and @PierreRambaud)

___
*[themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl, line 132 at r2](https://reviewable.io/reviews/prestashop/prestashop/15872#-LrDhO7V3K3vap3O28Pa:-LrDhO7W-G-iA2BUdibG:bo4d0se) ([raw file](https://github.com/prestashop/prestashop/blob/2360d5d5b2435ef3d632b033bca0c53d3bf2d4f3/themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl#L132)):*
> ```Smarty
>                 value="{$product.quantity}"
>                 name="product-quantity-spin"
>                 min="{$product.minimal_quantity}"
> ```

Still use by js, isn't a problem?

File: themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl (hunk 2 above)
Comment (matthieu-rolland) [reply themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl:132]: @PierreRambaud 

So I checked, and in those two files we are dealing with the 'min' attribute from the quantity input in the **product** page. I only removed the one on the **cart** page.

I was going to add a custom data attribute, but even in the cart.js, the min attribute that is used comes from the product page (the value is used when we click on "add to cart")

I couldn't find any use of the 'min' attribute for the quantity field in the **cart** page.

**Comment by matthieu-rolland** [comment]
> > _[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15872)_ status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @matthieu-rolland and @PierreRambaud)
> 
> _[themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl, line 132 at r2](https://reviewable.io/reviews/prestashop/prestashop/15872#-LrDhO7V3K3vap3O28Pa:-LrDhO7W-G-iA2BUdibG:bo4d0se) ([raw file](https://github.com/prestashop/prestashop/blob/2360d5d5b2435ef3d632b033bca0c53d3bf2d4f3/themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl#L132)):_
> 
> > ```
> >                 value="{$product.quantity}"
> >                 name="product-quantity-spin"
> >                 min="{$product.minimal_quantity}"
> > ```
> 
> Still use by js, isn't a problem?

@PierreRambaud  It's used in JS, and the attribute is here, so no I don't think it's a problem ? 🤔 
The attribute I removed is in `cart-detailed-product-line.tpl`, and this is one is not used in JS, as far as I know (do you think I missed something? )

File: themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl (hunk 2 above)
Comment (PierreRambaud) [reply themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl:132]: Good news :+1:

**Review by matthieu-rolland — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15872)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @PierreRambaud)

___
*[controllers/front/CartController.php, line 101 at r2](https://reviewable.io/reviews/prestashop/prestashop/15872#-LrDefy73BvPmeP9zGDg:-LrIoGiV0Dq-LCeLOoUJ:bs7wuaw) ([raw file](https://github.com/prestashop/prestashop/blob/ff8149b772a8db72ade7b974931f79e6bf1a49e2/controllers/front/CartController.php#L101)):*
<details><summary><i>Previously, PierreRambaud (GoT) wrote…</i></summary><blockquote>

 ```
        /** 
          * Check that minimal quantity conditions are respected for each product in the cart
          * (this is to be applied only on page load, not for ajax calls) 
          */
```

Watch out with comments indent, if you use multi lines you need to use this syntax 

</blockquote></details>

Done !

matthieu-rolland:
@PierreRambaud PrettyCi seems happy now 👍

PierreRambaud (approved):

matks:
Thank you @matthieu-rolland

PR #15887, opened 2019-10-09
Set value for list total automatically in HelperList generation

JBWModules:
| Questions     | Answers
| Branch?       | develop 
| Description?  | As the rows are defined by the function parameter $list anyway, we can use it to calculate and display the list total - no need to set this value outside when using the HelperList in this case.
| Type?         | improvement 
| Category?     | BO 
| How to test?  | Create and display a helper list - number of list entries should be visible without settting it

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15887)

### DIFF
--- a/classes/helper/HelperList.php
+++ b/classes/helper/HelperList.php
@@ -182,7 +182,7 @@ class HelperListCore extends Helper
 
         $this->orderWay = preg_replace('/^([a-z _]*!)/Ui', '', $this->orderWay);
 
-        if($this->listTotal == 0 && is_array($list) && count($list) > 0) {
+        if ($this->listTotal == 0 && is_array($list) && count($list) > 0) {
             $this->listTotal = count($list);
         }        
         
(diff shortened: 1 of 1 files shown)


Progi1984:
@jbroutier Hi, Thank you for your contribution. Could you create an issue for explaining your problem, please ? Thnaks

JBWModules:
@Progi1984  There is no issue - this is an improvement.

Progi1984:
@JBWModules I understand, but this require an issue in order to help us to see the initial problem and see if the fix is pertinent. Furthermore, could you fix Travis CI & PrettyCI ? Thanks

matks:
Hello, it seems this PR is now stale since 2019.

I close this PR, following our [review guidelines](https://devdocs.prestashop.com/1.7/project/maintainers-guide/reviewing-pull-requests/) "Pull Requests may be closed after 30 days of inactivity".

If however you are still interested in it, please ping us 😉  we can reopen.

PR #15902, opened 2019-10-10
Use the correct environment depending on what you are doing (dev, test, and prod)

PierreRambaud:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add `_PS_ENV_` for symfony and legacy context and make a real distinction between `dev`, `test` and `prod`.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Required for #14894 
| How to test?  | Travis must be green. You're able to install prestashop in dev and production env.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15902)


app/config/config_test.yml
```diff
@@ -1,6 +1,7 @@
 parameters:
     AdapterSecurityAdminClass:                LegacyTests\PrestaShopBundle\Mock\AdapterSecurityAdminMock
     prestashop.security.voter.product.class:  LegacyTests\PrestaShopBundle\Mock\PageVoter
+    database_name: "test_prestashop"
```
jolelievre: This means the test database will always be the same one.. It's a problem when you manage multiple projects/databases/test_databases in the same environment..

composer.json
```diff
@@ -155,8 +155,7 @@
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::prepareDeploymentTarget"
         ],
         "create-test-db": [
-            "@php ./tests-legacy/create-test-db.php",
-            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache"
+            "@php ./tests-legacy/create-test-db.php"
```
jolelievre, line 158: Why no more clear cache? I love clear cache!!

src/Adapter/Environment.php
```diff
@@ -58,12 +58,11 @@ public function __construct($isDebug = null, $name = null)
             $this->isDebug = $isDebug;
         }
 
-        $this->name = $name;
-        if (null === $this->name) {
-            if (!empty($_SERVER['APP_ENV'])) {
-                $this->name = $_SERVER['APP_ENV'];
-            } elseif (defined('_PS_IN_TEST_') && _PS_IN_TEST_) {
-                $this->name = 'test';
+        if (null !== $this->name) {
```
jolelievre:
```suggestion
        if (null !== $name) {
```

src/Adapter/Environment.php
```diff
@@ -91,6 +90,10 @@ public function isDebug()
      */
     public function getCacheDir()
     {
+        if (defined('_PS_CACHE_DIR_')) {
```
jolelievre, line 93: Noice!

src/PrestaShopBundle/Service/Command/AbstractCommand.php
```diff
@@ -50,8 +50,12 @@ public function __construct(AppKernel $kernel = null)
         set_time_limit(0);
 
         if (null === $kernel) {
-            require_once _PS_ROOT_DIR_ . '/app/AppKernel.php';
-            $kernel = new AppKernel(_PS_MODE_DEV_ ? 'dev' : 'prod', false);
+            global $kernel;
```
jolelievre, line 53: The return of the `global` ...

jolelievre: How come, since it's a Symfony command, kernel can be null? Isn't the command instanciated by the kernel anyway?

tests-legacy/PrestaShopBundle/Controller/ControllerTest.php
```diff
@@ -149,8 +149,7 @@ public function getControllersClasses()
 
     protected function declareRequiredConstants()
     {
-        $configDirectory = __DIR__ . '/../../../app/config';
-        $configuration = require_once $configDirectory . '/parameters.php';
+        $configuration = require_once _PS_CACHE_DIR_ . 'appParameters.php';
```
jolelievre, line 152: Why change this include? What if the cache folder has been emptied? Using the file in config seems safer no?

tests-legacy/PrestaShopBundle/Utils/DatabaseCreator.php
```diff
@@ -47,7 +47,11 @@ public static function createTestDB()
         $install = new Install();
         \DbPDOCore::createDatabase(_DB_SERVER_, _DB_USER_, _DB_PASSWD_, _DB_NAME_, false);
         $install->clearDatabase(false);
-        $install->installDatabase(true);
+        if (!$install->installDatabase(true)) {
+            // Something went wrong during installation
+            exit(1);
```
jolelievre, line 52: Can you throw an exception rather than using `exit` ? Too many unexpected behaviour happens from the use of exit.. (stoping tests suite, breaking ajax request, debug sessions, ruining friday nights...)

jolelievre (changes requested):
Nice improvement! A few things worry me though

jolelievre:
Oh and also I think you can find a better name for this PR

composer.json
PierreRambaud: We shouldn't need to clear the `dev` cache when running the `test` env :wink:

src/PrestaShopBundle/Service/Command/AbstractCommand.php
PierreRambaud: The kernel is always loaded. I will try to investigate to improve this.
Don't ask me why it was like that :sweat_smile:

tests-legacy/PrestaShopBundle/Controller/ControllerTest.php
PierreRambaud: parameters.php doesn't take in consideration what have be done in boostrap.
the `appParemeters.php` is the computed cache between `paremeters.yml`, and `parameters.php`, many configuration are missing if we don't use the right file. (needed for the utf8mb4 pull request)

tests-legacy/PrestaShopBundle/Utils/DatabaseCreator.php
PierreRambaud: We are in CLI context, the installDatabase already display an error message.

PierreRambaud:
> Oh and also I think you can find a better name for this PR

I press Enter too fast -_-

config/defines.inc.php
```diff
@@ -68,16 +68,30 @@
...
+) {
+    define('_PS_ENV_', 'test');
+} else {
+    define('_PS_ENV_', _PS_MODE_DEV_ ? 'dev': 'prod');
+}
 
-    $prestashopCacheDir =  _PS_ROOT_DIR_.'/var/cache/'.$env. DIRECTORY_SEPARATOR;
+if (!defined('_PS_CACHE_DIR_')) {
+    $prestashopCacheDir =  _PS_ROOT_DIR_.'/var/cache/' . _PS_ENV_ . DIRECTORY_SEPARATOR;
     define('_PS_CACHE_DIR_', $prestashopCacheDir);
```
Progi1984:
```suggestion
     define('_PS_CACHE_DIR_', _PS_ROOT_DIR_.'/var/cache/' . _PS_ENV_ . DIRECTORY_SEPARATOR);
```

src/PrestaShopBundle/Install/Install.php
```diff
@@ -103,7 +103,7 @@ public function setError($errors)
 
         if (null === $logger) {
             $cacheDir = _PS_ROOT_DIR_ . '/var/logs/';
-            $file = $cacheDir . (_PS_MODE_DEV_ ? 'dev' : 'prod') . '_' . @date('Ymd') . '_installation.log';
+            $file = $cacheDir . _PS_ENV_ . '_' . @date('Ymd') . '_installation.log';
```
Progi1984:
```suggestion
            $logDir = _PS_ROOT_DIR_ . '/var/logs/';
            $file = $logDir . _PS_ENV_ . '_' . @date('Ymd') . '_installation.log';
```

config/defines.inc.php
PierreRambaud: Done

src/PrestaShopBundle/Install/Install.php
PierreRambaud: Done

app/config/config_test.yml
PierreRambaud: It's the current behaviour (which add `test_` before the default database name). 
We will improve this later :)

src/Adapter/Environment.php
PierreRambaud: Done Thanks :)

Progi1984 (commented):
image
Why a right change ?

PierreRambaud:
> image
> Why a right change ?

Because we can't execute this file without `bash`, even if a shebang is present, which is totally stupid. (My IDE automatically change the chmod of files containing a shebang)

Progi1984 (approved):

classes/PrestaShopAutoload.php
```diff
@@ -104,13 +98,7 @@ public static function getCacheFileIndex()
...
     public static function getNamespacedStubFileIndex()
     {
-        if (defined('_PS_IN_TEST_')) {
-            $env = 'test';
-        } else {
-            $env = (_PS_MODE_DEV_) ? 'dev' : 'prod';
-        }
-
-        return _PS_ROOT_DIR_ . DIRECTORY_SEPARATOR . 'var' . DIRECTORY_SEPARATOR . 'cache' . DIRECTORY_SEPARATOR . $env . DIRECTORY_SEPARATOR . 'namespaced_class_stub.php';
+        return _PS_CACHE_DIR_ . 'namespaced_class_stub.php';
```
mickaelandrieu, line 101: is it the same directory ?

config/defines.inc.php
```diff
@@ -68,16 +68,29 @@
...
     define('_PS_BO_ALL_THEMES_DIR_', _PS_ADMIN_DIR_.'/themes/');
 }
-if (!defined('_PS_CACHE_DIR_')) {
-    if (defined('_PS_IN_TEST_')) {
-        $env = 'test';
-    } else {
-        $env = (_PS_MODE_DEV_) ? 'dev' : 'prod';
+
+// Find if we are running under a Symfony command
+$cliEnvValue = null;
```
mickaelandrieu, line 75: lets wrap this into a class with a public static method instead ?

install-dev/init.php
```diff
@@ -76,13 +76,8 @@
...
     require_once _PS_CORE_DIR_.'/config/bootstrap.php';
 
-    if (defined('_PS_IN_TEST_') && _PS_IN_TEST_) {
-        $env = 'test';
-    } else {
-        $env = _PS_MODE_DEV_ ? 'dev' : 'prod';
-    }
     global $kernel;
-    $kernel = new AppKernel($env, _PS_MODE_DEV_);
+    $kernel = new AppKernel(_PS_ENV_, _PS_MODE_DEV_);
```
mickaelandrieu, line 80: can we make it overridable using env variables?

Like we usually do in Symfony https://github.com/mickaelandrieu/resiliency-demo/blob/master/public/index.php#L23

src/PrestaShopBundle/Service/Command/AbstractCommand.php
mickaelandrieu: it wasn't the case before, doing that we have impacted the performances of PrestaShop... but now it's done, it's a good time to refactor it 👼 

these commands are not used using bin/console ^^tbh these are not real commands at all

classes/PrestaShopAutoload.php
PierreRambaud: yes

config/defines.inc.php
PierreRambaud: Not possible because we don't have access to the PrestaShop autoload at this moment

install-dev/init.php
PierreRambaud: Should be done in a new pull request because if we want to do it for the env mode, we should do it for debug and others

Progi1984 (dismissed):

app/config/config_test.yml
eternoendless: Why hard code this?

Progi1984 (dismissed):

app/config/config_test.yml
eternoendless: Why hard code this?

PierreRambaud: Was already half hard coded here https://github.com/PrestaShop/PrestaShop/pull/15902/files#diff-e24c8c5370061c36a5415c30ab98c8b4L56 :sweat_smile: 

And we can't add a circular reference 
```suggestion
    database_name: "test_%database_name%"
```

We should do a new PR to improve these configuration variables.

PierreRambaud: But we still must doing it for https://github.com/PrestaShop/PrestaShop/pull/15902/files#diff-2bad73d472db80e76e124fbfe9908964R100 because of legacy context and we must not load Symfony in front context right now :sob:

PierreRambaud: I changed it for https://github.com/PrestaShop/PrestaShop/pull/15902/commits/8b8585743b5ee146cc1811c0fa0e283732d64501 because it's currently not possible to have different parameters depending on the env.

Progi1984 (approved):

jolelievre (approved):

Issue #14663, opened 2019-07-15
Template \account.html.twig is not defined   (it's a bug ?)


fraria76:
Hi everyone,

I have this configuration:

WINDOWS SERVER 2019 with IIS 10
MySQL 8.0.16.0
PHP 7.2.16
PHPMYADMIN 4.9.0.1
Prestashop  1.7.6.0

I performed a new installation and at 23% appear a error message:

1: Template "C:\inetpub\wwwroot\XXXXXX\mails\themes\modern\core\account.html.twig" is not defined.

I specific than before I has Prestashop  v1.7.5.2 (with the same versions of MySQL, PHP, PHPMYADMIN) and the installation it's OK.

Thanks in advace for any help!﻿

![problema](https://user-images.githubusercontent.com/50572903/61229837-2f441700-a729-11e9-9c20-8d510bea10b5.jpg)

khouloudbelguith:
Hi @fraria76,

Thanks for your report.
I did not manage to reproduce the issue with PS1.7.6.0 stable version.
What is the exact Language/country did you used during the install?

Thanks!

fraria76:
Hi khouloudbelguith,

my installation language is ITALIAN.

I tried both to make a NEW installation Prestashop 1.7.6.0 and UPDATE from 1.7.5.2 but I have problems.

If I make a NEW installation Prestashop 1.7.6.0 at 23% appear a error message:    1: Template "C:\inetpub\wwwroot\XXXXXX\mails\themes\modern\core\account.html.twig" is not defined.

If I make a NEW installation Prestashop v1.7.5.2 the installation it's OK, but if I make a UPDATE to 1.7.6.0 appear a error message and If I refresh the page appear:   500 - Internal server error.

On PHPMYADMIN I use these character encoding for the database: utf8_general_ci

Kind regards﻿﻿

![error](https://user-images.githubusercontent.com/50572903/61293798-c619dd80-a7d4-11e9-840e-a59edf741841.jpg)

fraria76:
Hi khouloudbelguith,

I did a complete new installation test of Prestahop 1.7.6.0 with ENGLISH language: the result is the same.

But now when appear the error at 23%
1: Template "C:\inetpub\wwwroot\XXXXXX\mails\themes\modern\core\account.html.twig" is not defined.

I go on directory C:\inetpub\wwwroot\XXXXX\mails\themes and I delete MODERN directory.

The installation completed successfully.

I also send you the print-screen step by step.

[...]

PR #15906, opened 2019-10-10
Fixes issue Undefined index _route

jusurb:
| Questions     | Answers
| Branch?       | develop
| Descripton?| not sure why _route is expected to be part of caller parameters but it's not, the route is actually in the caller_route parameter, so we first merge parameters with defaults array to make sure _route key is set. + will not overwrite params with the same name
| Type?         | bug fix 
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes issue #14663
| How to test ? | Please check pagination BO pages not migrated to Symfony is working

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15906)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/CommonController.php
+++ b/src/PrestaShopBundle/Controller/Admin/CommonController.php
@@ -118,7 +118,7 @@ class CommonController extends FrameworkBundleAdminController
                 unset($callerParameters[$k]);
             }
         }
-        $routeName = $request->attributes->get('caller_route', ($callerParameters + ['_route' => false])['_route']);
+        $routeName = $request->attributes->get('caller_route', ($callerParameters + array('_route' => false))['_route']);
         $nextPageUrl = (!$routeName || ($offset + $limit >= $total)) ? false : $this->generateUrl($routeName, array_merge(
             $callerParameters,
             array(
--- a/src/PrestaShopBundle/Controller/Admin/CommonController.php
+++ b/src/PrestaShopBundle/Controller/Admin/CommonController.php
@@ -118,7 +118,7 @@ class CommonController extends FrameworkBundleAdminController
                 unset($callerParameters[$k]);
             }
         }
-        $routeName = $request->attributes->get('caller_route', $request->attributes->get('caller_parameters', ['_route' => false])['_route']);
+        $routeName = $request->attributes->get('caller_route', ($callerParameters + ['_route' => false])['_route']);
         $nextPageUrl = (!$routeName || ($offset + $limit >= $total)) ? false : $this->generateUrl($routeName, array_merge(
             $callerParameters,
             array(
--- a/src/PrestaShopBundle/Controller/Admin/CommonController.php
+++ b/src/PrestaShopBundle/Controller/Admin/CommonController.php
@@ -118,7 +118,9 @@ class CommonController extends FrameworkBundleAdminController
                 unset($callerParameters[$k]);
             }
         }
-        $routeName = $request->attributes->get('caller_route', ($callerParameters + array('_route' => false))['_route']);
+        
+        $callerParameters += array('_route' => false);
+        $routeName = $request->attributes->get('caller_route', $callerParameters['_route']);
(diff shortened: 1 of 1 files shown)


PierreRambaud (dismissed):

matks (dismissed):

jusurb:
@matks i don't understand what pretty is complaining about, it does not print what rule am i breaking or what needs to be changed, very unclear. Can you help me out and clarify?

src/PrestaShopBundle/Controller/Admin/CommonController.php
```diff
@@ -118,7 +118,9 @@ public function paginationAction(Request $request, $limit = 10, $offset = 0, $to
                 unset($callerParameters[$k]);
             }
         }
-        $routeName = $request->attributes->get('caller_route', $request->attributes->get('caller_parameters', ['_route' => false])['_route']);
+        
+        $callerParameters += array('_route' => false);
```
PierreRambaud, line 121: Always prefer short array syntax and use array_merge like this, otherwise `_route` will always be false.
```suggestion
        $callerParameters = array_merge(['_route' => false], $callerParameters);
```

src/PrestaShopBundle/Controller/Admin/CommonController.php
```diff
@@ -118,7 +118,9 @@ public function paginationAction(Request $request, $limit = 10, $offset = 0, $to
                 unset($callerParameters[$k]);
             }
         }
-        $routeName = $request->attributes->get('caller_route', $request->attributes->get('caller_parameters', ['_route' => false])['_route']);
+
```
PierreRambaud: PrettyCi doesn't like useless whitespaces :)
```suggestion

```

src/PrestaShopBundle/Controller/Admin/CommonController.php
jusurb: it won't that's the main difference between + union operator and array_merge:
```
>>> ['a' => 1] + ['a' => 2]
=> [
     "a" => 1,
   ]
```
```
>>> array_merge(['a' => 1], ['a' => 2])
=> [
     "a" => 2,
   ]
```
it does not override the already existing elements of array but will only append them if they don't exist. 
it's virtually the same as you wrote it with array_merge, just need to make sure the order of arguments is correct so the more important one always override the defaults, that's why i chose union

PierreRambaud: Oh my bad, always thought it'll be override :sweat_smile:

jusurb: I see we use array() everywhere, just wanted to be consistent, since it's not enforced by code style rules

PierreRambaud (approved):

sarahdib:
hello @Gamesh 

The issue linked seems to be the wrong one can you confirm ?

jusurb:
@sarahdib well it's correct kind of, the solution for the problem was provided by the last comment in that ticket
https://github.com/PrestaShop/PrestaShop/issues/14663#issuecomment-534450713

rather than changing the core files like it suggests i decided to make it a pull request, which was never done

matks:
Thank you @Gamesh

jusurb:
thanks

fabiorme:
hi, I found the same error, during a clean installation, from zero to prestashop.
I replaced the old part of the code with this new one but I still have the same problem, and I can't complete the installation
`
        $callerParameters = $request->attributes->get('caller_parameters', array());
        foreach ($callerParameters as $k => $v) {
            if (strpos($k, '_') === 0) {
                unset($callerParameters[$k]);
            }
        }
        $callerParameters += array('_route' => false);
        $routeName = $request->attributes->get('caller_route', $callerParameters['_route']);
        $nextPageUrl = (!$routeName || ($offset + $limit >= $total)) ? false : $this->generateUrl($routeName, array_merge(
            $callerParameters,
            array(
                $offsetParam => min($total - 1, $offset + $limit),
                $limitParam => $limit,
            )
        ));
`

config:
PHP: 7.1.19
Prestashop 1.7.6.1

presta_err_2

PR #15942, opened 2019-10-14
Fix docker compose not being able to stat files

### LINKED ISSUE #15941: Docker compose cp: cannot stat '/tmp/data-ps/prestashop/*'
docker-compose up 

**error message**
Reapplying PrestaShop files for enabled volumes ...
cp: cannot stat '/tmp/data-ps/prestashop/*': No such file or directory

**To Reproduce**
1. chown -hR www-data:www-data ./prestashop
2. git clone https://github.com/PrestaShop/PrestaShop.git
3. cd prestashop
4. docker-compose up

**Additional information**
docker
Docker version 19.03.2, build 6a30dfc
docker-compose
docker-compose version 1.24.1, build 4667896b

PierreRambaud:
| Questions     | Answers
| Branch?       | develop
| Description?  | Be sure we use the primary user (1000 in Unix environment)
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15941
| How to test?  | docker-compose up --build<br> Navigate to http://localhost:8001

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15942)

### DIFF
--- a/.docker/docker_run_git.sh
+++ b/.docker/docker_run_git.sh
@@ -36,9 +36,9 @@ if [ ! -f ./config/settings.inc.php ]; then
             echo "\n* Drop & recreate mysql database...";
             if [ $DB_PASSWD = "" ]; then
                 echo "\n* Dropping existing database $DB_NAME..."
-                mysql -h $DB_SERVER -P $DB_PORT -u $DB_USER -p$DB_PASSWD -e "drop database if exists $DB_NAME;"
+                mysql -h $DB_SERVER -P $DB_PORT -u $DB_USER -e "drop database if exists $DB_NAME;"
                 echo "\n* Creating database $DB_NAME..."
-                mysqladmin -h $DB_SERVER -P $DB_PORT -u $DB_USER create $DB_NAME -p$DB_PASSWD --force;
+                mysqladmin -h $DB_SERVER -P $DB_PORT -u $DB_USER create $DB_NAME --force;
             else
                 echo "\n* Dropping existing database $DB_NAME..."
                 mysql -h $DB_SERVER -P $DB_PORT -u $DB_USER -p$DB_PASSWD -e "drop database if exists $DB_NAME;"
--- a/.docker/docker_run_git.sh
+++ b/.docker/docker_run_git.sh
@@ -1,0 +1,76 @@
+#!/bin/sh
+
+if [ ! -f ./vendor/autoload.php  ]; then
+	echo "\n* Vendor autoloader not found, running composer ...";
+	runuser -g www-data -u www-data -- /usr/local/bin/composer install --no-interaction
+fi
+
+if [ "$DB_SERVER" = "<to be defined>" -a $PS_INSTALL_AUTO = 1 ]; then
+    echo >&2 'error: You requested automatic PrestaShop installation but MySQL server address is not provided '
+    echo >&2 '  You need to specify DB_SERVER in order to proceed'
+    exit 1
+elif [ "$DB_SERVER" != "<to be defined>" -a $PS_INSTALL_AUTO = 1 ]; then
+    RET=1
+    while [ $RET -ne 0 ]; do
+        echo "\n* Checking if $DB_SERVER is available..."
+        mysql -h $DB_SERVER -P $DB_PORT -u $DB_USER -p$DB_PASSWD -e "status" > /dev/null 2>&1
+        RET=$?
+
+        if [ $RET -ne 0 ]; then
+            echo "\n* Waiting for confirmation of MySQL service startup";
+            sleep 5
+        fi
+    done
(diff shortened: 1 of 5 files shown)


matks (dismissed):

matks:
Need one more review

@Quetzacoalt91 ?

docker-compose.yml
```diff
@@ -1,37 +1,31 @@
...
-            DB_SERVER: mysql
-            PS_DOMAIN: localhost:8001
-            PS_FOLDER_INSTALL: install-dev
-            PS_FOLDER_ADMIN: admin-dev
-        ports:
-            - "8001:80"
-        volumes:
-            - ./:/var/www/html/:delegated
-            - vendor:/var/www/html/vendor
-            - var:/var/www/html/var
```
Quetzacoalt91, line 35: These volumes are useful for people running Docker on Mac.

docker-compose.yml
```diff
@@ -1,37 +1,31 @@
...
+      DB_NAME: prestashop
+      DB_SERVER: mysql
+      PS_DOMAIN: localhost:8001
+      PS_FOLDER_INSTALL: install-dev
+      PS_FOLDER_ADMIN: admin-dev
+    ports:
+      - "8001:80"
+    volumes:
+      - ./:/var/www/html/
+    restart: always
```
Quetzacoalt91: `restart:always` is actually unecessary.We don't aim production environments with this stack so we do not need to start it automatically.

matks (dismissed):

boubkerbribri:
Error when running command `docker-compose up`

image

PierreRambaud:
No more in WIP.
Can a developer check if everything is ok :)

.docker/docker_run_git.sh
```diff
@@ -0,0 +1,76 @@
...
+if [ ! -f ./config/settings.inc.php ]; then
+    if [ $PS_INSTALL_AUTO = 1 ]; then
+
+        echo "\n* Installing PrestaShop, this may take a while ...";
+
+        if [ $PS_ERASE_DB = 1 ]; then
+            echo "\n* Drop & recreate mysql database...";
+            if [ $DB_PASSWD = "" ]; then
+                echo "\n* Dropping existing database $DB_NAME..."
+                mysql -h $DB_SERVER -P $DB_PORT -u $DB_USER -p$DB_PASSWD -e "drop database if exists $DB_NAME;"
```
Progi1984: @PierreRambaud The drop command is identical in the `if` and the `else`. You can share it before the `if`.

Progi1984 (commented):
A small change from my side. But I ask a second review (@PrestaShop/prestashop-core-developers).

.docker/docker_run_git.sh
PierreRambaud: the password was empty and the `-p$DB_PASSWD` shouldn't be used

Progi1984 (approved):
Need a second review

.gitignore
```diff
@@ -97,6 +97,7 @@ themes/*/cache/*
 
 # Config
 
+config/defines_custom.inc.php
```
jolelievre, line 100: What's the relation with this Docker PR?

docker-compose.yml
```diff
@@ -1,37 +1,35 @@
...
+    build: .docker
+    environment:
+      PS_INSTALL_AUTO: 1
+      DB_PASSWD: prestashop
+      DB_NAME: prestashop
+      DB_SERVER: mysql
+      PS_DOMAIN: localhost:8001
+      PS_FOLDER_INSTALL: install-dev
+      PS_FOLDER_ADMIN: admin-dev
+    command: ["/tmp/wait-for-it.sh", "--timeout=60", "--strict", "mysql:3306", "--", "/tmp/docker_run_git.sh"]
```
jolelievre, line 29: So you need to run this command in both `docker-compose.yml` and in `.docker/Dockerfile` ?

jolelievre (commented):
Just two questions

.gitignore
PierreRambaud: this file is created while running the container or in some situation https://github.com/PrestaShop/PrestaShop/blob/develop/config/config.inc.php#L30
but didn't know why it wasn't be ignored :thinking:

docker-compose.yml
PierreRambaud: Dockerfile is not related to the docker-compose behavior.
In `.docker/Dockerfile` we are building the container architecture / behavior, whereas when you use the docker-compose, you are waiting for some service to be up

jolelievre (approved):
Thanks @PierreRambaud

jolelievre:
Tested and approved!

PR #15965, opened 2019-10-15
Implement CQRS on Notifications

### LINKED ISSUE #12870: Implement CQRS on CommonController::notificationsAction()
Following review on #12708 and #10131, CQRS should be implemented in CommonController::notificationsAction()

_Originally posted by @Quetzacoalt91 in https://github.com/PrestaShop/PrestaShop/pull/12708_

atomiix:
| Questions     | Answers
| Branch?       | develop
| Description?  | Use CQRS to get and update notifications
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #12870
| How to test?  | Add a new customer, see if you get a notification then open it and see if it disappears. This PR changes no behavior, it's a refactoring, notifications should work as before. Examples of notifications: successful deletion of a product, or successful update of an element.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15965)

### DIFF
--- a/src/Adapter/Notification/CommandHandler/UpdateEmployeeNotificationLastElementHandler.php
+++ b/src/Adapter/Notification/CommandHandler/UpdateEmployeeNotificationLastElementHandler.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Adapter\Notification\CommandHandler;
+
+use Notification;
+use PrestaShop\PrestaShop\Core\Domain\Notification\Command\UpdateEmployeeNotificationLastElementCommand;
+use PrestaShop\PrestaShop\Core\Domain\Notification\CommandHandler\UpdateEmployeeNotificationLastElementCommandHandlerInterface;
+
+/**
+ * Handle update employee's last notification element of a given type
+ *
+ * @internal
+ */
+final class UpdateEmployeeNotificationLastElementHandler implements UpdateEmployeeNotificationLastElementCommandHandlerInterface
+{
+    /**
+     * @param UpdateEmployeeNotificationLastElementCommand $command
+     */
+    public function handle(UpdateEmployeeNotificationLastElementCommand $command)
+    {
+        (new Notification())->updateEmployeeLastElement($command->getType());
+    }
+}
--- a/src/Adapter/Notification/QueryHandler/GetNotificationLastElementsHandler.php
+++ b/src/Adapter/Notification/QueryHandler/GetNotificationLastElementsHandler.php
@@ -1,0 +1,77 @@
+<?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Adapter\Notification\QueryHandler;
+
+use Notification;
+use PrestaShop\PrestaShop\Core\Domain\Notification\Query\GetNotificationLastElements;
+use PrestaShop\PrestaShop\Core\Domain\Notification\QueryHandler\GetNotificationLastElementsHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Notification\QueryResult\NotificationResult;
+use PrestaShop\PrestaShop\Core\Domain\Notification\QueryResult\NotificationsResult;
+use PrestaShop\PrestaShop\Core\Domain\Notification\QueryResult\NotificationsResults;
+
+/**
+ * Get employee last notification elements
+ *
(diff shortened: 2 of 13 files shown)


src/Adapter/Notification/QueryHandler/GetNotificationLastElementsHandler.php
```diff
@@ -0,0 +1,49 @@
...
+{
+
+    /**
+     * @param GetNotificationLastElements $query
+     *
+     * @return array
+     */
+    public function handle(GetNotificationLastElements $query)
+    {
+        return (new \Notification())->getLastElements();
```
sarjon: I'm afraid you should not blindly return whatever legacy gives you. :confused:

eternoendless: Queries must return a [Data Transfer Object](https://devdocs.prestashop.com/1.7/development/architecture/domain/data-transfer-objects/).

See also [Query and QueryHandler principles](https://devdocs.prestashop.com/1.7/development/architecture/domain/cqrs/#query-and-queryhandler-principles):
> 8. A `QueryHandler` SHOULD return data objects that make sense to the domain, and SHOULD NOT leak internal objects.

src/Core/Domain/Notification/Command/UpdateEmployeeNotificationLastElementCommand.php
```diff
@@ -0,0 +1,55 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Notification\Command;
+
+/**
+ * ACK employee notification last element
```
eternoendless:
```suggestion
 * Updates the last notification element from a given type seen by the employee
```

src/Core/Domain/Notification/Command/UpdateEmployeeNotificationLastElementCommand.php
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * @var string
+     */
+    private $type;
+
+    /**
+     * UpdateEmployeeNotificationLastElementCommand constructor.
+     * @param string $type
+     */
+    public function __construct($type)
```
eternoendless: `$type` looks like a perfect candidate for a [Value Object](https://devdocs.prestashop.com/1.7/development/architecture/domain/value_objects/). You need to ensure that the given type is valid.

Looking a the implementation, valid types seem to be:
- order
- customer_message
- customer

One of our rules state:
> A Command receives only primitive types on input (int, float, string, bool and array).

So you can type-hint this parameter as a `string`, then create the ValueObject in the constructor itself. See an example in [AddCurrencyCommand](https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Domain/Currency/Command/AddCurrencyCommand.php#L67).

src/Core/Domain/Notification/Command/UpdateEmployeeNotificationLastElementCommand.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        $this->type = $type;
+    }
+
+    /**
+     * @return string
+     */
+    public function getType()
```
eternoendless, line 57: This would return a ValueObject.

Don't hesitate to use strict types everywhere.

src/Core/Domain/Notification/Query/GetNotificationLastElements.php
```diff
@@ -0,0 +1,34 @@
...
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Notification\Query;
+
+/**
+ * This Query return the last Notifications elements
+ */
+class GetNotificationLastElements
+{
```
eternoendless, line 36: I know that the current implementation works with the current employee only, but could you please add a requirement for the employeeId? This way a future better implementation won't be dependent on the context.

A context-dependent query is harder to put in cache.

eternoendless (dismissed):

src/Adapter/Notification/CommandHandler/UpdateEmployeeNotificationLastElementHandler.php
```diff
@@ -0,0 +1,47 @@
...
+ * @internal
+ */
+final class UpdateEmployeeNotificationLastElementHandler implements UpdateEmployeeNotificationLastElementCommandHandlerInterface
+{
+    /**
+     * @param UpdateEmployeeNotificationLastElementCommand $command
+     */
+    public function handle(UpdateEmployeeNotificationLastElementCommand $command)
+    {
+        (new \Notification())->updateEmployeeLastElement($command->getType());
```
Progi1984: Please use a `use Notification` in declarations.

src/Adapter/Notification/QueryHandler/GetNotificationLastElementsHandler.php
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Get employee last notification elements
+ *
+ * @internal
+ */
+final class GetNotificationLastElementsHandler implements GetNotificationLastElementsHandlerInterface
+{
+    /**
+     * @param GetNotificationLastElements $query
```
matks, line 44: You can put:
```
    /**
     * {@inheritdoc}
     */
```

src/Core/Domain/Notification/QueryResult/NotificationsResults.php
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Notification\QueryResult;
+
+class NotificationsResults
```
matks, line 32: When I saw NotificationsResults, NotificationResults and NotificationResult I was a bit confused 😅 

Do you think we can clear a bit this confusion ? People will understand after a while, but we can help them understand even faster. For example we could add a phpDoc here saying: "NotificationsResults is a collection of NotificationsResult". Simple, and gives immediately the answer.

src/Core/Domain/Notification/ValueObject/Type.php
```diff
@@ -0,0 +1,80 @@
...
+     * @return string
+     */
+    public function getValue(): string
+    {
+        return $this->type;
+    }
+
+    private function assertIsValidType(string $type)
+    {
+        $allowedTypes = [self::ORDER, self::CUSTOMER, self::CUSTOMER_MESSAGE];
```
matks, line 74: 👍

src/Core/Domain/Notification/ValueObject/Type.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * @return string
+     */
+    public function getValue(): string
+    {
+        return $this->type;
+    }
+
+    private function assertIsValidType(string $type)
```
matks, line 72: Even though the method signature says it all, please add the phpDoc above. I know this is redundant but this actually helps some developer IDEs to perform autocomplete / hints. We must address a large scope of developer and developer tools, and some of them rely only on phpDoc 😉

src/PrestaShopBundle/Controller/Admin/CommonController.php
```diff
@@ -56,20 +61,52 @@ class CommonController extends FrameworkBundleAdminController
...
     {
-        // TODO: Use CQRS
-        return new JsonResponse((new \Notification())->getLastElements());
+        $employeeId = Context::getContext()->employee->id;
+        /** @var NotificationsResults $elements */
+        $elements = $this->getQueryBus()->handle(new GetNotificationLastElements($employeeId));
+
+        $response = [];
+
+        foreach ($elements->getNotificationsResults() as $element) {
```
matks: I'm wondering 🤔 ... could we make this simpler ?

Some ideas:
- implements a `toArray()` function into the DTO NotificationsResult, NotificationsResults and NotificationResult
-  what happens if we use Symfony Serializer here ? I think he might be able to perform this operation without the need for us to write it (see below)

I'm not sure this is a good idea nor it's working 😅 please tell me what you think

Serializer usage:
```
// should work, not 100% sure, might need a few fix
$serializer = $this->get('serializer');
$notifications[] = $serializer->serialize(notification);
```

mickaelandrieu: this should not be done in the controller at all imho how about create an object with this responsibility ?

matks (approved):

matks:
Thank you @atomiix

PR #15974, opened 2019-10-16
Fix wrong verification of number of languages

fest73:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix condition check count languages in getAlternativeLangsUrl. Condition if ($alternative_links < 2) incorrect
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Here you can see that the array will simply be converted to a number to perform a comparison, which will be incorrect. leave enabled one language, open front office and open code in section head you will can see one tag: link rel="alternate" href="http://site.ru/" hreflang="en-us"

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15974)

### DIFF
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -1973,7 +1973,7 @@ class FrontControllerCore extends Controller
         $alternativeLangs = array();
         $languages = Language::getLanguages(true, $this->context->shop->id);
 
-        if ($languages < 2) {
+        if (count($languages) < 2) {
             // No need to display alternative lang if there is only one enabled
             return $alternativeLangs;
         }


Progi1984 (approved):

Progi1984:
Need a second approve.

jolelievre:
@dariusakafest Thanks for the fix, although it seems completely legit Do you have any test scenarii where this caused a bug please? So that the QA team can perform tests with/without the modification?

fest73:
@jolelievre leave enabled one language, open front office and open code in section head you will can see one tag:
link rel="alternate" href="http://site.ru/" hreflang="en-us"

jolelievre:
@dariusakafest Ok I get it better now, can you please detail this on the `How to test` section?

jolelievre (approved):

matks:
Thank you @dariusakafest

Uebix:
PrestaShop team, this bug fix could be inserted already in version 1.7.6.2 (we are now in 1.7.6.5), why was 1.7.7.0 milestone chosen?
It is very important for SEO.
@dariusakafest thanks for the fix.

jolelievre:
Hi @Uebix,

short answer: the reason why it's fixed in 1.7.7.x and not in 1.7.6.x is because it was not a regression from 1.7.6

A bit more details about the release process:
- when we develop a version we integrate new features and some bug fixes, those two are selected by priority among all the needed features and bug fixes (we can't fix all in a single version), and this defines the "scope" of a version
- once we arrive at then of a version we enter in a "feature freeze" period right before building the beta this is a time when we refuse any additional content to the version, only bugs for new features or regressions are accepted (a regression is when the behaviour compared to the previous version was broken in an undesired way)
- then we have build release, RC release and final release, but the principal of closing the scope is still there

The reason why we don't include all bug fixes in a version which scope has been closed is because if we did we would never be able to release it, the scope would always get bigger and bigger, and each time we release a new version (even a patch one) we need 1 to 2 weeks of QA validation before actually releasing it. Which is why a bug that is not a regression will be fixed for the next version.

So it might sound harsh and frustrating, I can get it, but it's the only we can move forward without constantly increase a version's scope, and all we would be doing would be releasing only patch versions without improving the rest of the project. (By the way rebuilding and migrating a feature is sometimes the most efficient way to fix the related bugs of a legacy, weirdly made and hard to maintain code.)

I hope this answer satisfies you and was clear enough, if not I believe we have a coming article that should be published on the build site that explains all this process with more details.

Uebix:
Hi @jolelievre , thanks for explaining about the release process and i know it.
The fix was commit at 16 Oct 2019 and the scope of 1.7.6.3, 1.7.6.4 and 1.7.6.5 was open. Indeed, they had not yet been opened.
This bug was very small and it didn't need testing because it was evident that it was a trivial mistake looking at the code.
I understand that for larger commits you have to be very careful (I'm a platform developer and integration specialist), but this was not the case, it could have been published without enlarging the scope much earlier.
I can assure you that the delay in publishing these small bug fixes is truly frustrating for the community and for the prestige of PrestaShop in the world.

Thanks anyway for all the work you do, I understand that it is not easy to manage a large cms like PrestaShop.

PR #15978, opened 2019-10-16
Functional Tests - Add BO tests for  taxes

boubkerbribri:
| Questions     | Answers
| Branch?       | develop
| Description?  | Adding Functional tests for BO taxes
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | URL_FO=SHOP_URL TEST_PATH="functional/BO/international/taxes/*" npm run specific-test

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15978)


tests/puppeteer/campaigns/data/faker/tax.js
```diff
@@ -0,0 +1,10 @@
+const faker = require('faker');
+
+module.exports = class Tax {
+  constructor(taxToCreate = {}) {
+    this.rate = taxToCreate.rate || faker.random.number({min: 1, max: 100}).toString();
```
SimonGrn: Maybe limit this to 40 :thinking: ? Having a 100% tax is kinda harsh :joy: 
```suggestion
    this.rate = taxToCreate.rate || faker.random.number({min: 1, max: 40}).toString();
```

SimonGrn (dismissed):
Thanks @boubkerbribri

tests/puppeteer/campaigns/data/faker/tax.js
boubkerbribri: done, thanks

tests/puppeteer/campaigns/data/faker/tax.js
```diff
@@ -0,0 +1,10 @@
+const faker = require('faker');
+
+module.exports = class Tax {
+  constructor(taxToCreate = {}) {
+    this.rate = taxToCreate.rate || faker.random.number({min: 1, max: 40}).toString();
+    this.name = taxToCreate.name || `TVA test ${this.rate}%`;
+    this.frName = taxToCreate.frName || this.name;
+    this.enabled = taxToCreate.enabled || 'Yes';
```
PierreRambaud: Should be a boolean :)

boubkerbribri: I have added a wip label since I changed this in customers, I should changed in this too

boubkerbribri: done

tests/puppeteer/campaigns/functional/BO/customers/customers/01_filterAndQuickEditCustomers.js
```diff
@@ -186,7 +186,7 @@ describe('Filter And Quick Edit Customers', async () => {
       await this.pageObjects.customersPage.filterCustomers(
         'select',
         'active',
-        DefaultAccount.enabled ? 'Yes' : 'NO',
+        DefaultAccount.enabled ? 'Yes' : 'No',
```
PierreRambaud: Why not using the `Yes` or `No`in the `filterCustomers` function?
So when you pass a boolean it consider it as `Yes/No`

boubkerbribri: It's just another way to do it, No preferences.
Do you think its easier to maintain when 'using the Yes or No in the filterCustomer' ?

PierreRambaud: depend on what the filterCustomer do.
If every time you think a boolean is passed to the function it must be convert to a string, go with it.
If it's really rare to have a real boolean, you can add an extra parameter late to don't convert the boolean into `yes|no`
You're also able to create a new method.
It's just to remove the ternary everywhere, it's considered as duplicate logic.

boubkerbribri: Yes, it's always a yes / no, and I was thinking, if the value of options in select changed (for YES/NO for example), it will be easier to changed it in filterCustomers than all tests
Thanks for suggestion :+1:

boubkerbribri: changes done, what do you think about it ?

PierreRambaud (approved):

Issue #10583, opened 2018-09-21
Migrate "Orders > Orders > Add new" page


matks:
Type: page
Part of Symfony migration project

## Status

Migration block per block

Todo:
- [x] Migrate Add action https://github.com/PrestaShop/PrestaShop/pull/12645
- [x] Migrate block "customer"
- [x] Migrate block "products" (a.k.a cart) https://github.com/PrestaShop/PrestaShop/pull/15814
- [x] Migrate block "cart rules" https://github.com/PrestaShop/PrestaShop/pull/15995
- [x] Migrate block "adresses"
- [x] Migrate block "shipping" https://github.com/PrestaShop/PrestaShop/pull/15899
- [x] Migrate "summary" block [WIP] https://github.com/PrestaShop/PrestaShop/pull/16207
- [x] Add "add an order" button to listing
- [x] Add a modale for "Details" of a customer #17090
- [x] Add a modale for "Add new voucher" #17090

Centralized Mail sendings: see https://github.com/PrestaShop/PrestaShop/pull/13554#discussion_r282534335

### Controller / template
- [ ] All controller actions are migrated and protected with `@AdminSecurity` annotations when eligible and `@DemoRestricted` annotations when eligible
- [ ] All links targeting this page are updated: either using _legacy_link or updating the code
- [ ] The new SF routes have `_legacy_link` and `_legacy_parameters` provided and explained
- [ ] Controller catches Domain Exception and translate them nicely
- [ ] Twig templates must contain relevant blocks to allow partial extension by modules
- [ ] JS being used on the page can be used as enabled extensions and use a PageMap for selectors
##### Conventions
- [ ] The new controller actions comply with our naming convention
- [ ] The new SF routes names and paths comply with our naming convention
- [ ] The new Twig template names comply actions with our naming convention
- [ ] Created files (controller, JS assets, CSS assets and twig views) must be in the right folder following the convention
- [ ] CSS being added must follow theme conventions

#### Core
- [ ] Commands and Queries only use native types but internally use VOs
- [ ] All Handlers must have an Interface
##### Conventions
- [ ] The created SF service names comply with our naming convention
- [ ] The created files must follow the folder convention (Domain, Core, Adapter...)

#### Polishing
- [ ] The legacy controller and templates files are gone
- [ ] If PR includes a new concept/mechanism, documentation must be written
- [ ] JS and CSS assets must be compiled and up-to-date
- [ ] License headers must be up-to-date
[...]

Issue #15977, opened 2019-10-16
Case style conventions


sarjon:
#### Is your feature request related to a problem?

Recently we've been having quite a few discussions on multiple PRs regarding case style (camelCase, snake_case or kebab-case) used in code, for example:
1. What style do we use for HTML id attribute?
2. What style do we use for JSON we send to API?

#### Describe the solution you'd like

To avoid wasting time on discussing and fixing these kind of issues, I suggest updating dev docs with whatever convention we choose for PS code base.

sarjon:
@matks wdyt?

matks:
Yes, it's a good thing to discuss and decide it now

ping @PierreRambaud

zuk3975:
Not sure if its the right place, but we could also decide yaml files ident.

matks:
OK, here's my suggestions:

YAML files indent: 2 spaces
JSON payloads to API and JSON responses: snake_case
HTML class and ids: kebab-case

No logic behind that, it just seems to be the most popular standards to me

matks:
ping @PrestaShop/prestashop-core-developers

sarjon:
> HTML class and ids: kebab-case

By default, Symfony Form component renders form elements with snake_case ids.

matks:
ping @jolelievre @PierreRambaud

PierreRambaud:
> OK, here's my suggestions:
> 
> YAML files indent: 2 spaces
> JSON payloads to API and JSON responses: snake_case
> HTML class and ids: kebab-case
> 
> No logic behind that, it just seems to be the most popular standards to me

Same as @matks, this is the most popular.
Plus, 2 spaces for HTML, JS & CSS (as airbnb guide style, one of the most used in VueJs community)

Progi1984:
Note :notebook: : 
* [Airbnb JavaScript Style Guide](https://github.com/airbnb/javascript)
* [Airbnb CSS / Sass Styleguide](https://github.com/airbnb/css) 

Not found for HTML

Progi1984:
For XML files, indent ? 
Actually, it is tabs ([Example](https://github.com/PrestaShop/PrestaShop/pull/13636#discussion_r336395861)).

zuk3975:
About the JSON API snake_case.
Currently we mostly use it for private api requests. For this reason I suggest using camelCase.
[...]

PR #15982, opened 2019-10-16
Migrate product component for Orders page

RaimondasSapola:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migration of order create page cart component search products and add to cart
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10583
| How to test?  | Create order cart block should be migrated and should have the same functionality as legacy

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15982)


src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -474,4 +524,16 @@ private function handleChangeOrderStatusException(ChangeOrderStatusException $e)
...
         }
     }
+
+    /**
+     * @param Request $request
+     * @return int|null
+     */
+    private function addCustomizationFields(Request $request): ?int
+    {
+        $customizationFields = $request->request->get(self::PRODUCT_CUSTOMIZATION_FIELDS_REQUEST_KEY);
```
RaimondasSapola: Hey @matks i need your guidance :) For add product to cart component in BO order creation i have to add product customization fields via ajax. They can be added by user in product page and can be file or text type. Im thinking of creating a separate service for validating and resolving data of these fields and use command to actually add them to database with legacy functions. Other idea would be to use form with collection type fields. Although form imo would look cleaner, i am not sure if it is going to work. Maybe you have some suggestions or insights about that?

admin-dev/themes/new-theme/js/pages/order/create/cart-editor.js
```diff
@@ -76,4 +74,57 @@ export default class CartEditor {
...
+   * @param {Number} cartId
+   * @param {FormData} product
+   */
+  addProduct(cartId, product) {
+    $.ajax(this.router.generate('admin_carts_add_product', {cartId}), {
+      method: 'POST',
+      data: product,
+      processData: false,
+      contentType: false,
+      cache: false,
```
sarjon: POST methods are not cached, you shouldn't need this option.

admin-dev/themes/new-theme/js/pages/order/create/cart-rule-manager.js
```diff
@@ -38,8 +39,8 @@ export default class CartRuleManager {
   constructor() {
     this.router = new Router();
     this.$searchInput = $(createOrderMap.cartRuleSearchInput);
-    this.$searchResultBox = $(createOrderMap.cartRulesSearchResultBox);
-    this.cartRulesRenderer = new CartRulesRenderer();
+    this.renderer = new CartRulesRenderer();
```
sarjon: Rename to `cartRulesRenderer`, it will be easier to understand when reading code below.

admin-dev/themes/new-theme/js/pages/order/create/cart-rule-manager.js
```diff
@@ -86,12 +87,13 @@ export default class CartRuleManager {
...
   _addCartRuleToCart(cartRuleId, cartId) {
-    $.post(this.router.generate('admin_carts_add_rule', {cartId}), {
-      cart_rule_id: cartRuleId,
-    }).then((cartInfo) => {
-      this.cartRulesRenderer.render(cartInfo.cartRules, cartInfo.products.length === 0);
-    }).catch((e) => {
-      this._displayErrorMessage(e.responseJSON.message);
+    this.cartEditor.addCartRuleToCart(cartRuleId, cartId);
+
+    EventEmitter.on(eventMap.cartRuleAdded, (cartInfo) => {
```
sarjon, line 84: You attach event every time cart rule is being added? Wouldn't it be enough to add event listener once?

admin-dev/themes/new-theme/js/pages/order/create/cart-rule-manager.js
```diff
@@ -86,12 +87,13 @@ export default class CartRuleManager {
...
+      this.renderer.renderCartRulesBlock(cartInfo.cartRules, cartInfo.products.length === 0);
+    });
+    EventEmitter.on(eventMap.cartRuleFailedToAdd, (message) => {
```
sarjon, line 95: Same here.

admin-dev/themes/new-theme/js/pages/order/create/cart-rule-manager.js
jolelievre, line 84: Shouldn't you subscribe the event before calling `addCartRuleToCart` ?

jolelievre: And if you don't unsusbcribe after result then the listener will be called multiple times after successive calls of `_addCartRuleToCart` no?

admin-dev/themes/new-theme/js/pages/order/create/create-order-page.js
```diff
@@ -100,14 +99,11 @@ export default class CreateOrderPage {
    *
    * @private
    */
-  _handleCustomerChooseForOrderCreation() {
+  _handleCustomerSelect() {
     this.$container.on('click', createOrderPageMap.chooseCustomerBtn, (event) => {
-      const customerId = this.customerManager.onCustomerChooseForOrderCreation(event);
-      this.data.customer_id = customerId;
+      const customerId = this.customerManager.onCustomerSelect(event);
```
jolelievre: I think `onCustomerSelect` is not useful, you could call `_selectCustomer` directly but maybe renaming it `selectCustomerByEvent`

admin-dev/themes/new-theme/js/pages/order/create/product-manager.js
```diff
@@ -0,0 +1,245 @@
...
+    this.selectedCombinationId = null;
+
+    this.renderer = new ProductRenderer();
+    this.router = new Router();
+    this.cartEditor = new CartEditor();
+
+    this._initEvents();
+
+    return {
+      onAddProductToCart: (cartId) => {
```
jolelievre: Maybe you could remove the `on` from these methods?

admin-dev/themes/new-theme/js/pages/order/create/create-order-page.js
```diff
@@ -301,7 +277,7 @@ export default class CreateOrderPage {
    * @private
    */
   _persistCartInfoData(cartInfo) {
-    this.data.cart_id = cartInfo.cart.id;
+    this.data.cartId = cartInfo.cart.id;
```
jolelievre: Mmm, I thought you removed the `this.data` object in favor of `this.cartId` ?

admin-dev/themes/new-theme/js/pages/order/create/customer-manager.js
```diff
@@ -74,130 +78,48 @@ export default class CustomerManager {
...
+   *@todo: fix showing not found customers and rerender after change customer
    * @private
    */
-  _doSearch() {
-    const name = this.$searchInput.val();
+  _search() {
+    const searchPhrase = this.$searchInput.val();
 
-    if (name.length < 4) {
+    if (searchPhrase.length < 3) {
```
jolelievre, line 149: That's the feature you were talking in the daily right?

I'm thinking about an improvement you should keep a reference on the query call so that you can stop the previous one each time a new letter is typed (this would avoid too many parallel calls) Using the `.abort()` function This way you could even get rid of this 2 characters limit

And the best would be to add a small delay (let's say 300-500ms) so that the request only starts after you stopped typing

admin-dev/themes/new-theme/js/pages/order/create/product-manager.js
```diff
@@ -0,0 +1,245 @@
...
+   * Searches for product
+   *
+   * @private
+   */
+  _search(event) {
+    const minSearchPhraseLength = 3;
+    const $productSearchInput = $(event.currentTarget);
+    const name = $productSearchInput.val();
+
+    if (name.length < minSearchPhraseLength) {
```
jolelievre: Same suggestion as `CustomerManager`, store the request to abort it and add a little delay

admin-dev/themes/new-theme/js/pages/order/create/product-manager.js
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  /**
+   * Adds selected product to current cart
+   *
+   * @private
+   */
+  _addProductToCart(cartId) {
+    this.cartEditor.addProduct(cartId, this._getProductData(cartId));
+    EventEmitter.on(eventMap.productAddedToCart, (cartInfo) => {
```
jolelievre: Again shouldn't you subscribe the event before calling `addProduct` ? And what about unlistening?

src/Adapter/Cart/CommandHandler/AddCustomizationFieldsHandler.php
```diff
@@ -0,0 +1,79 @@
...
+                        $productId,
+                        $customizationId,
+                        Product::CUSTOMIZE_TEXTFIELD,
+                        $customizations[$customizationId]
+                    );
+                    continue;
+                }
+
+                //@todo: file validation
+                $cart->addPictureToProduct(
```
jolelievre: I think `addTextFieldToProduct` would be more appropriate here
By the way I don't understand the difference between `addPictureToProduct /addTextFieldToProduct` I guess it was to clarify the use but then the type parameter is useless and should be set by force.. CUSTOMIZE_TEXTFIELD for addTextFieldToProduct and CUSTOMIZE_FILE for addPictureToProduct

src/Adapter/Product/QueryHandler/SearchProductsHandler.php
```diff
@@ -0,0 +1,196 @@
...
+     * @var string
+     */
+    private $locale;
+
+    /**
+     * @param LegacyContext $context
+     * @param RepositoryInterface $localeRepository
+     * @param string $locale
+     */
+    public function __construct(LegacyContext $context, RepositoryInterface $localeRepository, string $locale)
```
jolelievre: Try to avoid inserting the context when you can, it's preferable to insert a langId via the service definition (even though you use `@=service(...)` in the definition to get the same value
Besides you can inject the Locale object directly rather than the repo plus the locale string A new service has been added in this spirit:

https://github.com/Progi1984/PrestaShop/blob/d7b0b8e21af02a4059466b879f24fc863acf875e/src/PrestaShopBundle/Resources/config/services/core/cldr.yml#L24

Especially since you make repeated calls in a loop

src/Core/Domain/Cart/Command/AddCustomizationFieldsCommand.php
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * @return ProductId
+     */
+    public function getProductId(): ProductId
+    {
+        return $this->productId;
+    }
+
+    /**
+     * @return array
```
jolelievre, line 81: Can you define what this array is supposed to contain, Command objects represent the contract of our API and thus needs to be documented with details so we now what value needs to be inserted in them (ex: array of cusotmization ids? or here I believe it's customized texts indexed by customizationID?)

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
```diff
@@ -0,0 +1,349 @@
...
+      return;
+    }
+
+    const $customFieldsContainer = $(createOrderMap.productCustomFieldsContainer);
+
+    const $fileInputTemplate = $($(createOrderMap.productCustomFileTemplate).html());
+    const $textInputTemplate = $($(createOrderMap.productCustomTextTemplate).html());
+
+    const templateTypeMap = {
+      0: $fileInputTemplate,
```
jolelievre: I guess these values match ProductCustomizationField constants? I'm afraid the relation is not that clear at first, maybe we should add const values in a javascript class or use string const instead of integers?

src/PrestaShopBundle/Controller/Admin/ProductController.php
```diff
@@ -1253,4 +1255,28 @@ public function renderFieldAction($productId, $step, $fieldName)
...
+     *
+     * @return Response
+     */
+    public function searchProductsAction(Request $request): Response
+    {
+        try {
+            $searchPhrase = $request->query->get('search_phrase');
+
+            /** @var FoundProduct[] $foundProducts */
+            $foundProducts = $this->getQueryBus()->handle(new SearchProducts($searchPhrase, 10));
```
jolelievre, line 1270: So for now you limit to 10 results and the pagination will be added in another issue?

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
```diff
@@ -255,7 +259,56 @@ public function removeCartRuleFromCartAction(int $cartId, int $cartRuleId)
...
+                ['message' => $this->getErrorMessageForException($e, $this->getErrorMessages($e))],
+                Response::HTTP_INTERNAL_SERVER_ERROR
+            );
+        }
+    }
+
+    /**
+     * Adds product to cart
+     *
+     * @AdminSecurity("is_granted('update', request.get('_legacy_controller'))")
```
jolelievre, line 275: Not in the scope of this PR of course, but for API controller actions I think we should have a dedicated Annotation, because the default behaviour of this one is to redirect to the dashboard Returning a json response with 503 code and a quick message would allow us to create those ajax calls more easily

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
```diff
@@ -283,4 +336,34 @@ private function getErrorMessages(Exception $e)
...
+     *
+     * @return UpdateProductQuantityInCartCommand
+     */
+    private function getAddProductToCartCommand(Request $request, int $cartId): UpdateProductQuantityInCartCommand
+    {
+        $productId = $request->request->getInt('product_id');
+        $quantity = $request->request->getInt('quantity');
+        $combinationId = $request->request->getInt('combination_id');
+
+        if ($customizations = $request->request->get('customization')) {
```
jolelievre: The request parameter should `customizations` if it's an array

src/Adapter/Cart/CommandHandler/AddCustomizationFieldsHandler.php
```diff
@@ -0,0 +1,79 @@
...
+        $cart = $this->getCart($command->getCartId());
+        $product = new Product($productId);
+
+        $customizationFields = $product->getCustomizationFieldIds();
+        $customizations = $command->getCustomizations();
+
+        foreach ($customizationFields as $customizationField) {
+            $customizationId = (int) $customizationField['id_customization_field'];
+            //@todo validation
+            if (isset($customizations[$customizationId])) {
```
jolelievre:
```suggestion
            if (!isset($customizations[$customizationId])) {
                continue;
            }
```

src/Adapter/Cart/CommandHandler/AddCustomizationFieldsHandler.php
```diff
@@ -0,0 +1,79 @@
...
+                if ($customizationField['type'] == Product::CUSTOMIZE_TEXTFIELD) {
```
jolelievre: You you should use `if ($customizationField['type'] == Product::CUSTOMIZE_TEXTFIELD) { ... } else { ... }` rather than a `continue`, it's not very intuitive to read at first

src/Adapter/Cart/CommandHandler/AddCustomizationFieldsHandler.php
jolelievre: Ok I got it 😅 You manage the two cases between file and text customization Then as I said in the previous comment both methods do the same thing so in the second case you should use `CUSTOMIZE_FILE` as an argument (you already managed `CUSTOMIZE_TEXTFIELD` in the previous statement

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
jolelievre: Also you shouldn't execute a command in this method called `getAddProductToCartCommand`, the method should only create the object an not perform an action at the same time (which is not explicit in the method name)

You could rename the method into something like `getAddProductToCartCommandAndCreateOrUpdateCustomizations` but it would be clearer to have a dedicated method for this `updateProductCostomizations(Request $request, int $cartId)`

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/carts.yml
```diff
@@ -68,20 +68,44 @@ admin_carts_set_free_shipping:
   options:
     expose: true
 
-admin_carts_add_rule:
-  path: /{cartId}/rules/add
+admin_carts_add_cart_rule:
+  path: /{cartId}/cart-rules
```
jolelievre, line 74: I think you should keep the `add` prefix, so it is clear *one* cart rule is added, and not we update *all* the cart rules The `cart-rules` modification is better though! so it would give `/{cartId}/cart-rules/add` which matches the format of the delete urls

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/carts.yml
```diff
@@ -68,20 +68,44 @@ admin_carts_set_free_shipping:
...
     expose: true
+
+admin_carts_add_product:
+  path: /{cartId}/products
+  methods: [POST]
+  defaults:
+    _controller: PrestaShopBundle:Admin/Sell/Order/Cart:addProduct
+    _legacy_controller: AdminCarts
+  options:
+    expose: true
```
jolelievre, line 98: Just a question for my personal knowledge ^^, this `exposed` option is used for fos_js_routing?

src/PrestaShopBundle/Resources/config/services/adapter/order.yml
```diff
@@ -133,3 +133,13 @@ services:
     tags:
       - name: tactician.handler
         command: 'PrestaShop\PrestaShop\Core\Domain\Order\Query\GetOrderPreview'
+
+  prestashop.adapter.order.query_handler.search_products_for_order_creation:
+    class: PrestaShop\PrestaShop\Adapter\Order\QueryHandler\SearchProductsForOrderCreationHandler
+    arguments:
+      - '@prestashop.adapter.legacy.context'
+      - '@prestashop.core.localization.locale.repository'
```
jolelievre: You should use `prestashop.core.localization.locale.context_locale` here

src/PrestaShopBundle/Resources/config/services/adapter/product.yml
```diff
@@ -40,3 +40,13 @@ services:
         public: true
         tags:
             - { name: tactician.handler, command: PrestaShop\PrestaShop\Core\Domain\Product\Command\AssignProductToCategoryCommand }
+
+    prestashop.adapter.product.query_handler.search_products:
+        class: PrestaShop\PrestaShop\Adapter\Product\QueryHandler\SearchProductsHandler
+        arguments:
+            - '@prestashop.adapter.legacy.context'
+            - '@prestashop.core.localization.locale.repository'
```
jolelievre: You should use `prestashop.core.localization.locale.context_locale` here

src/PrestaShopBundle/Resources/config/services/bundle/serializer.yml
```diff
@@ -0,0 +1,17 @@
...
+    public: true
+
+  # Snake case normalizer definition
+  prestashop.bundle.snake_case_normalizer.object:
+    class: Symfony\Component\Serializer\Normalizer\ObjectNormalizer
+    arguments:
+      - ~
+      - '@serializer.name_converter.camel_case_to_snake_case'
+
+  # Snake case serializer definition
```
jolelievre, line 12: Nice!

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
```diff
@@ -154,3 +166,43 @@
     </div>
   </div>
 </div>
+
+<script id="js-product-custom-text-template" type="text/html">
```
jolelievre: I didn't know this method but it's really interesting!

admin-dev/themes/new-theme/js/pages/order/create/create-order-page.js
zuk3975: yea i did, this method is just a leftover, i didn't want to modify it at all, as its for summary part. It has a todo on top, so don't mind it 😄

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/carts.yml
zuk3975: yes :)

admin-dev/themes/new-theme/js/pages/order/create/product-manager.js
zuk3975: I guess you are right.

admin-dev/themes/new-theme/js/pages/order/create/product-manager.js
zuk3975: The `addProduct` method emits a `productAddedToCart` event after the ajax request ends and then I do `something` on that event. Not sure what you mean by `unlistening` ? 😄

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
zuk3975: sure

src/PrestaShopBundle/Controller/Admin/ProductController.php
zuk3975: I just added 10 because it was no limit at all in legacy 😄 I'm not sure this needs a pagination as it could only show like 10 most relative results because its a search. Anyway It should be discussed with product team i guess.
(the pagination i was talking at the meeting was about carts/orders lists which aren't searchable)

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
zuk3975: > The request parameter should `customizations` if it's an array

It comes as customization[1] customization[2] because of html input name. You think its ok to name the single input in plural form like `<input name=customizations[1]>` ?

> Also you shouldn't execute a command in this method called getAddProductToCartCommand

yes I agree, i should have left a todo here, It was just a quick draft, as cusomizations should be done in even separate pr maybe.

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/carts.yml
zuk3975: yes

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/carts.yml
zuk3975: idk, this way seemed to suit api standards better 😄 I was suggested by @sarjon, so wdyt ? 😄

src/PrestaShopBundle/Resources/config/services/bundle/serializer.yml
zuk3975: Not sure if its needed though 😄 more here https://github.com/PrestaShop/PrestaShop/issues/15977#issuecomment-545364613

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
```diff
@@ -154,3 +166,43 @@
...
+  <div class="row mt-3">
+    <div class="col">
+      <label for="" class="js-product-custom-input-label"></label>
+      <input name="" type="text" class="form-control js-product-custom-input">
+    </div>
+  </div>
+</script>
+
+<script id="js-product-custom-file-template" type="text/html">
```
PierreRambaud:
```suggestion
<script id="js-product-custom-file-template" type="text/template">
```

admin-dev/themes/new-theme/js/pages/order/create/create-order-page.js
jolelievre: Alright, I trust you 😉

admin-dev/themes/new-theme/js/pages/order/create/product-manager.js
jolelievre: Well events are like listeners, each time you call the `on` method you add a callback to the list of listeners for this method. But in this case each time `_addProductToCart` is called you add a new listener, so on second call you will have two listeners, on the third three listeners, and so on...
So the callback methods will be called multiple times each time you add a product. To avoid this you have different solutions:
- either you init your listener in a separate method and call it only *once*
- remove the listener after it's been called using removeListener but you need the reference on the callback so you can't use an arrow function, you will have to use a named class function
- in this case you are lucky because you only have one listener (no failure event) so you can use `EventEmitter.once` instead of `EventEmitter.on` It works the same way except it automatically removes the listener once it's called

The `once` can do the job here because you only have one event to listen and you need to be sure that it is *always* fired. But in other cases like in `_addCartRuleToCart` you listen to two events so it's not appropriate You have to init both events beforehand and only once Also be careful it's usually a good practice to clear the events when you stop using the objects because as long as they are present the object is kept in memory and the garbage collector can't remove it

Finally you should always init your listener before you call the method that emits them, if you're dealing with asynchronous methods you might be good But if the method is executed instantly the event will emitted before you register it and you won't be able to catch it any more

src/PrestaShopBundle/Controller/Admin/ProductController.php
jolelievre: Ok, indeed it should be discussed with @MatShir

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
jolelievre: Yes I think `<input name=customizations[1]>` is perfectly fine you see straight away that it contains an array

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/carts.yml
jolelievre: The thing is I wonder if we have a defined standard on these API calls ^^ This should be decided and documented I guess so that all our APIs follow the same convention

src/PrestaShopBundle/Resources/config/services/bundle/serializer.yml
jolelievre: Well I had a similar problem in one of my current PR actually.. We need to define a convention do we return json objects as snake case or camel case?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
```diff
@@ -40,80 +40,92 @@
...
+            </div>
           </div>
-          <div class="col-4">
-            <select class="form-control"></select>
+
+          {#@todo: max height and verical scroll#}
+          <div class="row mt-3 d-none" id="js-customization-container">
+            <div class="col-3">
+              <span class="float-right">
+                {{ 'Customization'|trans({}, 'Admin.Global') }}
```
LouiseBonnard:
```suggestion
                {{ 'Customization'|trans({}, 'Admin.Catalog.Feature') }}
```

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/carts.yml
sarjon: > I wonder if we have a defined standard on these API calls

Not that I aware of.

> This should be decided and documented I guess so that all our APIs follow the same convention

You are right, but for now I think we are trying to follow (more or less) standard Rest API naming, see https://restfulapi.net/resource-naming/

src/PrestaShopBundle/Resources/config/services/adapter/order.yml
zuk3975: maybe you know what is going on with it? I mean that Locale implements LocaleInterface which has no methods.

src/PrestaShopBundle/Resources/config/services/bundle/serializer.yml
zuk3975: I vote for camelCase :1st_place_medal:

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
matks: Not done yet 😄

src/Core/Domain/Cart/Command/AddCustomizationFieldsCommand.php
matks: 🆙

src/PrestaShopBundle/Resources/config/services/adapter/order.yml
matks: I'm sorry, I did not get the question

admin-dev/themes/new-theme/js/pages/order/create/cart-rule-manager.js
matks: not done yet

zuk3975: hmm, the `_onAddCartRuleToCart()` is the listener, it is initialized once to lisen for `'cartRuleAdded' event`. So I think this one should be resolved.

src/Core/Domain/Cart/Command/AddCustomizationFieldsCommand.php
zuk3975: I've already covered it in docblock above the property

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
matks: I think I answered that 🤔 maybe in gitter ?

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
matks: I'm not sure to understand 😅

matks (approved):

sarjon:
Thanks @zuk3975 

As discussed with @matks QA will happen on whole page after migrating remaining components.

PR #16014, opened 2019-10-18
Fix html being displayed in download virtual product text mail

### LINKED ISSUE #10915: Display issue in virtual product email
**Describe the bug**
Some html tags are displayed in email when you order virutal product

**To Reproduce**
1. Go to advanced parameters > Email and choose Send email in text format
2. Make sure you receive email (test it)
3. Go to Catalog > Product
4. Create a new virual product with a file
4. Order this product
5. Put the order in "payment accepted" status
6. See the txt email sent 

**Screenshots**

46542475-cb99dd00-c8be-11e8-932e-13fb208dacc0

**Additionnal information**
PrestaShop version: develop
PHP version: N/A

atomiix:
| Questions     | Answers
| Branch?       | develop
| Description?  | uses partial mail to render text/html list of downloadable virtual products in the download product mail
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #10915
| How to test?  | Order a virtual product with a file and see the text mail sent

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16014)

### DIFF
--- a/classes/order/OrderHistory.php
+++ b/classes/order/OrderHistory.php
@@ -23,6 +23,7 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
+use PrestaShop\PrestaShop\Adapter\MailTemplate\MailPartialTemplateRenderer;
 use PrestaShop\PrestaShop\Adapter\StockManager as StockManagerAdapter;
 use PrestaShop\PrestaShop\Core\Stock\StockManager;
 
@@ -133,27 +134,35 @@ class OrderHistoryCore extends ObjectModel
                 }
 
                 $customer = new Customer((int) $order->id_customer);
-
-                $links = '<ul>';
+                $links = array();
                 foreach ($assign as $product) {
-                    $links .= '<li>';
-                    $links .= '<a href="' . $product['link'] . '">' . Tools::htmlentitiesUTF8($product['name']) . '</a>';
+                    $text = Tools::htmlentitiesUTF8($product['name']);
                     if (isset($product['deadline'])) {
-                        $links .= '&nbsp;' . $this->trans('expires on %s.', array($product['deadline']), 'Admin.Orderscustomers.Notification');
+                        $text .= '&nbsp;' . $this->trans('expires on %s.', array($product['deadline']), 'Admin.Orderscustomers.Notification');
                     }
                     if (isset($product['downloadable'])) {
-                        $links .= '&nbsp;' . $this->trans('downloadable %d time(s)', array((int) $product['downloadable']), 'Admin.Orderscustomers.Notification');
+                        $text .= '&nbsp;' . $this->trans('downloadable %d time(s)', array((int) $product['downloadable']), 'Admin.Orderscustomers.Notification');
                     }
-                    $links .= '</li>';
+                    $links[] = array(
+                        'text' => $text,
+                        'url' => $product['link'],
+                    );
                 }
-                $links .= '</ul>';
+
+                $context = Context::getContext();
+                $partialRenderer = new MailPartialTemplateRenderer($context->smarty);
+
+                $links_txt = $partialRenderer->render('download_product_virtual_products.txt', $context->language, $links, true);
+                $links_html = $partialRenderer->render('download_product_virtual_products.tpl', $context->language, $links);
+
                 $data = array(
                     '{lastname}' => $customer->lastname,
                     '{firstname}' => $customer->firstname,
                     '{id_order}' => (int) $order->id,
                     '{order_name}' => $order->getUniqReference(),
                     '{nbProducts}' => count($virtual_products),
-                    '{virtualProducts}' => $links,
+                    '{virtualProducts}' => $links_html,
+                    '{virtualProductsTxt}' => $links_txt,
                 );
                 // If there is at least one downloadable file
                 if (!empty($assign)) {
--- a/mails/themes/classic/core/download_product.html.twig
+++ b/mails/themes/classic/core/download_product.html.twig
@@ -24,11 +24,16 @@
               <p style="border-bottom:1px solid #D6D4D4;">
                 {{ 'Product(s) now available for download'|trans({}, 'Emails.Body', locale)|raw }}
               </p>
-            
+
 {% endif %}
             <span>
               {{ 'You have [1]{nbProducts}[/1] product(s) now available for download using the following link(s):'|trans({'[1]': '<span><strong>', '[/1]': '</strong></span>'}, 'Emails.Body', locale)|raw }}<br/><br/>
-              {virtualProducts}
+              {% if templateType == 'html' %}
+                {virtualProducts}
+              {% endif %}
+              {% if templateType == 'txt' %}
+                {virtualProductsTxt}
+              {% endif %}
             </span>
           </font>
         </td>
--- a/mails/themes/modern/core/download_product.html.twig
+++ b/mails/themes/modern/core/download_product.html.twig
@@ -259,7 +259,14 @@
                                     </tr>
                                     <tr>
                                       <td align="left" style="font-size:0px;padding:10px 25px;word-break:break-word;">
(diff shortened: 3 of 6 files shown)


matks (dismissed):

classes/order/OrderHistory.php
```diff
@@ -615,4 +627,47 @@ public function addWs()
...
+        }
+
+        if (!isset($this->context)) {
+            $this->context = Context::getContext();
+        }
+
+        $content = $this->getPartialRenderer()->render($template_name, $this->context->language, $var);
+        if ($mail_type == Mail::TYPE_TEXT) {
+            // delete html comments in text mode so it's not visible in the mail
+            $content = preg_replace('/\n?<!--.*-->\n?/i', '', $content);
```
PierreRambaud: Maybe any spaces types? Don't know if all templates contains only `\n` notations, or if a developer add an email created on Mac / Windows.
Add global and dot can match new lines with ungreedy notation
for this kind of comments:
```html
<!-- blyat -->
<p>Htmlcontent</p>
<!-- 
test 
-->
```

```suggestion
            $content = preg_replace('/\s?<!--.*?-->\s?/gs', '', $content);
```

mails/themes/classic/core/download_product.html.twig
```diff
@@ -24,11 +24,16 @@
...
-            
+
 {% endif %}
             <span>
               {{ 'You have [1]{nbProducts}[/1] product(s) now available for download using the following link(s):'|trans({'[1]': '<span><strong>', '[/1]': '</strong></span>'}, 'Emails.Body', locale)|raw }}<br/><br/>
-              {virtualProducts}
+              {% if templateType == 'html' %}
+                {virtualProducts}
+              {% endif %}
+              {% if templateType == 'txt' %}
```
PierreRambaud, line 34: elseif instead

mails/themes/modern/core/download_product.html.twig
```diff
@@ -259,7 +259,14 @@
                                     </tr>
                                     <tr>
                                       <td align="left" style="font-size:0px;padding:10px 25px;word-break:break-word;">
-                                        <div style="font-family:Open sans, arial, sans-serif;font-size:14px;line-height:25px;text-align:left;color:#363A41;"> {virtualProducts} </div>
+                                        <div style="font-family:Open sans, arial, sans-serif;font-size:14px;line-height:25px;text-align:left;color:#363A41;">
+                                          {% if templateType == 'html' %}
+{virtualProducts}
+{% endif %}
+                                          {% if templateType == 'txt' %}
```
PierreRambaud, line 266: same here

mails/themes/classic/core/download_product.html.twig
PierreRambaud: Oh, just noticed these files are generated -_-

mails/themes/modern/core/download_product.html.twig
PierreRambaud: Oh, just noticed these files are generated -_-

classes/order/OrderHistory.php
atomiix: I did that mainly because the partial renderer add the partial file being rendered in html comment like that : 
```
return "\n<!-- begin $tpl -->\n"
            . parent::fetch($template, $cache_id, $compile_id, $parent, $display, $merge_tpl_vars, $no_output_filter)
            . "\n<!-- end $tpl -->\n";
```

PierreRambaud: Oh, I was more thinking about comments in custom html template ^^'

mails/themes/modern/core/download_product.html.twig
atomiix: This one is generated by the [mjml-theme-converter](https://github.com/PrestaShop/mjml-theme-converter), I can't change it

classes/order/OrderHistory.php
jolelievre: I don't understand the need to remove html comment tags, when I look at your txt template `download_product_virtual_products.txt` there is no html content, so why do you need to remove it?

classes/order/OrderHistory.php
```diff
@@ -615,4 +627,47 @@ public function addWs()
...
+     * current_theme/mails/current_iso_lang/ if found, otherwise in
+     * mails/current_iso_lang.
+     *
+     * @param string $template_name template name with extension
+     * @param int $mail_type Mail::TYPE_HTML or Mail::TYPE_TEXT
+     * @param array $var sent to smarty as 'list'
+     *
+     * @return string
+     */
+    protected function getEmailTemplateContent($template_name, $mail_type, $var)
```
jolelievre: As I told you I don't think this extra function is useful, it was already present in PaymentModule so I just modified the last part about partiel rendering. But anyway in the Mail::send function we already check the PS_MAIL_TYPE configuration and this is where the appropriate content is selected (or not) inside the mail body.

So I think you could simply (even drop completely) this method and simply call `$this->getPartialRenderer()->render($template_name, $this->context->language, $var)` to create you two variables
Sure in cas the mail config is not for BOTH you will have generated one for nothin but I the performance cost is really minimal compared to the complexity of the code

jolelievre: Actually it's probably useless in PaymentModule as well, but I wanted to make the minimum changes possible I guess

classes/order/OrderHistory.php
atomiix: That's because when  the `download_product_virtual_products.txt` template is rendered, the renderer adds comments. Here is what I get in my mailbox when I don't remove html comments : 
```
http://localhost/prestashop/index.php

Hi thomas baccelli,

Thank you for your order with the reference JRNKCBIHJ from PrestaShop

Product(s) to download

You have 1 product(s) now available for download using the following link(s):


<!-- begin /Users/tBaccelli/Sites/prestashop/mails/_partials/download_product_virtual_products.txt -->
  [Capture d&rsquo;écran 2019-10-15 à 17.40.05.png](http://localhost/prestashop/index.php?controller=get-file&key=756cddd141c35f16ab816bb301b14b79de34182f-6d13e7f3990af715aeb737adb923a46b5ee456fa&id_order=1932&secure_key=24926c1a734a9822e6ad4f535668f657)

<!-- end /Users/tBaccelli/Sites/prestashop/mails/_partials/download_product_virtual_products.txt -->


Follow your order and download your invoice on our shop, go to the [Order history and details](http://localhost/prestashop/index.php?controller=history) section of your customer account.

If you have a guest account, you can follow your order via the [Guest Tracking](http://localhost/prestashop/index.php?controller=guest-tracking) section on our shop.

[PrestaShop](http://localhost/prestashop/index.php)

Powered by [PrestaShop](https://www.prestashop.com/?utm_source=marchandprestashop&utm_medium=e-mail=utm_campaign=footer_1-7)

```

jolelievre: Oh damn, seems like it's coming from our own over layer on Smarty (probably SmartyDev in this case) https://github.com/PrestaShop/PrestaShop/blob/4ec595f06d52c5aa93b8dd930699f8f6d21cca58/classes/Smarty/SmartyDev.php#L39

This is completely not needed in this case, so we shouldn't use the context Smarty instance but a normal one instead.. but the thing is smarty is configured in https://github.com/PrestaShop/PrestaShop/blob/develop/config/smarty.config.inc.php

Then maybe we should add a `cleanComments` parameter to the `MailPartialTemplateRenderer::render` method

jolelievre (approved):
Nice job @atomiix

matks:
@Robin-Fischer-PS My bad, I told @atomiix to merge this because it was approved, I forgot to mention the QA

Robin-Fischer-PS:
Hi @matks :)

One small defect : The link on the mail includes number of DL and expiration date. The link still works.

pr16014 link

marionf:
@atomiix @matks 

I still have the issue with txt email

capture d'écran_2089

marionf:
My bad, I wasn't on the good store 
It's ok

capture d'écran_2090

PR #13670, opened 2019-05-04
Bug Fix: Multiple options with a couple in stock

tigeryan:
The proposed change will fix a bug where a product has multiple options, with one or more having stock, you can't select an option that does not have stock when you have set the preference to allow out of stock product options to be purchased.

The bug is you have a product with 5 options (S / M / L / XL / XXL) XL is in stock, the others are not in stock.  You have chosen to allow out of stock product to be purchased, because you have one product in stock, when selecting one of the out of stock options the page will reload and go back to the in-stock option as default.  

The bug is the filter that is used for quantity > 0 when the config option of PS_DISP_UNAVAILABLE_ATTR is set.

| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.4.X
| Description?  | The proposed change will fix a bug where a product has multiple options, with one or more having stock, you can't select an option that does not have stock when you have set the preference to allow out of stock product options to be purchased.

The bug is you have a product with 5 options (S / M / L / XL / XXL) XL is in stock, the others are not in stock.  You have chosen to allow out of stock product to be purchased, because you have one product in stock, when selecting one of the out of stock options the page will reload and go back to the in-stock option as default.  
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | No
| Deprecations? | No
| Fixed ticket? | This bug seems to be in various versions, the site I am working on is in the 1.7.4.x 
Here is a closed ticket: http://forge.prestashop.com/browse/BOOM-3120
| How to test?  | Run the code without the patch with a product with 5 options, one with stock with the configuration set as above and you will see the incorrect behavior

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/13670)

PR #14452, opened 2019-07-02
Allow to buy a product without stock when other combinations have stock and unavailable product attributes are not displayed

Progi1984:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Permit to buy a product without stock when other combination have stock and unavailable product attributes are not displayed
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9667 
| How to test?  | * Create a product with combinations. Put stock in some combination but not in all.<br>* Allow to buy products out of stock.<br>* Disable the option "Display unavailable product attributes on the product page" in Product Settings page<br>* In FO, combinations out of stock can be selected

@colinegin @TristanLDD I think that correct the bug #9667 but the behavior is not perfect : when you deny to buy products out of stock, the combination (which doesn't have stock) reset

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/14452)

Issue #9667, opened 2018-08-22
[BOOM-5345] It is not possible to buy a product without stock when other combinations have stock


prestonBot:
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-5345).

- _**Reporter:**_ JuanjoSC
- _**Created at:**_ Mon, 16 Apr 2018 18:12:27 +0200

<p>I can not select a combination that has no stock when another has it.</p>

<p>If no combination has stock it works correctly.</p>
        
* How to reproduce the issue ?

1-Create a product with combinations. Put stock in some combination but not in all.
<br/>
2-Allow to buy products out of stock.
<br/>
3-Disable the option &quot;Display unavailable product attributes on the product page&quot; in Product Settings page
<br/>
4-In FO, combinations out of stock can not be selected
<br/>

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-5345#comment-166270).

- _**Author:**_ khouloud.belguith
- _**Created at:**_ Mon, 16 Apr 2018 18:23:09 +0200

<p>Hi Juanjo,</p>

<p>I did not manage to reproduce the issue with the last version 1.7.3.1.<br/>
I attached a video record.<br/>
 <span class="nobr"><a href="http://forge.prestashop.com/secure/attachment/67001/67001_Apr+16+2018+6_20+PM.webm" title="Apr 16 2018 6_20 PM.webm attached to BOOM-5345">Apr 16 2018 6_20 PM.webm<sup><img class="rendericon" src="http://forge.prestashop.com/images/icons/link_attachment_7.gif" height="7" width="7" align="absmiddle" alt="" border="0"/></sup></a></span> <br/>
Thanks to check and feedback.</p>

<p>Best regards, Khouloud</p>

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-5345#comment-166272).

- _**Author:**_ juanjosc
- _**Created at:**_ Mon, 16 Apr 2018 18:40:24 +0200

<p>Thanks for the quick reply.</p>

<p>Attached video of what he does to me.</p>

[...]

PR #16016, opened 2019-10-18
Fix inability to order product variant out of stock for product accepting orders

Shoprunners:
| Questions     | Answers
| Branch?       | develop
| Description?  | Products that accept orders without stock do not function properly if some variants have stock and others don't. The variant selection always jumps to the variant with stock available, even if all variants would accept orders. The applied fix copies the same condition as from further down in the ProductController, where variants are filtered in order to be displayed. The same logic must apply here.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | This fixes the work started in #13670
| How to test?  | Have a product accepting orders without stock with some variants with stock and some without. The ajax selection will always deselect the variant without stock prior to this fix

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16016)

### DIFF
--- a/controllers/front/ProductController.php
+++ b/controllers/front/ProductController.php
@@ -955,7 +955,7 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
      */
     private function tryToGetAvailableIdProductAttribute($checkedIdProductAttribute)
     {
-        if (!Configuration::get('PS_DISP_UNAVAILABLE_ATTR')) {
+        if (!Product::isAvailableWhenOutOfStock($this->product->out_of_stock) && Configuration::get('PS_DISP_UNAVAILABLE_ATTR') == 0) {
             $availableProductAttributes = $this->product->getAttributeCombinations();
             if (!Product::isAvailableWhenOutOfStock($this->product->out_of_stock)) {
                 $availableProductAttributes = array_filter(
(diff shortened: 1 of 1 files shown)


matks (dismissed):

controllers/front/ProductController.php
```diff
@@ -955,7 +955,7 @@ private function getIdProductAttributeByRequestOrGroup()
      */
     private function tryToGetAvailableIdProductAttribute($checkedIdProductAttribute)
     {
-        if (!Configuration::get('PS_DISP_UNAVAILABLE_ATTR')) {
+        if (!Product::isAvailableWhenOutOfStock($this->product->out_of_stock) && Configuration::get('PS_DISP_UNAVAILABLE_ATTR') == 0) {
```
matks, line 959: Hi @Shoprunners, do you think we can improve the readability of this function as it gets more complex now ? Either using a code comment or using a variable name*

*Example
```
$myVar = (!Product::isAvailableWhenOutOfStock($this->product->out_of_stock) && Configuration::get('PS_DISP_UNAVAILABLE_ATTR') == 0);
if ($myVar) {
```

Shoprunners: I've now added a comment similar to the one at the other site using the exact logic while also referring to that function.

marionf:
Hello @Shoprunners 

Could you please open an issue and describe how to reproduce it ?

It seems related to this issue: https://github.com/PrestaShop/PrestaShop/issues/9667 which was already fixed by https://github.com/PrestaShop/PrestaShop/pull/14452

Shoprunners:
Hi @marionf ,
you are right, it is the same bugfix.

marionf:
Thank you for your answer @Shoprunners 
I close this PR as the issue has already been fixed

PR #16018, opened 2019-10-18
Sanity Tests - Refacto of all tests 

boubkerbribri:
| Questions     | Answers
| Branch?       | develop
| Description?  | Refactoring sanity tests, Adding expected results in each tests (mocha it), separate order and orders page.
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | See travis result

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16018)

### DIFF
--- a/tests/puppeteer/campaigns/sanity/03_ordersBO/01_filterOrders.js
+++ b/tests/puppeteer/campaigns/sanity/03_ordersBO/01_filterOrders.js
@@ -8,9 +8,10 @@ const loginCommon = require('@commonTests/loginBO');
 const LoginPage = require('@pages/BO/login');
 const DashboardPage = require('@pages/BO/dashboard');
 const BOBasePage = require('@pages/BO/BObasePage');
-const OrderPage = require('@pages/BO/order');
+const OrdersPage = require('@pages/BO/orders');
 const {Orders, Statuses} = require('@data/demo/orders');
 
+let numberOfOrders;
 let browser;
 let page;
 // creating pages objects in a function
@@ -19,7 +20,7 @@ const init = async function () {
     loginPage: new LoginPage(page),
     dashboardPage: new DashboardPage(page),
     boBasePage: new BOBasePage(page),
-    orderPage: new OrderPage(page),
+    ordersPage: new OrdersPage(page),
   };
 };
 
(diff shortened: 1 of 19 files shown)


tests/puppeteer/pages/Install/install.js
```diff
@@ -129,33 +129,35 @@ module.exports = class Install extends CommonPage {
...
+      this.page.waitForSelector(this.installDefaultDataStep, {visible: true, timeout: 120000}),
+      this.page.waitForSelector(this.populateDatabaseStep, {visible: true, timeout: 180000}),
+      this.page.waitForSelector(this.configureShopStep, {visible: true, timeout: 240000}),
+      this.page.waitForSelector(this.installModulesStep, {visible: true, timeout: 360000}),
+      this.page.waitForSelector(this.installModulesAddons, {visible: true, timeout: 360000}),
+      this.page.waitForSelector(this.installThemeStep, {visible: true, timeout: 360000}),
+      this.page.waitForSelector(this.installFixturesStep, {visible: true, timeout: 360000}),
+      this.page.waitForSelector(this.finalStepPageTitle, {visible: true, timeout: 360000}),
+    ]);
+    return this.checkStepTitle(this.finalStepPageTitle, this.finalStepEnTitle);
```
SimonGrn, line 160: Oof that's a big Promise you got there...

SimonGrn (dismissed):

tests/puppeteer/pages/Install/install.js
boubkerbribri: yep :smile:

tests/puppeteer/campaigns/sanity/03_ordersBO/01_filterOrders.js
```diff
@@ -43,48 +44,79 @@ describe('Filter the Orders table by ID, REFERENCE, STATUS', async () => {
...
-      this.pageObjects.orderPage.orderFilterReferenceInput,
+    await this.pageObjects.ordersPage.filterOrders(
+      'input',
+      'reference',
       Orders.fourthOrder.ref,
-      this.pageObjects.orderPage.searchButton,
     );
     const result = await this.pageObjects.boBasePage.checkTextValue(
-      this.pageObjects.orderPage.orderfirstLineReferenceTD,
+      this.pageObjects.ordersPage.orderfirstLineReferenceTD.replace('%ROW', '1'),
```
PierreRambaud, line 87: Why '1'?

tests/puppeteer/campaigns/sanity/03_ordersBO/02_editOrder.js
```diff
@@ -41,15 +43,17 @@ describe('Edit Order BO', async () => {
...
+    await this.pageObjects.boBasePage.goToSubMenu(
+      this.pageObjects.boBasePage.ordersParentLink,
+      this.pageObjects.boBasePage.ordersLink,
+    );
+    const pageTitle = await this.pageObjects.ordersPage.getPageTitle();
+    await expect(pageTitle).to.contains(this.pageObjects.ordersPage.pageTitle);
   });
   it('should go to the first order page', async function () {
-    await this.pageObjects.boBasePage.waitForSelectorAndClick(this.pageObjects.orderPage.orderfirstLineIdTD);
+    await this.pageObjects.ordersPage.goToOrder('1');
```
PierreRambaud, line 54: the '1' should be in configuration.

tests/puppeteer/pages/FO/cart.js
```diff
@@ -48,5 +48,20 @@ module.exports = class Cart extends FOBasePage {
...
+  /**
+   * To get a number from text
+   * @param selector
+   * @param timeout
+   * @return integer
+   */
+  async getPriceFromText(selector, timeout = 0) {
+    await this.page.waitFor(timeout);
+    const text = await this.getTextContent(selector);
+    const number = /[+-]?\d+(\.\d+)?/g.exec(text).toString();
```
PierreRambaud: Hum, depending on how the price if formatted :thinking: 
10 000.99 will be works

tests/puppeteer/campaigns/sanity/03_ordersBO/01_filterOrders.js
boubkerbribri: > Why '1'

We are sure that we will have a result (and it will be in first row)
> Why I add `.replace('%ROW', '1')`

Selector took first row by default, I changed it in orders page, now he have to choose in which row the check should be.
> Why string and not an Int

Can be an int

tests/puppeteer/pages/FO/cart.js
boubkerbribri: deleting spaces on string before extracting the number is better maybe ?

tests/puppeteer/campaigns/sanity/03_ordersBO/02_editOrder.js
boubkerbribri: What do you mean exactly by 'in configuration' ?

PierreRambaud: Exactly the same as you do for html selector, but for objects values.
If you use this id many times, maybe you should create
DEFAULT_ORDER_ID = 1
Because if one day you add scenario with many others id, it will be easier for everyone to understand why the default order id is X, and not Y.
Same for products, if you want a customizable product, it's maybe not 1 (I really don't know), and if you want a product with combination, you will take the 8 (Still don't know :sweat_smile: ), adding "words" to defined which object you'll use will help everybody to understand what you're doing, just by reading.

tests/puppeteer/pages/FO/cart.js
PierreRambaud: Depends on what text are returning, give me the string and I'll give you the best regex or code to sanitize it!

boubkerbribri: For now, `text` return something like `€57.41` (no spaces)
With actual regex, it's sure that we will have the value needed `57.41`.
but of course, if text return `€10 000.99`, number will be incorrect `0.99` 
That's why i suggested to remove spaces from text before getting our integer.

PierreRambaud: You could simply use
```suggestion
    const number = Number(text.replace(/[^0-9.-]+/g,''));
```
and everything will be ok :)

tests/puppeteer/campaigns/sanity/03_ordersBO/02_editOrder.js
boubkerbribri: `goToOrder` take `orderRow` as param (not an orderID).
In this test, we don"t care which order to edit.
If we want to choose a specific one, we will filter (with reference for example), check that we have the order in first row, and if passed, `goToOrder(row='1')`
we have a orders demo data, that contains ids but `id != row`

tests/puppeteer/pages/FO/cart.js
SimonGrn: Shouldn't you escape the dot in your regexp ?

tests/puppeteer/campaigns/sanity/03_ordersBO/02_editOrder.js
PierreRambaud: my bad, missreading :/

tests/puppeteer/pages/FO/cart.js
PierreRambaud: nope because it's in `[]`

PierreRambaud (approved):

Issue #15822, opened 2019-10-04
[ORDER PAGE] Status Tab Order - 5/11


MatShir:
Part of the order page migration and redesign project.

**User Story**

As a merchant, when I check an order, I need to be able to see the status history with the editor of the status changes.

**Detailed specification**
Specs on the repo prestashop-specs
[Status tab](https://github.com/PrestaShop/prestashop-specs/blob/cc476f1fc0c71c63bd3ba4879131f5163fca05e7/Order%20Page%20View.md#v-tabs-panel)

matks:
Fixed by https://github.com/PrestaShop/PrestaShop/pull/16022

Issue #15818, opened 2019-10-04
[ORDER PAGE] Order Actions (update status, view invoice, delivery slip) - 1/11


MatShir:
Part of the order page migration and redesign project.

**User Storie**

As a merchant, when I check an order, I need to be able to change the order status and to view the invoice & the delivery slip

**Detailed specification**
Specs on the repo prestashop-specs
[Order Action Specs](https://github.com/PrestaShop/prestashop-specs/blob/cc476f1fc0c71c63bd3ba4879131f5163fca05e7/Order%20Page%20View.md#i-action-panel
)

matks:
Fixed by https://github.com/PrestaShop/PrestaShop/pull/16022

PR #16022, opened 2019-10-18
Migrate order action and status block

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migrates "Update status", "View invoice" and "View delivery slip" actions. In addition, whole "Status" block should behave the same as legacy.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Partially #15818 and fully #15822
| How to test?  | :warning: Rebuild assets :warning: "Update status", "View invoice" and "View delivery slip" actions should behave the same as in legacy page. In addition, whole "Status" block should behave the same as legacy as well.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16022)

### DIFF
--- a/src/Core/Domain/Order/QueryResult/OrderCarrierForViewing.php
+++ b/src/Core/Domain/Order/QueryResult/OrderCarrierForViewing.php
@@ -78,10 +78,10 @@ class OrderCarrierForViewing
     /**
      * @param int $orderCarrierId
      * @param DateTimeImmutable $date
-     * @param string $name
+     * @param string $name Carrier name or null in case of virtual order
      * @param string $weight
      * @param int $carrierId
-     * @param string $price
+     * @param string $price Price or null in case of virtual order
      * @param string|null $trackingUrl
      * @param string|null $trackingNumber
      * @param bool $canEdit
@@ -89,10 +89,10 @@ class OrderCarrierForViewing
     public function __construct(
         int $orderCarrierId,
         DateTimeImmutable $date,
-        string $name,
+        ?string $name,
         string $weight,
         int $carrierId,
-        string $price,
+        ?string $price,
         ?string $trackingUrl,
         ?string $trackingNumber,
         bool $canEdit
@@ -127,7 +127,7 @@ class OrderCarrierForViewing
     /**
      * @return string
      */
-    public function getName(): string
+    public function getName(): ?string
     {
         return $this->name;
     }
@@ -143,7 +143,7 @@ class OrderCarrierForViewing
     /**
      * @return string
      */
-    public function getPrice(): string
+    public function getPrice(): ?string
     {
         return $this->price;
     }
--- a/src/PrestaShopBundle/Twig/Extension/LocalizationExtension.php
+++ b/src/PrestaShopBundle/Twig/Extension/LocalizationExtension.php
@@ -1,0 +1,64 @@
+<?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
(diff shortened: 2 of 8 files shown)


src/PrestaShopBundle/Twig/Extension/LocalizationExtension.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+namespace PrestaShopBundle\Twig\Extension;
+
+use DateTime;
+use DateTimeInterface;
+use Twig\Extension\AbstractExtension;
+use Twig_SimpleFilter;
+
+class LocalizationExtension extends AbstractExtension
```
sarjon, line 34: @matks wdyt about name?

Also, I think we should document this extension as it is very useful.

src/Core/Domain/Order/QueryResult/OrderCarrierForViewing.php
```diff
@@ -78,21 +78,21 @@ class OrderCarrierForViewing
     /**
      * @param int $orderCarrierId
      * @param DateTimeImmutable $date
-     * @param string $name
+     * @param string $name Carrier name or null in case of virtual order
```
PierreRambaud, line 81:
```suggestion
     * @param string|null $name Carrier name or null in case of virtual order
```

src/Core/Domain/Order/QueryResult/OrderCarrierForViewing.php
```diff
@@ -78,21 +78,21 @@ class OrderCarrierForViewing
...
      * @param string $weight
      * @param int $carrierId
-     * @param string $price
+     * @param string $price Price or null in case of virtual order
```
PierreRambaud, line 84:
```suggestion
     * @param string|null $price Price or null in case of virtual order
```

src/PrestaShopBundle/Twig/Extension/LocalizationExtension.php
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public function __construct(string $contextDateFormatFull)
+    {
+        $this->dateFormatFull = $contextDateFormatFull;
+    }
+
+    public function getFilters(): array
+    {
+        return [
+            new Twig_SimpleFilter('date_format_full', [$this, 'dateFormatFull']),
```
PierreRambaud, line 52: I can't consider `Y-m-d H:i:s` as a full date format when you can have `2004-02-12T15:19:21+00:00` or `Thu, 21 Dec 2000 16:01:07 +0200`.
Should we have `default_date_format` instead?

sarjon: I don't think so, we have 2 different settings, so date can be displayed in different formats: full & lite.

See https://prnt.sc/pm2fxt

matks: Indeed this is "full according to PrestaShop formatting conventions"
It makes sense 👍as the format is controlled by PS

src/PrestaShopBundle/Twig/Extension/LocalizationExtension.php
matks: A bit generic ... but it works for now. We should monitor what else we'll put in it (right now it's a DateFormatterExtension). If we add more and more localization-related functions in it, there it works 👍

matks (dismissed):

src/PrestaShopBundle/Twig/Extension/LocalizationExtension.php
sarjon: Yeah, you are right, it's a little generic, maybe we can rename it to `DateFormatterExtension` instead? Now when I think about it, it wouldn't make sense to inject multiple services, if we need more, lets say for price display, we could better create `PriceFormatterExtension`.

matks: It makes sense to have multiple services if they are part of the same component. See `PrestaShopBundle\Twig\Extension\GridExtension` 😉 

Price formats and date formats are both linked to localization so I think it's alright to have them side by side.

matks: (it's unlikely that you'd need to localize dates but not prices)

sarjon:
Rebased.

PierreRambaud (approved):

jolelievre (approved):

Robin-Fischer-PS:
Hi @sarjon !

It's OK for me, thanks :)

matks:
Thank you @sarjon

Issue #15823, opened 2019-10-04
[ORDER PAGE] Documents Tab Order - 6/11


MatShir:
Part of the order page migration and redesign project.

**User Story**

As a merchant, when I check an order, I need to be able to see all the documents generated by the order.

**Detailed specification**
Specs on the repo prestashop-specs
[Documents tab
](https://github.com/PrestaShop/prestashop-specs/blob/cc476f1fc0c71c63bd3ba4879131f5163fca05e7/Order%20Page%20View.md#v-tabs-panel)

matks:
Fixed by https://github.com/PrestaShop/PrestaShop/pull/16046

PR #16046, opened 2019-10-21
Migrate Documents block of Order View page

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | 
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/15823
| How to test?  | :warning: Rebuild assets :warning: Documents block should behave the same as it was in legacy.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16046)


matks (dismissed):

sarjon:
Rebased.

jolelievre (dismissed):

matks:
1 conflict fixed, be careful when you `git pull`

matks (dismissed):

Robin-Fischer-PS:
Hi @sarjon ! Thanks for the PR :smile: 

I've noticed two bugs : 

- When there is no document on an order, there should be a "generate invoice" button in document block.
Here in legacy : 
PR16046 generate invoice legacy

- When there is no payment associated with an invoice, there should be a "Enter payment" button.
Here in legacy : 
PR16046 enter payment legacy

sarjon:
> When there is no payment associated with an invoice, there should be a "Enter payment" button.
Here in legacy :

Regarding this one, do we really need it? In legacy, Payments panel were way below Documents panel, but in new page (thus new layout), Payments panel is just below Documents, so does it make sense to have this button and scroll user a few pixels below?

@matks wdyt? Maybe we should ask Product team as well? :thinking:

src/Core/Domain/Configuration/Configuration.php
```diff
@@ -0,0 +1,46 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Configuration;
+
+/**
+ * Defines available PrestaShop configurations.
+ */
+class Configuration
```
sarjon: @matks wdyt about this class? Is it worth having? If so, maybe name could be improved?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -277,6 +281,13 @@ public function viewAction(int $orderId, Request $request): Response
         $addProductToOrderForm = $this->createForm(AddProductToOrderType::class);
         $updateOrderProductForm = $this->createForm(UpdateProductInOrderType::class);
 
+        $invoiceManagementIsEnabled = $this->getQueryBus()->handle(
+            new GetConfiguration(
+                Configuration::INVOICE_MANAGEMENT,
+                $orderForViewing->getShopId()
+            )
+        );
```
sarjon: @matks wdyt about this one? Current `ConfigurationInterface` is just broken, so I've implemented new query to get configuration values. I have a few thoughts left to discuss:

1. Does it make sense?
2. Should we keep `mixed` value as return type?
    If not, I can see at least 2 options here:
    1. We could ask query for specific type, e.g. `new GetConfiguration('KEY', null, null, ConfigType::BOOL)`, so value conversion will happen inside query handler.
    2. Or we could return DTO instead of `mixed`, e.g. `ConfigValue` as we could use it as `$configValue->asString()` or `$configValue->asBool()`, wdyt? 
3. It's meant for single configuration value, if we need to get multi language config, then we could add `GetMultilangConfiguration` query, would it make sense? Or we could add `$langId` to query, but that would make return value more complex.

src/Core/Domain/Configuration/Configuration.php
matks: Is it intended to become a registry of all known configuration keys such as `PS_TAX` ?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
matks: I'm not sure :/ I'm afraid this is not a good usage of Queries

Commands and Queries are supposed to match user-oriented usecases. "Get all registered customers", "Delete order", "Get product informations (for edit form)". In theory* 1 page in the BO = 1 Query dispatched in the bus. Here you are not creating a Query because it matches a user need, but because you, as a developer, need this information.


*But in reality we cannot and dont want to follow this rule 100% we stay pragmatic

src/Core/Domain/Configuration/Configuration.php
sarjon: I don't know, but it very well could be.

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
sarjon: That is true, but on the other hand, we have both use case oriented queries and generic queries.

I was thinking about adding `GetInvoiceManagementIsEnabled` query, but not sure that's a good idea either. :/

How would you suggest going around it?

matks: Why dont we retrieve the data through `GetOrderForViewing` ?

sarjon: I don't think it would be right, this config is totally independable from order. But do you think it's ok? :thinking:

matks: I think it's OK indeed. A Query result is an aggregate of all the required data for the usecase. If `$invoiceManagementIsEnabled` is a required data to righfully display an order, then it's part of the API.

ping @eternoendless to be sure about that ?

matks: @sarjon Got a confirmation from Pablo: CQRS is a business-oriented interface, not a technical interface
so if `invoiceManagementIsEnabled` is required element for business in order to edit an order, then it must be returned by the query

sarjon: Ok, fixed!

matks (dismissed):

sarjon:
@matks we need to clarify this one https://github.com/PrestaShop/PrestaShop/pull/16046#issuecomment-547783410

matks:
Indeed, sorry
@TristanLDD @marionf @MatShir can you check https://github.com/PrestaShop/PrestaShop/pull/16046#issuecomment-547783410 😉 ?

marionf:
As it's about order page, I ping @MatShir

matks:
@MatShir 🆙

MatShir:
> > When there is no payment associated with an invoice, there should be a "Enter payment" button.
> > Here in legacy :
> 
> Regarding this one, do we really need it? In legacy, Payments panel were way below Documents panel, but in new page (thus new layout), Payments panel is just below Documents, so does it make sense to have this button and scroll user a few pixels below?
> 
> @matks wdyt? Maybe we should ask Product team as well?

Seen with Tristan, we should keep the same behaviours which are to scroll down and add the invoice amount in the amount input of the payment line with the right invoice.

matks:
@Robin-Fischer-PS @MatShir The 2 items reported by Robin have been fixed. Dont forget to rebuild the assets. Please check his PR again 😊

matks (dismissed):

PierreRambaud (dismissed):

Robin-Fischer-PS:
Hi @matks ! It's QA :heavy_check_mark: !

matks (approved):

matks:
Thank you @sarjon

Issue #16043, opened 2019-10-21
Customers created in another shop are not displayed in the list if the shop group share customers


marionf:
**Describe the bug**
Customers created in another shop are not displayed in the list if the shop group share customers

**To Reproduce**
1. Enable multistore in Shop parameters > General
2. Go in advanced parameters > Multistore 
3. Add a new shop group "group2" and share customers, share available quantity to sell and share orders
4. Add a new shop "shop1" in "group2"
5. Define an URL for "shop1"
6. Add a new shop "shop2" in group2"
7. Define an URL for "shop2"
8. Open "shop1" FO and create a customer account
9. Go in BO > Customers > Customers and select "Shop1": You see the customer created OK
10. Click on view the customer in "shop1": OK
11. Select "shop2", you still view the customer details: OK
12. Go in BO > Customers > Customers and select "Shop2": You don't see the customer created NOK
13. Open "shop2" FO and try to login with the customer created in "shop1": OK 

If the shop group share customers, I should see in the list customers created in another shop
It was working in 1.7.5.2

**Additional information**
PrestaShop version: 1.7.6.1
PHP version: N/A

marionf:
@Robin-Fischer-PS & @sarahdib 
You can take note of this scenario for the migration of order list & order view pages
As we also have the option to "share orders"
Which means if it's enabled for the shop group, I should be able to see in shop2 an order created in shop1 and vice versa

jolelievre:
Closed by #16047

PR #16047, opened 2019-10-21
Handle multistore customers sharing in customers listing at group level

matks:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Handle multistore customers sharing in customers listing at group level
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/16043
| How to test?  | Please see issue, and check multistore behavior for Customers listing. Please also check if "customers sharing" setting is working fine when disabled

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16047)

### DIFF
--- a/src/Adapter/Shop/Context.php
+++ b/src/Adapter/Shop/Context.php
@@ -73,6 +73,23 @@ class Context implements MultistoreContextCheckerInterface, ShopContextInterface
     }
 
     /**
+     * Return the result of getContextListShopID() for customers usecase
+     * This handles the "multishop sharing customer" feature setting
+     *
+     * @return array
+     */
+    public function getContextListShopIDUsingCustomerSharingSettings()
+    {
+        $groupSettings = Shop::getGroupFromShop(Shop::getContextShopID(), false);
+
+        if ($groupSettings['share_customer']) {
+            return Shop::getContextListShopID(Shop::SHARE_CUSTOMER);
+        } else {
+            return Shop::getContextListShopID();
+        }
+    }
+
+    /**
      * Get if it's a GroupShop context.
      *
      * @return bool
--- a/src/PrestaShopBundle/Resources/config/services/core/grid/doctrine_query_builder.yml
+++ b/src/PrestaShopBundle/Resources/config/services/core/grid/doctrine_query_builder.yml
@@ -66,7 +66,7 @@ services:
         arguments:
             - '@prestashop.core.query.doctrine_search_criteria_applicator'
             - "@=service('prestashop.adapter.legacy.context').getContext().language.id"
-            - '@=service("prestashop.adapter.shop.context").getContextListShopID()'
+            - '@=service("prestashop.adapter.shop.context").getContextListShopIDUsingCustomerSharingSettings()'
         public: true
 
     prestashop.core.grid.quer_.builder.language:


src/PrestaShopBundle/Resources/config/services/core/grid/doctrine_query_builder.yml
```diff
@@ -66,7 +66,7 @@ services:
         arguments:
             - '@prestashop.core.query.doctrine_search_criteria_applicator'
             - "@=service('prestashop.adapter.legacy.context').getContext().language.id"
-            - '@=service("prestashop.adapter.shop.context").getContextListShopID()'
+            - '@=service("prestashop.adapter.shop.context").getContextListShopID(parameter("const_shop_share_customer"))'
```
matks: Although this behavior was handled by a long if/else in legacy code:
https://github.com/PrestaShop/PrestaShop/blob/develop/classes/shop/Shop.php
It seems using a proper `Shop::getContextListShopID($share))` I can deal with it more simply
https://github.com/PrestaShop/PrestaShop/blob/develop/classes/shop/Shop.php#L923

matks: To sum it up:
It was implemented in another way in legacy customers listing
but ... it seems I can fix the bug passing a single param to what was used before
so: I was reused a legacy function, but different from the one that was used before

PierreRambaud (dismissed):

src/PrestaShopBundle/Resources/config/services/core/grid/doctrine_query_builder.yml
PierreRambaud: Really cool if we can fix it like that! :+1:

jolelievre: The only question is: are we sure that `prestashop.core.grid.query_builder.customer` is only used in this context and/or always needs to use the group sharing?

matks: @jolelievre ah you're right ... I think I "forced" the "shared data" behavior although this is an option that merchant can choose not to use

colinegin:
@Robin-Fischer-PS did you find any issue on this PR ?

Robin-Fischer-PS:
@colinegin No, but @jolelievre told me it wasn't ready for QA yet on Slack !

src/PrestaShopBundle/Resources/config/services/core/grid/doctrine_query_builder.yml
jolelievre: Mmm, so does it mean you should use the value from a `Configuration::get` (if it exists) instead of the const value?

matks: I'm going to need to re-implement legacy way to work
which means this `if/else`
https://github.com/PrestaShop/PrestaShop/blob/develop/classes/shop/Shop.php#L1077

to make it short
```
if ($group['share_customer']) {
    $ids = Shop::getContextListShopID($share));
} else {
    $ids = Shop::getContextListShopID())
}
```

and I need to do this at configuration/constructor level 🤔 

I'm afraid a BC break will be needed 😭

jolelievre: You mean a BC because you will have to change the constructor of `CustomerQueryBuilder`

Maybe you could instead create a new service `ShopRepository` with a `getContextListId()` method that would manage this if/else case Then you can simply change the service definition and there is no BC (all the naming can of course be improved, it's a quick proposition)

matks: Idea is good 👍

jolelievre (changes requested):
Just to clarify this PR status and write what we discussed:
- create a service to get the shop list
- this service is able to adapt this list to the ShopContext (especially the shared data depending on the group/shop)
- update `prestashop.core.grid.query_builder.customer` to use this new service for its dependencies
- check if some other services might have the same problem in their definition

matks:
I implemented @jolelievre idea and ammended the commit. Should be fine now.

src/Adapter/Shop/Context.php
```diff
@@ -72,6 +72,23 @@ public function getContextListShopID($share = false)
...
+     * This handles the "multishop sharing customer" feature setting
+     *
+     * @return array
+     */
+    public function getContextListShopIDUsingCustomerSharingSettings()
+    {
+        $groupSettings = Shop::getGroupFromShop(Shop::getContextShopID(), false);
+
+        if ($groupSettings['share_customer']) {
+            return Shop::getContextListShopID(Shop::SHARE_CUSTOMER);
```
jolelievre, line 86: Maybe we should manage the other two sharing types since we're at it ^^
Shop::SHARE_ORDER and SHOP::SHARE_STOCK
cd https://github.com/PrestaShop/PrestaShop/blob/2776a60d01dc3dac4fbfeb02b8bb67daedbed574/classes/shop/Shop.php#L900

jolelievre: I wonder if we should also take a look at other places where this method is used like:
https://github.com/PrestaShop/PrestaShop/blob/2776a60d01dc3dac4fbfeb02b8bb67daedbed574/src/Adapter/Module/ModuleDataProvider.php#L167

But I guess this bug only occurs for entities that have sharing capabilities, so Customer, Order, Stock Then we must be careful about (develop branch only):

https://github.com/PrestaShop/PrestaShop/blob/2776a60d01dc3dac4fbfeb02b8bb67daedbed574/src/PrestaShopBundle/Resources/config/services/core/grid/doctrine_query_builder.yml#L202

jolelievre (approved):
Thanks @matks I approve this one since I think it's enough for the bug
Although there might be other places in the code that could have the same issue (see my last comments)

Robin-Fischer-PS:
Hi here ! It's good for QA.

Thanks @matks :)

jolelievre:
Thank you @matks

Issue #15829, opened 2019-10-04
[EPIC] Localization packs rework - 1.7.7


LouiseBonnard:
Since the information in localization packs is not automatized yet, this issue is here to gather all the tasks to do for the next 1.7.7 PrestaShop version in order to update as much localized data as possible. 

## Countries

- [x] #15234 - Add states to India / Fixed in #15627
- [x] #15232 - Switch zone for Croatia / Fixed in #15618 
- [x] #13636 - Adding zones for: Austria, Switzerland, Denkmark, Netherlands, Thailand (the last one mandatory for Paypal Payments API)
- [x] #16001 - Add Ivory Coast as a supported country / Fixed in #16173 
- [x] #16833 - Switch zone for Bosnia and Herzegovina, Liechtenstein, and Ukraine / Fixed in #16834

## Currencies

- [x] #16020 - Update Turkish currency sign

## Flags

- [x] #16029 - Wrong flag and ISO code for Argentina / Fixed in #16334

## Installer

- [x] #15438 - Add Albanian and Korean to the installer / Fixed in #16416
- [x] ~~#16239 - Add Uzbek to the installer / Fixed in #16374~~

## Languages

- [x] ~~#15506 - Wrong language displaying for Slovenian~~
- [x] ~~#16053 - Update Hong Kongese language codes~~

## Localization packs

- [x] #15838 - Create a repository to stock localization files / Done [here](https://github.com/PrestaShop/LocalizationFiles)
- [x] #15857 - Add missing languages in the right localization packs / Fixed in #16292
- [x] #16321 - Clean languages (again) in localization packs / Fixed in  #16489
- [x] #16435 - Add extra localization packs / Fixed in #16927

## New supported languages

- [x] #15435 - Add Malagasy, Uzbek, and Tagalog as new supported languages / Fixed in #15484 
- [x] #16502 - Add Khmer as a new supported language / Fixed in #16690

## Update tax rates

- [x] #16122 - Update tax rates / Fixed in #16291
- [x] #16860 - Fix United Kindom VAT issue for all EU countries

LouiseBonnard:
Everything has been done so I close this EPIC. Many thanks to everyone involved here, see you in 1.7.7!
☀️

Navruzbek:
May I ask why it is removed?  Add Uzbek to the installer / Fixed in

PR #16053, opened 2019-10-21
Update Hong Kongese language codes

LouiseBonnard:
| Questions     | Answers
| Branch?       | develop
| Description?  | Update localization packs data, add two new language codes - feedback from the Hong Kongese ambassador ([here](https://docs.google.com/spreadsheets/d/1QvBo1q_ddSZU3wcawGN_oZVeutY-CB5eJaYfky6HXys/edit#gid=0))
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | It is part of this [EPIC](https://github.com/PrestaShop/PrestaShop/issues/15829).

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16053)

### DIFF
--- a/localization/hk.xml
+++ b/localization/hk.xml
@@ -5,6 +5,8 @@
 	</currencies>
 	<languages>
 		<language iso_code="tw" />
+    <language iso_code="zh" />
+		<language iso_code="hk" />
 	</languages>
 	<units>
 		<unit type="weight" value="市制" />


Progi1984:
@LouiseBonnard Which standard do you use for languages ? Is it  [ISO-639-1 ](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) ? If yes, `tw` is related Twi dialect (Niger/Togo). And `hk` doesn't exist.

LouiseBonnard:
@Progi1984, thanks a lot! It seems that the Chinese _localization_ file is also wrong (and thus the _install-dev/langs_ folder), we should only keep `zh` as the Chinese language code... right now we make the difference between Chinese simplified (`zh`) and Chinese traditional (`tw`) but if one of this code is wrong, I guess it will complicate things a bit. I'm going to check it all.

LouiseBonnard:
Well, there is a confusion between country (ISO 3166-1 alpha 2) and language (ISO 639-1) codes, I'll talk about it in another issue so it can be corrected. :-)

PR #16074, opened 2019-10-22
Prioritize customer search by id when saving an address

### LINKED ISSUE #16116: Newly created addresses are saved with the wrong `id_customer`
**Describe the bug**
Newly created addresses are saved with the wrong `id_customer`

**To Reproduce**
1. The system must have at least one guest account and one non-guest account (the guest account must be created earlier)
2. Go to BackOffice -> Orders -> Add new order
3. Type the customer name or email we created earlyer (two accounts with the same email)
4. Select the latest account (with the highest id)
5. Create a new address for this customer -> "Add a new address"
6. After we save the address, we will not see it in the list and cannot use it. Since it is not there. 
After the research I found out that the address was assigned to a user with the same email but with a guest account. 
There are two inaccuracies here: 
 - the user's search is done first by his email and not by the id, and therefore. This is not true, that is why the address was assigned to the wrong user. 
 - I don't think there should be any guest accounts on this list at all. You can't create a guest order from BackOffice.

The easiest solution is to swap conditions if we can get an id_customer, then we don't need to search for a user by email (see at the screenshot)
579b04c740

**Additional information**
PrestaShop version: 1.6.* - 1.7.*
PHP version: 5.6, 7.0, 7.1

The PR [https://github.com/PrestaShop/PrestaShop/pull/16074](https://github.com/PrestaShop/PrestaShop/pull/16074)

levyn:
| Questions     | Answers
| Branch?       | develop
| Description?  | When creating a new address from an order in the BackOffice, the address is assigned to the wrong user
| Type?         | bug fix 
| Category?     |  BO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16116 
| How to test?  | 1. Create few guest or non-guest accounts (with the same email address).<br>2. Start creating a new order from the Backoffice<br>3. Select the last account you have created<br>4. Create a new address for this user<br>Error: the address will be assigned to the first from previously created account

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16074)

### DIFF
--- a/controllers/admin/AdminAddressesController.php
+++ b/controllers/admin/AdminAddressesController.php
@@ -368,20 +368,20 @@ class AdminAddressesControllerCore extends AdminController
         }
 
         // Transform e-mail in id_customer for parent processing
-        if (Validate::isEmail(Tools::getValue('email'))) {
+        if ($id_customer = Tools::getValue('id_customer')) {
+            $customer = new Customer((int) $id_customer);
+            if (Validate::isLoadedObject($customer)) {
+                $_POST['id_customer'] = $customer->id;
+            } else {
+                $this->errors[] = $this->trans('This customer ID is not recognized.', array(), 'Admin.Orderscustomers.Notification');
+            }
+        } elseif (Validate::isEmail(Tools::getValue('email'))) {
             $customer = new Customer();
             $customer->getByEmail(Tools::getValue('email'), null, false);
             if (Validate::isLoadedObject($customer)) {
                 $_POST['id_customer'] = $customer->id;
             } else {
                 $this->errors[] = $this->trans('This email address is not registered.', array(), 'Admin.Orderscustomers.Notification');
-            }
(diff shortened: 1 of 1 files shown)


PierreRambaud:
Hi and thanks for your contribution.
Could you please create an issue to determine and better understand the problem with our Product team?

mBangiev (approved):

PierreRambaud:
@levyn Any news?

levyn:
@PierreRambaud 
I'm sorry, but what are you about?
The issue #16116 was created and linked to the current issue of the.
Is there something missing here?

PierreRambaud (approved):

PierreRambaud:
@levyn Didn't have answer from you, and GitHub doesn't notify us when a comment is edited, sorry :(
It's ok now :)

sarahdib:
Thank you @levyn  for this contribution :)

PierreRambaud:
Thanks @levyn

Issue #15825, opened 2019-10-04
[ORDER PAGE] Merchandise return Tab Order - 8/11


MatShir:
Part of the order page migration and redesign project.

**User Story**

As a merchant, when I check an order, I need to be able to see the returned merchandise interaction.

**Detailed specification**
Specs on the repo prestashop-specs
[Merchandise return Tab
](https://github.com/PrestaShop/prestashop-specs/blob/cc476f1fc0c71c63bd3ba4879131f5163fca05e7/Order%20Page%20View.md#v-tabs-panel)

PR #16084, opened 2019-10-23
Migrate Merchandise returns for Orders view

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migrates Merchandise returns for Orders view
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #15825 
| How to test?  | :warning: Rebuild assets :warning: Merchandise returns block should work the same as in legacy page.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16084)

### DIFF
--- a/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/merchandise_returns.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/merchandise_returns.html.twig
@@ -23,33 +23,41 @@
  * International Registered Trademark & Property of PrestaShop SA
  *#}
 
-{% if orderForViewing.returns.orderReturns is not empty %}
-  <table class="table">
-    <thead>
+{% if not orderForViewing.virtual %}
+  {% if orderForViewing.returns.orderReturns is not empty %}
+    <table class="table">
+      <thead>
       <tr>
         <th>{{ 'Date'|trans({}, 'Admin.Global') }}</th>
         <th>{{ 'Type'|trans({}, 'Admin.Global') }}</th>
         <th>{{ 'Carrier'|trans({}, 'Admin.Shipping.Feature') }}</th>
         <th>{{ 'Tracking number'|trans({}, 'Admin.Shipping.Feature') }}</th>
       </tr>
-    </thead>
-    <tbody>
-    {% for return in orderForViewing.returns.orderReturns %}
-      <tr>
-      <td>{{ return.date|date('Y-m-d H:i:s') }}</td>
-      <td>{{ return.type }}</td>
-      <td>{{ return.stateName }}</td>
-      <td>
-        {% if return.trackingNumber and return.trackingUrl %}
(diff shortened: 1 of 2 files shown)


matks (dismissed):

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/merchandise_returns.html.twig
```diff
@@ -23,33 +23,41 @@
...
+  {% else %}
+    <p class="text-center mb-0">{{ 'No merchandise returned yet'|trans({}, 'Admin.Orderscustomers.Notification') }}</p>
+  {% endif %}
+
+  {% if orderForViewing.shipping.carrierModuleInfo %}
+    {{ orderForViewing.shipping.carrierModuleInfo|raw }}
+  {% endif %}
 {% else %}
-  <p class="text-center mb-0">{{ 'No merchandise returned yet'|trans({}, 'Admin.Orderscustomers.Notification') }}</p>
+  <p class="text-center mb-0">{{ 'Merchandise returns are not applicable for virtual orders'|trans({}, 'Admin.Orderscustomers.Notification') }}</p>
```
LouiseBonnard, line 62: Is this notification displayed when an order contains only virtual products or when it contains some, among other standard products?

sarjon:
Rebased.

jolelievre (dismissed):

Robin-Fischer-PS:
Hi @sarjon , thanks for the PR ! It's OK for me :smile: 

However, I have a suggestion for our UX and product friends :stuck_out_tongue: : 

@MatShir @TristanLDD : 
The merchandise return block has two columns that have inconsistent labels (on Legacy and on Migrated page, and in the spec) : 
- "Carrier" which in fact display the status
- "Tracking number" which in fact display the Return ID

I think it could be a good idea to fix this with the migration, what do you think ?

Here a screenshot : 
pr16084 ux

Thanks !

TristanLDD:
Hi @Robin-Fischer-PS, 
I agree with you on the wording changes, Status and Return ID are much more relevant indeed.
However, considering the many last-minutes changes on order page migration and depending on workload estimations I don't know if we can prioritize it  (@sarjon @MatShir ) 
@LouiseBonnard what do you think of these wording suggestions?

sarjon:
Just add suggestions for wording and we can easily merge it. :)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/merchandise_returns.html.twig
Robin-Fischer-PS: This notification is only displayed on order page when an order contains only virtual products. When it contains some virtual and standard products, it is not displayed.

matks (approved):

matks:
Thank you @sarjon

PR #16088, opened 2019-10-23
Disable Customer group selection checkbox when single customer is set…

### LINKED ISSUE #15938: Badly configured vouchers can result in them being displayed by customer although they should not
**Describe the bug**
In the Back-Office, when a cart rule is created, and has **both** a customer condition (= should be available only for customer A) AND a customer group condition (= shoulb be available only for customer group B) then this voucher visibility will be wrongly computed and will be displayed to customer.

**To Reproduce**
1. Create a new customer group GA
2. Create a new customer C1 with Group access only GA
3. Create a new customer C2 with Group access only Customer & visitor & Guest
4. Create a cart rule CR1
4.1 this cart rule is only limited to a single customer ( pub@prestashop.com)
4.2 this cart rule have a restriction: customer group selection
4.3 unselected groups: GA
4.4 selected groups : Customer / visitor / Guest
5. Create another cart rule CR2 a restriction: customer group selection (selected groups : Customer / visitor / Guest)

In FO if we are signed with the customer C2 the cart rule CR1 is displayed but not applied ( In fact it should not be displayed because it is just limited to pub@prestashop.com)

PS: In the list voucher in FO, if we are signed with the customer C2 the cart rule is displayed =>, In fact, it should not be displayed because it is just limited to pub@prestashop.com)

**Screenshots**
Video record
https://drive.google.com/file/d/1lUySJIMwDVZJ_UxFsL548g8ZHP-cw8gL/view

**Additional information**
PrestaShop version: develop

atomiix:
| Questions     | Answers
| Branch?       | develop
| Description?  | Disable Customer group selection checkbox when single customer is set and disable single customer field when Customer group selection is checked
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15938
| How to test?  | In Sell > Catalog > Discounts, add a new Cart Rule, in conditions, set a single customer: Customer group selection checkbox should be disabled. Remove single customer, check Customer group selection: single customer field should be disabled

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16088)

### DIFF
--- a/admin-dev/themes/default/template/controllers/cart_rules/form.js
+++ b/admin-dev/themes/default/template/controllers/cart_rules/form.js
@@ -105,14 +105,22 @@ var restrictions = new Array('country', 'carrier', 'group', 'cart_rule', 'shop')
 for (i in restrictions)
 {
 	toggleCartRuleFilter($('#' + restrictions[i] + '_restriction'));
-	$('#' + restrictions[i] + '_restriction').click(function() {toggleCartRuleFilter(this);});
+	$('#' + restrictions[i] + '_restriction').change(function() {toggleCartRuleFilter(this);});
 	$('#' + restrictions[i] + '_select_remove').click(function() {removeCartRuleOption(this);});
 	$('#' + restrictions[i] + '_select_add').click(function() {addCartRuleOption(this);});
 }
 
 toggleCartRuleFilter($('#product_restriction'));
 
-$('#product_restriction').click(function() {
+$('#group_restriction').change(function() {
+  $('#customerFilter').prop('disabled', $(this).prop('checked'));
+}).change();
+
+$('#customerFilter').on('change keyup', function() {
+    $('#group_restriction').prop('disabled', $(this).val() !== '');
+}).change();
+
+$('#product_restriction').change(function() {
 	toggleCartRuleFilter(this);
 
 	if ($(this).prop('checked'))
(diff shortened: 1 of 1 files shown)


PierreRambaud (dismissed):

Robin-Fischer-PS:
Hi @atomiix ,

I've detected a defect on your fix : 
When editing a cart rule (which was previously set to single customer or to customer group), the other choice (customer group or single customer) is no longer disabled.
So it's still possible to select both single customer and customer group when editing.

Here you can find a screen record of the workflow : 
https://drive.google.com/file/d/1t0fhSfpkX2rBlrFmh8ZdTB4TNu_YZNWb/view?usp=sharing

Thanks :)

matks (dismissed):

admin-dev/themes/default/template/controllers/cart_rules/form.js
```diff
@@ -105,14 +105,22 @@ var restrictions = new Array('country', 'carrier', 'group', 'cart_rule', 'shop')
...
 
 toggleCartRuleFilter($('#product_restriction'));
 
-$('#product_restriction').click(function() {
+$('#group_restriction').change(function() {
+  $('#customerFilter').prop('disabled', $(this).prop('checked'));
+}).change();
+
+$('#customerFilter').on('change keyup', function() {
+    $('#group_restriction').prop('disabled', $(this).val() !== "");
```
PierreRambaud:
```suggestion
    $('#group_restriction').prop('disabled', $(this).val() !== '');
```

matks (dismissed):

matks (approved):

matks:
Thank you @atomiix

PR #16093, opened 2019-10-23
Add hooks on Presenter classes

djbuch:
| Questions     | Answers
| Branch?       | develop
| Description?  | Allow to add additionnal infos to each presenters thanks to new hooks
| Type?         | new feature
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | These are just hooks allowing module developpers to add additionnal infos to presented objects

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16093)

### DIFF
--- a/src/Adapter/Presenter/AbstractLazyArray.php
+++ b/src/Adapter/Presenter/AbstractLazyArray.php
@@ -143,6 +143,28 @@ abstract class AbstractLazyArray implements Iterator, ArrayAccess, Countable, Js
     }
 
     /**
+     * Set array key and callable from $array into the LazyArray.
+     * The callables are arrays where the elements are corresponding to the parameters of the call_user_func_array function
+     *
+     * @param array $array
+     */
+    public function appendLazyArray($array)
+    {
+        foreach ($array as $key => $value) {
+            // do not override any existing method
+            if (!$this->arrayAccessList->offsetExists($key)) {
+                $this->arrayAccessList->offsetSet(
+                    $key,
+                    array(
+                        'type' => 'callable',
+                        'value' => $value,
+                    )
+                );
+            }
+        }
+    }
+
+    /**
      * The number of keys defined into the lazyArray.
      *
      * @return int
@@ -239,6 +261,12 @@ abstract class AbstractLazyArray implements Iterator, ArrayAccess, Countable, Js
                 if (!isset($this->methodCacheResults[$index])) {
                     $methodName = $this->arrayAccessList[$index]['value'];
                     $this->methodCacheResults[$index] = $this->{$methodName}();
+                }
+                $result = $this->methodCacheResults[$index];
+            } elseif ($this->arrayAccessList[$index]['type'] === 'callable') {
+                if (!isset($this->methodCacheResults[$index])) {
+                    $callable = $this->arrayAccessList[$index]['value'];
+                    $this->methodCacheResults[$index] = call_user_func($callable[0], isset($callable[1]) ? $callable[1] : null);
                 }
                 $result = $this->methodCacheResults[$index];
             } else { // if the index is associated with a value, just return the value
--- a/src/Adapter/Presenter/AbstractLazyArray.php
+++ b/src/Adapter/Presenter/AbstractLazyArray.php
@@ -266,7 +266,7 @@ abstract class AbstractLazyArray implements Iterator, ArrayAccess, Countable, Js
             } elseif ($this->arrayAccessList[$index]['type'] === 'callable') {
                 if (!isset($this->methodCacheResults[$index])) {
                     $callable = $this->arrayAccessList[$index]['value'];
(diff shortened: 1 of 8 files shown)


kpodemski:
What about performance of such a solution? the more we have presented objects the more hooks are being triggered, right?

djbuch:
If there is no module hooked to the hook, there will be just very marignal performance change, since module list is cached for each hook.

If there is one hooked, then I think this is the best way for changing presented elements.

PierreRambaud:
This is exactly what we need for facetedsearch. ping @matthieu-rolland 
But, It could be cool to have a `before` and an `after` hook on each presenter :) Wdyt?

src/Adapter/Presenter/AbstractLazyArray.php
```diff
@@ -241,6 +263,12 @@ public function offsetGet($index)
                     $this->methodCacheResults[$index] = $this->{$methodName}();
                 }
                 $result = $this->methodCacheResults[$index];
+            } elseif ($this->arrayAccessList[$index]['type'] === 'callable') {
+                if (!isset($this->methodCacheResults[$index])) {
+                    $callable = $this->arrayAccessList[$index]['value'];
+                    $this->methodCacheResults[$index] = call_user_func($callable[0], isset($callable[1]) ? $callable[1] : null);
```
PierreRambaud:
```suggestion
                    $this->methodCacheResults[$index] = call_user_func($callable[0], $callable[1] ?: null);
```

PierreRambaud (changes requested):
Hooks names must change: https://devdocs.prestashop.com/1.7/modules/concepts/hooks/#naming-scheme

djbuch:
Hello, I agree the hook names should change, but what to choose, action or display ? Because it is quite at the crossroads of both behaviors

src/Adapter/Presenter/AbstractLazyArray.php
```diff
@@ -241,6 +263,12 @@ public function offsetGet($index)
                     $this->methodCacheResults[$index] = $this->{$methodName}();
                 }
                 $result = $this->methodCacheResults[$index];
+            } elseif ($this->arrayAccessList[$index]['type'] === 'callable') {
+                if (!isset($this->methodCacheResults[$index])) {
+                    $callable = $this->arrayAccessList[$index]['value'];
+                    $this->methodCacheResults[$index] = call_user_func($callable[0], $callable[1] ?: null);
```
PierreRambaud: My bad, wrong notation :D This one is better
```suggestion
                    $this->methodCacheResults[$index] = call_user_func($callable[0], $callable[1] ?? null);
```

eternoendless:
> I agree the hook names should change, but what to choose, action or display ? Because it is quite at the crossroads of both behaviors

If a hook returns HTML content, then it should be `display<...>`, whereas all other hooks should to be called `action<...>`.

djbuch:
OK thanks, I changed the names

src/Adapter/Presenter/AbstractLazyArray.php
```diff
@@ -142,6 +142,28 @@ public function appendArray($array)
...
+    /**
+     * Set array key and callable from $array into the LazyArray.
+     * The callables are arrays where the elements are corresponding to the parameters of the call_user_func_array function
+     *
+     * @param array $array
+     */
+    public function appendLazyArray($array)
+    {
+        foreach ($array as $key => $value) {
+            // do not override any existing method
```
PierreRambaud, line 154: I can't find where this method is used? :thinking:

eternoendless:
Looks like this PR may need a rebase, I see lots commits with a generic title "Update Somefile.php" plus a merge commit, and Travis is failing for some reason.

src/Adapter/Presenter/AbstractLazyArray.php
djbuch: This method is not used in the PR but is an addition to allow module developper to add "Lazy" values to an lazy array, like I showed it here : https://github.com/PrestaShop/PrestaShop/pull/16093/commits/9c32593c871a6d7e41eb1e87195ef4650c191467

PierreRambaud:
Pull request has been rebased and squashed.
And I fix the code for sanity tests.

PierreRambaud (approved):

PierreRambaud:
@djbuch Could you provide us a modules to tests all hooks? :)

matks:
@djbuch Hi ! I'm very sorry I just found this PR and it looks like I already implemented it myself (see https://github.com/PrestaShop/PrestaShop/pull/17203 and https://github.com/PrestaShop/PrestaShop/pull/17419/). I was unaware you submitted this PR.

Sorry for this 😞

djbuch:
Hello @matks I saw your implementation, but it lacks some elements, that where in mine.
For example the 'cart' param in the actionPresentCart hook in order to get the cart that is being presented (not necessarily the one of the context).

What do you think of this ?

djbuch:
Hello @matks have you seen my last comment ?

PR #16104, opened 2019-10-24
Update order status colors

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | See ticket.
| Type?         | improvement
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/15228
| How to test?  | After installing/upgrading PS, order colors should change. You must use the last version of autoupgrade 4.10.0

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16104)

### DIFF
--- a/install-dev/upgrade/php/ps_1770_update_order_status_colors.php
+++ b/install-dev/upgrade/php/ps_1770_update_order_status_colors.php
@@ -59,8 +59,24 @@ function ps_1770_update_order_status_colors() {
     foreach ($statusColorMap as $color => $statuses) {
         foreach ($statuses as $statusId) {
             Db::getInstance()->execute(
-                'UPDATE `'._DB_PREFIX_.'order_state` SET `color` = ' . pSQL($color) . ' WHERE `id_order_state` = ' . (int) $statusId
+                'UPDATE `'._DB_PREFIX_.'order_state` SET `color` = "' . pSQL($color) . '" WHERE `id_order_state` = ' . (int) $statusId
             );
         }
+    }
+
+    // Some of the statuses can be deduced by their parameters, this allows to update the modules status colors
+    $statusColorConditions = [
+        OrderStatusColor::ACCEPTED_PAYMENT => ['paid' => 1, 'shipped' => 0],
+        OrderStatusColor::COMPLETED => ['paid' => 1, 'shipped' => 1],
+        OrderStatusColor::AWAITING_PAYMENT => ['color' => '#4169E1'], // Former color of awaiting payment
+    ];
+    foreach ($statusColorConditions as $color => $conditions) {
+        $whereCondition = ' WHERE 1';
+        foreach ($conditions as $field => $expectedValue) {
+            $whereCondition .= ' AND `' . $field . '` = "' . pSQL($expectedValue) . '"';
+        }
+        Db::getInstance()->execute(
+            'UPDATE `'._DB_PREFIX_.'order_state` SET `color` = "' . pSQL($color) . '"' . $whereCondition
+        );
     }
 }
--- a/install-dev/upgrade/php/ps_1770_update_order_status_colors.php
+++ b/install-dev/upgrade/php/ps_1770_update_order_status_colors.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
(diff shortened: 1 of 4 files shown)


sarjon:
Hey @TristanLDD, here's how new colors look like

colors

Legacy has this feature, if background is bright it makes text dark, otherwise text is white. Brightness is calculated using some formula.

install-dev/upgrade/sql/1.7.7.0.sql
```diff
@@ -5,6 +5,7 @@ ALTER TABLE `PREFIX_search_word` MODIFY `word` VARCHAR(30) NOT NULL;
 INSERT INTO `PREFIX_configuration` (`name`, `value`, `date_add`, `date_upd`) VALUES ('PS_DISPLAY_MANUFACTURERS', '1', NOW(), NOW());
 
 /* PHP:ps_1770_preset_tab_enabled(); */;
+/* PHP:ps_1770_update_order_status_colors(); */;
```
sarjon: I saw methods are "registered" this way, I assume they are executed somehow? :thinking:

Progi1984 (dismissed):
Need second review. I'm not sure for the call from SQL file.

install-dev/upgrade/php/ps_1770_update_order_status_colors.php
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+use PrestaShop\PrestaShop\Core\Domain\Order\Status\OrderStatusColor;
+
+/**
+ * Updates order status colors according to new color schema
+ */
+function ps_1770_update_order_status_colors() {
+    $statusColorMap = [
+        OrderStatusColor::AWAITING_PAYMENT => Configuration::getMultiple([
```
PierreRambaud, line 34: Why not using const defined in config.inc.php?

sarjon: Because of deprecation, see

https://github.com/PrestaShop/PrestaShop/blob/develop/config/config.inc.php#L242

PierreRambaud: Oh, you're right, didn't saw it. Could we have class const at least? or is it too overkill?

install-dev/upgrade/sql/1.7.7.0.sql
PierreRambaud: yes, with this code you call the script `ps_1770_update_order_status_colors.php` and the function with the same name :)

sarjon: I'm confused, I don't know if that's cool or just a hack. :boom:

install-dev/upgrade/php/ps_1770_update_order_status_colors.php
sarjon: Hmmm, I guess it would make sense to have class constants for payment status config names, but it's beyond the scope of this PR imo.

PierreRambaud (dismissed):

Robin-Fischer-PS:
Hi @jolelievre @matks !

Some statuses are not displayed correctly :

- Partial refund should be "green HEX #2ECC71"
- Waiting capture (if it's a real status) should be "dark blue HEX #34219E"
- Waiting for Local Payment Method should be "dark blue HEX #34219E"
- Waiting for PayPal Payment should be "dark blue HEX #34219E"
- Waiting for Credit Card Payment should be "dark blue HEX #34219E"

Also, with a 1-click Upgrade, it did not worked, don't really know why.

Thanks

jolelievre:
Hi @Robin-Fischer-PS 

The problem here comes from the `ps_checkout` module, the order states you mentioned come from this module which insert them with his colors (matching legacy ones). The module should be updated to change these colors.

However I made some changes so now on upgrade almost all statuses are correctly updated, based on their parameters I was able to identify some of them and assigne the correct color. The only left one is `Partial refund` which can't be deduced and will have to be dealt with in the module.

I noticed an error on update on the menu, but it's not related to this PR, the problem comes the the new `enabled` field on Tabs, to test your back office correctly after the upgrade, either type the urls manually or you can update the database content manually and set the column `enabled` to 1 in all the `ps_tab` table.

I opened a dedicated issue for this problem: https://github.com/PrestaShop/PrestaShop/issues/16379

Progi1984 (dismissed):

matks (dismissed):

PierreRambaud (dismissed):

colinegin:
Hello @jolelievre and @Robin-Fischer-PS ,

FYI I've told @Darmona about these changes, he is going to apply the same changes in the PS Checkout module for more consistency. There is no ETA, but it will be done !

jolelievre:
I believe this PR can't be tested before #16379 is merged, it will also need a rebase for upgrade scripts and besides the ending `;` is missing in the PHP script calls in the sql file

matks (approved):

Robin-Fischer-PS:
Hi @jolelievre ,

With autoupgrade, the status colors are now OK (excepted Partial Refund as you said), so it's :heavy_check_mark: for QA !

Thanks

matks:
Thank you @sarjon and @jolelievre

PR #16126, opened 2019-10-25
Change checkout button wording

### LINKED ISSUE #15854: Replace the wording of the checkout button
**Is your feature request related to a problem? Please describe**
Issue described by @rdy4ever [here](https://github.com/PrestaShop/PrestaShop/issues/15406#issuecomment-531472765): "If you ask me, _Order with an obligation to pay_ button scares more customers than this kind of checkout process. The text on the button should be just _Complete order_ or _Finish order_. The term 'obligation' has no place here.

I’m pretty sure you cannot force anyone to pay the order just because they use your check-out - legally speaking. Until the very last moment when they actually pay for the merchandise (cash, card, bank wire, etc.) they are at the liberty to default on the payment and/or refuse to receive the order. And if they have any doubt about their rights, that text on the button will only make them go away."

It turns out we can modify this wording, I've checked the legal team and it is not mandatory; the only requirement is that the client must be aware of his/her obligation to pay but it doesn't have to be written inside the button, it can be displayed in the terms & conditions pages. But that is to be mentioned on the merchant's side, not ours.

So I suggest _Place order_ (localized in both _Shop.Theme.Actions_ and _Shop.Theme.Checkout_ domains) instead. :-)

atomiix:
| Questions     | Answers
| Branch?       | develop
| Description?  | Change the checkout button wording
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15854
| How to test?  | Order a product, see the last button in the checkout process when using check or bank wire payment method

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16126)

### DIFF
--- a/themes/classic/templates/checkout/_partials/steps/payment.tpl
+++ b/themes/classic/templates/checkout/_partials/steps/payment.tpl
@@ -128,7 +128,7 @@
   <div id="payment-confirmation">
     <div class="ps-shown-by-js">
       <button type="submit" {if !$selected_payment_option} disabled {/if} class="btn btn-primary center-block">
-        {l s='Order with an obligation to pay' d='Shop.Theme.Checkout'}
+        {l s='Place order' d='Shop.Theme.Checkout'}
       </button>
       {if $show_final_summary}
         <article class="alert alert-danger mt-2 js-alert-payment-conditions" role="alert" data-alert="danger">


Progi1984 (approved):

PierreRambaud (approved):

matks:
Thank you @atomiix

PR #16144, opened 2019-10-28
Migrate hooks for new Order view page

### LINKED ISSUE #16523: [ORDER PAGE] New hooks
**New hooks mapping**
We will add new hooks on the top with the other button. This hooks will force the module to have the same design as the other one.

68310376-30ba9600-00b0-11ea-86f3-52e5dffa2435

**Known limitation**
Mapping requires that "when there is not enough space for all buttons to be rendered, a dropdown list appears instead and buttons are put inside" => this is a responsive feature, need someone experienced in front (not @matks 😛 )

**Workaround for limitation instead:**

if 1 or 2 buttons are added (like module test) they are displayed
if more than 2 buttons are added, it displays 1st button THEN next buttons are in dropdown (see screenshot below)
=> these 2 usecases can be seen with the test module, when you refresh there is a 50% luck you have usecase 1 or 2

sarjon:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migrates olds hooks from legacy page to new one. See details below
| Type?         | improvement
| Category?     | CO
| BC breaks?    | yes, 2 hooks have been moved and renamed but aliases keep compatibility but some are really not compatible any more because of different parameters
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/16523
| How to test?  | See below

### New hooks mapping

See below https://github.com/PrestaShop/PrestaShop/pull/16144#issuecomment-550356204

### How to test

1. Build assets. Reinstall shop (SQL file modified).
2. Go to modules/folder in CLI
3. Copy [test module](https://github.com/matks/orderhooks) there: `git clone git@github.com:matks/orderhooks.git`
4. Go to BO "Module catalog" and search for "orderhook"
5. Install "orderhook" test module
6. Go to 1 order view page, example `/admin-dev/index.php/sell/orders/orders/1/view`
7. See how hooks are displayed and compare to mapping
8. Refresh. I added a "random 50%" to display the 2 usecases explained below

### Known limitation

Mapping requires that "when there is not enough space for all buttons to be rendered, a dropdown list appears instead and buttons are put inside". I cant do that as my integration skills are very low.

What I did instead:
- if 1 or 2 buttons are added (like module test) they are displayed
- if more than 2 buttons are added, I display 1st button THEN next buttons are in dropdown (see screenshot below)

=> these 2 usecases can be seen with the test module, when you refresh there is a 50% luck you have usecase 1 or 2

### Screenshots

Capture d’écran 2019-11-22 à 17 11 13
Capture d’écran 2019-11-22 à 17 11 21

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16144)


src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
```diff
@@ -45,6 +45,9 @@
         {{ 'Merchandise returns'|trans({}, 'Admin.Orderscustomers.Feature') }} ({{ orderForViewing.returns.orderReturns|length }})
       </a>
     </li>
+
+    {{ renderhook('displayAdminOrderTabShip', {'id_order': orderForViewing.id}) }}
+    {{ renderhook('displayAdminOrderTabOrder', {'id_order': orderForViewing.id}) }}
```
sarjon: There is kind of BC break with these 2 hooks (and 2 related hooks below). In legacy, three params where passed to these hooks: `order` object model, order products array and `customer` object model. 

See https://github.com/PrestaShop/PrestaShop/blob/develop/controllers/admin/AdminOrdersController.php#L1932

In new page, I pass only order id, but it's enough to retrieve all data about order (customer, products, order itself).

@matks wdty? Is it ok or should we keep legacy params?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
```diff
@@ -47,4 +47,6 @@
       {{ form_rest(updateOrderStatusActionBarForm) }}
     </div>
   {{ form_end(updateOrderStatusActionBarForm) }}
+
+  {{ renderhook('displayBackOfficeOrderActions', {'id_order': orderForViewing.id}) }}
```
sarjon: Regarding this hook, I don't know if we can implement "More actions", see below.

actions

PrestaShop does not define clear API for this, so modules can return anything (be it a button, multiple buttons or some other elements). What we know is that we get string in the end, but we don't know what's inside.

@matks your thoughts?

sarjon:
Hi @TristanLDD, regarding "Print" button,

print

why is it highlighted as a hook? I could not find any hooks around it, does it have other meaning? :)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
matks: Calling some friends developing modules on this one 😄 
@v4lux @apacios @Quetzacoalt91 wdyt ? is it better to keep as much as possible previous hook payload, or is it better to make it faster and lighter through the migration ?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
matks: Expected behavior is "if more than 2 modules are hooked on this button, then we show the 1st 2 buttons and the following buttons are put inside a drop-down "more actions" " is that right ?

Maybe we should drop the drop-down and go for a basic toggle

sarjon: > "if more than 2 modules are hooked on this button, then we show.......

Yup.

> Maybe we should drop the drop-down and go for a basic toggle

I'm afraid it might add unnecessary limitations, because as mentioned before, it can be more than a buttons, maybe form with input or anything else. :thinking:

matks: By toggle I meant this: "when I click on "more actions", it shows all previously hidden blocks added by hook modules. So they can be displayed in any way they want.

I think the toggle is more compatible that a dropdown, isn't it ? 🤔

sarjon: Ohhh, but I don't know tbh, let's wait for @TristanLDD feedback. :slightly_smiling_face:

TristanLDD: I think the question here is about how we want the modules to integrate this page: 
the name of the hook is "display back office order actions" so in my opinion, it should be a hook for action buttons only. If modules want to bring a form with inputs or other components there are more relevant hooks to do that, like display admin order.

sarjon: @TristanLDD thanks for reply, but PrestaShop itself show example of using form here, see https://prnt.sc/pq17mi so I think modules may have similar use cases.

Either way, we get HTML content from modules and it's nearly impossible for us to enforce some structure to it, so we cannot know whether modules returns 1 button or 10 buttons or maybe it's not button at all. :confused: 

So, wdyt? What should we do here? @matks your thoughts?

MatShir:
Here the new hooks mapping! :tada:  
New hooks

As discussed with @matks, we will add new hooks on the top with the other button. This hooks will force the module to have the same design as the other one.

Also, because there is no much space left and to not have lines of actions buttons, the ideal should be that it had them in a drop-list when there is not enough space.
Capture d’écran 2019-11-06 à 16 02 35

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
jolelievre: I can see the interest to have "smaller" hooks, but if the previous hooks had those three variables we' shouldn't change them Besides what would appear as a performance gain (only give the order id) might have drawbacks in the end (each module would have to request the same order)

src/PrestaShopBundle/Controller/Admin/Sell/Order/ActionsBarButton.php
```diff
@@ -0,0 +1,92 @@
...
+ *
+ */
+class ActionsBarButton
+{
+    /**
+     * Complete this property to add extra CSS classes
+     *
+     * @var string
+     */
+    public $class;
```
matks: Public properties to make them as easy to use as possible for modules

install-dev/data/xml/hook_alias.xml
```diff
@@ -337,5 +337,14 @@
       <alias>actionAfterDeleteProductInCart</alias>
       <name>actionDeleteProductInCartAfter</name>
     </hook_alias>
+
+    <hook_alias id="adminOrderLeft">
```
matks: Hooks have moved on Edit an Order page so I gave them new name that suit better for their positioning. But I kept backward compatibility with hook aliases.

matks (dismissed):

matks (dismissed):

matks (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Order/ActionsBarButton.php
PierreRambaud: Use accessor instead public property

src/PrestaShopBundle/Controller/Admin/Sell/Order/ActionsBarButtonsCollection.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public function isEmpty()
+    {
+        return empty($this->buttons);
+    }
+
+    /**
+     * @param ActionsBarButton $button
+     */
+    public function addButton(ActionsBarButton $button)
```
PierreRambaud: Must be an interface, same for the collection
```suggestion
    public function addButton(ActionsBarButtonInterface $button)
```

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -287,6 +287,18 @@ public function viewAction(int $orderId, Request $request): Response
             'order_id' => $orderId,
         ]);
 
+        $backOfficeOrderButtons = new ActionsBarButtonsCollection();
```
PierreRambaud, line 332: Must be a service or something extendable

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
```diff
@@ -62,11 +64,79 @@
...
+           class="btn btn-action">
           <i class="material-icons">local_shipping</i>
           {{ 'View delivery slip'|trans({}, 'Admin.Orderscustomers.Feature') }}
         </a>
       </div>
     </form>
   {% endif %}
+
+  {# Rendering of hook actionGetBackOfficeOrderButtons, see ActionsBarButtonsCollection #}
+  {% if (backOfficeOrderButtons.isEmpty() != true) %}
```
PierreRambaud: If this kind of block is going to be on many pages, it should be interesting to create a macro to do it :)

src/PrestaShopBundle/Controller/Admin/Sell/Order/ActionsBarButton.php
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Complete this property to add extra properties to <button> tag
+     * Each item of the array will be created as a tag
+     *
+     * Example: if $tags is ['href' => '/a/b', 'alt' => 'link']
+     * Then following tags will be added: href="/a/b", alt="link"
+     *
+     * @var string[]
+     */
+    public $tags;
```
jolelievre: I don't think it's tags, more like `attributes` no? Or `properties` at least

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
jolelievre: So we forget about the previous  behaviour? Assumed BC?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
```diff
@@ -31,27 +31,29 @@
...
+    <button class="btn btn-action ml-2"
+            id="update_order_status_action_btn"
+            disabled
+            data-order-status-id="{{ orderForViewing.history.currentOrderStatusId }}"
+    >
+      {{ 'Update status'|trans({}, 'Admin.Orderscustomers.Feature') }}
+    </button>
+  </div>
+
+  <div class="d-none">
```
jolelievre, line 47: Mmm, if all the rest is in a `d-none` div then you can never add extra fields? Or only invisible ones..

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
matks: I dont think 🤔 this is a collection of buttons, intended to be propagated through a hook to all modules which wants to add its button to be displayed. It's like DoctrineCollection, you would not make it a service, would you ?

matks: It's like an array but I did want to type-hint it

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
matks: Only here so far 🤔

src/PrestaShopBundle/Controller/Admin/Sell/Order/ActionsBarButtonsCollection.php
matks: I thought about it, but wouldn't it be a little overkill ? It's just some button with 3 properties ... almost like a VO. We dont put interfaces on VO, do we ?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
```diff
@@ -62,11 +64,79 @@
...
+                    data-toggle="dropdown"
+                    aria-haspopup="true" aria-expanded="false">
+              <i class="material-icons form-error-icon">more_horiz</i>
+
+              {{ 'More actions'|trans({}, 'Admin.Global') }}
+            </button>
+
+            <div class="dropdown-menu" aria-labelledby="dropdownOrderActionBar">
+              {% for button in buttons %}
+                <a class="dropdown-item btn {{ button.class }}"
```
jolelievre: Since the only change between these buttons and when there are only two is the extra class `dropdown-item` maybe you can create a twig template dedicated to render this button You can manage the extra class either through a parameter of the template Or by adding it in twig right before rendering the template `set button.class = button.class ~ ' dropdown-item'`

This would have few advantages, factorize this code, make the `order_actions.html.twig` template shorter and finally allow to override the button template if needed

jolelievre (commented):
I like the buttons feature But I'm wondering unless you add an href on those buttons they are likely to be used as "javascript buttons" then where are you supposed to integrate the javascript part?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: I also think the service definition would be overkill, however there should be an interface for the collection and button class (even though we provide concrete classes)

src/PrestaShopBundle/Controller/Admin/Sell/Order/ActionsBarButton.php
matks: The content is rendered as raw, so you can inject a `<script>` in it. Also you can choose to add the button with this hook and add a JS script in the header page using another hook.

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
PierreRambaud: Agreed with @jolelievre

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
matks: They can be toggled by JS if needed

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
matks: If I remember correctly, `include` statements in Twig are not the best optimized :/ so it would be better to use a macro here. Do you think it's worth it ? I've extracted all the "hook extra buttons" into a dedicated Twig template, is it enough ?

matks:
@PierreRambaud `ActionsBarButtonsCollection` now extends `AbstractTypedCollection` and code uses `AbstractTypedCollection` functions


@jolelievre You suggested using a reusable template for buttons, however the only common part between drop-down actions bar button and regular buttons is this part:
```
        <a class="btn {{ button.class }}"
          {% for propertyName, propertyContent in button.properties %}
            {{ propertyName }}="{{ propertyContent }}"
          {% endfor %}
          >
            {{ button.content|raw }}
        </a>
```
Again, I'm really not convinced a Twig file is necessary to reuse only 3 times, in the same file, such a small structure ? 🤔

matks:
@jolelievre Here is the BC break for hooks:
```
      {hook h="displayAdminOrderLeft" id_order=$order->id}
```
⬇️ 
```
      {{ renderhook('displayAdminOrderMain', {'id_order': orderForViewing.id}) }}
```
and I used a hook alias to ensure modules hooked on `displayAdminOrderLeft` will be hooked on `displayAdminOrderMain`

I feel like I have done all the steps required to maintain as much as possible the compatibilit 🤔

jolelievre:
@MatShir some decision needs to be made about the BC related to these three hooks:
https://github.com/PrestaShop/PrestaShop/pull/16144/files#r339418610

`displayAdminOrderContentShip` `displayAdminOrderTabOrder` `displayAdminOrderTabShip` were previously called in the legacy controller with three arguments:

```
Hook::exec(
                'displayAdminOrderTabShip',
                array(
                    'order' => $order,
                    'products' => $products,
                    'customer' => $customer, )
            )
```

We can't assure the retro compatibility of these arguments as the controller now handles DTO and not ObjectModels any more, so converting the elements in both sides seems complicated and not that useful (at best it would be risky..).

Keeping those three hooks to customize/extend the page however is needed but they can't work the same way they did, the new call simply uses the order_id as a parameter which is lighter and still allows improvement.

So I think we have two solutions:
- keep the same hook names with only one parameter and document the breaking change
- create three new hook names to separate both behavior and clearly state that the previous hooks are no longer called

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
```diff
@@ -47,6 +47,9 @@
         {{ 'Merchandise returns'|trans({}, 'Admin.Orderscustomers.Feature') }} ({{ orderForViewing.returns.orderReturns|length }})
       </a>
     </li>
+
+    {{ renderhook('displayAdminOrderTabShip', {'id_order': orderForViewing.id}) }}
```
jolelievre: Merge these two hooks into a sign `displayAdminOrderTabLink`

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
```diff
@@ -62,5 +65,8 @@
     <div class="tab-pane fade" id="orderReturnsTabContent" role="tabpanel" aria-labelledby="orderReturnsTab">
       {% include '@PrestaShop/Admin/Sell/Order/Order/Blocks/View/merchandise_returns.html.twig' %}
     </div>
+
+    {{ renderhook('displayAdminOrderContentShip', {'id_order': orderForViewing.id}) }}
```
jolelievre: Merge these two hooks into a single `displayAdminOrderTabContent`

jolelievre (changes requested):
Four hooks need to be modified (merged into two)

matks:
I noticed another hook whose name changed: `displayInvoice` is now `displayAdminOrderTop`. Same location: at the top of the order.

New name looks better to me as it's related to hook location and not hook purpose, and it can be used to display other things than an invoice. Wdyt @jolelievre @MatShir ?

jolelievre:
@matks yep seems perfect, it is the same parameter so no BC here, we juste need to add an alias!

MatShir:
Good for me as well :)

matks:
New hooks structure implemented

jolelievre (dismissed):
I approve these new hooks However the upgrade part and the new hooks for installation are missing But I guess we will deal with this once all other PRs are merged I believe we have tools now that help us build all the new hooks from grids and forms @matks Is there already a dedicated issue about this or should we create one?

matks:
> I approve these new hooks However the upgrade part and the new hooks for installation are missing But I guess we will deal with this once all other PRs are merged

Actually I forgot 😅I'm sorry and the tools will not be useful here as these hooks are custom to order page instead of generic

jolelievre:
> > I approve these new hooks However the upgrade part and the new hooks for installation are missing But I guess we will deal with this once all other PRs are merged
> 
> Actually I forgot 😅I'm sorry and the tools will not be useful here as these hooks are custom to order page instead of generic

Oh ok! So I was right to check with you 😛 No problem I'll add them tomorrow morning

jolelievre:
@matks @MatShir I just added the hooks on install and upgrade as it was the last missing piece
But it made me realize two hook names were not consistent with all the others and I suggest we rename them:
`displayBackOfficeOrderActions` -> `displayAdminOrderActions`
`actionGetBackOfficeOrderButtons` -> `actionGetAdminOrderButtons`

Other than that let me know if the description and title of the hooks seem good for you, thanks

sarahdib:
@matks @jolelievre 
There is a small display bug
Capture d’écran 2019-12-10 à 09 27 45

jolelievre:
@sarahdib the PR was not ready for QA but it's no big deal ^^
The bug you found comes from the module which doesn't respect all the css integration rules but it's his responsibility so it doesn't affect the validity of this PR which is to add hooks at the expected places

However since I'm nice I made a few corrections on the module ^^ Here is the PR with the corrections https://github.com/matks/orderhooks/pull/1

PierreRambaud (approved):

admin-dev/themes/new-theme/js/pages/order/view.js
```diff
@@ -55,7 +55,12 @@ $(() => {
   initAddCartRuleFormHandler();
   initAddProductFormHandler();
   initChangeAddressFormHandler();
-  
+  initHookTabs();
+
+  function initHookTabs() {
```
NeOMakinG, line 60: I think you should use a comment if the goal is to explain what your chain does,  exept if your function is used elsewhere to avoid code duplication

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/extra_order_button_actions.html.twig
```diff
@@ -0,0 +1,90 @@
...
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ *#}
+
+{# Rendering of hook actionGetAdminOrderButtons, see ActionsBarButtonsCollection #}
+{% if (backOfficeOrderButtons.isEmpty() != true) %}
```
NeOMakinG, line 27: Prefer strict operator maybe ? Like `!==`, I don't know, if isEmpty isn't returning strict value, then I'm wrong

admin-dev/themes/new-theme/js/pages/order/view.js
jolelievre: The thing is everywhere else in the file we have single use functions `initAddCartRuleFormHandler` `handlePaymentDetailsToggle` And as they say "in rome do as the romans do" so I don't want to start a new style in this file for such a small change It sure is not the best and this could justify to refacto it but it's not the purpose of this PR

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/extra_order_button_actions.html.twig
jolelievre: I don't think strict operators exist in twig

matks: I confirm, there is no Twig equivalent fro `!==` although it seems there is one for `===` which is `same as`
https://www.branchcms.com/learn/docs/developer/twig/operators

matks (approved):

matks:
@sarahdib Test module has been updated in order to allow you to validate this PR 😉

PierreRambaud:
Thanks everyone

PR #16172, opened 2019-10-29
Remove constraint on siret in CustomerType

### LINKED ISSUE #16167: Error when we add customer (with B2B enabled) if there is a specific character in field SIRET
Hi there all!
I have an issue that make me crazy. I need to know how prestashop 1.7.6 do validation of SIRET field into BackOffice (RUT in chilean).
In 1.7.5 version is Ok, but in the new version prestashop show me "content field is invalid", anyone knows where is the line to change the validation in 1.7.6 version?, I need complete this data with a "-" like "12345678-K"
errorpng

Cheers

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Fix error when adding a siret with a non numeric character in customer form
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16167
| How to test?  | In BO > Customer > Add a customer: fill siret field with this value: **12345678-K**, you should be able to save your modification. Please check that the SIRET input is still working in a relevant manner with B2B mode.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16172)

### DIFF
--- a/src/PrestaShopBundle/Form/Admin/Sell/Customer/CustomerType.php
+++ b/src/PrestaShopBundle/Form/Admin/Sell/Customer/CustomerType.php
@@ -202,12 +202,6 @@ class CustomerType extends AbstractType
                 ])
                 ->add('siret_code', TextType::class, [
                     'required' => false,
-                    'constraints' => [
-                        new Type([
-                            'type' => 'numeric',
-                            'message' => $this->trans('This field is invalid', [], 'Admin.Notifications.Error'),
-                        ]),
-                    ],
                 ])
                 ->add('ape_code', TextType::class, [
                     'required' => false,


matthieu-rolland:
Hello @colinegin , @marionf 

The siret field was expecting a fully numeric value, this is why the "-" would not be accepted.
Also, today this field is not required (even in B2B mode).

The thing is, I'm not sure we have a format that would work for siret codes in every countries, so I removed any constraint on this field, since it's not required anyway.

What do you think ? Is it ok for you ?

matks (approved):

PR #16175, opened 2019-10-29
Fix label not in place on BO products page in Filter by categories

### LINKED ISSUE #16106: Display regression on BO product page with the filter by categories
Display regression (develop) on BO product page with the filter by categories

Develop:

capture d'écran_2108

1.7.6.1:

capture d'écran_2109

atomiix:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix label not in place on BO products page in Filter by categories
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16106
| How to test?  | In Catalog > Products, click on  "Filter by categories"

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16175)


Progi1984 (approved):

admin-dev/themes/new-theme/scss/components/_category_tree.scss
```diff
@@ -91,6 +91,13 @@ ul.category-tree {
     }
   }
 
+  label.form-control-label {
+    color: $gray-medium;
+    width: 100%;
+    padding-left: 15px;
```
Progi1984, line 97: Please prefer use rem rather than px

Progi1984 (commented):
Small change

admin-dev/themes/new-theme/scss/components/_category_tree.scss
atomiix: Yes, I thought about doing that but as the list just below this label is placed in a `.col-sm` div which is itself padded by 15px I thought I should keep it that way.
Wdyt?

Progi1984: Ok. Logical.

Progi1984 (approved):

matks:
Thank you @atomiix

PR #16176, opened 2019-10-29
Increase maximum hook length to 191 characters

### LINKED ISSUE #16169: An error occurs when trying to register a hook whose name is longer than 64 characters
**Describe the bug**
Error!
Exception thrown by module on reset. Error when resetting module. The length of property Hook->name is currently 65 chars. It must be between 0 and 64 chars.

**To Reproduce**
Hook registration in the module.
$this->registerHook('actionBeforeAjaxDieOrderControllerdisplayAjaxselectDeliveryOption')

**Screenshots**
image

**Additional information**
PrestaShop version: 1.7.6.0
PHP version: 7.1

eternoendless:
| Questions     | Answers
| Branch?       | develop
| Description?  | Hook names were unnecessarily fixed at a maximum of 64 characters, which could be especially problematic for hooks whose names are dynamically generated (see fixed issue). This change simply increases that limit to 191 (maximum allowed by utf8mb4).
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16169
| How to test?  | Install or update PS, then install a module that simply registers a hook like this: `$this->registerHook('actionBeforeAjaxDieOrderControllerdisplayAjaxselectDeliveryOption')`

This PR also fixes an invalid SQL statement in the update script.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16176)

### DIFF
--- a/classes/Hook.php
+++ b/classes/Hook.php
@@ -62,7 +62,7 @@ class HookCore extends ObjectModel
         'table' => 'hook',
         'primary' => 'id_hook',
         'fields' => array(
-            'name' => array('type' => self::TYPE_STRING, 'validate' => 'isHookName', 'required' => true, 'size' => 255),
+            'name' => array('type' => self::TYPE_STRING, 'validate' => 'isHookName', 'required' => true, 'size' => 191),
             'title' => array('type' => self::TYPE_STRING, 'validate' => 'isGenericName'),
             'description' => array('type' => self::TYPE_HTML, 'validate' => 'isCleanHtml'),
             'position' => array('type' => self::TYPE_BOOL, 'validate' => 'isBool'),
--- a/classes/Hook.php
+++ b/classes/Hook.php
@@ -62,7 +62,7 @@ class HookCore extends ObjectModel
         'table' => 'hook',
         'primary' => 'id_hook',
         'fields' => array(
-            'name' => array('type' => self::TYPE_STRING, 'validate' => 'isHookName', 'required' => true, 'size' => 64),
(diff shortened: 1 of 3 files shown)


Progi1984 (dismissed):

atomiix (dismissed):

jolelievre (dismissed):

classes/Hook.php
```diff
@@ -62,7 +62,7 @@ class HookCore extends ObjectModel
         'table' => 'hook',
         'primary' => 'id_hook',
         'fields' => array(
-            'name' => array('type' => self::TYPE_STRING, 'validate' => 'isHookName', 'required' => true, 'size' => 64),
+            'name' => array('type' => self::TYPE_STRING, 'validate' => 'isHookName', 'required' => true, 'size' => 255),
```
PierreRambaud:
```suggestion
            'name' => array('type' => self::TYPE_STRING, 'validate' => 'isHookName', 'required' => true, 'size' => 191),
```
For utf8mb4 :roll_eyes:

eternoendless: I don't suppose anyone in their right mind would use emojis + long hook names?!

PierreRambaud: People like using emojis everywhere right now :(

eternoendless: Ok as far as I can tell the 191 limit applies for indexed columns regardless of their content, so I'll have to change it.

PierreRambaud (approved):

sarahdib:
@eternoendless there is an error when I try to re-install the module with a long hook. 
Capture d’écran 2019-11-06 à 16 44 33

PierreRambaud:
Thanks @eternoendless

PR #16180, opened 2019-10-29
Use Media Server for assets (CSS, JS, Images)

### LINKED ISSUE #9890: [BOOM-3849] css & js should use CDN when activated
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-3849).

- _**Reporter:**_ jmawad@bobbies.com
- _**Created at:**_ Fri, 15 Sep 2017 12:28:15 +0200

<p>Apparently, even if you use mediaserver, the css and js in cache are not using CDN feature, as the images do.</p>

<p>You should write a Htaccess rule to handle the /themes/<span class="error">&#91;theme&#93;</span>/assets/cache, and use media server url in header when calling the css</p>
        
- activate media server and put something.cdn.com
<br/>

<br/>
&gt; Images are now correctly using cdn url, but not the concatenated JS and CSS

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | Use Media Server for assets (CSS, JS, Images)
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9890
| How to test?  | **Test without SmartCache :**<br>1. FO : In your PrestaShop, without MediaServer defined, all CSS/JS/images are called in your MediaServer<br>2. BO : In "Configure > Advanced > Performance", disable Smart cache for CSS & JS, and define a media server (CDN)<br>3. FO : all CSS/JS/images from theme or favicon are called from your CDN (except from modules)<br><br>**Test with SmartCache :** <br>1. FO : In your PrestaShop, without MediaServer defined, all CSS/JS/images are called in your MediaServer<br>2. BO : In "Configure > Advanced > Performance", enable Smart cache for CSS & JS, and define a media server (CDN)<br>3. FO : all CSS/JS/images from theme or favicon are called from your CDN (except from modules) and must be compiled (1 CSS & 1 JS for the front)<br>

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16180)

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -2389,7 +2389,7 @@ class ToolsCore
      *
      * @return string
      */
-    public static function getMediaServer($filename)
+    public static function getMediaServer(string $filename): string
     {
         if (self::hasMediaServer() && ($id_media_server = (abs(crc32($filename)) % self::$_cache_nb_media_servers + 1))) {
             return constant('_MEDIA_SERVER_' . $id_media_server . '_');
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -2391,7 +2391,7 @@ class ToolsCore
      */
     public static function getMediaServer($filename)
     {
-        if ($filename && self::hasMediaServer() && ($id_media_server = (abs(crc32($filename)) % self::$_cache_nb_media_servers + 1))) {
+        if (self::hasMediaServer() && ($id_media_server = (abs(crc32($filename)) % self::$_cache_nb_media_servers + 1))) {
             return constant('_MEDIA_SERVER_' . $id_media_server . '_');
         }
 
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -2364,7 +2364,10 @@ class ToolsCore
 
     protected static $_cache_nb_media_servers = null;
 
-    public static function getMediaServer($filename)
+    /**
+     * @return bool
+     */
+    public static function hasMediaServer()
     {
         if (self::$_cache_nb_media_servers === null && defined('_MEDIA_SERVER_1_') && defined('_MEDIA_SERVER_2_') && defined('_MEDIA_SERVER_3_')) {
             if (_MEDIA_SERVER_1_ == '') {
@@ -2377,8 +2380,16 @@ class ToolsCore
                 self::$_cache_nb_media_servers = 3;
             }
         }
+        return (self::$_cache_nb_media_servers > 0);
+    }
 
-        if ($filename && self::$_cache_nb_media_servers && ($id_media_server = (abs(crc32($filename)) % self::$_cache_nb_media_servers + 1))) {
+    /**
+     * @param string $filename
+     * @return string
+     */
+    public static function getMediaServer($filename)
+    {
+        if ($filename && self::hasMediaServer() && ($id_media_server = (abs(crc32($filename)) % self::$_cache_nb_media_servers + 1))) {
             return constant('_MEDIA_SERVER_' . $id_media_server . '_');
         }
 
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -2380,11 +2380,13 @@ class ToolsCore
                 self::$_cache_nb_media_servers = 3;
             }
         }
-        return (self::$_cache_nb_media_servers > 0);
+
+        return self::$_cache_nb_media_servers > 0;
     }
 
     /**
      * @param string $filename
+     *
      * @return string
      */
     public static function getMediaServer($filename)
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -2367,7 +2367,7 @@ class ToolsCore
     /**
      * @return bool
      */
-    public static function hasMediaServer()
+    public static function hasMediaServer(): bool
     {
         if (self::$_cache_nb_media_servers === null && defined('_MEDIA_SERVER_1_') && defined('_MEDIA_SERVER_2_') && defined('_MEDIA_SERVER_3_')) {
             if (_MEDIA_SERVER_1_ == '') {
--- a/classes/assets/AbstractAssetManager.php
+++ b/classes/assets/AbstractAssetManager.php
@@ -52,6 +52,7 @@ abstract class AbstractAssetManagerCore
 
     /**
      * @param string $relativePath
+     *
      * @return bool|string
      */
     public function getFullPath($relativePath)
--- a/classes/assets/AbstractAssetManager.php
+++ b/classes/assets/AbstractAssetManager.php
@@ -50,10 +50,14 @@ abstract class AbstractAssetManagerCore
 
     abstract protected function getList();
 
-    protected function getFullPath($relativePath)
+    /**
+     * @param string $relativePath
+     * @return bool|string
+     */
+    public function getFullPath($relativePath)
     {
         foreach ($this->getDirectories() as $baseDir) {
-            $fullPath = $baseDir . ltrim($relativePath, '/'); // not DIRECTORY_SEPARATOR because, it's path included manualy
+            $fullPath = $baseDir . ltrim($relativePath, '/'); // not DIRECTORY_SEPARATOR because, it's path included manually
             if (file_exists($this->getPathFromUri($fullPath))) {
                 return $fullPath;
             }
--- a/classes/assets/AbstractAssetManager.php
+++ b/classes/assets/AbstractAssetManager.php
@@ -55,7 +55,7 @@ abstract class AbstractAssetManagerCore
      *
      * @return bool|string
      */
-    public function getFullPath($relativePath)
+    public function getFullPath(string $relativePath)
     {
         foreach ($this->getDirectories() as $baseDir) {
             $fullPath = $baseDir . ltrim($relativePath, '/'); // not DIRECTORY_SEPARATOR because, it's path included manually
--- a/classes/assets/CccReducer.php
+++ b/classes/assets/CccReducer.php
@@ -65,7 +65,7 @@ class CccReducerCore
         }
         if (Tools::hasMediaServer()) {
             $relativePath = _THEMES_DIR_ . _THEME_NAME_ . '/assets/cache/' . $cccFilename;
-            $destinationPath = $destinationUri = Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath) . $relativePath;
+            $destinationUri = Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath) . $relativePath;
         }
 
         $cssFileList['external']['theme-ccc'] = [
@@ -107,7 +107,7 @@ class CccReducerCore
             }
             if (Tools::hasMediaServer()) {
                 $relativePath = _THEMES_DIR_ . _THEME_NAME_ . '/assets/cache/' . $cccFilename;
-                $destinationPath = $destinationUri = Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath) . $relativePath;
+                $destinationUri = Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath) . $relativePath;
             }
 
             $cccItem = [];
--- a/classes/assets/CccReducer.php
+++ b/classes/assets/CccReducer.php
@@ -58,7 +58,6 @@ class CccReducerCore
         $version = Configuration::get('PS_CCCCSS_VERSION');
         $cccFilename = 'theme-' . $this->getFileNameIdentifierFromList($files) . $version . '.css';
         $destinationPath = $this->cacheDir . $cccFilename;
-        $destinationUri = $this->getFQDN() . $this->getUriFromPath($destinationPath);
 
         if (!$this->filesystem->exists($destinationPath)) {
             CssMinifier::minify($files, $destinationPath);
@@ -66,6 +65,8 @@ class CccReducerCore
         if (Tools::hasMediaServer()) {
             $relativePath = _THEMES_DIR_ . _THEME_NAME_ . '/assets/cache/' . $cccFilename;
             $destinationUri = Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath) . $relativePath;
+        } else {
+            $destinationUri = $this->getFQDN() . $this->getUriFromPath($destinationPath);
         }
 
         $cssFileList['external']['theme-ccc'] = [
@@ -100,7 +101,6 @@ class CccReducerCore
             $version = Configuration::get('PS_CCCJS_VERSION');
             $cccFilename = $position . '-' . $this->getFileNameIdentifierFromList($files) . $version . '.js';
             $destinationPath = $this->cacheDir . $cccFilename;
-            $destinationUri = $this->getFQDN() . $this->getUriFromPath($destinationPath);
 
             if (!$this->filesystem->exists($destinationPath)) {
                 JsMinifier::minify($files, $destinationPath);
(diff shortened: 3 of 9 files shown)


classes/controller/FrontController.php
```diff
@@ -1088,9 +1093,14 @@ public function registerJavascript($id, $relativePath, $params = array())
...
         ];
-
         $params = array_merge($default_params, $params);
 
+        if (Tools::hasMediaServer() && !Configuration::get('PS_JS_THEME_CACHE')) {
+            $addSlash = substr($relativePath, 0, 1) !== '/' ? '/' : '';
+            $themeDir = substr($id, 0, 6) == 'theme-' ? _THEMES_DIR_ . _THEME_NAME_: '';
+            $relativePath = Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath) . $addSlash . $themeDir . $relativePath;
+            $params['server'] = 'remote';
+        }
```
matthieu-rolland, line 1101: It looks like this could be externalized in a function (ex: buildRelativePath), for readability and DRY purpose? (it's very similar to the code at line 1069)

matthieu-rolland (dismissed):
minor change requested

classes/controller/FrontController.php
Progi1984: Fixed

matthieu-rolland:
@Progi1984  You just have to fix the prettyCI issue and it's a LGTM for me !

Progi1984:
@matthieu-rolland :ok_hand: for PrettyCI

classes/controller/FrontController.php
```diff
@@ -1099,6 +1105,20 @@ public function unregisterJavascript($id)
         $this->javascriptManager->unregisterById($id);
     }
 
+    /**
+     * @param string $id
+     * @param string $relativePath
+     *
+     * @return string
+     */
+    private function buildRelativePath(string $id, string $relativePath)
```
jolelievre:
```suggestion
    private function buildRelativePath(string $id, string $relativePath): string
```

classes/Tools.php
```diff
@@ -2364,7 +2364,10 @@ public static function replaceByAbsoluteURL($matches)
 
     protected static $_cache_nb_media_servers = null;
 
-    public static function getMediaServer($filename)
+    /**
+     * @return bool
+     */
+    public static function hasMediaServer()
```
jolelievre:
```suggestion
    public static function hasMediaServer(): bool
```

classes/controller/FrontController.php
```diff
@@ -1099,6 +1105,20 @@ public function unregisterJavascript($id)
...
+    {
+        $addSlash = substr($relativePath, 0, 1) !== '/' ? '/' : '';
+        $themeDir = substr($id, 0, 6) == 'theme-' ? _THEMES_DIR_ . _THEME_NAME_ : '';
```
jolelievre: This is used to detect the compiled assets?

classes/controller/FrontController.php
```diff
@@ -1605,9 +1625,9 @@ public function getTemplateVarShop()
             'long' => Configuration::get('PS_STORES_CENTER_LONG'),
             'lat' => Configuration::get('PS_STORES_CENTER_LAT'),
 
-            'logo' => (Configuration::get('PS_LOGO')) ? _PS_IMG_ . Configuration::get('PS_LOGO') : '',
-            'stores_icon' => (Configuration::get('PS_STORES_ICON')) ? _PS_IMG_ . Configuration::get('PS_STORES_ICON') : '',
-            'favicon' => (Configuration::get('PS_FAVICON')) ? _PS_IMG_ . Configuration::get('PS_FAVICON') : '',
+            'logo' => (Configuration::get('PS_LOGO')) ? Configuration::get('PS_LOGO') : '',
```
jolelievre, line 1616: Why do you need to do this? Is it because in the template `{$urls.img_ps_url}` is correctly set with the CDN url?
Because what worries me is that it changes the template vars used by all themes, so it's likely every themes will need to be updated to make their logo and favicon work again, so we'd better have a good reason for changing this

jolelievre (commented):
Wow, this must have not been easy to figure out!!

classes/controller/FrontController.php
Progi1984: Fixed

classes/Tools.php
Progi1984: Fixed

classes/controller/FrontController.php
Progi1984: No. Compiled assets has managed in CccReducer.php. $themeDir permits to define the real path to the asset in the theme (which is not given during the registration). Example :
```php
$this->registerStylesheet('theme-error', '/assets/css/error.css', ['media' => 'all', 'priority' => 50]);
```
The file is here : `themes/classic/assets/css/error.css`.

jolelievre: Ooohh, so to include a theme assets you need to use an ID starting with `theme-wtf` ?
You should add a comment about that then, it's not very intuitive (I wonder if it's documented)

Progi1984: I fixed it to a better way.

classes/controller/FrontController.php
Progi1984: In fact, we need to have a absolute URL : 
* In template, `$urls.img_ps_url` give me `http://127.0.0.1/PrestaShop/img/`
* While `_PS_IMG_` is just relative `/img/`

matthieu-rolland (dismissed):
lgtm

jolelievre (dismissed):

classes/Tools.php
```diff
@@ -2378,7 +2381,17 @@ public static function getMediaServer($filename)
...
+    }
+
+    /**
+     * @param string $filename
+     *
+     * @return string
+     */
+    public static function getMediaServer($filename)
+    {
+        if ($filename && self::hasMediaServer() && ($id_media_server = (abs(crc32($filename)) % self::$_cache_nb_media_servers + 1))) {
```
PierreRambaud: Otherwise you can't have a filename '0', which is a valid filename
```suggestion
        if (self::hasMediaServer() && ($id_media_server = (abs(crc32($filename)) % self::$_cache_nb_media_servers + 1))) {
```

PierreRambaud: Do you have the reason why we have the crc32 here?

classes/assets/CccReducer.php
```diff
@@ -58,16 +58,21 @@ public function reduceCss($cssFileList)
...
         $cccFilename = 'theme-' . $this->getFileNameIdentifierFromList($files) . $version . '.css';
         $destinationPath = $this->cacheDir . $cccFilename;
+        $destinationUri = $this->getFQDN() . $this->getUriFromPath($destinationPath);
 
         if (!$this->filesystem->exists($destinationPath)) {
             CssMinifier::minify($files, $destinationPath);
         }
+        if (Tools::hasMediaServer()) {
+            $relativePath = _THEMES_DIR_ . _THEME_NAME_ . '/assets/cache/' . $cccFilename;
+            $destinationPath = $destinationUri = Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath) . $relativePath;
```
PierreRambaud: Weird to have path and uri similar.. Something must be wrong here :thinking:

classes/assets/CccReducer.php
```diff
@@ -95,17 +100,22 @@ public function reduceJs($jsFileList)
...
             $cccFilename = $position . '-' . $this->getFileNameIdentifierFromList($files) . $version . '.js';
             $destinationPath = $this->cacheDir . $cccFilename;
+            $destinationUri = $this->getFQDN() . $this->getUriFromPath($destinationPath);
 
             if (!$this->filesystem->exists($destinationPath)) {
                 JsMinifier::minify($files, $destinationPath);
             }
+            if (Tools::hasMediaServer()) {
+                $relativePath = _THEMES_DIR_ . _THEME_NAME_ . '/assets/cache/' . $cccFilename;
+                $destinationPath = $destinationUri = Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath) . $relativePath;
```
PierreRambaud: same here

PierreRambaud (changes requested):
Some questions

classes/Tools.php
Progi1984: :white_check_mark:

Progi1984: Absolutely not ! I check log ;)

Progi1984: After blame... It's historical !

classes/assets/CccReducer.php
Progi1984: In template, only the URI & media is used.

When we used a CDN/Media Server, the URI is the path where the file is.

classes/assets/CccReducer.php
Progi1984: Answered above

classes/Tools.php
PierreRambaud: Dat sucks

classes/assets/CccReducer.php
PierreRambaud: In *our* templates :thinking: 
A URI = http://mydomain.fr/path/to/something.jpg
A path = /path/to/something.jpg

If your path integrate the scheme + authority, it's no longer a path

Progi1984: But you are totally right. I fix it now.

classes/Tools.php
Progi1984: I have a small idea :
 - `abs(crc32($filename)` transform a string in a integer
 - and the modulo permits to not use always the same media server
 - but always use the same media server for a specific filename.

Not ?

classes/Tools.php
```diff
@@ -2378,7 +2381,17 @@ public static function getMediaServer($filename)
...
-        if ($filename && self::$_cache_nb_media_servers && ($id_media_server = (abs(crc32($filename)) % self::$_cache_nb_media_servers + 1))) {
+        return self::$_cache_nb_media_servers > 0;
+    }
+
+    /**
+     * @param string $filename
+     *
+     * @return string
+     */
+    public static function getMediaServer($filename)
```
```suggestion
    public static function getMediaServer(string $filename): string
```

classes/assets/AbstractAssetManager.php
```diff
@@ -50,10 +50,15 @@ public function __construct(array $directories, ConfigurationInterface $configur
 
     abstract protected function getList();
 
-    protected function getFullPath($relativePath)
+    /**
+     * @param string $relativePath
+     *
+     * @return bool|string
+     */
+    public function getFullPath($relativePath)
```
PierreRambaud:
```suggestion
    public function getFullPath(string $relativePath)
```

classes/assets/CccReducer.php
```diff
@@ -58,16 +58,21 @@ public function reduceCss($cssFileList)
         $version = Configuration::get('PS_CCCCSS_VERSION');
         $cccFilename = 'theme-' . $this->getFileNameIdentifierFromList($files) . $version . '.css';
         $destinationPath = $this->cacheDir . $cccFilename;
+        $destinationUri = $this->getFQDN() . $this->getUriFromPath($destinationPath);
 
         if (!$this->filesystem->exists($destinationPath)) {
             CssMinifier::minify($files, $destinationPath);
         }
+        if (Tools::hasMediaServer()) {
```
PierreRambaud, line 65: Because getFQDN function do mysql requests.

```suggestion
        if (!Tools::hasMediaServer()) {
            $destinationUri = $this->getFQDN() . $this->getUriFromPath($destinationPath);
        } else {
            $relativePath = _THEMES_DIR_ . _THEME_NAME_ . '/assets/cache/' . $cccFilename;
            $destinationUri = Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath) . $relativePath;
        }
```

classes/assets/CccReducer.php
```diff
@@ -95,17 +100,22 @@ public function reduceJs($jsFileList)
             $version = Configuration::get('PS_CCCJS_VERSION');
             $cccFilename = $position . '-' . $this->getFileNameIdentifierFromList($files) . $version . '.js';
             $destinationPath = $this->cacheDir . $cccFilename;
+            $destinationUri = $this->getFQDN() . $this->getUriFromPath($destinationPath);
 
             if (!$this->filesystem->exists($destinationPath)) {
                 JsMinifier::minify($files, $destinationPath);
             }
+            if (Tools::hasMediaServer()) {
```
PierreRambaud, line 108: same here

classes/controller/FrontController.php
```diff
@@ -1064,9 +1064,13 @@ public function registerStylesheet($id, $relativePath, $params = array())
             'inline' => false,
             'server' => 'local',
         ];
-
         $params = array_merge($default_params, $params);
 
+        if (Tools::hasMediaServer() && !Configuration::get('PS_CSS_THEME_CACHE')) {
+            $relativePath = Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath)
```
PierreRambaud, line 1070: Not sure 
```php
Tools::getCurrentUrlProtocolPrefix() . Tools::getMediaServer($relativePath)
                . $this->stylesheetManager->getFullPath($relativePath)
``` 
can be empty / null :thinking:

classes/Tools.php
Progi1984: Fixed.

classes/assets/AbstractAssetManager.php
Progi1984: Fixed

classes/assets/CccReducer.php
Progi1984: Fixed

classes/controller/FrontController.php
Progi1984: @PierreRambaud It's not this part but `$this->stylesheetManager->getFullPath($relativePath)` which can be equals to false

classes/assets/CccReducer.php
Progi1984: Fixed

classes/controller/FrontController.php
PierreRambaud: So you need to add `()` because
```php
$a = 'test';
$b = 'test2';
$c = null;
$d = 'hoho';

var_dump($a . $b . $c ?? $d);
# => string(9) "testtest2"
```

Progi1984: Thanks ! I completly failed that :( ! Fixed & Pushed.

PierreRambaud (approved):

matks:
Thank you @Progi1984 and reviewers

micka-fdz:
Hello @Progi1984. This PR causes problems on previous versions of PrestaShop (< 1.7.7).
Image paths are broken (logo and favicon) because of the add of `{$urls.img_ps_url}`.

PierreRambaud:
> Hello @Progi1984. This PR causes problems on previous versions of PrestaShop (< 1.7.7).
> Image paths are broken (logo and favicon) because of the add of `{$urls.img_ps_url}`.

Unfortunately, you can't get a pull request and apply it directly to your codebase.
This PR targets the develop branch, regarding the time it means the 1.7.8.0 and files were changed for this pull request are not the same as yours.
If you need this change, you will have to upgrade your shop to the 1.7.8.0 or find all needed parts and apply it on your codebase, but you will not be able to upgrade your shop with manuals changes :/

ghost:
Nice catch @micka-fdz 

@PierreRambaud no logo when i do an upgrade 176 --> 177

micka-fdz:
I think it's a BC break.

If we use other theme than Classic (or update PS without update Classic theme), image paths are broken on logo and favicon.

PierreRambaud:
I agree with both of you if we consider the classic theme as part of our public API. It's a BC break.

SPACOMA:
Good evening,

firstly my best wishes 2021 to all!

Also a big thank you, I applied these mods and it's perfect on these points.

This concerns a prestashop 1.7.6.8 store, php 7.2.34 and 10.3.27-MariaDB.

Everything is more fluid whether it is the back office or the front office. On the other hand, I have high latencies when recording product modifications or even when connecting with the Merlin Back Office software.

After checking the database with the host, everything appears to be ok on this side.

Does anyone have similar feedback?

Thank you in advance for your help.

Issue #12772, opened 2019-03-02
Fonts are not loaded when not connected to internet


alireza2005:
hello.

I install prestashop 1.7.5.1 (the final version) in local server (in wamp server - Apache 2.4.37 - PHP 7.2.14 - MySQL 5.7.24). My front office is ok, but most pages in my back office have problem ( for example  picture below )(in this case My computer is not connected to the Internet).
![product problems](https://user-images.githubusercontent.com/48148331/53679000-f7898500-3cdb-11e9-9443-d490b7aba70a.PNG)

but when i connect my computer to the internet my back office all is ok (for example picture below).
![produck okey](https://user-images.githubusercontent.com/48148331/53679015-16881700-3cdc-11e9-8d41-e29e55527d73.PNG)

as if jquery  or css not load Properly in my back office (when My computer is not connected to the Internet).

please help me,very thanks.

PrestaShop version: 1.7.5.1
wamp server version: 3.1.7 64bit
PHP version: 7.2.14
MySQL version: 5.7.24

rblaurin:
Hi
It seems your browser is not able to access the material icons. Is there something in javascript console log?

alireza2005:
Hi
where is  javascript console log

rblaurin:
If you are using Google Chrome, just press F12 and click the Console tab

alireza2005:
![errors](https://user-images.githubusercontent.com/48148331/53682145-0d14a400-3d08-11e9-89ba-cc3de57cd7ed.PNG)

rblaurin:
Can you send also a full print of the network tab?

alireza2005:
![errors-network](https://user-images.githubusercontent.com/48148331/53682641-0177ac00-3d0d-11e9-983b-8faed61b329a.PNG)

rblaurin:
Try to clear cache and cookies, go to command prompt and enter the commands listed below, one by one:
ipconfig /flushdns
ipconfig /renew

alireza2005:
Hi
ipconfig donot solve my problem
when my computer is connected to internet there is no problem
in offline mode as if google font and btn (botton) classes have loading problem

rblaurin:
Ok, sorry, I misunderstood your problem. You are not connected.
[...]

PR #16209, opened 2019-10-30
Use local fonts in BO default theme

eternoendless:
| Questions     | Answers
| Branch?       | develop
| Description?  | The default theme loaded its fonts from the Google fonts API, which obviously doesn't work when offline. This PR includes those fonts for legacy pages in the BO. A follow-up PR should fix this on migrated pages as well.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Partial fix for #12772
| How to test?  | Load a legacy page while disconnected from the internet, see that the fonts are correct.

This PR also adds the webpack cleanup plugin on the default theme. Some assets were removed (see diff to learn which ones), which would indicate that they are no longer used. We should verify that they are indeed no longer used.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16209)


admin-dev/themes/default/sass/modules/_variables.sass
```diff
@@ -72,9 +72,8 @@ $body-bg: $bg-content-color
 $text-color: $gray
 
 // Typography
-$url-font-content: "//fonts.googleapis.com/css?family=Open+Sans:400,400i,600,700" !default
+$FontPathOpenSans: '../node_modules/open-sans-fonts/open-sans'
```
Progi1984, line 75:
```suggestion
$fontPathOpenSans: '../node_modules/open-sans-fonts/open-sans'
```

eternoendless: Unfortunately this variable must be defined exactly like that in order for the required library to work.

Progi1984: Ok. Sorry :pray:

Progi1984 (approved):

PierreRambaud (approved):

eternoendless:
Be aware that because the new theme is still loading fonts from Google, when switching from legacy pages to new pages fonts will have to be downloaded again by the client.

PierreRambaud:
Thanks @eternoendless

PR #16213, opened 2019-10-30
Make id_order available everywhere where order_name is available

### LINKED ISSUE #13964: {id_order} not available in order_conf
Only {order_name} is available as a variable to use in the template, but {id_order} is available in all other templates.
Could you please make this variable available also out-of-the-box ?

atomiix:
| Questions     | Answers
| Branch?       | develop
| Description?  | Make {id_order} available everywhere where {order_name} is available
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #13964
| How to test?  | replace `{order_name}` with `{id_order}` in `order_conf` mail template, buy any product on the store and look at the confirmation mail

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16213)

### DIFF
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -1135,7 +1135,7 @@ abstract class PaymentModuleCore extends Module
                         '{voucher_num}' => $voucher->code,
                         '{firstname}' => $this->context->customer->firstname,
                         '{lastname}' => $this->context->customer->lastname,
-                        '{id_order}' => $order->reference,
+                        '{id_order}' => $order->id,
                         '{order_name}' => $order->getUniqReference(),
                     );
                     Mail::Send(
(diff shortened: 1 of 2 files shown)


classes/PaymentModule.php
```diff
@@ -1134,7 +1135,7 @@ protected function createOrderCartRules(
                         '{voucher_num}' => $voucher->code,
                         '{firstname}' => $this->context->customer->firstname,
                         '{lastname}' => $this->context->customer->lastname,
-                        '{id_order}' => $order->reference,
+                        '{id_order}' => $order->id,
```
matks, line 1138: This is logical. However ... aren't we going to break modules that have been using this variable 😅 ?

ping @eternoendless : this would be kind of a BC break

PierreRambaud (approved):

classes/PaymentModule.php
PierreRambaud: It's not a BC, it's a bug fix. Otherwise all our bugs fixes are BC break :sweat_smile:

Progi1984:
Thanks @atomiix :)

Issue #9659, opened 2018-08-22
[BOOM-5403] Profiling for hooks and modules are broken


prestonBot:
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-5403).

- _**Reporter:**_ hubbobubbo
- _**Created at:**_ Tue, 24 Apr 2018 16:56:05 +0200

<p>No info is shown for hooks and modules when profiling is activated. This was very helpful to have and works in PS 1.6</p>

<p>Might be related to:<br/>
<a href="http://forge.prestashop.com/browse/BOOM-1242" class="external-link" rel="nofollow">http://forge.prestashop.com/browse/BOOM-1242</a></p>
        
* How to reproduce the issue ?

N/A

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-5403#comment-166722).

- _**Author:**_ khouloud.belguith
- _**Created at:**_ Tue, 24 Apr 2018 17:30:20 +0200

<p>Hi Johan,</p>

<p> Thanks for your report.<br/>
I manage to reproduce the issue with the last version 1.7.3.1.<br/>
 <span class="image-wrap" style=""><a id="67183_thumb" href="http://forge.prestashop.com/secure/attachment/67183/67183_screenshot-2.png" title="screenshot-2.png" file-preview-type="image" file-preview-id="67183" file-preview-title="screenshot-2.png"><img src="http://forge.prestashop.com/secure/thumbnail/67183/_thumb_67183.png" style="border: 0px solid black" /></a></span> <br/>
But in Ps1.6, it is OK.<br/>
 <span class="image-wrap" style=""><a id="67182_thumb" href="http://forge.prestashop.com/secure/attachment/67182/67182_screenshot-1.png" title="screenshot-1.png" file-preview-type="image" file-preview-id="67182" file-preview-title="screenshot-1.png"><img src="http://forge.prestashop.com/secure/thumbnail/67182/_thumb_67182.png" style="border: 0px solid black" /></a></span> </p>

<p>Best regards, Khouloud</p>

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-5403#comment-173307).

- _**Author:**_ marion_francois
- _**Created at:**_ Thu, 2 Aug 2018 11:40:24 +0200

[...]

PR #16231, opened 2019-10-31
Return of hook profiling

StanislovasRamanauskas:
| Questions     | Answers
| Branch?       | develop 
| Description?  | Added hook profiling which was removed in 1.7.0.0 version.
| Type?         | improvement
| Category?     | FO / BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/9659
| How to test?  | Turn on _PS_DEBUG_PROFILING_ in defines.inc.php and see profiling list in page bottom.<br>Screenshot from 2019-10-31 16-29-39

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16231)

### DIFF
--- a/classes/Hook.php
+++ b/classes/Hook.php
@@ -991,8 +991,18 @@ class HookCore extends ObjectModel
         $memory_end = memory_get_usage(true);
 
         Db::getInstance()->execute('
-		INSERT INTO '._DB_PREFIX_.'modules_perfs (session, module, method, time_start, time_end, memory_start, memory_end)
-		VALUES ('.(int)Module::$_log_modules_perfs_session.', "'.pSQL($module->name).'", "'.pSQL($method).'", "'.pSQL($time_start).'", "'.pSQL($time_end).'", '.(int)$memory_start.', '.(int)$memory_end.')');
+		    INSERT INTO 
+		    '._DB_PREFIX_.'modules_perfs (session, module, method, time_start, time_end, memory_start, memory_end)
+		    VALUES ('
+                .(int)Module::$_log_modules_perfs_session.', 
+                "'.pSQL($module->name).'", 
+                "'.pSQL($method).'", 
+                "'.pSQL($time_start).'", 
+                "'.pSQL($time_end).'", 
+                '.(int)$memory_start.', 
+                '.(int)$memory_end.'
+		    )'
+        );
 
         return $hookCallResult;
     }
--- a/classes/Hook.php
+++ b/classes/Hook.php
@@ -992,15 +992,15 @@ class HookCore extends ObjectModel
 
         Db::getInstance()->execute('
 		    INSERT INTO 
-		    '._DB_PREFIX_.'modules_perfs (session, module, method, time_start, time_end, memory_start, memory_end)
+		    ' . _DB_PREFIX_ . 'modules_perfs (session, module, method, time_start, time_end, memory_start, memory_end)
 		    VALUES ('
-                .(int)Module::$_log_modules_perfs_session.', 
-                "'.pSQL($module->name).'", 
-                "'.pSQL($method).'", 
-                "'.pSQL($time_start).'", 
-                "'.pSQL($time_end).'", 
-                '.(int)$memory_start.', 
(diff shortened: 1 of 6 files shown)


mickaelandrieu:
Hi @StanislovasRamanauskas, why not use the Hook profiler provided in the current debug toolbar ?

kpodemski:
> Hi @StanislovasRamanauskas, why not use the Hook profiler provided in the current debug toolbar ?

because you don't have symfony profiler on front-office @mickaelandrieu :)

mickaelandrieu:
> because you don't have symfony profiler on front-office @mickaelandrieu :)

I always forget the front office !

install-dev/upgrade/sql/1.7.0.0.sql
```diff
@@ -37,8 +37,6 @@ ALTER TABLE `PREFIX_customer` CHANGE COLUMN `lastname` `lastname` varchar(255) N
 /* Changes regarding modules */
 ALTER TABLE `PREFIX_module` ADD UNIQUE INDEX `name_UNIQUE` (`name` ASC);
 
-DROP TABLE IF EXISTS `PREFIX_modules_perfs`;
```
sarjon: Table was dropped in 1.7.0.0, you don't need to remove it from here, but create it once again in `1.7.7.0.sql` file.

StanislovasRamanauskas:
> Hi @StanislovasRamanauskas, why not use the Hook profiler provided in the current debug toolbar ?
@mickaelandrieu hook debugging is important feature for improving module speed. Also, not all hook profiling functionality was removed from 1.7.0.0 version. Till now PS have empty hook list in profiling and that is confusing. 
Screenshot from 2019-11-03 10-35-10

kpodemski:
@PrestaShop/prestashop-core-developers LGTM!

classes/Hook.php
```diff
@@ -967,7 +967,44 @@ public static function exec(
 
     public static function coreCallHook($module, $method, $params)
     {
-        return $module->{$method}($params);
+        // Define if we will log modules performances for this session
+        if (Module::$_log_modules_perfs === null) {
+            Module::$_log_modules_perfs = (_PS_DEBUG_PROFILING_ && mt_rand(0, _PS_DEBUG_PROFILING_ - 1) == 0);
+            if (Module::$_log_modules_perfs) {
+                Module::$_log_modules_perfs_session = mt_rand();
```
eternoendless, line 974: Looking at the code, it looks that the session ID is random and lasts for exactly one run. In that case, why bother to save results on the database?

eternoendless (changes requested):
I think this could be useful for some uses, but I think the implementation needs to be improved. For example, I'd rather see the whole profiling logic in a class of its own:

```php
function doSomething(...$params) {
  if ($this->profiler->shouldProfile()) {
    return $this->profiler->interceptDoSomething(...$params); // does the thing with profiling
  }

  return $this->doTheThing(...$params); // does the thing without profiling
}
```

or even better, use strategy pattern:

```php
public function doSomething(...$params)
{
  $this->thingExecutor->doTheThing(...$params);
}
```

In the latter case you just need to inject a different `thingExecutor` depending if you want to profile the thing or not.

Also, does it worth it adding this when you can use Blackfire.io?

mickaelandrieu:
Hi @StanislovasRamanauskas,

would you mind to contribute this to the [Developer Tools](https://github.com/friends-of-presta/developer_tools) module ?

I will take this code anyway, but I prefer to let you the attribution of this code 👍, the way I see it this won't be shipped in 1.7.7 so we won't have this in the Core before 2021...


@eternoendless the main use of this profiler is * not * the performance, but figure out which hooks are available: this is a limited version of the Hook Profiler I've contributed on the "modern" part of PrestaShop.

EDIT (13/05/20) => Now available under [Friends of Presta Developer Tools module](https://github.com/friends-of-presta/developer_tools/pull/12)

PierreRambaud:
Closed in favor of #20673
Thanks for your work, it helps us a lot :wink:

PR #16236, opened 2019-10-31
Fix price validation on 64-bit systems

ghost:
| Questions     | Answers
| Branch?       | develop
| Description?  | The number of possible decimals has been greatly increased in 64-bit systems and there is a big conflict in all objects that currently validate a price (_Validate::isPrice()_), such as when creating or editing an order. This could be considered a critical error instead of a simple bug.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Unknown
| How to test?  | Create a new order with a total of *45031.04* and then insert a coupon with a fixed amount discount of *45031.03*, of course this issue happen with many other amounts, here a quick sample of the issue: https://streamable.com/z2cs2

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16236)

### DIFF
--- a/classes/Validate.php
+++ b/classes/Validate.php
@@ -279,7 +279,7 @@ class ValidateCore
      */
     public static function isPrice($price)
     {
-        return preg_match('/^[0-9]{1,10}(\.[0-9]{1,9})?$/', $price);
+        return preg_match('/^[0-9]{1,10}(\.[0-9]+)?$/', $price);
     }
 
     /**
@@ -291,7 +291,7 @@ class ValidateCore
      */
     public static function isNegativePrice($price)
     {
-        return preg_match('/^[-]?[0-9]{1,10}(\.[0-9]{1,9})?$/', $price);
+        return preg_match('/^[-]?[0-9]{1,10}(\.[0-9]+)?$/', $price);
     }
 
     /**


classes/Validate.php
```diff
@@ -279,7 +279,7 @@ public static function isImageTypeName($type)
      */
     public static function isPrice($price)
     {
-        return preg_match('/^[0-9]{1,10}(\.[0-9]{1,9})?$/', $price);
+        return preg_match('/^[0-9]{1,10}(\.[0-9]*})?$/', $price);
```
PierreRambaud: Since it was `{1,9}`, we need One or more.
```suggestion
        return preg_match('/^[0-9]{1,10}(\.[0-9]+})?$/', $price);
```

PierreRambaud:
Little fix and looks ok for me

PierreRambaud (dismissed):
Need to be updated

Rolige:
You are right, changes committed!

PierreRambaud:
Oh, one more thing. I think the fix should also be done on `isNegativePrice` :)

Rolige:
Sorry, I forgot this, it's done!

PierreRambaud:
I've no idea why tests are failing.
Could you please rebase your branch? :thinking:

Rolige:
Check now, I have requested for review again, maybe this help to allow the approval.

PierreRambaud:
Seems this change breaks the installation due to wrong validation on ecotax field which is using the isPrice validate. But I don't manage to reproduce :thinking: 
It could be because of the size of this field :thinking:

Rolige:
Well, I have made an installation with this patch and nothing weird happened.

PierreRambaud:
I rebase and fix your branch. Wait for CI :)

Rolige:
@PierreRambaud Any update on this simple but needed fix?

PierreRambaud (approved):

Progi1984 (approved):

sarahdib:
Hello @Rolige 

Can you re-upload the video ? we can't found where is the issue.

Thank you

Rolige:
Hi @sarahdib 

I'm sorry but I dont have this video anymore, apparently is was deleted, but reproduce the error is really easy, just follow my steps.

Regards!

sarahdib:
Hello @Rolige 

I try to follow the steps but I don't know what is the bug. Everything looks fine. Could you check develop branch and see if the bug is already fixed ?

Thank you :)

Rolige:
@sarahdib have you tried with the exact amount of my "How to test?" ? This problem occur with different amounts but not for all, it only happens with some very particular amounts and also your environment needs to be 64-bit architecture.

Regards!

sarahdib:
Hello @Rolige 

Yes I try exactly and nothing was wrong. My env is a 64 bits 

In order to test the pr we need to know what is the exact problem and all step to reproduce.

Rolige:
@sarahdib Please try again with php 7.1.3, with php 7.3 is not happening any more, I'm not sure what happen with php 7.2.
Hope this help!

SimonGrn:
Hello, 
I tried to reproduce with PHP 7.2.24 on a 64bits environment.
Here is my product:
image
Here is my discount:
image
Here is my cart:
image

This is on the current `develop` branch. I see no difference with your PR (same result, 0.01€ total).

@sarahdib got the same results on a PHP v7.1.3.

A lot of work has been done on rounding and calculating these past months and I think this issue may have been fixed elsewhere.
I deeply apologise for not getting back to you as fast as we'd hoped for his PR.
**Once again, thanks for your contribution**, we really appreciate.

I close this PR as the underlying issue could not be reproduced on the current `develop` branch.

Rolige:
I see, in fact the problem came from php, not from PS, this PR is little old and maybe the issue was fixed on newer versions of php.

Thank you!

PR #16266, opened 2019-11-04
Fix ajax exception when search for a product in order creation page

### LINKED ISSUE #16259: Localization exception in order creation page after clean ps install
**Describe the bug**
Cannot search for products in Sell > Order > add new Order.
Keep getting Localization exception associated with empty currency, thrown in Locale::getPriceSpecification() which is apparently called in admin controller when preparing context.

**To Reproduce**
1. Install prestashop development mode (using /install-dev)
2. Click on Sell > Orders > Create new
3. Search and choose a customer
4. Search for products
5. No products shows up. In console you can see ajax request error. (screenshot 1)

**Screenshots**
1. 
Screenshot from 2019-11-04 13-05-27

**Additional information**
PrestaShop version: develop
PHP version: N/A

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix the broken ajax product search in order creation page
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16259
| How to test?  | See below:

**Steps to reproduce the behavior:**

- Install prestashop development mode (using /install-dev)
- Click on Sell > Orders > Create new
- Search and choose a customer
- Search for products (ex: 'mug')
- Products should be displayed in a select

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16266)

### DIFF
--- a/admin-dev/themes/default/template/controllers/orders/form.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/form.tpl
@@ -631,7 +631,7 @@
 				action: "searchProducts",
 				id_cart: id_cart,
 				id_customer: id_customer,
-				id_currency: id_currency,
+				id_currency: $('#id_currency option:selected').val(),
 				product_search: $('#product').val()},
 			success : function(res)
 			{


PierreRambaud (approved):

Robin-Fischer-PS:
Hi @matthieu-rolland !

Your fix is OK, but I've got a new bug, I do not know if it's caused by your fix : 

When I add a product, it's OK, but if I add a second product immediately after on the same order, it creates a new cart (with only the last product added) instead of adding the product to current cart.

See screen record : 
https://drive.google.com/file/d/191hrSUes1-eW16bpVYjwkBze1U2TbtTW/view?usp=sharing

If this is not caused by your fix, we should create a new issue...

matthieu-rolland:
> Hi @matthieu-rolland !
> 
> Your fix is OK, but I've got a new bug, I do not know if it's caused by your fix :
> 
> When I add a product, it's OK, but if I add a second product immediately after on the same order, it creates a new cart (with only the last product added) instead of adding the product to current cart.
> 
> See screen record :
> https://drive.google.com/file/d/191hrSUes1-eW16bpVYjwkBze1U2TbtTW/view?usp=sharing
> 
> If this is not caused by your fix, we should create a new issue...

This bug could not be reached because of the broken ajax query.... it seems that there's an issue bigger than the one fixed in this PR. I'll create an issue 👍

PierreRambaud:
Thanks @matthieu-rolland

PR #16271, opened 2019-11-05
Remove useless checkbox in product options tab for attached files

### LINKED ISSUE #9790: [BOOM-4458] Product attached files
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-4458).

- _**Reporter:**_ marion_francois
- _**Created at:**_ Mon, 11 Dec 2017 18:02:35 +0100

In product options tab, after adding many attached files, there is a checkbox but there isn't any bulk action possible.The checkbox before "Title" should be removed
        
Capture du 2017-12-11 18-05-32 (1)

marionf:
| Questions     | Answers
| Branch?       | develop
| Description?  | Remove useless checkbox in product options tab for attached files
| Type?         | bug fix 
| Category?     | BO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9790
| How to test?  | Edit a product, go to options tab, add 2 attached files, verify there is no checkbox before "Title"

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16271)

### DIFF
--- a/admin-dev/themes/default/js/bundle/product/form.js
+++ b/admin-dev/themes/default/js/bundle/product/form.js
@@ -1156,15 +1156,6 @@ var attachmentProduct = (function() {
       var buttonSave = $('#form_step6_attachment_product_add');
       var buttonCancel = $('#form_step6_attachment_product_cancel');
 
-      /** check all attachments files */
-      $('#product-attachment-files-check').change(function() {
-        if ($(this).is(":checked")) {
-          $('#product-attachment-file input[type="checkbox"]').prop('checked', true);
-        } else {
-          $('#product-attachment-file input[type="checkbox"]').prop('checked', false);
-        }
-      });
-
       buttonCancel.click(function (){
         resetAttachmentForm();
       });
--- a/src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
@@ -404,7 +404,7 @@
           <table id="product-attachment-file" class="table">
             <thead class="thead-default">
               <tr>
-                <th class="col-md-3"><input type="checkbox" id="product-attachment-files-check" /> {{ 'Title'|trans({}, 'Admin.Global') }}</th>
+                <th class="col-md-3">{{ 'Title'|trans({}, 'Admin.Global') }}</th>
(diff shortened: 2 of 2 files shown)


Progi1984:
Hi, Thank you for your contribution. Please can you remove the JS relative to the ID #product-attachment-files-check in admin-dev/themes/default/js/bundle/product/form.js ?

matks:
> Hi, Thank you for your contribution. Please can you remove the JS relative to the ID #product-attachment-files-check in admin-dev/themes/default/js/bundle/product/form.js ?

Euuuuh 😅 are we sure this JS is useless ?

Progi1984:
> > Hi, Thank you for your contribution. Please can you remove the JS relative to the ID #product-attachment-files-check in admin-dev/themes/default/js/bundle/product/form.js ?
> 
> Euuuuh sweat_smile are we sure this JS is useless ?

It seems there is only one element with ID product-attachment-files-check.

matthieu-rolland:
> > Hi, Thank you for your contribution. Please can you remove the JS relative to the ID #product-attachment-files-check in admin-dev/themes/default/js/bundle/product/form.js ?
> 
> Euuuuh 😅 are we sure this JS is useless ?

Yes, it's only related to this checkbox and will never be executed when the checkbox is gone.
The only thing I'm not sure of, is if this template is used only once or in other places...

If the checkbox has never a reason to exist then this PR is fine (there's just the JS related to the checkbox to be removed as well)

marionf:
Done, I deleted the related JS

PierreRambaud (approved):

Progi1984:
> Done, I deleted the related JS

Thanks @marionf. Don't forget to build assets.

PierreRambaud:
@Progi1984 No need to build asset for this file

matks:
@matthieu-rolland @Progi1984 I know the JS is useless if we remove the ID. What I mean is: are we sure they are both useless ? 😛 I find it hard to believe there is a JS + an HTML item that were added and are completely useless. I'm quite sure there is an expected behavior that rely on them.

So removing them is valid only if:
- the behavior is now obsolete
or
- the behavior is broken (then maybe we should fix it instead ?)
or
- this is dead code

PierreRambaud:
I think it's a dead code, we are talking about *our* backoffice theme, if something is not needed anymore, we must remove it.
It's not a BC since it's not a part of the Public API, it's a UI / UX problem :)

Progi1984 (approved):

PierreRambaud:
Thanks @marionf

PR #16285, opened 2019-11-05
Fix ajax spinner not disappearing after Ajax is done

eternoendless:
| Questions     | Answers
| Branch?       | develop
| Description?  | This fixes a regression in develop where the ajax spinner icon wouldn't stop spinning after the dashboard is loaded
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | N/A
| How to test?  | Open up the dashboard, the spinner icon at the top left corner of the screen should disappear after a couple seconds.

This bug was a regression due to the jQuery version being upgraded. As noted in [this page](https://api.jquery.com/ajaxStart/):

> As of jQuery 1.9, all the handlers for the jQuery global Ajax events, including those added with the .ajaxStart() method, must be attached to document.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16285)

### DIFF
--- a/js/admin.js
+++ b/js/admin.js
@@ -750,21 +750,24 @@ $(document).ready(function()
       return copyMeta2friendlyURL()
   });
 
-  $('#ajax_running').ajaxStart(function() {
-    ajax_running_timeout = setTimeout(function() {showAjaxOverlay()}, 1000);
-  });
-
-  $('#ajax_running').ajaxStop(function() {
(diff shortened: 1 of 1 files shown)


matks (approved):

PierreRambaud:
Thanks @eternoendless

PR #16287, opened 2019-11-06
Don't duplicate currencies in DB when creating a formerly deleted one 

### LINKED ISSUE #15934: Creating a once deleted currency causes duplicate rows in database
**Describe the bug**
When a currency is added in PrestaShop it is impossible to add it a second time,  thus it prevent having duplicate
However if you delete a currency and then add it again a new row is inserted and the former one is still present with its deleted column set to 1 (soft delete)
This means the database has duplicated rows, and that different orders with the same currency do not use the same one in database

**To Reproduce**
1. Go to Internationalisation > Localization > Currencies
2. Add a new currency (for example AUD)
3. Go back to the currencies list and delete the newly created currency
4. In your database you will see that the row is still present but the deleted column is now equal to `1`
5. Recreate this same currency (AUD)
6. Now if you check you database will see two rows for the same AUD currency (one is deleted, not the other)

**Expected behaviour**
Rather than adding a new row in database, the previously deleted one should be updated (use the data from the creation form to update its content) AND most importantly its deleted value should be revert back to 0 thus avoiding creating duplicate rows

**Additional information**
PrestaShop version: N/A
PHP version: N/A

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Update currency in DB when the one we are creating already exists in DB (as soft deleted)
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15934
| How to test?  | See just below:<br><br>1.  Go to Internationalisation > Localization > Currencies<br>2.  Add a new currency (for example AUD)<br>3.  Go back to the currencies list and delete the newly created currency<br>4.  In your database you will see that the row is still present but the deleted column is now equal to  `1`<br>5.  Recreate this same currency (AUD)<br>6.  Now if you check you database you should not have a new entry,  same AUD currency should have been updated with the values you pushed in the form, and the field "deleted" set to 0.<br><br>**sql query for checking**:  `SELECT * FROM ps_currency WHERE iso_code = 'AUD'`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16287)


jolelievre:
Hello @matthieu-rolland Sorry I didn't someone would deal with this issue so fast ^^
I actually think it's a bit early to make it as this PR is still not merged https://github.com/PrestaShop/PrestaShop/pull/14972
It includes lots of rules about currencies so it should be merged before we manage this feature Besides I'm not sure the fallback between creation and update should be managed with the FormDataHandler because most of the domain rules are checked in the command handlers and they are different from a creation and an update
So either the creation handler launches an update command as a fallback or the controller is responsible for it, it can detect if the currency is already in database Or many catch an exception thrown by the creation handler when you try to add an existing currency (even a deleted one)

matthieu-rolland:
> Hello @matthieu-rolland Sorry I didn't someone would deal with this issue so fast ^^
> I actually think it's a bit early to make it as this PR is still not merged #14972
> It includes lots of rules about currencies so it should be merged before we manage this feature Besides I'm not sure the fallback between creation and update should be managed with the FormDataHandler because most of the domain rules are checked in the command handlers and they are different from a creation and an update
> So either the creation handler launches an update command as a fallback or the controller is responsible for it, it can detect if the currency is already in database Or many catch an exception thrown by the creation handler when you try to add an existing currency (even a deleted one)

@jolelievre : 

alright, I had no idea, ok 👍 

About your last point, I thought about it, and in the end I concluded that even if it's an update from a database point of view, it is still a creation from a software and user point of view: setting the value of a 'deleted' field from 0 to 1 is a creation. 

But I totally understand your point: I strongly hesitated on this one... and I'm still not sure about my take on this concept.

PierreRambaud:
#14972 has been merge, can we continue this pull request, or this isn't needed anymore?

jolelievre:
It's still useful (nice to have) But it's preferable to merge this one first https://github.com/PrestaShop/PrestaShop/pull/16383
So that all the currency handlers are finalized

jolelievre:
@matthieu-rolland I modified the behaviour Instead of managing to "undeletion" in the DataFormHandler I moved it in the creation handlers, this way the process will work even if you don't use a form the command and the handlers hold the whole logic
This also allows to add behat tests on this feature

And I removed the setter for deleted in the command object, because you can't edit a deleted currency And when you create it it is necessarily not deleted

atomiix (commented):
I see a bit of code duplicated in `AddOfficialCurrencyHandler` and `AddUnofficialCurrencyHandler`. Can't we implement the logic in `AbstractCurrencyHandler` or use a `trait` instead?

jolelievre:
@atomiix I cleaned the handlers code and moved most of them in the `AbstractCurrencyHandler`

src/Adapter/Currency/CommandHandler/AbstractCurrencyHandler.php
```diff
@@ -122,6 +137,87 @@ protected function applyPatternTransformations(Currency $entity, array $localize
...
+     * @param Currency $entity
+     * @param AddCurrencyCommand $command
+     *
+     * @throws CannotCreateCurrencyException
+     * @throws LanguageNotFoundException
+     * @throws LocalizationException
+     * @throws PrestaShopDatabaseException
+     * @throws PrestaShopException
+     */
+    protected function createEntity(Currency $entity, AddCurrencyCommand $command)
```
atomiix: As this function is taking the entity as parameter and is not "creating" it itself, maybe the name could be `addEntity` or `saveEntity`? Wdyt?

PierreRambaud: `saveEntity` :+1:

jolelievre: Yeah you're right I prefer `addEntity` to emphasize the difference with `updateEntity` Because in object model `save` is a mix of `add/update`

jolelievre: Done

atomiix (dismissed):
LGTM !

sarahdib:
@matthieu-rolland  I can't create an order using the default currency if I delete a currency and create it again.

PierreRambaud (approved):

atomiix (approved):

SD1982:
This pull request correctly fix the [15934](https://github.com/PrestaShop/PrestaShop/issues/15934) bug thanks @matthieu-rolland

PierreRambaud:
Thanks @matthieu-rolland

PR #16294, opened 2019-11-06
Fix email not translated when installing a new language

### LINKED ISSUE #16273: When I install a new language, the generated emails are not translated on first generation
**Describe the bug**
When I install a new language, the generated emails are not translated on first generation

**To Reproduce**
1. Go to International > Translations
2. Install new language "Polish"
3. The mails are generated in `mails/pl` but they are in englisg
4. If you regenerate them manually after it works fine

**Solution**
Check the update package part of the code, seems like the translations are not available
Maybe the translations are cached on the start of the request, so when mails are generated the new one is not available, maybe we should clean the cache right after the new language is installed

**Additional information**
PrestaShop version: 1.7.6
PHP version: N/A

atomiix:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | This fixes emails not being translated when installing a new language
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16273
| How to test?  | See #16273 ⚠ As it modifies quite a lot the language management, we should also test if mails are correctly generated during a fresh install (using something else than english) from web, and with CLI And check that the manual generation from the form also still works  ⚠

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16294)

### DIFF
--- a/classes/Context.php
+++ b/classes/Context.php
@@ -422,6 +422,8 @@ class ContextCore
         $translator->clearLanguage($locale);
 
         $adminContext = defined('_PS_ADMIN_DIR_');
+        // Do not load DB translations when $this->language is PrestashopBundle\Install\Language
+        // because it means that we're looking for the installer translations, so we're not yet connected to the DB
         $withDB = !$this->language instanceof PrestashopBundle\Install\Language;
         $theme = $this->shop !== null ? $this->shop->theme : null;
         (new TranslatorLanguageLoader($adminContext))->loadLanguage($translator, $locale, $withDB, $theme);
--- a/classes/Context.php
+++ b/classes/Context.php
@@ -25,12 +25,11 @@
  */
 use PrestaShop\PrestaShop\Adapter\SymfonyContainer;
 use PrestaShop\PrestaShop\Core\Localization\Locale;
-use PrestaShopBundle\Translation\Loader\SqlTranslationLoader;
 use PrestaShopBundle\Translation\TranslatorComponent as Translator;
+use PrestaShopBundle\Translation\TranslatorLanguageLoader;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Finder\Finder;
-use Symfony\Component\Translation\Loader\XliffFileLoader;
 
 /**
  * Class ContextCore.
@@ -405,7 +404,7 @@ class ContextCore
 
         // In case we have at least 1 translated message, we return the current translator.
         // If some translations are missing, clear cache
-        if ($locale === '' || count($translator->getCatalogue($locale)->all())) {
+        if ($locale === '' || null === $locale || count($translator->getCatalogue($locale)->all())) {
             return $translator;
         }
 
@@ -420,49 +419,13 @@ class ContextCore
             (new Filesystem())->remove($cache_file);
         }
 
+        $translator->clearLanguage($locale);
+
         $adminContext = defined('_PS_ADMIN_DIR_');
-        $translator->addLoader('xlf', new XliffFileLoader());
-
-        $sqlTranslationLoader = new SqlTranslationLoader();
-        if (null !== $this->shop) {
-            $sqlTranslationLoader->setTheme($this->shop->theme);
-        }
-
-        $translator->addLoader('db', $sqlTranslationLoader);
-        $notName = $adminContext ? '^Shop*' : '^Admin*';
-
-        $finder = Finder::create()
-            ->files()
-            ->name('*.' . $locale . '.xlf')
-            ->notName($notName)
-            ->in($this->getTranslationResourcesDirectories());
-
-        foreach ($finder as $file) {
-            list($domain, $locale, $format) = explode('.', $file->getBasename(), 3);
-
-            $translator->addResource($format, $file, $locale, $domain);
-            if (!$this->language instanceof PrestashopBundle\Install\Language) {
-                $translator->addResource('db', $domain . '.' . $locale . '.db', $locale, $domain);
-            }
-        }
+        $withDB = !$this->language instanceof PrestashopBundle\Install\Language;
+        $theme = $this->shop !== null ? $this->shop->theme : null;
+        (new TranslatorLanguageLoader($adminContext))->loadLanguage($translator, $locale, $withDB, $theme);
 
         return $translator;
-    }
-
-    /**
-     * @return array
-     */
-    protected function getTranslationResourcesDirectories()
-    {
-        $locations = array(_PS_ROOT_DIR_ . '/app/Resources/translations');
-
-        if (null !== $this->shop) {
-            $activeThemeLocation = _PS_ROOT_DIR_ . '/themes/' . $this->shop->theme_name . '/translations';
-            if (is_dir($activeThemeLocation)) {
-                $locations[] = $activeThemeLocation;
-            }
-        }
-
-        return $locations;
     }
 }
--- a/classes/lang/DataLang.php
+++ b/classes/lang/DataLang.php
@@ -23,7 +23,9 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
+use PrestaShop\PrestaShop\Adapter\SymfonyContainer;
 use PrestaShopBundle\Translation\TranslatorComponent as Translator;
+use PrestaShopBundle\Translation\TranslatorLanguageLoader;
 
 class DataLangCore
 {
@@ -46,30 +48,12 @@ class DataLangCore
     {
         $this->locale = $locale;
 
-        $legacyTranslator = Context::getContext()->getTranslator();
-        $legacyLocale = $legacyTranslator->getLocale();
+        $this->translator = SymfonyContainer::getInstance()->get('translator');
+        $isAdminContext = defined('_PS_ADMIN_DIR_');
 
-        if ($legacyLocale === $this->locale) {
-            $this->translator = $legacyTranslator;
-        } else {
-            $this->translator = new Translator(
-                $this->locale,
-                null,
-                _PS_CACHE_DIR_ . '/translations/' . $this->locale,
-                false
-            );
-
-            $this->translator->addLoader('xlf', new \Symfony\Component\Translation\Loader\XliffFileLoader());
-
-            $finder = \Symfony\Component\Finder\Finder::create()
-                ->files()
-                ->name('*.' . $this->locale . '.xlf')
-                ->in((_PS_ROOT_DIR_ . '/app/Resources/translations'));
-
-            foreach ($finder as $file) {
-                list($domain, $locale, $format) = explode('.', $file->getBasename(), 3);
-                $this->translator->addResource($format, $file, $locale, $domain);
-            }
+        if (!$this->translator->isLanguageLoaded($this->locale)) {
+            (new TranslatorLanguageLoader($isAdminContext))->loadLanguage($this->translator, $this->locale);
+            $this->translator->getCatalogue($this->locale);
         }
     }
 
--- a/composer.json
+++ b/composer.json
@@ -157,9 +157,9 @@
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache"
         ],
         "test-all": [
+            "composer phpunit-admin",
             "composer phpunit-legacy",
             "composer unit-tests",
-            "composer phpunit-admin",
             "composer phpunit-sf",
             "composer integration-tests",
             "composer phpunit-controllers",
--- a/src/Core/Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler.php
+++ b/src/Core/Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler.php
@@ -33,7 +33,6 @@ use PrestaShop\PrestaShop\Core\Language\LanguageRepositoryInterface;
 use PrestaShop\PrestaShop\Core\MailTemplate\MailTemplateGenerator;
 use PrestaShop\PrestaShop\Core\MailTemplate\ThemeCatalogInterface;
 use PrestaShop\PrestaShop\Core\MailTemplate\ThemeInterface;
-use Symfony\Component\Translation\Loader\ArrayLoader;
 use Symfony\Component\Translation\TranslatorInterface;
 
 /**
@@ -99,34 +98,9 @@ class GenerateThemeMailTemplatesCommandHandler implements GenerateThemeMailTempl
         /** @var ThemeInterface $theme */
         $theme = $this->themeCatalog->getByName($command->getThemeName());
 
-        $this->cleanTranslatorLocaleCache($command->getLanguage());
-
         $coreMailsFolder = $command->getCoreMailsFolder() ?: $this->defaultCoreMailsFolder;
         $modulesMailFolder = $command->getModulesMailFolder() ?: $this->defaultModulesMailFolder;
 
         $this->generator->generateTemplates($theme, $language, $coreMailsFolder, $modulesMailFolder, $command->overwriteTemplates());
-    }
-
-    /**
-     * When installing a new Language, if it's a new one the Translator component can't manage it because its cache is
-     * already filled with the default one as fallback. We force the component to update its cache by adding a fake
-     * resource for this locale (this is the only way clean its local cache)
-     *
-     * @param string $locale
-     */
-    private function cleanTranslatorLocaleCache($locale)
-    {
-        if (!method_exists($this->translator, 'addLoader')
-            || !method_exists($this->translator, 'addResource')
-        ) {
-            return;
-        }
-
-        $this->translator->addLoader('array', new ArrayLoader());
-        $this->translator->addResource(
-            'array',
-            ['Fake clean cache message' => 'Fake clean cache message'],
-            $locale
-        );
     }
 }
--- a/src/PrestaShopBundle/Translation/TranslatorLanguageLoader.php
+++ b/src/PrestaShopBundle/Translation/TranslatorLanguageLoader.php
@@ -1,0 +1,109 @@
+<?php
+
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShopBundle\Translation;
+
+use PrestaShop\PrestaShop\Core\Addon\Theme\Theme;
+use PrestaShopBundle\Translation\Loader\SqlTranslationLoader;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Translation\Loader\XliffFileLoader;
+use Symfony\Component\Translation\TranslatorInterface;
+
+class TranslatorLanguageLoader
+{
+    const TRANSLATION_DIR = _PS_ROOT_DIR_ . '/app/Resources/translations';
+    /**
+     * @var bool
+     */
+    private $isAdminContext;
+
+    /**
+     * TranslatorLanguageLoader constructor.
+     *
+     * @param $isAdminContext
+     */
+    public function __construct($isAdminContext)
+    {
+        $this->isAdminContext = $isAdminContext;
+    }
+
+    /**
+     * Loads a language into a translator
+     *
+     * @param TranslatorInterface $translator Translator to modifiy
+     * @param string $locale Locale code for the language to load
+     * @param bool $withDB [default=true] Whether to load translations from the database or not
+     * @param Theme|null $theme [default=false] Currently active theme (Front office only)
+     */
+    public function loadLanguage(TranslatorInterface $translator, $locale, $withDB = true, Theme $theme = null)
+    {
+        if (!$translator->isLanguageLoaded($locale)) {
+            $translator->addLoader('xlf', new XliffFileLoader());
+
+            if ($withDB) {
+                $sqlTranslationLoader = new SqlTranslationLoader();
+                if (null !== $theme) {
+                    $sqlTranslationLoader->setTheme($theme);
+                }
+                $translator->addLoader('db', $sqlTranslationLoader);
+            }
+
+            $finder = Finder::create()
+                ->files()
+                ->name('*.' . $locale . '.xlf')
+                ->notName($this->isAdminContext ? '^Shop*' : '^Admin*')
+                ->in($this->getTranslationResourcesDirectories($theme));
+
+            foreach ($finder as $file) {
+                list($domain, $locale, $format) = explode('.', $file->getBasename(), 3);
+                $translator->addResource($format, $file, $locale, $domain);
+                if ($withDB) {
+                    $translator->addResource('db', $domain . '.' . $locale . '.db', $locale, $domain);
+                }
+            }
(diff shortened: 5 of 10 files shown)


src/Core/Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler.php
```diff
@@ -108,25 +108,20 @@ public function handle(GenerateThemeMailTemplatesCommand $command)
...
     /**
-     * When installing a new Language, if it's a new one the Translator component can't manage it because its cache is
-     * already filled with the default one as fallback. We force the component to update its cache by adding a fake
-     * resource for this locale (this is the only way clean its local cache)
+     * When installing a new language, we must add the lang pack to the translator
      *
      * @param string $locale
      */
-    private function cleanTranslatorLocaleCache($locale)
+    private function addLanguageToTranslator(string $locale)
```
PierreRambaud: 5.6compatible, you can't do that :)
```suggestion
    private function addLanguageToTranslator($locale)
```

src/Core/Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler.php
```diff
@@ -108,25 +108,20 @@ public function handle(GenerateThemeMailTemplatesCommand $command)
...
-        ) {
-            return;
+        $finder = Finder::create()
+            ->files()
+            ->name('*.' . $locale . '.xlf')
+            ->in(_PS_ROOT_DIR_ . '/app/Resources/translations/' . $locale);
+
+        foreach ($finder as $file) {
+            list($domain, $locale, $format) = explode('.', $file->getBasename(), 3);
+            $this->translator->addResource($format, $file, $locale, $domain);
```
eternoendless: You'll have to add this part as well in order to include translations from DB:

https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Context.php#L451-L453

eternoendless (changes requested):

src/Core/Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler.php
```diff
@@ -108,25 +108,21 @@ public function handle(GenerateThemeMailTemplatesCommand $command)
...
-     * resource for this locale (this is the only way clean its local cache)
+     * When installing a new language, we must add the lang pack to the translator
      *
-     * @param string $locale
+     * @param $locale
      */
-    private function cleanTranslatorLocaleCache($locale)
+    private function addLanguageToTranslator($locale)
     {
-        if (!method_exists($this->translator, 'addLoader')
```
jolelievre, line 119: Be careful, I remember this check was here for a reason because sometimes the injected translator was not `PrestaShopBundle\Translation\Translator` but `PrestaShopBundle\Translation\DataCollectorTranslator` which doesn't have the method `addRessource` nor `addLoader`
Although since there have been modifications on the translator (we had 2 different components) by @matthieu-rolland maybe this fixed the issue

src/Core/Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler.php
```diff
@@ -108,25 +108,21 @@ public function handle(GenerateThemeMailTemplatesCommand $command)
     }
 
     /**
-     * When installing a new Language, if it's a new one the Translator component can't manage it because its cache is
-     * already filled with the default one as fallback. We force the component to update its cache by adding a fake
-     * resource for this locale (this is the only way clean its local cache)
+     * When installing a new language, we must add the lang pack to the translator
      *
-     * @param string $locale
+     * @param $locale
```
eternoendless: Don't remove the parameter signature here!
```suggestion
     * @param string $locale
```

src/Core/Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler.php
```diff
@@ -108,25 +108,21 @@ public function handle(GenerateThemeMailTemplatesCommand $command)
...
      */
-    private function cleanTranslatorLocaleCache($locale)
+    private function addLanguageToTranslator($locale)
     {
-        if (!method_exists($this->translator, 'addLoader')
-            || !method_exists($this->translator, 'addResource')
-        ) {
-            return;
+        $finder = Finder::create()
```
eternoendless: You can avoid re-populating the translator if it's already done
```suggestion
		if (0 < count($this->translator->getCatalogue($locale)->all()) {
		    // avoid re-populating the translator
		    return;
		}

        $finder = Finder::create()
```

eternoendless (changes requested):

src/Core/Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler.php
eternoendless: I think it's fixed but it should be tested.

FYI you get DataCollectorTranslator when you're in dev mode (it tracks translation hits for the Symfony profiler)

atomiix: Yeah, actually if I check if those methods exists it will return false because I got a `DataCollectorTranslator` **but** as `DataCollectorTranslator` is wrapping the `Translator` class, those methods are still available via the magic function `__call` that call the translator functions and make it works.

src/Core/Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler.php
atomiix: What if we're updating the lang pack? Don't we need to update the translator with the latest translations?

eternoendless: I see what you mean. But in case of the current language being updated, the resources have already been added to the translator. What happens to the translator if you add the same resource twice?

I have the feeling that we need to factorize this translator refresh functionality out of this class and have it shared with the one in Context.

classes/Context.php
```diff
@@ -391,6 +392,16 @@ public function getTranslator($isInstaller = false)
         return $this->translator;
     }
 
+    /**
+     * @param string $locale
+     * @param string $notName
+     */
+    public function addLanguageToTranslator($locale, $notName = null)
```
jolelievre: It seems `$notName` is not really useful here (you don't use it in the `Language` class)

classes/Context.php
```diff
@@ -433,22 +444,29 @@ public function getTranslatorFromLocale($locale)
         $translator->addLoader('db', $sqlTranslationLoader);
         $notName = $adminContext ? '^Shop*' : '^Admin*';
 
+        $this->doAddLanguageToTranslator($translator, $locale, $notName);
```
atomiix: @jolelievre I'm using `$notName` there ;)

classes/Context.php
```diff
@@ -391,6 +392,15 @@ public function getTranslator($isInstaller = false)
         return $this->translator;
     }
 
+    /**
+     * @param string $locale
+     */
+    public function addLanguageToTranslator($locale)
```
eternoendless: You're adding responsibilities to Context that aren't theirs.

I think that you should extract a class `TranslationLanguageLoader` that would basically take over the responsibility for loading a language into an already instantiated translator.

It would work like this:

```
$isAdminContext = defined('_PS_ADMIN_DIR_');
$translator = Context::getContext()->getTranslator();
(new TranslatorLanguageLoader($isAdminContext))->loadLanguage($translator, $localeCode); // this returns void
```

This way, I don't think that you'd need to change the `getTranslator()` signature anymore.

eternoendless (changes requested):

classes/lang/DataLang.php
```diff
@@ -46,22 +48,13 @@ public function __construct($locale)
...
-        if ($legacyLocale === $this->locale) {
-            $this->translator = $legacyTranslator;
-        } else {
-            $this->translator = new Translator(
-                $this->locale,
-                null,
-                _PS_CACHE_DIR_ . '/translations/' . $this->locale,
-                false
-            );
+        if ($translatorLocale !== $this->locale) {
```
jolelievre: Waow! So fixing "only" this allows to manage the new language correctly?!

atomiix: Yes... Took time to find it though 😅
I just nedd to follow @eternoendless's recommandation by adding a new `TranslatorLanguageLoader` and we should be good with that PR

src/PrestaShopBundle/Translation/TranslatorLanguageLoader.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * @param TranslatorInterface $translator
+     * @param string $locale
+     * @param Theme|null $theme
+     */
+    public function loadLanguage(TranslatorInterface $translator, $locale, Theme $theme = null)
+    {
+        if ($translator->getLocale() !== $locale) {
+            $translator->addLoader('xlf', new XliffFileLoader());
```
jolelievre, line 65: Maybe you check if the loader are already present? If this method is called several times you will have multiple instances of the same loader

atomiix: From what I saw, `addLoader()` is doing `$this->loaders[$format] = $loader;` and there was no public method to get the loaders. But if you think it's better to check I can do it :)
Wdyt?

classes/lang/DataLang.php
```diff
@@ -46,31 +47,10 @@ public function __construct($locale)
...
-                ->files()
-                ->name('*.' . $this->locale . '.xlf')
-                ->in((_PS_ROOT_DIR_ . '/app/Resources/translations'));
-
-            foreach ($finder as $file) {
-                list($domain, $locale, $format) = explode('.', $file->getBasename(), 3);
-                $this->translator->addResource($format, $file, $locale, $domain);
-            }
-        }
+        (new TranslatorLanguageLoader($isAdminContext))->loadLanguage($this->translator, $this->locale);
```
eternoendless: As discussed, the problem with this is that the contextual translator is set to the current user's language. Doing this will take over the user's language in unexpected ways.

For example, let's say the user is in French and he installs Polish. Doing this will switch the user's translator to Polish until the next page reload.

This means that whenever a translation operation needs to be performed in a locale that's different from the user's current one, a new translator must be instantiated in that language and passed along as a parameter.

So you just need to call `Context::getContext()->getTranslatorFromLocale($this->locale)`

classes/Context.php
```diff
@@ -423,30 +422,8 @@ public function getTranslatorFromLocale($locale)
...
-        foreach ($finder as $file) {
-            list($domain, $locale, $format) = explode('.', $file->getBasename(), 3);
-
-            $translator->addResource($format, $file, $locale, $domain);
-            if (!$this->language instanceof PrestashopBundle\Install\Language) {
-                $translator->addResource('db', $domain . '.' . $locale . '.db', $locale, $domain);
-            }
-        }
+        $theme = $this->shop !== null ? $this->shop->theme : null;
+        (new TranslatorLanguageLoader($adminContext))->loadLanguage($translator, $locale, $theme);
```
eternoendless: Following my comment on DataLang, you may extract further logic from this method and create a static factory which would replace `TranslatorLanguageLoader` which would no longer be a necessary as a separate feature:

```php
$options = (new TranslationFactoryOptions())
  ->setIsBackoffice(true)
  ->setThemeName(null) // not needed when setIsBackoffice is true
  ->setCleanCache(true);
$newTranslator = TranslationFactory::buildTranslator($locale, $options);
```

eternoendless (changes requested):

classes/lang/DataLang.php
eternoendless: Nevermind, this is not necessary.

src/PrestaShopBundle/Translation/TranslatorLanguageLoader.php
eternoendless: You can also only add loaders if the catalogue is empty for that locale.

classes/Context.php
eternoendless: Nevermind, not necessary

src/PrestaShopBundle/Translation/TranslatorLanguageLoader.php
```diff
@@ -0,0 +1,82 @@
...
+
+            $sqlTranslationLoader = new SqlTranslationLoader();
+            if (null !== $theme) {
+                $sqlTranslationLoader->setTheme($theme);
```
jolelievre: I'm wondering since the loader depends on the theme, shouldn't the loader be also depends on it Something like:
```
$translator->addLoader('db-'.$theme, $sqlTranslationLoader);
```
Or are we sure that the theme stays unmodified during the whole process (for each calls)

src/PrestaShopBundle/Translation/PrestaShopTranslatorInterface.php
```diff
@@ -0,0 +1,42 @@
...
+
+use Symfony\Component\Translation\TranslatorInterface;
+
+/**
+ * Interface PrestaShopTranslatorInterface used by PrestaShop translators
+ */
+interface PrestaShopTranslatorInterface extends TranslatorInterface
+{
+    /**
+     * @param $locale
```
eternoendless:
```suggestion
     * @param string $locale
```

src/PrestaShopBundle/Translation/PrestaShopTranslatorTrait.php
```diff
@@ -130,6 +130,16 @@ public function transChoice($id, $number, array $parameters = array(), $domain =
         return vsprintf(parent::transChoice($id, $number, array(), $domain, $locale), $parameters);
     }
 
+    /**
+     * @param $locale
```
eternoendless:
```suggestion
     * @param string $locale
```

src/PrestaShopBundle/Translation/PrestaShopTranslatorTrait.php
```diff
@@ -130,6 +130,16 @@ public function transChoice($id, $number, array $parameters = array(), $domain =
...
+     *
+     * @return bool
+     */
+    public function isCatalogLoaded($locale)
+    {
+        return isset($this->catalogues[$locale]);
```
eternoendless: Shouldn't it be better to check for emptyness?
```suggestion
        return !empty($this->catalogues[$locale]);
```

src/PrestaShopBundle/Translation/TranslatorLanguageLoader.php
eternoendless: I don't think the theme is ever modified during runtime

eternoendless (commented):
Just a couple comments!

eternoendless (dismissed):

eternoendless:
Looks good to me, I'd like an approval from @jolelievre as well

src/PrestaShopBundle/Translation/TranslatorLanguageLoader.php
jolelievre: true!

jolelievre (dismissed):

Robin-Fischer-PS:
Hi,

Change of language on the install is KO with this PR.

Thanks ;)

jolelievre:
> Hi,
> 
> Change of language on the install is KO with this PR.
> 
> Thanks ;)

Woow that's a problem :P What do you mean by "change of language" Is it in the back office or the front office. Can you provide more detailed actions to test it and maybe a screenshot or stacktrace of the thrown exception (if any)

classes/Context.php
```diff
@@ -422,49 +422,15 @@ public function getTranslatorFromLocale($locale)
...
-        $translator->addLoader('db', $sqlTranslationLoader);
-        $notName = $adminContext ? '^Shop*' : '^Admin*';
-
-        $finder = Finder::create()
-            ->files()
-            ->name('*.' . $locale . '.xlf')
-            ->notName($notName)
-            ->in($this->getTranslationResourcesDirectories());
+        // only way to reset the catalogue
+        $translator->addLoader('Array', new ArrayLoader());
```
jolelievre: Maybe you should put this in the `TranslatorLanguageLoader::loadLanguage` method?

src/PrestaShopBundle/Translation/TranslatorLanguageLoader.php
```diff
@@ -0,0 +1,105 @@
...
+     */
+    public function __construct($isAdminContext)
+    {
+        $this->isAdminContext = $isAdminContext;
+    }
+
+    /**
+     * @param PrestaShopTranslatorInterface $translator
+     * @param string $locale
+     * @param Theme|null $theme
```
jolelievre: The PHPDoc is not in the right order (or the function's parameters)

classes/Context.php
```diff
@@ -422,49 +422,15 @@ public function getTranslatorFromLocale($locale)
...
-        foreach ($finder as $file) {
-            list($domain, $locale, $format) = explode('.', $file->getBasename(), 3);
-
-            $translator->addResource($format, $file, $locale, $domain);
-            if (!$this->language instanceof PrestashopBundle\Install\Language) {
-                $translator->addResource('db', $domain . '.' . $locale . '.db', $locale, $domain);
-            }
-        }
+        $adminContext = defined('_PS_ADMIN_DIR_');
+        $withDB = !$this->language instanceof PrestashopBundle\Install\Language;
```
jolelievre, line 427: I don't understand this check? In which case don't you want the database to be loaded? Maybe it's worth a comment here

Besides what other type can `$this->language` be? `\Language`? So it means you only disable the db loading during install process? Why?

classes/Context.php
```diff
@@ -422,49 +422,15 @@ public function getTranslatorFromLocale($locale)
...
-        $notName = $adminContext ? '^Shop*' : '^Admin*';
-
-        $finder = Finder::create()
-            ->files()
-            ->name('*.' . $locale . '.xlf')
-            ->notName($notName)
-            ->in($this->getTranslationResourcesDirectories());
+        // only way to reset the catalogue
+        $translator->addLoader('Array', new ArrayLoader());
+        $translator->addResource('Array', [], $locale);
```
eternoendless: This could be better managed inside the translator itself:

```php
public function resetCatalogue($locale)
{
    unset($this->catalogues[$locale]);
}
```

Just add this in the Translator trait and in the PrestaShopTranslatorInterface

src/PrestaShopBundle/Translation/TranslatorLanguageLoader.php
```diff
@@ -0,0 +1,105 @@
...
+     * @param bool $withDB
```
eternoendless: Arguments are not in the right order

eternoendless: Consider writing descriptions for public functions and arguments.

For example

```php
/**
 * Loads a language into a translator
 *
 * @param PrestaShopTranslatorInterface $translator Translator to modifiy
 * @param string $locale Locale code for the language to load
 * @param bool $withDB [default=true] Whether to load translations from the database or not
 * @param Theme|null $theme [default=false] Currently active theme (Front office only)
 */

**Review by matks — DISMISSED** [review]

File: src/PrestaShopBundle/Translation/PrestaShopTranslatorInterface.php [hunk 22]
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * @param string $locale
+     *
+     * @return bool
+     */
+    public function isCatalogLoaded($locale);
+
+    /**
+     * @param $locale
```
eternoendless:
```suggestion
     * @param string $locale
```

classes/Context.php
atomiix: We don't need to load DB translations when `$this->language` in an instance of `PrestashopBundle\Install\Language` because that's only translations for the web installer and when we load the installer language we didn't configured the DB yet...

src/PrestaShopBundle/Translation/PrestaShopTranslatorTrait.php
```diff
@@ -130,6 +130,24 @@ public function transChoice($id, $number, array $parameters = array(), $domain =
...
+     *
+     * @return bool
+     */
+    public function isCatalogLoaded($locale)
+    {
+        return !empty($this->catalogues[$locale]);
+    }
+
+    /**
+     * @param $locale
```
eternoendless:
```suggestion
     * @param string $locale
```

matks (approved):

matks (dismissed):

src/PrestaShopBundle/Translation/TranslatorLanguageLoader.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * @param Theme|null $theme
+     *
+     * @return array
+     */
+    protected function getTranslationResourcesDirectories(Theme $theme = null)
+    {
+        $locations = [_PS_ROOT_DIR_ . '/app/Resources/translations'];
```
PierreRambaud: This one can be a const (I mean `_PS_ROOT_DIR_ . '/app/Resources/translations'`)

classes/Context.php
jolelievre: oooookay, this makes sense but maybe you can add a comment explaining that, because it's not that intuitive at first ^^

Robin-Fischer-PS:
Hi,

I've a new bug : 

When I import a localization pack of a country with several official languages (for example, Belgium), only one language has the generated emails in the correct language. The other languages are in English.

**After manual regeneration (in Design > Email), the emails are still in English.**

Some details for my Belgium example : 

If only English is installed, French will be OK, German and Dutch will be NOK.
If English and French are already installed (and OK), German will also be OK, only Dutch will be NOK.

In 1752, it's OK.

matks:
> Hi,
> 
> I've a new bug :
> 
> When I import a localization pack of a country with several official languages (for example, Belgium), only one language has the generated emails in the correct language. The other languages are in English.
> 
> **After manual regeneration (in Design > Email), the emails are still in English.**
> 
> Some details for my Belgium example :
> 
> If only English is installed, French will be OK, German and Dutch will be NOK.
> If English and French are already installed (and OK), German will also be OK, only Dutch will be NOK.
> 
> In 1752, it's OK.

So, if I understand correctly, we have 2 reproductable failing scenarios (at least):

## Scenario 1

1. I install a shop in English. I download a localization pack for Belgium. This pack contains French, Dutch, and German.
2. I generate the mails for languages French, Dutch, German.
3. French, Dutch, German mails mails are generated but are not translated. They are in English.

## Scenario 2

1. I install a shop in French. This means 2 languages are available on my shop: French and English. I make sure both are activated.
2. I download a localization pack for Belgium. This pack contains French, Dutch, and German.
3. I generate the mails for languages French, Dutch, German.
4. French, and German mails are generated and translated. But Dutch mails are in English.

Is that correct @Robin-Fischer-PS ?

Robin-Fischer-PS:
Not exactly @matks 

Scenario 2 is OK. For Scenario 1, step 3 is : 
French mails are generated and translated. Dutch, German mails mails are generated too but are not translated. They are in English.

Robin-Fischer-PS:
Hi,

New bug. When I do a fresh install with a country with 3 or more official languages (Belgium, Switzerland), it is KO.

pr16294 fail install

PierreRambaud:
@Robin-Fischer-PS as defined in the documentation https://devdocs.prestashop.com/1.7/basics/installation/system-requirements/
The minimum memory_limit is 256M. Could you please increase the value and retry?

Robin-Fischer-PS:
With memory_limit = 256M, it's OK :wink:

matks (dismissed):

matks:
@atomiix Since this PR has been unscoped and we have a few more time to fix it, could you handle Pierre's feedback https://github.com/PrestaShop/PrestaShop/pull/16294#discussion_r349204994 ? 😉

classes/Context.php
matks: 🆙

composer.json
```diff
@@ -182,7 +182,7 @@
         ],
         "phpunit-admin": [
             "@composer create-test-db",
-            "@php -d date.timezone=UTC ./vendor/bin/phpunit -c tests-legacy/phpunit-admin.xml"
+            "php -d date.timezone=UTC ./vendor/bin/phpunit -c tests-legacy/phpunit-admin.xml"
```
matks: 👍

matks (dismissed):

composer.json
PierreRambaud: NOPE NOPE AND NOPE :scream:

PierreRambaud: NOPE NOPE AND NOPE !!!! :scream:

PierreRambaud (dismissed):
Change composer.json

PierreRambaud (approved):

Robin-Fischer-PS:
Hi @atomiix !

It gives me an immense joy and great satisfaction to announce you that this PR is QA Approved :heavy_check_mark: !

Congrats, now you can relax and enjoy the simple pleasures of life, like working on cart rules :trollface: 

(FYI : I tested with a 128MB configuration : It works fine on install with 3 languages)

Thanks !

PR #16307, opened 2019-11-07
Cast employeeId to int in the bulk update employee command

### LINKED ISSUE #16246: The bulk action doesn't work in the employees page
**Describe the bug**
In the BO => Advanced Parameters => Team => Employees page => we cannot delete or enable or disable employees with bulk action => an error is displayed
image

**To Reproduce**
1. Go to the BO => Advanced Parameters => Team => Employees page
2. Create a new Employee
3. Try to delete/Enable/Disable this new employee with bulk action

**Screenshots**
https://drive.google.com/file/d/1w3T6dtwnnuMdZsme3S9tDkA9aXwTzitl/view

**Additional information**
PrestaShop version: develop
PHP version: 7.1

atomiix:
| Questions     | Answers
| Branch?       | develop
| Description?  | This force `EmployeeId` constructor parameter to be an `int`
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16246
| How to test?  | Go to `Configure > Advanced Parameters > Team > Employees page` and try a bulk action

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16307)

### DIFF
--- a/src/Core/Domain/Employee/Command/BulkDeleteEmployeeCommand.php
+++ b/src/Core/Domain/Employee/Command/BulkDeleteEmployeeCommand.php
@@ -60,7 +60,7 @@ class BulkDeleteEmployeeCommand
     private function setEmployeeIds(array $employeeIds)
     {
         foreach ($employeeIds as $employeeId) {
-            $this->employeeIds[] = new EmployeeId($employeeId);
+            $this->employeeIds[] = new EmployeeId((int) $employeeId);
         }
     }
 }
--- a/src/Core/Domain/Employee/Command/BulkUpdateEmployeeStatusCommand.php
+++ b/src/Core/Domain/Employee/Command/BulkUpdateEmployeeStatusCommand.php
@@ -76,7 +76,7 @@ class BulkUpdateEmployeeStatusCommand
     private function setEmployeeIds(array $employeeIds)
     {
         foreach ($employeeIds as $employeeId) {
-            $this->employeeIds[] = new EmployeeId($employeeId);
+            $this->employeeIds[] = new EmployeeId((int) $employeeId);
         }
     }
 }


Progi1984 (dismissed):

Progi1984:
Ping @atomiix : Travis is broken in Unit Tests.

PierreRambaud (dismissed):
You also need to type int the constructor of EmployeeId ValueObject :thinking:

atomiix:
@PierreRambaud problem is if we type int the constructor of EmployeeId ValueObject, tests fail because they are testing that if we send a string to the constructor it should throw an exception but if the constructor is type int, the string will be casted to int and so no exception will be thrown.

To have an exception to be thrown, It would need to set `declare(strict_type=1)` in the EmployeeId ValueObject php file as well as in every php file that is instantiating an EmployeeId ValueObject.

We had a discussion with @eternoendless and decided that we should add `declare(strict_type=1)` only to new files. So, that's why I didn't type int the EmployeeId constructor.

What do you think?

PierreRambaud:
Why not. I was more thinking about adding `declare(strict_type=1)` each time we want to use type hint :) But he's the boss :sweat_smile: 
But you need to rename your pull request, you're not adding an `int` type, you're casting in `int`

PierreRambaud (dismissed):

Robin-Fischer-PS:
Hi @atomiix ,

The bug still appears when trying to delete the employee (for enable/disable bulk actions, it is OK) :/

Thanks !

jolelievre (approved):

PierreRambaud:
Thanks @atomiix

PR #16318, opened 2019-11-07
Make migrated order page's grid rows clickable

### LINKED ISSUE #14736: Right click "open in new tab" over the full width table line
**USER STORY**
As a merchant, I want to able to open in a new tab my the orders from the order list.

**WHY**
Merchants cannot open their orders in another tab even if the action is possible through the order view button. Making the entire line clickable would provide a better user experience and increase merchant productivity.

**SPECIFICATION:**
A right-click on any order then "open in a new tab"

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Make row clickable in migrated order page. Also make which column are clickable inside a clickable row configurable. 
| Type?         | Improvement
| Category?     | BO
| BC breaks?    | yes, se below
| Deprecations? | no
| Fixed ticket? | Fixes #14736
| How to test?  | See below

#### Feature description

Grid rows for Orders listing ( `/admin-dev/sell/orders/orders/` ) should be clickable. However the columns which have already some actions (links, dropdown, buttons ...) do not have the clickable behavior. So only columns without action are clickable.

#### PR Scope

We actually rewrote the "clickable row" behavior used in **all grids** to use the same everywhere.

#### How to test ?

Check that rows are clickable in migrated order pages `/admin-dev/sell/orders/orders/` except for columns. Check that other grids have a similar behavior.

#### BC Break ?

Before, the "clickable row" behavior was available on column/rows with selectors `td.data-type, td.identifier-type:not(:has(input)), td.badge-type, td.position-type` . We removed these to replace them with the `td.clickable` selector and made sure that standard PrestaShop columns have this class.

However if a developer used the Grid component to build its own column, this might be a BC break for him when he upgrades PS as the "clickable" behavior will only work if his column get the new selector too.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16318)


admin-dev/themes/new-theme/js/components/grid/extension/link-row-action-extension.js
```diff
@@ -70,8 +70,16 @@ export default class LinkRowActionExtension {
...
+         * @todo: only td.clickable should be in this list,
+         *  you could remove all the rest and it would work,
+         *  we keep it in 1.7.7 for retrocompatibilty purpose,
+         *  just in case some people extend the template and won't have td.clickable
+         *  (only td.indentifier-type was removed)
+         *
+         * Columns are know specified as clickable in the grid definitions
          */
-        const clickableCells = $('td.data-type, td.identifier-type, td.badge-type, td.position-type', $parentRow)
+        const clickableCells = $('td.data-type, td.badge-type, td.position-type, td.clickable', $parentRow)
```
atomiix: Why keeping `td.data-type, td.badge-type, td.position-type`? If those cells are set "clickable" in the grid definition, why `td.clickable` wouldn't be enough?

edit: my bad I didn't read the @todo above...

jolelievre: Why remove the `td.identifier-type` (there is a typo by the way)
I understand the issue when there is a checkbox, but it's not always the case

admin-dev/themes/new-theme/js/components/grid/extension/link-row-action-extension.js
```diff
@@ -70,8 +70,16 @@ export default class LinkRowActionExtension {
...
          * Only search for cells with non clickable contents to avoid conflicts with
          * previous cell behaviour (action, toggle, ...)
+         *
+         * @todo: only td.clickable should be in this list,
+         *  you could remove all the rest and it would work,
+         *  we keep it in 1.7.7 for retrocompatibilty purpose,
+         *  just in case some people extend the template and won't have td.clickable
+         *  (only td.indentifier-type was removed)
+         *
+         * Columns are know specified as clickable in the grid definitions
```
jolelievre: Columns are **now**

jolelievre (commented):
I don't like this solution because it force you to change ALL the available columns just to add a clickable class Whereas this behavior was already managed in the javascript which only targeted appropriate cells If some were missing then it would simpler to just update the `clickableCells` pattern in the javascript.

Besides even as is, I don't think it matches the expected behaviour. The `LinkRowActionExtension` is able to intercept the click on appropriate cells and then redirect to the appropriate link. But in this case we want more than that, we want to be able to open the link in another tab, which was not managed here.

There are few solutions:
- update `LinkRowActionExtension` to add a parameter that allows opening the link in another target (_blank) but then it will happen even on simple click
- create a custom right click menu for this list, but I don't if this what's expected by the product team, and if yes should it contain only `View in another page` or some other actions? (perhaps all those available for the row?)
- add a link on the row or cells (an `a` tag) but I don't think this is supposed to be authorized semantically speaking

matks:
As discussed with MatShir and @jolelievre:
- scope has changed: we aim to have clickable rows that open in the same tab
- we dont think keeping previous selectors is useful so we only apply the behavior on columns labeled with `clickable`

matks (dismissed):

matks:
@jolelievre I applied the changes we discussed together 💪

PierreRambaud (dismissed):

Robin-Fischer-PS:
Hi @matthieu-rolland :smile: , it's QA time !

On orders' grid it's OK.

However, I have some issues on others grid : 

- On International > Localization > Languages : Cannot click on "flag" column. It should be possible since there is no action on this column.

- On Catalog > Brands & Suppliers > Suppliers : "Logo" column should be clickable.

- On Catalog > Categories : "ID" column is clickable, it shouldn't be.


I tested these grids, please tell me if I miss something : 
logs
emails
employees
customers
currencies
webservice
traffic & SEO
categories
contacts
languages
taxes
profiles
catégories cms
pages cms
brands
brands addresses
suppliers

Thanks !

src/Core/Grid/Definition/Factory/OrderGridDefinitionFactory.php
```diff
@@ -195,6 +196,7 @@ protected function getColumns()
                     'field' => 'new',
                     'true_name' => $this->trans('Yes', [], 'Admin.Global'),
                     'false_name' => $this->trans('No', [], 'Admin.Global'),
+                    'clickable' => true,
```
jolelievre, line 199: I think it's useless here since `DataColumn` is already clickable by default

src/Core/Grid/Definition/Factory/OrderGridDefinitionFactory.php
```diff
@@ -211,6 +213,7 @@ protected function getColumns()
                 ->setOptions([
                     'field' => 'total_paid_tax_incl',
                     'is_paid_field' => 'paid',
+                    'clickable' => true,
```
jolelievre, line 216: Mmm, I don't get it a `LinkColumn` shouldn't be clickable.. unless this link matches the row link But still it seems weird

src/Core/Grid/Definition/Factory/OrderGridDefinitionFactory.php
```diff
@@ -236,6 +239,7 @@ protected function getColumns()
                 ->setOptions([
                     'field' => 'date_add',
                     'format' => $this->contextDateFormat,
+                    'clickable' => true,
```
jolelievre, line 242: Already clickable

jolelievre (dismissed):
No need to redefine default behavior when it's already good

src/Core/Grid/Definition/Factory/OrderGridDefinitionFactory.php
matks: BooleanColumn default value for `clickable` is false

src/Core/Grid/Definition/Factory/OrderGridDefinitionFactory.php
matks: Did you misread ? 😅 `LinkColumn` for customer name is not clickable, here it's `OrderPriceColumn` that is made clickable

src/Core/Grid/Definition/Factory/OrderGridDefinitionFactory.php
matks: Euuuuh no, `DateTimeColumn` default value for `clickable` is false

matks: See https://github.com/PrestaShop/PrestaShop/pull/16318/files#diff-4ecc9ae27c544deb5b13a5da1d8eec2fR56

matks:
@Robin-Fischer-PS QA feedbacks handled 😄 please re-check and rebuild assets

Robin-Fischer-PS:
Hi @matks ! There is a change requested, according to GitHub. Can I test it ?

matks (dismissed):

Robin-Fischer-PS:
Then it's good for QA :heavy_check_mark:

Progi1984 (approved):

PR #16334, opened 2019-11-08
Fix wrong flag on some languages

### LINKED ISSUE #16029: Wrong flag and ISO code for Argentina
**Describe the bug**
The Argentina localization pack creates a language with the ISO code "ag" which is for Antigua and Barbuda. Argentina's code is "ar".
The Antigua and Barbuda flag lang_mini_2_1 is also set instead of 23px-Flag_of_Argentina svg

**To Reproduce**
1. Install Argentina localization pack in International > Localization
2. Go to International > Localization > Language

**Additional information**
PrestaShop version: 1.7.6.1

eternoendless:
| Questions     | Answers
| Branch?       | develop
| Description?  | Some languages had the wrong flag. Read below to learn more.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | yes, protected method `Language::_copyNoneFlag()`
| Fixed ticket? | Fixes #16029
| How to test?  | Read below.

## Details

Up to now, language flag images were being chosen using the language's 2-letter ISO code. This worked fine for some combinations of language/countries where both the language ISO code and the country ISO code happen to be the same, like Spanish/Spain (es/es), French/France (fr/fr) and German/Germany (de/de). But this doesn't work well for some regionalized languages like Spanish/Argentina (es/ar), and is particularly bad for the reasons explained in [this comment](https://github.com/PrestaShop/PrestaShop/issues/16029#issuecomment-551079194).

Moreover, flags were being downloaded from prestashop.com every time.

This fix applies the following strategy:
- All flags are stored locally in `img/flags/` (I used the same libre [FamFamFam flag icons](http://www.famfamfam.com/lab/icons/flags/) as the ones served by prestashop.com)
- The flag is chosen using the following rule:
	1. If there's a `flag` property for the given language in `app/Resources/all_languages.json`, then the flag with that name is used
	2. Else, if a flag for the country name in the language locale code (the `AR` in `es-AR`) exists, then that flag is used
	3. Else, a generic flag with a question mark is used.

## There's no flag for my region!

If you're a Briton and you're mad at me for not adding the Brittany flag, this is solely because it wasn't included in the famfamfam pack. You can add it later by placing the image in the `admin-dev/themes/new-theme/static/flags/` directory, then edit `app/Resources/all_languages.json` to set up the `flag` parmeter to whatever your flag name is. Just keep in mind that the file must be a JPEG, and try to respect the same style as the other flags.

## Test cases

In all cases, you install the language in International > Translations > Add or update language, then visualize the list in International > Localization > Languages (yes that doesn't make much sense, eh @colinegin?)

1. Install Spanish (AR), you should see the Argentine flag (and not the Antigua and Barbuda one)
2. Install Arabic, you should see the Saudi Arabia flag
3. Install French (FR), you should see the French flag (and not the Belize one)
4. Install Catalan, you should see the Catalonian flag
5. Install Breton, you should see the French flag
6. Install French (CA), you should see the Canadian flag

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16334)


src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Handles language images (flag, "no image" placeholders)
+ */
+class LanguageImageManager
+{
+    /**
+     * Path where images are saved to
+     */
+    const IMG_PATH = _PS_ROOT_DIR_ . '/img/l/';
```
PierreRambaud:
```suggestion
    const IMG_PATH = _PS_IMG_DIR_ . '/l/';
```

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * Path where flags are stored
+     */
+    const FLAGS_SOURCE = _PS_BO_ALL_THEMES_DIR_ . 'new-theme/static/flags/%s.jpg';
+
+    const FALLBACK_FLAG_SOURCE = self::IMG_PATH . 'none.jpg';
```
PierreRambaud, line 55: If your class is not final

```suggestion
    const FALLBACK_FLAG_SOURCE = static::IMG_PATH . 'none.jpg';
```

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,165 @@
...
+    const IMAGE_DIRECTORIES = [_PS_CAT_IMG_DIR_, _PS_MANU_IMG_DIR_, _PS_PROD_IMG_DIR_, _PS_SUPP_IMG_DIR_];
+
+    /**
+     * Sets up the language flag image for the given language
+     *
+     * @param string $localeCode IETF language tag
+     * @param int $langId Language id
+     * @param string|null $flagCode If provided, use this flag code. By default, auto-detect using locale code.
+     *
+     * @throws \Exception
```
PierreRambaud: Don't see any throwing exception :thinking:

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,165 @@
...
+     */
+    public function setupLanguageFlag(string $localeCode, int $langId, ?string $flagCode = null): void
+    {
+        $flagCode = $flagCode ?? $this->getFlagCountryCodeFromLocale($localeCode);
+
+        $flagPath = $this->getFlagPath($flagCode);
+
+        if (!file_exists($flagPath)) {
+            $flagPath = self::FALLBACK_FLAG_SOURCE;
```
PierreRambaud: Same as previous about final class
```suggestion
            $flagPath = static::FALLBACK_FLAG_SOURCE;
```

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * Creates default copies for the "no image" image
+     *
+     * @param string $isoCode 2-lettre ISO code
+     *
+     * @throws \Exception
+     */
+    public function setupDefaultImagePlaceholder(string $isoCode): void
+    {
+        $filesToCopy = ['/en.jpg'];
```
PierreRambaud: Should be a const

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,165 @@
...
+
+        copy($flagPath, $destinationPath);
+    }
+
+    /**
+     * Creates default copies for the "no image" image
+     *
+     * @param string $isoCode 2-lettre ISO code
+     *
+     * @throws \Exception
```
PierreRambaud: same as previous about the throws

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    /**
+     * Deletes images associated with the language
+     *
+     * @param int $langId
+     * @param string $isoCode 2-letter ISO code
+     *
+     * @throws \Exception
```
PierreRambaud: same as previous about the throws

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,165 @@
...
+            $isoCode . '.jpg',
+            $isoCode . '-default-' . ImageType::getFormattedName('thickbox') . '.jpg',
+            $isoCode . '-default-' . ImageType::getFormattedName('home') . '.jpg',
+            $isoCode . '-default-' . ImageType::getFormattedName('large') . '.jpg',
+            $isoCode . '-default-' . ImageType::getFormattedName('medium') . '.jpg',
+            $isoCode . '-default-' . ImageType::getFormattedName('small') . '.jpg',
+        );
+        foreach (self::IMAGE_DIRECTORIES as $directory) {
+            foreach ($images as $image) {
+                if (file_exists($directory . $image)) {
```
PierreRambaud: You can create a method to don't repeat
```
if (file_exists($path)) {
 unlink($path);
}
```

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,165 @@
...
+            }
+        }
+    }
+
+    /**
+     * @param string $locale IETF language tag
+     *
+     * @return string
+     *
+     * @throws \Exception
```
PierreRambaud: same here

PierreRambaud (changes requested):
Some comments

src/Adapter/Language/LanguageImageManager.php
eternoendless: Oops, my bad, that's from an earlier implementation

src/Adapter/Language/LanguageImageManager.php
eternoendless: Done

src/Adapter/Language/LanguageImageManager.php
eternoendless: Done

src/Adapter/Language/LanguageImageManager.php
eternoendless: Done

src/Adapter/Language/LanguageImageManager.php
eternoendless: Done

src/Adapter/Language/LanguageImageManager.php
eternoendless: Done

src/Adapter/Language/LanguageImageManager.php
eternoendless: Done

src/Adapter/Language/LanguageImageManager.php
eternoendless: Done

eternoendless:
All changes done!

app/Resources/all_languages.json
```diff
@@ -87,7 +87,8 @@
     "date_format_full": "Y-m-d H:i:s",
     "is_rtl": "0",
     "language_code": "ca-es",
-    "locale": "ca-ES"
+    "locale": "ca-ES",
+    "flag": "catalonia"
```
PierreRambaud, line 91: Any reason to have `"flag"` only for `ca-es` and `en-ud`?

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,204 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Adapter\Language;
+
+use ImageType;
+
+/**
+ * Handles language images (flag, "no image" placeholders)
+ */
+class LanguageImageManager
```
PierreRambaud, line 35: Isn't redundant to have `Language\Language`?

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,204 @@
...
+                $formattedImageType = ImageType::getFormattedName($alias);
+                $from = $this->getPlaceholderImageFilename(static::DEFAULT_LANGUAGE_CODE, $formattedImageType);
+                $to = $this->getPlaceholderImageFilename($isoCode, $formattedImageType);
+                $filesToCopy[$from] = $to;
+            }
+        }
+
+        foreach (static::IMAGE_DIRECTORIES as $destinationDir) {
+            foreach ($filesToCopy as $sourceFile => $newFile) {
+                @copy(static::IMG_PATH . $sourceFile, $destinationDir . $newFile);
```
PierreRambaud, line 118: Why adding `@` here and not on line 92?

PierreRambaud (commented):
Just few questions and everything's ok for me

app/Resources/all_languages.json
eternoendless: Yes. By default, PrestaShop maps the country part of the locale to the country flag, which is in itself an approximation, but works fairly well for regional variations when they belong to a country, like Canadian French or Argentinian Spanish. However, in the case of ca-ES the country code is ES which would be mapped to the flag of Spain, but we'd want instead to use the Catalonian flag. Same goes for UD which is not a country in this case, but a variation.

src/Adapter/Language/LanguageImageManager.php
eternoendless: I think that once imported it makes more sense to see a `LanguageImageManager` rather than an `ImageManager`. In this case the `Language` part is not a namespace or contextual information but part of the class role: "it handles language images", not "images in the context of a language".

src/Adapter/Language/LanguageImageManager.php
eternoendless: No particular reason, I just kept it pretty much as it was.
I actually have the feeling that this piece of code (which are basically the thumbnails for the "no image" image) belongs elsewhere, but I didn't want to stray even farther from my original intention.

eternoendless:
Rebased with no changes

PierreRambaud (dismissed):

eternoendless:
Rebased again, no idea why Travis fails.

src/Adapter/Language/LanguageImageManager.php
```diff
@@ -0,0 +1,203 @@
...
+{
+    /**
+     * Path where images are saved to
+     */
+    const IMG_PATH = _PS_IMG_DIR_ . '/l/';
+
+    /**
+     * Path where flags are stored
+     */
+    const FLAGS_SOURCE = _PS_BO_ALL_THEMES_DIR_ . 'new-theme/static/flags/%s.jpg';
```
PierreRambaud: This const only exists in Admin context.

https://github.com/PrestaShop/PrestaShop/blob/develop/config/defines.inc.php#L70

eternoendless: I guess that needs to change!

config/defines.inc.php
```diff
@@ -67,9 +67,7 @@
 
 define('_PS_ALL_THEMES_DIR_', _PS_ROOT_DIR_.'/themes/');
 /* BO THEMES */
-if (defined('_PS_ADMIN_DIR_')) {
-    define('_PS_BO_ALL_THEMES_DIR_', _PS_ADMIN_DIR_.'/themes/');
-}
+define('_PS_BO_ALL_THEMES_DIR_', _PS_ADMIN_DIR_.'/themes/');
```
PierreRambaud: Now  ̀_PS_ADMIN_DIR_` is only available in `/admin` context :trollface:

eternoendless: Dammit!

PierreRambaud (approved):

matks:
Thank you @eternoendless

TecnoHands:
Hi, how I can solve this error in presta 1.7.6.8, Not only is the flag, is the code to The lang is Es (es-AR) in iso codes. But prestashop gives AG(es-AR) codes if I try change this code prestashop fuck off all the translation function.
If create a new lang and set with the correct iso code for Argentina and you try copy (Spanish) to the (new lang) fuck off all the translation function too. 

I dont see the changes in the change log for presta 1.7.7 beta or beta 2.
When will anybody correct this bug?. This exist from begin of presta 1.7.0 it's a shame certainly

PR #16335, opened 2019-11-08
Update id_order_state at the right place

### LINKED ISSUE #15834: Amount paid is no longer validated when creating an order
**Describe the bug**
When the amount paid sent in `validateOrder()` doesn't match with order total, order is created with state sent as a parameter (usually `Configuration::get('PS_OS_PAYMENT')`) instead of `Configuration::get('PS_OS_ERROR')` as in PS 1.7.5.2.

Since commit https://github.com/PrestaShop/PrestaShop/commit/57271b54b33503a31a103e95050f110f66c43dec#diff-0f162f4cf098397033f06b1cced01745, validation
```
// Amount paid by customer is not the right one -> Status = payment error	
// We don't use the following condition to avoid the float precision issues : http://www.php.net/manual/en/language.types.float.php	
// if ($order->total_paid != $order->total_paid_real)	
// We use number_format in order to compare two string	
if ($order_status->logable && number_format($cart_total_paid, _PS_PRICE_COMPUTE_PRECISION_) != number_format($amount_paid, _PS_PRICE_COMPUTE_PRECISION_)) {	
    $id_order_state = Configuration::get('PS_OS_ERROR');	
}
```

has been moved to `createOrderFromCart()`, so initial order state set in `validateOrder()` takes `$id_order_state` and is not changed to `Configuration::get('PS_OS_ERROR')`.

```
// Set the order status
$new_history = new OrderHistory();
$new_history->id_order = (int) $order->id;
$new_history->changeIdOrderState((int) $id_order_state, $order, true);
$new_history->addWithemail(true, $extra_vars);
```

**To Reproduce**
This can only be reproduced with an external payment method. Please contact me at info@idnovate.com or let me know your mail address so I can send you a module and credentials to test it.

**Additional information**
PrestaShop version: 1.7.6.1

atomiix:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | In case where `$cart_total_paid` is different from `$amount_paid`, the `$id_order_state` wasn't updated in the right context
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15834
| How to test?  | See #15834

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16335)

### DIFF
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -309,6 +309,14 @@ abstract class PaymentModuleCore extends Module
                 }
             }
 
+            // Amount paid by customer is not the right one -> Status = payment error
+            // We don't use the following condition to avoid the float precision issues : http://www.php.net/manual/en/language.types.float.php
+            // if ($order->total_paid != $order->total_paid_real)
+            // We use number_format in order to compare two string
+            if ($order_status->logable && number_format($cart_total_paid, _PS_PRICE_COMPUTE_PRECISION_) != number_format($amount_paid, _PS_PRICE_COMPUTE_PRECISION_)) {
+                $id_order_state = Configuration::get('PS_OS_ERROR');
+            }
+
             foreach ($package_list as $id_address => $packageByAddress) {
                 foreach ($packageByAddress as $id_package => $package) {
                     $orderData = $this->createOrderFromCart(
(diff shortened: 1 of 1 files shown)


PierreRambaud (approved):

PierreRambaud:
Thanks @atomiix

PR #16359, opened 2019-11-11
PassVsprintfValidator: Add separate regex for arguments only

### LINKED ISSUE #16421: Cannot add literal % to translation strings
**Describe the bug**
Any translation with a literal percent `%%` will fail because it is not an argument but is counted as one in `PrestaShopTranslatorTrait::$regexSprintfParams` and `PassVsprintfValidator::countArgumentsOfTranslation()`

**To Reproduce**
1. Go to any translatable string
2. Add a literal percent `%%` to it
3. Press save
4. Refresh page
5. String has reverted

**Additional information**
PrestaShop version: 1.7

Bugfix PR: #16359

jnvsor:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fixes a bug preventing the saving of translations containing a `%%` literal
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16421 
| How to test?  | Attempt a translation containing a literal `%%` - it will not be saved (Despite a green popup)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16359)

### DIFF
--- a/src/PrestaShopBundle/Command/CheckTranslationDuplicatesCommand.php
+++ b/src/PrestaShopBundle/Command/CheckTranslationDuplicatesCommand.php
@@ -114,7 +114,7 @@ class CheckTranslationDuplicatesCommand extends ContainerAwareCommand
         // Remove PrestaShop arguments %<arg>%
         $message = preg_replace(PrestaShopTranslatorTrait::$regexClassicParams, '~', $message);
         // Remove all related sprintf arguments
-        $message = preg_replace(PrestaShopTranslatorTrait::$regexSprintfParams, '~', $message);
+        $message = preg_replace(PrestaShopTranslatorTrait::$regexSprintfArgs, '~', $message);
 
         return $message;
     }
--- a/src/PrestaShopBundle/Translation/Constraints/PassVsprintfValidator.php
+++ b/src/PrestaShopBundle/Translation/Constraints/PassVsprintfValidator.php
@@ -54,7 +54,7 @@ class PassVsprintfValidator extends ConstraintValidator
     private function countArgumentsOfTranslation($property)
     {
         $matches = array();
-        if (preg_match_all(PrestaShopTranslatorTrait::$regexSprintfParams, $property, $matches) === false) {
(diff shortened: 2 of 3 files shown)


Progi1984:
@jnvsor Thank you for your contribution. Could you open an issue for describing the bug ? Thanks

jnvsor:
@Progi1984 Added a bug report

PierreRambaud (changes requested):
Hi and sorry for the delay,
why not updating the ̀regexSprintfParams` property?

jnvsor:
Because you still want `isSprintfString` to return true if the only argument is a `%%`, otherwise you might break existing strings

PierreRambaud:
@jnvsor This could be done with a single regex :thinking:

jnvsor:
> This could be done with a single regex

By using capture groups? I presume there was a performance reason not to, given that someone went to the effort of making the groups in this regex non-capturing in the first place.

The options as I see them are:

1. Have one regex that matches `%%` and ignore bug #16421
2. Have one regex that doesn't match `%%` and break existing translation strings with `%%` in them (Though given the existence of #16421 there may not be many of these)
3. Have two regexes
4. Use capture groups to distinguish between matches and argument counts (Would need more testing if we went that way)

PierreRambaud:
@jnvsor We should be able to use each characters we want :thinking: 
Having two regexes = two to maintain.
let's continue the discussion in the issue :)

matks:
@PierreRambaud Any update :) ?

PierreRambaud:
> @PierreRambaud Any update :) ?

Waiting for a feedback, this solution is not clean IMHO

matks:
Hello, it seems this PR has no activity since 2020.

I close this PR, following our [review guidelines](https://devdocs.prestashop.com/1.7/project/maintainers-guide/reviewing-pull-requests/) "Pull Requests may be closed after 30 days of inactivity".

If however you are still interested in it, please ping us 😉  we can reopen.

Issue #16377, opened 2019-11-12
Infinite crawl loop on module ps_currencyselector due to repetition of "SubmitCurrency=..&id_currency=.."


123monsite-regis:
**Describe the bug**
Links to other currencies on front office Home Page can lead to infinite crawl loop from search-engines as they generate new urls re-adding : "SubmitCurrency=..&id_currency=.." infinitly.

**To Reproduce**
1. Install a french shop , whith at less 2 currencies activated and ps_currencyselector activated
Your shop looks like the screen below :
![bug-currency-1](https://user-images.githubusercontent.com/3170104/68692687-30217400-0576-11ea-8994-3971a54d3807.png)

2. Go on home page Front office and click to change currency 
The urls of currencies are :
![bug-currency-2](https://user-images.githubusercontent.com/3170104/68692715-3adc0900-0576-11ea-9217-95175ea49863.png)

3. click again and you have again parameters repetead:
![bug-currency-3](https://user-images.githubusercontent.com/3170104/68693012-cbb2e480-0576-11ea-850d-10992e989783.png)

4. Try the pull request as a correction.

**Additional information**
PrestaShop version: 1.7.6.1

khouloudbelguith:
Hi @123monsite-regis,

We are aware of this issue, it is added to our debug roadmap.
Your ticket is a duplicate of #9481
So, I close it.
Thanks!

khouloudbelguith:
Duplicate of #9481

Issue #9481, opened 2018-08-22
[BOOM-6302] Changing currency URL handling


prestonBot:
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-6302).

- _**Reporter:**_ DennisPW
- _**Created at:**_ Tue, 21 Aug 2018 14:38:32 +0200

<p>Let's say I have two languages installed, English and Swedish.</p>

<p>Whenever I change the language using the ps_languageselector module the URL of the shop becomes either of the following:<br/>
myshop.com/en<br/>
myshop.com/sv</p>

<p>However, when changing the currency using the ps_currencyselector there seems to be an issue with the way it handles the urls. It appends "?SubmitCurrency=X&id_currency=X" instead of just reloading the page. The weird thing is, the first time you change the currency it appends it. The second time it appends it. The third time onwards it changes the <em>first</em></p>

<p>See attached video of the issue.</p>
        
* How to reproduce the issue ?

1. Add two or more languages
<br/>
2. Add two or more currencies
<br/>
3. Change language and/or currencies to see the issue happening in the URL

dennispw:
In the original ticket I attached a simple demo of how to reproduce: http://forge.prestashop.com/secure/attachment/70607/BOOM-6302.mp4

khouloudbelguith:
Hi @dennispw,

I manage to reproduce the issue with the last version 1.7.4.2.
![image](https://user-images.githubusercontent.com/16067358/44519090-48d11000-a6cc-11e8-94a3-9e3cf25ffd07.png)
We will see how to fix it.

Best regards, Khouloud

khouloudbelguith:
Hi,

Same issue with PS1.7.6.1.

Thanks!

123monsite-regis:
thanks @jolelievre for the code proposal on : https://github.com/PrestaShop/ps_currencyselector/pull/11

regards

PR #16378, opened 2019-11-12
repetition of "SubmitCurrency=..&id_currency=.."

123monsite-regis:
| Questions     | Answers
| Branch?       | develop
| Description?  | Infinite crawl loop on module ps_currencyselector due to repetition of "SubmitCurrency=..&id_currency=.."
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/16377 / #9481
| How to test?  | Please see related issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16378)

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -675,6 +675,8 @@ class ToolsCore
             $currency = Currency::getCurrencyInstance((int) $id_currency);
             if (is_object($currency) && $currency->id && !$currency->deleted && $currency->isAssociatedToShop()) {
                 $cookie->id_currency = (int) $currency->id;
+                unset($_GET['SubmitCurrency']);
+                unset($_GET['id_currency']);
             }
         }
 


jolelievre:
Hello @123monsite-regis 

Thank you for this proposal, I can see this bug can have drawbacks on SEO performance.
However it seems risky to make the modification here in a method that is supposed to set a cookie, not update the query It could cause problems for other part of the code or modules which would need this data

I think the problem should rather be fixed in the display, so in the module ps_currencyselector in the module class: https://github.com/PrestaShop/ps_currencyselector/blob/0985c9ab3aeeef4828ef9346672771015874b25f/ps_currencyselector.php#L84

Rather than concatenating the extra params, the request params should be parsed and then setting the submit parameters in the initial array, so it will either add them if not present or modify them if they were already in the query

123monsite-regis:
hello @jolelievre 

thanks, i will take a look at it, in fact, i tough the sooner the patch was done in the code ( in this case  ) the less impact it as (ps_currencyselector may have paid addons similar as it.)

the problem is more at https://github.com/PrestaShop/ps_currencyselector/blob/0985c9ab3aeeef4828ef9346672771015874b25f/ps_currencyselector.php#L77 than in https://github.com/PrestaShop/ps_currencyselector/blob/0985c9ab3aeeef4828ef9346672771015874b25f/ps_currencyselector.php#L87

jolelievre:
Are you sure it's simpler to modify `Link::getLanguageLink` I looked at it quickly it seems ok, and it's not that function which adds the parameter Since the extra parameters are added by the module he should be the one responsible of not adding them twice

Besides I understand what you mean about fixing the bug the sooner possible, but that's actually another reason to fix it in the module, especially native ones They can be released and distributed via addons any time, for small fixes like this it won't even break the backward compatibility so it will fix several versions of PrestaShop

Whereas fixing it in the core (and it's more a patch than a fix here) we will have to wait until the next release version (only one release every ~6 months)

123monsite-regis:
sorry, i does not see a solution , 
the module add parameters, but not at all case for the same currency we jsut jump into ,
for instance : `?SubmitCurrency=1&id_currency=1&SubmitCurrency=&id_currency=2` (duplicate parameters come from old url , not new generated by the module)

(but i agree, a module patch mean a quicker deployement)

regards

jolelievre:
@123monsite-regis I was thinking about this:

https://github.com/PrestaShop/ps_currencyselector/pull/11

123monsite-regis:
thanks @jolelievre 

I close this pull as https://github.com/PrestaShop/ps_currencyselector/pull/11 is better.

PR #16423, opened 2019-11-15
Use default attribute when no one are passed in front page

### LINKED ISSUE #16417: PS1.7.6.1 Quantity discounts not shown for default variant
The product page /{id_product}-xxx shows the default variant dropdowns but does not show the quantity discounts for the default variant

Only after selecting another variant or changing quantity the quantity discounts are shown, this is after the browser redirects to the variant page /{id_product}-{id_product_attribute}-xxx

[tested on a fresh PS1.7.6.1 instance]
**Steps to reproduce the issue**
1) install PS1.7.6.1
2) add quantity discount for product 3 "The best is yet to come' Framed poster" for combination "40x60"
3) Click on the Preview button => browse to http ://xxx.xxx/en/art/3-the-best-is-yet-to-come-framed-poster.html
table quantity discount prices for "40x60" is NOT shown
4) browse to http ://xxx.xxx/en/art/3-13-the-best-is-yet-to-come-framed-poster.html#/19-dimension-40x60cm
table quantity discount prices for "40x60" is shown

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Check for default attribute when no one are passed in the URL
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | /no
| Fixed ticket? | Fixes #16417 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16423)

### DIFF
--- a/controllers/front/ProductController.php
+++ b/controllers/front/ProductController.php
@@ -61,7 +61,7 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
                 unset($_GET['id_product_attribute']);
             } else {
                 //Only redirect to canonical (parent product without combination) when the requested combination is not valid
-                //In this case we are in a valid combination url and we must display it with redirection for SEO purpose
+                //In this case we are in a valid combination url and we must display it without redirection for SEO purpose
                 return;
             }
 
@@ -496,7 +496,7 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
 
         $id_currency = (int) $this->context->cookie->id_currency;
         $id_product = (int) $this->product->id;
-        $id_product_attribute = $this->getIdProductAttributeByRequestOrGroup();
+        $id_product_attribute = $this->getIdProductAttributeByGroupOrRequestOrDefault();
         $id_shop = $this->context->shop->id;
 
         $quantity_discounts = SpecificPrice::getQuantityDiscounts($id_product, $id_shop, $id_currency, $id_country, $id_group, $id_product_attribute, false, (int) $this->context->customer->id);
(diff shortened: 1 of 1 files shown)


matks (dismissed):

controllers/front/ProductController.php
```diff
@@ -999,6 +999,7 @@ private function getIdProductAttributeByRequestOrGroup()
      * if $checkedIdProductAttribute is available.
      * If not try to return the first available attribute, if none are available
```
matks: So this line is now false ?

controllers/front/ProductController.php
```diff
@@ -1026,6 +1027,10 @@ function ($elem) use ($checkedIdProductAttribute) {
             }
         }
 
+        if ($checkedIdProductAttribute === 0) {
```
jolelievre: I don't think this should be performed in `tryToGetAvailableIdProductAttribute` Considering the name of the method it does what it's supposed to do, which is checking that the combination is available in stock
Rather the problem here is that `$checkedIdProductAttribute` is equal to zero so this method can't fix anything

I'm looking at where it would make the most sense to get this default id

jolelievre: I'd say it should be in `getIdProductAttributeByRequestOrGroup` but then the name wouldn't make sense any more as well
I'd say we should add a new function `getIdProductAttributeByGroupRequestOrDefault` 😅 (the order reflects more the priority then the previous function) and replace all previous calls to `getIdProductAttributeByRequestOrGroup`

controllers/front/ProductController.php
PierreRambaud: Nope, it still try to get an attribute if possible :thinking:

controllers/front/ProductController.php
PierreRambaud: It still try to get an available attribute id. If there is no one, the getDefaultAttribute will also return 0 :)

PierreRambaud: Variable `$checkedIdProductAttribute` is equal to zero because of the request doesn't have any information about the attribute id. This is why in this context, when you try to get available product attribute id, it should test for the default one when it's possible.

jolelievre: Something like that

```php
    /**
     * Return id_product_attribute by id_product_attribute group parameter,
     * or request parameter, or the default attribute as a fallback.
     *
     * @return int|null
     *
     * @throws PrestaShopException
     */
    private function getIdProductAttributeByGroupOrRequestOrDefault()
    {
        $idProductAttribute = $this->getIdProductAttributeByGroup();
        if (null === $idProductAttribute) {
            $idProductAttribute = (int) Tools::getValue('id_product_attribute');
        }
        if (0 === $idProductAttribute) {
            $idProductAttribute = (int) Product::getDefaultAttribute($this->product->id);
        }

        return $this->tryToGetAvailableIdProductAttribute($idProductAttribute);
    }
```

jolelievre: Although the naming is lame 😅

jolelievre: Yes, I got the problem of the attribute missing in the request, I'm just saying the fallback shouldn't be managed in `tryToGetAvailableIdProductAttribute`

PierreRambaud: Oh, I now understand why I hate this part of code :sweat_smile: Thanks for explanation :+1:

jolelievre (approved):
Thanks @PierreRambaud

Progi1984:
Thanks @PierreRambaud

matks:
👏

PR #16426, opened 2019-11-15
Retrieve associated Specific Prices on the Product resource

rGaillard:
| Questions     | Answers
| Branch?       | develop
| Description?  | This improvement allow to retrieve the associated specific prices on the product resource using the webservices.
| Type?         | improvement 
| Category?     | WS
| BC breaks?    | no
| Deprecations? | no
| How to test?  | 1. Create a specific price for a product<br>2. Retrieve the product using the webservice

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16426)

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -417,6 +417,7 @@ class ProductCore extends ObjectModel
             'tax_rules_group' => array('type' => self::HAS_ONE),
             'categories' => array('type' => self::HAS_MANY, 'field' => 'id_category', 'object' => 'Category', 'association' => 'category_product'),
             'stock_availables' => array('type' => self::HAS_MANY, 'field' => 'id_stock_available', 'object' => 'StockAvailable', 'association' => 'stock_availables'),
+            'specific_prices' => array('type' => self::HAS_MANY, 'field' => 'id_specific_price', 'object' => 'SpecificPrice', 'association' => 'specific_prices'),
         ),
     );
 
@@ -536,6 +537,15 @@ class ProductCore extends ObjectModel
                     'id_product_attribute' => array(),
                     'quantity' => array(),
                 ),
+            ),
+            'specific_prices' => array(
+                'resource' => 'specific_price',
+                'getter' => 'getWsSpecificPricesIds',
+                'fields' => array(
+                    'id' => array('required' => true),
+                    'id_product_attribute' => array('required' => true),
+                ),
+                'setter' => false,
             ),
         ),
     );
@@ -6258,6 +6268,13 @@ class ProductCore extends ObjectModel
     public function getWsManufacturerName()
     {
         return Manufacturer::getNameById((int) $this->id_manufacturer);
+    }
+
+    public function getWsSpecificPricesIds()
+    {
+        return Db::getInstance()->executeS('SELECT `id_specific_price` id, `id_product_attribute`
+		FROM `' . _DB_PREFIX_ . 'specific_price`
+		WHERE `id_product`=' . (int) $this->id);
     }
 
     public static function resetEcoTax()


classes/Product.php
```diff
@@ -417,6 +417,7 @@ class ProductCore extends ObjectModel
             'tax_rules_group' => array('type' => self::HAS_ONE),
             'categories' => array('type' => self::HAS_MANY, 'field' => 'id_category', 'object' => 'Category', 'association' => 'category_product'),
             'stock_availables' => array('type' => self::HAS_MANY, 'field' => 'id_stock_available', 'object' => 'StockAvailable', 'association' => 'stock_availables'),
+            'specific_prices' =>    array('type' => self::HAS_MANY, 'field' => 'id_specific_price', 'object' => 'SpecificPrice', 'association' => 'specific_prices'),
```
PierreRambaud:
```suggestion
            'specific_prices' => array('type' => self::HAS_MANY, 'field' => 'id_specific_price', 'object' => 'SpecificPrice', 'association' => 'specific_prices'),
```

classes/Product.php
```diff
@@ -6260,6 +6269,13 @@ public function getWsManufacturerName()
         return Manufacturer::getNameById((int) $this->id_manufacturer);
     }
 
+	public function getWsSpecificPrices()
```
PierreRambaud: You mustn't use tabs :)
And could you please add phpdoc?

classes/Product.php
```diff
@@ -6260,6 +6269,13 @@ public function getWsManufacturerName()
...
+	{
+		return Db::getInstance()->executeS('SELECT `id_specific_price` id, `id_product_attribute`
```
PierreRambaud: The method name says it returns specifc prices, when the sql request return and `id` and an `id_product_attribute` :thinking:

PierreRambaud (changes requested):
You need to run php-cs-fixer :)

classes/Product.php
PierreRambaud: This is why a phpdoc block is required :smile:

classes/Product.php
rGaillard: My editor is configured with tabs and when i run php-cs-fixer on the file with the PrestaShop config it is changing lot of stuff in the file (like the short array syntax) :(

classes/Product.php
rGaillard: I can change the method name to getWsSpecificPricesIds() if you prefer... i named it like this to be consistent with the other webservice getter methods.

rGaillard:
> You need to run php-cs-fixer :)
It is changing a lot of stuff in the file but if it is OK, i will amend the PR.

classes/Product.php
PierreRambaud: Weird, are you sure you use the `.php_cs.dist` configuration?

```
[3.6.8] got@got:~/projects/prestashop/source/develop[develop]$ ./vendor/bin/php-cs-fixer fix classes/Product.php 
Loaded config default from "/home/got/projects/prestashop/source/develop/.php_cs.dist".
Using cache file "/home/got/projects/prestashop/source/develop/var/.php_cs.cache".
Paths from configuration file have been overridden by paths provided as command arguments.

Fixed all files in 0.001 seconds, 12.000 MB memory used
[3.6.8] got@got:~/projects/prestashop/source/develop[develop]$ git status
On branch develop
nothing to commit, working tree clean
```

classes/Product.php
PierreRambaud: If you can yes. The old way to defined method is not obvious, and we can't trust methods name :/

classes/Product.php
rGaillard: OK, not the same php_cs.dist, i was using the file at the prestashop root directory...

PierreRambaud: yes, this is what we are using too :thinking: 
https://github.com/PrestaShop/PrestaShop/blob/develop/.php_cs.dist

classes/Product.php
rGaillard: Updated

classes/Product.php
rGaillard: Done with the file inside the var instead of root directory.

classes/Product.php
```diff
@@ -6260,6 +6270,13 @@ public function getWsManufacturerName()
         return Manufacturer::getNameById((int) $this->id_manufacturer);
     }
 
+    public function getWsSpecificPricesIds()
+    {
+        return Db::getInstance()->executeS('SELECT `id_specific_price` id, `id_product_attribute`
+		FROM `' . _DB_PREFIX_ . 'specific_price`
+		WHERE `id_product`=' . (int) $this->id);
```
PierreRambaud, line 6277:
```php
return Db::getInstance()->executeS(
    'SELECT `id_specific_price` id, `id_product_attribute` ' .
    'FROM `' . _DB_PREFIX_ . 'specific_price` ' .
    'WHERE `id_product`=' . (int) $this->id
);
```

Progi1984:
@rGaillard Some news ?

PierreRambaud:
Hi @rGaillard ,

Thanks!

PR #16440, opened 2019-11-15
Added indexes for product reference and supplier_reference

### LINKED ISSUE #16323: Search by product reference causes full table scan
Searches by product reference are slow on large product table because it issues full table scan
```
explain SELECT `reference`
        FROM `product` p
        WHERE p.reference = "85450055675" LIMIT 1'
+----+-------------+-------+------------+------+---------------+------+---------+------+--------+----------+-------------+
| id | select_type | table | partitions | type | possible_keys | key  | key_len | ref  | rows   | filtered | Extra       |
+----+-------------+-------+------------+------+---------------+------+---------+------+--------+----------+-------------+
|  1 | SIMPLE      | p     | NULL       | ALL  | NULL          | NULL | NULL    | NULL | 365442 |    10.00 | Using where |
+----+-------------+-------+------------+------+---------------+------+---------+------+--------+----------+-------------+
```

product.reference column is VARCHAR(64) which is perfect for indexing, we should add an index on that column so imports and searches can benefit form it
same query but with index created
```
explain SELECT `reference`
    ->         FROM `product` p
    ->         WHERE p.reference = "85450055675" LIMIT 1;
+----+-------------+-------+------------+------+---------------+-----------+---------+-------+------+----------+-------------+
| id | select_type | table | partitions | type | possible_keys | key       | key_len | ref   | rows | filtered | Extra       |
+----+-------------+-------+------------+------+---------------+-----------+---------+-------+------+----------+-------------+
|  1 | SIMPLE      | p     | NULL       | ref  | reference     | reference | 195     | const |    1 |   100.00 | Using index |
+----+-------------+-------+------------+------+---------------+-----------+---------+-------+------+----------+-------------+
```
as you can see it only scanned one row instead of all rows
I would even consider adding an index for **supplier_reference**

if proposal sounds reasonable i can open a PR

jusurb:
| Questions     | Answers
| Branch?       | develop
| Description?  | Added indexes for product.reference and product.supplier_reference, so lookups in imports by reference no longer causes full table scans. Should also improve search by reference in BO
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16323 

**How to test?** 
- Don't apply create index SQL yet.
- Turn on slow query log / queries not using indexes log on mysql
- RUN import when product table contains at least 1000 entries before, 
- Notice there should be SELECT statement on product table where reference = "XXXX". 
- Run `explain` on that SQL query it should report using full table scan and scanning many rows. 
- Now add indexes and run the `explain` for the same query, it should now report using newly created index and scanning only 1 row if all references are unique

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16440)

### DIFF
--- a/install-dev/upgrade/sql/1.7.7.0.sql
+++ b/install-dev/upgrade/sql/1.7.7.0.sql
@@ -24,6 +24,11 @@ ALTER TABLE `PREFIX_hook` CHANGE `title` `title` VARCHAR(255) NOT NULL;
 ALTER TABLE `PREFIX_hook_alias` CHANGE `name` `name` VARCHAR(191) NOT NULL;
 ALTER TABLE `PREFIX_hook_alias` CHANGE `alias` `alias` VARCHAR(191) NOT NULL;
 
+/* improve performance of lookup by product reference/product_supplier avoiding full table scan */
+ALTER TABLE PREFIX_product
+    ADD INDEX reference_idx(reference),
+    ADD INDEX supplier_reference_idx(supplier_reference);
+
 /* Add fields for currencies */
 ALTER TABLE `PREFIX_currency` ADD `unofficial` TINYINT(1) UNSIGNED NOT NULL DEFAULT '0' AFTER `active`;
 ALTER TABLE `PREFIX_currency` ADD `modified` TINYINT(1) UNSIGNED NOT NULL DEFAULT '0' AFTER `unofficial`;


install-dev/upgrade/sql/1.7.7.0.sql
```diff
@@ -14,6 +14,14 @@ ALTER TABLE `PREFIX_product_attribute` ADD `mpn` VARCHAR(40) NULL AFTER `upc`;
 ALTER TABLE `PREFIX_product` ADD `mpn` VARCHAR(40) NULL AFTER `upc`;
 
 /* Delete price display precision configuration */
+
+DELETE `PREFIX_configuration` WHERE name = 'PS_PRICE_DISPLAY_PRECISION';
```
jusurb: this was added by a merge, but please note that **delete statement is missing FROM**

install-dev/upgrade/sql/1.7.7.0.sql
```diff
@@ -14,6 +14,14 @@ ALTER TABLE `PREFIX_product_attribute` ADD `mpn` VARCHAR(40) NULL AFTER `upc`;
...
+
+# improve performance of lookup by product reference/product_supplier avoiding full table scan
```
PierreRambaud: This is not an SQL comment

install-dev/upgrade/sql/1.7.7.0.sql
PierreRambaud: You need to use rebase command instead of `git pull` / `git merge`

install-dev/upgrade/sql/1.7.7.0.sql
jusurb: @PierreRambaud are you sure? it's been a mysql single line comment since v5.5 https://dev.mysql.com/doc/refman/5.5/en/comments.html
as is double dash --
maybe you're talking about ANSI SQL?

install-dev/upgrade/sql/1.7.7.0.sql
jusurb: fixed it

install-dev/upgrade/sql/1.7.7.0.sql
jusurb: either way changed it to be consistent with other comment styles, sorry for the trouble IDE by default uses #

PierreRambaud: unfortunately we are supposed to support 5.1 mysql version :(

jusurb: wow that's old, mysql official documentation doesn't even list anything past 5.5 :(

but hey a copy of mysql docs says # has been there since 5.0 http://download.nust.na/pub6/mysql/doc/refman/5.0/en/comments.html

PierreRambaud: Ohh, my bad, always use # in bash only :sweat_smile:

jusurb: well i made it consistent anyways so /**/ all the way 😄

jusurb:
build seems to be failing for random reasons like on files i didn't touch:
```
  ERROR  in src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/orders.yml

  >> Duplicate key "admin_search_product_for_order_creation" detected whilst parsing YAML. Silent handling of duplicate mapping keys in YAML is deprecated since Symfony 3.2 and will throw \Symfony\Component\Yaml\Exception\ParseException in 4.0 on line 187 at line 187.

                                                                                

 [WARNING] 232 YAML files have valid syntax and 1 contain errors.   
```            

or fails downloading translations 🤷‍♂

jf-viguier:
Very useful pr thanks @Gamesh

Progi1984:
@Gamesh I just relaunched the job in error. Could you rebase your PR for fixing conflict files, please ?

jusurb:
@Progi1984 done. i noticed we're adding a huge amount of alter tables 😮 
From practical perspective i would suggest to merge the same table's DDL into one, especially on larger databases people will feel the effect as the table will be modified only once and the tmp table will be created only once. Now if we leave as is some tables will get copied to tmp at least twice for each alter and that takes time.

Progi1984:
> @Progi1984 done. i noticed we're adding a huge amount of alter tables open_mouth
> From practical perspective i would suggest to merge the same table's DDL into one, especially on larger databases people will feel the effect as the table will be modified only once and the tmp table will be created only once. Now if we leave as is some tables will get copied to tmp at least twice for each alter and that takes time.

Good idea, but in this case, in an another pull request. Thanks

Progi1984 (dismissed):

jusurb:
sorry @Progi1984 had to rebase again because of conflicts, can we merge this quicker somehow i suspect that SQL file is modified frequently and we will keep getting conflicts

Progi1984 (dismissed):

Progi1984:
> sorry @Progi1984 had to rebase again because of conflicts, can we merge this quicker somehow i suspect that SQL file is modified frequently and we will keep getting conflicts

@Gamesh This PR needs to pass by QA before merge.

sarahdib:
hello @Gamesh 

Can you rebase the PR in order to test it ?

Thank you

jusurb:
@sarahdib done :/

Progi1984 (approved):

sam-pires:
Thanks @Gamesh 👍

PR #16465, opened 2019-11-19
Upgrade SQL scripts can use DB_NAME, release creator regex fix, Tab.enable upgrade script fixed

### LINKED ISSUE #16379: After upgrade to 177 the menu is empty
**Describe the bug**
After upgrade to 177 the menu is empty, this is linked to the new `Tab::enabled` field that is used to hide menu links when modules are disabled.

**To Reproduce**
Perform an upgrade to 177: 
- you need to use the autoupgrade module and build a release archive for 177
- or you can try by upgrading via the command `php install-dev/upgrade/upgrade.php`

**Analysis**
The problem is that the field is added in the database but not correctly initialized.
A dedicated upgrade script does this job (`ps_1770_preset_tab_enabled`) however the upgrade scripts are called BEFORE the doctrine update is called, and since Tab is a Doctrine entity when the PHP upgrade is called the table has not been updated yet.

**Solution**
Perform the table update in the `1.7.7.0.sql` file before executing the PHP script.

**Additional information**
PrestaShop version: N/A
PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | develop
| Description?  | The autoupgrade process failed because of previous modifications This PR fixes a bug related to the new charset utf8mb4 which required a function call to update the right database And somefix on the tabs because the menu was empty as every tab was init with `enabled=0` Finally the release creation process was removing the `_PS_ENV_` const because of a bad regexp
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16379
| How to test?  | This PR requires an update on autoupgrade PrestaShop/autoupgrade#341<br>Create a release thanks to the `php tools/build/CreateRelease.php --version="1.7.7.0"` command and then try to perform an upgrade with 1-click update module There should be no errors in the log, and the backoffice is accessible with correct menu

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16465)


tools/build/Library/ReleaseCreator.php
```diff
@@ -352,8 +352,8 @@ protected function setConfigDefinesConstants()
     {
         $configDefinesPath = $this->tempProjectPath . '/config/defines.inc.php';
         $configDefinesContent = file_get_contents($configDefinesPath);
-        $configDefinesNewContent = preg_replace('/(.*(define).*)_PS_MODE_DEV_(.*);/Ui', 'define(\'_PS_MODE_DEV_\', false);', $configDefinesContent);
-        $configDefinesNewContent = preg_replace('/(.*)_PS_DISPLAY_COMPATIBILITY_WARNING_(.*);/Ui', 'define(\'_PS_DISPLAY_COMPATIBILITY_WARNING_\', false);', $configDefinesNewContent);
+        $configDefinesNewContent = preg_replace('/(.*(define).*)["\']_PS_MODE_DEV_["\'](.*);/Ui', 'define(\'_PS_MODE_DEV_\', false);', $configDefinesContent);
```
jolelievre, line 355: Previous regexp was too wide (not checking the quotes) so it matched on expression like `define('_PS_ENV_', _PS_MODE_DEV_ ? 'dev': 'prod');` just because `_PS_MODE_DEV_` was on the line

install-dev/upgrade/php/ps_1770_preset_tab_enabled.php
```diff
@@ -28,9 +28,12 @@
  * Preset enabled new column in tabs to true for all (except for disabled modules)
  */
 function ps_1770_preset_tab_enabled() {
+    // Doctrine schema update is launched after this script so we need to, at least, add this field in database manually
+    $result = Db::getInstance()->execute('ALTER TABLE `' . _DB_PREFIX_ . 'tab` ADD enabled TINYINT(1) NOT NULL');
```
Progi1984: @jolelievre Don't you have any error if colum already exist ?

Progi1984 (commented):
Small question :question:

install-dev/upgrade/php/ps_1770_preset_tab_enabled.php
PierreRambaud: This kind of sql request must be done in 1.7.7.0.sql

install-dev/upgrade/php/ps_1770_update_charset.php
```diff
@@ -0,0 +1,32 @@
...
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+/**
+ * Update database to use utf8mb4 as a charset
+ */
+function ps_1770_update_charset() {
+    return Db::getInstance()->execute('ALTER DATABASE `' . _DB_NAME_ . '` CHARACTER SET = utf8mb4 COLLATE = utf8mb4_general_ci');
```
PierreRambaud: What is the reason of doing this? It was already in 1.7.7.0.sql?

install-dev/upgrade/php/ps_1770_preset_tab_enabled.php
jolelievre: Sure, but it doesn't exist at this moment Thats the problem because I can't preset it correctly

jolelievre: That's what we usually do, but since this case is special (partial update because doctrine update happens too late) I preferred to centralize all the related modification in the same php function
But if you insist I can move it in the SQL file

install-dev/upgrade/php/ps_1770_update_charset.php
jolelievre: Yes but it didn't work in the sql file because `_DB_NAME_` is not replaced automatically, nor can it be interpreted, so it tried to modify a database named `_DB_NAME_` which caused an error

install-dev/upgrade/php/ps_1770_preset_tab_enabled.php
PierreRambaud: AFAIK it was decided to centralized SQL :thinking: And use php only if needed

install-dev/upgrade/php/ps_1770_update_charset.php
PierreRambaud: Oh I see, this is something I forget to add. I only do that for modules: https://github.com/PrestaShop/PrestaShop/pull/14894/files#diff-8f670e6ff5e6b010dcbb019a08eb69adR456

Should be update here: https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Install/Upgrade.php#L423

install-dev/upgrade/php/ps_1770_preset_tab_enabled.php
jolelievre: Ok, no problem I'll move this one

install-dev/upgrade/php/ps_1770_update_charset.php
jolelievre: mmm, nice! I'll add this modification and revert to a full SQL then 😉

jolelievre:
@PierreRambaud Ok I modified as you suggested, all pure SQL queries are back in the SQL file, and now core is able to manage `DB_NAME` replacement inside them

However in the autoupgrade module there are other classes responsible for this SQL/PHP upgrade script execution here's the PR to manage this which will need to be merged for the QA team to test the whole process https://github.com/PrestaShop/autoupgrade/pull/341

install-dev/upgrade/php/ps_1770_preset_tab_enabled.php
```diff
@@ -29,8 +29,8 @@
  */
 function ps_1770_preset_tab_enabled() {
     //First set all tabs enabled
-    Db::getInstance()->execute(
-        'UPDATE `'._DB_PREFIX_.'tab` SET `enabled`= 1'
+    $result = Db::getInstance()->execute(
+        'UPDATE `'._DB_PREFIX_.'tab` SET `enabled` = 1'
```
PierreRambaud, line 33:
```suggestion
        'UPDATE `'._DB_PREFIX_.'tab` SET `enabled` = TRUE'
```
Could we start to use MySQL boolean instead integer. Will be really easier to play with another database engine (on day maybe)

PierreRambaud (approved):

jolelievre:
This PR requires an update on autoupgrade https://github.com/PrestaShop/autoupgrade/pull/341

PierreRambaud:
We can keep the `Waiting for QA` because it's the real status of this pull request :) 
I update the `test` part to be sure they notice an upgrade of the module is needed

Progi1984:
Thanks @jolelievre

PR #16150, opened 2019-10-28
Migrate order product management

sarjon:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Migrates product management in Order view page.
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #15826 #16512
| How to test?  | Not yet. => Rebuild assets 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16150)

PR #16467, opened 2019-11-19
Migrate partial refund

### LINKED ISSUE #15820: [ORDER PAGE] Partial Refund - 3/11

**User Story**

As a merchant, when I check an order, I need to be able refund my client.

**Detailed specification**
[Order Action Specs](https://github.com/PrestaShop/prestashop-specs/blob/cc476f1fc0c71c63bd3ba4879131f5163fca05e7/Order%20Page%20View.md#i-action-panel)

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | Migration of partial refund on order page
| Type?         | new feature
| Category?     | BO
| BC breaks?    |no
| Deprecations? | no
| Fixed ticket? | fixes #15820
| How to test?  | check that partial refund is working

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16467)


src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/view.html.twig
```diff
@@ -37,13 +37,13 @@
...
-    <div class="col-md-4">
+    <div class="col-md-4" style="order: 2;">
       {% include '@PrestaShop/Admin/Sell/Order/Order/Blocks/View/customer.html.twig' %}
 
       {% include '@PrestaShop/Admin/Sell/Order/Order/Blocks/View/messages.html.twig' %}
     </div>
 
-    <div class="col-md-8">
+    <div class="col-md-12" style="order: 1;">
       {% include '@PrestaShop/Admin/Sell/Order/Order/Blocks/View/products.html.twig' %}
```
matthieu-rolland, line 62: this is temporary, the action of making the products block bigger must have been done in another pr

jolelievre:
You should only compile required assets In this case I think order javascript is enough (maybe css of course) Besides try to have only ONE commit for building assets (drop the previous one with an interactive rebase) This makes things easier when other branches need to rebase themeselves

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -119,6 +130,32 @@
...
+                    <strong>{{ form_label(partialRefundForm['quantity_' ~ product.orderDetailId]) }}</strong>
+                    {{ form_widget(partialRefundForm['quantity_' ~ product.orderDetailId]) }}
+                    <div class="input-group-append"><div class="input-group-text">/ {{ product.quantityRefundable }}</div></div>
+                  </div>
+                </div>
+                <div class="col-lg-8">
+                  <div class="input-group">
+                    <strong>{{ form_label(partialRefundForm['amount_' ~ product.orderDetailId]) }}</strong>
+                    {{ form_widget(partialRefundForm['amount_' ~ product.orderDetailId]) }}
+                    <div class="input-group-append"><div class="input-group-text">€</div></div>
```
matthieu-rolland: just spotted this hardcoded currency symbol, I'll fix that (and the other one in the same file)

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
```diff
@@ -0,0 +1,86 @@
...
+        $products = $options['data']['products'];
+        $taxMethod = $options['data']['taxMethod'];
+        $translator = $options['data']['translator'];
+
+        foreach ($products as $product) {
+            $builder
+                ->add('quantity_' . $product->getOrderDetailId(), NumberType::class, [
+                    'attr' => ['max' => $product->getQuantity(), 'class' => 'refund-quantity'],
+                    'label' => $translator->trans('Qty', [], 'Admin.Orderscustomers.Feature'),
+                    'invalid_message' => $translator->trans('The "quantity" field must be a valid number', [], 'Admin.Orderscustomers.Feature'),
```
matthieu-rolland: hello @LouiseBonnard ,

I didn't find any similar error message, don't hesitate to tell me if you want me to change it (there are several error messages in this file, messages that are displayed when the user sends invalid data in the partial refund form in the order page.

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -563,13 +568,12 @@ private function getOrderDocuments(Order $order): OrderDocumentsForViewing
...
                 );
             }
-
             $documentsForViewing[] = new OrderDocumentForViewing(
                 $document->id,
                 $type,
                 new DateTimeImmutable($document->date_add),
                 $number,
-                $document->total_paid_tax_incl,
+                isset($document->total_paid_tax_incl) ? $document->total_paid_tax_incl : null,
```
PierreRambaud:
```suggestion
                $document->total_paid_tax_incl ?? null,
```

src/Core/Domain/Order/Command/IssuePartialRefundCommand.php
```diff
@@ -80,7 +79,7 @@ class IssuePartialRefundCommand
      * @param int $shippingCostRefund
      * @param bool $restockRefundedProducts
      * @param bool $generateCartRule
-     * @param int $taxMethod
+     * @param bool $taxMethod
```
PierreRambaud, line 82: Are you sure taxMethod is an int? :thinking:

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
```diff
@@ -69,4 +69,9 @@
       </div>
     </form>
   {% endif %}
+  {% if orderForViewing.hasInvoice() %}
+  <div class="form-inline d-inline-block form-horizontal ml-2">
```
PierreRambaud: indentation please :trollface:

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -197,11 +234,21 @@
...
               <p class="text-muted mb-0"><strong>{{ 'Shipping'|trans({}, 'Admin.Catalog.Feature') }}</strong></p>
               <strong>{{ orderForViewing.prices.shippingPriceFormatted }}</strong>
             </div>
 
+            <span class="partial-refund">
+              <p class="text-muted mb-0 text-center"><strong>{{ 'Shipping'|trans({}, 'Admin.Catalog.Feature') }}</strong></p>
+              <div class="col-sm text-center shipping-price-container" >
+                    <strong>{{ orderForViewing.prices.shippingPriceFormatted }}</strong>
+                    {{ form_widget(partialRefundForm.shipping) }}
+                    <div class="input-group-append"><div class="input-group-text">{{ orderCurrency.symbol }}</div></div>
```
PierreRambaud:
```suggestion
                    <div class="input-group-append">
                      <div class="input-group-text">{{ orderCurrency.symbol }}</div>
                    </div>
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -197,11 +234,21 @@
...
 
-            <div class="col-sm text-center">
+            <div class="col-sm text-center shipping-price">
               <p class="text-muted mb-0"><strong>{{ 'Shipping'|trans({}, 'Admin.Catalog.Feature') }}</strong></p>
               <strong>{{ orderForViewing.prices.shippingPriceFormatted }}</strong>
             </div>
 
+            <span class="partial-refund">
+              <p class="text-muted mb-0 text-center"><strong>{{ 'Shipping'|trans({}, 'Admin.Catalog.Feature') }}</strong></p>
+              <div class="col-sm text-center shipping-price-container" >
```
PierreRambaud:
```suggestion
              <div class="col-sm text-center shipping-price-container">
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -197,11 +234,21 @@
...
+                    <strong>{{ orderForViewing.prices.shippingPriceFormatted }}</strong>
```
PierreRambaud: Use two spaces for indentation please

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -119,6 +130,32 @@
...
+                  </div>
+                  <p class="text-center">{{ '(Max %amount_refundable% %tax_method%)'|trans({
+                      '%amount_refundable%': product.amountRefundable,
+                      '%tax_method%': orderForViewing.taxMethod
+                    }, 'Admin.Orderscustomers.Notification')|raw }}</p>
+                </div>
+              </div>
+            </td>
+            <td class="partial-refund">
+
```
PierreRambaud: empty block? what is the reason?

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
```diff
@@ -825,3 +825,13 @@
...
+
+{% block styled_checkbox_widget -%}
+  <div class="md-checkbox">
+    <label>
+      <input type="checkbox" name="{{ full_name }}"/>
+      <i class="md-checkbox-control"></i>
+      {{ label }}
+    </label>
+  </div>
+{%- endblock styled_checkbox_widget %}
```
PierreRambaud:
```suggestion
{%- endblock styled_checkbox_widget %}

```

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $products = $options['data']['products'];
+        $taxMethod = $options['data']['taxMethod'];
+        $translator = $options['data']['translator'];
+
+        foreach ($products as $product) {
+            $builder
+                ->add('quantity_' . $product->getOrderDetailId(), NumberType::class, [
+                    'attr' => ['max' => $product->getQuantity(), 'class' => 'refund-quantity'],
+                    'label' => $translator->trans('Qty', [], 'Admin.Orderscustomers.Feature'),
```
LouiseBonnard:
```suggestion
                    'label' => $translator->trans('Quantity', [], 'Admin.Global'),
```

LouiseBonnard: If we have enough space to do so, I'd prefer to have the word fully written.

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
LouiseBonnard: I have a generic error notification that could be used here: _This field is invalid, it must contain numeric values_ in _Admin.Notifications.Error_. Otherwise, I suggest _Quantity here must be a numeric value_.

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
```diff
@@ -0,0 +1,86 @@
...
+                    'required' => false,
+                ])
+                ->add('amount_' . $product->getOrderDetailId(), NumberType::class, [
+                    'attr' => ['max' => $product->getTotalPrice(), 'class' => 'refund-amount'],
+                    'label' => sprintf(
+                        '%s (%s)',
+                        $translator->trans('Amount', [], 'Admin.Global'),
+                        $taxMethod
+                    ),
+                    'invalid_message' => $translator->trans('The "amount" field must be a valid number', [], 'Admin.Orderscustomers.Feature'),
```
LouiseBonnard: Same here.

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -45,18 +46,25 @@
           <th>
             <p>{{ 'Qty'|trans({}, 'Admin.Orderscustomers.Feature') }}</p>
```
LouiseBonnard: Same as in previous comment: it could be nice to have it in full letters if we have enough space.

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -119,6 +130,32 @@
...
+                <div class="col-lg-8">
+                  <div class="input-group">
+                    <strong>{{ form_label(partialRefundForm['amount_' ~ product.orderDetailId]) }}</strong>
+                    {{ form_widget(partialRefundForm['amount_' ~ product.orderDetailId]) }}
+                    <div class="input-group-append"><div class="input-group-text">{{ orderCurrency.symbol }}</div></div>
+                  </div>
+                  <p class="text-center">{{ '(Max %amount_refundable% %tax_method%)'|trans({
+                      '%amount_refundable%': product.amountRefundable,
+                      '%tax_method%': orderForViewing.taxMethod
+                    }, 'Admin.Orderscustomers.Notification')|raw }}</p>
```
LouiseBonnard:
```suggestion
                    }, 'Admin.Orderscustomers.Help')|raw }}</p>
```

src/Core/Domain/Order/Command/IssuePartialRefundCommand.php
matthieu-rolland: it's indeed a bool, fixed !

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
matthieu-rolland: done !

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
matthieu-rolland: done !

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
matthieu-rolland: done !

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
matthieu-rolland: done !

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
matthieu-rolland: fixed ! (that was a bad way of dealing with the case when the cell is empty)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
matthieu-rolland: done !

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
matthieu-rolland: done !

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
matthieu-rolland: done !

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
matthieu-rolland: done !

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
matthieu-rolland: done !

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
matthieu-rolland: done !

matks (dismissed):

matks:
I checked: all code review feedbacks have been handled, thanks @matthieu-rolland

matks (dismissed):

jolelievre (dismissed):

sarahdib:
@matthieu-rolland 

- When I try to re-stock without fill the field quantity there is an error :
https://drive.google.com/file/d/1PdeUmPTWonnAW2zamvg-owrgKNI1k26I/view

- When I refund a product the price indicated is the wrong one :
https://drive.google.com/file/d/1PdeUmPTWonnAW2zamvg-owrgKNI1k26I/view

- In legacy the field quantity is already fill
Capture d’écran 2019-12-04 à 12 37 58
Capture d’écran 2019-12-04 à 12 38 44

sarahdib:
And the table display is not ok 

Capture d’écran 2019-12-04 à 13 42 17
Capture d’écran 2019-12-04 à 13 42 24

matthieu-rolland:
@sarahdib it's normal for the display, initially the table is supposed to be full width, do you want me to force it to be full width while waiting for the mechanism I need that is in another pr?

matthieu-rolland:
> @matthieu-rolland
> 
> * When I try to re-stock without fill the field quantity there is an error :
>   https://drive.google.com/file/d/1PdeUmPTWonnAW2zamvg-owrgKNI1k26I/view
> * When I refund a product the price indicated is the wrong one :
>   https://drive.google.com/file/d/1PdeUmPTWonnAW2zamvg-owrgKNI1k26I/view
> * In legacy the field quantity is already fill
> 
> Capture d’écran 2019-12-04 à 12 37 58
> 
> Capture d’écran 2019-12-04 à 12 38 44

Thanx @sarahdib for your feedback ! 

I finished fixing the first two issues.

About the third one, I think the legacy behavior is weird, don't hesitate to llet me know if you disagree @colinegin :

1/ From what I've seen in my tests: only customized products have their quantity field automatically filled, not the other products

2/ The given quantity is `1`, even if the customized product was bought in a higher quantity, why 1 ?It seems completely arbitrary.

In my opinion it's the shop owner's job to fill the wanted partial refund quantity field when needed.

colinegin:
@MatShir @marionf what do you think ?

MatShir:
Prefilling can be easier for the merchant. But compared to using the standard refund and cancel refund buttons, it can be a bad practice. Indeed, these buttons have the particularity of having a checkbox avoiding to refund a product unintentionally.

Seen with @matthieu-rolland, that's why I recommend that we set the default value to 0 to force the merchant to enter his quantity correctly. If the quantity is not entered during a refund then an error asking to enter the quantity is displayed. (Actually, it is the legacy behaviours)

admin-dev/themes/new-theme/js/pages/order/view.js
```diff
@@ -140,4 +140,9 @@ $(() => {
       $btn.prop('disabled', parseInt(selectedOrderStatusId, 10) === $btn.data('order-status-id'));
     });
   }
+
+  $('a.partial-refund-display, a.partial_refund_cancel').on('click', (event) => {
```
jolelievre: You should use orderMap values here

classes/order/OrderSlip.php
```diff
@@ -322,7 +321,7 @@ public static function create(Order $order, $product_list, $shipping_cost = fals
                 continue;
             }
 
-            if (!Tools::isSubmit('cancelProduct') && $order->hasBeenPaid()) {
+            if ($order->hasBeenPaid()) {
```
jolelievre: Are you sure about that? Although I find it ugly to have this check, maybe there's a reason for this in the "Cancel product" feature?

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -563,13 +569,12 @@ private function getOrderDocuments(Order $order): OrderDocumentsForViewing
...
                 );
             }
-
             $documentsForViewing[] = new OrderDocumentForViewing(
                 $document->id,
                 $type,
                 new DateTimeImmutable($document->date_add),
                 $number,
-                $document->total_paid_tax_incl,
+                isset($document->total_paid_tax_incl) ?? null,
```
jolelievre: Not sure you need both `isset` and `??` It should be one of those two:
- `isset($document->total_paid_tax_incl) ? $document->total_paid_tax_incl : null`
- `$document->total_paid_tax_incl ?? null` 

But the code here will only inject `true` at best

src/Core/Domain/Order/QueryResult/OrderProductForViewing.php
```diff
@@ -88,6 +88,21 @@ class OrderProductForViewing
      */
     private $orderDetailId;
 
+    /**
+     * @var string
+     */
+    private $amountRefund;
```
jolelievre, line 94: Are you sure this is a string? This seems like the float value

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -274,6 +277,13 @@ public function viewAction(int $orderId, Request $request): Response
     {
         /** @var OrderForViewing $orderForViewing */
         $orderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId));
+        $currencyDataProvider = $this->container->get('prestashop.adapter.data_provider.currency');
+        $orderCurrency = $currencyDataProvider->getCurrencyById($orderForViewing->getCurrencyId());
+
+        $orderDetailIds = [];
```
jolelievre: I don't see any use of `$orderDetailIds` so this whole code seems useless

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -304,6 +314,13 @@ public function viewAction(int $orderId, Request $request): Response
             'order_id' => $orderId,
         ]);
 
+        $translator = $this->get('translator');
+        $partialRefundForm = $this->createForm(PartialRefundType::class, [
```
jolelievre: You should create a service to avoid injecting the translator manually And maybe have a FormBuilder this would give hook available for modules

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -318,7 +335,61 @@ public function viewAction(int $orderId, Request $request): Response
             'addProductToOrderForm' => $addProductToOrderForm->createView(),
             'updateOrderProductForm' => $updateOrderProductForm->createView(),
             'updateOrderShippingForm' => $updateOrderShippingForm->createView(),
+            'partialRefundForm' => $partialRefundForm->createView(),
             'invoiceManagementIsEnabled' => $orderForViewing->isInvoiceManagementIsEnabled(),
+            'orderCurrency' => $orderCurrency,
+        ]);
+    }
+
+    public function partialRefundAction(int $orderId, Request $request)
```
jolelievre, line 381: Missing PHPDoc and security annotations

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -274,6 +277,13 @@ public function viewAction(int $orderId, Request $request): Response
     {
```
jolelievre, line 291: Missing PHPDoc and security annotation

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -318,7 +335,61 @@ public function viewAction(int $orderId, Request $request): Response
...
+    {
+        $orderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId));
+        $form = $this->createForm(PartialRefundType::class, [
+            'products' => $orderForViewing->getProducts()->getProducts(),
+            'taxMethod' => $orderForViewing->getTaxMethod(),
+            'translator' => $this->get('translator'),
+        ]);
+        $form->handleRequest($request);
+        if ($form->isSubmitted() && $form->isValid()) {
+            $data = $form->getData();
```
jolelievre: You should have a FormDataHandler to manage all this

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -318,7 +335,61 @@ public function viewAction(int $orderId, Request $request): Response
...
+                    $refunds[$orderDetailId]['quantity'] = $data['quantity_' . $orderDetailId];
+                }
+                if (!empty($data['amount_' . $orderDetailId])) {
+                    $refunds[$orderDetailId]['amount'] = $data['amount_' . $orderDetailId];
+                }
+            }
+
+            $status = 'success';
+            $message = $this->trans('A partial refund was successfully created.', 'Admin.Orderscustomers.Notification');
+            $command = new issuePartialRefundCommand(
```
```suggestion
            $command = new IssuePartialRefundCommand(
```

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
```diff
@@ -0,0 +1,87 @@
...
+
+namespace PrestaShopBundle\Form\Admin\Sell\Order;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\Extension\Core\Type\NumberType;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
+use PrestaShopBundle\Form\Type\StyledCheckboxType;
+
+class PartialRefundType extends AbstractType
```
jolelievre: Use `TranslatorAwareType` instead, you can then use `$this->trans` directly The translator shouldn't be injected as an option It's a dependency Not some contextual option

src/PrestaShopBundle/Form/Type/StyledCheckboxType.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+namespace PrestaShopBundle\Form\Type;
+
+use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
+
+/**
+ * Class StyledCheckboxType
+ */
+class StyledCheckboxType extends CheckboxType
```
jolelievre: I'm not sure you need to create a new FormType for this, you can try adding an attribute `material_design` Either in your `PartialRefundType ` like here https://github.com/PrestaShop/PrestaShop/blob/6d2f2405d6839357fe7dd5aa92144c5f63f4d515/src/PrestaShopBundle/Form/Admin/Improve/International/Currencies/CurrencyType.php#L104
Or inside the twig template https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/messages.html.twig

Although it seems weird that we need to explicitly add this, it should be somehow automatic

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
```diff
@@ -69,4 +69,9 @@
       </div>
     </form>
   {% endif %}
+  {% if orderForViewing.hasInvoice() %}
+    <div class="form-inline d-inline-block form-horizontal ml-2">
+      <a href="#" class="btn btn-action partial-refund-display">{{ 'Partial refund'|trans({}, 'Admin.Orderscustomers.Feature') }}</a>
```
jolelievre: Use a `<button type="button">` instead, then you can drop the useless `href="#"` and the `event.preventDefault();` in the javascript

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -89,9 +97,12 @@
                 {{ product.quantity }}
               {% endif %}
             </td>
+            <td>
+              <p>{{ product.quantityRefunded }}{% if product.amountRefund() %} ({{ product.amountRefund() }} {{ 'Refunded'|trans({}, 'Admin.Orderscustomers.Feature') }}){% endif %}</p>
```
jolelievre: Is the `()` necessary? It it's an `OrderProductForViewing` object then you can simply use `product.amountRefund`

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -119,6 +130,31 @@
                 </a>
               </td>
             {% endif %}
+
+            <td class="text-right">
+              <span class="partial-refund{% if product.isRefundable() == false %} hidden{% endif %}">
```
jolelievre: Same here I think you can drop the `()` Twig is able to call getters on object automatically as if they are public fields as long as they respect the convention `get**()/is**()`

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -230,7 +279,20 @@
...
+          <div class="form-group">
+            {{ form_widget(partialRefundForm.restock) }}
+          </div>
+          <div class="form-group">
+            {{ form_widget(partialRefundForm.voucher) }}
+          </div>
+        </div>
+      </div>
+      <div class="partial-refund form-submit text-right">
+        <a href="#" class="partial-refund partial_refund_cancel btn btn-outline-secondary">Cancel</a>
```
jolelievre: Use a `<button type="button">` here as well

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
```diff
@@ -0,0 +1,87 @@
...
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $products = $options['data']['products'];
+        $taxMethod = $options['data']['taxMethod'];
+        $translator = $options['data']['translator'];
+
+        foreach ($products as $product) {
```
jolelievre, line 42: Mmm, you should do this manually Symfony form have built in FormTypes to deal with collections
https://symfony.com/doc/current/reference/forms/types/collection.html

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -119,6 +130,31 @@
...
               </td>
             {% endif %}
+
+            <td class="text-right">
+              <span class="partial-refund{% if product.isRefundable() == false %} hidden{% endif %}">
+                <div class="form-group">
+                  <div class="col-lg-4">
+                    <div class="input-group">
+                      <strong>{{ form_label(partialRefundForm['quantity_' ~ product.orderDetailId]) }}</strong>
+                      {{ form_widget(partialRefundForm['quantity_' ~ product.orderDetailId]) }}
```
jolelievre, line 139: This is a weird syntax, I think if you use a CollectionType in the form type you can have cleaner way to access each product sub form

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/view.html.twig
```diff
@@ -39,13 +39,13 @@
   </div>
 
   <div class="row">
-    <div class="col-md-4">
+    <div class="col-md-4" style="order: 2">
```
jolelievre, line 51: Are you sure about this, you are breaking the two columns layout with this And inverting their order as well

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
```diff
@@ -825,3 +825,13 @@
     </datalist>
   {% endif %}
 {% endblock text_widget %}
+
+{% block styled_checkbox_widget -%}
```
jolelievre: This can be removed if you use the `material_design` attribute

jolelievre (dismissed):
A few things can be improved/cleaned Although it can take some times I think it's worth it

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/view.html.twig
matthieu-rolland: this was added temporarily, it allowed the QA team to review the feature, the mechanism for making the product block bigger is in another PR...

src/Core/Domain/Order/QueryResult/OrderProductForViewing.php
matthieu-rolland: yes, this variable contains a formatted price (with currency symbol)

classes/order/OrderSlip.php
matthieu-rolland: I removed it because this is a dependency to the legacy form (the cancel product form) that won't exist anymore. 

I just added a comment with a todo, since I'm the one who works on the cancel product feature, I'll deal with it in the corresponding pull request 👍

src/PrestaShopBundle/Form/Type/StyledCheckboxType.php
matks: It seems this class is not longer used but it's still in the PR

matks (dismissed):

Progi1984 (dismissed):

admin-dev/themes/new-theme/js/pages/order/view.js
```diff
@@ -160,4 +160,9 @@ $(() => {
       $modal.find(OrderViewPageMap.updateOrderAddressTypeInput).val($btn.data('address-type'));
     });
   }
+
+  $(`${OrderViewPageMap.displayPartialRefundBtn}, ${OrderViewPageMap.cancelPartialRefundBtn}`).on('click', (event) => {
+    event.preventDefault();
```
jolelievre: Since it's now a button no need for `event.preventDefault();` any more

classes/order/OrderSlip.php
jolelievre: Although I agree this should be managed better than this, but it a module uses the legacy page this could change its behaviour I'd say it would be safer to leave it unchanged in the legacy class But if you want to clean this code you extract it in a dedicated service that you can use in your handler Then you'll be free to make the improvements you want

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -446,7 +449,10 @@ private function getOrderProducts(Order $order): OrderProductsForViewing
                 $product['current_stock'],
                 $imagePath,
                 Tools::ps_round($product['unit_price_tax_excl'], 2),
-                Tools::ps_round($product['unit_price_tax_incl'], 2)
+                Tools::ps_round($product['unit_price_tax_incl'], 2),
```
jolelievre: Same feedback as for Franck's PR I'm not sure it's a good idea to round values like this, either not round at all Or if we round them we should use ComputePrecision and the appropriate currency precision I believe he has some modifications to make in the same handler in https://github.com/PrestaShop/PrestaShop/pull/16150

src/Core/Domain/Order/QueryResult/OrderProductForViewing.php
```diff
@@ -88,6 +88,21 @@ class OrderProductForViewing
...
+
+    /**
+     * @var int
+     */
+    private $quantityRefunded;
+
+    /**
+     * @var string
+     */
+    private $amountRefundable;
```
jolelievre, line 104: What's the difference between `amountRefund` and `amountRefundable` are they both formatted with currency symbol?

jolelievre: in the handler amountRefound doesn't seem formatted

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
```diff
@@ -0,0 +1,92 @@
...
+                    'invalid_message' => $this->trans('This field is invalid, it must contain numeric values', 'Admin.Notifications.Error', []),
+                    'required' => false,
+                    'data' => 0,
+                ])
+                ->add('amount_' . $product->getOrderDetailId(), NumberType::class, [
+                    'attr' => ['max' => $product->getTotalPrice(), 'class' => 'refund-amount'],
+                    'label' => sprintf(
+                        '%s (%s)',
+                        $this->trans('Amount', 'Admin.Global', []),
+                        $taxMethod
```
jolelievre, line 56: I believe `$taxMethod` is a boolean, then what will be displayed as a label? "Amount (true)" ? Shouldn't it be some text 'Tax excluded/tax included' ?

jolelievre (changes requested):

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
matthieu-rolland: It's a string containing the tax method (for example in french it will be TTC or HT)

src/Core/Domain/Order/QueryResult/OrderProductForViewing.php
matthieu-rolland: amountRefundable is a string*, containing the max value that can still be refunded for a product.

amountRefund is indeed a string, both are built via `$this->locale->formatPrice(`

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
matthieu-rolland: done !

admin-dev/themes/new-theme/js/pages/order/view.js
matthieu-rolland: done

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -425,14 +426,16 @@ private function getOrderProducts(Order $order): OrderProductsForViewing
                 $product['unit_price_tax_excl'] :
                 $product['unit_price_tax_incl']
             ;
-            $totalPrice = Tools::ps_round($unitPrice, 2) * ($product['product_quantity'] - $product['customizationQuantityTotal']);
+            $totalPrice = Tools::ps_round($unitPrice, Context::getContext()->getComputingPrecision()) * ($product['product_quantity'] - $product['customizationQuantityTotal']);
```
jolelievre: You should avoid using the context as much as possible, like ALWAYS if possible :P
And rounding should be done the later possible, you can check how it's done in the product PR: https://github.com/PrestaShop/PrestaShop/blob/03598a6060909e84f035393a16822c1b559b43b5/src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php#L427

src/Core/Form/IdentifiableObject/DataProvider/PartialRefundFormDataProvider.php
```diff
@@ -0,0 +1,73 @@
...
+    ) {
+        $this->queryBus = $queryBus;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getData($orderId)
+    {
+        /** @var EditableCategory $editableCategory */
```
jolelievre: This is not a `EditableCategory` it's an `OrderForViewing`

src/PrestaShopBundle/Form/Admin/Sell/Order/PartialRefundType.php
jolelievre: Oh right, it comes from `OrderForViewing` where it's a string My bad

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -425,14 +426,16 @@ private function getOrderProducts(Order $order): OrderProductsForViewing
...
 
             $unitPriceFormatted = $this->locale->formatPrice($unitPrice, $currency->iso_code);
             $totalPriceFormatted = $this->locale->formatPrice($totalPrice, $currency->iso_code);
 
             $imagePath = isset($product['image_tag']) ?
                 $this->imageTagSourceParser->parse($product['image_tag']) :
                 null;
+            $product['product_quantity_refunded'] = $product['product_quantity_refunded'] ?: false;
+            $product['amount_refund'] = $product['amount_refund'] ?: 0;
```
jolelievre: Since it's supposed to be a formatted price, should you use `$this->locale->formatPrice(0, $currency->iso_code);` as a fallback? Unless the field should be nullable?
Besides why reset this array field once again when it's already checked L363? These ternary checks should be used for the OrderProductForViewing  constructor, not modify the previously formatted array. Or you can correct the code L363 as well

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
matthieu-rolland: Done ! I didn't use the computingPrecision (since this is display), but the currency precision.

src/Core/Form/IdentifiableObject/DataProvider/PartialRefundFormDataProvider.php
matthieu-rolland: done !

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
matthieu-rolland: done !

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -426,16 +426,23 @@ private function getOrderProducts(Order $order): OrderProductsForViewing
...
             ;
-            $totalPrice = Tools::ps_round($unitPrice, 2) * ($product['product_quantity'] - $product['customizationQuantityTotal']);
 
-            $unitPriceFormatted = $this->locale->formatPrice($unitPrice, $currency->iso_code);
-            $totalPriceFormatted = $this->locale->formatPrice($totalPrice, $currency->iso_code);
+            $totalPrice = $unitPrice *
+                (!empty($product['customizedDatas']) ? $product['customizationQuantityTotal'] : $product['product_quantity']);
+
+            $unitPriceFormatted = $this->locale->formatPrice(
+                Tools::ps_round($unitPrice, $currency->precision),
```
jolelievre: Do not use `Tools::ps_round` when using formatPrice, it is redundant because format price already manages it internally

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -448,9 +455,9 @@ private function getOrderProducts(Order $order): OrderProductsForViewing
                 $totalPriceFormatted,
                 $product['current_stock'],
                 $imagePath,
-                Tools::ps_round($product['unit_price_tax_excl'], 2),
-                Tools::ps_round($product['unit_price_tax_incl'], 2),
-                $product['amount_refund'],
+                $product['unit_price_tax_excl'],
```
jolelievre: However here since the value is returned as raw float (not displayed price) you can use Tools::ps_round with appropriate computing precision Try to match what has been done in this PR https://github.com/PrestaShop/PrestaShop/pull/16150 as it will avoid conflicts

https://github.com/PrestaShop/PrestaShop/blob/03598a6060909e84f035393a16822c1b559b43b5/src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php#L448

src/Core/Form/IdentifiableObject/DataProvider/PartialRefundFormDataProvider.php
```diff
@@ -0,0 +1,72 @@
...
+    ) {
+        $this->queryBus = $queryBus;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getData($orderId)
+    {
+        $orderForViewing = $this->queryBus->handle(new GetOrderForViewing((int) $orderId));
```
jolelievre, line 57: Please add appropriate doc here
```suggestion
        /** @var OrderForViewing $orderForViewing */ 
        $orderForViewing = $this->queryBus->handle(new GetOrderForViewing((int) $orderId));
```

src/Core/Form/IdentifiableObject/DataProvider/PartialRefundFormDataProvider.php
```diff
@@ -54,6 +54,7 @@ public function __construct(
      */
     public function getData($orderId)
     {
+        /** @var OrderForViewing $orderForViewing */
```
jolelievre: It was a suggestion, but the `use` statement is missing I think

matthieu-rolland: Well I don't need the use statement for OrderForViewing, I need the one for `GetOrderForViewing`.

jolelievre (dismissed):
Thanks @matthieu-rolland

jolelievre (dismissed):

colinegin:
@sarahdib can you look at this PR in priority over other PRs in the QA column ?
Thanks

sarahdib:
@colinegin I'm already on it :)

sarahdib:
@matthieu-rolland 

When I try to re-stock without fill the field quantity there is an error :
https://drive.google.com/file/d/1XlhRgMCGv4iqq-tqVnAFXhk6-PqZwGGV/view

src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
```diff
@@ -310,4 +319,17 @@ public function handle(IssuePartialRefundCommand $command)
             );
         }
     }
+
+    /**
+     * @param float $taxRate
+     *
+     * @return TaxCalculator
+     */
+    private function getTaxCaculator(float $taxRate)
```
jolelievre:
```suggestion
    private function getTaxCalculator(float $taxRate)
```

matthieu-rolland:
> @matthieu-rolland
> 
> When I try to re-stock without fill the field quantity there is an error :
> https://drive.google.com/file/d/1XlhRgMCGv4iqq-tqVnAFXhk6-PqZwGGV/view

thanx for your feedback ! just fixed it

src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
```diff
@@ -74,14 +74,16 @@ public function handle(IssuePartialRefundCommand $command)
...
+        $taxCalculator = $this->getTaxCaculator($order->carrier_tax_rate);
 
         foreach ($refunds as $orderDetailId => $refund) {
-            $quantity = $refund['quantity'];
-
-            if (!$quantity) {
-                continue;
+            // this refund has an amount but no quantity, this should not happen
+            if (empty($refund['quantity'])) {
+                throw new OrderException('Please enter a quantity to proceed with your refund.');
```
jolelievre: Does the form type allows this? It seems the form type has required false which is why the value can be empty, you can add some constraints in the FormType to force values to be filled
The advantage is that you can then take advantage of Symfony form error display but then you need to modify your controller so that it renders the page with the form in case of validation error (for now it only use flash messages and redirects)

If you don't want to use the form validation but prefer to validate data in the handler through an exception then you need to use a specific exception so that you can manage its translation in the controller for your flash message The way it is now the error message will always be in english since you display the error message as is

You would have to use `$this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));` in the controller, and add the exception treatment in `getErrorMessages`

matthieu-rolland: So I noticed my mistake and fixed it while you wrote this message.

About the form type, the field is required => false because it can indeed be empty, the thing is: we can't have empty `quantity` AND a not empty `amount` field for a same product.

What I did so far in the handler is to throw an exception with an error message, and the error is caught in the controller, the message is displayed via the `addFlash` method.

src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
```diff
@@ -74,14 +84,19 @@ public function handle(IssuePartialRefundCommand $command)
...
 
         foreach ($refunds as $orderDetailId => $refund) {
-            $quantity = $refund['quantity'];
-
-            if (!$quantity) {
-                continue;
+            // this refund has an amount but no quantity, this should not happen
+            if (empty($refund['quantity'])) {
+                throw new OrderException(
+                    $translator->trans($this->quantityErrorMessage, array(), 'Admin.Orderscustomers.Notification')
```
jolelievre: You mustn't use `$this->trans` with variable or the TranslationTool is unable to scan them and add them in the catalog Besides you shouldn't translate the exception message, the exception message is always in english and not translated 
You need to create two dedicated exceptions for these two use cases, for example `EmptyRefundQuantityException` and `EmptyRefundAmountException` then you will be able to translate these exceptions **in the controller** using `$this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));` and by adding a use case in `getErrorMessages` for each of your exceptions

jolelievre: Besides getting the translator from the context is really ugly 😅

jolelievre (changes requested):

jolelievre (approved):

matks:
Huge congrats for @matthieu-rolland who picked this feature and jumped into the migration train with courage 😄

PR #16383, opened 2019-11-12
Currencies are customizable and prices are shown accordingly

jolelievre:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Currencies can be customized and have a specific localized pattern to modify their symbol position, this modification is then usable when displaying a price for a specific currency and language To simplify manipulation the command/handlers use a transformation type which is then applied and saved as a currency pattern
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14709
| How to test?  | The feature is not testable in the back-office yet, you can either rely on the Behat tests or update the database content manually To do that in the `currency_lang` table set the `pattern` column to one of these values: `¤ #,##0.00` `¤#,##0.00` `#,##0.00 ¤` `#,##0.00¤` (these are simple formats, the last PR will allow to test more advanced formats more easily like `¤ #,##,##0.00` or `¤#,##0.00;¤-#,##0.00`)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16383)

image

PR #16495, opened 2019-11-21
Remove key/value defined twice

mvorisek:
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16495)

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -26,7 +26,6 @@ return PhpCsFixer\Config::create()
         ],
         'function_to_constant' => false,
         'no_alias_functions' => false,
-        'non_printable_character' => false,
         'phpdoc_summary' => false,
         'phpdoc_align' => [
             'align' => 'left',


jolelievre:
Hi @mvorisek,

thank you for your contribution! Actually this doublon is gonna be clean soon in this coming PR https://github.com/PrestaShop/PrestaShop/pull/16383/files

We actually need to keep non printable character because we need them to build currency patterns used to display prices. So if it's ok I think we can close your PR and just wait for the other to be merged.

mvorisek:
I primary commited this as there was duplicate key in the config, but it seems not the best option to me to enable non-printable characters only for very few test cases. Commented also on the PR https://github.com/PrestaShop/PrestaShop/pull/16383.

mvorisek:
Closing as it will be resolved in PR https://github.com/PrestaShop/PrestaShop/pull/16383 .

PR #16506, opened 2019-11-22
Set minimum PHP version to 7.1.3 (and not 7.1)

eternoendless:
| Questions     | Answers
| Branch?       | develop
| Description?  | Set minimum PHP version to 7.1.3 (and not 7.1) + Extract version from wording
| Type?         | bug fix
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | N/A
| How to test?  | Run the install

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16506)


classes/ConfigurationTest.php
```diff
@@ -158,7 +158,7 @@ public static function test_apache_mod_rewrite()
 
     public static function test_new_phpversion()
     {
-        return version_compare(PHP_VERSION, '7.1.3', '>=');
+        return static::test_phpversion();
```
PierreRambaud, line 161: We can remove this function, edit the `getDefaultTestsOp` and use `'phpversion'` instead of `'new_phpversion`'?

eternoendless: Technically, it'd be a breaking change. I'll edit the consumer and deprecate the function though.

eternoendless: Technically, it'd be a breaking change. I'll edit the consumer and deprecate the function though.

eternoendless: Found the original intent for this test. It was for future-proofing: https://github.com/PrestaShop/PrestaShop/commit/b7753c1b547e3a455a6c691b8d90da1fc8639a07#diff-04f3684340e9328d0376ebec86a87523

Maybe it would be interesting to use it to push for PHP 7.2?

PierreRambaud: But methods name must change.
It's not about the new php version, but the recommended one :thinking:

eternoendless: How about we do that in a follow-up PR?

PierreRambaud (approved):

khouloudbelguith:
Hi @eternoendless,

It is ok.
1. Checked with PHP v5.6.40
image
2. Checked with PHP v7.1.33 => well installed
3. Checked with PHP v7.2.22 => well installed

Thanks!

PierreRambaud:
Thanks @eternoendless

PR #16520, opened 2019-11-25
Fix contact layout not working at all

### LINKED ISSUE #10140: contact page - full width - left column is still present without display: none;
Hi

**Describe the bug**
if you set "full width" for the contact page, the left column is still present without display: none;
even if the block is hidden

**Steps to reproduce the issue**
Click on Design > Theme & logo
Click on "Choose layout"
For contact us page, select "Full Width - No side columns, ideal for distraction-free pages such as product pages."

**Screenshots**

**Additionnal information**
PrestaShop version: 1.7.4.2
PHP version: 7.0

bye

NeOMakinG:
| Questions     | Answers
| Branch?       | develop
| Description?  | contact layout other than with left columns were not working at all, It may break theme users if these templates are in their child theme, I need an advice about this.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #10140 
| How to test?  | Just change the contact page layout in theme configuration and watch if every layouts are working

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16520)


themes/classic/templates/contact.tpl
```diff
@@ -26,12 +26,10 @@
...
-  <div id="left-column" class="col-xs-12 col-sm-4 col-md-3">
+{block name='column_contact'}
     {widget name="ps_contactinfo" hook='displayLeftColumn'}
-  </div>
 {/block}
 
 {block name='page_content'}
   {widget name="contactform"}
-{/block}
+{/block}
```
matthieu-rolland: Missing new line at end of file

themes/classic/templates/layouts/layout-right-column.tpl
```diff
@@ -35,3 +35,9 @@
     {hook h="displayContentWrapperBottom"}
   </div>
 {/block}
+
+{if $page.page_name == 'contact'}
+  {block name="right_column_contact"}
+    {widget name="ps_contactinfo" hook='displayLeftColumn'}
+  {/block}
+{/if}
```
matthieu-rolland: missing new line at end of file

matthieu-rolland (dismissed):
Two minor changes requested, until you get some feedback for the problem you mentioned in the PR description...

NeOMakinG:
For the reviewer : I don't know if the best solution is to do this, I think we could have a var from the backend which told the template in which type of layout he is so he can choose if he wants to populate a hook in case of a specific layout

themes/classic/templates/contact.tpl
```diff
@@ -26,10 +26,8 @@
 
 {block name='page_header_container'}{/block}
 
-{block name='left_column'}
-  <div id="left-column" class="col-xs-12 col-sm-4 col-md-3">
+{block name='column_contact'}
```
jolelievre: I don't know the layout architecture very well, but this seems like a BC, if someone has overridden this layout in a them then the block name is not the same any more

themes/classic/templates/layouts/layout-both-columns.tpl
```diff
@@ -61,9 +61,11 @@
 
           {block name="left_column"}
             <div id="left-column" class="col-xs-12 col-sm-4 col-md-3">
+              {block name="column_contact"}{/block}
+
               {if $page.page_name == 'product'}
                 {hook h='displayLeftColumnProduct'}
-              {else}
+              {else if $page.page_name != 'contact'}
```
jolelievre: And it feels weird to have special use case just for the contact page, since this is a generic global layout
I don't doubt something needs to be fixed but it should be managed in the contact page only not in the général layouts Or we could end up with multiple `elseif` for each use case or pages..

themes/classic/templates/layouts/layout-right-column.tpl
```diff
@@ -35,3 +35,9 @@
     {hook h="displayContentWrapperBottom"}
   </div>
 {/block}
+
+{if $page.page_name == 'contact'}
+  {block name="right_column_contact"}
+    {widget name="ps_contactinfo" hook='displayLeftColumn'}
```
jolelievre: And finally if this is the right layout then why is the `displayLeftColumn` hook called?

themes/classic/templates/contact.tpl
PierreRambaud: If we take in consideration it's a bc if we change template. We can't  fix css / js / tpl files :thinking: 
Since these files are not related to the `public API`, it's not a BC. ping @eternoendless

themes/classic/templates/layouts/layout-both-columns.tpl
PierreRambaud: Look line 66 :roll_eyes:

themes/classic/templates/contact.tpl
jolelievre: It's not related to the API but it can break any child theme of classic no? Maybe we don't have a choice, it already happened I'm just wondering if there is an alternative solution and if not we will have to warn about this BC

themes/classic/templates/layouts/layout-both-columns.tpl
jolelievre: True... but at least the hook is related to product But in this case I dont really understand the logic of adding a new block and ignoring the generic hook When the initial hook worked no?

themes/classic/templates/contact.tpl
jolelievre: Is it possible to detect which layout is used? And adapt this template rather than modifying the generic ones?

jolelievre: Maybe using the `$layout` variable that is used in `page.tpl`?

PierreRambaud: Everytime we add an html element or a css property we potentially break the child theme :trollface: 
I don't know if there is a way to detect a layout :/

themes/classic/templates/layouts/layout-both-columns.tpl
NeOMakinG: I agree with you, we need to handle these block populate in the contact tpl, but we need a way to detect which layout is used in order to populate the right block.. If there is a way I would love to know it

themes/classic/templates/layouts/layout-right-column.tpl
NeOMakinG: It's not even a real hook, it's just used to select the "rich" template

/modules/ps_contactinfo/ps_contactinfo.php

NeOMakinG:
$layout returns the layout selected, I refactored the PR to use it, it's way cleaner and less breaker I think :)

themes/classic/templates/contact.tpl
```diff
@@ -26,11 +26,19 @@
 
 {block name='page_header_container'}{/block}
 
-{block name='left_column'}
-  <div id="left-column" class="col-xs-12 col-sm-4 col-md-3">
-    {widget name="ps_contactinfo" hook='displayLeftColumn'}
-  </div>
-{/block}
+{if $layout ===  'layouts/layout-left-column.tpl'}
```
PierreRambaud:
```suggestion
{if $layout === 'layouts/layout-left-column.tpl'}
```

themes/classic/templates/contact.tpl
```diff
@@ -26,11 +26,19 @@
...
+  {block name="left_column"}
+    <div id="left-column" class="col-xs-12 col-sm-4 col-md-3">
+      {widget name="ps_contactinfo" hook='displayLeftColumn'}
+    </div>
+  {/block}
+{else if $layout ===  'layouts/layout-right-column.tpl'}
```
PierreRambaud:
```suggestion
{else if $layout === 'layouts/layout-right-column.tpl'}
```

themes/classic/templates/layouts/layout-right-column.tpl
```diff
@@ -34,4 +34,4 @@
     {/block}
     {hook h="displayContentWrapperBottom"}
   </div>
-{/block}
+{/block}
```
PierreRambaud:
```suggestion
{/block}

```

themes/classic/templates/contact.tpl
```diff
@@ -26,11 +26,19 @@
...
-    {widget name="ps_contactinfo" hook='displayLeftColumn'}
-  </div>
-{/block}
+{if $layout === 'layouts/layout-left-column.tpl'}
+  {block name="left_column"}
+    <div id="left-column" class="col-xs-12 col-sm-4 col-md-3">
+      {widget name="ps_contactinfo" hook='displayLeftColumn'}
+    </div>
+  {/block}
+{else if $layout === 'layouts/layout-right-column.tpl'}
```
PierreRambaud: What if it's a custom layout. No else statement needed?

PierreRambaud (approved):

Progi1984:
Thanks @NeOMakinG

PR #16527, opened 2019-11-25
Check memory_limit during installation

### LINKED ISSUE #16507: Add a check for minimum PHP memory on install
The [System requirements documentation](https://devdocs.prestashop.com/1.7/basics/installation/system-requirements/) states that the minimum recommended memory for PHP should be set to 256Mb. To my understanding, this is not enforced by the installer, putting the install process at risk of crashing.

A check on this requirement should be added to the install process.

PierreRambaud:
| Questions     | Answers
| Branch?       | develop
| Description?  | PHP's config memory_limit must be checked during installation process
| Type?         | improvement
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16507
| How to test?  | Decrease your memory_limit to 128M to see this message<br>image

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16527)

### DIFF
--- a/install-dev/controllers/http/process.php
+++ b/install-dev/controllers/http/process.php
@@ -297,60 +297,22 @@ class InstallControllerHttpProcess extends InstallControllerHttp implements Http
     public function display()
     {
         $memoryLimit = Tools::getMemoryLimit();
-        // The installer SHOULD take less than 32M, but may take up to 35/36M sometimes. So 42M is a good value :)
-        $lowMemory = ($memoryLimit != '-1' && $memoryLimit < Tools::getOctets('42M'));
-
         // We fill the process step used for Ajax queries
         $this->process_steps[] = array('key' => 'generateSettingsFile', 'lang' => $this->translator->trans('Create file parameters', array(), 'Install'));
         $this->process_steps[] = array('key' => 'installDatabase', 'lang' => $this->translator->trans('Create database tables', array(), 'Install'));
         $this->process_steps[] = array('key' => 'installDefaultData', 'lang' => $this->translator->trans('Create default shop and languages', array(), 'Install'));
 
-        // If low memory or big fixtures, create subtasks for populateDatabase step (entity per entity)
-        $populate_step = array('key' => 'populateDatabase', 'lang' => $this->translator->trans('Populate database tables', array(), 'Install'));
-        if ($lowMemory) {
-            $populate_step['subtasks'] = array();
-            $xml_loader = new XmlLoader();
-            $xml_loader->setTranslator($this->translator);
-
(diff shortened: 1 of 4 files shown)


install-dev/controllers/http/process.php
```diff
@@ -297,60 +297,22 @@ public function processInstallTheme()
...
-            $populate_step['subtasks'] = array();
-            $xml_loader = new XmlLoader();
-            $xml_loader->setTranslator($this->translator);
-
-            foreach ($xml_loader->getSortedEntities() as $entity) {
-                $populate_step['subtasks'][] = array('entity' => $entity);
-            }
-        }
-
-        $this->process_steps[] = $populate_step;
```
matthieu-rolland, line 320: Are all those deletions directly related to the check of php memory limit?

matthieu-rolland: answered 👍

matthieu-rolland (dismissed):
lgtm !

matthieu-rolland (approved):
lgtm, thanx @PierreRambaud !

Progi1984 (approved):

Robin-Fischer-PS:
Hi @PierreRambaud !

There is a way to skip this step : When there is an install cache on the browser, a new install starts at the "Store information" step, so the memory is not verified in this case. I don't know if there is an easy way to prevent users from skipping step "System compatibility" ?

Thanks

PierreRambaud:
@Robin-Fischer-PS If all tests are green, this step seems to be skipped, but in fact, no. If you change your memory limit and restart your apache server or your php-fpm, you can restart the installation and you'll see the warning.

Robin-Fischer-PS:
@PierreRambaud See on this screenrecord, the even if the memory limit is not high enough, I can skip the step (I just go back on step 3 to show the memory is not OK) : 

https://drive.google.com/file/d/1LXxvvqrZBsqpyNJAVR4S1J_bbTtICe7B/view?usp=sharing

PierreRambaud:
> @PierreRambaud See on this screenrecord, the even if the memory limit is not high enough, I can skip the step (I just go back on step 3 to show the memory is not OK) :
> 
> https://drive.google.com/file/d/1LXxvvqrZBsqpyNJAVR4S1J_bbTtICe7B/view?usp=sharing

This is a normal behaviour because of sessions :sweat_smile:

Robin-Fischer-PS:
Thanks, so it's QA approved :P

PierreRambaud:
Thanks me!

Robin-Fischer-PS:
Thanks @PierreRambaud :D

PR #16531, opened 2019-11-25
Round remaining value of reduction amount

### LINKED ISSUE #16491: Reduction error on invoice with amount reduction.
**Describe the bug**
Reduction error on invoice with amount reduction.

**To Reproduce**

1. Prepare a catalog with all following product VAT 20%

Product One: 46.90€
Product Two: 12,50€
Product Three: 12,50€
Product Four: 12,50€

2. Prepare a discount with:
- Information (all in Yes) > partial use: Yes
- Condition : no specific conditions
- Action
    - Free shipping: No
    - Apply a discount: amount
    - Amount 4.6€
    - Apply a discount to: Order (without shipping)
    - Send a free gift: No

3. Add in your cart:

- 1 x Product One: 46.90€
- 2 x Product Two: 12,50€
- 2 x Product Three: 12,50€
- 1 x Product Four: 12,50€

and add you discount to your cart.

Total : 104.60€

In your backoffice, you'll see
- Discount value = - €109.40 (Delete Voucher)
- Products:     €109.40
- Discounts     -€4.60
- Shipping  €0.00
- Total     €104.80

The bug is that in the "order_cart_rule" table, the value of the discount is the total amount of product.
You can note that a new discount is create with 0€ of discount.

We look for the origin of the bug and found there a bug with arround after the calculate classe of discounts on the PaymentModule classe.

We will propose to you a pull request.

Please note, if you add in your cart :
1x12.5€ + 1x12.5€ + 3x12.5€ + 1x27.9€ + 2x10€ + 1x1.0€ + 1x46,9€ with 10€ of discount you'll also have the bug

capture d'écran

**Additional information**
PrestaShop version: 1.7.6.x
PHP version: all

202-ecommerce:
| Questions     | Answers
| Branch?       | develop
| Description?  | Please see issue #16491 to see and reproduce the bug. Please note this proposal fix the issue but we also see the limit value of the method : here we generate a remaining reduction with a value 0€ (first effect of the bug) but it will not have an impact on the current value of reduction. So for us the code after line 1008 is quite dangerous. In fact, this bug highlight in this case. If you log $values on line 1007, you'll see the reduction amount is OK. $remaining value = 0.3xxxE-15 (that's why we propose to add round)/. After that log $values at line 1180, you'll see that $values had been corrupted in the if condition. That's the second perverse effect of this bug.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16491
| How to test?  | Please see issue #16491

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16531)

### DIFF
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -1104,7 +1104,7 @@ abstract class PaymentModuleCore extends Module
             // THEN
             //  The voucher is cloned with a new value corresponding to the remainder
             $remainingValue = $cartRule->reduction_amount - $values[$cartRule->reduction_tax ? 'tax_incl' : 'tax_excl'];
-            $remainingValue = Tools::ps_round($remainingValue, _PS_PRICE_COMPUTE_PRECISION_);
+            $remainingValue = Tools::ps_round($remainingValue, Context::getContext()->getComputingPrecision());
             if (count($order_list) == 1 && $remainingValue > 0 && $cartRule->partial_use == 1 && $cartRule->reduction_amount > 0) {
                 // Create a new voucher from the original
                 $voucher = new CartRule((int) $cartRule->id); // We need to instantiate the CartRule without lang parameter to allow saving it
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -1076,10 +1076,11 @@ abstract class PaymentModuleCore extends Module
         $id_order_state
     ) {
         $cart_rule_used = array();
+        $computingPrecision = Context::getContext()->getComputingPrecision();
 
         // prepare cart calculator to correctly get the value of each cart rule
         $calculator = $cart->newCalculator($order->product_list, $cart->getCartRules(), $order->id_carrier);
-        $calculator->processCalculation(Context::getContext()->getComputingPrecision());
+        $calculator->processCalculation($computingPrecision);
         $cartRulesData = $calculator->getCartRulesData();
 
         $cart_rules_list = array();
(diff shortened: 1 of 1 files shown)


classes/PaymentModule.php
```diff
@@ -1104,6 +1104,7 @@ protected function createOrderCartRules(
             // THEN
             //  The voucher is cloned with a new value corresponding to the remainder
             $remainingValue = $cartRule->reduction_amount - $values[$cartRule->reduction_tax ? 'tax_incl' : 'tax_excl'];
+            $remainingValue = Tools::ps_round($remainingValue, _PS_PRICE_COMPUTE_PRECISION_);
```
matthieu-rolland: Thank you for your contribution @202-ecommerce  !

The` _PS_PRICE_COMPUTE_PRECISION_` constant has been deprecated in 1.7.7 (see this PR #15084 for more informations)

You should be able to get the compute precision this way:

`Context::getContext()->getComputingPrecision()`

matthieu-rolland (changes requested):

classes/PaymentModule.php
matks: I see `Context::getContext()->getComputingPrecision()` is used at line 1182 in this file

matthieu-rolland (dismissed):
lgtm !  thank you @202-ecommerce

matthieu-rolland (dismissed):
checked, lgtm 👍

classes/PaymentModule.php
```diff
@@ -1104,6 +1104,7 @@ protected function createOrderCartRules(
             // THEN
             //  The voucher is cloned with a new value corresponding to the remainder
             $remainingValue = $cartRule->reduction_amount - $values[$cartRule->reduction_tax ? 'tax_incl' : 'tax_excl'];
+            $remainingValue = Tools::ps_round($remainingValue, Context::getContext()->getComputingPrecision());
```
PierreRambaud: `Context::getContext()->getComputingPrecision()` is already defined on line 1081.
Store it in a variable instead.

khouloudbelguith:
Hi @202-ecommerce,

This issue: https://github.com/PrestaShop/PrestaShop/issues/16491 is added to the PS1.7.6 backlog because it is regression, so it should be fixed in the Branch 1.7.6.x & not develop.
Ping @colinegin @marionf is it right?

Thanks!

marionf:
Let's keep it on `develop` for now as 1.7.6.2 is freezed, we will backport it on `1.7.6.x` branch if an 1.7.6.3 is needed

PierreRambaud (approved):

khouloudbelguith:
Hi @202-ecommerce,

It is OK
image
image

https://drive.google.com/file/d/1xn6X_A6LfxqXnIjcnQ7Nr1mgv1yN85L5/view
Thanks!

Progi1984:
Thanks @202-ecommerce

matks:
Another great win 💪 thank you @202-ecommerce ! 😄

PR #16543, opened 2019-11-26
Fixing menu animation to be opened properly

### LINKED ISSUE #10298: Menu animation doesn't open properly
In the BO, when you open the lateral menu, the animation is not the same as when you close the menu: 
Closing animation: 1 - text disappears; 2 - menu closes
Opening animation: 1 - text appears 2 - menu opens
Opening animation should be: 1 - menu opens; 2 - text appears when menu is fully opened
This result in a bad display of the text while the menu opens.

1. Go to the back office
2. Click the icon to close the menu
3. Click on the icon to open the menu 
4. See that text appears too early and makes the animation look strange

PrestaShop version: 1.7.4

NeOMakinG:
… same time

| Questions     | Answers
| Branch?       | develop
| Description?  | In the BO, when you open the lateral menu, the animation was not the same as when you close the menu
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #10298
| How to test?  | Go in the BO and just play with the menu opening and closing, the text when you open it should appear after the sliding

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16543)

### DIFF
--- a/admin-dev/themes/default/js/bundle/components/navbar-transition-handler.js
+++ b/admin-dev/themes/default/js/bundle/components/navbar-transition-handler.js
@@ -1,0 +1,63 @@
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+/**
+ * Toggle a class on $mainMenu after the end of an event (transition, animation...)
+ * @param {jQuery element} $navBar - The navbar item which get a css transition property.
+ * @param {jQuery element} $mainMenu - The menu inside the $navBar element.
+ * @param {string} endTransitionEvent - The name of the event.
+ * @param {jQuery element} $body - The body of the page.
+ * @method showNavBarContent - Toggle the class based on event and if body got a class.
+ * @method toggle - Add the listener if there is no transition launched yet.
+ * @return {Object} The object with methods wich permit to toggle on specific event.
+*/
+function NavbarTransitionHandler($navBar, $mainMenu, endTransitionEvent, $body) {
+  this.$body = $body;
+  this.transitionFired = false;
+  this.$navBar = $navBar.get(0);
+  this.$mainMenu = $mainMenu;
+  this.endTransitionEvent = endTransitionEvent;
+
+  this.showNavBarContent = (event) => {
+    if (event.propertyName !== 'width') {
+      return;
+    }
+
+    this.$navBar.removeEventListener(this.endTransitionEvent, this.showNavBarContent);
+    const isSidebarClosed = this.$body.hasClass('page-sidebar-closed');
+    this.$mainMenu.toggleClass('sidebar-closed', isSidebarClosed);
+    this.transitionFired = false;
+  };
+
+  this.toggle = () => {
+    if (!this.transitionFired) {
+      this.$navBar.addEventListener(this.endTransitionEvent, this.showNavBarContent.bind(this));
+    } else {
+      this.$navBar.removeEventListener(this.endTransitionEvent, this.showNavBarContent);
+    }
+
+    this.transitionFired = !this.transitionFired;
+  };
+}
--- a/admin-dev/themes/default/js/bundle/components/navbar-transition-handler.js
+++ b/admin-dev/themes/default/js/bundle/components/navbar-transition-handler.js
@@ -1,5 +1,5 @@
 /**
- * 2007-2019 PrestaShop SA and Contributors
+ * 2007-2020 PrestaShop SA and Contributors
  *
  * NOTICE OF LICENSE
  *
--- a/admin-dev/themes/default/sass/partials/_nav.sass
+++ b/admin-dev/themes/default/sass/partials/_nav.sass
@@ -147,7 +147,7 @@
 	&.sidebar-closed
 		.link-levelone
 			.link > span
-				display: none;
+				display: none
 
 				&.open
 					> .submenu
--- a/admin-dev/themes/default/sass/partials/_nav.sass
+++ b/admin-dev/themes/default/sass/partials/_nav.sass
@@ -144,6 +144,18 @@
 	padding: 0 0 5.313rem 0
 	margin: 0
 
+	&.sidebar-closed
+		.link-levelone
+			.link > span
+				display: none;
+
+				&.open
+					> .submenu
+						display: none;
+
+		.categort-title > .title
+			display: none;
+
 	.category-title > .title
 		text-transform: uppercase
 
--- a/admin-dev/themes/default/sass/partials/_nav.sass
+++ b/admin-dev/themes/default/sass/partials/_nav.sass
@@ -154,7 +154,7 @@
 						display: none;
 
 		.categort-title > .title
-			display: none;
+			display: none
 
 	.category-title > .title
 		text-transform: uppercase
@@ -429,4 +429,3 @@
 
 	&.expanded
 		display: block
-
--- a/admin-dev/themes/default/sass/partials/_nav.sass
+++ b/admin-dev/themes/default/sass/partials/_nav.sass
@@ -151,7 +151,7 @@
 
 				&.open
 					> .submenu
-						display: none;
+						display: none
 
 		.categort-title > .title
 			display: none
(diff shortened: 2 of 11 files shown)


admin-dev/themes/new-theme/js/nav_bar.js
```diff
@@ -91,6 +90,9 @@ export default class NavBar {
         '.menu-collapse',
         function onNavBarClick() {
           $('body').toggleClass('page-sidebar-closed');
+          setTimeout(() => {
+            $('.main-menu').toggleClass('sidebar-closed');
+          }, 400)
```
matthieu-rolland: Do you think it would be possible to have a callback mechanism here instead of a setTimeout ?

NeOMakinG: The callback mech wouldn't replace the setTimeout because sidebar-closed toggleClass needs to be executed 400ms after the first toggleClass, but you should be right the correct solution would be to include that setTimeout inside a promise resolve so it doesnt get executed if body doesnt exist, but do we really need to care if body isnt existing ?

matthieu-rolland: So if I get it right the 400ms are the time needed for the css animation to be executed.
I guess there's no other way (except to remake the whole animation system). 👍

matthieu-rolland (dismissed):
lgtm

admin-dev/themes/new-theme/js/nav_bar.js
NeOMakinG: `transition: all .5s ease-out;`

Thats set in the transition of .nav-bar yes, the substraction of 100ms is to show the text a bit before the animation's end

PierreRambaud (dismissed):
I'm not sure the usage of setTimeOut is the right way to fix the bug, since it depends on the browser performance.
What happened if you have an old computer without RAM and click on the navigation, are you sure 400ms will be enough?

admin-dev/themes/new-theme/js/nav_bar.js
```diff
@@ -24,24 +24,27 @@
...
+import getAnimationEvent from './app/utils/animations';
 
 const $ = window.$;
 
 export default class NavBar {
   constructor() {
-
     $(() => {
       const $navBar = $('.nav-bar');
+      const navBar = document.querySelector('.nav-bar');
```
eternoendless: Any reason to use this instead of `$navBar`?

admin-dev/themes/new-theme/scss/components/layout/_nav_bar.scss
```diff
@@ -213,6 +213,10 @@
         border-left: .25rem solid #25b9d7;
         padding-left: .638rem;
 
+        @at-root .sidebar-closed .link-levelone .link > span {
```
eternoendless: Why `at-root`? You could just put it at root

NeOMakinG: Seems clearer to me since it's under initial properties of the element

admin-dev/themes/new-theme/js/app/utils/animations.js
```diff
@@ -0,0 +1,48 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+const getAnimationEvent = (type, lifecycle) => {
```
eternoendless, line 32: Please document methods using jsdoc

I should be able to understand what the method does (or what it's for) as well as its parameters and return values without having to read the source code.

admin-dev/themes/new-theme/js/app/utils/animations.js
```diff
@@ -0,0 +1,48 @@
...
+    WebkitTransition: `webkit${typeUpper}${lifecycleUpper}`,
+  };
+
+  Object.keys(properties).forEach((e) => {
+    if (el.style[e] !== undefined) {
+      browserTransition = properties[e];
+    }
+  });
+
+  return browserTransition;
```
eternoendless: I think you mean this:

```js
for (const propName in properties) {
  if (el.style[propName] !== undefined) {
    return properties[propName];
  }
}

return ''; // or null
```

admin-dev/themes/new-theme/js/nav_bar.js
```diff
@@ -86,12 +88,30 @@ export default class NavBar {
...
+        }
+      }
+
       $navBar.on(
         'click',
         '.menu-collapse',
         function onNavBarClick() {
           $('body').toggleClass('page-sidebar-closed');
 
+          if(!transitionFired) {
```
eternoendless: Watch out for spaces
```suggestion
          if (!transitionFired) {
```

admin-dev/themes/new-theme/js/nav_bar.js
```diff
@@ -86,12 +88,30 @@ export default class NavBar {
...
+            transitionFired = true;
+            $navBar[0].addEventListener(endTransitionEvent, showNavBarContent);
+          }else {
```
eternoendless:
```suggestion
          } else {
```

admin-dev/themes/new-theme/js/nav_bar.js
```diff
@@ -86,12 +88,30 @@ export default class NavBar {
           $submenu.find('i.material-icons.sub-tabs-arrow').text('keyboard_arrow_up');
         });
 
+      function showNavBarContent(event) {
```
eternoendless: Watch out for function declaration style: https://github.com/airbnb/javascript#functions--declarations

admin-dev/themes/new-theme/js/nav_bar.js
```diff
@@ -24,24 +24,26 @@
...
 const $ = window.$;
 
 export default class NavBar {
   constructor() {
-
     $(() => {
       const $navBar = $('.nav-bar');
       new PerfectScrollbar($navBar.get(0));
+      let endTransitionEvent = getAnimationEvent('transition', 'end');
+      let transitionFired = false;
```
eternoendless: I think it would be best to have a stateful object that handles this all on its own. For example:

```js
class NavbarTransitionHandler {
  constructor($navBar, $mainMenu, endTransitionEvent) {
    this.transitionFired = false;
    this.$navBar = $navBar[0];
    this.$mainMenu = $mainMenu;
    this.endTransitionEvent = endTransitionEvent;
  }

  toggle() {
    if (!this.transitionFired) {
      transitionFired = true;
      this.$navBar.addEventListener(this.endTransitionEvent, this.showNavBarContent);
    } else {
      this.$navBar.removeEventListener(this.endTransitionEvent, this.showNavBarContent);
      transitionFired = false;
    }
  }

  _showNavBarContent(event) {
    if (event.propertyName == 'width') {
      this.$navBar.removeEventListener(endTransitionEvent, showNavBarContent);

      this.$mainMenu.toggleClass('sidebar-closed');

      this.transitionFired = false;
    }
  }
}
```

Then:

```js
$(() => {
  const $navBar = $('.nav-bar');
  new PerfectScrollbar($navBar.get(0));
  let navbarTransitionHandler = new NavbarTransitionHandler(
     $navBar,
     $('.main-menu'),
     getAnimationEvent('transition', 'end');
  );

  // ... snip

  $navBar.on(
    'click',
    '.menu-collapse',
    function onNavBarClick() {
      $('body').toggleClass('page-sidebar-closed');
 
      navbarTransitionHandler.toggle();
```

This yields more loosely-coupled code that's easier to understand.

admin-dev/themes/new-theme/scss/components/layout/_nav_bar.scss
eternoendless: Tihs is not the intended way of using at-root (https://sass-lang.com/documentation/at-rules/at-root)

Using it this way will make finding the scss declaration significantly harder to do when analyzing the final CSS styles, as it won't be placed in the expected part of the scss tree (the root).

eternoendless (dismissed):

admin-dev/themes/new-theme/js/app/utils/animations.js
PierreRambaud: Nope because of https://eslint.org/docs/rules/no-restricted-syntax which is an airbnb-base eslint configuration :)
image

But the solution should be
```js
return Object.values(properties).find(propName => el.style[propName] !== undefined);
```

admin-dev/themes/default/js/bundle/components/navbar-transition-handler.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * 2007-2019 PrestaShop SA and Contributors
```
Progi1984:
```suggestion
 * 2007-2020 PrestaShop SA and Contributors
```

Progi1984: Please update your headers

admin-dev/themes/default/sass/partials/_nav.sass
```diff
@@ -144,6 +144,18 @@
 	padding: 0 0 5.313rem 0
 	margin: 0
 
+	&.sidebar-closed
+		.link-levelone
+			.link > span
+				display: none;
```
Progi1984:
```suggestion
				display: none
```
You are in `.sass` file. Please don't use semicolon.

admin-dev/themes/default/sass/partials/_nav.sass
```diff
@@ -144,6 +144,18 @@
...
+
+				&.open
+					> .submenu
+						display: none;
```
Progi1984:
```suggestion
						display: none
```

admin-dev/themes/default/sass/partials/_nav.sass
```diff
@@ -144,6 +144,18 @@
...
+
+		.categort-title > .title
+			display: none;
```
Progi1984:
```suggestion
			display: none
```

Progi1984 (commented):
Some feedback

Progi1984 (approved):

PierreRambaud (approved):

SD1982:
This pull request correctly fix the described issue in [#10298](https://github.com/PrestaShop/PrestaShop/issues/10298). Thanks @NeOMakinG

Progi1984:
Thanks @NeOMakinG

PR #16553, opened 2019-11-26
Avoids webservice to create duplicated customers with same email

BlackKerio:
| Questions     | Answers
| Branch?       | develop
| Description?  | We can not create duplicated customers with same email with api
| Type?         | bug fix 
| Category?     | WS
| BC breaks?    | yes, We can not create duplicated customers with same email with api 
| Deprecations? | no
| Fixed ticket? | [12844](https://github.com/PrestaShop/PrestaShop/issues/12844)
| How to test?  | Send several same xml api request with the same email. Now, it is impossible to create 2 customers with same email.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16553)


classes/Customer.php
```diff
@@ -265,6 +269,33 @@ public function add($autoDate = true, $nullValues = true)
...
+     * @param bool $nullValues Whether we want to use NULL values instead of empty quotes values
+     *
+     * @return bool Indicates whether the Customer has been successfully added
+     *
+     * @throws PrestaShopDatabaseException
+     * @throws PrestaShopException
+     */
+    public function addWs($autodate = true, $null_values = false)
+    {
+        if (CustomerCore::customerExists($this->email)) {
```
eternoendless: You should use `Customer` instead of `CustomerCore` 
```suggestion
        if (Customer::customerExists($this->email)) {
```

classes/Customer.php
```diff
@@ -265,6 +269,33 @@ public function add($autoDate = true, $nullValues = true)
...
+            WebserviceRequest::getInstance()->setError(
+                500,
+                $this->trans('The email is already used, please choose another one',
+                array(), 'Admin.Notifications.Error'),
```
eternoendless: According to [PSR-2](https://www.php-fig.org/psr/psr-2/#44-method-arguments):

> Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line.

So arguments need to be split up like this:

```diff
-$this->trans('The email is already used, please choose another one',
-array(), 'Admin.Notifications.Error'),
+$this->trans(
+    'The email is already used, please choose another one',
+    [],
+    'Admin.Notifications.Error'
+),
```

classes/Customer.php
```diff
@@ -265,6 +269,33 @@ public function add($autoDate = true, $nullValues = true)
...
+                140
```
eternoendless, line 293: How did you come up with this number?

eternoendless (dismissed):

classes/Customer.php
PierreRambaud: Maybe one rule  we should add or find in php-cs-fixer, because everything seems ok for it :)

classes/Customer.php
BlackKerio: @eternoendless  I was looking for some info for the new issue of this error, I asked @matks  if there is any directive, the answer was no. So we agreed on a number that still does not exist .... and may change in the future

matks: Yep that sounds right to me. @eternoendless is that OK for you ?

classes/Customer.php
BlackKerio: @eternoendless  I made the changes you suggested

matks (approved):

matks:
Code seems ✅ to me but let's make sure @eternoendless agrees too 😉

PierreRambaud (dismissed):

matks (dismissed):

khouloudbelguith:
@BlackKerio, thanks for your contribution, would you mind to rebase your pull request.
PS: there is an error while creating a new customer via WS with an old address mail
image

Thanks!

BlackKerio:
@khouloudbelguith error 500 is an expected error .... if the sent data is incorrect, an existing email

matks:
> @khouloudbelguith error 500 is an expected error .... if the sent data is incorrect, an existing email

@khouloudbelguith this is what the PR is supposed to do 😉 (I updated the PR description to make it more clear)

khouloudbelguith:
Hi @BlackKerio,

It is OK.
image


Thanks!

matks (approved):

matks:
Thanks @BlackKerio

mowcixo:
I think I arrive too late, but... why HTTP 500? that http code should not be returned intentionally, why not an HTTP 400 or HTTP 409?

matks:
> I think I arrive too late, but... why HTTP 500? that http code should not be returned intentionally, why not an HTTP 400 or HTTP 409?

It's never too late until the release is shipped 😄 

ping @BlackKerio

PierreRambaud:
@mowcixo @matks The reason is old, it's the same behaviour as the Product.php file:
https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Product.php#L6127
It's ugly and not really logic, but we wanted to keep the same behaviour even if a 400 error is more legit

mowcixo:
Ok then, but I think it should be considered to change globally, because now it is so difficult to generize exception treatment all over webservice. It's needed to check the exception message to be sure if it's a breaking server error or just a "form" error.

Thanks for the PR though! :D

PR #16560, opened 2019-11-27
Fix header for migrated Orders pages

### LINKED ISSUE #16549: No create order button in orders listing for SF pages
## Describe the bug

I noticed an issue on migrated Orders listing (available at URL `/admin-dev/index.php/sell/orders/orders/`).

The "create an order" button is not there ... and actually the top item of each BO page is not there at all.
Capture d’écran 2019-11-26 à 14 19 35

This is actually because of PR https://github.com/PrestaShop/PrestaShop/pull/16090 who aimed to make "Edit an order" display as close as possible from original mock-up.

So this is the design, thanks to PR https://github.com/PrestaShop/PrestaShop/pull/16090
Capture d’écran 2019-11-25 à 18 22 00

Without this PR, "Edit an order" migrated page looks like this:
Capture d’écran 2019-11-25 à 18 21 52

However an unexpected side-effect of PR is that it modifies also "Create an order" and "Orders listing" migrated pages.

## Possible solutions

1) Put the banner back, so we get this display on Order page:
Capture d’écran 2019-11-25 à 18 21 52
This is the most simple because this way, the page behaves like all other BO pages.

2) Hack further into the legacy layout to have listing display correct, but edit order identical to mock-up. However this might break other things.

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix header for migrated Orders pages
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/16549
| How to test?  | Please see ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16560)

### DIFF
--- a/admin-dev/themes/new-theme/template/controllers/orders/page_header_toolbar.tpl
+++ b/admin-dev/themes/new-theme/template/controllers/orders/page_header_toolbar.tpl
@@ -25,7 +25,7 @@
 {extends file='page_header_toolbar.tpl'}
 
 {block name=pageTitle}
-  {if isset($use_regular_h1_structure) == false}
+  {if !isset($use_regular_h1_structure)}
     {assign var="use_regular_h1_structure" value=true}
   {/if}
 
--- a/admin-dev/themes/new-theme/template/controllers/orders/page_header_toolbar.tpl
+++ b/admin-dev/themes/new-theme/template/controllers/orders/page_header_toolbar.tpl
@@ -22,11 +22,18 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  *}
-<div class="header-toolbar">
-  <div class="container-fluid">
-      {literal}
-        {% block page_header_toolbar %}{% endblock %}
-      {/literal}
-  </div>
-  {hook h='displayDashboardTop'}
-</div>
+{extends file='page_header_toolbar.tpl'}
+
+{block name=pageTitle}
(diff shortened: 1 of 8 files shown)


admin-dev/themes/new-theme/template/controllers/orders/page_header_toolbar.tpl
```diff
@@ -22,11 +22,18 @@
...
-      {literal}
-        {% block page_header_toolbar %}{% endblock %}
-      {/literal}
-  </div>
-  {hook h='displayDashboardTop'}
-</div>
+{extends file='page_header_toolbar.tpl'}
+
+{block name=pageTitle}
+  {if isset($use_regular_h1_structure) == false}
```
NeOMakinG: Simple question

`{if !isset($use_regular_h1_structure)}` or `{if isset($use_regular_h1_structure) == false}`, one of these is best practice or we can use both ?

PierreRambaud:
```suggestion
  {if !isset($use_regular_h1_structure)}
```
otherwise use 
```
 {if isset($use_regular_h1_structure) === false}
```

NeOMakinG: Pierre told me

PierreRambaud (approved):

Robin-Fischer-PS:
Hi @matks ,
It looks fine to me now :)

Thanks !

PR #16573, opened 2019-11-27
CLOSED - Fix prestashop gifts and cart calculation when gift separation is enabled

### LINKED ISSUE #9757: [BOOM-4701] Free GIFT is not displayed correctly on cart
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-4701).

- _**Reporter:**_ tantan199
- _**Created at:**_ Fri, 26 Jan 2018 10:50:03 +0100

I added a gift to all products on my store, the problem is that the price of the gift is displayed on the shopping cart page, another problem is that the quantity of the gift is adjustable. See attached pictures.

Because the cart variable from the `controller/front/CartController.php` filed gets override by another cart variable from the assignGeneralPurposeVariables function in the `/classes/controller/ProductListingFrontController.php`

**Temporary solution is change this code**
`'cart' => $this->cart_presenter->present($this->context->cart),`

in the assignGeneralPurposeVariables function to 

`cart' => $this->cart_presenter->present($this->context->cart, $shouldSeparateGifts = true),`

The same problem also exist in the Shopping cart module.

---
When I apply a cart rule that offers an item, it should behave the same as in 1.6

- when adding a product that meets the conditions of a cart rule in the cart, do not take into account the gift price calculations. Currently, we add the product with its price and then apply a reduction of this amount.
- **The recycle bin to delete a gift does not work** (see #24831)
- display the product as a gift, remove the possibility of modifying the quantity and no longer display the price either in the basket and at the time of payment.

Capture du 2018-02-20 11-25-39

- If my cart rule offers an item & a reduction (in amount or %), a "Discount" field appears in the cart.
- This field must not include the amount of the product offered but only the reduction -> OK.
- Same for the orange line with the promo, it must not include the amount of the product offered but only the reduction -> NOK

Capture du 2018-04-17 15-48-03

Capture du 2018-04-17 15-47-50

- In the case that I buy a product that is offered, 2 lines must appear in the cart.
- The discount field should not include the amount of the product offered but only the discount -> NOK.
- Same for the orange line with the promo, it must not include the amount of the product offered but only the reduction -> NOK

Capture du 2018-04-17 16-12-58

Capture du 2018-04-17 15-55-47

We should also replace "Discount" by "Total vouchers" as in 1.6, it was clearer like this

### LINKED ISSUE #16486: Cart presenter totals wrong when shouldSeparateGifts is true
**Describe the bug**
In order to display cart gifts correctly, CartPresenter->present must be called with shouldSeparateGifts as true. This was proposed in this PR: #8706.

**To test, change bold to true:**
classes/controller/FrontController.php - Line 490 - 'cart' => $this->cart_presenter->present($this->context->cart, **true**),

Now, this works until you put another pieces of the _product you got as a gift_ to cart.

**1 - WRONG TOTALS**
From then, cart totals are wrong, because the totals are calculated from wrong quantity variable.

**Example**
Product ABC with price 100.
You add 5 to cart manually.
You get 1 free because you got over XXX amount or applied a voucher.
Total products should be 600, but it is 1200. 

**2 - WRONG DISCOUNTS**
Another bug - product gets discounted twice.

**Additional information**
PrestaShop version: 1.7.6.0.

Hlavtox:
| Questions     | Answers
| Branch?       | develop
| Description?  | This PR fixes display of gift products in cart and allows them to be displayed separately. The UI was provided, but it was disabled probably due to wrong calculations, which this PR fixes.
| Type?         | bug fix / improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16486 & Fixes #9757
| How to test?  | Enable some gift rules, enable some discount rules, add everything to cart, all calculations will be correct.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16573)


micka-fdz:
Any news about this PR? I tried on 1.7.6.1 and it seems work.

jf-viguier:
Nice work @Hlavtox I want this to be merged too, it's a really annoying bug.

Hlavtox:
I use it in production on 4 stores that make a living for several people. Works fine so far 👍

jf-viguier:
@Hlavtox thanks for your feedback, 4 stores, that should be enough.
There is some conflicts now, can you rebase ?

Hlavtox:
@jf-viguier Rebased 🍻

jf-viguier:
🍻

src/Core/Cart/CartRow.php
```diff
@@ -245,7 +245,7 @@ public function getFinalTotalPrice()
     public function processCalculation(Cart $cart)
     {
         $rowData = $this->getRowData();
-        $quantity = (int) $rowData['cart_quantity'];
+        $quantity = (int) $rowData['quantity'];
```
eternoendless, line 248: This change is puzzling, because here you're using the stock quantity instead of the number of items in the cart (see https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Cart.php#L642).

eternoendless (commented):
I'm struggling to understand this change. Could you please explain what was the problem and how does your change fix it?

src/Core/Cart/CartRow.php
Hlavtox: @eternoendless

It must be overwritten later in the code.

Quantity = quantity of items of that row.
Cart_quantity = total quantity of that product in cart.

Make a cart rule with gift and apply it. The price of the product is 100.
Then, add 3 more pieces of that product normally.

Without gift separation:
- Row 1 - cart_quantity = 4, quantity = 4
Everything is fine. Total is 4x100 = 400

With gift separation enabled:
- Row 1 - cart_quantity 4, quantity 1
- Row 2 - cart_quantity 4, quantity 3
Total is 4x100 + 4x100 = 800 = wrong

After changing cart_quantity to quantity, its 1x100 + 3x100 = 400 = correct

Hlavtox:
@eternoendless And regarding the second change in cart rule calculator:

The code loops through items in the cart, and if the id matches, it discounts it.
But, it does so both for the gift product and also for the normal product = discounts it twice.
A simple break will solve this for us.

The fixes  (at least for me) work correctly for both giftseparation turned on and off.

src/Core/Cart/CartRuleCalculator.php
```diff
@@ -117,6 +117,7 @@ protected function applyCartRule(CartRuleData $cartRuleData, $withFreeShipping =
                 ) {
                     $cartRuleData->addDiscountApplied($cartRow->getInitialUnitPrice());
                     $cartRow->applyFlatDiscount($cartRow->getInitialUnitPrice());
+                    break;
```
eternoendless, line 120: This means there can be only one gift

src/Core/Cart/CartRow.php
eternoendless: > It must be overwritten later in the code.

Found it! https://github.com/PrestaShop/PrestaShop/blob/1.7.6.0/classes/Cart.php#L842-L844

eternoendless:
I'm not extremely familiar with the cart calculation system, but your change _seems_ okay.

How do you feel about adding a [new test case here](https://github.com/PrestaShop/PrestaShop/blob/1.7.6.0/tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/CartRule/gift.feature) that validates this bug fix? It should fail without your change and succeed with it. Adding automated tests help us verify that your change does what it's supposed to do, and keeps doing it afterwards.
 

Also, I see you have merged develop into your branch. Please avoid doing that as a bad merge can inadvertently revert changes from others (this has already happened before). Please either revert the merge commit and rebase or simply squash all your commits into a single one.

jf-viguier:
@eternoendless IMO the cart calculation system needs a complete rewrite. It's so difficult to read, to debug and to extends this legacy code !

eternoendless:
@jf-viguier Of course. The cart calculation was partially rewritten in 1.7.4 but it was extremely hard to do it without rewriting everything around it as well. It has been vastly improved though, and above all, it has been made testable. We fixed a ton of bugs thanks to that.

jf-viguier:
@eternoendless  your plan is to rewrite from scratch for the next major release or keep all this mess for backward compatibility ? There is still a lot of bugs in calculation, cart rules, ecotax, ... IMO the only way is to rewrite from scratch, module devs will adapt.

Hlavtox:
@eternoendless @jf-viguier 

Kill me now. I totally fucked it up. 🤦‍♂ 

I made a new PR, continue there pls. https://github.com/PrestaShop/PrestaShop/pull/17825

Issue #16283, opened 2019-11-05
Unable to remove customized product from cart - minimum quantity error


Hlavtox:
**Describe the bug**
During testing, I got the store to a situation, where I couldn't delete product from cart. It happened only on my phone and I can't (for now) reproduce it on desktop.

- When I attempted to delete product from cart or reduce amount to 0, it threw **'You must add %quantity% minimum quantity'**.
- Adding and reducing count above 1 worked OK.
- Behavior was the same if it was only this product in cart, or I added some another.

**Steps to reproduce the issue:**

- Customize a product with a file.
- Add it to cart.
- Refresh the page, customize the product again, do not add it to cart.
- Go to cart, try to remove product. Will not work.
- After my fix, will work as it should and the logic will count only customized products that ARE in cart.

**Possible fix**
If we added two more conditions to the query, it should fix it.
`in_cart = 1 AND quantity > 0`
Another question is, why and how are these entries which are not in_cart and have zero quantity even created?

**How it looked (ignore the variables I output to the error)**

![image0](https://user-images.githubusercontent.com/6097524/68226186-e7a40c80-fff1-11e9-811f-e220f7245072.png)

**The code causing the error**

![code](https://user-images.githubusercontent.com/6097524/68226209-f2f73800-fff1-11e9-90d0-9a5d9374eb37.JPG)

**Database status for given id_cart and id_product**

![table](https://user-images.githubusercontent.com/6097524/68226246-05717180-fff2-11e9-8d80-a007553e7803.JPG)

**Database status for given query**

![select](https://user-images.githubusercontent.com/6097524/68226330-2f2a9880-fff2-11e9-9e09-2ee184296a6e.JPG)

**Additional information**
PrestaShop version: 1.7.6.0
PHP version: 7.2

khouloudbelguith:
Hi @Hlavtox,

Thanks for your report.
We have recently fixed two issues in PS1.7.7.0 similar to this problem.
1. Customized product can't be deleted from the cart reported in this ticket: https://github.com/PrestaShop/PrestaShop/issues/15986 & fixed by this PR: https://github.com/PrestaShop/PrestaShop/pull/15987
2. Removal of customization makes impossible to remove a product from the cart reported in this ticket: https://github.com/PrestaShop/PrestaShop/issues/15624 fixed by this PR: https://github.com/PrestaShop/PrestaShop/pull/15625

Thanks to check & feedback.

Hlavtox:
[...]

PR #16574, opened 2019-11-27
Fix product cannot be deleted from cart IF another customization is made and not added to cart yet

Hlavtox:
| Questions     | Answers
| Branch?       | develop
| Description?  | If customization for current product_id and cart_id exists and is not added to cart, it is not possible to remove any instance of this product from cart. The logic checks for amount of other customizations in cart, to check minimum quantity. BUT, somebody forgot to add a condition, so it checks only customizations that are actually in cart and have some quantity. This PR restores the correct function.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #16283
| How to test?  | -

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16574)


atomiix (approved):

Progi1984:
Thanks @Hlavtox

PR #6785, opened 2016-10-25
Fixed load of Jquery-ui JS/CSS cache (debug)

aleeks:
| Questions | Answers |
| --- | --- |
| Branch? | debug (develop) |
| Description? | Fixed load of Jquery-ui JS/CSS cache |
| Type? | bug fix |
| Category? | BO |
| BC breaks? | no |
| Deprecations? | no |
| Fixed ticket? |  |
| How to test? | ping @julienbourdeau |

#### Important guidelines
- Make sure [your local branch is up to date](https://help.github.com/articles/syncing-a-fork/) before commiting your changes!
- Your code MUST respect [our Coding Standards](http://doc.prestashop.com/display/PS16/Coding+Standards) (for code written in PHP, JavaScript, HTML/CSS/Smarty/Twig, SQL)!
- Your commit name MUST respect our [naming convention](http://doc.prestashop.com/display/PS16/How+to+write+a+commit+message)!

PR #6656, opened 2016-10-11
Rework asset management

julienbourdeau:
| Questions | Answers |
| --- | --- |
| Branch? | develop |
| Description? | Introduce a new asset management system to register assets (css and js). All previous method are deprecated BUT backward compatibility for `addCSS`, `addJS`, `removeCSS` and `removeJS` is maintained. I would very very strongly advise you to update your module to not rely on them. **This require new module version (PR in progress)** |
| Type? | new feature |
| Category? | FO |
| BC breaks? | no |
| Deprecations? | no |
| Fixed ticket? |  |
| How to test? | Try to add CSS/JS in your `theme.yml` or create a module that register assets the proper way |
## TODO
- [x] Introduce new simple method to register assets
- [x] Handle head and bottom position for javascript
- [x] Handle priority order for all assets
- [x] Allow theme developer to manage their assets in `theme.yml`
- [x] Allow module to inline JavaScript (for google analytics for example)
- [x] Allow module assets override
- [x] Work with parent theme
- [x] Ensure theme can unregister module assets by placing empty file in module override directory
- [x] Ensure CDN works (**note:** CDN 2 and 3 have been removed)
- [x] ~~Ensure all JS can be forced to load at the bottom (`PS_JS_DEFER`)~~ Feature became useless and have been removed
- [ ] Document everything
- [ ] Use new method in native modules
- [x] Ensure CCC is fully working
- [x] Allow javascript to be loaded in `async` or `defer`
- [x] Add backward compat' for `addCSS`, `addJS`, `removeCSS`, `removeJS`.
### Example to load assets depending on pages with your `theme.yml`

``` yaml
assets:
  css:
    all:
      - id: custom-lib-style
        path: assets/css/custom-lib.css
    product:
      - id: product-style
        path: assets/css/product.css
        media: all
        priority: 200
  js:
    cart:
      - id: cat-extra-lib
        path: assets/js/cart-lib.js
```
### Prority
- By default priority is set to 50.
- Priority zero is `core.js` and `theme.css`. `theme.js` is loaded with priority 50 in order to let you load libraries before.
- `custom.css` and `custom.js` are loaded with priority 1000 since they are meant to be edited by hand and override everything.
- By default javascript is loaded at the bottom. Modules should always try to load javascript at the bottom unless there is a special need (google analytics or modernizer for example). Remember jQuery2 is provided in core.js, at the bottom of the page.
### How to register assets for a module

In your module, use the `actionFrontControllerSetMedia` hook.

Example:

``` php
[...]

PR #16590, opened 2019-11-28
Update addJqueryUI - Fix Issue #12035

### LINKED ISSUE #12035: FrontController::addJqueryUI() is not working
**Describe the bug**
FrontController:: addJqueryUI() is not working, indeed, the $component variable is not used in the method....

**To Reproduce**
Try to register any jqueryui plugin.

**Additionnal information**
PrestaShop version: 1.7
PHP version: N/A

MarkALeonard:
Updates the function to add individual JqueryUI components and dependencies to the FO instead of the whole JqueryUI library. Eg. Category page ui-slider is used but entire ui library is loaded. This reduces the size of the page.
Function uses Media::getJqueryUIPath to return js and css paths required. 
Alters path and filename to use minified versions of stylesheets.
Creates Id for each file from the filename.
Registers stylesheet and javascript files with priority 51 so as to be added after core.js, theme.css and theme.js
Fix breaks ps_facetedsearch due to incorrect call of addJquery('slider') instead of addJqueryUi('ui.slider')

| Questions     | Answers
| Branch?       | develop 
| Description?  | Updates the FrontController::addJqueryUI function to add individual JqueryUI components and dependencies to the FO instead of the whole JqueryUI library. On category page for example ui-slider is used but entire ui library is loaded. This reduces the size of the page.
| Type?         |  improvement / bug fix
| Category?     | FO 
| BC breaks?    | ps_facetedsearch breaks due to incorrect call of function with 'slider' instead of 'ui-slider'
| Fixed ticket? | "Fixes #12035"
| How to test?  | Compare FO css and js files loaded on category page before and after change. 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16590)

### DIFF
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -1205,14 +1205,12 @@ class FrontControllerCore extends Controller
         }
 
         foreach ($component as $ui) {
-
             $ui_path = Media::getJqueryUIPath($ui, $theme, $check_dependencies);
 
             //Css array is an array of media types indexed by path in order: ui-theme-css, ui-component-css, ui-dependency-css
             foreach ($ui_path['css'] as $css_path => $css_media) {
                 $css_name = basename($css_path);
                 $css_id = str_replace('.', '-', $css_name);
-
                 $css_path_min = str_replace($css_name, 'minified/' . $css_name, $css_path);
                 $css_path_min = str_replace('.css', '.min.css', $css_path_min);
                 
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -1200,13 +1200,35 @@ class FrontControllerCore extends Controller
      */
     public function addJqueryUI($component, $theme = 'base', $check_dependencies = true)
     {
-        $css_theme_path = '/js/jquery/ui/themes/' . $theme . '/minified/jquery.ui.theme.min.css';
-        $css_path = '/js/jquery/ui/themes/' . $theme . '/minified/jquery-ui.min.css';
-        $js_path = '/js/jquery/ui/jquery-ui.min.js';
+        if (!is_array($component)) {
+            $component = array($component);
+        }
 
-        $this->registerStylesheet('jquery-ui-theme', $css_theme_path, ['media' => 'all', 'priority' => 95]);
-        $this->registerStylesheet('jquery-ui', $css_path, ['media' => 'all', 'priority' => 90]);
-        $this->registerJavascript('jquery-ui', $js_path, ['position' => 'bottom', 'priority' => 90]);
+        foreach ($component as $ui) {
+
+            $ui_path = Media::getJqueryUIPath($ui, $theme, $check_dependencies);
+
+            //Css array is an array of media types indexed by path in order: ui-theme-css, ui-component-css, ui-dependency-css
+            foreach ($ui_path['css'] as $css_path => $css_media) {
+
+                $css_name = basename($css_path);
+                $css_id  = str_replace(".", "-", $css_name);
+
+                $css_path_min = str_replace($css_name,'minified/' . $css_name, $css_path);
+                $css_path_min = str_replace('.css','.min.css', $css_path_min);
+                
+                $this->registerStylesheet($css_id, $css_path_min, ['media' => $css_media, 'priority' => 51]);
+            }
+
+            //Js array is an indexed array of paths in order of dependency
+            foreach ($ui_path['js'] as $js_path) {
+
+                $js_id = basename($js_path,".js");
+                $js_id  = str_replace(".", "-", $js_id);
+
+                $this->registerJavascript($js_id, $js_path, ['position' => 'bottom', 'priority' => 51]);
+            }
+        }
     }
 
     /**
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -1210,21 +1210,19 @@ class FrontControllerCore extends Controller
 
             //Css array is an array of media types indexed by path in order: ui-theme-css, ui-component-css, ui-dependency-css
             foreach ($ui_path['css'] as $css_path => $css_media) {
-
                 $css_name = basename($css_path);
-                $css_id  = str_replace(".", "-", $css_name);
+                $css_id  = str_replace('.', '-', $css_name);
 
-                $css_path_min = str_replace($css_name,'minified/' . $css_name, $css_path);
-                $css_path_min = str_replace('.css','.min.css', $css_path_min);
+                $css_path_min = str_replace($css_name, 'minified/' . $css_name, $css_path);
+                $css_path_min = str_replace('.css', '.min.css', $css_path_min);
                 
                 $this->registerStylesheet($css_id, $css_path_min, ['media' => $css_media, 'priority' => 51]);
             }
 
             //Js array is an indexed array of paths in order of dependency
             foreach ($ui_path['js'] as $js_path) {
-
-                $js_id = basename($js_path,".js");
-                $js_id  = str_replace(".", "-", $js_id);
+                $js_id = basename($js_path, '.js');
+                $js_id  = str_replace('.', '-', $js_id);
 
                 $this->registerJavascript($js_id, $js_path, ['position' => 'bottom', 'priority' => 51]);
             }
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -1201,23 +1201,23 @@ class FrontControllerCore extends Controller
     public function addJqueryUI($component, $theme = 'base', $check_dependencies = true)
     {
         if (!is_array($component)) {
-            $component = array($component);
+            $component = [$component];
         }
 
         foreach ($component as $ui) {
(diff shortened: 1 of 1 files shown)


PierreRambaud:
Facetedsearch use `addJqueryUi('slider')` and it's working https://github.com/PrestaShop/ps_facetedsearch/blob/master/src/Hook/ProductSearch.php#L55 :thinking: 
But the current code is loading the jquery-ui.min.js file. So everything is imported.
Could you please fix php-cs-fixer before beginning the review?

MarkALeonard:
Apologies for the delay. When running php-cs-fixer I was getting a lot of changes. I realised that when using php-cs-fixer globally it gave very different results to running the version in vendor/bin. They both used the same configuration file is it just the version (2.16.1 vs 2.15.1) that would make the fixes so different? Thank you.

PierreRambaud:
Everything looks ok for php-cs-fixer, but I'm not sure we can merge your PR because of this pull request: https://github.com/PrestaShop/PrestaShop/pull/6785

ping @eternoendless @matks wdyt?

MarkALeonard:
Ok, I understand that if this was merged it would be a step towards how the function previously worked. I can also see that the original code was a more elegant solution than mine although it would need updating. 

However, I don't understand why if the function was depreceated in 1.7 it was added back. Were there a lot of FO modules using it that required it to be reinstated? If so why was it done so with less functionallity and in doing so creating larger pages. 

Eg. On a category page using a basic Prestashop install, just to render the slider used by ps_facetedsearch, the current code loads: 

jquery.ui.theme.min.css
jquery-ui.min.css
jquery-ui.min.js 
(263kb) 

Using the dependency based version the following are loaded:

jquery.ui.theme.min.css
jquery.ui.slider.min.css
jquery.ui.core.min.css
jquery.ui.core.min.js
jquery.ui.widget.min.js
jquery.ui.mouse.min.js
jquery.ui.slider.min.js
(48kb)

I think that is a reasonable enough saving to warrant changing the function, but that depends on whether it is to be depreceated in the future again or not.

NeOMakinG:
ping @PrestaShop/prestashop-core-developers we would need some advices regarding this PR, it's a good front-end upgrade as it reduce load size but not sure about side effects

matks:
> ping @PrestaShop/prestashop-core-developers we would need some advices regarding this PR, it's a good front-end upgrade as it reduce load size but not sure about side effects

@NeOMakinG maybe now is the good time to do it 😉 as we prepare a new major version

eternoendless:
This was [originally meant to be removed](https://github.com/PrestaShop/PrestaShop/commit/29f6e81699092d527c630da2f1c18de15914876a#diff-dd62c91b35b9cbbc6ac792f1b44b24279c77f54c9cca97ea773ae297e93e9c61R1021) but was reintroduced later in the [same PR](https://github.com/PrestaShop/PrestaShop/pull/6656) to keep backward compatibility. The original intent was to make each module handle its own libraries. IMO that sounds like the better approach, right?

matks:
> The original intent was to make each module handle its own libraries. IMO that sounds like the better approach, right?

I agree. @Quetzacoalt91 @kpodemski I would be happy to have your feedback about this topic

kpodemski:
@matks 

To be honest, I'm waiting for the Jquery UI to be removed entirely. We should aim to terminate its use from the core and native modules.

> This was originally meant to be removed but was reintroduced later in the same PR to keep backward compatibility. The original intent was to make each module handle its own libraries. IMO that sounds like the better approach, right?

Of course, there's a risk of two modules loading the same library, but I want to believe in developers common sense 😁

matks (commented):
Hello everyone.

1. The PR is valid IMO. But as mentioned by many people, today is 2021 (sorry for the delay 😞 ) and we are preparing PrestaShop 8.0.0 . It might be a good time to get rid of jQueryUI and consequently get rid of this method.

2. So I think we should rather deprecate this function.

3. Can we even delete it? I think so but I dont want this to become a hinder to people who upgrade from 1.7 to 8. So I need an opinion from @PrestaShop/prestashop-maintainers .

kpodemski:
I think we should deprecate it. There is a lot of better alternatives. Also:

> jQuery UI is in maintenance-only mode. Users should not expect any new releases, though patches may be issued to resolve critical security, interoperability, or regression bugs. Trac, the project’s bug-tracking tool, has been put in read-only mode and developers are asked to file any critical issues on the project’s GitHub repository. 

The only native modules that use JqueryUI are `ps_searchbar`, and `ps_facetedsearch` if I remember correctly.

kpodemski:
I'll allow myself to ping some devs that might have feedback about it.

@PrestaEdit @idnovate @iqit-commerce @Oksydan @Prestaworks

Oksydan:
According to our vision of new theme, we are not using jQuery in theme. It is only included there for backwards compatibility. 
So we should get rid of jQuery.UI entirely, by rewriting native modules to don't use jQuery at all. 
Reworked native module will be working for legacy classic and new theme since it will not be relying on jQuery/jQuery UI at all.

idnovate:
> I'll allow myself to ping some devs that might have feedback about it.
> 
> @PrestaEdit @idnovate @iqit-commerce @Oksydan @Prestaworks

We don't use this library in the FO in any of our modules, only in the BO. 
Probably this change doesn't have many side effects (external modules affected).

> I think we should deprecate it. There is a lot of better alternatives. 

Do you mean to offer a better alternative library or leave each developer use its library?

Prestaworks:
> 
> 
> > I'll allow myself to ping some devs that might have feedback about it.
> > @PrestaEdit @idnovate @iqit-commerce @Oksydan @Prestaworks
> 
> We don't use this library in the FO in any of our modules, only in the BO. Probably this change doesn't have many side effects (external modules affected).
> 
> > I think we should deprecate it. There is a lot of better alternatives.
> 
> Do you mean to offer a better alternative library or leave each developer use its library?

Our frontend devs thinks it's a good idea to drop it, it's never used by them anyway they told me.
As a Backend dev I think it would be nice if there was some kind of "addLibrary(NAME, VERSION, URL)" function so we can avoid adding an already included library, in case of several modules depending on the same library being loaded in the frontend. But that's more of a "nice to have" feature.

kpodemski:
@Prestaworks 

We have `registerStylesheet`, and `registerJavascript` with the `id` provided. In an ideal world, the dev should check if the library is already available.

recent changes introduced the possibility of adding a version number too, but the entire mechanism would need to be checked if it is something that will help us to prevent double loads of the libraries 👍

kpodemski:
@idnovate 

> Do you mean to offer a better alternative library or leave each developer use its library?

I would leave the choice to the developer. Another option is to bundle preferred libraries just like we did with Fancybox.

idnovate:
> @idnovate
> 
> > Do you mean to offer a better alternative library or leave each developer use its library?
> 
> I would leave the choice to the developer. Another option is to bundle preferred libraries just like we did with Fancybox.

Then as you said previously we could have the problem with duplicated libraries included by different modules. But let's rely on the "ideal world" :blush:

Prestaworks:
> 
> 
> @Prestaworks
> 
> We have `registerStylesheet`, and `registerJavascript` with the `id` provided. In an ideal world, the dev should check if the library is already available.
> 
> recent changes introduced the possibility of adding a version number too, but the entire mechanism would need to be checked if it is something that will help us to prevent double loads of the libraries 👍

Maybe it could be a good idea to set a naming convention for common libraries so that the ids are names in the same way by different devs?
In the examples in the docs there are 'module-modulename-simple-lib' and 'google-analytics' mentioned as examples.

Then you also have this sentence
"However, avoid generic names when adding new JS and CSS files to avoid collision with other extensions. Prefixing with your module or theme name is a good start."

That if followed would give unique ids for the same library.

Oksydan:
Working on theme-refacto we could include some libraries and expose theme globally and make them something like standardized library list like we are doing with bootstrap. 
Everybody knows that bootstrap is available so they can use all of it's components.
What about making something like slider library available globally or event expose slider API that simplify building sliders on page. 
Thb most modules are including a lot of pointless js/css libraries. They try to solve trivial problems with huge libraries coz developers are just lazy. We can't do much that quality of modules are just poor.

Prestaworks:
> 
> 
> Working on theme-refacto we could include some libraries and expose theme globally and make them something like standardized library list like we are doing with bootstrap. Everybody knows that bootstrap is available so they can use all of it's components. What about making something like slider library available globally or event expose slider API that simplify building sliders on page. Thb most modules are including a lot of pointless js/css libraries. They try to solve trivial problems with huge libraries coz developers are just lazy. We can't do much that quality of modules are just poor.

I think its a good idea to try to avoid libraries as much as possible in the core. if a specific module or theme (including "classic") needs libraries to work, the theme/module should include that library. If not, you will just have to keep a bunch of libraries up2date. But if you do add libraries, do not change them like what happened with font-awesome :)

eternoendless:
Related: #26290

matks:
Hello everyone, I think this PR can be closed because the initial issue is being explored (and will be solved) in https://github.com/PrestaShop/PrestaShop/issues/26290 I think.

Thanks @MarkALeonard for starting the initial topic :) though

PR #16620, opened 2019-11-29
Prevent order controller from validating all guest addresses

### LINKED ISSUE #16584: [Bug][Performance] Guest checkout validates all the guest addresses (thousands)
**Describe the bug**
If you have a lot of guest checkouts, you'll have a lot of guest addresses (`ps_address.id_customer = 0`).

This causes a problem since, currently, the [checkout process validates all the customer's addresses](https://github.com/PrestaShop/PrestaShop/blob/a51084f29cb87627ad57fde9d7932057c225b4e1/controllers/front/OrderController.php#L230-L233). In the case you're a guest, that means **all guest addresses** are validated.

If you have a lot (thousands or tens of thousands) of guest checkouts, all the addresses are being validated at any guest checkout, making it ever more slower.

**To Reproduce**
1. Create 10k guest fake customer addresses, `ps_address.id_customer = 0`
2. Try to do a new guest checkout
4. Observe seconds of delay in the `POST /order` request

**Additional information**
PrestaShop version: latest, linked
PHP version: 7.2

slri:
| Questions     | Answers
| Branch?       | develop
| Description?  | Order controller has two address validation checks, one is run for the cart addresses and the other is run for all customer addresses (from context). For guest checkout there's no id_customer added to the address or the guest, therefore we shouldn't try to validate a non-existing customer's addresses.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16584
| How to test?  | 1. Create tens of thousands of addresses. <br> 2. Navigate to /order <br> 3. Observe the time it takes to resolve the request.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16620)

### DIFF
--- a/controllers/front/OrderController.php
+++ b/controllers/front/OrderController.php
@@ -238,10 +238,13 @@ class OrderControllerCore extends FrontController
             $checksum = $this->cartChecksum->generateChecksum($cart);
         }
 
-        // Prepare all other addresses' warning messages (if relevant).
-        // These messages are displayed when changing the selected address.
-        $allInvalidAddressIds = $addressValidator->validateCustomerAddresses($customer, $this->context->language);
-        $this->checkoutWarning['invalid_addresses'] = $allInvalidAddressIds;
+        // Prevent check for guests
+        if($customer->id) {
+            // Prepare all other addresses' warning messages (if relevant).
+            // These messages are displayed when changing the selected address.
+            $allInvalidAddressIds = $addressValidator->validateCustomerAddresses($customer, $this->context->language);
+            $this->checkoutWarning['invalid_addresses'] = $allInvalidAddressIds;
+        }
 
         if (isset($data['checksum']) && $data['checksum'] === $checksum) {
             $process->restorePersistedData($data);
--- a/controllers/front/OrderController.php
+++ b/controllers/front/OrderController.php
@@ -239,7 +239,7 @@ class OrderControllerCore extends FrontController
         }
 
         // Prevent check for guests
-        if($customer->id) {
+        if ($customer->id) {
             // Prepare all other addresses' warning messages (if relevant).
             // These messages are displayed when changing the selected address.
             $allInvalidAddressIds = $addressValidator->validateCustomerAddresses($customer, $this->context->language);


PierreRambaud (approved):

eternoendless:
Thank you @slri

ComonSoft:
Hello team,

we have two shops with more than  250K customers 390K address. It takes about 35 seconds on the order page on a high dedicated server ! How this bug can be considered as QA ? It's major or Prestashop only support small business ?
In PS 1.6 it was instantaneous, this is a regretssion still 1.7.0, and still not deployed. I really don't undertand priorities.

The 1rst shop that migrate from 1.6 to 1.7 recently has 210K customers and 345K addresses, about 30 seconds when a customer goto order page. We have lost 50% of turnover since migration. The transformation rate continues to fall. This generates 79889 SQL queries and slowdown the server !
image

Could you at least create an item in the repository for all problems related to large stores with large volumes of data and existing patches and solutions that are not yet reported in the current version. This will prevent users returning to 1.6 or worse looking for  other e-commerce solutions.

arius28:
Just to add another voice to this issue since this affected a store of a client as well. We recently migrated a client's eshop from an old magento (1.7) to prestashop (1.7.7.1). This issue caused quite some trouble until we stumbled on this pull request. We were also getting close to over 20 second delays while going to checkout as a guest. We have around 13.000 customer accounts and a bit more addresses, not even close to @ComonSoft's but still enough to have the same problem.

If I am seeing right, this has not been included in any of the 1.7.7.x versions so far since the commit was merged. Shouldn't this have been included in the next versions after it was accepted, QA'ed and merged? 

As a final note, thank you for all the work you & the contributors put into this open-source software.

matks:
Hello @arius28 thank you for your nice feedback, it's great to have constructive opinion from PrestaShop user 👍 

> If I am seeing right, this has not been included in any of the 1.7.7.x versions so far since the commit was merged. Shouldn't this have been included in the next versions after it was accepted, QA'ed and merged?

Yes, only regressions (= bugs observed on 1.7.7 but not in 1.7.6 which is previous version) are merged into 1.7.7.x . Basically, 1.7.7 patches aim to fix has been mistakenly broken in 1.7.7.0 and only this. This rule allows us to deliver small and robust patches on a regular basis (every 6 weeks, see https://devdocs.prestashop.com/1.7/project/release/patch-release-lifecycle/). In the past we tried to ship more bugfixes into a patch release but that produced instable and bigger patches, longer to build, test, deliver.

This is why this bugfix was delivered into 1.7.8 scope. That being said, you are free to use this fix on your shop while waiting for it to be distributed by upstream 😊  .

> As a final note, thank you for all the work you & the contributors put into this open-source software.

Thank you too! 👍

arius28:
Hi @matks, thanks for the response and the explanation.

I have already implemented this fix and the order page response times are now fine 👍.

Issue #11722, opened 2018-12-11
Discount is not computed correctly with free gift


idnovate:
**Describe the bug**
When 2 cart rules with a free gift are applied, discount is not computed correctly in cart summary

**To Reproduce**

1. Create 1 cart rule with a free gift
1. Create a 2nd rule with the same configuration than before
1. Check cart summary

**Screenshots**
Cart rule example
![Cart rule example](http://imagizer.imageshack.com/img922/3585/6F2fbI.png)

Cart problem
![Cart problem](http://imagizer.imageshack.com/img923/9499/iHFxr6.jpg)

**Additionnal information**
PrestaShop version: Reproduced in PS 1.7.4.4 and PS 1.7.5 RC1

khouloudbelguith:
Hi @idnovate,

Thanks for your report.
I manage to reproduce the issue with PS1.7.5.0rc1.
In the shopping cart page => NOK
![image](https://user-images.githubusercontent.com/16067358/49809815-27ba2b00-fd60-11e8-9a48-c709e75195d7.png)
In the order confirmation
![image](https://user-images.githubusercontent.com/16067358/49809863-47515380-fd60-11e8-9ddf-c632aa3a4807.png)
In the order detail BO => NOK
![image](https://user-images.githubusercontent.com/16067358/49809943-75369800-fd60-11e8-8560-4b94215c2c0b.png)

The discount should be 68,92€

In the invoice => NOK
![image](https://user-images.githubusercontent.com/16067358/49810004-939c9380-fd60-11e8-88be-cd92aba83407.png)
Thanks!

marionf:
Related to #9757

khouloudbelguith:
Hi,

This issue occurs also when we have two cart rules with free shipping vouchers as reported here #12024
![image](https://user-images.githubusercontent.com/16067358/50685793-e80ef800-101a-11e9-9773-23a46e18a85c.png)
Thanks!

webstartconsulting:
Hi!
Its an important bug, any solution?

khouloudbelguith:
Hi @webstartconsulting,

Sorry, not yet.

idnovate:
Hi!

Same problem when 2 cart rules with free gift is applied.

- Create 2 cart rules like this:

![image](https://user-images.githubusercontent.com/15104724/57214030-e107ee80-6fe8-11e9-9030-16cca52775f4.png)

- Shopping cart is €52.46 instead of €35.09:

[...]

Issue #15056, opened 2019-08-09
Faceted search wrong results


ghost:
Prestashop 1.7.6
PHP 7.2
Faceted module 3.2.0

![faceted](https://user-images.githubusercontent.com/8243273/62783364-8d5fe200-babc-11e9-904b-06d30cdf165a.png)

Module show Size 36 is available but after choosing product is only with size 39...

8ternity:
I was having issue with version 1.7.5.0 with Faceted Search. I was resolving it by purchase the module "[Amazzing filter Module](https://addons.prestashop.com/en/search-filters/18575-amazzing-filter.html)" and works really better ! Either if your native Prestahsop doesn't work. But you will need to disable or uninstall it after it. I know it's not related to resolving your issue, but after messing with Faceted Search on 1.7.5.0, 1.7.5.1 and 1.7.5.2 and having always an issue with it, i purchase a module which worth the money.

khouloudbelguith:
Hi @PrestaShark,

In fact, in your case, the quantity of the General Product > 0
![image](https://user-images.githubusercontent.com/16067358/62787126-f004ae00-babb-11e9-9aa9-6bc105a0661c.png)
But, in your case, only one combination is out of stock which is `36`

When you make a filter with the `36` filter => yes, we have a product with the size, you are searching with the filer size & not the availability
but we have another filter which i availability => check if the product available or in stock.
I attached a screen record
https://drive.google.com/file/d/1VwHjCatMsdNcFm2KGTaYrv-zyWibkeLE/view
Thanks to check & feedback.

ghost:
Hi.
Let me explain.

If someone search for exact size (no matter if its shoe or tshirt) then want to select wanted size and get matched products (in matter of UX need of use another filter like availability is highly unnecessary - searching within unavailble products is weird, and should be droped by the team working on UI/UX...)

So as customer i want to view shoes with size 42 then i want to click 42 and see all available shoes with this size. 

So i see (like in example) my size 36(1) click and get size 39 (as customer i can think that store is incompetent and leave it in future!). Customer is angry, admin is angry, both are wasting time.

It's obvious that the customer wants to look for only available products, so im treats my bug report very seriously - Sorry but it can't work like that anymore.

ghost:
@marionf @colinegin @PierreRambaud please take care about it. 

If Faceted Search need more attention and time please consider to Crowdfund it and do it right. 

Thanks.

ghost:
[...]

PR #16635, opened 2019-11-30
Fix discount calculation if two gift-cartrules exist for the same product.

Hlavtox:
| Questions     | Answers
| Branch?       | develop
| Description?  | This PR fixes a bug, where the discount was not calculated correctly if you specified two cart rules with the same gift product - as described in #11722.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes topwatcher issue #11722
| How to test?  | Make two identical cart rules, make them apply, see the wrong discount total.

**A fix was two lines of code:**
- The logic got the unit price of a product and added it to total discounts.
- BUT, it was using getFinalUnitPrice instead of getInitialUnitPrice. 
- The problem?
- You got 2 gift products in cart, unit price 100.
- In the first pass, it discounted 100, but in the second pass, it discounted only 50.
- After this fix, it gets the original price before discounts. Thus fixing the issue.

**Screenshot - correct function**
Výstřižek

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16635)

### DIFF
--- a/src/Core/Cart/CartRuleCalculator.php
+++ b/src/Core/Cart/CartRuleCalculator.php
@@ -115,8 +115,8 @@ class CartRuleCalculator
                     && ($product['id_product_attribute'] == $cartRule->gift_product_attribute
                         || !(int) $cartRule->gift_product_attribute)
                 ) {
-                    $cartRuleData->addDiscountApplied($cartRow->getFinalUnitPrice());
-                    $cartRow->applyFlatDiscount($cartRow->getFinalUnitPrice());
+                    $cartRuleData->addDiscountApplied($cartRow->getInitialUnitPrice());
+                    $cartRow->applyFlatDiscount($cartRow->getInitialUnitPrice());
                 }
             }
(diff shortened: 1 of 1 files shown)


PierreRambaud (approved):

Robin-Fischer-PS:
Thanks @Hlavtox for this fix ! :smile: 

It's QA approved :heavy_check_mark: !

matks:
Thank you @Hlavtox this is the 2nd PR you got merged in 1.7.7.0. This means that when 1.7.7.0 the whole prestashop community will benefit from your help, and you will be able to to benefit from all the community's work as well. This is opensource in its beauty 😉

I am very happy to see your contributions, a lot of things have changed since the heated discussions on https://github.com/PrestaShop/PrestaShop/issues/15056 😛 

So, again: a big THANK YOU from all of prestashop's community and core team

Issue #13002, opened 2019-03-22
Customer can see all vouchers


Hlavtox:
EDIT: see reproductable scenario at https://github.com/PrestaShop/PrestaShop/issues/13002#issuecomment-476516323

EDIT2: this issue had too much history so we closed it and opened https://github.com/PrestaShop/PrestaShop/issues/15938 instead

**Description**
Customers can see all vouchers and cart rules you have created, even if they are generic.

You create a 'hidden' voucher with a 10% discount, to use it in some promotional campaigns, or to give them from a good will as a gift.

Now every customer that has registered can see this voucher in his account? 

**How to reproduce**
1. Create some voucher NOT limited to a specific customer. Create some vouchers limited to a specific customer.
2. Log-in to the customer's account in front office.
3. Go to 'My vouchers'.
4. See all vouchers on the store.

**Possible fix**
Commenting out $sql .= ' UNION (SELECT '.$sql_part1.' WHERE cr.`group_restriction` = 1 '.$sql_part2.')'; from CartRule.php seems to fix it, but I don't know what else it could break. Customer can then see only his vouchers, the group resctrictions work as they should.

**Additionnal information**
PrestaShop version: 1.6.1.23
PHP version: 7.2

khouloudbelguith:
Hi @Hlavtox,

I manage to reproduce the issue with PS:1.7.5.1.
If you have already fixed it on your end or if you think you can do it, please do send us a pull request!
Ping @marionf what do you think? 

Thanks!

marionf:
@khouloudbelguith 

I can't reproduce the issue.

I created a voucher for every customers
I created a voucher for John Doe

![capture d'écran_1220](https://user-images.githubusercontent.com/13449658/54836242-1cae5980-4cc4-11e9-8906-acb144afdb80.png)

With John Doe account, I see only John Doe's voucher

![capture d'écran_1218](https://user-images.githubusercontent.com/13449658/54836260-26d05800-4cc4-11e9-88d4-71d79e2ff95b.png)

With another customer account, I don't see any voucher

![capture d'écran_1219](https://user-images.githubusercontent.com/13449658/54836276-2c2da280-4cc4-11e9-8a55-a1dd9ca26ef8.png)

khouloudbelguith:
@marionf 

Thanks to check the Customer group selection.

marionf:
@khouloudbelguith 

You mean, I need to select a specific customer group for the voucher instead of a specific customer ?

[...]

PR #16638, opened 2019-11-30
Fix logic and display of customer's cart rules

### LINKED ISSUE #10766: Logic for displaying some discounts in "My account -> Discounts" is broken
**Describe the bug**
Read below.

**To Reproduce**

1. Lets assume that you'll add some coupon code for your customers, free shipping for everyone within a valid code. Ok.
1. Add new voucher, with some generated code, free shipping, and - this is important - for Customer group only, Highlight option set to "No", other settings with default values.
1. Now login into your account on Front-Office, go to "Discounts" and... you'll see this code.

Is this the right behavior? In my opinion: no. On "Discounts" page we should have only have:

* active discounts for a specific customer
* active discounts with all rules valid for currenly logged in customer **AND**`Highlight` option set to "Yes".

Any scenario to prove my point? Lets imagine that you as a merchant is going to some event, you print "Welcome Vouchers", 100 generated codes added to PrestaShop back-office, free shipping, code, 1 available to use per customer, only for Customer group of course. Now ever customer who is logged in to his account will see those 100 discount codes on "My Account -> Discount"...

**Additionnal information**
PrestaShop version: 1.7.5.0
PHP version: N/A

Hlavtox:
| Questions     | Answers
| Branch?       | develop
| Description?  | This fixes a million-times reported issues of customers seeing cart rules they should not see.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #10766 #13002
| How to test?  | See below

**What's wrong**
There is a bug in getCustomerCartRules function. The function accidentaly includes ALL CART RULES THAT HAVE SOME KIND OF GROUP RESTRICTION.

If you set some kind of group restriction on a cart rule, it group_restriction in database to 1. Then it stores the rules in a separate table.

Now, no matter what parameters you specify when calling the function, this part of code gets done. Does the cart rule have any kind of group restriction? It gets included. **Even the generic ones, without a code.**
`$sql .= ' UNION (SELECT ' . $sql_part1 . ' WHERE cr.`group_restriction` = 1 ' . $sql_part2 . ')';`

Removing this part of code restores the correct function and logic.

The logic works fine, because it's done later in the code - near `// Remove cart rule that does not match the customer groups`

**Why it's important**
Imagine you have X of vouchers, that reduce amount of order by X percent. They are unlimited by all means and have a code. You have them ready to be told to customers if you need to persuade them about ordering. Now (if conditions are met), every customers sees all of that "hidden vouchers" and can use it. Do you realise how crucial it can be if some merchants don't know about it?

**How to test**
- Make a customer A. Customer group 3.
- Make a new customer group ID 4, for example "Business Customers". Do NOT assign the customer to it.
- Make a cart rule - limited to customer A, no group restrictions, voucher code ABCDE, discount order by 10 %.
- Make a cart rule - limited to groups 1,2,3 (not for 4), voucher code EFGHI, discount order by 10 %.
- Make a cart rule - no restrictions on customers or groups, voucher code JKLMN, discount order by 10 %.

**Results - before a fix**
ABCDE - Visible - OK
EFGHI - Visible - WRONG
JKLMN - Not visible - OK

**Results - before a fix**
ABCDE - Visible - OK
EFGHI - Not visible - OK
JKLMN - Not visible - OK

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16638)

### DIFF
--- a/classes/CartRule.php
+++ b/classes/CartRule.php
@@ -393,7 +393,6 @@ class CartRuleCore extends ObjectModel
         }
 
         $sql = '(SELECT SQL_NO_CACHE ' . $sql_part1 . ' WHERE cr.`id_customer` = ' . (int) $id_customer . ' ' . $sql_part2 . ')';
-        // $sql .= ' UNION (SELECT ' . $sql_part1 . ' WHERE cr.`group_restriction` = 1 ' . $sql_part2 . ')';
         if ($includeGeneric && (int) $id_customer != 0) {
             $sql .= ' UNION (SELECT ' . $sql_part1 . ' WHERE cr.`id_customer` = 0 ' . $sql_part2 . ')';
         }
--- a/classes/CartRule.php
+++ b/classes/CartRule.php
@@ -393,7 +393,7 @@ class CartRuleCore extends ObjectModel
         }
 
         $sql = '(SELECT SQL_NO_CACHE ' . $sql_part1 . ' WHERE cr.`id_customer` = ' . (int) $id_customer . ' ' . $sql_part2 . ')';
-        $sql .= ' UNION (SELECT ' . $sql_part1 . ' WHERE cr.`group_restriction` = 1 ' . $sql_part2 . ')';
+        // $sql .= ' UNION (SELECT ' . $sql_part1 . ' WHERE cr.`group_restriction` = 1 ' . $sql_part2 . ')';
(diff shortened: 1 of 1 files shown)


PierreRambaud (approved):

Progi1984:
Thanks @Hlavtox

PR #16566, opened 2019-11-27
Functional Tests - Add BO tests for generate invoice by date

nesrineabdmouleh:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Add Functional tests for generate invoice by date
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | TEST_PATH="functional/BO/orders/invoices/01_generateInvoiceByDate.js" URL_FO=yourShopURL DOWNLOAD_PATH=yourDownloadPath npm run specific-test

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16566)

PR #16647, opened 2019-12-02
Tests - Using pptruser to run tests with download

boubkerbribri:
| Questions     | Answers
| Branch?       | develop
| Description?  | Using pptruser to run tests with download
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | to be tested after merging #16566 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16647)

### DIFF
--- a/tests/puppeteer/.docker/Dockerfile
+++ b/tests/puppeteer/.docker/Dockerfile
@@ -22,4 +22,11 @@ WORKDIR /tests/
 COPY ["./", "./"]
 
 # Install dependencies
-RUN npm install
+RUN npm install \
+    && groupadd -r pptruser && useradd -r -g pptruser -G audio,video pptruser \
+    && mkdir -p /home/pptruser/Downloads \
+    && chown -R pptruser:pptruser /home/pptruser \
+    && chown -R pptruser:pptruser /tests
+
+# Run everything after as non-privileged user.
+USER pptruser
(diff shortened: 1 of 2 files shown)


tests/puppeteer/.docker/Dockerfile
```diff
@@ -22,4 +22,11 @@ WORKDIR /tests/
 COPY ["./", "./"]
 
 # Install dependencies
-RUN npm install
+RUN npm install \
+    && groupadd -r pptruser && useradd -r -g pptruser -G audio,video pptruser \
```
PierreRambaud, line 26:
```suggestion
    && groupadd -r pptruser && useradd -r -g pptruser -m -G audio,video pptruser
```

And no need mkdir & chown.

Plus, I don't think you need to change the `USER`

boubkerbribri: I will test like that.
This came from the official [DockerFile](https://github.com/puppeteer/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker)

PierreRambaud: Oh, funny they used this kind of behaviour.
Ignore my comment :D Always try to use what they show you in the official documentation

PierreRambaud (approved):

PR #16658, opened 2019-12-02
Use tab attribute as fallback when a module does not have category

### LINKED ISSUE #15298: Our Payment modules are being considered as non-compatible in Prestashop 1.7+

PierreRambaud:
| Questions     | Answers
| Branch?       | develop
| Description?  | Stop using the Addons Api to request module categories. We already have a yaml file with all information.
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15298
| How to test?  | Follow ticket instruction + nothing is broken in the MBO module.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16658)

### DIFF
--- a/src/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProvider.php
+++ b/src/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProvider.php
@@ -40,9 +40,6 @@ class CategoriesProvider
     const CATEGORY_THEME = 'theme_modules';
     const CATEGORY_THEME_NAME = 'Theme modules';
 
-    const CATEGORY_MY_MODULES = 'my_modules';
-    const CATEGORY_MY_MODULES_NAME = 'My Modules';
-
     /**
      * @var array
      */
@@ -84,10 +81,8 @@ class CategoriesProvider
             }
 
             // Clear custom categories if there is no module inside
-            foreach ([self::CATEGORY_THEME, self::CATEGORY_MY_MODULES] as $category) {
-                if (empty($categories['categories']->subMenu[$category]->modules)) {
-                    unset($categories['categories']->subMenu[$category]);
-                }
+            if (empty($categories['categories']->subMenu[self::CATEGORY_THEME]->modules)) {
+                unset($categories['categories']->subMenu[self::CATEGORY_THEME]);
             }
 
             $this->categories = $categories;
@@ -109,23 +104,6 @@ class CategoriesProvider
         $categories = [
             'categories' => $this->createMenuObject('categories', 'Categories'),
         ];
-
-        if (empty($categoriesListing)) {
-            $categories['categories']->subMenu[self::CATEGORY_THEME] = $this->createMenuObject(
-                self::CATEGORY_THEME,
-                self::CATEGORY_THEME_NAME,
-                [],
-                self::CATEGORY_THEME
-            );
-            $categories['categories']->subMenu[self::CATEGORY_MY_MODULES] = $this->createMenuObject(
-                self::CATEGORY_MY_MODULES,
-                self::CATEGORY_MY_MODULES_NAME,
-                [],
-                self::CATEGORY_MY_MODULES
-            );
-
-            return $categories;
-        }
 
         foreach ($categoriesListing as $category) {
             $categories['categories']->subMenu[$category->name] = $this->createMenuObject(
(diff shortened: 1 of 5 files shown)


matthieu-rolland (commented):
LGTM, but it seems that travis is not happy, not sure if it's about this PR or travis being annoying...

src/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProvider.php
```diff
@@ -197,14 +175,14 @@ public function getParentCategory($categoryName)
...
-     * @param null $tab
+     * @param string $tab
      *
-     * @return object
+     * @return stdClass
      */
-    private function createMenuObject($menu, $name, $moduleIds = [], $tab = null)
+    private function createMenuObject(string $menu, string $name, array $moduleIds = [], ?string $tab = null): stdClass
     {
         return (object) array(
```
jolelievre, line 165: 🤮 isn't it possible to create a class to manage this object?

jolelievre (commented):
Now that there are less dependencies in this class Maybe it could be good, and not too long, to add a few unit tests?

src/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProvider.php
PierreRambaud: Little bit overkill because it create and need an stdClass :/

PierreRambaud: But I'll add unit tests

PierreRambaud:
@jolelievre unit tests added

src/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProvider.php
jolelievre: You're right it's probably overkill, it just feel weird to use and object, if we don't use strong typed object I guess just using an array would do the job no? This would avoid weird syntaxes like `$categories['categories']->subMenu[self::CATEGORY_THEME]->modules` and the class could use `$categories = json_decode(json_encode($categories));` right before the return (since it's already present in the class)

tests/Unit/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProviderTest.php
```diff
@@ -0,0 +1,474 @@
...
+{
+    /**
+     * @var CategoriesProvider
+     */
+    private $provider;
+
+    public function setUp()
+    {
+        $yamlParser = new YamlParser(_PS_CACHE_DIR_);
+        $prestashopAddonsConfig = $yamlParser->parse(_PS_ROOT_DIR_ . '/app/config/addons/categories.yml');
```
jolelievre: You should copy this yml into the test resources, if we change this file for functional needs then the test will fail without valid reason

tests/Unit/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProviderTest.php
```diff
@@ -0,0 +1,474 @@
...
+
+    public function testGetParentCategoryWithoutResult()
+    {
+        $this->assertEquals('Test', $this->provider->getParentCategory('Test'));
+    }
+
+    public function testGetParentCategory()
+    {
+        $this->assertEquals('Administration', $this->provider->getParentCategory('Registration & Ordering Process'));
+    }
```
jolelievre, line 485: Maybe you can add a test related to `CATEGORY_THEME` (when it's present and when it's unset)

src/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProvider.php
PierreRambaud: This was the original behaviour, and I can make a json_deciode / json_encode because we store `Module` objects into the `modules` key :sweat_smile: 
Playing with collection would have been much better :(

jolelievre (approved):
Thanks @PierreRambaud

src/PrestaShopBundle/Service/DataProvider/Admin/CategoriesProvider.php
jolelievre: Yeah that's why I suggested creating a real class, but I guess this will be an improvement for later

PR #16668, opened 2019-12-03
Integrate a Vue component to manage Currency customization in the BO

### LINKED ISSUE #14750: [CLDR] Edit a currency display by language (2/3)
**USER STORY**
As a merchant, I want to modify the currency format if needed in order to adapt my shop to international norms.

**WHY**
The CLDR data doesn't always answer all the merchant's needs, and we've seen lots of tickets asking for currency modifications (not available in 1.7 so far).

- Here the first part #10052 and the related PR #10455

**SPECIFICATION**
This new page lists of the currency display according to all the languages of the shop with the possibility to edit them.

Add/Edit currency display by language:
- The symbol
- The symbol position around the amount

**DETAILED SPECIFICATION**

[Format Setting](https://github.com/PrestaShop/prestashop-specs/blob/master/core/back-office/Localization-Currencies-List-Add-Edit.md#currency-format-setting)

the first part of the CLDR Front End development -> https://github.com/nihco2/PrestaShop/tree/feat/currencies-page-rework

**DESIGN**
Link to the Invision models:
https://invis.io/SAPYJZCCR6K

The popup will display "Apply" to update the format on the row of languages table instead of "Save".

**ISSUES RELATED**
- #9714 
- #10014
- #10017 
- #10055
- #10054
- #10053
- #9885
- #10021

**OLD SPECIFICATION**
- #10058
- #10057

Blocked by [#15752 ](https://github.com/PrestaShop/PrestaShop/issues/15752)
RAF : 0.5 day

jolelievre:
| Questions     | Answers
| Branch?       | develop
| Description?  | Integrate Vue component into Currency form edition to customize the format for each language
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14750
| How to test?  | Go to International > Localization > Currencies, edit one of your currencies You can change the display (by language) and change the symbol and its position Once the form is saved the modification should work instantly in the front (the cache has been cleared)<br />Then you try the reset button (in the customization list) and the Reset default button (for official currencies only) and check that the format is correctly reset<br />Also try it with "unusual" currency formats like for languages Dutch (with negative pattern) or Hindi (the group size is 2 not 3)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16668)


admin-dev/themes/new-theme/.webpack/common.js
```diff
@@ -109,6 +109,7 @@ module.exports = {
     alias: {
       vue$: 'vue/dist/vue.common.js',
       app: path.resolve(__dirname, '../js/app'),
+      psvue: path.resolve(__dirname, '../js/vue'),
```
PierreRambaud: Just notice that `app` is already for vue components / apps and json app.
You should move everything in this `app` directory

admin-dev/themes/new-theme/.webpack/common.js
```diff
@@ -121,7 +122,11 @@ module.exports = {
           options: {
             presets: [
               ['es2015', {modules: false}],
+              ["env", {
```
PierreRambaud: This is not a Json
```suggestion
              ['env', {'useBuiltIns': 'usage'}],
```

admin-dev/themes/new-theme/.webpack/common.js
```diff
@@ -121,7 +122,11 @@ module.exports = {
...
+                "useBuiltIns": "usage"
+              }]
             ],
+            "plugins": ["transform-runtime"]
```
PierreRambaud:
```suggestion
            'plugins': ['transform-runtime'],
```

admin-dev/themes/new-theme/.webpack/prod.js
```diff
@@ -30,6 +30,13 @@ function prodConfig(analyze) {
     })
   );
 
+  // Required for Vue production environment
```
PierreRambaud, line 33: Why it's required? We already have vue app not requiring it :thinking:

admin-dev/themes/new-theme/js/app/cldr/number-formatter.js
```diff
@@ -80,7 +80,7 @@ class NumberFormatter {
     }
 
     // Get the good CLDR formatting pattern. Sign is important here !
-    const pattern = this.getCldrPattern(majorDigits < 0);
+    const pattern = this.getCldrPattern(number < 0);
```
PierreRambaud, line 83: Could you create a pull request in faceted search too? :heart:

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatForm.vue
```diff
@@ -0,0 +1,111 @@
...
+    data: () => {
+      return {
+        value: {
+          symbol: '',
+          transformation: '',
+        },
+      };
+    },
+    props: {
+      language: Object
```
PierreRambaud:
```suggestion
      language: {
        type: Object,
        required: true,
        default: () => {},
      },
```

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatForm.vue
```diff
@@ -0,0 +1,111 @@
...
+      }
+    },
+    mounted() {
+      this.customSymbol = this.language.priceSpecification.currencySymbol;
+      const currencyPattern = this.language.priceSpecification.positivePattern;
+
+      // Detect which transformation matches the language pattern
+      for (let transformation in this.language.transformations) {
+        let transformationPatterns = this.language.transformations[transformation].split(';');
+        if (transformationPatterns[0] === currencyPattern) {
```
PierreRambaud, line 108: Why only the first one?

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -0,0 +1,107 @@
...
+* versions in the future. If you wish to customize PrestaShop for your
+* needs please refer to https://www.prestashop.com for more information.
+*
+* @author    PrestaShop SA <contact@prestashop.com>
+* @copyright 2007-2019 PrestaShop SA and Contributors
+* @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+* International Registered Trademark & Property of PrestaShop SA
+*-->
+<template>
+  <div :id="id" class="card-block row">
```
PierreRambaud, line 26: does it need an id?

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -0,0 +1,107 @@
...
+  </div>
+</template>
+
+<script>
+  import LanguageList from './LanguageList.vue';
+  import CurrencyModal from './CurrencyModal';
+
+  export default {
+    name: 'currency-formatter',
+    data: () => {
```
PierreRambaud: if you use this syntax  you should remove the return statement (eslint airbnb)

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyModal.vue
```diff
@@ -0,0 +1,79 @@
...
+    computed: {
+      modalTitle() {
+        return this.$t('Customize symbol and format') + (null !== this.language ? ' + ' + this.language.name : '');
+      }
+    },
+  };
+</script>
+
+<style lang="scss" scoped>
+  @import "../../../../../scss/config/_settings.scss";
```
PierreRambaud:
```suggestion
  @import '../../../../../scss/config/_settings.scss';
```

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
```diff
@@ -32,19 +38,58 @@ export default class CurrencyForm {
...
     this._initListeners();
     this._initFields();
+    this._initState();
+    this._initCurrencyFormatter();
+  }
+
+  _initState() {
+    this.state = {
+      currencyData: this._getCurrencyDataFromForm(),
+      languages: JSON.parse(JSON.stringify(this.originalLanguages)),
```
PierreRambaud: Why convert it to string, and again to json? :thinking:

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
```diff
@@ -32,19 +38,58 @@ export default class CurrencyForm {
...
+    this.currencyFormatter = new Vue({
+        el: this.map.currencyFormatter,
+        i18n: new VueI18n({
+          locale: 'en',
+          formatter: new ReplaceFormatter(),
+          messages: { en: this.translations }
+        }),
+        components: {CurrencyFormatter},
+        data: this.state,
+        template: `<currency-formatter id="${this.currencyFormatterId}" :languages="languages" :currencyData="currencyData"></currency-formatter>`
```
PierreRambaud, line 87: Not sure the #id is needed in the compoment

admin-dev/themes/new-theme/js/vue/components/Modal.vue
```diff
@@ -0,0 +1,150 @@
...
+ *-->
+<template>
+  <div>
+    <transition name="fade">
+      <div class="modal show">
+        <div class="modal-dialog modal-dialog-centered" role="document">
+          <div class="modal-content"
+               aria-labelledby="modalTitle"
+               aria-describedby="modalDescription"
+               v-click-outside="close"
```
PierreRambaud, line 33: Must be add on the `modal-dialog` instead, otherwise you'll not be able to click on the footer :)

admin-dev/themes/new-theme/js/vue/i18n/replace-formatter.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+/**
+ * This formatter is used by VueI18n, the basic format for variables looks
+ * like 'Hi {name}' or 'Hi {0}' Sadly it doesn't match the PrestaShop usual
+ * placeholders format.
+ * So this custom formatter allows us to simple replace in order to use formats
+ * like 'Hi %name%' the parameters then should be an object like {'%name%': 'John'}
+ */
+export default class ReplaceFormatter {
```
PierreRambaud: Still needed? Wasn't for translations?

src/Core/Localization/Currency/PatternTransformer.php
```diff
@@ -92,6 +92,29 @@ public function transform(string $currencyPattern, string $transformationType):
...
+     * @return string
+     */
+    public function getTransformationType(string $currencyPattern)
+    {
+        $patterns = explode(';', $currencyPattern);
+        $pattern = str_replace(self::RTL_CHARACTER, '', $patterns[0]);
+
+        if (preg_match("/^¤[ \u{a0}].+/", $pattern)) {
+            return self::TYPE_LEFT_SYMBOL_WITH_SPACE;
+        } elseif (preg_match('/^¤.+/', $pattern)) {
```
PierreRambaud: Why elseif when you have return in each if :D

src/Core/Localization/Currency/PatternTransformer.php
```diff
@@ -92,6 +92,29 @@ public function transform(string $currencyPattern, string $transformationType):
...
+     * @param string $currencyPattern
+     *
+     * @return string
+     */
+    public function getTransformationType(string $currencyPattern)
+    {
+        $patterns = explode(';', $currencyPattern);
+        $pattern = str_replace(self::RTL_CHARACTER, '', $patterns[0]);
+
+        if (preg_match("/^¤[ \u{a0}].+/", $pattern)) {
```
PierreRambaud: Afaik, you created a const for the `\u{a0}`

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -29,7 +29,21 @@
   {{ form_start(currencyForm, {
     'attr': {
       'id': 'currency_form',
-      'data-get-cldr-data': path('admin_currencies_get_reference_data', {'currencyIsoCode': 'CURRENCY_ISO_CODE'}),
+      'data-reference-url': path('admin_currencies_get_reference_data', {'currencyIsoCode': 'CURRENCY_ISO_CODE'})|replace({'/CURRENCY_ISO_CODE': '{/id}'}),
+      'data-languages': languages|json_encode|replace("'", "&#39;")|raw,
+      'data-translations': {
+        '1. Enter symbol': '1. Enter symbol'|trans({}, 'Admin.International.Feature'),
```
PierreRambaud: Because of vue js, you can do something more easier
```suggestion
        'form.symbol': '1. Enter symbol'|trans({}, 'Admin.International.Feature'),
```
Use simple keys

tests-legacy/Unit/Core/Localization/Specification/FactoryTest.php
```diff
@@ -86,6 +86,19 @@ public function getNumberData()
                     'groupingUsed' => true,
                     'primaryGroupSize' => 3,
                     'secondaryGroupSize' => 3,
+                    'numberSymbols' => [
+                        ",",
```
PierreRambaud:
```suggestion
                        ',',
```

PierreRambaud: and others of course

tests-legacy/Unit/Core/Localization/Specification/FactoryTest.php
```diff
@@ -208,6 +234,19 @@ public function getPriceData()
                     'secondaryGroupSize' => 3,
                     'currencyCode' => 'EUR',
                     'currencySymbol' => '€',
+                    'numberSymbols' => [
+                        ",",
```
PierreRambaud: same here and below

admin-dev/themes/new-theme/.webpack/common.js
jolelievre: Done

admin-dev/themes/new-theme/.webpack/common.js
jolelievre: Done

admin-dev/themes/new-theme/.webpack/common.js
jolelievre: Yes but the `js/app` is quite old now and it also contains a `pages` folder

But all the recently migrated pages have their javascript in the `js/pages`, I don't want start splitting the place where the code is now that 40+ pages are in this folder (against 4 in the app folder..) Another reason is that the pages in the app folder are full-Vue pages (one Vue app to rule all the page) whereas the currency page is a mix between vanilla management of the form and one component in Vue

Maybe we'd rather move the content of the app folder in the `js/pages` folder (some of these pages don't even use Vue), not now of course but someday

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -29,7 +29,21 @@
...
+        '2. Choose format': '2. Choose format'|trans({}, 'Admin.International.Feature'),
+        'Customize symbol and format': 'Customize symbol and format'|trans({}, 'Admin.International.Feature'),
+        'Apply': 'Apply'|trans({}, 'Admin.Actions'),
+        'Cancel': 'Cancel'|trans({}, 'Admin.Actions'),
+        'Close': 'Close'|trans({}, 'Admin.Actions'),
```
LouiseBonnard: Do we need to have two buttons for the same action? Cancel and close mean the same to me, I'd keep only the 'Cancel' button and delete the 'Close' action. :-)

admin-dev/themes/new-theme/.webpack/prod.js
jolelievre: Well maybe it's related the new webpack config I'm not sure but all I know is that without this when I build for production `npm run build` I still have this warning in the console:
```
You are running Vue in development mode.
Make sure to turn on production mode when deploying for production.
See more tips at https://vuejs.org/guide/deployment.html
```
And the same happens for other Vue app like the translation page, with this code no more warning. But maybe there is a simpler/cleaner way to set this in the config.
I saw something about adding `mode: 'prod'` but it seems it's only from the v3 of webpack, this one is the only one I could make work

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -29,7 +29,21 @@
...
+        'Language': 'Language'|trans({}, 'Admin.International.Feature'),
```
LouiseBonnard:
```suggestion
        'Language': 'Language'|trans({}, 'Admin.Global'),
```

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -29,7 +29,21 @@
...
+        'Example': 'Example'|trans({}, 'Admin.International.Feature'),
```
LouiseBonnard:
```suggestion
        'Example': 'Example'|trans({}, 'Admin.Global'),
```

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -29,7 +29,21 @@
...
+        'Edit symbol / format': 'Edit symbol / format'|trans({}, 'Admin.International.Feature'),
+        'Reset settings': 'Reset settings'|trans({}, 'Admin.International.Feature'),
```
LouiseBonnard:
```suggestion
        'Reset settings': 'Reset settings'|trans({}, 'Admin.Actions'),
```

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -29,7 +29,21 @@
...
+        'Price: %price% (Special discount %discount%)': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature')
```
LouiseBonnard: 🤔

admin-dev/themes/new-theme/js/app/cldr/number-formatter.js
jolelievre: Sure 😉
https://github.com/PrestaShop/ps_facetedsearch/pull/151

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatForm.vue
jolelievre: Done, although I think there is no need for a default value if the prop is required right?

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatForm.vue
jolelievre: The CLDR patterns are formatted this way `{positivePatter};{negativePattern}`
So the first one is always the positive one and I think matching the positive one is enough to check that the transformation matches
Besides in 99% of the languages there isn't even a negative pattern (only found one for two languages I believe)

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
jolelievre: It's not required but I guess it will be much easier for automated test if they have an id to target no?
I could leave it optional but I thought at least this way we are sure to always have an id for test purpose

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
jolelievre: I wanted to but couldn't find the correct syntax I tried `data: () => {{selectedLanguage: null}},` or `data: () => {selectedLanguage: null},` but it didn't work Are you sure the short syntax for objects?
I wonder if it only works for literal values (string number, boolean)

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyModal.vue
jolelievre: Done

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
jolelievre: It's actually to copy the Object, do you know another way to do it?

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
jolelievre: Same answer as above it's for automated tests

admin-dev/themes/new-theme/js/vue/components/Modal.vue
jolelievre: No it's fine I tested it `modal-content` contains the header+body+footer it's perfect

`modal-dialog` is actually a div with big margin to center the modal, but since it contains the whole screen (including the modal content) everything becomes "out of click" so it didn't work

admin-dev/themes/new-theme/js/vue/i18n/replace-formatter.js
jolelievre: No it's actually for VueI18n as explained in the description the default format for VueI18n is not compatible with our translation keys So I added this formatter to allow simple "str_replace like" format so we can use the same format everywhere (stuff like `Hello %name%` `click on [1]this link[/1]`)

src/Core/Localization/Currency/PatternTransformer.php
jolelievre: Mmm indeed but it means I need to build the regexp with stuff like 
`$regexp = '/^' . self::CURRENCY_SYMBOL . '[' . self::REGULAR_SPACE . self::NO_BREAK_SPACE . '].+/'`

I think it becomes quite hard to read, wdyt?

src/Core/Localization/Currency/PatternTransformer.php
jolelievre: Right 😅

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
jolelievre: Well everywhere in PS we use these kind of keys.. Not a big fan but it works this way Or do you suggest I just change the javascript keys?

tests-legacy/Unit/Core/Localization/Specification/FactoryTest.php
jolelievre: But cs-fix sis happy with it 😅
Ok it's not enabled on this folder but still!!!
(I'll change it 😉)

tests-legacy/Unit/Core/Localization/Specification/FactoryTest.php
jolelievre: Done

admin-dev/themes/new-theme/.webpack/common.js
PierreRambaud: we should clean up these directories.
Having `js/app` storing all our BO apps, and `js/vue` storing all our vue apps / globals `components`, shared `store`, etc wdyt?

admin-dev/themes/new-theme/.webpack/prod.js
PierreRambaud: Better editing the `npm run build` command in the package.json. It's not a good idea to force an env variable in the code
```
"build": "NODE_ENV=production webpack --bail --progress --mode production",
```

src/Core/Localization/Currency/PatternTransformer.php
PierreRambaud: For someone who don't know what `\u{a0}` is, it's pretty eassier to see `self::NO_BREAK_SPACE` :D

admin-dev/themes/new-theme/.webpack/common.js
jolelievre: yep I agree with the `js/vue` folder but I would move the `js/app` into the `js/pages` folder which already has so many pages And move the component already present in `js/app` into `js/vue`

admin-dev/themes/new-theme/.webpack/prod.js
jolelievre: I agree but the problem is that this variable is alredy present in the package.json...
I don't know why it doesn't work.. It seems webpack ignores the global env variables, or maybe is it Vue, I didn't understand either why it is ignored so since I lack knowledge about webpack I stopped looking for the problem and focused on finding a (not ideal) solution 😅

src/Core/Localization/Currency/PatternTransformer.php
jolelievre: Yeah I guess.. 😅

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
jolelievre: Cancel is used for confirmation modal When you have two possibilities Cancel/Apply, and close for simple modal (like Alert modals)
In this cas I only use the confirmation modal, so I don't need close but since the Modal component is generic it also manages the Close label and I had a warning in the console

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
jolelievre: Done

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
jolelievre: Done

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
jolelievre: What's the problem? Is it the placeholders? Maybe use some `[1]` instead? I'm not sure about the usual convetion but I thought I had already some `%placeholder%` sometimes

admin-dev/themes/new-theme/.webpack/prod.js
PierreRambaud: It's weird if you have the problem and you're the only one with it :sweat_smile:  I tried and can't reproduce :/

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatForm.vue
PierreRambaud: Yes, the usual practice to do it with Object -_-

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
PierreRambaud: `data: () => ({selectedLanguage: null}),` is the solution :)

PierreRambaud: Otherwise 
```
data() {
  return {};
}
```

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
PierreRambaud: Forget automated tests :sweat_smile:

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
PierreRambaud: `Object.assign({}, this.originalLanguages);` is maybe enough here :thinking:

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
PierreRambaud: The difference is in PS we use the original english string, which is not the same in Vue apps :)

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
LouiseBonnard: Sorry, I should have explained: I don't understand why this price and discount wording appears at this point. It doesn't seem related to the currency block to me, does it?

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
jolelievre: Nice, I didn't think about it

jolelievre: Ohh, except it doesn't work here because it is an array of  objects, but I think this is the solution:
```js
languages: [...this.originalLanguages],
```

admin-dev/themes/new-theme/.webpack/prod.js
jolelievre: Mmm, indeed it's weird Here's how I tested, I removed this piec of code, I ran `npm run build` and then went to a page with vue component (currency, translation, ...) And then I saw the warning `You are running Vue in development mode...` in the console
Did you perform the same test? What could be wrong with in my case? I have `npm 6.4.1` and `node v8.14.0`

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
jolelievre: Yep; `data: () => ({selectedLanguage: null})` works like a charm

src/Core/Localization/Currency/PatternTransformer.php
jolelievre: Done

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
jolelievre: It's actually wording that is used by a javascript Vue component, but I insert it via the html data tag using twig this way I can translate it with our usual `|trans` helper

In this Vue component we display a preview of how the currency is gonna be displayed on the front office, the initial design only show a price but I wanted to display negative values as well since they can have specific display in some languages, so that's why I added a little sentence with a positive price and a negative discount so that it's easier to understand for the user

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
jolelievre: Yep, I'll change it

jolelievre (commented):
Ok @PierreRambaud everything should be good now, except this compilation plugin that I need in production But if we can't find why I guess we can leave it to be sure no one else hase the same issue, wdyt?

admin-dev/themes/new-theme/.webpack/common.js
```diff
@@ -126,7 +127,11 @@ module.exports = {
           options: {
             presets: [
               ['es2015', {modules: false}],
+              ['env', {
```
PierreRambaud:
```suggestion
             ['env', {useBuiltIns: 'usage'}],
```

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyModal.vue
```diff
@@ -0,0 +1,79 @@
...
+    computed: {
+      modalTitle() {
+        return this.$t('modal.title') + (null !== this.language ? ' + ' + this.language.name : '');
+      }
+    },
+  };
+</script>
+
+<style lang="scss" scoped>
+  @import '../../../../../scss/config/_settings.scss';
```
PierreRambaud:
```suggestion
  @import '@scss/config/_settings.scss';
```

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -0,0 +1,103 @@
...
+        @close="closeModal"
+        @applyCustomization="applyCustomization"
+      >
+      </currency-modal>
+    </div>
+  </div>
+</template>
+
+<script>
+  import LanguageList from './LanguageList.vue';
```
PierreRambaud:
```suggestion
  import LanguageList from './LanguageList';
```

admin-dev/themes/new-theme/js/pages/currency/form/index.js
```diff
@@ -23,8 +23,8 @@
  * International Registered Trademark & Property of PrestaShop SA
  */
 
-import ChoiceTree from '../../../components/form/choice-tree';
-import TranslatableInput from '../../../components/translatable-input';
+import ChoiceTree from '@components/form/choice-tree';
+import TranslatableInput from '@components/translatable-input';
 import currencyFormMap from "./currency-form-map";
```
PierreRambaud:
```suggestion
import currencyFormMap from './currency-form-map';
```

admin-dev/themes/new-theme/.webpack/common.js
jolelievre: Done

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
jolelievre: Done

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyModal.vue
jolelievre: Done but ti requires a `~` at the beginning to work

admin-dev/themes/new-theme/js/pages/currency/form/index.js
jolelievre: Done

jolelievre (commented):
@PierreRambaud all is fixed? Ready for QA?

admin-dev/themes/new-theme/.webpack/prod.js
NeOMakinG: You can use another way to achieve that, as Pierre said, --mode option can be retrieved in webpack.config.js if you want : https://github.com/webpack/webpack/issues/7074#issuecomment-399853701

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -0,0 +1,103 @@
...
+        this.currencyData.transformations[language.id] = '';
+        this.currencyData.symbols[language.id] = language.currencySymbol;
+      },
+      applyCustomization(customData) {
+        const selectedPattern = this.selectedLanguage.transformations[customData.transformation];
+        const patterns = selectedPattern.split(';');
+
+        this.selectedLanguage.priceSpecification.currencySymbol = customData.symbol;
+        this.selectedLanguage.priceSpecification.positivePattern = patterns[0];
+        this.selectedLanguage.priceSpecification.negativePattern = patterns.length > 1 ? patterns[1] : '-' + patterns[0];
```
NeOMakinG, line 94: I don't rly know what you did, but just a question, patterns cant be a simple string ? then patterns[0] will be the first letter and patterns[1] the second letter if it's not separated by `;` character, I may be totally wrong not understanding

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyModal.vue
```diff
@@ -0,0 +1,80 @@
...
+    props: {
+      language: {
+        type: Object,
+        required: false,
+        default: null
+      }
+    },
+    computed: {
+      modalTitle() {
+        return this.$t('modal.title') + (null !== this.language ? ' + ' + this.language.name : '');
```
NeOMakinG: Can language be null ? since it has default props value?

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
```diff
@@ -92,40 +138,96 @@ export default class CurrencyForm {
     }
   }
 
-  _onResetDefaultSettingsClick() {
-    this._resetCurrencyData(this.$isoCodeInput.val());
+  async _onResetDefaultSettingsClick() {
+    await this._resetCurrencyData(this.$isoCodeInput.val());
```
NeOMakinG, line 142: Depending of our choices about ESLint, this function will return an error about his name containing an underscore

admin-dev/themes/new-theme/.webpack/prod.js
PierreRambaud: `--mode` is not available for webpack 2 :sob:

jolelievre: `--mode` doesn't work with this version of webpack

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
jolelievre: No because patterns is a split return `const patterns = selectedPattern.split(';');` So it will always be an array even if it contains only one string

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
NeOMakinG: yeah, spread operator will populate languages with this.originalLanguages props

jolelievre: But not a reference to `this.originalLanguages` right? it's a copy?

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
jolelievre: Probably but it's not defined yet so not relevant for this PR ^^

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyModal.vue
jolelievre: Sure it can be `null` this is actually its default value, and what defined if the modal is displayed or not

PierreRambaud: Could you avoid the  `+` syntax for concatenation please?

jolelievre: @PierreRambaud Done

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -0,0 +1,103 @@
...
+<script>
+  import LanguageList from './LanguageList';
+  import CurrencyModal from './CurrencyModal';
+
+  export default {
+    name: 'currency-formatter',
+    data: () => ({selectedLanguage: null}),
+    props: {
+      id: {
+        type: String,
```
PierreRambaud:
```suggestion
        type: Number,
```
Shoudn't be a Number :thinking:

PierreRambaud (dismissed):

sarahdib:
@TristanLDD is it possible to add a message when we reset symbol and format. It's not very obvious that the symbol and format is reseted.

jolelievre:
@sarahdib If it's the only thing to improve we can deal with it in another PR and merge this one for now. Unless you found some bugs?

sarahdib:
@jolelievre I didn't found functional bug. only some UX improvement

TristanLDD:
@sarahdib Thanks for the feedback, we can indeed add a confirmation popup to reduce the risk of misaction. @jolelievre Do we do this in a separated PR as you said?

jolelievre:
If everything is ok, then yes it's preferable to merge this one and make a new one This will be faster to review and test.

sarahdib:
@TristanLDD thanks 

@jolelievre let's merge the improvement is created : #17024

matthieu-rolland (approved):
let's merge this 👍

PR #16696, opened 2019-12-05
Fix email theme permission

### LINKED ISSUE #16337: Email Theme permissions settings are not saved when trying to enable them
**Describe the bug**
It is impossible to re-enable the permissions settings for the Email Theme page **once they have been disabled**.
The modification does not seem to be saved when enabling the permissions.

When I go to the Permissions page, the Email Theme page permissions (first email theme row, the parent one) are unchecked, I checked them, See a confirmation message, but when I refresh the boxes are unchecked again...

See video:
https://drive.google.com/file/d/1aU2dbka6r1mFvhQpOYjaZg9C3pXz_nNg/view

⚠️**This means that any new profile created won't have access to Email Theme, as per default all permissions are unchecked when creating a new profile
SuperAdmin is the only account to have access to email Theme**

**To Reproduce**
1. Go to 'Advanced Parameters > Team > Permissions' and select a profile you can edit
2. Unchecked all the Email Theme permissions (2 rows) -> See confirmation message
3. Refresh the page, Check all boxes -> See confirmation message
4. Refresh the page, scroll to the Email Theme rows, the first one is unboxed
5. You can try to check this specific row's boxes again, but this time there is no confirmation message and after refreshing the boxes are still unchecked

**Additional information**
PrestaShop version: 1.7.6.1

atomiix:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Add missing AdminMailTheme parent tab rights
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | Fixes #16337
| How to test?  | See below

## How to test?
There is two ways to test that PR.

1. After a fresh install on branch 1.7.6.x
2. After an upgrade from 1.7.6.2 to 1.7.6.x

Either case, you should be able to change _Email Themes_ permissions in _configure > Advanced parameters > Team > Permissions_

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16696)


install-dev/upgrade/php/copy_tab_rights.php
```diff
@@ -0,0 +1,58 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
PierreRambaud:
```suggestion
```

install-dev/upgrade/php/copy_tab_rights.php
```diff
@@ -0,0 +1,58 @@
...
+function copy_tab_rights($fromTabName, $toTabName)
+{
+    foreach (array(PageVoter::CREATE, PageVoter::READ, PageVoter::UPDATE, PageVoter::DELETE) as $role) {
+        // 1- Add role
+        $roleToAdd = strtoupper('ROLE_MOD_TAB_'.$toTabName.'_'.$role);
+        Db::getInstance()->execute('INSERT IGNORE INTO `'._DB_PREFIX_.'authorization_role` (`slug`)
+            VALUES ("'.pSQL($roleToAdd).'")');
+        $newID = Db::getInstance()->Insert_ID();
+        if (!$newID) {
```
PierreRambaud, line 40:
```suggestion
        if ($newID) {
            continue;
        }
```

install-dev/upgrade/php/copy_tab_rights.php
```diff
@@ -0,0 +1,58 @@
...
+            ');
+        }
+
+        // 2- Copy access
+        if (!empty($fromTabName) && !empty($newID)) {
+            $parentRole = strtoupper('ROLE_MOD_TAB_'.pSQL($fromTabName).'_'.$role);
+            Db::getInstance()->execute(
+                'INSERT INTO `'._DB_PREFIX_.'access` (`id_profile`, `id_authorization_role`)
+                SELECT a.`id_profile`, '. (int)$newID .' as `id_authorization_role`
+                FROM `'._DB_PREFIX_.'access` a join `'._DB_PREFIX_.'authorization_role` ar on a.`id_authorization_role` = ar.`id_authorization_role`
```
PierreRambaud: You need to run cs fixer :)

install-dev/upgrade/php/move_tab.php
```diff
@@ -0,0 +1,36 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+function move_tab($className, $toParentClassName)
```
PierreRambaud: Seems overkill to create a new file here when you can use anonymoius functions in the original script

install-dev/upgrade/php/copy_tab_rights.php
atomiix: Shouldn't we trying to get an existing value instead of just skiping it?

PierreRambaud: Seen with you on slack. Keep the original condition but move the `empty($fromTabName)`

install-dev/upgrade/php/copy_tab_rights.php
```diff
@@ -0,0 +1,59 @@
...
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+function copy_tab_rights($fromTabName, $toTabName)
+{
+    if (empty($fromTabName) || empty($toTabName)) {
+        return;
+    }
+    foreach (array(PageVoter::CREATE, PageVoter::READ, PageVoter::UPDATE, PageVoter::DELETE) as $role) {
+        // 1- Add role
+        $roleToAdd = strtoupper('ROLE_MOD_TAB_' . $toTabName . '_' . $role);
```
jolelievre, line 36: You should try to get the id BEFORE you try to create it Because if the role is already in database since there's a unicity constraint on the roles it could cause an error

install-dev/data/xml/tab.xml
```diff
@@ -205,7 +205,7 @@
                     <class_name>AdminThemesCatalog</class_name>
                 </tab>
                 <tab id="Mail_Theme_Parent" id_parent="Look_feel" active="1" enabled="1" hide_host_mode="0" icon="">
-                  <class_name>AdminMailThemeParent</class_name>
+                  <class_name>AdminParentMailTheme</class_name>
```
jolelievre, line 209: Although I agree this name is better Now that it's been introduced in 176 you can't modify it any more, it creates even more risk of losing some permissions along the way

install-dev/upgrade/php/copy_tab_rights.php
```diff
@@ -0,0 +1,59 @@
...
+                SELECT `id_authorization_role`
+                FROM `' . _DB_PREFIX_ . 'authorization_role`
+                WHERE `slug` = "' . pSQL($roleToAdd) . '"
+            ');
+        }
+
+        // 2- Copy access
+        if (!empty($newID)) {
+            $parentRole = strtoupper('ROLE_MOD_TAB_' . pSQL($fromTabName) . '_' . $role);
+            Db::getInstance()->execute(
```
jolelievre, line 51: Same thing here always check if the role is present in database before inserting it So you need to first select parent accesses, and then add them one by one if they are not present

install-dev/upgrade/php/ps_1770_update_tabs.php
```diff
@@ -0,0 +1,52 @@
...
+    include_once 'add_new_tab.php';
+    include_once 'copy_tab_rights.php';
+
+    add_new_tab_17('AdminParentMailTheme', 'en:Email Themes', 0, false, 'AdminParentThemes');
+    Db::getInstance()->execute(
+        'UPDATE `' . _DB_PREFIX_ . 'tab` SET `active`= 1, `enabled`= 1, `position`= 2 WHERE `class_name` = "AdminParentMailTheme"'
+    );
+
+    // Move AdminMailTheme's parent from AdminMailThemeParent to AdminParentMailTheme
+    $toParentTabId = Db::getInstance()->getValue(
```
jolelievre: Ok I get it, you move children tabs from the former AdminMailThemeParent to AdminParentMailTheme
But if some modules already used AdminMailThemeParent as their parent tab then it will become messy, so we'd better keep the original ones and just fix the permissions

jolelievre (changes requested):
Some little adjustments to avoid losing permissions, or tabs
But one more important question: why does it target the 177? It should be a 176 fix for 1763 no?

jolelievre (approved):

PierreRambaud (approved):

Progi1984:
Thanks @atomiix

PR #16262, opened 2019-11-04
Migration of order view page printing functionality and layout changes

tomas862:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop 
| Description?  | When clicked "print" page display its version ready for print. There are some layout changes when print is being initialized.
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #15819 
| How to test?  | Try clicking print or ctrl + p while in this page **admin-dev/index.php/sell/orders/orders/1/view**. It should be similar to invision sketch of when page is printed. **Dont forget to rebuild assets**

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16262)

Issue #15819, opened 2019-10-04
[ORDER PAGE] Print Order - 2/11


MatShir:
Part of the order page migration and redesign project.

**User Storie**

As a merchant, when I check an order, I need to be able to print my order.

**Detailed specification**
Specs on the repo prestashop-specs
[Order Action Specs
](https://github.com/PrestaShop/prestashop-specs/blob/cc476f1fc0c71c63bd3ba4879131f5163fca05e7/Order%20Page%20View.md#i-action-panel)

matks:
Fixed by https://github.com/PrestaShop/PrestaShop/pull/17660/

PR #16700, opened 2019-12-05
Migration of order view page printing functionality and layout changes - part 2

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | As PR https://github.com/PrestaShop/PrestaShop/pull/16262/ stayed out-of-sync with develop for too long, I rebuilt it in this PR.  When clicked "print" page display its version ready for print. There are some layout changes when print is being initialized.
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #15819 
| How to test?  | Try clicking print or ctrl + p while in this page **admin-dev/index.php/sell/orders/orders/1/view**. It should be similar to invision sketch of when page is printed. **Dont forget to rebuild assets**

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16700)


src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
```diff
@@ -23,46 +23,90 @@
...
+{% set statusTitle %}
+  {{ 'Status'|trans({}, 'Admin.Global') }} ({{ orderForViewing.history.statuses|length }})
+{% endset %}
+
+{% set documentsTitle %}
+  {{ 'Documents'|trans({}, 'Admin.Orderscustomers.Feature') }} ({{ orderForViewing.documents.documents|length }})
+{% endset %}
+
+{% set carriersTitle %}
+  {{ 'Carriers'|trans({}, 'Admin.Orderscustomers.Feature') }} ({{ orderForViewing.shipping.carriers|length }})
```
LouiseBonnard:
```suggestion
  {{ 'Carriers'|trans({}, 'Admin.Shipping.Feature') }} ({{ orderForViewing.shipping.carriers|length }})
```

matks:
@MatShir Thanks to @NeOMakinG the "print" feature should be good now 😄 can you take a look ?

MatShir:
Let's try it :smile:

MatShir:
From a design, perspective is good for me, nice job.
[Orders • demo.pdf](https://github.com/PrestaShop/PrestaShop/files/3932522/Orders.demo.pdf)
@TristanLDD wdyt ?

@matks I noticed when I tested that some tabs are missing if you look from the screen above that the tab carriers are missing. Also that the payment tab goes on the second page, is it because of the hooks? 
To keep it as ISO, the buttons should be hidden as well :smile: 

It was tested with order 5 from the fixtures.

MatShir:
@TristanLDD building the asset is better
[Demo with mass product.pdf](https://github.com/PrestaShop/PrestaShop/files/3932774/Demo.with.mass.product.pdf)

@NeOMakinG for the adjustment.

admin-dev/themes/new-theme/template/controllers/orders/page_header_toolbar.tpl
```diff
@@ -37,3 +38,12 @@
     {$title}
   {/if}
 {/block}
+
+<div class="header-toolbar d-print-none">
```
jolelievre: I don't understand this new block? It wasn't there before and it's not used by any layout so what's its purpose?

admin-dev/themes/new-theme/scss/pages/_orders_view.scss
```diff
@@ -0,0 +1,72 @@
...
+      .print-title__id {
+        color: #8a8a8a;
+      }
+    }
+  }
+}
+
+.adminorders {
+  @media print {
+    .header-toolbar {
```
jolelievre, line 57: What's the use of this since all `.header-toolbar` seem to have the `d-print-none` class anyway

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
```diff
@@ -23,46 +23,90 @@
...
+        {{ merchantReturnsTitle }}
       </a>
     </li>
   </ul>
 
   <div class="tab-content">
-    <div class="tab-pane fade show active" id="historyTabContent" role="tabpanel" aria-labelledby="historyTab">
-      {% include '@PrestaShop/Admin/Sell/Order/Order/Blocks/View/history.html.twig' %}
+    <div class="tab-pane d-print-block fade show active" id="historyTabContent" role="tabpanel" aria-labelledby="historyTab">
+      {% embed '@PrestaShop/Admin/Sell/Order/Order/Blocks/View/details_card.html.twig' %}
```
jolelievre, line 72: This changes the display of the tabs right? not just for print view but in general (since it includes a card)
Is it what is expected in the design?

jolelievre: Unless the card is only visible for print, and in screen view it seems as if there is no change thanks to `card--details` ?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/view.html.twig
```diff
@@ -32,34 +32,48 @@
 
 
 {% block content %}
-  <div class="row">
-    <div class="col-md-12">
-      {% include '@PrestaShop/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig' %}
+  <div id="order-view-page">
```
jolelievre, line 35: Be careful with these modifications as they will conflict with https://github.com/PrestaShop/PrestaShop/pull/16144 if it is migrated sooner
This bring the question about how content from hooks will be interpreted during print.. Although you can't guess what they will print
Same question with the tabs from hooks https://github.com/PrestaShop/PrestaShop/pull/16144/files#diff-1a42492ba7a3b1fc9477d3ad77b8edb5R70
Which probably won't integrate card to optimize print display

admin-dev/themes/new-theme/scss/pages/_orders_view.scss
NeOMakinG: Probably not, we see the toolbar if we don't do this, I'll investigate a bit :D

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/view.html.twig
NeOMakinG: Well it will show the content of every tabs as it is doing with payment, shipping etc.. Should we force the display of default tabs and only these ?

admin-dev/themes/new-theme/scss/pages/_orders_view.scss
jolelievre: it just seems redundant

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
NeOMakinG: It just force the card to use `display: block` in print view so it take the full width space

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/view.html.twig
jolelievre: But I guess the card are used in order to have the title right before the content?
If so then I guess the tabs from hooks will look like:
```
Title 1
Title 2
Content 1
Content 2
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
jolelievre: So is the card really necessary to do that? If the aim is only to have a `display: block;` on tabs it could be managed by adding it on `.tab-pane` directly If it was as simple as that then we could also apply it for hooks tabs
But I have the feeling the card is used so that the tab title may be printed close to the tab content, just a guess though Although I get the idea what bothers me is that this solution is a bit complex and won't work for hooks tabs.. (and I hope it doesn't modify the screen display)

admin-dev/themes/new-theme/scss/components/_button.scss
```diff
@@ -1,6 +1,6 @@
 .btn {
   &.btn-action {
-    color: #576C72;
+    color: black;
```
NeOMakinG:
```suggestion
    color: #000;
```

Respecting bootstrap standards

admin-dev/themes/new-theme/scss/pages/_orders_view.scss
NeOMakinG: Ah no ok you didn't understand the purpose of that style, the header-toolbar doesnt have the d-print-none class because we need to see it on the first page to get the main title, but it is in `position: fixed` by default, we need to avoid this header bar to be seen on every print pages so we set it to `position: initial`

jolelievre: But yet in `page_header_toolbar.tpl` it has `d-print-none`

admin-dev/themes/new-theme/template/controllers/orders/page_header_toolbar.tpl
matks: It was a trick used to obtain this display:

However I reverted this trick (see https://github.com/PrestaShop/PrestaShop/pull/16560) as it was creating side-effects (see https://github.com/PrestaShop/PrestaShop/issues/16549)

matks:
@jolelievre @NeOMakinG It seems there are 2 types of questions without answers here:
- what is the `page_header_toolbar.tpl` file and is it useful
- should we display, in the "print" display, the data/widgets/tabs/blocks added by modules ?

I'll answer the 1st question by going deeper in the code because I know this topic. IMO I
it was a trick used to obtain this display:

However I reverted this trick (see https://github.com/PrestaShop/PrestaShop/pull/16560) as it was creating side-effects (see https://github.com/PrestaShop/PrestaShop/issues/16549). I think we can delete it but I need to check.

For 2nd question I would suggest NOT to show modules additional content as indeed it could break the display. Or maybe provide an option to choose whether or not to display them. However this is a question that needs to be answered by @MatShir

matks:
> * what is the `page_header_toolbar.tpl` file and is it useful
> * should we display, in the "print" display, the data/widgets/tabs/blocks added by modules ?
> 
> I'll answer the 1st question by going deeper in the code because I know this topic.

I confirm it's useless now. It was deleted by Valentin so we're OK on this.

colinegin:
Needs to be merged in the end after all other PRs.

PierreRambaud:
@matks You are trying to merge develop changes into the 1.7.7.x branch :sob:

NeOMakinG:
> @matks You are trying to merge develop changes into the 1.7.7.x branch 😭

No it's me and I reverted

matks:
PR rebuilt in https://github.com/PrestaShop/PrestaShop/pull/17617/ as we could not rebase it (too old)

PR #16717, opened 2019-12-05
Fix unassigned JS variable call in AdminOrdersController

didpoule:
| Questions     | Answers
| Branch?       | 1.7.6.x 
| Description?  | A JS error occures when function "updateCartProducts" is called from AdminOrdersController because "id_customization" is not assigned before usage. This has for consequence to make impossible to create an order from a cart in Back-Office
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | No
| Deprecations? | No
| Fixed ticket? | Didn't see, i experienced and fixed the issue myself
| How to test?  | Create a cart in Front Office then try to create an order from this cart in Back Office, you should have any error in console and validate form.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16717)

### DIFF
--- a/admin-dev/themes/default/template/controllers/orders/form.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/form.tpl
@@ -740,6 +740,7 @@
 		$.each(products, function() {
 			var id_product = Number(this.id_product);
 			var id_product_attribute = Number(this.id_product_attribute);
+			var id_customization = Number(this.id_customization);
 			cart_quantity[Number(this.id_product)+'_'+Number(this.id_product_attribute)+'_'+Number(this.id_customization)] = this.cart_quantity;
 			cart_content += '<tr><td><img src="'+this.image_link+'" title="'+this.name+'" /></td><td>'+this.name+'<br />'+this.attributes_small+'</td><td>'+this.reference+'</td><td><input type="text" rel="'+this.id_product+'_'+this.id_product_attribute+'" class="product_unit_price" value="' + this.numeric_price + '" /></td><td>';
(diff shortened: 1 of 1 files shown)


admin-dev/themes/default/template/controllers/orders/form.tpl
```diff
@@ -740,14 +740,15 @@
...
 			cart_content += (!this.id_customization ? '<input type="text" rel="'+this.id_product+'_'+this.id_product_attribute+'_'+(this.id_customization ? this.id_customization : 0)+'" class="cart_quantity" value="'+this.cart_quantity+'" />' : '');
 			cart_content += (!this.id_customization ? '<div class="input-group-btn"><a href="#" class="delete_product btn btn-default" rel="delete_'+this.id_product+'_'+this.id_product_attribute+'_'+(this.id_customization ? this.id_customization : 0)+'" ><i class="icon-remove text-danger"></i></a></div></div>' : '');
 			cart_content += '</td><td>' + formatCurrency(this.numeric_total, currency_format, currency_sign, currency_blank) + '</td></tr>';
 
-			if (this.id_customization && this.id_customization != 0)
+			if (id_customization && id_customization != 0)
```
NeOMakinG, line 751:
```suggestion
			if (id_customization && id_customization !== 0)
```

PierreRambaud (changes requested):
Waiting for change

matks:
As the "Create an order" page has been migrated with Symfony in 1.7.7.0 I think this bugfix is not needed anymore. The file being modified is deleted 😅 

I checked the bug report but it seems to be OK in PS 1.7.7.0

Thank you for your contribution, I will consequently close the PR 🙏

Issue #9586, opened 2018-08-22
[BOOM-5974] Free Shipping Line Removed After Free Gift Has Been Applied


prestonBot:
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-5974).

- _**Reporter:**_ Pentarou
- _**Created at:**_ Fri, 13 Jul 2018 09:49:42 +0200

<p>In the cart and during checkout, the free shipping line is removed after applying a second free gift cart rule. This is very confusing for the customer and will certainly reduce the number of sales.</p>
        
* How to reproduce the issue ?

Add cart rule free shipping from € 35,-
<br/>
Add cart rule free gift to the value of € 3,95,- with every order 
<br/>
Add cart rule free gift to the value of € 17,95,- for orders from € 75, -
<br/>

<br/>
1. Add a product to the shopping cart with a value of less than 35 euros to see the shipping costs.
<br/>

<br/>
2. Add a product to the shopping cart with a value over 35 euros to see the free shipping label.
<br/>

<br/>
3. Add a product to the shopping cart so that the total value is more than or equal to 75 euros to see the shipping costs again.

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-5974#comment-171586).

- _**Author:**_ khouloud.belguith
- _**Created at:**_ Fri, 13 Jul 2018 13:55:29 +0200

<p>Hi Ferry,</p>

<p>Thanks for your report.<br/>
I manage to reproduce this issue with PS1.7.4.0 / 1.7.4.1-rc-1.<br/>
I attached a screen record.<br/>
 <span class="nobr"><a href="http://forge.prestashop.com/secure/attachment/69445/69445_1740stable.webm" title="1740stable.webm attached to BOOM-5974">1740stable.webm<sup><img class="rendericon" src="http://forge.prestashop.com/images/icons/link_attachment_7.gif" height="7" width="7" align="absmiddle" alt="" border="0"/></sup></a></span> <br/>
We will see how to fix it.</p>

<p>Best regards, Khouloud</p>

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-5974#comment-171587).

- _**Author:**_ pentarou
- _**Created at:**_ Fri, 13 Jul 2018 14:10:51 +0200

[...]

PR #16724, opened 2019-12-06
Fix multiple cart rules calculation

### LINKED ISSUE #16673: The total is not well calculated with free gift
**Describe the bug**
After fixing this issue: https://github.com/PrestaShop/PrestaShop/issues/11722 in the develop branch.
2 uses cases are not ok in the develop branch as reported here: https://github.com/PrestaShop/PrestaShop/issues/11722#issuecomment-561220896
Issue 1:
Create the following rules:

- Rule 1
Amount discount:
image

- Rule 2
image

- Rule 3
Free gift
image

Cart:
image

Issue 2:
- Change the first rule:
image

- Discount is higher than the amount
image

In the PS1.7.6.2, the total is well calculated

**Additional information**
PrestaShop version: develop
PHP version: 7.1

atomiix:
| Questions     | Answers
| Branch?       | develop
| Description?  | This fix discounts calculation with multiple cart rules
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16673
| How to test?  | See #16673

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16724)

### DIFF
--- a/src/Core/Cart/Calculator.php
+++ b/src/Core/Cart/Calculator.php
@@ -150,10 +150,11 @@ class Calculator
 
         $amount = new AmountImmutable();
         foreach ($this->cartRows as $cartRow) {
-            $rowPrice = $cartRow->getFinalTotalPrice();
+            $rowPrice = $cartRow->getInitialTotalPrice();
             $amount = $amount->add($rowPrice);
         }
-        $shippingFees = $this->fees->getFinalShippingFees();
+        $amount = $amount->sub($this->getDiscountTotal());
+        $shippingFees = $this->fees->getInitialShippingFees();
         if (null !== $shippingFees) {
             $amount = $amount->add($shippingFees);
         }
@@ -185,11 +186,39 @@ class Calculator
      *
      * @throws \Exception
      */
+    public function getRowTotalWithoutDiscount()
+    {
+        $amount = new AmountImmutable();
+        foreach ($this->cartRows as $cartRow) {
+            $amount = $amount->add($cartRow->getInitialTotalPrice());
+        }
+
+        return $amount;
+    }
+
+    /**
+     * @return AmountImmutable
+     *
+     * @throws \Exception
+     */
     public function getDiscountTotal()
     {
         $amount = new AmountImmutable();
         foreach ($this->cartRules as $cartRule) {
             $amount = $amount->add($cartRule->getDiscountApplied());
+        }
+
+        $totalWithoutDiscount = $this->getRowTotalWithoutDiscount();
+        $initialShipping = $this->getFees()->getInitialShippingFees();
+        $finalShipping = $this->getFees()->getFinalShippingFees();
+
+        if (null !== $initialShipping && null !== $finalShipping) {
+            $totalWithoutDiscount = $totalWithoutDiscount->add($initialShipping);
+            $totalWithoutDiscount = $totalWithoutDiscount->sub($finalShipping);
+        }
+        // discount cannot be above total cart price
+        if ($totalWithoutDiscount < $amount) {
+            $amount = $totalWithoutDiscount;
         }
 
         return $amount;
--- a/src/Core/Cart/Calculator.php
+++ b/src/Core/Cart/Calculator.php
@@ -208,18 +208,18 @@ class Calculator
             $amount = $amount->add($cartRule->getDiscountApplied());
         }
 
-        $totalWithoutDiscount = $this->getRowTotalWithoutDiscount();
+        $allowedMaxDiscount = $this->getRowTotalWithoutDiscount();
 
         if (null !== $this->getFees()->getInitialShippingFees() && null !== $this->getFees()->getFinalShippingFees()) {
             $shippingDiscount = (new AmountImmutable())
                 ->add($this->getFees()->getInitialShippingFees())
                 ->sub($this->getFees()->getFinalShippingFees())
             ;
-            $totalWithoutDiscount = $totalWithoutDiscount->add($shippingDiscount);
+            $allowedMaxDiscount = $allowedMaxDiscount->add($shippingDiscount);
         }
         // discount cannot be above total cart price
-        if ($totalWithoutDiscount < $amount) {
-            $amount = $totalWithoutDiscount;
+        if ($amount > $allowedMaxDiscount) {
+            $amount = $allowedMaxDiscount;
         }
 
         return $amount;
--- a/src/Core/Cart/Calculator.php
+++ b/src/Core/Cart/Calculator.php
@@ -148,11 +148,7 @@ class Calculator
             throw new \Exception('Cart must be processed before getting its total');
         }
 
-        $amount = new AmountImmutable();
-        foreach ($this->cartRows as $cartRow) {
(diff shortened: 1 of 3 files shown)


src/Core/Cart/Calculator.php
```diff
@@ -192,6 +208,19 @@ public function getDiscountTotal()
             $amount = $amount->add($cartRule->getDiscountApplied());
         }
 
+        $totalWithoutDiscount = $this->getRowTotalWithoutDiscount();
+        $initialShipping = $this->getFees()->getInitialShippingFees();
+        $finalShipping = $this->getFees()->getFinalShippingFees();
+
+        if (null !== $initialShipping && null !== $finalShipping) {
```
jolelievre: I'm not sure I get this part Part of the shipping is added, the other one is subsided.. Which one is what? Maybe this justifies a comment

jolelievre: Since you compute the whole discount separately, is it still useful to `applyFlatDiscount` on rows during the computation process?

jolelievre (commented):
Is the issue https://github.com/PrestaShop/PrestaShop/issues/11722 still fixed with your modification? Is there a behat test dedicated to this previous issue?

src/Core/Cart/Calculator.php
atomiix: > I'm not sure I get this part Part of the shipping is added, the other one is subsided.. Which one is what? Maybe this justifies a comment

You're right, I'll add a comment explaining this part!

>Since you compute the whole discount separately, is it still useful to `applyFlatDiscount` on rows during the computation process?

I think it's not useful but I'm pretty sure that getting rid of it will introduce a BC break?

Edit: Actually, it's needed because otherwise you can't keep track of an already applied discount. Imagine 2 discounts of 50% each with 2 different priorities. If we do not keep track of the already applied discount the cart total will be 0€ instead of 50% of 50% of the amount.

atomiix:
> Is the issue #11722 still fixed with your modification? Is there a behat test dedicated to this previous issue?

Yes the issue is still fixed with my modifications. The behat tests I added handle that case too.

src/Core/Cart/Calculator.php
jolelievre: Ok for `applyFlatDiscount`

For the shipping part I didn't get it at first but with your comment it seems logical, but I think we could simplify this an remove the comment with this:
```php
if (null !== $initialShipping && null !== $finalShipping) {
    $shippingDiscount = (new AmountImmutable())->add($initialShipping)->sub($finalShipping);
    $totalWithoutDiscount = $totalWithoutDiscount->sub($shippingDiscount);
}
```
or to avoid the intermediate variables:
```php
if (null !== $this->getFees()->getInitialShippingFees() && null !== $this->getFees()->getFinalShippingFees()) {
    $shippingDiscount = (new AmountImmutable())
                         ->add($this->getFees()->getInitialShippingFees())
                         ->sub($this->getFees()->getFinalShippingFees())
    ;
    $totalWithoutDiscount = $totalWithoutDiscount->sub($shippingDiscount);
}
```
What do you think?

atomiix: Just updated the code 😉

src/Core/Cart/Calculator.php
```diff
@@ -192,6 +208,20 @@ public function getDiscountTotal()
...
         }
 
+        $totalWithoutDiscount = $this->getRowTotalWithoutDiscount();
+
+        if (null !== $this->getFees()->getInitialShippingFees() && null !== $this->getFees()->getFinalShippingFees()) {
+            $shippingDiscount = (new AmountImmutable())
+                ->add($this->getFees()->getInitialShippingFees())
+                ->sub($this->getFees()->getFinalShippingFees())
+            ;
+            $totalWithoutDiscount = $totalWithoutDiscount->add($shippingDiscount);
```
jolelievre: Shouldn't it be a substraction here since it's a discount?

jolelievre: Or maybe it should be added to `$amount` ? To have the addition of product discount plus shipping discount?
By the way does `$totalWithoutDiscount` already contains shipping fees?

atomiix: Actually it's the rest of the shipping discount so it should be added. Maybe I should rename the `$shippingDiscount`to better reflect what this variable is?

atomiix: As seen with you on Slack, I changed the variable name to better understand that part

src/Core/Cart/Calculator.php
```diff
@@ -150,10 +150,11 @@ public function getTotal($ignoreProcessedFlag = false)
 
         $amount = new AmountImmutable();
         foreach ($this->cartRows as $cartRow) {
-            $rowPrice = $cartRow->getFinalTotalPrice();
+            $rowPrice = $cartRow->getInitialTotalPrice();
```
jolelievre: I think you can use `getRowTotalWithoutDiscount` here instead of the loop

src/Core/Cart/Calculator.php
```diff
@@ -150,10 +150,11 @@ public function getTotal($ignoreProcessedFlag = false)
...
             $amount = $amount->add($rowPrice);
         }
-        $shippingFees = $this->fees->getFinalShippingFees();
+        $amount = $amount->sub($this->getDiscountTotal());
+        $shippingFees = $this->fees->getInitialShippingFees();
```
jolelievre, line 153: So if you add `getInitialShippingFees` when is the shipping discount applied? Also I don't know what wrapping fees are but should it be considered when computing the max authorized discount?

jolelievre: mmm, my bad it should already be inside the total discount! forget this comment

jolelievre (approved):
Nice job @atomiix

src/Core/Cart/Calculator.php
```diff
@@ -148,12 +148,9 @@ public function getTotal($ignoreProcessedFlag = false)
...
 
-        $amount = new AmountImmutable();
-        foreach ($this->cartRows as $cartRow) {
-            $rowPrice = $cartRow->getFinalTotalPrice();
-            $amount = $amount->add($rowPrice);
-        }
-        $shippingFees = $this->fees->getFinalShippingFees();
+        $amount = $this->getRowTotalWithoutDiscount();
+        $amount = $amount->sub($this->getDiscountTotal());
+        $shippingFees = $this->fees->getInitialShippingFees();
```
arouiadib, line 153: Hi Thomas, could you please explain  why using `getInitialShippingFees()` here instead of `getFinalShippingFees`? using  `getFinalShippingFees` makes avoiding [#9586 ](https://github.com/PrestaShop/PrestaShop/issues/9586)(at least).

khouloudbelguith:
Hi,

We have issues could be fixed with this PR:
-https://github.com/PrestaShop/PrestaShop/issues/16673#issuecomment-570246339
-https://github.com/PrestaShop/PrestaShop/issues/9586#issuecomment-571114824 (regression compared to PS1.7.6.2)

Thanks!

Progi1984:
Thanks @atomiix

PR #16731, opened 2019-12-06
Tests - Disable dev mode while testing with docker

boubkerbribri:
| Questions     | Answers
| Branch?       | develop
| Description?  | Disable dev mode while testing with docker
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | see how to test in Docker.md

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16731)

### DIFF
--- a/tests/puppeteer/.docker/prestashop/docker_run.sh
+++ b/tests/puppeteer/.docker/prestashop/docker_run.sh
@@ -24,13 +24,8 @@ if [ -f /var/www/html/config/settings.inc.php -o -f /var/www/html/app/config/par
   rm -rf /var/www/html/var/cache/*
 fi
 
-if [ $PS_DEV_MODE -ne 0 ]; then
-  echo "\n* Enabling DEV mode ...";
-  sed -ie "s/define('_PS_MODE_DEV_', false);/define('_PS_MODE_DEV_',\ true);/g" /var/www/html/config/defines.inc.php
-else
-  echo "\n* Disabling DEV mode ...";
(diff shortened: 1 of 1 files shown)


SimonGrn (dismissed):

tests/puppeteer/.docker/prestashop/docker_run.sh
```diff
@@ -27,6 +27,9 @@ fi
 if [ $PS_DEV_MODE -ne 0 ]; then
   echo "\n* Enabling DEV mode ...";
   sed -ie "s/define('_PS_MODE_DEV_', false);/define('_PS_MODE_DEV_',\ true);/g" /var/www/html/config/defines.inc.php
+else
+  echo "\n* Disabling DEV mode ...";
+  sed -ie "s/define('_PS_MODE_DEV_', true);/define('_PS_MODE_DEV_',\ false);/g" /var/www/html/config/defines.inc.php
```
PierreRambaud: Since this docker_run.sh is only for puppeteer, you can directly do 
`sed -ie "s/define('_PS_MODE_DEV_', true);/define('_PS_MODE_DEV_',\ false);/g" /var/www/html/config/defines.inc.php` and remove the condition about PS_DEV_MODE :)

boubkerbribri: Yes no need to :+1:

PierreRambaud (approved):

PR #16742, opened 2019-12-09
Can't export data more than filters limits

### LINKED ISSUE #16328: Customers export from the BO - Wrong 50 limit
Hello,

When you use the "export" button of the customer grid from the Backoffice:
image

The file is limited to 50 entries. It should extract all the Customers.

In the PrestaShopBundle\Controller\Admin\Sell\Customer\CustomerController::exportAction(CustomerFilters $filters)
There is no filter setting. So the default values of the Filter are the following:

    public static function getDefaults()
    {
        return [
            'limit' => 50,
            'offset' => 0,
            'orderBy' => 'date_add',
            'sortOrder' => 'DESC',
            'filters' => [],
        ];
    }

The issue can be handled adding: `$filters->set('limit', null);` at the top of the controller.

Best regards.

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Customers, manufacturers and Categories can't be export over the default limit.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16328 
| How to test?  | Follow ticket instruction, and repeat with Categories, manufacturers and customers.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16742)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
@@ -670,6 +670,7 @@ class CategoryController extends FrameworkBundleAdminController
      */
     public function exportAction(CategoryFilters $filters)
     {
+        $filters = new CategoryFilters(['limit' => null] + $filters->all());
         $categoriesGridFactory = $this->get('prestashop.core.grid.factory.category');
         $categoriesGrid = $categoriesGridFactory->getGrid($filters);
 
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php
@@ -399,6 +399,7 @@ class ManufacturerController extends FrameworkBundleAdminController
      */
     public function exportAction(ManufacturerFilters $filters)
     {
+        $filters = new ManufacturerFilters(['limit' => null] + $filters->all());
         $manufacturersGridFactory = $this->get('prestashop.core.grid.grid_factory.manufacturer');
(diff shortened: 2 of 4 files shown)


src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
```diff
@@ -664,12 +664,11 @@ public function deleteAction(Request $request)
...
      * @DemoRestricted(redirectRoute="admin_categories_index")
      *
-     * @param CategoryFilters $filters
-     *
      * @return Response
      */
-    public function exportAction(CategoryFilters $filters)
+    public function exportAction()
     {
+        $filters = new CategoryFilters(['limit' => null] + CategoryFilters::getDefaults());
```
jolelievre: By doing that you ignore all previously set filters in the grid, so the export action will simply return all the categories all the time (regardless of your current filtering) I'd recommend rebuild an object filter based on the provided values:
```php
$filters = new CategoryFilters($filters->all() + ['limit' => null], $filters->getFilterId());
```

jolelievre (changes requested):

Progi1984 (approved):

sarahdib:
@jolelievre it's ok for you ?

jolelievre (approved):

jolelievre:
@sarahdib yep all good

sarahdib:
@PierreRambaud It's fixed for all except for supplier

PierreRambaud:
@sarahdib The supplier has not been migrated yet :sweat_smile: This is why it's not working yet

Progi1984:
Thanks @PierreRambaud

PR #16765, opened 2019-12-11
Fix recursive check of updated files

Darhazer:
The return value of the recursive call is not used, so even if it
reports changed files in subfolders, they won't be returned at the end.

The URL to the xml file was incorrect, so it never got to the recursive
call in the first place.

| Questions     | Answers
| Branch?       | develop
| Description?  | In version 1.7 the RequiredFilesChecker is broken. See the PR description for details.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | No
| Deprecations? | No
| Fixed ticket? | N/A
| How to test?  | Modify a file in the core and go to Back Office -> Advanced-> System information. Your modified file should be on the list. Note: there is no XML file for the `develop` version, so you should cherry-pick this in an already released version to test (or just modify the AppKernel::VERSION constant to something released, e.g. `1.7.6.2`)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16765)

### DIFF
--- a/src/Adapter/Requirement/CheckMissingOrUpdatedFiles.php
+++ b/src/Adapter/Requirement/CheckMissingOrUpdatedFiles.php
@@ -49,7 +49,7 @@ class CheckMissingOrUpdatedFiles
         );
 
         if (null === $dir) {
-            $xml = @simplexml_load_file(_PS_API_URL_ . '/xml/md5-1' . AppKernel::MAJOR_VERSION . '/' . AppKernel::VERSION . '.xml');
+            $xml = @simplexml_load_file(_PS_API_URL_ . '/xml/md5-' . AppKernel::MAJOR_VERSION . '/' . AppKernel::VERSION . '.xml');
             if (!$xml) {
                 return $fileList;
             }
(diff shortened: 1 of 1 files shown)


PierreRambaud (approved):

sarahdib:
hello @Darhazer 

I just test the PR and the return value indicated that there is no file change even if I change a file.

Darhazer:
Hello @sarahdib

Did you make sure that `AppKernel::VERSION` points to a version that has an XML file with the checksums? As in dev mode / 1.7.7 you may not have a valid URL.

Unfortunately, there are a few things that might fail with loading the XML and there is no error message shown to the user. This is a separate improvement that could be made, but a decision regarding the UI and the wording should be made.

So to test this PR, please make sure it loads the xml file successfully and that the modified file is present in the list of files in the XML.

sarahdib:
@Darhazer thank you for your PR it's ok and thanks to @Progi1984 who help me with the test

Progi1984:
Thanks @Darhazer for your first contribution :tada:

PR #16757, opened 2019-12-10
Add behat tests for miscellaneous command/handlers: SwitchDebugMode, CloseShowcaseCard, GetShowcaseCardIsClosed, GenerateThemeMailTemplates, AddEmployee

tdavidsonas88:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | cover all CommandHandlers and QueryHandlers by behavior (behat) tests of the group MISC related to #14480
| Type?         | improvement
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/14480
| How to test?  | automated tests should pass

Covered handlers:

 Domain/Configuration/CommandHandler/SwitchDebugModeHandler
 Domain/ShowcaseCard/CommandHandler/CloseShowcaseCardHandler
 Domain/ShowcaseCard/QueryHandler/GetShowcaseCardIsClosedHandler
 Domain/MailTemplate/CommandHandler/GenerateThemeMailTemplatesCommandHandler
 Domain/Employee/CommandHandler/AddEmployeeHandler

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16757)

PR #16776, opened 2019-12-11
Cover Contact CommandHandlers and QueryHandlers by behat tests

tdavidsonas88:
| Questions     | Answers
| Branch?       | develop
| Description?  | Cover CommandHandlers and QueryHandlers by behat tests #16757
| Type?         | improvement
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/14480
| How to test?  | Travis automated tests passes in green

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16776)

 Domain/Contact/CommandHandler/AddContactHandler
 Domain/Contact/CommandHandler/EditContactHandler
 Domain/Contact/QueryHandler/GetContactForEditingHandler

### DIFF
--- a/src/Adapter/Contact/CommandHandler/AddContactHandler.php
+++ b/src/Adapter/Contact/CommandHandler/AddContactHandler.php
@@ -37,9 +37,6 @@ use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactConstraintExcepti
 use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactException;
 use PrestaShop\PrestaShop\Core\Domain\Contact\ValueObject\ContactId;
 use PrestaShopException;
-use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
-use Symfony\Component\Validator\Exception\InvalidOptionsException;
-use Symfony\Component\Validator\Exception\MissingOptionsException;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 /**
@@ -64,17 +61,6 @@ final class AddContactHandler extends AbstractObjectModelHandler implements AddC
 
     /**
      * {@inheritdoc}
-     *
-     * @param AddContactCommand $command
-     *
-     * @return ContactId
-     *
-     * @throws CannotAddContactException
-     * @throws ConstraintDefinitionException
-     * @throws ContactConstraintException
-     * @throws ContactException
-     * @throws InvalidOptionsException
-     * @throws MissingOptionsException
      */
     public function handle(AddContactCommand $command)
     {
@@ -120,9 +106,6 @@ final class AddContactHandler extends AbstractObjectModelHandler implements AddC
      * @param array $localisedTitle
      *
      * @throws ContactConstraintException
-     * @throws ConstraintDefinitionException
-     * @throws InvalidOptionsException
-     * @throws MissingOptionsException
      */
     private function assertLocalisedTitleContainsDefaultLanguage(array $localisedTitle)
     {
--- a/src/Adapter/Contact/CommandHandler/AddContactHandler.php
+++ b/src/Adapter/Contact/CommandHandler/AddContactHandler.php
@@ -61,6 +61,9 @@ final class AddContactHandler extends AbstractObjectModelHandler implements AddC
 
     /**
      * {@inheritdoc}
+     *
+     * @throws CannotAddContactException
+     * @throws ContactException
      */
     public function handle(AddContactCommand $command)
     {
--- a/src/Adapter/Contact/CommandHandler/AddContactHandler.php
+++ b/src/Adapter/Contact/CommandHandler/AddContactHandler.php
@@ -37,6 +37,9 @@ use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactConstraintExcepti
 use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactException;
 use PrestaShop\PrestaShop\Core\Domain\Contact\ValueObject\ContactId;
 use PrestaShopException;
+use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
+use Symfony\Component\Validator\Exception\InvalidOptionsException;
+use Symfony\Component\Validator\Exception\MissingOptionsException;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 /**
@@ -62,7 +65,16 @@ final class AddContactHandler extends AbstractObjectModelHandler implements AddC
     /**
      * {@inheritdoc}
      *
+     * @param AddContactCommand $command
+     *
+     * @return ContactId
+     *
+     * @throws CannotAddContactException
+     * @throws ConstraintDefinitionException
+     * @throws ContactConstraintException
      * @throws ContactException
+     * @throws InvalidOptionsException
+     * @throws MissingOptionsException
      */
     public function handle(AddContactCommand $command)
     {
@@ -108,6 +120,9 @@ final class AddContactHandler extends AbstractObjectModelHandler implements AddC
      * @param array $localisedTitle
      *
      * @throws ContactConstraintException
+     * @throws ConstraintDefinitionException
+     * @throws InvalidOptionsException
+     * @throws MissingOptionsException
      */
     private function assertLocalisedTitleContainsDefaultLanguage(array $localisedTitle)
     {
--- a/src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
+++ b/src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
@@ -32,8 +32,11 @@ use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactException;
 use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Contact\Query\GetContactForEditing;
 use PrestaShop\PrestaShop\Core\Domain\Contact\QueryHandler\GetContactForEditingHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Exception\DomainConstraintException;
+use PrestaShopBundle\Utils\BoolParser;
 use PrestaShopException;
 use Symfony\Component\Form\DataTransformerInterface;
+use Symfony\Component\Form\Exception\TransformationFailedException;
 
 /**
  * Class GetContactForEditingHandler is responsible for getting the data for contact edit page.
@@ -58,7 +61,14 @@ final class GetContactForEditingHandler implements GetContactForEditingHandlerIn
     /**
      * {@inheritdoc}
      *
+     * @param GetContactForEditing $query
+     *
+     * @return EditableContact
+     *
      * @throws ContactException
+     * @throws ContactNotFoundException
+     * @throws DomainConstraintException
+     * @throws TransformationFailedException
      */
     public function handle(GetContactForEditing $query)
     {
@@ -78,7 +88,7 @@ final class GetContactForEditingHandler implements GetContactForEditingHandlerIn
                 $query->getContactId()->getValue(),
                 $contact->name,
                 $contact->email,
-                $contact->customer_service,
+                BoolParser::castToBool($contact->customer_service),
                 $contact->description,
                 $this->stringArrayToIntegerArrayDataTransformer->reverseTransform($contact->getAssociatedShops())
             );
--- a/src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
+++ b/src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
@@ -27,12 +27,12 @@
 namespace PrestaShop\PrestaShop\Adapter\Contact\QueryHandler;
 
 use Contact;
-use PrestaShop\PrestaShop\Core\Domain\Contact\QueryResult\EditableContact;
 use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactException;
 use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Contact\Query\GetContactForEditing;
 use PrestaShop\PrestaShop\Core\Domain\Contact\QueryHandler\GetContactForEditingHandlerInterface;
-use PrestaShopBundle\Utils\BoolParser;
+use PrestaShop\PrestaShop\Core\Domain\Contact\QueryResult\EditableContact;
+use PrestaShop\PrestaShop\Core\Util\BoolParser;
 use PrestaShopException;
 use Symfony\Component\Form\DataTransformerInterface;
 
@@ -74,7 +74,6 @@ final class GetContactForEditingHandler implements GetContactForEditingHandlerIn
                     )
                 );
             }
-
             $editableContact = new EditableContact(
                 $query->getContactId()->getValue(),
                 $contact->name,
--- a/src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
+++ b/src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
@@ -32,11 +32,9 @@ use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactException;
 use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Contact\Query\GetContactForEditing;
 use PrestaShop\PrestaShop\Core\Domain\Contact\QueryHandler\GetContactForEditingHandlerInterface;
-use PrestaShop\PrestaShop\Core\Domain\Exception\DomainConstraintException;
 use PrestaShopBundle\Utils\BoolParser;
 use PrestaShopException;
 use Symfony\Component\Form\DataTransformerInterface;
-use Symfony\Component\Form\Exception\TransformationFailedException;
 
 /**
  * Class GetContactForEditingHandler is responsible for getting the data for contact edit page.
@@ -61,14 +59,7 @@ final class GetContactForEditingHandler implements GetContactForEditingHandlerIn
     /**
      * {@inheritdoc}
      *
-     * @param GetContactForEditing $query
-     *
-     * @return EditableContact
-     *
      * @throws ContactException
-     * @throws ContactNotFoundException
-     * @throws DomainConstraintException
-     * @throws TransformationFailedException
      */
     public function handle(GetContactForEditing $query)
     {
--- a/src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
+++ b/src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
@@ -32,7 +32,6 @@ use PrestaShop\PrestaShop\Core\Domain\Contact\Exception\ContactNotFoundException
 use PrestaShop\PrestaShop\Core\Domain\Contact\Query\GetContactForEditing;
 use PrestaShop\PrestaShop\Core\Domain\Contact\QueryHandler\GetContactForEditingHandlerInterface;
 use PrestaShop\PrestaShop\Core\Domain\Contact\QueryResult\EditableContact;
-use PrestaShop\PrestaShop\Core\Util\BoolParser;
 use PrestaShopException;
 use Symfony\Component\Form\DataTransformerInterface;
 
@@ -78,7 +77,7 @@ final class GetContactForEditingHandler implements GetContactForEditingHandlerIn
                 $query->getContactId()->getValue(),
                 $contact->name,
                 $contact->email,
-                BoolParser::castToBool($contact->customer_service),
+                (bool) $contact->customer_service,
                 $contact->description,
                 $this->stringArrayToIntegerArrayDataTransformer->reverseTransform($contact->getAssociatedShops())
             );
--- a/src/PrestaShopBundle/Utils/BoolParser.php
+++ b/src/PrestaShopBundle/Utils/BoolParser.php
@@ -19,8 +19,5 @@ class BoolParser
         }
 
         return strtolower($value) !== 'false';
-//        $lValue = strtolower($value);
-//
-//        return $lValue == 'false' ? false : true;
     }
 }
--- a/src/PrestaShopBundle/Utils/BoolParser.php
+++ b/src/PrestaShopBundle/Utils/BoolParser.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace PrestaShopBundle\Utils;
+
+class BoolParser
+{
+    /**
+     * @param string|int $value
+     *
+     * @return bool
+     */
+    public static function castToBool($value)
+    {
+        if (is_bool($value)) {
+            return $value;
+        }
+        if (is_numeric($value)) {
+            return (bool) $value; // 0 => false; all other true
+        } else {
+            $lValue = strtolower($value);
+
+            return $lValue == 'false' ? false : true;
+        }
+    }
+}
--- a/src/PrestaShopBundle/Utils/BoolParser.php
+++ b/src/PrestaShopBundle/Utils/BoolParser.php
@@ -16,10 +16,11 @@ class BoolParser
         }
         if (is_numeric($value)) {
             return (bool) $value; // 0 => false; all other true
-        } else {
-            $lValue = strtolower($value);
-
-            return $lValue == 'false' ? false : true;
         }
+
+        return strtolower($value) !== 'false';
+//        $lValue = strtolower($value);
+//
+//        return $lValue == 'false' ? false : true;
     }
 }
--- a/tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
@@ -66,9 +66,6 @@ abstract class AbstractDomainFeatureContext implements Context
 
     /**
      * @return CommandBusInterface
-     *
-     * @throws ServiceCircularReferenceException
-     * @throws ServiceNotFoundException
      */
     protected function getCommandBus()
     {
@@ -77,9 +74,6 @@ abstract class AbstractDomainFeatureContext implements Context
 
     /**
      * @return CommandBusInterface
-     *
-     * @throws ServiceCircularReferenceException
-     * @throws ServiceNotFoundException
      */
     protected function getQueryBus()
     {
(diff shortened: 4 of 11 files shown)


src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
```diff
@@ -78,7 +88,7 @@ public function handle(GetContactForEditing $query)
                 $query->getContactId()->getValue(),
                 $contact->name,
                 $contact->email,
-                $contact->customer_service,
+                BoolParser::castToBool($contact->customer_service),
```
tdavidsonas88: @rokaszygmantas @matks created BoolParser to handle cases like: "0", "false" returned from legacy classes/request handlers instead of boolean false;

tests/Unit/PrestaShopBundle/Utils/BoolParserTest.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Tests\Unit\PrestaShopBundle\Utils;
+
+use PHPUnit\Framework\TestCase;
+use PHPUnit_Framework_AssertionFailedError;
+use PrestaShopBundle\Utils\BoolParser;
+
+class BoolParserTest extends TestCase
```
tdavidsonas88: @rokaszygmantas @matks created BoolParser to handle cases like: "0", "false" returned from legacy classes/request handlers instead of boolean false; Tested it with phpunit for your convenience.

src/Adapter/Contact/CommandHandler/AddContactHandler.php
```diff
@@ -62,7 +65,16 @@ public function __construct(ValidatorInterface $validator)
...
      *
+     * @param AddContactCommand $command
+     *
+     * @return ContactId
+     *
+     * @throws CannotAddContactException
+     * @throws ConstraintDefinitionException
+     * @throws ContactConstraintException
      * @throws ContactException
+     * @throws InvalidOptionsException
```
PierreRambaud: many exceptions like this one are not throws in this function.

src/Adapter/Contact/CommandHandler/AddContactHandler.php
```diff
@@ -108,6 +120,9 @@ public function handle(AddContactCommand $command)
      * @param array $localisedTitle
      *
      * @throws ContactConstraintException
+     * @throws ConstraintDefinitionException
+     * @throws InvalidOptionsException
+     * @throws MissingOptionsException
```
PierreRambaud: same here

src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
```diff
@@ -58,7 +61,14 @@ public function __construct(DataTransformerInterface $stringArrayToIntegerArrayD
...
      * {@inheritdoc}
      *
+     * @param GetContactForEditing $query
+     *
+     * @return EditableContact
+     *
      * @throws ContactException
+     * @throws ContactNotFoundException
+     * @throws DomainConstraintException
+     * @throws TransformationFailedException
```
PierreRambaud: same here

src/PrestaShopBundle/Utils/BoolParser.php
```diff
@@ -0,0 +1,25 @@
...
+     * @return bool
+     */
+    public static function castToBool($value)
+    {
+        if (is_bool($value)) {
+            return $value;
+        }
+        if (is_numeric($value)) {
+            return (bool) $value; // 0 => false; all other true
+        } else {
```
PierreRambaud: else is useless

src/PrestaShopBundle/Utils/BoolParser.php
```diff
@@ -0,0 +1,25 @@
...
+            $lValue = strtolower($value);
+
+            return $lValue == 'false' ? false : true;
```
PierreRambaud:
```suggestion
            return strtolower($value) === 'true';
```

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
```diff
@@ -59,6 +63,9 @@ public static function prepare(BeforeSuiteScope $scope)
 
     /**
      * @return CommandBusInterface
+     *
+     * @throws ServiceCircularReferenceException
```
PierreRambaud: there is no exception thrown here

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
```diff
@@ -67,6 +74,9 @@ protected function getCommandBus()
...
```
PierreRambaud: same

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
```diff
@@ -146,4 +158,25 @@ protected function parseLocalizedArray(string $parsedArray): array
...
+     * @param TableNode $table
+     *
+     * @return array
+     *
+     * @throws RuntimeException
+     */
+    protected function extractFirstHorizontalRowFromProperties(TableNode $table): array
+    {
+        $hash = $table->getHash();
+        if (count($hash) != 1) {
```
PierreRambaud:
```suggestion
        if (count($hash) !== 1) {
```

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
```diff
@@ -146,4 +158,25 @@ protected function parseLocalizedArray(string $parsedArray): array
...
+            throw new RuntimeException('Properties are invalid');
+        }
+        /** @var array $data */
+        $data = $hash[0];
```
PierreRambaud:
```suggestion
        return $hash[0];
```

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
```diff
@@ -0,0 +1,187 @@
...
+     * @throws ContactException
+     * @throws ServiceCircularReferenceException
+     * @throws ServiceNotFoundException
+     */
+    public function thereIsNoContactWithId(int $contactId)
+    {
+        try {
+            $this->getQueryBus()->handle(new GetContactForEditing($contactId));
+        } catch (ContactNotFoundException $exception) {
+            return;
```
PierreRambaud:
```suggestion
```

src/PrestaShopBundle/Utils/BoolParser.php
tdavidsonas88: Thanks for the tip :) Actually better would be:
`        return strtolower($value) !== 'false';`

to pass these cases:
```
        $this->assertTrue(BoolParser::castToBool('anything else'));
        $this->assertTrue(BoolParser::castToBool('on'));
```

src/Adapter/Contact/CommandHandler/AddContactHandler.php
tdavidsonas88: have changed the PHPStorm setting for call tree analysis depth from 1 to 0. This will limit exceptions of just this function and will not look for them in the inner methods. So in the future, will be less of these to review.

tests/Unit/PrestaShopBundle/Utils/BoolParserTest.php
matks: It's a good idea but can we rather store it in `PrestaShop\PrestaShop\Core\Util` namespace (=> `src/Core/Util` folder) ?

`PrestaShopBundle` should contain all the code required to run PrestaShop as a Symfony application. This means that all the code in it must either serve as an util class to make prestashop domain code usable in Symfony or a class dedicated to app layer concerns, such as rendering twig views, handling sessions ...

The idea behind is to have a Core prestashop code folder which contains all the __logic__ of PrestaShop: Cart Rules, Customer logic, Multistore ... that is being wrapped by PrestaShopBundle to make it a Symfony application.
 In an ideal world, tomorrow if we decide to migrate from Symfony to Laravel, we might be able to create a new folder PrestaShopLaravel where we put all the code required to run PrestaShop as a Laravel application, delete PrestaShopBundle and reuse 100% of PrestaShop core folder.

This is why PrestaShop core folder must remain framework-agnostic.

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
```diff
@@ -146,4 +150,23 @@ protected function parseLocalizedArray(string $parsedArray): array
...
+    /**
+     * Used in scenarios which have horizontal properties table
+     *
+     * @param TableNode $table
+     *
+     * @return array
+     *
+     * @throws RuntimeException
+     */
+    protected function extractFirstHorizontalRowFromProperties(TableNode $table): array
```
rokaszygmantas: in which case a row is not horizontal? :thinking:

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * @Given there is no contact with id :contactId
+     *
+     * @param int $contactId
+     */
+    public function thereIsNoContactWithId(int $contactId)
+    {
+        try {
+            $this->getQueryBus()->handle(new GetContactForEditing($contactId));
```
rokaszygmantas: this test will not fail if contact exists with given id.
You could use `\Tests\Integration\Behaviour\Features\Context\Util\NoExceptionAlthoughExpectedException` to make sure it fails when Contact is found

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
```diff
@@ -0,0 +1,145 @@
...
+    {
+        $this->getQueryBus()->handle(new GetContactForEditing($contactId));
+    }
+
+    /**
+     * @Then I should be able to get contact with id :contactId for editing
+     *
+     * @param int $contactId
+     */
+    public function iShouldBeAbleToGetContactWithIdForEditing(int $contactId)
```
rokaszygmantas: same method as above, maybe can be joined into one?

tests/Integration/Behaviour/Features/Scenario/Contact/contact.feature
```diff
@@ -0,0 +1,27 @@
+@reset-database-before-feature
+Feature: Contact
+  In order to create customizable contact us form for customers
+  As a BO user
+  I should be able to add and edit new contact
+
+  Scenario: Add new contact
+    Given there is no contact with id 3
+    And there is contact with id 2
```
matks: I'm worried about relying on ID generation strategy 🤔
- it's not very user-oriented (customers will talk about contact "Rokas Z" or "Mathieu F" rather than contact "with ID 34")
- if one day we change the IDs generation strategy (for example to use UUIDs) all these tests will crash although the logic is still good

@PierreRambaud @jolelievre Any opinion ?
I guess we could start creating dedicated DataProviders / Repositories with search functions `getContactByName()`, `lookForContact()` and use them to find the IDs needed by commands and queries ?

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
```diff
@@ -0,0 +1,145 @@
...
+     * @param TableNode $table
+     */
+    public function iAddNewContactWithTheFollowingProperties(TableNode $table)
+    {
+        $data = $this->extractFirstHorizontalRowFromProperties($table);
+        /** @var EditableContact $editablContact */
+        $editableContact = $this->mapToEditableContact(self::DUMMY_CONTACT_ID, $data);
+
+        $addContactCommand = new AddContactCommand(
+            $editableContact->getLocalisedTitles(), $editableContact->isMessagesSavingEnabled()
```
rokaszygmantas:
```suggestion
            $editableContact->getLocalisedTitles(),
            $editableContact->isMessagesSavingEnabled()
```
each argument should be in separate line

tests/Unit/PrestaShopBundle/Utils/BoolParserTest.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * @throws PHPUnit_Framework_AssertionFailedError
+     */
+    public function testCastToBool()
+    {
+        $this->assertFalse(BoolParser::castToBool('false'));
```
rokaszygmantas: try casting an object or array :slightly_smiling_face:

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
tdavidsonas88: @rokaszygmantas https://github.com/PrestaShop/PrestaShop/blob/0facef127b6d512cce953c5e2c693705af46fdec/tests/Integration/Behaviour/Features/Scenario/Customer/customer_management.feature#L8-L12

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * @param int $contactId
+     * @param TableNode $table
+     */
+    public function iUpdateContactWithIdWithTheFollowingProperties(int $contactId, TableNode $table)
+    {
+        $data = $this->extractFirstHorizontalRowFromProperties($table);
+        $editableContact = $this->mapToEditableContact($contactId, $data);
+
+        $editContactCommand = new EditContactCommand($contactId);
```
tdavidsonas88, line 94: Can chain them here with fluent setters :)

tests/Integration/Behaviour/behat.yml
```diff
@@ -6,6 +6,12 @@ default:
             contexts:
                 - Tests\Integration\Behaviour\Features\Context\CommonFeatureContext
                 - Tests\Integration\Behaviour\Features\Context\SqlManagerFeatureContext
+        contact:
```
tdavidsonas88, line 9: @matks @rokaszygmantas had to include this here instead of the bottom after CLDR and some other suites are launched test database was not reset as expected; For example: ps_lang table had 7 different languages instead of just English.

src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
jolelievre: This seems really overkill, doesn't a simple cast work? `(bool) $contact->customer_service` The only use case that the class can handle is if the string is equal to `'false'`  But I don't think the legacy class ever return this and expect it to be considered false
As for the behat values then we should simply use 0

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
jolelievre: Actually all our behat tests are written this way and they use `$data = $node->getRowsHash();` to get the data, so I'm not sure it add interest to change this for those tests, especially since you only get the data for one row
This only multiply the way we build the test, so for somone who would want to add some it adds complexity without additional value

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
jolelievre: Indeed considering the name of the method, it means you expect the `ContactNotFoundException` to be thrown so if you want to trigger this test then you can throw an exception right after the handle call
```suggestion
            $this->getQueryBus()->handle(new GetContactForEditing($contactId));
            throw new RuntimeException(sprintf('Contact found whereas not contact with id %s was expected', $contactId));
```

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
```diff
@@ -0,0 +1,145 @@
...
+     * @param int $contactId
+     * @param TableNode $table
+     */
+    public function contactWithIdShouldHaveTheFollowingProperties(int $contactId, TableNode $table)
+    {
+        $data = $this->extractFirstHorizontalRowFromProperties($table);
+        $expectedEditableContact = $this->mapToEditableContact($contactId, $data);
+        /** @var EditableContact $editableContact */
+        $editableContact = $this->getQueryBus()->handle(new GetContactForEditing($contactId));
+        PHPUnit_Framework_Assert::assertEquals($expectedEditableContact, $editableContact);
```
jolelievre, line 75: I'm not sure we should mix Behat and PHPUnit

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * @param int $contactId
+     * @param array $data
+     *
+     * @return EditableContact
+     */
+    private function mapToEditableContact(int $contactId, array $data): EditableContact
+    {
+        $title = $data['title'];
```
jolelievre: You don't need this temporary variables you can use them in the constructor right away

tests/Integration/Behaviour/Features/Scenario/Contact/contact.feature
jolelievre: I agree the tests should rely on database IDs But maybe you don't need a repository In other behat context we use the `SharedStorage` to store references to objects Then you can get the id from the referenced saved object

tests/Integration/Behaviour/behat.yml
jolelievre: Then it means the database is not properly initialized or clean The order shouldn't matter as the tests are supposed to be independent so what you say is quite worrying as some other context could influence others.. That's not problem as long as it triggers an error ^^ But the risk is having false positive
I thought databases were automatically reset, but it seems it is linked to https://github.com/PrestaShop/PrestaShop/blob/0facef127b6d512cce953c5e2c693705af46fdec/tests/Integration/Behaviour/Features/Context/CommonFeatureContext.php#L65
So I guess you need to add `@reset-database-before-feature` in your feature file Or `@database-scenario` before a scenario (but this one is not used for now)

jolelievre: Oh you already have the `@reset-database-before-feature` so it means it doesn't work properly?

tests/Unit/PrestaShopBundle/Utils/BoolParserTest.php
jolelievre: Yep, the folder architecture is explained here https://devdocs.prestashop.com/1.7/development/architecture/understanding-src-folder/

Also maybe one day we will be able to extract the whole `Core` folder and have a dependency library with it So then we could have a Symfony implementation AND a laravel one Or maybe other project will simply use the Core library for their own purpose

tests/Unit/PrestaShopBundle/Utils/BoolParserTest.php
tdavidsonas88: array currently not supported :)

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
tdavidsonas88: getRowsHash() is for vertical table; As I continue to develop Behat tests I agree it is a better approach to use vertical tables (even sometimes horizontal one looks nicer).  Thank you for the notice.

tdavidsonas88: Thank you jolelievre for

> This seems really overkill, doesn't a simple cast work? (bool) $contact->customer_service The only use case that the class can handle is if the string is equal to 'false' But I don't think the legacy class ever return this and expect it to be considered false
> As for the behat values then we should simply use 0

The simple cast works. And for Behat tests I have found nice Ferment function in PrimiteUtils

```
    public static function castStringBooleanIntoBoolean($element)
    {
        if ($element === 'false') {
            return false;
        }

        return boolval($element);
    }

```
The only case it could be needed when the 'false' comes from the Request to Controller (for example from the checkbox). And if Symfony forms cover this case then it is really not needed.

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
tdavidsonas88: Yes, that's true Behat gives the ability to return table hash, formed by rows or columns. So this kind of method and it's naming I've created has no additional value. Thank you for noticing :)

src/Adapter/Contact/QueryHandler/GetContactForEditingHandler.php
tdavidsonas88: Yes, that's true, PrimitiveUtils class has 

```
    /**
     * @param string $element
     *
     * @return bool
     */
    public static function castStringBooleanIntoBoolean($element)
    {
        if ($element === 'false') {
            return false;
        }

        return boolval($element);
    }
```

tests/Unit/PrestaShopBundle/Utils/BoolParserTest.php
tdavidsonas88: I see from documentation that PrestaShopBundle/Utils should be move to Core eventually.

tests/Integration/Behaviour/behat.yml
tdavidsonas88: Yes, the Behat cases did not work properly when contact feature runs at the end. When I added new contact, instead of just only 1 item inside localisedTitles and localisedDescription, I've received 7. One for each language which were created with CLDR feature. 

Not sure why these languages were not cleaned with @reset-database-before-feature or if they were cleared how they reappeared. Also it could be that cache is not disabled properly but I see that it is disabled for the ObjectModel. Maybe object model disable Cache not working properly for Languages?

When DatabaseDump::restoreDb(); happens, I can see in ps_dump.sql that the ps_lang table is dropped though:

```
--
-- Table structure for table `ps_lang`
--

DROP TABLE IF EXISTS `ps_lang`;
/*!40101 SET @saved_cs_client     = @@character_set_client */;
/*!40101 SET character_set_client = utf8 */;
CREATE TABLE `ps_lang` (
  `id_lang` int(11) NOT NULL AUTO_INCREMENT,
  `name` varchar(32) COLLATE utf8_unicode_ci NOT NULL,
  `active` tinyint(1) NOT NULL,
  `iso_code` varchar(2) COLLATE utf8_unicode_ci NOT NULL,
  `language_code` varchar(5) COLLATE utf8_unicode_ci NOT NULL,
  `locale` varchar(5) COLLATE utf8_unicode_ci NOT NULL,
  `date_format_lite` varchar(32) COLLATE utf8_unicode_ci NOT NULL,
  `date_format_full` varchar(32) COLLATE utf8_unicode_ci NOT NULL,
  `is_rtl` tinyint(1) NOT NULL,
  PRIMARY KEY (`id_lang`)
) ENGINE=InnoDB AUTO_INCREMENT=2 DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
/*!40101 SET character_set_client = @saved_cs_client */;
```
So not sure why it happens like that:

```
    Then contact 4 should have the following details:  # Tests\Integration\Behaviour\Features\Context\Domain\ContactFeatureContext::contactShouldHaveTheFollowingDetails()
      | title                     | test service 2      |
      | email_address             | test@prestashop.com |
      | is_message_saving_enabled | true                |
      | description               | test12345           |
      | shop_id_association       | 1                   |
      Failed asserting that two objects are equal.
      --- Expected
      +++ Actual
      @@ @@
       PrestaShop\PrestaShop\Core\Domain\Contact\QueryResult\EditableContact Object (
           'contactId' => 4
           'localisedTitles' => Array (
               1 => 'test service 2'
      +        2 => 'test service 2'
      +        3 => 'test service 2'
      +        4 => 'test service 2'
      +        5 => 'test service 2'
      +        6 => 'test service 2'
      +        7 => 'test service 2'
           )
           'email' => PrestaShop\PrestaShop\Core\Domain\ValueObject\Email Object (...)
           'isMessagesSavingEnabled' => true
           'localisedDescription' => Array (
               1 => 'test12345'
      +        2 => ''
      +        3 => ''
      +        4 => ''
      +        5 => ''
      +        6 => ''
      +        7 => ''
           )
           'shopAssociation' => Array (...)
       )

```
When I query the tesdb at the end I see only one language in the output. So my guess is that object model cache for languages is not disabled properly in this case.

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
tdavidsonas88: @jolelievre @matks @rokaszygmantas In this Behat case, yes, just thought about this style if it is not the Behat test code but production code and the error happens on the live system. Let's imagine the situation, Prestashop shop is running on heavy loaded system and every minute gives profit for the company. Error happens and developer need to fix it ASAP. The case is specific to the LIVE system and in this style of code it won't be obvious which code caused the error - only the long error line will be visible and debugging might be needed.

So not sure what gives more business value to PrestaShop:

1. Quickly fixing production errors by developers Or
2. Speed / memory usage of software
3. My assumptions might be not very good also, because I came from the company with limited knowledge of E-commerce and it was more important to fix errors in production environment ASAP. Successful production environment running time was traded for money and customers value; It might be that I don't understand something about live system maintenance - as this style was forced to me by another more experienced developer.

tdavidsonas88:
Thank you for your professional feedback, did my best to make the code better.

tests/Integration/Behaviour/Features/Scenario/Contact/contact.feature
tdavidsonas88: Changed id generation strategy to references and SharedStorage in the last commit

tests/Integration/Behaviour/Features/Context/Domain/ContactFeatureContext.php
matks: @jolelievre we are only using PHPUnit assertion functions when they are useful. So far in our behat tests we have been using homemade assertion logic and throwing exceptions when the  result was bad ; PHPUnit provides some nice assertions that can save us from writing a lot of boilerplate code.

I've seen it multiple times 😊for projects using behat: using PHPUnit not as a unit test engine but as library full of useful assertions.

tests/Integration/Behaviour/behat.yml
matks: I think it's valid to have a contact dedicated behat suite.
About CLDR messing up our tests, we will need to explore this issue too 😋but current solution is relevant to me 👍

matks (dismissed):

matks:
@tdavidsonas88 the PR is approved 💚

However I cannot merge it because of a git conflict. Can you please rebase it, then we'll merge it ?

tdavidsonas88:
> @tdavidsonas88 the PR is approved 
> 
> However I cannot merge it because of a git conflict. Can you please rebase it, then we'll merge it ?

Thank you @matks for the review. Already rebased!

matks (approved):

matks:
Thank you and congratulations @tdavidsonas88

PR #16778, opened 2019-12-11
Fixed frontend 500 err. when deleting default cats

### LINKED ISSUE #19751: Error 500 when deleting the default category
#### Describe the bug

After bulk deleting the default Prestashop's categories, an error 500 occured on the frontend. The backend was still working. I found out with the debug mode that it cames from the following file: classes/Link.php
It seems that even after the deletion some data still remains and triggers the "throwError" when the category object is incorrectly formed. After that 1st fix, I had to fix a ternary condition for when the variable $category is not an object.

#### Expected behavior

Having no error when you're deleting the default category

#### Steps to Reproduce

Install an 1.7.6.x shop, delete the default category, you shouldn't have any error in FO and BO

#### Additional information

* PrestaShop version: 1.7.6.x

Created the issue for this PR : https://github.com/PrestaShop/PrestaShop/pull/16778

thomas-ddev:
| Questions     | Answers
| Branch?       |  develop
| Description?  | After bulk deleting the default Prestashop's categories, an error 500 occured on the frontend. The backend was still working. I found out with the debug mode that it cames from the following file: classes/Link.php <br /> It seems that even after the deletion some data still remains and triggers the "throwError" when the category object is incorrectly formed. After that 1st fix, I had to fix a ternary condition for when the variable $category is not an object.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19751 
| How to test?  | Can replicate only on one website, not sure if the bug occures on other ones, but if I put back the previous code the 500 error remains.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16778)

### DIFF
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -385,7 +385,7 @@ class LinkCore
             } elseif ((int) $category) {
                 $category = new Category((int) $category, $idLang);
             } else {
-                throw new PrestaShopException('Invalid category vars');
+		return null;
             }
         }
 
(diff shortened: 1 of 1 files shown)


Progi1984:
@thomas-ddev  Thank you for your contribution. :hugs: 

Thanks :+1:

NeOMakinG:
Hey @thomas-ddev, you noticed inside your PR's description that it's targeting 1.7.6.x version, having the target branch of develop is a mistake then or you don't mind if it's only fixed on 1.7.8 ?

NeOMakinG:
@thomas-ddev I created the issue, think about creating it next time before opening a PR for the changelog / test case :)

https://github.com/PrestaShop/PrestaShop/issues/19751

NeOMakinG:
@thomas-ddev since we've no news from you, I tried to reproduce it, couldn't reproduce it at all, I'm closing the issue !

Thanks anyway for the support :)

PR #13009, opened 2019-03-22
Change product canonical url to parent, no more redirection

jolelievre:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Change product canonical url to parent, no more redirection
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #12548
| How to test?  | Go to catalog pages, check that product with combinations have their link to the parent url (no attribute id in the url). Then go to the product page, check that you are not redirected (the url does not contain an attribute id) but the default combination is displayed. Check that combination urls are still accessible but their canonical all point to the parent one.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/13009)

PR #16806, opened 2019-12-13
[*] FO : Add default attribute to canonical url for better seo

cedricfontaine:
It's recommended to avoid sending 301 redirect to canonical, otherwise Google will decide for you which url is the best as the canonical. It's better to send the default attribute as the canonical.

| Questions     | Answers
| Branch?       | develop 
| Description?  | Send canonical with attribute
| Type?         | improvment
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | Verify that canonical includes attribute

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16806)

### DIFF
--- a/src/Adapter/Presenter/Product/ProductLazyArray.php
+++ b/src/Adapter/Presenter/Product/ProductLazyArray.php
@@ -735,7 +735,12 @@ class ProductLazyArray extends AbstractLazyArray
         $linkRewrite = isset($product['link_rewrite']) ? $product['link_rewrite'] : null;
         $category = isset($product['category']) ? $product['category'] : null;
         $ean13 = isset($product['ean13']) ? $product['ean13'] : null;
-
+        $attribute = null;
+        if (!$canonical && $product['id_product_attribute'] > 0) {
+            $attribute = $product['id_product_attribute'];
+        } elseif ($product['cache_default_attribute'] > 0){
+            $attribute = $product['cache_default_attribute'];
+        }
         return $this->link->getProductLink(
             $product['id_product'],
             $linkRewrite,
@@ -743,7 +748,7 @@ class ProductLazyArray extends AbstractLazyArray
             $ean13,
             $language->id,
             null,
-            !$canonical && $product['id_product_attribute'] > 0 ? $product['id_product_attribute'] : $product['cache_default_attribute'] > 0 ? $product['cache_default_attribute'] : null,
+            $attribute,
             false,
             false,
             true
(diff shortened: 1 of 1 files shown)


src/Adapter/Presenter/Product/ProductLazyArray.php
```diff
@@ -743,7 +743,7 @@ private function getProductURL(
             $ean13,
             $language->id,
             null,
-            !$canonical && $product['id_product_attribute'] > 0 ? $product['id_product_attribute'] : null,
+            !$canonical && $product['id_product_attribute'] > 0 ? $product['id_product_attribute'] : $product['cache_default_attribute'] > 0 ? $product['cache_default_attribute'] : null,
```
PierreRambaud: It's unreadable, could you use variable instead?

yannicka: "Nested ternary operators without explicit parentheses" is a deprecated feature for PHP 7.4: https://www.php.net/manual/en/migration74.deprecated.php#migration74.deprecated.core.nested-ternary

kpodemski:
I have the feeling that it's back and forth for this, every SEO expert says something different about this :)

matks:
> I have the feeling that it's back and forth for this, every SEO expert says something different about this :)

Same feeling 😐...

Sometimes I even feel like every SEO expert says something different about everything 😅. If I'm not wrong, SEO is actually a reverse-engineering of the google algorithm without having access to the code ?

cedricfontaine:
I see no reason to send a canonical that will generate a 301 redirect. It makes no sense, SEO or no SEO.

PierreRambaud:
Agree with @cedricfontaine There is no reason to send a 301 redirect (which is a move permanently) in a canonical url 🤔

matks:
ping @marionf and @colinegin

jolelievre:
Hello, indeed a 301 redirect is a bad practice which is why we removed it The parent url (without attribute) used to redirect to the canonical url (default attribute) but it has been changed in 176:
https://github.com/PrestaShop/PrestaShop/pull/13009

Now the parent url is the canonical url, is accessible and displays the default attribute and the redirection only happens when the combination is invalid So I don't understand the problem..

colinegin:
It would be easier for the product team to be aware of your work if this PR was associated to an issue (existing one or a new one that you create) which describes the purpose of your pr. Thanks !

cedricfontaine:
I'm not using the latest release and was not aware that the parent url is the canonical url is accessible and displays the default attribute. So it it's ok for me

colinegin:
This is great news, i'm glad we had the same ideas ! ;)

PR #16809, opened 2019-12-14
Fix date parsing in tests when TZ is changed to non-UTC during tests

mvorisek:
| Questions     | Answers
| Branch?       | develop
| Description?  | Tests are failing after midnight.
| Type?         | bug fix
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | Fork repo, setup Travis and push some commit after CEST midnight.

For some reasons, when I var_dumped the datetime used in the `Validate::isBirthDate` in the related `ValidateCoreTest::testIsBirthDate` test, the TZ was `US/Eastern`. This fixes it, the test data are generated in UTC.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16809)

### DIFF
--- a/tests-legacy/Unit/Classes/ValidateCoreTest.php
+++ b/tests-legacy/Unit/Classes/ValidateCoreTest.php
@@ -63,7 +63,14 @@ class ValidateCoreTest extends TestCase
      */
     public function testIsBirthDate($expected, $input)
     {
-        $this->assertSame($expected, Validate::isBirthDate($input));
+        // data from isBirthDateProvider provider are in UTC
+        $defaultTz = date_default_timezone_get();
+        date_default_timezone_set('UTC');
+        try {
+            $this->assertSame($expected, Validate::isBirthDate($input));
+        } finally {
+            date_default_timezone_set($defaultTz);
+        }
     }
 
     /**


Progi1984:
@mvorisek Thank your for your contribution :open_hands: 

Fixing a bug in a test is not a solution. A better solution could be to configure Travis to define the default timezone for PHP.

mvorisek:
@Progi1984 In Travis / PHP unit the TZ is already reset. But somewhere in the PHP unit tests the TZ is set to the mentioned `US/Eastern`.

This is how I fixed the test locally. If all PHP unit tests should be tested strictly isolated, then the test changing the TZ needs to be isolated and fixed like this one and this reverted.

matks:
> @mvorisek Thank your for your contribution 👐
> 
> Fixing a bug in a test is not a solution. A better solution could be to configure Travis to define the default timezone for PHP.

But tests must run in any environment, not only Travis. So we need to either setup such settings ourselves (like this PR does) or provide guidance for people for the right environment for these tests 🤔

matks:
What about we define globally a timezone ? Like in phpunit bootstrap file ? This fixes it once and for all and we do not care about "setting back the timezone" after the test was run ?

mvorisek:
> What about we define globally a timezone ? Like in phpunit bootstrap file ? This fixes it once and for all and we do not care about "setting back the timezone" after the test was run ?

There are two things:
1. TZ set/reset to UTC, currently TZ is set in https://github.com/PrestaShop/PrestaShop/blob/develop/composer.json . If you run tests thru composer command, the TZ is set corrently everywhere.
1. TZ changed during tests. It seems there is only one issue and this PR solves it. Or, as this issue is related with birth date, we can update the tests to not test values +- 24 hours around current time. Birth date in context of people and PrestaShop this makes sense too - is it better to solve this issue this way?

mvorisek:
@matks

> What about we define globally a timezone ? Like in phpunit bootstrap file ? This fixes it once and for all and we do not care about "setting back the timezone" after the test was run ?

There are two things:
1. TZ set/reset to UTC, currently TZ is set in https://github.com/PrestaShop/PrestaShop/blob/develop/composer.json . If you run tests thru composer command, the TZ is set corrently everywhere.
1. TZ changed during tests. It seems there is only one issue and this PR solves it. Or, as this issue is related with birth date, we can update the tests to not test values +- 24 hours around current time. Birth date in context of people and PrestaShop this makes sense too - is it better to solve this issue this way?

mvorisek:
@matks I have further analysed this issue as it breaks unit tests after UTC midnight on every test enviroment (incl. TravisCI, see https://travis-ci.org/mvorisek/PrestaShop/jobs/631449993#L683 )

I have run all unit tests with xdebug tracing enabled and the TZ is set to `US/Eastern` *only* at this line:
```
date_default_timezone_set() /builds/mvorisek/prestashop_eshop/config/config.inc.php:151
```

The issue is that phpunit data providers are run just after the phpunit bootstrap. The config file above, even it is a config file, was included at time `53.8167 sec` and `70156528 lines` after the phpunit has started. I do not think that this file can be moved into the bootstrap.php file easily.

As of now this is the only test failing due too data/dates without TZ from phpunit data provider. Can this PR be merged or can you advise another fix?

mvorisek:
@Progi1984, @matks I have further analysed this issue as it breaks unit tests after UTC midnight on every test enviroment (incl. TravisCI, see https://travis-ci.org/mvorisek/PrestaShop/jobs/631449993#L683 , test run on the latest `develop`)

I have run all unit tests with xdebug tracing enabled and the TZ is set to `US/Eastern` *only* at this line:
```
date_default_timezone_set() ./config/config.inc.php:151
```

The issue is that phpunit data providers are run just after the phpunit bootstrap. The config file above, even it is a config file, was included at time `53.8167 sec` and `70156528 lines` after the phpunit has started and it requires DB. I do not think that this file can be moved into the bootstrap.php file easily.

As of now this is the only test failing due too data/dates without TZ from phpunit data provider. Can this PR be merged or can you advise another fix?

mvorisek:
Any feedback?

PierreRambaud:
Why not using this in the travis.yml file?
```
before_install:
- export TZ=UTC
```

matks:
> Why not using this in the travis.yml file?
> 
> ```
> before_install:
> - export TZ=UTC
> ```

It means we only guarantee the tests work in Travis environment, not in another testing environment such as the developer computer 😥

PierreRambaud:
@matks This is because php, server and database are not configure properly, it's depending on the system and not the application :thinking:

matks:
> @matks This is because php, server and database are not configure properly, it's depending on the system and not the application 🤔

OK, but then it means "prestashop unit tests must be run only in an environment with UTC timezone". So either the developer makes the right settings for his local environment or we do this:

> What about we define globally a timezone ? Like in phpunit bootstrap file ? This fixes it once and for all and we do not care about "setting back the timezone" after the test was run ?

PierreRambaud:
> > @matks This is because php, server and database are not configure properly, it's depending on the system and not the application thinking
> 
> OK, but then it means "prestashop unit tests must be run only in an environment with UTC timezone". So either the developer makes the right settings for his local environment or we do this:
> 
> > What about we define globally a timezone ? Like in phpunit bootstrap file ? This fixes it once and for all and we do not care about "setting back the timezone" after the test was run ?

I was more thinking about mismatch configuration, many people are talking about timezone problem in Travis, maybe we can use both, the phpunit.xml with timezone in UTC, and the global travis UTC timezone, wdyt?

matks:
> > > @matks This is because php, server and database are not configure properly, it's depending on the system and not the application thinking
> > 
> > 
> > OK, but then it means "prestashop unit tests must be run only in an environment with UTC timezone". So either the developer makes the right settings for his local environment or we do this:
> > > What about we define globally a timezone ? Like in phpunit bootstrap file ? This fixes it once and for all and we do not care about "setting back the timezone" after the test was run ?
> 
> I was more thinking about mismatch configuration, many people are talking about timezone problem in Travis, maybe we can use both, the phpunit.xml with timezone in UTC, and the global travis UTC timezone, wdyt?

I like it 👍 it would be a clean solution

mvorisek:
No.

TZ for tests is already set to UTC in:
https://github.com/PrestaShop/PrestaShop/blob/707291944588cc633c91df26f54c85a58019479f/composer.json#L187

but later (re)set in the tests itself, more specifically in a config file, which is included during the tests:
https://github.com/PrestaShop/PrestaShop/blob/707291944588cc633c91df26f54c85a58019479f/config/config.inc.php#L151

So the enviroment settings are irrelevant and this issue is presented also in current TravisCI - see https://travis-ci.org/mvorisek/PrestaShop/jobs/631449993#L683

Can this PR be merged? Or is it better to not use `isBirthDateProvider` as a data provider at all (the only data provider which is affected by this issue)?

PierreRambaud (approved):
My bad, your change is totally legit :+1:

matks:
Since @PierreRambaud is 👍 LGTM

matks:
Thank you @mvorisek

PR #16818, opened 2019-12-15
Force Configuration::updateValue to return a bool

### LINKED ISSUE #14363: Configuration::updateValue can return wrong type
**Describe the bug**
`Configuration::updateValue` return wrong type when you update the value.
Looks ok when the value to update is equal to the value in DB.
Wrong doc is here : https://github.com/PrestaShop/PrestaShop/blob/1.7.5.2/classes/Configuration.php#L423

**To Reproduce**
1. Create a configuration key named 'myawesomeconfig' set to 'false';
2. `Configuration::updateValue('myawesomeconfig', 'false');` will return true
3. `Configuration::updateValue('myawesomeconfig', 'true');` will return 1 (and not a bool as stated in the docs)

**Additionnal information**
PrestaShop version: 1.7.5.2
PHP version: 7.2.17

zalexki:
| Questions     | Answers
| Branch?       | develop
| Description?  | Force Configuration::updateValue to return a boolean
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | yes - the method now returns bool instead of int
| Deprecations? | no
| Fixed ticket? | Fixes #14363
| How to test?  | check return of updateValue method.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16818)

### DIFF
--- a/classes/Configuration.php
+++ b/classes/Configuration.php
@@ -539,7 +539,7 @@ class ConfigurationCore extends ObjectModel
 
         Configuration::set($key, $values, $idShopGroup, $idShop);
 
-        return $result;
+        return (bool) $result;
     }
 
     /**


zalexki:
Since the operator `&=` return a int(1) instead of bool(true) we have to cast it before returning it

matks:
The idea is good but ... I'm afraid we cant merge this
- we would need to go through all usages of this function to make sure they are casting (and not checking `=== 1`)
- even if we did that, we would introduce a BC break for any 3rd party library relying on this being an integer instead of a boolean 😐

Although the benefit is obvious, it would be a lot cleaner, I'm afraid the cost of such a BC compared to the benefit is not a good balance ... 😢

PierreRambaud (approved):

PierreRambaud:
@matks I think it's the opposite, since the php doc says it should be a Boolean, it's not a BC, it's a bug.
Otherwise everytime we fix a bug on the cart, it must be consider as BC as keep the old behaviour :thinking:

zalexki:
as explained on the issue linked, the updateValue is inconsistent (it's returning a bool if the current value === new value and returning int(1) for all other cases).
Is this still really a BC break if we break code that do not properly work ?

About the usage of this function inside presta, i did a quick search and only saw unsafe checkings with `==` or just `!`.

matks:
> @matks I think it's the opposite, since the php doc says it should be a Boolean, it's not a BC, it's a bug.
> Otherwise everytime we fix a bug on the cart, it must be consider as BC as keep the old behaviour 🤔

You've seen this exact usecase with ps_facetedsearch. When a buggy behavior stays long enough, it becomes the expected behavior.

In the case of ps_facetedsearch, the fix was understandable as it was a new major version. This fix here is likely, if merged, to be shipped in a minor update.

PierreRambaud:
@matks For facetedsearch, it was more complicated, and even if people thought it was the expected behaviour, we changed it and keep it. More, you can't compare versions in modules and the Core, because we can do everything we want in modules :thinking: 

Wait for @eternoendless' feedback :)

matks:
@eternoendless 🆙

eternoendless:
I agree with @PierreRambaud, this is a bug fix because the signature indicated that it was supposed to return a bool. Let's just add this in the release notes.

PierreRambaud (approved):

Progi1984 (approved):

NeOMakinG:
Hem. hard to test for the QA, what's up with this PR then?

PierreRambaud:
Thanks @zalexki

Issue #14462, opened 2019-07-03
[EPIC] Add a modal to confirm or cancel delete in bulk actions


marionf:
**Describe the bug**
Add a modal like on product list (Catalog > Products) to confirm or cancel delete in bulk action and remove the other one

**Third part to do during the migration of each pages from 1.7.7**
Pages not yet migrated for which it must be added:
- [ ] Orders > Carts
- [ ] Catalog > Attributes
- [ ] Catalog > Features
- [ ] Catalog > Discounts > Cart rules
- [ ] Catalog > Discounts > Catalog price rules
- [ ] Customers > Customer service
- [ ] Design > Image settings
- [ ] Shipping > Carriers
- [ ] International > Locations > Zones
- [ ] International > Locations > Countries
- [ ] International > Locations > States
- [ ] International > Taxes > Tax rules
- [ ] International > Taxes > Tax rules > Edit a tax rule > Countries
- [ ] Shop parameters > Customer settings > Titles
- [ ] Shop parameters > Contact > Stores
- [ ] Shop parameters > Traffic & SEO > Referrers
- [ ] Shop parameters > Search > Search
- [ ] Shop parameters > Search > Tags

## Fixed
- [x] Shop parameters > Traffic & SEO > Search engines #10589
- [x] Shop parameters > Order settings > Statuses : #27606
- [x] Customers > Order messages - #18641
- [x] Customers > Addresses
- [x] #17816 - Add a modal to confirm or cancel delete in bulk actions 1.7.7 [1/2] 
- [x] #17845 - Add a modal to confirm or cancel delete in bulk actions 1.7.8 [2/2] 

**Screenshots**

Add this modal (@LouiseBonnard for the wording, do you think it's better to have the same general wording for all or to have a specific one for each pages ?)

![capture d'écran_1723](https://user-images.githubusercontent.com/13449658/60580979-4f1d2780-9d86-11e9-98b8-3129b605fd4b.png)

Remove this one

![capture d'écran_1724](https://user-images.githubusercontent.com/13449658/60581665-90620700-9d87-11e9-98d6-d8bb91b0c6d2.png)

marionf:
@matks FYI

LouiseBonnard:
@marionf, I think one global wording is better. Something like:

_Delete selection_
_Are you sure you want to (permanently) delete all selected items?_

[...]

PR #16820, opened 2019-12-15
Enable confirm modal for catalog > files listing for bulk delete action

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Introduce new type of Delete Bulk Action for grids, which uses a modal to confirm the user intent, and enable it on catalog > files listing
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes 1 item of https://github.com/PrestaShop/PrestaShop/issues/14462 and 1 item of https://github.com/PrestaShop/PrestaShop/issues/10550
| How to test?  | Please check that "bulk delete action" on Catalog > files listing uses a modal. No need to rebuild assets. Page can be accessed through BO URL `/admin-dev/index.php/sell/attachments/`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16820)


matks:
It looks like this:
Capture d’écran 2019-12-15 à 21 20 07

I used a red button instead of the blue from the initial mockup as it was the default bootstrap display and it makes sense for a "delete" action.

matks:
@marionf You will probably want to check it

marionf:
Thank you @matks !
I just ping @TristanLDD to be sure for the color of the delete button

src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Blocks/grid_actions.html.twig
```diff
@@ -67,5 +67,23 @@
           </div>
       {% endblock %}
     {% endembed %}
+
+    {% embed '@PrestaShop/Admin/Helpers/bootstrap_popup.html.twig' with {
+      'id': grid.id ~ '_grid_confirm_deletion_modal',
+      'title': "Delete selection"|trans({}, 'Admin.Global'),
```
LouiseBonnard:
```suggestion
      'title': "Delete selection"|trans({}, 'Admin.Actions'),
```

admin-dev/themes/new-theme/js/components/grid/extension/action/bulk/bulk-delete-with-confirm-modal-extension.js
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Invoked when clicking on the "delete selected items" toolbar button
+   *
+   * @param {Grid} grid
+   *
+   * @private
+   */
+  _onBulkDeleteClick(grid) {
+    const $deleteBtn = $('.js-common_bulk_delete-grid-action');
+    const $bulkDeleteForm = $('#' + grid.getId() + '_filter_form');
```
PierreRambaud:
```suggestion
    const $bulkDeleteForm = $(`#${grid.getId()}_filter_form`);
```

admin-dev/themes/new-theme/js/components/grid/extension/action/bulk/bulk-delete-with-confirm-modal-extension.js
```diff
@@ -0,0 +1,62 @@
...
+
+    $bulkDeleteForm.attr('action', $deleteBtn.data('form-url'));
+    $bulkDeleteForm.attr('method', $deleteBtn.data('form-method'));
+
+    const $modal = $('#' + grid.getId() + '_grid_confirm_deletion_modal');
```
PierreRambaud: same here

admin-dev/themes/new-theme/js/components/grid/extension/action/bulk/bulk-delete-with-confirm-modal-extension.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Invoked when clicking on the "delete selected items" toolbar button
+   *
+   * @param {Grid} grid
+   *
+   * @private
+   */
+  _onBulkDeleteClick(grid) {
```
PierreRambaud:
```suggestion
  onBulkDeleteClick(grid) {
```

admin-dev/themes/new-theme/js/components/grid/extension/action/bulk/bulk-delete-with-confirm-modal-extension.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+export default class BulkDeleteWithModalConfirmExtension {
+
+  /**
+   * Extend grid
+   *
+   * @param {Grid} grid
+   */
+  extend(grid) {
+    grid.getContainer().on('click', '.js-common_bulk_delete-grid-action', () => this._onBulkDeleteClick(grid));
```
PierreRambaud:
```suggestion
    grid.getContainer().on('click', '.js-common_bulk_delete-grid-action', () => this.onBulkDeleteClick(grid));
```

admin-dev/themes/new-theme/js/pages/attachment/index.js
```diff
@@ -33,6 +33,7 @@ import SubmitBulkExtension from '@components/grid/extension/submit-bulk-action-e
 import SubmitRowActionExtension from '@components/grid/extension/action/row/submit-row-action-extension';
 import LinkRowActionExtension from '@components/grid/extension/link-row-action-extension';
 import TranslatableInput from '@components/translatable-input';
+import BulkDeleteWithModalConfirmExtension from '@components/grid/extension/action/bulk/bulk-delete-with-confirm-modal-extension';
```
PierreRambaud:
```suggestion
import BulkDeleteWithModalConfirmExtension 
  from '@components/grid/extension/action/bulk/bulk-delete-with-confirm-modal-extension';
```

src/Core/Grid/Action/Bulk/Type/SubmitBulkDeleteAction.php
```diff
@@ -0,0 +1,59 @@
...
+     * {@inheritdoc}
+     */
+    protected function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver
+            ->setRequired([
+                'submit_route',
+            ])
+            ->setDefaults([
+                'submit_method' => 'POST',
```
PierreRambaud: please use const instead: https://github.com/symfony/http-foundation/blob/master/Request.php#L43
Same for 'GET'

jolelievre:
I think it's a pity to limit this modification to bulk delete It's almost exactly the same behaviour as a submit button except the modal is prettier I suggest making a more generic component to manage the modal I'll propose a PR based on your branch this will be easier to explain

matks:
> I think it's a pity to limit this modification to bulk delete It's almost exactly the same behaviour as a submit button except the modal is prettier I suggest making a more generic component to manage the modal I'll propose a PR based on your branch this will be easier to explain

I thought about it but I'm not sure making it generic is a good idea.

There is actually already 2 other "confirm before bulk delete" modals that already exist: [categories](https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Actions/Bulk/delete_categories.html.twig) and [customers](https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Actions/Bulk/delete_customers.html.twig). They have a few more specific logic but they are definitely "confirm before bulk delete" usecase.

When I started this feature I tried to make something generic, so that we would have a single configurable modal that would fit all usecases. I came close to doing it ... but the code was full of `if`, `options` and generic / badly named variable names. The code was hard to build, hard to read and I expected it to become hard to maintain.

That is why I took the exact reverse path and went for something 100% specific, following KISS principle: keep it simple and stupid.

If you want to make it a configurable modal, you'll probably need to pass, as options:
- selectors to target/listen for jQuery
- routes to call, with parameters and maybe arguments / parameters mapping
- wordings, either already translated or along with their translation domain
- classes to apply for different kind of buttons (in my case I had to apply `text-left font-weight-normal` CSS classes to modal body to make it look correct)
- accept different kind of content for the modal body, including html (such as `<p>` parts), and handle different sizes

... 😅 see what I mean ?

And if you define all these options from the GridDefinition (because I see no else location), it's one more step towards mixing backend logic and view configuration, violating the MVC principle (we already violate it for several items, but this would be one more)

jolelievre:
@matks Mmm, maybe I see it too simple in my mind and don't see the big picture But I thought of something really simple:

jolelievre:
@matks Mmm, maybe I see it too simple in my mind and don't see the big picture But I thought of something really simple:
- a generic modal template that would be present in gridActions template so available for every use case (some are already present)
- modify the SubmitBulkActionExtension js so that instead of using the native `confirm` it uses a modal
- the question is already in the options and would be inserted inside the modal content (just one text)
- we would only need to add an option for title to insert in the modal, it would be optional and we use the question/content as a default title if not present

The url and action already already managed anyway so nothing to add!

Then it wouldn't be that complicated, better it would make the modification for ALL submit actions at once, the only thing to do would be to add the title in each grid definitions (optional but prettier)

What do you think? As I suggested I can try a proposal on a branch based on your so you can see the difference and if it's too complicated

admin-dev/themes/new-theme/js/components/grid/extension/action/bulk/bulk-delete-with-confirm-modal-extension.js
```diff
@@ -0,0 +1,62 @@
...
+   * @param {Grid} grid
+   *
+   * @private
+   */
+  _onBulkDeleteClick(grid) {
+    const $deleteBtn = $('.js-common_bulk_delete-grid-action');
+    const $bulkDeleteForm = $('#' + grid.getId() + '_filter_form');
+
+    $bulkDeleteForm.attr('action', $deleteBtn.data('form-url'));
+    $bulkDeleteForm.attr('method', $deleteBtn.data('form-method'));
```
Progi1984:
```
$bulkDeleteForm.attr({
  action: $deleteBtn.data('form-url'),
  method: $deleteBtn.data('method')
});
```

src/Core/Grid/Action/Bulk/Type/SubmitBulkDeleteAction.php
Progi1984: (and for `submit_method`)

jolelievre:
@marionf I have a question for you, we integrated the possibility to use a modal instead of native `confirm` It can be activated on any bulk action now, with some options:
- modal title
- confirm button label
- confirm button class

Here is how it looks like for a customized one on delete:

Capture d’écran 2019-12-19 à 13 19 53

And the default one would like this:

Capture d’écran 2019-12-19 à 10 52 04

So the question is, do we enable the modal by default on every bulk action? This would enable it on all migrated pages. Or do we keep the native confirm?

marionf:
The idea is to have the same modal to confirm or cancel deletion on all migrated pages
The modal can be used only for delete action (in bulk or for only 1 item)

admin-dev/themes/new-theme/js/components/grid/extension/submit-bulk-action-extension.js
```diff
@@ -57,11 +57,48 @@ export default class SubmitBulkActionExtension {
...
+    const $modal = $('#' + grid.getId() + '_grid_confirm_modal');
+    $('.confirm-message', $modal).html(confirmMessage);
+    $('.modal-title', $modal).html(confirmTitle);
+    $('.modal-header', $modal).toggle(confirmTitle.length > 0);
 
+    const $confirmButton = $('.btn-confirm-submit', $modal);
+    $confirmButton.className = 'btn btn-confirm-submit btn-lg';
+    $confirmButton.addClass(confirmButtonClass);
+    $confirmButton.html(confirmButtonLabel);
+    $confirmButton.off('click').on('click', () => this.postForm($submitBtn, grid));
```
PierreRambaud: Still thiking `.off('click')` isn't a good idea and can break all modules javascript event on the confirm button..

jolelievre: It is required because the same modal can be shared between several submit actions, so we need to remove the previous callback to be sure only one will be called on confirm
If module need to use the same modal then they will have to do the same cleaning

PierreRambaud: A new modal must be created depending on the action you do. You shouldn't share the same modal context, more if you allow people to add module on your app :disappointed_relieved: 
If a module want to add a custom action before the `this.postForm` and just after the first `'click'` event, he must reimplement everything. Sad story :(

jolelievre: Mmm I think you are mixing too many features here for a use case that is much more simple in this case

This modal is simply a "pretty confirm" You can change the question, title, button style and that's it It's not a modal that any other module is supposed to listen to Just like when you use `confirm` in your code no modules will be waiting for the return

We don't want to module to integrate in this modal, since we have absolutely no JS hooks, event, or module integration available today anyway

However what this PR will allow is for module to add their SubmitBulkAction in the grid definition and be able to use a pretty modal only using a PHP hook

PierreRambaud: Even if we don't have hooks, we are able to add custom javascript files in the BO, extend the theme, or anything else. This means you're able to add a custom `.on('click')` on the confirm button to trigger a special api request or a custom effect, a new ajax box loaded.
This is also why we are not able to change all our html ids even if we don't have custom html hooks :trollface:

matks: @PierreRambaud But the `.on('click')` targets `.btn-confirm-submit` selector 🤔so if a module wants to customize the behavior of one button, he can remove the class for it and he's safe, isn't it ?

jolelievre: If a module wants to change the behavior of this button, the correct approach would be to change the grid definition, remove the existing button and add its custom button
Currently if we don't have a modal with the confirm that's what the module would need to do, it wouldn't be able to integrate its js in the middle of the execution

So yes of course any module could binf its actions on clicks, on change on any button, they can do whatever they want and break anything but we can't imagine all the unexpected behavior anyone could imagine or we can never change anything

jolelievre: And indeed as Mathieu said, if one wants to override the default behavior he can always unbind our actions on the submit button

PierreRambaud: This is exactly what I said. If he wants to keep the original behaviour and add a little action just before or after the event. He must reimplement what you've done, this means duplicate the `postForm` method.
@matks It's not targetting the `.btn-confirm-submit`, it targets the `'#GRID_ID_grid_confirm_modal .btn-confirm-submit` which is different.

As I said earlier, you disallow the capability to add an event when you click on the button, I keep thinking it's really bad for front developers :/
Two solutions:
1. Generate the modal dynamically, so no one can add custom event on it
2. One modal per action

Don't forget people can add javascript, and there is no hook / event in the BO for front developers, don't hand them the stick :(

NeOMakinG: I agree with Pierre, let peoples be ables to add events on buttons, don't kick events off, it won't change anything about your behavior no ?

Peoples may wants to add a behavior to this button and not delete your one, if they wants to do this, they need to remove your event, add your own behavior again and then add their behavior, it doesn't look like it's a good practice at all

jolelievre: But I don't remove anything!! Before you add just a `confirm` call, and absolutely NO way to integrate a behaviour before or after this confirm You are mixing things here, I didn't remove the possibility to add any behaviour on the bulk button
This confirm button is a new button in a new modal that no one has ever seen or has ever modified

jolelievre: This modal can be seen as a blackbox modal confirm component, and has no intention of allowing internal customization so indeed I don't want external modules to change its behaviour

admin-dev/themes/new-theme/js/components/grid/extension/submit-bulk-action-extension.js
```diff
@@ -57,11 +59,47 @@ export default class SubmitBulkActionExtension {
...
+   */
+  showConfirmModal($submitBtn, grid, confirmMessage, confirmTitle) {
+    const confirmButtonLabel = $submitBtn.data('confirmButtonLabel');
+    const closeButtonLabel = $submitBtn.data('closeButtonLabel');
+    const confirmButtonClass = $submitBtn.data('confirmButtonClass');
 
+    const modal = new ConfirmModal({
+      id: `${grid.getId()}_grid_confirm_modal`,
+      title: confirmTitle.length > 0 ? confirmTitle : '',
+      message: confirmMessage,
```
jolelievre: I like the quick style like you used for `confirmButtonClass` So maybe you can just rename your const/parameters names so you can do the same with  `message ` `confirmButtonText ` and `closeButtonText ` ?

admin-dev/themes/new-theme/js/components/modal.js
```diff
@@ -0,0 +1,127 @@
...
+ * ConfirmModal component
+ *
+ * @param {String} id
+ * @param {String} title
+ * @param {String} message
+ * @param {String} closeButtonText
+ * @param {String} confirmButtonText
+ * @param {String} confirmButtonClass
+ * @param {Boolean} closable
+ * @param {Function} action
```
jolelievre:
```suggestion
 * @param {Function} confirmCallback
```
So we understand clearly when the callback is called Do you think it could be useful to have a (optional) `cancelCallback` or `closeCallback` as well?

admin-dev/themes/new-theme/js/components/modal.js
```diff
@@ -0,0 +1,127 @@
...
+ *
+ */
+export default function ConfirmModal({id = 'confirm_modal', title = '', message = '', closeButtonText = 'Close', confirmButtonText = 'Accept', confirmButtonClass = 'btn-primary', closable = true}, action) {
+  // Main modal element
```
jolelievre, line 84: Could you place all this code for the modal creation in a dedicated `createModal` function and place it at the end of the file This way we can focus on the modal behaviour earlier when we read the file
Also this creation is quite verbose, since we are bound to use jQuery anyway here maybe we simplify it with something like:
```javascript
this.$modal = $(`<div class="modal fade" id="${id}">
<div class="modal-dialog">....</div>
</div>`);
```
It's easier to read What do you think?

admin-dev/themes/new-theme/js/components/modal.js
```diff
@@ -0,0 +1,127 @@
...
+  this.modalBody.appendChild(this.modalMessage);
+  this.modalFooter.append(this.modalCloseButton, this.modalConfirmButton);
+  this.modalContent.append(this.modalHeader, this.modalBody, this.modalFooter);
+  this.modalDialog.appendChild(this.modalContent);
+  this.modal.appendChild(this.modalDialog);
+
+  // jQuery modal object
+  this.$modal = $(this.modal);
+
+  this.toggle = () => {
```
jolelievre: Is it `toggle` or `show` here? Since the modal is removed right after it's hidden anyway

admin-dev/themes/new-theme/js/components/modal.js
```diff
@@ -0,0 +1,127 @@
...
+
+  // Modal dialog element
+  this.modalDialog = document.createElement('div');
+  this.modalDialog.classList.add('modal-dialog');
+
+  // Modal content element
+  this.modalContent = document.createElement('div');
+  this.modalContent.classList.add('modal-content');
+
+  // Modal header element
```
jolelievre, line 97: Only one feature missing compared to before, if the title is empty the header should be hidden (display: none) or not constructed (I used a toggle based on title.length)
Be cause an empty header is too big and not pretty

src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Actions/Bulk/submit.html.twig
```diff
@@ -29,6 +29,12 @@
         data-form-url="{{ path(action.options.submit_route, action.options.route_params) }}"
         data-form-method="{{ action.options.submit_method }}"
         data-confirm-message="{{ action.options.confirm_message }}"
+        data-close-button-label="{{ 'Close'|trans({}, 'Admin.Actions') }}"
```
jolelievre: I think you don't need `data-close-button-label` unless the modal is used, so this line should in the block with the `{% if action.options.modal_options %}` condition
Besides it's a nice touch to allow modifying this label, do you think you can add this field in the `ModalOptions` php object so that it becomes configurable (you can use `confirm_button_label` as an example)

jolelievre (commented):
Nice work @NeOMakinG This will be useful in some other places I'm sure Just a few feedback to make it even more awesome ;)

jolelievre (approved):

matthieu-rolland (approved):

boubkerbribri:
I just added a pull request #17066 to update test when this PR is merged. please make sure to validate it before merging this one (otherwise, functional test will fail)

admin-dev/themes/new-theme/js/components/modal.js
```diff
@@ -0,0 +1,154 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+const $ = window.$;
```
PierreRambaud, line 26:
```suggestion
const {$} = window;
```

TristanLDD:
I agree for the color of the delete button, it's more safe to have it in red.

admin-dev/themes/new-theme/js/components/modal.js
```diff
@@ -0,0 +1,154 @@
...
+ * @param {String} confirmTitle
+ * @param {String} confirmMessage
+ * @param {String} closeButtonLabel
+ * @param {String} confirmButtonLabel
+ * @param {String} confirmButtonClass
+ * @param {Boolean} closable
+ * @param {Function} confirmCallback
+ *
+ */
+function Modal({id = 'confirm_modal', confirmTitle, confirmMessage = '', closeButtonLabel = 'Close', confirmButtonLabel = 'Accept', confirmButtonClass = 'btn-primary'}) {
```
PierreRambaud, line 83:
```suggestion
function Modal(
  {
    id = 'confirm_modal', 
    confirmTitle, confirmMessage = '', 
    closeButtonLabel = 'Close', 
    confirmButtonLabel = 'Accept', 
    confirmButtonClass = 'btn-primary',
  },
) {
```

admin-dev/themes/new-theme/js/components/modal.js
```diff
@@ -0,0 +1,154 @@
...
+    document.querySelector(`#${id}`).remove();
+  });
+
+  document.body.appendChild(this.modal);
+}
+
+/**
+ * Modal component to improve lisibility by constructing the modal outside the main function
+ *
+ * @param {String} id
```
PierreRambaud, line 73: There is only one param, it's an object

admin-dev/themes/new-theme/js/components/modal.js
```diff
@@ -0,0 +1,154 @@
...
+ * @param {String} confirmButtonLabel
+ * @param {String} confirmButtonClass
+ * @param {Boolean} closable
+ * @param {Function} confirmCallback
+ *
+ */
+export default function ConfirmModal(params, confirmCallback) {
+  // Construct the modal
+  const {id, closable} = params;
+  const modal = new Modal(params);
```
PierreRambaud, line 44:
```suggestion
  this.modal = Modal(params);
```
No need new instruction here

admin-dev/themes/new-theme/js/components/modal.js
```diff
@@ -0,0 +1,154 @@
...
+ * @param {String} closeButtonLabel
+ * @param {String} confirmButtonLabel
+ * @param {String} confirmButtonClass
+ * @param {Boolean} closable
+ * @param {Function} confirmCallback
+ *
+ */
+function Modal({id = 'confirm_modal', confirmTitle, confirmMessage = '', closeButtonLabel = 'Close', confirmButtonLabel = 'Accept', confirmButtonClass = 'btn-primary'}) {
+  // Main modal element
+  this.modal = document.createElement('div');
```
PierreRambaud, line 85: no need to use `this` in this function

admin-dev/themes/new-theme/js/components/modal.js
```diff
@@ -0,0 +1,154 @@
...
+  Object.assign(this, modal);
+
+  // jQuery modal object
+  this.$modal = $(this.modal);
+
+  this.show = () => {
+    this.$modal.modal();
+  };
+
+  this.modalConfirmButton.addEventListener('click', confirmCallback);
```
PierreRambaud, line 54:
```suggestion
  this.modal.confirmation.addEventListener('click', confirmCallback);
```

admin-dev/themes/new-theme/js/components/modal.js
PierreRambaud: Return an object in this function. something like

```js
return {modal: {
  dialog: ...,
  confirmation: ....,
}
```
don't waste time by repeating  `modal`  in variables names

PR #16852, opened 2019-12-18
Fix a bug where changing quantity on product quickview adds products tags on every product

### LINKED ISSUE #9919: [BOOM-3664] Quickview - Related products with combinations change product page after changing a combination
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-3664).

- _**Reporter:**_ stkado
- _**Created at:**_ Thu, 17 Aug 2017 14:46:51 +0200

<p>Hello, i ve the same issue of this ticket <br/>
<a href="http://forge.prestashop.com/browse/BOOM-1878" class="external-link" rel="nofollow">http://forge.prestashop.com/browse/BOOM-1878</a></p>

<p>but here the product page also changes.</p>

<p>When i change quickview attribute i ve 2 POST (cf post.png)</p>
        
- Set product with attributes and related product
<br/>
- go to product page in FO
<br/>
- quickview on the related product
<br/>
- change attribute (S to M for exemple)

### LINKED ISSUE #16633: Quickview and product flags bug
When you are on the main page at the "popular products", and you have one of the products with "pack" label or with "on sale" label active, and you click the "quick view" button for the "pack" or "on sale" labeled product, then use the arrow buttons in quantity to change the quantity to 2, then the "on sale" or the "pack" label is applied to all products in "popular products" incorrectly.

See the screens.

**Screenshots**
scr1

scr2

PrestaShop version: 1.7.6.1

NeOMakinG:
| Questions     | Answers
| Branch?       |  1.7.6.x
| Description?  | When you were on product quickview, it was adding tags of the viewed product on every products on the page
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16633 and Fixes #9919 
| How to test?  | Go on the homepage, and click on quickview on a product with a promo tag for example, then change quantity and see if others product has the tag on the page

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16852)


themes/_core/js/product.js
```diff
@@ -166,13 +166,13 @@ function updateProduct(event, eventType, updateUrl) {
...
-        $('.product-discounts').replaceWith(data.product_discounts);
-        $('.product-additional-info').replaceWith(data.product_additional_info);
-        $('#product-details').replaceWith(data.product_details);
-        $('.product-flags').replaceWith(data.product_flags);
+        $('.quickview .product-prices, .page-product .product-prices').replaceWith(data.product_prices);
+        $('.quickview .product-customization, .page-product .product-customization').replaceWith(data.product_customization);
+        $('.quickview .product-variants .page-product .product-variants').replaceWith(data.product_variants);
+        $('.quickview .product-discounts, .page-product .product-discounts').replaceWith(data.product_discounts);
+        $('.quickview .product-additional-info, .page-product .product-additional-info').replaceWith(data.product_additional_info);
+        $('.quickview #product-details, .page-product #product-details').replaceWith(data.product_details);
```
PierreRambaud: It's an id, it must be unique, no need to target the quickview scope.
```suggestion
        $('#product-details').replaceWith(data.product_details);
```

NeOMakinG:
I should rebuild assets, don't approve yet

themes/_core/js/product.js
```diff
@@ -171,7 +171,7 @@ function updateProduct(event, eventType, updateUrl) {
         $('.quickview .product-variants .page-product .product-variants').replaceWith(data.product_variants);
         $('.quickview .product-discounts, .page-product .product-discounts').replaceWith(data.product_discounts);
         $('.quickview .product-additional-info, .page-product .product-additional-info').replaceWith(data.product_additional_info);
-        $('.quickview #product-details, .page-product #product-details').replaceWith(data.product_details);
+        $('#product-details').replaceWith(data.product_details);
```
PierreRambaud: But if the quickview also add a #product-details block, we are fucked :scream:

NeOMakinG: It doesnt has it

NeOMakinG: and it's ok if it change the modal product details, we are fucked if every product on product list has an id #product-details

PierreRambaud (dismissed):

khouloudbelguith:
Hi @NeOMakinG,

It is ok in the Categories page but if you follow these steps, the issue is reproduced:
1. Create a product
2. Add some related products, which they have some labels, to the last product created
3. Go to the FO => Product details page
4. `YOU MIGHT ALSO LIKE` section, click the "quick view" button for the "pack" or "on sale" labeled related product, 
5. See error

Before
image

After
image

https://drive.google.com/file/d/1XSu23m1bZeDPLysiMoMqpP1T7AvOt2h1/view

Thanks!

themes/_core/js/product.js
```diff
@@ -166,13 +166,14 @@ function updateProduct(event, eventType, updateUrl) {
...
-        $('#product-details').replaceWith(data.product_details);
-        $('.product-flags').replaceWith(data.product_flags);
+
+        $('.quickview .product-prices, .page-product:not(.modal-open) .row .product-prices').replaceWith(data.product_prices);
+        $('.quickview .product-customization, .page-product:not(.modal-open) .row .product-customization').replaceWith(data.product_customization);
+        $('.quickview .product-variants .page-product:not(.modal-open) .row .product-variants').replaceWith(data.product_variants);
+        $('.quickview .product-discounts, .page-product:not(.modal-open) .row .product-discounts').replaceWith(data.product_discounts);
+        $('.quickview .product-additional-info, .page-product:not(.modal-open) .row .product-additional-info').replaceWith(data.product_additional_info);
+        $('.quickview #product-details, #product-details').replaceWith(data.product_details);
+        $('.quickview .product-flags, .page-product:not(.modal-open) .row .product-flags').replaceWith(data.product_flags)
```
arouiadib, line 176: I don't find any element with class `.product-flags` inside the quickview. Consequently, I think this line should only be: `$('.page-product .product-flags').replaceWith(data.product_flags)`.

Same thing apply to `#product-details`, `.product-discounts` and `.product-additionnal-info`, because quickview doesn't contain them.

BUT this will not fully resolve the problem. One case is: when quantity is changed from a quickview inside a product page (when product miniatures exist in product page e.g similar products).

Let's take an example in detail:

- user is in product X page and there are other similar products Y and Z as miniatures in sidebar.
- user clicks in quickview of Y and changes quantity.

In this DOM, there are :
- 2 elements with class `product-variants` , one in quickview of Y and the other in product page X.
- 2 with `.product-prices`
- 1 element with `.product-customization` but for product X
- 1 element with `#product-details` but for product X
- 1 element with `.product-discounts` but for product X
- 1 element with `.product-additionnal-info` but for product X.


IMHO, the whole problem could be solved by re-naming classes (or adding discriminator ones) depending on the source (quickview or product page) and keeping javascript as it is for all classes.

NeOMakinG: I agree for the class not contained in quickview, gonna add a commit for it, but for the rest, did you see that I select elements on product X page only if there are no modals opened? It means that product infos will change only if there are no quickview opened

arouiadib: Thank you for the time you gave to my comment, and I am sorry for wrongly expressing the second point. 

I should have been only focusing on `.product-flags` when giving my example, so let's see an example with this class:

1- Install featured products module and hook it inside product page (footer for example).
2. From product page of product X, click on quickview of product Y( anyone from featured products and not necessarly one with special flags), change quantity...etc
3. Look at product flags of featured products and product X, they change to flags of product Y.

Now, let's keep modal closed and only change quantity for productX => all featured products have their flags changed. 

I agree with you that `:not(.open-modal)` is the solution for the majority of classes, but not all of them.

NeOMakinG: Hey, ok this time it looks like you didn't see the .row selector, featured products and product in the same category are not in a .row. I agree with you, thats not really solid, but atleast it avoid breaking changes, we are not able to change JS selectors for a new class because if we would do that, theme which herit these templates or rewrited these but didn't modify the JS would break with a new class, we're able to do it in a major version probably.

Thank you for your feedback and the time you take on this PR :)

matks (dismissed):

themes/_core/js/product.js
arouiadib: You pointed me to subject I am not always aware of. Thank you very much for such an invaluable info.

NeOMakinG: No problem, you'r welcome, thanks a lot for your feedback there, it was a pleasure !

Robin-Fischer-PS:
Hi @NeOMakinG !

The fix is OK for https://github.com/PrestaShop/PrestaShop/issues/16633, but I still have the bug for https://github.com/PrestaShop/PrestaShop/issues/9919#issuecomment-490441708
Each time I modify in quickview a "You might also like" product (quantity, attribute...), an error is displayed on the product.

NeOMakinG:
The error is now displayed in the right place, but the error is related to something else, submit an issue if this error should be fixed by a backend dev

themes/_core/js/product.js
```diff
@@ -151,10 +151,11 @@ function updateProduct(event, eventType, updateUrl) {
         }
       },
       error(jqXHR, textStatus, errorThrown) {
+        console.log(errorThrown);
```
PierreRambaud: Nope :trollface:

PierreRambaud (approved):

PierreRambaud:
I don't know why PrettyCi is failing when there are only js files edited :sweat_smile:

jolelievre:
Yep, that's weird.. @NeOMakinG maybe you can try to squash all the commits (this will make the PR cleaner) and re-push to relaunch the PrettyCi build

Robin-Fischer-PS:
Hi @NeOMakinG !

So, could you confirm me that the fact this error is still displayed (when changing combination of a "related product") should not be fixed by your PR ? If so, I would suggest to close the issue https://github.com/PrestaShop/PrestaShop/issues/16633 , which is totally fixed by your PR, but to keep https://github.com/PrestaShop/PrestaShop/issues/9919#issuecomment-490441708 open, since it's not totally fixed... WDYT ?

NeOMakinG:
image

We can keep #9919 as it's related to the url called using ajax and not misplacement of ajax calls as this issue should fix, but 16633 is fixed, the error is even displayed on the product it should be displayed now

Robin-Fischer-PS:
Ok ! So since the error of https://github.com/PrestaShop/PrestaShop/issues/9919 is now OK (only displayed on the good product), it's QA approved :)

PierreRambaud:
Thanks @NeOMakinG

Issue #16866, opened 2019-12-19
[HOOK | IMPROVEMENT] Add a actionWebserviceObjectAdd hook


BlackKerio:
I would like to ask you to add a hook to monitor the webservice API. Currently some hooks depend only on frontoffice and backoffice.

Insert one hook into the WebserviceRequestCore class with object parameter and sent xml file for any additional information

It is not necessarily used to override the function

Example: when creating a new customer using the webservice, the customer is not informed by email.

khouloudbelguith:
Hi @BlackKerio,

Thanks for your report.
Ping @PrestaShop/prestashop-product-team @PrestaShop/prestashop-core-developers what do you think?

Thanks!

PR #16867, opened 2019-12-19
[HOOK | IMPROVEMENT] Add a actionWebserviceObjectAdd hook

BlackKerio:
| Questions     | Answers
| Branch?       | develop
| Description?  | 
| Type?         |  improvement
| Category?     | WS
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes # [16866](https://github.com/PrestaShop/PrestaShop/issues/16866)
| How to test?  | 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16867)

### DIFF
--- a/classes/webservice/WebserviceRequest.php
+++ b/classes/webservice/WebserviceRequest.php
@@ -1622,7 +1622,7 @@ class WebserviceRequestCore
                     }
                     $result = $object->{$objectMethod}();
                     if ($result) {
-                        Hook::exec('actionWebserviceObjectAdd', array(
+                     Hook::exec('actionWebserviceObject' . ucfirst($objectMethod), array(
                             'object' => $object,
                             '_POST' => $this->_inputXml,
                         ));
--- a/classes/webservice/WebserviceRequest.php
+++ b/classes/webservice/WebserviceRequest.php
@@ -1622,6 +1622,10 @@ class WebserviceRequestCore
                     }
                     $result = $object->{$objectMethod}();
                     if ($result) {
+                        Hook::exec('actionWebserviceObjectAdd', array(
+                            'object' => $object,
+                            '_POST' => $this->_inputXml
+                        ));
                         if (isset($attributes->associations)) {
                             foreach ($attributes->associations->children() as $association) {
                                 /** @var SimpleXMLElement $association */
--- a/classes/webservice/WebserviceRequest.php
+++ b/classes/webservice/WebserviceRequest.php
@@ -1624,7 +1624,7 @@ class WebserviceRequestCore
                     if ($result) {
                         Hook::exec('actionWebserviceObjectAdd', array(
                             'object' => $object,
-                            '_POST' => $this->_inputXml
+                            '_POST' => $this->_inputXml,
                         ));
                         if (isset($attributes->associations)) {
                             foreach ($attributes->associations->children() as $association) {
--- a/classes/webservice/WebserviceRequest.php
+++ b/classes/webservice/WebserviceRequest.php
@@ -1622,9 +1622,10 @@ class WebserviceRequestCore
                     }
                     $result = $object->{$objectMethod}();
                     if ($result) {
-                     Hook::exec('actionWebserviceObject' . ucfirst($objectMethod), array(
+                        $objectMethod = ($this->method == 'POST' ? 'add' : 'update');
+                        Hook::exec('actionWebserviceObject' . get_class($object) . ucfirst($objectMethod), array(
                             'object' => $object,
-                            '_POST' => $this->_inputXml,
+                            'xml' => $this->_inputXml,
                         ));
                         if (isset($attributes->associations)) {
                             foreach ($attributes->associations->children() as $association) {


classes/webservice/WebserviceRequest.php
```diff
@@ -1622,6 +1622,10 @@ protected function saveEntityFromXml($successReturnCode)
                     }
                     $result = $object->{$objectMethod}();
                     if ($result) {
+                        Hook::exec('actionWebserviceObjectAdd', array(
```
PierreRambaud: There is three problems here for me.
1. the name of the hook is wrong, this can occurred when it's an update
2. what is the interest of having this kind of hook when there are already some in ObjectModel to retrieve the added objects :thinking:  
3. Why calling it `_POST` when it's for `$this->_inputXml`?

BlackKerio: 1. I agree, my fault... name should be without add
2. yes, there are already hooks in the ObjectModel , but I am only interested in changes in webservice with extra parameter $this->_inputXml . For example: there are also hooks such as actionCustomerAccountAdd, actionCustomerAccountUpdate ... instead of using them in the object model actionObjectCustomerAddAfter, actionObjectCustomerUpdateAfter .
3. ok, _inputXml = $this->_inputXml

PierreRambaud: Maybe something like?
1. ```suggestion
                        Hook::exec('actionWebserviceObject' . ucfirst($objectMethod), array(
```
3. `'xml' => $this->_inputXml`

Comment (BlackKerio) [reply classes/webservice/WebserviceRequest.php]: 1. Yes,  unfortunately, objectMethod varies depending on whether there is a special function for Webservice, add or update for any object that can be changed to addWs, updateWs ....

protected $webserviceParameters = array(
         'objectMethods' => array(
             'add' => 'addWs'
             'update' => 'updateWs'
         )

So I suggest to restore objectMethod before calling the hook
$objectMethod = ($this-> method == 'POST' ? 'add' : 'update');

we can leave the name general or add the name of the object 
'actionWebserviceObject' .  get_class($object) . ucfirst($objectMethod)

**Comment by BlackKerio** [comment]
@PierreRambaud is that OK for you ?

File: classes/webservice/WebserviceRequest.php [hunk 2]
```diff
@@ -1622,6 +1622,11 @@ protected function saveEntityFromXml($successReturnCode)
                     }
                     $result = $object->{$objectMethod}();
                     if ($result) {
+                        $objectMethod = ($this->method == 'POST' ? 'add' : 'update');
```
PierreRambaud, line 1625: The objectMethod variable is already defined above :)

BlackKerio: @PierreRambaud Yes, but it changes depending on webserviceParameters, which varies by object, I wrote it to you ...

PierreRambaud: I know, but you repeat the same condition. You maybe need to change the original variable name or reconsider to use the new `$objectMethod` variable 🤔 
Because if the `$objectMethod` is renamed and is now `updateTest` you're not able to properly hook it.
I'm more in favor of using the `$objectMethod` as it is

BlackKerio: @PierreRambaud  Let me summarize the idea.
we create a hook for object monitoring, which was created through webservice independent of internal processing:

1. general name 
result: one hook for all, in function parameters it is possible to find out if hook is called during add or update

2. specific name, we add only 'add' or 'update' to the object name.
result: two hooks, one for 'add' and one for 'update'

3. name completely generated by the object variable,.
result: there will be several hooks by which the names can change depending on the existence of a special function of the webservice object
example: 
the 'customer' object does not currently have any special function => hook 'add' and 'update'
the next time you add a special function 'addWs' and 'updateWs' or others 'xxxWs' => hook 'addWs' and 'updateWs' or 'xxxWs'
and the original hooks call code will not work, the name of hook will change

Which solution do you suggest?
According to my priorities the solution is 2, 1, 3

BlackKerio: Or your solution is 4 ?

Progi1984:
Ping @PierreRambaud

matks:
🆙

classes/webservice/WebserviceRequest.php
PierreRambaud: You should do something like this:
```suggestion
                    // Call alternative method for add/update
                    $event = $objectMethod = ($this->method == 'POST' ? 'add' : 'update');
                    if (!empty($this->resourceConfiguration['objectMethods'][$objectMethod])) {
                        $objectMethod = $this->resourceConfiguration['objectMethods'][$objectMethod];
                    }

                    $result = $object->{$objectMethod}();
                    Hook::exec(
                        'actionWebserviceObject' . get_class($object) . $event,
                        [
                            'object' => $object,
                            'xml' => $this->_inputXml,
                            'result' => $result,
                        ]
                    );
```
And the Hook needs to be outside  the `$result`

BlackKerio: @PierreRambaud I don't understand why you added **$event** = $object Method . If you don't use **$event**. I assume the resulting hook should be. Hook :: exec (
                         'actionWebserviceObject'. get_class ($object). ucfirst (**$event**),
                         [
                             'object' => $ object,
                             'xml' => $ this -> _ inputXml,
                             'result' => $ result,
                         ]
                     );

I agree with this proposal.

PierreRambaud: I update my comment, the $event needs to be on the Hook::exec . Sorry for the delay

BlackKerio: @PierreRambaud in the hook, you deleted function ucfirst ($event), the result is not camelCase

matks:
Hello everyone, I close this PR as it seems there was no activity since 2020. Sorry for the long time it took.

We can reopen if you are motivated to work on this PR again, feel free to ping me 😉

PR #16868, opened 2019-12-19
Show "No picture" image instead of broken one

prestamodule:
| Questions     | Answers
| Branch?       | develop
| Description?  | If product doesn't have cover, a broken image is show.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Add a product without cover image into cart and look at cart details

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16868)

### DIFF
--- a/themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl
+++ b/themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl
@@ -26,7 +26,11 @@
   <!--  product line left content: image-->
   <div class="product-line-grid-left col-md-3 col-xs-4">
     <span class="product-image media-middle">
-      <img src="{$product.cover.bySize.cart_default.url}" alt="{$product.name|escape:'quotes'}">
+      {if $product.cover}
+        <img src="{$product.cover.bySize.cart_default.url}" alt="{$product.name|escape:'quotes'}">
+      {else}
+        <img src="{$urls.no_picture_image.bySize.home_default.url}" />
+      {/if}
     </span>
   </div>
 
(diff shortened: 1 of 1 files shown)


themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl
```diff
@@ -26,7 +26,11 @@
   <!--  product line left content: image-->
   <div class="product-line-grid-left col-md-3 col-xs-4">
     <span class="product-image media-middle">
-      <img src="{$product.cover.bySize.cart_default.url}" alt="{$product.name|escape:'quotes'}">
+      {if $product.cover}
+        <img src="{$product.cover.bySize.cart_default.url}" alt="{$product.name|escape:'quotes'}">
+      {else}
+        <img src="{$urls.no_picture_image.bySize.home_default.url}" />
```
micka-fdz: Hello, I think you should use `` instead.

matks (approved):

matks:
Thanks @prestamodule for the nice PR. 

I validate this PR but I think we might even improve it further later 🤔because with this PR cart now looks like this:
Capture d’écran 2019-12-20 à 23 15 10

@TristanLDD Could we have a "no picture available" image that is more relevant for such views ? Right now it uses the default one that suits better a page where it's shown bigger.

themes/classic/templates/checkout/_partials/cart-detailed-product-line.tpl
matks: Thanks for the suggestion 😉

PR #16712, opened 2019-12-05
Add unit test for profile ID test

gavinkalika:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop 
| Description?  | Please be specific when describing the PR. <br> Every detail helps: versions, browser/server configuration, specific module/theme, etc.
| Type?         | improvement
| Category?     | TE
| BC breaks?    | NO
| Deprecations? | NO
| Fixed ticket? | [Issue](https://github.com/PrestaShop/PrestaShop/issues), please write "Fixes #[issue number]" here.
| How to test?  | Please indicate how to best verify that this PR is correct.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16028)

PR #16902, opened 2019-12-23
Modify ProfileId so that it accepts both int and string type as input

### LINKED ISSUE #16899: Exception displayed when creating new profile
**Describe the bug**
In page 'Advanced Parameters > Team > profile', when creating a new profile, an exception is displayed

**To Reproduce**
1. Go to 'Advanced Parameters > Team > profile'
2. Click on 'Add new profile'
3. Fill form and click on save'

**Screenshots**
image

**Additional information**
PrestaShop version: develop
PHP version: 7.2

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Modify ProfileId so that it accepts both int and string type as input but still validates them correctly
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16899
| How to test?  | See issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16902)

### DIFF
--- a/src/Core/Domain/Profile/ValueObject/ProfileId.php
+++ b/src/Core/Domain/Profile/ValueObject/ProfileId.php
@@ -43,6 +43,11 @@ class ProfileId
      */
     public function __construct($profileId)
     {
+        // Strict type should be used in next major
+        if (!is_int($profileId)) {
+            @trigger_error('Invalid type, int is expected', E_STRICT);
+        }
+
         $this->setProfileId($profileId);
     }
 
@@ -59,10 +64,12 @@ class ProfileId
      */
     private function setProfileId($profileId)
     {
-        if (!is_int($profileId) || 0 >= $profileId) {
-            throw new ProfileException(sprintf('Invalid Profile id %s supplied', var_export($profileId, true)));
+        if ((!is_int($profileId) && !ctype_digit($profileId)) || 0 >= $profileId) {
+            throw new ProfileException(
+                sprintf('Invalid Profile id %s supplied', var_export($profileId, true))
+            );
         }
 
-        $this->profileId = $profileId;
+        $this->profileId = (int) $profileId;
     }
 }
--- a/src/Adapter/Profile/CommandHandler/AddProfileHandler.php
+++ b/src/Adapter/Profile/CommandHandler/AddProfileHandler.php
@@ -53,6 +53,6 @@ final class AddProfileHandler implements AddProfileHandlerInterface
             throw new ProfileException('Failed to create Profile');
         }
 
-        return new ProfileId($profile->id);
+        return new ProfileId((int) $profile->id);
     }
 }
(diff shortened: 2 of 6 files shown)


jolelievre:
Hi @gavinkalika I just modified ProfileId a bit sot hat it accepts string input because this caused bug detected in our functional tests.
Although I would have preferred a stronger type checking (even explicit type would have been better), since the class was introduced in 1.7.6 we must keep string as allowed input to avoid BC.

jolelievre:
The previous message was a reference to this modification https://github.com/PrestaShop/PrestaShop/pull/16712

atomiix (dismissed):

src/Core/Domain/Profile/ValueObject/ProfileId.php
```diff
@@ -59,12 +59,12 @@ public function getValue()
...
     {
-        if (!is_int($profileId) || 0 >= $profileId) {
+        if ((!is_int($profileId) && !ctype_digit($profileId)) || 0 >= $profileId) {
             throw new ProfileException(
                 sprintf('Invalid Profile id %s supplied', var_export($profileId, true))
             );
         }
 
-        $this->profileId = $profileId;
+        $this->profileId = (int) $profileId;
```
eternoendless, line 73: This goes against strict typing. It's up to the consumer to send the right type.
 
Since this class has already been released with mixed type support and making it stricter would be a BC, we _can_ adapt it but in that case it **must** issue a weak error when receiving a type different than int  (same as we do with deprecations) so that we can change the interface later.

eternoendless (changes requested):

gavinkalika:
thanks @jolelievre  and sorry about 

I figure those functional test would run during the MR pre merge build.

To prevent this issue, next time I should run **composer test-all** ?

src/Core/Domain/Profile/ValueObject/ProfileId.php
jolelievre: @eternoendless I agree sadly we can't add strict typing here because the class was already released in 176. That's why we have so many tests to validate the type... Since we can't validate the strict type we can only validate that the context is a digit
I added the warning logs as you suggested though

jolelievre:
@gavinkalika No problem, you couldn't detect it with the travis tests
Here are information about how to launch functional tests https://github.com/PrestaShop/PrestaShop/tree/develop/tests/puppeteer#prestashop-tests-with-puppeteer

However they are quite long, which is why we can't launch all the tests for each pull request, only sanity tests are launched by travis (`npm run sanity-travis`) But every night (during nightly build) the full function test suite is launched (`npm run functional-tests`)

matks:
@eternoendless can you confirm the PR is valid now that it was changed ?

marionf:
@jolelievre Could you rebase on 1.7.7.x as you are fixing a regression ?

eternoendless (approved):

jolelievre:
Closed in favor of https://github.com/PrestaShop/PrestaShop/pull/17331 for incorrect rebase reasons

PR #16906, opened 2019-12-23
Handle profile name above 32 chars exception

### LINKED ISSUE #16725: Error - Unclear alert when adding a profile's name more than 32 chars
**Describe the bug**
 Unclear alert when adding a profile's name more than 32 chars

**To Reproduce**
1. Go to BO => Advanced Parameters => Team => Profiles
2. Create a profile with a name more than 32 chars
3. Save

With PS1.7.5.2, it was very clear
image

**Screenshots**
image

**Additional information**
PrestaShop version: 1.7.6.2, 1.7.6.0
PHP version: 7.1

atomiix:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | This handles exception when saving a profile name that is longer that 32 chars
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16725
| How to test?  | In _Advanced Parameters -> Team -> Profiles_ try adding a new profile with a name longer that 32 chars. Same thing when editing a profile name.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16906)

### DIFF
--- a/src/Core/Domain/Profile/Command/AddProfileCommand.php
+++ b/src/Core/Domain/Profile/Command/AddProfileCommand.php
@@ -27,6 +27,7 @@
 namespace PrestaShop\PrestaShop\Core\Domain\Profile\Command;
 
 use PrestaShop\PrestaShop\Core\Domain\Profile\Exception\ProfileConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Profile\ProfileSettings;
 
 /**
  * Adds new profile
@@ -64,11 +65,11 @@ class AddProfileCommand
      */
     private function assertNameIsStringAndRequiredLength($name)
     {
-        if (!is_string($name) || strlen($name) !== ProfileConstraintException::NAME_MAX_LENGTH) {
+        if (null !== $name && !is_string($name) || strlen($name) > ProfileSettings::NAME_MAX_LENGTH) {
             throw new ProfileConstraintException(
                 sprintf(
                     'Profile name should not exceed %d characters length but %s given',
-                    ProfileConstraintException::NAME_MAX_LENGTH,
+                    ProfileSettings::NAME_MAX_LENGTH,
                     var_export($name, true)
                 ),
                 ProfileConstraintException::INVALID_NAME
--- a/src/Core/Domain/Profile/Command/AddProfileCommand.php
+++ b/src/Core/Domain/Profile/Command/AddProfileCommand.php
@@ -26,54 +26,9 @@
 
 namespace PrestaShop\PrestaShop\Core\Domain\Profile\Command;
 
-use PrestaShop\PrestaShop\Core\Domain\Profile\Exception\ProfileConstraintException;
-use PrestaShop\PrestaShop\Core\Domain\Profile\ProfileSettings;
-
 /**
  * Adds new profile
  */
-class AddProfileCommand
+class AddProfileCommand extends AbstractProfileCommand
 {
-    /**
-     * @var string[] As langId => name
-     */
-    private $localizedNames;
-
-    /**
-     * @param string[] $localizedNames
-     *
-     * @throws ProfileConstraintException
-     */
-    public function __construct(array $localizedNames)
-    {
-        foreach ($localizedNames as $localizedName) {
-            $this->assertNameIsStringAndRequiredLength($localizedName);
-        }
-        $this->localizedNames = $localizedNames;
-    }
-
-    /**
-     * @return string[]
-     */
-    public function getLocalizedNames()
-    {
-        return $this->localizedNames;
-    }
-
-    /**
-     * @param string $name
-     */
-    private function assertNameIsStringAndRequiredLength($name)
-    {
-        if (null !== $name && !is_string($name) || strlen($name) > ProfileSettings::NAME_MAX_LENGTH) {
-            throw new ProfileConstraintException(
-                sprintf(
-                    'Profile name should not exceed %d characters length but %s given',
-                    ProfileSettings::NAME_MAX_LENGTH,
-                    var_export($name, true)
-                ),
-                ProfileConstraintException::INVALID_NAME
-            );
-        }
-    }
 }
--- a/src/Core/Domain/Profile/Command/AddProfileCommand.php
+++ b/src/Core/Domain/Profile/Command/AddProfileCommand.php
@@ -26,6 +26,8 @@
 
 namespace PrestaShop\PrestaShop\Core\Domain\Profile\Command;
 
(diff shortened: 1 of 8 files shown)


src/Core/Domain/Profile/Exception/ProfileConstraintException.php
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Is thrown when some constraint is violated in Profile subdomain
+ */
+class ProfileConstraintException extends ProfileException
+{
+    /**
+     * Profile name max length as defined in the ObjectModel
+     */
+    const NAME_MAX_LENGTH = 32;
```
eternoendless: I think it would make more sense to put this in AddProfileCommand

atomiix: Yes, I thought about it but since we need to check the length in `EditProfileCommand` too, I thought the common place would have been the Exception linked to that check.
So should I use `AddProfileCommand::MAX_LENGTH` in `EditProfileCommand` or should I add a second constante in `EditProfileCommand`?

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/ProfileController.php
```diff
@@ -264,6 +265,13 @@ public function bulkDeleteAction(Request $request)
     protected function getErrorMessages()
     {
         return [
+            ProfileConstraintException::class => [
+                ProfileConstraintException::INVALID_NAME => $this->trans(
+                    'Your entry in field name exceeds max length %d chars (incl. HTML tags)',
```
LouiseBonnard: It works for both first and last name fields, right? In this case, I would write _Your entry in the %s field cannot be longer than %limit% characters (incl. HTML tags)_, localized in the same domain.

atomiix: It's not a person name, it's a name we give to a profile, like `SuperAdmin`, `Logistician`, `Salesman`, etc... But if you still think I should change it I'll do it! Wdyt?

LouiseBonnard: My bad, I was on the employee tab... What about _This field cannot be longer than %limit% characters (incl. HTML tags)_ in Admin.Notifications.Error?

src/Core/Domain/Profile/Exception/ProfileConstraintException.php
matks: For this kind of constants, we have also been using "Settings" class
See https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Domain/Category/SeoSettings.php

src/Core/Domain/Profile/Command/AddProfileCommand.php
```diff
@@ -51,4 +58,21 @@ public function getLocalizedNames()
     {
         return $this->localizedNames;
     }
+
+    /**
+     * @param string $name
+     */
+    private function assertNameIsStringAndRequiredLength($name)
+    {
+        if (!is_string($name) || strlen($name) !== ProfileConstraintException::NAME_MAX_LENGTH) {
```
jolelievre:
```suggestion
        if (!is_string($name) || strlen($name) > ProfileConstraintException::NAME_MAX_LENGTH) {
```

src/Core/Domain/Profile/Command/EditProfileCommand.php
```diff
@@ -75,4 +80,21 @@ public function getLocalizedNames()
     {
         return $this->localizedNames;
     }
+
+    /**
+     * @param string $name
+     */
+    private function assertNameIsStringAndRequiredLength($name)
+    {
+        if (!is_string($name) || strlen($name) !== ProfileConstraintException::NAME_MAX_LENGTH) {
```
jolelievre:
```suggestion
        if (!is_string($name) || strlen($name) > ProfileConstraintException::NAME_MAX_LENGTH) {
```

src/Core/Domain/Profile/Exception/ProfileConstraintException.php
jolelievre: Agreed! This kind of domain specification should not be stored in the exception

jolelievre (changes requested):
Careful your condition is a strict equality It should check the string max length instead

src/Core/Domain/Profile/Command/AddProfileCommand.php
```diff
@@ -51,4 +59,21 @@ public function getLocalizedNames()
     {
         return $this->localizedNames;
     }
+
+    /**
+     * @param string $name
+     */
+    private function assertNameIsStringAndRequiredLength($name)
+    {
+        if (null !== $name && !is_string($name) || strlen($name) > ProfileSettings::NAME_MAX_LENGTH) {
```
PierreRambaud:
```suggestion
        if (!is_string($name) || strlen($name) > ProfileSettings::NAME_MAX_LENGTH) {
```
No need to check for null if you check for is_string & strlen :)

src/Core/Domain/Profile/Command/EditProfileCommand.php
```diff
@@ -75,4 +81,21 @@ public function getLocalizedNames()
     {
         return $this->localizedNames;
     }
+
+    /**
+     * @param string $name
+     */
+    private function assertNameIsStringAndRequiredLength($name)
```
PierreRambaud: Duplicate code, should we provide an abstract method?

src/Core/Domain/Profile/Command/AddProfileCommand.php
atomiix: For TranslatableField type, empty strings are sent as null so I think we need this test

PierreRambaud: Not needed `is_string(null) => false`

atomiix: Yes, that's the problem, we should allow the value null

PierreRambaud: Ooooooooooohhhhhhhhhhhhhhhhhhh :sweat_smile:

jolelievre (approved):

PierreRambaud (approved):

matks:
Thank you @atomiix and thank you everybody for validation

PR #16940, opened 2019-12-27
Functional tests - Add init pages file for basic pages

boubkerbribri:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add init pages file for basic pages
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | `TEST_PATH="sanity/02_productsBO/02_CRUDStandardProductInBO.js" URL_FO=Shop_Url npm run specific-test`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16940)


tests/puppeteer/campaigns/utils/initPages.js
```diff
@@ -0,0 +1,47 @@
...
+        homePage: new HomePage(page),
+      };
+      break;
+    case 'Install':
+      baseObject = {
+        installPage: new InstallPage(page),
+        homePage: new HomePage(page),
+      };
+      break;
+    case 'BO + FO':
```
SimonGrn:
```suggestion
    case 'BO+FO':
```
Maybe avoid spaces in arguments ?

NeOMakinG: Maybe we could call it something else too, like BFO, it looks a bit touchy to write, if someone adds a missy space or remove one

tests/puppeteer/campaigns/utils/initPages.js
```diff
@@ -0,0 +1,47 @@
+require('module-alias/register');
+const BOBasePage = require('@pages/BO/BObasePage');
+const FOBasePage = require('@pages/FO/FObasePage');
+const LoginPage = require('@pages/BO/login');
+const DashboardPage = require('@pages/BO/dashboard');
+const HomePage = require('@pages/FO/home');
+const InstallPage = require('@pages/Install/install');
+
+module.exports =  async function initBasicPages(context = 'BO', pagesObjects, page) {
+  let baseObject;
```
NeOMakinG:
```suggestion
  let baseObject;
  
```

I love to get some spaces in my code, it increase readability as hell, before boucles, conditions...

tests/puppeteer/campaigns/utils/initPages.js
```diff
@@ -0,0 +1,47 @@
+require('module-alias/register');
+const BOBasePage = require('@pages/BO/BObasePage');
+const FOBasePage = require('@pages/FO/FObasePage');
+const LoginPage = require('@pages/BO/login');
+const DashboardPage = require('@pages/BO/dashboard');
+const HomePage = require('@pages/FO/home');
+const InstallPage = require('@pages/Install/install');
+
+module.exports =  async function initBasicPages(context = 'BO', pagesObjects, page) {
```
NeOMakinG: page is required to use this function, so context and pagesObjects are by default required too, I think context doesn't need a default value, btw you can use object destructuring if you want to let peoples the liberty to use this function without calling a context

That would be something like this :

```suggestion
module.exports =  async function initBasicPages({context = 'BO', pagesObjects = {}, page}) {
```

And that would be usable like this :

```
let object = init({page});
```

and with context and adding new pages :
```
let object = init({page, context: 'BO', pagesObjects});
```

tests/puppeteer/campaigns/sanity/02_productsBO/02_CRUDStandardProductInBO.js
```diff
@@ -20,14 +18,12 @@ let editedProductData;
 
 // creating pages objects in a function
 const init = async function () {
```
NeOMakinG: I would completly delete this function and directly use initBasicPages with an object as an argument

tests/puppeteer/campaigns/sanity/02_productsBO/02_CRUDStandardProductInBO.js
```diff
@@ -2,12 +2,10 @@ require('module-alias/register');
 // Using chai
 const {expect} = require('chai');
 const helper = require('@utils/helpers');
+const initBasicPages = require('@utils/initPages');
```
NeOMakinG: Bad naming, it's not only basicPages but whole pages, this should be called initPages

```suggestion
const initPages = require('@utils/initPages');
```

NeOMakinG (changes requested):

NeOMakinG:
ping @PierreRambaud what do you think about this ? Even if the object instanciation looks crappy, its a big gain on every test files, winning around 7 lines by test files and also centralizing basic testing contexts in case they would change in some years.

tests/puppeteer/campaigns/utils/init.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  Object.keys(pagesObjects).forEach((item) => {
+    const returnItem = {
+      [item]: new pagesObjects[item](page),
+    };
+    returnItems = {
+      ...returnItems,
+      ...returnItem,
+    };
```
NeOMakinG, line 52:
```suggestion
    Object.assign(returnItems, returnItem);
```

NeOMakinG: Instead of 
`returnItems = {
      ...returnItems,
      ...returnItem,
    };`

NeOMakinG: You can also do a .map in a var and object.assign the var which contains the map datas

Issue #16482, opened 2019-11-20
[EPIC] Symfony Forms simplification using PrestaShop Form Theme


matks:
## Documentation

We need to document

- it is now possible to use `form_widget(form)` to render whole symfony forms
- explain why `form_widget()` is so much better (especially the fact it can be used to completely reorder/modify template rendering without template override)
- explain how tu use CustomContentType, when to use it, when not to use it

## Uniformize rendering of all forms

And we need to modify the twig templates for all Symfony forms to use a single `form_widget()` call, instead of rendering each widget in a dedicated manner.

## Progress monitoring (last update: 07/05/2021)

- 50 pages with forms to simplify
- 47 done ✅ 
- 3 WIP

### PRs left to be merged

https://github.com/PrestaShop/PrestaShop/pull/20243 Invoices form
https://github.com/PrestaShop/PrestaShop/pull/22472 Payment preferences

### Full roadmap

Page | Status | PR | Refactored in version
-------|--|--------|---
Advanced Parameters > Performance                        |✅ | https://github.com/PrestaShop/PrestaShop/pull/16964 | 1.7.8.0
Shop parameters > General > General                      |✅ | https://github.com/PrestaShop/PrestaShop/pull/21219 | 1.7.8.0
Advanced Parameters > Administration                     |✅ | https://github.com/PrestaShop/PrestaShop/pull/21243 | 1.7.8.0
Advanced Parameters > Import (1/2 pages)                 |✅ | https://github.com/PrestaShop/PrestaShop/pull/21245 | 1.7.8.0
Shop parameters > General > Maintenance                  |✅ | https://github.com/PrestaShop/PrestaShop/pull/22379/ | 1.7.8.0
Shop parameters > Product settings                       |✅ | https://github.com/PrestaShop/PrestaShop/pull/27892 | 8.0.0
Shop parameters > Customer settings > Customers          |✅  | https://github.com/PrestaShop/PrestaShop/pull/21234 | 1.7.8.0
Advanced Parameters > Logs                               |✅  | https://github.com/PrestaShop/PrestaShop/pull/21075 | 1.7.8.0
Orders > Invoices                                       |✅ | https://github.com/PrestaShop/PrestaShop/pull/39203 | 9.0.1
Orders > Delivery Slips                                 |✅  | https://github.com/PrestaShop/PrestaShop/pull/27851  | 8.0.0
[...]

PR #16964, opened 2019-12-30
Refactor UI kit form theme and apply it to the Performance page

eternoendless:
| Questions     | Answers
| Branch?       | develop
| Description?  | Read below
| Type?         | improvement
| Category?     | BO
| Deprecations? | no
| Fixed ticket? | Partially, #16482
| How to test?  | The "Advanced Parameters" > Performance page should work as before. All other pages should be untouched.

## What this change is about

This change does two main things:

1. It declares a new form theme named "prestashop_ui_kit", based on the UI kit and Symfony's Bootstrap 4 theme.
2. It updates the "Advanced Parameters" > "Performance" page to render each form in one shot using `form_widget()` instead of individual fields.

## Why I did this

When I started exploring the changes needed for #16482, I realized that we should have used `form_widget()` instead of `form_rest()` to render forms. However, I found that rendering a form using `form_widget()` yielded a different result than doing it through `form_rest()` (and that shouldn't happen).

It looks like during the early days of 1.7, Symfony's original form theme files ware copied over to `src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/` instead of using them as a base. This form theme is composed of several files, some of them override each other:

- bootstrap_4_horizontal_layout
- bootstrap_4_layout
- form_div_layout
- typeahead
- material

In time, these files were modified time and again, resulting in an unmaintainable mess that produced the problem explained above.

So I started a new form theme called `prestashop_ui_kit` from scratch, using Symfony's own form theme as a base (located in `/vendor/symfony/symfony/src/Symfony/Bridge/Twig/Resources/views/Form/bootstrap_4_horizontal_layout.html.twig`), applying changes from the previously used form theme as overrides, until I got what I was looking for.

## How it looks

Before:
before

After:
after

## The way forward

In order to minimize risk of side effects and allow for quicker integration, I didn't replace the previous form theme yet. Each updated page should explicitly use the new form theme by adding this line in its template:

```twig
{% form_theme form 'PrestaShopBundle:Admin/TwigTemplateForm:prestashop_ui_kit.html.twig' %}
```

Once all pages have been updated, we can safely change the default in `config.yml` and remove the old files.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16964)


src/PrestaShopBundle/Resources/views/Admin/Configure/AdvancedParameters/memcache_servers.html.twig
```diff
@@ -36,31 +36,7 @@
...
-    <div class="form-group row">
-        <label class="form-control-label">{{ 'Weight'|trans }}</label>
-        <div class="col-sm">
-          {{ form_errors(form.memcache_weight) }}
-          {{ form_widget(form.memcache_weight) }}
-        </div>
-    </div>
-
-    {{ form_rest(form) }}
+    {{ form_widget(form) }}
```
matks, line 39: Is it a good practice to use form_widget on the whole form ? I know it works but I never understood why 😅.

IMO it feels more logical to have:
```
form_start(form)

// specific rendering of the widgets you want
form_widget(field1)
form_widget(field2)
form_widget(field3)

// "misc" rendering of whatever is left
form_rest(form)
```

It's even in the name of the function: form_widget aims to render a single widget, form_rest aims to render whatever has not been rendered left (= the rest) 😅?

matks:
So will we deprecate the other form themes ?

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/PerformanceController.php
```diff
@@ -44,12 +44,11 @@ class PerformanceController extends FrameworkBundleAdminController
     /**
      * Displays the Performance main page.
      *
-     * @Template("@PrestaShop/Admin/Configure/AdvancedParameters/performance.html.twig")
```
mickaelandrieu: 👍

mickaelandrieu, line 47: 👍

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/PerformanceController.php
```diff
@@ -44,12 +44,11 @@ class PerformanceController extends FrameworkBundleAdminController
...
      * @AdminSecurity("is_granted('read', request.get('_legacy_controller'))", message="Access denied.")
      *
      * @param FormInterface $form
      *
-     * @return array
+     * @return Response
```
mickaelandrieu, line 51: can we imagine to force the type hint instead ?

```diff
- public function indexAction(FormInterface $form = null)
+ public function indexAction(FormInterface $form = null) : Response
```

src/PrestaShopBundle/Form/Admin/AdvancedParameters/Performance/CachingType.php
```diff
@@ -50,8 +50,11 @@ class CachingType extends TranslatorAwareType
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
-            ->add('use_cache', SwitchType::class)
+            ->add('use_cache', SwitchType::class, [
```
mickaelandrieu, line 52: you should disable the translation of the form component: otherwise the translation will be done twice => https://symfony.com/doc/3.4/reference/forms/types/form.html#translation-domain

you can control the behavior in the Translations profiler (this is something you can actually almost "unit test").

src/PrestaShopBundle/Form/Admin/AdvancedParameters/Performance/CachingType.php
```diff
@@ -85,8 +88,10 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                     return $disabled === true ? ['disabled' => $disabled] : [];
                 },
                 'expanded' => true,
-                'required' => false,
                 'placeholder' => false,
+                'row_attr' => [
```
mickaelandrieu, line 92: ~~this option doesn't exists in Symfony 3, is it something PrestaShop have backported into its own form theme ?~~ ok  :)

mickaelandrieu (commented):
This will make the enrichment of this page easier for sure, I hope we can learn something here to dramatically improve the Product form page 👍

src/PrestaShopBundle/Form/Admin/Extension/RowAttributesExtension.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        $view->vars['row_attr'] = isset($options['row_attr']) ? $options['row_attr'] : [];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getExtendedType()
+    {
+        return FormType::class;
```
mickaelandrieu, line 74: maybe ``ChoiceType`` is enough here as you probably don't want to add this option to every Form Type ?

src/PrestaShopBundle/Resources/config/services/bundle/form/form_extension.yml
```diff
@@ -21,3 +21,9 @@ services:
         class: 'PrestaShopBundle\Form\Extension\DataListExtension'
         tags:
             - { name: form.type_extension, extended_type: Symfony\Component\Form\Extension\Core\Type\TextType }
+
+    form.extension.row_attributes:
+      class: 'PrestaShopBundle\Form\Admin\Extension\RowAttributesExtension'
+      tags:
+        - { name: form.type_extension, extended_type: Symfony\Component\Form\Extension\Core\Type\FormType }
```
mickaelandrieu, line 28: to be updated here, too

src/PrestaShopBundle/Resources/views/Admin/Configure/AdvancedParameters/memcache_servers.html.twig
eternoendless: `form_widget` is used to render both the form and its widgets (see [form_widget documentation](https://symfony.com/doc/3.4/reference/forms/twig_reference.html#form-widget-view-variables)).

Moreover, check out what `form()` does: [source code](https://github.com/symfony/twig-bridge/blob/master/Resources/views/Form/form_div_layout.html.twig#L359-L363)

`form_rest` is clearly meant to be used when coupled with several `form_row` in case you forget to render a field.

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/PerformanceController.php
eternoendless: I was convinced this was a breaking change, but apparently it's possible!

src/PrestaShopBundle/Form/Admin/Extension/RowAttributesExtension.php
eternoendless: This extension allows you to add a class to the form row, so it could be applied to any type. Thanks for the suggestion though!

src/PrestaShopBundle/Form/Admin/AdvancedParameters/Performance/CachingType.php
eternoendless: Good tip 👍

src/PrestaShopBundle/Resources/views/Admin/Configure/AdvancedParameters/memcache_servers.html.twig
matks: OK, you convinced me 👍

admin-dev/themes/default/js/bundle/admin_parameters/performancePageUI.js
```diff
@@ -46,7 +46,7 @@ var PerformancePageUI = {
 
         if(cacheEnabledInput.value === CACHE_ENABLED) {
             for (var i = 0; i < cachingElements.length; i++) {
-                cachingElements[i].style.display = "block";
+                cachingElements[i].style.display = "";
```
PierreRambaud, line 49: Is this block still needed?

eternoendless: Yep, it's used for toggling related options visible/invisible depending of the chosen option. Ofc this script should be moved to the new theme, but that's a subject for another PR.

PierreRambaud (dismissed):

sarahdib:
@eternoendless 
When I save a change an the performance page, there is a message that server media need to be set up : 
https://drive.google.com/file/d/1br_xqOt67URxhQ8CR19orjyM7IHULmIT/view

eternoendless:
@sarahdib Fixed!

PierreRambaud (approved):

Progi1984:
Thanks @eternoendless

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/prestashop_ui_kit.html.twig
```diff
@@ -22,4 +22,107 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  *#}
-{% use '@PrestaShop/Admin/TwigTemplateForm/bootstrap_4_horizontal_layout.html.twig' %}
+{% use 'bootstrap_4_horizontal_layout.html.twig' %}
```
tdavidsonas88: @eternoendless  As this one is not used because of the wrong path: bootstrap_4_horizontal_layout.html.twig - can we delete this line?

tdavidsonas88, line 25: @eternoendless on my phpstorm setup it expects full path for 'bootstrap_4_horizontal_layout.html.twig'. Better to modify phpstorm settings or add a full path here?

tdavidsonas88: Just a thought that as we already have it: https://github.com/PrestaShop/PrestaShop/blob/f05a93b4434388fbeceb149331524bc3918574e2/app/config/config.yml#L55-L61

so, maybe we can remove bootstrap_4_horizontal_layout import here?

eternoendless: Actually leaving the path relative in prestashop_ui_kit will actually make it use the `bootstrap_4_horizontal_layout.html.twig` file from Symfony instead of the one from PrestaShop. I didn't find any way to make it explicit :\

We need to keep the default form_theme like that until all forms have been migrated to the new form theme. Once that's done, we can switch the default and remove the old files, as I said in the PR description.

PR #17009, opened 2020-01-02
Functional tests - add test 'Preview Email Themes'

boubkerbribri:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add new functional test : **Preview Email Themes**
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | `TEST_PATH="functional/BO/design/emailThemes/01_previewEmailThemes" URL_FO=Shop_Url npm run specific-test`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17009)

### DIFF
--- a/tests/puppeteer/pages/BO/BObasePage.js
+++ b/tests/puppeteer/pages/BO/BObasePage.js
@@ -64,6 +64,8 @@ module.exports = class BOBasePage extends CommonPage {
 
     // Design
     this.designParentLink = '#subtab-AdminParentThemes';
+    // Email theme
+    this.emailThemeLink = '#subtab-AdminMailThemeParent';
     // Pages
     this.pagesLink = '#subtab-AdminCmsContent';
     // Link widget
--- a/tests/puppeteer/pages/BO/design/emailThemes/index.js
+++ b/tests/puppeteer/pages/BO/design/emailThemes/index.js
@@ -1,0 +1,45 @@
+require('module-alias/register');
+const BOBasePage = require('@pages/BO/BObasePage');
+
+module.exports = class EmailThemes extends BOBasePage {
+  constructor(page) {
+    super(page);
+
+    this.pageTitle = 'Email Theme •';
+
+    // Selectors
+    this.blockDiv = 'div.content-div div.justify-content-center:nth-child(%ID)';
+    // Email Theme table selectors
+    this.emailThemeTable = `${this.blockDiv.replace('%ID', 5)} table.grid-table`;
+    this.tableBody = `${this.emailThemeTable} tbody`;
+    this.tableRows = `${this.tableBody} tr`;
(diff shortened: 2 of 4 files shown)


tests/puppeteer/pages/BO/BObasePage.js
```diff
@@ -64,6 +64,8 @@ module.exports = class BOBasePage extends CommonPage {
 
     // Design
     this.designParentLink = '#subtab-AdminParentThemes';
+    // Email theme
+    this.emailThemeLink = '#subtab-AdminMailThemeParent';
```
jolelievre, line 68: There is a pending PR about this https://github.com/PrestaShop/PrestaShop/pull/16696
`AdminMailThemeParent ` is gonna be renamed into `AdminParentMailTheme`
Do you prefer to merge this one as is and fix it once the bugfix PR will be merged, or wait for the proper name to be available?

tests/puppeteer/pages/BO/design/emailThemes/index.js
```diff
@@ -0,0 +1,45 @@
...
+module.exports = class EmailThemes extends BOBasePage {
+  constructor(page) {
+    super(page);
+
+    this.pageTitle = 'Email Theme •';
+
+    // Selectors
+    this.blockDiv = 'div.content-div div.justify-content-center:nth-child(%ID)';
+    // Email Theme table selectors
+    this.emailThemeTable = `${this.blockDiv.replace('%ID', 5)} table.grid-table`;
```
jolelievre: Seems a bit unstable to rely on children order no? couldn't you use some class or id which would have less chance to be modified?

boubkerbribri: in this case there is only one table, so `table.grid-table` should do it.
but it would be better to have id on the block or a name like in the bloc translate emails `form[name='generate_mails']`.
do you think i should delete `${this.blockDiv.replace('%ID', 5)}` ?

tests/puppeteer/pages/BO/BObasePage.js
boubkerbribri: We can pass this one and and fixing the id after that

tests/puppeteer/pages/BO/design/emailThemes/index.js
jolelievre: Indeed it's too bad there is no ID to target this specific table, so if we add one it won't work any more
But for now yes I think you should only use `table.grid-table` to select it, because the chances are high that we add/remove/switch a block in this page The chances of a new table are low, but when it happens we will add the appropriate IDs

boubkerbribri: done

jolelievre (approved):

PR #17018, opened 2020-01-03
Cover Category CommandHandlers and QueryHandlers by behat tests  - Part 2

tdavidsonas88:
| Questions     | Answers
| Branch?       | develop
| Description?  | increases behat tests coverage for Category Management feature related handlers
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #14480
| How to test?  | CI tests should pass

AddRootCategory
EditRootCategory - fixes mistake in previous commit
DeleteCategoryCoverImage
DeleteCategoryMenuThumbnailImage
SetCategoryIsEnabled
BulkUpdateCategoriesStatus
GetCategoryIsEnabledHandler

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17018)

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -15,8 +15,8 @@ parameters:
     env(PS_LOG_OUTPUT): "%kernel.logs_dir%/%kernel.environment%.log"
     mail_themes_uri: "/mails/themes"
     mail_themes_dir: "%kernel.project_dir%%mail_themes_uri%"
-    ps_tmp_img_dir: "/var/www/PrestaShop/img/tmp/"
-    ps_cat_img_dir: "/var/www/PrestaShop/img/c/"
+    ps_tmp_img_dir: "%kernel.project_dir%/img/tmp/"
+    ps_cat_img_dir: "%kernel.project_dir%/img/c/"
 
 framework:
     assets:
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -15,8 +15,6 @@ parameters:
     env(PS_LOG_OUTPUT): "%kernel.logs_dir%/%kernel.environment%.log"
     mail_themes_uri: "/mails/themes"
     mail_themes_dir: "%kernel.project_dir%%mail_themes_uri%"
-    prestashop_temporary_image_directory: "%kernel.project_dir%/img/tmp/"
-    prestashop_category_image_directory: "%kernel.project_dir%/img/c/"
 
 framework:
     assets:
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -16,6 +16,7 @@ parameters:
     mail_themes_uri: "/mails/themes"
     mail_themes_dir: "%kernel.project_dir%%mail_themes_uri%"
     ps_tmp_img_dir: "/var/www/PrestaShop/img/tmp/"
+    ps_cat_img_dir: "/var/www/PrestaShop/img/c/"
 
 framework:
     assets:
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -15,8 +15,8 @@ parameters:
     env(PS_LOG_OUTPUT): "%kernel.logs_dir%/%kernel.environment%.log"
     mail_themes_uri: "/mails/themes"
     mail_themes_dir: "%kernel.project_dir%%mail_themes_uri%"
-    ps_tmp_img_dir: "%kernel.project_dir%/img/tmp/"
-    ps_cat_img_dir: "%kernel.project_dir%/img/c/"
+    prestashop_temporary_image_directory: "%kernel.project_dir%/img/tmp/"
+    prestashop_category_image_directory: "%kernel.project_dir%/img/c/"
 
 framework:
     assets:
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -15,6 +15,7 @@ parameters:
     env(PS_LOG_OUTPUT): "%kernel.logs_dir%/%kernel.environment%.log"
     mail_themes_uri: "/mails/themes"
     mail_themes_dir: "%kernel.project_dir%%mail_themes_uri%"
+    ps_tmp_img_dir: "/var/www/PrestaShop/img/tmp/"
 
 framework:
     assets:
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -1,7 +1,7 @@
 parameters:
     AdapterSecurityAdminClass:                LegacyTests\PrestaShopBundle\Mock\AdapterSecurityAdminMock
     prestashop.security.voter.product.class:  LegacyTests\PrestaShopBundle\Mock\PageVoter
-    ps_tmp_img_dir: "%kernel.project_dir%/img/tmp/"
+    prestashop_temporary_image_directory: "%kernel.project_dir%/img/tmp/"
 
 imports:
     - { resource: config_dev.yml }
@@ -27,12 +27,12 @@ services:
     prestashop.adapter.image.uploader.category_cover_image_uploader:
       class: 'PrestaShop\PrestaShop\Adapter\Image\Uploader\CategoryCoverImageUploader'
       arguments:
-        - '%ps_tmp_img_dir%'
+        - '%prestashop_temporary_image_directory%'
 
     prestashop.adapter.category.command_handler.edit_category_handler:
       class: PrestaShop\PrestaShop\Adapter\Category\CommandHandler\EditCategoryHandler
       arguments:
-        - '%ps_tmp_img_dir%'
+        - '%prestashop_temporary_image_directory%'
       public: true
       tags:
         - { name: tactician.handler, command: PrestaShop\PrestaShop\Core\Domain\Category\Command\EditCategoryCommand }
@@ -41,7 +41,7 @@ services:
       class: PrestaShop\PrestaShop\Adapter\Category\QueryHandler\GetCategoryForEditingHandler
       arguments:
         - '@prestashop.core.image.parser.image_tag_source_parser'
-        - '%ps_tmp_img_dir%'
+        - '%prestashop_temporary_image_directory%'
       tags:
         - { name: tactician.handler, command: PrestaShop\PrestaShop\Core\Domain\Category\Query\GetCategoryForEditing }
 
@@ -51,7 +51,7 @@ services:
         - '@filesystem'
         - '@prestashop.adapter.legacy.configuration'
         - '@prestashop.adapter.cache.clearer.smarty_cache_clearer'
-        - '%ps_tmp_img_dir%'
+        - '%prestashop_temporary_image_directory%'
       tags:
         - name: tactician.handler
           command: PrestaShop\PrestaShop\Core\Domain\Category\Command\DeleteCategoryMenuThumbnailImageCommand
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -1,7 +1,7 @@
 parameters:
     AdapterSecurityAdminClass:                LegacyTests\PrestaShopBundle\Mock\AdapterSecurityAdminMock
     prestashop.security.voter.product.class:  LegacyTests\PrestaShopBundle\Mock\PageVoter
-    ps_tmp_img_dir: "/tmp/"
+    ps_tmp_img_dir: "%kernel.project_dir%/img/tmp/"
 
 imports:
     - { resource: config_dev.yml }
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -1,6 +1,7 @@
 parameters:
     AdapterSecurityAdminClass:                LegacyTests\PrestaShopBundle\Mock\AdapterSecurityAdminMock
     prestashop.security.voter.product.class:  LegacyTests\PrestaShopBundle\Mock\PageVoter
+    ps_tmp_img_dir: "/tmp/"
 
 imports:
     - { resource: config_dev.yml }
@@ -22,6 +23,19 @@ swiftmailer:
 services:
     prestashop.bundle.routing.converter.cache:
         class: Symfony\Component\Cache\Adapter\ArrayAdapter
+
+    prestashop.adapter.image.uploader.category_cover_image_uploader:
+      class: 'PrestaShop\PrestaShop\Adapter\Image\Uploader\CategoryCoverImageUploader'
+      arguments:
+        - '%ps_tmp_img_dir%'
+
+    prestashop.adapter.category.command_handler.edit_category_handler:
+      class: PrestaShop\PrestaShop\Adapter\Category\CommandHandler\EditCategoryHandler
+      arguments:
+        - '%ps_tmp_img_dir%'
+      public: true
+      tags:
+        - { name: tactician.handler, command: PrestaShop\PrestaShop\Core\Domain\Category\Command\EditCategoryCommand }
 
 # Doctrine Configuration
 doctrine:
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -37,6 +37,25 @@ services:
       tags:
         - { name: tactician.handler, command: PrestaShop\PrestaShop\Core\Domain\Category\Command\EditCategoryCommand }
 
+    prestashop.adapter.category.query_handler.get_category_for_editing:
+      class: PrestaShop\PrestaShop\Adapter\Category\QueryHandler\GetCategoryForEditingHandler
+      arguments:
+        - '@prestashop.core.image.parser.image_tag_source_parser'
+        - '%ps_tmp_img_dir%'
+      tags:
+        - { name: tactician.handler, command: PrestaShop\PrestaShop\Core\Domain\Category\Query\GetCategoryForEditing }
+
+    prestashop.adapter.category.command_handler.delete_category_menu_thumbnail_image_handler:
+      class: PrestaShop\PrestaShop\Adapter\Category\CommandHandler\DeleteCategoryMenuThumbnailImageHandler
+      arguments:
+        - '@filesystem'
+        - '@prestashop.adapter.legacy.configuration'
+        - '@prestashop.adapter.cache.clearer.smarty_cache_clearer'
+        - '%ps_tmp_img_dir%'
+      tags:
+        - name: tactician.handler
+          command: PrestaShop\PrestaShop\Core\Domain\Category\Command\DeleteCategoryMenuThumbnailImageCommand
+
 # Doctrine Configuration
 doctrine:
     dbal:
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -1,7 +1,6 @@
 parameters:
     AdapterSecurityAdminClass:                LegacyTests\PrestaShopBundle\Mock\AdapterSecurityAdminMock
     prestashop.security.voter.product.class:  LegacyTests\PrestaShopBundle\Mock\PageVoter
-    prestashop_temporary_image_directory: "%kernel.project_dir%/img/tmp/"
 
 imports:
     - { resource: config_dev.yml }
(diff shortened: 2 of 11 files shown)


tests/Integration/Behaviour/Features/Context/Domain/CategoryFeatureContext.php
```diff
@@ -381,4 +545,103 @@ private function mapDataToEditableCategory(
...
+        /** @var Kernel $kernel */
+        $kernel = $this->getContainer()->get('kernel');
+        copy(
+            $kernel->getRootDir() . '/../img/' . $categoryCoverImage,
+            _PS_CAT_IMG_DIR_ . $categoryId . '.jpg'
+        );
+        copy(
+            $kernel->getRootDir() . '/../img/' . $categoryCoverImage,
+            _PS_CAT_IMG_DIR_ . $categoryId . '-' . stripslashes($categoryCoverImage) . '.jpg'
+        );
```
tomas862: I dont think its a good idea to actually move images in the test scenarios - the best option would be to mock and imitate actual image moving from source to destination. If its not possible when better not test this case at all

tests/Integration/Behaviour/Features/Context/Domain/CategoryFeatureContext.php
```diff
@@ -342,24 +500,30 @@ private function mapDataToEditableCategory(
...
+        }
+        if (isset($testCaseData['Menu thumbnails'])) {
+            $categoryCoverImage = $testCaseData['Menu thumbnails'];
+            // could not use handler because it uses move_uploaded_file in Uploader.php which allows only POST upload
+            /** @var Kernel $kernel */
+            $kernel = $this->getContainer()->get('kernel');
+            copy(
+                $kernel->getRootDir() . '/../img/' . $categoryCoverImage,
+                _PS_CAT_IMG_DIR_ . $categoryId . '-' . MenuThumbnailId::ALLOWED_ID_VALUES[0] . '_thumb.jpg'
+            );
```
tomas862: same here as in comment https://github.com/PrestaShop/PrestaShop/pull/17018/files#r362855042

tests/Integration/Behaviour/Features/Context/Domain/CategoryFeatureContext.php
tdavidsonas88: Thanks Tomas, for now commented out scenarios using this method

Also some useful info:

     * @param bool        $test         Whether the test mode is active
     *                                  Local files are used in test mode hence the code should not enforce HTTP uploads

in our case the code enforces HTTP uploads in UploadFile inside handlers and because of this reason cannot use test mode. @matks do we really need to use: move_uploaded_file php function which forces only HTTP uploads.

tests/Integration/Behaviour/Features/Context/Domain/CategoryFeatureContext.php
tdavidsonas88: Thanks Tomas, for now commented out scenarios using this method. While will figure out better approach.

tests/Integration/Behaviour/Features/Context/Domain/CategoryFeatureContext.php
tdavidsonas88: https://github.com/symfony/symfony/blob/8e9aafc84fc741d966607bff5b18fdaaf0ed2987/src/Symfony/Component/HttpFoundation/File/UploadedFile.php#L19-L53

matks:
@tdavidsonas88 You have found that current handler implementation is hard to test because of move_uploaded_file usage.

If the code is hard to test, it means the code is badly written. So the solution is to refactor this code into an usage that makes testing easier.

vfsStream is a standard virtual file system library used in php tests to simulate such usecases. We already have a few tests that use it, see [ImageFileDeleterTest](https://github.com/PrestaShop/PrestaShop/blob/develop/tests/Unit/Core/Image/Deleter/ImageFileDeleterTest.php) for example.

matks:
We will probably need to remove hardcoded `_PS_TMP_IMG_DIR_` constants and inject them through the constructor. This will allow us to modify the working directory used by CoverImageUploader and replace it by a virtual one provided by vFsStream.

We will be able to inject our virtual directory by 3 possible ways:
- either override Symfony service declaration for this service, only in test environment
- control how `_PS_TMP_IMG_DIR_` is injected into the service, for example if we use a Symfony parameter placeholder like `%prestashop.temporary.image.diectoryr% we could modify the value of this placeholder for test environment
- or modify the raw value of _PS_TMP_IMG_DIR_ with a define 😅but it looks a bit overkill

In test environment, we can modify symfony service container configuration in file app/config/config_test.yml

Ping @jolelievre and @PierreRambaud for additional ideas how to handle this topic

tests/Integration/Behaviour/Features/Context/Domain/CategoryFeatureContext.php
tdavidsonas88: @jolelievre, @PierreRambaud - currently our legacy classes forces to upload the file via HTTP form. That's why cannot use UploadedFile in test mode. Any creative ideas to overcome this issue would be helpful. 

Also @matks suggested a solution below.

PierreRambaud: Because we are not using `\copy` we can mock with namespace, but it's quite ugly.
I'm more in favor of refactoring like he said: https://github.com/PrestaShop/PrestaShop/pull/17018#issuecomment-571985497

tdavidsonas88:
> @tdavidsonas88 You have found that current handler implementation is hard to test because of move_uploaded_file usage.
> 
> If the code is hard to test, it means the code is badly written. So the solution is to refactor this code into an usage that makes testing easier.
> 
> vfsStream is a standard virtual file system library used in php tests to simulate such usecases. We already have a few tests that use it, see [ImageFileDeleterTest](https://github.com/PrestaShop/PrestaShop/blob/develop/tests/Unit/Core/Image/Deleter/ImageFileDeleterTest.php) for example.

@matks Example from the CoverImageUploader using vFsStream is great one. Though going deeper it sets default mime/type: text/plain and our code needs the image/jpg. Have difficulties finding how to set this mime/type using vFsStream. Still have some hope to find it soon!

tdavidsonas88:
> We will probably need to remove hardcoded `_PS_TMP_IMG_DIR_` constants and inject them through the constructor. This will allow us to modify the working directory used by CoverImageUploader and replace it by a virtual one provided by vFsStream.

Going @matks your suggested way. Had some creative searching setting the correct mime type for file. Hope to deliver results soon.

matks:
> > We will probably need to remove hardcoded `_PS_TMP_IMG_DIR_` constants and inject them through the constructor. This will allow us to modify the working directory used by CoverImageUploader and replace it by a virtual one provided by vFsStream.
> 
> Going @matks your suggested way. Had some creative searching setting the correct mime type for file. Hope to deliver results soon.

🎉
Thanks for being proactive 😉

tdavidsonas88:
@matks  Continuing from yesterday - Some PHP functions do not work with any other URLs than pure filenames, as they are just thin wrappers around the native C functions and therefore do not support userland stream wrappers like vfs. Currently these functions are known:
chdir()
realpath()
link(), symlink(), readlink(), linkinfo()
SplFileInfo::getRealPath()
tempnam()
There may be more - in case a function you use does not work correctly with vfsStream URLs check the PHP manual for this function if it mentions a similar native C function. For more information: https://github.com/bovigo/vfsStream/wiki/Known-Issues

It looks like it is not enough only to inject _PS_TMP_IMG_DIR_ value. More creative refactoring is needed

tdavidsonas88:
actually we can do it simply with the hardcoded tempnam('/tmp', 'test') injected for tests where '/tmp' comes from test config. For more info: https://stackoverflow.com/questions/36609046/phpunit-is-it-possible-to-test-mkdir-tempnam-realpath-and-move-uploaded-file

tdavidsonas88:
@matks Even though we can create images in /tmp/ manually. It still does not pass move_uploaded_file validation. It can be done with curl to imitate the form POST request. But then when deleting the file there's another issue:

I think it is too early to test these handlers because DeleteCategoryCoverImageHandler is not using any service we can inject or mock but instead using deleteImage from ObjectModel. Looks like these handlers are still not migrated enough to be tested with Behat.

Any thoughts?

tdavidsonas88:
Injected the ps_tmp_img_dir through services and even was able to cover one delete handler. Though the GUI testing of uploading image have not passed - that's why have this still in WIP.

tdavidsonas88:
Fixed the UI part. Ready to be merged.

app/config/config.yml
```diff
@@ -15,6 +15,8 @@ parameters:
     env(PS_LOG_OUTPUT): "%kernel.logs_dir%/%kernel.environment%.log"
     mail_themes_uri: "/mails/themes"
     mail_themes_dir: "%kernel.project_dir%%mail_themes_uri%"
+    ps_tmp_img_dir: "/var/www/PrestaShop/img/tmp/"
```
PierreRambaud: Why?? :thinking:

app/config/config_test.yml
```diff
@@ -1,6 +1,7 @@
 parameters:
     AdapterSecurityAdminClass:                LegacyTests\PrestaShopBundle\Mock\AdapterSecurityAdminMock
     prestashop.security.voter.product.class:  LegacyTests\PrestaShopBundle\Mock\PageVoter
+    ps_tmp_img_dir: "/tmp/"
```
PierreRambaud: Nope :scream:

composer.json
```diff
@@ -131,7 +131,8 @@
         "mikey179/vfsstream": "^1.6",
         "phake/phake": "@stable",
         "phpunit/phpunit": "~5.7",
-        "symfony/phpunit-bridge": "^3.4.31"
+        "symfony/phpunit-bridge": "^3.4.31",
+        "ext-gd": "*"
```
PierreRambaud: It's not a dev requirement

app/config/config.yml
tdavidsonas88: To be able to inject other directory from config_test.yml by overriding services. Any better ideas?

app/config/config_test.yml
tdavidsonas88: Did I misunderstood @matks here ?

We will probably need to remove hardcoded _PS_TMP_IMG_DIR_ constants and inject them through the constructor. This will allow us to modify the working directory used by CoverImageUploader and replace it by a virtual one provided by vFsStream.

We will be able to inject our virtual directory by 3 possible ways:

either override Symfony service declaration for this service, only in test environment
control how _PS_TMP_IMG_DIR_ is injected into the service, for example if we use a Symfony parameter placeholder like `%prestashop.temporary.image.diectoryr% we could modify the value of this placeholder for test environment

app/config/config.yml
tdavidsonas88: relatave path needed here ofcourse :)

app/config/config_test.yml
tdavidsonas88: Talked with Rokas and yes we agree it is a bad idea to use "/tmp/". 

Actually could use relative /img/tmp/ path from root and while vfstream does not support tempnam() function which is used by our legacy code it seems there's no benefit of injecting different urls for test environment then. Do we understand it correctly? 

Also now we see that we have two constants to maintain instead of one. So without vfstream does injecting the another "/testtemp/" folder has any meaning?

@PierreRambaud @matks would be nice to confirm the solution now because it is already clear what kind of problems we face.

tdavidsonas88: Actually it has some meaning injecting parameters via services. While injecting %prestashop.temporary.image.diectoryr% is not so important because it's already temporary. 

Also in order to get the EditableCategory GetCategoryForEditing there is one more constant _PS_CAT_IMG_DIR_ which points to the /img/c/ folder and injecting the parameter here is useful for test to be able to confirm image was uploaded or already deleted.

PierreRambaud: Always use the `var/` for temporary files, the  `img` directory is not for temporary resources./
Maybe time to have a `var/tmp`

app/config/config.yml
```diff
@@ -15,6 +15,8 @@ parameters:
     env(PS_LOG_OUTPUT): "%kernel.logs_dir%/%kernel.environment%.log"
     mail_themes_uri: "/mails/themes"
     mail_themes_dir: "%kernel.project_dir%%mail_themes_uri%"
+    prestashop_temporary_image_directory: "%kernel.project_dir%/img/tmp/"
```
PierreRambaud:
```suggestion
    temp_dir: "%kernel.project_dir%/var/tmp/"
```

app/config/config_test.yml
```diff
@@ -1,6 +1,7 @@
 parameters:
     AdapterSecurityAdminClass:                LegacyTests\PrestaShopBundle\Mock\AdapterSecurityAdminMock
     prestashop.security.voter.product.class:  LegacyTests\PrestaShopBundle\Mock\PageVoter
+    prestashop_temporary_image_directory: "%kernel.project_dir%/img/tmp/"
```
PierreRambaud:
```suggestion
```
Already loaded in config.yml

app/config/config.yml
PierreRambaud: We already know we are in a prestashop context :)

tdavidsonas88: @PierreRambaud  @matks as we with @rokaszygmantas  see without vfstream using these injected parameters cause more problems / refactoring / creating non existing folder structures. So we suggest to not use these injected parameters now and use folder structures like we currently have /img/tmp and img/c

PierreRambaud:
```suggestion
    image_temp_dir: "%kernel.project_dir%/img/tmp/"
```
be consistent with above parameters (like `mail_themes_dir`)

app/config/config.yml
```diff
@@ -15,6 +15,8 @@ parameters:
...
+    prestashop_category_image_directory: "%kernel.project_dir%/img/c/"
```
PierreRambaud:
```suggestion
    image_cat_dir: "%kernel.project_dir%/img/c/"
```

tdavidsonas88:
@PierreRambaud @matks - with some guidance of @rokaszygmantas chosen to create file in /img/c 
 and similar for thumbnail folders from the test avoiding all the services injections. It was enough for test scenarios to pass. :slightly_smiling_face:

PierreRambaud:
@tdavidsonas88 Since it's related to test env, it shouldn't be created in `img/c` :thinking:

tdavidsonas88:
> @tdavidsonas88 Since it's related to test env, it shouldn't be created in `img/c`

Me too wanted that - after couple of days of efforts - saw that can't do that with a predictable deadline. Basically there are couple of reasons for that:

1. move_uploaded_file expects POST form upload so decided to fake file upload.
2. DeleteCategoryCoverImageHandler is not refactored enough and uses 
`$category->deleteImage(true)` from object model so no service parameter can be injected here as I used to do in symfony services.

Knowing that can we somehow solve this issue without too much efforts?

PierreRambaud:
We shouldn't think about effort, but, are  they real test and are they easily maintainable? 
If we mix data with another environment, we will have some trouble :sweat_smile:

tdavidsonas88:
> We shouldn't think about effort, but, are they real test and are they easily maintainable?
> If we mix data with another environment, we will have some trouble

Actually we could use vfstream faking the image upload (faking means - not using the handler which expects file upload via POST form submit and does not use methods not supported by vfstream). 

Then inject vfstream path to symfony service like @matks recommended. That would be enough to cover DeleteCategoryMenuThumbnailImageHandler.

For DeleteCategoryCoverImageHandler - it needs some creative idea how  we can handle $this->deleteCoverImage($category). As I see it will throw exception:   'Cannot delete cover image for category with id "%s"', :

https://github.com/PrestaShop/PrestaShop/blob/675626f689becd95c9bf3161e368f74dcfbf11c0/src/Adapter/Category/CommandHandler/DeleteCategoryCoverImageHandler.php#L72-L83

https://github.com/PrestaShop/PrestaShop/blob/675626f689becd95c9bf3161e368f74dcfbf11c0/src/Adapter/Category/CommandHandler/DeleteCategoryCoverImageHandler.php#L101-L117

Also @rokaszygmantas might have some insight as I already included him in helping me.

My assumption is that deleteCoverImage shoud not be used in DeleteCategoryCoverImageHandler because it depends on the ObjectModel and changed with

            if ($this->filesystem->exists($categoryCoverImagePath)) {
                $this->filesystem->remove($categoryCoverImagePath);
            }
            // and also
            if ($this->filesystem->exists($tempCoverImagePath)) {
                $this->filesystem->remove($tempCoverImagePath);
            }

And then it can be done :slightly_smiling_face:   The only not tested part would be file upload using form and HTTP POST request as we would fake it for now.

PierreRambaud:
>Also @rokaszygmantas might have some insight as I already included him in helping me.
>
> My assumption is that deleteCoverImage shoud not be used in DeleteCategoryCoverImageHandler because it depends on the ObjectModel and changed with

We should decide because the ObjectModel doesn't only delete the image, it also checks for multishop instances.

src/Adapter/Category/CommandHandler/DeleteCategoryMenuThumbnailImageHandler.php
```diff
@@ -92,7 +92,7 @@ public function handle(DeleteCategoryMenuThumbnailImageCommand $command)
 
         $thumbnailPath = sprintf(
             '%s%s-%s_thumb.jpg',
-            $this->configuration->get('_PS_CAT_IMG_DIR_'),
+            _PS_CAT_IMG_DIR_,
```
matks: Woah that is a nasty bug that you have found 😮(and fixed)

So it means the feature "delete category menu thumbnail image" was broken ?

matks:
@tdavidsonas88 Code looks ✅ to me however because of the git conflict, Travis did not run. Can you please rebase ?

src/Adapter/Category/CommandHandler/DeleteCategoryMenuThumbnailImageHandler.php
tdavidsonas88: Actually I refactored code several times searching for a good solution how to imitate file upload; And this code I've change was created by me.

PierreRambaud: It wasn't a bug: https://github.com/PrestaShop/PrestaShop/blob/develop/src/Adapter/Configuration.php#L101

tests/Integration/Behaviour/Features/Context/Domain/CategoryFeatureContext.php
```diff
@@ -29,27 +29,41 @@
 use Behat\Gherkin\Node\TableNode;
 use Configuration;
 use PHPUnit\Framework\Assert as Assert;
+use PHPUnit_Framework_Assert;
```
PierreRambaud:
```suggestion
use PHPUnit\Framework\Assert;
```

tdavidsonas88: I see, there is Assert :slightly_smiling_face:

tests/Integration/Behaviour/Features/Context/Domain/CategoryFeatureContext.php
```diff
@@ -285,8 +270,190 @@ public function editRootCategoryWithFollowingDetails(string $rootCategory, Table
...
+     * @param string $categoryReference
+     * @param TableNode $table
+     */
+    public function addNewRootCategoryWithFollowingDetails(string $categoryReference, TableNode $table)
+    {
+        $testCaseData = $table->getRowsHash();
+        $editableRootCategoryTestCaseData = $this->mapDataToEditableCategory($testCaseData);
+
+        /** @var EditCategoryCommand $command */
+        $command = new AddRootCategoryCommand(
```
tdavidsonas88, line 287: @matks - general question related to command. Would be interesting to know why we are passing primitive types to commands constructors instead of objects and entities? Looking far from now - if one day we will change object model to doctrine - then we'll have lots of refactoring once again. Also Symfony forms are capable of handling into objects/entities (possibility of less code).

What part of the PrestaShop migration puzzle am I missing?

matks: It's intended 😊this is one of CQRS rules.
Commands and Queries must contain only primitive types, although a few exceptions are possible but only bu providing custom immutable types (ex: `Money`).

The idea is that Commands and Queries represent an intention of the user: `ViewUser`, `AddOrderFromBO`, `DisableWarehouse`. These intentions are constructed by a user using the types he has: strings, integers, booleans, usually passed by HTTP requests.

As https://stackoverflow.com/questions/15891143/cqrs-command-constructs states:
"A command is data-oriented, not object-oriented. Its properties should be mostly primitives. You're not modeling anything here, just sending flat data around."

And all these types are immutable which helps making sure a Command or Query is immutable too.

Moreover we want to make dispatching Commands and Queries as easy as possible for developers so using primitives only make it very easy to build one.

src/Adapter/Category/CommandHandler/DeleteCategoryMenuThumbnailImageHandler.php
matks: 😆 @PierreRambaud thank you legacy configuration with fallback behavior 🙏

tests/Integration/Behaviour/Features/Context/Domain/CategoryFeatureContext.php
tdavidsonas88: Thank you for explaining to me that. :)

src/Adapter/Category/CommandHandler/DeleteCategoryMenuThumbnailImageHandler.php
tdavidsonas88: That's not careful enough refactoring style from me. Thank you for showing it to me :slightly_smiling_face:

tdavidsonas88:
cannot reproduce the failure locally with: ./vendor/bin/php-cs-fixer fix --dry-run

matks:
Closing and reopening to re-trigger Travis

matks:
> cannot reproduce the failure locally with: ./vendor/bin/php-cs-fixer fix --dry-run

This is @PierreRambaud fault :trollface: (see https://github.com/PrestaShop/PrestaShop/commit/51b8bfa2213ea97392ffb397ed284b20e9b077c1) so I expect him to help us 😈

PierreRambaud:
@tdavidsonas88 Don't forget to clear your php-cs-fixer cache file, and run `composer install` to get the latest version :)

matks (approved):

matks:
Thank you @tdavidsonas88 !

PR #17028, opened 2020-01-06
Behat tests for Customer, Address features handlers

tdavidsonas88:
| Questions     | Answers
| Branch?       | develop
| Description?  | Customer, Address feature behat tests
| Type?         | improvement
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/14480
| How to test?  | CI should pass

AddManufacturerAddress
GetManufacturerAddressForEditing
AddCustomerAddress
GetCustomerAddressForEditing
DeleteAddress
BulkDeleteAddressHandler
EditManufacturerAddress

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17028)

### DIFF
--- a/src/Core/Domain/Address/Command/AddCustomerAddressCommand.php
+++ b/src/Core/Domain/Address/Command/AddCustomerAddressCommand.php
@@ -147,14 +147,14 @@ class AddCustomerAddressCommand
         string $city,
         int $countryId,
         string $postcode,
-        string $dni,
-        string $company,
-        string $vat_number,
-        string $address2,
-        int $id_state,
-        string $phone,
-        ?string $phone_mobile,
-        string $other
+        string $dni = null,
+        string $company = null,
+        string $vat_number = null,
+        string $address2 = null,
+        int $id_state = 0,
+        string $phone = null,
+        ?string $phone_mobile = null,
+        string $other = null
     ) {
         $this->customerId = new CustomerId($customerId);
         $this->addressAlias = $addressAlias;
--- a/src/Core/Domain/Address/QueryResult/EditableManufacturerAddress.php
+++ b/src/Core/Domain/Address/QueryResult/EditableManufacturerAddress.php
@@ -128,13 +128,13 @@ class EditableManufacturerAddress
         $city,
         $manufacturerId,
         $countryId,
-        $address2,
-        $postCode,
-        $stateId,
-        $homePhone,
-        $mobilePhone,
-        $other,
-        $dni
+        $address2 = null,
+        $postCode = null,
+        $stateId = null,
+        $homePhone = null,
+        $mobilePhone = null,
+        $other = null,
+        $dni = null
     ) {
         $this->addressId = $addressId;
         $this->lastName = $lastName;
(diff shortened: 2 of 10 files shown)


tdavidsonas88:
@rokaszygmantas @matks One important lesson from this pull request - Try not to mix using Domain and without domain contexts. For example CustomerFeatureContext (not inside domain folder) uses domain handlers and also is heavily based on Legacy classes. Which is not the best approach because refactoring it and getting rid of legacy classes is super hard and also causes code duplication inside Domain folder.

Suggested approach to this problem all the new handlers tests context methods put to Domain contexts and avoid using contexts and methods not from the Domain folder in the new tests. At some point we would have duplicated code but when everything will be done with Domain it should be possible to remove the not domain related contexts.

Also this is related to writing code in style avoiding using legacy classes as much as possible.

matthieu-rolland:
Hi @tdavidsonas88, thank you for your contribution ! 

It looks like behavior tests are not green, but I'm not sure why 🤔

matks:
> @rokaszygmantas @matks One important lesson from this pull request - Try not to mix using Domain and without domain contexts. For example CustomerFeatureContext (not inside domain folder) uses domain handlers and also is heavily based on Legacy classes. Which is not the best approach because refactoring it and getting rid of legacy classes is super hard and also causes code duplication inside Domain folder.
> 
> Suggested approach to this problem all the new handlers tests context methods put to Domain contexts and avoid using contexts and methods not from the Domain folder in the new tests. At some point we would have duplicated code but when everything will be done with Domain it should be possible to remove the not domain related contexts.
> 
> Also this is related to writing code in style avoiding using legacy classes as much as possible.

You're right.

The original mix was a consequence of us still looking for the right setup for our tests. When we started using behat we had a lot of unanswered questions 😅and the code of the 1st behat tests is quite messy.

We'll try to follow this for now 👍

matks:
> Hi @tdavidsonas88, thank you for your contribution !
> 
> It looks like behavior tests are not green, but I'm not sure why 🤔

```
323 scenarios (319 passed, 4 undefined)
3808 steps (3794 passed, 4 undefined, 10 skipped)
3m1.01s (91.57Mb)
Script @php -d date.timezone=UTC ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml --no-snippets --strict handling the integration-behaviour-tests event returned with error code 1
```

In Travis, we run behat with `--strict` flag which requires all tests to pass. Skipped and undefined files are considered an issue and prevent Travis to accept the PR.

tdavidsonas88:
@matks corrected the accidental typo in the context for tests to pass

tdavidsonas88:
@matks thank you for merging AddCustomerAddressHandler so quickly. :) Needed it to stop using legacy classes in domain test cases.

src/Core/Domain/Address/Command/AddCustomerAddressCommand.php
```diff
@@ -147,14 +147,14 @@ public function __construct(
...
         string $postcode,
-        string $dni,
-        string $company,
-        string $vat_number,
-        string $address2,
-        int $id_state,
-        string $phone,
-        ?string $phone_mobile,
-        string $other
+        string $dni = null,
```
matks, line 150: Makes sense 🤔indeed these parameters are mandatory for a B2B business, but not a B2C business

matks (approved):

matks:
Thank you @tdavidsonas88

PR #17036, opened 2020-01-06
Protect modules vendor folder on install/upgrade/enable

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Create HtaccessFolderProtector to protect folder, it is bound to module install, enable and upgrade
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | Install a module that contains a vendor folder (`welcome` for example) Check that the vendor folder contains a `.htaccess` file after installation (should also work with upgrade, enable and on fresh install)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17036)

### DIFF
--- a/src/Core/Exception/IOException.php
+++ b/src/Core/Exception/IOException.php
@@ -1,0 +1,59 @@
+<?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Exception;
+
+use Throwable;
+
+/**
+ * Exception class thrown when a filesystem operation failure happens.
+ */
+class IOException extends CoreException
+{
+    private $path;
+
+    /**
+     * @param string $message
+     * @param int $code
+     * @param Throwable $previous
+     * @param string|null $path
+     */
+    public function __construct($message = '', $code = 0, Throwable $previous = null, $path = null)
+    {
+        parent::__construct($message, $code, $previous);
+        $this->path = $path;
+    }
+
+    /**
+     * Returns the associated path for the exception.
+     *
+     * @return string|null
+     */
+    public function getPath()
+    {
+        return $this->path;
+    }
+}
--- a/src/Core/Exception/IOException.php
+++ b/src/Core/Exception/IOException.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * 2007-2019 PrestaShop SA and Contributors
+ * 2007-2020 PrestaShop SA and Contributors
  *
  * NOTICE OF LICENSE
  *
@@ -19,7 +19,7 @@
  * needs please refer to https://www.prestashop.com for more information.
  *
  * @author    PrestaShop SA <contact@prestashop.com>
- * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @copyright 2007-2020 PrestaShop SA and Contributors
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
--- a/src/Core/Security/HtaccessFolderProtector.php
+++ b/src/Core/Security/HtaccessFolderProtector.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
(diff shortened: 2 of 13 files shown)


eternoendless:
Needs to target 1.7.6.x

tests/Unit/Resources/security/htaccess
```diff
@@ -0,0 +1,10 @@
+# Apache 2.2
```
Quetzacoalt91: Missing dot in filename `htaccess`

src/Core/Security/HtaccessFolderProtector.php
```diff
@@ -0,0 +1,72 @@
...
+        }
+        $this->htaccessContent = file_get_contents($htaccessTemplatePath);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function protectFolder($folderPath)
+    {
+        if (!file_exists($folderPath) || !is_dir($folderPath)) {
```
PierreRambaud:
```suggestion
        if (!is_dir($folderPath)) {
```

src/Core/Security/HtaccessFolderProtector.php
```diff
@@ -0,0 +1,72 @@
...
+            throw new FileNotFoundException(sprintf('Cannot protect nonexistent folder %s', $folderPath));
+        }
+
+        $htaccessPath = $folderPath . DIRECTORY_SEPARATOR . '.htaccess';
+        if (!file_exists($htaccessPath)) {
+            if (!@file_put_contents($htaccessPath, $this->htaccessContent)) {
```
PierreRambaud: use `is_writable($folderPath)` instead of hiding error with `@`

src/PrestaShopBundle/Resources/config/services/core/security.yml
```diff
@@ -4,3 +4,8 @@ services:
 
   prestashop.core.crypto.hashing:
     class: 'PrestaShop\PrestaShop\Core\Crypto\Hashing'
+
+  prestashop.core.security.folder_protector.vendor:
+    class: 'PrestaShop\PrestaShop\Core\Security\HtaccessFolderProtector'
+    arguments:
+      - '%kernel.project_dir%/vendor/.htaccess'
```
PierreRambaud: it's not really a template. template must be store in resources directory instead

src/Core/Security/HtaccessFolderProtector.php
jolelievre: Done

src/Core/Security/HtaccessFolderProtector.php
jolelievre: Sadly I can't do that `is_writable` returns false on non existing files (which is always the case for this use case)
So I can only check if `file_put_contents` was successful

src/PrestaShopBundle/Resources/config/services/core/security.yml
jolelievre: Moved into `app/Resources/security/.htaccess.dist`

src/Core/Security/HtaccessFolderProtector.php
PierreRambaud: This is why I talked about the `$folderPath` :)

jolelievre: Ooooohh! My bad 😅
Done!

src/Core/Exception/IOException.php
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
```
Progi1984:
```suggestion
 * 2007-2020 PrestaShop SA and Contributors
```

src/Core/Exception/IOException.php
```diff
@@ -0,0 +1,59 @@
...
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
```
Progi1984:
```suggestion
 * @copyright 2007-2020 PrestaShop SA and Contributors
```

and so on...

src/PrestaShopBundle/EventListener/ModuleProtectorListener.php
```diff
@@ -0,0 +1,103 @@
...
+namespace PrestaShopBundle\EventListener;
+
+use PrestaShop\PrestaShop\Core\Exception\IOException;
+use PrestaShop\PrestaShop\Core\Security\FolderProtectorInterface;
+use PrestaShopBundle\Event\ModuleManagementEvent;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Listens to install/upgrade events from module manager, and protects the module vendor
```
Progi1984:
```suggestion
 * Listen install/upgrade events from module manager, and protect the module vendor
```

src/Core/Security/HtaccessFolderProtector.php
```diff
@@ -0,0 +1,74 @@
...
+namespace PrestaShop\PrestaShop\Core\Security;
+
+use PrestaShop\PrestaShop\Core\Exception\FileNotFoundException;
+use PrestaShop\PrestaShop\Core\Exception\IOException;
+
+/**
+ * Class HtaccessFolderProtector protects a designated folder by inserting an htaccess file in it
+ * which prevents access from an external call.
+ */
+class HtaccessFolderProtector implements FolderProtectorInterface
```
Progi1984:
```suggestion
class HtaccessFolderGuard implements FolderProtectorInterface
```
Sounds more :gb:

Progi1984 (commented):
@jolelievre Some feedback

Progi1984 (approved):

PierreRambaud (approved):

Progi1984:
Thanks @jolelievre

matks:
Sorry guys, I'm quite late for the party 😅I just saw this PR.

In order to ease unit testing, do you think it would make sense, in a new PR, to use `vFsStream` to handle I/O simulations ?
And maybe also use a FileSystem implementation to handle file contents inside the source code, instead of raw `file_put_contents()` ?

The reason I say this is that we have issues in the migration with classes which perform files/folders operations, the issues mainly arise when writing integration tests.

PR #17050, opened 2020-01-07
Quick access error when having a wrong url

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Thanks to Charlie Hosier for the report. Url are not escaped when using the quick add button. CVE-2020-6632
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Quick access must have urls escaped with token even after an add<br> image
.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17050)


jolelievre (dismissed):
I don't understand everything but I trust you ^^
I guess the main difference is that `getQuickAccessesWithToken` rebuilds the link using `context->link`

jolelievre (changes requested):
Ok for me, but you don't need to publish all the built assets
I think main.bundle.js is enough

PierreRambaud:
@jolelievre Not my fault if all assets changed when I built them :sweat_smile:

jolelievre:
@PierreRambaud true! but you can choose to commit only some of them ^^ The one related to what you changed

jolelievre (dismissed):
Ok, since we are preparing a fix version let's publish all assets just in case 😉

matks (approved):

NeOMakinG:
Thanks @PierreRambaud

admin-dev/themes/default/template/header.tpl
```diff
@@ -119,97 +119,98 @@
...
-                    $.each(data, function(index,value){
-                      if (typeof data[index]['name'] !== 'undefined')
-                        quicklink_list += '<li><a href="' + data[index]['link'] + '&token=' + data[index]['token'] + '">' + data[index]['name'] + '</a></li>';
-                    });
+            if(method == 'add' && name || method == 'remove')
+            {
+              $.ajax({
+                type: 'POST',
+                headers: { "cache-control": "no-cache" },
+                async: false,
```
Eolia, line 181: Not a good idea^^
   async: false,

jolelievre: Not sure I understand why this is not a good idea Besides this code has been changed, only the indentation So it's been here for quite some time I think

Eolia: Synchronous XMLHttpRequest outside of workers is in the process of being removed from the web platform as it has detrimental effects to the end user’s experience. (This is a long process that takes many years.) Developers must not pass false for the async argument when current global object is a Window object. User agents are strongly encouraged to warn about such usage in developer tools and may experiment with throwing an "InvalidAccessError" DOMException when it occurs.

The code being refactored it would be good to remove these depreciated elements, right?

PR #17069, opened 2020-01-08
Fix Currency i18n fields issue in webservice

### LINKED ISSUE #16760: [WS] Currency translatable fields empty in webservice 1.7.6.1
**Describe the bug**

When listing currencies using the web service, the translatable fields are empty.

Also when updating currencies using web service, the translatable fields are cleared/emptied even if no **name** or **symbol** element is present (removed from xml).

This has implications when doing patch-work using external services.
For example loading existing currency data, re-populating currency rate and then pushing it back to prestashop. All translatable fields are emptied.

**To Reproduce**

Browse your currencies endpoint:
yourdomain.com/api/currencies?display=full

**Screenshots**

Before PUT

Web service response

Example GET/PUT

Cleared tables after PUT

**Additional information**
PrestaShop version: 1.7.6.1 release
PHP version: 7.0.33

atomiix:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Correctly show `name` and `symbol` in the xml output of the currency webservice and prevent updating those fields when updating the currency.
| Type?         | bug fix
| Category?     | WS
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16760
| How to test?  | See below

# How to test

1. After enabling webservice in `Advanced Parameters -> Webservice` and adding a key with full access on the currency endpoint, send a GET request to `/api/currencies?display=full`. The field `name` and `symbol` should be present with the correct value.

2. Send a PUT request on `/api/currencies` with the necessary XML body : 
```
<?xml version="1.0" encoding="UTF-8"?>
<prestashop
  xmlns:xlink="http://www.w3.org/1999/xlink">
  <currency>
    <id><![CDATA[1]]></id>
    <iso_code><![CDATA[EUR]]></iso_code>
    <numeric_iso_code><![CDATA[978]]></numeric_iso_code>
    <precision><![CDATA[2]]></precision>
    <conversion_rate><![CDATA[1.000000]]></conversion_rate>
    <deleted><![CDATA[0]]></deleted>
    <active><![CDATA[1]]></active>
  </currency>
</prestashop>
```

In the database, the field `name` and `symbol` should not be empty in the `ps_currency_lang` table.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17069)

### DIFF
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -244,6 +244,10 @@ class CurrencyCore extends ObjectModel
     public function getWebserviceParameters($ws_params_attribute_name = null)
     {
         $parameters = parent::getWebserviceParameters($ws_params_attribute_name);
+        // name & symbol are i18n fields but casted to single string in the constructor
+        // so we need to force the webservice to consider those fields as non-i18n fields.
+        // Also, in 1.7.5 the field symbol didn't exists and name wasn't an i18n field so in order
+        // to keep 1.7.6 backward compatible we need to make those fields non-i18n.
         $parameters['fields']['name']['i18n'] = false;
         $parameters['fields']['symbol']['i18n'] = false;
 
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -37,6 +37,13 @@ class CurrencyCore extends ObjectModel
     public $name;
 
     /**
+     * Localized names of the currency
+     *
+     * @var string[]
+     */
+    protected $localizedNames;
+
+    /**
      * Alphabetic ISO 4217 code of this currency.
      *
      * @var string
@@ -94,6 +101,13 @@ class CurrencyCore extends ObjectModel
     public $symbol;
 
     /**
+     * Localized Currency's symbol.
+     *
+     * @var string[]
+     */
+    private $localizedSymbol;
+
+    /**
      * CLDR price formatting pattern
      * e.g.: In french (fr-FR), price formatting pattern is : #,##0.00 ¤.
      *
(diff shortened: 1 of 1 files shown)


classes/Currency.php
```diff
@@ -379,6 +434,20 @@ public function getName()
         return Tools::ucfirst($this->name[$id_lang]);
     }
 
+    public function getSymbol()
+    {
+        if (is_string($this->symbol)) {
+            return $this->symbol;
+        }
+
+        $id_lang = $this->id_lang;
```
Progi1984, line 447: Please use the Null Coalescing Operator
```suggestion
        $id_lang = $this->id_lang ?? Configuration::get('PS_LANG_DEFAULT');
```

Progi1984 (commented):
A small feedback

classes/Currency.php
PierreRambaud: No because we must be compatible with php5.6 and this one has been added in 7.0 https://www.php.net/manual/en/migration70.new-features.php

classes/Currency.php
```diff
@@ -205,6 +241,25 @@ public function __construct($id = null, $idLang = null, $idShop = null)
         }
     }
 
+    public function getWebserviceParameters($ws_params_attribute_name = null)
+    {
+        $parameters = parent::getWebserviceParameters($ws_params_attribute_name);
+        $parameters['fields']['name']['i18n'] = false;
```
PierreRambaud, line 251: Could you explain why you add this?

atomiix: Done

classes/Currency.php
```diff
@@ -379,6 +438,20 @@ public function getName()
         return Tools::ucfirst($this->name[$id_lang]);
     }
 
+    public function getSymbol()
+    {
+        if (is_string($this->symbol)) {
```
PierreRambaud, line 443: What if `$this->symbol` is empty?

atomiix: It will return an empty string then but I think that's what we want as long as the value comes from the DB. Wdyt?

PierreRambaud: Oh yes, I think I'm really tired today :sweat_smile:

PierreRambaud (approved):

Progi1984:
Thanks @atomiix

PR #17092, opened 2020-01-10
Functional tests - Fix other options test

nesrineabdmouleh:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix the min value of invoice number
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | TEST_PATH="functional/BO/orders/invoices/03_invoiceOptions/06_otherOptions" URL_FO=yourFoURL npm run specific-test

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17092)

### DIFF
--- a/tests/puppeteer/campaigns/data/faker/invoice.js
+++ b/tests/puppeteer/campaigns/data/faker/invoice.js
@@ -2,7 +2,7 @@ const faker = require('faker');
 
 module.exports = class Invoice {
   constructor(invoiceOptions = {}) {
-    this.invoiceNumber = invoiceOptions.invoiceNumber || faker.random.number({min: 20, max: 200}).toString();
+    this.invoiceNumber = invoiceOptions.invoiceNumber || faker.random.number({min: 100, max: 200}).toString();
     this.legalFreeText = invoiceOptions.legalFreeText || faker.lorem.sentence();
     this.footerText = invoiceOptions.footerText || faker.lorem.word();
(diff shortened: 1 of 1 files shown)


tests/puppeteer/campaigns/data/faker/invoice.js
```diff
@@ -2,7 +2,7 @@ const faker = require('faker');
 
 module.exports = class Invoice {
   constructor(invoiceOptions = {}) {
-    this.invoiceNumber = invoiceOptions.invoiceNumber || faker.random.number({min: 1, max: 200}).toString();
+    this.invoiceNumber = invoiceOptions.invoiceNumber || faker.random.number({min: 20, max: 200}).toString();
```
Progi1984: @nesrineabdmouleh Why do you change it to 20 ?

nesrineabdmouleh: There is an error when we put a number lower than the last invoice number, in our case, there are 8 invoice creations before this test, then we should enter a value upper than 9, I chose 20 because maybe we will add more tests in the future :)

PierreRambaud: Why not directly doing this?
```suggestion
    this.invoiceNumber = invoiceOptions.invoiceNumber || faker.random.number({min: 100, max: 200}).toString();
```

PierreRambaud (approved):

Issue #17099, opened 2020-01-10
Update the number of PrestaShop stores in the installer


LouiseBonnard:
When installing PrestaShop, the number of PrestaShop stores mentioned is _250,000_ while it should be _300,000_ as officially communicated.

<img width="618" alt="Capture d’écran 2020-01-10 à 16 06 38" src="https://user-images.githubusercontent.com/32565890/72163034-52dbd780-33c3-11ea-9488-c4ddebdf00d1.png">

Progi1984:
Fixed in #17100

PR #17100, opened 2020-01-10
Update the number of PrestaShop stores

LouiseBonnard:
| Questions     | Answers
| Branch?       | develop
| Description?  | Update the number of PrestaShop stores mentioned in the installer
| Type?         | improvement
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #17099
| How to test?  | Install PrestaShop and check the number of stores mentioned

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17100)


Progi1984 (approved):

PierreRambaud (approved):

matks:
If I'm not wrong, the hardcoded number will need to be translated everytime we modify the value. Maybe we could make it a parameter ?

PR #17123, opened 2020-01-12
Fix default country and country detection

mvorisek:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix default country and country detection
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | All tests should pass.

Improve country detection and make it working with GeoIP.

The original code has even a major typo:
https://github.com/PrestaShop/PrestaShop/commit/83ab48464929447c54ea1bd2104f3b07502185be#diff-a5aacc71d32391733634cc71652c7bc7L338

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17123)

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -647,14 +647,19 @@ class ToolsCore
 
     public static function getCountry($address = null)
     {
-        $req_id_country = Tools::getValue('id_country');
-        if (Validate::isInt($req_id_country) && (int) $req_id_country > 0 && !empty(Country::getIsoById((int) $req_id_country))) {
-            return (int) $req_id_country;
-        } elseif ($address !== null && (int) $address->id_country > 0) {
-            return (int) $address->id_country;
-        } else {
-            return (int) Configuration::get('PS_COUNTRY_DEFAULT');
+        $reqIdCountry = Tools::getValue('id_country');
+        if (Validate::isInt($reqIdCountry)
+            && (int) $reqIdCountry > 0
+            && !empty(Country::getIsoById((int) $reqIdCountry))
+        ) {
+            return (int) $reqIdCountry;
         }
+
+        if (!empty($address->id_country) && (int) $address->id_country > 0) {
+            return (int) $address->id_country;
+        }
+
+        return (int) Configuration::get('PS_COUNTRY_DEFAULT');
     }
 
     /**
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -647,20 +647,14 @@ class ToolsCore
 
     public static function getCountry($address = null)
     {
-        $id_country = (int) Tools::getValue('id_country');
-        if (!$id_country && isset($address, $address->id_country) && $address->id_country) {
-            $id_country = (int) $address->id_country;
-        } elseif (Configuration::get('PS_DETECT_COUNTRY') && isset($_SERVER['HTTP_ACCEPT_LANGUAGE'])) {
-            preg_match('#(?<=-)\w\w|\w\w(?!-)#', $_SERVER['HTTP_ACCEPT_LANGUAGE'], $array);
-            if (is_array($array) && isset($array[0]) && Validate::isLanguageIsoCode($array[0])) {
-                $id_country = (int) Country::getByIso($array[0], true);
-            }
+        $req_id_country = Tools::getValue('id_country');
+        if (Validate::isInt($req_id_country) && (int) $req_id_country > 0 && !empty(Country::getIsoById((int) $req_id_country))) {
+            return (int) $req_id_country;
+        } elseif ($address !== null && (int) $address->id_country > 0) {
+            return (int) $address->id_country;
+        } else {
+            return (int) Configuration::get('PS_COUNTRY_DEFAULT');
         }
-        if (!isset($id_country) || !$id_country) {
-            $id_country = (int) Configuration::get('PS_COUNTRY_DEFAULT');
-        }
-
-        return (int) $id_country;
     }
 
     /**
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -647,12 +647,12 @@ class ToolsCore
 
     public static function getCountry($address = null)
     {
-        $reqIdCountry = Tools::getValue('id_country');
-        if (Validate::isInt($reqIdCountry)
-            && (int) $reqIdCountry > 0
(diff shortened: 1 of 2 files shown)


classes/Tools.php
```diff
@@ -647,20 +647,14 @@ public static function switchLanguage(Context $context = null)
...
-            if (is_array($array) && isset($array[0]) && Validate::isLanguageIsoCode($array[0])) {
-                $id_country = (int) Country::getByIso($array[0], true);
-            }
-        }
-        if (!isset($id_country) || !$id_country) {
-            $id_country = (int) Configuration::get('PS_COUNTRY_DEFAULT');
+        $req_id_country = Tools::getValue('id_country');
+        if (Validate::isInt($req_id_country) && (int) $req_id_country > 0 && !empty(Country::getIsoById((int) $req_id_country))) {
+            return (int) $req_id_country;
+        } elseif ($address !== null && (int) $address->id_country > 0) {
```
```suggestion
        }

        if ($address !== null && (int) $address->id_country > 0) {
```

classes/Tools.php
```diff
@@ -647,20 +647,14 @@ public static function switchLanguage(Context $context = null)
...
+            return (int) $address->id_country;
+        } else {
```
```suggestion
        }
```

classes/Tools.php
mvorisek: I find if else ok, is that really an issue? We already has this integrated in our project, and I would like to prevent as much merge conflicts as possible.

PierreRambaud: It's not necessary, and it's easier to read without the esleif :)

mvorisek:
Can this PR be merged?

PierreRambaud:
@mvorisek Still comment unresolved :thinking:

mvorisek:
> 
> 
> @mvorisek Still comment unresolved 🤔

I just marked them as resolved.

PierreRambaud:
@mvorisek You marked  them as resolved, but they are not :thinking:

mvorisek:
> 
> 
> @mvorisek You marked them as resolved, but they are not 🤔

Only cosmetic change which we have resolved in the comments. In our company we have like 50+ bigger and smaller core changes - that is why I prefer to avoid later cosmetic changes and why I also asked if we can close this to have a free mind for another batch of PRs :)

PierreRambaud:
I'm in favor of the Boy scout rule `Leave your code better than you found it.`
So I can't approve something as it is :/

mvorisek:
> 
> 
> I'm in favor of the Boy scout rule `Leave your code better than you found it.`
> So I can't approve something as it is :/

I like it, this is why I also submitted PR 17120, but here I do not think that the if elseif else construct is bad (can you post a link to a code/style guildeline with some reasoning?) and from my point of view it will only create internal conflicts at our side. Anyway, this is my contribution and feel free to change the code/PR.

PierreRambaud:
@mvorisek I update your pull request and add the PS_DETECT_COUNTRY, tell me if everything is ok for you?

mvorisek:
> 
> 
> @mvorisek I update your pull request and add the PS_DETECT_COUNTRY, tell me if everything is ok for you?

PS_DETECT_COUNTRY / HTTP_ACCEPT_LANGUAGE should be done solely from Front Controller. The whole point is that:
- the GeoIP detection is there too (otherwise the country selection with GeoIP will work differently)
- we do not want any auto country/lang auto settings for other modules like API or backoffice

I meant you meant by "It's not necessary, and it's easier to read without the esleif :)" only cosmetic change :)

PierreRambaud:
Oh you're right about the PS_DETECT_COUNTRY, but since `Tools::getCountry`  is also use in ProductController, you should update your check here too, wdyt?

mvorisek:
> Oh you're right about the PS_DETECT_COUNTRY, but since Tools::getCountry is also use in ProductController, you should update your check here too, wdyt?

It should be fine, as most users currently use GeoIP and no issues are reported.

Your change is now fine, but still wondering if doubling the lines (at least in the first if codition) makes sense :)

matthieu-rolland (dismissed):
@PierreRambaud , the PR seems fine to me in its current state, what do you think?

PierreRambaud:
> @PierreRambaud , the PR seems fine to me in its current state, what do you think?

LGTM, but since I touche the code I can't approve :sweat_smile:

classes/Tools.php
```diff
@@ -647,20 +647,19 @@ public static function switchLanguage(Context $context = null)
...
     {
-        $id_country = (int) Tools::getValue('id_country');
-        if (!$id_country && isset($address, $address->id_country) && $address->id_country) {
-            $id_country = (int) $address->id_country;
-        } elseif (Configuration::get('PS_DETECT_COUNTRY') && isset($_SERVER['HTTP_ACCEPT_LANGUAGE'])) {
-            preg_match('#(?<=-)\w\w|\w\w(?!-)#', $_SERVER['HTTP_ACCEPT_LANGUAGE'], $array);
-            if (is_array($array) && isset($array[0]) && Validate::isLanguageIsoCode($array[0])) {
-                $id_country = (int) Country::getByIso($array[0], true);
-            }
+        $reqIdCountry = Tools::getValue('id_country');
```
sowbiba: why not just $countryId ? I'm doing the modification

sowbiba (changes requested):

sowbiba (approved):

mvorisek:
@matthieu-rolland can you reapprove?

Progi1984 (approved):

mvorisek:
> LGTM, but since I touche the code I can't approve 😅

@PierreRambaud after 1/2 year all approvals collected 🚀

PierreRambaud:
> > LGTM, but since I touche the code I can't approve sweat_smile
> 
> @PierreRambaud after 1/2 year all approvals collected

Oh yeah :sweat:

matks:
> > > LGTM, but since I touche the code I can't approve sweat_smile
> > 
> > 
> > @PierreRambaud after 1/2 year all approvals collected
> 
> Oh yeah 😓

We had a rough year 😅

matthieu-rolland (approved):

matthieu-rolland:
thank you for your contribution @mvorisek  👍

PR #17135, opened 2020-01-13
Orders > Product : Display "Stock Location" column only if it is not empty

### LINKED ISSUE #16391: Missing columns on new order page
**Describe the bug**
Missing columns on the new order page

**To Reproduce**
Steps to reproduce the behaviour:
1. Edit a product and fill the field stock location
2. Order this product in FO
3. See the order in BO
4. Stock location and Available quantity are missing

**Screenshots**
capture d'écran_2142

Legacy

capture d'écran_2141

**Solution**
The stock location column is hidden only when none of the products has a stock location only on the website. The column is only displayed at the corresponding pagination. If only a product on page 2 of the list has a stock location then the stock location column appears on page 2 and hidden in the other pages.
PrestaShop version: develop
PHP version: N/A

Progi1984:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Orders > Product : Display "Stock Location" column only if it is not empty
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16391
| How to test?  | * Build assets<br>* In BO, in an order, if stock location is defined, the stock location column is displayed else not<br>* Column is displayed in add form <br>* Works well

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17135)


admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
```diff
@@ -190,4 +196,18 @@ export default class OrderProductRenderer {
       $(OrderViewPageMap.productAddNewInvoiceInfo).addClass('d-none');
     }
   }
+
+  toggleColumnLocation(forceDisplay = null) {
+    let isColumnLocationDisplayed = false;
+    if (forceDisplay === null) {
+      $(OrderViewPageMap.productsCellLocation).filter('td').each(function() {
+        if ($(this).html() !== '') {
+          isColumnLocationDisplayed = true;
```
PierreRambaud, line 205: No need to continue
```suggestion
          isColumnLocationDisplayed = true;
          return false;
```

Progi1984: :heavy_check_mark: Thanks

admin-dev/themes/new-theme/js/pages/order/OrderViewPageMap.js
```diff
@@ -71,6 +71,7 @@ export default {
   productsTablePaginationTemplate: '#orderProductsTablePagination .page-item.d-none',
   productsTableRow: productId => `#orderProduct_${productId}`,
   productsTableRowEdited: productId => `#editOrderProduct_${productId}`,
+  productsCellLocation: '.cellProductLocation',
```
PierreRambaud:
```suggestion
  productsCellLocation: '.cell-product-location',
```

Progi1984: Note : https://codeguide.co/#css-classes

PierreRambaud (dismissed):

Robin-Fischer-PS:
Hi @Progi1984 !

Two issues on this PR : 

- When adding a product with combination, the stock location is not displayed in the add form. It is however well displayed once added.
- When there is a pagination, according to the main issue, the stock location column should not be displayed if on this page, there is no stock location.

Thanks :)

admin-dev/themes/new-theme/js/pages/order/view/order-product-add.js
```diff
@@ -135,9 +136,10 @@ export default class OrderProductAdd {
   setCombinations(combinations) {
     this.combinationsSelect.empty();
     Object.values(combinations).forEach((val) => {
-      this.combinationsSelect.append(`<option value="${val.attributeCombinationId}" data-price-tax-excluded="${val.priceTaxExcluded}" data-price-tax-included="${val.priceTaxIncluded}" data-stock="${val.stock}">${val.attribute}</option>`);
+      this.combinationsSelect.append(`<option value="${val.attributeCombinationId}" data-price-tax-excluded="${val.priceTaxExcluded}" data-price-tax-included="${val.priceTaxIncluded}" data-stock="${val.stock}" data-location="${val.location}" >${val.attribute}</option>`);
```
jolelievre: Remove extra space after data-location

matks (dismissed):

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
```diff
@@ -22,7 +22,9 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  *#}
-
+{% if isColumnLocationDisplayed is not defined %}
+  {% set isColumnLocationDisplayed=(product.availableQuantity is empty) %}
```
jolelievre: So the column is displayed when there is no stock? Shouldn't it be the opposite?

jolelievre: In products.html.twig you check `product.location is not empty` which seems more logical

jolelievre: I guess you add the `isColumnLocationDisplayed ` here because when adding a product the variable is not defined like it is in the products loop
The it would be cleaner to define this variable in the addProduct controller

admin-dev/themes/new-theme/js/pages/order/view/order-product-add.js
Progi1984: :heavy_check_mark:

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
Progi1984: The variable `product` is added in  `src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php` at l.542

Progi1984: Fixed for the bad variable used

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
```diff
@@ -22,7 +22,9 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  *#}
-
+{% if isColumnLocationDisplayed is not defined %}
+  {% set isColumnLocationDisplayed=(product.location is empty) %}
```
jolelievre:
```suggestion
  {% set isColumnLocationDisplayed=(product.location is not empty) %}
```

jolelievre: But again it's cleaner to do this in the controller (if it's related to the add product feature html return)

jolelievre (changes requested):

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
Progi1984: Fixed ;)

admin-dev/themes/new-theme/js/pages/order/view/order-view-page.js
```diff
@@ -199,6 +199,7 @@ export default class OrderViewPage {
 
     EventEmitter.on(OrderViewEventMap.productListPaginated, (event) => {
       this.orderProductRenderer.paginate(event.numPage);
+      this.orderProductRenderer.toggleColumnLocation();
```
jolelievre: This would be more logical to do this in the `orderProductRenderer.paginate` function

jolelievre (changes requested):
last change ;)

admin-dev/themes/new-theme/js/pages/order/view/order-view-page.js
Progi1984: Fixed

jolelievre (dismissed):
Thank @Progi1984

Robin-Fischer-PS:
Hi @Progi1984 !

The display of stock location seems to be OK, but when I install your PR, the display price and quantity of "basic" product when adding is broken ("NaN") : 
pr17135 NaN

I do not have this problem with combination product (I did not try any other type, no time unfortunately).

jolelievre (dismissed):

Robin-Fischer-PS:
Hi @Progi1984 

Still one small bug, when I add a new product, if no product already in the order has a stock location, and if there is more than 8 product (with pagination), the display of the product of the second page is shifted on the left (the "stock location" column is not taken into account for this page).
See screenshot, the two last product are the products on page 2 : 
pr17135 pagination decale

matks (dismissed):

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
```diff
@@ -200,15 +200,15 @@ export default class OrderProductRenderer {
...
+      $(OrderViewPageMap.productsCellLocationDisplayed).filter('td').each(function() {
         if ($(this).html() !== '') {
           isColumnLocationDisplayed = true;
           return false;
         }
       });
+      $(OrderViewPageMap.productsCellLocationDisplayed).toggleClass('d-none', !isColumnLocationDisplayed);
     } else {
-      isColumnLocationDisplayed = forceDisplay;
+      $(OrderViewPageMap.productsCellLocation).toggleClass('d-none', !forceDisplay);
```
jolelievre: I'd say you're gonna have the same problem here than with the previous bug, you must only toggle `productsCellLocationDisplayed`

jolelievre (changes requested):

jolelievre (dismissed):
Nice job @Progi1984

PierreRambaud (dismissed):
Don't forget to build assets please :)

Robin-Fischer-PS:
Hi @Progi1984 ! Could you resolve the conflicts on the two files ? Thx !

Progi1984:
@Robin-Fischer-PS Fixed

jolelievre (approved):

matthieu-rolland (approved):

PR #17140, opened 2020-01-13
Update modules to latest

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Update native modules to latest version
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | N/A
| How to test?  | Install shop, it works

## Updated modules

- prestashop/ps_contactinfo: v3.1.0 -> v3.2.0
- prestashop/ps_currencyselector: v2.0.0 -> v2.0.1
- prestashop/ps_emailsubscription: v2.3.0 -> v2.5.0
- prestashop/ps_facetedsearch: v3.0.6 -> v3.4.1
- prestashop/ps_linklist: 3.0.3 -> 3.0.4
- prestashop/ps_shoppingcart: v2.0.2 -> v2.0.3
- prestashop/ps_themecusto: v1.0.10 -> v1.1.0

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17140)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3597,16 +3597,16 @@
         },
         {
             "name": "prestashop/ps_contactinfo",
-            "version": "v3.1.0",
+            "version": "v3.2.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/PrestaShop/ps_contactinfo.git",
-                "reference": "e28551a9c857262b214775312ce195818933bd7e"
+                "reference": "407bd65291e8f3c64084ab267af1d8c9476872ea"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/PrestaShop/ps_contactinfo/zipball/e28551a9c857262b214775312ce195818933bd7e",
-                "reference": "e28551a9c857262b214775312ce195818933bd7e",
+                "url": "https://api.github.com/repos/PrestaShop/ps_contactinfo/zipball/407bd65291e8f3c64084ab267af1d8c9476872ea",
+                "reference": "407bd65291e8f3c64084ab267af1d8c9476872ea",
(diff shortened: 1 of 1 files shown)


Progi1984 (approved):

PierreRambaud:
Could we update composer.json too?

eternoendless:
> Could we update composer.json too?

IMO it's better to keep constraints for native modules up to the major version. That way, it's easier to update them when we bump them to the next major version. Or do you reckon there's a good reason for setting the exact version?

PierreRambaud:
It's only because the composer.lock is updated depending on who trigger it. 
This means, depending on the environment we can have different result, more if someone built it with PHP7.4, when other will update it with 7.1. We already have some issues because of this.
Put the most accurate version will help us to prevent "surprise" downgrade.
More, we must use the composer.json as an app, not as a package, so we can be very specific.

eternoendless:
I don't think you can be downgraded accidentally if you have a composer.lock

PierreRambaud:
@eternoendless You can if you run `composer update`, it happened when we built `twig` on `5.6` and `7.3`, the `composer.lock` was changing everytime (same with swift mailer)

matks:
🆙 Please reach an agreement @eternoendless @PierreRambaud 

Else we'll have you two fight until there is only one standing 👀

PierreRambaud:
@matks Let's do it on Tuesday? I'm ready :muscle:

Progi1984:
@eternoendless What do you want we do with this PR ?

PierreRambaud:
We can close this PR, the update has already been done.

PR #17155, opened 2020-01-14
Refacto partial refund along with behat tests

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Add behat test for partial refund in order to allow a refacto
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | partially #15820
| How to test?  | Behat tests must be green in travis

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17155)

### DIFF
--- a/src/Adapter/Order/AbstractOrderHandler.php
+++ b/src/Adapter/Order/AbstractOrderHandler.php
@@ -26,6 +26,8 @@
 
 namespace PrestaShop\PrestaShop\Adapter\Order;
 
+use Customer;
+use Group;
 use Order;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Order\ValueObject\OrderId;
@@ -52,5 +54,19 @@ abstract class AbstractOrderHandler
         }
 
         return $order;
+    }
+
+    /**
+     * @param Order $order
+     *
+     * @return bool
+     */
+    protected function isTaxIncludedInOrder(Order $order): bool
+    {
+        $customer = new Customer($order->id_customer);
+
+        $taxCalculationMethod = Group::getPriceDisplayMethod((int) $customer->id_default_group);
+
+        return ($taxCalculationMethod == PS_TAX_INC);
     }
 }
--- a/src/Adapter/Order/AbstractOrderHandler.php
+++ b/src/Adapter/Order/AbstractOrderHandler.php
@@ -67,6 +67,6 @@ abstract class AbstractOrderHandler
 
         $taxCalculationMethod = Group::getPriceDisplayMethod((int) $customer->id_default_group);
 
-        return ($taxCalculationMethod == PS_TAX_INC);
+        return $taxCalculationMethod == PS_TAX_INC;
     }
 }
--- a/src/Adapter/Order/AbstractOrderHandler.php
+++ b/src/Adapter/Order/AbstractOrderHandler.php
@@ -67,6 +67,6 @@ abstract class AbstractOrderHandler
 
         $taxCalculationMethod = Group::getPriceDisplayMethod((int) $customer->id_default_group);
 
-        return $taxCalculationMethod == PS_TAX_INC;
+        return $taxCalculationMethod === PS_TAX_INC;
     }
 }
--- a/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
@@ -169,6 +169,7 @@ final class AddOrderFromBackOfficeHandler implements AddOrderFromBackOfficeHandl
      * @param Cart $cart
      *
      * @return Country
+     *
      * @throws \PrestaShopDatabaseException
      * @throws \PrestaShopException
      */
--- a/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
@@ -176,7 +176,7 @@ final class AddOrderFromBackOfficeHandler implements AddOrderFromBackOfficeHandl
     private function getTaxCountry(Cart $cart)
     {
         $taxAddressType = Configuration::get('PS_TAX_ADDRESS_TYPE');
-        $taxAddressId = $cart->{$taxAddressType};
+        $taxAddressId = property_exists($cart, $taxAddressType) ? $cart->{$taxAddressType} : $cart->id_address_delivery;
         $taxAddress = new Address($taxAddressId);
 
         return new Country($taxAddress->id_country);
--- a/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
@@ -86,8 +86,8 @@ final class AddOrderFromBackOfficeHandler implements AddOrderFromBackOfficeHandl
             ->setCurrency(new Currency($cart->id_currency))
             ->setCustomer(new Customer($cart->id_customer))
             ->setLanguage(new Language($cart->id_lang))
-            //@TODO: investigate further which address should be used when setting context country (invoice or delivery)
-            ->setCountry(new Country((new Address($cart->id_address_invoice))->id_country));
+            ->setCountry($this->getTaxCountry($cart))
+        ;
 
         $translator = Context::getContext()->getTranslator();
         $employee = new Employee($command->getEmployeeId()->getValue());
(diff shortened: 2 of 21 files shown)


src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
```diff
@@ -103,18 +98,16 @@ public function handle(IssuePartialRefundCommand $command)
...
+                $productRefundAmount = $isTaxIncluded ?
                     $orderDetail->unit_price_tax_excl :
                     $orderDetail->unit_price_tax_incl;
-                $refund['amount'] *= $quantity;
+                $productRefundAmount *= $quantity;
             } else {
-                $refund['amount'] = $command->getTaxMethod() ?
-                    $taxCalculator->removeTaxes($refund['amount']) :
-                    $taxCalculator->addTaxes($refund['amount']);
+                $productRefundAmount = $orderDetailRefund->getRefundAmount();
```
Progi1984:
```suggestion
                $productRefundAmount = $orderDetailRefund->getRefundAmount() ?? (($isTaxIncluded ? $orderDetail->unit_price_tax_excl : $orderDetail->unit_price_tax_incl) * $quantity);
```

jolelievre: Usually I like ternary assignation, but in this case it's a double ternary and it deals with a code/feature that is already complicated enough to grasp So I would prefer to keep the explicit notation, in order to focus on the expected behaviour

Progi1984: :ok_hand:

Progi1984 (dismissed):

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -53,4 +55,18 @@ protected function getOrderObject(OrderId $orderId)
...
+     *
+     * @return bool
+     */
+    protected function isTaxIncludedInOrder(Order $order): bool
+    {
+        $customer = new Customer($order->id_customer);
+
+        $taxCalculationMethod = Group::getPriceDisplayMethod((int) $customer->id_default_group);
+
+        return $taxCalculationMethod == PS_TAX_INC;
```
PierreRambaud:
```suggestion
        return $taxCalculationMethod === PS_TAX_INC;
```

src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
```diff
@@ -164,4 +164,21 @@ private function assertAddressesAreNotDeleted(Cart $cart)
...
+     *
+     * @return Country
+     *
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
+    private function getTaxCountry(Cart $cart)
+    {
+        $taxAddressType = Configuration::get('PS_TAX_ADDRESS_TYPE');
+        $taxAddressId = $cart->{$taxAddressType};
```
PierreRambaud: Since `$taxAddressType` can be set by configuration, you shoud check if the attribute is available  :thinking:

src/Core/Domain/Order/ValueObject/OrderDetailRefund.php
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * @var int
+     */
+    private $productQuantity;
+
+    /**
+     * @var float|null
+     */
+    private $refundAmount;
```
PierreRambaud: why not `$amountRefunded` like in previous files? :thinking:

src/Core/Domain/Order/ValueObject/OrderDetailRefund.php
```diff
@@ -0,0 +1,143 @@
...
+     * @throws EmptyRefundAmountException
+     * @throws OrderException
+     */
+    public static function createPartialRefund(int $orderDetailId, int $productQuantity, float $refundAmount): self
+    {
+        if (0 >= $refundAmount) {
+            throw new EmptyRefundAmountException();
+        }
+
+        return new self($orderDetailId, $productQuantity, $refundAmount);
```
PierreRambaud: Is this method really needed? Why not using the constructor, and add the check above in a validator method?

src/Core/Domain/Order/ValueObject/OrderDetailRefund.php
```diff
@@ -0,0 +1,143 @@
...
+    {
+        return $this->refundAmount;
+    }
+
+    /**
+     * @param int $orderDetailId
+     *
+     * @throws OrderException
+     */
+    private function assertIntegerIsGreaterThanZero(int $orderDetailId)
```
PierreRambaud:
```suggestion
    private function assertOrderDetailIdsGreaterThanZero(int $orderDetailId)
```
if we look the code, you're not checking for integer, but for `$orderDetailId` (same in the exception message)

tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
```diff
@@ -137,6 +138,28 @@ public function assertPrivateNoteAboutCustomer($reference, $privateNote)
...
+    {
+        /** @var Customer $customer */
+        $customer = SharedStorage::getStorage()->get($reference);
+        $cartRules = CartRule::getCustomerCartRules((int) Configuration::get('PS_LANG_DEFAULT'), $customer->id, true);
+        if (empty($cartRules)) {
+            throw new RuntimeException('Cannot find any cart rules for customer');
+        }
+
+        $voucher = $cartRules[0];
+        if ($voucherAmount != (float) $voucher['reduction_amount']) {
```
PierreRambaud:
```suggestion
        if ($voucherAmount !== (float) $voucher['reduction_amount']) {
```

src/Adapter/Order/AbstractOrderHandler.php
jolelievre: Done

src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
jolelievre: Yep, good idea

src/Core/Domain/Order/ValueObject/OrderDetailRefund.php
jolelievre: Done

src/Core/Domain/Order/ValueObject/OrderDetailRefund.php
jolelievre: It's not mandatory, but the idea is to close the constructor access, and use static functions which clearly define the context because this whole feature is a bit difficult to understand

A partial refund MUST have a defined amount, just like a standard refund has NO defined amount (it is the product price by design), so I wanted the available code to translate these ideas

This could be done through different ValueObject classes as well (`OrderDetailPartialRefund extends OrderDetailStandardRefund` or vice versa), but it seemed overkill

src/Core/Domain/Order/ValueObject/OrderDetailRefund.php
jolelievre: Done

tests/Integration/Behaviour/Features/Context/CustomerFeatureContext.php
jolelievre: Done

jolelievre (commented):
Thanks @PierreRambaud I addresses your feedback
Except the one about the static functions, this was intended, but I'm open to discussion

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -596,18 +584,36 @@ private function getOrderDocuments(Order $order): OrderDocumentsForViewing
...
+                $number = sprintf(
+                    '%s%06d',
+                    Configuration::get('PS_CREDIT_SLIP_PREFIX', $this->contextLanguageId),
+                    $document->id
+                );
                 $amount = $this->locale->formatPrice(
                     $document->total_products_tax_incl + $document->total_shipping_tax_incl,
                     $currency->iso_code
                 );
+                $numericAmount = $document->total_products_tax_incl + $document->total_shipping_tax_incl;
```
matthieu-rolland, line 608: Just curious: was the numericalAmount wrong in some cases before this modification? 🤔

matthieu-rolland: got my answer 👍

jolelievre: Yes it was missing for some documents, and inconsistent For example the formatted value was the sum of  `$document->total_products_tax_incl + $document->total_shipping_tax_incl` but the numeric amount was just `$document->total_products_tax_incl`

matthieu-rolland (approved):
lgtm !

PR #17169, opened 2020-01-14
Refacto partial refund handler

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Split partial refund handler into three reusable services `OrderRefundCalculator` `OrderSlipCreator` and `VoucherGenerator`
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | partially #15820 
| How to test?  | This PR does require QA validation, it is mostly refacto so the tests need to be green in travis (the whole behaviour will be validate by QA team in the frontend integration PR anyway)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17169)

### DIFF
--- a/src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
+++ b/src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
@@ -36,18 +36,15 @@ use Mail;
 use OrderCarrier;
 use OrderDetail;
 use OrderSlip;
+use PrestaShop\PrestaShop\Adapter\Order\Refund\OrderRefundCalculator;
+use PrestaShop\PrestaShop\Adapter\Order\Refund\OrderRefundDetail;
 use PrestaShop\PrestaShop\Core\Domain\Order\Command\IssuePartialRefundCommand;
 use PrestaShop\PrestaShop\Core\Domain\Order\CommandHandler\IssuePartialRefundHandlerInterface;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
-use PrestaShop\PrestaShop\Core\Domain\Order\Exception\EmptyRefundQuantityException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\EmptyRefundAmountException;
-use PrestaShop\PrestaShop\Core\Domain\Order\ValueObject\OrderDetailRefund;
-use PrestaShop\PrestaShop\Core\Domain\Order\VoucherRefundType;
 use PrestaShop\PrestaShop\Core\Localization\Locale;
 use StockAvailable;
 use Symfony\Component\Translation\TranslatorInterface;
-use Tax;
-use TaxCalculator;
 use Validate;
 
 /**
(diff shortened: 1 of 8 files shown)


src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
```diff
@@ -56,23 +43,41 @@
...
+     */
+    private $voucherGenerator;
+
+    /**
+     * @param ConfigurationInterface $configuration
+     * @param OrderRefundCalculator $orderRefundCalculator
+     * @param OrderSlipCreator $orderSlipCreator
+     * @param VoucherGenerator $voucherGenerator
+     */
+    public function __construct(
```
PierreRambaud, line 71: Even if the class is final, it's a BC :thinking:

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -0,0 +1,189 @@
...
+            $isTaxIncluded
+        );
+    }
+
+    /**
+     * @param array $orderDetailRefunds
+     *
+     * @return OrderDetail[]
+     *
+     * @throws \PrestaShopDatabaseException
```
PierreRambaud:
```suggestion
     * @throws PrestaShopDatabaseException
```

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -0,0 +1,189 @@
...
+     * @throws \PrestaShopException
```
PierreRambaud:
```suggestion
     * @throws PrestaShopException
```

src/Adapter/Order/Refund/OrderRefundDetail.php
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * @param int $orderDetailId
+     *
+     * @return OrderDetail|null
+     */
+    public function getOrderDetailById(int $orderDetailId): ?OrderDetail
+    {
+        return isset($this->orderDetails[$orderDetailId]) ? $this->orderDetails[$orderDetailId] : null;
```
PierreRambaud:
```suggestion
        return $this->orderDetails[$orderDetailId] ?: null;
```

src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
jolelievre: Mmm, I wouldn't say it's a BC for a few reasons:
- the class is in the `Adapter` namespace, so it's supposed to respect the interface but the implementation is not fixed
- it's not supposed to be extended, or used "manually" it's a service so it should be retrieved via the container
- it's not a service that is supposed to be accessed directly, it's an handler that is automatically called via the command bus when its command is launched
- last but not least, it was introduced in 177, so no previous version had this class hence it can't break anything

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: Done

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: Done

src/Adapter/Order/Refund/OrderRefundDetail.php
jolelievre: I tried but if the index is not present it throws a `Notice: Undefined offset`

jolelievre (commented):
Thanks @PierreRambaud most feedback addressed

src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
PierreRambaud: Not a BC only for the latest reason :sweat_smile:

src/Adapter/Order/Refund/OrderRefundDetail.php
PierreRambaud: Oh you're right, I mean :sweat_smile: 
```suggestion
        return $this->orderDetails[$orderDetailId] ?? null;
```

matthieu-rolland (approved):
lgtm

PierreRambaud (approved):

Progi1984 (approved):

Issue #17085, opened 2020-01-09
 Migration - Create order bugs


Robin-Fischer-PS:
## What's left

- #17179

## Done / cannot reproduce

**Describe the bug**
Following the merge of https://github.com/PrestaShop/PrestaShop/pull/16207, here is the list of the remaining bugs on this page :

1. Voucher with gift product does not work : The gift should appear in the cart, with price replaced by "Gift", but nothing happens. However, once the order is created, the gift is in the order.

➡️ PR https://github.com/PrestaShop/PrestaShop/pull/17179

2. The combinations of product with combination are no longer loaded immediately on search : 
https://drive.google.com/file/d/1gWseCzlCvy7I4e1ijpRcnlafCaquZD1R/view?usp=sharing

➡️ fixed by https://github.com/PrestaShop/PrestaShop/pull/17465 ✅ 

3. When adding a "Free Shipping" voucher on an order (NOT created by checking the "Free Shipping" on the page), and then, checking the "Free Shipping" option on the order, the Shipping price is deducted twice.
See this 1.7.7.x invoice : 
[Invoice two free ship 177.pdf](https://github.com/PrestaShop/PrestaShop/files/4147295/Invoice.two.free.ship.177.pdf)
And the same action on 1.7.6 (free shipping is only deducted once) : 
[Invoice two free ship 176.pdf](https://github.com/PrestaShop/PrestaShop/files/4147322/Invoice.two.free.ship.176.pdf)

➡️ we failed to reproduce it

4. file name is not displayed in migrated when adding file for a custom product 
screenshot 1.7.7 : 
![upload_custom_file_name_1 7 7](https://user-images.githubusercontent.com/40635814/73676182-3b3afa80-46b4-11ea-9de9-7958691bdea2.png)
in legacy 1.7.6 : 
![upload_custom_file_name_1 7 6](https://user-images.githubusercontent.com/40635814/73676234-51e15180-46b4-11ea-8bfb-7bdde7d4bb74.png)

Please note that there might still be some bugs out there...

**Additional information**
PrestaShop version: 1.7.7.x
PHP version: N/A

matks:
For the record: "gift products" display is actually a totally different listing display
so this bug is actually a not-implemented feature that we did not know about 😛 

So it'll take some time to do
[...]

PR #17179, opened 2020-01-15
Fix free gift product in Order create page

### LINKED ISSUE #18119: [CREATE ORDER] Issues on vouchers adding gift products to the manually created cart
**ISSUE UPDATED**

Several issues related to vouchers with gift products added to the manually created order have appeared. To make it clear what are the expected behaviours, I've updated the [specifications on the create order page](https://github.com/PrestaShop/prestashop-specs/blob/MatShir-create-order/back-office/orders/orders/Create%20new%20order.md) (available in section III - voucher). Please review it and tell me if you have any comments or questions.

ping @matks @zuk3975 @Robin-Fischer-PS @MatShir 

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -- - - - - - - - - - - - - - - 

**Current behavior**

Small difference between Legacy and Migrated page.

If I add a voucher with a Gift and another discount (amount, percentage...), if I remove the voucher : 
- Legacy : Voucher AND gift are deleted from the order
- Migrated page : Only the voucher is deleted, the gift stays in the order

For reminder, when adding a voucher only with a Gift, it is not possible to remove it either in Legacy or in Migrated page. The voucher is not displayed in the list.

Screenrecord of Legacy behavior : 
https://drive.google.com/file/d/1ShPcag0h5oNLtV4EpYCokMvtuwu8jEcs/view?usp=sharing

Screenrecord of Migrated behavior : 
https://drive.google.com/file/d/1IO8hy7bJoj9vyY1KYRaMfck_nJdyQSjg/view?usp=sharing

**Expected behavior**

Voucher gifts with or without other cart rules actions should be displayed on the voucher list.
If the voucher is deleted, it deletes the gift from the product list.

_Originally posted by @Robin-Fischer-PS in https://github.com/PrestaShop/PrestaShop/pull/17179#issuecomment-598200694_

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Adds following feature to order create page in Back office - Once you add a cart rule with free-product action, the free product appears in products list, marked as a Gift and gift cart rule appears in vouchers list. Once you delete free gift voucher from cart, the gift product also dissapears. :warning: `There is another known issue - cart rule without code cannot be deleted, this pr doesn't fix that`.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #17085 (1st item), #18119
| How to test?  | Create a cart rule with action "send free gift = YES". Add that cart rule in your cart of Order create page in back office. You should see the gift product being added to a list marked as a gift instead of product price.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17179)

Screenshot from 2020-01-15 13-33-11


matks (dismissed):

admin-dev/themes/new-theme/js/pages/order/create/cart-rule-manager.js
```diff
@@ -31,6 +31,7 @@ import eventMap from '@pages/order/create/event-map';
 import Router from '@components/router';
 import SummaryRenderer from '@pages/order/create/summary-renderer';
 import ShippingRenderer from '@pages/order/create/shipping-renderer';
+import ProductRenderer from "./product-renderer";
```
PierreRambaud:
```suggestion
import ProductRenderer from './product-renderer';
```

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
```diff
@@ -47,10 +47,16 @@ export default class ProductRenderer {
     }
 
     const $productsTableRowTemplate = $($(createOrderMap.productsTableRowTemplate).html());
+    const $productsTableGiftRowTemplate = $($(createOrderMap.productsTableGiftRowTemplate).html());
 
     for (const key in products) {
       const product = products[key];
-      const $template = $productsTableRowTemplate.clone();
+
+      let $template = $productsTableRowTemplate.clone();
```
PierreRambaud: it's useless to clone twice, consume resource for nothing
```suggestion
      const $template = product.gift === true ? $productsTableGiftRowTemplate.clone() : $productsTableRowTemplate.clone();
```

src/Adapter/Cart/QueryHandler/GetCartInformationHandler.php
```diff
@@ -203,18 +203,13 @@ private function extractProductsFromLegacySummary(Cart $cart, array $legacySumma
...
-                \Tools::ps_round($product['price'], $currency->precision),
-                (int) $product['quantity'],
-                \Tools::ps_round($product['total'], $currency->precision),
-                $this->contextLink->getImageLink($product['link_rewrite'], $product['id_image'], 'small_default'),
-                $this->getProductCustomizedData($cart, $product)
-            );
+            $products[] = $this->buildCartProduct($cart, $currency, $product);
+        }
+
+        if (isset($legacySummary['gift_products'])) {
```
PierreRambaud:
```suggestion
        if (isset($legacySummary['gift_products']) && is_array($legacySummary['gift_products'])) {
```
I'm afraid about legacy data

src/Adapter/Cart/QueryHandler/GetCartInformationHandler.php
```diff
@@ -409,4 +404,21 @@ private function getProductCustomizedFieldsData(array $customizations, array $pr
 
         return $customizationFieldsData;
     }
+
+    private function buildCartProduct(Cart $cart, Currency $currency, array $product, bool $isGift = false): CartProduct
```
PierreRambaud:
```suggestion
    private function buildCartProduct(
        Cart $cart, 
        Currency $currency, 
        array $product, 
        bool $isGift = false
    ): CartProduct
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
```diff
@@ -218,6 +218,21 @@
   </tr>
 </script>
 
+<script id="products-table-gift-row-template" type="text/template">
```
PierreRambaud, line 228: Why not `<template>`? Should we support IE11?

PierreRambaud (changes requested):
Question and little refactoring

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
zuk3975: i guess we should :smile:

zuk3975:
> Question and little refactoring

@PierreRambaud thanks for feedback. Addressed your comments

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
PierreRambaud: I hope this kind of template is working everywhere :)

PierreRambaud (dismissed):

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
zuk3975: If Im right you offered to use this instead of text/html that i was using before in some other pr :smile:

PierreRambaud: You're maybe right :sweat_smile:

matks (dismissed):

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
```diff
@@ -47,10 +47,13 @@ export default class ProductRenderer {
     }
 
     const $productsTableRowTemplate = $($(createOrderMap.productsTableRowTemplate).html());
+    const $productsTableGiftRowTemplate = $($(createOrderMap.productsTableGiftRowTemplate).html());
 
     for (const key in products) {
       const product = products[key];
-      const $template = $productsTableRowTemplate.clone();
+
+      const $template = product.gift === true ? $productsTableGiftRowTemplate.clone() : $productsTableRowTemplate.clone();
```
Progi1984:
```suggestion
      const $template = product.gift === true ? $(createOrderMap.productsTableGiftRowTemplate).clone() : $(createOrderMap.productsTableRowTemplate).clone();
```

Progi1984 (commented):
A small feedback @zuk3975

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
matks: Why ? Your suggestion is making the line longer 😮 it's harder to read now.
Why clutter this nice statement with meaningful and shortened variable names ?

And why block the PR for such a small thing 😞

Progi1984: Because it's a memory optimization : that avoids to recreate a second jquery object for the same element. With this change, we can remove `$productsTableGiftRowTemplate` which become unuseful and memory hungry.

matks: Source: http://www.commitstrip.com/en/2015/07/23/for-a-few-bytes-more/

matks: Also: [Make it work, make it right, make it fast](https://tknilsson.com/2018/05/25/programmer-friday-make-it-work-make-it-right-make-it-fast/)

The "make it fast" should happen **at the end**. Right now a lot of code on PrestaShop does not work, and even more is not right.

Moreover, the provided article clearly states it:
_"When you start to optimize for speed, your codebase often takes on more complexity and becomes harder to understand and work with. A highly optimized loop is harder to read. A cache adds complexity and moving parts._

_Many software systems only need to work, they don’t require any special tuning for performance. Consider the usage scenario, the target audience, and do performance testing to figure out if more performance is necessary."_

I think this optimization is decreasing the code readability and that the memory consumption benefit is insignificant. Moreover you are blocking this PR which is a must-have item for 1.7.7 for such an insignificant part of the PR 😞 . Please think about it, I'm asking you to change your mind 🙏  .

matthieu-rolland: I think both of you are right, on different perspectives. ;) 

It's true that @zuk3975 's proposal makes it easier to read, it's also true that it increases browser's memory consumption, even though it's probably not that bad (given this is BO, not FO).

A third option, that would both keep @Progi1984 's optimization and @zuk3975 's readability, would be to extract the condition in a properly named function, for example **getProductsTemplate**, since the purpose of this condition is to define products's template (maybe you can think of a better name?)

```js
  const $template = this._getProductsTemplate();
```

And the function:

```js
  _getProductsTemplate() {
    return product.gift === true ? $(createOrderMap.productsTableGiftRowTemplate).clone() : $(createOrderMap.productsTableRowTemplate).clone();
  }
```
Extracting hard to read conditions into properly named functions is generally accepted as a good practice (notably in the book **clean code**).

But I really think we are nitpicking over tiny details here 😃 😃

jolelievre: Yep @matthieu-rolland proposal seems the cleaner, the code readability is important but jquery objects are quite heavy and shouldn't be created without reason.

jolelievre: The method name might be something like `cloneProductTemplate(product)` though

Progi1984: I approve it ;) Thanks @matthieu-rolland

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
```diff
@@ -218,6 +218,21 @@
...
+  <tr>
+    <td><img class="js-product-image" src="" alt=""></td>
+    <td class="js-product-definition-td">
+      <p class="mb-0 js-product-name"></p>
+      <p class="js-product-attr"></p>
+    </td>
+    <td class="js-product-ref"></td>
+    <td>{{ 'Gift'|trans({}, 'Admin.Catalog.Feature') }}</td>
```
LouiseBonnard:
```suggestion
    <td>{{ 'Gift'|trans({}, 'Admin.Orderscustomers.Feature') }}</td>
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
```diff
@@ -218,6 +218,21 @@
...
+    <td class="js-product-gift-qty"></td>
+    <td class="js-product-total-price">{{ 'Gift'|trans({}, 'Admin.Catalog.Feature') }}</td>
```
LouiseBonnard:
```suggestion
    <td class="js-product-total-price">{{ 'Gift'|trans({}, 'Admin.Orderscustomers.Feature') }}</td>
```

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
```diff
@@ -536,6 +536,10 @@ private function getErrorMessages(Exception $e)
                     'Positive product quantity is required',
                     'Admin.Notifications.Error'
                 ),
+                CartConstraintException::UNCHANGED_QUANTITY => $this->trans(
+                    'Product quantity is already the same as in cart',
```
LouiseBonnard: What action leads to this notification?

zuk3975: Example:

- Product A exists in cart.

- Product A has quantity of 10.

- I search for Product A again and try to add it to cart with quantity of 10.

matks (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
LouiseBonnard:
```suggestion
                    'Same product quantity is already in cart',
```

Progi1984 (dismissed):

Progi1984:
@zuk3975 Pretty CI is :red_circle: ; Could you fix it ? And fixing this wording change https://github.com/PrestaShop/PrestaShop/pull/17179#discussion_r375861428 ? Thanks
@LouiseBonnard Could you check to wording ? Thanks :)

LouiseBonnard:
@Progi1984, it is all fine for me except this suggestion: https://github.com/PrestaShop/PrestaShop/pull/17179#discussion_r375861428. :-)

zuk3975:
@Progi1984 Fixed.

matks (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/17179/commits/07613459013fa55a9b49537973a31f6f427ed252

Progi1984 (dismissed):

zuk3975:
reopen to restart prettyCI

zuk3975:
reopen last try for prettyCI

matks (approved):

matks (approved):

matks (approved):

matks (approved):

matks (approved):

matks (dismissed):

NeOMakinG (dismissed):

PierreRambaud (dismissed):

Robin-Fischer-PS:
Hi @zuk3975 !

I have a small difference between Legacy and Migrated page.

If I add a voucher with a Gift and another discount (amount, percentage...), if I remove the voucher : 
- Legacy : Voucher AND gift are deleted from the order
- Migrated page : Only the voucher is deleted, the gift stays in the order

For reminder, when adding a voucher only with a Gift, it is not possible to remove it either in Legacy or in Migrated page.

Screenrecord of Legacy behavior : 
https://drive.google.com/file/d/1ShPcag0h5oNLtV4EpYCokMvtuwu8jEcs/view?usp=sharing

Screenrecord of Migrated behavior : 
https://drive.google.com/file/d/1IO8hy7bJoj9vyY1KYRaMfck_nJdyQSjg/view?usp=sharing

I would like a PM validation for the good behavior, since (in my opinion) both are "a bit strange".

@MatShir , I need you !

MatShir:
Thanks @Robin-Fischer-PS for your report. Indeed, the behaviors are not ideal. It is a bug and a regression. I created a new issue https://github.com/PrestaShop/PrestaShop/issues/18119

MatShir:
@zuk3975 since QA has validated this PR, would you prefer to fix the bug on this one or create a new one?

zuk3975:
> @zuk3975 since QA has validated this PR, would you prefer to fix the bug on this one or create a new one?

@MatShir , Ill stay on this one. Added wip label and referenced the issue already.

src/Core/Domain/Cart/QueryResult/CartInformation/CartProduct.php
```diff
@@ -197,4 +205,12 @@ public function getCustomization(): ?Customization
     {
         return $this->customization;
     }
+
+    /**
+     * @return bool
+     */
+    public function isGift(): bool
```
jolelievre, line 212: I don't see any use off this getter, and I also wonder where in the template or the javascript is this data injected?

admin-dev/themes/new-theme/js/pages/order/create/cart-rule-manager.js
```diff
@@ -90,6 +90,7 @@ export default class CartRuleManager {
     EventEmitter.on(eventMap.cartRuleAdded, (cartInfo) => {
       const cartIsEmpty = cartInfo.products.length === 0;
       this.cartRulesRenderer.renderCartRulesBlock(cartInfo.cartRules, cartIsEmpty);
+      this.productRenderer.renderList(cartInfo.products);
```
jolelievre, line 93: Why do you need to update the product list here? Does the cart rule affect the product rendering?

jolelievre: Oh I get it, it's cart rule that adds the gift product right? Hence the products re-rendering

jolelievre (commented):
Hi @zuk3975 is this PR still in WIP?

admin-dev/themes/new-theme/js/pages/order/create/cart-rule-manager.js
zuk3975: yes

zuk3975:
> Hi @zuk3975 is this PR still in WIP?

Yes, Its wip because of the last found bug.

src/Core/Domain/Cart/QueryResult/CartInformation/CartProduct.php
zuk3975: most of this class getters have no usage. I generated all of them i think, its basically a DTO which is used (after converting to json) in javascript.
I think json serializing would fail due to private properties without getters behind the scenes.

zuk3975:
WIP check if classes/Cart::getSummaryDetails changes didn't ruin other pages where its used.
EDIT: removed changes from legacy class. Retrieved gift cart rules in getCartInformationHandler instead.

zuk3975:
Done :heavy_check_mark:

Progi1984 (dismissed):

matks (dismissed):

Robin-Fischer-PS:
Hi @zuk3975 !

Four issues with the new behaviour : 

- If I have a paid product, and the same product, but gifted, in my cart, if I delete the paid product, it will also delete the gifted product (the voucher is still displayed). The deletion should only delete the paid product. See screenrecord at the bottom of the comment for this bug (between 01:00 and 01:20).

- If I have a gifted product in my cart (with quantity "1"), and then I try to add the same product, but paid, an error is displayed "Same product quantity is already in cart" and nothing is added.
If I have two gifted products (same product), and then I try to add the same product, but paid, the two gifted product are replaced by only one gifted product (and no paid product). See screenrecord at the bottom of the comment for this bug (between 00:20 and 00:55).

- When adding several vouchers that offers the same product as a gift, there is one line for each gift (like in Legacy, so OK), but the quantity displayed is wrong : The total quantity of this product is displayed, instead of "1".

On my example below, I added the "Best is yet to come Mug" once as a paid item, and four times as a gift (with 4 different vouchers). There is one line for each mug in the cart, with a quantity set to "5" each time, instead of being set to "1" for each. When I create the order, the total quantity for the mug is OK (5 products, and not 25).
pr17179 gift product bad quantity

- When a voucher effect is "Free Gift" + something else (Free shipping, Amount discount or percent discount), the voucher is displayed two times : One with the "gift" discount", and one for the other effect.
pr17179 duplicated vouchers

Screenrecord for these behaviours : 
https://drive.google.com/file/d/1Pz9-0xBO1a1FaKk2qmdFyWSa-c25oLqM/view?usp=sharing

If something is not clear, don't hesitate to ping me, it's quite complex to explain sometimes ^^

zuk3975:
> Four issues with the new behaviour :

Oh my god, here we go again :sob:

matks:
@Robin-Fischer-PS and @MatShir This PR was started on 15th of January. We are once again in the nightmare scenario of endless QA feedbacks. We must do something, else we will continue consuming @zuk3975 motivation and time for weeks and 1.7.7 will never be out.

@Robin-Fischer-PS @MatShir Could you please write a dedicated set of specifications for this bug, and then we implement **exactly** what is described in the document, @Robin-Fischer-PS will test **exactly** what is described in the document and we can merge this ?

Take all the time you need to write this document, it will never be wasted time and it will save all of us lot of time.

zuk3975:
1 bug -WIP: behats passes, but getCartInformation in ajax call is still returning cartrules, which shouldnt exist.

zuk3975:
done :heavy_check_mark:

colinegin:
@zuk3975 did you check all behaviours I wrote in the specifications ? It's not limited to what Mateus had previously written in the issue.

zuk3975:
> @zuk3975 did you check all behaviours I wrote in the specifications ? It's not limited to what Mateus had previously written in the issue.

In progress. Found at least one bug already.

matks (dismissed):

matks (dismissed):

matks (dismissed):

classes/Cart.php
```diff
@@ -756,7 +756,7 @@ public function getProducts($refresh = false, $id_product = false, $id_country =
             $givenAwayProductsIds = [];
 
             // Do not recalculate in case of refresh
```
Progi1984: The comment seems not sync with the `if`

zuk3975: yea thanks for noticing. In fact i need an advice here. This if stmt was not making any sense, but it was made on purpose by someone, so idk if i didn't break anything :( . There is only one public action which changes the $splitFiftProductsQuantity to true (cart::getProductsWithSeparatedGifts), but it also forces the $refresh to be true, so the gifts were never recalulated..
So I've decided that was just a mistake and removed the "!" from refresh.
In this case I should also align the comment to absolute opposite (as in the code too.) "Recalculate in case of refresh" Do these changes make sense to you?

Progi1984 (commented):
A small feedback

classes/Cart.php
Progi1984: For me, yes :+1:

zuk3975:
> A small feedback

deleted above mentioned useless comment from code.

matks (dismissed):

classes/Cart.php
```diff
@@ -3838,7 +3837,7 @@ public function getSummaryDetails($id_lang = null, $refresh = false)
 
             if ($cart_rule['gift_product']) {
                 foreach ($products as $key => &$product) {
-                    if (empty($product['gift']) && $product['id_product'] == $cart_rule['gift_product'] && $product['id_product_attribute'] == $cart_rule['gift_product_attribute']) {
+                    if (empty($product['is_gift']) && $product['id_product'] == $cart_rule['gift_product'] && $product['id_product_attribute'] == $cart_rule['gift_product_attribute']) {
```
jolelievre, line 3839: Isn't this a BC break?

zuk3975: not 100 % sure, but it seems just a mistake. Everywhere else is "is_gift", thats why this if stmt seems to be impossible to break through.

matks (dismissed):

classes/Cart.php
```diff
@@ -3866,7 +3865,7 @@ public function getSummaryDetails($id_lang = null, $refresh = false)
                         $gift_product['price_wt'] = 0;
                         $gift_product['total_wt'] = 0;
                         $gift_product['total'] = 0;
-                        $gift_product['gift'] = true;
+                        $gift_product['is_gift'] = true;
```
jolelievre, line 3868: To answer your comment https://github.com/PrestaShop/PrestaShop/pull/17179/files#r405261136
The problem is more on this line because `Product::getSummaryDetails` is used elsewhere, so if the array field `gift` becomes `is_gift` it changes the expected data returned by this function no?
Unless this was a bug to use `gift` and the expected result is actually to use `is_gift` ?

jolelievre: From what I can see in `Cart::getProducts` it seems like `is_gift` is expected you're right! But then I think there is the same mistake in CartRule https://github.com/PrestaShop/PrestaShop/blob/8415bca965c74ff96e211f987c98b648fa8bd0d8/classes/CartRule.php#L767

matks (dismissed):

classes/Cart.php
Progi1984: Ping @zuk3975

zuk3975: changed that too. ['gift'] to ['is_gift'].

matks (dismissed):

Progi1984 (dismissed):

Progi1984:
Hello @zuk3975 , thanks for your contribution, would you mind to rebase your pull request and rebuild assets before merging ? Thanks

matks:
@prestashop-issue-bot :trollface: :trollface: :trollface:

Progi1984 (approved):

jolelievre (approved):

Issue #16569, opened 2019-11-27
Error in the console when creating a supplier


khouloudbelguith:
**Describe the bug**
When we try to create a supplier without description & we save => an error in the console is displayed.
There is no alert that it is required.

**To Reproduce**
1. Go to BO => Catalog => Suppliers page
2. Create a supplier without a description & save
3. See an error in the console "An invalid form control with name='supplier[description][1]' is not focusable."
![image](https://user-images.githubusercontent.com/16067358/69744608-c93cb700-1140-11ea-8f4e-1e9670653ab2.png)

PS1.7.6.1 => OK
**Screenshots**
https://drive.google.com/file/d/1f3r_xo_gSlT9q1i18RtLaOEgILLg0dnh/view

**Additional information**
PrestaShop version: develop
PHP version: 7.1

Progi1984:
Fixed in #17180

PR #17180, opened 2020-01-15
Fix suppliers form description to be optional

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Suppliers form `description` field should have been optional according to legacy behavior. Furthermore the required description caused some conflict between tinyMCE and browser (can read more here https://stackoverflow.com/questions/22148080/an-invalid-form-control-with-name-is-not-focusable). This pr only adds required => false in SupplierType and that solves the issues.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #16569
| How to test?  | Try to add new supplier without filling description. It allows to create one, and doesn't print any error in console any more.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17180)

### DIFF
--- a/src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
+++ b/src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
@@ -137,6 +137,7 @@ class SupplierType extends TranslatorAwareType
                 ],
             ])
             ->add('description', TranslateType::class, [
+                'required' => false,
                 'type' => FormattedTextareaType::class,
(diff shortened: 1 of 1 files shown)


Progi1984 (approved):

PierreRambaud (approved):

Robin-Fischer-PS:
Tested, OK ! Thanks @zuk3975 :)

Progi1984:
Thanks @zuk3975

Issue #17032, opened 2020-01-06
Adresses page - Required Fields cannot be selected which occurs issues in the BO & the FO


khouloudbelguith:
**Describe the bug**
In the BO => Customers => Addresses page => the required Fields cannot be selected which occurs issues in the BO & the FO

**To Reproduce**
1. Go to BO => Customers => Addresses page
2. Click on the button "Set required fields for this section"
3. Select all options
4. Save
5. A green alert is displayed
6. Click on the button "Set required fields for this section" again => the files are not selected => **NOK**
https://drive.google.com/file/d/1sdn7gWMhIA42IbjrH0zoA-rhE93l6RBP/view
7. In the **BO** => Click Add a new Address & set all required fields with the required symbol
8. Save => see error
![image](https://user-images.githubusercontent.com/16067358/71822394-06c42680-3095-11ea-9fe7-b334640e88f4.png)
9. In the **FO** => try to create an order with `John Doe` => In the checkout page, we have an exception
![image](https://user-images.githubusercontent.com/16067358/71823399-66233600-3097-11ea-884e-31f4b98ff8d6.png)

**Additional information**
PrestaShop version: develop
PHP version: 7.1

zuk3975:
I think this is related https://github.com/PrestaShop/PrestaShop/issues/9845

MatShir:
fixed by #17526

colinegin:
Hello @khouloudbelguith the bug is on develop or on the 1.7.7.x branch ?

khouloudbelguith:
Hi @colinegin,

Same issue with branch 1.7.7.x.

Thanks!

Progi1984:
Fixed by #17526

PR #17190, opened 2020-01-15
[WIP] Fix customer address required fields

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Required fields in customer address was not able to be saved. This pr fixes it by createing CustomerAddress object instead of Address according to legacy page (so it passes the correct object name to db). `There are still some behavior to consider`: currently if you disable mobile_phone the field will not appear in form at all instead of just being not required. All other fields even when selected as non required still appears in form, just doesn't have the required attribute.. So which should be the correct behavior? `And another issue`: postcode required attribute appears by using javascript and it depends from country, so if the country requires postcode should it override the Required fields selection? (some info for user would be great in any case).
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #17032
| How to test?  | GO BO Customers->addresses and select required fields whichever you want and save. Then go to Create new address and you should be able to create address without filling fields which are not required.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17190)

### DIFF
--- a/src/Adapter/Address/CommandHandler/SetRequiredFieldsForAddressHandler.php
+++ b/src/Adapter/Address/CommandHandler/SetRequiredFieldsForAddressHandler.php
@@ -26,7 +26,7 @@
 
 namespace PrestaShop\PrestaShop\Adapter\Address\CommandHandler;
 
-use Address;
+use CustomerAddress;
 use PrestaShop\PrestaShop\Core\Domain\Address\Command\SetRequiredFieldsForAddressCommand;
 use PrestaShop\PrestaShop\Core\Domain\Address\CommandHandler\SetRequiredFieldsForAddressHandlerInterface;
 use PrestaShop\PrestaShop\Core\Domain\Address\Exception\CannotSetRequiredFieldsForAddressException;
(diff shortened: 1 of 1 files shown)


zuk3975:
There are 2 questions that are stopping this pr, described in description

matks:
@MatShir We need your answers 😄

zuk3975:
> @MatShir We need your answers

As agreed during standup, I should get concrete specifications by 22 of January, so until then this pr stays wip

zuk3975:
@colinegin As I know you are preparing some specs for addresses regarding required fields.
Could you please also consider `Brands &Suppliers` addresses too, because (thanks to this issue https://github.com/PrestaShop/PrestaShop/issues/17268) I've noticed it isn't fully migrated and should consider required fields too (my bad)

It would be nice to know which addresses(customer, supplier, manufacturer) and how depends on required fields in `Customer -> Adddresses -> set Required Field` 
and with `Location->country->[select country to edit] -> required fields`.

According to legacy all of those addresses depends from changes of required fields in Customer->addresses.

colinegin:
Hello @zuk3975 , i'm sorry i did not have time to get back to you with some feedbacks about this issue. We will be working with Mateus on this today, we will get back to you asap.

zuk3975:
Moving this to related PR. https://github.com/PrestaShop/PrestaShop/pull/17293

Issue #17089, opened 2020-01-09
Migration - Product block on order page bugs


Robin-Fischer-PS:
**Describe the bug**
 Following the merge of https://github.com/PrestaShop/PrestaShop/pull/16150, here is the list of the remaining bugs on this page :

Adding several "small" discount with a total greater than the order total will raise another Symfony Exception, with a small difference : The last discount will still be added even if it raise an exception, but it will raise another exception when trying to delete it
* **Bug** :negative_squared_cross_mark: : I still have the exception when adding free shipping discount (it's now OK when adding only "amount" discounts) => This bug also exist in Legacy

* **New bug** :negative_squared_cross_mark: When there is a product which is out of stock, it is impossible to delete it from the order (this also happens in Legacy)
* **New bug** :negative_squared_cross_mark: When editing a product, if the quantity is set higher than available quantity, the field "available" should becomes red. Can you confirm @TristanLDD ? 
* **New bug** :negative_squared_cross_mark: It is not possible to edit a virtual product : https://drive.google.com/file/d/1QtMeLer98qEFD5jXEEO0kBUHswnr4Wcn/view?usp=sharing (this also happens in Legacy)
 * When adding a discount on an order on which at least one invoice has already been created, the user can choose the invoice on which the discount must apply. In the dropdown, in Legacy, each line is "[Invoice Number] - [Invoice Amount]. In migrated page, Invoice Amount is missing.
* A typo (in Legacy and migrated page) bug, on adding a discount popin, under "Apply to all invoices", in the help text :
"If you chooses to create this discount for all invoices, only one discount will be created per order invoice." should be "If you choose to create this discount for all invoices, only one discount will be created per order invoice." => This bug also exist in Legacy,.
* New bug : On pagination, when I have 8 product, if I add a 9th product, it will be added on the bottom of page 1, AND on page 2 (the sum below the product block is still OK). See screenshots : 
![pr16150 pagination 9e element](https://user-images.githubusercontent.com/55688534/72056147-56406780-32cc-11ea-8ab9-31e33bc53e02.png)
![pr16150 pagination 9e element 2](https://user-images.githubusercontent.com/55688534/72056160-5b9db200-32cc-11ea-8718-e03414624e17.png)
* When adding a product, if a new invoice is created, the "shipping block" displayed is not deleted after creating the invoice

Please note that there might still be some bugs out there...

**Additional information**
PrestaShop version: develop
PHP version: N/A

[...]

PR #17206, opened 2020-01-16
Fix order editing in order view page

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | 1. In Order view page was possible to update quantity without checking stockAvailability. This pr adds a check of product stock and does not allow to add more quantity than available in stock.(same behavior as adding a product to order). 2. It was also failing to update virtual product base price - this fixes it by disabling shpping price update for virtual products. 3. When applying discount  the invoice choice total value was missing- fixed by adding option in form. 4.Fix typo in discount creation Help 'If you chooses..' => 'if you choose...' after invoice choice field
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #17089
| How to test?  | use `admin-dev/index.php/sell/orders/orders` to reach migrated orders list. 1. Select one of orders to open orders view page (order must not be delivered, to be able to edit). Edit product and change its quantity higher than available. The `Available quantity text should become RED` and `save button disabled`. 2. Open orders with virtual products, update virtual product price. should work as expected. 3. Try to add discount on order which has invoice already created, you should see choices of invoices shown like '[invoice nr] - ['total']. 4. See discount creation Help 'If you chooses..' => 'if you choose...' after invoice choice field

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17206)

Screenshot from 2020-01-16 15-56-19


src/Core/Domain/Order/QueryResult/OrderProductForViewing.php
```diff
@@ -123,6 +123,11 @@ class OrderProductForViewing
      */
     private $orderInvoiceNumber;
 
+    /**
+     * @var bool
+     */
+    private $availableOutOfStock;
```
tdavidsonas88, line 129: Suggestion for future. For bool primitives suggest to use names like isAvailableOutOfStock or isOutOfStock. Then even without annotations it's clear that this variable is a boolean.

admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
```diff
@@ -39,13 +39,19 @@ export default class OrderProductEdit {
...
     this.priceTaxCalculator = new OrderPrices();
+    this.productEditSaveBtn = $(OrderViewPageMap.productEditSaveBtn);
+    this.quantityInput = $(OrderViewPageMap.productEditQuantityInput);
   }
 
   setupListener() {
     this.quantityInput.on('change keyup', (event) => {
       this.quantity = parseInt(event.target.value ? event.target.value : 0, 10);
-      this.availableText.html(this.totalQuantity - this.quantity);
+      const available = $(event.currentTarget).data('stock') - this.quantity;
```
PierreRambaud:
```suggestion
      const available = parseInt($(event.currentTarget).data('stock'), 10) - this.quantity;
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
```diff
@@ -46,7 +46,17 @@ export default class OrderProductRenderer {
...
+  editProductFromList(
+    orderDetailId,
+    quantity,
+    priceTaxIncl,
+    priceTaxExcl,
+    taxRate,
+    location,
+    availableQuantity,
+    availableOutOfStock,
+    orderInvoiceId
```
PierreRambaud:
```suggestion
    orderInvoiceId,
```

src/Core/Domain/Order/QueryResult/OrderProductForViewing.php
zuk3975: imo `is` should be added as a prefix in method only, not in proeperty itself. But I agree that this could be better named, like $availableWhenOutOfStock, but i left it like that unchanged for convienience because another object has the same property. Maybe I could rename them both :thinking:

tdavidsonas88: From where comes the rule that "is should be added as a prefix in method only, not in proeperty itself"?

Some suggestions: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

matks: I like @tdavidsonas88 suggestion 👍

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
matks: Euuuuh 😮 in a function call ? A trailing comma for last argument ? Even if it's syntaxically valid, is it something we want ?

matks:
Apart from @PierreRambaud feedback, code looks ✅ to me

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
zuk3975: It works in both cases and in both cases js linter is unhappy, so.. Its your call :smile:

matks: I will consider @PierreRambaud knowledge of JS code style better than me 👍

matks (approved):

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
PierreRambaud: eslint-airbnb recommandation :)

matks:
All bugs fixed in https://github.com/PrestaShop/PrestaShop/pull/17220

PR #17211, opened 2020-01-16
Reintegrate partial refund in the order view page

### LINKED ISSUE #15820: [ORDER PAGE] Partial Refund - 3/11

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Reintegrate partial refund in the order view page, add new fields into OrderDetail to store the refunded amount by product instead of relying on OrderSlip which is not always created As a result `OrderSlip::getProductSlipResume` is now deprecated in favor of the fields in OrderDetail
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | yes
| Fixed ticket? | Fixes #15820
| How to test?  | See issue ^^

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17211)


admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
```diff
@@ -184,7 +184,7 @@ export default class OrderProductRenderer {
   }
 
   toggleProductAddNewInvoiceInfo() {
-    if ($(OrderViewPageMap.productAddInvoiceSelect).val() == 0) {
+    if (0 === $(OrderViewPageMap.productAddInvoiceSelect).val()) {
```
PierreRambaud: Airbnb prefer this style :)
```suggestion
    if ($(OrderViewPageMap.productAddInvoiceSelect).val() === 0) {
```

admin-dev/themes/new-theme/js/pages/order/view/order-view-page.js
```diff
@@ -204,6 +204,29 @@ export default class OrderViewPage {
     });
   }
 
+  listenForRefund() {
+    $(OrderViewPageMap.displayPartialRefundBtn).on('click', () => {
+      this.orderProductRenderer.moveProductsPanelToRefundPosition();
+      $(OrderViewPageMap.toggleStandardRefundForm).hide();
```
PierreRambaud: Can you keep the same order for `toggleStandardRefundForm`, `togglePartialRefundForm` and `actionColumnElements` in the three functions?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
```diff
@@ -73,8 +73,22 @@
     </form>
   {% endif %}
   {% include '@PrestaShop/Admin/Sell/Order/Order/Blocks/View/extra_order_button_actions.html.twig' %}
+  {% if merchandiseReturnEnabled %}
+    {% if orderForViewing.hasInvoice() %}
```
PierreRambaud, line 77: Can be a single if no? :)

admin-dev/themes/new-theme/js/pages/order/OrderViewPageMap.js
```diff
@@ -113,9 +114,17 @@ export default {
...
-  cancelPartialRefundBtn: 'button.partial-refund-cancel',
   actionColumnElements: 'td.cellProductActions, th.product_actions',
-  togglePartialRefundForm: '.partial-refund:not(.hidden), .shipping-price, .refund-checkboxes-container',
-  productAddNewInvoiceInfo: '#addProductNewInvoiceInfo',
+  // Product cancel/refund elements
+  togglePartialRefundForm: '.cancel-product-element:not(.hidden):not(.shipping-refund), .cancel-product-amount',
+  toggleStandardRefundForm: '.cancel-product-element:not(.hidden):not(.shipping-refund), .cancel-product-selector',
+  displayPartialRefundBtn: 'button.partial-refund-display',
+  displayStandardRefundBtn: 'button.standard-refund-display',
+  cancelProductForm: 'form[name="cancel_product"]',
```
PierreRambaud:
```suggestion
  cancelProduct: {
    form: 'form[name="cancel_product"]',
    ...
  }
```

Adding a depth level is easier to read.

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,92 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+import Router from '@components/router';
+import OrderViewPageMap from '@pages/order/OrderViewPageMap';
+
+const $ = window.$;
```
PierreRambaud:
```suggestion
const {$} = window;
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,92 @@
...
+    this.cancelProductForm = $(OrderViewPageMap.cancelProductForm);
+    this.orderId = this.cancelProductForm.data('orderId');
+  }
+
+  showPartialRefund() {
+    $(OrderViewPageMap.toggleStandardRefundForm).hide();
+    $(OrderViewPageMap.togglePartialRefundForm).show();
+    $(OrderViewPageMap.actionColumnElements).hide();
+    this.listenForInputs();
+    console.log(this.orderId);
```
PierreRambaud: Nope

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,92 @@
...
+    $(OrderViewPageMap.cancelProductColumnHeader).html(actionName);
+  }
+
+  listenForInputs() {
+    $(OrderViewPageMap.cancelProductQuantityInput).off('change').on('change', (event) => {
+      const $productQuantityInput = $(event.target);
+      const $parentCell = $productQuantityInput.parents(OrderViewPageMap.cancelProductTableCell);
+      const $productAmount = $parentCell.find(OrderViewPageMap.cancelProductAmountInput);
+      const productQuantity = parseInt($productQuantityInput.val(), 10);
+      if (productQuantity === 0) {
```
PierreRambaud:
```suggestion
      if (productQuantity <= 0) {
```

src/Core/Form/IdentifiableObject/DataProvider/CancelProductFormDataProvider.php
```diff
@@ -56,10 +66,13 @@ public function getData($orderId)
     {
         /** @var OrderForViewing $orderForViewing */
         $orderForViewing = $this->queryBus->handle(new GetOrderForViewing((int) $orderId));
+        $computingPrecision = new ComputingPrecision();
```
PierreRambaud, line 69: Must be injected :thinking:

src/PrestaShopBundle/Form/Admin/Sell/Order/CancelProductType.php
```diff
@@ -56,24 +57,38 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                     'invalid_message' => $this->trans('This field is invalid, it must contain numeric values', 'Admin.Notifications.Error', []),
                     'required' => false,
                     'data' => 0,
+                    'scale' => 0,
                 ])
                 ->add('amount_' . $product->getOrderDetailId(), NumberType::class, [
                     'attr' => ['max' => $product->getTotalPrice(), 'class' => 'refund-amount'],
                     'label' => sprintf(
-                        '%s (%s)',
+                        "%s<br>(%s)",
```
PierreRambaud:
```suggestion
                        '%s<br>(%s)',
```

classes/order/OrderSlip.php
```diff
@@ -192,6 +192,19 @@ public static function getProductSlipResume($id_order_detail)
             WHERE `id_order_detail` = ' . (int) $id_order_detail);
     }
 
+    /**
+     * Get resume of all shipping refund for one order.
+     *
+     * @param $id_order
+     */
+    public static function getShippingSlipResume($id_order)
```
PierreRambaud:
```suggestion
    public static function getShippingSlipResume(int $idOrder)
```

classes/order/OrderSlip.php
```diff
@@ -192,6 +192,19 @@ public static function getProductSlipResume($id_order_detail)
             WHERE `id_order_detail` = ' . (int) $id_order_detail);
     }
 
+    /**
+     * Get resume of all shipping refund for one order.
+     *
+     * @param $id_order
```
PierreRambaud:
```suggestion
     * @param int $idOrder
```

classes/order/OrderSlip.php
```diff
@@ -192,6 +192,19 @@ public static function getProductSlipResume($id_order_detail)
...
+     */
+    public static function getShippingSlipResume($id_order)
+    {
+        return Db::getInstance(_PS_USE_SQL_SLAVE_)->getRow('
+            SELECT SUM(total_shipping_tax_incl) total_shipping_tax_incl, SUM(total_shipping_tax_excl) total_shipping_tax_excl, SUM(shipping_cost_amount) shipping_cost_amount
```
PierreRambaud:
```suggestion
            SELECT 
                SUM(total_shipping_tax_incl) total_shipping_tax_incl,
                SUM(total_shipping_tax_excl) total_shipping_tax_excl,
                SUM(shipping_cost_amount) shipping_cost_amount
```

classes/order/OrderSlip.php
```diff
@@ -192,6 +192,19 @@ public static function getProductSlipResume($id_order_detail)
...
+            FROM `' . _DB_PREFIX_ . 'order_slip`
+            WHERE `id_order` = ' . (int) $id_order);
```
PierreRambaud:
```suggestion
            WHERE `id_order` = ' . (int) $idOrder);
```

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -71,6 +72,12 @@ public function computeOrderFund(
 
         $shippingCostAmount = $shippingRefund ?: false;
         if ($shippingCostAmount > 0) {
+            $shippingMaxRefund = $isTaxIncluded ? $order->total_shipping_tax_incl : $order->total_shipping_tax_excl;
+            $shippingSlipResume = OrderSlip::getShippingSlipResume($order->id);
+            $shippingMaxRefund -= isset($shippingSlipResume['total_shipping_tax_incl']) ? $shippingSlipResume['total_shipping_tax_incl'] : 0;
```
PierreRambaud:
```suggestion
            $shippingMaxRefund -= $shippingSlipResume['total_shipping_tax_incl'] ?? 0;
```

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -134,20 +142,25 @@ private function flattenProductRefunds(array $orderDetailRefunds, bool $isTaxInc
                 'id_order_detail' => $orderDetailId,
             ];
 
+            // Compute max refund by product (based on quantity and already refunded amount)
+            $productMaxRefund = $isTaxIncluded ? $orderDetail->unit_price_tax_excl : $orderDetail->unit_price_tax_incl;
+            $productMaxRefund *= $quantity;
+            $productSlipResume = OrderSlip::getProductSlipResume($orderDetailId);
+            $productMaxRefund -= $isTaxIncluded ? $productSlipResume['amount_tax_incl'] : $productSlipResume['amount_tax_excl'];
```
PierreRambaud: look like a duplicate logic with line 75

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
jolelievre: Done, but it's too bad Yoda style is cool and Yoda style is safe ^^

classes/order/OrderSlip.php
jolelievre: Done

classes/order/OrderSlip.php
jolelievre: Done

classes/order/OrderSlip.php
jolelievre: Done

classes/order/OrderSlip.php
jolelievre: Done

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: Done

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: Mmm, the file has been modified so I don't see the duplicate, but maybe it's on another line now?

Unless you are referring to line 80 now but it's different because it's about shipping cost (`OrderSlip::getShippingSlipResume`) although the code looks alike indeed as it's inspired from this code

src/Core/Form/IdentifiableObject/DataProvider/CancelProductFormDataProvider.php
jolelievre: Probably, but then it would be worth a true refacto, so that the injected service can have a method like `$this->precisionComputing->getPrecisionByCurrencyId($currencyId)`
For now this code is present in many handlers so I'm not gonna refacto it in this PR

src/PrestaShopBundle/Form/Admin/Sell/Order/CancelProductType.php
jolelievre: Waiting for the cs fixer PR to update/fix this

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
jolelievre: No it's actually a portion a the full template, when merchandise return is enabled We can have three different buttons:
- stantard refund
- return products
- cancel products

Depending on the order state they are displayed or not, but they are all conditioned to the merchandise return, that's why it's not a single if
Besides this feature (standard refund) doesn't work yet but I needed it to test the elements toggling in advance

admin-dev/themes/new-theme/js/pages/order/OrderViewPageMap.js
jolelievre: Done

src/PrestaShopBundle/Form/Admin/Sell/Order/CancelProductType.php
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-view-page.js
jolelievre: Actually I can't, I added a comment to explain why `// Always start by hiding elements then show the others, since some elements are common`

atomiix (dismissed):
LGTM but probably need a second review

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  initForm(actionName, formAction) {
+    this.listenForInputs();
+    this.updateVoucherRefund();
+
+    this.cancelProductForm.attr('action', formAction);
+    $(OrderViewPageMap.cancelProduct.buttons.save).html(actionName);
+    $(OrderViewPageMap.cancelProduct.table.header).html(actionName);
+    $(OrderViewPageMap.cancelProduct.checkboxes.restock).attr('checked', this.orderDelivered);
```
PierreRambaud:
```suggestion
    $(OrderViewPageMap.cancelProduct.checkboxes.restock).prop('checked', this.orderDelivered);
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+    $(OrderViewPageMap.cancelProduct.checkboxes.creditSlip).attr('checked', true);
```
PierreRambaud:
```suggestion
    $(OrderViewPageMap.cancelProduct.checkboxes.creditSlip).prop('checked', true);
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  listenForInputs() {
+    $(OrderViewPageMap.cancelProduct.inputs.quantity).off('change').on('change', (event) => {
```
PierreRambaud: Is it mandatory to `off('change')`?

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+      if (productQuantity <= 0) {
+        $productAmount.val(0);
+        this.updateVoucherRefund();
+
+        return;
+      }
+      const priceFieldName = this.isTaxIncluded ? 'productPriceTaxIncl' : 'productPriceTaxExcl';
+      const productUnitPrice = parseFloat($productQuantityInput.data(priceFieldName));
+      const amountRefundable = parseFloat($productQuantityInput.data('amountRefundable'));
+      const guessedAmount = productUnitPrice * productQuantity < amountRefundable ?
```
PierreRambaud: Make it more readable
```suggestion
      const guessedAmount = (productUnitPrice * productQuantity) < amountRefundable ?
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+        productUnitPrice * productQuantity : amountRefundable;
```
PierreRambaud:
```suggestion
        (productUnitPrice * productQuantity) : amountRefundable;
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+      const amountValue = parseFloat($productAmount.val());
+      if ($productAmount.val() === '' || amountValue === 0 || amountValue > guessedAmount) {
+        $productAmount.val(guessedAmount);
+        this.updateVoucherRefund();
+      }
+    });
+
+    $(OrderViewPageMap.cancelProduct.inputs.amount).off('change').on('change', () => {
```
PierreRambaud: Same here about the `off`

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+      this.updateVoucherRefund();
+    });
+  }
+
+  updateVoucherRefund() {
+    let totalAmount = 0;
+    $(OrderViewPageMap.cancelProduct.inputs.amount).each((index, amount) => {
+      const floatValue = parseFloat(amount.value);
+      totalAmount += !isNaN(floatValue) ? floatValue : 0;
```
PierreRambaud:
```suggestion
      totalAmount += !Number.isNaN(floatValue) ? floatValue : 0;
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+    });
+
+    this.updateVoucherRefundTypeLabel($(OrderViewPageMap.cancelProduct.radios.voucherRefundType.productPrices), totalAmount);
```
PierreRambaud:
```suggestion
    this.updateVoucherRefundTypeLabel(
      $(OrderViewPageMap.cancelProduct.radios.voucherRefundType.productPrices), 
      totalAmount
    );
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+    const refundVoucherExcluded = totalAmount - this.discountsAmount;
+    this.updateVoucherRefundTypeLabel($(OrderViewPageMap.cancelProduct.radios.voucherRefundType.productPricesVoucherExcluded), refundVoucherExcluded);
```
PierreRambaud:
```suggestion
    this.updateVoucherRefundTypeLabel(
      $(OrderViewPageMap.cancelProduct.radios.voucherRefundType.productPricesVoucherExcluded),
      refundVoucherExcluded
    );
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+    if (refundVoucherExcluded < 0) {
+      $(OrderViewPageMap.cancelProduct.radios.voucherRefundType.productPricesVoucherExcluded).prop('checked', false).prop('disabled', true);
```
PierreRambaud:
```suggestion
      $(OrderViewPageMap.cancelProduct.radios.voucherRefundType.productPricesVoucherExcluded)
        .prop('checked', false)
        .prop('disabled', true);
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  updateVoucherRefundTypeLabel($input, refundAmount) {
+    const defaultLabel = $input.data('defaultLabel');
+    const $label = $input.parents('label');
+    const formattedAmount = this.currencyFormatter.format(refundAmount);
+
+    // Change the ending text part only to avoid removing the input
+    $label.get(0).lastChild.nodeValue = `
+    ${defaultLabel} ${formattedAmount}`;
```
PierreRambaud, line 151: Is line break wanted?

admin-dev/themes/new-theme/package-lock.json
```diff
@@ -4161,8 +4161,7 @@
                 "ansi-regex": {
                     "version": "2.1.1",
                     "bundled": true,
-                    "dev": true,
-                    "optional": true
+                    "dev": true
```
PierreRambaud: You should update your npm & node version. Please use nodejs 10.

admin-dev/themes/new-theme/scss/pages/_orders.scss
```diff
@@ -40,6 +40,14 @@
   }
 }
 
+#addProductTableRow,
+.editProductRow {
+  // div selector is important here, it is used for empty labels regular labels must not be hidden
+  div.form-control-label {
```
PierreRambaud, line 51: Ping @NeOMakinG

install-dev/upgrade/sql/1.7.7.0.sql
```diff
@@ -540,3 +540,22 @@ INSERT IGNORE INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `po
 ;
 
 INSERT INTO `PREFIX_hook_alias` (`name`, `alias`) VALUES ('displayAdminOrderTop', 'displayInvoice');
+
+/* Add refund amount on order detail, and fill new columns via data in order_slip_detail table */
+ALTER TABLE `PREFIX_order_detail` ADD `total_refunded_tax_excl` DECIMAL(20, 6) NOT NULL AFTER `original_wholesale_price`;
```
PierreRambaud: You forget the DEFAULT value :thinking:

src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
```diff
@@ -112,15 +115,38 @@ public function handle(IssuePartialRefundCommand $command)
...
         // Create order slip
-        $this->orderSlipCreator->createOrderSlip($order, $orderRefundDetail);
+        if ($command->generateCreditSlip()) {
+            $this->orderSlipCreator->create($order, $orderRefundSummary);
+        }
+
+        // Update order details (after credit slip to avoid updating refunded quantities while the credit slip fails)
+        foreach ($orderRefundSummary->getProductRefunds() as $orderDetailId => $productRefund) {
+            $orderDetail = $orderRefundSummary->getOrderDetailById($orderDetailId);
+            if ($order->hasBeenPaid()) {
```
PierreRambaud: Prevent useless indentation
```suggestion
            if (!$order->hasBeenPaid()) {
                continue;
            }
```

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -356,13 +356,12 @@ private function getOrderProducts(Order $order): OrderProductsForViewing
...
-            $product['quantity_refundable'] = $product['product_quantity'] - $resume['product_quantity'];
-            $product['amount_refundable'] = $product['total_price_tax_excl'] - $resume['amount_tax_excl'];
-            $product['amount_refundable_tax_incl'] = $product['total_price_tax_incl'] - $resume['amount_tax_incl'];
+            $product['quantity_refundable'] = $product['product_quantity'] - $product['product_quantity_return'] - $product['product_quantity_refunded'];
+            $product['amount_refundable'] = $product['total_price_tax_excl'] - $product['total_refunded_tax_excl'];
+            $product['amount_refundable_tax_incl'] = $product['total_price_tax_incl'] - $product['total_refunded_tax_incl'];
             $product['displayed_max_refundable'] = $order->getTaxCalculationMethod() ? $product['amount_refundable'] : $product['amount_refundable_tax_incl'];
-            $resumeAmount = $order->getTaxCalculationMethod() ? 'amount_tax_excl' : 'amount_tax_incl';
-            $product['amount_refunded'] = $resume[$resumeAmount] ?? 0;
+            $resumeAmount = $order->getTaxCalculationMethod() ? 'total_refunded_tax_excl' : 'total_refunded_tax_incl';
```
PierreRambaud: wdyt?
```suggestion
            $resumeAmountKey = $order->getTaxCalculationMethod() ? 'total_refunded_tax_excl' : 'total_refunded_tax_incl';
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: It's better because the listeners are added each time we click on the "Display partial refund button" and I don't want an input to have 5 duplicated listeners Besides this needs to be done because some rows may have been added (or removed) via javascript and ajax actions

Of course we could add the listener on the go each time a row is added, but it seems simpler to do it this way and less prone to error We already had a similar discussion about how it prevents some javascript overriding, which is partly true but none of our frontend code is designed for override right now anyway

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Same answer on the other comment

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -26,34 +26,42 @@
...
         ?float $chosenVoucherAmount
     ): OrderRefundSummary {
         $isTaxIncluded = $this->isTaxIncludedInOrder($order);
+        $precision = $this->getPrecision($order);
 
-        $orderDetailList = $this->getOrderTailList($orderDetailRefunds);
-        $productRefunds = $this->flattenProductRefunds($orderDetailRefunds, $isTaxIncluded, $orderDetailList);
+        $orderDetailList = $this->getOrderDetailList($orderDetailRefunds);
+        $taxCalculator = $this->getOrderTaxCalculator($order);
+        $productRefunds = $this->flattenCheckedProductRefunds($orderDetailRefunds, $isTaxIncluded, $orderDetailList, $taxCalculator, $precision);
```
PierreRambaud:
```suggestion
        $productRefunds = $this->flattenCheckedProductRefunds(
            $orderDetailRefunds, 
            $isTaxIncluded, 
            $orderDetailList, 
            $taxCalculator, 
            $precision
        );
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Yep it's on purpose, this is how the twig widget is displayed It avoids to text to be too close to the radio

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -114,44 +128,64 @@ private function getOrderTailList(array $orderDetailRefunds)
...
+     * @param TaxCalculator $taxCalculator
+     * @param int $precision
      *
      * @return array
      *
      * @throws PrestaShopDatabaseException
      * @throws PrestaShopException
      */
-    private function flattenProductRefunds(array $orderDetailRefunds, bool $isTaxIncluded, array $orderDetails)
+    private function flattenCheckedProductRefunds(array $orderDetailRefunds, bool $isTaxIncluded, array $orderDetails, TaxCalculator $taxCalculator, int $precision)
```
PierreRambaud:
```suggestion
    private function flattenCheckedProductRefunds(
        array $orderDetailRefunds, 
        bool $isTaxIncluded, 
        array $orderDetails, 
        TaxCalculator $taxCalculator, 
        int $precision
    ) {
```

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -114,44 +128,64 @@ private function getOrderTailList(array $orderDetailRefunds)
...
         $productRefunds = [];
         /** @var OrderDetailRefund $orderDetailRefund */
         foreach ($orderDetailRefunds as $orderDetailRefund) {
             $orderDetailId = $orderDetailRefund->getOrderDetailId();
+            /** @var OrderDetail $orderDetail */
             $orderDetail = $orderDetails[$orderDetailId];
             $quantity = $orderDetailRefund->getProductQuantity();
+            $quantityLeft = (int) $orderDetail->product_quantity - (int) $orderDetail->product_quantity_refunded - (int) $orderDetail->product_quantity_return;
+            if ($quantity > $quantityLeft) {
+                throw new InvalidRefundQuantityException(InvalidRefundQuantityException::QUANTITY_TOO_HIGH, $quantityLeft);
```
PierreRambaud:
```suggestion
                throw new InvalidRefundQuantityException(
                    InvalidRefundQuantityException::QUANTITY_TOO_HIGH, 
                    $quantityLeft
                );
```

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -172,18 +206,31 @@ private function isTaxIncludedInOrder(Order $order): bool
...
      * @return TaxCalculator
      *
-     * @throws PrestaShopDatabaseException
      * @throws PrestaShopException
      */
-    private function getTaxCalculator(float $taxRate)
+    private function getOrderTaxCalculator(Order $order): TaxCalculator
+    {
+        $carrier = new Carrier((int) $order->id_carrier);
+        // @todo: define if we use invoice or delivery address, or we use configuration PS_TAX_ADDRESS_TYPE
```
PierreRambaud, line 241: PIng @eternoendless @colinegin 
I think it's mandatory before merge.
AFAIK, it's the delivery address, no?

admin-dev/themes/new-theme/scss/pages/_orders.scss
jolelievre: Yes, thats because in the twig template, empty labels are replaced with an empty div with this class I think the purpose is to keep the element aligned in a regular form, even when some of the form inputs have no labels

src/Adapter/Order/Refund/OrderSlipCreator.php
```diff
@@ -138,13 +147,184 @@ public function createOrderSlip(
...
+     * @param bool $amount_choosen
+     * @param bool $add_tax
+     * @param int $precision
+     *
+     * @return bool
+     *
+     * @throws PrestaShopDatabaseException
+     * @throws PrestaShopException
+     */
+    private function createOrderSlip(Order $order, array $product_list, float $shipping_cost = 0, float $amount = 0, bool $amount_choosen = false, bool $add_tax = true, int $precision = 6)
```
PierreRambaud:
```suggestion
    private function createOrderSlip(
        Order $order, 
        array $product_list, 
        float $shipping_cost = 0, 
        float $amount = 0, 
        bool $amount_choosen = false, 
        bool $add_tax = true, 
        int $precision = 6
    ) {
```

src/Adapter/Order/Refund/OrderSlipCreator.php
```diff
@@ -138,13 +147,184 @@ public function createOrderSlip(
...
+    {
+        $currency = new Currency((int) $order->id_currency);
+        $order_slip = new OrderSlip();
```
PierreRambaud:
```suggestion
        $orderSlip = new OrderSlip();
```
You can use camelCase :)

src/Adapter/Order/Refund/OrderSlipCreator.php
```diff
@@ -138,13 +147,184 @@ public function createOrderSlip(
...
+            $add_or_remove = 'add';
+            $inc_or_ex_1 = 'excl';
+            $inc_or_ex_2 = 'incl';
+        } else {
+            $add_or_remove = 'remove';
+            $inc_or_ex_1 = 'incl';
+            $inc_or_ex_2 = 'excl';
+        }
+
+        $order_slip->{'total_shipping_tax_' . $inc_or_ex_1} = 0;
```
PierreRambaud:
```suggestion
        $order_slip->total_shipping_tax_excl = 0;
```

src/Adapter/Order/Refund/OrderSlipCreator.php
```diff
@@ -138,13 +147,184 @@ public function createOrderSlip(
...
+        $order_slip->{'total_shipping_tax_' . $inc_or_ex_2} = 0;
```
PierreRambaud:
```suggestion
        $order_slip->total_shipping_tax_incl = 0;
```

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: I believe cs-fixer changed this for me 😅 But I'll try again

src/Adapter/Order/Refund/OrderSlipCreator.php
PierreRambaud: Not needed, because both are set up to 0

src/Adapter/Order/Refund/OrderSlipCreator.php
```diff
@@ -138,13 +147,184 @@ public function createOrderSlip(
...
+                $order_slip->{'total_products_tax_' . $inc_or_ex_2} += $price;
+            }
+        }
+
+        $order_slip->{'total_products_tax_' . $inc_or_ex_2} -= $amount && !$amount_choosen ? $amount : 0;
+        $order_slip->amount = $amount_choosen ? $amount : $order_slip->{'total_products_tax_' . $inc_or_ex_1};
+        $order_slip->shipping_cost_amount = $order_slip->total_shipping_tax_incl;
+
+        if ((float) $amount && !$amount_choosen) {
+            $order_slip->order_slip_type = 1;
```
PierreRambaud: Could we use a constant here, what is 1? what is 2? what is love?
https://www.youtube.com/watch?v=HwVh8pmOot4

src/Adapter/Order/Refund/OrderSlipCreator.php
```diff
@@ -138,13 +147,184 @@ public function createOrderSlip(
...
+            return false;
+        }
+
+        $res = true;
+
+        foreach ($product_list as $product) {
+            $res &= $this->addProductOrderSlip((int) $order_slip->id, $product);
+        }
+
+        return $res;
```
PierreRambaud:
```suggestion
        return (bool) $res;
```
Must be cast in bool

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: I don't know if it's mandatory before merging this PR, but it definitely is before the freeze so that we clean up the miscellaneous part of the code with this behaviour It is a legal/functional issue
I think that's why there is a configuration for this, so that the merchant can choose Maybe the rules are different for professional customers for example (you wan the invoice to be taxed depending on your office not where you will ship the order)

Maybe it should be related to the customer group so that you can have different settings depending on your customers (not possible for now, but it is possible to define if tax is included or not, so why not add a "tax_address_type" field?)

src/Adapter/Order/Refund/OrderSlipCreator.php
jolelievre: Yes it was copy/pasted from legacy code but I will clean it a bit ;)

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
```diff
@@ -453,6 +478,17 @@ protected function getContextShopId()
         return $this->getContext()->shop->id;
     }
 
+    /**
+     * @param FormInterface $form
+     */
+    protected function addFlashFormErrors(FormInterface $form)
+    {
+        /** @var FormError $formError */
+        foreach ($form->getErrors() as $formError) {
```
PierreRambaud:
```suggestion
        foreach ($form->getErrors(true, false) as $formError) {
```
Maybe something more generic?  https://symfony.com/doc/current/components/form.html#accessing-form-errors

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -468,6 +472,8 @@ public function partialRefundAction(int $orderId, Request $request)
             $result = $formHandler->handleFor($orderId, $form);
             if ($result->isSubmitted() && $result->isValid()) {
                 $this->addFlash('success', $this->trans('A partial refund was successfully created.', 'Admin.Orderscustomers.Notification'));
+            } else {
+                $this->addFlashFormErrors($form);
```
PierreRambaud: Generate flash messages only when the form is submitted.

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -450,6 +451,45 @@ public function orderContainsRefundedProductWithReference(string $orderReference
...
+     * @param string $productName
+     */
+    public function checkProductDetailsWithReference(string $orderReference, string $productName, TableNode $table)
+    {
+        $productId = (int) $this->getProductIdByName($productName);
+        $order = new Order(SharedStorage::getStorage()->get($orderReference));
+        $orderDetails = $order->getProducts();
+        $productOrderDetail = null;
+        foreach ($orderDetails as $orderDetail) {
+            if ((int) $orderDetail['product_id'] == $productId) {
```
PierreRambaud:
```suggestion
            if ((int) $orderDetail['product_id'] === $productId) {
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done

src/Adapter/Order/Refund/OrderRefundCalculator.php
PierreRambaud: Weird, fuck php-cs-fixer sometimes :/

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
PierreRambaud: Could you add a little comment, look like it's a missclick :sweat_smile:

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
PierreRambaud: So instead of adding listener each time you click on the displya button. use the document syntax
```suggestion
    $(document).on('change', OrderViewPageMap.cancelProduct.inputs.quantity, (event) => {
```

PierreRambaud: It will be faster than creating an unlimited amount of js events

admin-dev/themes/new-theme/scss/pages/_orders.scss
PierreRambaud: Weird to add an empty div :(

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Mmm I like that I'm gonna try!

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: It works!

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done with `$(document).on` as well

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-cancel.js
jolelievre: Yep, done

admin-dev/themes/new-theme/package-lock.json
jolelievre: Done as well ^^

admin-dev/themes/new-theme/scss/pages/_orders.scss
jolelievre: I guess if you had only the input element on the line it would be aligned on the left, the empty div is used to fill the space on the left (like a label would if present) so that all inputs and labels stay aligned Just a guess but that's why I added a comment so it's not removed
And I guess this will require some special comment to avoid eslint to remove it @NeOMakinG ?

install-dev/upgrade/sql/1.7.7.0.sql
jolelievre: Thanks, done

src/Adapter/Order/CommandHandler/IssuePartialRefundHandler.php
jolelievre: Actually moved it in a method so it's cleaner

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
jolelievre: Yep! Done

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: Done

src/Adapter/Order/Refund/OrderSlipCreator.php
jolelievre: Done

src/Adapter/Order/Refund/OrderSlipCreator.php
jolelievre: Done

src/Adapter/Order/Refund/OrderSlipCreator.php
jolelievre: Yep, nice catch Done

src/Adapter/Order/Refund/OrderSlipCreator.php
jolelievre: Done

src/Adapter/Order/Refund/OrderSlipCreator.php
```diff
@@ -138,13 +147,184 @@ public function createOrderSlip(
...
+                $order_slip->{'total_shipping_tax_' . $inc_or_ex_2} = Tools::ps_round($tax_calculator->{$add_or_remove . 'Taxes'}($order_slip->{'total_shipping_tax_' . $inc_or_ex_1}), $precision);
+            } else {
+                $order_slip->{'total_shipping_tax_' . $inc_or_ex_2} = $order_slip->{'total_shipping_tax_' . $inc_or_ex_1};
+            }
+        } else {
+            $order_slip->shipping_cost = false;
+        }
+
+        $order_slip->amount = 0;
+        $order_slip->{'total_products_tax_' . $inc_or_ex_1} = 0;
```
jolelievre: @PierreRambaud  Same here!

src/Adapter/Order/Refund/OrderSlipCreator.php
jolelievre: That's a very good question!! 😅
A new mystery from the legacy code, I can only try to guess..

jolelievre: Ok found it, it matches with the voucher refund type which was already created

src/Adapter/Order/Refund/OrderSlipCreator.php
jolelievre: Mmm, it's already adding boolean only anyway But if we want to be sure I prefer to add it explicitly on the `addProductOrderSlip` method

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
jolelievre: I think `$form->getErrors(true)` is enough, so we have a one depth array with all the errors

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: Nice catch Done

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
jolelievre: Done

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: Done

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: Yep I confirm cs-fix changes it back to one line, I won't search for hours why for now ^^

jolelievre (commented):
@PierreRambaud I think I addressed all your feedback

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
PierreRambaud: Good to me if we use at least the first parameter

src/Adapter/Order/Refund/OrderRefundCalculator.php
PierreRambaud: FUCK cs fixer again

src/Adapter/Order/Refund/OrderSlipCreator.php
PierreRambaud: Because of bitwise, it's not a boolean, it's an integer :) 
https://www.php.net/manual/en/language.operators.bitwise.php

If you want to keep the integer, you need to change the php doc block

PierreRambaud (dismissed):
LGTM, more review please :pray:

PierreRambaud (dismissed):

matthieu-rolland (dismissed):

src/PrestaShopBundle/Form/Admin/Sell/Order/CancelProductType.php
```diff
@@ -94,8 +121,37 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                     ],
                 ]
             )
+            ->add('voucher_refund_type', ChoiceType::class,
+                [
+                    'required' => true,
+                    'multiple' => false,
+                    'expanded' => true,
+                    'choices' => [
+                        $this->trans('Product(s) price: ', 'Admin.Orderscustomers.Feature') => VoucherRefundType::PRODUCT_PRICES_REFUND,
```
LouiseBonnard, line 130:
```suggestion
                        $this->trans('Product(s) price:', 'Admin.Orderscustomers.Feature') => VoucherRefundType::PRODUCT_PRICES_REFUND,
```

src/PrestaShopBundle/Form/Admin/Sell/Order/CancelProductType.php
```diff
@@ -94,8 +121,37 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+                        $this->trans('Product(s) price, excluding amount of initial voucher: ', 'Admin.Orderscustomers.Feature') => VoucherRefundType::PRODUCT_PRICES_EXCLUDING_VOUCHER_REFUND,
```
LouiseBonnard, line 131: _.Feature_ or _.Help_, is it a tooltip?
🤔

Robin-Fischer-PS:
Hi there !

Some feedbacks on this PR : 

- When using demo order (ID 1), the "Max Amount tax included" is different from the "Total tax included" of the product. This also happens on Legacy (177x and 176x), I can't find out why, and can't reproduce with "real" orders.
pr17211 max different
- Some graphical bugs : The "0 (€0.00 Refunded)" should be aligned with others info (Available, Quantity...), Quantity in Partial Refund too, and Amount (Tax included) is a bit "messy". In the "Shipping" info below, when the partial refund block is displayed, the Shipping amount is not aligned. The "Partial refund" button is not aligned with other buttons.
pr17211 graphical bugs
pr17211 partial refund button

- There is a "Generate a credit slip" choice that should not be there (according to Legacy, Spec and mockup). If neither "credit slip" or "voucher" are enabled when doing a partial refund, the refund is just saved in DB but is nowhere to be found. If we want to keep the "credit slip" choice, there should be a check in order to be sure that at least either voucher or credit slip are checked.
pr17211 generate credit slip
- When not filling the "Quantity" field, the error message should be "Please enter a quantity to proceed with your refund." instead of "Please enter a positive amount to proceed with your refund."
- When doing a refund with an amount superior to the tax excluded price of the product, the refunded price will be capped to the tax excluded price. It should not.
https://drive.google.com/file/d/1-9gMhPg4KhY8V4JKC4irve7F9XlT1-RA/view?usp=sharing
- When an order is partially or totally paid with a voucher and you haven't filled the form yet, there is an error that might confuse the merchant (this error was less "scary" in Legacy and did not blocked the process) : 
pr17211 paid with voucher
- As discussed together, there is a missing choice when an order is paid with voucher : "Amount of your choice"

jolelievre:
@Robin-Fischer-PS all the bugs have been fixed except the one about `"Amount of your choice"`
I thought I understood it fully  but it turns out the feature really is not clear and requires more detailed specifications, so it will be dealt with in another issue

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1224,21 +1223,29 @@ private function getErrorMessages(Exception $e)
...
                     'Please enter a positive quantity to proceed with your refund.',
                     'Admin.Orderscustomers.Notification'
                 ),
-                InvalidRefundQuantityException::QUANTITY_TOO_HIGH => $this->trans(
+                InvalidRefundException::QUANTITY_TOO_HIGH => $this->trans(
                     'Please enter a maximum quantity of [1] to proceed with your refund.',
                     'Admin.Orderscustomers.Notification',
                     ['[1]' => $refundableQuantity]
                 ),
+                InvalidRefundException::INVALID_AMOUNT => $this->trans(
```
jolelievre, line 1236: Hello @LouiseBonnard I added these new wordings, let me know if you think they need to be modified Thank you

install-dev/fixtures/fashion/data/order_detail.xml
```diff
@@ -53,7 +53,7 @@
                       product_ean13="" product_upc="" product_reference="demo_2" product_supplier_reference=""
                       product_weight="0.000000" tax_computation_method="0" tax_name="" tax_rate="0.000"
                       ecotax="0.000000" ecotax_tax_rate="0.000" discount_quantity_applied="0" download_nb="0"
-                      download_deadline="0000-00-00 00:00:00" total_price_tax_incl="27.000000"
+                      download_deadline="0000-00-00 00:00:00" total_price_tax_incl="23.900000"
                       total_price_tax_excl="23.900000" unit_price_tax_incl="23.900000" unit_price_tax_excl="23.900000"
```
matthieu-rolland, line 57: Why do you need to change the tax included price on a fixture? The bug came from the fixture itself?

src/Core/Domain/Order/Exception/InvalidRefundException.php
```diff
@@ -31,7 +31,7 @@
 /**
  * Throw new when a partial refund's is asked without a specified quantity
  */
-class InvalidRefundQuantityException extends OrderException
+class InvalidRefundException extends OrderException
```
matthieu-rolland, line 34: Do you consider that having a different exception for quantity and amount was overkill? (or is there another reason for this slight refactor ?)

install-dev/fixtures/fashion/data/order_detail.xml
jolelievre: Yes, this was the only field where tax included was different from tax excluded Which is why we didn't understand why this bug only happened on this specific order We could ignore it but I think it's better to fix it
Although the real best fix would be to use real tax included values but I think this will be done some other time

>

src/Core/Domain/Order/Exception/InvalidRefundException.php
jolelievre: Well, fo two scenarii I didn't mind But then I added more specific use case codes, so having 5 different exceptions related to a refund validation seemed overkill indeed So I preferred to mix them all

src/PrestaShopBundle/Form/Admin/Sell/Order/CancelProductType.php
jolelievre: No it's a radio button selection
Capture d’écran 2020-01-24 à 11 24 35

I also need the extra space to concatenate the price after, but I can remove it from the wording and add it programmatically It's just that this wording was already used in legacy

src/Core/Domain/Order/Exception/InvalidRefundException.php
matthieu-rolland: alright it makes sense 👍

matthieu-rolland (dismissed):
LGTM, but we probably need to wait Louise's feedback before sending this PR to the QA.

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1203,14 +1223,29 @@ private function getErrorMessages(Exception $e)
...
+                    'Please enter a positive amount to proceed with your refund.',
+                    'Admin.Orderscustomers.Notification'
+                ),
+                InvalidRefundException::NO_REFUNDS => $this->trans(
+                    'Please enter at least one refund.',
+                    'Admin.Orderscustomers.Notification'
+                ),
+                InvalidRefundException::NO_GENERATION => $this->trans(
+                    'Please generate at least a credit slip or a voucher.',
```
```suggestion
                    'Please generate at least one credit slip or voucher.',
```

LouiseBonnard: It seems more consistent this way.

src/PrestaShopBundle/Form/Admin/Sell/Order/CancelProductType.php
LouiseBonnard: Good for me, thanks Jo!

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: Done

matthieu-rolland (dismissed):
LGTM

admin-dev/themes/new-theme/scss/pages/_orders.scss
NeOMakinG: Could you wrap these inside stylelint comments like this ?

```
/* stylelint-disable */
  div.form-control-label {
   ...
  }
  /* stylelint-enable */
```

admin-dev/themes/new-theme/scss/pages/_orders.scss
```diff
@@ -166,33 +174,30 @@ td .editProductInvoice {
...
   width: 100%;
   margin-right: 20px;
   margin-top: -10px;
 }
 
 .refund-checkboxes-container {
-  max-width: 90px;
+  width: fit-content;
   margin: 16px auto;
   display: none;
```
NeOMakinG: display should be first there before the width

NeOMakinG (changes requested):

jolelievre:
@NeOMakinG I made the changes you suggested, can you please re-approve the PR (only lint changes, no impact for QA)

NeOMakinG (dismissed):

matks (dismissed):

PierreRambaud (dismissed):

matthieu-rolland (dismissed):

matks (dismissed):

matks (approved):

PierreRambaud (approved):

PierreRambaud:
Thanks @jolelievre

PR #16722, opened 2019-12-06
Orders > Products : Display content of a pack

Progi1984:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | "Orders > Products" : Display content of a pack
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16393
| How to test?  | Check beahavior in BO > Orders > Product when you have a pack (design in  #16393)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16722)

PR #17220, opened 2020-01-17
Fix multiple bugs in order view page

### LINKED ISSUE #17089: Migration - Product block on order page bugs

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | last product of pagination were shown on both pages. Fixed
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes all bugs of #17089
| How to test?  | https://github.com/PrestaShop/PrestaShop/issues/17089

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17220)


matks (dismissed):

matks:
All bugs fixed in https://github.com/PrestaShop/PrestaShop/pull/17206

atomiix (dismissed):

Robin-Fischer-PS:
Hi @zuk3975 ! Could you fix the conflict on the "UpdateProductInOrderHandler.php" file ? Thanks !

matks:
Needs rebase

zuk3975:
@Robin-Fischer-PS rebased.
Conflicts were related to this pr https://github.com/PrestaShop/PrestaShop/pull/16722 so it would be nice to check if I didn't break anything. (I've checked standard case, seems to be ok)

admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
```diff
@@ -127,7 +134,6 @@ export default class OrderProductEdit {
 
     // Init editor data
     this.taxRate = product.tax_rate;
-    this.totalQuantity = product.availableQuantity + product.quantity;
```
matks, line 130: @zuk3975 Expected ? No rebase mistake ?

matks (approved):

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -240,5 +246,14 @@ private function assertProductCanBeUpdated(
 //                }
 //            }
 //        }
+
+        //check if product is available in stock
+        if (!\Product::isAvailableWhenOutOfStock(StockAvailable::outOfStock($orderDetail->product_id))) {
```
Progi1984, line 251:
```suggestion
        if (!Product::isAvailableWhenOutOfStock(StockAvailable::outOfStock($orderDetail->product_id))) {
```

Progi1984 (commented):
A feedback and the same question as @matks

NeOMakinG (dismissed):
Except the @matks question, front-end part LGTM

admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
zuk3975: It was intentional. Check lines (-47) and 49. As I remember it had something to do with dynamic stock value checking instead of product.availableQuantity (or something like that, i don't remember)

matks (dismissed):

matks:
Confirmed by @zuk3975 👍

Robin-Fischer-PS:
I tested this PR, no major bug appeared, so once the conflict is solved, merge will be possible : 
Even if we might still have some bugs out there, I will create one issue per bug, in order to have a clearer backlog, and a better view on what's left to do in the version.

Thanks !

matks (approved):

jolelievre (approved):

PR #17241, opened 2020-01-20
Do not allow negative amount in order payment block

### LINKED ISSUE #16582: It is possible to add negative Amount for Order Payment from the Back Office
**Describe the bug**
It is possible to add negative Amount for Order Payment from the Back Office. Then message is thrown: An unexpected error occurred. [PrestaShopException code 0]. The actual exception is: Property Order->total_paid_real is not valid.

According to as I understand negative payments should not be saved to the database or at least total sum of payments from the db should not be negative.

**To Reproduce**
1. Go to Front Office and create new order.
2. Go to Back Office and go into that order page
3. Scroll down to Payment section and add new payment with negative amount: -5.548
4. See error: An unexpected error occurred. [PrestaShopException code 0]

**Screenshots**
Screenshot from 2019-11-28 12-51-45
Screenshot from 2019-11-28 12-52-21

**Additional information**
Have Behat integration 2 failing tests written in file /Order/add_payment.feature on the Github Pull request: https://github.com/PrestaShop/PrestaShop/pull/16525

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Do not allow negative amount in order payment
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16582
| How to test?  | In the migrated order/view page, try to add a payment with a negative amount, an error should be displayed and the payment should not be added

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17241)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * 2007-2019 PrestaShop SA and Contributors
+ * 2007-2020 PrestaShop SA and Contributors
  *
  * NOTICE OF LICENSE
  *
@@ -19,7 +19,7 @@
  * needs please refer to https://www.prestashop.com for more information.
  *
  * @author    PrestaShop SA <contact@prestashop.com>
- * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @copyright 2007-2020 PrestaShop SA and Contributors
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
@@ -819,6 +819,10 @@ class OrderController extends FrameworkBundleAdminController
                 $this->addFlash('success', $this->trans('Successful update.', 'Admin.Notifications.Success'));
             } catch (Exception $e) {
                 $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));
+            }
+        } elseif (!$form->isValid()) {
+            foreach ($form->getErrors(true) as $error) {
+                $this->addFlash('error', $error->getMessage());
             }
         }
 
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
@@ -1260,7 +1260,7 @@ class OrderController extends FrameworkBundleAdminController
                 'Admin.Notifications.Error'
             ),
             NegativePaymentAmountException::class => $this->trans(
-                'The amount should be greater than 0.',
+                'Invalid value: the payment must be a positive amount.',
                 'Admin.Notifications.Error'
             ),
         ];
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
@@ -46,6 +46,7 @@ use PrestaShop\PrestaShop\Core\Domain\Order\Command\UpdateOrderStatusCommand;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\CannotEditDeliveredOrderProductException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\ChangeOrderStatusException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\InvalidRefundException;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\NegativePaymentAmountException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderEmailSendException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderNotFoundException;
@@ -1256,6 +1257,10 @@ class OrderController extends FrameworkBundleAdminController
             ],
             ProductOutOfStockException::class => $this->trans(
                 'There are not enough products in stock',
+                'Admin.Notifications.Error'
+            ),
+            NegativePaymentAmountException::class => $this->trans(
+                'The amount should be greater than 0.',
                 'Admin.Notifications.Error'
             ),
         ];
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
@@ -800,29 +800,31 @@ class OrderController extends FrameworkBundleAdminController
         ]);
         $form->handleRequest($request);
 
-        if ($form->isSubmitted() && $form->isValid()) {
-            $data = $form->getData();
+        if ($form->isSubmitted()) {
+            if ($form->isValid()) {
+                $data = $form->getData();
 
-            try {
-                $this->getCommandBus()->handle(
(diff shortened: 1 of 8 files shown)


src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -803,6 +803,10 @@ public function addPaymentAction(int $orderId, Request $request): RedirectRespon
             } catch (Exception $e) {
                 $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));
             }
+        } elseif (!$form->isValid()) {
```
PierreRambaud:
```php
       if ($form->isSubmitted()) {
            if (!$form->isValid()) {
                foreach ($form->getErrors(true) as $error) {
                    $this->addFlash('error', $error->getMessage());
                }
            } else {
                $data = $form->getData();

                try {
                        new AddPaymentCommand(
                            $orderId,
                            $data['date'],
                            $data['payment_method'],
                            $data['amount'],
                            $data['id_currency'],
                            $data['id_invoice'],
                            $data['transaction_id']
                        )
                    );

                    $this->addFlash('success', $this->trans('Successful update.', 'Admin.Notifications.Success'));
                } catch (Exception $e) {
                    $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));
                }
            }
        }

**Comment by matks** [comment]
Can you turn on the test we disabled until this bug is fixed ? 😄this will help validating the feature

https://github.com/tdavidsonas88/PrestaShop/commit/ea4251123a6c43eaf53b9a1fb13006751b434765

**Review by jolelievre — APPROVED** [review]
Thanks @atomiix

**Review by PierreRambaud — DISMISSED** [review]

**Review by jolelievre — CHANGES_REQUESTED** [review]
Sorry I missed the behat test part ^^ Please add:
- a check for negative amount in the handler (or the command via ValueObject maybe easier) which throws an exception
- reactivate the behat test and add a line to check that the appropriate exception has been thrown

File: src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php [hunk 2]
```diff
@@ -1215,6 +1222,10 @@ private function getErrorMessages(Exception $e)
                 'There are not enough products in stock',
                 'Admin.Notifications.Error'
             ),
+            NegativePaymentAmountException::class => $this->trans(
+                'The amount should be greater than 0.',
```
```suggestion
                'The amount of a payment should be strictly positive.',
```

Suggestion @LouiseBonnard what do you think?

tests/Integration/Behaviour/Features/Context/Domain/OrderPaymentFeatureContext.php
```diff
@@ -129,15 +130,19 @@ public function addPaymentToOrderWithTheInvalidFollowingProperties(string $order
...
                 )
             );
-        } catch (PrestaShopException $exception) {
-            $msg = $exception->getMessage();
-            $expectedMsg = 'Property Order->total_paid_real is not valid';
-            if ($msg !== 'Property Order->total_paid_real is not valid') {
-                throw new RuntimeException(sprintf('Not expected exception is thrown "%s" but "%s" was expected', $msg, $expectedMsg));
-            }
+        } catch (NegativePaymentAmountException $exception) {
+            $this->lastException = $exception;
```
jolelievre, line 134: Be careful you removed an expected behaviour (with Property Order->total_paid_real is not valid), although the way it was managed was weird
You need to find the related behat test and fix/refacto it By catching the `PrestaShopException` exception and maybe only store it when the message matches so that other use cases correctly fail

matthieu-rolland (dismissed):
@jolelievre 's feedback has been dealt with properly, LGTM for me 👍

PierreRambaud (dismissed):

jolelievre (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
LouiseBonnard: It sounds a bit authoritarian, aha! Why changing the wording, isn't it clear enough?

If you want to emphasize the mandatory side of it, I would write _Invalid value: the payment must be a positive amount._, what do you think?

Robin-Fischer-PS:
Hi @atomiix ! Could you resolve the conflict on this PR ? Thanks !

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: It's not the mandatory side that I wanted to emphasize, but more having a less "technical" message I just though "positive amount" is maybe more understandable than "greater to 0"

LouiseBonnard: I agree, thanks @jolelievre. I still suggest _Invalid value: the payment must be a positive amount._ in this case, it looks clearer.

jolelievre (approved):
Nice @atomiix

PierreRambaud (approved):

Progi1984:
Thanks @atomiix

djbuch:
This is completely wrong... The majority of e-commercant we are helping are typing negative amounts in case of reimbursement... This should be rolled back as soon as possible.

matks:
Hi @djbuch there are 4 possible ways to refund an order:
- cancel a product
- standard refund
- partial refund
- return products

You can see the details in https://github.com/PrestaShop/prestashop-specs/blob/master/content/1.7/back-office/orders/orders/Order%20Page%20View.md#merchandise-return-has-to-be-enabled-if-the-merchant-want-to-use-the-standard-refund-partial-refund-and-return-product-feature

Why not use them rather than typing a negative amount in a "payment" field?

djbuch:
Because those use cases only create the OrderSlip if the Invoice has already been created or cancel a product, but it won't change the payment.

But imagine a workflow were the invoice is created when the Order is already sent out.
Then the seller sees he has not enough of one product, so he cancels the product or lowers the quantity, the order total is so decreased, and he will have an error in the payment section, saying that there was too much paid.

So when he really makes the reimbursement to the buyer with his bank system, he could note this in here with a negative amount to keep track of financial mouvement for this order.

matks:
@djbuch So if I understand correctly, here is a valid scenario

1. I buy 4 boxes of Sugar, each costs 15€ => my order is 60€. I pay 60€. No shipping fees. No VAT, it's B2B.
2. I receive my invoice (it mentions 60€).
3. Merchant finds out OH NO! he has only 3 boxes
4. Merchant wants to modify the order, he modifies the quantity to 3. Order total decreases to 45€.
5. But I already paid 60€ ! So this order is in "overdue" status.
6. What you want to do is "Merchant sends a 15€ check to me, and marks the order as 'paid' (not overdue)" in the Back Office because now the amount I have paid and the amount of my order are equal

Is that a good scenario ?

@MatShir @marionf what do you think about this usecase?

@djbuch I still think that adding negative amount payments is not a good solution 😅  but maybe there is a scenario that is not handled by PrestaShop, indeed. Let's see what Product Team can say.

djbuch:
@matks Yes this is a valide use case.

Maybe the negative could be checked agains total payments no becoming negative. But a reimbursement is indeed a negative payment.

matks:
AFAIK Product team is actually in the process of getting some law compliance people so that they can tell us what can be done or what cannot be done in order to comply with the law.

PrestaEdit:
The main issue there is not that you disallow to add the negative price on the Back Office, but that the model have been change (it's a BC break, as far I know) on a minor version (1.7.7) and you are now not allowed to make a refund with a specific transaction ID and set the negative amount using the ObjectModel related, so.

PR #17288, opened 2020-01-21
Upgrade Symfony 3.4

### LINKED ISSUE #16736: Installer is consistently failing with NTS PHP on Windows
**Describe the bug**
When PHP uses only 1 thread, i.e. CGI or CLI, it should produce exactly the same output with TS AND NTS build, right?

But after days of trouble shooting I finally find the issue with PrestaShop installer. The issue is NTS PHP on Windows.

The PrestaShop install always fail with NTS PHP (with error Template "C:\wamp64\www\ps\mails\themes\modern\core\account.html.twig" is not defined.), but always succeed when run with TS PHP.
image

**To Reproduce**

0. download NTS PHP for Windows, for ex. download https://windows.php.net/downloads/releases/php-7.2.25-nts-Win32-VC15-x64.zip
1. download the latest version of PrestaShop https://github.com/PrestaShop/PrestaShop/releases/download/1.7.6.2/prestashop_1.7.6.2.zip and, unzip it
2. a) install it from CLI using admin/index_cli.php
    b) OR install thru web using PHP CGI

**Additional information**
PrestaShop version: 1.7.6.2 rel or git develop branch
PHP version: 7.2.25, 7.3.9, 7.3.12
OS: Windows 10 x64

I tested it on multiple Windows 10 systems/PCs and with multiple versions of PHP. The install always completes with the TS version of PHP, but always fail with the NTS one.

Can you please reproduce and fix?

mvorisek:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fix Installer is consistently failing with NTS PHP on Windows
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16736
| How to test?  | All tests needs to pass

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17288)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "000c20400b663f7d57cea6c420b7e4cc",
+    "content-hash": "956f5b7a0863ccdbec31308601948de9",
     "packages": [
         {
             "name": "beberlei/doctrineextensions",
@@ -7502,16 +7502,16 @@
         },
         {
             "name": "symfony/symfony",
-            "version": "v3.4.36",
+            "version": "v3.4.37",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/symfony.git",
-                "reference": "0a6fccb5772ad2467253e6849d589bd09d9eb195"
+                "reference": "1bd873459b36cf505c7b515ba6e0e2ee40890b8a"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/symfony/symfony/zipball/0a6fccb5772ad2467253e6849d589bd09d9eb195",
-                "reference": "0a6fccb5772ad2467253e6849d589bd09d9eb195",
+                "url": "https://api.github.com/repos/symfony/symfony/zipball/1bd873459b36cf505c7b515ba6e0e2ee40890b8a",
+                "reference": "1bd873459b36cf505c7b515ba6e0e2ee40890b8a",
                 "shasum": ""
             },
             "require": {
@@ -7654,7 +7654,7 @@
             "keywords": [
                 "framework"
             ],
-            "time": "2019-12-01T13:50:53+00:00"
+            "time": "2020-01-21T12:30:09+00:00"
         },
         {
             "name": "tecnickcom/tcpdf",
--- a/composer.json
+++ b/composer.json
@@ -115,7 +115,7 @@
         "symfony/polyfill-mbstring": "^1.12",
         "symfony/polyfill-php73": "^1.10",
         "symfony/swiftmailer-bundle": "^3.1",
-        "symfony/symfony": ">=3.4.26",
+        "symfony/symfony": ">=3.4.37",
         "tecnickcom/tcpdf": "^6.2.12",
         "tijsverkoyen/css-to-inline-styles": "^2.2",
         "twig/twig": "^1.38"


mvorisek:
Does anyone know why the PHP 5.6 unit test is failing and how to fix it?

matks:
> Does anyone know why the PHP 5.6 unit test is failing and how to fix it?

Looks like a false positive due to Travis env.
We have observed multiple failures on PHP 5.6 of our unit tests, including segfault. Travis team told us extensive phpunit dataproviders usage can create this kind of issue (because apparently it consumes a lot of memory).

On 1.7.7, since we have dropped the PHP5.6 compatibility, the issue is gone 😅 

Also, thanks for being so fast, I saw the Symfony release yesterday evening and here you are ! 😄

composer.lock
```diff
@@ -6557,31 +6558,31 @@
         },
         {
             "name": "twig/twig",
-            "version": "v1.38.4",
+            "version": "v1.42.4",
```
matks: Seems like a good thing but we'll need to perform a full check

tests-legacy/Integration/PrestaShopBundle/Controller/Api/StockManagementControllerTest.php
```diff
@@ -214,7 +214,7 @@ private function assertResponseHasTotalPages($parameters, $expectedTotalPages)
         $this->assertTrue($headers->has('Total-Pages'), 'The response headers should contain the total pages.');
         $this->assertSame(
             $expectedTotalPages,
-            $headers->get('Total-Pages'),
+            (int) $headers->get('Total-Pages'),
```
matks: This is necessary for tests to pass ?

mvorisek: Yes, otherwise there is datatype mismatch.

composer.lock
mvorisek: Yes, change of twig version was needed to upgrade the symfony version, but the test failure origin is not much clearer...

PierreRambaud: because of this: https://github.com/PrestaShop/PrestaShop/pull/17288/files#diff-1da2c7edc898c70e5a79a9997c98ceccR6324

mvorisek: @PierreRambaud Of course, but how to solve it? Or change the target from 1.7.6.x to 1.7.7.x which should be released soon?

matks: Does not look like an issue to me 😄 

I checked twig changelog, nothing suspicious in this update, I think we can do it

matks:
2nd failure of Travis for php5.6
I restart it again 🤔 but maybe we're observing an exotic issue for Symfony v3.4.37 and php5.6

mvorisek:
Updated to apply on `1.7.7.x` branch instead `1.7.6.x`.

PierreRambaud (approved):

Progi1984 (approved):

PierreRambaud (approved):

khouloudbelguith:
Hi @mvorisek,

It is OK.
1. Using `windows 10 & PHP nts`
[phpinfo()_pr.pdf](https://github.com/PrestaShop/PrestaShop/files/4123387/phpinfo._pr.pdf)
Install PS with CLI => OK / Front Office => OK / Back Office OK
image

Install PS using steps => OK
2. Using `Linux #1 SMP Debian 3.16.39-1`
[phpinfo().pdf](https://github.com/PrestaShop/PrestaShop/files/4123401/phpinfo.pdf)
Install PS with steps => OK / Front Office => OK / Back Office => OK

Thanks!

matks:
Thank you @mvorisek

Issue #17282, opened 2020-01-21
Finish @todo in 1.7.7 migrated code


matks:
We put some `@todo` in 1.7.7 migrated code to be handled later.

Later is now 💪 !

For each `@todo`, we must
- [remove it if irrelevant
or
- fix it
or
- answer it
or
- [create an issue about it

### TODO List

**In BO Theme JS:**
- [x] js/pages/order/create/customer-manager.js
- [x] js/pages/order/message/order-view-page-messages-handler.js

**In Adapter namespace:**
* [x] src/Core/Domain/Address/Command/SetRequiredFieldsForAddressCommand - #17526
* [x]  src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php - #17348
* [ ]  src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
* [x]  src/Adapter/Order/CommandHandler/ChangeOrderCurrencyHandler.php - #17672
* [ ]  src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
* [ ]  src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
* [ ]  src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
* [x]  src/Adapter/Order/Refund/OrderRefundCalculator.php - #17850
* [x]  src/Adapter/Order/Refund/OrderSlipCreator.php - `associated with mailer. As we don't have a new mailer yet, these todo's should be left for future`
* [x]  src/Adapter/Order/Refund/VoucherGenerator.php - `associated with mailer. As we don't have a new mailer yet, these todo's should be left for future`
* [x]  src/Adapter/Product/QueryHandler/SearchProductsHandler.php - #17736 

**In Core namespace:** 
* [ ]  src/Core/Grid/Data/Factory/CatalogPriceRuleGridDataFactory.php
* [ ]  src/Core/Grid/Definition/Factory/Monitoring/EmptyCategoryGridDefinitionFactory.php

**In PrestaShopBundle:**

* [ ]  src/Core/Grid/Data/Factory/CatalogPriceRuleGridDataFactory.php
* [ ]  src/Core/Grid/Definition/Factory/Monitoring/EmptyCategoryGridDefinitionFactory.php
* [ ]  src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
[...]

PR #17293, opened 2020-01-22
[POC] Fix address required field and country address format relation

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fix todo - Add required fields validation in SetRequiredFieldsForAddressCommand. TODO - still have to consider validation fields added by modules. refer to https://github.com/PrestaShop/PrestaShop/pull/15229#discussion_r329488344
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #17282 and finishes https://github.com/PrestaShop/PrestaShop/issues/10558
| How to test?  | Not testable with ui.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17293)


Progi1984 (dismissed):

matks (dismissed):

PierreRambaud (dismissed):

matks:
As discussed together:
- it seems legacy page was not considering at all extra fields provided through the hook `additionalCustomerFormFields`
- so we will not handle it as well as this topic would require a full refacto o the system both on FO and BO
- instead, what we will add is a notice warning, see below

### Notice warning

If hook `additionalCustomerFormFields` returns something not empty (= one module uses it), we display in Customer address add/edit form a warning: "Warning, extra form fields are added to customer address by modules through the hook `additionalCustomerFormFields`"

zuk3975:
> As discussed together:
> 
> * it seems legacy page was not considering at all extra fields provided through the hook `additionalCustomerFormFields`
> * so we will not handle it as well as this topic would require a full refacto o the system both on FO and BO
> * instead, what we will add is a notice warning, see below
> 
> ### Notice warning
> If hook `additionalCustomerFormFields` returns something not empty (= one module uses it), we display in Customer address add/edit form a warning: "Warning, extra form fields are added to customer address by modules through the hook `additionalCustomerFormFields`"

hmm, but that hook is called in FO and we want a notice in BO.

matks:
> hmm, but that hook is called in FO and we want a notice in BO.

Why is this an issue ? 😄

colinegin:
@zuk3975 @matks is there still PM info missing to finish this PR ?

MatShir:
So since the address format in the country doesn't match the form but it does in the FO (except for phone mobile), we can keep as Legacy. 
@matks and I suggest to take out the required fields in the BO (except for the real mandatory information). They will affect only the FO, as it works fine. So the merchant won't be surprised by having required fields to fulfill when the country doesn't require it. 

@zuk3975 what do you think?

zuk3975:
As discussed this feature is more complicated than expected, so it should be done in next release. Closing to leave the code for future when AddressFormat in Country settings will be migrated. 
Moved fixing mandatory bugs for 1..7.7.x to https://github.com/PrestaShop/PrestaShop/pull/17526

PR #17302, opened 2020-01-22
Fix duplicated orders for the same cart

### LINKED ISSUE #17266: Duplicated orders from the same cart
We have found a bug which allows to create duplicated orders from the same cart. In order to check if the shop has this kind of problem, you can execute the following MySQL query:

SELECT 
	id_order,
	GROUP_CONCAT(id_order), 
	count(id_cart), 
	GROUP_CONCAT(reference),
	date_add
FROM ps_orders
GROUP BY id_cart
HAVING COUNT(id_cart) > 1
ORDER BY `ps_orders`.`id_order` DESC

In the following screenshoot you can see that two identical orders have been created at the same time:
http://soy.es/screenshots/issues_17266_01.jpg

If we look at the carts, we see that two different orders were created from the same cart:
http://soy.es/screenshots/issues_17266_02.jpg

When we revise the code, in order to find the error, we see that in the Cart.php class there is a function which checks if there is already an order for that cart. This function, orderExists(), has a cache:
http://soy.es/screenshots/issues_17266_03.jpg

It is quite likely that the bug comes from this cache. We belive that each time the shop checks if an order exists from a cart may be verified with the data base directly, not with a cache:
http://soy.es/screenshots/issues_17266_04.jpg

This function is called before creating an order in the validateOrder() function from the PaymentModule class:
http://soy.es/screenshots/issues_17266_05.jpg

soyes:
Fix function for not using the cache and avoid the issue with duplicated orders.

| Questions     | Answers
| Branch?       | develop 
| Description?  | The function orderExists() from Cart.php has been modified in order to use only the cache when the order exists, avoiding the issue with duplicated orders
| Type?         | bug fix 
| Category?     | FO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17266 
| How to test?  | See #17266 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17302)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1628,15 +1628,9 @@ class CartCore extends ObjectModel
      */
     public function orderExists()
     {
-        $cache_id = 'Cart::orderExists_' . (int) $this->id;
-        if (!Cache::isStored($cache_id)) {
-            $result = (bool) Db::getInstance()->getValue('SELECT count(*) FROM `' . _DB_PREFIX_ . 'orders` WHERE `id_cart` = ' . (int) $this->id);
-            Cache::store($cache_id, $result);
+        $result = (bool) Db::getInstance()->getValue('SELECT count(*) FROM `' . _DB_PREFIX_ . 'orders` WHERE `id_cart` = ' . (int) $this->id);
 
-            return $result;
-        }
-
-        return Cache::retrieve($cache_id);
+        return $result;
     }
 
     /**
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -377,7 +377,7 @@ abstract class PaymentModuleCore extends Module
                     $transaction_id = null;
                 }
 
-                if (!$order->addOrderPayment($amount_paid, null, $transaction_id)) {
+                if (!isset($order) || !$order->addOrderPayment($amount_paid, null, $transaction_id)) {
                     PrestaShopLogger::addLog('PaymentModule::validateOrder - Cannot save Order Payment', 3, null, 'Cart', (int) $id_cart, true);
 
                     throw new PrestaShopException('Can\'t save Order Payment');
@@ -398,7 +398,7 @@ abstract class PaymentModuleCore extends Module
                         $message .= '<br />' . $this->trans('Warning: the secure key is empty, check your payment account before validation', [], 'Admin.Payment.Notification');
                     }
                     // Optional message to attach to this order
-                    if (isset($message) & !empty($message)) {
+                    if (!empty($message)) {
                         $msg = new Message();
                         $message = strip_tags($message, '<br>');
                         if (Validate::isCleanHtml($message)) {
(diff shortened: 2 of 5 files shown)


soyes:
We are having a problem with this pull request, because our code is correct, but the testing is not well designed.

The test is trying to create more than one order using the same cart; something that PrestaShop should not allow, but due to a bug, it is possible.

We have fixed this bug in this pull request, but now the problem is that the test fails because it needs the bug to work. The test should use different carts to create different orders, instead of just the same cart to create different orders.

PierreRambaud:
Look like, tests were malformed :thinking: 
ping @matks

matks:
> The test is trying to create more than one order using the same cart; something that PrestaShop should not allow, but due to a bug, it is possible.

If we are more specific, I think this makes sense 👍. "If customer X checkouts and pays for cart A (which creates an order), then cart A should not be used anymore" looks valid.
However what should be possible instead would be to duplicate a cart that has already been used to create an order, and use this duplicated cart to create another order.

So we can fix the tests indeed, by using duplicate carts.

But I'm not sure the PR is valid. The cache here has a purpose: to avoid unncessary SQL queries. It might be wrong if not handled correctly (it needs to be invalidated properly) but removing it might create performance issues 😢

soyes:
> But I'm not sure the PR is valid. The cache here has a purpose: to avoid unncessary SQL queries. It might be wrong if not handled correctly (it needs to be invalidated properly) but removing it might create performance issues 😢

As you say, the purpose of having a cache is to avoid unnecessary SQL queries. But in this case, those queries are necessary, because the function needs to check every time if the cart has an order associated to it, to avoid creating 2 or more orders using the same cart.

Furthermore, it is not usual to call this function twice or more using the same cart, so it should not create performance issues.

It is true that if the cart already has an associated order, we could use a cache, since once an order exists, it cannot stop existing, so it is not needed to run another query in that case. But if the cart does not have an associated order, it should be queried again, to avoid this bug.

matks:
> It is true that if the cart already has an associated order, we could use a cache, since once an order exists, it cannot stop existing, so it is not needed to run another query in that case. But if the cart does not have an associated order, it should be queried again, to avoid this bug.

I like this idea. So "if result is false, do not activate cache ; but if result is true, activate cache to return true immediately" 👍

soyes:
We have updated the code of the function in order to make it use the cache only when the cart has an associated order, and avoiding using the cache when the cart does not have an associated order.

With this change, the bug is fixed and the cache is used the best as possible.

However, the testing still fails due to the reasons mentioned above.

Matt75:
Warning a Cart can have multiple Order with same reference : https://github.com/PrestaShop/docs/issues/508

PierreRambaud:
> Warning a Cart can have multiple Order with same reference : [PrestaShop/docs#508](https://github.com/PrestaShop/docs/issues/508)

We should do it in a separated pull request. First merge this one, then fix the problem about cart having multiple orders. Wdyt @eternoendless ?

Progi1984 (approved):

eternoendless:
@Matt75 I agree with @PierreRambaud as this bug fix is correct (in terms that we shouldn't put this kind of data in cache without a proper invalidation strategy). Fixing this bug only  put in evidence other problems that were already present before, but hidden. This doesn't mean that your point is invalid, only that it should be resolved independently from this one.

@PierreRambaud would you mind rebasing this change into 1.7.7.x?

PierreRambaud:
Hi @soyes 

We are sorry, because you use your `develop` branch, we are not able to rebase it properly.
I cherry-pick your commit and things I added in a new PR.

Closed in favor of #19139

PR #17337, opened 2020-01-23
Order: avoid Notice error in case of missing state with PHP 7.4+

### LINKED ISSUE #13838: ContextErrorException while hide the only order state associate to an order
**Describe the bug**
If you edit an Order State to be hidden inside the client account and that it's the only one associated to the order, the getCustomerOrders() thrown an error.

**To Reproduce**
1. Go to Back Office and edit an order state (Payment by check, for demo datas)
2. Disable the "Show to customer" option
3. Save
4. Go to the customer account and check it's history

**Screenshots**
image

**Additionnal information**
PrestaShop version: 1.7.6.0-beta1
PHP version: N/A

rmilecki:
```
Starting with PHP 7.4 code like null['foo'] causes a notice error.
```

| Questions     | Answers
| Branch?       | develop
| Description?  | Fix PHP 7.4 notice caused by PHP backward incompatible change, see:https://www.php.net/manual/en/migration74.incompatible.php ("Array-style access of non-arrays"). It's not allowed to use `null['foo']` anymore.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #13838 [ContextErrorException while hide the only order state associate to an order](https://github.com/PrestaShop/PrestaShop/issues/13838).
| How to test?  | Open orders history page in FO

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17337)

### DIFF
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -952,11 +952,11 @@ class OrderCore extends ObjectModel
         }
 
         foreach ($res as $key => $val) {
-            // In case order creation crashed midway some data might be absent
+            // In case state is hidden or order creation crashed midway some data might be absent
             $orderState = !empty($val['id_order_state']) ? $indexedOrderStates[$val['id_order_state']] : null;
-            $res[$key]['order_state'] = $orderState['name'] ?: null;
-            $res[$key]['invoice'] = $orderState['invoice'] ?: null;
-            $res[$key]['order_state_color'] = $orderState['color'] ?: null;
+            $res[$key]['order_state'] = !empty($orderState['name']) ? $orderState['name'] : null;
+            $res[$key]['invoice'] = !empty($orderState['invoice']) ? $orderState['invoice'] : null;
+            $res[$key]['order_state_color'] = !empty($orderState['color']) ? $orderState['color'] : null;
         }
 
         return $res;


classes/order/Order.php
```diff
@@ -952,11 +952,11 @@ public static function getCustomerOrders($id_customer, $show_hidden_status = fal
...
         foreach ($res as $key => $val) {
-            // In case order creation crashed midway some data might be absent
+            // In case state is hidden or order creation crashed midway some data might be absent
             $orderState = !empty($val['id_order_state']) ? $indexedOrderStates[$val['id_order_state']] : null;
-            $res[$key]['order_state'] = $orderState['name'] ?: null;
-            $res[$key]['invoice'] = $orderState['invoice'] ?: null;
-            $res[$key]['order_state_color'] = $orderState['color'] ?: null;
+            $res[$key]['order_state'] = $orderState ? $orderState['name'] : null;
+            $res[$key]['invoice'] = $orderState ? $orderState['invoice'] : null;
+            $res[$key]['order_state_color'] = $orderState ? $orderState['color'] : null;
```
Progi1984:
```suggestion
            $res[$key]['order_state_color'] = $orderState ? ($orderState['color'] ?: null) : null;
            $res[$key]['invoice'] = $orderState ? ($orderState['invoice'] ?: null) : null;
            $res[$key]['order_state_color'] = $orderState ? ($orderState['color']  ?: null) : null;
```

Progi1984 (changes requested):
Check if if variable exists and if key exists.

classes/order/Order.php
rmilecki: Thanks for your review @Progi1984 ! You're right I should better check that `$orderState` array.

I have to use `!empty()` or `isset()` instead of your suggestion though. Code you suggested may still result in `Notice`, please consider following:
```
$orderState = [ ];
var_dump($orderState['color'] ?: null);
```
```
PHP Notice:  Undefined index: color
```

rmilecki:
I updated my branch with extra checks for `$orderState` array keys as suggested by @Progi1984

Progi1984 (approved):

classes/order/Order.php
```diff
@@ -952,11 +952,11 @@ public static function getCustomerOrders($id_customer, $show_hidden_status = fal
         }
 
         foreach ($res as $key => $val) {
-            // In case order creation crashed midway some data might be absent
+            // In case state is hidden or order creation crashed midway some data might be absent
             $orderState = !empty($val['id_order_state']) ? $indexedOrderStates[$val['id_order_state']] : null;
-            $res[$key]['order_state'] = $orderState['name'] ?: null;
-            $res[$key]['invoice'] = $orderState['invoice'] ?: null;
-            $res[$key]['order_state_color'] = $orderState['color'] ?: null;
+            $res[$key]['order_state'] = !empty($orderState['name']) ? $orderState['name'] : null;
```
PierreRambaud, line 957:
```suggestion
            $res[$key]['order_state'] = $orderState['name'] ?? null;
```

PierreRambaud (changes requested):
Do not use empty please. Use isset or `??` instead

classes/order/Order.php
Progi1984: Ping @rmilecki

PierreRambaud:
Hi @rmilecki ,

Thanks!

PR #17343, opened 2020-01-24
Improving the ps_newproducts module

helgvor-stoll:
Hi,
I think it's better to embrace the section with the "if $products" condition. So, we can let the module overriding in case of there no products

| Questions     | Answers
| Branch?       | 1.7.6.x
| Type?         | improvement
| Category?     | FO
| BC breaks?    | Does it break backward compatibility? yes/no
| Deprecations? | Does it deprecate an existing feature? yes/no

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17343)

### DIFF
--- a/themes/classic/modules/ps_newproducts/views/templates/hook/ps_newproducts.tpl
+++ b/themes/classic/modules/ps_newproducts/views/templates/hook/ps_newproducts.tpl
@@ -22,7 +22,7 @@
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  * International Registered Trademark & Property of PrestaShop SA
  *}
-
+{if $products}
 <section class="featured-products clearfix mt-3">
   <h2 class="h2 products-section-title text-uppercase">
     {l s='New products' d='Shop.Theme.Catalog'}
@@ -36,4 +36,4 @@
     {l s='All new products' d='Shop.Theme.Catalog'}<i class="material-icons">&#xE315;</i>
   </a>
 </section>
-
+{/if}


themes/classic/modules/ps_newproducts/views/templates/hook/ps_newproducts.tpl
```diff
@@ -36,4 +36,4 @@
     {l s='All new products' d='Shop.Theme.Catalog'}<i class="material-icons">&#xE315;</i>
   </a>
 </section>
```
NeOMakinG, line 38: Could you add some padding to the section ? Thanks

NeOMakinG (changes requested):
Hey @helgvor-stoll 

Can you update the title of the PR so we can be able to see what the PR really does, for example :
Hide new products if there are no new products using a condition in ps_newproducts

Also, can you fill the description, how to test, bc breaks, and deprecation section ? Feel free to read the contribution guideline : https://devdocs.prestashop.com/1.7/contribute/contribution-guidelines/

Also, when you modify a module, you need to create a PR on the module's repo too, see there : 
https://github.com/PrestaShop/ps_newproducts

Thanks for your contribution !

themes/classic/modules/ps_newproducts/views/templates/hook/ps_newproducts.tpl
```diff
@@ -22,7 +22,7 @@
  * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
  * International Registered Trademark & Property of PrestaShop SA
  *}
-
+{if $products}
```
jolelievre, line 25: Not sure it's the best solution to simply hide/remove the whole section
It might be your way of displaying it, but some other merchants would prefer a text in place of product list

Both are possible and valid point of views, but it shouldn't be force in the core template It would be more interesting to manage this through a configuration in the module to adapt the behaviour

besides as you said in the description this can be overridden in a module or a theme template, so if you need to change this behaviour in one your shop that would be more fitting for now no?

Progi1984:
Ping @helgvor-stoll

Progi1984:
@helgvor-stoll Coud you target the `develop` branch ? Please :)

matks:
Hi, unfortunately since we got no answer back from author and @jolelievre feedback is relevant (it cannot be merged like this) I close the PR

Feel free to open it again if you think it's needed 😄 

PR #17348, opened 2020-01-24
Rework OrderHandlers calculations to use Prestashop Number

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | See below
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #17282 and #18020
| How to test?  | See below

### Initial scope

- fix `@todo` in **AddCartRuleToOrderHandler** ('move to private method' in couple places)

### Scope expansion

- refactors **AddCartRuleToOrderHandler**, **AddProductToOrderHandler**, **DeleteProductFromOrderHandler**, and **UpdateProductInOrderHandler** handlers into private methods
- implement Number calculations instead of floats for handlers **AddCartRuleToOrderHandler**, **AddProductToOrderHandler**, **DeleteProductFromOrderHandler**, and **UpdateProductInOrderHandler**
- added some behat tests
- removed the ability to update invoices when adding discount to order _as requested by PM_
- fixed bug where "free shipping" checkbox wasn't displayed when adding product with new invoice

## How to test ?

Use `sell/orders/orders/[orderId]/view` to reach migrated order view page. Make sure `Add discount` is working as expected.
**Tip:** to add product with new invoice the order status must be "payment accepted" (works with some others as well).
Be sure to check the product total prices (shipping, total, taxes, discounts- _this one also increments when product is added with free shipping option_).
:warning:  **WARNING**
Pr included a heavy rebase associated with price calculations from #17877 and #17476. So its important to test according to those PR's as well.

## Known issues
* when adding product with free shipping, the discounts are increased, but not shipping price, which leads to wrong total numbers. (the overall total for the order is correct though). _(this one is hard to fix and is delayed to at least 1.7.8)_
* After adding product with new invoice and free shipping, the page needs to be refreshed to see the new specific price rule in list. _(this should be solved in another PR)_

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17348)


src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -49,82 +50,137 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
...
     {
         $order = $this->getOrderObject($command->getOrderId());
+        $orderInvoice = $this->createSingleOrderInvoice($order, $command);
 
-        // If the discount is for only one invoice
-        if ($order->hasInvoice() && null !== $command->getOrderInvoiceId()) {
-            $orderInvoice = new OrderInvoice($command->getOrderInvoiceId()->getValue());
-            if (!Validate::isLoadedObject($orderInvoice)) {
-                throw new OrderException('Can\'t load Order Invoice object');
+        $discountValue = (float) str_replace(',', '.', $command->getDiscountValue());
```
matks: 😱

matks:
That would be very cool to have some behat tests to help us validate this complex and critical code. For a later PR maybe ?

matks (dismissed):

matks:
Closing and reopening, it seems Travis is checking against `develop` and not against `1.7.7.x`

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -49,82 +50,137 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
...
+        $cartRules = $this->buildCartRulesByType($command, $order, $discountValue, $orderInvoice);
+
+        $result = true;
+
+        foreach ($cartRules as &$cartRule) {
+            $cartRuleObj = $this->createCartRule($command, $order, $cartRule, $discountValue);
+
+            if (false === $cartRuleObj->add()) {
+                break;
```
PierreRambaud: should we really need to break?

PierreRambaud: what if we add nine rules, but the ten one is broken, we will have an exception, what about the nine rules?

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -49,82 +50,137 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
...
+            }
+
+            $cartRule['id'] = $cartRuleObj->id;
+        }
+
+        if ($result) {
```
PierreRambaud: $result is always true here :thinking:

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -49,82 +50,137 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
...
+
+    /**
+     * @param array $cartRules
+     * @param Order $order
+     * @param float $discountValue
+     * @param OrderInvoice|null $orderInvoice
+     */
+    private function buildPercentTypeCartRules(array &$cartRules, Order $order, float $discountValue, ?OrderInvoice $orderInvoice): void
+    {
+        if ($discountValue < 100) {
```
PierreRambaud: why 100, must be a constant

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
zuk3975: Sorry guys, I have no knowledge about whats happening there deep, I just tried to at least separate to private methods, didn't wan't to break the logic.

zuk3975:
addressed comments.

PierreRambaud:
Thanks @zuk3975. Because it's a migration work, I'll not annoy you anymore :sweat_smile:

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
zuk3975: Fixed. (not sure why its not updating here):
But it looks like it previously was:
```
            if ($result = $cartRuleObj->add()) {
                $cartRule['id'] = $cartRuleObj->id;
            } else {
                break;
            }
```

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
zuk3975: Fixed above:
```
            if ($result = $cartRuleObj->add()) {
                $cartRule['id'] = $cartRuleObj->id;
            } else {
                break;
            }
```
Original stmt from previous pr.

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
zuk3975: As I understand the command adds only one cart rule.
I think its all wrong with the naming, It should be singular.
buildCartRulesByType is only checking which type of cartRule should be built.

Ill make sure its that, and update it..

zuk3975:
> Thanks @zuk3975. Because it's a migration work, I'll not annoy you anymore

hmm sorry, but you should :smile: I've noticed It still needs some changes. see https://github.com/PrestaShop/PrestaShop/pull/17348#discussion_r372266207

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
PierreRambaud: It's disturbing `AddCartRuleToOrderHandler` should be `AddCartRulesToOrderHandler` or the `buildCartRulesByType` should be singular and receive only one cart rule (and the foreach became useless) :sweat_smile:

zuk3975: yep, trying to figure out whats happening, its confusing cuz its mostly legacy code pasted to handler :smile:

It should create only one cartRule, but multiple invoices has something to do with multiple cartRules, so I don't really know how should it behave and how the order gets multiple invoices in first place.

zuk3975:
@matks, the code is crappy inside the adapter (it was like pasted straight from legacy before I touched it - most confusing part is that command says its adding single CartRule, but it does create an array of cart rules based on invoices inside..), but I don't know how it suppose to work, so I think I will not try to improve it anymore, as I might break smth, and its is not what we want in featurefreeze.
However, I could rework this if i have clear specifications about discount creation (and use cases with single/multiple invoices)

matks:
> @matks, the code is crappy inside the adapter (it was like pasted straight from legacy before I touched it - most confusing part is that command says its adding single CartRule, but it does create an array of cart rules based on invoices inside..), but I don't know how it suppose to work, so I think I will not try to improve it anymore, as I might break smth, and its is not what we want in featurefreeze.
> However, I could rework this if i have clear specifications about discount creation (and use cases with single/multiple invoices)

Let's do our best 😢 

If PM writes such specs, indeed we'll be able to improve this

matks (dismissed):

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
jolelievre: It was used in the legacy code to reformat the received string, but since the command has a strong typed float it's useless

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
jolelievre: It happens when you add a product to an order which already has an invoice You can chosse to add the new product to a new invoice..

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -49,82 +51,137 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
     public function handle(AddCartRuleToOrderCommand $command): void
     {
         $order = $this->getOrderObject($command->getOrderId());
+        $orderInvoice = $this->createSingleOrderInvoice($order, $command);
```
jolelievre: It would be clearer if you managed the `single order invoice` case here, if `createSingleOrderInvoice` return null it means you have to apply the discount to all its invoices. So you could do:
```php
$orderInvoice = $this->createSingleOrderInvoice($order, $command);
if (null !== $orderInvoice) {
    $orderInvoices = [$orderInvoice];
} else {
    $orderInvoices = $order-> getInvoicesCollection()->getAll();
}
```

This way you always have an array of OrderInvoice objects, and you can simplify most of your methods, you don't need to transfer the Order and a potentially null OrderInvoice. The method `buildCartRulesByType` (and its subsequent methods) can be simplified into `buildCartRulesByType($orderInvoices, $discountValue, $cartRuleType)`

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -49,82 +51,137 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
...
-                        );
-                        $cartRules[0]['value_tax_excl'] = Tools::ps_round(
-                            $order->total_paid_tax_excl * $discountValue / 100,
-                            2
-                        );
-                    }
-                } else {
-                    throw new OrderException('Percentage discount value cannot be higher than 100%.');
-                }
+                $this->buildPercentTypeCartRules($cartRules, $order, $discountValue, $orderInvoice);
```
jolelievre: I don't think you need to instanciate `$cartRules` beforehand and pass it around in each method Since you only deal with one cart rule type, only one method will be called, so each methods `buildPercentTypeCartRules`,  `buildAmountTypeCartRules` and `buildPercentTypeCartRules` can create its own internal array and return it.

This way (if you already made the simplification I suggested earlier with the array of invoices) you can simply do:

```php
switch ($cartRuleType) {
    case OrderDiscountType::DISCOUNT_PERCENT:
        return $this->buildPercentTypeCartRules($orderInvoices, $discountValue);
        break;
    ....
}
```

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -49,82 +51,137 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
...
+    /**
+     * @param array $cartRules
+     * @param Order $order
+     * @param float $discountValue
+     * @param OrderInvoice|null $orderInvoice
+     */
+    private function buildPercentTypeCartRules(array &$cartRules, Order $order, float $discountValue, ?OrderInvoice $orderInvoice): void
+    {
+        if ($discountValue < PercentageDiscount::MAX_PERCENTAGE) {
+            if (isset($orderInvoice)) {
```
jolelievre: If you only deal with an array of OrderInvoice this method can be simplified (the other two as well) you wouldn't need to deal with three different cases each time only:
```php
if (count($orderInvoices) > 0) {
 ...
} else {
 ...
}
```

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -134,150 +191,196 @@ public function handle(AddCartRuleToOrderCommand $command): void
...
+    /**
+     * @param AddCartRuleToOrderCommand $command
+     * @param Order $order
+     * @param array $cartRule
+     * @param float $discountValue
+     *
+     * @return CartRule
+     */
+    private function createCartRule(
+        AddCartRuleToOrderCommand $command,
```
jolelievre, line 180: You can simply use the cart rule type as a string instead of passing around the command all the time

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -134,150 +191,196 @@ public function handle(AddCartRuleToOrderCommand $command): void
...
+        $cartRuleObj->active = 0;
+
+        return $cartRuleObj;
     }
 
     /**
-     * @param OrderInvoice $orderInvoice
-     * @param float $valueTaxIncl
-     * @param float $valueTaxExcl
+     * @param AddCartRuleToOrderCommand $command
```
jolelievre: You can simplify by using:

```php
/**
 * @param string $cartRuleName
 * @param int $orderId
*/
```

jolelievre (changes requested):
This is not an easy code to say the least ^^
I made a few suggestion that will allow to reduce the size of the code and will make it more readable I think, let me know what you think of it

matks:
As said by @zuk3975, 
> but I don't know how it suppose to work, so I think I will not try to improve it anymore, as I might break something, and its is not what we want in feature freeze

This is unreadable and untested legacy code that we want only to move. I strongly suggest not to touch it.

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
zuk3975: It is typed as null|Number . If I use Number I need to rework all calculations, thats risky and alot of code. You think its worth it ?

zuk3975:
@jolelievre addressed comments, I think its enough for this one, a bit clearer now :smile: need QA to retest (although i tried some usecases and it seem to work)

matks (dismissed):

zuk3975:
I will add some behats for this in same pr

zuk3975:
@matks, @MatShir I've added some behat but noticed kinda strange behavior, and wanted to clarify.
So, when adding a product to order with new invoice there is a checkbox for "free shipping". 
1.Should the shipping price be applied to the product alone or override the whole order shipping and recalculate?
Currently it is applied only to added product.
2. If I select free shipping, shouldn't it be reflected in discount total and shipping total of the whole order?
Currently it is only reflected in Total price of order, but doesn't add up to total shipping.

Need your answers to fix the behavior (if needed) and finish the pr, thanks.

zuk3975:
> @matks, @MatShir I've added some behat but noticed kinda strange behavior, and wanted to clarify.
> So, when adding a product to order with new invoice there is a checkbox for "free shipping".
> 1.Should the shipping price be applied to the product alone or override the whole order shipping and recalculate?
> Currently it is applied only to added product.
> 2. If I select free shipping, shouldn't it be reflected in discount total and shipping total of the whole order?
> Currently it is only reflected in Total price of order, but doesn't add up to total shipping.
> 
> Need your answers to fix the behavior (if needed) and finish the pr, thanks.

Ok so, we had sesion with @MatShir, we found out that the calculations of shipping, discounts and sometimes a total price is wrong. Going to figure it out and fix it in the same PR.

zuk3975:
@jolelievre, @matks ready for review again

matks (dismissed):

Progi1984:
@zuk3975 Please fix Travis CI tests : Behat test from `Features/Scenario/Order/add_cart_rule_to_order.feature` are all broken.

zuk3975:
> @zuk3975 Please fix Travis CI tests : Behat test from `Features/Scenario/Order/add_cart_rule_to_order.feature` are all broken.

Yeah im on it, it was something wrong with local test database fixtures.

zuk3975:
@jolelievre , @Progi1984 , @matks, @PierreRambaud waiting for approve

matks (approved):

tests/Integration/Behaviour/Features/Scenario/Order/add_cart_rule_to_order.feature
```diff
@@ -0,0 +1,325 @@
...
+    Given the current currency is "USD"
+    And country "US" is enabled
+    And there is customer "testCustomer" with email "pub@prestashop.com"
+    And customer "testCustomer" has address in "US" country
+    And customer "testCustomer" has an empty cart "dummy_cart"
+    And cart "dummy_cart" delivery and invoice address country for customer "testCustomer" is "US"
+    And the module "dummy_payment" is installed
+    And I am logged in as "test@prestashop.com" employee
+    And I add 2 products "Mug The best is yet to come" to the cart "dummy_cart"
+    And I add order "bo_order1" with the following details:
```
rokaszygmantas, line 17:
```suggestion
    And I place order "bo_order1" with the following details:
```

tests/Integration/Behaviour/Features/Scenario/Order/add_cart_rule_to_order.feature
```diff
@@ -0,0 +1,325 @@
...
+  Background:
+    Given the current currency is "USD"
+    And country "US" is enabled
+    And there is customer "testCustomer" with email "pub@prestashop.com"
+    And customer "testCustomer" has address in "US" country
+    And customer "testCustomer" has an empty cart "dummy_cart"
+    And cart "dummy_cart" delivery and invoice address country for customer "testCustomer" is "US"
+    And the module "dummy_payment" is installed
+    And I am logged in as "test@prestashop.com" employee
+    And I add 2 products "Mug The best is yet to come" to the cart "dummy_cart"
```
rokaszygmantas, line 16:
```suggestion
    And I add 2 quantity of product "Mug The best is yet to come" to the cart "dummy_cart"
```

rokaszygmantas (dismissed):

tests/Integration/Behaviour/Features/Scenario/Order/add_cart_rule_to_order.feature
zuk3975: I doubt its in pr scope, because those were implemented earlier and just reused here.

tests/Integration/Behaviour/Features/Scenario/Order/add_cart_rule_to_order.feature
zuk3975: I doubt its in pr scope, because those were implemented earlier and just reused here.

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -49,235 +52,296 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
...
+
+        $result = true;
+
+        foreach ($reductionValuesByInvoiceId as &$reducedValues) {
+            $cartRuleObj = $this->createCartRule($command, $order, $reducedValues, $discountValue);
+
+            if ($result = $cartRuleObj->add()) {
+                $reducedValues['cart_rule_id'] = $cartRuleObj->id;
+            } else {
+                break;
```
jolelievre: This `break` happens when a cart rule could not be added, then `$result == false` and an exception is thrown L95
To make the code clearer could you please throw the exception here instead of the `break` ?

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -49,235 +52,296 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
...
-            $orderInvoice = new OrderInvoice($command->getOrderInvoiceId()->getValue());
-            if (!Validate::isLoadedObject($orderInvoice)) {
-                throw new OrderException('Can\'t load Order Invoice object');
+        if (null !== $orderInvoice) {
+            $orderInvoices = [$orderInvoice];
+        } else {
+            $orderInvoices = $order->getInvoicesCollection()->getResults();
+        }
+
+        $discountValue = (float) $command->getDiscountValue()->toPrecision(CommonAbstractType::PRESTASHOP_DECIMALS);
```
jolelievre: This should be related to the currency, using the `ComputingPrecision` class, but this can be improved later 😉

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -498,27 +414,91 @@ private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping)
...
+        $order->total_wrapping_tax_incl += (float) abs($cart->getOrderTotal(true, Cart::ONLY_WRAPPING));
+        $order->total_wrapping += (float) abs($cart->getOrderTotal(true, Cart::ONLY_WRAPPING));
+
+        //shipping
+        $order->total_shipping += (float) abs($cart->getOrderTotal(true, Cart::ONLY_SHIPPING));
+        $order->total_shipping_tax_incl += (float) abs($cart->getOrderTotal(true, Cart::ONLY_SHIPPING));
+        $order->total_shipping_tax_excl += (float) abs($cart->getOrderTotal(false, Cart::ONLY_SHIPPING));
+    }
+
+    private function createFreeShippingCartRule(Cart $cart, Order $order): CartRule
```
jolelievre, line 511:
```suggestion
    private function addFreeShippingCartRule(Cart $cart, Order $order): CartRule
```

tests/Integration/Behaviour/Features/Scenario/Order/add_cart_rule_to_order.feature
jolelievre: Besides it's not easier to read

tests/Integration/Behaviour/Features/Scenario/Order/add_cart_rule_to_order.feature
jolelievre: This one makes sense, but indeed not in this PR scope I think

jolelievre (commented):
A few feedback but nothing blocking

jolelievre (dismissed):

matks (dismissed):

Robin-Fischer-PS:
Hi @zuk3975 :)

I have two bugs : 
- When adding a Free Shipping, an Exception is displayed
pr17348 free ship
- When choosing the "Apply on all invoices", the discount is applied only once, but displayed once for each invoice. It should be applied as many times as there is invoices
pr17348 all invoice voucher

Thanks !

zuk3975:
> When choosing the "Apply on all invoices", the discount is applied only once, but displayed once for each invoice. It should be applied as many times as there is invoices

Sorry this can't be right ( I know it worked like this in legacy, but it doesn't sound right).
@colinegin, @sam-pires could you please take a look? (_We have discussed it abit in chat, but it would be great to see some recap here_)

colinegin:
So just a feedback about invoice edition : this is normal that we allow it, as it is not illegal to edit an invoice (for instance delivery address) even though the payment has been accepted. It should not be done when the order is too old, but there is no precise information about what details can be edited and until when it can be modified. So the only solution we have is to let the user edit what he wants, it's his responsibility to be compliant with the law.

About the "apply to all invoices" button, I agree with @jolelievre, I recommend to drop this feature as : 

- we are not sure how it is supposed to work (especially the several invoices on one order feature) and i'm afraid everything we do could add new undesired behaviour changes
- there is still a workaround for the merchant : he can always modify invoices one by one manually

Is this ok for you @zuk3975 ?

zuk3975:
> Is this ok for you @zuk3975 ?

yep, so Ill drop the "update all invoices feature"

matks (dismissed):
Code looks good to me 👍 
Re-running PrettyCI and Travis to see if issues are real issues or random glabberish from drunken CI

zuk3975:
Removed the feature "Update all invoices". Reworked calculations using Number().

matks (dismissed):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -170,49 +146,40 @@ public function handle(AddProductToOrderCommand $command)
             $specificPrice->delete();
         }
 
-        $order = $order->refreshShippingCost();
```
zuk3975, line 174: todo - recheck this. Depends from configurations

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
```diff
@@ -44,53 +46,70 @@
...
+    ) {
+        $this->updateOrderStatusHandler = $updateOrderStatusHandler;
+    }
+
     /**
      * {@inheritdoc}
      */
     public function handle(DeleteProductFromOrderCommand $command)
     {
+        //@todo: what if its the last product, should order stay with zero prices or be cancelled?
```
zuk3975: @matks need help about this one.

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -498,27 +440,163 @@ private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping)
...
+    /**
+     * couldn't use Order::refreshShippingCost() with free shipping because it recalculates whole cart shipping
+     *  and then total numbers doesn't fit together.
+     *  e.g initial shipping was $7 and discounts $0
+     *  we add product with free shipping (behind the scenes it adds a discount of shipping price ($7))
+     *  so for total prices to fit - shipping should be $14 and discounts $7
+     *  using  Order::refreshShippingCost() would end up shipping being $7
+     * @todo: If i do it manually, i don't update OrderCarrier shipping price.
+     *      Difference occurs between order shipping price and carrier shipping price
+     *      That might introduce problems. How To fix it?
```
zuk3975: @matks I have a problem with this (explained in TODO). If its unclear ill explain in chat. Could you help?

zuk3975: It would be easily solved if we apply shipping price to whole cart instead of one product. Or completely remove the option to create new invoice when adding product, or remove the free shipping option when adding product with new invoice. Idk all those features are very strange and doesn't comply with the system design

zuk3975: this was discussed in standup and chat. Decided to leave is at it was in legacy and dive deeper in later versions.

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
```diff
@@ -216,7 +216,7 @@ export default class OrderProductRenderer {
   }
 
   toggleProductAddNewInvoiceInfo() {
-    if ($(OrderViewPageMap.productAddInvoiceSelect).val() === 0) {
+    if (parseInt($(OrderViewPageMap.productAddInvoiceSelect).val()) === 0) {
```
Progi1984:
```suggestion
    if (parseInt($(OrderViewPageMap.productAddInvoiceSelect).val(), 10) === 0) {
```

src/Core/Domain/CartRule/ValueObject/PercentageDiscount.php
```diff
@@ -33,6 +33,8 @@
  */
 class PercentageDiscount
 {
+    const MAX_PERCENTAGE = 100;
```
Progi1984, line 36: Why don't add a MIN_PERCENTAGE ?

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -530,6 +662,89 @@ public function checkProductStockDifference(int $productDifference, string $fact
...
+    }
+
+    /**
+     * @Given /^I (enable|disable) shipping recalculation config$/
+     */
+    public function setShippingRecalculationConfig(bool $status): void
+    {
+        $configName = 'PS_ORDER_RECALCULATE_SHIPPING';
+
+        Configuration::set($configName, $status);
```
Progi1984:
```suggestion
        Configuration::set('PS_ORDER_RECALCULATE_SHIPPING', $status);
```

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
zuk3975: is it necessary ? I get same results with or without it :thinking:

src/Core/Domain/CartRule/ValueObject/PercentageDiscount.php
zuk3975: what would it be? Min percentage should be just greater than zero, but not the zero itself. So would it be good to make MIN_PERCENTAGE = 0.000001 ?

zuk3975:
The conflicts are too harsh. I don't know how to fix it anymore! :bomb:

matks (approved):

matks:
@zuk3975 After deep analysis of this PR, I suggest the following plan to re-build it and deliver the improvements

## Scope summary

First, I summarized the large scope of this PR like this:

**Initial scope**

- fix `@todo` in **AddCartRuleToOrderHandler** ('move to private method' in couple places)

**Scope expansion**

- refactors **AddCartRuleToOrderHandler**, **AddProductToOrderHandler**, **DeleteProductFromOrderHandler**, and **UpdateProductInOrderHandler** handlers into private methods
- implement Number calculations instead of floats for handlers **AddCartRuleToOrderHandler**, **AddProductToOrderHandler**, **DeleteProductFromOrderHandler**, and **UpdateProductInOrderHandler**
- added some behat tests
- removed the ability to update invoices when adding discount to order _as requested by PM_
- fixed bug where "free shipping" checkbox wasn't displayed when adding product with new invoice

Please tell me if I was wrong on this scope.

## Roadmap to merge

I suggest to pick pieces of this PR and deliver them as new PRs, from new branches, one piece at a time. What I mean is in fact __looking at the diff and copy-pasting pieces of code into new branches created from 1.7.7.X__.

I suggest doing this in this order:
1. Pick the code necessary to remove the ability to update invoices when adding discount to order ; pick related behat tests if there is one ; create a PR for it and only for this. Pick **the smallest possible amount of code** following a TDD fashion. No refactoring in this PR, no conversion from float to numbers.
2. Pick the code necessary to fix the bug where "free shipping" checkbox was not displayed after adding product with new invoice. Create a PR with this piece of code. Again pick **the smallest possible amount of code**. I would like that **this is impossible to remove any of the committed lines without breaking the bugfix.**

These 2 items are the necessary items for 177x. Following items can be done 178 (develop). We'll see when we reach this step how far we are in 177 release process.

3. Pick the code to refactor AddCartRuleToOrderHandler using more functions (split big functions into small ones ; pick related behat tests if there are some. Use the behat tests to validate we break nothing. Create a PR for that. No other refactoring in this PR, no conversion from float to numbers.
4. Do the same for AddProductToOrderHandler, DeleteProductFromOrderHandler, and UpdateProductInOrderHandler, one PR for each.
5. Finally pick the code to refactor AddCartRuleToOrderHandler using Numbers. At this point I expect all behat tests to have been merged before, allowing them to make sure we break nothing. Create a PR for that.
6. Do the same for AddProductToOrderHandler, DeleteProductFromOrderHandler, and UpdateProductInOrderHandler, one PR for each.

@zuk3975 This is a long process and you might be too busy to do it. I will share the workload if necessary.

zuk3975:
> I suggest doing this in this order:
> 
> 1. Pick the code necessary to remove the ability to update invoices when adding discount to order ...
> 2. Pick the code necessary to fix the bug where "free shipping" checkbox was not displayed after adding product with new invoice....
> These 2 items are necessary for 1.7.7

1. #18524
2. #18544

zuk3975:
Created a reminder issue for 1.7.8 #18545. Closing this pr.

Issue #17347, opened 2020-01-24
Changing order products randomly corrupts order prices & breaks editing


rmilecki:
**Describe the bug**
PrestaShop allows editing order products. We use that if one of products is unavailable or customer made a mistake. **Sometimes** editing an order results in **database corruption**. Order prices get invalid and they don't match order products. Further order products modifications may also stop working.

**To Reproduce**
We **don't** know how to reproduce this. This bug definitely requires some very specific circumstances that remain unknown to us. That means we can't debug this issue properly.

I was only able to analyze one order case as I happened to have database snapshot before the corruption. I'm providing that case details in hope someone can conclude what has happened meanwhile.

**Case XHGOUAGRP before corruption**
Shop customer ordered 11 Lasagne LV products and 7 Tortilla RP products. We set 10% discount for that order. It was all good:
![before-good](https://user-images.githubusercontent.com/452972/73077386-5b063d80-3ec0-11ea-91bd-a476b958fc8f.png)
Result of `SELECT * FROM orders/order_detail/order_invoice WHERE id_order = 11077;`:
[before-good.txt](https://github.com/PrestaShop/PrestaShop/files/4109090/before-good.txt)

**Case XHGOUAGRP after corruption**
We decided to replace "Lasagne LV" product with "Lasagne SCH" due to unavailability of the first one. That has failed. We couldn't remove "Lasagne LV" and order prices got wrong:
![after-bad](https://user-images.githubusercontent.com/452972/73077385-5b063d80-3ec0-11ea-80fc-e67fe4930067.png)
Result of `SELECT * FROM orders/order_detail/order_invoice WHERE id_order = 11077;`:
[after-bad.txt](https://github.com/PrestaShop/PrestaShop/files/4109093/after-bad.txt)

If you look closely you'll see that new `order_detail` entry doesn't have `id_order_invoice` set. This causes further problems, e.g. trying to change quantity of "Lasagne SCH" products:
![after-edit](https://user-images.githubusercontent.com/452972/73079926-48dace00-3ec5-11ea-90d3-c7236c3006dd.png)
results in sending request to `AdminOrders` with following data:
```
ajax: 1
token: 3a88a4780bfca3c624a07585f40a487b
action: editProductOnOrder
id_order: 11077
product_id_order_detail: 131129
product_price_tax_excl: 14.14
product_price_tax_incl: 14.85
product_quantity: 9
product_invoice: 10548
```
but getting following reply:
```
[...]

PR #17354, opened 2020-01-25
Display error alert if editing order products in BO fails

rmilecki:
```
Both actions: editProductOnOrder and deleteProductLine may fail. If that
happens it's important for user experience & debugging purposes to
display error alert.
```

| Questions     | Answers
| Branch?       | develop
| Description?  | In some corer cases database may get corrupted (e.g. due to bugs in code). Example: #17347. If that happens AJAX requests may fail and AJAX response may differ from what is expected (e.g. `editProductOnOrder` and `deleteProductLine`). It's important then to report a problem so it can be noticed/reported/debugged.<br>There are already few JavaScript files reporting `TECHNICAL ERROR`s like this.
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | This doesn't **fix** any real bug. It **improves** BO behavior when dealing with bugs like #17347
| How to test?  | On your **experimental** shop execute<br>`UPDATE orders SET total_paid = 1 WHERE 0;`<br>(remove that `WHERE 0` to make it work), open any order in BO and try to delete a single product from it. Delete action should fail & you should see error alert.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17354)

### DIFF
--- a/js/admin/orders.js
+++ b/js/admin/orders.js
@@ -754,6 +754,9 @@ function init()
 					} else {
             jAlert(data.error);
           }
+				},
+				error: function(jqXHR, textStatus, errorThrown) {
+					jAlert('TECHNICAL ERROR:\n\nDetails:\n\nText status: ' + textStatus + '\n\nError thrown: ' + errorThrown + '\n\nResponse text:\n' + jqXHR.responseText);
 				}
 			});
 		}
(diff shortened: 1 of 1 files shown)


matthieu-rolland (commented):
Hello @rmilecki and thank you very much for your contribution 👍 

You are making a modification on the legacy BO order page... but this page is being remade (symfony migration), and will be available in the next release.... (1.7.7).

PR #17360, opened 2020-01-26
Add transactions support to Db and use them when moving carrier

rmilecki:
```
Carrier: use transaction when moving carrier

It's required to keep database in consistent state if a later query
fails for whatever reason. As much as it's unlikely in this simple case
it's still a good practice and provides an example of using
transactions.
```
```
Db: add support for SQL transactions

Transactions should be used when executing multiple related queries.
They are required to keep database in consistent state in case of any
errors.

MySQL supports transactions only when using InnoDB however this code is
safe even for MyISAM. PHP "checks for transaction capabilities" for a
given driver only. So this code won't cause exception on MyISAM (it'll
just do nothing).

Ref: https://www.php.net/manual/en/pdo.transactions.php
```

| Questions     | Answers
| Branch?       | develop
| Description?  | All related database change queries should be executed in a single transaction. This way we make sure database won't get in inconsistent state if processing fails at some point.<br>This Pull Request adds transaction support to `Db` API and uses it when moving carrier. PrestaShop code should be updated incrementally to use transactions in all critical actions.
| Type?         | bug fix / improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Those changes are **inspired** by my problems with **orders** inconsistenty as reported in the [#17347](https://github.com/PrestaShop/PrestaShop/issues/17347). As this point I focused on moving carrier though (to start with something simple).
| How to test?  | Navigate to Carrier & make sure carriers can be moved up/down.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17360)

### DIFF
--- a/controllers/admin/AdminOrdersController.php
+++ b/controllers/admin/AdminOrdersController.php
@@ -2125,6 +2125,8 @@ class AdminOrdersControllerCore extends AdminController
         // Total method
         $total_method = Cart::BOTH_WITHOUT_SHIPPING;
 
+        Db::getInstance()->beginTransaction();
+
         // Create new cart
         $cart = new Cart();
         $cart->id_shop_group = $order->id_shop_group;
@@ -2420,6 +2422,8 @@ class AdminOrdersControllerCore extends AdminController
         // Update Order
         $res &= $order->update();
 
+        Db::getInstance()->commit();
+
         die(json_encode([
             'result' => true,
             'view' => $this->createTemplate('_product_line.tpl')->fetch(),
@@ -2491,6 +2495,8 @@ class AdminOrdersControllerCore extends AdminController
 
         // Check fields validity
         $this->doEditProductValidation($order_detail, $order, isset($order_invoice) ? $order_invoice : null);
+
+        Db::getInstance()->beginTransaction();
 
         // If multiple product_quantity, the order details concern a product customized
         $product_quantity = 0;
@@ -2595,6 +2601,8 @@ class AdminOrdersControllerCore extends AdminController
         // Update product available quantity
         StockAvailable::updateQuantity($order_detail->product_id, $order_detail->product_attribute_id, ($old_quantity - $order_detail->product_quantity), $order->id_shop);
 
+        Db::getInstance()->commit();
+
         $products = $this->getProducts($order);
         // Get the last product
         $product = $products[$order_detail->id];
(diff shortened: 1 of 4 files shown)


rmilecki:
> This is your first pull request on the PrestaShop project.

Dear @prestonBot (maintainer), it's my [4th pull request](https://github.com/PrestaShop/PrestaShop/pulls?q=author%3Armilecki), you may stop welcoming me now I think :wink:

controllers/admin/AdminOrdersController.php
```diff
@@ -2125,6 +2125,8 @@ public function ajaxProcessAddProductOnOrder()
         // Total method
         $total_method = Cart::BOTH_WITHOUT_SHIPPING;
 
+        Db::getInstance()->beginTransaction();
```
PierreRambaud, line 2128: This kind of translations must be done only if innodb is used

rmilecki: Good point! I'll fix that!
> If the underlying driver does not support transactions, a PDOException will be thrown (regardless of your error handling settings: this is always a serious error condition).

Reference: https://www.php.net/manual/en/pdo.transactions.php

PierreRambaud (changes requested):
We still support MyISAM

Progi1984:
@PrestaShop/prestashop-core-developers What do you think about this PR ?

Progi1984:
Hello @rmilecki, thanks for your contribution, would you mind to rebase your pull request for fixing conflicting file and fixing the check for InnoDB ? Thanks

PierreRambaud:
Hi @rmilecki ,

Thanks!

controllers/admin/AdminOrdersController.php
rmilecki: @PierreRambaud: I've checked PHP documentation and it says it's actually **safe** to use transactions even with MyISAM. This code **won't** trigger an exception. I've tested it with PrestaShop `develop` branch running on MySQL 5.7.35 with all tables using MyISAM.

rmilecki:
> Feel free to reopen or open another one if you think it's still relevant.

@PierreRambaud, @Progi1984: I can't find a way to reopen this pull request. Would you be so kind to reopen it for me, please? I've updated my branch.

Progi1984:
@rmilecki I can't too. Could you recreate a new PR ?

PierreRambaud:
@rmilecki the branch was force-pushed or recreated that's why we are not able to reopen the pr ^^

PR #17373, opened 2020-01-27
Fix currency localization

### LINKED ISSUE #16686: When I Install a new Language currency_lang values are not correctly set
**Describe the bug**
During installation the `currency_lang` values are not correctly set.

**To Reproduce**
I already have two languages in my shop (French, English), and two currencies (Euro, Dollar). And I have customization on one of them (let's say Euro).
If I install a new Language the currency_lang for the new Language are correctly added in the database, but their value are incorrect. They don't match the default value of the new language (name, symbol and pattern are not correct). It seems the data from the default language are copied in the other languages instead of using the CLDR data.

Capture d’écran 2019-12-04 à 16 33 26

If I go and edit the incorrect currency the symbol and pattern are a bit better (they still use the transformation used but at least they respect the language specification).

Capture d’écran 2019-12-04 à 16 36 47

And if i use the 'Reset default settings' (on official currencies) the custom values are reset and the names are also correct.

**Solution**
All these behavior are logical from the form point of view, the issue is not on the form but on the Language installation. When a Language/Localization pack is installed the currency_lang should have correct values:
 - name: translated name of the currency (or ISO code for unofficial currency)
 - symbol: symbol from CLDR for the installed language
 - pattern: empty

**Additional information**
PrestaShop version: 1.7.7
PHP version: N/A

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16686
| How to test?  | With an modified **default** currency, try adding a new language in *International* -> *Translations*. The default currency should have the correct translated name/symbol for the new language.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17373)

### DIFF
--- a/classes/Language.php
+++ b/classes/Language.php
@@ -1258,6 +1258,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
         $lang_pack = self::getLangDetails($iso);
         self::installSfLanguagePack(self::getLocaleByIso($iso), $errors);
         self::updateMultilangTable($iso);
+        self::updateCurrenciesCldr(new Language(self::getIdByIso($iso, true)));
         self::generateEmailsLanguagePack($lang_pack, $errors, true);
 
         return count($errors) ? $errors : true;
@@ -1270,15 +1271,19 @@ class LanguageCore extends ObjectModel implements LanguageInterface
         $container = SymfonyContainer::getInstance();
         /** @var LocaleRepository $localeRepoCLDR */
         $localeRepoCLDR = $container->get('prestashop.core.localization.cldr.locale_repository');
+        $localeCLDR = $localeRepoCLDR->getLocale($language->locale);
 
         foreach ($currencies as $currency) {
             $names = $currency->getLocalizedNames();
             $symbols = $currency->getLocalizedSymbols();
 
-            $cldrCurrency = $localeRepoCLDR->getLocale($language->locale)->getCurrency($currency->iso_code);
+            $currencyCLDR = $localeCLDR->getCurrency($currency->iso_code);
+            if (null === $currencyCLDR) {
+                continue;
+            }
 
-            $names[$language->id] = $cldrCurrency->getDisplayName();
-            $symbols[$language->id] = $cldrCurrency->getSymbol();
+            $names[$language->id] = $currencyCLDR->getDisplayName();
+            $symbols[$language->id] = $currencyCLDR->getSymbol();
 
             $currency->setLocalizedNames($names);
             $currency->setLocalizedSymbols($symbols);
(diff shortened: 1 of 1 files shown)


matks (dismissed):

classes/Language.php
```diff
@@ -1262,6 +1263,30 @@ public static function installLanguagePack($iso, $params, &$errors = [])
...
+        $currencies = Currency::getCurrencies(true, false, false);
+        $container = SymfonyContainer::getInstance();
+        /** @var LocaleRepository $localeRepoCLDR */
+        $localeRepoCLDR = $container->get('prestashop.core.localization.cldr.locale_repository');
+
+        foreach ($currencies as $currency) {
+            $names = $currency->getLocalizedNames();
+            $symbols = $currency->getLocalizedSymbols();
+
+            $cldrCurrency = $localeRepoCLDR->getLocale($language->locale)->getCurrency($currency->iso_code);
```
Progi1984: You can extract this : `$localeRepoCLDR->getLocale($language->locale)` from the loop foreach.

Progi1984 (changes requested):
A small feeback

classes/Language.php
```diff
@@ -1262,6 +1263,30 @@ public static function installLanguagePack($iso, $params, &$errors = [])
...
 
+    private static function updateCurrenciesCldr(Language $language)
+    {
+        /** @var Currency[] $currencies */
+        $currencies = Currency::getCurrencies(true, false, false);
+        $container = SymfonyContainer::getInstance();
+        /** @var LocaleRepository $localeRepoCLDR */
+        $localeRepoCLDR = $container->get('prestashop.core.localization.cldr.locale_repository');
+
+        foreach ($currencies as $currency) {
```
jolelievre, line 1276: You are recoding a part of what is present in `Currency::refreshLocalizedCurrencyData`, the problem is that the code is duplicated/needs to be modified at several places And you are forgetting some part of (what if the cldr currency is not available  - custom currencies)
You'd better use the existing method with a languages array that will only contain one language (even though it's not optimized it will reduce the risk of splitting the behaviour in case of a feature update)

jolelievre (changes requested):

classes/Language.php
atomiix: Done

Progi1984 (approved):

PierreRambaud (approved):

jolelievre (approved):
Nice work @atomiix

matks:
Thank you @atomiix

PR #17385, opened 2020-01-28
Confirm modal on restoring currency settings

### LINKED ISSUE #17024: Add a message when symbol and format are reseted in currencies
In international -> localizations -> currencies 
Add a message when we reset symbol and format. It's not very obvious that the symbol and format is reseted.

sowbiba:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Add confirmation modal when user click on the "Restore default settings" CTA. Add title and help before format by language settings
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17024
| How to test?  | -

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17385)


admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
```diff
@@ -138,6 +139,23 @@ export default class CurrencyForm {
     }
   }
 
+  _showResetDefaultSettingsConfirmModal() {
+    const confirmTitle = this.translations['confirmTitle'];
+    const confirmMessage = this.translations['confirmMessage'];
+    const confirmButtonLabel = this.translations['confirmButtonLabel'];
+    const closeButtonLabel = this.translations['closeButtonLabel'];
+
+    const modal = new ConfirmModal({
```
matks, line 153: @NeOMakinG looks like the modal you built is nicely reused 👍

matks (dismissed):

sarahdib:
@sowbiba 
When I click in the restore option 
Capture d’écran 2020-01-28 à 16 20 38

there isn't any modal

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -42,7 +42,11 @@
         'list.example': 'Example'|trans({}, 'Admin.Global'),
         'list.edit': 'Edit symbol / format'|trans({}, 'Admin.International.Feature'),
         'list.reset': 'Reset settings'|trans({}, 'Admin.Actions'),
-        'list.example.format': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature')
+        'list.example.format': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature'),
+        'confirmTitle': 'Restore currencies default settings'|trans({}, 'Admin.International.Feature'),
+        'confirmButtonLabel': 'Restore'|trans({}, 'Admin.Actions'),
+        'closeButtonLabel': 'Cancel'|trans({}, 'Admin.Actions'),
+        'confirmMessage': 'Are you sure you want to restore default settings?'|trans({}, 'Admin.Actions')
```
LouiseBonnard:
```suggestion
        'confirmMessage': 'Are you sure you want to restore default settings?'|trans({}, 'Admin.International.Feature')
```

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -42,7 +42,11 @@
         'list.example': 'Example'|trans({}, 'Admin.Global'),
         'list.edit': 'Edit symbol / format'|trans({}, 'Admin.International.Feature'),
         'list.reset': 'Reset settings'|trans({}, 'Admin.Actions'),
-        'list.example.format': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature')
+        'list.example.format': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature'),
+        'confirmTitle': 'Restore currencies default settings'|trans({}, 'Admin.International.Feature'),
```
LouiseBonnard: _Restore default settings_ instead?

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -42,7 +42,11 @@
         'list.example': 'Example'|trans({}, 'Admin.Global'),
         'list.edit': 'Edit symbol / format'|trans({}, 'Admin.International.Feature'),
         'list.reset': 'Reset settings'|trans({}, 'Admin.Actions'),
```
LouiseBonnard: Isn't this wording supposed to be deleted?

sowbiba: It's the title for the column of the grid

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
sowbiba: done

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
sowbiba: done

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -46,6 +47,7 @@
 <script>
   import LanguageList from './LanguageList';
   import CurrencyModal from './CurrencyModal';
+  import ConfirmModal from '@components/modal';
```
jolelievre: You are mixing a jquery modal within this vue component This should be avoided at all cost, there is a modal vue component that you should use instead:
`import Modal from '@vue/components/Modal';`

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -28,8 +28,9 @@
       <language-list
         v-if="languagesCount"
         :languages="languages"
+        :translations="translations"
         @selectLanguage="selectLanguage"
-        @resetLanguage="resetLanguage"
+        @confirmResetLanguage="confirmResetLanguage"
```
jolelievre: You don't need to change the event name in `LanguageList` The action requested by the list is still `resetLanguage` the fact that there is a confirmation is the responsibility of this component, so the callback name however makes sense

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -28,8 +28,9 @@
       <language-list
         v-if="languagesCount"
         :languages="languages"
+        :translations="translations"
```
jolelievre: You're not supposed to inject the translations, Vue integrate a mixin to deal with translations thanks to the `this.$t()` function

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -76,6 +82,22 @@
       selectLanguage(language) {
         this.selectedLanguage = language;
       },
+      confirmResetLanguage(language) {
+        const confirmTitle = this.translations['confirmTitle'];
```
jolelievre: With the vue modal you should be able to get rid of this translation object

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
jolelievre: Oh one thing that could be interesting is to create a `ConfirmModal` vue component (in the `vue/component` folder) It could be based on the already existing modal component and allows a few customization (like the jquery one):
- title
- button labels
- modal description content
- confirm/cancel callbacks

The good thing is that thanks to the slots and properties you can have default behaviour that are easily overridable. This is the kind of component we will definitely need in the future so I think it's an interesting topic :)

admin-dev/themes/new-theme/js/pages/currency/form/components/LanguageList.vue
```diff
@@ -64,7 +64,7 @@
       <td>
         <div class="btn-group-action text-right">
           <div class="btn-group">
-            <button type="button" class="btn" @click.prevent.stop="$emit('resetLanguage', language)">
+            <button type="button" class="btn" @click.prevent.stop="$emit('confirmResetLanguage', language)">
```
jolelievre: As I said no need to rename the event here

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
jolelievre: This part being in jquery however it makes total sens to use this component here :)

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -42,7 +42,11 @@
...
-        'list.example.format': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature')
+        'list.example.format': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature'),
+        'confirmTitle': 'Restore default settings'|trans({}, 'Admin.International.Feature'),
```
jolelievre: Please use the same format for translation keys, something like `modal.confirm.title`, `modal.confirm.restore`, `modal.confirm.message`, ...
And `closeButtonLabel` is redundant with the already present `modal.cancel`

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
jolelievre: Maybe we could have something more related to the component like `Remove customizations`

jolelievre (changes requested):
Nice work @sowbiba 
I have a few feedback though, especially avoid mixing vue and jquery components

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
LouiseBonnard: I guess we should turn it into _Restore settings_ to stay consistent?

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -39,17 +39,34 @@
         @applyCustomization="applyCustomization"
       >
       </currency-modal>
+
+      <modal
+        confirmation
+        v-if="showConfirmationModal"
+        @confirm="applyResetLanguage"
+        @close="closeConfirmationModal"
+        :modalTitle="modalTitle"
```
PierreRambaud:
```suggestion
        :modalTitle="$t('modal.confirm.title')"
```
No need to use computed here, it's wasting memory

jolelievre: Oh right, then I guess it's the same for all other computed fields `cancelLabel`, `confirmLabel`, ...

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
```diff
@@ -64,26 +81,47 @@
...
+      },
     }, methods: {
       closeModal() {
         this.selectedLanguage = null;
       },
       selectLanguage(language) {
         this.selectedLanguage = language;
       },
+      closeConfirmationModal() {
+        this.showConfirmationModal = false;
```
jolelievre: You can simplify this by removing the boolean, and use `this.showConfirmationModal` as your condition, in the template simply use `v-if="confirmationLanguage !== null"`

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
jolelievre: But the action is a bit different, the button 'Restore settings' reset all the data for the currency (symbols, names, patterns, exchange rate...) for ALL languages Whereas the button in the list is only used for ONE language, and only reset symbol and pattern (the two customizable fields) Which is why, although it's in a list and you can guess the context, maybe using different wording would emphasize that the two actions are different

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -42,7 +42,11 @@
         'list.example': 'Example'|trans({}, 'Admin.Global'),
         'list.edit': 'Edit symbol / format'|trans({}, 'Admin.International.Feature'),
         'list.reset': 'Reset settings'|trans({}, 'Admin.Actions'),
-        'list.example.format': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature')
+        'list.example.format': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature'),
+        'modal.confirm.title': 'Are you sure you want to restore default settings?'|trans({}, 'Admin.International.Feature'),
```
jolelievre: On second thought, maybe `modal.confirm.*` is a bit generic, especially for `modal.confirm.body` Maybe we should use something a bit more related to context, in case some day another modal is added Like `modal.restore.title` and `modal.restore.body` As for apply and cancel since they are quite generic it doesn't shock me that much

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
PierreRambaud: Yes :)

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
LouiseBonnard: Got it. But the symbol is the same, it is not ideal. I'm okay to figure out another wording for the listing action but I'm afraid it will be quite confusing for the users if they have to click the same icon. @TristanLDD, @prestascott, any recommendation?

cf. the initial mockup [here](https://projects.invisionapp.com/d/main?origin=v7#/console/11916080/386388367/preview?scrollOffset=1416) for more context

TristanLDD: In my opinion, we should get rid of the icon on the reset / restore button to avoid any confusion (in the case we change wording)

LouiseBonnard: Alright, let's do that: 

- main CTA does not display the icon and its wording remains _Restore default setting_

- when in the second section, 'Format Settings', we keep the icon at the end of each currency line and the wording is _Reset_ (Admin.Actions)

We must keep it short, and since the previous column talks about editing both the symbol and the format, users integrate that the following reset action is about the two same fields.

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
sowbiba: done

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
sowbiba: done

admin-dev/themes/new-theme/js/pages/currency/form/components/CurrencyFormatter.vue
sowbiba: Done. thanks for the tip

atomiix (dismissed):
LGTM !

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
```diff
@@ -102,7 +103,7 @@ export default class CurrencyForm {
   _initListeners() {
     this.$currencySelector.change(this._onCurrencySelectorChange.bind(this));
     this.$isUnofficialCheckbox.change(this._onIsUnofficialCheckboxChange.bind(this));
-    this.$resetDefaultSettingsButton.click(this._onResetDefaultSettingsClick.bind(this));
+    this.$resetDefaultSettingsButton.click(this._showResetDefaultSettingsConfirmModal.bind(this));
```
Progi1984:
```suggestion
    this.$resetDefaultSettingsButton.click(this.showResetDefaultSettingsConfirmModal.bind(this));
```

admin-dev/themes/new-theme/js/pages/currency/form/currency-form.js
```diff
@@ -144,6 +145,23 @@ export default class CurrencyForm {
     }
   }
 
+  _showResetDefaultSettingsConfirmModal() {
```
Progi1984:
```suggestion
  showResetDefaultSettingsConfirmModal() {
```

Progi1984 (commented):
Small change

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -41,8 +41,12 @@
...
         'list.example': 'Example'|trans({}, 'Admin.Global'),
         'list.edit': 'Edit symbol / format'|trans({}, 'Admin.International.Feature'),
-        'list.reset': 'Reset settings'|trans({}, 'Admin.Actions'),
-        'list.example.format': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature')
+        'list.reset': 'Reset'|trans({}, 'Admin.Actions'),
+        'list.example.format': 'Price: %price% (Special discount %discount%)'|trans({}, 'Admin.International.Feature'),
+        'modal.restore.title': 'Are you sure you want to restore default settings?'|trans({}, 'Admin.International.Feature'),
+        'modal.restore.apply': 'Restore'|trans({}, 'Admin.Actions'),
+        'modal.restore.cancel': 'Cancel'|trans({}, 'Admin.Actions'),
+        'modal.restore.body': 'Restoring your currencies’ default settings will delete all the customizations you made before. Parameters will look just the same as when a currency is freshly imported.'|trans({}, 'Admin.International.Feature')
```
jolelievre: The same body message is used for the general restore button (which restore default settings AND customizations for all the languages) and the specific one (which restores ONLY customization for ONE language)
The title can be the same it doesn't shock me, however I think the body message should be adapted for those two use cases to avoid misunderstanding.
@LouiseBonnard do you have an idea what messages we could use to differentiate both use cases?

jolelievre (changes requested):
The same message is used for both restore actions
Despite that, the rest seems good to me ^^

MatShir:
Hi @sowbiba,
Sorry for the misunderstanding, there is some wording to adjust and some button to move.
- So the 'Restore default settings' button should go on the button left of the form. 
- At the end of the page, the cancel button will go next to the save button. 
- The name of the form header should be "Edit {name of the currency}" instead of "Currency".
- The 'Reset' (and not 'Reset settings') button on the list will **Not** display a modal but **confirmation notification.** 
- On the format settings list, the example should be **only** the currency and the price **without**
 "Price:" and "(Special Discount: -price)"

@LouiseBonnard will provide the right wording for the confirmation notification.

---

EDIT by @LouiseBonnard 

**Regarding the modal, here is the expected wording:**

Are you sure you want to restore default settings?

Restoring your currency's default settings will delete all the customizations you made before. Parameters will look just the same as when the currency is freshly imported.

Cancel + Restore (in _Admin.Actions_)

**Regarding the title and subtitle, it disappears.**

~~Format settings~~

~~You might need to adapt the predefined format display of your currencies, or edit their symbol. Note that restoring default settings will erase symbol and format customizations~~

**Regarding the list and the success notification displayed after the reset action:**

Write 'Reset' instead of 'Reset settings' as the last column of the list

Your symbol and format customizations have been successfully reset for this language. (in _Admin.Notifications.Success_)

Progi1984 (dismissed):

Progi1984 (dismissed):

src/PrestaShopBundle/Controller/Admin/Improve/International/CurrencyController.php
```diff
@@ -153,6 +153,7 @@ public function createAction(Request $request)
         return $this->render('@PrestaShop/Admin/Improve/International/Currency/create.html.twig', [
             'isShopFeatureEnabled' => $multiStoreFeature->isUsed(),
             'currencyForm' => $currencyForm->createView(),
+            'currencyName' => null !== $currencyForm ? array_values($currencyForm->getData()['names'])[0] : null,
```
jolelievre: I'm not sure it is really useful for the creation form You can't know in advance which currency will be created so in the data will be empty Maybe we should have a text like `Create new currency` or something similar
@LouiseBonnard what do you think?

jolelievre (commented):
Small comment about the creation page title, besides this it's ok!

Be careful when you commit the assets (even though it soon won't be needed any more) to only commit the modified assets This avoids conflicts with features you didn't change and it makes PR smaller (you even pushed the built assets of the front)

Oh and you have a PrettyCI error related to `OrderRefundFeatureContext` but it's weird because you didn't modify it Maybe a rebase problem?

src/PrestaShopBundle/Controller/Admin/Improve/International/CurrencyController.php
LouiseBonnard: I suggest either _Currency_ or _Name_. _Name_ is the most consistent option since most of the creation forms in the software start with this wording.

sowbiba: I dont really get your point @LouiseBonnard

sowbiba: Decision have been made to keep *Currencies* for creation and edit forms

Done

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Currency/Blocks/form.html.twig
```diff
@@ -41,15 +41,20 @@
...
+        'modal.restore.cancel': 'Cancel'|trans({}, 'Admin.Actions'),
+        'modal.restore.body': 'Restoring your currency\'s default settings will delete all the customizations you made before. Parameters will look just the same as when the currency is freshly imported.'|trans({}, 'Admin.International.Feature')
       }|json_encode|replace("'", "&#39;")|raw
     }
   }) }}
 
   <div class="card">
     <div class="card-header">
-      {{ 'Currency'|trans({}, 'Admin.Global') }}
+      {{ 'Currencies'|trans({}, 'Admin.Global') }}
```
jolelievre, line 57: Ok for keeping a generic name but why plural since the form is dedicated to only one currency?

LouiseBonnard: It is pure consistency; it is the case for all the localization pages of PrestaShop, when you click the 'Add new something' button in the top right-hand corner, see examples below.

Capture d’écran 2020-02-19 à 18 18 53

Capture d’écran 2020-02-19 à 18 19 09

Capture d’écran 2020-02-19 à 18 19 19

sowbiba:
@jolelievre looks good for you ?

NeOMakinG (dismissed):
LGTM

matthieu-rolland (dismissed):

matthieu-rolland:
@sowbiba  I aproved this PR, but I think we should wait for @jolelievre 's requested change to be validated by him before sending it to QA 👍

jolelievre (dismissed):
Nice work @sowbiba

PierreRambaud (dismissed):

boubkerbribri:
@PrestaShop/prestashop-core-developers please resolve conflicts before we test it

NeOMakinG:
Hello @boubkerbribri as we talked with the team today, it's assets conflict, you can test it without a rebase, after your approve, @sowbiba will be able to rebase, approve and merge :)

Robin-Fischer-PS:
Hi @sowbiba !

Since now there is also conflict on  themes/core.js , I need the conflicts to be resolved before testing.

Thanks !

SD1982:
Hi @sowbiba it seems that this PR need some reviews before QA can treat it.

jolelievre (approved):

NeOMakinG (approved):

PierreRambaud (approved):

matks:
Thank you @sowbiba

PR #12928, opened 2019-03-14
Allow multiple grids on same page

jolelievre:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | The grid system, and more specifically the automatic SearchParametersResolver don't allow to have two grids on the same controller This PR changes the system so that the Filter classes can use their own "scope" of parameters The resolver is able to populate the different filters appropriately
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #10820 #11082
| How to test?  | The page "Brands & Suppliers Brands" has 2 listings on the same page (URL: `/admin-dev/index.php/sell/catalog/brands`). Try to filter, search and paginate on both listings. They must remain independent. As we modified the search and filter system globally, please test that another SF listing such as "International > Localization > Languages" page works as before.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/12928)

PR #17399, opened 2020-01-29
Expand Categories search capabilities for categories listing home page and avoid going from subcategory to root category when using grid filters

### LINKED ISSUE #15588: Subcategory not found with category id when searched for - redirection with the reset button also
In categories, when searching for a category based on category ID, name, description, position or displayed no results are found for subcategories.

**To Reproduce**
1. Go to BO => Catalog => Categories 
2. Click on "View button", Check that you have more than a category
3. Search for a subcategory by ID, name, description, position or displayed 
4. You are redirected to the Category list page

Prestashop 1.7.6
PHP 7.2

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | See below
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/15588
| How to test?  | See ticket is fixed and description below

#### TLDR

Short summary: I enabled "full search" on home Category page only, not subcategories. This fixes the issue when the search is done on Home Category page.

I also fixed a bug: avoid going from subcategory to root category when using grid filters. See https://github.com/PrestaShop/PrestaShop/pull/17399#issuecomment-581831954

#### What happened in the bug

In the current implementation of Categories grid listing, search only targets direct children of current category. So indeed subcategories could not be found.

#### Complex topic

If we allow full categories search (so we dont check for direct children but target all categories), crazy side-effects happen. For example you view category "Men" and you search "Home" and it appears, although Home is a parent category way higher than "Men".

#### Ideal behavior

The ideal behavior would be "when I am on listing, current category is A, if I search for "blablabla" the search will be done against all subcategories of A (either direct or deeper children). This requires complex tree computations that are prone to errors and long to do.

**So I built a workaround**

#### My workaround

I have enabled "full depth search" only on Home category listing page, which is likely to be the main place where such searches will happen. This removes the crazy side-effects on subcategories grid listings as it is not enabled on those pages.

# Note
:notebook: When this PR is merged, please notify @boubkerbribri as this will allow him to enable a test (the test is skipped today)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17399)

### DIFF
--- a/src/Core/Grid/Factory/CategoryGridFactoryDecorator.php
+++ b/src/Core/Grid/Factory/CategoryGridFactoryDecorator.php
@@ -76,7 +76,7 @@ final class CategoryGridFactoryDecorator implements GridFactoryInterface
         if ($this->isHomeCategory($filters) && ($this->filterFormFactory !== null)) {
             return $categoryGrid;
         }
-        
+
         $this->injectCategoryIdIntoSearchTypeOptions($categoryGrid, $filters);
 
         $filterForm = $this->rebuildFilterForm($searchCriteria, $categoryGrid);
--- a/src/Core/Grid/Factory/CategoryGridFactoryDecorator.php
+++ b/src/Core/Grid/Factory/CategoryGridFactoryDecorator.php
@@ -57,8 +57,8 @@ final class CategoryGridFactoryDecorator implements GridFactoryInterface
      */
     public function __construct(
         GridFactoryInterface $categoryGridFactory,
-        GridFilterFormFactoryInterface $filterFormFactory = null)
-    {
+        GridFilterFormFactoryInterface $filterFormFactory = null
+    ) {
         $this->categoryGridFactory = $categoryGridFactory;
         $this->filterFormFactory = $filterFormFactory;
     }
@@ -109,8 +109,8 @@ final class CategoryGridFactoryDecorator implements GridFactoryInterface
      */
     private function removePositionDragColumnIfEligible(
         SearchCriteriaInterface $searchCriteria,
-        $categoryGrid)
-    {
+        $categoryGrid
+    ) {
         if ('position' !== $searchCriteria->getOrderBy() ||
             'asc' !== $searchCriteria->getOrderWay()) {
             $categoryGrid->getDefinition()
--- a/src/Core/Grid/Factory/CategoryGridFactoryDecorator.php
+++ b/src/Core/Grid/Factory/CategoryGridFactoryDecorator.php
@@ -26,8 +26,13 @@
 
 namespace PrestaShop\PrestaShop\Core\Grid\Factory;
 
+use PrestaShop\PrestaShop\Core\Grid\Filter\FilterInterface;
+use PrestaShop\PrestaShop\Core\Grid\Filter\GridFilterFormFactoryInterface;
+use PrestaShop\PrestaShop\Core\Grid\Grid;
 use PrestaShop\PrestaShop\Core\Grid\GridFactoryInterface;
+use PrestaShop\PrestaShop\Core\Grid\GridInterface;
 use PrestaShop\PrestaShop\Core\Grid\Search\SearchCriteriaInterface;
+use Symfony\Component\Form\FormInterface;
 
 /**
  * Decorates Category grid factory
@@ -40,31 +45,108 @@ final class CategoryGridFactoryDecorator implements GridFactoryInterface
     private $categoryGridFactory;
 
     /**
-     * @param GridFactoryInterface $categoryGridFactory
+     * @var GridFilterFormFactoryInterface
      */
-    public function __construct(GridFactoryInterface $categoryGridFactory)
+    private $filterFormFactory;
+
+    /**
+     * @param GridFactoryInterface $categoryGridFactory
+     * @param GridFilterFormFactoryInterface $filterFormFactory optional
+     *
+     * $filterFormFactory is optional in order to comply with SemVer
+     */
+    public function __construct(
+        GridFactoryInterface $categoryGridFactory,
+        GridFilterFormFactoryInterface $filterFormFactory = null)
     {
         $this->categoryGridFactory = $categoryGridFactory;
+        $this->filterFormFactory = $filterFormFactory;
     }
 
     /**
-     * Position can only be changed when grid is
-     * ordered by "position" in "asc" way.
-     *
      * {@inheritdoc}
      */
     public function getGrid(SearchCriteriaInterface $searchCriteria)
     {
         $categoryGrid = $this->categoryGridFactory->getGrid($searchCriteria);
 
+        $this->removePositionDragColumnIfEligible($searchCriteria, $categoryGrid);
+
+        $filters = $searchCriteria->getFilters();
+        if ($this->isHomeCategory($filters) && ($this->filterFormFactory !== null)) {
+            return $categoryGrid;
+        }
+        
+        $this->injectCategoryIdIntoSearchTypeOptions($categoryGrid, $filters);
+
+        $filterForm = $this->rebuildFilterForm($searchCriteria, $categoryGrid);
+
+        return new Grid(
+            $categoryGrid->getDefinition(),
+            $categoryGrid->getData(),
+            $searchCriteria,
+            $filterForm
+        );
+    }
+
+    /**
+     * @param array $filters
+     *
+     * @return bool
+     */
+    private function isHomeCategory(array $filters)
+    {
+        return isset($filters['is_home_category'])
+            && $filters['is_home_category'] === true;
+    }
+
+    /**
+     * Position can only be changed when grid is
+     * ordered by "position" in "asc" way.
+     *
+     * @param SearchCriteriaInterface $searchCriteria
+     * @param GridInterface $categoryGrid
+     */
+    private function removePositionDragColumnIfEligible(
+        SearchCriteriaInterface $searchCriteria,
+        $categoryGrid)
+    {
         if ('position' !== $searchCriteria->getOrderBy() ||
             'asc' !== $searchCriteria->getOrderWay()) {
             $categoryGrid->getDefinition()
                 ->getColumns()
-                ->remove('position_drag')
-            ;
+                ->remove('position_drag');
         }
+    }
 
-        return $categoryGrid;
+    /**
+     * @param GridInterface $categoryGrid
+     * @param array $filters
+     */
+    private function injectCategoryIdIntoSearchTypeOptions($categoryGrid, array $filters)
+    {
+        /** @var FilterInterface $actionsFilter */
+        $actionsFilter = $categoryGrid->getDefinition()
+            ->getFilters()->get('actions');
+
+        $typeOptions = $actionsFilter->getTypeOptions();
+        $typeOptions['redirect_route_params'] = ['categoryId' => $filters['id_category_parent']];
+        $actionsFilter->setTypeOptions($typeOptions);
+    }
+
+    /**
+     * @param SearchCriteriaInterface $searchCriteria
+     * @param GridInterface $categoryGrid
+     *
+     * @return FormInterface
+     */
+    private function rebuildFilterForm(
+        SearchCriteriaInterface $searchCriteria,
+        GridInterface $categoryGrid)
+    {
+        $filterForm = $this->filterFormFactory->create($categoryGrid->getDefinition());
+        $filterForm->setData($searchCriteria->getFilters());
+
+        return $filterForm;
     }
 }
--- a/src/Core/Grid/Filter/FilterCollection.php
+++ b/src/Core/Grid/Filter/FilterCollection.php
@@ -61,7 +61,7 @@ final class FilterCollection implements FilterCollectionInterface
     /**
      * @param string $filterName
      *
-     * @return null|FilterInterface return null if no filter with given filter name
+     * @return FilterInterface|null return null if no filter with given filter name
      */
     public function get($filterName)
     {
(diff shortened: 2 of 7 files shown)


src/Core/Grid/Query/CategoryQueryBuilder.php
```diff
@@ -208,4 +212,15 @@ private function getQueryBuilder(array $filters)
...
     }
+
+    /**
+     * @param array $filters
+     *
+     * @return bool
+     */
+    private function isSearchRequestOnHomeCategory(array $filters)
+    {
+        return (array_key_exists('is_home_category', $filters) && ($filters['is_home_category'] === true))
```
PierreRambaud: Isset is really faster than array_key_exists (and because we don't care about the null value)

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
```diff
@@ -823,4 +828,22 @@ private function getBulkCategoriesFromRequest(Request $request)
...
+     *
+     * @return bool
+     */
+    private function requestHasSearchParameters(Request $request)
+    {
+        if (!$request->query->has('category')) {
+            return false;
+        }
+
+        if (!array_key_exists('filters', $request->query->get('category'))) {
```
PierreRambaud: Unnecessary conditions, the `!empty` is already doing theses checks

```suggestion
    private function requestHasSearchParameters(Request $request)
    {
        return !empty($request->query->get('category')['filters']);
    }
```

is enough

src/Core/Grid/Query/CategoryQueryBuilder.php
Progi1984: Please use `isset` rather than `array_key_exists`

src/Core/Search/Filters.php
```diff
@@ -101,6 +101,16 @@ public function getLimit()
         return $this->get('limit');
     }
 
+    /**
+     * @param array $parameters
+     */
+    public function addFilter(array $parameters = [])
+    {
+        $filters = $this->getFilters();
+        $filters = array_replace($filters, $parameters);
```
Progi1984:
```suggestion
        $filters = array_replace($this->getFilters(), $parameters);
```

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
Progi1984: Please use `isset` rather than `array_key_exists`

Progi1984 (commented):
Some feedback

matks:
Code review feedbacks handled

src/Core/Grid/Query/CategoryQueryBuilder.php
```diff
@@ -208,4 +212,15 @@ private function getQueryBuilder(array $filters)
...
+
+    /**
+     * @param array $filters
+     *
+     * @return bool
+     */
+    private function isSearchRequestOnHomeCategory(array $filters)
+    {
+        return (isset($filters['is_home_category']) && ($filters['is_home_category'] === true))
+            && (isset($filters['is_search_request']) && ($filters['is_search_request'] === true));
```
Progi1984:
```suggestion
        return isset($filters['is_home_category'], $filters['is_search_request'])
             && $filters['is_home_category'] === true && $filters['is_search_request'] === true;
```

Progi1984 (commented):
A last feedback

src/Core/Grid/Query/CategoryQueryBuilder.php
matks: Please tell me if I'm wrong, but I think this last feedback is not an issue in the PR but rather a personal preference 😉 and hence we should not block PRs for that. The goal of review is to validate the quality of a PR, not that it is written exactly as you would have written it yourself 😉

Progi1984: It's rather an optimization : check if all exists before tests. In addition, that removes some code, and less code, less bugs.

PierreRambaud (dismissed):

src/Core/Grid/Query/CategoryQueryBuilder.php
matks: You are killing me 🙄

matks: Done

Progi1984 (dismissed):

matks:
When this PR is merged, please notify @boubkerbribri as this will allow him to enable a test (the test is skipped today)

PierreRambaud (dismissed):

sarahdib:
@matks 
If I search access in Home it's ok
If I search access in the category Accessoire or vêtement I found accessoire and accessoire de maison

matks:
OK, I found the issue. When you are on "Accessoires" category, if you search "Vetements" **you are directed to home page THEN the search is performed**. And since the home page search looks in all categories, it looks like a bug.

You can see it by looking at the URL, you move from `/admin-dev/index.php/sell/catalog/categories/6` to `/admin-dev/index.php/sell/catalog/categories`

The real bug is that when you are on "Accessoires" category (= subcategory), if you search "Vetements" you are directed to home page then the search is performed. We should stay on the subcategory page. I'll fix it.

matks:
Ready for review !

I fixed this:
> When you are on "Accessoires" category, if you search "Vetements" **you are directed to home page THEN the search is performed**. And since the home page search looks in all categories, it looks like a bug.
> 
> You can see it by looking at the URL, you move from `/admin-dev/index.php/sell/catalog/categories/6` to `/admin-dev/index.php/sell/catalog/categories`
> 
> The real bug is that when you are on "Accessoires" category (= subcategory), if you search "Vetements" you are directed to home page then the search is performed. We should stay on the subcategory page. I'll fix it.

PierreRambaud (dismissed):

src/Core/Grid/Filter/FilterCollection.php
```diff
@@ -58,6 +58,20 @@ public function remove($filterName)
...
+     *
+     * @return FilterInterface|null return null if no filter with given filter name
+     */
+    public function get($filterName)
+    {
+        if (isset($this->filters[$filterName])) {
+            return $this->filters[$filterName];
+        }
+
+        return null;
```
Progi1984, line 72:
```suggestion
        return $this->filters[$filterName] ?? null;
```

Progi1984 (commented):
A feedback

matthieu-rolland (dismissed):

src/Core/Grid/Filter/FilterCollection.php
matks: The null coalescing operator is a new PHP short operator that kindof serves the same purpose that the ternary operator, which is to avoid heavy `if/else` statements.

However the use of these short operators has been debated for a long time (and will probably continue for years) because it might hurt the code readability.

Stackoverflow is full of fiery debates about whether or not to use ternary operators, with both sides having both passionate and logical arguments 😄 (see [this one](https://stackoverflow.com/questions/694814/ternary-operator-bad-or-good-practice) or [this one](https://softwareengineering.stackexchange.com/questions/28314/ternary-operator-considered-harmful)).

I personally side with the people suggesting not to use it, except for very obvious usecases. I like [this position](https://pressupinc.com/blog/2015/03/ternary-operators-considered-harmful/): _"If conciseness is more valuable, I should use a single-line ternary operation. If clarity is more valuable, I should use the English keywords of if and else."_

Considering that PrestaShop is a wide and complex project which can be used by both junior and senior developers, I suggest we keep the if/else structure as it's the most common and widely known, and we keep the maximum readability (with the cost of a few more lines indeed) 🤔

boubkerbribri:
With this categories tree 
**Home
-> Clothes
 -> -> Men
-> -> -> t-shirts
-> -> -> shoes
-> -> Women
-> Art**

When I search for `Clothes` or `Art`: 

- In home categories page, the category is displayed => **OK**, wanted behavior
- In a subcategory page, no result found => **OK**, wanted behavior

When I search for `Men` : 
- In home category, the category is displayed => **OK**, wanted behavior
- In subcategory page `Clothes`, the category is displayed => **OK**, wanted behavior
- In subcategory page `Women`, no result found => **OK**, wanted behavior

When I search for `t-shirts`:
- In home category, the category is displayed => **OK**, wanted behavior
- In subcategory page `Clothes`,  no result found => **OK**, wanted behavior
- In subcategory page `Men`, the category is displayed => **OK**, wanted behavior

The only behavior to Maybe fix is : 
When I reset filter in a subcategory page, it redirect me to home Categories page, @matks WDYT ?
1.
image
2. 
image

matks:
> When I reset filter in a subcategory page, it redirect me to home Categories page, @matks WDYT ?

Nice catch, will try to fix it as well

matks:
Added https://github.com/PrestaShop/PrestaShop/pull/17399/commits/49898f96b4d204d1efc7f487c4a0758ff460681e to follow @boubkerbribri request.

Please review it, it was actually quite hard to do 😄

Progi1984:
@matks Could you fix PHPCS fixer ? Thanks :)

Progi1984 (dismissed):

matks:
> @matks Could you fix PHPCS fixer ? Thanks :)

Done, thanks

Progi1984 (dismissed):

src/Core/Grid/Factory/CategoryGridFactoryDecorator.php
```diff
@@ -39,32 +44,109 @@ final class CategoryGridFactoryDecorator implements GridFactoryInterface
...
+     * Position can only be changed when grid is
+     * ordered by "position" in "asc" way.
+     *
+     * @param SearchCriteriaInterface $searchCriteria
+     * @param GridInterface $categoryGrid
+     */
+    private function removePositionDragColumnIfEligible(
+        SearchCriteriaInterface $searchCriteria,
+        $categoryGrid)
+    {
```
PierreRambaud:
```suggestion
        $categoryGrid
    ) {
```

src/Core/Grid/Factory/CategoryGridFactoryDecorator.php
```diff
@@ -39,32 +44,109 @@ final class CategoryGridFactoryDecorator implements GridFactoryInterface
...
      * @param GridFactoryInterface $categoryGridFactory
+     * @param GridFilterFormFactoryInterface $filterFormFactory optional
+     *
+     * $filterFormFactory is optional in order to comply with SemVer
      */
-    public function __construct(GridFactoryInterface $categoryGridFactory)
+    public function __construct(
+        GridFactoryInterface $categoryGridFactory,
+        GridFilterFormFactoryInterface $filterFormFactory = null)
     {
```
PierreRambaud:
```suggestion
        GridFilterFormFactoryInterface $filterFormFactory = null
    ) {
```

PierreRambaud (dismissed):

Progi1984 (dismissed):

sarahdib:
@matks 
There is a weird behaviour

1. Go to category "accessoires" & search in the description field "maison" you have two results -> It's OK
2. Click on the previous button of the browser you have one result -> Not OK
3. Reset and search in the description field "maison" in Home category you have three result -> It's Ok

See the screen record : 
https://drive.google.com/file/d/1bmiG8cfDH2KvKc5IbtnG6WcJsTNMittm/view

Progi1984:
Ping @matks

matks:
> Ping @matks

I have no time for this PR right now 😞

If you want to make it go forward, be my guest 👍 

Else I will pick it up later when I find some time 🙏

Progi1984:
@matks Ok. I'll take it ;)

matks:
> @matks Ok. I'll take it ;)

## You are the true MVP



(this is a high five)

Progi1984:
> @matks
> There is a weird behaviour
> 
>     1. Go to category "accessoires" & search in the description field "maison" you have two results -> It's OK
> 
>     2. Click on the previous button of the browser you have one result -> Not OK
> 
>     3. Reset and search in the description field "maison" in Home category you have three result -> It's Ok
> 
> 
> See the screen record :
> https://drive.google.com/file/d/1bmiG8cfDH2KvKc5IbtnG6WcJsTNMittm/view

@Sarah : it seems that it's a behavior which has been in grid since one year : the persistence of search (#12928 since 1.7.6) by @jolelievre 

Make a search on Grid : 
- Category or Brand : the search applies
- Try to come back on the page from the menu : it doesn't reset the filter

Do you want I create a new issue ?

Progi1984 (approved):

matthieu-rolland (approved):

sarahdib:
@Progi1984 I create the issue for the last bug. 

Everything is ok for me

Progi1984:
Thanks @matks & @sarahdib

src/Core/Grid/Query/CategoryQueryBuilder.php
```diff
@@ -195,6 +195,10 @@ private function getQueryBuilder(array $filters)
             }
 
             if ('id_category_parent' === $filterName) {
+                if ($this->isSearchRequestOnHomeCategory($filters)) {
```
arouiadib, line 198: Thanks for the efforts in this complex matter. 
This `if` check in line 198 is what creates the problem raised here https://github.com/PrestaShop/PrestaShop/issues/20067
The system assumes that, if parent ID is sent in query string  (i.e key `id_category_parent` exists in `$filters` array), then automatically the merchant is not in home category.  So checking again if he is in Home just to skip the "WHERE" clause in query builder will put inconsistency in database results.
The "WHERE id_parent= X" is necessary whenever the `$filters['id_category_parent']` is set.
I am really interested in starting a PR about this complex matter, but I wanted to discuss with you about this choice before making first commit. Is there any specific reason for creating method `isSearchRequestOnHomeCategory`?

Issue #16654, opened 2019-12-02
[ORDER PAGE] Return products


MatShir:
**User story**

As a merchant, I want to restock or refund when my customer returns his orders.

**Specs**

Return product only **appears when the merchandise returns are activated and the order is considered as shipped**.

When the user presses the button a column "Refund" is added with a checkbox and the input quantity.

A quantity must be selected to refund a product. 
A product can be return if it has been re-stock before by another process has standard, cancel or partial refund button.
A product can't be refund multiple times. 

Before refunding the products, the user can choose to **generate a credit slip** and/or **voucher** or/and **re-stock** and **repay the shipping costs** (option appears if you select generate a voucher or credit slip).

![before return product](https://user-images.githubusercontent.com/43613217/69962213-cd385280-150d-11ea-8eca-1e245e5e408f.png)

After returning the products, it shows on a column "Return" and "Refund" with the return quantity and the price of the refund per products if a generated slip was checked in the option.

![after return product](https://user-images.githubusercontent.com/43613217/69962570-9878cb00-150e-11ea-89a0-85d97a681244.png)

MatShir:
One PR might work for this issue as well #16653, to study!

PR #17149, opened 2020-01-13
Cancel product

matthieu-rolland:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | Cancel product feature on migrated order detail page.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | Does it break backward compatibility? yes/no
| Deprecations? | Does it deprecate an existing feature? yes/no
| Fixed ticket? | #16650
| How to test?  | Make an order in frontend, then go the order page in BO, click on "cancel product" at the top of the page, then select one or more product and save. New Order page URL:  **/admin-dev/sell/orders/orders/8/view** . (where 8 is the order ID)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17149)

PR #17416, opened 2020-01-29
Migrate standard refund and return products in order page

### LINKED ISSUE #16653: [ORDER PAGE] Standard Refund button
**User story**

As a merchant, I want to refund my orders to my customer before there are shipped in the case for order mistaken.

**Specs**

Standard refund only **appears when the merchandise returns are activated and the order is considered as paid** and **disappears when the order is considered as shipped.**

When the user presses the button a column "Refund" is added with a checkbox and the input quantity.

A quantity must be selected to refund a product. 

Before refunding the products, the user can choose to **generate a credit slip** and/or **voucher** and **repay the shipping costs** (option appears if you select generate a voucher or credit slip)

before standard refund

When all the products are refunded order status change to Refunded.
A product can't be refund multiple times. 

After refunding, it shows on a column "Refunded", the quantity and the price of the refund per products.

after standard refund

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Manage standard refund and return product in migrated order page, add a new Command/Handler couple, the form is shared with partial refund but the dedicated controller has its own FormDataHandler so it ignores the data not relevant for it, in front the useless inputs are hidden and the form action is updated Plus add a bunch of behat tests
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16653 and #16654
| How to test?  | You need to enable Merchandise return to have access to this feature (in Customer services section), the button for this standard refund will only appear if the order has an invoice AND has not been delivered yet The button for return products only appears when order has a delivery slip (status Processing in progress)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17415)


matks (dismissed):

PierreRambaud (dismissed):

matks (dismissed):

PierreRambaud (dismissed):

Robin-Fischer-PS:
Hi there,

Some returns : 

On Standard Refund : 
- According to the spec, there is a behaviour missing : 
"When all the products are refunded order status change to Refunded."
=> This behaviour should probably also occurs on Return Product, Cancel product and Partial Refund. Is the "Shipping" taken into account for the change of Status ?

On all refund : 
- If we click on modify a product in the order and then partial or standard refund there is a mix between modify and refund
Capture d’écran 2020-01-31 à 14 44 22


Suggested behaviour (on all refund/return) : 
- When checking the product "Refund" checkbox, if there is already a number filled by the merchant in the "Quantity" field, it should not be overwritten by the max Quantity.

On Partial Refund : 
- There is more than 2 digits on product price
- When changing quantity of a product from 2 to 1, the "Amount tax" is updated, but when changing again from 1 to 2, it is not. It should.
- After a standard refund/Retunr product of only some of the product on a product line with Quantity > 1, the info "Max €XX.XX tax included/excluded" on Partial refund is false (for example, I have 3 Mug Today, I Standard refunded 1, but the Max is still Product price*3 instead of Product price*2
pr refund killmenowplease


Some graphical bugs :
- The "0 (€0.00 Refunded)" should be aligned with others info (Available, Quantity...)
- Quantity should be aligned with others info
- Amount (Tax included) in Partial Refund is a bit "messy".
- In the "Shipping" info below, when the partial refund block is displayed, the Shipping amount is not aligned.
- The Refund buttons are not aligned with other buttons.

MatShir:
> Hi there,
> 
> Some returns :
> 
> On Standard Refund :
> 
> * According to the spec, there is a behaviour missing :
>   "When all the products are refunded order status change to Refunded."
>   => This behaviour should probably also occurs on Return Product, Cancel product and Partial Refund. Is the "Shipping" taken into account for the change of Status ?

~~We don't care about the shipping cost and check the quantity refunded and if the total refunded is not equal it will a be partial refund status~~
No automation, the modules can change status after refund, so we will let the responsibility in the module side.

> On all refund :
> 
> * If we click on modify a product in the order and then partial or standard refund there is a mix between modify and refund
>   Capture d’écran 2020-01-31 à 14 44 22
> 
> Suggested behaviour (on all refund/return) :
> 
> * When checking the product "Refund" checkbox, if there is already a number filled by the merchant in the "Quantity" field, it should not be overwritten by the max Quantity.

Replace only when the value is 0. If unselect value is 0
> 
> On Partial Refund :
> 
> * There is more than 2 digits on the product price
Need it for computing calculation

**Be able to refund only the shipping cost**

> * When changing quantity of a product from 2 to 1, the "Amount tax" is updated, but when changing again from 1 to 2, it is not. It should.
> * After a standard refund/Return product of only some of the product on a product line with Quantity > 1, the info "Max €XX.XX tax included/excluded" on Partial refund is false (for example, I have 3 Mug Today, I Standard refunded 1, but the Max is still Product price_3 instead of Product price_2
>   pr refund killmenowplease
> 
> Some graphical bugs :
> 
> * The "0 (€0.00 Refunded)" should be aligned with others info (Available, Quantity...)
> * Quantity should be aligned with others info
> * Amount (Tax included) in Partial Refund is a bit "messy".
> * In the "Shipping" info below, when the partial refund block is displayed, the Shipping amount is not aligned.
> * The Refund buttons are not aligned with other buttons.

matthieu-rolland (dismissed):
last 4 commits LGTM 👍

atomiix (dismissed):
LGTM !

src/Adapter/Order/CommandHandler/IssueStandardRefundHandler.php
```diff
@@ -0,0 +1,154 @@
+<?php
```
Progi1984, line 1: On new files, please don't forget  `declare(strict_types=1);`

tests/Integration/Behaviour/Features/Context/Domain/OrderRefundFeatureContext.php
```diff
@@ -27,22 +27,35 @@
...
 class OrderRefundFeatureContext extends AbstractDomainFeatureContext
 {
+    /**
+     * @BeforeScenario
+     */
+    public function before()
+    {
+        // Merchandise return is disabled by default, use enabledReturnProduct() to enable it
+        Configuration::set('PS_ORDER_RETURN', 0);
+    }
```
Progi1984, line 57: Is there an `after` for reset the configuration to its default value ?

Progi1984 (commented):
Small feedback

tests/Integration/Behaviour/Features/Context/Domain/OrderRefundFeatureContext.php
jolelievre: Actually it should be 0 already with the default sql dump But I change this value with a Behat statement to enable it, and in following scenarii it was not reset to 0 I don't understand exactly when the dump is reinserted in the database..
This might be worth resetting on an after indeed, for now it doesn't create side effect on the behat suite so maybe we can fix this in the last PR of cancel product (to avoid restarting the whole review + QA process)

@matthieu-rolland could you integrate an `after` in your PR https://github.com/PrestaShop/PrestaShop/pull/17149 please?

src/Adapter/Order/CommandHandler/IssueStandardRefundHandler.php
jolelievre: You're right I never think about it But I don't want to lose the approved status just for this
That said maybe we could make a script to automatically add this in all the new files in 177?!

matks (approved):

matthieu-rolland (approved):

Progi1984 (approved):

PR #17419, opened 2020-01-29
Enable closure to be appended into LazyArray

### LINKED ISSUE #11901: Allow dynamic lazy arrays properties
@123monsite-regis has the following usecase: he receives a`ProductLazyArray` from hook hookFilterSupplierContent and wishes to add a property to it, **making it lazy as well**

So this does not work:
```
public function hookFilterSupplierContent($params) {
        $params['object']->appendArray(array('MymoduleVars' => array('MyVAriable' => 'MyVAriable'));
}
```
Added data is not lazy

A solution would be to create a LazyArray able to use closures.

```
$myLazyArray = new AdvancedClosureLazyArray();
$myLazyArray->appendArray(array(
    'get_my_url' => function () use ($link) {
        return getPageLink('foo', true, null);
    }
));
```
And when calling `$myLazyArray['get_my_url']` it would execute the closure then return the result

What do you think @jocel1 ? 😄

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Enable closures to be appended into LazyArray
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/11901
| How to test?  | No need for QA it's technical only. I covered it with unit tests. Please see ticket to understand the topic.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17419)

### DIFF
--- a/src/Adapter/Presenter/AbstractLazyArray.php
+++ b/src/Adapter/Presenter/AbstractLazyArray.php
@@ -143,6 +143,21 @@ abstract class AbstractLazyArray implements Iterator, ArrayAccess, Countable, Js
     }
 
     /**
+     * @param mixed $key
+     * @param \Closure $closure
+     */
+    public function appendClosure($key, \Closure $closure)
+    {
+        $this->arrayAccessList->offsetSet(
+            $key,
+            [
+                'type' => 'closure',
+                'value' => $closure,
+            ]
+        );
+    }
+
+    /**
      * The number of keys defined into the lazyArray.
      *
      * @return int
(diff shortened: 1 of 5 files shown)


matks:
Why does PrettyCI complain about a file that is totally unrelated to this PR 😓 ?

Progi1984:
I have the problem on another PR

matks:
@Progi1984 anyway I "fixed" it in this PR ... by linting Product.php file althought it was not related to my PR

Progi1984:
Thanks ;)

src/Adapter/Presenter/AbstractLazyArray.php
```diff
@@ -234,15 +249,33 @@ public function __clone()
...
-                    $methodName = $this->arrayAccessList[$index]['value'];
-                    $this->methodCacheResults[$index] = $this->{$methodName}();
-                }
-                $result = $this->methodCacheResults[$index];
-            } else { // if the index is associated with a value, just return the value
-                $result = $this->arrayAccessList[$index]['value'];
+            $type = $this->arrayAccessList[$index]['type'];
+            switch ($type) {
+                case 'method':
+                    $isResultAvailableInCache = (isset($this->methodCacheResults[$index]));
```
PierreRambaud, line 255:
```suggestion
                    $isResultAvailableInCache = isset($this->methodCacheResults[$index]);
```

PierreRambaud: Is this variable really needed, can be use directly in the `if` like on line 251 :thinking:

tests/Unit/Adapter/Presenter/LazyArrayImplementation.php
```diff
@@ -0,0 +1,57 @@
...
+namespace Tests\Unit\Adapter\Presenter;
+
+use PrestaShop\PrestaShop\Adapter\Presenter\AbstractLazyArray;
+
+/**
+ * Used for testing abstract class AbstractLazyArray
+ */
+class LazyArrayImplementation extends AbstractLazyArray
+{
+    private $propertOneWasCalled = false;
```
PierreRambaud:
```suggestion
    private $propertyOneWasCalled = false;
```

src/Adapter/Presenter/AbstractLazyArray.php
matks: When I modified this code, I was unsure about what `methodCacheResults` was. Was it a caching array ? Was it some methods name related to caching ? So I checked what it was and added this variable in order to make crystal clear for the next developer who will read this file.

matks: Also I personally like adding `(` and `)` around booleans statements 😛 so I guess it's a personal preference, not related to whether or not this PR is valid.

tests/Unit/Adapter/Presenter/LazyArrayImplementation.php
matks: Fixed

src/Adapter/Presenter/AbstractLazyArray.php
PierreRambaud: No problem, do what you think is the best for you :heart:

PierreRambaud (approved):

Progi1984 (approved):

matks:
Thanks for review !

Issue #17276, opened 2020-01-21
Free Shipping voucher not deleted after creating an order on BO


Robin-Fischer-PS:
**Describe the bug**
When creating an order on BO, if the merchant select the "Free shipping" option, then unselect it, a voucher is created. Since it's not used, the customer can see this voucher on FO in the "Voucher" section, and it can be used, because the "Total available" is still at "1".

**To Reproduce**
1. Go to BO page "Create order"
2. On the form, select then unselect the "Free shipping" option
3. Go to FO (with ou without creating the order on BO, it does not change the behaviour), on the Voucher list of the customer
4. See the "Free Shipping" voucher created by the "Create order" page on BO
5. You can use this voucher on FO

This happens on both Legacy and Migrated page.

**Screenshots**
![free shipping voucher](https://user-images.githubusercontent.com/55688534/72816123-fd77b400-3c67-11ea-913e-b81b6d776202.png)

**Additional information**
PrestaShop version: 1.7.6.x - 1.7.7.x
PHP version: N/A

khouloudbelguith:
Same issue with PS1.7.5.2 & PS1.7.6.3
https://drive.google.com/file/d/1kTYFWE2ZUSiklS1_ZIRcgUs8OzqihGGF/view
I’ll add this to the debug roadmap.
Thanks!

Progi1984:
Fixed by #17421

PR #17421, opened 2020-01-30
Delete free shipping voucher after creating an order on BO

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When creating an order on BO, if the merchant select the "Free shipping" option, a voucher is created, but is never removed even when free shipping option is switched back to false. This Pr fixes that, so the cart rule is deleted if free shipping cart rule exists by code and free shipping is switched to false. The fix is in BO, but affects FO too
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #17276
| How to test?  | see https://github.com/PrestaShop/PrestaShop/issues/17276

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17421)

### DIFF
--- a/src/Core/Domain/CartRule/Exception/CannotDeleteCartRuleException.php
+++ b/src/Core/Domain/CartRule/Exception/CannotDeleteCartRuleException.php
@@ -1,4 +1,28 @@
 <?php
+/**
+ * 2007-2019 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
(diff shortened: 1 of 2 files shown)


Progi1984 (dismissed):

matks (dismissed):

src/Core/Domain/CartRule/Exception/CannotDeleteCartRuleException.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
```
atomiix, line 26: Missing license header 🙂

atomiix (changes requested):

src/Core/Domain/CartRule/Exception/CannotDeleteCartRuleException.php
atomiix: And maybe add `declare(strict_types=1);`

zuk3975: strict types is useless here as its empty class. (unless we require it in all files by linter)

atomiix (approved):

matks (approved):

Progi1984:
Thanks @zuk3975

PR #17422, opened 2020-01-30
Check if id_cart  greater than 0 before deleting specific prices on Order add

### LINKED ISSUE #17427: Specific prices are deleted on manual order add from code
I had an issue while I was creating order manually from code. I was creating order without cart and by default id_cart is 0 in new Order object. After calling add function to insert an order all specific prices were deleted. I found out that after order add function calls SpecificPrice class method deleteByIdCart which in my case id_cart was 0 and because of that all specific prices were deleted. I know that I should not create an order without cart but in my opinion there should also be a check if id_cart > 0 and only then delete cart by id.

lukaslau:
| Questions     | Answers
| Branch?       | develop
| Description?  | Id_cart should be checked if it is greater than zero or not before deleting cart by id.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17427
| How to test?  | Create order manually from code without cart.

I had an issue while I was creating order manually from code. I was creating order without cart and by default id_cart is 0 in new Order object. After calling add function to insert an order all specific prices were deleted. I found out that after order add function calls SpecificPrice class method deleteByIdCart which in my case id_cart was 0 and because of that all specific prices were deleted. I know that I should not create an order without cart but in my opinion there should also be a check if id_cart > 0 and only then delete cart by id.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17422)


PierreRambaud:
Hi and thanks for your contribution,
Could you create an issue and also squash your commits?
Kind regards,

lukaslau:
https://github.com/PrestaShop/PrestaShop/issues/17427

PierreRambaud:
Thanks for the issue @zapper Can you squash your commit please?

lukaslau:
Is it good right now?

Matt75:
I think cache issue should be handled to another Pull Request and an Issue should be created for it.
I see some discussions related on PrestaShop Forums with APC :
https://www.prestashop.com/forums/topic/985490-cannot-use-a-scalar-value-as-an-array/

lukaslau:
> I think cache issue should be handled to another Pull Request and an Issue should be created for it.
> I see some discussions related on PrestaShop Forums with APC :
> https://www.prestashop.com/forums/topic/985490-cannot-use-a-scalar-value-as-an-array/

How cache issue is related to this issue with specific prices?

PierreRambaud:
@zapper To squash your commit: https://www.internalpointers.com/post/squash-commits-into-one-git :)

Matt75:
> > I think cache issue should be handled to another Pull Request and an Issue should be created for it.
> > I see some discussions related on PrestaShop Forums with APC :
> > https://www.prestashop.com/forums/topic/985490-cannot-use-a-scalar-value-as-an-array/
> 
> How cache issue is related to this issue with specific prices?

You have commited a file modification for it in this pull request.
As you said, it's not related to issue with specific prices, that's why I think you should create another issue and pull request for it.

matks:
Hi @lukaslau some time has passed 😅  but the content of the PR still contains a change related to cache https://github.com/PrestaShop/PrestaShop/pull/17422/files

That being said,

> I know that I should not create an order without cart but in my opinion there should also be a check if id_cart > 0 and only then delete cart by id.

We now have explored the structure of Order and I think it makes sense that we cannot create an order without a cart. Even in Back office a Cart is created. Without a cart I'm afraid the data model is invalid and this could create unwanted bugs where totals are wrong or order cannot be used correctly.

I don't think we should allow creation of an order without a cart, it might lack some important data. Maybe we should forbid it, to avoid inconsistencies?

kpodemski:
@matks 

definitely a good idea tho this issue is very, very specific, it makes no sense to create an order without an associated cart

lukaslau:
> @matks
> 
> definitely a good idea tho this issue is very, very specific, it makes no sense to create an order without an associated cart

I agree with you that it makes no sense to create an order without an associated cart but also it makes no sense for me not to check if cart id is greater than zero because otherwise most of the specific prices will be deleted (accidently). It would be just an extra line of protection for developers who write some custom code and creating orders manually.

PierreRambaud:
Hi @lukaslau 
Unfortunately, we took too long to answer you, and now, the fork is no longer available to rebase your PR.
I'm closing it, feel free to reopen if you're able to rebase it, and I'll be glad to review and merge it fast!

Kind regards

PR #9428, opened 2018-08-10
Added a new hook to custom SwiftMessage before sending mail

tonyyb:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  |  Adding hook to `Mail::send()` method to be able to custom SwiftMessage before sending mail.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | -
| How to test?  | Follow example bellow.

**Example of usage in a module :** 

```php
public function hookActionMailAlterMessageBeforeSend($params)
{
     /**
     * @var Swift_Message $message
     */
     $message = $params['message'];
     $headers = $message->getHeaders();
     $headers->addTextHeader('MY-CUSTOM-HEADER-KEY', 'MY-VALUE']);
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/9428)

PR #17431, opened 2020-01-30
Add  hook to alter Mail subject before send

davideapvd:
| Questions     | Answers
| Branch?       | develop 
| Description?  | I would like the ability to change the email subject without creating classes/Mail.php overrides, and given that the subject is changed in the send function, adding a [sitename] prefix, this would lead to a lot of repeated code
| Type?         | improvement
| BC breaks?    | no
| category | CO
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Creating a new hook listener for actionMailAlterSubjectBeforeSend and altering subject, this mimics the actionMailAlterMessageBeforeSend hook already available

example taken from https://github.com/PrestaShop/PrestaShop/pull/9428 

```
public function hookActionMailAlterSubjectBeforeSend($params)
{
     /**
     * @var String $subject
     */
     $shop_name = '[' . Tools::safeOutput(Configuration::get('PS_SHOP_NAME')) . '] ';
     if (substr($params['subject'], 0, strlen($shop_name)) == $shop_name) {
         $params['subject'] = substr_replace($params['subject'], '', 0, strlen($shop_name));
     }
}
````

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17431)

### DIFF
--- a/classes/Mail.php
+++ b/classes/Mail.php
@@ -598,6 +598,12 @@ class MailCore extends ObjectModel
                 'message' => &$message,
             ]);
 
+            // Hook to alter Swift Message before sending mail
+            Hook::exec('actionMailAlterSubjectBeforeSend', [
+                'subject' => &$subject,
+            ]);
+
+
             $send = $swift->send($message);
 
             ShopUrl::resetMainDomainCache();


Progi1984:
@davideapvd Hi, but normally, you can edit your subject with the hook `actionMailAlterMessageBeforeSend` with a `$message->setSubject('My new subject')` because $message is a `Swift Message`.

davideapvd:
> @davideapvd Hi, but normally, you can edit your subject with the hook `actionMailAlterMessageBeforeSend` with a `$message->setSubject('My new subject')` because $message is a `Swift Message`.

right, then I think we can discard this PR, thanks.

davideapvd:
@Progi1984 I just noticied that the message gets logged wrongly because the $subject is taken as is before processing, should we do something about it? https://github.com/davideapvd/PrestaShop/blob/3fbc073f9a3b866d855894ea795496698e5473ea/classes/Mail.php#L614

EDIT: ok in case we should I just created https://github.com/PrestaShop/PrestaShop/pull/17439

Issue #9948, opened 2018-08-22
[BOOM-3420] FO: Unable to select product combination, because you need to go via a combination that is out of stock.


prestonBot:
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-3420).

- _**Reporter:**_ marion_francois
- _**Created at:**_ Thu, 6 Jul 2017 17:20:32 +0200

**Steps to reproduce the issue:**
1-Create a product with combinations. Put stock in some combination but not in all.
2- Disable the option "Allow to buy products out of stock."
3-Disable the option "Display unavailable product attributes on the product page" in the Product Settings page
4-In FO, combinations out of stock can not be selected

https://drive.google.com/file/d/1YoTsG5tSydgnhtQAYhWWXiMhbWOTXTd_/view
<p>Dans Paramètres > Produits, je met Afficher les attributs indisponibles sur la fiche produit sur Non.<br/>
Ceci fonctionne si il y a qu'un seul attribut 

![2](https://user-images.githubusercontent.com/13449658/60576961-b800a180-9d7e-11e9-8c2d-9e0fda846b07.png)

![1](https://user-images.githubusercontent.com/13449658/60576959-b505b100-9d7e-11e9-9c0b-8f45cadcfbda.png)

Cependant, ça ne fonctionne plus quand il y a plusieurs attributs 

![4](https://user-images.githubusercontent.com/13449658/60576969-bdf68280-9d7e-11e9-80d7-d3eec5da7afd.png)

![3](https://user-images.githubusercontent.com/13449658/60576966-bb942880-9d7e-11e9-93b1-1301d7881bf1.png)

https://drive.google.com/file/d/1GYgWfsCiLa_YAZkShBhzxFf7TixHLdiN/view?usp=sharing

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-3420#comment-150096).

- _**Author:**_ aleeks
- _**Created at:**_ Thu, 13 Jul 2017 11:28:59 +0200

<p>On a déjà un ticket sur ça, vieux comme le monde. C'est plus compliqué que ça pour résoudre le problème avec le système actuelle <img class="emoticon" src="http://forge.prestashop.com/images/icons/emoticons/sad.png" height="16" width="16" align="absmiddle" alt="" border="0"/><br/>
[...]

PR #17457, opened 2020-01-31
Disable add-to-cart button when product/attribute combination is not available

### LINKED ISSUE #17345: Product attributes combination not working in 1.7.6.3
Hi,
Found HUGE bug in product attribute combination... :(
Fresh install 1.7.6.3,
PHP 7.2.11, 
Classic theme

**1) Setup Hummingbird printed t-shirt product as below**

**2) Setup - do not show unavailable product attributes on product page**

**3) Go to product and notice that black color in S size should't be available below**

**4) Then change size to M and notice that nothing happend color not changed to Black and White is still available as combination. Also links are the same place**

**5) And boom! Add product to cart...**

OMFG :(

### LINKED ISSUE #17437: The combination B is not refreshed when I change the combination A
**Describe the bug**
The combination B is not refreshed when I change the combination A if the same product has a completely different combination as Tshirt S white and Tshirt M black. 

**To Reproduce**
1. Go to Product Page
2. Create a combination t-shirt with **size S** and **black** as default
3. Create a combination t-shirt with **size M** and **white**
4. Save
5. Go in the FO
6. Change the size S to **M**
The color doesn't refresh and when clinking on the add to cart, it says the color isn't available

**Screenshots**
If applicable, add screenshots or screen records to help explain your problem.

**Additional information**
PrestaShop version: 1.7.6.3
PHP version: N/A

sowbiba:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | <br>#17345 : On 1.7.6.x version, when a product attributes combination doesnt have available quantity, you can click on Add-To-Cart button and 0 product are added. The reason is that, when there is 0 quantity, we selected the first available combination, So the button is displayed based on the wrong combination<br>#17437 : The DOM element refreshed doesnt exist in the product view page
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17345, #17437
| How to test?  | Please take a look at https://github.com/PrestaShop/PrestaShop/issues/17345#issuecomment-580663472 and https://github.com/PrestaShop/PrestaShop/issues/17437

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17457)


controllers/front/ProductController.php
```diff
@@ -976,6 +976,27 @@ private function getIdProductAttributeByRequest()
...
+        if (null === $idProductAttribute) {
+            $idProductAttribute = (int) Tools::getValue('id_product_attribute');
+        }
+
+        if (0 === $idProductAttribute) {
+            $idProductAttribute = (int) Product::getDefaultAttribute($this->product->id);
+        }
+
+        return $idProductAttribute;
+    }
```
PululuK: Hi @sowbiba,
I think we can gain a few milliseconds, in case it is zero for example.

```php
    private function getIdProductAttributeByGroupOrRequest()
    {
        $idProductAttribute = $this->getIdProductAttributeByGroup();
        if (0 === $idProductAttribute) {
            return (int) Product::getDefaultAttribute($this->product->id);
        } elseif (null === $idProductAttribute) {
			return (int) Tools::getValue('id_product_attribute');
        } else {
            return $idProductAttribute;
        }
    }
```
What do you think ?

PierreRambaud (dismissed):

Progi1984 (dismissed):

controllers/front/ProductController.php
```diff
@@ -1064,7 +1085,7 @@ public function getTemplateVarProduct()
         $product['id_product'] = (int) $this->product->id;
         $product['out_of_stock'] = (int) $this->product->out_of_stock;
         $product['new'] = (int) $this->product->new;
-        $product['id_product_attribute'] = $this->getIdProductAttributeByGroupOrRequestOrDefault();
+        $product['id_product_attribute'] = $this->getIdProductAttributeByGroupOrRequest();
```
jolelievre: Be careful you are reverting the behavior that was introduced by this PR: https://github.com/PrestaShop/PrestaShop/pull/16423

Careful not to reintroduce the previously fixed bug, this attribute fetching feature is a hell I know.. It has too many sub cases and the code is poorly written (sadly we don't have much valid way of doing this better in this legacy controller context).

Besides you only changed this call in one place in this controller, whereas the initial `getIdProductAttributeByGroupOrRequestOrDefault` is called in several places, theoraticaly it should be called everywhere for consistency. But maybe that's the point of your fix? In ajax we shouldn't fallback on available product?

jolelievre (changes requested):
Careful not to break the previous fix

controllers/front/ProductController.php
sowbiba: I noticed that taking the available id_product_attribute made us take the wrong product quantity
https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Product.php#L4978
By re-testing it right now, I see that the other attributes are changed too.

So making the "product-vaiants" container replaced well do the job. I will just keep the JS modification

themes/_core/js/product.js
```diff
@@ -169,7 +169,7 @@ function updateProduct(event, eventType, updateUrl) {
 
         $('.quickview .product-prices, .page-product:not(.modal-open) .row .product-prices').replaceWith(data.product_prices);
         $('.quickview .product-customization, .page-product:not(.modal-open) .row .product-customization').replaceWith(data.product_customization);
-        $('.quickview .product-variants .page-product:not(.modal-open) .row .product-variants').replaceWith(data.product_variants);
+        $('.quickview .product-variants .page-product:not(.modal-open) .row .product-variants, .page-product:not(.modal-open) .product-information .product-variants').replaceWith(data.product_variants);
```
jolelievre: How come the previous selector didn't work any more? Did the markup changed? Does it mean the other selector could potentially not work as well?

sowbiba: By other selectors you mean the other product-variants selectors ?
Actually, I do reproduce the quickview context but I dont know if there are other ways to display it

And all the other selectors had quickview and classic page selectors except "product-variants"

jolelievre: Ohhh I think I see the issue, seems like there was a missing `,`
```suggestion
        $('.quickview .product-variants, .page-product:not(.modal-open) .row .product-variants').replaceWith(data.product_variants);
```

themes/core.js
```diff
@@ -1,42 +1,174 @@
...
- * jQuery JavaScript Library v2.2.4
- * http://jquery.com/
- *
- * Includes Sizzle.js
- * http://sizzlejs.com/
- *
- * Copyright jQuery Foundation and other contributors
- * Released under the MIT license
- * http://jquery.org/license
+/******/ (function(modules) { // webpackBootstrap
```
atomiix: I think this is not the minified version, did you run `npm run build`? 🤔

sowbiba: Yes I did. I will do a new modification and run it again ;)

themes/_core/js/product.js
sowbiba: DOne, and you were right, there was just a missing comma

themes/core.js
sowbiba: done

jolelievre (dismissed):

NeOMakinG (dismissed):
LGTM

SD1982:
The pull request does disable "add to cart" button when a product is unavailable  but it not fix the [#17345](https://github.com/PrestaShop/PrestaShop/issues/17345), and [#17437](https://github.com/PrestaShop/PrestaShop/issues/17437) issues can you check it @sowbiba

sowbiba:
> The pull request does disable "add to cart" button when a product is unavailable but it not fix the [#17345](https://github.com/PrestaShop/PrestaShop/issues/17345), and [#17437](https://github.com/PrestaShop/PrestaShop/issues/17437) issues can you check it @sowbiba

The issue https://github.com/PrestaShop/PrestaShop/issues/17345
says that if you have
S - White => 1
S - Black => 0
M - White => 0
M - Black => 1
When you select S - Black or M - White, you have the button Add-to-cart active and you can add a quantity of 0 items
Now when you select a non available combination, the button is just disabled

sowbiba:
> The pull request does disable "add to cart" button when a product is unavailable but it not fix the [#17345](https://github.com/PrestaShop/PrestaShop/issues/17345), and [#17437](https://github.com/PrestaShop/PrestaShop/issues/17437) issues can you check it @sowbiba

For https://github.com/PrestaShop/PrestaShop/issues/17437

If you just have
S - White => 1
M - Black => 1
as combinations. When you arrive on the front page, you have the size selector with S and M and just a white box for colors
When you change the size to M, the white box is replaced by a black one

Robin-Fischer-PS:
Hi @sowbiba @SD1982 ! I have this behaviour with the PR (when I follow the steps in https://github.com/PrestaShop/PrestaShop/issues/17345) : 
I cannot change anymore the size of the t-shirt, I'm always redirected to the "S" size. This is similar to the behaviour described in this bug, which is still not fixed yet : https://github.com/PrestaShop/PrestaShop/issues/9948

Since we're back to 1762 (and previous versions) behaviour, I think it's OK, but I would like to be sure that's what expected with this PR. Could you confirm @sowbiba ? Thanks !

The screenrecord of the behaviour I have : 

https://drive.google.com/file/d/10X8PFj2zeJHCSrDP0GokzdxNCtCLkeAN/view?usp=sharing

sowbiba:
Ok, this behaviour was the same in 1.7.6.2, you could not change the size nor the color. 
In 1.7.6.3, you can change values and add to cart a non available combination.
The comment from @khouloudbelguith  in #17345 describes the difference.
I focused on the case when "show unavailable product attributes on product page" is ON and didn't see the behaviour when it is OFF

I suggest to try to fix #9948 too to make it clean once

Robin-Fischer-PS:
I'm okay with your suggestion, since you're working on this topic, it would be nice to fix it entirely, even if it's a bit out of scope for a patch version ! (Maybe a PM validation here could be useful)

Emiliotorrecilla:
Hi @sowbiba ,
I've tried changes you say but they don't work in my store. I have the problem mentioned in #17437 and I need to fix it because I can't go back to version 1.7.6.2 and I can't wait for the new version either. Please, can you help me?

sowbiba:
> Hi @sowbiba ,
> I've tried changes you say but they don't work in my store. I have the problem mentioned in #17437 and I need to fix it because I can't go back to version 1.7.6.2 and I can't wait for the new version either. Please, can you help me?

Hello @Emiliotorrecilla 
The file I modified `themes/_core/js/product.js` is the source file. The one used by the frontend is the "compiled" one `themes/core.js`

You can edit it and add the missing comma like suggested here
https://github.com/PrestaShop/PrestaShop/issues/17345#issuecomment-583353708

And after that, you have to hit ctrl+F5 on your browser to be sure the JS is reloaded

Progi1984 (dismissed):

PierreRambaud (dismissed):

eternoendless (dismissed):

PierreRambaud (dismissed):

jolelievre (dismissed):

Progi1984:
@sowbiba Travis is  :red_circle:

matks:
It's a PrettyCI issue, @atomiix is working on the fix

atomiix (approved):

atomiix (approved):

matks (approved):

Robin-Fischer-PS:
Tested by @SD1982 and myself, it's QA :heavy_check_mark: !

atomiix:
Thanks @sowbiba, @SD1982 & @Robin-Fischer-PS !

PR #17461, opened 2020-01-31
Fix wrong wordings on "New" translations

### LINKED ISSUE #17455: Use different strings for "New" translation
Is it possible to use different strings for _New_ translation?
This wording is used in 2 different cases:

- on new products (means _Nouveauté_ in french)
- on product condition (means _Neuf_ in french)

Currently, there is only one string and if I want to change translation for new products, it also changes wording for product condition (FO and BO).

In my opinion, we need to use two strings:

- `{l s='New product' d='Shop.Theme.Catalog'}` = _New_ in english = _Nouveauté_ in french
- `{l s='New' d='Shop.Theme.Catalog'}` = _New_ in english = _Neuf_ in french

@LouiseBonnard what do you think?

micka-fdz:
Questions | Answers
-- | --
Branch? | develop
Description? | This PR fixes wrong wordings on _New_ translations (details below)
Type? | bug fix
Category? | FO
BC breaks? | no
Deprecations? | no
Fixed ticket? | Fixes #17455
How to test? | - _New product_ should be use for new products (_Nouveauté_ in french)<br>- _New_ should be use for product condition (_Neuf_ in french)
| Sponsor company | @Creabilis

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17461)

### DIFF
--- a/src/Adapter/Presenter/Product/ProductLazyArray.php
+++ b/src/Adapter/Presenter/Product/ProductLazyArray.php
@@ -223,7 +223,7 @@ class ProductLazyArray extends AbstractLazyArray
             case 'new':
                 return [
                     'type' => 'new',
-                    'label' => $this->translator->trans('New product', [], 'Shop.Theme.Catalog'),
+                    'label' => $this->translator->trans('New', [], 'Shop.Theme.Catalog'),
                     'schema_url' => 'https://schema.org/NewCondition',
                 ];
             case 'used':
@@ -465,9 +465,9 @@ class ProductLazyArray extends AbstractLazyArray
         }
 
         if ($this->product['new']) {
-            $flags['new'] = [
+            $flags['new'] = [f
                 'type' => 'new',
-                'label' => $this->translator->trans('New', [], 'Shop.Theme.Catalog'),
+                'label' => $this->translator->trans('New product', [], 'Shop.Theme.Catalog'),
             ];
         }
 
--- a/src/Adapter/Presenter/Product/ProductLazyArray.php
+++ b/src/Adapter/Presenter/Product/ProductLazyArray.php
@@ -465,7 +465,7 @@ class ProductLazyArray extends AbstractLazyArray
         }
 
         if ($this->product['new']) {
-            $flags['new'] = [f
+            $flags['new'] = [
                 'type' => 'new',
                 'label' => $this->translator->trans('New product', [], 'Shop.Theme.Catalog'),
             ];


Progi1984:
@micka-fdz Could you update your PR for using the contribution table ? Thanks

micka-fdz:
@Progi1984 done !

PierreRambaud (approved):

Progi1984 (approved):

Progi1984:
Need "Wording :heavy_check_mark:" by @LouiseBonnard Before QA

LouiseBonnard:
It is not wording approved yet! I will ask for the QA to test it once the wording validated. ;-)

micka-fdz:
_New_ wording is always used on product condition except on `ProductLazyArray.php` file.
In this file, _New_ and _New product_ wordings are reversed. My PR is only about to fix that.

LouiseBonnard:
@micka-fdz, thanks for the explanation, it is perfectly clear now! I therefore approve the wording. ;-)

As I said in https://github.com/PrestaShop/PrestaShop/issues/17455#issuecomment-580828779, the best solution might be to use _New condition_ instead in both the back office (when selecting the condition on the 'Options' tab in the 'Product' page) and the front office (when on the 'Product Details' tab at the end of a product page). It will allow us to make sure we get a relevant translation, what do you think @micka-fdz?

micka-fdz:
@LouiseBonnard if I understand your proposal, you want to use:
- _New condition_ = on product condition
- _New_ = on new/recent products

I think it can be a good alternative but in this case, wordings have to be changed everywhere (in core but also in native modules like `ps_facetedsearch`).

LouiseBonnard:
@micka-fdz, yep, you got it, thanks a lot for the feedback! I'll create another issue to bring this idea and see how it can be applied.

matks (approved):

matks:
@LouiseBonnard Can we go for QA now, or not yet ?

PierreRambaud (approved):

SD1982:
Hello @micka-fdz Your PR seems to do correctly what it is supposed to do but before validating it I would like to ask @LouiseBonnard  opinion about the French translation for the product condition because I am not sure that this is the best term in this situation I think the translation should be "neuf" rather than "nouveau"

**In English :**
pr#17461 english

**In French :**
pr#17461 french

What do you think @LouiseBonnard ?

LouiseBonnard:
@SD1982, thanks for the ping. It is normal, it has not been merged yet so the translation is expected to be the one that formerly labels new products. :-)

I've just updated the translation on Crowdin so that _New_ can equal _Neuf_ instead of _Nouveau_, it should be okay in a couple of days.

SD1982:
Thanks @micka-fdz

Progi1984:
Thanks @micka-fdz

PR #17467, opened 2020-01-31
Category admin jumps to home when deleting a category.

### LINKED ISSUE #17463: [BO] Category admin jumps to home when deleting a category
You have categories: 
Home > Cars > SUVs > Small
Home > Cars > SUVs > Medium
Home > Cars > SUVs > Big

You are editing categories.
You are in SUVs category.
You delete Small category and the view jumps all the way back to Home category, annoying.

**Additional information**
PrestaShop version: 1.7.6.2

PululuK:
| Questions     | Answers
| Branch?       | develop
| Description?  | Category admin jumps to home when deleting a category.                           
| Type?         | bug fix 
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17463
| How to test?  | **Considere this category three :** <br>- Home > Cars > SUVs > Small<br>- Home > Cars > SUVs > Medium<br>- Home > Cars > SUVs > Big.<br><br>1. You are editing categories.<br>2. You are in SUVs category.<br>3. You delete Small category and the view jumps all the way back to Home category, annoying.<br><br>By @Hlavtox 

Sorry @khouloudbelguith, but these two issues (https://github.com/PrestaShop/PrestaShop/issues/17463 and https://github.com/PrestaShop/PrestaShop/issues/15588) are very different.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17467)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
@@ -629,9 +629,11 @@ class CategoryController extends FrameworkBundleAdminController
     {
         $deleteCategoriesForm = $this->createForm(DeleteCategoriesType::class);
         $deleteCategoriesForm->handleRequest($request);
+        $idParent = $this->configuration->getInt('PS_HOME_CATEGORY');
 
         if ($deleteCategoriesForm->isSubmitted()) {
             $categoriesDeleteData = $deleteCategoriesForm->getData();
+            $idParent = (int) $categoriesDeleteData['categories_to_delete_parent'];
 
             try {
                 $command = new DeleteCategoryCommand(
@@ -647,7 +649,7 @@ class CategoryController extends FrameworkBundleAdminController
             }
         }
 
-        return $this->redirectToRoute('admin_categories_index');
+        return $this->redirectToRoute('admin_categories_index', ['categoryId' => $idParent]);
     }
 
     /**
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
@@ -92,7 +92,7 @@ class CategoryController extends FrameworkBundleAdminController
         $categoryViewDataProvider = $this->get('prestashop.adapter.category.category_view_data_provider');
         $categoryViewData = $categoryViewDataProvider->getViewData($currentCategoryId);
 
-        $deleteCategoriesForm = $this->createForm(DeleteCategoriesType::class);
+        $deleteCategoriesForm = $this->createForm(DeleteCategoriesType::class, ['categories_to_delete_parent' => (int) $currentCategoryId], []);
 
         $showcaseCardIsClosed = $this->getQueryBus()->handle(
             new GetShowcaseCardIsClosed((int) $this->getContext()->employee->id, ShowcaseCard::CATEGORIES_CARD)
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
@@ -584,10 +584,12 @@ class CategoryController extends FrameworkBundleAdminController
     {
         $deleteCategoriesForm = $this->createForm(DeleteCategoriesType::class);
         $deleteCategoriesForm->handleRequest($request);
+        $idParent = $this->configuration->getInt('PS_HOME_CATEGORY');
 
         if ($deleteCategoriesForm->isSubmitted()) {
             try {
                 $categoriesDeleteData = $deleteCategoriesForm->getData();
+                $idParent = (int) $categoriesDeleteData['categories_to_delete_parent'];
                 $categoryIds = array_map(function ($categoryId) {
                     return (int) $categoryId;
                 }, $categoriesDeleteData['categories_to_delete']);
@@ -608,7 +610,7 @@ class CategoryController extends FrameworkBundleAdminController
             }
         }
 
-        return $this->redirectToRoute('admin_categories_index');
+        return $this->redirectToRoute('admin_categories_index', ['categoryId' => $idParent]);
     }
 
     /**
--- a/src/PrestaShopBundle/Form/Admin/Sell/Category/DeleteCategoriesType.php
+++ b/src/PrestaShopBundle/Form/Admin/Sell/Category/DeleteCategoriesType.php
@@ -67,6 +67,9 @@ class DeleteCategoriesType extends AbstractType
                 'entry_type' => HiddenType::class,
                 'label' => false,
                 'allow_add' => true,
+            ])
+            ->add('categories_to_delete_parent', HiddenType::class, [
+                'label' => false,
             ]);
     }
 }
--- a/src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Categories/Blocks/delete_categories_modal.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Categories/Blocks/delete_categories_modal.html.twig
@@ -47,7 +47,7 @@
         {{ form_widget(deleteCategoriesForm.categories_to_delete) }}
       </div>
       <div class="d-none">
-        {{ form_widget(deleteCategoriesForm.categories_to_delete_parent, { 'attr': {'value': app.request.get('categoryId') } }) }}
+        {{ form_widget(deleteCategoriesForm.categories_to_delete_parent) }}
       </div>
     </div>
   {% endblock %}
--- a/src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Categories/Blocks/delete_categories_modal.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Categories/Blocks/delete_categories_modal.html.twig
@@ -46,6 +46,9 @@
       <div class="d-none">
         {{ form_widget(deleteCategoriesForm.categories_to_delete) }}
       </div>
+      <div class="d-none">
+        {{ form_widget(deleteCategoriesForm.categories_to_delete_parent, { 'attr': {'value': app.request.get('categoryId') } }) }}
+      </div>
     </div>
   {% endblock %}
 {% endembed %}


src/PrestaShopBundle/Form/Admin/Sell/Category/DeleteCategoriesType.php
```diff
@@ -31,6 +31,7 @@
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\CollectionType;
 use Symfony\Component\Form\Extension\Core\Type\HiddenType;
+use Symfony\Component\Form\Extension\Core\Type\IntegerType;
```
Progi1984: @PululuK Could you remove it, please ? It's not used.

PululuK: Hi @Progi1984 
Sorry, I'm getting old and I'm going blind 😅
Done
Thank you

Progi1984 (dismissed):

PierreRambaud (dismissed):

khouloudbelguith:
Hi @PululuK,

Could you please try to rebase your branch.
After running the `composer install`, PrestaShop version is PS1.7.6.1.
image

Thanks!

PululuK:
Hi @khouloudbelguith 
Thank for you feedback, but i don't understand the question, there is no conflict on the modified files. @PierreRambaud  @Progi1984 can you check and help please ? 
I'm getting really old, and I'm having more and more trouble with technology 😁

Progi1984:
Hi @PululuK, after discussion with @khouloudbelguith, I understand the problem. Your branch `PululuK:PululuK-patch9` is an "old" version (https://github.com/PululuK/PrestaShop/blob/PululuK-patch9/app/AppKernel.php#L34). Could you rebase it on the latest version of the `develop` branch of `PrestaShop/PrestaShop` ? Thanks

PululuK:
> Hi @PululuK, after discussion with @khouloudbelguith, I understand the problem. Your branch `PululuK:PululuK-patch9` is an "old" version (https://github.com/PululuK/PrestaShop/blob/PululuK-patch9/app/AppKernel.php#L34). Could you rebase it on the latest version of the `develop` branch of `PrestaShop/PrestaShop` ? Thanks

Hi @Progi1984 
Can you check please !
Thank you

Progi1984 (dismissed):

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Categories/Blocks/delete_categories_modal.html.twig
```diff
@@ -46,6 +46,9 @@
       <div class="d-none">
         {{ form_widget(deleteCategoriesForm.categories_to_delete) }}
       </div>
+      <div class="d-none">
+        {{ form_widget(deleteCategoriesForm.categories_to_delete_parent, { 'attr': {'value': app.request.get('categoryId') } }) }}
```
jolelievre: You should inject this data in the controller when creating the form rather than in the template (in `PrestaShopBundle\Controller\Admin\Sell\Catalog\CategoryController::indexAction`)

jolelievre (changes requested):

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Categories/Blocks/delete_categories_modal.html.twig
PululuK: Hi @jolelievre,
Thanks for feedback !
Could you explain why ? Because in the other templates, it is injected as well
See : https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Categories/Blocks/cover_image.html.twig#L35


With your suggestion, I think we would have something like this ...

**Controller**
```php
$deleteCategoriesForm = $this->createForm(
    DeleteCategoriesType::class, 
    array(
        'categories_to_delete_parent'=>$request->query->get('categoryId')
    ),
    [],
);

```
**Template**

```html
{{ form_widget(deleteCategoriesForm.categories_to_delete_parent) }}
```
It's beautiful all that ... But why not the other?

Thanks

jolelievre: Yes that's exactly what I thought (only prefer short syntax for array `[]` or it will fail the code style checker)

The reason I advise this is mainly because we use a Symfony form generated from the backoffice, so the whole form (including its default data) should be created on the server side (in the controller) and the template (the view) is only responsible for displaying/rendering it
But it shouldn't inject data in it at the last moment Even the code/hack you needed to make it work is a hint it's not supposed to work that way (inject via attr => value used by the `form_widget` macro)

The other example you refer to is different because:
- the form is completely generated in the template, so no Symfony form in which you could properly inject the data
- it's actually not the form data that is modified, it's a form attribute used in js in which we inject a generate url

Using `app.request.get(*)` in the template might be useful and sometimes necessary, no problem with that (also one should be careful with some potential cache issues) However we are trying to clean all the forms in order to lighten the template as much as possible (see this PR as an example of what we aim https://github.com/PrestaShop/PrestaShop/pull/17558)

And to do that we need clean form types, correctly set in the controller so that you wouldn't even need to modify the template, the idea is that modifying the form type alone will be enough to update the page

I hope this (long) explanation makes sense ^^

jolelievre: Oh and I'm not 100% sure what you proposed was never done and can't be found in the core code if you search for it ^^ But it still wouldn't be a good practice

That said, thanks a lot for your contribution in any case 😉

PululuK: It seems legitimate to me. Done !
Thank you @jolelievre

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
```diff
@@ -92,7 +92,7 @@ public function indexAction(Request $request, CategoryFilters $filters)
         $categoryViewDataProvider = $this->get('prestashop.adapter.category.category_view_data_provider');
         $categoryViewData = $categoryViewDataProvider->getViewData($currentCategoryId);
 
-        $deleteCategoriesForm = $this->createForm(DeleteCategoriesType::class);
+        $deleteCategoriesForm = $this->createForm(DeleteCategoriesType::class, ['categories_to_delete_parent' => (int) $request->query->get('categoryId')], []);
```
jolelievre: Mmm, I believe `$request->query` is the ParameterBag for POST request, for GET request it's supposed to be stored in `$request->request`
Or alternatively you can do `$request->get()` directly which combines both (along with route attributes)

jolelievre: My bad it's the opposite ^^ `$request->query` => GET, `$request->request` => POST

jolelievre (dismissed):
Thank you @PululuK

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
PululuK: Hi @jolelievre 
We don't really need all of that ... I already have a variable that gives me what I want !
See : https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php#L91

Thank you all the same

PululuK:
> Thank you @PululuK

Hi 
Sorry, can you check again ?
Thanks

PierreRambaud (dismissed):

jolelievre (approved):

khouloudbelguith:
Hi @PululuK,

It is ok.
But, we should test this case:
1. If we create these categories `Home > Cars > SUVs > Small`
2. We are in `SUVs` category
3. Delete the `Small` category

=> we should be redirected to the `Cars` category, because, `SUVs` has no subcategories => so we can't view the `SUVs` category.

Or we should be in the `SUVs` category like this
https://drive.google.com/file/d/1jPP6Sazz2fC1w6cfcR0e5LhNRUr3vb__/view

Ping @PrestaShop/prestashop-product-team what do you think?

Thanks!

jolelievre:
Hi @khouloudbelguith 

I'm not sure it's a problem, because navigating into an empty category is possible I believe?
I even think it's one of the way to create a sub category:
- navigate to the parent category (event empty)
- click on the "Add new category" button
- the creation form should have preselected the parent category

But maybe I'm mistaking

jolelievre (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
```diff
@@ -647,7 +649,7 @@ public function deleteAction(Request $request)
             }
         }
 
-        return $this->redirectToRoute('admin_categories_index');
+        return $this->redirectToRoute('admin_categories_index', ['categoryId' => $idParent]);
```
arouiadib: Is this the desired behaviour? IMHO, it is preferable to stay in current index page when I delete a category, especially when  I delete one category (which is the aim of this action). wdyt?

colinegin:
I agree with @jolelievre, I think you should be able to see an empty list if there is no subcategory. So second option suits me better, as this is the current behaviour (if you click on the link of a category which does not have any subcategory, you can see an empty list).

Thanks !

matks (dismissed):

SD1982:
Hi @PululuK You PR correctly fix #17463 like it was described but there is still a case that is not covered by your PR as you can see in my screen record if we delete subcategories by bulk action we are still redirected to home category and i think we shouldn't. Ping @PrestaShop/prestashop-product-team what do you think?

**Screen record :** 
https://drive.google.com/file/d/1S4w4oaptUx1wOGnoFl6_wWqQnAHWMb-k/view?usp=sharing

PululuK:
> Hi @PululuK You PR correctly fix #17463 like it was described but there is still a case that is not covered by your PR as you can see in my screen record if we delete subcategories by bulk action we are still redirected to home category and i think we shouldn't. Ping @PrestaShop/prestashop-product-team what do you think?
> 
> **Screen record :**
> https://drive.google.com/file/d/1S4w4oaptUx1wOGnoFl6_wWqQnAHWMb-k/view?usp=sharing

Hi @SD1982  
Thank you so much man !

Hi guys @jolelievre @matks @PierreRambaud 
I think I should handle this case in my PR even if it is not described here https://github.com/PrestaShop/PrestaShop/issues/17463 instead of creating a new PR. What do you think ?

colinegin:
Hello @SD1982 and @PululuK ,

Using the bulk action should have the same effect than having an action on one single category, meaning you should stay on the category listing.
And deleting all subcategories at once with the bulk action should leave the category list empty.

Thanks !

PululuK:
> Hello @SD1982 and @PululuK ,
> 
> Using the bulk action should have the same effect than having an action on one single category, meaning you should stay on the category listing.
> And deleting all subcategories at once with the bulk action should leave the category list empty.
> 
> Thanks !

Thanks @colinegin  
I will look at it!

PululuK:
> Hi @PululuK You PR correctly fix #17463 like it was described but there is still a case that is not covered by your PR as you can see in my screen record if we delete subcategories by bulk action we are still redirected to home category and i think we shouldn't. Ping @PrestaShop/prestashop-product-team what do you think?
> 
> **Screen record :**
> https://drive.google.com/file/d/1S4w4oaptUx1wOGnoFl6_wWqQnAHWMb-k/view?usp=sharing

@SD1982 @colinegin 
Done ! I'm waiting for Reviewers  now ... can you update the labels please ? Thanks

Progi1984 (approved):

jolelievre (approved):

SD1982:
Hi @PululuK Now your pull request correctly delete subcategory without redirecting to "home" category in bulk action to . Thanks @PululuK !

Progi1984:
Thanks @PululuK

PR #17476, opened 2020-02-03
Don't allow comma in order payment amount field

### LINKED ISSUE #17475: Can't add an amount using "," in block payment in order page migrated
**Describe the bug**

**To Reproduce**
1. Go to the order page migrated 
2. Select an order 
3. Try to add a payment with a "," like 5,80

**Screenshots**
Migrated : https://drive.google.com/file/d/1Yq-H3O9_Yk2ZKE3FO5uDmrF3vPF8wklj/view
Legacy : https://drive.google.com/file/d/1Gjo_9kVEREleGFhzArcJt7Hbro5WyA6T/view

**Additional information**
PrestaShop version: 1.7.7.x
PHP version: N/A

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When adding an amount in the migrated order page (block payment, partial refound, add/edit product), it was sometime possible to use a comma instead of a dot, sometime impossible. This PR fixes it by authorizing only dots.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17475
| How to test?  | 1. Go to the order page migrated<br>2. Select an order<br>3. Try to add a payment with a "," like 5,80<br>4. You should have an error message.<br><br>Same thing when adding/editing a product, partial refound

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17476)


jolelievre (commented):
It feels weird to use a `NumberType` in the form so it should return a float value, to then recast it into a string no?

That being said, should this behavior (using `Number` in the command to automatically parse it) be used everywhere in other commands? This would make them all consistent and probably avoid some bugs.
Besides I wondered what happens if you use a value with groups separator? like `1,025.48` ?

atomiix:
@jolelievre I agree but `NumberType` is needed to parse numbers in different format, but return `float` and `Number` takes only `string`. Do you think of a better way of doing it?

I just tried with `1,025.48` and got
>This value is not valid.

eternoendless:
Have you had a look at `\PrestaShopBundle\Utils\FloatParser`?

atomiix:
@eternoendless I'm not sure that it's useful because `NumberType` [already does the parsing](https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Form/Extension/Core/Type/NumberType.php#L26). WDYT?

eternoendless:
Money amounts should always be strings, never floats, because of rounding errors.

admin-dev/themes/new-theme/js/pages/order/view/order-product-add.js
```diff
@@ -169,8 +169,12 @@ export default class OrderProductAdd {
         newRow: response,
       });
     }, (response) => {
-      if (response.message) {
-        $.growl.error({message: response.message});
+      this.productAddActionBtn.prop('disabled', false);
+      this.invoiceSelect.prop('disabled', false);
+      this.combinationsSelect.prop('disabled', false);
+
+      if (response.responseJSON !== undefined && response.responseJSON.message !== undefined) {
```
sowbiba: Is it enough to do ```if (response.responseJSON && response.responseJSON.message)``` ?

admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
```diff
@@ -173,8 +173,8 @@ export default class OrderProductEdit {
         newRow: response
       });
     }, (response) => {
-      if (response.message) {
-        $.growl.error({message: response.message});
+      if (response.responseJSON !== undefined && response.responseJSON.message !== undefined) {
```
sowbiba: same here

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -97,7 +98,7 @@ public function computeOrderRefund(
             $refundedAmount = $voucherAmount = $chosenVoucherAmount;
         }
 
-        $shippingCostAmount = $shippingRefund ?: false;
+        $shippingCostAmount = (float) (string) $shippingRefund;
```
sowbiba, line 101: I understand you need to cast twice. the first one will call __toString method and the second cast it to float.

Suggestion : Can't we add a method toFloat in Number class to do it in one place ?

atomiix: @sowbiba yes, that could be a solution. However it would means that we should update the Decimal library and IMO it's not worth it as we're planning to use Decimal everywhere we can instead of float in the coming releases.

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1475,6 +1495,16 @@ private function getErrorMessages(Exception $e)
         ];
     }
 
+    private function getPaymentErrorMessages(Exception $e)
+    {
+        return array_merge($this->getErrorMessages($e), [
+            InvalidArgumentException::class => $this->trans(
+                'Only numbers and decimal points (".") are allowed in the amount fields of the payment block, e.g. 10.50 or 1050.',
```
LouiseBonnard, line 1513: Why is it (slightly) different from the first exception above, at line 1443? Can't we use the same string twice?

atomiix: @LouiseBonnard Because the exception line 1443 is not related to the payment block, it's a more global error message for other amount fields

matthieu-rolland (dismissed):

matthieu-rolland (dismissed):

PierreRambaud (dismissed):

Progi1984:
@atomiix Please rebuild assets before merge

matks:
Thank you @atomiix

matks:
Rebased on 1.7.7.x

matks:
Rebased again

PierreRambaud (approved):

matthieu-rolland (approved):

PR #17410, opened 2020-01-29
Add FO hook displayFooterCategory

matks:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | Restore 1.6 hook displayFooterCategory
| Type?         | new feature
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/16628
| How to test?  | No QA needed I think

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17410)

PR #17203, opened 2020-01-16
Add override hooks into Presenters

matks:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Add override hooks into Presenters, allowing to enrich the data built by these services
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/11125
| How to test?  | No need for QA I guess 🤔

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17203)

PR #17477, opened 2020-02-03
Register new 1.7.8 hooks into XML and SQL files

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Register new 1.7.8 hooks into XML and SQL files
| Type?         | new feature
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 

Finishes the job for PRs https://github.com/PrestaShop/PrestaShop/pull/17410/ and https://github.com/PrestaShop/PrestaShop/pull/17203/

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17477)

### DIFF
--- a/install-dev/upgrade/sql/1.7.8.0.sql
+++ b/install-dev/upgrade/sql/1.7.8.0.sql
@@ -2,9 +2,9 @@ SET SESSION sql_mode='';
 SET NAMES 'utf8';
 
 INSERT IGNORE INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `position`) VALUES
-  (NULL, 'displayFooterCategory', 'Category footer', 'This hook adds new blocks under the products listing in a category/search', '1 '),
-  (NULL, 'actionPresentCart', 'Cart Presenter', 'This hook is called before a cart is presented', '1 '),
-  (NULL, 'actionPresentOrder', 'Order footer', 'This hook is called before an order is presented', '1 '),
-  (NULL, 'actionPresentOrderReturn', 'Order Return Presenter', 'This hook is called before an order return is presented', '1 '),
-  (NULL, 'actionPresentProduct', 'Product Presenter', 'This hook is called before a product is presented', '1 ')
+  (NULL, 'displayFooterCategory', 'Category footer', 'This hook adds new blocks under the products listing in a category/search', '1'),
+  (NULL, 'actionPresentCart', 'Cart Presenter', 'This hook is called before a cart is presented', '1'),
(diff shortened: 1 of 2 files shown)


install-dev/upgrade/sql/1.7.8.0.sql
```diff
@@ -0,0 +1,10 @@
+SET SESSION sql_mode='';
+SET NAMES 'utf8';
+
+INSERT IGNORE INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `position`) VALUES
+  (NULL, 'displayFooterCategory', 'Category footer', 'This hook adds new blocks under the products listing in a category/search', '1 '),
+  (NULL, 'actionPresentCart', 'Cart Presenter', 'This hook is called before a cart is presented', '1 '),
```
jolelievre: I just noticed, I believe we made a typo error here (the same can be found in upgrade scripts up to `1.7.4.0` 😅)
The position inserted is `'1 '` (string with extra space), I guess mysql or the sql interpreter is smart enough to set it correctly but still it would be better to simply use an int `1` What do you think?

jolelievre (changes requested):
Woups ^^

install-dev/upgrade/sql/1.7.8.0.sql
matks: I thought it was on purpose 😄 also indeed I was wondering what the hell it was

Well, I told myself "must be a legacy rule ..."

jolelievre: Well I can't be sure but it seems like a mistake, especially since the database field is a TINYINT

matks: Fixed

Progi1984 (dismissed):

install-dev/upgrade/sql/1.7.8.0.sql
```diff
@@ -0,0 +1,10 @@
+SET SESSION sql_mode='';
+SET NAMES 'utf8';
```
PierreRambaud:
```suggestion
SET NAMES 'utf8mb4';
```

matks: Indeed ! nice catch

Progi1984 (approved):

Progi1984:
Fixed Travis

PierreRambaud (approved):

jolelievre (approved):

matks:
Thanks !

PR #17482, opened 2020-02-03
Allow token for TokenisedUrlsListener class and in form using the method=GET attribute

### LINKED ISSUE #17188: Token invalid when trying to edit a translation
**Describe the bug**
Token invalid when trying to edit a translation

**To Reproduce**
1. Go to BO > international > translations
2. Select a translation type and a language (no matter which one)
3. Click on edit
4. See the page invalid token

**Screenshots**
capture d'écran_2274

**Additional information**
PrestaShop version: 1.7.7.x, develop
PHP version: 7.2

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Form CsrfToken should also work when working with a form using `method="GET"`. Also the `prestashop.tokenized_url_listener` service wasn't able to work because in `onRequest` context the `prestashop.user_provider` service is never initialized, so the `getUsername` is `null`. 
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | Fixes #17188 
| How to test?  | Forms in Translations page should work as before. Ticket should be fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17482)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/TranslationsController.php
+++ b/src/PrestaShopBundle/Controller/Admin/TranslationsController.php
@@ -26,6 +26,7 @@
 
 namespace PrestaShopBundle\Controller\Admin;
 
+use PrestaShop\PrestaShop\Core\Form\FormHandlerInterface;
 use PrestaShop\PrestaShop\Core\Language\Copier\LanguageCopierConfig;
 use PrestaShopBundle\Security\Annotation\AdminSecurity;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
@@ -99,15 +100,20 @@ class TranslationsController extends FrameworkBundleAdminController
         $legacyController = $request->attributes->get('_legacy_controller');
         $legacyContext = $this->get('prestashop.adapter.legacy.context');
         $kpiRowFactory = $this->get('prestashop.core.kpi_row.factory.translations_page');
-        $formHandler = $this->get('prestashop.admin.translations_settings.form_handler');
-        $form = $formHandler->getForm();
+        $modifyTranslationsForm = $this->getModifyTranslationsFormHander()->getForm();
+        $addUpdateLanguageForm = $this->getAddUpdateLanguageTranslationsFormHander()->getForm();
+        $exportLanguageForm = $this->getExportLanguageTranslationsFormHander()->getForm();
+        $copyLanguageForm = $this->getCopyLanguageTranslationsFormHander()->getForm();
 
         return [
             'layoutTitle' => $this->trans('Translations', 'Admin.Navigation.Menu'),
             'enableSidebar' => true,
             'help_link' => $this->generateSidebarLink($legacyController),
             'kpiRow' => $kpiRowFactory->build(),
-            'translationSettingsForm' => $form->createView(),
+            'copyLanguageForm' => $copyLanguageForm->createView(),
+            'exportLanguageForm' => $exportLanguageForm->createView(),
+            'addUpdateLanguageForm' => $addUpdateLanguageForm->createView(),
+            'modifyTranslationsForm' => $modifyTranslationsForm->createView(),
             'addLanguageUrl' => $legacyContext->getAdminLink('AdminLanguages', true, ['addlang' => '']),
         ];
     }
@@ -142,13 +148,13 @@ class TranslationsController extends FrameworkBundleAdminController
      */
     public function addUpdateLanguageAction(Request $request)
     {
-        $formHandler = $this->get('prestashop.admin.translations_settings.form_handler');
+        $formHandler = $this->getAddUpdateLanguageTranslationsFormHander();
         $addUpdateLanguageForm = $formHandler->getForm();
         $addUpdateLanguageForm->handleRequest($request);
 
         if ($addUpdateLanguageForm->isSubmitted()) {
             $data = $addUpdateLanguageForm->getData();
-            $isoCode = $data['add_update_language']['iso_localization_pack'];
+            $isoCode = $data['iso_localization_pack'];
 
             $languagePackImporter = $this->get('prestashop.adapter.language.pack.importer');
             $errors = $languagePackImporter->import($isoCode);
@@ -181,15 +187,15 @@ class TranslationsController extends FrameworkBundleAdminController
      */
     public function exportThemeLanguageAction(Request $request)
     {
-        $formHandler = $this->get('prestashop.admin.translations_settings.form_handler');
+        $formHandler = $this->getExportLanguageTranslationsFormHander();
         $exportThemeLanguageForm = $formHandler->getForm();
         $exportThemeLanguageForm->handleRequest($request);
 
         if ($exportThemeLanguageForm->isSubmitted()) {
             $data = $exportThemeLanguageForm->getData();
 
-            $themeName = $data['export_language']['theme_name'];
-            $isoCode = $data['export_language']['iso_code'];
+            $themeName = $data['theme_name'];
+            $isoCode = $data['iso_code'];
 
             $langRepository = $this->get('prestashop.core.admin.lang.repository');
             $locale = $langRepository->getLocaleByIsoCode($isoCode);
@@ -219,7 +225,7 @@ class TranslationsController extends FrameworkBundleAdminController
      */
     public function copyLanguageAction(Request $request)
     {
-        $formHandler = $this->get('prestashop.admin.translations_settings.form_handler');
+        $formHandler = $this->getCopyLanguageTranslationsFormHander();
         $form = $formHandler->getForm();
         $form->handleRequest($request);
 
@@ -227,10 +233,10 @@ class TranslationsController extends FrameworkBundleAdminController
             $languageCopier = $this->get('prestashop.adapter.language.copier');
             $data = $form->getData();
             $languageCopierConfig = new LanguageCopierConfig(
-                $data['copy_language']['from_theme'],
-                $data['copy_language']['from_language'],
-                $data['copy_language']['to_theme'],
-                $data['copy_language']['to_language']
+                $data['from_theme'],
+                $data['from_language'],
+                $data['to_theme'],
+                $data['to_language']
             );
 
             if ($errors = $languageCopier->copy($languageCopierConfig)) {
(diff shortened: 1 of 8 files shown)


src/PrestaShopBundle/Controller/Admin/TranslationsController.php
```diff
@@ -99,15 +99,20 @@ public function showSettingsAction(Request $request)
         $legacyController = $request->attributes->get('_legacy_controller');
         $legacyContext = $this->get('prestashop.adapter.legacy.context');
         $kpiRowFactory = $this->get('prestashop.core.kpi_row.factory.translations_page');
-        $formHandler = $this->get('prestashop.admin.translations_settings.form_handler');
-        $form = $formHandler->getForm();
+        $modifyTranslationsForm = $this->get('prestashop.admin.translations_settings.modify_translations.form_handler')->getForm();
```
jolelievre: Maybe you can add private getters for these handlers:

```php
private function getModyfyTranslationFormHander(): FormHandlerInterface
{
    return $this->get('prestashop.admin.translations_settings.modify_translations.form_handler');
}
```

This would the code of this controller and the corresponding action more readable?

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
```diff
@@ -28,12 +28,12 @@
...
-use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormFactoryInterface;
 
 final class TranslationsSettingsFormHandler implements FormHandlerInterface
 {
     /**
-     * @var FormBuilderInterface the form builder
+     * @var FormFactoryInterface the form builder
      */
     protected $formBuilder;
```
jolelievre:
```suggestion
    protected $formFactory;
```

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
```diff
@@ -50,23 +50,23 @@ final class TranslationsSettingsFormHandler implements FormHandlerInterface
     /**
      * @var array the list of Form Types
      */
-    protected $formTypes;
+    protected $form;
 
     /**
-     * @param FormBuilderInterface $formBuilder
+     * @param FormFactoryInterface $formBuilder
```
jolelievre:
```suggestion
     * @param FormFactoryInterface $formFactory
```

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
```diff
@@ -50,23 +50,23 @@ final class TranslationsSettingsFormHandler implements FormHandlerInterface
...
-        FormBuilderInterface $formBuilder,
+        FormFactoryInterface $formBuilder,
         HookDispatcherInterface $hookDispatcher,
-        array $formTypes,
-        $hookName
+        string $form,
+        string $hookName
     ) {
-        $this->formBuilder = $formBuilder;
+        $this->formFactory = $formBuilder;
```
jolelievre: Seems like you are using an nonexistent field `$this->formFactory`

src/PrestaShopBundle/EventListener/TokenizedUrlsListener.php
```diff
@@ -101,9 +101,14 @@ public function onKernelRequest(GetResponseEvent $event)
             }
         }
 
-        $token = urldecode($request->query->get('_token', false));
+        $token = false;
+        if ($request->query->has('_token')) {
+            $token = new CsrfToken($this->username, $request->query->get('_token'));
+        } elseif ($request->query->has('form')) {
```
jolelievre: I'm not at ease with this hard coded `'form'` name, where does it come from? Besides you don't test if `_token` is present

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
```diff
@@ -75,18 +75,16 @@ public function __construct(
...
-        foreach ($this->formTypes as $formName => $formType) {
-            $this->formBuilder->add($formName, $formType);
-        }
+        $form = $this->formFactory->createNamed('form', $this->form);
 
         $this->hookDispatcher->dispatchWithParameters(
             "action{$this->hookName}Form",
             [
-                'form_builder' => $this->formBuilder,
+                'form_builder' => $form,
```
jolelievre: This completely changes the hook behavior, quite a serious BC in my opinion since using a form builder is so much easier than the form itself Even though the hook is poorly named (it should have been `"action{$this->hookName}FormBuilder"` I'm not at ease with modifying this hook, isn't it possible to have a single FormType but still keep the formBuilder in the hook?

src/PrestaShopBundle/EventListener/TokenizedUrlsListener.php
```diff
@@ -101,9 +101,14 @@ public function onKernelRequest(GetResponseEvent $event)
...
+            $token = new CsrfToken('form', $request->query->get('form')['_token']);
```
eternoendless, line 108: Are you sure there's always a `_token` inside `form`?

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
jolelievre: The form builder is also used in hooks in other classes in the core `FormHandler` `(Indentifiable)FormBuilder` So it would not be consistent (and not handy) if translation handler is the only to use the form directly

jolelievre (dismissed):

src/PrestaShopBundle/EventListener/TokenizedUrlsListener.php
PierreRambaud: yes because of this https://github.com/PrestaShop/PrestaShop/blob/develop/app/config/config.yml#L30 but I'll add a new check to be sure the key exists

src/PrestaShopBundle/EventListener/TokenizedUrlsListener.php
PierreRambaud: This is our generic form name and also the same as Symfony form when form are created without a form factory / form builder: https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Form/FormHandler.php#L85
We currently don't have any choice because form are badly created, and csrf tokens are badly used :sweat_smile: 
Must be fixed in 1.7.8.0

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
PierreRambaud: Oops, forget to push files :sweat_smile:

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
PierreRambaud: The problem is, we don't have any form builder yet and we should never have used the `form_builder` here.  Ping @eternoendless

jolelievre: Maybe you can do something similar to the form builder for identifiable builder: https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Form/IdentifiableObject/Builder/FormBuilder.php#L120

```php
$formBuilder = $this->formFactory->createBuilder($formType, $data, $options);

        $this->hookDispatcher->dispatchWithParameters('action' . Container::camelize($formBuilder->getName()) . 'FormBuilderModifier', [
            'form_builder' => $formBuilder,
            'data' => &$data,
            'id' => $id,
        ]);

        return $formBuilder->getForm();
```

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
```diff
@@ -50,23 +50,23 @@ final class TranslationsSettingsFormHandler implements FormHandlerInterface
...
+     * @param string $form
      * @param string $hookName
      */
     public function __construct(
-        FormBuilderInterface $formBuilder,
+        FormFactoryInterface $formFactory,
         HookDispatcherInterface $hookDispatcher,
-        array $formTypes,
-        $hookName
+        string $form,
```
jolelievre, line 64: Maybe you should keep `$formType` ?

PierreRambaud: not needed anymore in this one, we already have an `array $formTypes` in the `FormHandler`

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
matks: @PierreRambaud We can duplicate Symfony `createNamedBuilder()` function and use it instead of `CreateNamed()` which allows us to call `createBuilder()` ourselves. This allows us to fix the issue with token AND keep sending the `form_builder` in the hook. Wdyt ?

PierreRambaud: Thanks @matks it's done :)

Progi1984 (dismissed):

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
```diff
@@ -75,18 +75,16 @@ public function __construct(
...
         $this->hookDispatcher->dispatchWithParameters(
             "action{$this->hookName}Form",
             [
-                'form_builder' => $this->formBuilder,
+                'form_builder' => $form_builder,
             ]
         );
 
-        return $this->formBuilder->getForm();
+        return $form->getForm();
```
matks: Sorry I think this variable is undefined 😅

src/PrestaShopBundle/Form/Admin/Improve/International/Translations/TranslationsSettingsFormHandler.php
```diff
@@ -75,18 +75,16 @@ public function __construct(
      */
     public function getForm()
     {
-        foreach ($this->formTypes as $formName => $formType) {
-            $this->formBuilder->add($formName, $formType);
-        }
+        $form_builder = $this->formFactory->createNamedBuilder('form', $this->form);
```
matks: Do we use snake_case now ? 😅

PierreRambaud: You saw nothing!

Progi1984 (dismissed):

Progi1984:
Travis is fixed

matks (dismissed):

SD1982:
Ping @PierreRambaud  I have this error  at the first step ( Go to BO > international > translations )

#PR17482 Allow token for TokenisedUrlsListener class and in form using the method=GET attribute after pr step 1 of reproduce procedure

PierreRambaud:
@SD1982 My bad, forget to commit the interface :thinking:

Robin-Fischer-PS:
Need reviews before QA :)

matthieu-rolland (approved):

atomiix (approved):

khouloudbelguith:
Hi @PierreRambaud,

It is Ok.
https://drive.google.com/file/d/1oTBsxe_lcHxgITY-KIHEv1YxAc6FRDLf/view

Thanks!

Progi1984:
Thanks @PierreRambaud

PR #17498, opened 2020-02-04
[Blocked] Allow injection of extra variables inside twig

### LINKED ISSUE #13462: [SF Migration] Allow injection of extra variables inside twig
Usecase: developers can override twig templates in order to modify a page display. For example if I prefer flat design over material design, I can re-style my BO with my design preference.

What happens if, while re-styling one template, I want to display an additional information using a variable that was not passed to Twig by the Controller ? As Symfony controllers are not overridable in PS 1.7, I might have to create a new Controller even though I just wanted to add 1 single extra variable. For example I just wanted to replace
```
return $this->render('@PrestaShop/Admin/Configure/AdvancedParameters/LogsPage/index.html.twig', [
            'layoutHeaderToolbarBtn' => [],
            'layoutTitle' => $this->trans('Logs', 'Admin.Navigation.Menu'),
            'requireAddonsSearch' => true,
            'grid' => $this->presentGrid($grid),
        ]);
```
by
```
return $this->render('@PrestaShop/Admin/Configure/AdvancedParameters/LogsPage/index.html.twig', [
            'layoutHeaderToolbarBtn' => [],
            'layoutTitle' => $this->trans('Logs', 'Admin.Navigation.Menu'),
            'requireAddonsSearch' => true,
            'grid' => $this->presentGrid($grid),
            'extra_var' => $myExtraVariable
        ]);
```

Possible solutions:
- Declare a twig global variable, but this impacts all templates, not just the one I wanted (suggestion by @mickaelandrieu)
- Create a dedicated Twig Extension to be used in my template (suggestion by @mickaelandrieu)
- Add a listener to KernelEVENTS::CONTROLLER and KernelEvents::VIEW and perform a hook call there to enable variables injection

tdavidsonas88:
| Questions     | Answers
| Branch?       | develop
| Description?  | Module developers want to display an additional information using a variable that was not passed to Twig by the Controller. This PR gives the possibility to hook additional parameter directly to the controller twig template.
| Type?         | improvement
| Category?     | BO
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/13462
| How to test?  | register a hook 'actionBackOfficeControllerSetVariables' in the PrestaShop module. Write hook function 'hookActionBackOfficeControllerSetVariables' returning the extra variable. Use the injected variable inside any migrated rendered BO controller twig file {{ moduleName }} or {% moduleName %}

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17498)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
+++ b/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
@@ -33,6 +33,7 @@ use PrestaShop\PrestaShop\Core\ConfigurationInterface;
 use PrestaShop\PrestaShop\Core\Grid\GridInterface;
 use PrestaShop\PrestaShop\Core\Hook\HookDispatcherInterface;
 use PrestaShop\PrestaShop\Core\Hook\RenderedHook;
+use PrestaShop\PrestaShop\Core\Hook\RenderedHookInterface;
 use PrestaShop\PrestaShop\Core\Localization\Locale;
 use PrestaShop\PrestaShop\Core\Localization\Locale\Repository as LocaleRepository;
 use PrestaShop\PrestaShop\Core\Module\Exception\ModuleErrorInterface;
@@ -531,10 +532,10 @@ class FrameworkBundleAdminController extends Controller
     {
         /** @var HookDispatcherInterface $hookDispatcher */
         $hookDispatcher = $this->get('prestashop.core.hook.dispatcher');
-        /** @var RenderedHook $renderedHook */
+        /** @var RenderedHookInterface $renderedHook */
         $renderedHook = $hookDispatcher->dispatchRenderingWithParameters('actionBackOfficeControllerSetVariables');
 
-        $parameters = array_merge($renderedHook->getContent(), $parameters);
+        $parameters = array_merge($parameters, $renderedHook->getContent());
 
         return parent::render($view, $parameters, $response);
     }
--- a/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
+++ b/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
@@ -31,6 +31,8 @@ use PrestaShop\PrestaShop\Adapter\Configuration;
 use PrestaShop\PrestaShop\Adapter\Shop\Context;
 use PrestaShop\PrestaShop\Core\ConfigurationInterface;
 use PrestaShop\PrestaShop\Core\Grid\GridInterface;
+use PrestaShop\PrestaShop\Core\Hook\HookDispatcherInterface;
+use PrestaShop\PrestaShop\Core\Hook\RenderedHook;
 use PrestaShop\PrestaShop\Core\Localization\Locale;
 use PrestaShop\PrestaShop\Core\Localization\Locale\Repository as LocaleRepository;
 use PrestaShop\PrestaShop\Core\Module\Exception\ModuleErrorInterface;
@@ -523,5 +525,17 @@ class FrameworkBundleAdminController extends Controller
             $exceptionCode,
             $e->getMessage()
         );
+    }
+
+    protected function render($view, array $parameters = [], Response $response = null)
+    {
+        /** @var HookDispatcherInterface $hookDispatcher */
+        $hookDispatcher = $this->get('prestashop.core.hook.dispatcher');
+        /** @var RenderedHook $renderedHook */
+        $renderedHook = $hookDispatcher->dispatchRenderingWithParameters('actionBackOfficeControllerSetVariables');
+
+        $parameters = array_merge($renderedHook->getContent(), $parameters);
+
+        return parent::render($view, $parameters, $response);
     }
 }
--- a/install-dev/data/xml/hook.xml
+++ b/install-dev/data/xml/hook.xml
@@ -2145,5 +2145,10 @@
       <title>Admin Order Buttons</title>
       <description>This hook is used to generate the buttons collection on the order view page thanks (see ActionsBarButtonsCollection)</description>
     </hook>
+    <hook id="actionBackOfficeControllerSetVariables">
+      <name>actionBackOfficeControllerSetVariables</name>
+      <title>Admin extra variables inside twig</title>
+      <description>This hook injects extra variables inside twig via controller parameters</description>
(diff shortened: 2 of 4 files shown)


tdavidsonas88:
Returned array from the module which is indexed bymodule name like: moduleName => array (
  'exVar1' => 'testvar1',
)
Thinking if it could be better

atomiix (dismissed):

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
```diff
@@ -524,4 +526,16 @@ protected function getErrorMessageForException(Exception $e, array $messages)
             $e->getMessage()
         );
     }
+
+    protected function render($view, array $parameters = [], Response $response = null)
+    {
+        /** @var HookDispatcherInterface $hookDispatcher */
+        $hookDispatcher = $this->get('prestashop.core.hook.dispatcher');
+        /** @var RenderedHook $renderedHook */
```
jolelievre: You can use `RenderedHookInterface` instead of the concrete class

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
```diff
@@ -524,4 +526,16 @@ protected function getErrorMessageForException(Exception $e, array $messages)
...
+        $renderedHook = $hookDispatcher->dispatchRenderingWithParameters('actionBackOfficeControllerSetVariables');
+
+        $parameters = array_merge($renderedHook->getContent(), $parameters);
```
jolelievre:
```suggestion
        $parameters = array_merge($parameters, $renderedHook->getContent());
```
I'm not sure but I think the idea is to allow the modules override the core variables, if so then the hook content must be the final argument of array merge

jolelievre (changes requested):

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
tdavidsonas88: As I understood the idea is to add extra variable/s to the twig along with the existing ones (passed from Controller)

matks: Mmmm ... indeed @jolelievre's idea is good. And we could even pass `$parameters` to the hook to allow it to enrich the data or compute data from it !

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
jolelievre: Also a thing I'm not sure of, but I thought a `RenderedHook` is used for display hooks (as its name would suggest), so it will only contain a string to display whereas here we are aiming for an action hook which would return an array of variables

tdavidsonas88:
> Returned array from the module which is indexed bymodule name like: moduleName => array (
> 'exVar1' => 'testvar1',
> )
> Thinking if it could be better

Tested it with both cases when array or string is returned - and it might be that the current state of code is quite good as it accept both string and array. Just it's important notice that the extra variables are wrapped by the module name which is also nice because variables might come from different modules, :slightly_smiling_face:

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
tdavidsonas88: also have a small comment at the bottom about the structure the variables are returned. If we take @jolelievre's idea of overriding the variables then we need some changes in the code I guess. Separated parsing for array and string case;

jolelievre:
> > Returned array from the module which is indexed bymodule name like: moduleName => array (
> > 'exVar1' => 'testvar1',
> > )
> > Thinking if it could be better
> 
> Tested it with both cases when array or string is returned - and it might be that the current state of code is quite good as it accept both string and array. Just it's important notice that the extra variables are wrapped by the module name which is also nice because variables might come from different modules, 🙂

I looked into the details of the hook dispatcher, so I think this array divided by module is normal for a rendering hook: https://github.com/PrestaShop/PrestaShop/blob/develop/src/Adapter/Hook/HookDispatcher.php#L123

But it confirms what I was fearing, I don't think we should use `dispatchRenderingWithParameters` for an action hook, it is dedicated to display hooks where the module displays a part of the template (which is caught by `ob_start/ob_get_clean` and then formatted into an array

For an action hook like here we should use `$hookDispatcher->dispatch('actionBackOfficeControllerSetVariables')` although I'm not sure exactly how the modules can then set extra field in the array..

Progi1984 (dismissed):

tdavidsonas88:
If to use the `$hookDispatcher->dispatch('actionBackOfficeControllerSetVariables')` then we can add listener to it, sth. like:

```
        $dispatcher = new EventDispatcher();
        $dispatcher->addListener(KernelEvents::VIEW, function ($event) {
            $event->setResponse(new Response($event->getControllerResult()));
        });
```
and then try to add extra parameters there. Maybe this idea is worth trying it out.

matks:
> If to use the `$hookDispatcher->dispatch('actionBackOfficeControllerSetVariables')` then we can add listener to it, sth. like:
> 
> ```
>         $dispatcher = new EventDispatcher();
>         $dispatcher->addListener(KernelEvents::VIEW, function ($event) {
>             $event->setResponse(new Response($event->getControllerResult()));
>         });
> ```
> 
> and then try to add extra parameters there. Maybe this idea is worth trying it out.

It will not work I think. Hooking system has specific behavior that Symfony listener does not handle.

Legacy hook system has been wrapped up into Symfony listener but this is only an extra layer.

Indeed the HookDispatcher, which wraps legacy Hook system, is unfortunately not completely implemented 😞 we might need to fix the remaining issues.

tdavidsonas88:
By the way, the current implementation works only on the migrated Controllers; Legacy ones do not tirgger this - but probably that is what was expected as the Legacy ones use Smarty

matks:
This PR is blocked by https://github.com/PrestaShop/PrestaShop/issues/17513

PierreRambaud (dismissed):

matks:
@tdavidsonas88 I close this PR as it currently appears in our backlog of PRs and until I (or someone else) find some time to finish it, it won't move 😞  .

Issue #9702, opened 2018-08-22
[BOOM-5101] Voucher free shipping show 0.00% instead of free sipping


prestonBot:
## SPECS

Cart rule with free shipping only: Display free shipping instead of 0,00%
Cart rule with free shipping + discount (amount or %): Display price of the carrier + Discount amount added
Cart rule with % discount: Display the discount amount            

![image](https://user-images.githubusercontent.com/13449658/67876296-f5f8b100-fb37-11e9-81b2-076c75f642c0.png)

You should always see the same value between the "discount(s)" line and the "cart rule" line if there is only one cart rule in the cart

----------------------
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-5101).

- _**Reporter:**_ PN
- _**Created at:**_ Thu, 15 Mar 2018 09:27:45 +0100

<p>Hello,</p>

<p> </p>

<p><b>Type of view discounts in cart when is discount action set as free shipping. Why is showed: -0.00%? Better will be show free shipping instead of -0.00%.</b></p>

<p>Thank you</p>

<p>Best regards</p>

<p>PN</p>
        
* How to reproduce the issue ?

Catalog, Discounts, Cart Rules
<br/>

<br/>
for:
<br/>
1) Set Highlight: YES
<br/>
2) Leave field code blank
<br/>
3) Set actions: Free shipping: YES, none

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-5101#comment-164244).

- _**Author:**_ khouloud.belguith
- _**Created at:**_ Thu, 15 Mar 2018 10:42:20 +0100

<p>Salut,</p>

<p>Que pensez-vous de sa remarque ?<br/>
J'ai pu reproduire le problème avec 1.7.3.0<br/>
 <span class="image-wrap" style=""><a id="66034_thumb" href="http://forge.prestashop.com/secure/attachment/66034/66034_1730.png" title="1730.png" file-preview-type="image" file-preview-id="66034" file-preview-title="1730.png"><img src="http://forge.prestashop.com/secure/thumbnail/66034/_thumb_66034.png" style="border: 0px solid black" /></a></span> </p>

<p>Merci!</p>

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-5101#comment-164253).

[...]

PR #17502, opened 2020-02-04
Fix cart voucher values in cart summary and checkout summary

### LINKED ISSUE #16842: FO - The promo value is not correct in the cart summary & checkout summary
**Describe the bug**
The promo value is not correct in the cart summary & checkout summary

**To Reproduce**
1. Go to the BO => Catalog => Discounts => Create a new cart rule
1.1 has no code
1.2 there is no condition
1.3 As an acrtion=> Apply a discount Percent = 100% => Apply a discount to:"Cheapest product"
image

2. Go to the FO => Add to the cart demo_11 & demo_6
3. The cart rule will be applied to the demo_11
4. Go to the cart page
5. See error
image

With PS1.7.6.2 was ok
image

**Additional information**
PrestaShop version: develop
PHP version: 7.1

Edit:
**Specs:**
The discount should display **the value of the discount** and nor the percentage.
And if the value can't be retrieved for a technical reason, it should display the percentage.

arouiadib:
| Questions     | Answers
| Branch?       | 1.7.7.x 
| Description?  | This PR fixes the issues of the amounts of promo codes in cart summary and checkout summary.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? |  Fixes #16842 
| How to test?  | Check the amount of reduction of each cart voucher, now it depends of the actions of the cart voucher and is not equal to all products total.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17502)


MatShir:
@arouiadib: "the PR I created is a "white-flag" PR for me, because I have spend hours trying to create a cleaner and better solution to no avail. My one-line PR uses an existing method getContextualValue in CartRule.php. Does it go against migration principles? Better not to use methods from that class? Am I right? Additionally, the processing in that mothod almost exists in CartRuleCalculator also, but not easy to re-use ( as I see it addresses another purpose IMO, and re-factoring should be done carefully).

Another possible solution is use quantity from product list in Presenter, but it would be a second duplication for that processing, and maybe is not the role of cartPresenter?!

I think for a cleaner solution is to make a re-factoring in CartRuleCalculator. But more experienced people from the community will suggest me the right direction."

@PrestaShop/prestashop-core-developers What do you think ?

marionf:
For @PrestaShop/prestashop-qa 
Could you verify if this PR brings no regressions with a free shipping cart rule (https://github.com/PrestaShop/PrestaShop/issues/9702) ?

src/Adapter/Presenter/Cart/CartPresenter.php
```diff
@@ -640,6 +637,16 @@ private function cartVoucherHasAmountReduction($cartVoucher)
...
     }
 
+    /**
+     * @param array $cartVoucher
+     *
+     * @return bool
+     */
+    private function cartVoucherHasGiftProductReduction($cartVoucher)
+    {
+        return array_key_exists('gift_product', $cartVoucher) && $cartVoucher['gift_product'];
```
PierreRambaud:
```suggestion
        return !empty($cartVoucher['gift_product']);
```

Is it really needed to add a method only for this check?

arouiadib: Not really needed, but just to follow suit with current code.

arouiadib: I see what you mean now. Well, the only reason is to make the `if` statement more self-explanatory  ( the naming `cartVoucherHasGiftProductReduction` inside `if` emphasizes the fact that we need the 3 conditions that have "similar" weight). But you know, I am still learning the codebase so I can't tell which one is more suitable until you directly tell me ;-)

Progi1984:
@arouiadib Hi Could you fix the PHPCSFixer ? Thanks

arouiadib:
@Progi1984 , I tried today to no avail. I will retry ASAP. Thank you very much for your time.

arouiadib:
Hi @Progi1984 , I finally managed to fix code styles issue.  Messed up the branch a little bit :baby:  
Thank you very much for your time.

Progi1984 (dismissed):

matks (dismissed):

src/Adapter/Presenter/Cart/CartPresenter.php
PierreRambaud: In fact, this is the same thing.
```
!empty($cartVoucher['gift_product']); // => Not false, 0, '', null, []
array_key_exists('gift_product', $cartVoucher) && $cartVoucher['gift_product'] // => Check if the key exists in the table and is not null, false, 0, '', etc
```

This is the same check, but it's faster due to only one call

arouiadib: Thank you very much :). 
If I understand well (because I really have a slow brain these days), we can we go further by adding [one another commit](https://github.com/PrestaShop/PrestaShop/pull/17502/commits/099a863b16a8a3405af20d28269b483ccb015cb0).

SD1982:
hi @arouiadib  Can you please resolve the conflict ?

Progi1984:
Hello @arouiadib, thanks for your contribution, would you mind to rebase your pull request? Thanks

arouiadib:
Hello @SD1982 @Progi1984 , I think this PR is no longer taking the direction that the code source is taking. I will be rethinking things to bring a more appropriate change. Thank you very much guys for all your time.

Progi1984 (dismissed):

matthieu-rolland (dismissed):

matthieu-rolland:
@arouiadib you're almost done, one last conflict to fix ;)

Progi1984 (approved):

atomiix (approved):

jolelievre:
Hi @PrestaShop/prestashop-qa 

In addition to what @marionf said about checking no regression regarding the issue #9702 about free shipping cart rule

You should also check this issue https://github.com/PrestaShop/PrestaShop/issues/18065 since lots of the code that fixed this in https://github.com/PrestaShop/PrestaShop/pull/18144 has been replaced

So it would seem all this calculation was useless because it is present in `$cartVoucher` which would be great! @Progi1984 since you approved this PR and coded the previous one i guess it's ok but since we don't have much automated test on this code better safe than sorry ^^

SD1982:
LGTM I have also checked for #9702 and #18065 as suggested by @jolelievre and i haven't notified any unwanted behaviour or regression ! Thanks @arouiadib !

Progi1984:
Thanks @arouiadib

arouiadib:
Thank you all guys.

ComonSoft:
Hi, this fix the issue #16842 but the reduction summury is still false in cart summury see screenshot
image

MatShir:
Hi @ComonSoft,
Could you create a new issue with the detailed steps to helps us reproduce it?
We will be happy to test and check it out.

ComonSoft:
@MatShir this is the same issue as describre in ticket #19090 The solution proposed here only solve the amount of each voucher in cart summury but not the sum of all voucher. 1rst circle in my screen shot.

MatShir:
Okay thanks for your feedbacks @ComonSoft, the bug is placed on the Roadmap 1.7.8 ;-)

PR #17526, opened 2020-02-06
Fix customer address form

### LINKED ISSUE #17032: Adresses page - Required Fields cannot be selected which occurs issues in the BO & the FO

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | After trying to fix addresses required fields and format relation in  https://github.com/PrestaShop/PrestaShop/pull/17293 we found out it needs more work than expected, so this pr only fixes mandatory bugs and leaves the BO address form simple, without considering the required fields(in customer->addresses settings) or Address format (in country settings). The above mentioned relation of addressFormat and fields is planned to be reworked in 1.7.8
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #17282, finishes #10558, improve #17549, fix #17032 #17549
| How to test?  | Customer address form in BO now should always render all fields as optional (= you can leave blank). Only firstname, lastname, address1, country and city fields are mandatory. We dont try to guess what fields are required or not based on country on whatever. We keep it simple 😄  **NOTE** I had to leave some of required fields check in BO anyway to ensure that it works like before and nothing breaks in FO.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17526)


matks (dismissed):

src/Adapter/Form/ChoiceProvider/CountryStateByIdChoiceProvider.php
```diff
@@ -47,14 +48,21 @@ public function getChoices(array $options)
...
-            $states = State::getStatesByIdCountry($resolvedOptions['id_country'], $resolvedOptions['only_active']);
+            $countryId = $resolvedOptions['id_country'];
+            $countryHasStates = (new Country($countryId))->contains_states;
+
+            if (!$countryHasStates) {
+                return [];
+            }
+
+            $states = State::getStatesByIdCountry($countryId, $resolvedOptions['only_active']);
             $choices = [];
```
PierreRambaud: `$choices` must be decalred above. If there is an exception in `new Country`, `$choices` will be undefined

PierreRambaud (changes requested):
see my comment

src/Adapter/Address/CommandHandler/SetRequiredFieldsForAddressHandler.php
```diff
@@ -46,7 +46,7 @@ final class SetRequiredFieldsForAddressHandler implements SetRequiredFieldsForAd
      */
     public function handle(SetRequiredFieldsForAddressCommand $command)
     {
-        $address = new Address();
+        $address = new CustomerAddress();
```
jolelievre, line 49: One question, how does it work for brand or manufacturer address? Because I believe they all share the same `Address` class But using `CustomerAddress ` here means it's only dedicated to customer addresses, but not the other ones? Then shouldn't the handler name and namespace reflect this limitation?
=> `SetRequiredFieldsForCustomerAddressHandler::handle(SetRequiredFieldsForCustomerAddressCommand $command)`

src/Adapter/Form/ChoiceProvider/CountryStateByIdChoiceProvider.php
jolelievre: I get what you mean it would be cleaner, but it shoudn't be a problem here since an exception is thrown in the catch block It will never reach the `return $choices;`

jolelievre (commented):
Nothing blocking for me But I always have concerns regarding this "Address" feature, it is mixed/common with brand addresses, manufacturer addresses But each one has been developed separately, so we fix bugs separately as well And the improvement are never ported from one to another
Since you said some improvement should be made in 178 about the customer address, maybe it would be worth refactoring all the address management so all rules are clear everywhere?

src/Adapter/Form/ChoiceProvider/CountryStateByIdChoiceProvider.php
PierreRambaud: What if it's not a `PrestaShopException` :thinking:

jolelievre: Then it's not caught by this class and either the process crashes or it's caught by something higher In both cases it still won't reach the `return`

PierreRambaud: Could you just move the `return $choices` after the foreach to have a cleaner code :D :pray:

src/Adapter/Address/CommandHandler/SetRequiredFieldsForAddressHandler.php
zuk3975: Only customer address is affected :man_shrugging:
Maybe as you say it could be reworked in 178.
It came from legacy that in db required_fields are saved by classname of the object (in this case CustomerAddress. By providing Address it doesn't work at all).

zuk3975:
> see my comment

moved var declaration up before try-catch

zuk3975:
Related issue: https://github.com/PrestaShop/PrestaShop/issues/17647

zuk3975:
@jolelievre, @PierreRambaud addressed.
also i put back checks for required DNI and POSTCODE fields in js 
31f6e85 . Somehow those checks are essential only for these 2 fields so those cannot be always optional, but no other configurable requiredFields need it :man_shrugging:

PierreRambaud (dismissed):

zuk3975:
Reopen to restart travis

PierreRambaud (dismissed):

zuk3975:
reopen last try for prettyCI

matthieu-rolland (dismissed):

matks (dismissed):

Robin-Fischer-PS:
@zuk3975 Could you resolve conflicts ? (QA is back :D )

zuk3975:
> @zuk3975 Could you resolve conflicts ? (QA is back :D )

done

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -89,10 +86,14 @@ public function __construct(
...
     {
         $data = $builder->getData();
-        $requiredFields = $data['required_fields'];
         $countryId = 0 !== $data['id_country'] ? $data['id_country'] : $this->contextCountryId;
         $stateChoices = $this->stateChoiceProvider->getChoices(['id_country' => $countryId]);
 
+        $hideStates = false;
+        if (empty($stateChoices)) {
+            $hideStates = true;
+        }
```
Progi1984:
```suggestion
        $hideStates = (empty($stateChoices));
```

Progi1984 (dismissed):

PierreRambaud (dismissed):

NeOMakinG:
Is this ready to QA then?

zuk3975:
> Is this ready to QA then?

yes lets go :smile:

Robin-Fischer-PS:
Hi !

I've two issues : 

> Only firstname, lastname, address1, country and city fields are mandatory. We dont try to guess what fields are required or not based on country on whatever.
- When installing Spain for example, I have all the required fields of the country (DNI, First Name, Last Name, Address, Zip Code, City, Country, State). If I understand well your "how to test", we should not have DNI and Zip Code set as required.

- On FO, when having required fields such as "DNI" or "Mobile Phone", which are not displayed by default, Address are marked as not valid, but the required fields do not appear in Edit Address, so it's impossible to make an order.
**=> This issue was already there in Legacy, so it's OK for me if we leave it there in 177**

Screenshot : Required fields DNI on BO : 
pr17526 DNI required

Screenshot : Address incomplete on FO : 
pr17526 address incomplete

Screenshot : DNI and mobile phone field not displayed in FO
pr17526 missing field address checkout

Thanks !

zuk3975:
@Robin-Fischer-PS 

> Hi !
> 
> I've two issues :
> 
> > Only firstname, lastname, address1, country and city fields are mandatory. We dont try to guess what fields are required or not based on country on whatever.
> 
> * When installing Spain for example, I have all the required fields of the country (DNI, First Name, Last Name, Address, Zip Code, City, Country, State). If I understand well your "how to test", we should not have DNI and Zip Code set as required.
> * On FO, when having required fields such as "DNI" or "Mobile Phone", which are not displayed by default, Address are marked as not valid, but the required fields do not appear in Edit Address, so it's impossible to make an order.
> ...
> 
> Thanks !

Hi, sorry for the first one, I just forgot to update description. I had to leave the legacy behavior to avoid FO issues. So its kinda how it should work until we fix whole addresses stuff in upcoming versions.

zuk3975:
:heavy_check_mark:

Progi1984 (dismissed):

PierreRambaud (dismissed):

Robin-Fischer-PS:
I tested again, with France and Spain addresses, both BO & FO are OK (except the "issue" which was already there in 176) !

Thanks @zuk3975 :)

zuk3975:
> I tested again, with France and Spain addresses, both BO & FO are OK (except the "issue" which was already there in 176) !
> 
> Thanks @zuk3975 :)

Thank you, doing a great job!

Progi1984:
Hello @zuk3975, thanks for your contribution, would you mind to rebase your pull request? Thanks

Progi1984 (approved):

PierreRambaud (approved):

PierreRambaud:
Thanks @zuk3975

matks (approved):

PR #17535, opened 2020-02-06
Fix deleted customer's order preview/details

### LINKED ISSUE #17311: Error while trying to preview or view an order of a deleted customer
**Describe the bug**
Error while trying to preview or view an order of a deleted customer

**To Reproduce**
1. Go to BO > Customers > Customers
2. Delete John Doe
3. Go on the migrated order list
4. Try to click on the preview, nothing is happening
5. Try to view the order 
6. See errors

**Screenshots**
**Error in the console after clicking on the preview**
capture d'écran_2301

**Error after clicking on view**
capture d'écran_2299

**Additional information**
PrestaShop version: 1.7.7.x
PHP version: N/A

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When a customer was hard deleted, you couldn't see his orders preview and details. This PR fixes it.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17311
| How to test?  | 1. Go to BO > Customers > Customers<br>2. Delete John Doe<br>3. Go on the migrated order list<br>4. See this [issue comment](https://github.com/PrestaShop/PrestaShop/issues/17311#issuecomment-581481515) to see what should be displayed on the list and on the order details.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17535)

### DIFF
--- a/src/Core/Grid/Column/Type/DisableableLinkColumn.php
+++ b/src/Core/Grid/Column/Type/DisableableLinkColumn.php
@@ -1,0 +1,92 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Grid\Column\Type;
+
+use PrestaShop\PrestaShop\Core\Grid\Column\AbstractColumn;
+use PrestaShop\PrestaShop\Core\Grid\Column\Type\Common\LinkColumn;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+final class DisableableLinkColumn extends AbstractColumn
+{
+    /**
+     * @var LinkColumn
+     */
+    private $linkColumn;
+
+    public function __construct($id)
+    {
+        parent::__construct($id);
+        $this->linkColumn = new LinkColumn($id);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getType()
+    {
+        return 'disableable_link';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver
+            ->setRequired(['disabled_field'])
+            ->setAllowedTypes('disabled_field', ['string', 'null'])
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setOptions(array $options)
+    {
+        $disabledOptions = [];
+
+        if (isset($options['disabled_field'])) {
+            $disabledOptions['disabled_field'] = $options['disabled_field'];
+            unset($options['disabled_field']);
+        }
+
+        $this->linkColumn->setOptions($options);
+        parent::setOptions($disabledOptions);
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getOptions()
+    {
+        return array_merge($this->linkColumn->getOptions(), parent::getOptions());
+    }
+}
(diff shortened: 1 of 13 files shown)


src/Core/Grid/Column/Type/DisableableLinkColumn.php
```diff
@@ -0,0 +1,92 @@
...
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getOptions()
+    {
+        return array_merge($this->linkColumn->getOptions(), parent::getOptions());
```
Progi1984, line 90:
```suggestion
        return array_merge(parent::getOptions(), $this->linkColumn->getOptions());
```

Progi1984 (commented):
Feedback

src/Core/Grid/Column/Type/DisableableLinkColumn.php
Progi1984: Cancelled

Progi1984 (dismissed):

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/Modal/view_all_messages_modal.html.twig
```diff
@@ -30,7 +30,12 @@
         <div class="row">
           <div class="col">
             <h3 class="card-header-title">
-              {{ 'Message history with %name%'|trans({'%name%': orderForViewing.customer.firstName ~ ' ' ~ orderForViewing.customer.lastName}, 'Admin.Global') }} ({{ orderForViewing.messages.messages|length }})
+              {% if orderForViewing.customer is not null %}
+                {{ 'Message history with %name%'|trans({'%name%': orderForViewing.customer.firstName ~ ' ' ~ orderForViewing.customer.lastName}, 'Admin.Global') }}
```
LouiseBonnard, line 34: Is this string necessary? What will %name% indicate? We are talking about a deleted customer, we are not supposed to have its name anymore.

src/Core/Grid/Column/Type/DisableableLinkColumn.php
```diff
@@ -0,0 +1,92 @@
...
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Grid\Column\Type;
+
+use PrestaShop\PrestaShop\Core\Grid\Column\AbstractColumn;
+use PrestaShop\PrestaShop\Core\Grid\Column\Type\Common\LinkColumn;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+final class DisableableLinkColumn extends AbstractColumn
```
matks, line 35: It's not necessarily an issue but ... did you consider modifying LinkColumn to add an option, rather than creating a new class ? If yes, what are the reasoning behind this choice ? Just curious

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/Modal/view_all_messages_modal.html.twig
atomiix: Actually, "Message history with %name%" is the sentence when the customer is not deleted. %name% will "firstname lastname". When the customer is deleted, the sentence will be "Message history" as you can see just below. 🙂

src/Core/Grid/Column/Type/DisableableLinkColumn.php
atomiix: @matks The main purpose was to keep the `LinkColumn` as simple as possible knowing that `DisableableLinkColumn` is only used once (for now at least...). But your suggestion seems good to me too. WDYT?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/customer.html.twig
```diff
@@ -49,8 +50,14 @@
             {{ 'View full details'|trans({}, 'Admin.Actions') }}
           </a>
         </div>
+        {% else %}
+        <div class="col">
+          <h2 class="mb-0">{{ 'Deleted customer'|trans({}, 'Admin.Global') }}</h2>
```
NeOMakinG, line 55: Could we think about a better workflow when we've to disable margin-bottom ? Using class like mb-0 is against the CSS principle :/

NeOMakinG: I would prefer something like customer-title-delete and adding a margin-bottom: 0; somewhere in the css, or disable margin-bottom on h2 inside a certain class...

What do you think @PierreRambaud

PierreRambaud: Agree with @NeOMakinG It's also easier for someone to do something more specific with a custom class than a class like `mb-0`

atomiix: @PierreRambaud @NeOMakinG Maybe we can open another issue for that because all that page is build like that and I was just reusing a part to adapt it when the customer doesn't exist. WDYT?

NeOMakinG: Ok, but I think we should talk about it with the team before next migrations

Progi1984 (approved):

NeOMakinG (approved):

SD1982:
The pull request correctly fix #17311  : 

**Order list after deleting a customer :** 

Fix deleted customer's order preview details #17535 after PR1

**Order view after deleting a customer :** 

Fix deleted customer's order preview details #17535 after PR2

PR #17547, opened 2020-02-07
Fix grids, when search finds no result we still display filters and reset button

matks:
This reverts commit 8647e5b514df931e06a9426a52fd573b1ee42439.

| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Reverts 8647e5b514df931e06a9426a52fd573b1ee42439. Fix grids, when search finds no result we still display filters and reset button
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/17548
| How to test?  | See ticket

Downside of this PR: when an empty grid (= no data, like no brands or no suppliers) is displayed, we display filters even though there is no data to filter. But it's better this way.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17547)

### DIFF
--- a/src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Blocks/Table/filters_row.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Blocks/Table/filters_row.html.twig
@@ -23,7 +23,7 @@
  * International Registered Trademark & Property of PrestaShop SA
  *#}
 
-<tr class="column-filters {% if 0 == grid.data.records_total or (grid.filters is empty and grid.actions.bulk|length < 0) %}d-none{% endif %}">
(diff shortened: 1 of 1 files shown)


matthieu-rolland (approved):

atomiix (approved):

Robin-Fischer-PS:
Hi @matks ! As discussed together, one bug linked to the PR that created the issue : 
When sorting a tab which has bulk action and position selector, the display of the tab is broken : 
pr17547 bad display

Do you think this can be fixed is this PR ?

**EDIT** : Will be fixed in another PR, see issue :
https://github.com/PrestaShop/PrestaShop/issues/17563

Issue #16017, opened 2019-10-18
Call to private method when trying to update shop language or e-mails themes


Klemart3D:
**Describe the bug**

I tried to update a language setting (French) from BO: International > translations > Add / Update a language and click button. It returned me:

```
Call to private Pelago\Emogrifier\HtmlProcessor\AbstractHtmlProcessor::__construct() from context 'PrestaShop\PrestaShop\Core\MailTemplate\Transformation\CSSInlineTransformation'

Uncaught PHP Exception Symfony\Component\Debug\Exception\FatalThrowableError: "Call to private Pelago\Emogrifier\HtmlProcessor\AbstractHtmlProcessor::__construct() from context 'PrestaShop\PrestaShop\Core\MailTemplate\Transformation\CSSInlineTransformation'" at /var/www/c-piscine/html/src/Core/MailTemplate/Transformation/CSSInlineTransformation.php line 70
```

**Screenshots**
![bug-language-update](https://user-images.githubusercontent.com/7419525/67084750-df4c7480-f19d-11e9-8a86-9b5b808ecbab.png)

![bug-language-update2](https://user-images.githubusercontent.com/7419525/67085133-8fba7880-f19e-11e9-9397-823f8b28f9b9.png)

**Additional information**
PrestaShop version: 1.7.6.1
PHP version: 7.2

khouloudbelguith:
Hi @Klemart3D,

Thanks for your report.
I did not manage to reproduce the issue with PS1.7.6.1.
What is the exact version of the ps_emailsubscription module did you use?

Thanks!

Klemart3D:
Hi @khouloudbelguith,

Thanks for tour answer.

Module ps_emailsubscription:

**Inscription à la newsletter**
**v2.5.1** - par PrestaShop 

And when I'm trying to Configure this module I have the following:
```
Notice à la ligne 74 du fichier /var/www/prestashop/html/src/Adapter/Routing/LegacyHelperLinkBuilder.php
[8] Undefined index: id_merged
```

![Capture d’écran de 2019-10-18 12-12-18](https://user-images.githubusercontent.com/7419525/67086274-ac57b000-f1a0-11e9-915e-ef52a17b11f0.png)

khouloudbelguith:
@Klemart3D, this is adifferent issue.
[...]

PR #17554, opened 2020-02-07
fix pelago/emogrifier version spech in composer.json

davideapvd:
I think the composer.json package version should be forced to be 2.1.1 or this bug could happen, see https://github.com/PrestaShop/PrestaShop/issues/16017

composer upgrade will bring 2.2.0

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17554)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
         "nikic/php-parser": "^4.0",
         "paragonie/random_compat": "^2.0",
         "pear/archive_tar": "^1.4.6",
-        "pelago/emogrifier": "^2.1",
+        "pelago/emogrifier": "2.1.1",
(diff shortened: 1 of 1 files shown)


PierreRambaud (changes requested):
We already target the 2.2.0: https://github.com/PrestaShop/PrestaShop/blob/develop/composer.lock#L3538

Progi1984:
Ping @davideapvd

drc0:
sorry but I don't know how composer pulled the wrong version there. I'm unable to test this right now.

Progi1984:
@drc0 @davideapvd Do you want we close this PR ?

drc0:
yes please, let see if anyone else have this problem in the future.

PR #17533, opened 2020-02-06
Only compute assets for Travis for tests which need it

matks:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Only compute assets for Travis for tests which need it
| Type?         | improvement
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Travis is green

This PR is done here to avoid that check on files syntax is performed on dependency files:
Capture d’écran 2020-02-06 à 15 58 28

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17533)

PR #17556, opened 2020-02-07
Improve Travis config file to avoid green builds if File Syntax fails

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Improve Travis config file to avoid green builds if File Syntax fails
| Type?         | bug fix
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Travis is green and ITS FOR GOOD REASONS

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17556)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ script:
         bash tests-legacy/check_phpunit.sh;
         phpunit_result=$?;
 
-        if [[ "file_syntax_result" == "0" && "phpunit_result" == "0" ]]; then
+        if [ $file_syntax_result -eq 0 ] && [ $phpunit_result -eq 0 ]; then
               exit 0;
         else
               exit 255;
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,15 +85,10 @@ install:
 script:
   - if [ "$PRESTASHOP_TEST_TYPE" = "unit" ]; then
         bash tests/check_file_syntax.sh;
-        file_syntax_result=$?;
-        bash tests-legacy/check_phpunit.sh;
-        phpunit_result=$?;
+    fi
 
-        if [ $file_syntax_result -eq 0 ] && [ $phpunit_result -eq 0 ]; then
-              exit 0;
-        else
-              exit 255;
-        fi
+  - if [ "$PRESTASHOP_TEST_TYPE" = "unit" ]; then
+        bash tests-legacy/check_phpunit.sh;
     fi
 
   - if [ "$PRESTASHOP_TEST_TYPE" = "unit" ]; then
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -86,7 +86,15 @@ install:
 script:
   - if [ "$PRESTASHOP_TEST_TYPE" = "unit" ]; then
         bash tests/check_file_syntax.sh;
+        file_syntax_result=$?;
         bash tests-legacy/check_phpunit.sh;
+        phpunit_result=$?;
+
+        if [[ "file_syntax_result" == "0" && "phpunit_result" == "0" ]]; then
+              exit 0;
+        else
```
PululuK: Hi @matks 
Some feedbacks ... Wdyf ?
```bash
        if [ $file_syntax_result -eq 0 ] && [ $phpunit_result -eq 0 ]; then
              exit 0;
```

matks: Looks better indeed 😄 I'm afraid I'm a bad Bash developer

Thanks for the help 👍

PululuK: 😔 but no not at all ... you just forgot the `$` in your vars : 
  `file_syntax_result` and `phpunit_result`
so it will never be true:
```bash
if [[ "file_syntax_result" == "0" && "phpunit_result" == "0" ]]
```

Because you compare the `"file_syntax_result"` string with `"0"`

matks:
Travis fails because https://github.com/PrestaShop/PrestaShop/pull/17533 needs to be merged first

.travis.yml
matks: 😄 
I am not proud of this PR

Fixed with your suggestion 😉

.travis.yml
```diff
@@ -85,7 +85,15 @@ install:
 script:
   - if [ "$PRESTASHOP_TEST_TYPE" = "unit" ]; then
```
PierreRambaud, line 86:
```suggestion
  - if [ "$PRESTASHOP_TEST_TYPE" = "unit" ]; then
        bash tests/check_file_syntax.sh;
    fi

  - if [ "$PRESTASHOP_TEST_TYPE" = "unit" ]; then
        bash tests-legacy/check_phpunit.sh;
    fi

  - if [ "$PRESTASHOP_TEST_TYPE" = "unit" ]; then
        bash tests/check_unit.sh;
    fi

```

Like below

matks: Feels strange to repeat this `if` ... although I know it's to avoid the hardcoded exit

PierreRambaud: Two proper solutions:
 * repeat if statement
 * chain scripts with `&&`

matks: Fixed

PierreRambaud (approved):

Progi1984 (approved):

Progi1984:
Travis is now :green_heart:

PierreRambaud:
Thanks @matks

PR #16938, opened 2019-12-27
Simplify rendering of forms for 1.7.7 pages

tdavidsonas88:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Simplify rendering of forms for 1.7.7 pages
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/16891
| How to test?  | See below

## How to test

The best way to test is
- open the modified page with the branch `1.7.7.x`
- open the same form page with this pull request branch
=> **Their display and HTML/JS behavior should be the same.** We modified the code but not the behavior.
Also if possible, run browser based UI tests.

Nice to have: check if other not related forms to this  correctly - regression UI automated tests would be perfect. This would be nice because we changed the general templates elements rendering system (for example: how country state is rendered - appear/disappear).

Also some bug possibility might have been introduced because we started simplifying with form_rest and then refactored to use form_widget for rendering instead.

## List of modified pages (all 1.7.7.x pages)

- Catalog > Files add/edit form (migrated by @RaimondasSapola) (see form)
- Catalog > Supplier add/edit form (migrated by @zuk3975) (see form)
- "Improve > International > Taxes" options form
- Form type for tax add/edit
- Customers > Addresses -> add/edit form
- Customer Service > Order Messages > Add new / edit order message form
- Credit Slips (can only be added, no edited) form

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16938)

Issue #16891, opened 2019-12-20
Simplify rendering of forms for 1.7.8 pages


matks:
Hello @tdavidsonas88 I have another topic you can tackle in order to get more familiar with PrestaShop stack and that would be really useful for 1.7.7 😄 .

Since @sarjon 's PR https://github.com/PrestaShop/PrestaShop/pull/15130 we are able to render entire forms inside migrated pages using the single twig statement ` {{ form_rest(...) }}`.

**Before**
```
        {{ ps.form_group_row(cmsPageForm.content, {}, {
          'label': 'Page content'|trans({}, 'Admin.Design.Feature'),
          }) }}

        {{ ps.form_group_row(cmsPageForm.is_indexed_for_search, {}, {
          'label': 'Indexation by search engines'|trans({}, 'Admin.Design.Feature'),
          }) }}

        {{ ps.form_group_row(cmsPageForm.is_displayed, {}, {
          'label': 'Displayed'|trans({}, 'Admin.Global'),
          }) }}

        {% if cmsPageForm.shop_association is defined %}
          {{ ps.form_group_row(cmsPageForm.shop_association, {}, {
            'label': 'Shop association'|trans({}, 'Admin.Global')
            }) }}
        {% endif %}
```

**After**
```
        {% block cms_page_form_rest %}
          {{ form_rest(cmsPageForm) }}
        {% endblock %}
```

Basically, since we updated our [Twig Form Theme](https://symfony.com/doc/current/form/form_themes.html), the `{{ form_rest(...) }}` statement is able to construct the full html markup for our forms without us having to write it ourselves as we used to do for all pages before.

So benefit n°1 is "we write less code". Benefit n°2 is that modules are able to modify the structure of the form, so they can alter the form display without having to override any twig statement if the full form is rendered using `{{ form_rest(cmsPageForm) }}`

So your assignment is to modify all forms created during 1.7.7 development phase and constructed with the standard "I write the html markup myself" and modify them so that they are rendered by `{{ form_rest(cmsPageForm) }}` instead, although we want to keep the exact same display. @sarjon did it for 2 pages in https://github.com/PrestaShop/PrestaShop/pull/15130/ so you can use it as an example.

You might need to improve our Form Theme while doing it (https://github.com/PrestaShop/PrestaShop/pull/15130/ does it a little).

Target pages:
[...]

PR #17558, opened 2020-02-07
Simplify rendering of forms for 1.7.7 pages

tdavidsonas88:
| Questions     | Answers
| Branch?       | develop
| Description?  | Simplify rendering of forms for 1.7.7 pages. Cherry picked all commits from Develop branch https://github.com/PrestaShop/PrestaShop/pull/16938
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/16891
| How to test?  | See below

## How to test

The best way to test is
- open the modified page with the branch `1.7.7.x`
- open the same form page with this pull request branch
=> **Their display and HTML/JS behavior should be the same.** We modified the code but not the behavior.
Also if possible, run browser based UI tests.

Nice to have: check if other not related forms to this  correctly - regression UI automated tests would be perfect. This would be nice because we changed the general templates elements rendering system (for example: how country state is rendered - appear/disappear).

Also some bug possibility might have been introduced because we started simplifying with form_rest and then refactored to use form_widget for rendering instead.

## List of modified pages (all 1.7.7.x pages)

- Catalog > Files add/edit form (migrated by @RaimondasSapola) (see form)
- Catalog > Supplier add/edit form (migrated by @zuk3975) (see form)
- "Improve > International > Taxes" options form
- Form type for tax add/edit
- Customers > Addresses -> add/edit form
- Customer Service > Order Messages > Add new / edit order message form
- Credit Slips (can only be added, no edited) form

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16938)

### DIFF
--- a/src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxType.php
+++ b/src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxType.php
@@ -61,8 +61,24 @@ class TaxType extends AbstractType
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        $invalidCharsText = sprintf(
+            '%s <>;=#{}',
+            $this->translator->trans('Invalid characters:', [], 'Admin.Notifications.Info')
+        );
+
+        $nameHintText =
+            $this->translator->trans(
+                'Tax name to display in carts and on invoices (e.g. "VAT").',
+                [],
+                'Admin.International.Help'
+            )
+            . PHP_EOL
+            . $invalidCharsText;
+
         $builder
             ->add('name', TranslatableType::class, [
+                'label' => $this->translator->trans('Name', [], 'Admin.Global'),
+                'help' => $nameHintText,
                 'options' => [
                     'constraints' => [
                         new Length([
@@ -83,6 +99,12 @@ class TaxType extends AbstractType
                 ],
             ])
             ->add('rate', TextType::class, [
+                'label' => $this->translator->trans('Rate', [], 'Admin.International.Feature'),
+                'help' => $this->translator->trans(
+                    'Format: XX.XX or XX.XXX (e.g. 19.60 or 13.925)',
+                    [],
+                    'Admin.International.Help'
+                ),
                 'constraints' => [
                     new NotBlank([
                         'message' => $this->translator->trans(
@@ -114,6 +136,7 @@ class TaxType extends AbstractType
                 ],
             ])
             ->add('is_enabled', SwitchType::class, [
+                'label' => $this->translator->trans('Enable', [], 'Admin.Actions'),
                 'required' => false,
             ])
         ;
--- a/src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxType.php
+++ b/src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxType.php
@@ -28,6 +28,7 @@ namespace PrestaShopBundle\Form\Admin\Improve\International\Tax;
 
 use PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints\DefaultLanguage;
 use PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints\TypedRegex;
+use PrestaShop\PrestaShop\Core\ConstraintValidator\TypedRegexValidator;
 use PrestaShopBundle\Form\Admin\Type\SwitchType;
 use PrestaShopBundle\Form\Admin\Type\TranslatableType;
 use Symfony\Component\Form\AbstractType;
@@ -62,7 +63,7 @@ class TaxType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $invalidCharsText = sprintf(
-            '%s <>;=#{}',
+            '%s ' . TypedRegexValidator::CATALOG_CHARS,
             $this->translator->trans('Invalid characters:', [], 'Admin.Notifications.Info')
         );
 
--- a/src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
+++ b/src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
@@ -371,12 +371,14 @@ class SupplierType extends TranslatorAwareType
                 ],
             ])
             ->add('is_enabled', SwitchType::class, [
+                'label' => $this->trans('Enabled', 'Admin.Global'),
                 'required' => false,
             ])
         ;
 
         if ($this->isMultistoreEnabled) {
             $builder->add('shop_association', ShopChoiceTreeType::class, [
+                'label' => $this->trans('Shop association', 'Admin.Global'),
                 'required' => false,
                 'constraints' => [
                     new NotBlank([
--- a/src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
+++ b/src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
@@ -293,10 +293,7 @@ class SupplierType extends TranslatorAwareType
                 'translation_domain' => false,
                 'choices' => $this->statesChoiceProvider->getChoices([
                     'id_country' => $countryIdForStateChoices,
-                ]),
-                'attr' => [
-                    'class' => 'js-supplier-state',
-                ],
+                ])
             ])
             ->add('logo', FileType::class, [
                 'label' => $this->trans('Logo', 'Admin.Global'),
--- a/src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
+++ b/src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
@@ -166,7 +166,6 @@ class SupplierType extends TranslatorAwareType
             ])
             ->add('description', TranslateType::class, [
                 'label' => $this->trans('Description', 'Admin.Global'),
-                'required' => false,
                 'help' => sprintf(
                     '%s %s',
                     $this->trans('Will appear in the list of suppliers.', 'Admin.Catalog.Help'),
--- a/src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
+++ b/src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
@@ -115,8 +115,16 @@ class SupplierType extends TranslatorAwareType
         $data = $builder->getData();
         $countryId = 0 !== $data['id_country'] ? $data['id_country'] : $this->contextCountryId;
 
+        $invalidCharsText = sprintf(
+            '%s <>;=#{}',
+            $this->trans('Invalid characters:', 'Admin.Notifications.Info')
+        );
+
         $builder
             ->add('name', TextType::class, [
+                'label' => $this->trans('Name', 'Admin.Global'),
+                'required' => true,
+                'help' => $invalidCharsText,
                 'constraints' => [
                     new NotBlank([
                         'message' => $this->trans(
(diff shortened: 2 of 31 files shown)


matks (dismissed):

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxType.php
```diff
@@ -61,8 +61,24 @@ public function __construct(TranslatorInterface $translator)
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        $invalidCharsText = sprintf(
+            '%s <>;=#{}',
+            $this->translator->trans('Invalid characters:', [], 'Admin.Notifications.Info')
```
jolelievre, line 67: Should the placeholder be inside the translation message? @LouiseBonnard

jolelievre: Maybe you could add a const `TypedRegexValidator::CATALOG_CHARS` for `<>;=#{}` ? Like you did for name chars

src/PrestaShopBundle/Form/Admin/Sell/Supplier/SupplierType.php
```diff
@@ -108,15 +115,36 @@ public function __construct(
...
         $data = $builder->getData();
         $countryId = 0 !== $data['id_country'] ? $data['id_country'] : $this->contextCountryId;
 
+        $invalidCharsText = sprintf(
+            '%s <>;=#{}',
+            $this->trans('Invalid characters:', 'Admin.Notifications.Info')
+        );
+
+        $invalidGenericNameHint = sprintf(
+            '%s <>={}',
```
jolelievre, line 133: Maybe you should use `TypedRegexValidator::GENERIC_NAME_CHARS` here? Although if we use these const in multiple places maybe they should be in a dedicated settings class instead of `TypedRegexValidator` which is linked to a form constraint

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Tax/Blocks/form.html.twig
```diff
@@ -24,6 +24,7 @@
  *#}
 
 {% import '@PrestaShop/Admin/macros.html.twig' as ps %}
+{% form_theme taxForm 'PrestaShopBundle:Admin/TwigTemplateForm:prestashop_ui_kit.html.twig' %}
```
jolelievre, line 26: Then the macro is no useful any more? Since we don't use `ps.form_group_row`

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Tax/Blocks/form.html.twig
```diff
@@ -35,31 +36,8 @@
...
-        'help': rateHint,
-      }) }}
-
-      {{ ps.form_group_row(taxForm.is_enabled, {}, {
-        'label': 'Enable'|trans({}, 'Admin.Actions')
-      }) }}
-
       {% block tax_form_rest %}
-        {{ form_rest(taxForm) }}
+        {{ form_widget(taxForm) }}
```
jolelievre, line 39: Maybe the block name should be `tax_form_widget` ? Based on what you did for `attachment_form_widget`

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Tax/Blocks/tax_options.html.twig
```diff
@@ -32,59 +32,9 @@
...
-        {{ ps.form_group_row(taxOptionsForm.options.eco_tax_rule_group, {}, {
-          label: 'Ecotax'|trans({}, 'Admin.International.Feature'),
-          help: 'Define the ecotax (e.g. French ecotax: 19.6%).'|trans({}, 'Admin.International.Help')
-        }) }}
-      {% endif %}
-
-      {% block tax_options_rest %}
-        {{ form_rest(taxOptionsForm) }}
-      {% endblock %}
+      {% for child in taxOptionsForm -%}
```
jolelievre, line 35: Maybe add a block here? `{% block tax_options_form_widget %}`

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
```diff
@@ -53,91 +54,8 @@
...
-        }) }}
-      {% endif %}
-
-      {{ ps.form_group_row(addressForm.other, {}, {
-        'label': 'Other'|trans({}, 'Admin.Global'),
-        'help' : 'Invalid characters:'|trans({}, 'Admin.Notifications.Info') ~ ' <>;=#{}'
-      }) }}
-
-      {{ form_rest(addressForm) }}
+      {{ form_widget(addressForm) }}
```
jolelievre: Add block `address_form_widget` or `manufacturer_address_form_widget` ?

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Attachment/Blocks/form.html.twig
```diff
@@ -24,6 +24,7 @@
  *#}
 
 {% import '@PrestaShop/Admin/macros.html.twig' as ps %}
+{% form_theme attachmentForm 'PrestaShopBundle:Admin/TwigTemplateForm:prestashop_ui_kit.html.twig' %}
```
jolelievre, line 26: Is the macros import removable?

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Suppliers/Blocks/form.html.twig
```diff
@@ -114,38 +47,8 @@
...
-
-      {% if supplierForm.shop_association is defined %}
-        {{ ps.form_group_row(supplierForm.shop_association, {}, {
-          'label': 'Shop association'|trans({}, 'Admin.Global')
-        }) }}
-      {% endif %}
-
       {% block supplier_form_rest %}
-        {{ form_rest(supplierForm) }}
+        {{ form_widget(supplierForm) }}
```
jolelievre, line 47: Rename the block here as well `supplier_form_widget`

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Suppliers/Blocks/form.html.twig
```diff
@@ -34,77 +34,10 @@
...
-      {% set invalidCatalogNameHint %}
-        {{ 'Invalid characters:'|trans({}, 'Admin.Notifications.Info') ~ ' <>;=#{}' }}
-      {% endset %}
-
-      {{ ps.form_group_row(supplierForm.name, {}, {
-        'label': 'Name'|trans({}, 'Admin.Global'),
-        'help': invalidCatalogNameHint
-      }) }}
-
       {% set invalidGenericNameHint %}
```
jolelievre: This hint should be in the form directly now no?

src/PrestaShopBundle/Resources/views/Admin/Sell/CustomerService/OrderMessage/Blocks/form.html.twig
```diff
@@ -32,17 +33,8 @@
...
-        'label': 'Name'|trans({}, 'Admin.Global')
-      }) }}
-
-      {{ ps.form_group_row(orderMessageForm.message, {}, {
-        'label': 'Message'|trans({}, 'Admin.Global')
-      }) }}
-
       {% block order_message_form_rest %}
-        {{ form_rest(orderMessageForm) }}
+        {{ form_widget(orderMessageForm) }}
```
jolelievre, line 37: Rename block `order_message_form_widget`

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/prestashop_ui_kit.html.twig
```diff
@@ -53,13 +60,20 @@
...
 {% block form_row -%}
   {% spaceless %}
-    <div class="{{ block('form_row_class') }}{% if (not compound or force_error|default(false)) and not valid %} has-error{% endif %}">
-      {{ form_label(form) }}
-      <div class="{{ block('form_group_class') }}">
-        {{ form_widget(form) }}
-        {{ form_errors(form) }}
-      </div>
-    </div>
+        {# could be refactored better if we search id_state in block_prefixes and there's only one js-state class  #}
```
jolelievre, line 63: Or maybe it should be managed in the FormType rather than in the generic template?

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/prestashop_ui_kit_base.html.twig
```diff
@@ -71,20 +71,27 @@
         {% set label = name|humanize %}
       {%- endif -%}
     {%- endif -%}
-    <label{% for attrname, attrvalue in label_attr %} {{ attrname }}="{{ attrvalue }}"{% endfor %}>{{ translation_domain is same as(false) ? label|raw : label|raw }}
-    {% if label_attr['tooltip'] is defined %}
-      {% set placement = label_attr['tooltip_placement'] is defined ? label_attr['tooltip_placement'] : 'top' %}
-      <i class="icon-question" data-toggle="pstooltip" data-placement="{{ placement }}"
-         title="{{ label_attr['tooltip'] }}"></i>
-    {% endif %}
+    {# options 'comes' from service: prestashop.admin.credit_slip_options.form_handler in .yml file and is added as additional field; #}
```
jolelievre, line 74: So we need to make "buggy" modification in the template because the service was weirdly configured 😅

This feels weird.. If there is a problem with the `options` field of `prestashop.admin.credit_slip_options.form_handler` then I guess there will be a similar issue with the `configuration` field of `prestashop.admin.mail_theme.form_handler` or the `shop_logos` of `prestashop.admin.shop_logos_settings.form_handler`, and so on..

We can't keep piling ignore cases as we migrate the configuration forms..

jolelievre (changes requested):
I made a few similar comments about the block in which the widget is integrated I think it's a very good idea to have one at each time but we would need to define a convention because sometimes:
- it ends by `_widget`
- it ends by `_form`
- it ends by `_rest`
- there is no block

The last two are no good ^^ But we can choose either of the first two I don't mind

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxType.php
LouiseBonnard: Nope, it's fine. ;-)

tdavidsonas88:
> *I made a few similar comments about the block in which the widget is integrated I think it's a very good idea to have one at each time but we would need to define a convention because sometimes:
it ends by _widget
it ends by _form
it ends by _rest
there is no block
The last two are no good ^^ But we can choose either of the first two I don't mind

Well according to symfony documenation: 

**form_widget(form_view, variables)¶**
Renders the HTML widget of a given field. If you apply this to an entire form or collection of fields, each underlying form row will be rendered.

Not sure why to complicate it by creating additional blocks like _order_message_form_widget_  ? I think @eternoendless suggested form_widget rendering is good enough.

[2020.02.12] @jolelievre  I am sorry for this comment: the green and red colors made me misunderstand which line you recommended to change. It's about the block line and then your changes makes sense. Had a lack of understanding your good thought.

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxType.php
tdavidsonas88: This idea came when was facing errors like - we have one help message but in reality another characters were validated than in the help message of the form. Keeping it in one place, helps avoiding issues of this type.

tdavidsonas88:
this can be closed and instead work on: https://github.com/PrestaShop/PrestaShop/pull/16938  Cherry picked the needed changes already

PR #12377, opened 2019-01-30
Improve email validation

ghost:
Prevents logs and exceptions returned by the Swift

| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Avoid saving incorrect email addresses.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Simply try to use an email address like this `mail..fine@mail.com` in any form that you subsequently have to send an email to that email address, what's the problem?, PrestaShop will save perfectly the email without any error message for the user, but once an attempt is made to send the email, Swift will throw an exception and obviously the email will never be sent. With this modification we take advantage of Swift's own validation and avoid storing incorrect emails.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/12377)

PR #17561, opened 2020-02-07
Allow cyrilic domain names in email addresses

### LINKED ISSUE #11588: Email validation are not checked cyrillic characters before the @ sign. Also there is no support for Cyrillic domains.
**Describe the bug**
At the php level, the email validation form skips the cyrillic characters before the @ sign. Also, the form has no support for cyrillistic domains.

**How it should work**
Cyrillistic domains are supported by some email providers, such as Yandex or Google (Gmail).
Cyrillic characters should not be before the @ sign, only after it in domain, if the domain is converted to punycode.
For example:
test@site.com - this is a valid email address.
test@сайт.рф - this is a valid email address ( after in conversion to pinycode: test@xn--80aswg.xn--p1ai )
иван@сайт.рф - this is not a valid email address.
иван@site.com - this is not a valid email address too.
It should work that way. 

**To Reproduce**

1. Go to the [Create Account page](http://fo.demo.prestashop.com/en/login?create_account=1)
2. First, we need to disable html5 validation (because we will test only server validation).
For that we need to add 'novalidate' atribute in `<form>` tag. 
3. On the page in the Email field we write: test@сайт.рф
4. Click "Save" button. And we will see the message: 
> Invalid format. 
This is the first problem. The second is cyrillic characters before the @ sign.
5. Now repeat step 2 and enter another email address: тест@site.com
6. After that we will see the message: 
> Fatal error: Uncaught exception 'Swift_RfcComplianceException' with message 'Address in mailbox given [тест@test.site] does not comply with RFC 2822, 3.6.2.'

**Screenshots**
Step2:
Step2
Step3:
Step3
Step6:
Step6

**Additionnal information**
PrestaShop version: from 1.6 to 1.7
PHP version: all

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Before this change, it was impossible to use email addresses that contains cyrillic characters in the host part of the address.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #11588
| How to test?  | Try to create an account (BO and/or FO) with an email address containing cyrillic characters in the host part (**after** the `@`): it should work. Try with cyrillic characters **before** the `@`, it should **not** work.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17561)


matks (dismissed):

src/Core/Util/InternationalizedDomainNameConverter.php
```diff
@@ -0,0 +1,49 @@
...
+     * @return string
+     */
+    public function emailToUtf8(string $email): string
+    {
+        $parts = explode('@', $email);
+        if (count($parts) !== 2) {
+            return $email;
+        }
+
+        return $parts[0] . '@' . idn_to_utf8($parts[1], 0, INTL_IDNA_VARIANT_UTS46);
```
PierreRambaud, line 47: Watch out with `INTL_IDNA_VARIANT_UTS46` it's only available as of ICU 4.6
Because we now use SwiftMailer > 6.2 we should have a look on what SwiftMailer can do 
https://github.com/swiftmailer/swiftmailer/pull/1044

PierreRambaud (changes requested):
Wait for SwiftMailer feedback

src/Core/Util/InternationalizedDomainNameConverter.php
atomiix: @PierreRambaud I think we're pretty safe because we have the [Symfony polyfill](https://github.com/symfony/polyfill-intl-idn). WDYT?

PierreRambaud: You're right :+1:

classes/Validate.php
```diff
@@ -53,7 +53,7 @@ public static function isAnything()
      */
     public static function isEmail($email)
     {
-        return !empty($email) && preg_match(Tools::cleanNonUnicodeSupport('/^[a-z\p{L}0-9!#$%&\'*+\/=?^`{}|~_-]+[.a-z\p{L}0-9!#$%&\'*+\/=?^`{}|~_-]*@[a-z\p{L}0-9]+(?:[.]?[_a-z\p{L}0-9-])*\.[a-z\p{L}0-9]+$/ui'), $email);
+        return !empty($email) && filter_var($email, FILTER_VALIDATE_EMAIL);
```
PierreRambaud: https://github.com/PrestaShop/PrestaShop/pull/12377#issuecomment-584182583 we should consider using this package instead.

One reason is `FILTER_VALIDATE_EMAIL` will not consider `test@mydomain` as a real address (type of address you can have in your server)

atomiix: Done @PierreRambaud

composer.json
```diff
@@ -22,6 +22,7 @@
         "doctrine/cache": "^1.8",
         "doctrine/doctrine-bundle": "^1.6",
         "doctrine/orm": "^2.7",
+        "egulias/email-validator": "^2.1",
```
Progi1984, line 25: License of the package ?

Progi1984 (commented):
A question ?

composer.json
atomiix: MIT ;)

Progi1984:
@atomiix All unit tests are :red_circle:

atomiix:
@Progi1984 yeah I know, I put the issue back in In Progress on the kanban, I'm working on it 😉

Progi1984:
Please use the label WIP in this case. Thanks

atomiix:
@Progi1984 It should be good now. Only have a prettyCI error but it's not coming from the code I pushed 🤔

Progi1984 (dismissed):

src/PrestaShopBundle/Resources/config/services/core/grid/grid_data_factory.yml
```diff
@@ -86,6 +86,19 @@ services:
             - "@prestashop.core.localization.locale.context_locale"
             - "@=service('prestashop.adapter.legacy.context').getContext().currency.iso_code"
 
+    prestashop.core.grid.data_provider.email_logs:
```
jolelievre: This looks like a copy of existing service `prestashop.core.grid.data_factory.email_logs` Why did you need to duplicate it?
I think you can keep/use the original and define your decorator along with it in `src/PrestaShopBundle/Resources/config/services/core/grid/grid_data_factory.yml`

classes/Customer.php
```diff
@@ -446,6 +447,16 @@ public function getByEmail($email, $plaintextPassword = null, $ignoreGuest = tru
...
     }
 
+    /**
+     * Convert the host part of the email from punycode to utf8 (e.g,. email@xn--e1aybc.xn--p1ai -> email@тест.рф)
+     *
+     * @return string
+     */
+    public function getEmailForDisplay(): string
+    {
+        return (new InternationalizedDomainNameConverter())->emailToUtf8($this->email);
```
jolelievre: I'm not sure it's a good idea to instanciate the class at each call of this method..

src/Core/Grid/Data/Factory/CustomerGridDataFactoryDecorator.php
```diff
@@ -91,6 +92,7 @@ public function getData(SearchCriteriaInterface $searchCriteria)
     private function applyModifications(RecordCollectionInterface $customers)
     {
         $modifiedCustomers = [];
+        $idn = new InternationalizedDomainNameConverter();
```
jolelievre: Since you use this class in several services, I guess you could also declare it as a service and inject it?

jolelievre (commented):
Little proposal, you could declare a service for `InternationalizedDomainNameConverter`
This would avoid instanciating it in several services, you could also inject it in the handler that use `getEmailForDisplay` for display (which also instanciate the class at each call), the preview handler code would look like something like that:

```php
'email' => $this->converter->emailToUtf8($customer->email),
```

jolelievre (commented):
Nice update with service, seems much cleaner now. I have two other feedbacks:
1- regarding the `SwiftMailerValidation` class, since it implements an external interface maybe it should be in the Adapter namespace instead of Core (along with the related exception)

2- this fixes the bug in front office when we create an account, but should we backport this new validation in the CQRS commands as well, for customers created in the back office. We have an `Email` value object that is used in several handlers
The value object already has a private method to validate the format using `filter_var($email, FILTER_VALIDATE_EMAIL)` but I don't think this allows cyrilic characters?

src/Core/Domain/Customer/ValueObject/Email.php
```diff
@@ -26,97 +26,11 @@
 
 namespace PrestaShop\PrestaShop\Core\Domain\Customer\ValueObject;
 
-use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\CustomerConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\ValueObject\Email as BaseEmail;
 
 /**
- * Stores customer's email address
+ * @deprecated use PrestaShop\PrestaShop\Core\Domain\ValueObject\Email instead
```
eternoendless: Missing since, missing deprecation notice

src/Core/Util/InternationalizedDomainNameConverter.php
```diff
@@ -0,0 +1,49 @@
...
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Util;
+
+final class InternationalizedDomainNameConverter
```
eternoendless: Only classes implementing interfaces should be final

src/Core/Domain/Customer/ValueObject/Email.php
```diff
@@ -26,97 +26,16 @@
...
 
-use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\CustomerConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\ValueObject\Email as BaseEmail;
 
 /**
- * Stores customer's email address
+ * @deprecated since version 1.7.7 use PrestaShop\PrestaShop\Core\Domain\ValueObject\Email instead
  */
-class Email
+class Email extends BaseEmail
```
jolelievre: Since this class is now deprecated, it would be better if we stop using it in the core code, I only found a reference in `GetCustomerForAddressCreation` but maybe there are more

src/Core/Domain/ValueObject/Email.php
```diff
@@ -86,7 +86,7 @@ public function isEqualTo(Email $email)
      */
     private function assertEmailIsValid($email)
     {
-        if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
+        if (!preg_match('/^.+\@.+$/', $email)) {
```
PierreRambaud: Is this preg_match really useful?
Because "` azd    @  `"  will return true :thinking:

atomiix: @PierreRambaud Yeah, it will check if it looks like an email but it's not really trustworthy, maybe just an indication. Do you think we should get rid of this check in this ValueObject?

src/Core/Domain/Customer/ValueObject/Email.php
atomiix: @jolelievre Yeah, I didn't find this class elsewhere, and even in `GetCustomerForAddressCreation`, it's not really used because phpdoc of the function `getCustomerEmail()` is says that it returns a `Email` but the function declares returning a `string`. I'll update the phpddoc!

src/Core/Domain/ValueObject/Email.php
PierreRambaud: I prefer having no check instead of a useless one :/

matthieu-rolland (dismissed):

jolelievre (dismissed):

Progi1984 (dismissed):

PierreRambaud (dismissed):

boubkerbribri:
@atomiix  please resolve conflicts

Progi1984 (dismissed):

matks (dismissed):

Robin-Fischer-PS:
It's OK for creation and edit (on BO and FO).

But the display of the address is not OK on Advanced Parameters > Team > Employee in BO. Please tell me if this is related to this PR or not : 

pr17561 display mail employee BO

Thanks !

atomiix:
@Robin-Fischer-PS Nice catch, forgot about this place! I'll fix this ;)

Robin-Fischer-PS:
Oddly, there is another bug : The dropdown in order list do not work anymore (it work on the branch).

pr17561 orders dropdown

matks (dismissed):

Progi1984 (dismissed):

Robin-Fischer-PS:
Hi @atomiix !

Adding a customer or employee with Cyrilic mail is OK.

But after adding, it is not possible to login : 

pr17561 login

And when editing the employee/customer, this message is displayed (even if I did not change the mail) : 

pr17561 edit

Sorry :stuck_out_tongue:

admin-dev/themes/new-theme/js/app/utils/email-idn.js
```diff
@@ -0,0 +1,41 @@
...
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+import punycode from 'punycode';
+
+const $ = global.$;
```
PierreRambaud:
```suggestion
const {$} = window;
```

admin-dev/themes/new-theme/js/app/utils/email-idn.js
```diff
@@ -0,0 +1,41 @@
...
+
+const init = function initEmailFields() {
+  const $emailFields = $('input[type="email"]');
```
PierreRambaud: To be reusable, `'input[type="email"]'` should be in parameter

admin-dev/themes/new-theme/js/app/utils/email-idn.js
```diff
@@ -0,0 +1,41 @@
...
+  $.each($emailFields, (i, field) => {
+    if (!field.checkValidity()) {
+      const parts = field.value.split('@');
+      if (punycode.toASCII(parts[0]) === parts[0]) {
+        field.value = punycode.toASCII(field.value);
```
PierreRambaud, line 36: Only the domain needs to be in punycode. This means `parts[1]`   :thinking:

classes/form/CustomerForm.php
```diff
@@ -73,6 +77,17 @@ public function setPasswordRequired($passwordRequired)
         return $this;
     }
 
+    public function fillWith(array $params = [])
+    {
+        if (isset($params['email'])) {
```
PierreRambaud:
```suggestion
        if (!empty($params['email'])) {
```
We don't care about 0, false and empty string :)

classes/form/CustomerLoginForm.php
```diff
@@ -83,6 +90,17 @@ public function submit()
         return !$this->hasErrors();
     }
 
+    public function fillWith(array $params = [])
+    {
+        if (isset($params['email'])) {
```
PierreRambaud: same here

admin-dev/themes/new-theme/js/app/utils/email-idn.js
atomiix: @PierreRambaud it's a very specific case actually. This would work only for input[type="email"]. Do you see another use case for this?

admin-dev/themes/new-theme/js/app/utils/email-idn.js
atomiix: @PierreRambaud right, but as I'm checking that parts[0] won't be changed and the puycode library doesn't touch the local part of an email adresse, I think we should be pretty safe here. [https://github.com/bestiejs/punycode.js#punycodetoasciiinput](https://github.com/bestiejs/punycode.js#punycodetoasciiinput)

PierreRambaud: Add a comment to explain maybe?

admin-dev/themes/new-theme/js/app/utils/email-idn.js
PierreRambaud: If you have ten forms on your page, you maybe don't want to check them all, they are maybe not using the same validator as you do :thinking:

jolelievre (approved):
You're becoming quite the expert in Form transformation 😅

Progi1984 (approved):

PierreRambaud (approved):

Robin-Fischer-PS:
It's finally QA :heavy_check_mark: @atomiix !

Progi1984:
Thanks @atomiix

PR #17580, opened 2020-02-10
Module healthcheck creation. Prevent redirections

### LINKED ISSUE #12982: Application healthcheck
More often than not, we need to make sure that a specific PrestaShop instance is healthy and working as expected, that it is accessible, can connect to the DB and generally able to serve as expected.
Thing is, if we want to check a given instance (in HTTP) status (accessing a given page), we'll be redirected to the shop's main domain, which is not the check's purpose. 
If I have 3 PrestaShop instances (ps1, ps2 and ps3) for meshop.com, I want to ensure that my 3 instances are up and running, not just meshop.com (where I'm redirected to if not careful).

     
                         meshop.com    <----------      redirected to main domain
                             |                    |     then can fall to either 3 of the
                            LB                    |     prestashop instances
                             |                    |
                           / | \                  | 
                         /   |   \                | 
                     P1    P2   P3    <------ Checking instance P3

There are workarounds (playing with Host header, etc..) but those are not always applicable and are still workarounds.

Some healthcheck's feature would be :
  - A dedicated path (/health for example) to access the healthcheck
  - No shop's main domain redirection (accessible through private IP, localhost, internal domain, whatever)
  - Checking some PS internals
  - Preferably connecting to the shop's DB and performing a simple request (sysdate or else)
  - Returning the result OK or NOK following the results of the checks (with a proper HTTP Code)
  - Maintenance mode should not disable the healthcheck (for example)
 
As said, their are some workarounds to perform healthchecks, for example : working with Host header to avoid redirect, which implies access to the shop's DB prior to the call, which can be prone to error if the configuration file is not the same as the application's. 
Again, this is a workaround, far from a proper solution : it does not manage the shops' maintenance mode, for example.
Thing is, those may be hard to implement (especially in a containerized environment) and are difficult to maintain (and potentially prone to errors).

**Additional context**
Nope, but let me know if you need anything else.

sowbiba:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | The need is to perform healthcheck by calling an endpoint secured by IP filtering and a token in the request headers.<br />A module "ps_healthcheck" have been created. The security is handled by this module and some checks are already registered.<br />The user can add any check he wants by tagging his service with "admin.prestashop.healthcheck"<br />For LB and proxies, we need to prevent redirections when call is made from different URL than the configured BASE_URL. To do it, we avoid redirection when the route has the option "_prevent_redirect"<br />Linked to https://github.com/PrestaShop/ps_healthcheck/pull/1
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #12982
| How to test?  | Install and configure ps_healthcheck module.<br />Configure 2 apache virtual hosts with the same Prestashop instance and 2 different ServerNames<br />You can call the healthcheck URL (http://BASE_URL/admin-dev/modules/healthcheck) with the 2 BASE_URL without redirection

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17580)


src/PrestaShopBundle/DependencyInjection/Compiler/LoadCheckServicesPass.php
```diff
@@ -0,0 +1,37 @@
...
+class LoadCheckServicesPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasParameter('kernel.active_modules')) {
+            return;
+        }
+
+        $activeModules = $container->getParameter('kernel.active_modules');
+        if (!in_array('ps_healthcheck', $activeModules, true)) {
```
eternoendless: Avoid direct references to modules in the core

src/PrestaShopBundle/DependencyInjection/Compiler/LoadCheckServicesPass.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace PrestaShopBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class LoadCheckServicesPass implements CompilerPassInterface
```
eternoendless: Please describe what this class is for

eternoendless (changes requested):
1.7.7 is in feature freeze, please target the develop branch for any new features.

Please describe the problem and your proposed solution.

src/PrestaShopBundle/DependencyInjection/Compiler/LoadCheckServicesPass.php
sowbiba: OK. I added a way to load compiler pass defined in the module, as suggested

src/PrestaShopBundle/DependencyInjection/Compiler/LoadCheckServicesPass.php
sowbiba: moved to ps_healthcheck module and I add a description to it ;)

src/PrestaShopBundle/PrestaShopBundle.php
```diff
@@ -71,5 +74,52 @@ public function build(ContainerBuilder $container)
...
+    private function loadModuleCompilerPass(SplFileInfo $moduleFolder, ContainerBuilder $container)
+    {
+        $finder = new Finder();
+        $finder->files()
+            ->in($moduleFolder->getRealPath() . '/src/DependencyInjection/Compiler')
+            ->name('*.php')
+        ;
+        foreach ($finder as $phpFile) {
+            $phpContent = file_get_contents($phpFile->getRealPath());
+            if (preg_match('~namespace[ \t]+(.+)[ \t]*;~Um', $phpContent, $matches)) {
```
PierreRambaud: This function needs unit test :thinking:

app/AppKernel.php
```diff
@@ -175,6 +181,53 @@ private function enableComposerAutoloaderOnModules($modules)
...
+
+        $finder = new Finder();
+        $finder->files()
+            ->in($moduleSrcDirectory)
+            ->depth(0)
+            ->name('*Bundle.php')
+        ;
+        foreach ($finder as $phpFile) {
+            $className = PhpFileParser::getCLassName($phpFile);
+            if (is_a($className, Bundle::class, true)) {
```
PierreRambaud: Prefer instanceof, it's an instruction which is faster
```suggestion
            if ($className instanceof Bundle) {
```

sowbiba: instanceof requires to have an instance. I did not want to instanciate the class only to make sure it is a bundle or not

sowbiba: I dont know if `is_a` do instanciate the class behind the scene

eternoendless: It doesn't call the constructor: https://3v4l.org/RUWpH

PierreRambaud: You're right, forget the class isn't instantiated yet :sweat_smile:

app/AppKernel.php
```diff
@@ -79,6 +83,8 @@ public function registerBundles()
             }
         }
 
+        $this->loadModulesBundles($bundles);
```
eternoendless, line 83: Consider extracting a `ModuleBundleLoader` class

classes/module/Module.php
```diff
@@ -340,6 +348,58 @@ public function __construct($name = null, Context $context = null)
...
+            $moduleBundles[] = $bundle;
+        }
+        $registeredBundles[$this->name] = $moduleBundles;
+
+        $this->writeBundlesFile($registeredBundles, $configFile);
+    }
+
+    private function writeBundlesFile(array $moduleBundles, string $configFile)
+    {
+        $content = "<?php\nreturn [\n";
```
sowbiba, line 377: check var_export

Progi1984:
@sowbiba I add "waiting for author" waiting you fix ;)

classes/module/Module.php
sowbiba: I tried with var_export. Too complicated. 

The output I try to have is 
```
<?php
return [
	'ps_healthcheck' => [
		new PrestaShop\Module\HealthCheck\PsHealthcheckBundle(),
	],
];
```

If I take the instantiated bundle, I have an output like
```
<?php
return array (
  'ps_healthcheck' => 
  array (
    0 => 
    PrestaShopModuleHealthCheckPsHealthcheckBundle::__set_state(array(
       'name' => NULL,
       'extension' => NULL,
       'path' => NULL,
       'namespace' => NULL,
       'container' => NULL,
    )),
  ),
);
```

So a representation of the bundle object itself

If I set to the array the instantiation, 
`$modules[$moduleName][] = sprintf("new %s()", get_class($bundle));`

The output is 
```
<?php
return array (
  'ps_healthcheck' => 
  array (
    0 => 'new PrestaShop\Module\HealthCheck\PsHealthcheckBundle()',
  ),
);
```
It's a string not an instruction to create the bundle object

PierreRambaud: the `set_state` is a magic method for PHP to restore the statement of a class, it allows you to export your class easily at a moment. (https://www.php.net/manual/en/language.oop5.magic.php#object.set-state) 
It should be enough :thinking: 

Otherwise you can store something like this:
```php
return [
  'ps_healtcheck' => [
    PsHealthcheckBundle::class,
  ],
];
```
But need an adaptation in the bundle runner et not sure it's the best idea :sweat_smile:

matks:
Since https://github.com/PrestaShop/ps_healthcheck has been archived, I think we can safely close this PR?

Sorry for the waste of your time @sowbiba  😢

PierreRambaud:
We are sorry for the waste of time, the module is no longer required and @PrestaShop/product-team ask us to archive the repository.

saikarthikp9:
How do you host PrestaShop behind AWS load balancer without a healthcheck route?

PR #17611, opened 2020-02-12
Avoid bug shipping price being deducted twice

### LINKED ISSUE #17085: Migration - Create order bugs

tdavidsonas88:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fix. bug "shipping price is deducted twice". When FREE shipping voucher is added and then FREE SHIPPING option is added. Also vice versa.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/18551
| How to test?  | See below

## On BO 

https://drive.google.com/file/d/1sXSULMDGfz1TMAAqY0noKBKomAXnT41W/view?usp=sharing and the opposite order for free shipping and then voucher. It is important to rebuild the javascripts for it to work correctly: npm run build before testing

## On FO
Steps from https://github.com/PrestaShop/PrestaShop/issues/17875
1. Go to BO => Catalog => Discounts, create two cart rules with free shipping
2. Go to FO, create an order with paid shipping
3. Add the two cart rules
4. Shipping price is deducted twice

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17611)


admin-dev/themes/new-theme/js/pages/order/create/shipping-renderer.js
```diff
@@ -190,4 +204,22 @@ export default class ShippingRenderer {
   _hideNoCarrierBlock() {
     this.$noCarrierBlock.addClass('d-none');
   }
+
+  /**
+   * Show warning message block which warns that free shipping is already added
+   *
+   * @private
+   */
+  _showNoFreeShippingSwitchHelp() {
```
PierreRambaud: Please don't use underscore, the merge on develop will be hard instead :(

admin-dev/themes/new-theme/js/pages/order/create/shipping-renderer.js
```diff
@@ -62,26 +63,39 @@ export default class ShippingRenderer {
...
-  _renderFreeShippingSwitch(isFreeShipping) {
-    $(createOrderMap.freeShippingSwitch).each((key, input) => {
-      if (input.value === '1') {
-        input.checked = isFreeShipping;
-      } else {
-        input.checked = !isFreeShipping;
-      }
-    });
+  _renderFreeShippingSwitch(shipping) {
+    let isFreeShipping = shipping.freeShipping;
```
PierreRambaud:
```suggestion
    const isFreeShipping = shipping.freeShipping;
```

admin-dev/themes/new-theme/js/pages/order/create/shipping-renderer.js
```diff
@@ -62,26 +63,39 @@ export default class ShippingRenderer {
...
+    let isFreeShippingCartRuleAdded = shipping.freeShippingCartRuleAdded;
```
PierreRambaud:
```suggestion
    const isFreeShippingCartRuleAdded = shipping.freeShippingCartRuleAdded;
```

admin-dev/themes/new-theme/js/pages/order/create/shipping-renderer.js
tdavidsonas88: thank you :slightly_smiling_face: :

src/Core/Domain/Cart/QueryResult/CartInformation/CartShipping.php
```diff
@@ -53,22 +53,30 @@ class CartShipping
...
-        ?int $selectedCarrierId
+        ?int $selectedCarrierId,
+        bool $hasFreeShippingCartRule
     ) {
         $this->shippingPrice = $shippingPrice;
         $this->freeShipping = $freeShipping;
         $this->deliveryOptions = $deliveryOptions;
         $this->selectedCarrierId = $selectedCarrierId;
+        $this->isFreeShippingCartRuleAdded = $hasFreeShippingCartRule;
     }
```
matthieu-rolland, line 80: The property name is `isFreeShippingCartRuleAdded` and the injected variable is called `$hasFreeShippingCartRule` .

That's a minor thing but is it possible to give it the same name ? (I'm not blocking this PR for this though)

tdavidsonas88: Yes, it is possible. It happened because .js trims  the prefix of object property. That's why I though it must start with 'is' but also proved to be correct with 'has'. Big thanks! :slightly_smiling_face:

matks (dismissed):

matthieu-rolland (dismissed):

src/Adapter/Cart/QueryHandler/GetCartInformationHandler.php
```diff
@@ -120,16 +122,23 @@ public function handle(GetCartInformation $query): CartInformation
 
         $legacySummary = $cart->getSummaryDetails(null, true);
         $addresses = $this->getAddresses($cart);
+        $cartRules = $this->extractCartRulesFromLegacySummary($legacySummary, $currency);
+        $hasFreeShippingCartRule = false;
+        foreach ($legacySummary['discounts'] as $discount) {
+            if ((bool) $discount['free_shipping'] && $discount['name'] != self::FREE_SHIPPING_NAME) {
```
jolelievre: I don't think you can rely on the value of a cart rule name, this can be edited in the back office So it's not a reliable information

jolelievre (changes requested):
Relying on discount name seems risky

src/Adapter/Cart/QueryHandler/GetCartInformationHandler.php
tdavidsonas88: Ah, this one is tricky,  when FREE SHIPPING for order is toggled in the UI, it actually creates it always with the same name _Free Shipping_  (so it is hard coded in code): https://github.com/PrestaShop/PrestaShop/blob/d3099d26dff606fa4d38bcecdfccddfdeb1f80f5/src/Adapter/Cart/CommandHandler/SetFreeShippingToCartHandler.php#L114-L137 
The only other unique info is the code BO_ORDER (well it can also be changed in the the back office creating new cart rule). So have no idea, how make it more reliable!?

matks: I see a `free_shipping` SQL column into Cart Rule SQL table. Can we check what is inside this column ? it looks like a reliable information

tdavidsonas88: Looking to select * from ps_cart_rule results I don't see possibility to make it reliable without adding one more not changeable db column which defines what this cart rule came from toggling the Order FREE shipping button. But maybe we should think out of this box and search for different solution @matks ?

tdavidsonas88: @matks @jolelievre I think I'll try leave only ` if ((bool) $discount['free_shipping']` here because it makes sense. And will search a way to modify javascripts for page to render as expected

tdavidsonas88: @jolelievre @matks Thank you both, provided a cleaner solution - no longer relying on the Discount/Cart Rule name!

src/Adapter/Cart/QueryHandler/GetCartInformationHandler.php
```diff
@@ -120,16 +123,24 @@ public function handle(GetCartInformation $query): CartInformation
 
         $legacySummary = $cart->getSummaryDetails(null, true);
         $addresses = $this->getAddresses($cart);
+        $cartRules = $this->extractCartRulesFromLegacySummary($legacySummary, $currency);
+        $hasFreeShippingCartRule = false;
+        foreach ($legacySummary['discounts'] as $discount) {
+//            if ((bool) $discount[OrderDiscountType::FREE_SHIPPING] && $discount['name'] != self::FREE_SHIPPING_NAME) {
+            if ((bool) $discount[OrderDiscountType::FREE_SHIPPING]) {
```
jolelievre:
```suggestion
            if ((bool) $discount['free_shipping']) {
```

You shouldn't use `OrderDiscountType::FREE_SHIPPING` because it is related to the cart rule type Although it does match the `free_shipping` property, it's preferable to use the table field hard coded here

src/Adapter/Cart/QueryHandler/GetCartInformationHandler.php
```diff
@@ -120,16 +123,24 @@ public function handle(GetCartInformation $query): CartInformation
 
         $legacySummary = $cart->getSummaryDetails(null, true);
         $addresses = $this->getAddresses($cart);
+        $cartRules = $this->extractCartRulesFromLegacySummary($legacySummary, $currency);
+        $hasFreeShippingCartRule = false;
+        foreach ($legacySummary['discounts'] as $discount) {
+//            if ((bool) $discount[OrderDiscountType::FREE_SHIPPING] && $discount['name'] != self::FREE_SHIPPING_NAME) {
```
jolelievre: You can remove this comment now

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
```diff
@@ -289,6 +290,10 @@ public function addCartRuleAction(Request $request, int $cartId): JsonResponse
     {
         $cartRuleId = $request->request->getInt('cartRuleId');
         try {
+            $cartInfo = $this->getCartInfo($cartId)->getShipping();
```
jolelievre: It's weird to block the cart rule addition if the order has free shipping It means as long as you don't pay the shipping you can't add have additional discounts on products?

Are you sure this code shouldn't be in the `setFreeShippingAction` instead? In any case it would be better to manage this check in a handler rather than in the controller, the domain rules should be implemented by the handler It makes behat testing easier, and if the handler is later used in an API it contains the whole domain assertions

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
```diff
@@ -289,6 +290,10 @@ public function addCartRuleAction(Request $request, int $cartId): JsonResponse
...
+            if ($cartInfo->hasFreeShippingCartRule() || $cartInfo->isFreeShipping()) {
+                throw new FreeShippingCartRuleAlreadyExistException('Free shipping cart rule already exist!');
```
jolelievre: You should add a translated message matching this exception

jolelievre (changes requested):

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
tdavidsonas88: You just spotted a bug I've left: It means as long as you don't pay the shipping you can't add have additional discounts on products! :man_astronaut:

tdavidsonas88: Just I think the corrected code should go to `AddCartRuleToCartHandler` because that handler is actually invoked when adding any cart rule as a voucher

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
tdavidsonas88: ok, translated :)

jolelievre: I was referring to the `getErrorMessages` that makes the mapping between exception type and a translated message

src/Adapter/Cart/CommandHandler/AddCartRuleToCartHandler.php
```diff
@@ -72,8 +73,19 @@ public function __construct(
      */
     public function handle(AddCartRuleToCartCommand $command)
     {
+        /** @var Cart $cart */
         $cart = $this->getCart($command->getCartId());
+
         $cartRule = new CartRule($command->getCartRuleId()->getValue());
+        $freeShippingCartRules = $cart->getCartRules(CartRule::FILTER_ACTION_SHIPPING);
+        if (count($freeShippingCartRules) && $cartRule->getFields()['free_shipping']) {
```
jolelievre:
```suggestion
        if (count($freeShippingCartRules) && (bool) $cartRule->free_shipping) {
```

src/Adapter/Cart/QueryHandler/GetCartInformationHandler.php
```diff
@@ -59,6 +59,8 @@
  */
 final class GetCartInformationHandler extends AbstractCartHandler implements GetCartInformationHandlerInterface
 {
+    private const FREE_SHIPPING_NAME = 'Free Shipping';
```
jolelievre: this const is now useless

jolelievre (changes requested):
I believe you changed the chmod of `CartController.php` file to 755, revert it back to 644

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
tdavidsonas88: fixing this one now :)

tdavidsonas88: @jolelievre done, ready for review again!

src/Adapter/Cart/CommandHandler/AddCartRuleToCartHandler.php
```diff
@@ -80,11 +80,8 @@ public function handle(AddCartRuleToCartCommand $command)
         $freeShippingCartRules = $cart->getCartRules(CartRule::FILTER_ACTION_SHIPPING);
         if (count($freeShippingCartRules) && (bool) $cartRule->free_shipping) {
             throw new FreeShippingCartRuleAlreadyExistException(
-                $this->translator->trans(
-                    'Free shipping cart rule already exist!',
-                    [],
-                    'Admin.Notifications.Error'
-            ));
+                'Free shipping cart rule is already added to this cart!'
```
zuk3975: Why have you removed the translation :smile: ?

src/PrestaShopBundle/Controller/Admin/Sell/Order/CartController.php
```diff
@@ -578,6 +579,10 @@ private function getErrorMessages(Exception $e)
                     'Admin.Notifications.Error'
                 ),
             ],
+            FreeShippingCartRuleAlreadyExistException::class => $this->trans(
+                'Free shipping cart rule is already added to this cart!',
```
zuk3975: @marionf could you check the wording, please ? :cat:

LouiseBonnard:
```suggestion
                'Free shipping has already been taken into account!',
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/shipping.html.twig
```diff
@@ -56,13 +56,14 @@
...
-          <span class="slide-button"></span>
-        </span>
+          <span class="ps-switch">
+            <input id="free-shipping_0" class="js-free-shipping-switch" name="free_shipping" value="0" type="radio">
+            <label for="free-shipping_0">{{ 'No'|trans({}, 'Admin.Global') }}</label>
+            <input id="free-shipping_1" class="js-free-shipping-switch" name="free_shipping" value="1" checked="" type="radio">
+            <label for="free-shipping_1">{{ 'Yes'|trans({}, 'Admin.Global') }}</label>
+            <span class="slide-button"></span>
+          </span>
+          <small class="js-free-shipping-switch-help form-text d-none">{{ 'Free shipping is already enabled by voucher'|trans({}, 'Admin.Shipping.Feature') }}</small>
```
LouiseBonnard: I cannot see where this text is displayed. Could you please make a screenshot? It might need harmonization, to match with the previous error notification, but I prefer having the screenshot first. :-)

tdavidsonas88: It displays when you have voucher added with free shipping enabled here:
image

tdavidsonas88: Another possible case is when free shipping is toggled in the shipping section and you try to add the voucher with Free shipping, then different error message is displayed here: 
image

LouiseBonnard: Noted, thanks a lot, @tdavidsonas88! I prefer the first case, it looks more user-friendly. Let's pick this wording twice: _Free shipping voucher is already in your cart_ in _Admin.Orderscustomers.Help_ or _Admin.Notifications.Error_.

tdavidsonas88: Thank you Louise, changed it.

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/shipping.html.twig
```diff
@@ -56,13 +56,14 @@
...
-          <span class="slide-button"></span>
-        </span>
+          <span class="ps-switch">
+            <input id="free-shipping_0" class="js-free-shipping-switch" name="free_shipping" value="0" type="radio">
+            <label for="free-shipping_0">{{ 'No'|trans({}, 'Admin.Global') }}</label>
+            <input id="free-shipping_1" class="js-free-shipping-switch" name="free_shipping" value="1" checked="" type="radio">
+            <label for="free-shipping_1">{{ 'Yes'|trans({}, 'Admin.Global') }}</label>
+            <span class="slide-button"></span>
+          </span>
+          <small class="js-free-shipping-switch-help form-text d-none">{{ 'Free shipping voucher is already in your cart'|trans({}, 'Admin.Notifications.Error') }}</small>
```
LouiseBonnard:
```suggestion
          <small class="js-free-shipping-switch-help form-text d-none">{{ 'Free shipping voucher is already in your cart'|trans({}, 'Admin.Orderscustomers.Help') }}</small>
```

jolelievre (dismissed):
Thank you @tdavidsonas88

matks (dismissed):

src/Adapter/Cart/QueryHandler/GetCartInformationHandler.php
```diff
@@ -120,16 +120,23 @@ public function handle(GetCartInformation $query): CartInformation
 
         $legacySummary = $cart->getSummaryDetails(null, true);
         $addresses = $this->getAddresses($cart);
+        $cartRules = $this->extractCartRulesFromLegacySummary($legacySummary, $currency);
```
matks, line 123: 👍

Robin-Fischer-PS:
I've got a problem with the behaviour implemented with your PR : 

If a merchant has several cart rules with Free Shipping included, but not only that (Free Shipping + Gift, Free Shipping + 5% off), he cannot add two of these cart rules.

pr17611 free ship voucher cumulated

I'm not sure if this is a problem, and what should be done.

So dear @PrestaShop/prestashop-product-team , I need your help !

(the original bug : 
When adding a "Free Shipping" voucher on an order (NOT created by checking the "Free Shipping" on the page), and then, checking the "Free Shipping" option on the order, the Shipping price is deducted twice.
See this 1.7.7.x invoice : 
[Invoice two free ship 177.pdf](https://github.com/PrestaShop/PrestaShop/files/4147295/Invoice.two.free.ship.177.pdf)
And the same action on 1.7.6 (free shipping is only deducted once) : 
[Invoice two free ship 176.pdf](https://github.com/PrestaShop/PrestaShop/files/4147322/Invoice.two.free.ship.176.pdf)
)

MatShir:
The 2 vouchers should be applied but the shipping should be deducted once.
If I applied 2 vouchers in a cart:
- Voucher 1: Free Shipping + 5%
- Voucher 2: Free Shipping + Gift
Then
Free shipping is applied once with a 5 % discount and a gift.

tdavidsonas88:
> The 2 vouchers should be applied but the shipping should be deducted once. If I applied 2 vouchers in a cart: - Voucher 1: Free Shipping + 5% - Voucher 2: Free Shipping + Gift Then Free shipping is applied once with a 5 % discount and a gift.

Ok, will start rethinking the solution. So, I have to reduce validation and change the backend logic of how free shipping is applied. Haven't thought about this case possible to be accepted. As I understand free shipping can be applied many times by the user but only once to the final price.

tdavidsonas88:
Also noticed that it's possible to add cart rules and apply them to already created carts from the Discounts -> Cart Rules pages and that proves that the code needs to change - @MatShir suggested way.

MatShir:
If there is new behavior between cart rules, we will reference them in this file to keep track of it.
https://github.com/PrestaShop/prestashop-specs/blob/master/Cart%20Rule.md

tdavidsonas88:
I think the bug is related with this function: https://github.com/PrestaShop/PrestaShop/blob/30024b0d2a0612cc24497ef3b0971cfb27e2c379/src/Core/Cart/Calculator.php#L200-L222

And this debug data applied twice: 
image

tdavidsonas88:
Provided possible fix according to the specs, however I'm not sure if I haven't broken other related functionality which we haven't covered with the Behats!

LouiseBonnard (dismissed):

LouiseBonnard (dismissed):

PierreRambaud (dismissed):

src/Adapter/Cart/QueryHandler/GetCartInformationHandler.php
```diff
@@ -120,16 +120,23 @@ public function handle(GetCartInformation $query): CartInformation
...
+        $hasFreeShippingCartRule = false;
+        foreach ($legacySummary['discounts'] as $discount) {
+            if ((bool) $discount['free_shipping']) {
+                $hasFreeShippingCartRule = true;
```
Progi1984, line 127:
```suggestion
                $hasFreeShippingCartRule = true;
                break;
```

Progi1984 (approved):

jolelievre (approved):

jolelievre:
@tdavidsonas88 what is the bug you are talking about in the Calculator?
Is it related to this PR, or is it about the bug in front office in this issue? https://github.com/PrestaShop/PrestaShop/issues/17875

tdavidsonas88:
> @tdavidsonas88 what is the bug you are talking about in the Calculator?
> Is it related to this PR, or is it about the bug in front office in this issue? #17875

this PR. I tested it only in the Back Office and the total discount calculation was modified. Not sure if it also solves the front office bug.

matthieu-rolland:
you're right @jolelievre it seems like this PR fixes #17875 as well I just tested it and the bug disappear.

matthieu-rolland (approved):

Robin-Fischer-PS:
Hi @tdavidsonas88 

There is an issue with your PR : 

On FO : If I try to add too many vouchers with free shipping and other actions (for example Free Shipping only, then Free Shipping + Amount discount, then Free Shipping + Percent discount), the last voucher is not added and an error is displayed.

On BO, same thing, but the only "visible" behavior is that the voucher is not added. There is probably an error in the console.

Screenrecord of FO behaviour : 
https://drive.google.com/file/d/17KkAbrSg1CPAO8Xd0tf7OO3J1Rp1lMqO/view?usp=sharing

Screenshot of the error : 
pr17611 error crash

matks:
@Robin-Fischer-PS This PR does 2 things:
- it makes sure, when 2 "free shipping" vouchers are used, that the shipping fees are not removed twice from cart total
- it improves "create order" interface by preventing user to add both a "free shipping" rule and a "free shipping" voucher

As the 1st item is already a complex one, as you found bu running tests against the PR, I will fix the 1st item ("when 2 free shipping vouchers are used, that the shipping fees are not removed twice from cart total") in PR https://github.com/PrestaShop/PrestaShop/pull/18321 **then** when it's merged I will finish this PR with 2nd item.

tdavidsonas88:
> Hi @tdavidsonas88
> 
> There is an issue with your PR :
> 
> On FO : If I try to add too many vouchers with free shipping and other actions (for example Free Shipping only, then Free Shipping + Amount discount, then Free Shipping + Percent discount), the last voucher is not added and an error is displayed.
> 
> On BO, same thing, but the only "visible" behavior is that the voucher is not added. There is probably an error in the console.
> 
> Screenrecord of FO behaviour :
> https://drive.google.com/file/d/17KkAbrSg1CPAO8Xd0tf7OO3J1Rp1lMqO/view?usp=sharing
> 
> Screenshot of the error :
> pr17611 error crash

Testing BO part now because I understand it a little better than FO.

matks:
https://github.com/PrestaShop/PrestaShop/pull/18321 was merged so bug is now fixed into core backend code. What's left to do is to improve the interface to avoid the issue to happen

matks:
Closing now I will reopen it when needed

PR #17650, opened 2020-02-13
Fixed display for first cell (select all & input) in grid

### LINKED ISSUE #17563: Bad display in categories page when sorting data
**Describe the bug**
When sorting a tab which has bulk action and position selector, the display of the tab is broken.

**To Reproduce**
1. Go to Categories
2. Sort the tab (for example, by ID)
3. Bulk action checkbox display is not OK

**Screenshots**
pr17547 bad display

**Additional information**
PrestaShop version: 1.7.7.x
PHP version: N/A

Progi1984:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fixed display for first cell (select all & input) in grid
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17563 
| How to test?  | Steps to reproduce the behavior:<br>0. Build assets<br>1. Go to Categories<br>2. Sort the tab (for example, by ID)<br>3. **Before :** Bulk action checkbox display is not OK<br>3. **After :** Bulk action checkbox display is OK

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17650)

### DIFF
--- a/admin-dev/themes/new-theme/scss/components/_grid.scss
+++ b/admin-dev/themes/new-theme/scss/components/_grid.scss
@@ -93,6 +93,15 @@ table .column-filters td:last-child .grid-reset-button {
             display: inline-block;
             width: 60px;
           }
+          .md-checkbox {
+            &:not(:last-child) {
+              display: inline-block;
+            }
+            & ~ .form-control {
+              display: inline-block;
+              width: calc(100% - 29px);
+            }
+          }
         }
       }
     }


atomiix (dismissed):

admin-dev/themes/new-theme/scss/components/_grid.scss
```diff
@@ -93,6 +93,9 @@ table .column-filters td:last-child .grid-reset-button {
             display: inline-block;
             width: 60px;
           }
+          .d-inline-block.md-checkbox ~ .d-inline-block.form-control {
```
NeOMakinG: Can you try to select it using another class? It's dirty to select by a modifier class, and anyway these classes should'nt exist...
```suggestion
          .md-checkbox ~ .form-control {
```

Isn't that enough ?

src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Blocks/Table/filters_row.html.twig
```diff
@@ -26,14 +26,17 @@
 <tr class="column-filters {% if 0 == grid.data.records_total and grid.filters is empty and grid.actions.bulk|length < 0 %}d-none{% endif %}">
   {% for column in grid.columns %}
     <td>
+      {% set cssClassFormWidget = ''  %}
       {% if loop.first and grid.actions.bulk|length > 0 %}
+        {% set cssClassFormWidget = ' d-inline-block' %}
```
NeOMakinG: @eternoendless I know what do you think seeing . this, should we change this class ?

NeOMakinG (changes requested):

admin-dev/themes/new-theme/scss/components/_grid.scss
Progi1984: They can't not exist because they are in Prestashop-Ui-Kit because it is base on Bootstrap.

NeOMakinG: But you can add classes to components, even if they are from the UIKit ?

tests/Integration/Behaviour/Features/Context/Domain/OrderRefundFeatureContext.php
```diff
@@ -35,8 +35,8 @@
 use PrestaShop\PrestaShop\Core\Domain\Order\Command\IssuePartialRefundCommand;
 use PrestaShop\PrestaShop\Core\Domain\Order\Command\IssueReturnProductCommand;
 use PrestaShop\PrestaShop\Core\Domain\Order\Command\IssueStandardRefundCommand;
-use PrestaShop\PrestaShop\Core\Domain\Order\Exception\InvalidOrderStateException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\InvalidCancelProductException;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\InvalidOrderStateException;
```
matks: [Please rebase your PR : this fix is always fixed in 1.7.7.x branch](https://github.com/PrestaShop/PrestaShop/pull/17660#discussion_r379302373). :trollface:

admin-dev/themes/new-theme/scss/components/_grid.scss
jolelievre: But maybe there is a reason why it's applied only in `.d-inline-block` maybe there are several ways to display a `.md-checkbox` Sometines it's inline, sometimes not and it makes sense to have different display no?

src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Blocks/Table/filters_row.html.twig
jolelievre: So I'm not sure I get it, if you have bulk actions AND it's in the first column, then the filters displayed in this column should have an `inline` display?

jolelievre: I actually wonder, isn't it simpler if we always have inline blocks?

admin-dev/themes/new-theme/scss/components/_grid.scss
```diff
@@ -93,6 +93,9 @@ table .column-filters td:last-child .grid-reset-button {
...
+            width: calc(100% - 29px);
```
jolelievre: Maybe you could use `width: fit-content;` ?

src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Blocks/Table/filters_row.html.twig
NeOMakinG: The correct thing would be to have display: inline-block; on a proper selector css-sided, that's it so it would be something like 
```suggestion
        {% set cssClassFormWidget = ' grid-column-checkbox' %}
```

NeOMakinG: The correct thing would be to have display: inline-block; on a proper selector css-sided, that's it so it would be something like 
```suggestion
        {% set cssClassFormWidget = ' grid-column-checkbox' %}
```

and in the css of the grid :
```
  .grid-column-checkbox {
    display: inline-block;
  }
```

and if it not display block in all cases, there should be a modifier class added using a condition like `.grid-column-checkbox-inline`

admin-dev/themes/new-theme/scss/components/_grid.scss
NeOMakinG: https://caniuse.com/#search=fit-content

I think no as it represent almost 4% of the web traffic (and as @matks think, maybe 8% 🤣 ) where it will break

tests/Integration/Behaviour/Features/Context/Domain/OrderRefundFeatureContext.php
Progi1984: Done

src/PrestaShopBundle/Resources/views/Admin/Common/Grid/Blocks/Table/filters_row.html.twig
Progi1984: I don't understand why we can't use classes from PrestaShop UI Kit based on the framework Twitter Bootstrap. It would be recreate classes because we don't want to use Symfony framework.

tdavidsonas88: @NeOMakinG any explanation to move the PR forward?

NeOMakinG: @Progi1984 "helpers classes" are against the base of CSS and webcomponents principles, why even using `d-inline-block` classes as it is the same as using a `style=""` attribute. Creating a class to use only one CSS property is too weird, if we follow that idea, then let's create a class for every properties and never touch stylesheets anymore.

To be honest, that's just laziness. I'm actually working on doing returns from the product team on View Order pages and theses classes make it hard to customize almost everything because they are often using !important and they make it really harder.

For this time as I said on some others PR's, let's stay like this, but before next migrations, we have to speak of a proper way to integrate design's using good webcomponents guidelines and workflow.

@eternoendless @PierreRambaud help...

PierreRambaud: I totally agree what @NeOMakinG said :+1: 
Using these kinds of "classes" is a bad practice and make the customisation harder to be done!

I'm 100% sure, if our front-end developer says it's a bad practice and these classes shouldn't be used, we must trust him! He's the one here who's working with CSS & HTML, we shouldn't argue and listen him!

eternoendless: I agree with @NeOMakinG on this. Helper classes are useful for quick prototyping but are actually antipatterns, because they violate the principle of separating style from semantics, which is the reason behind CSS files being a thing.

If changing the style of an element requires you to change the object's markup, **_you're doing it wrong_**.

Progi1984: I do it, but always not agree

admin-dev/themes/new-theme/scss/components/_grid.scss
```diff
@@ -93,6 +93,15 @@ table .column-filters td:last-child .grid-reset-button {
             display: inline-block;
             width: 60px;
           }
+          .md-checkbox {
+            &:not(:last-child) {
+              display: inline-block;;
```
NeOMakinG:
```suggestion
              display: inline-block;
```

admin-dev/themes/new-theme/scss/components/_grid.scss
```diff
@@ -93,6 +93,15 @@ table .column-filters td:last-child .grid-reset-button {
...
+            }
+            & ~ .form-control {
+              display: inline-block;;
```
NeOMakinG:
```suggestion
              display: inline-block;
```

admin-dev/themes/new-theme/scss/components/_grid.scss
```diff
@@ -93,6 +93,15 @@ table .column-filters td:last-child .grid-reset-button {
             display: inline-block;
             width: 60px;
           }
+          .md-checkbox {
+            &:not(:last-child) {
+              display: inline-block;;
+            }
```
NeOMakinG, line 99:
```suggestion
            }
            
```

Add a space to increase readability (would not ask you if I had nothing else to ask you 😄  )

NeOMakinG (changes requested):

NeOMakinG (approved):

jolelievre (approved):

khouloudbelguith:
Hi @Progi1984,

It is OK
image

Thanks!

matks:
Thank you @Progi1984

PR #17651, opened 2020-02-13
Update tinymce from 4.0.16 to 4.9.8

### LINKED ISSUE #11011: JavaScript error in console when editing an image in TinyMCE - BO CMS Page
**Describe the bug**
issue JS when editing a CMS page with an image in the wysiwyg

**To Reproduce**

1. Go to Design > Pages
1. Edit a CMS page
1. Add an image in the wysiwyg editor & save
1. Edit the CMS page and click on the image in the wysiwyg editor
1. See error, you can't edit the image

**Screenshots**
capture d ecran!UNITO-UNDERSCORE!437

**Additionnal information**
PrestaShop version: 1.7.4.3
PHP version: N/A

### LINKED ISSUE #17376: [BO] JS error in tinymce on "add new supplier" page
**Describe the bug**
When going on the "add a supplier page" (the page with the form to add a supplier), we sometimes have a javascript error related to tinymce.

It's not all the time, I had to refresh the page at least 20 times to reproduce this 😓 

This is blocking some automated tests made by the QA team, so this issue should be taken seriously.

**Steps to reproduce the behavior:**
1. Go to **admin-dev/sell/catalog/suppliers/new**
2. Check for the error in the console
3. Refresh until you have the error

**Screenshots**
image (1)

**Additional information**
PrestaShop version: 1.7.7.x

NeOMakinG:
| Questions     | Answers
| Branch?       | develop
| Description?  | We wants to update tinymce as it fix #11011 and maybe #17376, I updated from 4.0.16 as 5+ are major versions and it may break everything, this version is not adding breaking changes, thats already a good thing because I removed a modification made in the tinymce core by PrestaShop team in the past, this tweak is not needed anymore :)
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | yes (But this is uncertain at all, and it could only be design break on certains old modules)
| Deprecations? | no
| Fixed ticket? | #11011 and maybe #17376 
| How to test?  | Build assets and go on tinymce pages, tinymce should work as expected, also you should be able to select an image that you already saved / added which is was not possible before. It should use the same language package as before too, it means that if you are using the BO in a certain lang, the language should be right.

<p align="center">

</p>

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17651)


js/admin/tinymce.inc.js
```diff
@@ -45,40 +45,58 @@ function tinySetup(config) {
     config.selector = '.' + config.editor_selector;
   }
 
-
   var default_config = {
-    selector: ".rte",
-    plugins: "align colorpicker link image filemanager table media placeholder advlist code table autoresize",
+    selector: '.rte',
+    plugins:
+      'align colorpicker link image filemanager table media placeholder advlist code table autoresize',
```
eternoendless: Weird line break here

NeOMakinG:
Just tested it on a legacy tiny mce, that looks fine, except some layout that should be fixed on the side of modules and not on our side, but this cost a potential BC Break

PierreRambaud (dismissed):
Sometimes we need to take risks!

Progi1984:
@PrestaShop/prestashop-core-developers  If it's a dependancy, why not remove files and move it to dependency manager (npm / package.json) ?
It's a bad practice to use and to show to our community to put assets file in repository.

PierreRambaud:
@Progi1984 this dependency has been added in a specific path, if we remove files, we create a huge BC breaks, modules are using this library :/

matks (dismissed):

matks:
> Just tested it on a legacy tiny mce, that looks fine, except some layout that should be fixed on the side of modules and not on our side, but this cost a potential BC Break

You mean "if we fix the weird display, this would be a BC break ; if we dont fix the display (no BC break) then modules will have to do it" ?

sarahdib:
There is some display problem  : 

Capture d’écran 2020-06-03 à 10 59 29

Capture d’écran 2020-06-03 à 11 34 19

PierreRambaud (approved):

Progi1984 (approved):

Progi1984:
Thanks @NeOMakinG & @sarahdib

Issue #17665, opened 2020-02-13
Calculation of discounts  - Discount of a cart rule with percentage reduction restricted to specific product


arouiadib:
When a cart rule is created with percentage reduction on specific product, the discounts subtotal is calculated on final price and not on unit price of that specific product.

**To Reproduce**

1. Go to BO => Catalog => Discounts => Create a new cart rule
   1.1 Give a name to the cart rule.
   1.2 As an action => Apply a discount Percent = 100% => Apply a discount to: "Specific product".
   1.3 Choose a product (`demo_1` hummingbird printed t-shirt).
2. Go to FO, add whatever products in cart, but including `demo_1`.
3. Increase the quantity of `demo_1` to 10 for example.
4. See behaviour in screenshot.

**Screenshots**
![specific product - percentage reduction cart rule - discounts calculation](https://user-images.githubusercontent.com/23559010/74469966-ebbbb200-4e9d-11ea-8a45-dbb6fabcd672.png)

**Additional information**
PrestaShop version: 1.7.7.x, 1.7.6.3
PHP version:  7.2

khouloudbelguith:
Hi @arouiadib,

Thanks for your report.
I have the same behavior as PS1.7.5.2, PS1.7.6.3 & branch 1.7.7.x.
But, I think, it is normal behavior.
Ping @PrestaShop/prestashop-product-team @MatShir what do you think?
We apply 100% for a product demo_1 & there is no quantity specified in this case
![image](https://user-images.githubusercontent.com/16067358/74522487-945c2700-4f1b-11ea-84cf-36c877afcdb7.png)

Thanks!

arouiadib:
Thank you very much @khouloudbelguith for your time.

Let me please add another example, and I hope my confusions will be cleared this time.

A shop with 5 products: Product1, ...Product5 with an available quantity of 100 item each.
The merchant wants to give a percentage discount restricted to `Product5`.
And a cart that contains:
Product1 with quantity= 1
Product5  with quantity = 20

You can see that the customer is able to have a discount on many items while buying only one product from Product[1..4]. If additionally the percentage reduction is set to 100% then, after 5 similar carts , `product5` is out of stock and only 4 items are really purchased (Merchant looking for bankruptcy). 

[...]

PR #17667, opened 2020-02-13
WIP - Fix discounts subtotal - percentage reduction on specific product/selection of products

### LINKED ISSUE #17666: Specs for calculation of discounts  - Discount of a cart rule with percentage reduction restricted to selection of products
When a cart rule is created with percentage reduction on a selection of products, the discounts subtotal is calculated on final price and not on unit price of that selection of products.

**To Reproduce**

1. Go to BO => Catalog => Discounts => Create a new cart rule
   1.1 Give a name to the cart rule.
   1.2 As conditions => Restrictions => product selection => Choose 2 products (`demo_1` Hummingbird printed t-shirt and `demo_3` Hummingbird printed sweater).
   1.3 As an action => Apply a discount Percent = 100% => Apply a discount to: "Selection of products".
2. Go to FO, add whatever products in cart, but including `demo_1` and/or `demo_3`.
3. Increase the quantity of `demo_1` and `demo_3` to 10 for example.
4. See behaviour in screenshot.

**Screenshots**
selection of products - percentage reduction cart rule - discounts calculation

**Additional information**
PrestaShop version: 1.7.7.x, 1.7.6.3
PHP version:  7.2

arouiadib:
For a percentage reduction, there are 4 types:
- On whole order.
- On cheapest product.
- On specific product.
- On selection of products.

For the whole order, it is obvious that the calculation should be performed on the final prices. ( perent * sum of final prices of rows)
But for cheapest product, specific product and selection of products, the calculation should be performed on unit price. This is already implemented for cheapest product. This PR tries to follow suit regarding specific product and selection of products.

This PR tries also to trigger discussion about cart rule calculations specs. An initial draft is [here](https://github.com/PrestaShop/prestashop-specs/blob/845b50998c2a5f0c783c72ae6806a92b0beecbaa/core/front-office/cart%20rules%20calculations.md) with hope of it being reviewed, corrected, tagged and helpful.

| Questions     | Answers
| Branch?       | 1.7.7.x 
| Description?  | 
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17666 and #17665 
| How to test?  | Reproduce issues and increase quantities of specific product in cart (or selection of products).

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17667)

### DIFF
--- a/src/Core/Cart/CartRow.php
+++ b/src/Core/Cart/CartRow.php
@@ -447,7 +447,7 @@ class CartRow
 
     /**
      * @param float $percent 0-100
-     * @param boolean $onUnitPrice;
+     * @param bool $onUnitPrice;
      *
      * @return AmountImmutable
      */
@@ -462,8 +462,7 @@ class CartRow
             $discountTaxIncluded = $this->initialUnitPrice->getTaxIncluded() * $percent / 100;
             $discountTaxExcluded = $this->initialUnitPrice->getTaxExcluded() * $percent / 100;
             $amount = new AmountImmutable($discountTaxIncluded, $discountTaxExcluded);
-        }
-        else{
+        } else{
             $discountTaxIncluded = $this->finalTotalPrice->getTaxIncluded() * $percent / 100;
             $discountTaxExcluded = $this->finalTotalPrice->getTaxExcluded() * $percent / 100;
             $amount = new AmountImmutable($discountTaxIncluded, $discountTaxExcluded);
--- a/src/Core/Cart/CartRow.php
+++ b/src/Core/Cart/CartRow.php
@@ -447,7 +447,7 @@ class CartRow
 
     /**
      * @param float $percent 0-100
-     * @param boolean $onUnitPrice;
+     * @param bool $onUnitPrice;
      *
      * @return AmountImmutable
      */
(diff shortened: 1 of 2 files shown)


PierreRambaud:
We can't approve your fix since it breaks integration tests.
And because it's related to the Cart, it needs more tests to be sure everything is ok :)

arouiadib:
@PierreRambaud, thank you very much for your usual help. In front of TravisCI huge wall of logs, It wasn't obvious for a Behat newbie to find out what are  failed scenarios.

Here, broken tests seems to be a normal thing, because this PR is changing the business logic of a calculation. [I am still not sure if that is what should be done.](https://github.com/PrestaShop/PrestaShop/issues/17665#issuecomment-586744326). Once I get the answer and if the issue is valid, I will rewrite existing integration tests and add another scenario.

But, this current PR contains also a refactoring, so I have split the two things. I have now a [separate refacto PR  that breaks no integration tests](https://github.com/PrestaShop/PrestaShop/pull/17687), and will hopefully simplify this PR.

Again, thanks for your help.


**A note for Behat new kids in the block:**

* If locally `composer create-test-db` doesn't work, you can directly create test database using `php tests-legacy/create-test-db.php`.
* For slow machines, running locally `php vendor/bin/behat -c tests/Integration/Behaviour/behat.yml` will take a whole life. using `-s` will execute specific test suite, and relatively be faster: `php ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml -s cart` for example.

arouiadib:
I close this PR because the corresponding issues need more study. During the migration of cart rule pages in back office, many answers will be easily found ;-).

In the future it will be reopened, but with some behat tests and without breaking current one.

Thank you very much.

PR #17672, opened 2020-02-14
Rework ChangeOrderCurrencyHandler to private methods

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fix multiple todo's `handle in private method` in ChangeOrderCurrencyHandler
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Make sure Changing order currency in BO order view page works as expected. You can change order currency in order view page payments block. There should be multiple active currencies in BO and the order must be valid to see that block (although I am not aware about precise specification what is a "valid" order, but I can see it being rendered on all fixture orders regardless their status :man_shrugging:) Screenshot from 2020-03-12 11-15-47

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17672)

### DIFF
--- a/src/Adapter/Order/AbstractOrderHandler.php
+++ b/src/Adapter/Order/AbstractOrderHandler.php
@@ -29,8 +29,10 @@ namespace PrestaShop\PrestaShop\Adapter\Order;
 use Customer;
 use Group;
 use Order;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Order\ValueObject\OrderId;
+use PrestaShopException;
 
 /**
  * Reusable methods for Order subdomain command/query handlers.
@@ -44,7 +46,18 @@ abstract class AbstractOrderHandler
      */
     protected function getOrderObject(OrderId $orderId)
     {
-        $order = new Order($orderId->getValue());
+        try {
+            $order = new Order($orderId->getValue());
+        } catch (PrestaShopException $e) {
+            throw new OrderException(
+                sprintf(
+                    'Error occured when trying to get order object #%s',
+                    $orderId->getValue()
+                ),
+                0,
+                $e
+            );
+        }
 
         if ($order->id !== $orderId->getValue()) {
             throw new OrderNotFoundException($orderId, sprintf('Order with id "%d" was not found.', $orderId->getValue()));
--- a/src/Adapter/Order/CommandHandler/ChangeOrderCurrencyHandler.php
+++ b/src/Adapter/Order/CommandHandler/ChangeOrderCurrencyHandler.php
@@ -88,20 +88,22 @@ final class ChangeOrderCurrencyHandler extends AbstractOrderHandler implements C
      */
     private function updateOrderCarrier(int $orderCarrierId, Currency $oldCurrency, Currency $newCurrency): void
     {
-        if ($orderCarrierId) {
-            $order_carrier = new OrderCarrier($orderCarrierId);
-            $order_carrier->shipping_cost_tax_excl = (float) Tools::convertPriceFull(
-                $order_carrier->shipping_cost_tax_excl,
-                $oldCurrency,
-                $newCurrency
-            );
-            $order_carrier->shipping_cost_tax_incl = (float) Tools::convertPriceFull(
-                $order_carrier->shipping_cost_tax_incl,
-                $oldCurrency,
-                $newCurrency
-            );
-            $order_carrier->update();
+        if (!$orderCarrierId) {
+            return;
         }
+
+        $order_carrier = new OrderCarrier($orderCarrierId);
+        $order_carrier->shipping_cost_tax_excl = (float) Tools::convertPriceFull(
+            $order_carrier->shipping_cost_tax_excl,
+            $oldCurrency,
+            $newCurrency
+        );
+        $order_carrier->shipping_cost_tax_incl = (float) Tools::convertPriceFull(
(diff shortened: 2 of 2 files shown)


src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -44,7 +46,18 @@ abstract class AbstractOrderHandler
      */
     protected function getOrderObject(OrderId $orderId)
     {
-        $order = new Order($orderId->getValue());
+        try {
+            $order = new Order($orderId->getValue());
+        } catch (PrestaShopException $e) {
+            throw new OrderException(
```
PierreRambaud, line 52: Should be OrderNotFoundException? :thinking:

src/Adapter/Order/CommandHandler/ChangeOrderCurrencyHandler.php
```diff
@@ -82,33 +130,64 @@ public function handle(ChangeOrderCurrencyCommand $command)
...
+     * @param Currency $newCurrency
+     */
+    private function updateOrder(Order $order, Currency $oldCurrency, Currency $newCurrency): void
+    {
+        $fields = $this->getSharedAmountFields();
 
-        // Update order && order_invoice amount
-        // @todo: use private method to handle this
-        $fields = [
+        foreach ($fields as $field) {
```
PierreRambaud: Duplicate from line 153, you maybe can add a method `updateFields($object, $fields, $oldCurrency, $newCurrency)` maybe?

src/Adapter/Order/AbstractOrderHandler.php
matthieu-rolland: is this try/catch even necessary, given there's an OrderNotFoundException thrown at line 62 if we don't have the right Order object ?

zuk3975: This one might be thrown if something unexpected happens, like wrong query, or smth with db connection etc., Ideally we want to surround all Prestashop exceptions and replace by new core exceptions.  Practically it will never happen i guess.

src/Adapter/Order/CommandHandler/ChangeOrderCurrencyHandler.php
zuk3975: https://github.com/PrestaShop/PrestaShop/pull/17672/commits/65c612888ad0037d53868c1661136caf7dc66add implemented your suggestion. Although its bad desing vs duplication (couldn't type the object class well), I think its worth it.

src/Adapter/Order/CommandHandler/ChangeOrderCurrencyHandler.php
```diff
@@ -52,17 +57,62 @@ public function handle(ChangeOrderCurrencyCommand $command)
...
 
-        $oldCurrency = new Currency($order->id_currency);
-        $currency = new Currency($command->getNewCurrencyId()->getValue());
+        try {
+            $oldCurrency = new Currency($order->id_currency);
+            $newCurrency = new Currency($command->getNewCurrencyId()->getValue());
 
-        if (!Validate::isLoadedObject($currency)) {
-            throw new OrderException('Can\'t load Currency object');
+            if (!Validate::isLoadedObject($newCurrency)) {
```
Progi1984: And why not validate the `$oldCurrency` ?

src/Adapter/Order/CommandHandler/ChangeOrderCurrencyHandler.php
```diff
@@ -52,17 +57,62 @@ public function handle(ChangeOrderCurrencyCommand $command)
...
-            $order_detail = new OrderDetail($orderDetail['id_order_detail']);
-            // @todo: use private method to handle this
+    /**
+     * @param int $orderCarrierId
+     * @param Currency $oldCurrency
+     * @param Currency $newCurrency
+     */
+    private function updateOrderCarrier(int $orderCarrierId, Currency $oldCurrency, Currency $newCurrency): void
+    {
+        if ($orderCarrierId) {
```
Progi1984:
```suggestion
        if (!$orderCarrierId) {
            return;
        }
```

src/Adapter/Order/CommandHandler/ChangeOrderCurrencyHandler.php
```diff
@@ -81,34 +131,50 @@ public function handle(ChangeOrderCurrencyCommand $command)
...
-            );
-            $order_carrier->update();
+    /**
+     * @param PrestaShopCollection $invoices
+     * @param Currency $oldCurrency
+     * @param Currency $newCurrency
+     */
+    private function updateInvoices(PrestaShopCollection $invoices, Currency $oldCurrency, Currency $newCurrency): void
+    {
+        if ($invoices->count()) {
```
Progi1984:
```suggestion
        if (!$invoices->count()) {
            return;
        }
```

zuk3975:
@Progi1984 thanks for feedback, addressed.

Progi1984 (approved):

matks (approved):

SD1982:
Hi @zuk3975  can you give us some more info about the "how to test" when you talk about "Changing order currency in BO order view page" are you referring to this area of ​​the page ? 
capture d'écran_60
Because this is the only place on this page where it's possible to change the currency

zuk3975:
> Hi @zuk3975 can you give us some more info about the "how to test"

@SD1982 , sorry my bad. When you have multiple currencies in BO and the order is valid (although   I am not aware about precise specification when its considered valid) you should be able to see another select box in Payments block. (Updated description with screenshot)

SD1982:
This PR works as expected . Thanks @zuk3975

PR #17679, opened 2020-02-14
Remove empty option in select brand of product page

### LINKED ISSUE #9678: [BOOM-5269] BO - Product page > Brand select dropdown
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-5269).

- _**Reporter:**_ khouloud.belguith
- _**Created at:**_ Fri, 6 Apr 2018 18:53:53 +0200

<p>Hi,</p>

<p>In the product page, when you add a brand, a drop-down list is displayed, where there is an empty select option by default that is not correct.<br/>
 <span class="image-wrap" style=""><a id="66760_thumb" href="http://forge.prestashop.com/secure/attachment/66760/66760_image-2018-04-06-18-52-24-153.png" title="image-2018-04-06-18-52-24-153.png" file-preview-type="image" file-preview-id="66760" file-preview-title="image-2018-04-06-18-52-24-153.png"></a></span> </p>     

## SPECS

The currenct empty option should be "Choose a brand"

marionf:
| Questions     | Answers
| Branch?       | develop 
| Description?  | Remove empty option in select brand of product page
| Type?         | bug fix 
| Category?     | BO
| Fixed ticket? | Fixes #9678
| How to test?  | See issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17679)

### DIFF
--- a/src/PrestaShopBundle/Form/Admin/Product/ProductInformation.php
+++ b/src/PrestaShopBundle/Form/Admin/Product/ProductInformation.php
@@ -222,6 +222,7 @@ class ProductInformation extends CommonAbstractType
                 'allow_delete' => true,
             ])
             ->add('id_manufacturer', FormType\ChoiceType::class, [
+                'placeholder' => false,
                 'choices' => $this->manufacturers,
                 'required' => false,
                 'attr' => [


src/PrestaShopBundle/Form/Admin/Product/ProductInformation.php
```diff
@@ -222,6 +222,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'allow_delete' => true,
             ])
             ->add('id_manufacturer', FormType\ChoiceType::class, [
+                'placeholder' => false,
```
arouiadib, line 225: I have tested under branch 1.7.7.x and it successfully removes the empty option in select brand of product page. However, when I create new product, the first brand in list is now automatically applied to product. That empty option is (maybe) what tells Prestashop to not store any brand for the product until the merchant decides it. What do you think of using the `attr` to add a class (with some CSS properties that only hide and not remove option from DOM)

Progi1984 (approved):

src/PrestaShopBundle/Form/Admin/Product/ProductInformation.php
NeOMakinG: That's too tricky for me, maybe we can manage to get an option displaying "Select brand" which is empty ? That would be better than an empty option and avoid tricky css things

NeOMakinG (changes requested):

NeOMakinG:
As the @arouiadib comment say, it looks like it's now auto adding a brand to a product, that's probably not the behavior we wants

marionf:
> However, when I create new product, the first brand in list is now automatically applied to product.

Nice catch @arouiadib !

> That's too tricky for me, maybe we can manage to get an option displaying "Select brand" which is empty ? That would be better than an empty option and avoid tricky css things

Yes good idea @NeOMakinG 
I will add this in the specs to get it done during the migration of the page

src/PrestaShopBundle/Form/Admin/Product/ProductInformation.php
arouiadib: @NeOMakinG , you are right, your option is better than CSS "hiding".

PR #17702, opened 2020-02-17
Order Page: Keep email failure from blocking product cancellation (and some refactor)

### LINKED ISSUE #17615: Error when canceling last product of an order
**Describe the bug**
When using the "Cancel product" functionality, if all products are canceled (in one or multiple times), there is an error showing, even if the cancellation is OK. The error prints "An error occurred while attempting to delete the product X".

This is present in Legacy 176, Legacy 177 and Migrated page 177.

**To Reproduce**
1. Activate merchandise return
2. Go to an unpaid order
3. Cancel all products of the order (using the "Cancel products" button)
4. An error is displayed

**Screenshots**
pr17149 cancel all legacy

**Additional information**
PrestaShop version: 1.7.6, 1.7.7
PHP version: N/A

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When all products from an order are cancelled in the back office, and the email service is not configured properly on the user's machine, we should not have an error message, just a warning log instead. 
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? |no
| Fixed ticket? | Fixes #17615
| How to test?  | - In FO, create an order<br>- In BO, cancel all products from this order<br> - You should not have any error message even if your mail is not working

Migrated order page:  **/admin-dev/sell/orders/orders/15/view** (replace 15 with your order id)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17702)


src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -0,0 +1,254 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Adapter\Order\Refund;
```
matthieu-rolland, line 27: This file contains the legacy logic of product deletion from an order, for now it is used only by the cancel product feature on order page.

I added dedicated exceptions instead of returning false when something went wrong, and split the logic into several functions, renamed variables to follow our naming standards.

I also logged an error message when the email could not be sent, making it non blocking. (that was the initial goal of this PR)

src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -0,0 +1,254 @@
...
+        $cart = new Cart($order->id_cart);
+        $this->updateCart($cart, $orderDetail, $quantity);
+
+        $packageShippingCostTaxIncl = $cart->getPackageShippingCost(
+            $order->id_carrier,
+            true,
+            null,
+            $order->getCartProducts()
+        );
+        $packageShippingCostTaxExcl = $cart->getPackageShippingCost($order->id_carrier,
```
Progi1984:
```suggestion
        $packageShippingCostTaxExcl = $cart->getPackageShippingCost(
            $order->id_carrier,
```

src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -0,0 +1,254 @@
...
+            if (count($order->getProductsDetail()) == 0) {
+                $history = new OrderHistory();
+                $history->id_order = (int) $order->id;
+                $history->changeIdOrderState(Configuration::get('PS_OS_CANCELED'), $order);
+                if (!$history->addWithemail()) {
+                    // email failure must not block order update process
+                    $this->logger->warning(
+                        'Order history email could not be sent, test your email configuration in the Advanced Parameters > E-mail section of your back office'
+                    );
+                    $order->update();
```
Progi1984:
```suggestion
```

src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -0,0 +1,254 @@
...
+     * @param OrderDetail $orderDetail
+     * @param int $quantity
+     */
+    private function deleteCustomization(Order $order, OrderDetail $orderDetail, int $quantity)
+    {
+        if (!(int) $order->getCurrentState()) {
+            throw new DeleteCustomizedProductFromOrderException('Could not get a valid Order state before deletion');
+        }
+
+        if ($order->hasBeenDelivered()) {
```
Progi1984:
```suggestion
        if ($order->hasBeenDelivered() || $order->hasBeenPaid()) {
```

src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -0,0 +1,254 @@
...
+            return Db::getInstance()->execute('UPDATE `' . _DB_PREFIX_ . 'customization` SET `quantity_returned` = `quantity_returned` + ' . (int) $quantity . ' WHERE `id_customization` = ' . (int) $orderDetail->id_customization . ' AND `id_cart` = ' . (int) $order->id_cart . ' AND `id_product` = ' . (int) $orderDetail->product_id);
+        } elseif ($order->hasBeenPaid()) {
```
Progi1984: Remove next three lines

Progi1984 (commented):
A feedback

Progi1984:
And PrettyCI is :red_circle:

src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -0,0 +1,265 @@
...
+
+    /**
+     * @param Order $order
+     * @param OrderDetail $orderDetail
+     * @param int $quantity
+     * @param bool $isCustomized
+     */
+    public function deleteProductFromOrder(Order $order, OrderDetail $orderDetail, int $quantity, $isCustomized = false)
+    {
+        if ($isCustomized) {
```
jolelievre: You don't need this boolean since you have the info in `$orderDetail->id_customization`

src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -0,0 +1,265 @@
...
+     * @param int $quantity
+     */
+    private function deleteCustomization(Order $order, OrderDetail $orderDetail, int $quantity)
+    {
+        if (!(int) $order->getCurrentState()) {
+            throw new DeleteCustomizedProductFromOrderException('Could not get a valid Order state before deletion');
+        }
+
+        if ($order->hasBeenDelivered()) {
+            return Db::getInstance()->execute('UPDATE `' . _DB_PREFIX_ . 'customization` SET `quantity_returned` = `quantity_returned` + ' . (int) $quantity . ' WHERE `id_customization` = ' . (int) $orderDetail->id_customization . ' AND `id_cart` = ' . (int) $order->id_cart . ' AND `id_product` = ' . (int) $orderDetail->product_id);
```
jolelievre, line 253: Wouldn't it be simpler to use the object to update this data?

Progi1984 (dismissed):

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
```diff
@@ -138,9 +136,7 @@ public function handle(CancelOrderProductCommand $command)
         if (!empty($orderDetails['customizedProductsOrderDetail'])) {
             foreach ($orderDetails['customizedProductsOrderDetail'] as $orderDetail) {
                 $qtyCancelProduct = abs($orderDetails['customizedCancelQuantity'][$orderDetail->id_customization]);
-                if (!$order->deleteCustomization($orderDetail->id_customization, $qtyCancelProduct, $orderDetail)) {
-                    $this->errors[] = $this->translator->trans('An error occurred while attempting to delete product customization.', [], 'Admin.Orderscustomers.Notification') . ' ' . $id_customization;
-                }
+                $this->orderProductRemover->deleteProductFromOrder($order, $orderDetail, $qtyCancelProduct, true);
```
jolelievre: You could get rid of the last parameter (since the info is already in the the OrderDetail object)
This way you could even get rid of this loop

Besides I'm worried that calling this method twice (in case there is a customization) will decrement the quantity twice as well (unless the `$orderDetails['productsOrderDetails']` only contains OrderDetail without customization, then they will not overlap)

matks (dismissed):

src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -0,0 +1,265 @@
...
+            if (!$orderDetail->delete()) {
+                throw new DeleteProductFromOrderException('Could not delete order detail');
+            }
+            if (count($order->getProductsDetail()) == 0) {
+                $history = new OrderHistory();
+                $history->id_order = (int) $order->id;
+                $history->changeIdOrderState(Configuration::get('PS_OS_CANCELED'), $order);
+                if (!$history->addWithemail()) {
+                    // email failure must not block order update process
+                    $this->logger->warning(
```
jolelievre, line 222: This logging is nice, but do you think there would be a way to display it to the user?
The delete action would work fine (displaying a success block), but it would also display a warning block regarding the mail sending failure

src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -0,0 +1,265 @@
...
+                        $this->translator->trans(
+                            'Order history email could not be sent, test your email configuration in the Advanced Parameters > E-mail section of your back office.',
+                            [],
```
matthieu-rolland, line 225: ping @LouiseBonnard   

So I closed the old PR with the new error message when the mail is not working.

In this PR, we do not block the feature if mail notification is not working, and we do not display an error message (a shop is not forced to have emails working). Instead we log a warning message dedicated to the mail issue.

I was inspired by your old email error message, I only kept the part about checking the mail configuration.

matthieu-rolland: (this is a "technical" warning message that appears in the logs only)

src/Adapter/Order/Refund/OrderProductRemover.php
matthieu-rolland: This class is a wrapper for the legacy code that is actually being used today when you cancel a product. 

The scope of this PR is only to make email failure non blocking 😓

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
matthieu-rolland: Well this process is called once per product today (both in legacy and on 1.7.7.x) , so we are iso legacy. 
But I'll still have a look at what you say, in case I could fix an existing bug 🤔

src/Adapter/Order/Refund/OrderProductRemover.php
matthieu-rolland: As seen with @eternoendless and the product team , the rule we stood by is:

If mail failure is blocking the feature: error message
If not:  logging message as a warning

(we are in case 2)

Being able to send emails is not mandatory for a shop.

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
matthieu-rolland: yes `$orderDetails['productsOrderDetails']` only contains OrderDetail without customization.

src/Adapter/Order/Refund/OrderProductRemover.php
LouiseBonnard: Do you know what is the name of the email that could not have been sent in this case?

matks (approved):

Progi1984 (approved):

src/Adapter/Order/Refund/OrderProductRemover.php
matthieu-rolland: It seems like the title of the mail is `[name of the store] Canceled"`

It is a mail that is sent when the status of an order has changed.

Progi1984:
Thanks @matthieu-rolland

PR #17660, opened 2020-02-13
Migration of order view page printing functionality and layout changes - part 4

matks:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | Rebuild of PR https://github.com/PrestaShop/PrestaShop/pull/17617 . When clicked "print" page display its version ready for print. There are some layout changes when print is being initialized.
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #15819 
| How to test?  | Try clicking print or ctrl + p while in this page **admin-dev/index.php/sell/orders/orders/1/view**. It should be similar to invision sketch of when page is printed. **Dont forget to rebuild assets**

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17660)

PR #17715, opened 2020-02-18
Fix print css

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Remove useless faulty CSS class introduced in https://github.com/PrestaShop/PrestaShop/pull/17660/
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 

I can confirm this CSS class is useless. It was introduced in https://github.com/PrestaShop/PrestaShop/pull/16262 which was used to build https://github.com/PrestaShop/PrestaShop/pull/17660/ but it's actually never used as markup as changed since.

Old markup :
https://github.com/PrestaShop/PrestaShop/pull/16262/files#diff-0df196bc43fe2c160b71a12725624750R26

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17715)


Progi1984 (approved):

matthieu-rolland (approved):

matks:
Thanks for review ! It's a harmless refacto, no need for QA

Issue #15649, opened 2019-09-23
State input disappearing shorly after page is loaded


matks:
On migrated page "Catalog > Brand & Suppliers > Suppliers create/edit action" **migrated** page, the "State" input disappears shortly after page has opened.

This is due to the fact that selected Country has no States, so javascript is supposed to hide this input (so the State drop-down is useless). However javascript loads a bit late and hides it a bit late, which leads to this visual bug.

matks:
This issue is part of migration issue https://github.com/PrestaShop/PrestaShop/issues/10541

matks:
See video https://drive.google.com/file/d/106r7hStShrJYZw-yLGMxr6ftkKDdwEru/view

matks:
Fixed by https://github.com/PrestaShop/PrestaShop/pull/17742

PR #17742, opened 2020-02-19
Fix country states input toggle on page load

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | On page load of address form (manufacturer, customer, supplier) js used to fadeOut the states field if the country doesn't have states, but small flickering was noticable. Added d-none classes in formType of tempalte depending from states, so it doesn't need the javascript on page load, but uses it on change only. For more info check https://github.com/PrestaShop/PrestaShop/issues/15649
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #15649
| How to test?  | Open addresses (manufacturer, customer, supplier) for country other than United states. Country states input shouldn't be visible. If you select United States, then states input should fadeIn. More here #15649

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17742)


Progi1984 (dismissed):

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Manufacturer/Address/Blocks/form.html.twig
```diff
@@ -72,7 +72,7 @@
           'label': 'Country'|trans({}, 'Admin.Global'),
       }) }}
 
-      <div class="js-manufacturer-address-state">
+      <div class="js-manufacturer-address-state {{ addressForm.children.id_state.vars.attr.class }}">
```
NeOMakinG: Would it be possible to avoid front-end classes coming from the back ? Like this :
```suggestion
      <div class="js-manufacturer-address-state{% if addressForm.children.id_state.vars.isHidden %} d-none{% endif %}">
```

NeOMakinG (changes requested):

matthieu-rolland (dismissed):
LGTM ! I don't put the waiting for qa label since there is still a requested changes from @NeOMakinG

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Manufacturer/Address/Blocks/form.html.twig
zuk3975: https://github.com/PrestaShop/PrestaShop/pull/16938/files#diff-bd2da52b9459a118b619e517064ca188R431
It seems that with new form rendering we are still doing all of that in backend (even tho I agree it doesn't sound right) :man_shrugging: .
Another way would be to pass vars from controller to twig, but thats an overhead.
And yet another way is to pass Hidden attribute, but its also kinda same as class, just this case would require even more code to make js toggle component work.

NeOMakinG: It's not as same as a class,  because it means we can still modify the class and css linked to it without going to backend, while passing a complete class name force you to work on the backend if you want to modify the class. Even if we already did it somewhere, it doesn't mean that it make sense 😄 

We should try to separate as much as possible the front-end job from the backend one, as we are thinking of a transition to VueJS, this is such a thing we have to avoid absolutely in order to split as much as possible the back from the front event if we're on a monolithic application.

Also an other point it that we don't know at all what that `{{ addressForm.children.id_state.vars.attr.class }}` does while reading the twig, but with `{% if addressForm.children.id_state.vars.isHidden %} d-none{% endif %}` we instantly know what this does, it's just a question about splitting backend logic and front-end one :)

NeOMakinG: @PrestaShop/prestashop-core-developers wdyt ?

PierreRambaud: Agree with @NeOMakinG 

but please use that:
```suggestion
      <div class="js-manufacturer-address-state{% if addressForm.children.id_state.vars.choices is empty %} d-none{% endif %}">
```

And please remove useless `class` in the Type files.

PierreRambaud (changes requested):
Don't set Css class in php if it's possible please

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Manufacturer/Address/Blocks/form.html.twig
zuk3975: so your final suggestion is to try work it out with hidden attribute? This would mean some tweaking in js component (but shouldn't be very hard) ?

zuk3975:
> Don't set Css class in php if it's possible please

Ok. 
bit offtopic, but what about the new form rendering where we want to render it by single form_widget() and everything else is hidden in php formType?

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Manufacturer/Address/Blocks/form.html.twig
zuk3975: that works. thanks

zuk3975:
> Don't set Css class in php if it's possible please

Ok. Done.
bit offtopic, but what about the new form rendering where we want to render it by single form_widget() and everything else is hidden in php formType?

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
```diff
@@ -112,15 +112,11 @@
...
-      {% if addressForm.id_state.vars.disabled is defined and addressForm.id_state.vars.disabled %}
-        {% set rowAttr = {'style': 'display:none'} %}
-      {% endif %}
-
-      {{ ps.form_group_row(addressForm.id_state, {}, {
-        'label': 'State'|trans({}, 'Admin.Global'),
-        'class': 'js-address-state-select',
-        'row_attr': rowAttr|default({})
-      }) }}
+      <div class="js-address-state-select {% if addressForm.id_state.vars.choices is empty %} d-none{% endif %}">
```
PierreRambaud:
```suggestion
      <div class="js-address-state-select{% if addressForm.id_state.vars.choices is empty %} d-none{% endif %}">
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Manufacturer/Address/Blocks/form.html.twig
```diff
@@ -72,7 +72,7 @@
           'label': 'Country'|trans({}, 'Admin.Global'),
       }) }}
 
-      <div class="js-manufacturer-address-state">
+      <div class="js-manufacturer-address-state {% if addressForm.id_state.vars.choices is empty %} d-none{% endif %}">
```
PierreRambaud:
```suggestion
      <div class="js-manufacturer-address-state{% if addressForm.id_state.vars.choices is empty %} d-none{% endif %}">
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Catalog/Suppliers/Blocks/form.html.twig
```diff
@@ -90,7 +90,7 @@
         'label': 'Country'|trans({}, 'Admin.Global'),
       }) }}
 
-      <div class="js-supplier-state">
+      <div class="js-supplier-state {% if supplierForm.id_state.vars.choices is empty %} d-none{% endif %}">
```
PierreRambaud:
```suggestion
      <div class="js-supplier-state{% if supplierForm.id_state.vars.choices is empty %} d-none{% endif %}">
```

NeOMakinG (approved):
LGTM for the front-end part

matks (approved):

PierreRambaud (approved):

matks:
Thank you @zuk3975

PR #17752, opened 2020-02-20
Add Prev/Next order button in the Order page

### LINKED ISSUE #17648: [ORDER PAGE] Prev/Next order button
Missing part of the order page migration and redesign project.

**User Storie**

As a merchant, I want to go to the next order without leaving the page.

**Detailed specification**
On the action bar button, add the arrow next prev on the left
image

WIP specs to update it
https://invis.io/YKTGGFZAMCB#/388660077__Order_Details_-_Private_Note

tdavidsonas88:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | As a merchant, I want to go to the next order without leaving the page. On the action bar button, added the arrow next prev on the left
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/17648
| How to test?  | select order and interact with next/prev buttons on the top right of the Order view page

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17752)


src/Core/Domain/Order/Presenter/OrderLinkPresenter.php
```diff
@@ -0,0 +1,56 @@
...
+    private $urlGenerator;
+
+    public function __construct(UrlGeneratorInterface $urlGenerator)
+    {
+        $this->urlGenerator = $urlGenerator;
+    }
+
+    public function present(?int $orderId): array
+    {
+        $order = new Order($orderId);
```
matks: What happens if you pass `null` to `new Order()` ? Nothing ?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1523,4 +1528,40 @@ private function getCustomerMessageErrorMapping(Exception $exception): array
...
+    /**
+     * @param int $orderId
+     *
+     * @return int|null
+     */
+    private function getPreviousOrderId(int $orderId)
+    {
+        try {
+            /** @var OrderForViewing $previousOrderForViewing */
+            $previousOrderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId - 1));
```
matks: We cannot use `$orderId - 1)` that's the tricky part of this feature 😄 
- if someone manually deletes a row from mysql it does not work
- if mysql jumps one increment (can happen) it does not work

We need to retrieve the real `previous MySQL ID` which is __probably__ "$orderId - 1" but we need to be sure.

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1523,4 +1528,40 @@ private function getCustomerMessageErrorMapping(Exception $exception): array
...
+    /**
+     * @param int $orderId
+     *
+     * @return int|null
+     */
+    private function getNextOrderId(int $orderId)
+    {
+        try {
+            /** @var OrderForViewing $nextOrderForViewing */
+            $nextOrderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId + 1));
```
matks: Same for `$orderId + 1` : we cannot rely on MySQL properly incrementing IDs

tdavidsonas88:
finished with design: 
image

@matks Ready for review!

src/Adapter/Presenter/Order/OrderLinkPresenter.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * @param int $orderId
+     *
+     * @return array
+     */
+    public function present($orderId): array
+    {
+        $order = new Order($orderId);
```
matks: What happens if `$orderId` is null ?
What happens for `$order->reference` ?

src/Adapter/Order/OrderSiblingProvider.php
```diff
@@ -0,0 +1,29 @@
...
+class OrderSiblingProvider implements OrderSiblingProviderInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getNextOrder(int $orderId): int
+    {
+        $order = new Order($orderId);
+
+        return $order->getNextOrderId();
```
matks: Nice finding !

matks (dismissed):

src/Adapter/Order/OrderSiblingProvider.php
tdavidsonas88: it's @jolelievre finding yesterday :slightly_smiling_face:

jolelievre (dismissed):
I think this solution is overkill, we add a new layer of OrderPresenter, the solution we discussed about extending OrderForViewing seemed more straightforward
Besides in this cas you have to instanciate two Order object to fetch two IDs which also seems overkill

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_navigation.html.twig
```diff
@@ -0,0 +1,42 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ *#}
+
+<div class="col-xs-2 mb-3 row ml-auto">
+  <div class="col-xs-1 d-inline mr-1">
+    {% if previousOrder.enabled %}
+      <a href="{{ previousOrder.href }}" class="btn btn-action">
```
jolelievre: When the previous/next order is not present the button must be `disabled` not hidden

admin-dev/themes/new-theme/scss/pages/_orders_view.scss
```diff
@@ -64,6 +64,7 @@
       font-weight: normal;
     }
   }
+
```
Progi1984:
```suggestion
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_navigation.html.twig
tdavidsonas88: changed it: 
image

src/Adapter/Order/OrderSiblingProvider.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
```
Progi1984, line 28: Miss license headers & declare strict_types to 1

src/Adapter/Presenter/Order/OrderLinkPresenter.php
```diff
@@ -0,0 +1,62 @@
...
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+class OrderLinkPresenter implements PresenterInterface
+{
+    /**
+     * @var UrlGeneratorInterface
+     */
+    private $urlGenerator;
+
+    public function __construct(UrlGeneratorInterface $urlGenerator)
```
Progi1984:
```suggestion
    public function __construct(UrlGeneratorInterface $urlGenerator): void
```

src/Core/Order/OrderSiblingProviderInterface.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
```
Progi1984, line 28: Miss license headers & declare strict_types to 1

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_navigation.html.twig
```diff
@@ -0,0 +1,42 @@
...
+        <span class="material-icons">arrow_back</span>
+      </a>
+    {% endif %}
+  </div>
```
Progi1984:
```suggestion
  {% if previousOrder.enabled %}
    <div class="col-xs-1 d-inline mr-1">
      <a href="{{ previousOrder.href }}" class="btn btn-action">
        <span class="material-icons">arrow_back</span>
      </a>
    </div>
  {% endif %}
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_navigation.html.twig
```diff
@@ -0,0 +1,42 @@
...
+
+  <div class="col-xs-1 ui-button d-inline">
+    {% if nextOrder.enabled %}
+      <a href="{{ nextOrder.href }}" class="btn btn-action">
+        <span class="material-icons">arrow_forward</span>
+      </a>
+    {% endif %}
+  </div>
```
Progi1984:
```suggestion
  {% if nextOrder.enabled %}
    <div class="col-xs-1 ui-button d-inline">
      <a href="{{ nextOrder.href }}" class="btn btn-action">
        <span class="material-icons">arrow_forward</span>
      </a>
    </div>
  {% endif %}
```

admin-dev/themes/new-theme/scss/components/_button.scss
```diff
@@ -5,10 +5,11 @@
     border-radius: 4px;
     box-shadow: 0 0 10px 0 rgba(0, 0, 0, 0.1);
 
-    &:disabled {
+    &:disabled, &.disabled {
       color: #fff;
       background-color: #dddddd;
       box-shadow: none;
     }
+
```
Progi1984:
```suggestion
```

src/Adapter/Presenter/Order/OrderLinkPresenter.php
tdavidsonas88: sorry but for constructor return type declaration is not allowed

tdavidsonas88:
> I think this solution is overkill, we add a new layer of OrderPresenter, the solution we discussed about extending  seemed more straightforward
> Besides in this cas you have to instanciate two Order object to fetch two IDs which also seems overkill

It seems like it's simpler to add params to GetOrderForViewing and use them without DataProvider. However OrderForViewing constructor is already very big and adding two extra params with null default values might just bring some more complexity to Behat tests. Both solutions are good, it would be nice if @matks and @jolelievre would stick with one of them!

src/Adapter/Presenter/Order/OrderLinkPresenter.php
matks: @tdavidsonas88 I still need the answer 😄

matks (dismissed):

src/Adapter/Presenter/Order/OrderLinkPresenter.php
tdavidsonas88: I think, it says it expected integer and breaks that's why I've used (int) : return (int) $order->getPreviousOrderId(); Let me improve this tomorrow morning :slightly_smiling_face:

tdavidsonas88: But most likely creates the new object instead of loading the existing one.

tdavidsonas88: Removed the Order object creation because it was no longer used after changes in the UI design according to task

matks (approved):

Progi1984 (approved):

Issue #17747, opened 2020-02-19
Exception when adding a new discount with invalid amount - invalid discount


khouloudbelguith:
**Describe the bug**
In the BO => Orders => new migrated Order details page or legacy page when we add a new discount with an invalid amount for example `-10` => an exception is displayed

**To Reproduce**
1. Go to BO => Orders => new migrated Order details page `admin-dev/index.php/sell/orders/orders/5/view` or the legacy page
2. Click on "Add a new discount" button
2.1. Type = Amount
2.2. Value = -10
3. Click Add
![image](https://user-images.githubusercontent.com/16067358/74853905-da135800-533e-11ea-8006-6f9496b2e6ba.png)
![image](https://user-images.githubusercontent.com/16067358/74853952-ea2b3780-533e-11ea-8c23-5527f7b66fc7.png)

**Additional information**
PrestaShop version: 1.7.7.x, 1.7.6.3, 1.7.5.2
PHP version: 7.1

khouloudbelguith:
Ping @colinegin new issue in the Order page

khouloudbelguith:
Hi,

Same issue with an invalid percent value
An exception detected when we add a new discount = `-10%`
![image](https://user-images.githubusercontent.com/16067358/74854942-4b9fd600-5340-11ea-811f-d4f34e51abe7.png)

![image](https://user-images.githubusercontent.com/16067358/74854891-3b87f680-5340-11ea-9b21-e82e3f0424f3.png)

Reproduced with 177x branch, PS1.7.6.3 and PS1.7.5.2

Thanks!

Hlavtox:
Isn't this the same as this? https://github.com/PrestaShop/PrestaShop/issues/13809

khouloudbelguith:
Hi @Hlavtox, 

We have different exceptions:
#13809: Property Order->total_paid is not valid
#17747: Property CartRule->reduction_prcent is not valid / Property Order->total_discounts is not valid

They are not similar but they are just related.

Thanks!

colinegin:
The behaviour should be the same as in the cart rule page. When adding a discount value (in %) containing a "-", this error message is displayed : 
![image](https://user-images.githubusercontent.com/35263827/75331055-5dbece80-5882-11ea-9907-ba790ebb60a9.png)

And when adding a discount value (in amount) containing a "-", this error message is displayed : 
[...]

PR #17790, opened 2020-02-23
Fix Exception when adding a new discount with invalid amount

arouiadib:
| Questions     | Answers
| Branch?       | 1.7.7.x 
| Description?  | This PR is about validation of form "add discount", in order view page. It is inspired by the existing job done in `Reduction` and `ReductionValidator` classes. It rewrites them so that constraints are directly used in FormType. It also adds conditional constraint (because value validation depends on type) using `PropretyAccess`.  
| Type?         | bug fix 
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #17747
| How to test?  | Try to add a discount from the new order view page (`/sell/orders/orders/3/view`) with invalid values, like percentage `<0` or `>100`, or amount `<0` or `>total`. You will see corresponding error messages on the page instead of PrestashopException page.

PS: This PR needs a decision about the naming, whether it is `percent` or `percentage`.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17790)

### DIFF
--- a/src/Core/ConstraintValidator/Constraints/ValidReductionType.php
+++ b/src/Core/ConstraintValidator/Constraints/ValidReductionType.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints;
+
+use PrestaShop\PrestaShop\Core\ConstraintValidator\ValidReductionTypeValidator;
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Constraint for validating reduction type
+ */
+final class ValidReductionType extends Constraint
+{
+    public $message = 'Reduction type "%type%" is invalid. Allowed types are: %types%.';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function validatedBy()
+    {
+        return ValidReductionTypeValidator::class;
+    }
+}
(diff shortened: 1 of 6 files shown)


matks:
@arouiadib This is a very nice PR and a very nice idea 😊 thanks !

src/Core/ConstraintValidator/Constraints/ValidReductionValue.php
```diff
@@ -0,0 +1,58 @@
...
+
+use PrestaShop\PrestaShop\Core\ConstraintValidator\ValidReductionValueValidator;
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Constraint for validating reduction value based on reduction type
+ */
+final class ValidReductionValue extends Constraint
+{
+    public $invalidAmountValueMessage = 'Reduction value "%value%" is invalid. Value cannot be negative';
```
PierreRambaud: This variable are not going to change, why not using const?

arouiadib: Way better if constants are used :+1: . I changed the code for this one.
Thank you very much @PierreRambaud

src/Core/ConstraintValidator/Constraints/ValidReductionType.php
```diff
@@ -0,0 +1,46 @@
...
+
+use PrestaShop\PrestaShop\Core\ConstraintValidator\ValidReductionTypeValidator;
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Constraint for validating reduction type
+ */
+final class ValidReductionType extends Constraint
+{
+    public $message = 'Reduction type "%type%" is invalid. Allowed types are: %types%.';
```
PierreRambaud, line 37: This one is really used? if yes, can be a const :thinking:

arouiadib: Good point @PierreRambaud , I didn't change it because I feel the whole class is to be removed. I just need your confirmation. In fact, do we really need to validate the type? The type come from an `option` in a `select` which is rendered based on predefined values (`percent`,`amount`,`free_shipping`) so there is a very small probability that the value sent back to server doesn't belong to this set. In the case someone plays with an `option` value client-side, there is an existing exception that prints `Invalid value`. So I think there is no need of this class.  But if you tell me it is necessary to validate this field with custom error message, I can try  first with the Symfony built-in constraint named [`Choice`](https://symfony.com/doc/current/reference/constraints/Choice.html). If it doesn't satisfy the need, I can keep this custom constraint (with constant of course). The truth is that I created this class because I based my work on existing `Reduction` constraint without giving it a deep thought.  Wdyt? Thank you very much for your time

PierreRambaud: I think this kind of validation is needed because even if we use form when it's possible, changes can be done by CQRS commands for example. But I also think @matks or @jolelievre are better than me to answer you correctly (they're our CQRS experts!)
(really sorry for the delay, little bit busy :sweat_smile: )

matks: I think it would be nice to have this ValidReductionType class. It could be a helper class for developers who want to use it and easily strengthen the user experience of their custom form.

However if we can reuse `Choice` logic it's better indeed. So maybe solution is a simple children class:
```
<?php

class ValidReductionType extends Choice
{
    public $choices = ['percentage', 'amount', 'free_shipping'];

}
```

Something like that ?

Another nice point of having a dedicated Constraint class is that @LouiseBonnard can customize the error message we want to display to user.

marionf:
Guys, maybe we can close this PR as the issue #17747 has been fixed by #17852 ?

Progi1984:
ping @atomiix

arouiadib:
Totally agree. @atomiix 's PR serves the same purpose.

Issue #14388, opened 2019-06-27
Order details only 2 customer messages appear


siomosp:
**Describe the bug**

**To Reproduce**
1. Place an order
2. Add 3 or more  messages to order
3. Only the first two messages appear at order details page

**Screenshots**
If applicable, add screenshots or screenrecords to help explain your problem.

**Additionnal information**
PrestaShop version: 1.7.5.2
PHP version: 7.0.x
![PS1752_Order_details_messages](https://user-images.githubusercontent.com/4071032/60254139-0a9d1200-98d6-11e9-8ad2-d01145a425d5.jpg)

khouloudbelguith:
Hi @siomosp,

No, It is not the first & second message.
They are the last two messages, it is normal behavior, you can click on "Show all message" to show them all.
I attached a screen record
https://drive.google.com/file/d/1_NdOLd9mHfmMBLWJzG5oDh0crEGjEXJT/view
Thanks to check & feedback.

siomosp:
Hi,
thanks for the reply!
There is an option to display more than last 2 meesages?
For example last 4 messages?
Thank you

khouloudbelguith:
@siomosp, sorry no.
But this is could be an improvement added to the order page after the migration to Symfony: #13878.
ping @matks, @colinegin what do you think?
Can we have an option to display the number of the latest message?

Thanks!

TristanLDD:
Hi @siomosp, thank you for your feedback, we are currently working on the order page and this problem has been signaled by several merchants. We will do our best to address it as soon as possible.

colinegin:
@MatShir could you give us an update on the status of this issue ?

MatShir:
@colinegin We consider adding the feature in the order page redesign. To check the status of the issue and the design results follow the issues #14745, it's currently in work in progress in @TristanLDD hand's.

MatShir:
Fixed by #17801

PR #17734, opened 2020-02-19
Register new hook displayFooterCategory for 1770 and rename order hooks

matks:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | Register new hook displayFooterCategory introduced in https://github.com/PrestaShop/PrestaShop/pull/17410. It's [already registered](https://github.com/PrestaShop/PrestaShop/pull/17477) for `develop` but not for `1.7.7.x`. Also rename some of the order view page hooks
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18084
| How to test?  | To validate all the hooks you will need the order example module using the branch from this PR https://github.com/PrestaShop/example-modules/pull/15

⚠️ **Merge this PR AFTER https://github.com/PrestaShop/PrestaShop/pull/17733** ✓
⚠️ **When merging 177x into develop, this will likely create conflicts with https://github.com/PrestaShop/PrestaShop/pull/17477**

To validate the `displayFooterCategory` hook you can use this module:
[footercategory.zip](https://github.com/PrestaShop/PrestaShop/files/4317802/footercategory.zip)

To validate the order view page hooks you can read their description in the test module Readme file:
https://github.com/PrestaShop/example-modules/blob/ed93cbd9757c851f8efc3c6e8fecbc31ebcb0847/demovieworderhooks/README.md

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17734)

PR #17801, opened 2020-02-24
View and Create Order feedbacks corrections

### LINKED ISSUE #17655: [ORDER PAGE] Display correction
Go through in the missed display and see what could be improved

## Part 1 
**ORDER VIEW**

Need a dev back on these issues : 

- [x] Empty rectangle below the tabs (status, transport...)

- [x] The stock location is displayed in cancel, partial, standard and return products while none of the products have a stock location. It should only display it if only one of them has a stock location.

- [x] The refunded column is always displayed by default even if the merchandise return is deactivated and if no refund has been made.

- [x] Border between `%` (or `€`) and the value field in the "Add a discount" popup is too thick. Adding the class `input-group-text` on the value field should solve it.

- [x] Check if we can dynamically change the color of the top left status dropdown according to the selected status. 

Front-end issues :

- [x] Put the label of the Order list on the Order page in the status history

- [x] The label is not well rounded in a stock greater than two in the product list of the order page.

- [x] The status change does not correspond to the expected display.

- [x] The button to edit the merchant's address page goes below the label when the screen size is reduced.

- [x] In a reduced screen, the button that goes to the line is too close to the drop-down menu to update the status (see on the mobile view).

- [x] Pop-in of messages in a reduced screen makes the profiles icons all line up on the left while some should stay on the right.

- [x] The block of messages in a reduced screen puts the icons below the message when it should stay above the message.
Change the number of messages displayed to 4 instead of 2.

- [x] Adding a discount.
The discount value when you want to increase the value of the field using the top arrow increases the value by 0.01. Can it be changed to 1 or maybe 10?

- [x] The trash icon in the discount list is in blue, whereas it should only be displayed in blue when the mouse passes over it.

- [x] Review the display (alignment of the input with the label) of the amount to be entered when the refund, partial refund and standard refund buttons interact.

- [x] The numbers labeled in black in the customer block have a size that looks smaller than normal.
- [x] The numbers labeled in black in the pop-in of the product packs are smaller than normal.
- [x] The link to see the packs is too close to the product name.

- [x] Added ":" after custom text labels in the product list when displaying a customized product.
- [x] Review the partial refund shipping field.

- [x] All the fields in add and edit are not aligned
- [x] The "configure message" link is too much of a dropdown in the message block.
- [x] See if there are any changes or comments to be made in the mobile view. 
- [ ] ~~The radio does not look like in the ui kit design when using a partial refund with a discount~~ --> not to be done in 177

- [x] Mobile: center the content of the main buttons (top of the page buttons)

### LINKED ISSUE #17656: [CREATE ORDER] Display correction #2
Review the missed display and see what could be improved.
**CREATE ORDER**

- [x] The button to unfold the old baskets and control is reversed.
- [x] The file name is not displayed after selecting the file.

### LINKED ISSUE #17867: Order view all products becomes refunded
**Describe the bug**
If I open order in migrated page and use partial refund of one product, then trying to add new products results in all of them being `refunded`

**To Reproduce**
1. Go to `admin-dev/orders/orders/[orderId]/view` to reach migrated order view page
2. Make sure status is payment accepted.
3. Click partial refund -> fill quantity -> refund.
4. See that product is marked with 'refunded' in list.
5. Search for another product -> add
6. See that this added product is also marked with 'refunded'

**Expected behavior**
Only refunded product must be marked with 'refunded'.

**Screenshots**
Migrated page: Screenshot from 2020-02-27 14-43-41 Legacy page: Screenshot from 2020-02-27 14-44-20

**Additional information**
PrestaShop version: 1.7.7.x
PHP version: N/A

NeOMakinG:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | There are a lot of corrections to do regarding the view order pages
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #17655 #17656 #17867 #18378 #14388 
| How to test?  | Go on issues linked before and point by point, verify if it is correctly adjusted. Please rebuild assets.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17801)


matks (dismissed):
Although I am very bad at frontend, I think you code looks cool  😎

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
```diff
@@ -252,17 +248,29 @@ export default class ProductRenderer {
...
-      $template.find(createOrderMap.productCustomInput)
+      if (customField.type === fieldTypeFile) {
+        $template.on('change', e => {
+          const fileName = e.target.files[0].name;
+
+          $(e.target)
+            .next('.custom-file-label')
+            .html(fileName);
+        });
+      }
```
NeOMakinG, line 266: This is the modification, others is linting/correcting, should have pushed on another commit, sorry for the reviewers

admin-dev/themes/new-theme/scss/config/_settings.scss
```diff
@@ -8,6 +7,7 @@ $background-grey: #eaebec;
 $gray-dark-text-hover: #ffffff;
 $gray-dark-divider: #eeeeee;
 $color-separator: #dfdfdf;
+$purple: #34219e;
```
NeOMakinG, line 10: Here is the modification too, others are linting

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/msg_list_item.html.twig
```diff
@@ -23,13 +23,14 @@
  * International Registered Trademark & Property of PrestaShop SA
  *#}
 
-<li class="mb-2">
-  <div class="row">
+ <li class="mb-2{% if message.employeeId %} messages-block-employee{% else %} messages-block-customer{% endif %}">
```
Progi1984:
```suggestion
<li class="mb-2 {% if message.employeeId %}messages-block-employee{% else %}messages-block-customer{% endif %}">
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
```diff
@@ -23,17 +23,31 @@
...
     'action': path('admin_orders_update_status', {'orderId': orderForViewing.id}),
     'attr': {
-      'class': 'form-inline d-inline-block',
       'id': 'update_order_status_action_form'
     }
   }) }}
 
   <div class="input-group">
-    {{ form_widget(updateOrderStatusActionBarForm.new_order_status_id, {'attr': {'class': 'mr-1'}, 'id': 'update_order_status_action_input'}) }}
+    {% set backgroundColor = 'white' %}
```
Progi1984:
```suggestion
    {% set backgroundColor = '#ffffff' %}
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
```diff
@@ -160,7 +162,7 @@
         </div>
       {% endif %}
       {% for customization in product.customizations.textCustomizations %}
-        <p><strong>{{ customization.name }}</strong> {{ customization.value }}</p>
+        <p><strong>{{ customization.name }} :</strong> {{ customization.value }}</p>
```
Progi1984, line 165: :warning: In english, there is no space before `:`

LouiseBonnard: Franck is right: English grammar requires no space between the last letter of a word and the following punctuation sign. ;-)

src/Core/Form/ChoiceProvider/OrderStateByIdChoiceProvider.php
```diff
@@ -70,4 +81,22 @@ public function getChoices()
...
+     *
+     * @return array
+     */
+    public function getChoicesAttributes()
+    {
+        $orderStates = $this->orderStateDataProvider->getOrderStates($this->languageId);
+        $attrs = [];
+
+        foreach ($orderStates as $orderState) {
+            $attrs[$orderState['name']]['data-background-color'] = $orderState['color'];
```
PierreRambaud, line 96: Should be better to use an id instead of a name here :thinking:  (`$orderState['name']`)
i'm afraid about having the same name twice

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
PierreRambaud, line 165:
```suggestion
        <p><strong>{{ '%name%:'|trans({'%name%': customization.name}, 'DOMAINYOUWANTTOUSE) }} :</strong> {{ customization.value }}</p>
```

src/Core/Form/ChoiceProvider/OrderStateByIdChoiceProvider.php
NeOMakinG: @atomiix

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
```diff
@@ -160,7 +162,7 @@
         </div>
       {% endif %}
       {% for customization in product.customizations.textCustomizations %}
-        <p><strong>{{ customization.name }}</strong> {{ customization.value }}</p>
+      <p><strong>{{ '%name%:'|trans({'%name%': customization.name}, 'Admin.Ordercustomers.Feature') }} </strong> {{ customization.value }}</p>
```
NeOMakinG: @LouiseBonnard Need wording there, not sure of the domain

LouiseBonnard: @NeOMakinG, this one is fine. :-)

src/Core/Form/ChoiceProvider/OrderStateByIdChoiceProvider.php
atomiix: I'm afraid we can't. The way Symfony works with `select` (and `option`) is that it use the `value` of the array as the `name` property of the `option` element and the `key` of the array as the `option` text. [See here](https://symfony.com/doc/current/reference/forms/types/choice.html#choices)

PierreRambaud (dismissed):

admin-dev/themes/new-theme/js/pages/order/view.js
```diff
@@ -159,9 +163,19 @@ $(() => {
...
+      $wrapper.toggleClass('select-status-opened');
+    });
+
+    $(OrderViewPageMap.updateOrderStatusActionInput).on('change', event => {
+      const $element = $(event.currentTarget);
+      const $option = $('option:selected', $element);
+      const selectedOrderStatusId = $element.val();
+
+      $wrapper.css('background-color', $option.data('background-color'));
+      $wrapper.toggleClass('is-bright', $option.data('is-bright') !== undefined);
```
Progi1984, line 173:
```suggestion
      $wrapper
        .css('background-color', $option.data('background-color'))
        .toggleClass('is-bright', $option.data('is-bright') !== undefined);
```

Progi1984 (commented):
A last comment

Progi1984:
@NeOMakinG Please rebase your PR :)

admin-dev/themes/new-theme/js/pages/order/view.js
NeOMakinG: It's less than 100 character, eslint will inline it after mergin it on develop

Progi1984 (dismissed):

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
```diff
@@ -23,17 +23,30 @@
...
-    {{ form_widget(updateOrderStatusActionBarForm.new_order_status_id, {'attr': {'class': 'mr-1'}, 'id': 'update_order_status_action_input'}) }}
+    {% set backgroundColor = '#ffffff' %}
+    {% set isBright = true %}
+    {% for choice in updateOrderStatusActionBarForm.new_order_status_id.vars.choices %}
+      {% if choice.value == updateOrderStatusActionBarForm.new_order_status_id.vars.data %}
+        {% set backgroundColor = choice.attr['data-background-color'] %}
+        {% set isBright = choice.attr['data-is-bright'] %}
+      {% endif %}
+    {% endfor %}
+    <div class="select-status{% if isBright %} is-bright{% endif %} select-status-colored" id="update_order_status_action_input_wrapper" style="background-color: {{ backgroundColor }}">
```
jolelievre: I think you should deal with this in the form type directly, in the class `UpdateOrderStatusType`, you already have all the data available thanks to `$this->orderStatusChoiceAttributeProvider->getChoicesAttributes()` So you can preset the correct background-color directly in the form type

This avoids having too much logic in the twig (which is not the appropriate place), and is consistent with what weare trying to do in new form We manage as much as we can in the form type which allows to simply use `form_widget` on the form, this will help extend the forms by modules

That said you are using the color on a prevent div that was added here, is this necessary to have this extra div layer? Ideally we should rely on the regular form display as much as we can (for mentioned reasons)

jolelievre: From what I understand with this feature, it allows to use the status color as a background for the selector Isn't it possible to manage this only thanks to the basic select input?

admin-dev/themes/new-theme/js/pages/order/view.js
```diff
@@ -159,9 +163,19 @@ $(() => {
 
   function handleUpdateOrderStatusButton() {
     const $btn = $(OrderViewPageMap.updateOrderStatusActionBtn);
+    const $wrapper = $(OrderViewPageMap.updateOrderStatusActionInputWrapper);
```
jolelievre, line 165: As explained in another comment, is this wrapper div necessary It would be better if we could rely on default input rendering without wrapping/customizing it in the twig Or it will make this form harder to extend/customize for module developer

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
```diff
@@ -223,18 +225,18 @@ export default class OrderProductRenderer {
     }
   }
 
-  toggleColumnLocation(target, forceDisplay = null) {
-    let isColumnLocationDisplayed = false;
+  toggleColumn(target, forceDisplay = null) {
```
jolelievre, line 230: Nice reuse of the method, but maybe we should add a description in a doc to understand what it does and when it should be used

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
```diff
@@ -92,8 +92,9 @@ export default class OrderProductRenderer {
     $(OrderViewPageMap.productsPanel).detach().appendTo($modificationPosition);
     $modificationPosition.closest('.row').removeClass('d-none');
 
-    // Show column location
-    this.toggleColumnLocation(OrderViewPageMap.productsCellLocation, true);
+    // Show column location & refunded
+    this.toggleColumn(OrderViewPageMap.productsCellLocation);
```
jolelievre, line 96: Is it normal that the second parameter (used to force display) has disappeared?

src/PrestaShopBundle/Form/Admin/Sell/Order/UpdateOrderStatusType.php
```diff
@@ -38,12 +39,20 @@ class UpdateOrderStatusType extends AbstractType
...
+     * @var FormChoiceAttributeProviderInterface
+     */
+    private $orderStatusChoiceAttributeProvider;
+
     /**
      * @param FormChoiceProviderInterface $orderStatusChoiceProvider
      */
-    public function __construct(FormChoiceProviderInterface $orderStatusChoiceProvider)
-    {
+    public function __construct(
```
jolelievre, line 49: Nice use of the `FormChoiceAttributeProviderInterface`!!
However the form should be injected with the attributes as an array, to avoid coupling it  with the `FormChoiceProviderInterface` and `FormChoiceAttributeProviderInterface` directly Especially since in this case it is the same service that is injected..

This also allows to use the form in more various ways. The service must be used in the service definition directly. You can look at `ManufacturerAddressType` class and definition for an example

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/Modal/add_order_discount_modal.html.twig
```diff
@@ -66,7 +66,7 @@
 
                 <div class="input-group-append">
                   <div class="input-group-text" id="add_order_cart_rule_value_unit" data-currency-symbol="{{ orderCurrency.symbol }}">%</div>
-                  {{ form_widget(addOrderCartRuleForm.value, {'type':'number'}) }}
+                  {{ form_widget(addOrderCartRuleForm.value, { 'type':'number', 'attr': { 'step': 1, 'class': 'input-group-text' }}) }}
```
jolelievre: These arguments should be integrated in the form type rather than in the twig Especially for the type number

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
```diff
@@ -111,14 +111,16 @@
   </div>
 </div>
 
-<div class="mt-2" id="order_hook_tabs">
-  <ul class="nav nav nav-tabs" role="tablist">
-    {# Rendering of hook displayAdminOrderTabLink, we expect tab links #}
-    {{ renderhook('displayAdminOrderTabLink', {'id_order': orderForViewing.id}) }}
-  </ul>
+{% if hookcount('displayAdminOrderTabLink') > 0 %}
```
jolelievre: Although both hooks are supposed to be related and have the same count, maybe it's better to separate them with each one checking its own count This way if a module forgets to add the appropriate link it will at least display the content, which is less confusing for debugging

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
```diff
@@ -81,44 +94,31 @@
...
     {% if orderForViewing.isDelivered() %}
-      <div class="form-inline d-inline-block form-horizontal ml-2">
-        <div class="input-group">
-          <button class="btn btn-action return-product-display" type="button"{% if not orderForViewing.isRefundable() %} disabled{% endif %}>
-            <i class="material-icons">swap_horiz</i>
-            {{ 'Return products'|trans({}, 'Admin.Orderscustomers.Feature') }}
-          </button>
-        </div>
-      </div>
+      <button class="btn btn-action return-product-display" type="button"{% if not orderForViewing.isRefundable() %} disabled{% endif %}>
```
jolelievre, line 102: I believe I added this because there was a small offset compared with the buttons on the left of this block This was spotted by @Robin-Fischer-PS with the accurate use of a ruler!! Like a real life ruler!!
If you managed to get rid of all those extra divs and keep all the buttons aligned that's great!! Just be careful as it was a really small offset (1px)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -25,10 +25,14 @@
 
 {% set paginationNum = 8 %}
 {% set isColumnLocationDisplayed = false %}
+{% set isColumnRefundedDisplayed = false %}
```
jolelievre, line 28: I would be better if we could integrated this in the controller instead of the twig (same for `isColumnLocationDisplayed`)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
```diff
@@ -64,8 +64,10 @@
     </div>
   </td>
   <td class="cellProductLocation{% if not isColumnLocationDisplayed %} d-none{% endif %}">{{ product.location }}</td>
-  <td>
-    {{ product.quantityRefunded }}{% if product.amountRefunded %} ({{ product.amountRefunded }} {{ 'Refunded'|trans({}, 'Admin.Orderscustomers.Feature') }}){% endif %}
+  <td class="cellProductRefunded{% if not isColumnRefundedDisplayed %} d-none{% endif %}">
```
jolelievre, line 67: Be careful, you need to add this in the product add controller as well, we had a bug related to this when `isColumnLocationDisplayed` was integrated Because when the product is added via ajax, the controller returns the displayed row and uses this template directly, so the variable will be missing as is

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
jolelievre: That's weird this data is supposed to be translated in the database (in the `customization_field_lang`) I guess this can be useful for our fixtures data (and only them), but they'd rather be correctly inserted in the database rather than rely on the Translator
Well it will probably work anyway, but still a weird use

jolelievre (dismissed):
Most of my comments are meant for improvement, or discussion
But at least the one regarding `isColumnRefundedDisplayed` in product add cotroller is mandatory to avoid a bug

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
NeOMakinG: Ahahahah, it should be fine, it looks aligned :)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
NeOMakinG: @Progi1984 did a review on this because we needed to manage the `:` spacing between EN and other languages

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
NeOMakinG: we absolutly need this surrounding div as we are placing a pseudo-element and we can't customize the background of the select as it also set the option background if we do that

matthieu-rolland: @jolelievre  I'll have a look into it 👍 

Edit: oops @NeOMakinG 's answer arrived juste before mine. I checked it and I concluded the same thing, we can't really do that with the generic ChoiceType. We probably would have to create a dedicated type, that integrates the surrounding div.

@jolelievre what do you think?

jolelievre:
@NeOMakinG as I wrote in this comment https://github.com/PrestaShop/PrestaShop/pull/17734#issuecomment-597682952

The template `src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/extra_order_button_actions.html.twig` needs to be updated as well so that all buttons are correctly aligned, once the hooks PR is merged you will be able to rebase and use the module example to test this feature

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
jolelievre: Oooh I get it! Smart!

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/order_actions.html.twig
jolelievre: Yeah since indeed we need this div, it's an "enriched" choice type so it's worth creating a dedicated form field type I guess

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/product.html.twig
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
jolelievre, line 96:
```suggestion
    this.toggleColumn(OrderViewPageMap.productsCellLocation, true);
```
@NeOMakinG are you sure you don't need this parameter to force display here? Same on the line under

src/PrestaShopBundle/Form/Admin/Sell/Order/UpdateOrderStatusType.php
jolelievre: Done

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/Modal/add_order_discount_modal.html.twig
jolelievre: Done

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
jolelievre: Added an or condition

admin-dev/themes/new-theme/scss/pages/_orders_view.scss
```diff
@@ -1,4 +1,143 @@
...
+      box-shadow: 0 0 10px 0 rgba(0, 0, 0, 0.1);
+    }
+
+    &::after {
+      content: "keyboard_arrow_down";
+      position: absolute;
+      font-family: "Material Icons";
+      cursor: pointer;
+      font-size: 1.25rem;
+      color: white;
```
Progi1984:
```suggestion
      color: #ffffff;
```

admin-dev/themes/new-theme/scss/pages/_orders_view.scss
```diff
@@ -1,4 +1,143 @@
...
+      }
+    }
+  }
+
+  // Todo: Looks like we don't have any custom select without JS to add in a form
+  .select-status {
+    position: relative;
+
+    .select-status-colored {
+      background: white;
```
Progi1984:
```suggestion
      background: #ffffff;
```

admin-dev/themes/new-theme/scss/pages/_orders_view.scss
```diff
@@ -1,4 +1,143 @@
...
+      margin-top: 5px;
+    }
+  }
+
+  #product-pack-modal-table {
+    .cell-product-quantity {
+      .badge {
+        font-size: 100%;
+        min-width: 26px;
+        border-radius: 1000px !important;
```
Progi1984, line 103: 1000px isn't huge ?

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
NeOMakinG: @atomiix as you worked on this, what do you think ?

atomiix: @jolelievre @NeOMakinG Yes I'm sure, it should not be forced! :)

admin-dev/themes/new-theme/scss/pages/_orders_view.scss
NeOMakinG: No as I tested a lot of values, it's common to have rounded at right and left properly

Progi1984 (dismissed):

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/messages.html.twig
```diff
@@ -78,9 +78,9 @@
       {% endfor %}
     </div>
 
-    <div class="form-group row">
+    <div class="form-group row configure">
       <div class="col-sm">
-        <a href="{{ path('admin_order_messages_index') }}" class="mb-1">
+        <a href="{{ path('admin_order_messages_index') }}" class="configure-link">
           {{ 'Configure predefined messages'|trans({}, 'Admin.Orderscustomers.Feature') }}
```
LouiseBonnard, line 84: I have a slight doubt. @MatShir, did we change this wording into something shorter during the first reviews? It kind of rings a bell to me but I can't put my finger on it...
🤔

NeOMakinG: This PR is coming from 1.7.7.x, it means that this wording has been merged on 1.7.7.x, maybe the change has been forgot in the past or hasn't been done

LouiseBonnard: Okay, well, let's keep it as is!

jolelievre (dismissed):

matks:
@NeOMakinG I'm very sorry there is a conflict on `admin-dev/themes/new-theme/scss/pages/_orders.scss` this means QA cannot recompile assets 😢

Progi1984 (dismissed):

Progi1984 (dismissed):

matthieu-rolland (dismissed):

Progi1984:
@NeOMakinG Travis is :red_circle:

matthieu-rolland:
> @NeOMakinG Travis is 🔴

hm... it breaks at the same place on another PR of mine 🤔

Progi1984 (dismissed):

matthieu-rolland (dismissed):

Robin-Fischer-PS:
Hi @NeOMakinG !

3 issues on this PR, listed below, the rest is OK !

#17867 is :heavy_check_mark: 

#17656 : 
- The button to unfold the old baskets and control is reversed. : :heavy_check_mark: 
- The file name is not displayed after selecting the file. : :x: 
When file name is really long, it goes on the label under it. This is not blocking the merge.
pr17801 long file name

#17655 : 

- Empty rectangle below the tabs (status, transport...) :heavy_check_mark: 
- The stock location is displayed in cancel, partial, standard and return products while none of the products have a stock location. It should only display it if only one of them has a stock location : :heavy_check_mark: 
- The refunded column is always displayed by default even if the merchandise return is deactivated and if no refund has been made. :heavy_check_mark: 
- Border between % (or €) and the value field in the "Add a discount" popup is too thick. Adding the class input-group-text on the value field should solve it. :heavy_check_mark: 
- Check if we can dynamically change the color of the top left status dropdown according to the selected status. :x: The whole dropdown is colored, only the status should be.
Capture d’écran de 2020-03-23 11-56-54


Front-end issues :

- Put the label of the Order list on the Order page in the status history : :heavy_check_mark: 
- The label is not well rounded in a stock greater than two in the product list of the order page : :heavy_check_mark: 
- The status change does not correspond to the expected display : :x: (see "Check if we can dynamically change the color of the top left status dropdown according to the selected status")
- The button to edit the merchant's address page goes below the label when the screen size is reduced : :heavy_check_mark: 
- In a reduced screen, the button that goes to the line is too close to the drop-down menu to update the status (see on the mobile view) : :heavy_check_mark: 
- Pop-in of messages in a reduced screen makes the profiles icons all line up on the left while some should stay on the right : :heavy_check_mark: 
- The block of messages in a reduced screen puts the icons below the message when it should stay above the message : :heavy_check_mark: 
- Change the number of messages displayed to 4 instead of 2 : :heavy_check_mark: 
- Adding a discount.
The discount value when you want to increase the value of the field using the top arrow increases the value by 0.01. Can it be changed to 1 or maybe 10? :heavy_check_mark: 
- The trash icon in the discount list is in blue, whereas it should only be displayed in blue when the mouse passes over it : :heavy_check_mark: 
- Review the display (alignment of the input with the label) of the amount to be entered when the refund, partial refund and standard refund buttons interact : :heavy_check_mark: 
I don't know what was expected, but it looks fine to me.
pr17801 refund display

- The numbers labeled in black in the customer block have a size that looks smaller than normal : :heavy_check_mark: 
- The numbers labeled in black in the pop-in of the product packs are smaller than normal :heavy_check_mark: 
- The link to see the packs is too close to the product name :heavy_check_mark: 
- Added ":" after custom text labels in the product list when displaying a customized product : :heavy_check_mark: 
- Review the partial refund shipping field : :heavy_check_mark: 
- All the fields in add and edit are not aligned : :x: The Invoice, Cancel and Add/Edit buttons are not aligned
pr17801 add alignement
pr17801 edit alignement

- The "configure message" link is too much of a dropdown in the message block : :heavy_check_mark: 
- See if there are any changes or comments to be made in the mobile view : :heavy_check_mark: 
- Mobile: center the content of the main buttons (top of the page buttons) : :heavy_check_mark:

matks (dismissed):

matthieu-rolland (dismissed):

Robin-Fischer-PS:
Hi @NeOMakinG !

I have one issue left : On Edit order, the available and Total column are not aligned with the numbers.

pr17801 horizontal align edit

jolelievre (dismissed):

matthieu-rolland (dismissed):

Robin-Fischer-PS:
@NeOMakinG I still have the alignment issue on the edit page :/

NeOMakinG:
Fixed another visual bug on add product alignment @Robin-Fischer-PS 

It was the Available and Total case on the added product which were not aligned with the rest

Progi1984 (dismissed):

matks (dismissed):

Robin-Fischer-PS:
It's OK for me ! (But I'm not a UX expert ^^ )

Thanks @NeOMakinG

matks:
> It's OK for me ! (But I'm not a UX expert ^^ )

Trust yourself !

Progi1984 (approved):

matthieu-rolland (approved):

matks (approved):

Progi1984:
Thanks @NeOMakinG

PR #17797, opened 2020-02-24
Keep BO order page from displaying an exception when accessing a non existing order via URL

matthieu-rolland:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | When accessing a non existing order via URL (order is doesn't exist), we display an exception, we should redirect to order list page instead. 
| Type?         | bug fix
| Category?     | BO
| BC breaks?    |no
| Deprecations? |no
| Fixed ticket? | Fixes #17732 
| How to test?  | This url: `/admin-dev/sell/orders/orders/999999/view` should redirect to the list order page (given there is no order with id **999999**.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17797)

PR #17839, opened 2020-02-26
Wrap OrderController::viewAction in try-catch

### LINKED ISSUE #17732: Exception when accessing a non existing order via URL
**Describe the bug**
When a user try to access a non existing order (with a bad id) via URL in migrated page, there is an exception. On Legacy, there is a redirection on the order list.

**To Reproduce**
1. Go to a non existing order via URL, for example : /admin-dev/index.php/sell/orders/orders/647/view
2. An exception is displayed

**Screenshots**
non existing order

**Additional information**
PrestaShop version: 1.7.7
PHP version: N/A

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | viewAction of Sell/Orders/OrderController didn't catch any exception, so in case of error the page would break. For example passing non existing order id would result in following error: Screenshot from 2020-02-26 13-22-35. After the fix it will redirect to index page with error like: Screenshot from 2020-02-26 13-23-30
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17732
| How to test?  | follow `admin-dev/sell/orders/orders` to see migrated order list. Select any order for view. Provoke the page to throw unexpected error (like write not existing order id in url). You should be redirected to index page with error message like mentioned in description. (error message might vary, depending if its defined in controller or not)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17839)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
@@ -365,113 +365,120 @@ class OrderController extends FrameworkBundleAdminController
         try {
             /** @var OrderForViewing $orderForViewing */
             $orderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId));
-            $addOrderCartRuleForm = $this->createForm(AddOrderCartRuleType::class, [], [
-                'order_id' => $orderId,
-            ]);
-            $updateOrderStatusForm = $this->createForm(UpdateOrderStatusType::class, [
-                'new_order_status_id' => $orderForViewing->getHistory()->getCurrentOrderStatusId(),
-            ]);
-            $updateOrderStatusActionBarForm = $this->createForm(UpdateOrderStatusType::class, [
-                'new_order_status_id' => $orderForViewing->getHistory()->getCurrentOrderStatusId(),
-            ]);
-            $addOrderPaymentForm = $this->createForm(OrderPaymentType::class, [
-                'id_currency' => $orderForViewing->getCurrencyId(),
-            ], [
-                'id_order' => $orderId,
-            ]);
+        } catch (OrderException $e) {
+            $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));
 
-            $orderMessageForm = $this->createForm(OrderMessageType::class, [], [
-                'action' => $this->generateUrl('admin_orders_send_message', ['orderId' => $orderId]),
-            ]);
-            $orderMessageForm->handleRequest($request);
+            return $this->redirectToRoute('admin_orders_index');
+        }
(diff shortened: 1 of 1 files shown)


matks (dismissed):

matthieu-rolland:
@matks , @zuk3975 there is already a PR in QA for the case when order is is not found: 

https://github.com/PrestaShop/PrestaShop/pull/17797

It was decided that we should redirect to order list when the order id doesn't exist.

matthieu-rolland:
As discussed with @zuk3975 I'll close the other PR and we'll use this one, that also deals with other kind of exceptions 👍

matthieu-rolland (dismissed):

jolelievre (changes requested):
Sorry for being annoying, but as we discussed:
1. The try/catch scope is way too wide, we should only encapsulate the code we are expecting to throw an exception, and each piece of code should have its dedicated try/catch block:
- the query handler part
- the order message handle request part
- the dispatch hook part

I don't think the form creation are supposed to throw an exception, but maybe there are hooks during form creation as well? In which case they should be protected as well

2. The exception caught is too wide, we can't catch a basic `Exception` because it would hide many unexpected errors that need to be displayed

3. Nice to have (but not required), when using dev mode maybe we should let the exception be thrown instead of displaying it?

zuk3975:
> Sorry for being annoying, but as we discussed:
> 
> 1. The try/catch scope is way too wide, we should only encapsulate the code we are expecting to throw an exception, and each piece of code should have its dedicated try/catch block:
> 
> * the query handler part
> * the order message handle request part
> * the dispatch hook part
> 
> I don't think the form creation are supposed to throw an exception, but maybe there are hooks during form creation as well? In which case they should be protected as well
> 
> 1. The exception caught is too wide, we can't catch a basic `Exception` because it would hide many unexpected errors that need to be displayed
> 2. Nice to have (but not required), when using dev mode maybe we should let the exception be thrown instead of displaying it?

Ok im gonna split multiple exceptions.
Still going to catch the generic Exception **where hooks are called** because we dont want to break the page if module is throwing some random exception as @matks said.
Where no hooks are called, ill catch domain exceptions. Thats how it should be I suppose

jolelievre (approved):
Thanks @zuk3975

zuk3975:
> Thanks @zuk3975

thanks for attentive reviews :P

atomiix (approved):

PierreRambaud (approved):

matks (approved):

SD1982:
This PR correctly fix #17732 . Thanks @zuk3975

Progi1984:
Thanks @zuk3975

Issue #17138, opened 2020-01-13
[SF Migration] Build an example module for 1.7.7


matks:
## Build an example module for 1.7.7

This modules must demonstrate:
- how to use the new hooks in "Edit an order" page
- how to insert a "upload image" input inside the Symfony form managing "Edit a Supplier" page

This module must promote the best practices.

**Update:**

We built 2 modules. See https://github.com/PrestaShop/PrestaShop/issues/17138#issuecomment-599570114

jf-viguier:
yes please

rokaszygmantas:
I'm on it :slightly_smiling_face:

matks:
@eternoendless I need the "all example modules" repository we discussed about, to host all of our example modules 😄

matks:
Linked to https://github.com/PrestaShop/PrestaShop/issues/17275

tdavidsonas88:
> @eternoendless I need the "all example modules" repository we discussed about, to host all of our example modules

Yup, that would be helpful for SEO purposes and backlinking when writing documentation

jf-viguier:
@matks any news for uploading stuff ? We need this

matks:
> @matks any news for uploading stuff ? We need this

WIP, but dont worry it will be delivered, at max, the same day we deliver 1.7.7.0 😊

matks:
@olecorre @jf-viguier I remember you asked for this "upload image" example 😊 

Could you tell us more specifically what you had in mind ?
To me, a "upload image" input
- allows preview of already registered image (and you can replace it)
- maybe resize the image ?

So "image upload" inputs of Product Categories page are eligible, then ?

jf-viguier:
I don't need the resize stuff but uploading, deleting, replacing and preview.
Thanks @matks

matks:
@jf-viguier @olecorre 2 example modules for 1.7.7.x have been delivered in repository https://github.com/PrestaShop/example-modules
- Order Page hooks module 
- Extend a Symfony Form module 2

Documentation has been written for Order Page hooks module => https://devdocs.prestashop.com/1.7/modules/sample_modules/order_view_page_new_hooks/
Documentation for Extend a Symfony Form module 2 is WIP

Feel free to try (must be installed on `1.7.7.x` shop) and tell us your feedback 😊

jf-viguier:
Thanks

PR #17842, opened 2020-02-26
File type with image

tdavidsonas88:
| Questions     | Answers
| Branch?       | develop
| Description?  | ImageTypeExtension to render file with image. Also image Delete Button. Needed for module developers and maybe could be used to simplify the forms.
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/17138
| How to test?  | Can be tested together with the install demo module: https://github.com/PrestaShop/example-modules/pull/2 in the Supplier Edit form

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17842)


matks (dismissed):

matks:
@tdavidsonas88 Simple, efficient : good !

tdavidsonas88:
> @tdavidsonas88 Simple, efficient : good !

Thank you, also users asked for additional functionality like delete image (that is why it is still in WIP) : https://github.com/PrestaShop/PrestaShop/issues/17138#issuecomment-590924978

tdavidsonas88:
@matks Thinking about adding the Delete button for image also. Then one more option needs to be added to the ImageTypeExtension like `is_delete_button_enabled`

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
```diff
@@ -730,6 +730,26 @@
...
     </label>
   </div>
+  {# is defined in ImageTypeExtension #}
+  {% if image_url is defined and image_url is not null %}
+    <div class="form-group row mt-3">
+      <div class="col-sm">
+        <figure class="figure">
+          <img src="{{ image_url }}" class="figure-img img-fluid img-thumbnail">
+        </figure>
+        {% if delete_action_route is defined and id is defined %}
```
jolelievre, line 740: `id_property` is never used, maybe it's supposed to be used here?

PierreRambaud: The **id_property** is used here https://github.com/PrestaShop/PrestaShop/pull/17842/files#diff-e560981ab87968f4ec8283bef85cef23R55

tdavidsonas88: it is used in the ImageTypeExtension. The solution came from https://symfony.com/doc/3.4/form/create_form_type_extension.html The only difference that used Hidden fields instead of Entity methods to get the values.

Progi1984 (dismissed):

PierreRambaud (dismissed):

tdavidsonas88:
Thought about some possible changes to do it a bit better when will start to work on 1.7.8 version. Could use for generating the link and remove it's generation from twig.
```
         /** @var \Symfony\Bundle\FrameworkBundle\Routing\Router $router */
        $router = $this->get('router');
```

And also received suggestions to implement `delete` button as a separate block.

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
```diff
@@ -730,6 +730,26 @@
...
+        <div class="d-block">
```
NeOMakinG, line 741: What ? Does this div has something else than a display block CSS sided? This class looks unnecessary

tdavidsonas88: It looks, but I was not able to make it other way this time. If you have a better suggestion, would be nice to hear it

tdavidsonas88: without it the button is aligned to the right of the other elements. With it, it is aligned below other elements of the block

tdavidsonas88: @NeOMakinG also this upload block has a strange behavior on it's left on the whole PrestaShop - all edit form using uploads. If I click it, it popups upload file dialog, even if I click not on the browse button. Could you check it out as a separate issue?

NeOMakinG: > @NeOMakinG also this upload block has a strange behavior on it's left on the whole PrestaShop - all edit form using uploads. If I click it, it popups upload file dialog, even if I click not on the browse button. Could you check it out as a separate issue?

That behavior looks fine, it's same on basical html upload input : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file

Or I missed something of you explaination

NeOMakinG: Do you have a simple way of testing your block so I can check the behavior of your div with that class ?

Did you test with a div without any classes?

tdavidsonas88: Without the div: 
image

With the div:
image

Issue with the click outside is that clicking at the form label area invokes upload popup. And if I extend the Upload block, the popup dialog opens even several centimeters from the upload input below the : Upload image file label.

tdavidsonas88: To easily test it you can try installing this module: https://github.com/PrestaShop/example-modules/pull/2 and go to Suppliers edit form

NeOMakinG: Isn't that because the image file label has a margin maybe or something extending his height?

NeOMakinG: The div is fine, but the class looks not, a div is naturally in display: block; that's why I commented : 
image

PierreRambaud:
Hi @tdavidsonas88 ,

For your information, the original issue has been closed.

Thanks!

PR #17846, opened 2020-02-26
Fix exception thrown when editing a customer with a bad email address

### LINKED ISSUE #17809: Bad validation of customer emails in BO
**Describe the bug**

When we have wrong object value in back-office, in some controllers, we can't tell exactly what's wrong, FormBuilder doesn't give us anything even tho we have `try` `catch`

>Edit by Khouloud

**Steps to reproduce the issue:**
1. Go to the FO => Create a new account with an email: `test@test.com.777` => OK
2. Go to the BO => Customers page => Edit the last customer-created
3. See error

It is ok with PS1.7.5.2.
It is a regression.

With PS1.7.6.3 and PS1.7.6.0, when we try to create a customer with the email `test@test.com.777`
=> we have an error
image

With PS1.7.5.2 => it is successfully added.

----------------------------------------------------------------
**To Reproduce**
1. go to ps_customers
2. change e-mail address for some customer to for example some.email@gemail.com.77
3. go to back-office
4. try to edit this customer

check screenshot

**Screenshots**

image

We should have exact information what's wrong here, not a notice because of failure of FormBuilder.

**Additional information**
PrestaShop version: N/A
PHP version: N/A

atomiix:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | The email address format verification was too strict on the BO so a valid email address saved in the FO could throw an exception on the BO when editing the customer. This PR fixes it.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17809
| How to test?  | 1. Go to the FO => Create a new account with an email: test@test.com.777<br>2. Go to the BO => Customers page => Edit the last customer created.<br>3. No error should be thrown

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17846)

### DIFF
--- a/src/Core/Domain/Customer/ValueObject/Email.php
+++ b/src/Core/Domain/Customer/ValueObject/Email.php
@@ -26,106 +26,14 @@
 
 namespace PrestaShop\PrestaShop\Core\Domain\Customer\ValueObject;
 
-use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\CustomerConstraintException;
-
 /**
- * Stores customer's email address
+ * @deprecated since version 1.7.6.4 use PrestaShop\PrestaShop\Core\Domain\ValueObject\Email instead
  */
 class Email
 {
-    /**
-     * @var int Maximum allowed length for customer email
-     */
-    const MAX_LENGTH = 255;
-
-    /**
-     * @var string
-     */
-    private $email;
-
-    /**
-     * @param string $email
-     */
     public function __construct($email)
     {
-        $this->assertEmailIsString($email);
-        $this->assertEmailDoesNotExceedAllowedLength($email);
-        $this->assertEmailIsValid($email);
-
-        $this->email = $email;
-    }
-
-    /**
-     * @return string
-     */
-    public function getValue()
-    {
-        return $this->email;
-    }
-
-    /**
-     * Check if given email is the same as current
-     *
-     * @param Email $email
-     *
-     * @return bool
-     */
-    public function isEqualTo(Email $email)
-    {
-        return $email->getValue() === $this->getValue();
-    }
-
-    /**
-     * Assert that email is in valid format
-     *
-     * @param string $email
-     *
-     * @throws CustomerConstraintException
-     */
-    private function assertEmailIsValid($email)
-    {
-        if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
-            throw new CustomerConstraintException(
-                sprintf('Customer email %s is invalid.', var_export($email, true)),
-                CustomerConstraintException::INVALID_EMAIL
-            );
-        }
-    }
-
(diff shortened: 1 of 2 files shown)


NeOMakinG:
Can you rename the title please ?

src/Core/Domain/Customer/ValueObject/Email.php
```diff
@@ -26,106 +26,14 @@
 
 namespace PrestaShop\PrestaShop\Core\Domain\Customer\ValueObject;
 
-use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\CustomerConstraintException;
-
 /**
- * Stores customer's email address
+ * @deprecated since version 1.7.6.4 use PrestaShop\PrestaShop\Core\Domain\ValueObject\Email instead
```
matks, line 30: @atomiix Sure about that ?

src/Core/Domain/ValueObject/Email.php
```diff
@@ -78,19 +78,16 @@ public function isEqualTo(Email $email)
...
      *
-     * @param string $email
+     * @param $email
      *
      * @throws DomainConstraintException
      */
-    private function assertEmailIsValid($email)
+    public function assertEmailIsNotEmpty($email)
     {
-        if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
```
matks, line 89: @atomiix Can you tell us why `filter_var($email, FILTER_VALIDATE_EMAIL)` is not right 😮 ?! it's both surprising and frightening somehow

src/Core/Domain/Customer/ValueObject/Email.php
atomiix: @matks yep, this class is not used. Also, this PR is a cherry-picking of another PR on the 1.7.7.x branch that makes email validation coherent.

src/Core/Domain/ValueObject/Email.php
atomiix: `FILTER_VALIDATE_EMAIL` is not rfc 6531 compliant. It doesn't allow email address like `prestashop@localhost` for instance. Also, as the real email validation is done by the `Validation::isEmail()` method, we should not be validating the email in another place but check for coherence in the ValueObject.

matks: Can we add this explanation into the code as a comment ? I am sure other developers will have the same confusion that I had, and in order to prevent them from thinking "omg prestashop is bad 🙄 " it would be nice to provide them the full context

matks (approved):

PierreRambaud (approved):

matks:
Thank you @atomiix for PR and @Robin-Fischer-PS for QA

PR #17850, opened 2020-02-26
Rework OrderRefundCalculator to use Number

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fixes todo's in OrderRefundCalculator.php.`@todo: Check if it matches order_discount_price in legacy`, `@todo: use https://github.com/PrestaShop/decimal for price computations`
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Order refunds should work as expected in BO Order view page. Behavior tests covers most use-cases, so there shouldn't be any new bugs introduced by this pr.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17850)

### DIFF
--- a/src/Adapter/Order/Refund/OrderRefundCalculator.php
+++ b/src/Adapter/Order/Refund/OrderRefundCalculator.php
@@ -89,7 +89,6 @@ class OrderRefundCalculator
         $voucherChosen = false;
         $voucherAmount = 0;
         if ($voucherRefundType === VoucherRefundType::PRODUCT_PRICES_EXCLUDING_VOUCHER_REFUND) {
-            //@todo: Check if it matches order_discount_price in legacy
             $refundedAmount -= $voucherAmount = (float) $order->total_discounts;
         } elseif ($voucherRefundType === VoucherRefundType::SPECIFIC_AMOUNT_REFUND) {
             $voucherChosen = true;
--- a/src/Adapter/Order/Refund/OrderRefundCalculator.php
+++ b/src/Adapter/Order/Refund/OrderRefundCalculator.php
@@ -82,13 +82,15 @@ class OrderRefundCalculator
             $precision
         );
 
-        $refundedAmount = new Number('0');
+        $numberZero = new Number('0');
+
+        $refundedAmount = $numberZero;
         foreach ($productRefunds as $orderDetailId => $productRefund) {
             $refundedAmount = $refundedAmount->plus(new Number((string) $productRefund['amount']));
         }
 
         $voucherChosen = false;
-        $voucherAmount = new Number('0');
+        $voucherAmount = $numberZero;
         if ($voucherRefundType === VoucherRefundType::PRODUCT_PRICES_EXCLUDING_VOUCHER_REFUND) {
             $voucherAmount = new Number((string) $order->total_discounts);
             $refundedAmount = $refundedAmount->minus($voucherAmount);
@@ -97,7 +99,7 @@ class OrderRefundCalculator
             $refundedAmount = $voucherAmount = $chosenVoucherAmount;
         }
 
-        $shippingCostAmount = new Number((string) ($shippingRefund ?? 0));
+        $shippingCostAmount = $shippingRefund ?? $numberZero;
         if ($shippingCostAmount->isPositive()) {
             $shippingMaxRefund = new Number(
                 $isTaxIncluded ?
@@ -123,7 +125,7 @@ class OrderRefundCalculator
         }
 
         // Something has to be refunded (check refunds count instead of the sum in case a voucher is implied)
-        if (count($productRefunds) <= 0 && $refundedAmount->isLowerOrEqualThan(new Number('0'))) {
+        if (count($productRefunds) <= 0 && $refundedAmount->isLowerOrEqualThan($numberZero)) {
             throw new InvalidCancelProductException(InvalidCancelProductException::NO_REFUNDS);
         }
 
--- a/src/Adapter/Order/Refund/OrderRefundCalculator.php
+++ b/src/Adapter/Order/Refund/OrderRefundCalculator.php
@@ -40,6 +40,7 @@ use PrestaShop\PrestaShop\Core\Domain\Order\Exception\InvalidCancelProductExcept
 use PrestaShop\PrestaShop\Core\Domain\Order\ValueObject\OrderDetailRefund;
 use PrestaShop\PrestaShop\Core\Domain\Order\VoucherRefundType;
 use PrestaShop\PrestaShop\Core\Localization\CLDR\ComputingPrecision;
+use PrestaShopBundle\Form\Admin\Type\CommonAbstractType;
 use PrestaShopDatabaseException;
 use PrestaShopException;
 use TaxCalculator;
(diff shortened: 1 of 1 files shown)


src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -82,48 +84,56 @@ public function computeOrderRefund(
...
             throw new InvalidCancelProductException(InvalidCancelProductException::NO_REFUNDS);
         }
 
         return new OrderRefundSummary(
             $orderDetailList,
             $productRefunds,
-            $refundedAmount,
-            $shippingCostAmount,
-            $voucherAmount,
+            (float) $refundedAmount->toPrecision(CommonAbstractType::PRESTASHOP_DECIMALS),
```
jolelievre: I don't think you should use `CommonAbstractType` since this class has nothing related with form types

You should use `ComputingPrecision` instead to compute the precision related to this order's currency
However since the use of `Number` is located in this class it might be risky to perform any rounding at this point, we might even lose some precision here that we had with float (that's the opposite of what we are aiming for using `Number`)

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -82,48 +84,56 @@ public function computeOrderRefund(
...
+                $refundedAmount = $refundedAmount->plus($taxes);
             } else {
-                $refundedAmount += $shippingCostAmount;
+                $refundedAmount = $refundedAmount->plus($shippingCostAmount);
             }
         }
 
         // Something has to be refunded (check refunds count instead of the sum in case a voucher is implied)
-        if (count($productRefunds) <= 0 && $shippingCostAmount <= 0) {
+        if (count($productRefunds) <= 0 && $refundedAmount->isLowerOrEqualThan(new Number('0'))) {
```
jolelievre: I believe @atomiix also needed to perform these kind of checks, it's a pity that we need to instanciate a `new Number('0')` to perform this check Maybe we could add a new method to the `Number` class but it requires updating our `prestashop/decimal` library
@eternoendless do you think it could be possible?

jolelievre (commented):
I get why you wanted to change only this class However I'm afraid this risks to degrade the precision of the results I'm afraid using `Number` only has meaning if it is used in the whole computing chain and rounded at the end of the process

src/Adapter/Order/Refund/OrderRefundCalculator.php
zuk3975: so should i do something like (float) (string) $refundedAmount ? I dont see how to easily convert it back to float.
Or you suggest to try wider the scope and implement Number types in this DTO constructor ?

src/Adapter/Order/Refund/OrderRefundCalculator.php
zuk3975: then maybe we could also implement toFloat() ? because its still needed for some legacy class computations. And now it seems possible to do (float) (string) $number which looks strange

atomiix: @jolelievre @zuk3975 I agree, doing `new Number('0')` for each check is a pity... As we'll be using more this library I think it would be useful to have methods to make that easier.

In the other hand, the need for a `toFloat()` method will be less useful for the same reasons: if we use `Number` everywhere we will (theoretically 😄) not need to cast to float anymore.

zuk3975:
@jolelievre removed the rounding and usage of CommonAbstractType. Implemented (float) (string) instead :man_astronaut:

Progi1984:
Hello @zuk3975, thanks for your contribution, would you mind to rebase your pull request? Thanks

jolelievre (dismissed):

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: Well in the long term when everything uses `Number` you might be right, but it would still be more handy to use than this ugly `(float) (string)` ... And since the purpose of this class is to deal with decimal numbers (aka float) I don't think it's shocking to allow an output as a float (but it's just my opinion)
However having a static method to build from a float would be even more handy!! Along with an helper to compare to zero which is a very specific and used value

I'd say we could add a const `Number::ZERO` which would contains the 0 value in a number would be ideal, but unless I'm mistaken, we can't use objects as const So maybe a static function? with a static (🙀) var to avoid instanciating the object a thousand time.

eternoendless: @jolelievre Create an issue in PrestaShop/decimal and we can discuss there.

jolelievre:
@zuk3975 some conflicts appeared, sorry

src/Adapter/Order/Refund/OrderRefundCalculator.php
jolelievre: Hop, here's the issue https://github.com/PrestaShop/decimal/issues/6

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -81,49 +81,58 @@ public function computeOrderRefund(
...
+            $refundedAmount = $refundedAmount->minus($voucherAmount);
         } elseif ($voucherRefundType === VoucherRefundType::SPECIFIC_AMOUNT_REFUND) {
             $voucherChosen = true;
             $refundedAmount = $voucherAmount = $chosenVoucherAmount;
         }
 
-        $shippingCostAmount = (float) (string) $shippingRefund;
-        if ($shippingCostAmount > 0) {
-            $shippingMaxRefund = $isTaxIncluded ? $order->total_shipping_tax_incl : $order->total_shipping_tax_excl;
+        $shippingCostAmount = new Number((string) ($shippingRefund ?? 0));
```
jolelievre:
```suggestion
        $shippingCostAmount = $shippingRefund ?? new Number('0');
```

Since `$shippingRefund` is already a Number (and Number is immutable)

src/Adapter/Order/Refund/OrderRefundCalculator.php
```diff
@@ -81,49 +81,58 @@ public function computeOrderRefund(
...
             throw new InvalidCancelProductException(InvalidCancelProductException::NO_REFUNDS);
         }
 
         return new OrderRefundSummary(
             $orderDetailList,
             $productRefunds,
-            $refundedAmount,
-            $shippingCostAmount,
-            $voucherAmount,
+            (float) (string) $refundedAmount,
```
jolelievre, line 135: Not in this PR but maybe those should be `Number` as well (in a further refacto)

zuk3975: yeah I guess, this one might have bigger scope, lets leave it for another time

src/Adapter/Order/Refund/OrderRefundCalculator.php
zuk3975: :+1:

jolelievre (approved):

Progi1984 (approved):

Progi1984 (approved):

PierreRambaud (approved):

matks (approved):

Robin-Fischer-PS:
Thanks @zuk3975 , it's QA :heavy_check_mark: !

matks:
Thank you @zuk3975 🚀

PR #17852, opened 2020-02-26
Fix exception when adding negative amount/percentage discount in order view

### LINKED ISSUE #17747: Exception when adding a new discount with invalid amount - invalid discount

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | An exception was thrown when adding a negative amount/percentage in an order view. This PR fixes it by showing an error message.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17747
| How to test?  | In the migrated order view page, click `Add a discount`. Set discount type to "Amount" and put a negative number in the "value" field. Click "add" -> You should have an error message.<br><br>Same thing with discount type "Percent"

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17852)


src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -63,7 +64,7 @@ public function handle(AddCartRuleToOrderCommand $command): void
         switch ($command->getCartRuleType()) {
             // Percent type
             case OrderDiscountType::DISCOUNT_PERCENT:
-                if ($discountValue < 100) {
+                if ($discountValue < 100 && $discountValue >= 0) {
```
arouiadib: Thank you for this PR, I wasn't aware of this file while dealing with same issue :) .
I am a bit confused about `0` and `100` values.
From `order page` legacy: 
* possible to create percent discount with value 0%
* impossible to create percent discount with value 100%

From `catalog add cart rule` legacy:
* imPossible to create percent discount with value 0%.
* possible to create percent discount with value  100%
If you create discount from  `order page` legacy with 99% value, it is possible to edit it to 100% in  `catalog edit cart rule`. 

So what are possible values for percentage reduction?

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -111,12 +112,21 @@ public function handle(AddCartRuleToOrderCommand $command): void
...
+                    throw new NegativeCartRuleValueException(
+                        'Reduction percentage must be between 0% and 100%.',
+                        NegativeCartRuleValueException::NEGATIVE_PERCENTAGE
+                    );
                 }
 
                 break;
             // Amount type
             case OrderDiscountType::DISCOUNT_AMOUNT:
+                if (0 > $discountValue) {
```
arouiadib: Same here.
From `order page` legacy, it is possible to create anount discount with value 0, but when visiting this cart rule later in `catalog edit cart rule` legacy, it has no amount reduction at all. In fact you can't create amount discount with value 0 in `catalog>discounts> edit cart rule` legacy. Wdyt? Is there any change in specs?

arouiadib (dismissed):

matthieu-rolland (dismissed):

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
MatShir: Thanks @arouiadib for the discovery. The logic should respect the catalog and cart rule behaviors such as :
- impossible to create a percent discount with value of 0%.
- possible to create a percent discount with value of 100%

@atomiix Could you fixed in this PR ?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1423,6 +1424,16 @@ private function getErrorMessages(Exception $e)
                 $e->getMessage(),
                 'Admin.Orderscustomers.Notification'
             ),
+            NegativeCartRuleValueException::class => [
+                NegativeCartRuleValueException::NEGATIVE_AMOUNT => $this->trans(
+                    'Reduction amount cannot be lower than zero.',
+                    'Admin.Orderscustomers.Notification'
+                ),
+                NegativeCartRuleValueException::NEGATIVE_PERCENTAGE => $this->trans(
```
jolelievre: It's not only for negative value since you also check that it is inferior to 100 So you should rename this const to `NegativeCartRuleValueException::INVALID_PERCENTAGE` And for consistency the exception name should also be renamed to `InvalidCartRuleValueException`

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
jolelievre: Indeed this need to be confirmed and specified with the product team. Besides would it be possible to change the detect and detect the exception cases rather than managing this in an `else` statement.
Something like:
```php
if ($discountValue > 100 || $discountValue <= 0) {
    throw new NegativeCartRuleValueException(
        'Reduction percentage must be between 0% and 100%.',
        NegativeCartRuleValueException::NEGATIVE_PERCENTAGE
    );
}
// Do the action here
```

This would reduce the indentation which makes the code more readable And the condition for error is more straight forward.

Oh and I can't comment the appropriate line, but I think `$discountValue = (float) str_replace(',', '.', $command->getDiscountValue());` is useless since the command contains a `Number` we need to get rid of these nasty/useless `str_replace`

jolelievre (changes requested):
A few changes suggested And the authorized values need to be specified with the product team because @arouiadib suggestions make sense

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
atomiix: @arouiadib @jolelievre @MatShir I'm taking everything into account, I'm just waiting for new exception messages before pushing new changes ;)

Progi1984 (dismissed):

matks (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1454,6 +1455,10 @@ private function getErrorMessages(Exception $e)
                 'Only numbers and decimal points (".") are allowed in the amount fields, e.g. 10.50 or 1050.',
                 'Admin.Orderscustomers.Notification'
             ),
+            InvalidCartRuleValueException::class => $this->trans(
+                'It looks like the value is invalid: <ol><li>Percent or amount value must be greater than 0.</li><li>Percent value cannot exceed 100.</li><li>Amount value cannot exceed the total price of this order.</li></ol>',
```
matks: I validate the PR but I'm not a fan of embedding html inside translation chains 🤔

jolelievre: I know the PR has already been validated, but wouldn't it be better to manage dedicated exception codes for each use case, this would allow to have a precise message for each context
And it removes the html from wording which is really not a good practice

jolelievre (commented):
@atomiix suggestion for improvement

atomiix:
@jolelievre @matks I agree that html in wordings are not ideal (and as I should rebase this PR anyway, I'll try to do it differently) but those wordings have been choose after a discussion with @LouiseBonnard. WDYT?

LouiseBonnard:
@atomiix, @jolelievre, @matks, actually Crowdin catalogs already contain a few strings with HTML tags so we can use HTML if there is no other possible solution. I guess it is better to replace it with placeholders whenever it is possible but it does not work for formatting issue, does it?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: At least if we don't manage each use case with an exception we can break this wording into three to avoid having html in the wording:

```php
sprintf('%s<ol><li>%s</li><li>%s</li></ol>',
                $this->trans(
                    'It looks like the value is invalid:',
                ),
                $this->trans(
                    'Percent or amount value must be greater than 0.',
                ),
                $this->trans(
                    'Amount value cannot exceed the total price of this order.',
                )
            )
```

atomiix: @jolelievre done 🙂

jolelievre (dismissed):

matks (dismissed):

SD1982:
Hi @atomiix is it still possible to reproduce the error : 

capture d'écran_12

Step to reproduce : 
1. Add a discount with the amount of the total order ( shipping included )
2. Add a free shipping discount

**Screen-record**
https://drive.google.com/file/d/1YLqI1rXBdKwuQ00WSGafbtUNyzgiuC1l/view?usp=sharing

atomiix:
@PrestaShop/prestashop-product-team the exception that @SD1982 found was already present in 1.7.6.x. 
What should be the behavior for this specific case?

colinegin:
@atomiix @SD1982 same as when you try to add a discount which is higher than the order value : you should get this error message on top of the order page : "The discount value is greater than the order total."

atomiix:
@colinegin ok, so an error message! Thing is, we regrouped error messages into only one as follow:
```
It looks like the value is invalid:
- Percent or amount value must be greater than 0.
- Percent value cannot exceed 100.
- Amount value cannot exceed the total price of this order.
```
Should we reuse this list, change this list, add a case to this list or add a  separate error message?

colinegin:
After discussing with @atomiix , we think that the error message "- Amount value cannot exceed the total price of this order." should be replaced by "- Discount value cannot exceed the total price of this order." to be more generic and take into account situations where it's not an amount but a free shipping for instance. Are you ok with this @LouiseBonnard ?

LouiseBonnard:
@colinegin, @atomiix, it suits me fine! Let's keep this list and update the last line as follows:

```
It looks like the value is invalid:
- Percent or amount value must be greater than 0.
- Percent value cannot exceed 100.
- Discount value cannot exceed the total price of this order.
```

matks (dismissed):

jolelievre:
@atomiix  the behat tests have been fixed on 1.7.7.x branch you need to rebase your PR

matks (dismissed):

Progi1984 (dismissed):

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -65,12 +66,33 @@ public function handle(AddCartRuleToOrderCommand $command): void
         }
 
         $cartRules = [];
-        $discountValue = (float) str_replace(',', '.', $command->getDiscountValue());
+        $discountValue = (float) (string) $command->getDiscountValue();
```
PierreRambaud, line 69: I'm not a big fan of double cast without comment.

src/Core/Domain/Order/Exception/InvalidCartRuleValueException.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Order\Exception;
```
PierreRambaud:
```suggestion
namespace PrestaShop\PrestaShop\Core\Domain\CartRules\Exception;
```
Should be in this namespace, it's not only related to Orders :thinking:

PierreRambaud (dismissed):

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
jolelievre: It's used in a lot of handlers that use `Decimal\Number` Sadly, if we add a comment each time this is used the codebase is gonna be crowded pretty quick
That's why I wanted to update the library https://github.com/PrestaShop/decimal/issues/6
But as @eternoendless mentioned, the core issue is that we mix float and Number for now

jolelievre (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1428,6 +1429,13 @@ private function getErrorMessages(Exception $e)
                 'Only numbers and decimal points (".") are allowed in the amount fields, e.g. 10.50 or 1050.',
                 'Admin.Orderscustomers.Notification'
             ),
+            InvalidCartRuleDiscountValueException::class => sprintf(
+                '%s<ol><li>%s</li><li>%s</li><li>%s</li></ol>',
+                $this->trans('It looks like the value is invalid:', 'Admin.Orderscustomers.Notification'),
+                $this->trans('Discount value must be greater than 0.', 'Admin.Orderscustomers.Notification'),
```
```suggestion
                $this->trans('Percent or amount value must be greater than 0.', 'Admin.Orderscustomers.Notification'),
```

LouiseBonnard: cf. https://github.com/PrestaShop/PrestaShop/pull/17852/#issuecomment-604471739

atomiix: @LouiseBonnard my bad, I didn't see you updated [your comment](https://github.com/PrestaShop/PrestaShop/pull/17852#issuecomment-604471739)!

LouiseBonnard: No worries, thanks @atomiix!

PierreRambaud (dismissed):

jolelievre (dismissed):

arouiadib (dismissed):

matks (dismissed):

Progi1984 (dismissed):

jolelievre (dismissed):

matks (dismissed):

SD1982:
Hello @atomiix I still have some exception when : 
1 - add a negative value
2 - add discount for a amount greater than the order price
3 - If adding a discount of the total amount order then adding a free shipping discount 

In the screen record I've made you can the that there is still exception but they are not the same than before : 
https://drive.google.com/file/d/112css0th-cKUVMmoZxMPPkUTHAZmAR3O/view?usp=sharing

Progi1984 (dismissed):

Progi1984 (approved):

PierreRambaud (approved):

matks (approved):

SD1982:
Hi @atomiix this time I think it's the good one ! I still detect a error but it's not really related with your PR I have already declared a new issue #18576 for the error i detected (not validated yet ) . Thanks @atomiix

Progi1984:
Thanks @atomiix

PR #17858, opened 2020-02-26
handle exception properly in modern customer controller

### LINKED ISSUE #17809: Bad validation of customer emails in BO

kpodemski:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | refs #17809
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17809
| How to test?  | Check screenshots, before & after

Before:

image

After:
image

If this change will be approved, some other controllers needs to be changed too.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17811)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
@@ -26,7 +26,7 @@
 
 namespace PrestaShopBundle\Controller\Admin\Sell\Customer;
 
-use Exception;
+use PrestaShopException as Exception;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\BulkDeleteCustomerCommand;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\BulkDisableCustomerCommand;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\BulkEnableCustomerCommand;


matks (changes requested):
I answered in the issue why I think this might not be the best solution 😉 see https://github.com/PrestaShop/PrestaShop/issues/17809#issuecomment-592430684

marionf:
The issue #17809 is fixed by the PR #17846, so I close this PR

PR #17864, opened 2020-02-27
Extract order products to separate query

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fix todo in OrderController: `Optimize this by using a GetPartialOrderForViewing query which loads only the relevant products` and in GetOrderForViewingHandler: `todo $this->displayWarning($this->trans('This product, included in package (' . $product['product_name'] . ') is out of stock: ', array(), 'Admin.Orderscustomers.Notification') . ' ' . $pack_item['product_name']);`. Extracts the products query to allow getting `all` and `paginated` products. `The pagination is kinda fake`, because in handler it still gets all products from Order object model and filters them in php - added todo there for future improvements. The warning of product stock is already handled in controller so that todo is removed.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | No feature changes (only some code refactoring). Make sure that order view page is working as expected. (to be more precise - products listing after various actions behaving as expected)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17864)


src/Adapter/Order/QueryHandler/GetOrderProductsForViewingHandler.php
```diff
@@ -0,0 +1,318 @@
...
+use StockAvailable;
+use Tools;
+use Validate;
+use Warehouse;
+use WarehouseProductLocation;
+
+/**
+ * Handles GetOrderProductsForViewing query using legacy object models
+ */
+final class GetOrderProductsForViewingHandler extends AbstractOrderHandler implements GetOrderProductsForViewingHandlerInterface
```
zuk3975, line 57: Dont judge me on this class content. Its purely copy/paste from GetOrderForViewingHandler which was already merged.

zuk3975:
Found unused var $stockLocationIsAvailable when moving code from GetOrderForVieningHandler.
Investigating if its not forgotten feature.

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -66,4 +66,16 @@ protected function isTaxIncludedInOrder(Order $order): bool
...
+    /**
+     * @param Order $order
+     *
+     * @return int
+     */
+    protected function getOrderTaxCalculationMethod(Order $order): int
+    {
+        $customer = new Customer($order->id_customer);
+
+        return Group::getPriceDisplayMethod((int) $customer->id_default_group);
```
jolelievre, line 79: I like the idea of having this method in the abstract class
Could you add another one to detect if tax is included? Which is often what we really want to know
```php
    /**
     * @param Order $order
     *
     * @return bool
     */
    protected function isTaxIncludedInOrder(Order $order): bool
    {
        return $this-> getOrderTaxCalculationMethod($order) === PS_TAX_INC;
    }
```

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -188,7 +175,9 @@ public function handle(GetOrderForViewing $query): OrderForViewing
             $this->getOrderCustomer($order),
             $this->getOrderShippingAddress($order),
             $this->getOrderInvoiceAddress($order),
-            $this->getOrderProducts($order),
+            $this->getOrderProductsForViewingHandler->handle(
```
jolelievre: I think you can keep this call in a dedicated method `getOrderProducts` The content will be short but it will be cleaner

jolelievre (commented):
Two improvement comments

But what surprises me is that you actually don't use the limit/offset properties from the command, so there is never a pagination Is it planned to add it in another PR?

You asked how to test this feature I guess you could add some behat tests to validate the `GetOrderProductsForViewing`, but then you will need to implement the pagination in this PR

Besides these behat test I guess the QA team should validate that the products block still works correctly

Oh and I just noticed, this pagination url is actually not used any more, because the pagination in the order view is now managed on the frontend..

src/Adapter/Order/QueryHandler/GetOrderProductsForViewingHandler.php
```diff
@@ -263,7 +263,10 @@ public function handle(GetOrderProductsForViewing $query): OrderProductsForViewi
             );
         }
 
-        if (isset($offset, $limit)) {
+        $offset = $query->getOffset();
+        $limit = $query->getLimit();
+
+        if (null !== $offset && $limit) {
```
jolelievre: Why not `null !== $limit` as well? Or maybe `$limit > 0` ?

jolelievre, line 265: Why not `null !== $limit` as well? Or maybe `$limit > 0` ?

jolelievre: Why not `null !== $limit` as well? Or maybe `$limit > 0` ?

jolelievre: Why not `null !== $limit` as well? Or maybe `$limit > 0` ?

zuk3975: in either case if limit is null or 0 we still gonna return all products.

matks (approved):

matks (dismissed):

src/Adapter/Order/QueryHandler/GetOrderProductsForViewingHandler.php
matks: I will never judge you, I swear 👍

src/Adapter/Order/AbstractOrderHandler.php
zuk3975: Its there already as I see, git says you've already added it at 2020-01-09, but its not being used anywhere though :smile:

matks (approved):

jolelievre (approved):

src/Adapter/Order/AbstractOrderHandler.php
jolelievre: Damn you're right 😅 I guess many code has been refacto and thus this function is not used any more

Robin-Fischer-PS:
Thanks @zuk3975 !

Progi1984:
Thanks @zuk3975

PR #17877, opened 2020-02-27
Fix bug when adding a discount on order page

### LINKED ISSUE #13809: Can't add a promo in order page
_Originally posted by @Hlavtox in https://github.com/PrestaShop/PrestaShop/issues/11059#issuecomment-492771674_

I don't know if somebody is trying to make jokes of people or WTF is happening here. I am opening this issue for second time and always somebody quiets it down or says it's fixed when it's not. 

Now there is another issue and it throws 500 even when creating the voucher.
https://youtu.be/MHbHYnx5UAo

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | In some cases, adding a discount on order page throws an exception because the total paid price is not valid, because it was not rounded.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #13809
| How to test?  |  See this: https://github.com/PrestaShop/PrestaShop/issues/13809#issuecomment-492970674. (the test must be done on the migrated page)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17877)

### DIFF
--- a/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
@@ -28,6 +28,7 @@ namespace PrestaShop\PrestaShop\Adapter\Order\CommandHandler;
 
 use CartRule;
 use Configuration;
+use Currency;
 use OrderCartRule;
 use OrderInvoice;
 use PrestaShop\PrestaShop\Adapter\Order\AbstractOrderHandler;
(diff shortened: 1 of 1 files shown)


matthieu-rolland:
The bug happened because the value `$order->total_paid` was not rounded, sometimes it would have many digits after coma, making it invalid.

In the mean time, I rounded other order values, and replaced hardcoded precision of 2 with proper computing precision.

atomiix (approved):

PierreRambaud (approved):

SD1982:
This pull request correctly fix the #13809 bug on migrated page. Thanks @matthieu-rolland

Issue #17817, opened 2020-02-25
BO - We can create a new customer successfully without setting required field "Partner offers"


khouloudbelguith:
**Describe the bug**
We can create from the BO a new customer successfully without setting required field "Partner offers"
![image](https://user-images.githubusercontent.com/16067358/75238651-a82c4680-57c1-11ea-8180-88faf3dd824f.png)

**To Reproduce**
1. Go to BO => Customers => Click on  Set required fields for this section
2. Check the "Partner offers" option
3. Save
4. Create a new customer from BO without enabling Partner offers => the customer is successfully saved => NOK
We should have a red alert like in the last previous version PrestaShop
 
![image](https://user-images.githubusercontent.com/16067358/75238496-60a5ba80-57c1-11ea-8220-ed556383f67d.png)

**Additionnal information**
PrestaShop version: 1.7.6.3
PHP version: 7.2

khouloudbelguith:
This issue is fixed previously in the PS1.7.6.0 with this PR: https://github.com/PrestaShop/PrestaShop/pull/13887

It is ok with PS1.7.6.0 and PS1.7.6.1 and PS1.7.6.2

khouloudbelguith:
Ping @nesrineabdmouleh

colinegin:
We need to check what happened in 1763 which introduced this bug again.

matks:
It seems this is because of PR https://github.com/PrestaShop/PrestaShop/pull/16648 which fixed issue https://github.com/PrestaShop/PrestaShop/issues/16509 (milestone 1.7.6.3)

matks:
@colinegin There is a misunderstanding here.

**When I look into the code**, I can see that, when we create a customer, **we check whether or not in the form the "required field"s have been given to the BO admin** and indeed it works.

**Setting a field to "required" means "we require the BO admin to fill this field".**

There is a misunderstanding on "what is a required field".
- 1st version: "A required field must be filled in by the user" (so the opposite is an optional field). For example "birthday" input, on customer form, is optional. You can fill it with a date or leave it blank.
- 2nd version: "A required field must be provided with a valid value" (= for a checkbox, it must be checked).

[...]

PR #17887, opened 2020-02-28
Make sure checkbox must be checked for required fields for customers

matks:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | For customer required fields, if they are checkboxes they need to be checked for the form to be valid.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/17817
| How to test?  | See ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17887)

### DIFF
--- a/classes/Customer.php
+++ b/classes/Customer.php
@@ -774,6 +774,45 @@ class CustomerCore extends ObjectModel
     }
 
     /**
+     * Validate that fields from Customer that are
+     * - required
+     * - boolean
+     * Must be true
+     *
+     * @return array of errors
+     *
+     * @throws PrestaShopException
+     */
+    public function validateCheckboxRequiredFields()
+    {
+        $this->cacheFieldsRequiredDatabase();
+        $errors = [];
+        $requiredFields = $this->getCachedFieldsRequiredDatabase();
+
+        foreach ($this->def['fields'] as $field => $fieldConfiguration) {
+            if (!in_array($field, $requiredFields)) {
+                continue;
+            }
+            if ($fieldConfiguration['type'] !== self::TYPE_BOOL) {
+                continue;
+            }
+
+            $value = Tools::getValue($field, null);
+
+
+            if ($value === false) {
+                $errors[$field] = $this->trans(
+                    'The field %s is required.',
+                    [self::displayFieldName($field, get_class($this), true)],
+                    'Admin.Notifications.Error'
+                );
+            }
+        }
+
+        return $errors;
+    }
+
+    /**
      * Light back office search for customers.
      *
      * @param string $query Searched string
--- a/src/Adapter/Customer/CommandHandler/AbstractCustomerHandler.php
+++ b/src/Adapter/Customer/CommandHandler/AbstractCustomerHandler.php
@@ -27,6 +27,7 @@
 namespace PrestaShop\PrestaShop\Adapter\Customer\CommandHandler;
 
 use Customer;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\BadCustomerRequiredFieldsException;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\CustomerNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\MissingCustomerRequiredFieldsException;
 use PrestaShop\PrestaShop\Core\Domain\Customer\ValueObject\CustomerId;
@@ -71,6 +72,20 @@ abstract class AbstractCustomerHandler
                 sprintf(
                     'One or more required fields for customer are missing. Missing fields are: %s',
                     implode(',', $missingFields)
+                )
+            );
+        }
+
+        $checkboxErrors = $customer->validateCheckboxRequiredFields();
+
+        if (!empty($checkboxErrors)) {
+            $badFields = array_keys($checkboxErrors);
+
+            throw new BadCustomerRequiredFieldsException(
+                $badFields,
+                sprintf(
+                    'One or more required checkboxes for customer are not checked. Faulty fields are: %s',
+                    implode(',', $badFields)
                 )
             );
         }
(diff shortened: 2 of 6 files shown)


src/Adapter/Customer/CommandHandler/EditCustomerHandler.php
```diff
@@ -82,7 +82,8 @@ public function handle(EditCustomerCommand $command)
         // to check if required fields are set
         if ($command->isPartnerOffersSubscribed() !== null) {
             $_POST[RequiredField::PARTNER_OFFERS] = $command->isPartnerOffersSubscribed();
-        } elseif ($command->isNewsletterSubscribed() !== null) {
+        }
+        if ($command->isNewsletterSubscribed() !== null) {
```
matks, line 86: Not related to the issue, but I think it makes sense 🤔

src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
```diff
@@ -838,6 +839,16 @@ private function getErrorMessages(Exception $e)
                     ),
                 ]
             ),
+            BadCustomerRequiredFieldsException::class => $this->trans(
+                'The required field %s is not valid.',
```
matks, line 843: @LouiseBonnard If someone sets "newsletter partners" as required field and then attempts to create a customer WITHOUT checking the checkbox, I display this message to him.

"The required field optin is not valid."

It's a bit generic, but since this could be used for other required-fields-that-is-not-properly-filled usecases (not only checkboxes) it's the best I can think of.

src/PrestaShopBundle/Resources/views/Admin/Sell/Customer/Blocks/Index/required_fields.html.twig
```diff
@@ -34,6 +34,7 @@
         <div class="alert-text">
           <p>{{ 'Select the fields you would like to be required for this section.'|trans({}, 'Admin.Orderscustomers.Help') }}</p>
           <p>{{ 'Please make sure you are complying with the opt-in legislation applicable in your country.'|trans({}, 'Admin.Orderscustomers.Help') }}</p>
+          <p>{{ "Warning: setting a checkbox to 'required' will require users, from back-office and front-office, to check it."|trans({}, 'Admin.Orderscustomers.Help') }}</p>
```
matks, line 37: @LouiseBonnard I try to explain this https://github.com/PrestaShop/PrestaShop/issues/17817#issuecomment-591514972

But it's hard with so few words 😅 . The idea is to explain the confusion between a "required to be checked" checkbox and a "mandatory/necessary" field (like birthdate).

classes/Customer.php
```diff
@@ -773,6 +773,45 @@ public static function checkPassword($idCustomer, $passwordHash)
...
 
+    /**
+     * Validate that fields from Customer that are
+     * - required
+     * - boolean
+     * Must be true
+     *
+     * @return array of errors
+     *
+     * @throws PrestaShopException
```
PierreRambaud, line 784: You throw nothing Jon Throw :thinking:

classes/Customer.php
```diff
@@ -773,6 +773,45 @@ public static function checkPassword($idCustomer, $passwordHash)
...
+        foreach ($this->def['fields'] as $field => $fieldConfiguration) {
+            if (!in_array($field, $requiredFields)) {
+                continue;
+            }
+            if ($fieldConfiguration['type'] !== self::TYPE_BOOL) {
+                continue;
+            }
+
+            $value = Tools::getValue($field, null);
+
```
PierreRambaud, line 801:
```suggestion
```

classes/Customer.php
```diff
@@ -773,6 +773,45 @@ public static function checkPassword($idCustomer, $passwordHash)
...
+
+            if ($value === false) {
```
PierreRambaud, line 803:
```suggestion
            if ($value !== true) {
```
What if the $field is null or 'blyat' ?

src/Adapter/Customer/CommandHandler/AbstractCustomerHandler.php
```diff
@@ -74,5 +75,19 @@ protected function assertRequiredFieldsAreNotMissing(Customer $customer)
                 )
             );
         }
+
+        $checkboxErrors = $customer->validateCheckboxRequiredFields();
+
+        if (!empty($checkboxErrors)) {
+            $badFields = array_keys($checkboxErrors);
```
PierreRambaud, line 82: You never use `$checkboxErrors`, is it needed to return translated string?

src/PrestaShopBundle/Controller/Admin/Sell/Customer/CustomerController.php
LouiseBonnard: Indeed! But actually it is not a box to check but a switch button, right? In this case, I'd display a more specific notification like _You set the newsletter partners as a required option for customers, enable it to continue._, what do you think?

src/PrestaShopBundle/Resources/views/Admin/Sell/Customer/Blocks/Index/required_fields.html.twig
LouiseBonnard:
```suggestion
          <p>{{ 'Note that setting an option as required enjoins users to complete it, whether it is on the back office or the front office.'|trans({}, 'Admin.Orderscustomers.Help') }}</p>
```

LouiseBonnard: It is clearer with this formulation? You tell me. :-)

classes/Customer.php
matks: Copy-paste from similar method in ObjectModel. It might throw, I dont know, it's legacy 😅

classes/Customer.php
matks: If the field is null, then another check will take care of this. We are only looking for `false` values.
For `blyat` ... well I have no idea ! What are we supposed to do with a checkbox whose provided value is `blyat` ..? It could be a valid answer 😆

src/Adapter/Customer/CommandHandler/AbstractCustomerHandler.php
matks: You mean I never use the values of `$checkboxErrors` ? Somehow I thought "eyh maybe a module developer could need it" but now it looks like over-engineering, right ?

matks:
Closing the PR as PM decided this behavior is actually wanted (see https://github.com/PrestaShop/PrestaShop/issues/17817#issuecomment-592515362)

PR #17907, opened 2020-03-02
Fix TinyMCEEditor that shouldn't handle language switching by adding a TranslatableField

### LINKED ISSUE #17855: Fix TinyMCEEditor that shouldn't handle language switching
**Describe the bug**
To fix this initial issue https://github.com/PrestaShop/PrestaShop/issues/16646
A feature was added into TinyMCEEditor in this PR https://github.com/PrestaShop/PrestaShop/pull/17444
So it is now able to dispatch an event `selectedLangauge` when its parent navigation switches tab
However this feature is already managed by the `TranslatableField` component
https://github.com/PrestaShop/PrestaShop/blob/develop/admin-dev/themes/new-theme/js/components/translatable-field.js#L31

This breaks the separation of responsibility:
- TinyMCEEditor is responsible of initiating rich text areas, it is also able to detect a `'shown.bs.tab'` event from its parent This is necessary because tab switching breaks the editor content, so it reset its content to respond correctly to the tab switching (that may or may not be related to a language switch)
- TranslatableField is able to manage tab navigation for languages (to my knowledge it is the equivalent of TranslatableInput but for text areas)

**Refacto**
Steaps to fix this architectural mix:
- remove from TinyMCEEditor the event features
- to fix the initial issue (brand/supplier forms) add the TranslatableField in the form js (for supplier it's https://github.com/PrestaShop/PrestaShop/blob/develop/admin-dev/themes/new-theme/js/pages/supplier/supplier-form.js#L53)

Here is an example of a correct form https://github.com/PrestaShop/PrestaShop/blob/develop/admin-dev/themes/new-theme/js/pages/manufacturer/index.js#L63

**Possible Improvement**
It appears in this case a component was simply forgotten, to avoid this problem in the future maybe we could create component that would include both `TranslatableInput` and `TranslatableField` Or we could merge the two components into one

**Additional information**
PrestaShop version: 1.7.7
PHP version: N/A

NeOMakinG:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Switch of lang should not be handled in TinyMCE as it is already managed by TranslatableField
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17855.
| How to test?  | Steps to reproduce the behavior:<br>- Go to back office<br>- Click on catalog -> brands & suppliers -> suppliers add new supplier<br>- Switch language in the description area<br>- See if tinymce lang changed too

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17907)


jolelievre (dismissed):
Nice job @NeOMakinG

matks (dismissed):

PierreRambaud (dismissed):

Progi1984 (dismissed):

boubkerbribri:
@NeOMakinG please resolve conflicts

Progi1984 (dismissed):

matthieu-rolland (dismissed):

jolelievre (dismissed):

matthieu-rolland (dismissed):

jolelievre (dismissed):

matks (dismissed):

matks (approved):

Progi1984 (approved):

Progi1984:
Thanks @NeOMakinG

PR #17914, opened 2020-03-02
Add index on product table for fresh install

### LINKED ISSUE #16323: Search by product reference causes full table scan

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Some indexes have been added in 1.7.7 upgrade scripts, their equivalent need to be added in initial script used in fresh install
| Type?         | improvement
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Addition to PR https://github.com/PrestaShop/PrestaShop/pull/16440 Fixes https://github.com/PrestaShop/PrestaShop/issues/16323
| How to test?  | Perform a fresh install of PrestaShop, everything needs to work correctly Check in database that the new indexes are present on product table (Also check the initial PR for more feature oriented tests)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17914)

### DIFF
--- a/install-dev/data/db_structure.sql
+++ b/install-dev/data/db_structure.sql
@@ -1666,6 +1666,8 @@ CREATE TABLE `PREFIX_product` (
   `pack_stock_type` int(11) unsigned DEFAULT '3' NOT NULL,
   `state` int(11) unsigned NOT NULL DEFAULT '1',
   PRIMARY KEY (`id_product`),
+  INDEX reference_idx(`reference`),
+  INDEX supplier_reference_idx(`supplier_reference`),
   KEY `product_supplier` (`id_supplier`),
   KEY `product_manufacturer` (`id_manufacturer`, `id_product`),
   KEY `id_category_default` (`id_category_default`),


atomiix (approved):

PierreRambaud (approved):

Robin-Fischer-PS:
It's OK, after install I have the index : 
pr17914 index

Thanks @jolelievre !

Progi1984:
Thanks @jolelievre

PR #17920, opened 2020-03-02
Wrong redirection when using the quick search for a category

### LINKED ISSUE #17908: BO - Quick search - redirect to the Categories list page
#### Describe the bug

In the BO, When we search a category with the quick search, we are always redirected to the Categories list page instead of the Edit category page

#### Expected behavior
Redirection to the Categories list page
**Explain what you expected to happen instead.**
Redirection to the Edit category page

#### Steps to Reproduce

1. Go to BO
2. Search with the "Women" category for example in the top of the page
image

3. Click on the "Edit" button
image

3. Redirection to the Categories list page  instead of the Edit category page

It is ok with PS1.7.5.2
https://drive.google.com/file/d/1yZ9d7uD_NwqDhRP5irBRR-cVIgTlTZz3/view

#### Additional information

* PrestaShop version: 1.7.6.0, PS1.7.6.3, Branch 1.7.6.x
* PHP version: 7.2

PululuK:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | In the BO, When we search a category with the quick search, we are always redirected to the Categories list page instead of the Edit category page. `Tools::getPath` Became obsolete after migrating the category page. You must update the URLs to the new symfony controller.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17908.
| How to test?  | See https://github.com/PrestaShop/PrestaShop/issues/17908 by @khouloudbelguith .

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17920)

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -2861,10 +2861,10 @@ FileETag none
         $tab['Files'] = array();
         if (Configuration::get('PS_REWRITING_SETTINGS')) {
             $sql = 'SELECT DISTINCT ml.url_rewrite, l.iso_code
-					FROM ' . _DB_PREFIX_ . 'meta m
-					INNER JOIN ' . _DB_PREFIX_ . 'meta_lang ml ON ml.id_meta = m.id_meta
-					INNER JOIN ' . _DB_PREFIX_ . 'lang l ON l.id_lang = ml.id_lang
-					WHERE l.active = 1 AND m.page IN (\'' . implode('\', \'', $disallow_controllers) . '\')';
+                FROM ' . _DB_PREFIX_ . 'meta m
+                INNER JOIN ' . _DB_PREFIX_ . 'meta_lang ml ON ml.id_meta = m.id_meta
+                INNER JOIN ' . _DB_PREFIX_ . 'lang l ON l.id_lang = ml.id_lang
+                WHERE l.active = 1 AND m.page IN (\'' . implode('\', \'', $disallow_controllers) . '\')';
             if ($results = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS($sql)) {
                 foreach ($results as $row) {
                     $tab['Files'][$row['iso_code']][] = $row['url_rewrite'];
@@ -4264,10 +4264,11 @@ exit;
                 $n = 1;
                 $n_categories = (int) count($categories);
                 foreach ($categories as $category) {
-                    $link = Context::getContext()->link->getAdminLink('AdminCategories');
-                    $edit = '<a href="' . Tools::safeOutput($link . '&id_category=' . (int) $category['id_category'] . '&' . (($category['id_category'] == 1 || $home) ? 'viewcategory' : 'updatecategory')) . '" title="' . ($category['id_category'] == Category::getRootCategory()->id_category ? 'Home' : 'Modify') . '"><i class="icon-' . (($category['id_category'] == Category::getRootCategory()->id_category || $home) ? 'home' : 'pencil') . '"></i></a> ';
+                    $action = (($category['id_category'] == (int) Configuration::get('PS_HOME_CATEGORY') || $home) ? 'index' : 'updatecategory');
+                    $link = Context::getContext()->link->getAdminLink('AdminCategories', true, ['action' => $action, 'id_category' => (int) $category['id_category']]);
+                    $edit = '<a href="' . Tools::safeOutput($link) . '" title="' . ($category['id_category'] == Category::getRootCategory()->id_category ? 'Home' : 'Modify') . '"><i class="icon-' . (($category['id_category'] == Category::getRootCategory()->id_category || $home) ? 'home' : 'pencil') . '"></i></a> ';
                     $full_path .= $edit .
-                                  ($n < $n_categories ? '<a href="' . Tools::safeOutput($url_base . '&id_category=' . (int) $category['id_category'] . '&viewcategory&token=' . Tools::getAdminToken('AdminCategories' . (int) Tab::getIdFromClassName('AdminCategories') . (int) $context->employee->id)) . '" title="' . htmlentities($category['name'], ENT_NOQUOTES, 'UTF-8') . '">' : '') .
+                                  ($n < $n_categories ? '<a href="' . Tools::safeOutput($link) . '" title="' . htmlentities($category['name'], ENT_NOQUOTES, 'UTF-8') . '">' : '') .
                                   (!empty($highlight) ? str_ireplace($highlight, '<span class="highlight">' . htmlentities($highlight, ENT_NOQUOTES, 'UTF-8') . '</span>', $category['name']) : $category['name']) .
                                   ($n < $n_categories ? '</a>' : '') .
                                   (($n++ != $n_categories || !empty($path)) ? ' > ' : '');
--- a/src/PrestaShopBundle/Resources/config/routing/admin/sell/catalog/categories.yml
+++ b/src/PrestaShopBundle/Resources/config/routing/admin/sell/catalog/categories.yml
@@ -5,6 +5,8 @@ admin_categories_index:
         _controller: PrestaShopBundle:Admin\Sell\Catalog\Category:index
         _legacy_controller: AdminCategories
         _legacy_link: AdminCategories
+        _legacy_parameters:
+            id_category: categoryId
         categoryId: null
     requirements:
         categoryId: '\d+'


classes/Tools.php
```diff
@@ -4264,10 +4264,11 @@ public static function getPath($url_base, $id_category, $path = '', $highlight =
                 $n = 1;
                 $n_categories = (int) count($categories);
                 foreach ($categories as $category) {
-                    $link = Context::getContext()->link->getAdminLink('AdminCategories');
-                    $edit = '<a href="' . Tools::safeOutput($link . '&id_category=' . (int) $category['id_category'] . '&' . (($category['id_category'] == 1 || $home) ? 'viewcategory' : 'updatecategory')) . '" title="' . ($category['id_category'] == Category::getRootCategory()->id_category ? 'Home' : 'Modify') . '"><i class="icon-' . (($category['id_category'] == Category::getRootCategory()->id_category || $home) ? 'home' : 'pencil') . '"></i></a> ';
+                    $action = (($category['id_category'] == (int) Configuration::get('PS_HOME_CATEGORY') || $home) ? 'admin_categories_index' : 'admin_categories_edit');
```
jolelievre: Nice work @PululuK Although in this case you can take advantage of the `_legacy_link` feature https://devdocs.prestashop.com/1.7/development/architecture/migration-guide/controller-routing/#the-legacy-link-property

Which is able to automatically convert a legacy link into a symfony route, provided you correctly use the arguments and not copy paste them by hand. So you could use the actions `index` and `updatecategory` in place of (respectively) the routes `admin_categories_index` and `admin_categories_edit`

This feature also manages an automatic redirection even with urls formatted like legacy, however in this case it doesn't work because the mapping of the `id_category` is missing for the `admin_categories_index` route (probably because the parameter was managed after the initial creation).

Anyway I believe you could fix this whole bug just fixing the routing configuration in `src/PrestaShopBundle/Resources/config/routing/admin/sell/catalog/categories.yml`

```yaml
admin_categories_index:
    path: /{categoryId}
    methods: [GET, POST]
    defaults:
        _controller: PrestaShopBundle:Admin\Sell\Catalog\Category:index
        _legacy_controller: AdminCategories
        _legacy_link: AdminCategories
        _legacy_parameters:
            id_category: categoryId
        categoryId: null
    requirements:
        categoryId: '\d+'
```
Only by adding the missing `_legacy_parameters` part

jolelievre (commented):
Thanks @PululuK 
Nice handling of the `Link::getAdminLink` feature ^^
I have a suggestion for a fix that might be a bit simpler though

PululuK:
> Thanks @PululuK
> Nice handling of the `Link::getAdminLink` feature ^^
> I have a suggestion for a fix that might be a bit simpler though

classes/Tools.php
PululuK: Hi @jolelievre 
Thank for you feedback ....


Very interesting your proposal ... Effectively is missing the `_legacy_parameters` (in `src/PrestaShopBundle/Resources/config/routing/admin/sell/catalog/categories.yml::admin_categories_index`), adding this will prevent the redirection from always being on the category home, but this will not solve this problem for editing (which is the case with this issue : https://github.com/PrestaShop/PrestaShop/issues/17908).

Just with that :

```yml
admin_categories_edit:
    path: /{categoryId}/edit
    methods: [GET, POST]
    defaults:
        _controller: PrestaShopBundle:Admin\Sell\Catalog\Category:edit
        _legacy_controller: AdminCategories
        _legacy_link: AdminCategories:updatecategory
        _legacy_parameters:
            id_category: categoryId
    requirements:
        categoryId: \d+
```

`index.php/sell/catalog/categories?_token=xxxx&id_category=xxx&updatecategory` should redirect here `/index.php/sell/catalog/categories/xxx/edit?_token=xxxxx`, But this is not the case ... see https://github.com/PrestaShop/PrestaShop/issues/17908

I think in this case the problem is much more complex then ...

jolelievre: Oh you're right!! That's because the initial call to `getAdminLink` doesn't have the action parameter So it converts it to the index route (which makes sense), and then append the additional parameter

This would work well if the parameters were inject completely in `getAdminLink` So I guess the solution is a mix between what you did and fixing the routing

Once the routing is fixed, then you can really use the parameter `action` instead of `route` so you would need to replace:

```php
$action = (($category['id_category'] == (int) Configuration::get('PS_HOME_CATEGORY') || $home) ? 'index' : 'updatecategory');
```

and 

```php
$link = Context::getContext()->link->getAdminLink('AdminCategories',true, array('action'=> $action,'categoryId'=> (int) $category['id_category']));
```

PululuK: It also works, I just tested! Could you tell me how it is faster than injecting symfony parameters directly? Because it seems more logical to directly inject the parameters for the new controller instead of legacy. See https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/classes/Link.php#L705

PululuK: Done !

jolelievre: The sf route params were initially introduced to allow creating symfony url through the `Link::getAdminLink` function (at the time the `_legacy_link` feature didn't exist)

Whereas the `_legacy_link` feature along with the `LegacyUrlConverter` service was specifically designed from a migration point of view. It has two main advantages:
- once the `_legacy_link` is correctly set the legacy urls are automatically redirected (when they find a matching route), so that even hard coded urls are correctly redirected to the new migrated page
- for modules developer you can use the getAdminLink in any version, if the controller has been migrated it will output the symfony url, if not it will output the legacy controller one

The only drawback (as you could see), is when parameters are concatenated after the getAdminLink call, because they are concatenated after the symfony url. We have no solution for this problem today, except correctly using the function parameters (which is sadly not possible in some early 1.7 versions).


As for this special case, since it's inside the core you are right: it doesn't bring much values It might be slightly less performant because it needs to find the matching route, but some cache policy has been implemented so the matching process is not that long

You might wonder, what is the advantage because this small performance drawback then? If I could inject the appropriate route directly. I would see two side advantages:
- if the symfony route is modified somehow (its name, parameters, ...) the modification would be automatically updated via this call
- for people using the previous versions it also gives them a valid example of how they should use the getAdminLink correctly and how they could integrate it (wether in the core or in their module)

Also, but this one is personal, I don't think mixing symfony routes inside the legacy classes is a good thing to do if we can avoid it.

PululuK: Thank you @jolelievre  ! 
I agree on some points!
Done

classes/Tools.php
```diff
@@ -4264,10 +4264,11 @@ public static function getPath($url_base, $id_category, $path = '', $highlight =
...
+                    $link = Context::getContext()->link->getAdminLink('AdminCategories',true, array('route'=>$action,'categoryId'=>(int) $category['id_category']));
```
jolelievre: Sorry I missed this one, last modif you should use the legacy parameter `id_category` and let the converter rename it into `categoryId` for symfony routing

This is one of the advantages I mentioned, you can always inject the legacy parameter and if we, someday, change the parameter name in the routing it will still work.

PululuK: Done !
Thank you @jolelievre

jolelievre: Thank YOU! I just approved the PR!

jolelievre (dismissed):
Thank you @PululuK

PierreRambaud (changes requested):
Indentation is broken, please use php-cs-fixer

src/PrestaShopBundle/Resources/config/routing/admin/sell/catalog/categories.yml
```diff
@@ -5,6 +5,8 @@ admin_categories_index:
         _controller: PrestaShopBundle:Admin\Sell\Catalog\Category:index
         _legacy_controller: AdminCategories
         _legacy_link: AdminCategories
+        _legacy_parameters:
+            id_category: categoryId
```
matks, line 9: @jolelievre this parameter will not become mandatory for other URL generation statements ?

jolelievre: Good question! But I don't think so the converter only uses them to try and rename the parameters in their routing equivalent:
https://github.com/PrestaShop/PrestaShop/blob/7633be8294d90148d2bdfe3fd1c2d403e72ec91b/src/PrestaShopBundle/Routing/Converter/LegacyUrlConverter.php#L151

Nothing special happens if they are not present, or if any additional parameter is used. At worst they will be appended to the generated url

matks: Ok 👍 good to me then

PululuK:
> Indentation is broken, please use php-cs-fixer

Hi @PierreRambaud 
Can you tell me where? In php file or yml ?

jolelievre:
@PululuK We recently disabled PrettyCI because it had some issues, and it will soon stop to work (bu the end of the year)
But we haven't reintroduced the Github action that validates code style on the 176 branch, so @PierreRambaud detects them with his super eyes!! ^^

Anyway to fix it you can just run `php ./vendor/bin/php-cs-fixer fix` from the root folder

PululuK:
> @PululuK We recently disabled PrettyCI because it had some issues, and it will soon stop to work (bu the end of the year)


Yes I am aware, I received their email! We must quickly find a tool to replace, even if I sincerely believe that the eyes of @PierreRambaud  are more effective

PululuK:
Done @PierreRambaud (AKA PrettyCI-Man)  @jolelievre  !

jolelievre (dismissed):

classes/Tools.php
```diff
@@ -4264,10 +4264,11 @@ public static function getPath($url_base, $id_category, $path = '', $highlight =
                 $n = 1;
                 $n_categories = (int) count($categories);
                 foreach ($categories as $category) {
-                    $link = Context::getContext()->link->getAdminLink('AdminCategories');
-                    $edit = '<a href="' . Tools::safeOutput($link . '&id_category=' . (int) $category['id_category'] . '&' . (($category['id_category'] == 1 || $home) ? 'viewcategory' : 'updatecategory')) . '" title="' . ($category['id_category'] == Category::getRootCategory()->id_category ? 'Home' : 'Modify') . '"><i class="icon-' . (($category['id_category'] == Category::getRootCategory()->id_category || $home) ? 'home' : 'pencil') . '"></i></a> ';
+                    $action = (($category['id_category'] == (int) Configuration::get('PS_HOME_CATEGORY') || $home) ? 'index' : 'updatecategory');
+                    $link = Context::getContext()->link->getAdminLink('AdminCategories',true, array('action'=> $action,'id_category'=> (int) $category['id_category']));
```
Progi1984:
```suggestion
                    $link = Context::getContext()->link->getAdminLink('AdminCategories', true, ['action'=> $action,'id_category'=> (int) $category['id_category']]);
```

classes/Tools.php
```diff
@@ -4264,10 +4264,11 @@ public static function getPath($url_base, $id_category, $path = '', $highlight =
...
+                    $edit = '<a href="' . Tools::safeOutput($link).'" title="' . ($category['id_category'] == Category::getRootCategory()->id_category ? 'Home' : 'Modify') . '"><i class="icon-' . (($category['id_category'] == Category::getRootCategory()->id_category || $home) ? 'home' : 'pencil') . '"></i></a> ';
```
Progi1984:
```suggestion
                    $edit = '<a href="' . Tools::safeOutput($link) . '" title="' . ($category['id_category'] == Category::getRootCategory()->id_category ? 'Home' : 'Modify') . '"><i class="icon-' . (($category['id_category'] == Category::getRootCategory()->id_category || $home) ? 'home' : 'pencil') . '"></i></a> ';
```

Progi1984 (dismissed):

PierreRambaud:
Can you please  target the 1.7.7.x branch instead of 1.7.6.x?

PululuK:
> Can you please target the 1.7.7.x branch instead of 1.7.6.x?

hi @PierreRambaud 
Why and how should I do that please?

PierreRambaud:
@PululuK Because this one is not major and we will release a 1.7.7.0 really soon :wink:

PululuK:
> @PululuK Because this one is not major and we will release a 1.7.7.0 really soon

Ok ... I updated the tag to 1.7.7.X.

jolelievre:
@PululuK you also need to change the target branch (actually that's the most important)
You can do it by editing the PR and changing the base, although ideally you should rebase because there might be some conflict If it's too complicated you can recreate a branch from 1.7.7.x branch and cherry pick the commit
I can do it for you if you want?

PululuK:
> @PululuK you also need to change the target branch (actually that's the most important)
> You can do it by editing the PR and changing the base, although ideally you should rebase because there might be some conflict If it's too complicated you can recreate a branch from 1.7.7.x branch and cherry pick the commit
> I can do it for you if you want?

Yes you can do it ! thank you so much @jolelievre

jolelievre:
Ok @PululuK , the PR is rebased and the target is 1.7.7.x

Progi1984:
Please fix PHPCSFixer issues.

PululuK:
> Please fix PHPCSFixer issues.

Hi @Progi1984 
Sorry! but your test tool is not verbose enough, I have no information on the lines concerned ... I miss PrettyCi 😔

PululuK:
> Please fix PHPCSFixer issues.

Done !

Progi1984 (dismissed):

jolelievre:
Hello @PululuK 
So it's been decided that there will be a 1.7.6.5 version in the end 😅
So this PR can go back to the 1.7.6.x branch, I'm gonna make the rebase to avoid bothering you 😉

jolelievre (dismissed):
Thank you @PululuK 
The PR is back on track for 1.7.6!

Progi1984 (dismissed):

PierreRambaud (dismissed):

PululuK:
Hi @PierreRambaud 

Is it normal that travisCI is blocked? I have to do something ?

PierreRambaud:
@PululuK It's not normal, it happens sometimes :sob:

SD1982:
Hi @PululuK It seems that there is an issue with Travis can you check this please ?

PululuK:
> @PululuK It's not normal, it happens sometimes sob

Hi @PierreRambaud 
Can you relaunch Travis please? It was green before rebase to 1.7.6.x

jolelievre:
Actually the problem is not from the rebase nor this PR but some fixtures in the tests suites which are not valid any more
Here is the PR that fixes it https://github.com/PrestaShop/PrestaShop/pull/18309

jolelievre:
@PululuK  I'll rebase this PR as soon as the other one is merged 😉

PululuK:
> Actually the problem is not from the rebase nor this PR but some fixtures in the tests suites which are not valid any more
Here is the PR that fixes it #18309

You are in the future ! 
Thanks @jolelievre  @PierreRambaud

jolelievre:
Rebased! should be good this time

jolelievre (approved):

PululuK:
> Hi @PululuK It seems that there is an issue with Travis can you check this please ?


Hi @SD1982 
It is Done .. Big thanks guys @jolelievre @PierreRambaud !

PierreRambaud (approved):

SD1982:
This PR correctly fix #17908 ! Thanks @PululuK !

Progi1984:
Thanks @PululuK

Gipielle:
Hello, this fix works but is not complete,

if you click in the category link now it goes always in modify both (pencil right) and also category name (wrong) category name should go into the cateogoty so you can see list of subcategory.

PululuK:
> Hello, this fix works but is not complete,
> 
> if you click in the category link now it goes always in modify both (pencil right) and also category name (wrong) category name should go into the cateogoty so you can see list of subcategory.

Hi @Gipielle ,

Could you open a new issue dedicated to this? Do not hesitate to tag me please!

Thank you !

Gipielle:
@PululuK done here #18732
Thanks

PR #17928, opened 2020-03-02
Fix 'Invalid order->total_discounts after deleteting voucher from backoffice

### LINKED ISSUE #11059: Invalid order->total_discounts after deleteting voucher from backoffice
Sometimes, when you delete a voucher from an order in back office, you will get an exception and the order gets messed up.

**How to reproduce:**
- Rouding setup: (mathematically up (default), each line rounding, zero decimal places.
- Make a product with price for example 5800
- Make a cart rule: 5% discount with tax for all orders
- Go to admin, delete the voucher, exception

**Cause:**
Total discounts with tax is rounded to 290; without tax to 240, 
Cart rule is NOT ROUNDED - with tax 290 and without tax 239,65.
When you delete the voucher, Prestashop does this:
`// Update amounts of order`
`$order->total_discounts -= $order_cart_rule->value;`
`$order->total_discounts_tax_incl -= $order_cart_rule->value;`
`$order->total_discounts_tax_excl -= $order_cart_rule->value_tax_excl;`

The resulting total discounts W/O TAX is 0,35, which is bigger than WITH TAX, which is invalid. Sometimes total_discounts is negative, which is also invalid.

**Possible resolution:**
Instead of subtracting the voucher value from total discounts...
1. Set total discounts to 0. 
2. Check for remaining vouchers of this order, sum their values > get 100% correct result of total_discounts. 
3. Recalculate the order total and update it.

**Versions:**
Happens in all 1.6 and 1.7 versions, the code is still the same.
PHP 5.6, 7.1 and currently 7.2

**Screenshots:**
update - exception
screenshot

arouiadib:
| Questions     | Answers
| Branch?       |  1.7.7.x
| Description?  | This PR is a follow to the foot-steps of @matthieu-rolland in fixing a similar issue (13809), but somewhere else.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    |  no
| Deprecations? | no
| Fixed ticket? | Fixes #11059
| How to test?  | Reproduce the error as mentioned by Hlavox in issue ticket, but in migrated Order View page.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17928)

### DIFF
--- a/src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
@@ -47,6 +47,10 @@ final class DeleteCartRuleFromOrderHandler extends AbstractOrderHandler implemen
         $order = $this->getOrderObject($command->getOrderId());
         $orderCartRule = new OrderCartRule($command->getOrderCartRuleId());
 
+        $computingPrecision = new ComputingPrecision();
+        $currency = new Currency((int) $order->id_currency);
+        $precision = $computingPrecision->getPrecision($currency->precision);
+
         if (!Validate::isLoadedObject($orderCartRule) || $orderCartRule->id_order != $order->id) {
             throw new OrderException('Invalid order cart rule provided.');
         }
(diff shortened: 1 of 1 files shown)


arouiadib:
I closed this because I am still not sure how to fix it in the best way. I also lack some knowledge. I will be back asap.

PR #17938, opened 2020-03-03
Fix conversion when adding payment to order

### LINKED ISSUE #9850: [BOOM-4098] CLONE - Currency issue while adding payment to order
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-4098).

- _**Reporter:**_ salma.moakhar
- _**Created at:**_ Mon, 23 Oct 2017 11:57:53 +0200

<p>PS 1.7.2.3, with Euros and Dollars available.<br/>
Default currency is Euros.</p>

<p>In Front, I place an order, paid in dollar through bankwire.<br/>
In back-office, in the order page, I add a payment of 10 euro to this order.</p>

<p>In database, the field total_paid_real is equal to 10.</p>

<p>In my opinion, it should be 11,75, to be in the same currency than than the order itself</p>

sowbiba:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Conversion issue when adding a payment in another currency than the order one
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9850 http://forge.prestashop.com/browse/BOOM-4098
| How to test?  | <ul><li>Create a shop with Euro as default currency and USD as secondary</li><li>Create an order in USD with bankwire as payment method</li><li>In the BO, edit the order and add a new payment in Euro</li><li>In the Database the value for total_paid_real is incorrect. You can see it in the header of the payments section too</li></ul>

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17938)

### DIFF
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1852,7 +1852,7 @@ class OrderCore extends ObjectModel
         if ($order_payment->id_currency == $this->id_currency) {
             $this->total_paid_real += $order_payment->amount;
         } else {
-            $this->total_paid_real += Tools::ps_round(Tools::convertPrice($order_payment->amount, $order_payment->id_currency, false), 2);
+            $this->total_paid_real += Tools::ps_round(Tools::convertPrice($order_payment->amount, $this->id_currency, false), 2);
         }
 
         // We put autodate parameter of add method to true if date_add field is null
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1848,11 +1848,35 @@ class OrderCore extends ObjectModel
             $order_payment->date_add .= ' ' . date('H:i:s');
         }
 
+        /*
+         * 4 cases
+         *
+         * Order is in default_currency + Payment is in Order currency
+         *    for example default = 1, order = 1, payment = 1
+         *    ==> NO conversion to do
+         * Order is in default_currency + Payment is NOT in Order currency
+         *    for example default = 1, order = 1, payment = 2
+         *    ==> convert payment in order's currency
+         * Order is NOT in default_currency + Payment is in Order currency
+         *    for example default = 1, order = 2, payment = 2
+         *    ==> NO conversion to do
+         * Order is NOT in default_currency + Payment is NOT in Order currency
+         *    for example default = 1, order = 2, payment = 3
+         *    ==> As conversion rates are set regarding the default currency,
+         *        convert payment to default and from default to order's currency
+         */
+
         // Update total_paid_real value for backward compatibility reasons
         if ($order_payment->id_currency == $this->id_currency) {
             $this->total_paid_real += $order_payment->amount;
         } else {
-            $this->total_paid_real += Tools::ps_round(Tools::convertPrice($order_payment->amount, $this->id_currency, false), 2);
+            $default_currency = (int) Configuration::get('PS_CURRENCY_DEFAULT');
+            if ($this->id_currency === $default_currency) {
+                $this->total_paid_real += Tools::ps_round(Tools::convertPrice($order_payment->amount, $this->id_currency, false), 2);
+            } else {
+                $amountInDefaultCurrency = Tools::convertPrice($order_payment->amount, $order_payment->id_currency, false);
+                $this->total_paid_real += Tools::ps_round(Tools::convertPrice($amountInDefaultCurrency, $this->id_currency, true), 2);
+            }
         }
 
         // We put autodate parameter of add method to true if date_add field is null
--- a/tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
@@ -650,6 +650,18 @@ class OrderFeatureContext extends AbstractDomainFeatureContext
     }
 
     /**
+     * @Then order :orderReference should have the following details:
+     *
+     * @param string $orderReference
+     * @param TableNode $table
+     */
+    public function queryOrderToGetTheFollowingProperties(string $orderReference, TableNode $table)
+    {
+        $orderId = SharedStorage::getStorage()->get($orderReference);
+        $this->assertOrderPropertiesEquals(new Order($orderId), $table->getRowsHash());
+    }
+
+    /**
      * @param string $productName
      *
      * @return int
@@ -695,5 +707,25 @@ class OrderFeatureContext extends AbstractDomainFeatureContext
         }
 
         return $productQuantities;
+    }
+
+    /**
+     * @param Order $order
+     * @param array $data
+     */
+    private function assertOrderPropertiesEquals(Order $order, array $data): void
+    {
+        foreach (array_keys($data) as $property) {
+            if (!property_exists($order, $property) || $data[$property] !== $order->$property) {
+                throw new RuntimeException(
+                    sprintf(
+                        'Expected %s value to be equal to %s, got %s instead',
+                        $property,
+                        $data[$property],
+                        $order->$property
+                    )
+                );
+            }
+        }
     }
 }
--- a/tests/Integration/Behaviour/Features/Context/Domain/OrderPaymentFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/Domain/OrderPaymentFeatureContext.php
@@ -52,13 +52,17 @@ class OrderPaymentFeatureContext extends AbstractDomainFeatureContext
 
         $data = $table->getRowsHash();
 
+        $currencyId = is_numeric($data['id_currency']) ?
+            (int) $data['id_currency'] :
+            SharedStorage::getStorage()->get($data['id_currency'])->id;
+
         $this->getCommandBus()->handle(
             new AddPaymentCommand(
                 $orderId,
                 $data['date'],
                 $data['payment_method'],
                 $data['amount'],
-                (int) $data['id_currency'],
+                $currencyId,
                 isset($data['id_invoice']) ? (int) $data['id_invoice'] : null,
                 $data['transaction_id']
             )
--- a/tests/Integration/Behaviour/Features/Context/Domain/OrderPaymentFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/Domain/OrderPaymentFeatureContext.php
@@ -52,17 +52,13 @@ class OrderPaymentFeatureContext extends AbstractDomainFeatureContext
 
         $data = $table->getRowsHash();
 
-        $currencyId = is_numeric($data['id_currency']) ?
-            (int) $data['id_currency'] :
-            SharedStorage::getStorage()->get($data['id_currency'])->id;
-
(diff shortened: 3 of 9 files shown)


classes/order/Order.php
```diff
@@ -1852,7 +1852,7 @@ public function addOrderPayment($amount_paid, $payment_method = null, $payment_t
         if ($order_payment->id_currency == $this->id_currency) {
             $this->total_paid_real += $order_payment->amount;
         } else {
-            $this->total_paid_real += Tools::ps_round(Tools::convertPrice($order_payment->amount, $order_payment->id_currency, false), 2);
+            $this->total_paid_real += Tools::ps_round(Tools::convertPrice($order_payment->amount, $this->id_currency, false), 2);
```
jolelievre: I'm not sure this is correct Because `$order_payment->id_currency` is already initialized with `($currency ? $currency->id : $this->id_currency);`
I think in your case the fix should be at the moment this method is called, not changing this method behaviour as it could cause some BC breaks

jolelievre (changes requested):
I'm not sure it's the right place to fix this

classes/order/Order.php
jolelievre: Ok so after digging into the code I think I finally understood your modif which seems correct

The confusion comes from the fact that OrderPayment's currency matches the `$currency` parameter (this is correct), but `total_paid_real` must be expressed as the Order's currency, hence the use of `$this->id_currency` The gymnastic is not so simple at first I think a comment describing this behavoiur would be a good thing :)

jolelievre (commented):
I validate the fix
However I would recommend adding a comment
And even more important some behat tests shoud be added, the good news is there are already some existing about this feature in `tests/Integration/Behaviour/Features/Scenario/Order/order_from_bo.feature`

So you can add a dedicated scenario that adds a payment in a currency different from the Order's
Then you should check both the order payment value AND the order total paid value

classes/order/Order.php
jolelievre: I just thought of an additional use case, what happens when the Order AND the Order payment are different from the default currency I think the conversion rate won't match

The amount should first be converted back to the default currency value, and then converted into the third currency value

jolelievre: Unless the Order->total_paid_real value is stored as the default value? (as the third parameter `$to_currency` suggests
Then it should be correct, but out of safety we should have a behat test to valide this:
- create an order with currency different from the default one
- pay with a currency different from the default one
- check that everything is correctly converted into the default currency

jolelievre: But since Order has a currency_id as well, I would tend to think that its value is also expressed as this currency value, not the default one

jolelievre:
```suggestion
            // Convert payment value as default currency
            $paymentDefaultValue = Tools::ps_round(Tools::convertPrice($order_payment->amount, $order_payment->id_currency, false), 2);
            // Then convert into the order currency's value with initially saved conversion rate
            $paymentOrderValue = $paymentDefaultValue * $this->conversion_rate;
            $this->total_paid_real += $paymentOrderValue;
```

I also noticed that conversion rate is saved when the order is created, I'm not sure if we should use the rate at the moment the order was created or when it was paid..

sowbiba: > I also noticed that conversion rate is saved when the order is created, I'm not sure if we should use the rate at the moment the order was created or when it was paid..

the conversion rate is the rate between order's currency and the default currency. But the default currency can be changed, so we cannot use this conversion rate
Validated with @MatShir

sowbiba: > Unless the Order->total_paid_real value is stored as the default value? (as the third parameter `$to_currency` suggests
> Then it should be correct, but out of safety we should have a behat test to valide this:
> 
> * create an order with currency different from the default one
> * pay with a currency different from the default one
> * check that everything is correctly converted into the default currency

I'm creating the behat test but the conversion made is not from "payment currency" to "default currency", but from "payment currency" to "order currency". So whatever the default currency is, I think it will be ok

But you're right to ask "what about the conversion_rate ?" My opinion is we don't need to change it because order's currency doesn't change, default currency doesn't change.


==> I will check at invoices and refunds if everything is OK

sowbiba: @jolelievre You was absolutely right. We need to convert to default currency before converting to the order one, because of the logic of convertPrice and how the conversion rate is set to the currencies

And that make me understand why you said that changing a shop default currency is not a good idea

classes/order/Order.php
```diff
@@ -1848,11 +1848,35 @@ public function addOrderPayment($amount_paid, $payment_method = null, $payment_t
...
+         */
+
         // Update total_paid_real value for backward compatibility reasons
         if ($order_payment->id_currency == $this->id_currency) {
             $this->total_paid_real += $order_payment->amount;
         } else {
-            $this->total_paid_real += Tools::ps_round(Tools::convertPrice($order_payment->amount, $order_payment->id_currency, false), 2);
+            $default_currency = (int) Configuration::get('PS_CURRENCY_DEFAULT');
+            if ($this->id_currency === $default_currency) {
+                $this->total_paid_real += Tools::ps_round(Tools::convertPrice($order_payment->amount, $this->id_currency, false), 2);
```
Progi1984, line 1875: The precision should not be to the default currency precision ?

jolelievre: Indeed the precision should be related to the currency However I don't think we need to deal with it in this PR's scope
Because there are actually MANY places where precision is not managed correctly, a PR should handle all these rounding at once

But it's worth question in however, in case of conversion which currency's precision should we use? I'd say it should be the "destination" precision And the rounding should only be performed in the last operation
Because if we perform several conversion we might lose some precision during the different computing (especially if one of the currency has a 0 precision..)

classes/order/Order.php
```diff
@@ -1848,11 +1848,35 @@ public function addOrderPayment($amount_paid, $payment_method = null, $payment_t
...
+            } else {
+                $amountInDefaultCurrency = Tools::convertPrice($order_payment->amount, $order_payment->id_currency, false);
```
jolelievre, line 1877: Careful with the third parameter `$to_currency` In first conversion it should be `false` (converting **from** currency to default one) But in the second call it must be true (converting from default currency **to** currency)

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -577,4 +592,15 @@ private function getProductQuantitiesByReference(string $orderReference, string
 
         return $productQuantities;
     }
+
+    private function assertOrderPropertiesEquals(Order $order, array $data): bool
+    {
+        foreach (array_keys($data) as $property) {
+            if (!property_exists($order, $property) || $data[$property] !== $order->$property) {
+                return false;
```
jolelievre: Instead of returning tru, and using `Assert::assertTrue` Maybe you can throw a `RuntimeException` with an explicit message? (like which field is incorrect, expected value and found result...)
This will make debugging easier when something wrongs happens, instead of one boolean for the whole object

Progi1984 (dismissed):

Progi1984:
Hello @sowbiba , thanks for your contribution, would you mind to rebase your pull request? Thanks

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
sowbiba: I moved the behat test to https://github.com/sowbiba/PrestaShop/pull/6 to unblock 1.7.7 release

tests/Integration/Behaviour/Features/Context/Domain/OrderPaymentFeatureContext.php
```diff
@@ -52,13 +52,17 @@ public function addPaymentToOrderWithTheFollowingDetails(string $orderReference,
 
         $data = $table->getRowsHash();
 
+        $currencyId = is_numeric($data['id_currency']) ?
```
jolelievre: I don't think we should use IDs here to fetch the currency, as ID are not set in stone and can change depending on how we will build the fixtures in the future But it's not to be fixed now anyway, let's see this matter in the PR that will fix the tests

tests/Integration/Behaviour/behat.yml
```diff
@@ -73,8 +74,12 @@ default:
                 - Tests\Integration\Behaviour\Features\Context\Domain\CustomerFeatureContext
                 - Tests\Integration\Behaviour\Features\Context\Domain\AddressFeatureContext
                 - Tests\Integration\Behaviour\Features\Context\Domain\OrderRefundFeatureContext
-                - Tests\Integration\Behaviour\Features\Context\Domain\TaxFeatureContext
-                - Tests\Integration\Behaviour\Features\Context\Domain\ProductFeatureContext
+                    - Tests\Integration\Behaviour\Features\Context\Domain\TaxFeatureContext
```
jolelievre: Careful with the indentation, although if the tests are green this can be fixed later

jolelievre (dismissed):

tests/Integration/Behaviour/behat.yml
sowbiba: done ;)

tests/Integration/Behaviour/Features/Context/Domain/OrderPaymentFeatureContext.php
sowbiba: fixed ;)

jolelievre (dismissed):

jolelievre (approved):

matthieu-rolland (approved):

tests/Integration/Behaviour/Features/Scenario/Order/order_add_payment.feature
```diff
@@ -0,0 +1,136 @@
...
+      | amount         | 6.00                |
+    Then order "bo_order1" payments should have the following details:
+      | date           | 2019-11-26 13:56:23 |
+      | payment_method | Payments by check   |
+      | transaction_id | test123             |
+      | amount         | €6.00               |
+    And order "bo_order1" should have the following details:
+      | total_paid_real           | 6.820000 |
+
+#  Scenario: Add a payment when Order is NOT in default_currency and Payment is in Order currency
```
PierreRambaud, line 94: Are these commented scenarios useful? :thinking:

PierreRambaud (commented):
Just one comment, I dislike useless comment :sweat_smile:

tests/Integration/Behaviour/Features/Scenario/Order/order_add_payment.feature
sowbiba: They are commented in order to fix them for the next release or if we have time to fix after the Beta

Robin-Fischer-PS:
Thanks @sowbiba ! It's OK, I tested with multiple currencies and payment.

Progi1984:
Thanks @sowbiba

PR #17975, opened 2020-03-04
Convert arabic/persian digits to latin digits in DatePickerType

### LINKED ISSUE #17916: Can't add a payment in migrated view order page when using an RTL language
#### Describe the bug

Can't add a pyament in new order page when using an RTL language because of the date

#### Steps to Reproduce

1. Install Arabic
2. Open an order with the migrated page
3. Try to add a payment

**Screenshots**
capture d'écran_2410

capture d'écran_2409

It was ok in legacy

capture d'écran_2411

#### Additional information

* PrestaShop version: 1.7.7.x

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When using arabic language in the BO, you weren't able to add a payment in the order page. This PR fixes it.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17916
| How to test?  | 1. Install & set Arabic as language in the BO<br>2. Open an order with the migrated page<br>3. Try to add a payment<br>4. The date & time should be displayed in arabic in the field and when adding the payment, no error should happend.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17975)

### DIFF
--- a/admin-dev/themes/default/js/bundle/product/form.js
+++ b/admin-dev/themes/default/js/bundle/product/form.js
@@ -63,12 +63,6 @@ $(document).ready(function() {
     $(this).val(parsedValue);
   });
 
-  /** Attach date picker */
-  $('.datepicker').datetimepicker({
-    locale: full_language_code,
-    format: 'YYYY-MM-DD'
-  });
-
   /** tooltips should be hidden when we move to another tab */
   $('#form-nav').on('click','.nav-item', function clearTooltipsAndPopovers() {
     $('[data-toggle="tooltip"]').tooltip('hide');
--- a/src/Adapter/Product/AdminProductWrapper.php
+++ b/src/Adapter/Product/AdminProductWrapper.php
@@ -79,17 +79,23 @@ class AdminProductWrapper
     private $employeeAssociatedShops;
 
     /**
+     * @var FloatParser
+     */
+    private $floatParser;
+
+    /**
      * Constructor : Inject Symfony\Component\Translation Translator.
      *
      * @param object $translator
      * @param array $employeeAssociatedShops
      * @param Locale $locale
      */
-    public function __construct($translator, array $employeeAssociatedShops, Locale $locale)
+    public function __construct($translator, array $employeeAssociatedShops, Locale $locale, FloatParser $floatParser)
     {
         $this->translator = $translator;
         $this->employeeAssociatedShops = $employeeAssociatedShops;
         $this->locale = $locale;
+        $this->floatParser = $floatParser;
     }
 
     /**
@@ -260,8 +266,6 @@ class AdminProductWrapper
      */
     public function processProductSpecificPrice($id_product, $specificPriceValues, $idSpecificPrice = null)
     {
-        $floatParser = new FloatParser();
-
         // ---- data formatting ----
         $id_product_attribute = $specificPriceValues['sp_id_product_attribute'];
         $id_shop = $specificPriceValues['sp_id_shop'] ? $specificPriceValues['sp_id_shop'] : 0;
@@ -269,9 +273,9 @@ class AdminProductWrapper
         $id_country = $specificPriceValues['sp_id_country'] ? $specificPriceValues['sp_id_country'] : 0;
         $id_group = $specificPriceValues['sp_id_group'] ? $specificPriceValues['sp_id_group'] : 0;
         $id_customer = !empty($specificPriceValues['sp_id_customer']['data']) ? $specificPriceValues['sp_id_customer']['data'][0] : 0;
-        $price = isset($specificPriceValues['leave_bprice']) ? '-1' : $floatParser->fromString($specificPriceValues['sp_price']);
+        $price = isset($specificPriceValues['leave_bprice']) ? '-1' : $this->floatParser->fromString($specificPriceValues['sp_price']);
         $from_quantity = $specificPriceValues['sp_from_quantity'];
-        $reduction = $floatParser->fromString($specificPriceValues['sp_reduction']);
+        $reduction = $this->floatParser->fromString($specificPriceValues['sp_reduction']);
         $reduction_tax = $specificPriceValues['sp_reduction_tax'];
         $reduction_type = !$reduction ? 'amount' : $specificPriceValues['sp_reduction_type'];
         $reduction_type = $reduction_type == '-' ? 'amount' : $reduction_type;
--- a/src/Adapter/Product/AdminProductWrapper.php
+++ b/src/Adapter/Product/AdminProductWrapper.php
@@ -89,13 +89,14 @@ class AdminProductWrapper
      * @param object $translator
      * @param array $employeeAssociatedShops
      * @param Locale $locale
+     * @param FloatParser|null $floatParser
      */
-    public function __construct($translator, array $employeeAssociatedShops, Locale $locale, FloatParser $floatParser)
+    public function __construct($translator, array $employeeAssociatedShops, Locale $locale, FloatParser $floatParser = null)
     {
         $this->translator = $translator;
         $this->employeeAssociatedShops = $employeeAssociatedShops;
         $this->locale = $locale;
-        $this->floatParser = $floatParser;
+        $this->floatParser = $floatParser ?? new FloatParser();
     }
 
     /**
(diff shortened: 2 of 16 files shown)


src/PrestaShopBundle/Resources/config/services/bundle/form/form_type.yml
```diff
@@ -1135,3 +1137,5 @@ services:
         - '@translator'
       tags:
         - { name: form.type }
+
+
```
Progi1984:
```suggestion
```

Progi1984 (dismissed):

Progi1984 (dismissed):

src/Core/Util/ArabicToLatinDigitConverter.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * Convert from latin digits to arabic digits
+     *
+     * @param string $str
+     *
+     * @return string
+     */
+    public function reverseConvert(string $str): string
```
jolelievre: What about reverse into persian? I know it looks the same and only UTF codes change, but still Why favor arabic over persian?

src/PrestaShopBundle/Form/Admin/Type/DatePickerType.php
```diff
@@ -45,6 +57,11 @@ public function getParent()
         return TextType::class;
     }
 
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder->addViewTransformer($this->arabicToLatinDigitDataTransformer);
```
jolelievre, line 62: One question, since the transformer converts arabic to latin, back and forth Doesn't it risk to transform latin number into arabic ones even when not needed? How is it able to detect when it must be used or not?

src/Core/Util/ArabicToLatinDigitConverter.php
atomiix: @jolelievre I had to make a choice 😄 I can separate in 2 methods (`latinToArabic()` and `latinToPersan()`) in this class and only use one in `DatePickerType` though! And because it will be replaced by javascript in the end that would be ok. WDYT?

src/PrestaShopBundle/Form/Admin/Type/DatePickerType.php
atomiix: 🤯 You are right! I'll have to find a way 😅

src/Core/Util/ArabicToLatinDigitConverter.php
```diff
@@ -0,0 +1,86 @@
...
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Util;
+
+/**
+ * Utility class to convert arabic/persian digits to latin digits
+ */
+class ArabicToLatinDigitConverter
+{
+    private static $translationTable = [
```
jolelievre: This could be a const

src/Core/Util/ArabicToLatinDigitConverter.php
jolelievre: Either two method or a parameter to chose the translation table But yeah I get that one will be automatically used in the `DatePickerType` But at least this class will be more versatile from a stand alone point of view

admin-dev/themes/default/js/bundle/product/form.js
```diff
@@ -63,12 +63,6 @@ $(document).ready(function() {
     $(this).val(parsedValue);
   });
 
-  /** Attach date picker */
```
jolelievre, line 66: Why removing this datetimepicker? Especially in the product js while the issue is related to the order view page?

atomiix: @jolelievre Because the problem is relative to all datetimepickers, not only the one in the order view page, and this one caused a bug because datetimepickers are already initialized [here](https://github.com/PrestaShop/PrestaShop/blob/1.7.7.x/admin-dev/themes/new-theme/js/app/utils/datepicker.js#L67)

src/Adapter/Product/AdminProductWrapper.php
```diff
@@ -78,18 +78,24 @@ class AdminProductWrapper
...
+
     /**
      * Constructor : Inject Symfony\Component\Translation Translator.
      *
      * @param object $translator
      * @param array $employeeAssociatedShops
      * @param Locale $locale
      */
-    public function __construct($translator, array $employeeAssociatedShops, Locale $locale)
+    public function __construct($translator, array $employeeAssociatedShops, Locale $locale, FloatParser $floatParser)
```
eternoendless: This is a breaking change, I guess you can just add a hard dependency and instantiate in the constructor.

eternoendless (dismissed):

Progi1984 (dismissed):

src/Adapter/Product/AdminProductWrapper.php
eternoendless: Oh I see why you're injecting it now

src/PrestaShopBundle/Form/Admin/Type/DatePickerType.php
jolelievre: @atomiix You didn't change anything about this issue, so how do you transform in one way only?

src/PrestaShopBundle/Form/Admin/Type/DatePickerType.php
```diff
@@ -37,6 +39,16 @@
  */
 class DatePickerType extends AbstractType
 {
+    /**
+     * @var ArabicToLatinDigitDataTransformer
+     */
+    private $arabicToLatinDigitDataTransformer;
+
+    public function __construct(ArabicToLatinDigitDataTransformer $arabicToLatinDigitDataTransformer)
```
eternoendless: Shouldn't this be a DataTransformerInterface?

```suggestion
    public function __construct(DataTransformerInterface $arabicToLatinDigitDataTransformer)
```

src/PrestaShopBundle/Form/Admin/Type/DatePickerType.php
```diff
@@ -26,8 +26,10 @@
 
 namespace PrestaShopBundle\Form\Admin\Type;
 
+use PrestaShopBundle\Form\DataTransformer\ArabicToLatinDigitDataTransformer;
```
eternoendless: Shouldn't this be a DataTransformerInterface?
```suggestion
use Symfony\Component\Form\DataTransformerInterface;
```

src/PrestaShopBundle/Form/DataTransformer/ArabicToLatinDigitDataTransformer.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $this->arabicToLatinDigitConverter = $arabicToLatinDigitConverter;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function transform($value)
+    {
+        return $value;
```
jolelievre, line 57: I the dual transformation is managed by doing nothing here (I see you don't call `$this->arabicToLatinDigitConverter->reverseConvert`) Then this needs to be thoroughly documented and unit tested to avoid anybody changing this behaviour

src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
```diff
@@ -194,7 +200,8 @@ public function __construct(
         PackDataProvider $packDataProvider,
         ShopContext $shopContext,
         TaxRuleDataProvider $taxRuleDataProvider,
-        Router $router
+        Router $router,
+        FloatParser $floatParser
```
jolelievre: Another BC left to fix here

jolelievre (dismissed):
Almost done ^^

src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
atomiix: Should be good now 😄

Progi1984 (approved):

matthieu-rolland (approved):

matks (approved):

Progi1984:
Thanks @atomiix

PR #18004, opened 2020-03-05
Replace alt image by product title if not set

Klemart3D:
| Questions     | Answers
| Branch?       | develop
| Description?  | Replace alt image by product title if not set for better SEO
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | No
| How to test?  | Check "alt" field for a product image without legend field set.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18004)

### DIFF
--- a/themes/classic/templates/catalog/_partials/product-cover-thumbnails.tpl
+++ b/themes/classic/templates/catalog/_partials/product-cover-thumbnails.tpl
@@ -26,7 +26,7 @@
   {block name='product_cover'}
     <div class="product-cover">
       {if $product.cover}
-        <img class="js-qv-product-cover" src="{$product.cover.bySize.large_default.url}" alt="{$product.cover.legend}" title="{$product.cover.legend}" style="width:100%;" itemprop="image">
+        <img class="js-qv-product-cover" src="{$product.cover.bySize.large_default.url}" alt="{($product.cover.legend != '') ? $product.cover.legend : $product.name}" title="{$product.cover.legend}" style="width:100%;" itemprop="image">
         <div class="layer hidden-sm-down" data-toggle="modal" data-target="#product-modal">
           <i class="material-icons zoom-in">search</i>
         </div>
--- a/themes/classic/templates/catalog/_partials/product-cover-thumbnails.tpl
+++ b/themes/classic/templates/catalog/_partials/product-cover-thumbnails.tpl
@@ -26,7 +26,7 @@
   {block name='product_cover'}
     <div class="product-cover">
       {if $product.cover}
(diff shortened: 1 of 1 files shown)


themes/classic/templates/catalog/_partials/product-cover-thumbnails.tpl
```diff
@@ -26,7 +26,7 @@
   {block name='product_cover'}
     <div class="product-cover">
       {if $product.cover}
-        <img class="js-qv-product-cover" src="{$product.cover.bySize.large_default.url}" alt="{$product.cover.legend}" title="{$product.cover.legend}" style="width:100%;" itemprop="image">
+        <img class="js-qv-product-cover" src="{$product.cover.bySize.large_default.url}" alt="{($product.cover.legend != '') ? $product.cover.legend : $product.name}" title="{$product.cover.legend}" style="width:100%;" itemprop="image">
```
arouiadib: Thank you for this PR. I totally agree that image `alt` enhance page SEO, but the question is : Do classic theme specs include such improvements ? Why not to recommend to merchant to fill the legend of cover image for better SEO instead of more code. If not, my suggestion is to avoid inline stuff, maybe better to do:
`{assign $imageAlt = $product.cover.legend != '' ? $product.cover.legend : $product.name }`
what do you think?

NeOMakinG:
```suggestion
        <img class="js-qv-product-cover" src="{$product.cover.bySize.large_default.url}" alt="{$product.cover.legend not same as '' ? $product.cover.legend : $product.name}" title="{$product.cover.legend}" style="width:100%;" itemprop="image">
```

NeOMakinG (changes requested):

HartLarsson:
what about if the title has a text like: Product "example" for demo

the "example" with the ' " ' char will be a problem if is rendered inside:

`alt="Product "example" for demo"`

matthieu-rolland (commented):
@HartLarsson is right, it could be problematic if the submitted string contains `"` , twig provides a solution for this:

`{{ variable|escape('html_attr') }}`

NeOMakinG:
It looks like this has been already fixed on develop, wdyt @PrestaShop/prestashop-core-developers ?

matks:
Indeed https://github.com/PrestaShop/PrestaShop/blob/develop/themes/classic/templates/catalog/_partials/product-cover-thumbnails.tpl

Sorry @Klemart3D 😞 we handled poorly this PR. Thank you for the contribution though.

PR #18009, opened 2020-03-05
Change and rename order view page hooks

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Change and rename order view page hooks, reorder upgrade scripts request to group them by scope
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | To validate all the hooks you will need the order example module https://github.com/PrestaShop/example-modules/tree/master/demovieworderhooks But it needs to integrate these modifications first

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18009)

### DIFF
--- a/install-dev/data/xml/hook.xml
+++ b/install-dev/data/xml/hook.xml
@@ -141,6 +141,11 @@
       <title>Product footer</title>
       <description>This hook adds new blocks under the product's description</description>
     </hook>
+    <hook id="displayFooterCategory">
+      <name>displayFooterCategory</name>
+      <title>Category footer</title>
+      <description>This hook adds new blocks under the products listing in a category/search</description>
+    </hook>
     <hook id="displayInvoice">
       <name>displayInvoice</name>
       <title>Invoice</title>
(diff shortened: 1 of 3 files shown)


jolelievre:
Note: the `displayAdminOrderMainBottom` additional hook is just a suggestion and needs to be validated
@MatShir @matks do you think it's useful? I proposed it to keep the same principle as in the side column

matks (approved):

install-dev/data/xml/hook.xml
```diff
@@ -141,6 +141,11 @@
       <title>Product footer</title>
       <description>This hook adds new blocks under the product's description</description>
     </hook>
+    <hook id="displayFooterCategory">
```
matks, line 144: This will break puppeter tests if I'm not wrong

atomiix (approved):

PierreRambaud (approved):

jolelievre:
Incorporated into https://github.com/PrestaShop/PrestaShop/pull/17734

PR #18063, opened 2020-03-09
Use localized dates for order page

### LINKED ISSUE #16226: Fix dates format in new Order page
All dates should be formatted according to logged in employee's language, currently format is hardcoded.

For this, `PrestaShopBundle\Twig\Extension\LocalizationExtension` extension was introduced which adds Twig filter `date_format_full` to ease the work. 

So, instead of this:
`{{ some_object.createdAt|date('Y-m-d H:i:s') }}`
it should become like this:
`{{ some_object.createdAt|date_format_full }}`

matks:
| Questions     | Answers
| Branch?       |  1.7.7.x
| Description?  | Use twig LocalizationExtension to print dates using localized format in Order page. Enhanced LocalizationExtension to provide also "light" format (days only, no hours/minutes/seconds). No BC break as LocalizationExtension was created for 1770 so we can still modify its signature.
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/16226
| How to test?  | Login to BO with employee using French language => all the dates displayed in Order page are displayed using French format. Modify Employee language to English and see dates displayed using US format.

Some dates might not be localized because the format was hardcoded for UI purposes. We can challenge this choice by discussing with PM and UX.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18063)

### DIFF
--- a/src/PrestaShopBundle/Resources/config/services/bundle/twig.yml
+++ b/src/PrestaShopBundle/Resources/config/services/bundle/twig.yml
@@ -110,5 +110,6 @@ services:
       class: 'PrestaShopBundle\Twig\Extension\LocalizationExtension'
       arguments:
         - '@=service("prestashop.adapter.legacy.context").getContext().language.date_format_full'
+        - '@=service("prestashop.adapter.legacy.context").getContext().language.date_format_lite'
       tags:
         - { name: twig.extension }
--- a/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/customer.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/customer.html.twig
@@ -76,7 +76,7 @@
           <p class="mb-1">
             <strong>{{ 'Account registered:'|trans({}, 'Admin.Orderscustomers.Feature') }}</strong>
           </p>
-          <p>{{ orderForViewing.customer.accountRegistrationDate|date("m/d/Y H:i:s") }}</p>
+          <p>{{ orderForViewing.customer.accountRegistrationDate|date_format_full }}</p>
         {% endif %}
       </div>
       <div class="col-md-6">
--- a/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/documents.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/documents.html.twig
@@ -48,7 +48,7 @@
     {% for document in orderForViewing.documents.documents %}
       <tr>
         <td>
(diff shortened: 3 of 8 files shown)


Progi1984 (dismissed):

MatShir:
@matks How the calendar format is chosen? CLDR?

matks:
> @matks How the calendar format is chosen? CLDR?

It's stored in database into `ps_lang` table. This table is filled when we download language packs, which I guess is defined from CLDR indeed.

MatShir:
Okay, thanks @matks

PierreRambaud (dismissed):

jolelievre (dismissed):
Great feature! I'll add it to the new features for 177

boubkerbribri:
@matks can you please resolve conflicts

matks:
> @matks can you please resolve conflicts

For this PR I can try, but for Russia/USA conflicts I'm sorry I wont be able to

Progi1984 (approved):

jolelievre (approved):

Robin-Fischer-PS:
Hello my dear friend @matks !

On the order list and order view page, it's OK.

However, on Create an order, the dates are with "Y-m-d H:i:s" format, even in US Language.

Is this OK or should it be handled by this PR ?

pr18063 create order

Thanks !

matks:
Indeed I only applied this on "View an order page" but you're right, Create an order page is relevant too

I'll do it

matks:
> However, on Create an order, the dates are with "Y-m-d H:i:s" format, even in US Language.
> 
> Is this OK or should it be handled by this PR ?

@Robin-Fischer-PS After exploring how this data is fetched, it seems that customer information is provided by a Search endpoint on customer controller and carts informations are provided by a "Get carts from customer" endpoint on customer controller.

Making sure these dates are localized is then located in a very different area of code. Is this OK if I fix this in another PR ? This will make sure I modify code only in specific locations, lowering the risk of unwanted side-effects.

Robin-Fischer-PS:
@matks It's OK for me ! Can you create the issue for this new fix ?

Thanks ! I'll validate this PR now

Progi1984:
Thanks @matks

dn1353:
@matks
Greate feature! Thank you!
Does it support JalaliDate for localized dates as well?

matks:
> @matks
> Greate feature! Thank you!
> Does it support JalaliDate for localized dates as well?

If I am not wrong, the format is being fetched from [CLDR](https://build.prestashop.com/news/cldr-composer-jshint/) database. Thus, this will depend on what CLDR says for the language you are using.

Here is the reference for Persian => http://cldr.unicode.org/translation/language-specific/persian

Issue #17890, opened 2020-02-28
Fix Serbian locale in PrestaShop


LouiseBonnard:
> message from Zoran, Serbian translator on Crowdin

#### Describe the bug

In PrestaShop 1.7.6, some code is changed and now is not possible to translate to Serbian language from BO. I found that the have added some code that says locale should be a combination of country code and language code. In Serbian case, it would be Sr-RS. That is why old Sr-CS is not working anymore. It works actually everywhere but not on BO when you want to translate anything. 

Fortunately I have fond solution for the problem and it works for me. But every time I have to install new PrestaShop I have to go to some modifications in order that everything works.

#### Expected behavior

Locale SR-CS is wrong, it should be called SR-RS (Combination of country code and language code according to some lines in PrestaShop cod). This new map on Crowdin for presta  1.7.7 is already there and it is named Sr-CS 

1.	Map should be renamed  from Sr-CS to Sr-RS 

2.	In all files in that map, on many places Sr-CS should be renamed to Sr-RS (please save all these files because I have translated everything myself and it took me so much time) 

3.	In this folder : \localization\CLDR\core\common\main there are two files:  sr_RS.xml file and another file sr.xml. In both files Sr-CS shoud be changed in Sr-RS

4.	In this folder \install\langs\sr language.xml is also one place with Sr-CS. It should be changed to Sr-RS. During the installation this will and up in database.

This is all. After these 4 steps PrestaShop could be used in Serbia without any problem.

#### One more thing

I reported long ago that even file: \js\jquery\plugins\validate\localization\ messages_sr.js is wrong. Language used in this file is Serbian Cyrillic and it should be Latinic. This is not changed yet. I have provided correct file but nothing has been done. I can provide correct file again if someone is willing to do these changes. 

#### Additional information

* PrestaShop version: 1.7.6
* PHP version: N/A

LouiseBonnard:
@PrestaShop/prestashop-core-developers, are you aware of this issue?

matks:
Related to https://github.com/jquery-validation/jquery-validation/issues/2244

LouiseBonnard:
> answer from the core team

So this issue comes from jQuery plugin for translation which does not handle properly Serbian language. If you really want this issue to be fixed you have to consider other solutions:

1. Stop using jQuery Validation in PrestaShop (drop this piece of software) and use a replacement that handles Serbia language correctly.

[...]

PR #11321, opened 2018-11-08
Translation Manager of Back Office can now manage legacy translations of modules

mickaelandrieu:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Until then, the support of legacy translation files was not supported by PrestaShop. To help people migrate to the new system (xliff files) we make the new translations manager compatible with the old format. 
| Type?         | new feature
| Category?     | CO
| BC breaks?    | hmm... probably some BC in the translations manager but this page is not hookable at all, so... who cares? :angel: 
| Deprecations? | yes
| Issue? | https://github.com/PrestaShop/PrestaShop/issues/11841
| How to test?  | Modules with legacy translations files should be translatable in the back office, take a look at the module `ps_themecusto` in both "french" and "english" languages.

### Legacy file support

- [X] legacy files are extractable as a Message Catalogue; ("READ")
- [X] we can save translations in the database; ("WRITE")
- [x] we have access to the "default" catalog (ie, translations from the module contents); ("REFERENCE")
- [x] We retrieve all values from the default catalog (ie the LegacyModuleExtractor is incomplete);
- [X] misses and completions works as expected; ("TREE")

state

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/11321)

PR #18064, opened 2020-03-09
Fix error when trying to translate Serbian using the BO interface

### LINKED ISSUE #18062: Cannot translate Serbian in 1.7.6.0
#### Describe the bug

Cannot access the translation interface for the Serbian language:
Screencap 1.7.6

#### Expected behavior

You should be able to access like in < 1.7.5

Screencap 1.7.5

#### Steps to Reproduce

1. Install serbian
2. Go to International > Translations
3. Try to translate BO wordings in Serbian

#### Additional information

* PrestaShop version: 1.7.6.0 and later
* PHP version: Indifferent

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | This change makes it so that it is possible to translate Serbian again using the BO interface. Read below for more.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixed #18062
| How to test?  | See steps in issue.

## More information

In 1.7.6, a [change was introduced](https://github.com/PrestaShop/PrestaShop/pull/11321/files#diff-f0c2048d6122a903ce9515acfd72098dR74-R77) where the locale code provided to the controller in charge of listing the translations to display would be verified against the server's ICU list instead of the list of installed languages. 

This was a double mistake because not only would it let a non-installed locale pass through (it would throw an unhandled exception later in code resulting in a 500 error), but it would also invalidate nonstandard locale codes like `sr-CS` (which we use for Serbian for [some unknown reason](https://github.com/PrestaShop/PrestaShop/issues/17890#issuecomment-596583744)).

This PR changes this verification so that the locale code is compared against the list of installed locales.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18064)

### DIFF
--- a/src/PrestaShopBundle/Exception/InvalidLanguageException.php
+++ b/src/PrestaShopBundle/Exception/InvalidLanguageException.php
@@ -33,6 +33,6 @@ class InvalidLanguageException extends \RuntimeException
 
     public static function localeNotFound($locale)
     {
-        return new self(sprintf('The locale "%s" is not available', $locale), self::LOCALE_NOT_FOUND);
+        return new static(sprintf('The locale "%s" is not available', $locale), self::LOCALE_NOT_FOUND);
     }
 }
--- a/src/PrestaShopBundle/Exception/InvalidLanguageException.php
+++ b/src/PrestaShopBundle/Exception/InvalidLanguageException.php
@@ -1,0 +1,38 @@
+<?php
+
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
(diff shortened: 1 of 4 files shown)


matks:
Some unit tests fail

Progi1984 (dismissed):

eternoendless:
Tests have been fixed!

Progi1984 (dismissed):

src/PrestaShopBundle/Exception/InvalidLanguageException.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace PrestaShopBundle\Exception;
+
+class InvalidLanguageException extends \RuntimeException
+{
+    const LOCALE_NOT_FOUND = 1;
+
+    public static function localeNotFound($locale)
+    {
+        return new self(sprintf('The locale "%s" is not available', $locale), self::LOCALE_NOT_FOUND);
```
PierreRambaud:
```suggestion
        return new static(sprintf('The locale "%s" is not available', $locale), self::LOCALE_NOT_FOUND);
```

PierreRambaud (dismissed):
use static instead of self

eternoendless:
Changes applied

Progi1984 (approved):

matthieu-rolland (approved):

PierreRambaud (approved):

SD1982:
This PR correctly fix #18062 thanks @eternoendless

Progi1984:
Thanks @eternoendless

PR #18082, opened 2020-03-10
Make BO menu translatable

### LINKED ISSUE #9762: [BOOM-4692] Menu translation
## Log
24/03/2021 - remaining - Fix the lastest bug from the PR

## Description

> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-4692).

- _**Reporter:**_ marion_francois
- _**Created at:**_ Thu, 25 Jan 2018 09:25:29 +0100

### Steps to reproduce
1. Install shop in French. 
2. Import EN language 
3. Set your account on it 
4. See error > some menu texts are not translated correctly ("Order Settings" instead of "Orders").

### Additional infos

- PS version: 1.7.7.0
- PHP version: 7.2

### LINKED ISSUE #9816: [BOOM-4258] On BO orders menu I have item 'discount' instead of 'credit slips'
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-4258).

- _**Reporter:**_ ldifaz
- _**Created at:**_ Tue, 7 Nov 2017 15:36:19 +0100

<p>With prestashop 1.7.2 in the BO main menu is show 'discount' (italian 'buoni sconto') in menu Orders instead of 'credit slips' (italian 'note di credito'). The traslation is correct (in the language module), and the form show the right content but also the first box show the wrong label 'discount' (italian 'buoni sconto') see the image attached.<br/>
Now my version is 1.7.2.4 and the problem is still there!<br/>
There is anyone can help me?</p>

<p>Thanks in advance. </p>
        
N/A

### LINKED ISSUE #9893: [BOOM-3832] "Stock" menu cannot be translated
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-3832).

- _**Reporter:**_ alexeven
- _**Created at:**_ Wed, 13 Sep 2017 11:53:21 +0200

<p>Hello, </p>

<p>The labels "Stocks" in the menu isn't translated.</p>

<p>A few things things:</p>

<p>1/ It should be "Stock" and not "Stocks"<br/>
2/ We have translations for "Stock" and Link using the domain Admin.Navigation.Menu.<br/>

<p>Please make sure this is translated + translatable, like the other labels from the menu.</p>

<p>Thanks,</p>
        
N/A

eternoendless:
| Questions     | Answers
| Branch?       | develop
| Description?  | Make the BO menu translatable through the translation interface and get rid of translation problems when installing other languages.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9762 & Fixes #9893
| How to test?  | See below.

## About this change

This PR introduces several changes to the way automatic database content translation works.

### Understanding multi language entities, the menu, installer fixtures and automatic translation

Multi language entities in PrestaShop manage translated content by means of a second table, called `{entity_name}_lang`, where a copy of each one of the entity's multilingual fields are stored, translated to each of the installed languages.

Elements from the back office menu are stored in database and are managed by an entity called Tab. In order to have elements from the tab menu translated, the wordings from those "tabs" are multilingual.

On install, PrestaShop uses "data fixtures" (XML files containing predefined data) to populate certain tables, like shop configuration and elements of the BO menu, with default data. As a side note, this system is used for demo products as well.

During this process, wordings for multilingual entities are inserted on the database in the language chosen during install, or in English if translated fixtures for said data are not available (this is the case for Tab, which only has fixtures in English).

But as you have probably noticed, elements in the menu appear translated after you finish the translation in another language, like French. How does this happen?

PrestaShop has an _"automagic translation"_ feature for most multilingual entities. `Language::updateMultilangFromClass()` will loop through all tables finishing in `_lang`, and will try to translate their content using the appropriate class for that entity that extends `DataLang` (like `TabLang`, `ConfigurationLang`, `OrderStateLang`, and so forth) and is located in `classes/lang/`. 

Content is translated by performing a technique accurately described by @tomlev as _"untranslating and re-translating"_. It consists in taking the already translated content in the `lang` table, finding the original translation key by performing a reverse lookup in the translator component's dictionary, then using that key to translate the content into another language. 

This process is triggered during the install and whenever a new language is installed.

During install, Tab content is initially in English, regardless of your language of choice, because of fixtures. If you are installing in, say, French, the translator will contain a dictionary with pairs like `"Some Wording" => "Le wording en français"`. When the install process triggers this translation, it will find wordings like _"Some Wording"_ and will try to untranslate them. Since it won't find them, it will fall back to the original key. Then, it will perform a translation, resulting in _"Le wording en français"_, which is the value it will save.

If later on you install another language, like Spanish, then your dictionary will have pairs like `"Some Wording" => "El wording en español"`. Since content is saved in French, it will find values like _"Le wording en français"_ which results in _"Some Wording"_ when performing a reverse translation using the French dictionary. Using that key, we can translate the wording again using the Spanish dictionary, and we get _"El wording en español"_, which is saved in database.

### The problems

This sounds like a clever hack, and in most cases it works fine. However, in some cases two different wordings translate to the same phrase in a particular language. In those cases, a reverse translation could yield the wrong key, resulting in an erroneous translation to the target language.

The second problem appears in the specific case of the menu. Default menu entries are  supposed to be translatable via the translation interface in the Back Office (wordings are placed in the `Admin.Navigation.Menu` translation domain). However, translations modified through this interface are not reflected in the menu. This is because the automatic translation is not triggered after editing a translation, but even if it was, it would never work: the dictionary we're using to _"untranslate and retranslate"_ is the same we're updating. If the dictionary was updated before retranslation process, then the reverse lookup wouldn't work (because the menu name stored in database would be different from the one in the dictionary and therefore never found), and if the dictionary was updated after the retranslation process, then the previous value would be inserted in database, resulting in no change.

### The solutions

In my view, multilanguage data should be classified in two distinct groups. In the first one, wordings are stored as translation key and translation domain pairs, and are translated on the fly, exclusively via the translator component. In the second one, content is translated exclusively by the user via form (like, for example, a product's description). Mixing the two by performing untranslation and retranslation is an absurdity that results in too much complexity, which is at best unreliable.

This means that content which is not directly user-editable, like menu names, should always be translated using the translator, and never stored translated.

Ideally, this could be solved simply by adding a "wording" and "wording_domain" to each element in `tab`, and getting rid of `tab_lang` altogether. After that, it would be up to the presentation layer to get these properties and translate them on the fly using the translator. However, this would introduce a breaking change, since modules rely on the "name" property being stored translated. Therefore, this required a more complex solution.

In this change, I introduced the aforementioned columns, but kept the automatic translation feature so that the `name` column is still stored translated in `tab_lang` just as before. However, I added a specific behavior for Tab so that when the automatic translation is triggered, instead of performing a reverse translation lookup, it uses the provided wording and wording domain as basis for the retranslation. In order to provide zero-effort retrocompatibility, Tabs with no `wording` and `wording_domain` are translated using reverse lookup, just like before.

In addition, I refactored the retranslation feature a little so that I could trigger it on a single entity. This allows me to trigger a retranslation of Tabs whenever wordings from the menu domain (`Admin.Navigation.Menu`) are updated in the translation interface. 

Modules can also benefit from this system: they just need to fill out the `wording` and `wording_domain` fields on the tabs they create. If the wording exists in the catalogue, it will be translated. However, translating it from the back office interface won't trigger a retranslation unless it's attached to the `Admin.Navigation.Menu` domain. A future improvement could add a hook so that modules can add elements to the list of domains that trigger this.

**Updates on November 9, 2020:**

- Reworked how language packs are set up during the install 
  ⚠️ **WARNING:** This also includes that language packs are no longer downloaded again during install if already present in the `/translations` folder
- Removed a reference to "safe mode" during install, which no longer exists

**As noted on [April 8, 2021](https://github.com/PrestaShop/PrestaShop/pull/18082#issuecomment-815829907):**

Modules [need to be updated](https://devdocs.prestashop.com/1.7/modules/concepts/controllers/admin-controllers/tabs/#since-prestashop-178) to allow their tabs to be translated automatically when installing a new language.

## How to test

1. Install PrestaShop in French. Menu and other auto translatable things like order states should be in French.
2. Install PrestaShop in English, then add French. Menu and other auto translatable things like order states should be in French or English depending on your choice of language.
3. Edit a menu item in the BO translation page. After refreshing the page, the menu item should appear as translated.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18082)

### DIFF
--- a/classes/Language.php
+++ b/classes/Language.php
@@ -23,7 +23,7 @@
  * @copyright Since 2007 PrestaShop SA and Contributors
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  */
-use PrestaShop\PrestaShop\Adapter\EntityTranslation\DataLangClassNameNotFoundException;
+use PrestaShop\PrestaShop\Adapter\EntityTranslation\Exception\DataLangClassNameNotFoundException;
 use PrestaShop\PrestaShop\Adapter\EntityTranslation\DataLangFactory;
 use PrestaShop\PrestaShop\Adapter\EntityTranslation\EntityTranslatorFactory;
 use PrestaShop\PrestaShop\Adapter\Language\LanguageImageManager;
--- a/classes/Language.php
+++ b/classes/Language.php
@@ -23,6 +23,9 @@
  * @copyright Since 2007 PrestaShop SA and Contributors
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  */
+use PrestaShop\PrestaShop\Adapter\EntityTranslation\DataLangClassNameNotFoundException;
+use PrestaShop\PrestaShop\Adapter\EntityTranslation\DataLangFactory;
+use PrestaShop\PrestaShop\Adapter\EntityTranslation\EntityTranslatorFactory;
 use PrestaShop\PrestaShop\Adapter\Language\LanguageImageManager;
 use PrestaShop\PrestaShop\Adapter\SymfonyContainer;
 use PrestaShop\PrestaShop\Core\Addon\Theme\ThemeManagerBuilder;
@@ -1427,8 +1430,11 @@ class LanguageCore extends ObjectModel implements LanguageInterface
      */
     public static function updateMultilangTables(Language $language, array $tablesToUpdate)
     {
+        $translator = SymfonyContainer::getInstance()->get('translator');
+
         foreach ($tablesToUpdate as $tableName) {
-            $className = ucfirst(Tools::toCamelCase(str_replace(_DB_PREFIX_, '', $tableName)));
+            $className = (new DataLangFactory(_DB_PREFIX_, $translator))
+                ->getClassNameFromTable($tableName);
 
             if (_DB_PREFIX_ . 'country_lang' === $tableName) {
                 self::updateMultilangFromCldr($language);
@@ -1484,12 +1490,14 @@ class LanguageCore extends ObjectModel implements LanguageInterface
      */
     public static function updateMultilangFromClass($table, $className, $lang)
     {
-        if (!class_exists($className)) {
+        $translator = SymfonyContainer::getInstance()->get('translator');
+
+        try {
+            $classObject = (new DataLangFactory(_DB_PREFIX_, $translator))
+                ->buildFromClassName($className, $lang->locale);
+        } catch (DataLangClassNameNotFoundException $e) {
             return;
         }
-
-        /** @var DataLangCore $classObject */
-        $classObject = new $className($lang->locale);
 
         $keys = $classObject->getKeys();
         $fieldsToUpdate = $classObject->getFieldsToUpdate();
@@ -1497,7 +1505,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
         if (!empty($keys) && !empty($fieldsToUpdate)) {
             $shops = Shop::getShopsCollection(false);
             foreach ($shops as $shop) {
-                static::updateMultilangFromClassForShop($table, $classObject, $lang, $shop, $keys, $fieldsToUpdate);
+                static::updateMultilangFromClassForShop($classObject, $lang, $shop);
             }
         }
     }
@@ -1505,83 +1513,26 @@ class LanguageCore extends ObjectModel implements LanguageInterface
     /**
      * untranslate then re-translate duplicated rows in tables with pattern xxx_lang.
      *
-     * @param string $tableName
-     * @param DataLang $classObject
+     * @param DataLangCore $classObject
      * @param Language $lang
      * @param Shop $shop
-     * @param array $keys
-     * @param array $fieldsToUpdate
      *
      * @throws \PrestaShopDatabaseException
+     * @throws PrestaShopException
      */
-    private static function updateMultilangFromClassForShop($tableName, $classObject, $lang, $shop, $keys, $fieldsToUpdate)
+    private static function updateMultilangFromClassForShop(DataLangCore $classObject, Language $lang, Shop $shop)
     {
-        $shopFieldExists = false;
-        $columns = Db::getInstance()->executeS('SHOW COLUMNS FROM `' . $tableName . '`');
-        foreach ($columns as $column) {
-            $fields[] = '`' . $column['Field'] . '`';
-            if ($column['Field'] == 'id_shop') {
-                $shopFieldExists = true;
-            }
+        $shopDefaultLangId = Configuration::get('PS_LANG_DEFAULT', null, $shop->id_shop_group, $shop->id);
+        $shopDefaultLanguage = new Language($shopDefaultLangId);
+
+        $translator = SymfonyContainer::getInstance()->get('translator');
+        if (!$translator->isLanguageLoaded($shopDefaultLanguage->locale)) {
+            (new TranslatorLanguageLoader(true))->loadLanguage($translator, $shopDefaultLanguage->locale);
         }
 
-        // get table data
-        $tableData = Db::getInstance()->executeS(
-            'SELECT * FROM `' . bqSQL($tableName) . '`
-            WHERE `id_lang` = "' . (int) $lang->id . '"'
-            . ($shopFieldExists ? ' AND `id_shop` = ' . (int) $shop->id : ''),
-            true,
-            false
-        );
-
-        if (!empty($tableData)) {
-            $shopDefaultLangId = Configuration::get('PS_LANG_DEFAULT', null, $shop->id_shop_group, $shop->id);
-            $shopDefaultLanguage = new Language($shopDefaultLangId);
-
-            $translator = SymfonyContainer::getInstance()->get('translator');
-            if (!$translator->isLanguageLoaded($shopDefaultLanguage->locale)) {
-                (new TranslatorLanguageLoader(true))->loadLanguage($translator, $shopDefaultLanguage->locale);
-            }
-
-            foreach ($tableData as $data) {
-                $updateWhere = '';
-                $updateField = '';
-
-                // Construct update where
-                foreach ($keys as $key) {
-                    if (!empty($updateWhere)) {
-                        $updateWhere .= ' AND ';
-                    }
-                    $updateWhere .= '`' . bqSQL($key) . '` = "' . pSQL($data[$key]) . '"';
-                }
-
-                // Construct update field
-                foreach ($fieldsToUpdate as $toUpdate) {
-                    if ('url_rewrite' === $toUpdate && self::$locale_crowdin_lang === $lang->locale) {
-                        continue;
-                    }
-
-                    $untranslated = $translator->getSourceString($data[$toUpdate], $classObject->getDomain());
-                    $translatedField = $classObject->getFieldValue($toUpdate, $untranslated);
-
-                    if (!empty($translatedField) && $translatedField != $data[$toUpdate]) {
-                        if (!empty($updateField)) {
-                            $updateField .= ' , ';
-                        }
-                        $updateField .= '`' . bqSQL($toUpdate) . '` = "' . pSQL($translatedField) . '"';
-                    }
-                }
-
-                // Update table
-                if (!empty($updateWhere) && !empty($updateField)) {
-                    $sql = 'UPDATE `' . bqSQL($tableName) . '` SET ' . $updateField . '
-                    WHERE ' . $updateWhere . ' AND `id_lang` = "' . (int) $lang->id . '"
-                    ' . ($shopFieldExists ? ' AND `id_shop` = ' . (int) $shop->id : '') . '
-                    LIMIT 1;';
-                    Db::getInstance()->execute($sql);
-                }
-            }
-        }
+        (new EntityTranslatorFactory($translator))
+            ->build($classObject)
+            ->translate($lang->id, $shop->id);
     }
 
     /**
--- a/classes/Language.php
+++ b/classes/Language.php
@@ -139,13 +139,13 @@ class LanguageCore extends ObjectModel implements LanguageInterface
 
     public static function resetCache()
     {
-        self::$_checkedLangs = null;
-        self::$_LANGUAGES = null;
-        self::$countActiveLanguages = null;
-        self::$_cache_language_installation = null;
-        self::$_cache_language_installation_by_locale = null;
-        self::$_cache_all_language_json = null;
-        self::$_cache_all_languages_iso = null;
+        static::$_checkedLangs = null;
+        static::$_LANGUAGES = null;
+        static::$countActiveLanguages = null;
+        static::$_cache_language_installation = null;
+        static::$_cache_language_installation_by_locale = null;
+        static::$_cache_all_language_json = null;
+        static::$_cache_all_languages_iso = null;
     }
 
     /**
@@ -157,9 +157,9 @@ class LanguageCore extends ObjectModel implements LanguageInterface
      */
     private static function loadAllLanguagesDetails(): array
     {
-        if (null === self::$_cache_all_languages_iso) {
+        if (null === static::$_cache_all_languages_iso) {
             $allLanguages = file_get_contents(_PS_ROOT_DIR_ . self::ALL_LANGUAGES_FILE);
-            self::$_cache_all_languages_iso = json_decode($allLanguages, true);
+            static::$_cache_all_languages_iso = json_decode($allLanguages, true);
 
             if (JSON_ERROR_NONE !== json_last_error()) {
                 throw new RuntimeException(
@@ -168,7 +168,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
             }
         }
 
-        return self::$_cache_all_languages_iso;
+        return static::$_cache_all_languages_iso;
     }
 
     /**
@@ -244,7 +244,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
         }
 
         if ($this->is_rtl) {
-            self::getRtlStylesheetProcessor()
+            static::getRtlStylesheetProcessor()
                 ->setIsInstall(defined('PS_INSTALLATION_IN_PROGRESS'))
                 ->setProcessBOTheme(true)
                 ->setProcessDefaultModules(true)
@@ -272,7 +272,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
 
         // Generate RTL stylesheets if language is_rtl parameter changes
         if ($this->is_rtl) {
-            self::getRtlStylesheetProcessor()
+            static::getRtlStylesheetProcessor()
                 ->setProcessBOTheme(true)
                 ->setProcessDefaultModules(true)
                 ->process();
@@ -304,7 +304,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
      */
     public static function checkFilesWithIsoCode($iso_code)
     {
-        if (isset(self::$_checkedLangs[$iso_code]) && self::$_checkedLangs[$iso_code]) {
+        if (isset(static::$_checkedLangs[$iso_code]) && static::$_checkedLangs[$iso_code]) {
             return true;
         }
 
@@ -313,7 +313,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
                 return false;
             }
         }
-        self::$_checkedLangs[$iso_code] = true;
+        static::$_checkedLangs[$iso_code] = true;
 
         return true;
     }
@@ -693,13 +693,13 @@ class LanguageCore extends ObjectModel implements LanguageInterface
      */
     public static function getLanguages($active = true, $id_shop = false, $ids_only = false)
     {
-        if (!self::$_LANGUAGES) {
+        if (!static::$_LANGUAGES) {
             Language::loadLanguages();
         }
 
         $languages = [];
-        foreach (self::$_LANGUAGES as $language) {
-            if ($active && !$language['active'] || ($id_shop && !isset($language['shops'][(int) $id_shop])) || self::$locale_crowdin_lang === $language['locale']) {
+        foreach (static::$_LANGUAGES as $language) {
+            if ($active && !$language['active'] || ($id_shop && !isset($language['shops'][(int) $id_shop])) || static::$locale_crowdin_lang === $language['locale']) {
                 continue;
             }
 
@@ -719,7 +719,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
      */
     public static function getIDs($active = true, $id_shop = false)
     {
-        return self::getLanguages($active, $id_shop, true);
+        return static::getLanguages($active, $id_shop, true);
     }
 
     /**
@@ -731,14 +731,14 @@ class LanguageCore extends ObjectModel implements LanguageInterface
      */
     public static function getLanguage($id_lang)
     {
-        if (!self::$_LANGUAGES) {
+        if (!static::$_LANGUAGES) {
             Language::loadLanguages();
         }
-        if (!array_key_exists((int) $id_lang, self::$_LANGUAGES)) {
+        if (!array_key_exists((int) $id_lang, static::$_LANGUAGES)) {
             return false;
         }
 
-        return self::$_LANGUAGES[(int) ($id_lang)];
+        return static::$_LANGUAGES[(int) ($id_lang)];
     }
 
     /**
@@ -750,11 +750,11 @@ class LanguageCore extends ObjectModel implements LanguageInterface
      */
     public static function getIsoById($id_lang)
     {
-        if (!self::$_LANGUAGES) {
+        if (!static::$_LANGUAGES) {
             Language::loadLanguages();
         }
-        if (isset(self::$_LANGUAGES[(int) $id_lang]['iso_code'])) {
-            return self::$_LANGUAGES[(int) $id_lang]['iso_code'];
+        if (isset(static::$_LANGUAGES[(int) $id_lang]['iso_code'])) {
+            return static::$_LANGUAGES[(int) $id_lang]['iso_code'];
         }
 
         return false;
@@ -791,16 +791,16 @@ class LanguageCore extends ObjectModel implements LanguageInterface
      */
     public static function getJsonLanguageDetails($locale)
     {
-        if (self::$_cache_all_language_json === null) {
-            self::$_cache_all_language_json = [];
-            $allLanguages = self::loadAllLanguagesDetails();
+        if (static::$_cache_all_language_json === null) {
+            static::$_cache_all_language_json = [];
+            $allLanguages = static::loadAllLanguagesDetails();
 
             foreach ($allLanguages as $isoCode => $langDetails) {
-                self::$_cache_all_language_json[$langDetails['locale']] = $langDetails;
+                static::$_cache_all_language_json[$langDetails['locale']] = $langDetails;
             }
         }
 
-        return isset(self::$_cache_all_language_json[$locale]) ? self::$_cache_all_language_json[$locale] : false;
+        return isset(static::$_cache_all_language_json[$locale]) ? static::$_cache_all_language_json[$locale] : false;
     }
 
     /**
@@ -872,7 +872,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
     {
         $iso = (string) $iso; // $iso often comes from xml and is a SimpleXMLElement
 
-        $allLanguages = self::loadAllLanguagesDetails();
+        $allLanguages = static::loadAllLanguagesDetails();
 
         return isset($allLanguages[$iso]) ? $allLanguages[$iso] : false;
     }
@@ -892,7 +892,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
             throw new Exception('The ISO code ' . $isoCode . ' is invalid');
         }
 
-        if ($details = self::getLangDetails($isoCode)) {
+        if ($details = static::getLangDetails($isoCode)) {
             return $details['locale'];
         }
 
@@ -914,7 +914,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
             throw new Exception('The locale ' . $locale . ' is invalid');
         }
 
-        if ($details = self::getJsonLanguageDetails($locale)) {
+        if ($details = static::getJsonLanguageDetails($locale)) {
             return $details['iso_code'];
         }
 
@@ -1038,7 +1038,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
      */
     public static function loadLanguages()
     {
-        self::$_LANGUAGES = [];
+        static::$_LANGUAGES = [];
 
         $sql = 'SELECT l.*, ls.`id_shop`
 				FROM `' . _DB_PREFIX_ . 'lang` l
@@ -1053,26 +1053,26 @@ class LanguageCore extends ObjectModel implements LanguageInterface
         foreach ($result as $row) {
             $idLang = (int) $row['id_lang'];
 
-            if (!isset(self::$_LANGUAGES[$idLang])) {
-                self::$_LANGUAGES[$idLang] = $row;
+            if (!isset(static::$_LANGUAGES[$idLang])) {
+                static::$_LANGUAGES[$idLang] = $row;
             }
-            self::$_LANGUAGES[$idLang]['shops'][(int) $row['id_shop']] = true;
+            static::$_LANGUAGES[$idLang]['shops'][(int) $row['id_shop']] = true;
         }
     }
 
     public static function loadLanguagesLegacy()
     {
-        self::$_LANGUAGES = [];
+        static::$_LANGUAGES = [];
 
         $result = Db::getInstance()->executeS('SELECT * FROM `' . _DB_PREFIX_ . 'lang`');
 
         foreach ($result as $row) {
             $idLang = (int) $row['id_lang'];
 
-            if (!isset(self::$_LANGUAGES[$idLang])) {
-                self::$_LANGUAGES[$idLang] = $row;
+            if (!isset(static::$_LANGUAGES[$idLang])) {
+                static::$_LANGUAGES[$idLang] = $row;
             }
-            self::$_LANGUAGES[$idLang]['shops'][1] = true;
+            static::$_LANGUAGES[$idLang]['shops'][1] = true;
         }
     }
 
@@ -1103,7 +1103,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
 
         // If the language pack has not been provided, retrieve it from prestashop.com
         if (!$lang_pack) {
-            $lang_pack = self::getLangDetails($iso_code);
+            $lang_pack = static::getLangDetails($iso_code);
         }
 
         // If a language pack has been found or provided, prefill the language object with the value
@@ -1140,7 +1140,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
         $languageManager->setupLanguageFlag($lang->locale, $lang->id, $lang_pack['flag'] ?? null);
         $languageManager->setupDefaultImagePlaceholder($lang->iso_code);
 
-        self::loadLanguages();
+        static::loadLanguages();
 
         return true;
     }
@@ -1160,28 +1160,28 @@ class LanguageCore extends ObjectModel implements LanguageInterface
 
     public static function isInstalled($iso_code)
     {
-        if (self::$_cache_language_installation === null) {
-            self::$_cache_language_installation = [];
+        if (static::$_cache_language_installation === null) {
+            static::$_cache_language_installation = [];
             $result = Db::getInstance()->executeS('SELECT `id_lang`, `iso_code` FROM `' . _DB_PREFIX_ . 'lang`');
             foreach ($result as $row) {
-                self::$_cache_language_installation[$row['iso_code']] = $row['id_lang'];
+                static::$_cache_language_installation[$row['iso_code']] = $row['id_lang'];
             }
         }
 
-        return isset(self::$_cache_language_installation[$iso_code]) ? self::$_cache_language_installation[$iso_code] : false;
+        return isset(static::$_cache_language_installation[$iso_code]) ? static::$_cache_language_installation[$iso_code] : false;
     }
 
     public static function isInstalledByLocale($locale)
     {
-        if (self::$_cache_language_installation_by_locale === null) {
-            self::$_cache_language_installation_by_locale = [];
+        if (static::$_cache_language_installation_by_locale === null) {
+            static::$_cache_language_installation_by_locale = [];
             $result = Db::getInstance()->executeS('SELECT `id_lang`, `locale` FROM `' . _DB_PREFIX_ . 'lang`');
             foreach ($result as $row) {
-                self::$_cache_language_installation_by_locale[$row['locale']] = $row['id_lang'];
+                static::$_cache_language_installation_by_locale[$row['locale']] = $row['id_lang'];
             }
         }
 
-        return isset(self::$_cache_language_installation_by_locale[$locale]);
+        return isset(static::$_cache_language_installation_by_locale[$locale]);
     }
 
     public static function countActiveLanguages($id_shop = null)
@@ -1190,15 +1190,15 @@ class LanguageCore extends ObjectModel implements LanguageInterface
             $id_shop = (int) Context::getContext()->shop->id;
         }
 
-        if (!isset(self::$countActiveLanguages[$id_shop])) {
-            self::$countActiveLanguages[$id_shop] = Db::getInstance()->getValue('
+        if (!isset(static::$countActiveLanguages[$id_shop])) {
+            static::$countActiveLanguages[$id_shop] = Db::getInstance()->getValue('
 				SELECT COUNT(DISTINCT l.id_lang) FROM `' . _DB_PREFIX_ . 'lang` l
 				JOIN ' . _DB_PREFIX_ . 'lang_shop lang_shop ON (lang_shop.id_lang = l.id_lang AND lang_shop.id_shop = ' . (int) $id_shop . ')
 				WHERE l.`active` = 1
 			');
         }
 
-        return self::$countActiveLanguages[$id_shop];
+        return static::$countActiveLanguages[$id_shop];
     }
 
     public static function downloadAndInstallLanguagePack($iso, $version = _PS_VERSION_, $params = null, $install = true)
@@ -1224,11 +1224,11 @@ class LanguageCore extends ObjectModel implements LanguageInterface
     {
         $iso = (string) $iso; // $iso often comes from xml and is a SimpleXMLElement
 
-        $lang_pack = self::getLangDetails($iso);
+        $lang_pack = static::getLangDetails($iso);
         if (!$lang_pack) {
             $errors[] = Context::getContext()->getTranslator()->trans('Sorry this language is not available', [], 'Admin.International.Notification');
         } else {
-            self::downloadXLFLanguagePack($lang_pack['locale'], $errors, self::PACK_TYPE_SYMFONY);
+            static::downloadXLFLanguagePack($lang_pack['locale'], $errors, self::PACK_TYPE_SYMFONY);
         }
 
         return count($errors) === 0;
@@ -1236,7 +1236,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
 
     public static function downloadXLFLanguagePack($locale, &$errors = [], $type = self::PACK_TYPE_SYMFONY)
     {
-        $file = self::getPathToCachedTranslationPack($type, $locale);
+        $file = static::getPathToCachedTranslationPack($type, $locale);
         $url = (self::PACK_TYPE_EMAILS === $type) ? self::EMAILS_LANGUAGE_PACK_URL : self::SF_LANGUAGE_PACK_URL;
         $url = str_replace(
             [
@@ -1253,7 +1253,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
         if (!is_writable(dirname($file))) {
             // @todo Throw exception
             $errors[] = Context::getContext()->getTranslator()->trans('Server does not have permissions for writing.', [], 'Admin.International.Notification') . ' (' . $file . ')';
-        } elseif ($content = Tools::file_get_contents($url, false, null, self::PACK_DOWNLOAD_TIMEOUT)) {
+        } elseif ($content = Tools::file_get_contents($url, false, null, static::PACK_DOWNLOAD_TIMEOUT)) {
             file_put_contents($file, $content);
         } else {
             $errors[] = Context::getContext()->getTranslator()->trans('Language pack unavailable.', [], 'Admin.International.Notification') . ' ' . $url;
@@ -1264,12 +1264,12 @@ class LanguageCore extends ObjectModel implements LanguageInterface
 
     public static function installSfLanguagePack($locale, &$errors = [])
     {
-        if (!self::translationPackIsInCache($locale)) {
+        if (!static::translationPackIsInCache($locale)) {
             // @todo Throw exception
             $errors[] = Context::getContext()->getTranslator()->trans('Language pack unavailable.', [], 'Admin.International.Notification');
         } else {
             $zipArchive = new ZipArchive();
-            $zipArchive->open(self::getPathToCachedTranslationPack($locale));
+            $zipArchive->open(static::getPathToCachedTranslationPack($locale));
             $zipArchive->extractTo(_PS_ROOT_DIR_ . '/app/Resources/translations');
             $zipArchive->close();
         }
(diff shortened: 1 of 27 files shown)


Progi1984:
@eternoendless Your PR put Travis in the :red_circle: zone !

classes/Language.php
```diff
@@ -1342,75 +1345,70 @@ public static function getLanguagePackListContent($iso, $tar)
...
      *
-     * @param $iso_code
+     * @param string $iso_code 2-letter language code
      *
      * @return bool
      */
     public static function updateMultilangTable($iso_code)
     {
-        $useLang = Db::getInstance()->getRow('SELECT * FROM `' . _DB_PREFIX_ . 'lang` WHERE `iso_code` = "' . pSQL($iso_code) . '" ', true, false);
+        $langId = Language::getIdByIso($iso_code);
```
PierreRambaud:
```suggestion
        $langId = static::getIdByIso($iso_code);
```

classes/Language.php
```diff
@@ -1342,75 +1345,70 @@ public static function getLanguagePackListContent($iso, $tar)
...
 
-        if (!empty($useLang)) {
-            $lang = new Language($useLang['id_lang']);
+        if (!empty($langId)) {
+            $lang = new Language($langId);
```
PierreRambaud:
```suggestion
            $lang = new static($langId);
```

classes/Language.php
```diff
@@ -1342,75 +1345,70 @@ public static function getLanguagePackListContent($iso, $tar)
...
+    public static function updateMultilangTables(Language $language, array $tablesToUpdate)
+    {
+        $translator = SymfonyContainer::getInstance()->get('translator');
+
+        foreach ($tablesToUpdate as $tableName) {
+            $className = (new DataLangFactory(_DB_PREFIX_, $translator))
+                ->getClassNameFromTable($tableName);
+
+            if (_DB_PREFIX_ . 'country_lang' === $tableName) {
+                self::updateMultilangFromCldr($language);
```
PierreRambaud:
```suggestion
                static::updateMultilangFromCldr($language);
```

classes/Language.php
```diff
@@ -1342,75 +1345,70 @@ public static function getLanguagePackListContent($iso, $tar)
...
+            } else {
+                self::updateMultilangFromClass($tableName, $className, $language);
```
PierreRambaud:
```suggestion
                static::updateMultilangFromClass($tableName, $className, $language);
```

classes/Language.php
```diff
@@ -1342,75 +1345,70 @@ public static function getLanguagePackListContent($iso, $tar)
...
+            }
+        }
+
+        Hook::exec('actionUpdateLangAfter', ['lang' => $language]);
```
PierreRambaud: Maybe add a 
`Hook::exec('actionUpdateLangBefore', ['lang' => $language]);` ?

classes/Language.php
```diff
@@ -1447,43 +1445,43 @@ public static function updateMultilangFromCldr($lang)
      *
      * @param string $table
      * @param string $className
-     * @param string $lang
+     * @param Language $lang
```
PierreRambaud:
```suggestion
     * @param static $lang
```
?

classes/Tab.php
```diff
@@ -77,6 +83,8 @@ class TabCore extends ObjectModel
             'enabled' => ['type' => self::TYPE_BOOL, 'validate' => 'isBool'],
             'hide_host_mode' => ['type' => self::TYPE_BOOL, 'validate' => 'isBool'],
             'icon' => ['type' => self::TYPE_STRING, 'size' => 64],
+            'wording' => ['type' => self::TYPE_STRING, 'validate' => 'isString', 'size' => 194],
```
PierreRambaud: No need to put `194` if you don't add index on it :thinking:

classes/lang/TabLang.php
```diff
@@ -23,12 +23,30 @@
...
+    public function getFieldValue($field, $value)
+    {
+        $domain = '';
+        if (is_array($value)) {
+            list($message, $domain) = $value;
+        } else {
+            $message = $value;
+        }
+
+        return $this->translator->trans($message, [], (!empty($domain)) ? $domain : $this->domain, $this->locale);
```
PierreRambaud:
```suggestion
        return $this->translator->trans($message, [], (!empty($domain) ? $domain : $this->domain), $this->locale);
```
For a better readability

composer.json
```diff
@@ -12,6 +12,7 @@
         "ext-json": "*",
         "ext-gd": "*",
         "ext-dom": "*",
+        "ext-simplexml": "*",
```
PierreRambaud: composer.lock needs to be updated

install-dev/data/xml/tab.xml
```diff
@@ -5,450 +5,697 @@
...
 
         <tab id="Dashboard" id_parent="" active="1" enabled="1" hide_host_mode="0" icon="trending_up">
             <class_name>AdminDashboard</class_name>
+            <wording>Dashboard</wording>
+            <wording_domain>Admin.Navigation.Menu</wording_domain>
         </tab>
 
         <!--SELL-->
         <tab id="SELL" id_parent="" active="1" enabled="1" hide_host_mode="0" icon="">
             <class_name>SELL</class_name>
```
PierreRambaud, line 21: Is these  `class_name` and `id` are ok? :thinking: 
```suggestion
        <tab id="Sell" id_parent="" active="1" enabled="1" hide_host_mode="0" icon="">
            <class_name>Sell</class_name>
```

install-dev/data/xml/tab.xml
```diff
@@ -5,450 +5,697 @@
...
+                    <wording_domain>Admin.Navigation.Menu</wording_domain>
                 </tab>
 
-                    <tab id="Stock_Management" id_parent="Stock_Management" active="1" enabled="1" hide_host_mode="0" icon="">
+                    <tab id="Stock_Management" id_parent="Stock_Management_Parent" active="1" enabled="1" hide_host_mode="0" icon="">
                         <class_name>AdminStockManagement</class_name>
+                        <wording>Stock Management</wording>
+                        <wording_domain>Admin.Navigation.Menu</wording_domain>
                     </tab>
                     <tab id="Stock_Movement" id_parent="Stock_Management" active="1" enabled="1" hide_host_mode="0" icon="">
```
PierreRambaud, line 218:
```suggestion
                    <tab id="Stock_Movement" id_parent="Stock_Management_Parent" active="1" enabled="1" hide_host_mode="0" icon="">
```
?

PierreRambaud: And for others

install-dev/data/xml/tab.xml
```diff
@@ -5,450 +5,697 @@
...
                 </tab>
                 <tab id="Configuration" id_parent="Advanced_stock_management" active="1" enabled="1" hide_host_mode="0" icon="">
                     <class_name>AdminStockConfiguration</class_name>
+                    <wording>Configuration</wording>
+                    <wording_domain>Admin.Navigation.Menu</wording_domain>
                 </tab>
 
 
         <!--IMPROVE-->
         <tab id="IMPROVE" id_parent="" active="1" enabled="1" hide_host_mode="0" icon="">
```
PierreRambaud, line 247: Again, an uppercase id :thinking:

install-dev/data/xml/tab.xml
```diff
@@ -5,450 +5,697 @@
...
+                    <wording_domain>Admin.Navigation.Menu</wording_domain>
                 </tab>
                 <tab id="Mail_Theme_Parent" id_parent="Look_feel" active="1" enabled="1" hide_host_mode="0" icon="">
-                  <class_name>AdminParentMailTheme</class_name>
+                    <class_name>AdminParentMailTheme</class_name>
+                    <wording>Email Theme</wording>
+                    <wording_domain>Admin.Navigation.Menu</wording_domain>
                 </tab>
                     <tab id="Mail_Theme" id_parent="Mail_Theme_Parent" active="1" enabled="1" hide_host_mode="0" icon="">
-                      <class_name>AdminMailTheme</class_name>
```
PierreRambaud, line 211: Indentation in the whole file is totally broken :sweat_smile:

install-dev/upgrade/sql/1.7.8.0.sql
```diff
@@ -9,3 +9,6 @@ INSERT IGNORE INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `po
   (NULL, 'actionPresentProduct', 'Product Presenter', 'This hook is called before a product is presented', '1')
   (NULL, 'displayBanner', 'Display Banner', 'Use this hook for banners on top of every pages', '1')
 ;
+
+ALTER TABLE `PREFIX_tab` ADD `wording` VARCHAR(196) DEFAULT NULL AFTER `icon`;
```
PierreRambaud:
```suggestion
ALTER TABLE `PREFIX_tab` ADD `wording` VARCHAR(256) DEFAULT NULL AFTER `icon`;
```

install-dev/upgrade/sql/1.7.8.0.sql
```diff
@@ -9,3 +9,6 @@ INSERT IGNORE INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `po
...
+ALTER TABLE `PREFIX_tab` ADD `wording_domain` VARCHAR(196) DEFAULT NULL AFTER `wording`;
```
PierreRambaud:
```suggestion
ALTER TABLE `PREFIX_tab` ADD `wording_domain` VARCHAR(256) DEFAULT NULL AFTER `wording`;
```

src/Adapter/EntityTranslation/DataLangClassNameNotFoundException.php
```diff
@@ -0,0 +1,32 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to http://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Adapter\EntityTranslation;
```
PierreRambaud:
```suggestion
namespace PrestaShop\PrestaShop\Adapter\EntityTranslation\Exception;
```

Could it be in a separated directory?

src/Adapter/EntityTranslation/EntityTranslator.php
```diff
@@ -0,0 +1,223 @@
...
+
+        $shopFieldExists = $this->shopFieldExists($tableNameSql);
+
+        // get table data
+        $sql = "SELECT * FROM `$tableNameSql` WHERE `id_lang` = $languageId"
+            . ($shopFieldExists ? sprintf(' AND `id_shop` = %d', $this->shopId) : '');
+
+        $tableData = $this->db->executeS($sql, true, false);
+
+        if (!empty($tableData)) {
```
PierreRambaud:
```suggestion
        if (empty($tableData)) {
            return;
        }
```

src/Adapter/EntityTranslation/EntityTranslator.php
```diff
@@ -0,0 +1,223 @@
...
+                        $updateField[] = '`' . bqSQL($fieldName) . '` = "' . pSQL($translatedField) . '"';
+                    }
+                }
+
+                // Update table
+                if (!empty($updateWhere) && !empty($updateField)) {
+                    $updateWhere = implode(' AND ', $updateWhere);
+                    $updateField = implode(', ', $updateField);
+
+                    $sql = "UPDATE `$tableNameSql` SET $updateField
```
PierreRambaud: Prepare the string outside the loop, and use sprintf instead. 
The `$shopFieldExists` shouldn't be executed on each loop.

src/Adapter/EntityTranslation/EntityTranslator.php
```diff
@@ -0,0 +1,223 @@
...
+     * Retrieves the original wording via reverse dictionary search (aka "untranslation")
+     *
+     * @param array $data
+     * @param string $fieldName
+     *
+     * @return string
+     */
+    protected function getSourceString(array $data, string $fieldName): string
+    {
+        return $this->translator->getSourceString($data[$fieldName], $this->dataLang->getDomain());
```
PierreRambaud, line 211: Make sure `$data[$fieldName]` exists?

src/PrestaShopBundle/Entity/Tab.php
```diff
@@ -108,6 +108,20 @@ class Tab
      */
     private $icon;
 
+    /**
+     * @var string
+     *
+     * @ORM\Column(name="wording", type="string", length=194, nullable=true)
```
PierreRambaud:
```suggestion
     * @ORM\Column(name="wording", type="string", length=256, nullable=true)
```

src/PrestaShopBundle/Entity/Tab.php
```diff
@@ -108,6 +108,20 @@ class Tab
...
+     */
+    private $wording;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(name="wording_domain", type="string", length=194, nullable=true)
```
PierreRambaud:
```suggestion
     * @ORM\Column(name="wording_domain", type="string", length=256, nullable=true)
```

src/PrestaShopBundle/Install/EntityLoader/FileLoader.php
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * @return string
+     */
+    private function getFallBackToDefaultEntityLanguage($iso, $entity)
+    {
+        if ($this->getFallbackToDefaultLanguage($iso) === self::FALLBACK_LANGUAGE_CODE) {
+            return self::FALLBACK_LANGUAGE_CODE;
+        }
+
+        return file_exists($this->langPath . $this->getFallbackToDefaultLanguage($iso) . '/data/' . $entity . '.xml') ? $iso :
```
PierreRambaud: You did thee times `file_exists` here. Why not checking only:
```suggestion
        return file_exists($this->langPath . $iso . '/data/' . $entity . '.xml') ? $iso : self::FALLBACK_LANGUAGE_CODE;
```

And you can remove `getFallbackToDefaultLanguage` method, and the previous check :thinking:

marionf:
@eternoendless Does this PR will fix also #9816 and #9893 ?

eternoendless:
@marionf I think it should. However the "Link widget" may need an update in the module

classes/Language.php
eternoendless: What for?

install-dev/data/xml/tab.xml
eternoendless: I don't remember what I found out about these but since it's unrelated I'd prefer to leave it as is for now. Be aware that this tab exists as it is since 1.7.0.

install-dev/data/xml/tab.xml
eternoendless: Again, this is shadowy terrain that should definitely be explored, but not in this PR!

install-dev/data/xml/tab.xml
eternoendless: Yes because the structure makes no sense. This indentation is meant to reflect the menu structure which doesn't follow the xml structure (or is it the other way around? 🤪) 

Anyway I spent some hours trying to fix it and I found myself falling through the "refactor ALL the things!" rabbit hole, so I backtracked and decided to focus on the original intent.

src/Adapter/EntityTranslation/EntityTranslator.php
eternoendless: I'm refactoring the $shopFieldExists but leaving the string interpolation. AFAIK `sprintf` won't add anything and replacing named tokens with `%s` will decrease readability.

src/Adapter/EntityTranslation/EntityTranslator.php
eternoendless: Okay

src/PrestaShopBundle/Install/EntityLoader/FileLoader.php
eternoendless: Okay, but if for any reason this stops working you'll fix it 😛

eternoendless:
@PierreRambaud rebased and changes applied.

src/Adapter/EntityTranslation/EntityTranslator.php
```diff
@@ -0,0 +1,238 @@
...
+     * Builds the table name using the DataLang class as source
+     *
+     * @param DataLangCore $dataLang
+     *
+     * @return string The table name, including prefix
+     */
+    private function buildTableNameFromDataLang(DataLangCore $dataLang): string
+    {
+        $tableName = Inflector::tableize(get_class($dataLang));
+        if (substr($tableName, 0, strlen(_DB_PREFIX_)) !== _DB_PREFIX_) {
```
PierreRambaud: Use `construct` parameter for `dbprefix` like in `DataLangFactory`

eternoendless: Done!

PierreRambaud (dismissed):

install-dev/data/xml/tab.xml
```diff
@@ -5,450 +5,697 @@
...
                     </tab>
                     <tab id="Marketing_tools" id_parent="Module_Catalog" active="1" enabled="1" hide_host_mode="0" icon="">
                         <class_name>AdminAddonsCatalog</class_name>
+                        <wording>Module Selections</wording>
+                        <wording_domain>Admin.Navigation.Menu</wording_domain>
                     </tab>
                 <tab id="Module_Page_old" id_parent="Modules" active="0" enabled="1" hide_host_mode="0" icon="">
                     <class_name>AdminModules</class_name>
+                    <wording/>
+                    <wording_domain/>
```
Progi1984, line 297: Normal ?

src/Adapter/EntityTranslation/DataLangFactory.php
```diff
@@ -0,0 +1,142 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to http://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 28:
```suggestion
declare(strict_types=1);
```

src/Adapter/EntityTranslation/EntityTranslator.php
```diff
@@ -0,0 +1,246 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to http://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 28:
```suggestion
declare(strict_types=1);

```

src/Adapter/EntityTranslation/EntityTranslatorFactory.php
```diff
@@ -0,0 +1,107 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 28:
```suggestion
declare(strict_types=1);

```

src/Adapter/EntityTranslation/Exception/DataLangClassNameNotFoundException.php
```diff
@@ -0,0 +1,32 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to http://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 28:
```suggestion
declare(strict_types=1);

```

src/Adapter/EntityTranslation/TabTranslator.php
```diff
@@ -0,0 +1,84 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to http://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 28:
```suggestion
declare(strict_types=1);

```

src/Core/Translation/EntityTranslatorInterface.php
```diff
@@ -0,0 +1,41 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27:
```suggestion
declare(strict_types=1);

```

src/PrestaShopBundle/Install/EntityLoader/FileLoader.php
```diff
@@ -0,0 +1,164 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to http://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 28:
```suggestion
declare(strict_types=1);

```

src/PrestaShopBundle/Translation/TranslatorInterface.php
```diff
@@ -0,0 +1,59 @@
...
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShopBundle\Translation;
+
```
Progi1984, line 29:
```suggestion

use Symfony\Component\Translation\TranslatorInterface as SymfonyTranslatorInterface;
```

src/PrestaShopBundle/Translation/TranslatorInterface.php
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Interface for PrestaShop translators
+ */
+interface TranslatorInterface extends \Symfony\Component\Translation\TranslatorInterface
```
Progi1984:
```suggestion
interface TranslatorInterface extends SymfonyTranslatorInterface
```

src/PrestaShopBundle/Translation/TranslatorInterface.php
```diff
@@ -0,0 +1,59 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27:
```suggestion
declare(strict_types=1);

```

tests/Unit/Adapter/Translations/DataLangFactoryTest.php
```diff
@@ -0,0 +1,59 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27:
```suggestion
declare(strict_types=1);

```

Progi1984 (commented):
Some changes

matthieu-rolland (dismissed):

install-dev/data/xml/tab.xml
eternoendless: Yes, this tab was never translated. I don't really know why it's here, notice it's `active="0"`.

Progi1984: Ok for me ;)

src/PrestaShopBundle/Translation/TranslatorInterface.php
```diff
@@ -0,0 +1,61 @@
...
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+declare(strict_types=1);
+
+namespace PrestaShopBundle\Translation;
+
+use Symfony\Component\Translation\TranslatorInterface as SymfonyTranslatorInterface;
```
Progi1984, line 30:
```suggestion
use Symfony\Component\Translation\TranslatorInterface as SymfonyTranslatorInterface;

```

Progi1984:
@eternoendless Some PHPCS fix errors ;) and Travis is now :green_circle:

Progi1984 (dismissed):

PierreRambaud (dismissed):

SD1982:
> ## How to test
> 1. Install PrestaShop in French. Menu and other auto translatable things like order states should be in French.
> 2. Install PrestaShop in English, then add French. Menu and other auto translatable things like order states should be in French or English depending on your choice of language.
> 3. Edit a menu item in the BO translation page. After refreshing the page, the menu item should appear as translated.

Hi @eternoendless If I reproduce step 2 everything is ok but if I do it in reverse i mean install in French => add English it doesn't seem to work properly the menu stays in french
https://drive.google.com/file/d/1U49TDB1kaQChUUp-7wR9p1DA-0kLeGy4/view?usp=sharing

I also have an exception when trying to update a langage : 
https://drive.google.com/file/d/1AO7eV7S_dqvPEZFsSEp0_JlRDXMXnfZl/view?usp=sharing

src/PrestaShopBundle/Translation/DataCollectorTranslator.php
```diff
@@ -29,7 +29,23 @@
...
+    {
+        return $this->__call('isLanguageLoaded', [$locale]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function clearLanguage($locale)
+    {
+        return  $this->__call('clearLanguage', [$locale]);
```
Pierstoval: I'm not sure about these two methods. They seem to use the [base translator's __call() method](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Translation/DataCollectorTranslator.php#L113-L116) but it needs a decorated translator.

Why don't you call `$this->translator->clearLanguage($locale)`, since the decorated translator should anyway be an instance of Prestashop's one?
You could even type-hint this in the class with `/** @​property PrestashopTranslator $translator */` (of course you would need to import the namespace for auto-completion in IDEs).

eternoendless: I had forgotten about this, the answer is pretty simple: the translator property is private on the original DataCollectorTranslator, so I can't access it 😢 

That said I think I could have used `parent::isLanguageLoaded()` and it would have been routed automagically all the same. But I don't know if it would have been clearer that way either anyway 😁

Pierstoval: Calling `parent::isLanguageLoaded()` is definitely the way to go. Relying on magic methods is almost never a good option. Especially when the base method is reachable through `parent::` 🙂  (and especially in this case where it's more a hack than a true feature/pattern usage)

eternoendless:
Rebased and bugs fixed.

- Reworked how language packs are set up during the install 
  ⚠️ **WARNING:** This also includes that language packs are no longer downloaded again during install if already present in the `/translations` folder
- Removed a reference to "safe mode" during install, which no longer exists
- Shuffled code around in Language class

### It works on my computer!

- Tested install in French, adding English, switching to that language, changing menu translations, installing more languages, uninstalling languages and reinstalling them again...
- Also verified that mail templates are in the correct language, as well as currencies
- Note: I installed English through localization pack, for all other languages I used _International > Translations > Add / update a language_

classes/Language.php
```diff
@@ -39,12 +42,23 @@ class LanguageCore extends ObjectModel implements LanguageInterface
     const ALL_LANGUAGES_FILE = '/app/Resources/all_languages.json';
     const SF_LANGUAGE_PACK_URL = 'https://i18n.prestashop.com/translations/%version%/%locale%/%locale%.zip';
     const EMAILS_LANGUAGE_PACK_URL = 'https://i18n.prestashop.com/mails/%version%/%locale%/%locale%.zip';
+    const PACK_TYPE_EMAILS = 'emails';
+    const PACK_TYPE_SYMFONY = 'sf';
```
Progi1984:
```suggestion
    public const PACK_TYPE_EMAILS = 'emails';
    public const PACK_TYPE_SYMFONY = 'sf';
```

src/PrestaShopBundle/Translation/DataCollectorTranslator.php
eternoendless: I had to revert to this because PHPStan doesn't support calls to `parent` if the method is not declared  in the parent and routed with `__call()` ☹️ 

See https://github.com/PrestaShop/PrestaShop/pull/18082/checks?check_run_id=1375456805

classes/Language.php
```diff
@@ -1256,31 +1335,87 @@ public static function installEmailsLanguagePack($lang_pack, &$errors = [])
...
+     * @param string $iso Language ISO code
+     * @param array $params Optional parameters for self::checkAndAddLanguage
+     * @param array $errors
+     *
+     * @return bool
+     *
+     * @throws PrestaShopDatabaseException
+     * @throws PrestaShopException
+     */
+    public static function installFirstLanguagePack(string $iso, array $params = [], &$errors = []): bool
```
Progi1984:
```suggestion
    public static function installFirstLanguagePack(string $iso, array $params = [], array &$errors = []): bool
```

classes/lang/DataLang.php
```diff
@@ -36,26 +38,33 @@
...
-    /** @var string */
+    /** @var string Default translation domain */
     protected $domain;
 
-    public function __construct($locale)
+    /**
+     * @param string $locale
+     * @param TranslatorInterface|null $translator If defined, use this translator
+     */
+    public function __construct($locale, $translator = null)
```
Progi1984:
```suggestion
    public function __construct($locale, ?TranslatorInterface $translator = null)
```

classes/Tab.php
```diff
@@ -80,6 +86,8 @@ class TabCore extends ObjectModel
             'enabled' => ['type' => self::TYPE_BOOL, 'validate' => 'isBool'],
             'hide_host_mode' => ['type' => self::TYPE_BOOL, 'validate' => 'isBool'],
             'icon' => ['type' => self::TYPE_STRING, 'size' => 64],
+            'wording' => ['type' => self::TYPE_STRING, 'validate' => 'isString', 'size' => 255],
+            'wording_domain' => ['type' => self::TYPE_STRING, 'validate' => 'isString', 'size' => 255],
```
Progi1984:
```suggestion
            'wording' => ['type' => self::TYPE_STRING, 'validate' => 'isString', 'allow_null' => true, 'size' => 255],
            'wording_domain' => ['type' => self::TYPE_STRING, 'validate' => 'isString', 'allow_null' => true, 'size' => 255],
```

classes/Tab.php
```diff
@@ -58,6 +58,12 @@ class TabCore extends ObjectModel
     /** @var string Icon font */
     public $icon;
 
+    /** @var string Wording to use for the display name */
+    public $wording;
+
+    /** @var string Wording domain to use for the display name */
+    public $wording_domain;
```
Progi1984:
```suggestion
    /** @var string|null Wording to use for the display name */
    public $wording;

    /** @var string|null Wording domain to use for the display name */
    public $wording_domain;
```

src/Adapter/EntityTranslation/EntityTranslator.php
```diff
@@ -0,0 +1,246 @@
...
+     * @param TranslatorInterface $translator
+     * @param DataLangCore $dataLang
+     */
+    public function __construct(
+        Db $db,
+        string $dbPrefix,
+        TranslatorInterface $translator,
+        DataLangCore $dataLang
+    ) {
+        //$this->translatorLanguageLoader = new TranslatorLanguageLoader(true);
```
Progi1984:
```suggestion
```

src/PrestaShopBundle/Entity/Tab.php
```diff
@@ -108,6 +108,20 @@ class Tab
      */
     private $icon;
 
+    /**
+     * @var string
```
Progi1984:
```suggestion
     * @var string|null
```

src/PrestaShopBundle/Entity/Tab.php
```diff
@@ -108,6 +108,20 @@ class Tab
...
+     *
+     * @ORM\Column(name="wording", type="string", length=255, nullable=true)
+     */
+    private $wording;
+
+    /**
+     * @var string
```
Progi1984:
```suggestion
     * @var string|null
```

src/PrestaShopBundle/Entity/Tab.php
```diff
@@ -158,6 +172,22 @@ public function getTabLangs()
         return $this->tabLangs;
     }
 
+    /**
+     * @return string|null
+     */
+    public function getWording()
```
Progi1984:
```suggestion
    public function getWording(): ?string
```

src/PrestaShopBundle/Entity/Tab.php
```diff
@@ -158,6 +172,22 @@ public function getTabLangs()
...
+    {
+        return $this->wording;
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getWordingDomain()
```
Progi1984:
```suggestion
    public function getWordingDomain(): ?string
```

Progi1984 (dismissed):

src/PrestaShopBundle/Translation/DataCollectorTranslator.php
Pierstoval: @eternoendless PHPStan is not a true god that you have to follow! If you're certain that you have the right class at the right moment, and you tested the feature, then it works, and you can add an exception to PHPStan. If you do so, remember to add a small comment to explain why there's an exception :)

Julievrz:
Hi @eternoendless, the wordling looks good to me but I only spotted 2 strings. Is there something else that I might be missing? Thanks

eternoendless:
Don't worry @Julievrz I didn't change any of the existing wordings (nor did I add any new one).

eternoendless:
Rebased again

src/Adapter/EntityTranslation/EntityTranslator.php
```diff
@@ -0,0 +1,245 @@
...
+            // Update table
+            if (!empty($updateWhere) && !empty($updateFields)) {
+                $updateWhere = implode(' AND ', $updateWhere);
+                $updateFields = implode(', ', $updateFields);
+
+                $sql = "UPDATE `$tableNameSql`
+                    SET $updateFields
+                    WHERE $updateWhere AND `id_lang` = $languageId"
+                    . $shopWhere
+                    . ' LIMIT 1';
```
PierreRambaud, line 161:
```suggestion
                $sql = 'UPDATE `' . $tableNameSql . '` '
                    'SET $updateFields ' .
                    'WHERE ' . $updateWhere . ' AND `id_lang` = ' . $languageId . ' ' .
                    . $shopWhere
                    . ' LIMIT 1';
```

src/Adapter/EntityTranslation/EntityTranslator.php
```diff
@@ -0,0 +1,245 @@
...
+     * @param string $tableNameSql
+     *
+     * @return bool
+     *
+     * @throws \PrestaShopDatabaseException
+     */
+    protected function shopFieldExists(string $tableNameSql): bool
+    {
+        $columns = $this->db->executeS(
+            sprintf('SHOW COLUMNS FROM `%s`', $tableNameSql)
```
PierreRambaud:
```suggestion
            sprintf('SHOW COLUMNS FROM `%s`', bqSQL($tableNameSql))
```

src/Adapter/EntityTranslation/EntityTranslator.php
```diff
@@ -0,0 +1,245 @@
...
+        );
+
+        foreach ($columns as $column) {
+            if ($column['Field'] == 'id_shop') {
```
PierreRambaud:
```suggestion
            if ($column['Field'] === 'id_shop') {
```

src/PrestaShopBundle/Install/EntityLoader/FileLoader.php
```diff
@@ -0,0 +1,165 @@
...
+     * Returns the lookup paths for the provided entity / iso.
+     *
+     * @param string $entity Entity name
+     * @param string|null $iso [default=null] 2-letter language code
+     *
+     * @return string[] List of lookup paths
+     */
+    private function getLookupPaths(string $entity, ?string $iso = null): array
+    {
+        $fileName = "$entity.xml";
```
PierreRambaud, line 144:
```suggestion
        $fileName = $entity . '.xml';
```

eternoendless: I don't get this obsession with single quotes. I find interpolation much more readable.

src/Adapter/EntityTranslation/EntityTranslator.php
eternoendless: There is an error in your suggestion. IMO I find the original more readable.

PierreRambaud: I just don't like mixing ways to concatenate variables in strings :roll_eyes:

Progi1984 (dismissed):

Progi1984:
@eternoendless Please rebase (PHPStan is :red_circle: & composer has conflicts).

src/PrestaShopBundle/Install/EntityLoader/FileLoader.php
PierreRambaud: Just a habit, make sure the code is always written in the same way

eternoendless:
Rebased once again

Progi1984 (dismissed):

src/PrestaShopBundle/Translation/TranslatorInterface.php
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Interface for PrestaShop translators
+ */
+interface TranslatorInterface extends SymfonyTranslatorInterface
+{
+    /**
+     * Performs a reverse search in the catalogue and returns the translation key if found.
+     * AVOID USING THIS, IT PROVIDES APPROXIMATE RESULTS.
```
sowbiba, line 39: ???

sowbiba (dismissed):
Rebase needed

eternoendless:
I'm not rebasing this again until it's been validated by QA

khouloudbelguith:
Hi @eternoendless,

This issue https://github.com/PrestaShop/PrestaShop/issues/9893 is not fixed with this PR
I cannot translate the `Stock` menu because it doesn't exist 

This issue: https://github.com/PrestaShop/PrestaShop/issues/9816 is not fixed with this PR
image

But I can change the translation successfully 
image

Also, I found an error

1. Install shop with French language & France country
2. Go to BO > International > Localization page
3. Upload United states pack
4. Change employee language to English
5. The menu Design > Themes Catalog is still displayed in French Catalogue de thèmes
6. I tried to translate it but I found 
image

I tried to change the translation of this expression in French, I found two similar expressions of `Catalogue de thèmes`
image

Thank you!

eternoendless:
I Rebased the PR and added two commits:
- 15ede64 - which fixes the Stock wording
- 6c594e5 - which adds support for registering wording and wording domain in module tabs

@khouloudbelguith please test my PR again.

> This issue #9893 is not fixed with this PR
> I cannot translate the `Stock` menu because it doesn't exist

* The `Stock` wording is now fixed in this PR. 
* The other problem is fixed by in the Link Widget module: https://github.com/PrestaShop/ps_linklist/pull/116 (which requires this PR to work).

Regarding the "Theme Catalog" not being translated - this is actually due to the same problem that I fixed in Link Widget, but in this case it must be fixed in MBO. In fact, every module that adds a menu item will display the same problem and need to be fixed individually (ie. the module must register the Tab adding `wording` and `wording_domain`). The problem itself is unrelated to the PR, but this PR makes it possible to fix it!

composer.lock
```diff
@@ -58,6 +58,9 @@
                 "doctrine",
                 "orm"
             ],
+            "support": {
```
Progi1984: Please remove modifications on composer.lock

Progi1984 (commented):
A small feedback

composer.lock
eternoendless: These are added automatically by composer when you run `composer update nothing`
If I change it then composer.lock will become out of sync with composer.json

Progi1984 (dismissed):

composer.lock
Progi1984: composer.lock is again out of sync

eternoendless: I will update it again once QA has validated it

khouloudbelguith:
Hi @eternoendless,

It is ok :heavy_check_mark: 
I make these tests:
1. Install Shop with Arabic language + Saudi Arabia country
2. Add some language via Translation page & pack via Localization page
3. Translate some menus
4. Install Shop with English language + France country
5. Install Shop with French language + France country
6. #9762 => fixed / #9893 (Stock menu) => fixed
7. Add Japan language via Translations page (translate some menus)
8. Import Algeria pack
10. Update all Core Menu 
image

PS: https://github.com/PrestaShop/PrestaShop/issues/9816 still exists
https://github.com/PrestaShop/PrestaShop/issues/23156#issuecomment-774157511 still exists

Thank you!

eternoendless:
Rebased

Progi1984 (dismissed):

PierreRambaud (dismissed):

eternoendless:
Re rebased. Hopefully tests will pass now?

Progi1984 (dismissed):

matks (dismissed):

eternoendless:
After rebase, a new Tab had been introduced in fixtures, breaking the install.

Progi1984 (approved):

matks (approved):

eternoendless:
Thanks everyone! 🎉

PR #18097, opened 2020-03-11
Remove unused result in for statement

Arman-Hosseini:
| Questions     | Answers
| Branch?       | develop
| Description?  | Remove unused result in for statement
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | -

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18097)

### DIFF
--- a/controllers/admin/AdminCartRulesController.php
+++ b/controllers/admin/AdminCartRulesController.php
@@ -170,9 +170,9 @@ class AdminCartRulesControllerCore extends AdminController
                     // Add a new rule group
                     $rule_group_id = 1;
                     if (is_array($rule_group_array)) {
-                        // Empty for (with a ; at the end), that just find the first rule_group_id available in rule_group_array
-                        for ($rule_group_id = 1; in_array($rule_group_id, $rule_group_array); ++$rule_group_id) {
-                            42;
+                        // Find the first rule_group_id that is not available in the array
+                        while (in_array($rule_group_id, $rule_group_array)) {
+                            ++$rule_group_id;
                         }
                         $_POST['product_rule_group'][] = $rule_group_id;
                     } else {


controllers/admin/AdminCartRulesController.php
```diff
@@ -171,9 +171,7 @@ public function postProcess()
                     $rule_group_id = 1;
                     if (is_array($rule_group_array)) {
                         // Empty for (with a ; at the end), that just find the first rule_group_id available in rule_group_array
-                        for ($rule_group_id = 1; in_array($rule_group_id, $rule_group_array); ++$rule_group_id) {
-                            42;
```
matks, line 175: What is that 😅

matks:
Hi @Arman-Hosseini thanks for the PR. I am sorry, honestly I dont understand neither the previous statement nor the new statement you wrote 😅it might be valid php syntax but it's unusual.

Can you please explain the change you introduced ?

Also, maybe if I cant understand this code, it means a lot more people will not be able to understand it either. So maybe we can replace this code by something simpler, easier (maybe less optimized), or add a comment explaining the logic ?

controllers/admin/AdminCartRulesController.php
```diff
@@ -171,9 +171,7 @@ public function postProcess()
...
-                        }
```
Arman-Hosseini: >     // Empty for (with a ; at the end), that just find the first rule_group_id available in rule_group_array 

@matks Now I look at the comment and think that the logic of the code is not correct. The comment states that it finds the first value available in the array.

PierreRambaud (approved):

controllers/admin/AdminCartRulesController.php
```diff
@@ -171,9 +171,7 @@ public function postProcess()
...
+                        for (; in_array($rule_group_id, $rule_group_array); ++$rule_group_id);
```
PierreRambaud: Should be done with a while for a better readability? :thinking:

PierreRambaud (changes requested):
Wrong button, clicked too fast. Read comment :sweat_smile:

controllers/admin/AdminCartRulesController.php
Arman-Hosseini: @PierreRambaud What do you think here?

PierreRambaud: I agree with you, the statement is useless, but if you're thinking algorithm, it should be a whil it's in array, then increase the var

controllers/admin/AdminCartRulesController.php
Arman-Hosseini: Looks good for transparency. But I have a better idea.
The logic of the code gives us The first incremental result that is not in the array.
Isn't this better? 
`$rule_group_id = end($rule_group_array) + 1;`

PululuK: But ... why making it simple when we can make it complicated ? 😁

PierreRambaud: I think they thought about something like this:
```php
$rule_group_array = [1, 2, 3, 4, 5, 9, 10];
for (; in_array($rule_group_id, $rule_group_array); ++$rule_group_id);
// => $rule_group_id = 6;
```

PierreRambaud: Otherwise an autoincrement field should have been used

matks: The very best code change would be something that, when we read it, we *immediately* understand *exactly* what is going on 😄. Here, with the current implementation, we are "I think it does this ..." ; "I think they wanted to do that ..." => such code is a bringer of complexity and bugs

Progi1984 (dismissed):

Progi1984:
@Arman-Hosseini Thanks for the contribution. Could you fix PHPCS Fixer ? Thanks

controllers/admin/AdminCartRulesController.php
PierreRambaud: The whole legacy code is like that :sweat_smile:

Arman-Hosseini:
@Progi1984 Logically there is no error.

PululuK:
> @Progi1984 Logically there is no error.

Hi @Arman-Hosseini ... Thanks man !
To fix it you can just run` php ./vendor/bin/php-cs-fixer fix` from the root folder

Arman-Hosseini:
> > @Progi1984 Logically there is no error.
> 
> Hi @Arman-Hosseini ... Thanks man !
> To fix it you can just run` php ./vendor/bin/php-cs-fixer fix` from the root folder

Hi @PululuK !
There is no difference here between $rule_group_id++ or ++$rule_group_id
But anyway it was done =))

Progi1984 (approved):

PierreRambaud (approved):

PierreRambaud:
No need QA. Thanks @Arman-Hosseini

PR #18098, opened 2020-03-11
Add return to Tools::deleteFile() method

Arman-Hosseini:
| Questions     | Answers
| Branch?       | develop
| Description?  | Adds the return statement in Tools::deleteFile() method.
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | -

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18098)

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -1106,6 +1106,8 @@ class ToolsCore
      *
      * @param string $file File path
      * @param array $exclude_files Excluded files
+     *
+     * @return bool
      */
     public static function deleteFile($file, $exclude_files = [])
     {
@@ -1114,8 +1116,10 @@ class ToolsCore
         }
 
         if (file_exists($file) && is_file($file) && array_search(basename($file), $exclude_files) === false) {
-            unlink($file);
+            return unlink($file);
         }
+
+        return false;
     }
 
     /**


Progi1984 (approved):

matthieu-rolland (dismissed):

matks:
I think this would be considered a BC break. Can you confirm @eternoendless ?

Arman-Hosseini:
@matks It returns no value and is void. As a result, it does not break backward compatibility.

matks:
> @matks It returns no value and is void. As a result, it does not break backward compatibility.

## Before

```
$result = Tools::deleteFile($file);
// $result is null
```

## After

```
$result = Tools::deleteFile($file);
// $result is false or true
```

If someone's code rely on the void result* it will break his code.

*Example

```
$result = Tools::deleteFile($file);

id (null !== $result) {
    // do something
}
```
The code above works based on a mistake (= developer thinks `Tools::deleteFile()` returns something) but works.

If we follow SemVer this is a BC break, even though it's a code improvement.

As OSS maintainers, "we have to live with our mistakes until the major" 😢

## How to avoid breaking SemVer

### Use flag

Use flag to return bool, so only people with knowledge of the flag can use it

```
public static function deleteFile($file, $exclude_files = [], $returnResultAsBool = false)
```

### Create new, better function

```
public static function deleteFileV2($file, $exclude_files = []) : bool
```

Arman-Hosseini:
Why should a developer check it with a null/non-null value? When its value is always null.
Whenever I want to use a method, I first refer to its documentation. Any developer should check the documentation, even if it's a novice.
When you don't know how a method works and what kind of data it returns, then how do you want to use it?

matthieu-rolland:
> Why should a developer check it with a null/non-null value? When its value is always null.
> Whenever I want to use a method, I first refer to its documentation. Any developer should check the documentation, even if it's a novice.
> When you don't know how a method works and what kind of data it returns, then how do you want to use it?

Yes we know it's bad...  Your modification is cleaner but we can't implement it:

@matks is right, changing what a function returns in a given circumstance is a breaking change. For example if a module expects this method to return nothing/null all the time, then it will stop working when the function will return true.  

We can only do breaking changes on major versions, this is not a new major version.

PierreRambaud:
It's more a bug than a BC, otherwise all changes are BC because the behavior isn't the same :thinking: 

And here @matks you're wrong it's an improvement :sweat_smile:

PierreRambaud (approved):

PierreRambaud:
No need QA. @eternoendless confirmed us, it's not a BC because the return was void :)
Thanks @Arman-Hosseini

matks:
> No need QA. @eternoendless confirmed us, it's not a BC because the return was void :)

I learned something today 😊thanks !

Sorry for confusion @Arman-Hosseini 🙏

PR #18116, opened 2020-03-12
Avoid crash because of image timeout

### LINKED ISSUE #18115: "Modify module positions" BO page fails if an image from Addons is too long to load
#### Describe the bug

I was working on "Modify module positions" BO page and my page crashed because a module image failed to load fast enough: see screenshot

URL http://localhost/admin-dev/index.php?controller=AdminModulesPositions&addToHook=&show_modules=69&token=XXXX

#### Expected behavior

A BO page should never crash because of a slow image load 😅 

**Screenshots**

Capture d’écran 2020-03-12 à 11 54 35

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Improve `Module::copyModAddonsImg()` to avoid crashes if timeout
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/18115
| How to test?  | I dont know 😓 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18116)

### DIFF
--- a/classes/module/Module.php
+++ b/classes/module/Module.php
@@ -23,7 +23,6 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
-use Exception;
 use PrestaShop\PrestaShop\Adapter\LegacyLogger;
 use PrestaShop\PrestaShop\Adapter\Module\ModuleDataProvider;
 use PrestaShop\PrestaShop\Adapter\ServiceLocator;
--- a/classes/module/Module.php
+++ b/classes/module/Module.php
@@ -31,6 +31,7 @@ use PrestaShop\PrestaShop\Core\Foundation\Filesystem\FileSystem;
 use PrestaShop\PrestaShop\Core\Module\ModuleInterface;
 use PrestaShop\PrestaShop\Core\Module\WidgetInterface;
 use PrestaShop\TranslationToolsBundle\Translation\Helper\DomainHelper;
+use Exception;
 
 abstract class ModuleCore implements ModuleInterface
 {
@@ -1613,16 +1614,17 @@ abstract class ModuleCore implements ModuleInterface
         if (!Validate::isLoadedObject($modaddons)) {
             return null;
         }
+        
+        $filename = md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg';
+        $filepath = _PS_TMP_IMG_DIR_ . $filename;
+        $fileExist = file_exists($filepath);
 
-        $filepath = _PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg';
-        $fileDoesNotExist = !file_exists($filepath);
-
-        if ($fileDoesNotExist) {
+        if (!$fileExist) {
             $remoteDownloadWasASuccess = false;
             try {
                 $remoteImage = Tools::file_get_contents($modaddons->img);
                 $remoteDownloadWasASuccess = true;
-            } catch (\Exception $e) {
+            } catch (Exception $e) {
                 copy(_PS_IMG_DIR_ . '404.gif', $filepath);
             }
 
@@ -1631,8 +1633,8 @@ abstract class ModuleCore implements ModuleInterface
             }
         }
 
-        if (file_exists($filepath)) {
-            return '../img/tmp/' . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg';
+        if ($fileExist) {
+            return '../img/tmp/' . $filename;
         }
     }
 
--- a/classes/module/Module.php
+++ b/classes/module/Module.php
@@ -1603,16 +1603,35 @@ abstract class ModuleCore implements ModuleInterface
         return $module_list;
     }
 
+    /**
+     * @param \StdClass $modaddons Addons Module object, provided by XML stream
+     *
+     * @return null|string
+     */
     public static function copyModAddonsImg($modaddons)
     {
         if (!Validate::isLoadedObject($modaddons)) {
-            return;
+            return null;
         }
-        if (!file_exists(_PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg') &&
-        !file_put_contents(_PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg', Tools::file_get_contents($modaddons->img))) {
-            copy(_PS_IMG_DIR_ . '404.gif', _PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg');
+
+        $filepath = _PS_TMP_IMG_DIR_ . md5((int)$modaddons->id . '-' . $modaddons->name) . '.jpg';
+        $fileDoesNotExist = !file_exists($filepath);
+
+        if ($fileDoesNotExist) {
+            $remoteDownloadWasASuccess = false;
+            try {
+                $remoteImage = Tools::file_get_contents($modaddons->img);
+                $remoteDownloadWasASuccess = true;;
+            } catch (\Exception $e) {
+                copy(_PS_IMG_DIR_ . '404.gif', $filepath);
+            }
+
+            if ($remoteDownloadWasASuccess && !file_put_contents($filepath, $remoteImage)) {
+                copy(_PS_IMG_DIR_ . '404.gif', $filepath);
+            }
         }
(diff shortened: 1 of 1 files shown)


classes/module/Module.php
```diff
@@ -1603,16 +1603,35 @@ public static function getModulesOnDisk($use_config = false, $logged_on_addons =
...
+     * @param \StdClass $modaddons Addons Module object, provided by XML stream
+     *
+     * @return string|null
+     */
     public static function copyModAddonsImg($modaddons)
     {
         if (!Validate::isLoadedObject($modaddons)) {
-            return;
+            return null;
         }
```
PululuK, line 1615: Returning null is too silent ... 
Rather check if it is a module instance instead of isLoadedObject and throw an exception.

something like that :
```php
         if (!($modaddons instanceof Module)) {
            throw new InvalidModuleException('Invalud module ... etc etc');
        }
```

matks: I'm afraid of modifying the code of such an old class used so deeper 😅

This part of the code is very tangled and buried deep inside layers of layers of code. I'm afraid of breaking something.

matthieu-rolland (dismissed):

classes/module/Module.php
```diff
@@ -1603,16 +1603,35 @@ public static function getModulesOnDisk($use_config = false, $logged_on_addons =
...
+            }
+
+            if ($remoteDownloadWasASuccess && !file_put_contents($filepath, $remoteImage)) {
+                copy(_PS_IMG_DIR_ . '404.gif', $filepath);
+            }
         }
-        if (file_exists(_PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg')) {
+
+        if (file_exists($filepath)) {
             return '../img/tmp/' . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg';
```
PierreRambaud: Don't do `md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg'` twice please :sweat_smile:

classes/module/Module.php
```diff
@@ -1603,16 +1603,35 @@ public static function getModulesOnDisk($use_config = false, $logged_on_addons =
...
+
+        $filepath = _PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg';
+        $fileDoesNotExist = !file_exists($filepath);
+
+        if ($fileDoesNotExist) {
+            $remoteDownloadWasASuccess = false;
+            try {
+                $remoteImage = Tools::file_get_contents($modaddons->img);
+                $remoteDownloadWasASuccess = true;
+            } catch (\Exception $e) {
```
PululuK: Please add a use Exception; instead of the \ notation

classes/module/Module.php
```diff
@@ -1603,16 +1603,35 @@ public static function getModulesOnDisk($use_config = false, $logged_on_addons =
...
         if (!Validate::isLoadedObject($modaddons)) {
-            return;
+            return null;
         }
-        if (!file_exists(_PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg') &&
-        !file_put_contents(_PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg', Tools::file_get_contents($modaddons->img))) {
-            copy(_PS_IMG_DIR_ . '404.gif', _PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg');
+
+        $filepath = _PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg';
+        $fileDoesNotExist = !file_exists($filepath);
```
PululuK: some optimizations

PululuK (changes requested):
@see : https://github.com/matks/PrestaShop/pull/4/files

PierreRambaud (changes requested):
See comments

classes/module/Module.php
matks: This is 2020 ! Hardware is cheap ! 😄

matks: (the code, before my changes, did it 3 or 4 times)

Anyway @PululuK fixed my code and addresses your feedback

matks:
Big thanks to @PululuK for not only reviewing this PR but also submitting a PR to my fork with valid (and better) changes / suggestions

@PierreRambaud your word, now 😄

classes/module/Module.php
```diff
@@ -1603,17 +1604,37 @@ public static function getModulesOnDisk($use_config = false, $logged_on_addons =
...
+            }
+
+            if ($remoteDownloadWasASuccess && !file_put_contents($filepath, $remoteImage)) {
+                copy(_PS_IMG_DIR_ . '404.gif', $filepath);
+            }
         }
-        if (file_exists(_PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg')) {
-            return '../img/tmp/' . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg';
+
+        if ($fileExist) {
```
PierreRambaud:
```suggestion
        if (file_exists($filepath)) {
```
You need to do it one more time, otherwise it just keep the result before trying to copy

matks: @PierreRambaud Done. But now I guess we should add a `else()` right ? Else function returns void

classes/module/Module.php
```diff
@@ -1603,17 +1604,37 @@ public static function getModulesOnDisk($use_config = false, $logged_on_addons =
...
+            if ($remoteDownloadWasASuccess && !file_put_contents($filepath, $remoteImage)) {
+                copy(_PS_IMG_DIR_ . '404.gif', $filepath);
+            }
         }
-        if (file_exists(_PS_TMP_IMG_DIR_ . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg')) {
-            return '../img/tmp/' . md5((int) $modaddons->id . '-' . $modaddons->name) . '.jpg';
+
+        if (file_exists($filepath)) {
+            return '../img/tmp/' . $filename;
         }
```
Progi1984:
```suggestion
        if ($fileExist) {
            return '../img/tmp/' . $filename;
        }
```

classes/module/Module.php
```diff
@@ -1603,17 +1604,37 @@ public static function getModulesOnDisk($use_config = false, $logged_on_addons =
...
+            try {
+                $remoteImage = Tools::file_get_contents($modaddons->img);
+                $remoteDownloadWasASuccess = true;
+            } catch (Exception $e) {
+                copy(_PS_IMG_DIR_ . '404.gif', $filepath);
+            }
+
+            if ($remoteDownloadWasASuccess && !file_put_contents($filepath, $remoteImage)) {
+                copy(_PS_IMG_DIR_ . '404.gif', $filepath);
+            }
```
Progi1984:
```suggestion
            }
            $fileExist = file_exists($filepath);
```

Progi1984 (dismissed):

classes/module/Module.php
PierreRambaud: The function was already returning void :thinking:

PierreRambaud (dismissed):

classes/module/Module.php
```diff
@@ -23,6 +23,7 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
+use Exception;
```
Progi1984: To remove ;)

matks: I have no opinion on this (use `use` for native php classes) but I think @PierreRambaud asked me to do the opposite a long time ago 🤔

Progi1984: It's just for fixing Travis ;)

PierreRambaud: I asked to use namespace object when it's needed (in Symfony pages for example) :sweat_smile: 
It's not mandatory to remove the use here, it can help to see which Exception are thrown in this class.

matks: My bad, sorry @Progi1984

PierreRambaud (approved):

Progi1984 (approved):

matthieu-rolland (approved):

matks:
I deleted the branch by mistake. Restoring it and reopening PR

PululuK (approved):

PierreRambaud (approved):

SD1982:
LGTM Thanks @matks !

Progi1984:
Thanks @matks

Issue #17993, opened 2020-03-05
Optimize the 1.7.7 translation catalog


LouiseBonnard:
### Admin.Advparameters.Feature

- [x] ‘MPN’ to be localized in **Admin.Catalog.Feature** instead
cf. https://crowdin.com/translate/prestashop-official/35796/en-en

### Admin.Advparameters.Notification

- [x] ‘Are you sure that you would like to delete this entity: ‘ to be changed into **Are you sure you want to delete this entity:** without the extra space at the end
cf. https://crowdin.com/translate/prestashop-official/35798/en-en#5831278

- [x] ‘Advanced stock management has incorrect value. Not set for product %name% ‘ to be written without the extra space at the end

- [x] ‘Advanced stock management is not enabled, cannot enable on product %name% ‘ to be written without the extra space at the end

- [x] ‘Advanced stock management is not enabled, warehouse not set on product %name% ‘ to be written without the extra space at the end

- [x] ‘Warehouse did not exist, cannot set on product  %name% ‘ to be written without the extra space at the end

- [x] ‘Incorrect value for "Depends on stock" for product %name% ‘ to be written without the extra space at the end

- [x] ‘Advanced stock management is not enabled, cannot set "Depends on stock" for product %name% ‘ to be written without the extra space at the end

### Admin.Catalog.Feature

- [x] ‘MPN’ is written three times, is there any reason to have so many duplicates?
cf. https://crowdin.com/translate/prestashop-official/35748/en-en#q=mpn

- [x] ‘Payment: ‘ to be written without the extra space at the end

- [x] ‘Message to: ‘ to be written without the extra space at the end

- [x] ‘Product: ‘ to be written without the extra space at the end

### Admin.Catalog.Notification

- [x] ‘The display of your brands is enabled on your store. Go to %sShop Parameters > General to edit settings.%s’ needs its last variable changed into **The display of your brands is enabled on your store. Go to %sShop Parameters > General%s to edit settings.**

- [x] ‘The display of your brands is disabled on your store. Go to %sShop Parameters > General to edit settings.%s’ needs its last variable changed into **The display of your brands is disabled on your store. Go to %sShop Parameters > General%s to edit settings.**

- [x] ‘The display of your suppliers is enabled on your store. Go to %sShop Parameters > General to edit settings%s.’ needs its last variable changed into **The display of your suppliers is enabled on your store. Go to %sShop Parameters > General%s to edit settings.**

[...]

PR #17819, opened 2020-02-25
Split configuration forms into single form management

PierreRambaud:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Configuration forms are badly designed. The problem is, it creates a form with subforms, and these subforms are used like a single form. <br>Each time a back office user submit a form, the whole page, meaning all forms are submitted even if they are split, and it's harder to override with services, templates, or form handler.<br>The solution is to create single form management each time we can by adding a route per action, a form handler management per form.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | yes
| How to test?  | Forms must worked as before.<br>- Configure -> Advanced -> Administration<br>- Configure -> Advanced -> Backup <br>- Configure -> Advanced -> Logs<br>- Configure -> Advanced -> Performance<br>- Configure -> Advanced -> SqlManager<br>- Configure -> Advanced -> Webservices<br>- Configure -> Shop parameters -> Préferences<br>- Configure -> Shop parameters -> Maintenance<br>- Configure -> Shop parameters -> Order Preferences<br>- Configure -> Shop parameters -> Customer Preferences<br>- Improve -> Design -> Mail Themes<br>- Improve -> Design -> Themes<br>- Improve -> International -> Geolocation<br>- Improve -> International -> Localization<br>- Improve -> International -> Tax<br>- Improve -> Payment -> Payment<br>- Improve -> Shipping -> Preferences<br>- Sell -> Customer -> Merchandise return<br>- Sell -> Order -> Delivery<br>- Sell -> Order -> Invoices

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17819)

PR #18124, opened 2020-03-12
Modifying address from order should duplicate it and soft delete former address

### LINKED ISSUE #17978: Editing an address in the order should create a new address
#### Describe the bug

If you change an address (billing or shipping address) in the order, the new address will be updated in all customer orders.

#### Why 

The order page does not dissociate between modifying an erroneous address and entering a new address.

#### Expected behavior

Editing an address in the order should:
- create a new address 
- it only updates the address of the current order

#### Steps to Reproduce

1. Go to BO > orders > orders > View order
2. Click on edit address
3. Change the address and save
4. The address has been edited and changed for both billing and delivery

**Screenshots**

https://drive.google.com/file/d/1Rx7oDD_8YLr0cMDwLDGm9-qw8aTFfN6Y/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When editing an address from the order view page the edited address data was then impacted on every cart and orders This PR includes two fixes:<br/>1- when an address is edited if it is used in an order (as delivery or invoice address) then the modification are saved into a duplicated address row, and the former one is soft deleted (via a database field)<br/>2- a new controller has been added to edit order's addresses, so the duplication process natively occurs though address editing but then the order (and its cart) must be updated with the new address, and the order shipping cost is updated
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17978
| How to test?  | See issue's problems Basically edit an order delivery address, when going back to the order view only the delivery address must have been updated Any other orders that used this address is unchanged In the Customer's addresses list you should not see more addresses, the former one has been soft deleted and replaced by the new one

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18124)


src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
```diff
@@ -57,12 +60,32 @@ public function handle(EditCustomerAddressCommand $command): void
                 throw new AddressConstraintException('Address contains invalid field values');
             }
 
-            if (false === $address->update()) {
+            // The address is used by an order so it is not edited directly, instead a copy is created and
+            if ($address->isUsed()) {
+                // Get a copy of current address
+                $old_address = new Address($address->id);
```
PierreRambaud:
```suggestion
                $copyAddress = new Address($address->id);
```

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
```diff
@@ -57,12 +60,32 @@ public function handle(EditCustomerAddressCommand $command): void
...
+
+                // Reset ID to force recreating a new address
+                $address->id = $address->id_address = null;
```
PierreRambaud: I don't understand why we must reset the current address id and not the copied one? :thinking:

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
```diff
@@ -57,12 +60,32 @@ public function handle(EditCustomerAddressCommand $command): void
...
+
+                // We consider this address as necessarily NOT deleted, in case you were editing a deleted address
+                // from an order then the newly edited address should not be deleted, so that you can select it
+                $address->deleted = 0;
```
PierreRambaud: This means, if we edit a deleted address, it's automatically undeleted? should be `if ($address->isUsed() && !$address->deleted)` or something similar?

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
```diff
@@ -57,12 +60,32 @@ public function handle(EditCustomerAddressCommand $command): void
...
+                if (false === $address->save()) {
+                    throw new CannotAddAddressException(sprintf('Failed to add new address "%s"', $command->getAddress()));
+                }
+                // Soft delete the former address
+                if (false === $old_address->delete()) {
```
PierreRambaud: delete $address to be consistent and save the copy instead

src/Core/Form/IdentifiableObject/DataHandler/OrderAddressFormDataHandler.php
```diff
@@ -0,0 +1,133 @@
...
+
+        if (null !== $data['id_state']) {
+            $editAddressCommand->setStateId((int) $data['id_state']);
+        }
+
+        if (null !== $data['phone']) {
+            $editAddressCommand->setHomePhone($data['phone']);
+        }
+
+        if (isset($data['phone_mobile']) && null !== $data['phone_mobile']) {
```
PierreRambaud:
```suggestion
        if (isset($data['phone_mobile'])) {
```
Why doing the same check? :sweat_smile:

src/Core/Form/IdentifiableObject/DataHandler/OrderAddressFormDataHandler.php
```diff
@@ -0,0 +1,133 @@
...
+     * @throws CountryConstraintException
+     * @throws StateConstraintException
+     * @throws InvalidAddressTypeException
+     * @throws OrderException
+     */
+    public function update($orderId, array $data)
+    {
+        $editAddressCommand = new EditOrderAddressCommand($orderId, $data['address_type']);
+
+        if (null !== $data['alias']) {
```
PierreRambaud: why not using isset instead?

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -401,6 +404,90 @@ public function editAction(int $addressId, Request $request): Response
...
+
+            // Addres form is built based on address id to fill the data related to this address
+            $addressForm = $addressFormBuilder->getFormFor($addressId, $formData);
+            $addressForm->handleRequest($request);
+
+            // Form is handled based on Order ID because that's the order that needs update
+            $result = $addressFormHandler->handleFor($orderId, $addressForm);
+
+            if ($result->isSubmitted() && $result->isValid()) {
+                $this->addFlash('success', $this->trans('Successful update.', 'Admin.Notifications.Success'));
```
LouiseBonnard:
```suggestion
                $this->addFlash('success', $this->trans('Update successful', 'Admin.Notifications.Success'));
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/customer.html.twig
```diff
@@ -110,14 +110,7 @@
...
-                  'realedit': 1,
-                  'updateaddress': 1,
-                  'address_type': 1,
-                  'id_order': orderForViewing.id,
-                  'id_address': orderForViewing.shippingAddress.addressId,
-                  'back': path('admin_orders_view', {'orderId': orderForViewing.id})
-                }) }}"
+                 href="{{ path('admin_order_addresses_edit', {'orderId': orderForViewing.id, 'addressType': 'delivery'}) }}"
               >
                 {{ 'Edit address'|trans({}, 'Admin.Actions') }}
```
LouiseBonnard: It might be related to the changes requested by @colinegin and mentioned [here](https://github.com/PrestaShop/PrestaShop/issues/17993), in the **Admin.Actions** domain: _Edit address_ will be clearer if turned into _Edit existing address_ - as well as _Change address_ into _Select another address_.

FYI @marionf, @MatShir

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/customer.html.twig
```diff
@@ -166,14 +159,7 @@
...
-                  'realedit': 1,
-                  'updateaddress': 1,
-                  'address_type': 2,
-                  'id_order': orderForViewing.id,
-                  'id_address': orderForViewing.invoiceAddress.addressId,
-                  'back': path('admin_orders_view', {'orderId': orderForViewing.id})
-                }) }}"
+                 href="{{ path('admin_order_addresses_edit', {'orderId': orderForViewing.id, 'addressType': 'invoice'}) }}"
               >
                 {{ 'Edit address'|trans({}, 'Admin.Actions') }}
```
LouiseBonnard: Same here, I guess.

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
matthieu-rolland: Here the $address object is the new one, it contains all the data of the address we are creating, so its id must be null so that a new address is created. 

$old_address is what you call the "copied one", this one must keep its id, it's the one that will be soft deleted (without the data modifications)

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
jolelievre: Done

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
jolelievre: Exactly `$address` contains the edited data so this is the one that needs to be inserted in the DB, setting null to the id is kind of a trick but it makes the job

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
jolelievre: Yeh I wasn't 100% sure about this one, but here is the use case: let's you have an order with invoice and delivery address with id 4. You edit the delivery address, so the address with id 4 becomes deleted, and a new address is inserted

Now you also want to edit the invoice address which is still the address with id 4 (deleted), you create a new address by editing it, but since this new address is based on a deleted address the new address is also deleted.

I don't think this is what the user expects, a newly edited and created address should not be deleted I guess. And worst case scenario the new address can always be deleted in the list anyway.

What do you think?

jolelievre: Oh and by the way regarding the condition you proposed, it changes nothing as we are already in a block `if ($address->isUsed())` And i we only set `deleted = 0` when `deleted == 0` then it changes nothing
The real question is indeed should we undelete it by default? as explained in the use case above.

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
jolelievre: As explained above `$address` contains the edited data so it makes more sense to save/insert this one Besides `$address->delete()` sets deleted to true and then performs an update. So it would erase the previous data and save the edited data in the deleted row.

But I understand that the code is not really instinctive.. Maybe if we renamed `$address` into `$editedAddress` it would be easier to understand?

src/Core/Form/IdentifiableObject/DataHandler/OrderAddressFormDataHandler.php
jolelievre: It is a dumb copy/paste from `AddressFormDataHandler` 😛
I think the isset is necessary because in some cases the field is not present, but then the latter part is useless indeed, I'll remove it

src/Core/Form/IdentifiableObject/DataHandler/OrderAddressFormDataHandler.php
jolelievre: I guess we can replace with isset everywhere indeed

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
jolelievre: Done

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/customer.html.twig
jolelievre: Ok, I changed these two wordings

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/customer.html.twig
jolelievre: Done as well

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
jolelievre: I only changed them in this controller, but there are many other use of this wrong wording in other controllers I don't want to make all the updates in this PR because it's not related to the issue, but should we change all the present wordings in another PR to be consistent everywhere?

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
```diff
@@ -48,21 +51,41 @@ final class EditCustomerAddressHandler extends AbstractAddressHandler implements
...
 
-            if (false === $address->validateFields(false)) {
+            if (false === $editedAddress->validateFields(false)) {
                 throw new AddressConstraintException('Address contains invalid field values');
             }
 
-            if (false === $address->update()) {
-                throw new CannotUpdateAddressException(sprintf('Failed to update address "%s"', $address->id));
+            // The address is used by an order so it is not edited directly, instead a copy is created and
+            if ($editedAddress->isUsed()) {
```
jolelievre, line 64: @PierreRambaud do you think it's clearer with this naming? Also I thought maybe `$backupAddress` instead of `$copyAddress` would be better?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/Modal/update_customer_address_modal.html.twig
```diff
@@ -50,7 +50,7 @@
           <button type="button" class="btn btn-outline-secondary" data-dismiss="modal">
             {{ 'Cancel'|trans({}, 'Admin.Actions') }}
           </button>
-          
+
           <button type="submit" class="btn btn-primary">
             {{ 'Change'|trans({}, 'Admin.Actions') }}
```
LouiseBonnard:
```suggestion
            {{ 'Select'|trans({}, 'Admin.Actions') }}
```

LouiseBonnard: In order to stay consistent. ;-)

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
LouiseBonnard: Absolutely! I think a part of it has already been done in PR #17819, can you check and let me know if all controllers are changed? If not, I'll create an issue to have it all changed. :-)

src/Adapter/Address/CommandHandler/EditCustomerAddressHandler.php
PierreRambaud: It's not really a backup but a real copy.
But it's clearer now :)

PierreRambaud (dismissed):

src/Adapter/Address/CommandHandler/EditOrderAddressHandler.php
```diff
@@ -0,0 +1,168 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: Miss `declare(strict_types=1);`

src/Core/Domain/Address/Command/AbstractEditAddressCommand.php
```diff
@@ -0,0 +1,414 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: Miss `declare(strict_types=1);`

src/Core/Domain/Address/Command/EditOrderAddressCommand.php
```diff
@@ -0,0 +1,103 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: Miss `declare(strict_types=1);`

src/Core/Domain/Address/CommandHandler/EditOrderAddressHandlerInterface.php
```diff
@@ -0,0 +1,43 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: Miss `declare(strict_types=1);`

src/Core/Domain/Address/Exception/CannotUpdateOrderAddressException.php
```diff
@@ -0,0 +1,34 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: Miss `declare(strict_types=1);`

src/Core/Domain/Cart/Exception/CannotUpdateCartException.php
```diff
@@ -0,0 +1,34 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: Miss `declare(strict_types=1);`

src/Core/Domain/Order/Exception/CannotUpdateOrderException.php
```diff
@@ -0,0 +1,34 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: Miss `declare(strict_types=1);`

src/Core/Domain/Order/Exception/InvalidAddressTypeException.php
```diff
@@ -0,0 +1,34 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: Miss `declare(strict_types=1);`

src/Core/Domain/Order/OrderAddressType.php
```diff
@@ -0,0 +1,37 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: Miss `declare(strict_types=1);`

src/Core/Form/IdentifiableObject/DataHandler/OrderAddressFormDataHandler.php
```diff
@@ -0,0 +1,133 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: Miss `declare(strict_types=1);`

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -401,6 +404,90 @@ public function editAction(int $addressId, Request $request): Response
...
+            $addressId = null;
+            switch ($addressType) {
+                case 'delivery':
+                    $addressType = OrderAddressType::DELIVERY_ADDRESS_TYPE;
+                    $addressId = $order->id_address_delivery;
+                    break;
+                case 'invoice':
+                    $addressType = OrderAddressType::INVOICE_ADDRESS_TYPE;
+                    $addressId = $order->id_address_invoice;
+                    break;
```
Progi1984, line 434: default ? An exception ?

jolelievre: The exception is already managed in the command class which checks allowed address types

jolelievre:
Thank you @Progi1984 I addressed all your feedbacks

Progi1984 (dismissed):

Progi1984:
@jolelievre You have some errors in Behat tests : https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/299023607#L7674

jolelievre:
@Progi1984 thanks it was related to strict typing, it's ok now

Progi1984 (approved):

matks (approved):

SD1982:
This PR correctly fix #17978 . Thanks @jolelievre

PR #18130, opened 2020-03-13
Use grid common search action where possible (1.7.7.x controllers only)

### LINKED ISSUE #13448: [SF Migration] Harmonize controller grid search action
We have a generic and reusable [searchGridAction](https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Controller/Admin/CommonController.php#L335) in CommonController but some Controllers use their own `searchAction` rather than the reusable one.

For each Controller which has a custom searchAction, check if there is a valid need/reason for it, else make it use the generic `searchGridAction`.

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Implements CommonController::searchGridAction (only of 1.7.7.x controllers)  where it was possible to do easily. Custom search actions are only left for multi-grid pages.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #13448
| How to test?  | Check that grid filtering works as before (if these changes broke something, then filtering would not work at all, not necessary to test all the edge-cases). Modified grids list below.<br>- orders,<br>- attributes and attribute groups (`sell/catalog/attribute-groups` view action opens attributes list),<br>- customer addresses,<br>- order messages,<br>- merchandise returns,<br>- suppliers ( this was migrated already in 1.7.6, but was still hidden)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18123)

### DIFF
--- a/src/Core/Grid/Definition/Factory/CmsPageCategoryDefinitionFactory.php
+++ b/src/Core/Grid/Definition/Factory/CmsPageCategoryDefinitionFactory.php
@@ -62,6 +62,7 @@ final class CmsPageCategoryDefinitionFactory extends AbstractGridDefinitionFacto
      * @var int
      */
     private $cmsCategoryParentId;
+
     /**
      * @var MultistoreContextCheckerInterface
      */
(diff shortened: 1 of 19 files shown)


Progi1984 (approved):

PierreRambaud (approved):

SD1982:
This PR is working as expected, no unwanted behavior has been noticed. Thanks @zuk3975

Progi1984:
Thanks @zuk3975

PR #18135, opened 2020-03-13
Add some phpdoc to Tools

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add, fix and enhance phpDoc inside Tools class
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18135)

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -2147,7 +2147,7 @@ class ToolsCore
      * @param int $curl_timeout
      * @param array $opts
      *
-     * @return bool|mixed
+     * @return bool|string
      *
      * @throws Exception
      */
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -52,6 +52,9 @@ class ToolsCore
 
     public static $round_mode = null;
 
+    /**
+     * @param Request $request
+     */
     public function __construct(Request $request = null)
     {
         if ($request) {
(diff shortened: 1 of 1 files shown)


classes/Tools.php
```diff
@@ -1337,7 +1359,13 @@ public static function getAdminTokenLite($tab, Context $context = null)
         return Tools::getAdminToken($tab . (int) Tab::getIdFromClassName($tab) . (int) $context->employee->id);
     }
 
-    public static function getAdminTokenLiteSmarty($params, &$smarty)
+    /**
+     * @param array $params
+     * @param $smarty unused parameter, please ignore (@todo: remove in next major)
```
matks, line 1364: This parameter is not used in the function 🙄

Progi1984 (dismissed):

matthieu-rolland (dismissed):

classes/Tools.php
```diff
@@ -2110,6 +2142,15 @@ public static function refreshCACertFile()
         }
     }
 
+    /**
+     * @param string $url
+     * @param int $curl_timeout
+     * @param array $opts
+     *
+     * @return bool|mixed
```
```suggestion
     * @return bool|string
```
Set CURLOPT_RETURNTRANSFER to TRUE to return the transfer as a string of the return value of curl_exec() instead of outputting it out directly.

matks: Thanks I apply it

Progi1984 (approved):

PierreRambaud (approved):

PierreRambaud:
No need QA. Thanks @matks

PR #18136, opened 2020-03-13
Use grid common search action where possible (1.7.8 target)

### LINKED ISSUE #13448: [SF Migration] Harmonize controller grid search action

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Implements CommonController::searchGridAction where it was possible to do easily. Custom search actions are only left for multi-grid pages. Deprecated unusable search actions
| Type?         | refacto
| Category?     | BO
| BC breaks?    | yes, some constructors have been modified (php classes and Symfony service definition) but it was only `final` classes
| Deprecations? | no
| Fixed ticket? | #13448
| How to test?  | Check that grid filtering works as before (if these changes broke something, then filtering would not work at all, not necessary to test all the edge-cases). Modified grids list below.<br>- categories,<br>- credit slips,<br>- customers,<br>- languages,<br>- currencies,<br>- taxes,<br>- log,<br>- meta,<br>- requestsql,<br>- webserviceKey

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18123)


Progi1984 (dismissed):

PierreRambaud (dismissed):

zuk3975:
@jolelievre mentioned that some bugs have occurred in older grids due to missing grid_ids when merged similar pr to 1.7.7.x (https://github.com/PrestaShop/PrestaShop/pull/18130). And some BC breaks might be introduced. This pr shares the same changes of CommonController (as in abovementioned pr) so we need his input before merging this one.

jolelievre:
Yes indeed, there is a breaking change on old grids (using controller/action matching instead of filterId)
Here is the PR with fix proposal: https://github.com/PrestaShop/PrestaShop/pull/18228

jolelievre (changes requested):
Hi @zuk3975 
Sorry but I block this PR only so we don't forget to update it, as we discussed there was a breaking change in CommonController which is fixed in this PR https://github.com/PrestaShop/PrestaShop/pull/18228

So a new interface `IdentifiableGridDefinitionFactoryInterface` has been introduced so that `CommonController` can know how to render the response. For easier integration an abstract class `AbstractIdentifiableGridDefinitionFactory` implementing this interface has been introduced So you just need to change the parent class of your grid definitions to make it all work fine

jolelievre:
Blocked until https://github.com/PrestaShop/PrestaShop/pull/18228 is merged (and merged into develop)

zuk3975:
> Sorry but I block this PR only so we don't forget to update it, as we discussed there was a breaking change in CommonController which is fixed in this PR #18228

Sure, no problem, its for develop, so no need to rush.

khouloudbelguith:
Hi @zuk3975,

Can you please resolve conflicts.

Thanks!

src/Core/Grid/Definition/Factory/LogGridDefinitionFactory.php
```diff
@@ -229,10 +231,11 @@ public function getFilters()
             ->add(
                 (new Filter('actions', SearchAndResetType::class))
                     ->setTypeOptions([
-                        'attr' => [
-                            'data-url' => $this->resetActionUrl,
-                            'data-redirect' => $this->redirectionUrl,
+                        'reset_route' => 'admin_common_reset_search_by_filter_id',
```
jolelievre, line 216: I think you could clean the constructor (since the arguments are not used any more) and the service definition no?

It might be considered a BC break although it's arguable since this is a final class not destined to be extended nor constructed manually, so at worst it's a small one

src/Core/Grid/Definition/Factory/WebserviceKeyDefinitionFactory.php
```diff
@@ -205,10 +207,11 @@ protected function getFilters()
             ->add(
                 (new Filter('actions', SearchAndResetType::class))
                     ->setTypeOptions([
-                        'attr' => [
-                            'data-url' => $this->resetActionUrl,
-                            'data-redirect' => $this->redirectionUrl,
+                        'reset_route' => 'admin_common_reset_search_by_filter_id',
```
jolelievre, line 194: Same as `LogGridDefinitionFactory` construction could be cleaned

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/LogsController.php
```diff
@@ -83,22 +84,16 @@ public function indexAction(LogsFilters $filters)
...
-
-        $searchParametersForm->handleRequest($request);
-        $filters = [];
-
         $this->dispatchHook('actionAdminLogsControllerPostProcessBefore', ['controller' => $this]);
 
-        if ($searchParametersForm->isSubmitted()) {
-            $filters = $searchParametersForm->getData();
-        }
+        $responseBuilder = $this->get('prestashop.bundle.grid.response_builder');
```
jolelievre, line 89: Why modify this controller and not use the common search one instead? This could simply be deprecated and not modified right?

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/SqlManagerController.php
```diff
@@ -113,20 +113,14 @@ public function indexAction(Request $request, RequestSqlFilters $filters)
...
-        $filtersForm->handleRequest($request);
-
-        $filters = [];
-
-        if ($filtersForm->isSubmitted()) {
-            $filters = $filtersForm->getData();
-        }
-
-        return $this->redirectToRoute('admin_sql_requests_index', ['filters' => $filters]);
+        $responseBuilder = $this->get('prestashop.bundle.grid.response_builder');
```
jolelievre: Same here, deprecation isn't enough?

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/LogsController.php
jolelievre: By the way the deprecate comment is missing

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/WebserviceController.php
```diff
@@ -163,20 +161,14 @@ public function editAction($webserviceKeyId, Request $request)
...
     {
-        $definitionFactory = $this->get('prestashop.core.grid.definition.factory.webservice_key');
-        $webserviceDefinition = $definitionFactory->getDefinition();
-
-        $gridFilterFormFactory = $this->get('prestashop.core.grid.filter.form_factory');
-        $searchParametersForm = $gridFilterFormFactory->create($webserviceDefinition);
-
-        $searchParametersForm->handleRequest($request);
-        $filters = [];
+        $responseBuilder = $this->get('prestashop.bundle.grid.response_builder');
```
jolelievre: Missing `@deprecated` comment, and modification shouldn't be required

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/MetaController.php
```diff
@@ -143,19 +145,14 @@ public function indexAction(MetaFilters $filters, Request $request)
...
-
-        $gridFilterFormFactory = $this->get('prestashop.core.grid.filter.form_factory');
-        $searchParametersForm = $gridFilterFormFactory->create($definitionFactory);
-        $searchParametersForm->handleRequest($request);
-
-        $filters = [];
-        if ($searchParametersForm->isSubmitted()) {
-            $filters = $searchParametersForm->getData();
-        }
+        $responseBuilder = $this->get('prestashop.bundle.grid.response_builder');
```
jolelievre: Modification not required

src/PrestaShopBundle/Controller/Admin/Improve/International/TaxController.php
```diff
@@ -124,19 +126,14 @@ public function saveOptionsAction(Request $request)
...
-        $searchParametersForm = $gridFilterFormFactory->create($definitionFactory);
-        $searchParametersForm->handleRequest($request);
-
-        $filters = [];
-        if ($searchParametersForm->isSubmitted()) {
-            $filters = $searchParametersForm->getData();
-        }
-
-        return $this->redirectToRoute('admin_taxes_index', ['filters' => $filters]);
+        $responseBuilder = $this->get('prestashop.bundle.grid.response_builder');
```
jolelievre: Modification not required

src/PrestaShopBundle/Resources/config/routing/admin/sell/catalog/categories.yml
```diff
@@ -12,16 +12,16 @@ admin_categories_index:
...
-    methods: [POST]
-    defaults:
-        _controller: PrestaShopBundle:Admin/Sell/Catalog/Category:searchGrid
-        _legacy_controller: AdminCategories
-        _legacy_link: AdminCategories:submitFiltercategory
-        gridDefinitionFactoryServiceId: prestashop.core.grid.definition.factory.category
-        redirectRoute: admin_categories_index
-        redirectQueryParamsToKeep:
-            - 'categoryId'
+  path: /
```
jolelievre: Changing the path might be a BC break (although I'm in favor of it)

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/LogsController.php
Progi1984: Because there is a dispatchHook in this case.

src/Core/Grid/Definition/Factory/RequestSqlGridDefinitionFactory.php
```diff
@@ -201,10 +203,11 @@ public function getFilters()
             ->add(
                 (new Filter('actions', SearchAndResetType::class))
                     ->setTypeOptions([
-                        'attr' => [
-                            'data-url' => $this->resetSearchUrl,
-                            'data-redirect' => $this->redirectionUrl,
+                        'reset_route' => 'admin_common_reset_search_by_filter_id',
```
atomiix, line 180: As @jolelievre stated, the constructor can be cleaned too.

src/Core/Grid/Definition/Factory/LogGridDefinitionFactory.php
Progi1984: Fixed

src/Core/Grid/Definition/Factory/RequestSqlGridDefinitionFactory.php
Progi1984: Fixed. Thanks

src/Core/Grid/Definition/Factory/WebserviceKeyDefinitionFactory.php
Progi1984: Fixed

src/PrestaShopBundle/Controller/Admin/Improve/International/TaxController.php
Progi1984: Fixed

src/PrestaShopBundle/Resources/config/routing/admin/sell/catalog/categories.yml
Progi1984: Fixed

atomiix (dismissed):

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/LogsController.php
jolelievre: Oh I get it, so you didn't need to modify the routing either and it uses this specific search controller All good! my bad

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php
```diff
@@ -53,6 +53,7 @@
 use PrestaShop\PrestaShop\Core\Grid\Definition\Factory\CategoryGridDefinitionFactory;
 use PrestaShop\PrestaShop\Core\Search\Filters\CategoryFilters;
 use PrestaShopBundle\Component\CsvResponse;
+use PrestaShopBundle\Controller\Admin\CommonController;
```
jolelievre, line 56: I don't think we need the use statement here, just for the deprecation comment
I'm even surprised CSFixer doesn't shout at this but it's probably because it takes the comment into account, anyway this is probably overkill but not worth blocking the PR either

jolelievre (dismissed):
Only one extra `use` statement But I leave you the choice to remove it or not

khouloudbelguith:
Hi @Progi1984  (edited)

I found an issue

1. Go to BO > Shop Parameters > SEO & Urls page
2. Try to filter by any filed
3. See error

https://drive.google.com/file/d/16KRdZ2A98DfR8b4InAiSFUePVVBwztgd/view?usp=sharing

PS: I found this regression: https://github.com/PrestaShop/PrestaShop/issues/22224 but it is reproduced without PR.
It could be fixed here?
Thanks!

PululuK:
> PS: I found this regression: #22224 but it is reproduced without PR.
> It could be fixed here?
> Thanks!

Hi @khouloudbelguith 

This bug was introduced here : https://github.com/PrestaShop/PrestaShop/pull/21075 (I'm sorry)
Will be fixed here : https://github.com/PrestaShop/PrestaShop/pull/22229

Nice catch ! Thank you so much @khouloudbelguith

Progi1984:
@khouloudbelguith If you test #22229 and if we merge it, I could rebase this PR after.

khouloudbelguith:
Hi @Progi1984,

About this PR: https://github.com/PrestaShop/PrestaShop/pull/22229, it is validated.

Thanks!

atomiix (dismissed):

matks (dismissed):
I'm a bit not confortable with the constructor BC breaks and the services BC break 😅  but I respect the work of @jolelievre and @Progi1984 👍

khouloudbelguith:
Hi @zuk3975, @Progi1984,

The issue of the SEO & URLs page is not fixed.

1. Go to BO > Shop Parameters > SEO & Urls page
2. Try to filter by any filed
3. See error

https://drive.google.com/file/d/1tLquW7Bb4wMNuFEEeam14gwWawPGBNEe/view?usp=sharing

Thanks!

Progi1984:
Thanks @khouloudbelguith. It's fixed.

atomiix (dismissed):

Progi1984:
Thanks @matks for the label

sowbiba (dismissed):
LGTM, unless you want to fix the comment by @jolelievre

sowbiba (approved):

atomiix (approved):

src/PrestaShopBundle/Controller/Admin/Configure/AdvancedParameters/WebserviceController.php
```diff
@@ -59,8 +59,23 @@ class WebserviceController extends FrameworkBundleAdminController
     public function indexAction(WebserviceKeyFilters $filters, Request $request)
     {
         $form = $this->getFormHandler()->getForm();
+        $gridWebserviceFactory = $this->get('prestashop.core.grid.factory.webservice_key');
```
jolelievre, line 62: One think I don't get here is that this is the exact same code as in the `renderPage` So what's the point of duplicating this code instead of using the existing method?

khouloudbelguith:
Hi @zuk3975, @Progi1984,

It is ok :heavy_check_mark: 
https://drive.google.com/file/d/1u1kTevSI8bAdc1OZTXn1DGBWCEVsrGKV/view?usp=sharing

Thank you!

PR #18164, opened 2020-03-16
Use hasBeenPaid instead of hasInvoice to enable refund feature

### LINKED ISSUE #18157: Refund buttons/actions should rely on paid status instead of invoice
#### Describe the bug

For now the condition to display and authorize the partial refund button, and the standard refund button is to check if order has an invoice. This was done because when the order is set a paid status the invoice is automatically generated. But the two conditions are not equivalent because an order can be manually generated, in which case you can then refund it even though it has not been paid.

#### Expected behavior

The buttons should be visible:
- when the order has been paid.
- when at least one payment is added to the order

The invoice is not a guaranty that any payment has been validated.

#### Steps to Reproduce

Steps to reproduce the **expected** behavior:

1. Go to an unpaid order
2. In the Documents Tab click on the "Generate an invoice" button
3. The refund buttons **must not** be visible
4. Add a payment to the order, the buttons are now visible

Second scenario

1. Go to another unpaid order
2. Switch the status to "Payment accepted"
3. The buttons are visible

Third scenario

1. Go to an unpaid order
2. Add a partial payment
3. The buttons are visible even though there is No invoice and no status as paid

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Refund buttons and actions/handlers were based on `Order::hasInvoice` to enable their feature, it is more logical to use `Order::hasBeenPaid` because an invoice can be generated manually without any payment previously made
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18157
| How to test?  | See issue, but basically create an invoice manually the refund buttons should not be visible They only appear once the status has been changed to a paid status

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18164)

### DIFF
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1817,6 +1817,16 @@ class OrderCore extends ObjectModel
     }
 
     /**
+     * Indicates if order has any associated payments.
+     *
+     * @return bool
+     */
+    public function hasPayments(): bool
+    {
+        return $this->getOrderPaymentCollection()->count() > 0;
+    }
+
+    /**
      * This method allows to add a payment to the current order.
      *
      * @since 1.5.0.1
--- a/src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
+++ b/src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
@@ -184,7 +184,7 @@ final class CancelOrderProductHandler extends AbstractOrderCommandHandler implem
      */
     private function checkOrderState(Order $order)
     {
-        if ((bool) $order->hasBeenPaid() === true) {
+        if ($order->hasBeenPaid() || $order->hasPayments()) {
             throw new InvalidOrderStateException(
                 InvalidOrderStateException::ALREADY_PAID,
                 'Can not cancel product on an order which is already paid'
--- a/src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
+++ b/src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
@@ -184,8 +184,11 @@ final class CancelOrderProductHandler extends AbstractOrderCommandHandler implem
      */
     private function checkOrderState(Order $order)
     {
-        if ($order->hasInvoice() !== false) {
-            throw new InvalidOrderStateException(InvalidOrderStateException::UNEXPECTED_INVOICE);
+        if ((bool) $order->hasBeenPaid() === true) {
+            throw new InvalidOrderStateException(
+                InvalidOrderStateException::ALREADY_PAID,
+                'Can not cancel product on an order which is already paid'
+            );
         }
     }
 }
--- a/src/Core/Domain/Order/Exception/InvalidOrderStateException.php
+++ b/src/Core/Domain/Order/Exception/InvalidOrderStateException.php
@@ -33,14 +33,14 @@ namespace PrestaShop\PrestaShop\Core\Domain\Order\Exception;
 class InvalidOrderStateException extends OrderException
 {
     /**
-     * Used when the order has no invoice (and it should have)
+     * Used when the order is not paid (and it should be)
      */
-    const INVOICE_NOT_FOUND = 1;
+    const NOT_PAID = 1;
 
     /**
-     * Used when the order has an invoice (and it should not)
+     * Used when the order is already paid (and it should not be)
      */
-    const UNEXPECTED_INVOICE = 2;
+    const ALREADY_PAID = 2;
 
     /**
      * Used when the order has not been delivered (and it should have)
--- a/src/Core/Domain/Order/QueryResult/OrderForViewing.php
+++ b/src/Core/Domain/Order/QueryResult/OrderForViewing.php
@@ -397,6 +397,14 @@ class OrderForViewing
     }
 
     /**
+     * @return bool
+     */
+    public function hasPayments(): bool
+    {
+        return count($this->payments->getPayments()) > 0;
+    }
+
+    /**
      * @return OrderMessagesForViewing
      */
     public function getMessages(): OrderMessagesForViewing
--- a/src/Core/Domain/Order/QueryResult/OrderForViewing.php
+++ b/src/Core/Domain/Order/QueryResult/OrderForViewing.php
@@ -131,6 +131,11 @@ class OrderForViewing
     /**
      * @var bool
      */
+    private $hasBeenPaid;
+
+    /**
+     * @var bool
+     */
     private $hasInvoice;
 
     /**
@@ -179,6 +184,7 @@ class OrderForViewing
      * @param string $taxMethod
      * @param bool $isTaxIncluded
      * @param bool $isValid
+     * @param bool $hasBeenPaid
      * @param bool $hasInvoice
      * @param bool $isDelivered
      * @param bool $isShipped
@@ -208,6 +214,7 @@ class OrderForViewing
         string $taxMethod,
         bool $isTaxIncluded,
         bool $isValid,
+        bool $hasBeenPaid,
         bool $hasInvoice,
         bool $isDelivered,
         bool $isShipped,
@@ -245,6 +252,7 @@ class OrderForViewing
         $this->isShipped = $isShipped;
         $this->prices = $prices;
         $this->isTaxIncluded = $isTaxIncluded;
+        $this->hasBeenPaid = $hasBeenPaid;
         $this->hasInvoice = $hasInvoice;
         $this->discounts = $discounts;
         $this->createdAt = $createdAt;
@@ -426,6 +434,14 @@ class OrderForViewing
     public function isTaxIncluded(): bool
     {
         return $this->isTaxIncluded;
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasBeenPaid(): bool
+    {
+        return $this->hasBeenPaid;
     }
 
     /**
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
@@ -1487,8 +1487,8 @@ class OrderController extends CommonController
                 'Admin.Notifications.Error'
             ),
             InvalidOrderStateException::class => [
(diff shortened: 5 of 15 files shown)


matks (commented):
Looks good.
👍 for the behat tests

What about ... partially paid orders 😅?

Progi1984 (dismissed):

jolelievre:
> Looks good.
> 👍 for the behat tests
> 
> What about ... partially paid orders 😅?

@matks I thought about it, I actually thought that partially paid orders would have their status updated But it turns out adding a payment manually only.. adds a payment 😅

The feature is based on ORDER STATUS ONLY, this can be perceived as a limitation because the merchant will have to perform an extra action (update the status manually), but I think it's safer to keep the feature based on this order status data.

jolelievre:
Unless we change this again, and base the feature toggling on the presence of payments and not on the status order any more
@MatShir @matks what do you think?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1486,8 +1486,8 @@ private function getErrorMessages(Exception $e)
                 'Admin.Notifications.Error'
             ),
             InvalidOrderStateException::class => [
-                InvalidOrderStateException::UNEXPECTED_INVOICE => $this->trans(
-                    'Invalid action: this order already has an invoice.',
+                InvalidOrderStateException::ALREADY_PAID => $this->trans(
+                    'Invalid action: this order is already paid.',
```
```suggestion
                    'Invalid action: this order has already been paid.',
```

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1498,8 +1498,8 @@ private function getErrorMessages(Exception $e)
                     'Invalid action: this order has already been delivered.',
                     'Admin.Notifications.Error'
                 ),
-                InvalidOrderStateException::INVOICE_NOT_FOUND => $this->trans(
-                    'Invalid action: this order has no invoice.',
+                InvalidOrderStateException::NOT_PAID => $this->trans(
+                    'Invalid action: this order is not paid.',
```
```suggestion
                    'Invalid action: this order has not been paid.',
```

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: Damn.. that's what I originally used I was *this close* 👌 to have it right on the first try!! 😛

jolelievre: Done

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: Done

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
```diff
@@ -184,8 +184,11 @@ private function checkInput(CancelOrderProductCommand $command)
      */
     private function checkOrderState(Order $order)
     {
-        if ($order->hasInvoice() !== false) {
-            throw new InvalidOrderStateException(InvalidOrderStateException::UNEXPECTED_INVOICE);
+        if ((bool) $order->hasBeenPaid() === true) {
```
Arman-Hosseini:
```suggestion
        if ($order->hasBeenPaid()) {
```

src/Core/Domain/Order/Exception/InvalidOrderStateException.php
```diff
@@ -33,14 +33,14 @@
...
      */
-    const INVOICE_NOT_FOUND = 1;
+    const NOT_PAID = 1;
 
     /**
-     * Used when the order has an invoice (and it should not)
+     * Used when the order is already paid (and it should not be)
      */
-    const UNEXPECTED_INVOICE = 2;
+    const ALREADY_PAID = 2;
```
Arman-Hosseini, line 43: BC may break.

Progi1984: It's not a BC... This class and thoses constants don't exist in 176x.

jolelievre: Indeed, this class has been introduced in 177, so we can still modify it without BC break for now

matks:
> Unless we change this again, and base the feature toggling on the presence of payments and not on the status order any more
> @MatShir @matks what do you think?

This is a PM decision, as for me I dont know 😄 

Ping @colinegin @marionf

MatShir:
The payment tab is not used a lot by merchants, only if they rely on it. It is not a reliable criterion. I suggest that we stick on the hasPaid or hasInvoice.

Before, the merchant has to generate manually the invoice to have the button displayed. 

The status "payment accepted" does the same, generates the invoice and flags the order hasPaid. IMO if the order is half paid, it usually means that they are processing a B2B order. So the merchant can deliver the product even if it is haft paid. And to go further on the delivery process, the merchant has to consider its order as paid in PrestaShop. So I guess PrestaShop handles the partial payment and total payment the same way. IMO, it makes sense to keep hasPaid. WDYT? @jolelievre @matks

Matt75:
Payment tab is used for merchant use third party payment modules like PrestaShop Checkout to get transaction_id
But sometime, Order can be paid but without OrderPayment associated.
Caused by a bug when OrderPayment is saved or maybe others troubles in PaymentModule::validateOrder()
We have some cases with PrestaShop Checkout
So It is not a reliable criterion.
I recommended to use `$order->hasBeenPaid()`

MatShir:
thanks @Matt75 for your clarification,

jolelievre:
So they way I understand it we have two use cases:
- partially paid order, so OrderPayment are present but not the status hasBeenPaid
- the OrderPayment (for some reason) are absent, but the status hasBeenPaid can be set to allow refunding

So maybe the solution is to use both those information to enable the refund feature, something like `if ($order->hasBeenPaid() || $order->getOrderPaymentCollection()->count() > 0)`

@Matt75 what do you think of this compromise? Do you think it could cause other unexpected problems?

Matt75:
I think yes, but I'm not comfortable with case partially paid.
I never use it and never see usage of it so I don't know how it work.
I checked in our modules of payment Nx but I see only usage of OrderState paid...
Maybe it's a manual OrderState ?
We can ask some others community developers of modules Payment Nx to check if they know a usecase.

jolelievre:
I guess ideally, there should be an order state `Partially paid` Module can add their own status so maybe some use this technic

But anyway, even natively in the core it is possible to add payments manually, and the status is not updated in those cases So you can't refund even though some payments have been made, which can be a problem I think

That's why I proposed using both conditions as an activation criteria for refund feature

Matt75:
Yes I think it's the best compromise

jolelievre:
Ok I updated the feature, so we can now make refunds even if order status is not paid
And even for partially paid orders

However 😅 @MatShir this brings new questions now! How do we handle the max allowed refund for partially paid orders?
But maybe it should be dealt with in another issue?

classes/order/Order.php
```diff
@@ -1816,6 +1816,16 @@ public function getOrderPaymentCollection()
...
     }
 
+    /**
+     * Indicates if order has any associated payments.
+     *
+     * @return bool
+     */
+    public function hasPayments(): bool
+    {
+        return $this->getOrderPaymentCollection()->count() > 0;
```
Arman-Hosseini, line 1826:
```suggestion
        return $this->getOrderPaymentCollection()->count();
```

src/Core/Domain/Order/QueryResult/OrderForViewing.php
```diff
@@ -396,6 +396,14 @@ public function getPayments(): OrderPaymentsForViewing
         return $this->payments;
     }
 
+    /**
+     * @return bool
+     */
+    public function hasPayments(): bool
+    {
+        return count($this->payments->getPayments()) > 0;
```
Arman-Hosseini, line 404:
```suggestion
        return count($this->payments->getPayments());
```

jolelievre: Even though a count can't be negative and the code you propose will also work I prefer to be explicit Besides comparing to 0 has the advantage to be directly a boolean wheres your suggestion should actually look like `return (bool) count($this->payments->getPayments());`

Arman-Hosseini: Be explicit is great but when you declare a return type in php 7, it will eventually be typecasting.
Your code look like this: `return (bool) ((bool) (count($this->payments->getPayments()) > 0));`

matks (approved):

classes/order/Order.php
matthieu-rolland: this modification will not return a boolean anymore 🤔

matthieu-rolland (approved):

matks:
> Ok I updated the feature, so we can now make refunds even if order status is not paid
> And even for partially paid orders
> 
> However 😅 @MatShir this brings new questions now! How do we handle the max allowed refund for partially paid orders?
> But maybe it should be dealt with in another issue?

In order for this topic to move on, I suggest we follow @jolelievre suggestion: lets discuss, specify and fix this other topic in another issue

EDIT: issue created https://github.com/PrestaShop/PrestaShop/issues/18286

PierreRambaud (approved):

SD1982:
Hi @jolelievre  I have reproduced the steps described in #18157 and indeed now with the PR the refund button doesn't appear when the invoice is generated but it appear when a paiement is added and the order is still in "await check paiement": 
https://drive.google.com/file/d/1dRtOu57iQTw0OwWJa42SL1YUeSDkIFbW/view?usp=sharing

jolelievre:
@SD1982 Sorry the specs of this feature have been changed but I didn't update the issue
The description of the issue is now up to date with three possible scenarii

SD1982:
This PR work as expected ! Thanks @jolelievre

Progi1984:
Thanks @jolelievre

PR #18166, opened 2020-03-16
Use Tools::deleteFile() instead of unlink() method

Arman-Hosseini:
| Questions     | Answers
| Branch?       | develop
| Description?  | Use Tools::deleteFile() instead of unlink() method
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | -

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18166)

### DIFF
--- a/classes/Attachment.php
+++ b/classes/Attachment.php
@@ -83,7 +83,7 @@ class AttachmentCore extends ObjectModel
      */
     public function delete()
     {
-        @unlink(_PS_DOWNLOAD_DIR_ . $this->file);
+        @Tools::deleteFile(_PS_DOWNLOAD_DIR_ . $this->file);
 
         $sql = new DbQuery();
         $sql->select('pa.`id_product`');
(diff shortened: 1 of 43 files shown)


matks:
Hi @Arman-Hosseini I must warn you that this PR seems already very big and I'm especially worried about testing it. It seems the "How to test" description will look like "Check that all areas of PrestaShop where we delete files work as expected" and this seems like the kind of PR very hard to be sure it's 100% tested. It would be easier to test if you split it into multiple small PRs, and each PR modifies delete statements in a very accurate location.

I say this, but the best person to tell this would be QA specialist @Robin-Fischer-PS or @SimonGrn

Arman-Hosseini:
Hi @matks 
don't worry. I'll do the same ;)
Thanks for the review.

SimonGrn:
Thanks @matks , that's actually true. It would be easier to split it. Be sure to write in the "how to test" what you changed and how we are supposed to test it. Your PRs will be validated faster.

PR #18167, opened 2020-03-16
Avoiding to call several times the actionProductUpdate hook

bertoferrero:
If we save or update a product with any combination, the for loop that calls procdessProductAttribute will finally execute the line Hook::exec('actionProductUpdate', array('id_product' => (int) $product->id, 'product' => $product)); one time per each combination. It can be understood as a great performance problem if the user has a very big module that, for example, regenerates an index per each call.

The same Hook is also called in update and uptadeWS Product's functions, I do not know if could be better to just delete the line.

| Questions     | Answers
| Branch?       | develop 
| Description?  | Avoiding to call many times the same hook
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Just save a product with combinations.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18167)

### DIFF
--- a/src/Adapter/Product/AdminProductWrapper.php
+++ b/src/Adapter/Product/AdminProductWrapper.php
@@ -215,7 +215,7 @@ class AdminProductWrapper
         static $hookExecuted = [];
         if (!in_array($product->id, $hookExecuted)) {
             $hookExecuted[] = $product->id;
-            Hook::exec('actionProductUpdate', array('id_product' => (int) $product->id, 'product' => $product));
+            Hook::exec('actionProductUpdate', ['id_product' => (int) $product->id, 'product' => $product]);
         }
     }
 
--- a/src/Adapter/Product/AdminProductWrapper.php
+++ b/src/Adapter/Product/AdminProductWrapper.php
@@ -212,7 +212,11 @@ class AdminProductWrapper
     {
         // Hook triggered by legacy code below: actionUpdateQuantity('id_product', 'id_product_attribute', 'quantity')
         StockAvailable::setQuantity((int) $product->id, $forAttributeId, $quantity);
-        Hook::exec('actionProductUpdate', ['id_product' => (int) $product->id, 'product' => $product]);
+        static $hookExecuted = [];
+        if (!in_array($product->id, $hookExecuted)) {
+            $hookExecuted[] = $product->id;
+            Hook::exec('actionProductUpdate', array('id_product' => (int) $product->id, 'product' => $product));
+        }
     }
 
     /**


src/Adapter/Product/AdminProductWrapper.php
```diff
@@ -212,7 +212,11 @@ public function processQuantityUpdate(Product $product, $quantity, $forAttribute
     {
         // Hook triggered by legacy code below: actionUpdateQuantity('id_product', 'id_product_attribute', 'quantity')
         StockAvailable::setQuantity((int) $product->id, $forAttributeId, $quantity);
-        Hook::exec('actionProductUpdate', ['id_product' => (int) $product->id, 'product' => $product]);
+        static $hookExecuted = [];
+        if (!in_array($product->id, $hookExecuted)) {
+            $hookExecuted[] = $product->id;
+            Hook::exec('actionProductUpdate', array('id_product' => (int) $product->id, 'product' => $product));
```
Arman-Hosseini:
```suggestion
            Hook::exec('actionProductUpdate', ['id_product' => (int) $product->id, 'product' => $product]);
```

src/Adapter/Product/AdminProductWrapper.php
```diff
@@ -212,7 +212,11 @@ public function processQuantityUpdate(Product $product, $quantity, $forAttribute
     {
         // Hook triggered by legacy code below: actionUpdateQuantity('id_product', 'id_product_attribute', 'quantity')
         StockAvailable::setQuantity((int) $product->id, $forAttributeId, $quantity);
-        Hook::exec('actionProductUpdate', ['id_product' => (int) $product->id, 'product' => $product]);
+        static $hookExecuted = [];
```
Progi1984, line 215: Please put out the method the attribute and define it a description. Thanks

bertoferrero: sure? the attribute $hookExecuted is just a variable. Its scope is just this method. The use of the definition static here is just for notifying the compiler to recycle the same memory for every execution and it is totally accepted.
Here is explained perfectly: https://www.php.net/manual/en/language.variables.scope.php (search for Example #5 Example use of static variables)
I think that putting out this attribute it will become into garbage for the class, it is totally useless for any other code line.
However, I may be wrong. If you confirm me I will put this attribute outside

matks: Hi @prestaquality the idea is quite good however I see one possible issue / limitation: testing.

If we decide tomorrow to write a set of unit tests for this class, this static variable will be shared among them, which will break the necessary isolation property of these unit tests. It could make our tests very hard to maintain.

So indeed having the variable as a class instance will be a bigger memory consumption, but it will make the code a little more maintainable. It seems like a nice compromise 🤔wdyt ? An array of IDs is not something we should be afraid of I think.

Btw if we are talking about optimization, maybe using `array_key_exists()` or `isset()` is faster than `in_array()` ? (so we would transform `$hookExecuted[] = $product->id` into `$hookExecuted[$product->id] = true;`)

bertoferrero: OMG you are right with the in_array statement... but... before to find a perfect way for this. What if this hook calling were removed? I created this fix because I am not really sure if this calling is here because of any legacy trouble... The hook productUpdate is also call in update() function inside of the Product entity (and also in updateWS). And there is other hook for notifiying when stock is changed fired by StockAvailable... so, may we are discussing here for solve this performance trouble finding the perfect way and it is just removing the line.

PierreRambaud:
Hi @prestaquality ,

Thanks!

PR #18182, opened 2020-03-17
CO : Correct phpDoc

djbuch:
| Questions     | Answers
| Branch?       | develop 
| Description?  | Simply correct phpDoc and ensure the return value is an int.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18182)

### DIFF
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1131,7 +1131,7 @@ class OrderCore extends ObjectModel
                     . Shop::addSqlRestriction();
         $result = Db::getInstance(_PS_USE_SQL_SLAVE_)->getRow($sql);
 
-        return isset($result['nb']) ? (int)$result['nb'] : 0;
+        return isset($result['nb']) ? (int) $result['nb'] : 0;
     }
 
     /**
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1121,7 +1121,7 @@ class OrderCore extends ObjectModel
      *
      * @param int $id_customer Customer id
      *
-     * @return array Customer orders number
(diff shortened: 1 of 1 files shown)


classes/order/Order.php
```diff
@@ -1131,7 +1131,7 @@ public static function getCustomerNbOrders($id_customer)
                     . Shop::addSqlRestriction();
         $result = Db::getInstance(_PS_USE_SQL_SLAVE_)->getRow($sql);
 
-        return isset($result['nb']) ? $result['nb'] : 0;
+        return isset($result['nb']) ? (int)$result['nb'] : 0;
```
Progi1984:
```suggestion
        return isset($result['nb']) ? (int) $result['nb'] : 0;
```

Progi1984: Small fix. But :ok_hand: for me.

PierreRambaud (dismissed):
PHP Cs fixer must be ran

Progi1984 (approved):

matks (approved):

Progi1984:
Only PHPDoc. No need QA.

matks:
> Only PHPDoc. No need QA.

Plus a type casting 😄this could break some bad code.
However I checked for this function usage in standard prestashop install and found only usages into "Search (customers) page" and "GDPR Customer Search page".

Progi1984:
Fail of my part :(

PR #18185, opened 2020-03-17
Add a few phpdoc

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add a few phpdoc
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18185)

### DIFF
--- a/classes/assets/CccReducer.php
+++ b/classes/assets/CccReducer.php
@@ -29,11 +29,18 @@ use Symfony\Component\Filesystem\Filesystem;
 
 class CccReducerCore
 {
+    /** @var string */
     private $cacheDir;
+    /** @var Filesystem */
     protected $filesystem;
 
     use PrestaShop\PrestaShop\Adapter\Assets\AssetUrlGeneratorTrait;
 
+    /**
+     * @param string $cacheDir
+     * @param ConfigurationInterface $configuration
+     * @param Filesystem $filesystem
+     */
     public function __construct($cacheDir, ConfigurationInterface $configuration, Filesystem $filesystem)
     {
         $this->cacheDir = $cacheDir;
@@ -45,6 +52,11 @@ class CccReducerCore
         }
     }
 
+    /**
+     * @param array $cssFileList
+     *
+     * @return array same list, reduced
+     */
     public function reduceCss($cssFileList)
     {
         $files = [];
@@ -81,6 +93,11 @@ class CccReducerCore
         return $cssFileList;
     }
 
+    /**
+     * @param array $jsFileList
+     *
+     * @return array same list, reduced
+     */
     public function reduceJs($jsFileList)
     {
         foreach ($jsFileList as $position => &$list) {
@@ -127,6 +144,11 @@ class CccReducerCore
         return $jsFileList;
     }
 
+    /**
+     * @param string[] $files
+     *
+     * @return string
+     */
     private function getFileNameIdentifierFromList(array $files)
     {
         return substr(sha1(implode('|', $files)), 0, 6);
--- a/classes/assets/CccReducer.php
+++ b/classes/assets/CccReducer.php
@@ -57,7 +57,7 @@ class CccReducerCore
      *
      * @return array Same list, reduced
      */
-    public function reduceCss(array $cssFileList)
+    public function reduceCss($cssFileList)
     {
         $files = [];
         foreach ($cssFileList['external'] as $key => &$css) {
@@ -98,7 +98,7 @@ class CccReducerCore
      *
      * @return array Same list, reduced
      */
-    public function reduceJs(array $jsFileList)
+    public function reduceJs($jsFileList)
     {
         foreach ($jsFileList as $position => &$list) {
             $files = [];
--- a/classes/assets/CccReducer.php
+++ b/classes/assets/CccReducer.php
@@ -55,9 +55,9 @@ class CccReducerCore
     /**
      * @param array $cssFileList
      *
-     * @return array same list, reduced
+     * @return array Same list, reduced
      */
-    public function reduceCss($cssFileList)
+    public function reduceCss(array $cssFileList)
     {
         $files = [];
         foreach ($cssFileList['external'] as $key => &$css) {
@@ -96,9 +96,9 @@ class CccReducerCore
     /**
      * @param array $jsFileList
      *
-     * @return array same list, reduced
+     * @return array Same list, reduced
      */
-    public function reduceJs($jsFileList)
+    public function reduceJs(array $jsFileList)
     {
         foreach ($jsFileList as $position => &$list) {
             $files = [];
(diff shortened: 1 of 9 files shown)


classes/assets/CccReducer.php
```diff
@@ -45,6 +52,11 @@ public function __construct($cacheDir, ConfigurationInterface $configuration, Fi
         }
     }
 
+    /**
+     * @param array $cssFileList
+     *
+     * @return array same list, reduced
+     */
     public function reduceCss($cssFileList)
```
Arman-Hosseini, line 60:
```suggestion
    public function reduceCss(array $cssFileList)
```

classes/assets/CccReducer.php
```diff
@@ -81,6 +93,11 @@ public function reduceCss($cssFileList)
         return $cssFileList;
     }
 
+    /**
+     * @param array $jsFileList
+     *
+     * @return array same list, reduced
+     */
     public function reduceJs($jsFileList)
```
Arman-Hosseini, line 101:
```suggestion
    public function reduceJs(array $jsFileList)
```

classes/assets/CccReducer.php
```diff
@@ -45,6 +52,11 @@ public function __construct($cacheDir, ConfigurationInterface $configuration, Fi
         }
     }
 
+    /**
+     * @param array $cssFileList
+     *
+     * @return array same list, reduced
```
Arman-Hosseini:
```suggestion
     * @return array Same list, reduced
```

classes/assets/CccReducer.php
```diff
@@ -81,6 +93,11 @@ public function reduceCss($cssFileList)
         return $cssFileList;
     }
 
+    /**
+     * @param array $jsFileList
+     *
+     * @return array same list, reduced
```
Arman-Hosseini:
```suggestion
     * @return array Same list, reduced
```

classes/assets/CssMinifier.php
```diff
@@ -28,6 +28,12 @@
 
 class CssMinifierCore
 {
+    /**
+     * @param string[] $files
+     * @param string $destination
+     *
+     * @return string minified data
```
Arman-Hosseini:
```suggestion
     * @return string Minified data
```

Arman-Hosseini:
Doc block descriptions should also be completed.
Of course with multiple PRs.

matks:
> Doc block descriptions should also be completed.
> Of course with multiple PRs.

I dont know if I'll have either the time or the courage 😅 

I think arguments typehints and return typehint is the 1st needed item for people using these classes. Description can come 2nd, with throws.

Arman-Hosseini:
> > Doc block descriptions should also be completed.
> > Of course with multiple PRs.
> 
> I dont know if I'll have either the time or the courage sweat_smile
> 
> I think arguments typehints and return typehint is the 1st needed item for people using these classes. Description can come 2nd, with throws.

Maybe we need someone who has the courage to do it :grin:

classes/assets/CssMinifier.php
```diff
@@ -28,6 +28,12 @@
...
+     */
     public static function minify(array $files, $destination)
```
Progi1984, line 37:
```suggestion
    public static function minify(array $files, string $destination)
```

matks:
Applied @Arman-Hosseini suggestions

classes/assets/CccReducer.php
```diff
@@ -45,7 +52,12 @@ public function __construct($cacheDir, ConfigurationInterface $configuration, Fi
         }
     }
 
-    public function reduceCss($cssFileList)
+    /**
+     * @param array $cssFileList
+     *
+     * @return array Same list, reduced
+     */
+    public function reduceCss(array $cssFileList)
```
PierreRambaud: It's a BC :thinking: 
What if the parameter was `ArrayAccess` or something similar?

classes/assets/CssMinifier.php
```diff
@@ -28,7 +28,13 @@
...
 class CssMinifierCore
 {
-    public static function minify(array $files, $destination)
+    /**
+     * @param string[] $files
+     * @param string $destination
+     *
+     * @return string Minified data
+     */
+    public static function minify(array $files, string $destination)
```
PierreRambaud: same here, it's a BC, could have been an object with `__toString`

classes/assets/CccReducer.php
matks: Ay 😅 I didnt think about `ArrayAccess`

classes/assets/CssMinifier.php
matks: You mean that an object, when `foreach` is applied to it, if it has `__toString()` function, this will return `__toString()` result and it will loop over it ?

... seriously 😭 😅 ?

PierreRambaud: I don't know, but as @eternoendless said, changing the signature is a BC :/

Matt75: Have a lot of phpdoc update in stock on my side.
Modules code analysis with phpstan return to many error due to bad phpdoc
Prefer to not make code changes to avoid tests from QA.
So what to do ?
Not change type of parameters and returns and add comment to explain expected type ?

matks: @Matt75 We should probably ignore the issues from the past using [PHPStan baseline](https://medium.com/@ondrejmirtes/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard-e77d815a5dff). Only new code should be 100% bugfree according to phpstan.

Because legacy code is too big, too messy and sometimes (like here) cannot be updated because of SemVer 😭

Matt75: Ok but phpdoc is useful for all developers have to interact with legacy classes, so I would like to share my personal work on it.
I just want to know what to do for type not matches signature, should I use mixed and add a comment to explain expected type ?
I made a PR for Product class but I have a lot of PR in stock ^^

matks: > should I use mixed and add a comment to explain expected type ?

I think this is the best solution. We are sure there is no BC break and you can explain what content is "most of the time".

classes/assets/CccReducer.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/18185/commits/e495740557e289cac08d92f6319badd4233dbd63

classes/assets/CssMinifier.php
matks: @PierreRambaud BC break fixed in https://github.com/PrestaShop/PrestaShop/pull/18185/commits/e495740557e289cac08d92f6319badd4233dbd63

PierreRambaud (approved):

Progi1984 (approved):

PierreRambaud:
Thanks @matks

Issue #10258, opened 2018-09-03
Allow to choose a different language for order documents


CArlvJ1N0N8:
**Describe the bug**

Hello, 
I am using Prestashop in association with Amazon marketplace module. 
So Amazon orders are created inside of Prestashop by this module every time an order is available. 
Client is created at the same time. 
I have checked with the developer for this module and the client language is also imported. 
Although if i go to the client i can't see his language. 
My issue is i am connecting to the BO using french language in order to print the shipping notes for the Amazon orders. but all of them are printed in french. Even the emails sent to the customers are always in french. Although customers are US based or Canadian based (english talking).
I even changed the default language for my store to English. 
What do i need to do to be able to print or send emails to customers in their language.
How can i check a customer language? (maybe a feature to show customer language on customer page?)

---
**Edited on October 23rd by Louise B.**

A. ✅ If as a customer you log to the system and view your invoices/delivery slip it definitely needs to be in the customers language.
--> this should use the language used by the customer in FO (works OK already)

**Specifications:** https://github.com/PrestaShop/prestashop-specs/pull/161/files

B. ➡️ But if the dispatcher is preparing the shipping and printing the invoice/delivery slip, then does it need to be in his language or in the customer language since it will be put in the colis to be shipped.
--> we could provide a new option to let the merchant decide if he wants to use customer language or employee language to download delivery slips and invoices when preparing orders. This option is to be added in Shop parameters > Order settings with the following wording:

**Label**: `Default language for order documents` in `Admin.Shopparameters.Feature`
**Tooltip**: `Select in what language you want to download invoices and delivery slips.` in `Admin.Shopparameters.Help`
**Dropdown selector**: `Customer language` / `Logistician language` in `Admin.Shopparameters.Feature`

By default, the customer language is selected.

**Specifications:** https://github.com/PrestaShop/prestashop-specs/pull/152/commits/464be4a0ee88da1bbc3659e487d8640ac0726ee7

C. ✅ Then if it is an accountancy person who is printing the invoices/delivery slips i can understand it is printed in the accountant's language.
--> this should use the language used by the employee in BO (works OK already)

**Specifications:** https://github.com/PrestaShop/prestashop-specs/pull/161/files

**Additionnal information**

PrestaShop version: 1.7.4.2
PHP version: 7.0.27-0+deb9u1

CArlvJ1N0N8:
Sorry a little error in what i said... 
[...]

PR #18188, opened 2020-03-17
Fix language used for PDF sent by mail

Matt75:
| Questions     | Answers
| Branch?       | develop
| Description?  | Invoice or DeliverySlip sent by mail are not using Language used for Order
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10258 (Partial : fixed only for PDF sent by mail)
| How to test?  |  Change OrderState to Payment accepted or change OrderState to Processing in progress, PDF sent by mail use Language used in BO, not Language used in Order.

I use a way without BC Break

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18188)

### DIFF
--- a/classes/pdf/PDF.php
+++ b/classes/pdf/PDF.php
@@ -119,6 +119,7 @@ class PDFCore
             if (!$template) {
                 continue;
             }
+            $template->setLangId((int) $this->language->id);
 
             if (empty($this->filename)) {
                 $this->filename = $template->getFilename();
--- a/classes/pdf/PDF.php
+++ b/classes/pdf/PDF.php
@@ -35,11 +35,6 @@ class PDFCore
     public $template;
     public $send_bulk_flag = false;
 
-    /**
-     * @var Language
-     */
-    private $language;
-
     const TEMPLATE_INVOICE = 'Invoice';
     const TEMPLATE_ORDER_RETURN = 'OrderReturn';
     const TEMPLATE_ORDER_SLIP = 'OrderSlip';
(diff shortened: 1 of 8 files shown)


classes/pdf/PDF.php
```diff
@@ -105,7 +111,7 @@ public function __construct($objects, $template, $smarty, $orientation = 'P')
     public function render($display = true)
     {
         $render = false;
-        $this->pdf_renderer->setFontForLang(Context::getContext()->language->iso_code);
+        $this->pdf_renderer->setLanguage($this->language);
```
Progi1984: Personnally, I will setRTL & setFontForLang... because it's not the responsability to PDFGenerator to use Language. What do you think @PrestaShop/prestashop-core-developers ?

Matt75:
Problem with translations is here : https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Translate.php#L299
Static function, no existing params...

I think we have to change Language in Context before generate PDF then restore previous Language... 🤭

Progi1984 (dismissed):

matthieu-rolland (approved):

Progi1984 (approved):

PierreRambaud (approved):

SD1982:
Hello @Matt75  Can you please be more explicit about the "how to test" ? 

Because what you are describing here : 

capture d'écran_3

Is already the behavior I get without applying PR

Matt75:
@SD1982 In FO use French language
Make an Order in french

In BO use English language
Change OrderState to `Payment accepted`
Mail sent to Customer have attachment in English (BO language) instead of Customer language used to make the Order.

With this change, mail attachement should be in language used by customer on FO

SD1982:
Hi @Matt75  Thank you for the additional informations . After applying the PR the pdf sent with the mail is still in BO language . I have tried with the FO in french and in spanish and the BO in english.

Matt75:
Hummm `$_LANGPDF` is a Superglobal... already initialized with translations of BO... 😱
Damned 😩

PR #18192, opened 2020-03-18
Fix end EOL and convert CRLF to LF in all text files

mvorisek:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fix missing EOL
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | CI tests should pass

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18192)


Progi1984:
@mvorisek Hello, why do you change this file ? Thanks.

mvorisek:
> 
> 
> @mvorisek Hello, why do you change this file ? Thanks.

I received an warning from my dev. setup as this is a text file not properly terminated by EOL which is an industry standard for text files. All other text files are fine except this.

matks:
I think it's alright 😅because this XML file is not used for the code. It's the XML data used to generate [this image](https://github.com/PrestaShop/PrestaShop/raw/develop/tools/build/pre-installer-diagram.jpg) using www.draw.io (the source is provided in order to help developers re-generate and update the image if needed).

The XML was generated by draw.io not a developer, so if draw.io chooses not to add an empty line at the end, maybe it's needed for draw.io 🤔

mvorisek:
> The XML was generated by draw.io not a developer, so if draw.io chooses not to add an empty line at the end, maybe it's needed for draw.io 🤔

It is xml, so formatting does not matter like in json :)

PierreRambaud (changes requested):
I don't see any reason to update this file, it's generated by an external tool :thinking:

mvorisek:
@PierreRambaud The issue is that if I clone the repo, the file is changed by our company git filter - our setup fixes all text files, i.e. it adds end NL which is anoying. Why not to fix this and store this file in a canonical form? It's like saying you do care about code syntax ;-)

PierreRambaud:
I care about syntax, but not when the code from a generated tool.
If you have problem with this file, you should configure your tool to don't check files in `tools` for example. 
If tomorrow we generate this file one more time, even if we change nothing, we'll have a difference, the EOL will disappear. But we'll not see it directly because we don't have your git filter tool, and we are not using the same tools as your company. :thinking: 
Maybe time to add an equivalent of .gitignore for your company tool :wink:

mvorisek:
@PierreRambaud I will update this PR like this one - https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4753

mvorisek:
How can I obtain a write/maintainer access to PrestaShop so I can participate on PRs like this?

mvorisek:
PR complete, I fixed all source text files, minimized/generated files are excluded.

Feel free to update/commit the exclude list if needed.

PierreRambaud:
Hi @mvorisek, I'm sorry but as I said ealier, this doesn't bring any values and it's not test by the CI.
Some updated files came from dependencies and we don't want to change the code from them.
I'm against doing this kind of work.
If the problem from a git configuration, don't forget you can add a global git config on your system by using `~/.gitconfig` :wink:

PierreRambaud (changes requested):
Against this change

mvorisek:
> Some updated files came from dependencies and we don't want to change the code from them.
> I'm against doing this kind of work.
> 
@PierreRambaud I agree, please add these locations and amend the first commit. I do not want to change dependencies, but normalize source files within this project.

PierreRambaud:
@mvorisek Since we have no automated checks on these files, there is no interested to merge this change :/

mvorisek:
@PierreRambaud do I understand it correctly that you have concerns about maintaining the clean state by CI tests? If the tests will be added, is normalizing EOL to LF and requiring end EOL welcomed for this project?

Of course, for source, non generated files only.

There are a lot of files in the following directories:
- `js/tiny_mce/*`
- `js/jquery/plugins/*`
what is the origin of these files? Are they 1:1 to source public source?

PierreRambaud:
@mvorisek Yes, it's not tested, and regarding what you've changed, it's principally related to external dependencies.
For these directories, I don't know where they  come from, legacy you know :sweat_smile:

PierreRambaud:
Hi @mvorisek ,

Thanks!

PR #18204, opened 2020-03-18
BO : Order > Products : The block discounts was not displayed after adding a product

### LINKED ISSUE #13880: Bug in displayed discount when adding product (linked to a coupon) to an existing order from BO

Progi1984:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When you add a product (which add a discount), the discount block is not added & product can't be deleted
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Partially fix #13880
| How to test?  | See https://github.com/PrestaShop/PrestaShop/issues/13880#issuecomment-595189713<br>0. Build assets from **admin-dev/themes/new-theme**<br>1. Create a new Product (Price 100€, Quantity 100)<br>2. Create a new cart rule<br>2.1. has no code<br>2.2. product selection a product with 100 euro tax included<br>2.3. discount amount of 500 EUR tax included<br>3. add the product to the old order<br>4. **Before** The discount block is not displayed<br>4. **Before** A product can't be deleted<br>5. **After** The discount block is displayed<br>5. **After** A product can be deleted<br>

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18204)


matthieu-rolland (dismissed):

admin-dev/themes/new-theme/js/pages/order/view/order-prices-refresher.js
```diff
@@ -37,6 +37,8 @@ export default class OrderPricesRefresher {
     $.ajax(this.router.generate('admin_orders_get_prices', {orderId}))
       .then((response) => {
         $(OrderViewPageMap.orderTotal).text(response.orderTotalFormatted);
+        $(OrderViewPageMap.orderDiscountsTotal).text(response.discountsAmountFormatted);
+        $(OrderViewPageMap.orderDiscountsTotal).parent().toggleClass('d-none', !response.discountsAmountDisplayed);
```
PierreRambaud: Could you use something different than a simple `.parent()`, better to target a class instead. Otherwise if we change the template, we're f***ed :sweat_smile:

Progi1984: Done :)

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1277,12 +1278,16 @@ public function getPricesAction(int $orderId): JsonResponse
...
+        $zeroNumber = new Number('0');
 
         return $this->json([
-            'orderTotalFormatted' => $orderForViewing->getPrices()->getTotalAmountFormatted(),
-            'productsTotalFormatted' => $orderForViewing->getPrices()->getProductsPriceFormatted(),
-            'shippingTotalFormatted' => $orderForViewing->getPrices()->getShippingPriceFormatted(),
-            'taxesTotalFormatted' => $orderForViewing->getPrices()->getTaxesAmountFormatted(),
+            'orderTotalFormatted' => $orderForViewingPrices->getTotalAmountFormatted(),
+            'discountsAmountFormatted' => $orderForViewingPrices->getDiscountsAmountFormatted(),
+            'discountsAmountDisplayed' => $orderForViewingPrices->getDiscountsAmountRaw()->isGreaterThan($zeroNumber),
```
jolelievre: Maybe you should simply return `getDiscountsAmountRaw` It will provide more info than just a boolean (the test can be performed in js).
All those formatted prices are a bad idea in my opinion.. Very often we need to return the formatted price along with the raw one.. We have a CLDR component on front side as well but it's never been used sadly making the whole PHP code more heavy
But that's not the question here anyway 😅

jolelievre (dismissed):

matthieu-rolland (dismissed):

Robin-Fischer-PS:
Hi @Progi1984 !

With your fix, the discount block is not added when I add the product (even after refresh).

Other issue : It is not longer possible to delete the product from the order (it works without the cart rule).

See screenrecord : 

https://drive.google.com/file/d/1DZxWt9FcMGX_t6GLlGp9DVt3pjr6hLm6/view?usp=sharing

Thanks !

Progi1984:
@Robin-Fischer-PS OK. I didn't understand which Block it was.

matks (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1267,6 +1268,25 @@ public function deleteProductAction(int $orderId, int $orderDetailId): JsonRespo
...
+     * @param int $orderId
+     *
+     * @return JsonResponse
+     */
+    public function getDiscountsAction(int $orderId): Response
+    {
+        /** @var OrderForViewing $orderForViewing */
+        $orderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId));
+
+        dump($orderForViewing);
```
sowbiba: forgotten ?

sowbiba: otherwise, seems good to me

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
```diff
@@ -26,41 +26,70 @@
...
+        $this->contextStateManager = $contextStateManager;
+    }
     /**
      * {@inheritdoc}
      */
     public function handle(DeleteProductFromOrderCommand $command)
     {
         $orderDetail = new OrderDetail($command->getOrderDetailId());
         $order = new Order($command->getOrderId()->getValue());
+        dump($order);
```
jolelievre: Don't forget to remove this

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
matthieu-rolland: isn't it one of the the goals of "xxxForViewing" objects to give us ready to use values so that it doesn't have to be caculated in frontend?

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
```diff
@@ -70,6 +100,49 @@ public function handle(DeleteProductFromOrderCommand $command)
...
+        CartRule::autoAddToCart();
+
+        // Remove previous cart rules applied to order
+        $result = array_diff(
+            array_map('serialize', $oldCartRules),
+            array_map('serialize', $cart->getCartRules())
+        );
+        $result = array_map('unserialize', $result);
+
+        foreach ($order->getCartRules() as $orderCartRule) {
```
jolelievre, line 154: This is technically not updating the cart, but I get that you need to do it here because you need the diff of cart rules
Maybe rename the method then `updateCartAndRules` ?

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
```diff
@@ -96,17 +169,47 @@ private function updateOrderInvoice(OrderDetail $orderDetail)
...
-    private function updateOrder(Order $order, OrderDetail $orderDetail)
+    private function updateOrder(Order $order, OrderDetail $orderDetail, Cart $cart)
     {
         // @todo: use https://github.com/PrestaShop/decimal for price computations
-        $order->total_paid -= $orderDetail->total_price_tax_incl;
-        $order->total_paid_tax_incl -= $orderDetail->total_price_tax_incl;
-        $order->total_paid_tax_excl -= $orderDetail->total_price_tax_excl;
-        $order->total_products -= $orderDetail->total_price_tax_excl;
-        $order->total_products_wt -= $orderDetail->total_price_tax_incl;
+        $computingPrecision = new ComputingPrecision();
```
jolelievre, line 199: I think you copied this code from the `AddProductToOrderHandler` right?
Then maybe it would be worth having a dedicated service/class to perform those updates? Some kind `OrderProductUpdater` which would update the cart at the same time (allowing you to store cart rules before/after update in a class field).

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/discount_list.html.twig
```diff
@@ -0,0 +1,49 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ *#}
+<table class="table discountList{% if orderForViewing.discounts.discounts is empty %} d-none{% endif %}">
```
jolelievre: If you create a template dedicated to display discounts it shouldn't rely on `orderForViewing` But simply use a `discounts` variable

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -147,33 +147,9 @@
...
-              </a>
-            </td>
-          </tr>
-        {% endfor %}
-        </tbody>
-      </table>
-    {% endif %}
+    {% include '@PrestaShop/Admin/Sell/Order/Order/Blocks/View/discount_list.html.twig' with {
+      'orderForViewing': orderForViewing
+    } %}
```
jolelievre:
```suggestion
    {% include '@PrestaShop/Admin/Sell/Order/Order/Blocks/View/discount_list.html.twig' with {
      'discounts': orderForViewing.discounts.discounts
    } %}
```

admin-dev/themes/new-theme/js/pages/order/view/order-discounts-refresher.js
```diff
@@ -0,0 +1,42 @@
...
+
+export default class OrderDiscountsRefresher {
+  constructor() {
+    this.router = new Router();
+  }
+
+  refresh(orderId) {
+    $.ajax(this.router.generate('admin_orders_get_discounts', {orderId}))
+      .then((response) => {
+        $(OrderViewPageMap.productDiscountList.list).replaceWith(response);
```
jolelievre, line 39: Do you think it would be complicated to return a JSON instead in the controller? And manage the update via javascript?
Instead of adding a new ajax controller that returns some HTML code

jolelievre (commented):
Careful with the remainin `dump`

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
```diff
@@ -70,6 +100,49 @@ public function handle(DeleteProductFromOrderCommand $command)
...
+            array_map('serialize', $cart->getCartRules())
+        );
+        $result = array_map('unserialize', $result);
+
+        foreach ($order->getCartRules() as $orderCartRule) {
+            foreach ($result as $itemCartRule) {
+                if ($itemCartRule['id_cart_rule'] == $orderCartRule['id_cart_rule']) {
+                    $orderCartRule = new OrderCartRule($orderCartRule['id_order_cart_rule']);
+                    $orderCartRule->delete();
+                }
```
matthieu-rolland, line 161: I'm not sure I understand what is being done here (the whole foreach), I'm not saying it's wrong, I just don't understand/know what you are doing exactly and why ?

EDIT: ok I get it 👍

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
matthieu-rolland: I agree with @jolelievre , the name of the method misled my understanding when reading this code for the first time

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
```diff
@@ -26,41 +26,71 @@
...
     {
         $orderDetail = new OrderDetail($command->getOrderDetailId());
         $order = new Order($command->getOrderId()->getValue());
+        dump($order);
 
         $this->assertProductCanBeDeleted($order, $orderDetail);
 
+        $cart = new Cart($order->id_cart);
+
+        $this->contextStateManager
```
sowbiba, line 81: Maybe restore the context if an exception is thrown

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
```diff
@@ -70,6 +100,49 @@ public function handle(DeleteProductFromOrderCommand $command)
...
+    private function updateCart(Order $order, OrderDetail $orderDetail, Cart &$cart)
+    {
+        $oldCartRules = $cart->getCartRules();
+        // Update Product Quantity in the cart
+        $return = $cart->updateQty(
+            $orderDetail->product_quantity,
+            $orderDetail->product_id,
+            $orderDetail->product_attribute_id,
+            false,
+            'down'
```
sowbiba, line 128: Can we put the operators in constants ? (not mandatory for this PR)

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1280,12 +1300,16 @@ public function getPricesAction(int $orderId): JsonResponse
     {
         /** @var OrderForViewing $orderForViewing */
         $orderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId));
+        $orderForViewingPrices = $orderForViewing->getPrices();
+        $zeroNumber = new Number('0');
```
sowbiba: used just once. Maybe you dont need to put it in a variable

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
Progi1984: Fixed

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
Progi1984: Fixed ;)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
Progi1984: Fixed

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
Progi1984: ;)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/discount_list.html.twig
Progi1984: Fixed

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
Progi1984: Fixed

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
Progi1984: It's legacy... so I can't do this... (may be in an another PR :wink:)

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
Progi1984: Fixed

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
Progi1984: A new issue has been created : https://github.com/PrestaShop/PrestaShop/issues/18432

matthieu-rolland (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1269,6 +1270,24 @@ public function deleteProductAction(int $orderId, int $orderDetailId): JsonRespo
...
+     * @return JsonResponse
+     */
+    public function getDiscountsAction(int $orderId): Response
+    {
+        /** @var OrderForViewing $orderForViewing */
+        $orderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId));
+
+        return $this->render('@PrestaShop/Admin/Sell/Order/Order/Blocks/View/discount_list.html.twig', [
+            'discounts' => $orderForViewing->getDiscounts()->getDiscounts(),
+            'orderForViewingId' => $orderForViewing->getId(),
```
jolelievre: `orderId` is enough here I think
Since it's only the ID it no longer carries any viewing info

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -175,6 +176,49 @@ public function addProductsToOrderWithNewInvoiceAndTheFollowingDetails(string $o
         }
     }
 
+    /**
+     * @When I remove product :productReference from order :orderReference
+     *
+     * @param string $productReference
+     * @param string $orderReference
+     */
+    public function removeProductsToOrderWithNewInvoiceAndTheFollowingDetails(string $productReference, string $orderReference)
```
jolelievre:
```suggestion
    public function removeProductsFromOrderWithNewInvoiceAndTheFollowingDetails(string $productReference, string $orderReference)
```

jolelievre (commented):
Two small feedbacks

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
Progi1984: Fixed

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
Progi1984: Fixed

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
```diff
@@ -54,22 +76,91 @@ public function handle(DeleteProductFromOrderCommand $command)
...
+            array_map('serialize', $oldCartRules),
+            array_map('serialize', $newCartRules)
+        );
+        $result = array_map('unserialize', $result);
+
+        foreach ($order->getCartRules() as $orderCartRule) {
+            foreach ($result as $itemCartRule) {
+                if ($itemCartRule['id_cart_rule'] == $orderCartRule['id_cart_rule']) {
+                    $orderCartRule = new OrderCartRule($orderCartRule['id_order_cart_rule']);
+                    $orderCartRule->delete();
```
sowbiba: return false if the deletion of orderCartRule returns false ?

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -175,6 +176,49 @@ public function addProductsToOrderWithNewInvoiceAndTheFollowingDetails(string $o
...
+                    $orderReference
+                )
+            );
+        }
+
+        try {
+            $this->getCommandBus()->handle(
+                new DeleteProductFromOrderCommand($orderId, $orderDetailId)
+            );
+        } catch (InvalidProductQuantityException $e) {
```
sowbiba: catch (InvalidProductQuantityException|ProductOutOfStockException $e) ?

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
sowbiba: I dont know what's the consequence of partially deleted cart rules

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
Progi1984: Fixed

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
Progi1984: Fixed

admin-dev/themes/new-theme/js/pages/order/view/order-discounts-refresher.js
matks: It would be a big change of scope for this PR.
This requires heavy javascript logic and it means that modules that would customize/override the Twig template will not see their changes being kept after ajax refresh (this is why we dont vue.js yet in BO). I suggest we stick to HTML replace 😉

Progi1984: The problem is that (if we do it) we need to do it before beta release, for not changing the signature of the API.

matks: Not an issue. If we want to modify it later, we can create a new API endpoint and deprecate the precious one. Let's continue this on slack 😉

sowbiba (approved):

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/discount_list.html.twig
```diff
@@ -0,0 +1,49 @@
...
+        {% if discount.amountRaw.greaterThan(number(0)) %}-{% endif %}
+        {{ discount.amountFormatted }}
+      </td>
+      <td class="text-right d-print-none">
+        <a class="delete-cart-rule btn btn-text" href="{{ path('admin_orders_remove_cart_rule', {'orderId': orderId, 'orderCartRuleId': discount.orderCartRuleId}) }}">
+          <i class="material-icons">delete</i>
+        </a>
+      </td>
+    </tr>
+  {% endfor %}
```
NeOMakinG, line 47: Isn't this indent missing ?

NeOMakinG (approved):
Looks fine to me, event if I posted a comment on an indent thing

jolelievre (approved):

matks (approved):

matthieu-rolland (approved):

Robin-Fischer-PS:
Except for the discount deletion (which will be treated in https://github.com/PrestaShop/PrestaShop/issues/18379), it's OK for me !

Thanks @Progi1984 !

Progi1984:
Thanks me :)

PR #17927, opened 2020-03-02
Wrong prices in edit order with secondary currency

sowbiba:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | Fixes total price error when adding a product to an order with secondary currency (not default one)
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #12521
| How to test?  | <ul><li>Add a secondary currency to the shop</li><li>Create an order using this secondary currency</li><li>Go to admin and add a product to the order</li><li>Result : The total price is incorrect</li></ul>

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17927)

PR #18205, opened 2020-03-18
BO : Add Product on Order : Use previous cart rather than creating a new one

### LINKED ISSUE #13880: Bug in displayed discount when adding product (linked to a coupon) to an existing order from BO

Progi1984:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When you add a product (which add a discount), the discount block is not added
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Partially fix #13880
| How to test?  | See https://github.com/PrestaShop/PrestaShop/issues/13880#issuecomment-595189713 from @khouloudbelguith

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18205)

### DIFF
--- a/classes/order/OrderDetail.php
+++ b/classes/order/OrderDetail.php
@@ -755,7 +755,7 @@ class OrderDetailCore extends ObjectModel
 
         unset(
             $this->vat_address,
-            $products, $this->customer
+            $this->customer
         );
     }
 
--- a/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
@@ -44,6 +44,7 @@ use OrderDetail;
 use OrderInvoice;
 use PrestaShop\Decimal\Number;
 use PrestaShop\PrestaShop\Adapter\Order\AbstractOrderHandler;
+use PrestaShop\PrestaShop\Adapter\Order\Invoice;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Product\Command\AddProductToOrderCommand;
 use PrestaShop\PrestaShop\Core\Domain\Order\Product\CommandHandler\AddProductToOrderHandlerInterface;
@@ -116,23 +117,24 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
 
         $this->addProductToCart($cart, $product, $combination, $command->getProductQuantity());
 
+        // Fetch Cart Product
+        $productCart = $this->getCartProduct($cart, $product, $command->getProductQuantity());
+
         $invoice = $this->createNewOrEditExistingInvoice(
             $command,
             $order,
             $cart,
-            $product
+            [$productCart]
         );
 
         $totalMethod = $command->getOrderInvoiceId() ? Cart::BOTH_WITHOUT_SHIPPING : Cart::BOTH;
 
         // Create Order detail information
-        $orderDetail = new OrderDetail();
-        $orderDetail->createList(
+        $orderDetail = $this->createOrderDetail(
             $order,
+            $invoice,
             $cart,
-            $order->getCurrentOrderState(),
-            [$product],
-            !empty($invoice->id) ? $invoice->id : 0
+            [$productCart]
         );
 
         // update totals amount of order
@@ -313,6 +315,31 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
     }
 
     /**
+     * @param Order $order
+     * @param OrderInvoice|null $invoice
+     * @param Cart $cart
+     * @param array $productCart
+     *
+     * @return OrderDetail
+     *
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
+    private function createOrderDetail(Order $order, ?OrderInvoice $invoice, Cart $cart, array $productCart): OrderDetail
+    {
+        $orderDetail = new OrderDetail();
+        $orderDetail->createList(
+            $order,
+            $cart,
+            $order->getCurrentOrderState(),
+            $productCart,
+            !empty($invoice->id) ? $invoice->id : 0
+        );
+
+        return $orderDetail;
+    }
+
+    /**
      * @param AddProductToOrderCommand $command
      * @param Order $order
      * @param Cart $cart
@@ -372,6 +399,59 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
     /**
      * @param Cart $cart
      * @param Product $product
+     * @param int $quantity
+     *
+     * @return array
+     */
+    private function getCartProduct(Cart $cart, Product $product, int $quantity): array
+    {
+        $productItem = array_reduce($cart->getProducts(), function ($carry, $item) use ($product) {
+            if (null !== $carry) {
+                return $carry;
+            }
+
+            return $item['id_product'] == $product->id ? $item : null;
+        });
+        $productItem['cart_quantity'] = $quantity;
+
+        switch (Configuration::get('PS_ROUND_TYPE')) {
+            case Order::ROUND_TOTAL:
+                $productItem['total'] = $productItem['price_with_reduction_without_tax'] * $quantity;
+                $productItem['total_wt'] = $productItem['price_with_reduction'] * $quantity;
+
+                break;
+            case Order::ROUND_LINE:
+                $productItem['total'] = Tools::ps_round(
+                    $productItem['price_with_reduction_without_tax'] * $quantity,
+                    Context::getContext()->getComputingPrecision()
+                );
+                $productItem['total_wt'] = Tools::ps_round(
+                    $productItem['price_with_reduction'] * $quantity,
+                    Context::getContext()->getComputingPrecision()
+                );
+
+                break;
+
+            case Order::ROUND_ITEM:
+            default:
+                $productItem['total'] = Tools::ps_round(
+                        $productItem['price_with_reduction_without_tax'],
+                        Context::getContext()->getComputingPrecision()
+                    ) * $quantity;
+                $productItem['total_wt'] = Tools::ps_round(
+                        $productItem['price_with_reduction'],
+                        Context::getContext()->getComputingPrecision()
+                    ) * $quantity;
+
+                break;
+        }
+
+        return $productItem;
+    }
+
+    /**
+     * @param Cart $cart
+     * @param Product $product
      * @param Combination|null $combination
      * @param int $quantity
      */
@@ -414,11 +494,11 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
         AddProductToOrderCommand $command,
         Order $order,
         Cart $cart,
-        Product $product
+        array $products
     ) {
         if ($order->hasInvoice()) {
             return $command->getOrderInvoiceId() ?
-                $this->updateExistingInvoice($command->getOrderInvoiceId(), $cart, $product) :
+                $this->updateExistingInvoice($command->getOrderInvoiceId(), $cart, $products) :
                 $this->createNewInvoice($order, $cart, $command->isFreeShipping());
         }
 
@@ -527,21 +607,21 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
      *
      * @return OrderInvoice
      */
-    private function updateExistingInvoice($orderInvoiceId, Cart $cart, Product $product)
+    private function updateExistingInvoice($orderInvoiceId, Cart $cart, array $products)
     {
         $precision = $this->getPrecisionFromCart($cart);
         $invoice = new OrderInvoice($orderInvoiceId);
 
         $invoice->total_paid_tax_excl += Tools::ps_round(
-            (float) $cart->getOrderTotal(false, Cart::BOTH_WITHOUT_SHIPPING, [$product]),
+            (float) $cart->getOrderTotal(false, Cart::BOTH_WITHOUT_SHIPPING, $products),
             $precision
         );
         $invoice->total_paid_tax_incl += Tools::ps_round(
-            (float) $cart->getOrderTotal(true, Cart::BOTH_WITHOUT_SHIPPING, [$product]),
+            (float) $cart->getOrderTotal(true, Cart::BOTH_WITHOUT_SHIPPING, $products),
             $precision
         );
-        $invoice->total_products += (float) $cart->getOrderTotal(false, Cart::ONLY_PRODUCTS, [$product]);
-        $invoice->total_products_wt += (float) $cart->getOrderTotal(true, Cart::ONLY_PRODUCTS, [$product]);
+        $invoice->total_products += (float) $cart->getOrderTotal(false, Cart::ONLY_PRODUCTS, $products);
+        $invoice->total_products_wt += (float) $cart->getOrderTotal(true, Cart::ONLY_PRODUCTS, $products);
 
         $invoice->update();
 
--- a/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
@@ -118,7 +118,7 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
         $this->addProductToCart($cart, $product, $combination, $command->getProductQuantity());
 
         // Fetch Cart Product
-        $productCart = $this->getCartProduct($cart, $product, $command->getProductQuantity());
+        $productCart = $this->getCartProductData($cart, $product, $command->getProductQuantity());
 
         $invoice = $this->createNewOrEditExistingInvoice(
             $command,
@@ -397,13 +397,16 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
     }
 
     /**
+     * This function extracts the newly added product from the cart and reformat the data in order to create a
+     * dedicated OrderDetail with appropriate amounts
+     *
      * @param Cart $cart
      * @param Product $product
      * @param int $quantity
      *
      * @return array
      */
-    private function getCartProduct(Cart $cart, Product $product, int $quantity): array
+    private function getCartProductData(Cart $cart, Product $product, int $quantity): array
     {
         $productItem = array_reduce($cart->getProducts(), function ($carry, $item) use ($product) {
             if (null !== $carry) {
@@ -486,7 +489,7 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
      * @param AddProductToOrderCommand $command
      * @param Order $order
      * @param Cart $cart
-     * @param Product $product
+     * @param array $products
      *
      * @return OrderInvoice|null
      */
@@ -499,7 +502,7 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
         if ($order->hasInvoice()) {
             return $command->getOrderInvoiceId() ?
                 $this->updateExistingInvoice($command->getOrderInvoiceId(), $cart, $products) :
-                $this->createNewInvoice($order, $cart, $command->isFreeShipping());
+                $this->createNewInvoice($order, $cart, $command->isFreeShipping(), $products);
         }
 
         return null;
@@ -509,8 +512,10 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
      * @param Order $order
      * @param Cart $cart
      * @param bool $isFreeShipping
+     * @param array $newProducts
+     * @param
      */
-    private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping)
+    private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping, array $newProducts)
     {
         $invoice = new OrderInvoice();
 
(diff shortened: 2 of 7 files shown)


matthieu-rolland (dismissed):
..

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -131,12 +131,12 @@ public function handle(AddProductToOrderCommand $command)
...
         // @todo: use https://github.com/PrestaShop/decimal for prices computations
-        $order->total_products += (float) $cart->getOrderTotal(false, Cart::ONLY_PRODUCTS);
-        $order->total_products_wt += (float) $cart->getOrderTotal(true, Cart::ONLY_PRODUCTS);
+        $order->total_products = (float) $cart->getOrderTotal(false, Cart::ONLY_PRODUCTS);
+        $order->total_products_wt = (float) $cart->getOrderTotal(true, Cart::ONLY_PRODUCTS);
 
-        $order->total_paid += Tools::ps_round((float) $cart->getOrderTotal(true, $totalMethod), 2);
-        $order->total_paid_tax_excl += Tools::ps_round((float) $cart->getOrderTotal(false, $totalMethod), 2);
-        $order->total_paid_tax_incl += Tools::ps_round((float) $cart->getOrderTotal(true, $totalMethod), 2);
+        $order->total_paid = Tools::ps_round((float) $cart->getOrderTotal(true, $totalMethod), 2);
```
matthieu-rolland:
```suggestion
        $order->total_paid = Tools::ps_round((float) $cart->getOrderTotal(true, $totalMethod), Cart::ONLY_DISCOUNTS);
```

And same for other hardcoded 2 values ;)

matthieu-rolland (changes requested):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: @matthieu-rolland The 2 value is for the `Tools::ps_round`, not the `$cart->getOrderTotal`

matthieu-rolland: ah damn, I badly read, my mistake .... in this case we need to use the proper computing precision:


```
        $computingPrecision = new ComputingPrecision();
        $precision = $computingPrecision->getPrecision($currency->precision);
```
Then replace 2 with $precision ...  (we're supposed to get rid of hardcoded precision everywhere)

matthieu-rolland: But I'm not blocking the PR for this since you didn't introduce this, I can do it in another PR 👍

matthieu-rolland (dismissed):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: @matthieu-rolland I do the move now ;)

matthieu-rolland (dismissed):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -131,12 +131,21 @@ public function handle(AddProductToOrderCommand $command)
...
-        $order->total_products_wt += (float) $cart->getOrderTotal(true, Cart::ONLY_PRODUCTS);
+        $order->total_products = (float) $cart->getOrderTotal(false, Cart::ONLY_PRODUCTS);
+        $order->total_products_wt = (float) $cart->getOrderTotal(true, Cart::ONLY_PRODUCTS);
 
-        $order->total_paid += Tools::ps_round((float) $cart->getOrderTotal(true, $totalMethod), 2);
-        $order->total_paid_tax_excl += Tools::ps_round((float) $cart->getOrderTotal(false, $totalMethod), 2);
-        $order->total_paid_tax_incl += Tools::ps_round((float) $cart->getOrderTotal(true, $totalMethod), 2);
+        $order->total_paid = Tools::ps_round(
+            (float) $cart->getOrderTotal(true, $totalMethod),
+            $this->context->getComputingPrecision()
```
jolelievre: I don't think you should use the precision rom context It will use the context's currency which might be different from the order one Rather use a dedicated ComputingPrecision object along with the precision o the currency in the order

jolelievre (changes requested):
In the same spirit as the comment I let, we shouldn't rely on context values What you changed or the cart is even more true!
In this handler we use cart rules from the cart present in the context, shouldn't we use the cart associated to the order like you did to get the real total?

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: Thanks. I applied your review ;) (Thanks a lot dude ;))

matthieu-rolland (dismissed):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -505,15 +529,16 @@ private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping)
      */
     private function updateExistingInvoice($orderInvoiceId, Cart $cart)
     {
+        $precision = $this->getPrecisionFromCart($cart);
         $invoice = new OrderInvoice($orderInvoiceId);
 
         $invoice->total_paid_tax_excl += Tools::ps_round(
```
jolelievre, line 635: Something feels off For the order you replaced the `+=` part with a simple `=` I guess it's because previously the cart only contained the new product but now it contains all o them
So it should be the same thing when you update the invoice right?

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -462,8 +479,15 @@ private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping)
         $taxCalculator = $carrier->getTaxCalculator($invoice_address);
 
         // @todo: use https://github.com/PrestaShop/decimal to compute prices and taxes
-        $invoice->total_paid_tax_excl = Tools::ps_round((float) $cart->getOrderTotal(false, $totalMethod), 2);
-        $invoice->total_paid_tax_incl = Tools::ps_round((float) $cart->getOrderTotal(true, $totalMethod), 2);
+        $precision = $this->getPrecisionFromCart($cart);
+        $invoice->total_paid_tax_excl = Tools::ps_round(
```
jolelievre, line 591: Same here i you create a new invoice with a cart containing ALL the products the total won't be correct

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: Correct. I fixed it.

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: Changed the behavior.

classes/order/OrderDetail.php
```diff
@@ -755,7 +755,7 @@ public function createList(Order $order, Cart $cart, $id_order_state, $product_l
 
         unset(
             $this->vat_address,
-            $products, $this->customer
```
matks, line 758: I see that `OrderDetail::createList()` is used in PaymentModule.
Modifying its unspecified behavior is highly dangerous.
May I suggest some other ideas ?
- create a new function instead of using/modifying `createList()`
- add a flag (yes I know 😱) to modify this one

I am very scared of this change

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -117,40 +119,60 @@ public function handle(AddProductToOrderCommand $command)
...
-        $order->total_products_wt += (float) $cart->getOrderTotal(true, Cart::ONLY_PRODUCTS);
-
-        $order->total_paid += Tools::ps_round((float) $cart->getOrderTotal(true, $totalMethod), 2);
-        $order->total_paid_tax_excl += Tools::ps_round((float) $cart->getOrderTotal(false, $totalMethod), 2);
-        $order->total_paid_tax_incl += Tools::ps_round((float) $cart->getOrderTotal(true, $totalMethod), 2);
+        $orderProducts = $order->getCartProducts();
+        $order->total_products = (float) $cart->getOrderTotal(false, Cart::ONLY_PRODUCTS, $orderProducts);
+        $order->total_products_wt = (float) $cart->getOrderTotal(true, Cart::ONLY_PRODUCTS, $orderProducts);
+
+        $precision = $this->getPrecisionFromCart($cart);
```
matks: Suggestion, but not mandatory: use Decimal instead of `ps_round()`

classes/order/OrderDetail.php
Progi1984: Please check again... The variable isn't used in  the method and it's the alone reference. :wink:

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -349,6 +396,59 @@ private function createSpecificPriceIfNeeded(
         return null;
     }
 
+    /**
+     * @param Cart $cart
```
jolelievre, line 418: Add comment, something like: `This function extracts the newly added product from the cart and reformat the data in order to create a dedicated OrderDetail with the appropriate amounts`
And maybe rename into `getNewCartProduct` or `getNewCartProductData`

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -465,8 +567,15 @@ private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping)
         $taxCalculator = $carrier->getTaxCalculator($invoice_address);
 
         // @todo: use https://github.com/PrestaShop/decimal to compute prices and taxes
-        $invoice->total_paid_tax_excl = Tools::ps_round((float) $cart->getOrderTotal(false, $totalMethod), 2);
-        $invoice->total_paid_tax_incl = Tools::ps_round((float) $cart->getOrderTotal(true, $totalMethod), 2);
+        $precision = $this->getPrecisionFromCart($cart);
+        $invoice->total_paid_tax_excl = Tools::ps_round(
+            (float) $cart->getOrderTotal(false, $totalMethod),
```
jolelievre: Be careful I think here you need to use the new products as well
The new invoice can be created after one already existed, in which case you don't want the newly created invoice to contain the total of the previous invoice

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -503,23 +603,25 @@ private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping)
     /**
      * @param int $orderInvoiceId
      * @param Cart $cart
+     * @param Product $product
      *
      * @return OrderInvoice
      */
-    private function updateExistingInvoice($orderInvoiceId, Cart $cart)
+    private function updateExistingInvoice($orderInvoiceId, Cart $cart, array $products)
```
jolelievre:
```suggestion
    private function updateExistingInvoice($orderInvoiceId, Cart $cart, array $newProducts)
```

for better understanding

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -503,23 +603,25 @@ private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping)
     /**
      * @param int $orderInvoiceId
      * @param Cart $cart
+     * @param Product $product
```
jolelievre: this param doesn't match the new argument

jolelievre (changes requested):
Some small fix left, but congrats on this fix!!

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: Fixed

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: Fixed

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: Thanks. I check it ;)

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: Fixed

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -102,7 +105,7 @@ public function handle(AddProductToOrderCommand $command)
 
         $this->checkProductInStock($command->getProductId()->getValue(), $command->getCombinationId(), $command->getProductQuantity());
 
-        $cart = $this->createNewCart($order);
+        $cart = $this->createNewOrEditExistingCart($order);
```
jolelievre: Maybe you could move this here: `$oldCartRules = $cart->getCartRules();`
Instead of L95, this would seem more logical and I think it will ease the merge with @sowbiba 's PR: https://github.com/PrestaShop/PrestaShop/pull/17927

Progi1984: :+1:

classes/order/OrderDetail.php
matks: You're right, I was confused by this stupid code 😓

matks (dismissed):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -199,22 +216,33 @@ public function handle(AddProductToOrderCommand $command)
...
+            (float) $cart->getOrderTotal(true, $totalMethod, $orderProducts),
+            $precision
+        );
+        $order->total_discounts = (float) abs($cart->getOrderTotal(true, Cart::ONLY_DISCOUNTS, $orderProducts));
+        $order->total_discounts_tax_excl = (float) abs($cart->getOrderTotal(false, Cart::ONLY_DISCOUNTS, $orderProducts));
+        $order->total_discounts_tax_incl = (float) abs($cart->getOrderTotal(true, Cart::ONLY_DISCOUNTS, $orderProducts));
+
         $order->update();
+
+        $this->contextStateManager->restoreContext();
```
sowbiba, line 250: Consider restore the context if there is an exception between the moment it has been modified and the moment it is restored

try {} catch{ restore context }

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -114,46 +124,52 @@ public function handle(AddProductToOrderCommand $command)
 
         $this->addProductToCart($cart, $product, $combination, $command->getProductQuantity());
 
+        // Fetch Cart Product
+        $productCart = $this->getCartProductData($cart, $product, $command->getProductQuantity());
+
         $invoice = $this->createNewOrEditExistingInvoice(
```
sowbiba: I maybe dont get it. We have 1 invoice per product ?

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -523,21 +581,39 @@ public function duplicateOrder($orderReference, $duplicatedCartReference)
...
+            }
         }
 
-        throw new RuntimeException(sprintf('Order was expected to have "%d" products "%s" in it. Instead got "%s"', $quantity, $productName, $productQuantities['quantity']));
+        if ($productQuantity == $quantity) {
+            return;
+        }
+        throw new RuntimeException(
+            sprintf(
+                'Order was expected to have "%d" products "%s" in it. Instead got "%s"',
```
sowbiba: [Minor] we can use %d for the last placeholder

Progi1984: Fixed

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: Done

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: No it's depending of the input ;)

sowbiba (dismissed):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -89,132 +99,155 @@ public function handle(AddProductToOrderCommand $command)
...
+            ->setCurrency(new Currency($order->id_currency))
+            ->setCustomer(new Customer($order->id_customer));
 
-        $product = $this->getProductObject($command->getProductId(), (int) $order->id_lang);
-        $combination = $this->getCombination($command->getCombinationId());
+        try {
+            $this->assertOrderWasNotShipped($order);
 
-        $this->checkProductInStock($command->getProductId()->getValue(), $command->getCombinationId(), $command->getProductQuantity());
+            $oldCartRules = (new Cart($order->id_cart))->getCartRules();
```
jolelievre: Only a suggestion if you feel like it, this could be done right after the call to `createNewOrEditExistingCart` L116:
`$oldCartRules = $cart->getCartRules();`

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -89,132 +99,155 @@ public function handle(AddProductToOrderCommand $command)
...
+            }
 
-        $order = $order->refreshShippingCost();
+            $order = $order->refreshShippingCost();
 
-        Hook::exec('actionOrderEdited', ['order' => $order]);
+            Hook::exec('actionOrderEdited', ['order' => $order]);
 
-        $newCartRules = $this->context->cart->getCartRules();
+            $newCartRules = $this->context->cart->getCartRules();
```
jolelievre: Seems like `$this->context->cart = $cart;` (L327)
So actually `$this->context->cart === $cart` So maybe we could remove all direct references to `$this->context` ?

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -89,132 +99,155 @@ public function handle(AddProductToOrderCommand $command)
...
-        if ($this->context->cart !== null) {
-            $oldCartRules = $this->context->cart->getCartRules();
-        } else {
-            $oldCartRules = [];
-        }
-
-        $this->assertOrderWasNotShipped($order);
+        $this->contextStateManager
+            ->setCurrency(new Currency($order->id_currency))
+            ->setCustomer(new Customer($order->id_customer));
```
jolelievre, line 104: Thanks to this I think `$this->context->customer = new Customer($order->id_customer);` L328 is now useless

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -89,132 +99,155 @@ public function handle(AddProductToOrderCommand $command)
...
 
-        Hook::exec('actionOrderEdited', ['order' => $order]);
+            Hook::exec('actionOrderEdited', ['order' => $order]);
 
-        $newCartRules = $this->context->cart->getCartRules();
+            $newCartRules = $this->context->cart->getCartRules();
 
-        sort($oldCartRules);
-        sort($newCartRules);
+            sort($oldCartRules);
```
jolelievre, line 200: Since old and new cart rules now come from the same cart instance, `$newCartRules` are only the ones added from the product, maybe this could be fetched more easily

matks (dismissed):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -503,23 +624,33 @@ private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping)
...
+        );
+        $invoice->total_products += (float) $cart->getOrderTotal(
+            false,
+            Cart::ONLY_PRODUCTS,
+            $newProducts
+        );
+        $invoice->total_products_wt += (float) $cart->getOrderTotal(
+            true,
+            Cart::ONLY_PRODUCTS,
+            $newProducts
```
jolelievre, line 651: No need to update `$invoice->total_wrapping` like in `createInvoice` here?

jolelievre (dismissed):
Great improvement @Progi1984 

I only added some comments because we are a few lines away from getting completely rid of `$this->context`

Only thing I'm worried is the comment about wrapping in `updateExistingInvoice `

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: :+1:

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: :+1:

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: :+1:

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: I don't know... but not the cas in legacy :tongue:

jolelievre (approved):

sowbiba (approved):

Robin-Fischer-PS:
Hi @Progi1984 !

I have two errors with your PR : 

- I cannot delete the product which is linked to the discount : When I click on delete and confirm, nothing happens. After refresh, the product is still in the cart.
- When I try to delete the voucher, there is an error

See the errors in this screenrecord : 
https://drive.google.com/file/d/1yXPaN5rDVTXcLychf1ft3e_3RbblR3QT/view?usp=sharing

Thanks !

Progi1984:
@Robin-Fischer-PS
> I cannot delete the product which is linked to the discount : When I click on delete and confirm, nothing happens. After refresh, the product is still in the cart.

Will be fixed in PR #18204 

> When I try to delete the voucher, there is an error

A new issue has been created : #18379

Robin-Fischer-PS:
Alright, then it's OK for me ! :heavy_check_mark: 

Thanks @Progi1984  :)

Progi1984:
Thanks @Robin-Fischer-PS

PR #18220, opened 2020-03-19
Add phpdoc for PDF classes

Matt75:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add phpdoc for PDF classes
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Everything should work as before in PDF generation

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18220)

### DIFF
--- a/classes/pdf/HTMLTemplateInvoice.php
+++ b/classes/pdf/HTMLTemplateInvoice.php
@@ -40,18 +40,22 @@ class HTMLTemplateInvoiceCore extends HTMLTemplate
     public $order_invoice;
 
     /**
+     * @var bool
+     */
+    public $available_in_your_account = false;
+
+    /**
      * @param OrderInvoice $order_invoice
      * @param Smarty $smarty
      * @param bool $bulk_mode
      *
      * @throws PrestaShopException
      */
-    public function __construct(OrderInvoice $order_invoice, $smarty, $bulk_mode = false)
+    public function __construct(OrderInvoice $order_invoice, Smarty $smarty, $bulk_mode = false)
     {
         $this->order_invoice = $order_invoice;
         $this->order = new Order((int) $this->order_invoice->id_order);
         $this->smarty = $smarty;
-        $this->available_in_your_account = false;
 
         // If shop_address is null, then update it with current one.
         // But no DB save required here to avoid massive updates for bulk PDF generation case.
@@ -92,7 +96,7 @@ class HTMLTemplateInvoiceCore extends HTMLTemplate
      *
      * @return array Layout elements columns size
      */
-    protected function computeLayout($params)
+    protected function computeLayout(array $params)
     {
         $layout = [
             'reference' => [
@@ -379,7 +383,7 @@ class HTMLTemplateInvoiceCore extends HTMLTemplate
     /**
      * Returns the tax tab content.
      *
-     * @return string|array Tax tab html content (Returns an array if debug)
+     * @return string|array Tax tab html content (Returns an array if debug params used in request)
      */
     public function getTaxTabContent()
     {
--- a/classes/pdf/HTMLTemplateInvoice.php
+++ b/classes/pdf/HTMLTemplateInvoice.php
@@ -29,13 +29,20 @@
  */
 class HTMLTemplateInvoiceCore extends HTMLTemplate
 {
+    /**
+     * @var Order
+     */
     public $order;
+
+    /**
+     * @var OrderInvoice
+     */
     public $order_invoice;
-    public $available_in_your_account = false;
 
     /**
      * @param OrderInvoice $order_invoice
-     * @param $smarty
+     * @param Smarty $smarty
+     * @param bool $bulk_mode
      *
      * @throws PrestaShopException
      */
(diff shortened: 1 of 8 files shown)


classes/pdf/HTMLTemplateInvoice.php
```diff
@@ -29,13 +29,20 @@
...
+    /**
+     * @var Order
+     */
     public $order;
+
+    /**
+     * @var OrderInvoice
+     */
     public $order_invoice;
-    public $available_in_your_account = false;
```
matks, line 34: Although it seems to be useless, we cannot remove it, it would be a breaking change if someone extended this class 😢

classes/pdf/HTMLTemplateInvoice.php
```diff
@@ -80,7 +88,7 @@ public function getHeader()
     /**
      * Compute layout elements size.
      *
-     * @param $params Array Layout elements
+     * @param array $params Layout elements
```
matks, line 95: 👍

classes/pdf/HTMLTemplateInvoice.php
```diff
@@ -29,13 +29,20 @@
...
+    /**
+     * @var OrderInvoice
+     */
     public $order_invoice;
-    public $available_in_your_account = false;
 
     /**
      * @param OrderInvoice $order_invoice
-     * @param $smarty
+     * @param Smarty $smarty
```
PierreRambaud, line 49: The signature is `mixed` not `Smarty` even if it should be a `Smarty` instance :/

classes/pdf/HTMLTemplateInvoice.php
PierreRambaud, line 95: Same here, if we're sure it's an array, we should do `computeLayout(array $params)`

classes/pdf/PDF.php
```diff
@@ -145,7 +169,7 @@ public function render($display = true)
     /**
      * Get correct PDF template classes.
      *
-     * @param mixed $object
+     * @param ObjectModel $object
```
PierreRambaud, line 172: We can't change here because the signature indicated `mixed` not ObjectModel

classes/pdf/HTMLTemplateOrderSlip.php
```diff
@@ -29,12 +29,24 @@
...
 
+    /**
+     * @var int
+     */
+    private $id_cart;
+
     /**
      * @param OrderSlip $order_slip
-     * @param $smarty
+     * @param Smarty $smarty
```
PierreRambaud, line 49:
```suggestion
     * @param mixed $smarty
```

classes/pdf/HTMLTemplateInvoice.php
PierreRambaud: It can be changed, the signature is `mixed` not array :thinking:

classes/pdf/HTMLTemplateInvoice.php
Matt75: This property is in parent HTMLTemplate and moved assignation in constructor, so no breaking change for existing override

classes/pdf/PDF.php
Matt75: If it's not an ObjectModel, it will fail...

classes/pdf/HTMLTemplateInvoice.php
Matt75: It's sure, so I can change to `computeLayout(array $params)` or it's a BC ?
Change only phpdoc don't change behavior

Matt75: Can or not ?

classes/pdf/HTMLTemplateInvoice.php
Matt75: It must be a Smarty instance to work

classes/pdf/HTMLTemplateOrderSlip.php
Matt75: It must be a Smarty instance to work

classes/pdf/HTMLTemplateInvoice.php
PierreRambaud: Can't sorry forget a part :D

classes/pdf/HTMLTemplateInvoice.php
PierreRambaud: So you can change the method signature too :wink:

classes/pdf/HTMLTemplateInvoice.php
PierreRambaud: Having a phpdoc different than the signature is weird.

classes/pdf/PDF.php
PierreRambaud: So you need to change the method signature too

classes/pdf/HTMLTemplateOrderSlip.php
PierreRambaud: You need to change the method signature too :)

Progi1984 (approved):

PierreRambaud (approved):

matks (approved):

khouloudbelguith:
Hi @Matt75,

It is OK :heavy_check_mark: 

Thanks!

PierreRambaud:
Thanks @Matt75

presteus:
Please, respect class name and filename recommandation, you have several files that do not respect this simple recommandation.

**Class name:** HTMLTemplateInvoiceCore
**Filename:** HTMLTemplateInvoice.php

Matt75:
@presteus This files classe name and files name are same since 1.3 or less. Change it will be a breaking change 🤔

PR #18228, opened 2020-03-19
Improve common search controller for backward compatibility

### LINKED ISSUE #18218: BO - Filter doesn't work in the Categories page and the Employees - an exception displayed
#### Describe the bug

In the BO => Catalog => Categories page & the BO => Team => Employees page,
when we try to search by any filter => an exception is displayed

1. In the BO => Catalog => Categories page 
image
2. In the BO => Team => Employees page
image

#### Steps to Reproduce

1. Go to BO => Catalog => Categories page
2. Search by any filter
3. See error
4. Go to BO => Team => Employees page
5. Search by any filter
6. See error

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: 7.2

### LINKED ISSUE #18271: BO - Files page - Filter doesn't work, it just refresh the page
#### Describe the bug

In the BO => Catalog => Files page, when we try to search by any filter => nothing happens, the page is just refreshed.

#### Steps to Reproduce

BO - Files page - Filter doesn't work, it just refresh the page
1. Go to BO => Catalog => Files page
2. Search by any filter
3. See error

**Screenshots**
[Files • 177x.webm.zip](https://github.com/PrestaShop/PrestaShop/files/4370230/Files.177x.webm.zip)

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: 7.2

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Introduce ControllerResponseBuilder so that `CommonController::searchAction` works on both type of grids (with or without `filterId`) To identify which kind of definition is being used a new `IdentifiableGridDefinitionFactoryInterface` has been introduced along with its matching abstract class `AbstractIdentifiableGridDefinitionFactory` for easier use<br>Also fix the Files filtering
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18218 and #18271
| How to test?  | Check that the search feature is fixed on Employee and Category grid. As some grid have been modified like in this PR https://github.com/PrestaShop/PrestaShop/pull/18130 it might be safer to check those lists<br>- addresses<br>- attachments<br>- orders,<br>- attributes <br>- attribute groups (`sell/catalog/attribute-groups` view action opens attributes list),<br>- customer addresses,<br>- order messages,<br>- merchandise returns,<br>- suppliers ( this was migrated already in 1.7.6, but was still hidden)<br>- CMS page category<br>- tax rules

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18228)

### DIFF
--- a/src/Core/Grid/Definition/Factory/AbstractIdentifiableGridDefinitionFactory.php
+++ b/src/Core/Grid/Definition/Factory/AbstractIdentifiableGridDefinitionFactory.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Grid\Definition\Factory;
+
+/**
+ * Class AbstractIdentifiableGridDefinitionFactory implements identified grid definition creation.
+ */
+abstract class AbstractIdentifiableGridDefinitionFactory extends AbstractGridDefinitionFactory implements IdentifiableGridDefinitionFactoryInterface
+{
+    /**
+     * {@inheritdoc}
+     *
+     * Grid definition already has an internal id used or hooks, that can also be used as an identifier for filters
+     */
+    public function getGridId(): string
+    {
+        return $this->getId();
+    }
+}
--- a/src/PrestaShopBundle/Controller/Admin/CommonController.php
+++ b/src/PrestaShopBundle/Controller/Admin/CommonController.php
@@ -37,7 +37,9 @@ use PrestaShop\PrestaShop\Core\Grid\Definition\Factory\GridDefinitionFactoryInte
 use PrestaShop\PrestaShop\Core\Kpi\Row\KpiRowInterface;
 use PrestaShopBundle\Security\Annotation\AdminSecurity;
 use PrestaShopBundle\Service\DataProvider\Admin\RecommendedModules;
+use PrestaShopBundle\Service\Grid\ControllerResponseBuilder;
 use PrestaShopBundle\Service\Grid\ResponseBuilder;
+use ReflectionClass;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\RedirectResponse;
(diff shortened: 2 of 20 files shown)


src/PrestaShopBundle/Service/Grid/ControllerResponseBuilder.php
```diff
@@ -0,0 +1,105 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
Progi1984, line 27: declare(strict_types=1);

src/PrestaShopBundle/Controller/Admin/CommonController.php
```diff
@@ -357,15 +359,30 @@ public function searchGridAction(
         $redirectRoute,
         array $redirectQueryParamsToKeep = []
     ) {
-        /** @var ResponseBuilder $responseBuilder */
-        $responseBuilder = $this->get('prestashop.bundle.grid.response_builder');
         /** @var GridDefinitionFactoryInterface $definitionFactory */
         $definitionFactory = $this->get($gridDefinitionFactoryServiceId);
+        $reflectionClass = new ReflectionClass($definitionFactory);
```
matks: I have always been told to never use Reflection in production 😮(ping @PierreRambaud any opinion on this ?)

jolelievre: I'm not a big fan either but I only see two alternatives:
- a try/catch block witch would catch a very generic `FatalThrowaleError` (if ti works)
- adding an interface to all grid definitions which manage the GRID_ID which would allow to use `instanceof`

matks: I ❤️ the interface solution

PierreRambaud: `Interface` FTW

jolelievre: The only issue about the interface solution is that it brings more complexity for module developers, we will have to update the doc accordingly

And I'm afraid it could create a BC break 🤔 as I'm not sure exactly what the interface should contain:
`public function getFilterId(): string` maybe?

PierreRambaud: This complexity already exist in PrestaShop, and if you want to help developers feel free to add an abstract class in addition of the Interface.

matks: No BC break I think 🤔
- create new Interface `IdentifiedGridDefinitionFactoryInterface` that extends existing `GridDefinitionFactoryInterface`
- apply to all existing grids in Core
- check if given definition is instance of `IdentifiedGridDefinitionFactoryInterface` to use the ID

jolelievre:
@PierreRambaud @matks it should be good, I added the interface and an abstract class Which makes it simpler to update the definition (only change the parent class)
Like in the initial PR https://github.com/PrestaShop/PrestaShop/pull/18130 I only updated the grid that use the common controller, as the remaining grids are planned to be udpated in the next version (in this PR https://github.com/PrestaShop/PrestaShop/pull/18136 I added a comment so that it uses this new parent class)

Progi1984 (dismissed):

Progi1984 (dismissed):

matks (dismissed):

src/Core/Grid/Definition/Factory/AbstractIdentifiableGridDefinitionFactory.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    /**
+     * {@inheritdoc}
+     *
+     * Grid definition already has an internal id used or hooks, that can also be used as an identifier for filters
+     */
+    public function getGridId(): string
+    {
+        return $this->getId();
+    }
```
PierreRambaud: Still thinking it's confusing, we are already in a Grid context, the getId should be the getGridId. Should be renamed getFilterId

jolelievre:
@PierreRambaud you're right I modified the interface and methods, I'm not 100% satisfied by it but I could't find better naming.
So the interface is `FilterableGridDefinitionFactoryInterface` and the method `getFilterId`

PierreRambaud (approved):

Progi1984 (approved):

matks (approved):

SD1982:
The filters seems to work correctly ! Thanks @jolelievre

matks:
Thank you @jolelievre

Progi1984:
Thank you @jolelievre

PR #18230, opened 2020-03-19
Avoid fatal error when module hookDisplayPaymentEU returns another thing instead of array

### LINKED ISSUE #18229: [Checkout][FO] Fatal error when module hookDisplayPaymentEU returns another thing instead of array
#### Describe the bug

Fatal error when module `hookDisplayPaymentEU` returns another thing instead of array 

#### Expected behavior

Avoid fatal error when module `hookDisplayPaymentEU` returns another thing instead of array

#### Steps to Reproduce

1. Go to checkout page in FO
2. See this is Ok !
3. Now ... install this module  : [pululuk.zip](https://drive.google.com/open?id=1DeCGZWtkShyjJqIUvQNf9z_nXrV3_bQW) or in another module add : 
```php
public function install(){ 
      ... etc etc
      $this->registerHook('displayPaymentEU')
      ... etc etc
}
```

and 
```php
public function hookDisplayPaymentEU($params){
	return null;
} 
```
4. Return to checkout page in FO, See fatal error

**Screenshots**

https://drive.google.com/file/d/1cV4VzhN9NA9uYwTmTDj8RDjuGW1mBWZH/view
#### Additional information

* PrestaShop version: 1.7.6.3
* PHP version: N/A

PululuK:
| Questions     | Answers
| Branch?       | develop
| Description?  | Avoid fatal error `Catchable Fatal Error: Argument 1 passed to PrestaShop\PrestaShop\Core\Payment\PaymentOption::convertLegacyOption() must be of the type array, null given.` when module **hookDisplayPaymentEU** returns another thing instead of array. See [PaymentOption::convertLegacyOption()](https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Payment/PaymentOption.php#L337) method signature.
| Type?         | improvement 
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18229.
| How to test?  | See #18229.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18230)

### DIFF
--- a/classes/checkout/PaymentOptionsFinder.php
+++ b/classes/checkout/PaymentOptionsFinder.php
@@ -40,7 +40,7 @@ class PaymentOptionsFinderCore extends HookFinder
     {
         // Payment options coming from intermediate, deprecated version of the Advanced API
         $this->hookName = 'displayPaymentEU';
-        $rawDisplayPaymentEUOptions = parent::find();
+        $rawDisplayPaymentEUOptions = array_filter(parent::find(), 'is_array');
         $paymentOptions = array_map(
             ['PrestaShop\PrestaShop\Core\Payment\PaymentOption', 'convertLegacyOption'],
             $rawDisplayPaymentEUOptions


Progi1984 (approved):

classes/checkout/PaymentOptionsFinder.php
```diff
@@ -40,7 +40,7 @@ public function find() //getPaymentOptions()
     {
         // Payment options coming from intermediate, deprecated version of the Advanced API
         $this->hookName = 'displayPaymentEU';
-        $rawDisplayPaymentEUOptions = parent::find();
+        $rawDisplayPaymentEUOptions = array_filter(parent::find(), 'is_array');
```
PierreRambaud, line 43: I'm not sure it's the good way to fix the bug.
Regarding the issue the problem come from the module, this one shouldn't return a null value for this hook.
It's the module developer responsibility to use the correct behaviour :thinking:

PululuK: But that doesn't stop us from checking if it returns the expected value!

jolelievre: You can check the return and throw a more adapted exception to help identify the problem, but instead the problem should be ignored/hidden

PululuK: @PierreRambaud This is not a bug fix, but rather an improvement to avoid fatal errors, in case the developer has obulated to return the expected value ... wdyf ?

PululuK: Hi @jolelievre  

Maybe trigger a php warning?

PululuK: Maybe trigger a php warning?

jolelievre: The thing is that the error should not be hidden if there is one, and a warning will be ignored by most merchants/devs as they don't look into the logs
If you ignore the problem, then the only thing that will happen is people won't understand that it doesn't work. Usually when an unexpected behaviour is occurring and there is no possible fallback it's best to have an error, the more detailed the error the easier the debugging can then be done but it's still better to have an error

PierreRambaud: I don't consider this as an improvement, because you add treatments and use memory for something that should never happened. Wasting memory if the first step of bad performance :/

PululuK: > ... for something that should never happened




Are you sure @PierreRambaud  🤔 ? See this case please https://github.com/PrestaShop/PrestaShop/issues/18229

PierreRambaud: This is what I said, if you use this hook, you know you shouldn't return null but an array.

PierreRambaud:
Hi @PululuK,

I'm closing it as we won't accept this pull request, it's up to module developers to be aware of return types. Feel free to reopen or open another one if you think it's still relevant.

Thanks!

PR #18235, opened 2020-03-20
Improve prestashop:linter:security-annotation

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | See below this table.
| Type?         | new feature
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | No need for QA. Try CLI Command `php bin/console prestashop:linter:security-annotation list`

## Description

I refactored and improve command `prestashop:linter:security-annotation`
Previous behavior (listing all Symfony routes without security annotation provided) is still available but now you must run `php bin/console prestashop:linter:security-annotation find-missing` to obtain it.

New behavior added: `php bin/console prestashop:linter:security-annotation list` will now output **all Symfony routes** available and display the controller, the action, the HTTP methods, whether it's secured using Security annotation and the permissions needed to use it.

This new behavior should help us to find unproperly configured routes with strange permission sets (example: a POST route modifying data* accepting READ permission).

*There is a few exceptions: for example some routes which are used to perform data search or data exports use POST.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18235)

### DIFF
--- a/src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
+++ b/src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
@@ -111,7 +111,7 @@ final class SecurityAnnotationLinterCommand extends ContainerAwareCommand
 
         switch ($actionToPerform) {
             case self::ACTION_LIST_ALL:
-                $this->listAllRoutesAndTheirPermissions($input, $output);
+                $this->listAllRoutesAndRelatedPermissions($input, $output);
                 break;
             case self::ACTION_FIND_MISSING:
                 $this->findRoutesWithMissingSecurityAnnotations($input, $output);
@@ -126,7 +126,7 @@ final class SecurityAnnotationLinterCommand extends ContainerAwareCommand
      * @param InputInterface $input
      * @param OutputInterface $output
      */
-    private function listAllRoutesAndTheirPermissions(InputInterface $input, OutputInterface $output)
+    private function listAllRoutesAndRelatedPermissions(InputInterface $input, OutputInterface $output)
     {
         $container = $this->getContainer();
 
@@ -139,7 +139,7 @@ final class SecurityAnnotationLinterCommand extends ContainerAwareCommand
         $listing = [];
 
         foreach ($adminRouteProvider->getRoutes() as $routeName => $route) {
-            /* @var \Symfony\Component\Routing\Route $route */
+            /* @var Route $route */
             try {
                 $annotation = $securityAnnotationLinter->getRouteSecurityAnnotation($routeName, $route);
                 $listing[] = [
--- a/src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
+++ b/src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
@@ -57,11 +57,11 @@ final class SecurityAnnotationLinterCommand extends ContainerAwareCommand
         $matches1 = [];
         $matches2 = [];
         preg_match($pattern1, $expression, $matches1);
-        preg_match($pattern2, $expression, $matches2);
 
         if (count($matches1) > 1) {
             return $matches1[1];
         }
+        preg_match($pattern2, $expression, $matches2);
         if (count($matches2) > 1) {
             return $matches2[1];
         }
--- a/src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
+++ b/src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
@@ -27,7 +27,9 @@
 namespace PrestaShopBundle\Command;
 
 use PrestaShopBundle\Routing\Linter\Exception\LinterException;
+use PrestaShopBundle\Routing\Linter\SecurityAnnotationLinter;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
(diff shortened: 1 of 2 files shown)


src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
```diff
@@ -27,35 +27,155 @@
...
+            default:
+                throw new \RuntimeException(sprintf('Unknown action %s', $actionToPerform));
+        }
+    }
+
+    /**
+     * @param InputInterface $input
+     * @param OutputInterface $output
+     */
+    private function listAllRoutesAndTheirPermissions(InputInterface $input, OutputInterface $output)
```
Progi1984:
```suggestion
    private function listAllRoutesAndRelatedPermissions(InputInterface $input, OutputInterface $output)
```

src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
```diff
@@ -27,35 +27,155 @@
...
+        $adminRouteProvider = $container
+            ->get('prestashop.bundle.routing.linter.admin_route_provider');
+        /** @var SecurityAnnotationLinter $securityAnnotationLinter */
+        $securityAnnotationLinter = $container
+            ->get('prestashop.bundle.routing.linter.security_annotation_linter');
+
+        $listing = [];
+
+        foreach ($adminRouteProvider->getRoutes() as $routeName => $route) {
+            /* @var \Symfony\Component\Routing\Route $route */
```
Arman-Hosseini:
```suggestion
            /** @var Route $route */
```

matks: Thanks, applied in https://github.com/PrestaShop/PrestaShop/pull/18235/commits/001e445569e9009913564470183c238688b26456

src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
matks: Thanks, applied in https://github.com/PrestaShop/PrestaShop/pull/18235/commits/001e445569e9009913564470183c238688b26456

src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
```diff
@@ -27,35 +27,155 @@
...
+     * @param string $expression
+     *
+     * @return string
+     */
+    public static function parseExpression($expression)
+    {
+        $pattern1 = '#\[(.*)\]#';
+        $pattern2 = '#is_granted\((.*),#';
+        $matches1 = [];
+        $matches2 = [];
```
PierreRambaud, line 58:
```suggestion
```

src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
```diff
@@ -27,35 +27,155 @@
...
+            'Action to perform, must be one of: %s',
+            implode(', ', self::getAvailableActions())
+        );
+
         $this
             ->setName('prestashop:linter:security-annotation')
-            ->setDescription('Checks if Back Office route controllers has configured Security annotations.')
-        ;
+            ->setDescription($description)
+            ->addArgument('action', InputArgument::REQUIRED, $actionDescription);
```
PierreRambaud, line 94: Don't use argument, prefer option so you will be able to do:
```
>addOption(
        'action',
        null,
        InputOption::VALUE_REQUIRED,
        'Action to perform',
        [self::ACTION_LIST_ALL, self::ACTION_FIND_MISSING]
    )

**Review by PierreRambaud — CHANGES_REQUESTED** [review]
Great work, few comments :+1:

File: src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php (hunk 4 above)
Comment (matks) [reply src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php:94]: But I expect an argument, not an option 😛 
We run `php vendor/bin/php-cs-fixer fix` and not `php vendor/bin/php-cs-fixer --action=fix`
Same here

Comment (PierreRambaud) [reply src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php:94]: You're right 👍

**Review by PierreRambaud — DISMISSED** [review]

File: src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php [hunk 5]
```diff
@@ -27,35 +27,155 @@
...
+        $matches2 = [];
+        preg_match($pattern1, $expression, $matches1);
+        preg_match($pattern2, $expression, $matches2);
+
+        if (count($matches1) > 1) {
+            return $matches1[1];
+        }
+        if (count($matches2) > 1) {
+            return $matches2[1];
+        }
```
Progi1984:
```suggestion

        if (count($matches1) > 1) {
            return $matches1[1];
        }
        preg_match($pattern2, $expression, $matches2);
        if (count($matches2) > 1) {
            return $matches2[1];
        }
```

matks: Still a fan of micro-optimization right 😁?

Progi1984: If we can avoid a regex, it's not an optimization.

matks: It's a linter command that is currently being run once per release (= once or twice every 6 months) 😄

Progi1984: Optimized code is good code, another code is code to optimize.

matks: We've had this discussion before 😛. If you want to say beautiful sentences, I will do it too:
- "Good is better than perfect"
- "Make it work, make it right, make it fast"

PierreRambaud: Octogone?

matks: 😄no I will do the requested changes because I want this approval 😛

PierreRambaud: Not funny :(

matks: I know you want to see BLOOD ALL OVER THE PLACE @PierreRambaud but not this time 😄

matks:
Applied @Progi1984 suggestion

Progi1984 (approved):

Progi1984 (approved):

src/PrestaShopBundle/Command/SecurityAnnotationLinterCommand.php
PierreRambaud: You don't need to declare these variables :thinking:

matks: Not declaring variables 🤔? I use them after to be given back the matches by `preg_match` . Even if it's possible, I think tools like phpstan will tell us it's not good

PierreRambaud: Phpstan will not trigger any errors because variables passed by reference are automatically created by PHP.

PierreRambaud (approved):

PierreRambaud:
Thanks @matks

PR #18200, opened 2020-03-18
Fix "weird behaviour when filling the quantity field with letters of symbols"

arouiadib:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       |1.7.7.x
| Description?  | This PR contains a front-end validation for the quantity input when adding/edditing product in order view page (migrated).
| Type?         | bug fix 
| Category?     |  BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18031
| How to test?  | In order view page (`sell/orders/orders/6/view`), add product to order , then fill the quantity field with a negative number or characters other than numerical characters, you will see the add button disabled. And so, it makes avoiding the creation of fake product.

PS:<del> I am also trying to find a good way to make a server-side validation also. Any feedback regarding this part is highly appreciated.</del>. I think server-side validation is already done by @jolelievre in his PR (just discovered this fact).

⚠️ Note: this PR https://github.com/PrestaShop/PrestaShop/pull/18168 (server side validation) should be tested and merged before this front validation, or it won't be possible to validate the server side easily any more. ✓

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18200)

PR #18254, opened 2020-03-21
Modify the quantity input type from text type to number type

arouiadib:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | This PR is a suggestion of the use of HTML5 number input type instead of text input type for quantity input. This input type has some advantages for front and back validation as discussed [here.](https://github.com/PrestaShop/PrestaShop/pull/18200)
| Type?         | improvement
| Category?     | BO 
| BC breaks?    |no
| Deprecations? | no
| Fixed ticket? |  Fixes nothing, but will help fixing another issue number 18031
| How to test?  | Go to migrated order view page and add a product to order.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18254)

PS: If suggestion approved, this feature should also be ported to edit product.

### DIFF
--- a/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/add_product_row.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/add_product_row.html.twig
@@ -44,7 +44,7 @@
     {{ form_row(addProductRowForm.price_tax_included) }}
   </td>
   <td class="pr-2">
-    {{ form_row(addProductRowForm.quantity) }}
+    {{ form_row(addProductRowForm.quantity, {'type':'number'}) }}
   </td>
   <td id="addProductLocation"></td>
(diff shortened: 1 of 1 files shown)


Arman-Hosseini (approved):

NeOMakinG (approved):

PierreRambaud (changes requested):
The type Number must be used in the FormType and not in the template :thinking:

arouiadib:
Hello @PierreRambaud , the `numberType` is already used in form building, but I don't know why it is not enough to make it work.  When adding it here in template, it works! This is similar to other situation (add discount in order view). I need some time to inspect more about this question because I don't understand what is happening! Thank you very much for raising this point.

PierreRambaud:
@arouiadib Thanks to you for helping us :wink:

arouiadib:
I close this PR because its only purpose was to test a server-side validation while the front-end one doesn't exist. No longer useful.

PR #13008, opened 2019-03-22
Adds phpdoc on range objectmodel

Matt75:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Adds phpdoc on range objectmodel. Hope this can avoid mistake with usage of rangeExist() who doesn't return id of existing range. Helpful for carrier module developers too. It should return boolean but I don't want make a BC.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | Should work as before

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/13008)

PR #18259, opened 2020-03-22
Update phpdoc Product class

Matt75:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add and fix phpdoc for Product class
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | No need for QA : no change

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18259)

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -24,13 +24,10 @@
  * International Registered Trademark & Property of PrestaShop SA
  */
 
-/**
- * @deprecated 1.5.0.1
- */
+// Deprecated since 1.5.0.1 use Product::CUSTOMIZE_FILE
 define('_CUSTOMIZE_FILE_', 0);
-/*
- * @deprecated 1.5.0.1
- */
+
+// Deprecated since 1.5.0.1 use Product::CUSTOMIZE_TEXTFIELD
 define('_CUSTOMIZE_TEXTFIELD_', 1);
 
 use PrestaShop\PrestaShop\Adapter\ServiceLocator;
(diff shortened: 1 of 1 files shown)


classes/Product.php
```diff
@@ -6945,11 +7647,19 @@ public function modifierWsLinkRewrite()
         return true;
     }
 
+    /**
+     * @return array
+     */
     public function getWsProductBundle()
     {
         return Db::getInstance()->executeS('SELECT id_product_item as id, id_product_attribute_item as id_product_attribute, quantity FROM ' . _DB_PREFIX_ . 'pack WHERE id_product_pack = ' . (int) $this->id);
     }
```
Arman-Hosseini, line 7651: https://github.com/PrestaShop/PrestaShop/blob/41951d81761cbed89c711c3be96073c091ad1e92/classes/db/Db.php#L574-L632

Arman-Hosseini: See the return type

Matt75: https://github.com/PrestaShop/PrestaShop/pull/13008#discussion_r269555970

Arman-Hosseini: I do not agree with it.
The code should look like this:

```
    /**
     * @return array
     */
    public function getWsProductBundle()
    {
        $result = Db::getInstance()->executeS('SELECT id_product_item as id, id_product_attribute_item as id_product_attribute, quantity FROM ' . _DB_PREFIX_ . 'pack WHERE id_product_pack = ' . (int) $this->id);
        if (!is_array($result))
            $result = [];

        return $result;
    }
```

Matt75: Core team have final word, so let's wait core team review ;)

Goal is only to add phpdoc and not to make code changes because all code change need tests from QA and we don't have enough resources in QA team.

So I'm more in favor of use `mixed` return type here and add a comment with expected result if everything work as expected.

Arman-Hosseini: Yes, the core team review certainly says the last word, but I want to say something that is reasonably correct. We come together to get feedback and create a quality product.

Matt75: Legacy classes must die but not yet because migration is not achieved.
So we have to do with, invest time to refactor this is waste time and we can't make breaking changes.
Best practice is to use it inside an Adapter to rely on more trusted results and it helps with retrocompatiblity when breaking change has been made in the past but it need an extra work from developers.

classes/Product.php
```diff
@@ -293,29 +347,61 @@ class ProductCore extends ObjectModel
...
+     * @deprecated Since 1.5.6.1
      *
      * @var array is deprecated since 1.5.6.1
      */
     protected static $_cart_quantity = [];
 
+    /**
+     * @deprecated Since 1.5.0.9
+     *
+     * @var array is deprecated since 1.5.0.9
```
Progi1984:
```suggestion
     * @var array
```

PierreRambaud (approved):

Progi1984 (approved):

Progi1984:
Thanks @Matt75

PR #18293, opened 2020-03-24
Fix translation export not working

### LINKED ISSUE #16555: Export language seems to be KO in 176
**Describe the bug**
In PS 1761 and 1762 build 2, the export of language seems to be KO, the zip file created does not contains any translations.

**To Reproduce**
1. Go to International > Translations > Export a language
2. Export a language
3. ZIP file is created, with only one xlf file in it (without any translation in it). In 1.7.5.2, there is around 10 files (with translations).
See ZIP files.

**Files**

[classic.ru-RU 175.zip](https://github.com/PrestaShop/PrestaShop/files/3893168/classic.ru-RU.175.zip)

[classic.ru-RU 1762.zip](https://github.com/PrestaShop/PrestaShop/files/3893169/classic.ru-RU.1762.zip)

**Additional information**
PrestaShop version: 1.7.6.1 - 1.7.6.2 build 2
PHP version: N/A

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | This change fixes the theme translation export feature. Read below for more.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16555
| How to test?  | Read below

## The issue

- The translation export feature was exporting a single file called `fr-FR.xlf` (for the French example). 
- Already existing translations were being ignored
- Only translations performed through the database were actually exported

## Why it was happening

First, there were multiple problems regarding conventions:

- paths in return values (sometimes it expected the path to include the locale, sometimes not)
- domain names in message catalogues (with or without dots, with or without the locale suffix...)

Then, when loading theme translations for export, the suite went like this:

`ThemeExporter::exportCatalogues()` calls `ThemeExtractor::extract()`, which extracts the theme's "default catalogue" by interpreting .tpl files using TranslationToolsBundle. After that, `extract()` overrides that catalogue using the core's "default catalogue" – more on this in a moment. Finally, it dumps this catalogue in XLF files.

Back to `exportCatalogues()`, the extracted and enriched catalogue is loaded from the XLF files built above, and further enriched by loading translations found in the theme itself (if XLF files are found), then in the database. This catalogue is dumped again to XLF files, zipped and sent away.

So, one of the problems was that the inconsistencies in conventions led to problems in the name of the output files (hence the file named `fr-FR.xlf`). The other problem was during the enrichment process. 

Earlier I described that the catalogue was enriched using the core's "default catalogue". This is actually a language abuse, because it actually pulls the _translated_ catalogue, not _default_ one. This catalogue was being emptied out for some reason, by default[1]. This resulted in all wordings from the theme also found in the core catalogue being emptied out from the theme's catalogue. This also meant that wordings were translated to an empty string, therefore removed from the exported catalogue.

Since we're actually loading the translated catalogue, not the default one, I think it would make more sense to _keep_ those translations. I solved this by simply not emptying out that "default catalogue" loaded from the core.

[1]: Update: looks like this was done to empty out not translated fields in the translation interface, so this PR introduces a regression where you can't know if a theme is fully translated or not (not translated fields will be filled with the original wording).

#### First case: using the classic theme

1. Export the theme's translations in a given language. You should basically get a zip file containing several files, all of them start with "Shop", and they are all translated to that language (or at least to the same extent as the core).
2. Now go to the back office's translation section for that theme, and change some translations. Export the theme again. The xlf file for the related domain should be impacted by the modifications you made.
3. Export the theme. It should contain all of the translation files for all the installed languages and include any modifications you made. FYI: this uses the same code as the "export theme translations" feature, only that it does it for ALL the installed languages at once.

#### Second case: using a custom theme (NOT classic), that includes translation files

Repeat steps from the classic theme. The exported files should be the same as the ones you already had for that theme, and when you modify things in the translation interface, they should be exported as well.

#### Third case: using a custom theme (NOT classic), NO translation files included in it

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18293)

### DIFF
--- a/src/PrestaShopBundle/Service/TranslationService.php
+++ b/src/PrestaShopBundle/Service/TranslationService.php
@@ -36,6 +36,7 @@ use Symfony\Component\Validator\Validation;
 
 class TranslationService
 {
+    const DEFAULT_THEME = 'classic';
     /**
      * @var Container
      */
@@ -114,7 +115,7 @@ class TranslationService
         $factory = $this->container->get('ps.translations_factory');
 
         if ($this->requiresThemeTranslationsFactory($theme, $type)) {
-            if ('classic' === $theme) {
+            if ($this->isDefaultTheme($theme)) {
                 $type = 'front';
             } else {
                 $type = $theme;
@@ -154,7 +155,7 @@ class TranslationService
      */
     public function listDomainTranslation($locale, $domain, $theme = null, $search = null, $module = null)
     {
-        if (!empty($theme) && 'classic' !== $theme) {
+        if (!empty($theme) && !$this->isDefaultTheme($theme)) {
             $translationProvider = $this->container->get('prestashop.translation.theme_provider');
             $translationProvider->setThemeName($theme);
         } else {
@@ -179,7 +180,7 @@ class TranslationService
             'data' => array(),
         );
         $treeDomain = preg_split('/(?=[A-Z])/', $domain, -1, PREG_SPLIT_NO_EMPTY);
-        if (!empty($theme) && 'classic' !== $theme) {
+        if (!empty($theme) && !$this->isDefaultTheme($theme)) {
             $defaultCatalog = current($translationProvider->getThemeCatalogue()->all());
         } else {
             $defaultCatalog = current($translationProvider->getDefaultCatalogue()->all());
@@ -351,5 +352,15 @@ class TranslationService
         }
 
         return $resetTranslationSuccessfully;
+    }
+
+    /**
+     * @param string $theme
+     *
+     * @return bool
+     */
+    private function isDefaultTheme($theme)
+    {
+        return static::DEFAULT_THEME === $theme;
     }
 }
--- a/src/Core/Addon/Theme/ThemeExporter.php
+++ b/src/Core/Addon/Theme/ThemeExporter.php
@@ -114,11 +114,7 @@ class ThemeExporter
                 $catalogueDir = $this->translationsExporter->exportCatalogues($theme->getName(), $locale);
             }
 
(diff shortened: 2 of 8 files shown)


matks:
Failing unit test
```
1) LegacyTests\Unit\Classes\ValidateCoreTest::testIsBirthDate with data set #3 (false, '2020-03-19')
Failed asserting that true matches expected false.
/home/travis/build/PrestaShop/PrestaShop/tests-legacy/Unit/Classes/ValidateCoreTest.php:66
```

src/PrestaShopBundle/Service/TranslationService.php
```diff
@@ -35,6 +35,7 @@
 
 class TranslationService
 {
+    const DEFAULT_THEME = 'classic';
```
matks, line 39: There is a similar constant in `FrontOfficeProvider` you might want to reuse it.

eternoendless:
> Failing unit test
> 
> ```
> 1) LegacyTests\Unit\Classes\ValidateCoreTest::testIsBirthDate with data set #3 (false, '2020-03-19')
> Failed asserting that true matches expected false.
> /home/travis/build/PrestaShop/PrestaShop/tests-legacy/Unit/Classes/ValidateCoreTest.php:66
> ```

Lol, absolutely unrelated test... hadn't we fix that test already? Was in 1.7.7.x maybe?

eternoendless:
Rebased to fix the failing unrelated test

eternoendless:
This PR introduces regressions, I will have to refactor the translation interface for 1.7.7. Read this comment for more: https://github.com/PrestaShop/PrestaShop/issues/16555#issuecomment-617677165

Issue #18312, opened 2020-03-25
[Create order] Product images are wrong in cart details


colinegin:
In Create order page, when clicking on Details of a previously created cart, product images displayed are not correct. See screenshot below : the products in the cart use the combination "White", but associated images show black products. 

![image](https://user-images.githubusercontent.com/35263827/77568364-cf615b00-6ec8-11ea-8219-baf9de981f4d.png)

It is correct once you use the cart in the new cart.

I have tested it in 1.7.6.4, it works correctly.

zuk3975:
just a note - technically its more like Order->Shipping Carts page bug (as the `details` redirects to cart view page), which is not yet enabled. Anyway, its only accessible from order create page in 1.7.7.x so I guess its not wrong to call it order bug. Ill take a look.

**Btw the page doesnt seem to show gift products either.**

zuk3975:
blocked by https://github.com/PrestaShop/PrestaShop/pull/17179

colinegin:
@zuk3975 it does show gift products. You don't see it on the screenshot but in this cart I had added 2 cart rules with the t-shirt as a gift product.

zuk3975:
> @zuk3975 it does show gift products. You don't see it on the screenshot but in this cart I had added 2 cart rules with the t-shirt as a gift product.

Oh right, gift products show up, but those supposed to be marked as GIFT I guess instead of having prices. Same as in create order page.

colinegin:
About your last comment, we need to discuss with @MatShir to see what's the expected behaviour on this page. Thanks !

Progi1984:
Fixed by #18334

PR #18318, opened 2020-03-26
Fix wrong images for combinations in Cart view page

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | In Orders -> Shipping carts -> Cart view page, product images displayed are not correct when product contains combinations.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | issued by #18312, but moved to develop because carts page is not enabled yet in 1.7.7
| How to test?  | Go to Orders -> Shipping carts and select a cart which has products with combinations. (you can create a cart in create order page). Make sure that product images are correct. ( e.g. if you have product `.. black cushion` and `...white cushion` then you should see images with black and white cushions accordingly. Previously it would show only black ones.)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18315)

### DIFF
--- a/src/Adapter/Cart/QueryHandler/GetCartForViewingHandler.php
+++ b/src/Adapter/Cart/QueryHandler/GetCartForViewingHandler.php
@@ -33,6 +33,7 @@ use Customer;
 use DateTime;
 use Gender;
 use Group;
+use Link;
 use Order;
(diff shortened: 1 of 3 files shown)


zuk3975:
closing due to changed PM decision.
Final fix here #18334

PR #18321, opened 2020-03-26
Do not stack free shipping vouchers on total: shipping fees applied once

### LINKED ISSUE #17875: Free shipping vouchers in FO and BO should not be cumulative
**Describe the bug**
When adding two (or more) cart rules in FO and BO which include Free Shipping, they are both added, only on should be added.

**To Reproduce on FO**
1. Go to BO => Catalog => Discounts, create two cart rules with free shipping
2. Go to FO, create an order with paid shipping
3. Add the two cart rules
4. Shipping price is deducted twice

**Screenshots**
two shipping vouchers

**Additional information**
PrestaShop version: 1.7.7
PHP version: N/A

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Do not remove shipping fees twice even if multiple 'free shipping' vouchers are used
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/17875
| How to test?  | See below

## How to test

Steps from https://github.com/PrestaShop/PrestaShop/issues/17875
1. Go to BO => Catalog => Discounts, create two cart rules with free shipping
2. Go to FO, create an order with paid shipping
3. Add the two cart rules
4. Shipping price is deducted twice

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18321)

### DIFF
--- a/src/Core/Cart/Calculator.php
+++ b/src/Core/Cart/Calculator.php
@@ -200,8 +200,14 @@ class Calculator
     public function getDiscountTotal()
     {
         $amount = new AmountImmutable();
+        $isFreeShippingAppliedToAmount = false;
         foreach ($this->cartRules as $cartRule) {
+            if ($isFreeShippingAppliedToAmount && (bool) $cartRule->getRuleData()['free_shipping']) {
+                $initialShippingFees = $this->getFees()->getInitialShippingFees();
+                $amount = $amount->sub($initialShippingFees);
+            }
             $amount = $amount->add($cartRule->getDiscountApplied());
+            $isFreeShippingAppliedToAmount = (bool) $cartRule->getRuleData()['free_shipping'];
         }
 
         $allowedMaxDiscount = $this->getRowTotalWithoutDiscount();
--- a/src/Core/Cart/Calculator.php
+++ b/src/Core/Cart/Calculator.php
@@ -202,12 +202,14 @@ class Calculator
         $amount = new AmountImmutable();
         $isFreeShippingAppliedToAmount = false;
         foreach ($this->cartRules as $cartRule) {
-            if ($isFreeShippingAppliedToAmount && (bool) $cartRule->getRuleData()['free_shipping']) {
-                $initialShippingFees = $this->getFees()->getInitialShippingFees();
-                $amount = $amount->sub($initialShippingFees);
+            if ((bool) $cartRule->getRuleData()['free_shipping']) {
+                if ($isFreeShippingAppliedToAmount) {
+                    $initialShippingFees = $this->getFees()->getInitialShippingFees();
+                    $amount = $amount->sub($initialShippingFees);
+                }
+                $isFreeShippingAppliedToAmount = true;
             }
             $amount = $amount->add($cartRule->getDiscountApplied());
-            $isFreeShippingAppliedToAmount = (bool) $cartRule->getRuleData()['free_shipping'];
         }
 
         $allowedMaxDiscount = $this->getRowTotalWithoutDiscount();
--- a/src/Core/Cart/Fees.php
+++ b/src/Core/Cart/Fees.php
@@ -64,6 +64,12 @@ class Fees
      */
     protected $isProcessed = false;
 
+    public function __construct()
+    {
+        $this->shippingFees = new AmountImmutable();
+    }
+
+
     /**
      * @param Cart $cart
      * @param CartRowCollection $cartRowCollection
--- a/src/Core/Cart/Fees.php
+++ b/src/Core/Cart/Fees.php
@@ -69,7 +69,6 @@ class Fees
         $this->shippingFees = new AmountImmutable();
     }
 
-
     /**
      * @param Cart $cart
      * @param CartRowCollection $cartRowCollection
--- a/tests/Integration/Behaviour/Features/Context/AbstractPrestaShopFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/AbstractPrestaShopFeatureContext.php
@@ -41,11 +41,19 @@ abstract class AbstractPrestaShopFeatureContext implements BehatContext
 
     protected function checkFixtureExists(array $fixtures, $fixtureName, $fixtureIndex)
     {
+        $searchLength = 10;
+
         if (!isset($fixtures[$fixtureIndex])) {
             $fixtureNames = array_keys($fixtures);
-            $firstFixtureNames = array_splice($fixtureNames, 0, 5);
+            $firstFixtureNames = array_splice($fixtureNames, 0, $searchLength);
             $firstFixtureNamesStr = implode(',', $firstFixtureNames);
-            throw new \Exception($fixtureName . ' named "' . $fixtureIndex . '" was not added in fixtures. First 5 added are: ' . $firstFixtureNamesStr);
+            throw new \Exception(sprintf(
+                '%s named "%s" was not added in fixtures. First %d added are: %s',
+                $fixtureName,
+                $fixtureIndex,
+                $searchLength,
+                $firstFixtureNamesStr
+            ));
         }
     }
 }
--- a/tests/Integration/Behaviour/Features/Context/AbstractPrestaShopFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/AbstractPrestaShopFeatureContext.php
@@ -47,7 +47,7 @@ abstract class AbstractPrestaShopFeatureContext implements BehatContext
             $fixtureNames = array_keys($fixtures);
             $firstFixtureNames = array_splice($fixtureNames, 0, $searchLength);
             $firstFixtureNamesStr = implode(',', $firstFixtureNames);
-            throw new \Exception(sprintf(
+            throw new \RuntimeException(sprintf(
                 '%s named "%s" was not added in fixtures. First %d added are: %s',
                 $fixtureName,
                 $fixtureIndex,
--- a/tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/CartRule/free_shipping.feature
+++ b/tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/CartRule/free_shipping.feature
@@ -41,7 +41,6 @@ Feature: Cart rule (amount) calculation with one cart rule offering free shippin
     Given cart rule "cartrule6" has a discount code "foo6"
     Given cart rule "cartrule7" offers free shipping
     Given cart rule "cartrule7" has a discount code "foo7"
-    Given cart rule "cartrule8" offers free shipping
     Given cart rule "cartrule8" has a discount code "foo8"
     When I add 1 items of product "product1" in my cart
     When I use the discount "cartrule5"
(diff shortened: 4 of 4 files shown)


src/Core/Cart/Fees.php
```diff
@@ -64,6 +64,12 @@ class Fees
      */
     protected $isProcessed = false;
 
+    public function __construct()
+    {
+        $this->shippingFees = new AmountImmutable();
```
matks, line 69: When `$this->getFees()->getInitialShippingFees()` is called on `Calculator::getDiscountTotal()` it's going to go through a huge number of legacy and complex function calls which is going to try to compute fees by getting carriers list, delivery options, carrier whatever-I-dont-know ... and at some point it's going to call `Order::getTotal()` which is going to call `Calculator::getDiscountTotal()` 😱 this is a recursion loop.

At this moment `$this->getFees()->getInitialShippingFees()` will return `null` and next line will execute `$amount = $amount->sub(null);` which will crash.

From my debugging process it seems that when not computed yet `$this->shippingFees` should be an `AmountImmutable` instance with properties set  to `0.0`. So I initialize this property here in order for it to never ne null, and I hope it works 🙏 .

According to our behat tests it works 🤔

tests/Integration/Behaviour/Features/Context/AbstractPrestaShopFeatureContext.php
```diff
@@ -41,11 +41,19 @@ abstract class AbstractPrestaShopFeatureContext implements BehatContext
...
     {
+        $searchLength = 10;
+
         if (!isset($fixtures[$fixtureIndex])) {
             $fixtureNames = array_keys($fixtures);
-            $firstFixtureNames = array_splice($fixtureNames, 0, 5);
+            $firstFixtureNames = array_splice($fixtureNames, 0, $searchLength);
             $firstFixtureNamesStr = implode(',', $firstFixtureNames);
-            throw new \Exception($fixtureName . ' named "' . $fixtureIndex . '" was not added in fixtures. First 5 added are: ' . $firstFixtureNamesStr);
+            throw new \Exception(sprintf(
```
matks: Improved a bit this helper/debug function

matks:
@Robin-Fischer-PS This PR does make sure, when 2 "free shipping" vouchers are used, that the shipping fees are not removed twice from cart total both on FO and BO

tests/Integration/Behaviour/Features/Context/AbstractPrestaShopFeatureContext.php
jolelievre: Do you think this could be removed by `RuntimeException` to homogenize with all the other context classes? (with use statement to avoid the \)

jolelievre (dismissed):
I added a suggestion or context class but it's not mandatory, so I still approve the PR

Feel free to change the exception or not ^^

src/Core/Cart/Calculator.php
```diff
@@ -200,8 +200,14 @@ public function getRowTotalWithoutDiscount()
...
     {
         $amount = new AmountImmutable();
+        $isFreeShippingAppliedToAmount = false;
         foreach ($this->cartRules as $cartRule) {
+            if ($isFreeShippingAppliedToAmount && (bool) $cartRule->getRuleData()['free_shipping']) {
+                $initialShippingFees = $this->getFees()->getInitialShippingFees();
+                $amount = $amount->sub($initialShippingFees);
+            }
             $amount = $amount->add($cartRule->getDiscountApplied());
+            $isFreeShippingAppliedToAmount = (bool) $cartRule->getRuleData()['free_shipping'];
```
atomiix: I'm not sure to fully understand but if i'm not wrong, you want to keep `isFreeShippingAppliedToAmount` to true as soon as the first free shipping cart rule is applied until the end?

```suggestion
            if ((bool) $cartRule->getRuleData()['free_shipping']) {
                if ($isFreeShippingAppliedToAmount) {
                    $initialShippingFees = $this->getFees()->getInitialShippingFees();
                    $amount = $amount->sub($initialShippingFees);
                }
                $isFreeShippingAppliedToAmount = true;
            }
            $amount = $amount->add($cartRule->getDiscountApplied());
```

jolelievre: That makes sense!

src/Core/Cart/Fees.php
arouiadib: Finally I found a comment which describes me in front of Calculator and CartRuleCalculator files . This re-comfort me now, I was thinking I am stupid in front of the the loop:D

matks: @arouiadib Unfortunately no you are not insane. It's this code which is insane. Recursion loop happen in a lot of computations for product prices, cart total prices and cart rules prices and is making any attempt to fix part of the code an herculean task.

Unfortunately we cannot delete all this code because of SemVer although it seems to be the only relevant solution to fix all these crazy cart rules bugs. It must wait for PS 8.0.0 where we will be allowed to perform such a huge BC break.

We had the same issue for PR https://github.com/PrestaShop/PrestaShop/pull/12878

src/Core/Cart/Calculator.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/18321/commits/6dd5baa266fbae61f58215040303d7d664442769
I modified the behat test in order to reproduce @atomiix usecase and indeed shipping fees were deducted multiple times. Suggested code fix fixes the usecase.

tests/Integration/Behaviour/Features/Context/AbstractPrestaShopFeatureContext.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/18321/commits/6dd5baa266fbae61f58215040303d7d664442769

atomiix (approved):

src/Core/Cart/Fees.php
arouiadib: @matks , thank you for your usual informative comments. I totally agree with you in parts I understand.

I think one of the good points is that there is a number of behat tests now. Till the day comes, we will be adding more tests, and no matter is the complexity of the current system, the swap to a new system will be easier in 1.8. I hope I am not wrong!
For sure, you are not meaning PS 8 but PS 1.8 :p :)
One another thing I see about cart rules, is that, still we need to discuss them before fixing any issue (in current system or in 1.8). If we take the current bug you are fixing as example, some merchants may think that free shipping should be cumulative and that free shipping vouchers should be stack in total (we can discuss this point later if you want but it is not the purpose of my comment). 

The problem is when that merchant raises a ticket here in github, that might end up with a lot of time wasted if there is no link to send back telling him how Prestashop is supposed to calculate his cart rule. In other words, what is the equation Prestashop will be using to calculate that cart rule, for all times.

He can decide then whether to wait for the fix from github or to hire a developer. For Prestashop, it will result in less questions to POs , less rethininking of things, less...less... all the advantages of specs in software industry.

I still believe that cart rules specs must be the first written in the whole specs of Prestashop due to its complexity, relativity with total amount and invoices....

Sorry for being long

arouiadib: What I wrote is not maybe in the appropriate time and tone. So don't take this matter seriously :redface

matks: > @matks , thank you for your usual informative comments. I totally agree with you in parts I understand.
> 
> I think one of the good points is that there is a number of behat tests now. Till the day comes, we will be adding more tests, and no matter is the complexity of the current system, the swap to a new system will be easier in 1.8. I hope I am not wrong!
> For sure, you are not meaning PS 8 but PS 1.8 :p :)

I mean PS 8.0.0 .
Although it is not decided 100% yet we think that, for next major version, we will stop using the 1.X.Y.Z version numbers and start using the X.Y.Z version number standard.

So if PS 1.7.9 is the last PS 1.7 minor release, next release will be PS 8.0.0 .
It changes nothing about the project, it's just a version number. PS 1.8.0.0 or PS 8.0.0 is the same.

> One another thing I see about cart rules, is that, still we need to discuss them before fixing any issue (in current system or in 1.8). If we take the current bug you are fixing as example, some merchants may think that free shipping should be cumulative and that free shipping vouchers should be stack in total (we can discuss this point later if you want but it is not the purpose of my comment).

@Robin-Fischer-PS (who is the author of the issue which states free shipping vouchers should not be cumulative) is always in discussion with Product Team and Product Team is writing down these specs into https://github.com/PrestaShop/prestashop-specs

> The problem is when that merchant raises a ticket here in github, that might end up with a lot of time wasted if there is no link to send back telling him how Prestashop is supposed to calculate his cart rule. In other words, what is the equation Prestashop will be using to calculate that cart rule, for all times.

Indeed it must go into https://github.com/PrestaShop/prestashop-specs . Product Team is on this topic

> Sorry for being long

I actually appreciate you took so much time to write this comment, it means you care about the project and we ❤️ having dedicated contributors

> What I wrote is not maybe in the appropriate time and tone. So don't take this matter seriously 😊

Your tone and time are appropriate and the topic you raise is very important. If you have other similar topics / feedbacks please share them with us, it's very important and we appreciate this. Your feedbacks have a lot of value 👍

arouiadib: I thank you very much @matks . 
It is always a great pleasure to be part of a big masonry. It is a beautiful building, where just one brick contribution makes you feel helpful. As you know, Prestashop is a great source of income in normal days, and a rescue for many businesses in special days. For me, it a also special building: a university where I always learn new things from experts.

I will be glad to contribute in "free Shipping discount" specs. I will be writing comments and sharing them with @Robin-Fischer-PS so that we can think of all possible scenarios.

Again thank you all guys :)

arouiadib: Hello @Robin-Fischer-PS @matks,
I have gathered what I think about the shipping discount calculation in: https://github.com/PrestaShop/prestashop-specs/issues/107. I will be glad to explain more if something is unclear :)
Regards

src/Core/Cart/Calculator.php
```diff
@@ -200,7 +200,15 @@ public function getRowTotalWithoutDiscount()
     public function getDiscountTotal()
     {
         $amount = new AmountImmutable();
+        $isFreeShippingAppliedToAmount = false;
         foreach ($this->cartRules as $cartRule) {
+            if ((bool) $cartRule->getRuleData()['free_shipping']) {
+                if ($isFreeShippingAppliedToAmount) {
+                    $initialShippingFees = $this->getFees()->getInitialShippingFees();
+                    $amount = $amount->sub($initialShippingFees);
+                }
```
Progi1984, line 209:
```suggestion
                if (!$isFreeShippingAppliedToAmount) {
                    $initialShippingFees = $this->getFees()->getInitialShippingFees();
                    $amount = $amount->sub($initialShippingFees);
                }
```

This must be applied only the first time when the variable is defined at `false`

Progi1984 (approved):

SD1982:
This PR correctly fix #17875 ! Thanks @matks .

matks:
Thank you for review and QA

Issue #18030, opened 2020-03-06
Bug when adding more than 600 characters in message block 


colinegin:
On the order page, we have this text displayed on top of the page when adding more than 600 chars in the message block : 
`<h1 class="d-inline"> <strong class="text-muted">#5</strong> <strong>KHWLILZLL</strong> </h1> <p class="lead d-inline"> from John DOE </p> <span class="badge rounded badge-dark ml-2 mr-2 font-size-100"> 20,90 € </span> <p class="lead d-inline"> 2020-02-28 <span class="text-muted"> at 17:15:36 </span> </p>`

jolelievre:
Hi @colinegin,
I tried to reproduce this bug but it seems impossible to add more than 600 characters, the input blocks adding more characters, so how did you manage to add more than 600?

![Capture d’écran 2020-03-18 à 11 06 19](https://user-images.githubusercontent.com/13801017/76949232-862d6c00-6908-11ea-8df8-7a4f8e579c4f.png)

Also do you have a screenshot of the bug? Thank you

colinegin:
I can try to reproduce it, but what you can try is use a predefined message and then copy paste it a few times, and it this case you should be able to add more than 600 characters. Otherwise, by just writing text, indeed you are blocked when reaching the max limit.

arouiadib:
I agree with the fact that it is not always reproducible and that you need copy-paste to see the error
![image](https://user-images.githubusercontent.com/23559010/76968714-a3befd80-6929-11ea-9379-f3802ec9f77a.png)

colinegin:
Thanks for the screenshot @arouiadib ;)

zuk3975:
Couldn't reproduce it nor by writing manually nor by pasting, however I noticed the following inconsistency:
Predefined message (when editing in sell/customer-service/order-messages) allows to fill up to 1200 characters, but here in order message its limited to 600. We must equalize these numbers IMO., someone just needs to decide to leave 600 or 1200

colinegin:
After more investigation, let me be more precise about how to reproduce this bug :
- copy paste a few times a preset message - indeed the block seems to be blocking the number of characters
- save
- check in the header, you have the error message as on @arouiadib's screenshot.

About your 2nd comment, you're right @zuk3975, let's be more consistent. I would recommend to accept 600 characters everywhere, but what would happen to merchants who had messages with more characters and update to 177 ?

MatShir:
Hi @zuk3975,

[...]

PR #18326, opened 2020-03-26
Fix messages block in order view page

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Customer service->message and Order view ->message length constraints were different (one being 600 and another 1200). Now both will have max length of 1200 symbols. Also bug mentioned in #18030 is fixed.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #18030
| How to test?  | **Check length:**`Sell->Customer service->Order messages edit` And `Sell->Orders->View order messages block` must have the same maximum length. **Check bug:** go to order view page messages block and do following: inspect the textarea element and remove the 'max-length' attribute. Then fill in random text with more than 1200 symbols and submit. Page should display error message that maximum allowed length is only 1200 and strange html should not appear at the top of the page anymore (reference #18030 for more info).

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18326)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
@@ -1096,16 +1096,16 @@ class OrderController extends CommonController
 
         $routesCollection = $this->get('router')->getRouteCollection();
 
-        if (
-            null !== $routesCollection &&
+        if (null !== $routesCollection &&
             !$orderMessageForm->isValid() &&
-            $viewRoute = $routesCollection->get('admin_orders_view')) {
+            $viewRoute = $routesCollection->get('admin_orders_view')
+        ) {
+            $attributes = $viewRoute->getDefaults();
+            $attributes['orderId'] = $orderId;
+
             return $this->forward(
                 $viewRoute->getDefault('_controller'),
-                [
-                    'orderId' => $orderId,
-                ],
-                $viewRoute->getDefaults()
+                $attributes
             );
         }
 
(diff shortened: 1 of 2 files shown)


matks (dismissed):

zuk3975:
rr travis

zuk3975:
blocking #18339

matks (approved):

PierreRambaud (approved):

Robin-Fischer-PS:
Hi @zuk3975 !

Your fix is almost OK, only one question : When I remove the 'max-length' attribute in the textarea element, there is no error message at the top. There is only the error message under the "message" block.

Is this the expected behavior or should we also have a message at the top of the screen ?

pr18326 no error message

Thanks !

arouiadib (approved):

zuk3975:
> Your fix is almost OK, only one question : When I remove the 'max-length' attribute in the textarea element, there is no error message at the top. There is only the error message under the "message" block.

Hi, @Robin-Fischer-PS , Its expected, I don't think we need additional message
Is this the expected behavior or should we also have a message at the top of the screen ?

Robin-Fischer-PS:
Hi @zuk3975 ! I discussed this with @MatShir , we are okay with the current behaviour, there is no need to add an additional message.

So this PR is QA :heavy_check_mark: , thanks !

MatShir:
FYI: [The related specs ](https://github.com/PrestaShop/prestashop-specs/commit/741807e8e68f7a21517f28f2828e6810415390bd) 
Thanks, @zuk3975 for your contribution ;)

Progi1984:
Thanks @zuk3975

PR #18334, opened 2020-03-26
Add FancyBox in create order page details button

### LINKED ISSUE #18303: [Create order] Issue when clicking on Customer details
New issue to fix the last bug related to the issue #17223.
when clicking "Details" for a customer, a cart or a previous order, it takes me out of the window, when it should rather open the details with minimal view inside a modal or similar?

### LINKED ISSUE #18312: [Create order] Product images are wrong in cart details

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Use fancy box on Customer, Order and Cart details buttons in the create order page Also add some loading messages to improve user experience Finally fix the cart view to display the combination image
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18303 and #18312
| How to test?  | Create a new order, you can now see the order, cart and customer details inside a fancy box (no more redirections to those pages)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18334)


admin-dev/themes/new-theme/js/pages/order/create/create-order-map.js
```diff
@@ -74,6 +75,7 @@ export default {
   orderPaymentMethod: '.js-order-payment-method',
   orderStatusField: '.js-order-status',
   emptyListRowTemplate: '#js-empty-list-row',
+  loadingListRowTemplate: '#js-loading-list-row',
```
NeOMakinG, line 78: I think a JS selector should never be an ID as it should be reusable :/

What do you think ?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/customer.html.twig
```diff
@@ -222,3 +234,14 @@
...
+<script id="js-loading-list-row" type="text/template">
+  <tr class="empty_row js-empty-row">
+    <td colspan="8" class="border-0">
+      <div class="text-center grid-table-empty">
+        <p class="mb-0 mt-2"><i class="material-icons">autorenew</i></p>
+        <p class="mb-2">{{ 'Loading'|trans({}, 'Admin.Global') }}</p>
+      </div>
+    </td>
+  </tr>
+</script>
```
NeOMakinG: Would this be usefull on some other pages ? If yes, that would be cool to add it to components folders to make it reusable using a twig include with parameters, with a generic class + class parameter if you want to select it more precisely

NeOMakinG (commented):
Added some comments as it looks like there is a reusable component that could be done, that's a suggestion, feel free to explain it :)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/customer.html.twig
```diff
@@ -222,3 +234,14 @@
     </td>
   </tr>
 </script>
+
+<script id="js-loading-list-row" type="text/template">
+  <tr class="empty_row js-empty-row">
+    <td colspan="8" class="border-0">
+      <div class="text-center grid-table-empty">
+        <p class="mb-0 mt-2"><i class="material-icons">autorenew</i></p>
+        <p class="mb-2">{{ 'Loading'|trans({}, 'Admin.Global') }}</p>
```
LouiseBonnard:
```suggestion
        <p class="mb-2">{{ 'Loading...'|trans({}, 'Admin.Global') }}</p>
```

admin-dev/themes/new-theme/js/pages/order/create/create-order-map.js
zuk3975: indeed if template will be used multiple times, it will generate html with duplicated ids, which will issue warnings. Same mistake was done in order view page i guess (see screenshot).
Screenshot from 2020-03-27 11-08-06
On the other hand, if it only stores the template inside and the id is not rendered multiple times, then it might be the best option.

jolelievre: This one is a bit special, it is used to get the template that is cloned It maps a `script` tag, and we only duplicate its content (which has no ids, only classes):
`const $emptyTableRow = $($(createOrderMap.loadingListRowTemplate).html()).clone();`

I could use a class to identify this script but it wouldn't make much sense because it really is not a component, more like a component resource at worst And it's probably better to identify it via an id to emphasize its difference

Now the whole script/template may not be a good practice may not be the best, but it's been used a lot in all pages so it's not the purpose o this PR to refacto all this

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/customer.html.twig
jolelievre: I'm not sure it could be used like this in other pages, because it is really dependent on where it will be included (in that case a table with 8 columns..) Besides we don't have much `Loading` templates or now (sadly)

So I'm not sure any other pages use this kind o feature, but i we start generalizing this kind of display I guess the message part could be extracted in a common template (just the div part in that case) But it might be too early or this as it's not use somewhere else for now

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/customer.html.twig
jolelievre: Done

admin-dev/themes/new-theme/js/pages/order/create/create-order-map.js
zuk3975: I guess its the case
> On the other hand, if it only stores the template inside and the id is not rendered multiple times, then it might be the best option.

So it seems all good to me

> Now the whole script/template may not be a good practice may not be the best, but it's been used a lot in all pages so it's not the purpose o this PR to refacto all this

IMO its the only available option unless we want to write html in javascript :smile:

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/customer.html.twig
NeOMakinG: Too early then, but hope that won't be forgotten

NeOMakinG (dismissed):
LGTM after some talks

admin-dev/themes/new-theme/js/pages/order/create/customer-renderer.js
```diff
@@ -223,6 +287,18 @@ export default class CustomerRenderer {
...
   }
 
+  /**
+   * Renders 'Loading' message in list
+   *
+   * @param $table
+   *
+   * @private
+   */
+  _renderLoading($table) {
```
PierreRambaud:
```suggestion
  renderLoading($table) {
```

admin-dev/themes/new-theme/js/pages/order/create/customer-renderer.js
```diff
@@ -95,6 +107,15 @@ export default class CustomerRenderer {
     this.$container.find(createOrderMap.customerSearchRow).removeClass('d-none');
   }
 
+  /**
+   * Empty the cart list and display a loading message.
+   */
+  showLoadingCarts() {
+    const $cartsTable = $(createOrderMap.customerCartsTable);
+    $cartsTable.find('tbody').empty();
+    this._renderLoading($cartsTable);
```
matks:
```suggestion
    this.renderLoading($cartsTable);
```

admin-dev/themes/new-theme/js/pages/order/create/customer-renderer.js
```diff
@@ -129,14 +150,33 @@ export default class CustomerRenderer {
...
+    });
+  }
+
+  /**
+   * Empty the order list and display a loading message.
+   */
+  showLoadingOrders() {
+    const $ordersTable = $(createOrderMap.customerOrdersTable);
+    $ordersTable.find('tbody').empty();
+    this._renderLoading($ordersTable);
```
matks:
```suggestion
    this.renderLoading($ordersTable);
```

matks (dismissed):
Added @PierreRambaud suggestion through github UI

jolelievre:
Thanks @matks Since there was a conflict and I had to rebase I squashed the commit
Needs to reapprove @NeOMakinG

matks (approved):

PierreRambaud (approved):
LGTM

NeOMakinG (approved):

Robin-Fischer-PS:
Hi @jolelievre , it's :heavy_check_mark: for QA !

Thanks :)

Progi1984:
Thanks @jolelievre

Issue #15671, opened 2019-09-24
Allow employee to choose their own avatar


Matt75:
**Is your feature request related to a problem? Please describe.**
Since employee cannot choose a theme variation, it becomes hard to know which employee is logged. The only one distinctive element is employee avatar.
![employee-top](https://user-images.githubusercontent.com/5262628/65513261-3e8ad200-dedb-11e9-8984-0fef8c5c4cb1.png)

Currently employee avatar still rely on third party service linked to PrestaShop Forum account. It's not possible to choose an avatar until create an account with same email on PrestaShop Forum.
So it's harder to quickly know who is currently logged as nobody goes to PrestaShop.com to create an account for upload their own avatar.

![employee-avatar](https://user-images.githubusercontent.com/5262628/65512252-24e88b00-ded9-11e9-8653-7f6654ddc64b.png)

**Describe the solution you'd like**
Remove usage of profile.prestashop.com and add an upload field to send a custom image and add a way for module developers to replace the way to retrieve employee avatar by their own implementation.

**Describe alternatives you've considered**
Keep usage of profile.prestashop.com as default implementation but allow module developers to replace it by their own implementation easier.
Currently it require some override to Employee ObjectModel to change behavior of function getImage()
https://github.com/PrestaShop/PrestaShop/blob/10825700128e838a87d91f524d6c529c56a816a4/classes/Employee.php#L609

**Additional context**
On a PrestaShop instance with multiple employee who share same computer, they must open dropdown of employee avatar to see who is currently logged. (Or use different browser...)
By default there are no way to difference who is logged as there are anymore theme variation by employee and process to use a custom avatar is weird.
![employee-dropdown](https://user-images.githubusercontent.com/5262628/65514958-a7277e00-dede-11e9-9409-f1c0399e7d8b.png)

khouloudbelguith:
Ping @PrestaShop/prestashop-product-team

ghost:
Wow @Matt75, that's the kind of feature request I like, clean, complete, lots of context, and some alternatives 👍 

We talked about it yesterday within the team, this avatar is also the only way to identify who answered to a customer in the customer service block in the new Orders page.

I totally agree with you, this is a great idea!


Thanks!

ttoine:
could be interesting to support gravatar, too, like many other cms and web applications.
Thanks @Matt75, +1

toutantic:
[...]

PR #18346, opened 2020-03-27
Fetch BO employee avatar from a hook

### LINKED ISSUE #18503: Remove call to profile.prestashop.com
#### Describe the bug

I can't access the community forum of Prestashop, I get this message error 

> Banned You do not have permission to view this site.

#### Steps to Reproduce

Access to multiple back-office websites at the same time, 
I should be log in one single website and log out from other websites then restart my router to get a new IP Adress so I can access the forum.

**Screenshots**
image

#### Additional information

* PrestaShop version: 1.7.6.4
* Version de PHP : 7.2.29

Regards,

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | In this PR, I modify how BO Profile avatar URL is fetched to retrieve the information from a module. This is done to follow @toutantic [request](https://github.com/PrestaShop/PrestaShop/issues/15671#issuecomment-603236906) and remove one more call from inside the BO to a PrestaShop service. I also created a [Module](https://github.com/matks/ps_boprofile) that performs the previous behavior. If no module is installed which uses the new hook "actionGetProfileImageUrl", default Preston image is used
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18503 and partially https://github.com/PrestaShop/PrestaShop/issues/15671
| How to test?  | See below

## What should be done, if we merge this PR ?

- I need to create another PR to register the new hook into database and hook.xml file
- I need to update the devdocs
- we should transfer https://github.com/matks/ps_boprofile module into PrestaShop organization
- we should decide what we want for default shop behavior: do we provide the module as a native module ? do we push it through API ? or do we let people who want it download it, and we let default Preston avatar
- we can also provide other modules to provide an avatar URL

## How to test

1. Checkout this branch
2. Check that now the default Preston avatar is used.
3. Install module https://github.com/matks/ps_boprofile (I provide a ZIP archive below)
4. Refresh
5. Check that now the standard PrestaShop avatar is used in BO: image is retrieved from profile.prestashop.com service and image used should be http://profile.prestashop.com/[your email].jpg

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18346)

## Test module ZIP archive

[ps_boprofile.zip](https://github.com/PrestaShop/PrestaShop/files/4394608/ps_boprofile.zip)

### DIFF
--- a/classes/Employee.php
+++ b/classes/Employee.php
@@ -584,11 +584,19 @@ class EmployeeCore extends ObjectModel
      */
     public function getImage()
     {
+        $defaultAvatar = Tools::getAdminImageUrl('prestashop-avatar.png');
+
         if (!Validate::isLoadedObject($this)) {
-            return Tools::getAdminImageUrl('prestashop-avatar.png');
+            return $defaultAvatar;
         }
 
-        return Tools::getShopProtocol() . 'profile.prestashop.com/' . urlencode($this->email) . '.jpg';
+        $profileImageUrl = Hook::exec('actionGetProfileImageUrl');
+
+        if (is_string($profileImageUrl) && '' !== $profileImageUrl) {
+            return $profileImageUrl;
+        }
+
+        return $defaultAvatar;
     }
 
     /**


matks:
Ping @toutantic @ttoine you should like this PR

ttoine:
Thanks for the ping.

This raises many questions:
- Is it an open source feature, or should this be part of the SSO with Addons and PSX in a commercial distribution available on prestashop.com?
- What is the real value of this service/feature for the open source project? Why not implement Gravatar instead? (and yes I know that Gravater is an Automattic service)
- Why not just remove this "yet another" link to a prestashop.com service/API and let people add manually a picture, simply in their shop BO, so we decrease collected and managed personal data and simplify GDPR for the PrestaShop company?

If this is confirmed that this feature doesn't belong to the open source project and should be removed (this is my feeling), this can go in PrestaShopCorp organisation.

matks:
> Thanks for the ping.
> 
> This raises many questions:
> 
> * Is it an open source feature, or should this be part of the SSO with Addons and PSX in a commercial distribution available on prestashop.com?

If we merge my PR, SSO / Addons PSX / whatever can make a module to provide the image from any external source

> * What is the real value of this service/feature for the open source project? Why not implement Gravatar instead? (and yes I know that Gravater is an Automattic service)

If we merge my PR, it becomes possible to create an distribute a Gravatar provider module that uses Gravatar

> * Why not just remove this "yet another" link to a prestashop.com service/API and let people add manually a picture, simply in their shop BO, so we decrease collected and managed personal data and simplify GDPR for the PrestaShop company?

It's longer to do (add new field to Employee form, store/resize image, allow people to edit/delete image ...). If we merge my PR, it becomes possible to create a "manage my avatar myself" module that does all of this.

PululuK:
Hi @matks 

What will happen if multiple modules are hooked here **`actionGetProfileImageUrl`**  🤔 ?
An employee could therefore have several profile photos ... Voilà Prestashop has become facebook 😂

In my opinion, this should be a simple input file in Employee form !

matks:
> What will happen if multiple modules are hooked here **`actionGetProfileImageUrl`** 🤔 ?

Well I guess indeed this hook is built with the "only 1 module installed" in mind 🤔.

The input file is a good idea and it could be put inside a module. The module would
- add the input into Employee form
- handle the image storage and management
- provide the image

Why restrict the avatar to single solution when we can allow Module developers to do whatever they want ?

PululuK:
> Well I guess indeed this hook is built with the "only 1 module installed" in mind thinking.

And 

> Why restrict the avatar to single solution when we can allow Module developers to do whatever they want ?

A contradiction 🤔 ?

ttoine:
So, per default, having a simple field to add a picture in the employee form is mandatory. I agree with @PululuK. That's expected by any user of any web application nowadays. And a basic feature in all open source CMS.

The current problem is that a user can not change their picture on their employee profile. With this PR, if I understand well (which I am not sure), this does not change, you just remove the current broken picture feature. And the module you propose keep the old behavior. So at the end, we would still have that in the BO (see screenshot below). Am I right? Or, did I miss something? And if I miss something, could you please clarify?

Selection_697
Selection_696

Your work is really interesting so it is possible to use a SSO (Google's, Prestashop's, ...), or Gravatar, or any other custom or public avatar implementations. And your module could be a good a good base for PSX's SSO.

But the software would still miss a way to add a picture manually if there is no module installed to change the behavior.

Progi1984 (approved):

PierreRambaud (approved):

ttoine:
So it is approved, and will be merged but without any default solution to add picture to the profile ?

eternoendless (changes requested):
I like this solution but it must be validated by a PM

ping @PrestaShop/prestashop-product-team

matks:
> So it is approved, and will be merged but without any default solution to add picture to the profile ?

PrestaShop 1.7 did not allow employees to upload their own picture for years. I think it can stay like this a little more as it does not seem to be a feature really wanted by people (there are other items in the backlog a lot more expected) ?

This PR allows people unhappy about the current behavior to disable it and allows module developers to provide some crafty alternatives. It's better than current implementation.

PululuK:
> This PR allows people unhappy about the current behavior to disable it and allows module developers to provide some crafty alternatives. It's better than current implementation.

So I ask my question again.
Suppose two developers hook their modules here: **`actionGetProfileImageUrl`**
How will you handle this?

ttoine:
@PululuK interesting point.
Maybe a choice should be proposed if many ways of adding a picture are available (this is what other CMS does)

classes/Employee.php
```diff
@@ -584,11 +584,19 @@ public function isSuperAdmin()
...
+            return $defaultAvatar;
+        }
+
+        $profileImageUrl = Hook::exec('actionGetProfileImageUrl');
+
+        if (is_string($profileImageUrl) && '' !== $profileImageUrl) {
+            return $profileImageUrl;
         }
 
-        return Tools::getShopProtocol() . 'profile.prestashop.com/' . urlencode($this->email) . '.jpg';
```
toutantic, line 591: Yyyeeeeeesssssssssssssss

matks:
> > This PR allows people unhappy about the current behavior to disable it and allows module developers to provide some crafty alternatives. It's better than current implementation.
> 
> So I ask my question again.
> Suppose two developers hook their modules here: **`actionGetProfileImageUrl`**
> How will you handle this?

Three possible solutions come in my mind:
1. We do not handle it. We tell merchants "you can only install 1 'employee profile picture provider module' so choose the one you want to use". It's simple, efficient and probably adresses 90% of "standard" merchant usage. This is current implementation so no extra work needed.
2. We allow modules to return `null` or an URL for each call, and modules must "agree" on which employees pictures they serve. When hook is called, it returns the 1st not null result. Every module is called on hook execution but we expect all of them but one to return `null`. A few extra work needed but not a lot.
3. We implement the [Chain of Responsibility](https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern) pattern with a dedicated system for modules to handle this usecase. Each module is first given the username of the employee and must return true or false. If false it means module cannot provide an URL for this employee and hook must call next module on list. If yes it means module "supports" this employee and is capable of providing an URL for this. Lot of extra work needed for this.

Question is: although options 2 and 3 are better than option 1, is the extra work needed to do options 2 or 3 worth it ?
If for example I spend 3 days working on the [Chain of Responsibility](https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern) pattern for option 3, are these 3 days "invested in a good way" 
 It could be 3 days I spend on other major bugs in PrestaShop. Or 3 days I spend reviewing / approving contributors Pull Requests 😄. I need to dedicate my time wisely 😉.

PululuK:
> > > This PR allows people unhappy about the current behavior to disable it and allows module developers to provide some crafty alternatives. It's better than current implementation.
> > 
> > 
> > So I ask my question again.
> > Suppose two developers hook their modules here: **`actionGetProfileImageUrl`**
> > How will you handle this?
> 
> Three possible solutions come in my mind:
> 
>     1. We do not handle it. We tell merchants "you can only install 1 'employee profile picture provider module' so choose the one you want to use". It's simple, efficient and probably adresses 90% of "standard" merchant usage. This is current implementation so no extra work needed.
> 
>     2. We allow modules to return `null` or an URL for each call, and modules must "agree" on which employees pictures they serve. When hook is called, it returns the 1st not null result. Every module is called on hook execution but we expect all of them but one to return `null`. A few extra work needed but not a lot.
> 
>     3. We implement the [Chain of Responsibility](https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern) pattern with a dedicated system for modules to handle this usecase. Each module is first given the username of the employee and must return true or false. If false it means module cannot provide an URL for this employee and hook must call next module on list. If yes it means module "supports" this employee and is capable of providing an URL for this. Lot of extra work needed for this.
> 
> 
> Question is: although options 2 and 3 are better than option 1, is the extra work needed to do options 2 or 3 worth it ?
> If for example I spend 3 days working on the [Chain of Responsibility](https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern) pattern for option 3, are these 3 days "invested in a good way"
> It could be 3 days I spend on other major bugs in PrestaShop. Or 3 days I spend reviewing / approving contributors Pull Requests smile. I need to dedicate my time wisely wink.


Anyway why the user should buy a module to manage a profile photo when it should be an expected feature in all CMS by default ? Maybe offer a native module for this? I will be interested in this case to work in the development of this module if necessary.

ttoine:
> Anyway why the user should buy a module to manage a profile photo when it should be an expected feature in all CMS by default ? Maybe offer a native module for this? I will be interested in this case to work in the development of this module if necessary.

+1

this is an expected feature and is a basic feature for years in all CMS.

ttoine:
Let's wait for a PM to answer.

LouiseBonnard:

matks:
As the scope of the PR changed, I created a new PR to apply specs
=> https://github.com/PrestaShop/PrestaShop/pull/18622

Progi1984:
@LouiseBonnard : @matks has made a PR to remove the call. We are waiting for us for a better built-in feature to handle avatars (new issue ? specs ?). Thanks

PR #18347, opened 2020-03-27
Hide order preview stock location column if empty

### LINKED ISSUE #18104: [Order page] The stock location column shouldn't be displayed if no location has been indicated
Currently on the new order page for 1.7.7, the stock location column is displayed even though no stock location has been indicated for the product. This occurs in :
- the order preview from the order listing
- the cancel product block
- the standard refund block
- the partial refund block
- the return product block

It should not be displayed when the field is empty. The column should be displayed only if one product of the order has a stock location value.

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | In order listing page, when previewing an order, the column 'stock location' must be hidden if no showed product has this information. That wasn't the case.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18104
| How to test?  | Create 3 different orders with more that 10 products:<br>- One with the first product having a stock location.<br>- One with the 11th (or more) product having a stock location.<br>- One with no product having a stock location.<br>For every preview of these orders, the column 'stock location' should be shown only when a product with a stock location is shown.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18347)


matks (approved):

Progi1984 (approved):

NeOMakinG (approved):

PierreRambaud (approved):

Robin-Fischer-PS:
It's OK, thanks for the nice How to test @atomiix :D

Progi1984:
Thanks @atomiix

PR #18348, opened 2020-03-27
Cart products update

### LINKED ISSUE #17958: The product name in the cart isn't changed according to the selected language
#### Describe the bug

The product name in the cart isn't changed according to the selected language

#### Expected behavior

The product name in the cart changes accordingly to the selected language

#### Steps to Reproduce

1. Go to international > localization > add a second language (ex: EN)
2. Go to catalog > products > edit a product (ex: mug the best is yet to come)
3. Enter a different product name in each languages
4. Go to orders > orders > add new order
5. Select a customer
6. Add the product "mug the best is yet to come" to the cart
7. Select english language
8. The product name is still in FR instead of EN

**Screenshots**
https://drive.google.com/file/d/1gIBfRjlptHgXfLf69wk60GOD8K1iih_g/view?usp=sharing

capture d'écran_2429

capture d'écran_2430

capture d'écran_2431

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | In order create page when the cart language was changed the cart preview was not updated so product names didn't change This PR fixes this behaviour It also adds a searching notice when products search request is running
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17958
| How to test?  | Follow the issue's scenario, but mainly you need two languages, correctly translated products (at least their names) and try switching the cart language The product names must update as well

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18348)


admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
```diff
@@ -454,4 +461,22 @@ export default class ProductRenderer {
...
   }
+
+  /**
+   * Shows searching product notice
+   *
+   * @private
+   */
+  _showSearching() {
+    $(createOrderMap.searchingProductsNotice).removeClass('d-none');
+  }
```
Progi1984:
```suggestion
  _setSearchingState(state) {
    $(createOrderMap.searchingProductsNotice).toggle('d-none', !state);
  }
```

matks: I would actually advocate for `showSearching()` and `hideSearching()` because functions with no arguments are easier to use and maintain than functions with N arguments (the higher N is, the more complex it is)

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
```diff
@@ -454,4 +461,22 @@ export default class ProductRenderer {
   _hideNotFound() {
     $(createOrderMap.noProductsFoundWarning).addClass('d-none');
   }
+
+  /**
+   * Shows searching product notice
+   *
+   * @private
+   */
+  _showSearching() {
```
matks:
```suggestion
  showSearching() {
```
I think @PierreRambaud will tell you not tu use `_` prefix and it will be removed when merged into `develop` because ESLint will refuse it

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
```diff
@@ -454,4 +461,22 @@ export default class ProductRenderer {
...
+    $(createOrderMap.searchingProductsNotice).removeClass('d-none');
+  }
+
+  /**
+   * Hides searching product notice
+   *
+   * @private
+   */
+  _hideSearching() {
```
matks:
```suggestion
  hideSearching() {
```

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
```diff
@@ -138,6 +144,7 @@ export default class ProductRenderer {
     this._cleanSearchResults();
     this._hideTaxWarning();
     this._hideResultBlock();
+    this._hideSearching();
```
matks:
```suggestion
    this.hideSearching();
```

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
```diff
@@ -113,13 +113,19 @@ export default class ProductRenderer {
...
+  }
+
   /**
    * Renders cart products search results block
    *
    * @param foundProducts
    */
   renderSearchResults(foundProducts) {
     this._cleanSearchResults();
+    this._hideSearching();
```
matks:
```suggestion
    this.hideSearching();
```

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
```diff
@@ -113,13 +113,19 @@ export default class ProductRenderer {
     }
   }
 
+  renderSearching() {
+    this.reset();
+    this._showSearching();
```
matks:
```suggestion
    this.showSearching();
```

admin-dev/themes/new-theme/js/pages/order/create/product-manager.js
```diff
@@ -182,7 +182,7 @@ export default class ProductManager {
    * @private
    */
   _search(searchPhrase) {
-    if (searchPhrase.length < 3) {
+    if (searchPhrase.length < 2) {
```
PierreRambaud, line 185: Should be define in a const, or with a comment to explain why we set this value (linked to a spec file maybe?)

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
PierreRambaud: It's just depending on you :sweat_smile: 
Ping @NeOMakinG  wdyt?
I really prefer having one method doing
```js
this.display(something);
```

instead of doing

```js
if (something) {
  this.show();
} else {
  this.hide();
}

File: admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js (hunk 2 above)
Comment (jolelievre) [reply admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js]: I know I anticipated this kind off feedback 😁 It just feels weird to have a mix in 177.. some start by _ some not..
But I totally understand how it will ease the merge in develop so who cares 😅

File: admin-dev/themes/new-theme/js/pages/order/create/product-manager.js (hunk 7 above)
Comment (jolelievre) [reply admin-dev/themes/new-theme/js/pages/order/create/product-manager.js:185]: I guess you're right about the spec
@MatShir what do you think? This defines how many characters we need to type before searching for products
In the customer search this was reduced to 1 character previously (nice when you need to look or "jo" 😅), did you write all this down somewhere in the spec?

I hesitated between 1 or 2, but I thought 1 would match too many products. What limit do you think we should use here?

File: admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js (hunk 1 above)
Comment (jolelievre) [reply admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js]: I copied the existent and didn't really think about it, but I actually kind of like the toggle feature better as well But indeed I guess it's a matter of preference, let's leave the final decision to our expert @NeOMakinG ^^

However I wouldn't use `display` because using `display(false)` to hide something feels counter intuitive, Franck's proposal `_setSearchingState` seems more logic, or something like `toggleSearchingNotice` maybe

If we decide to use a toggle though, should it be done the same way or other block toggling (in this class), I also mean the one in this PR https://github.com/PrestaShop/PrestaShop/pull/18334 which are very similar

File: src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig [hunk 8]
```diff
@@ -49,6 +49,14 @@
       </div>
     </div>
 
+    <div class="row js-searching-products d-none">
+      <div class="col">
+        <div class="alert alert-info" role="alert">
+          <p class="alert-text">{{ 'Searching for products'|trans({}, 'Admin.Orderscustomers.Notification') }}</p>
```
LouiseBonnard, line 55: It seems identical but, just to make sure, it'd need to be harmonized with the same alert in #18334. :-)

admin-dev/themes/new-theme/js/pages/order/create/product-manager.js
PierreRambaud: Maybe we should look for a debounce instead of min characters :thinking:

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
jolelievre: Yeah I did the same thing

admin-dev/themes/new-theme/js/pages/order/create/product-manager.js
jolelievre: There's already a debounce of 300ms managed in `CreateOrderPage::_initProductSearch`
And in this _search function the current request (if present) is aborted

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/cart.html.twig
LouiseBonnard: You rock! 👌

admin-dev/themes/new-theme/js/pages/order/create/product-renderer.js
NeOMakinG: I agree with Pierre on this, I would prefer a toggle function as it decrease complexity to understand it, but that's a matter of preference, except if that's reusable in the future

Progi1984 (approved):

admin-dev/themes/new-theme/js/pages/order/create/product-manager.js
matks: ping @MatShir

matks (approved):

khouloudbelguith:
Hi @jolelievre,

It is OK :heavy_check_mark: 
https://drive.google.com/file/d/10m4xmJthPhfITKCm4jkeWNR40d42WGgJ/view?usp=sharing

Thanks!

Progi1984:
Thanks @jolelievre

Issue #9817, opened 2018-08-22
[BOOM-4250] Some characters (Cyrillic, Persian...) in invoice pdf show as undefined characters


prestonBot:
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-4250).

- _**Reporter:**_ salma.moakhar
- _**Created at:**_ Mon, 6 Nov 2017 17:08:28 +0100

<p>Title is self-explanatory. Japanese characters in the invoice PDF are showing as undefined characters . In the attachment these correspond to customization text field values. </p>

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-4250#comment-157398).

- _**Author:**_ vincentbz
- _**Created at:**_ Fri, 24 Nov 2017 15:41:47 +0100

<p>Cela doit s'afficher correctement quelle que soit la langue.</p>

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-4250#comment-172775).

- _**Author:**_ marion_francois
- _**Created at:**_ Fri, 27 Jul 2018 12:13:11 +0200

<p>Hello <a href="http://forge.prestashop.com/secure/ViewProfile.jspa?name=khouloud.belguith" class="user-hover" rel="khouloud.belguith">Khouloud BELGUITH</a> </p>

<p>Can you tell me if there is still this issue on 1.7.4.1 ?</p>

<p>Thanks <img class="emoticon" src="http://forge.prestashop.com/images/icons/emoticons/smile.png" height="16" width="16" align="absmiddle" alt="" border="0"/></p>

prestonBot:
> This comment has been migrated from the Forge. Read the [original comment here](http://forge.prestashop.com/browse/BOOM-4250#comment-172778).

- _**Author:**_ khouloud.belguith
- _**Created at:**_ Fri, 27 Jul 2018 12:50:42 +0200

<p>Hi <a href="http://forge.prestashop.com/secure/ViewProfile.jspa?name=marion_francois" class="user-hover" rel="marion_francois">Marion F</a>,</p>

<p>I manage to reproduce the issue with 1.7.4.1.<br/>
[...]

PR #18352, opened 2020-03-28
Fix language used in pdf attachement sent by mail

Matt75:
| Questions     | Answers
| Branch?       | develop
| Description?  | Invoice or DeliverySlip sent by mail are not using Language used for Order
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10258 (Partial : fixed only for PDF sent by mail)
| How to test?  |  Change OrderState to Payment accepted or change OrderState to Processing in progress, attachment PDF sent by mail use Language used in BO, not Language used in Order.

### Test scenario
In FO use French language
Make an Order in french

In BO use English language
Change OrderState to Payment accepted
Mail sent to Customer have attachment in English (BO language) instead of Customer language used to make the Order.

With this change, mail attachement should be in language used by customer on FO

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18352)

### DIFF
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -646,17 +646,19 @@ abstract class PaymentModuleCore extends Module
                             $data = array_merge($data, $extra_vars);
                         }
 
+                        $contextLanguage = Context::getContext()->language;
                         $orderLanguage = new Language((int) $order->id_lang);
 
                         // Join PDF invoice
                         if ((int) Configuration::get('PS_INVOICE') && $order_status->invoice && $order->invoice_number) {
                             $order_invoice_list = $order->getInvoicesCollection();
+                            Context::getContext()->language = $orderLanguage;
                             Hook::exec('actionPDFInvoiceRender', ['order_invoice_list' => $order_invoice_list]);
                             $pdf = new PDF($order_invoice_list, PDF::TEMPLATE_INVOICE, $this->context->smarty);
-                            $pdf->setLanguage($orderLanguage);
                             $file_attachement['content'] = $pdf->render(false);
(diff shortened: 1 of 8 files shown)


classes/Context.php
```diff
@@ -467,4 +467,17 @@ public function getComputingPrecision()
...
     }
+
+    /**
+     * Change context language and locale used in translator
+     *
+     * @since 1.7.8
+     *
+     * @param Language $language
+     */
+    public function setLanguageAndTranslatorLocale(Language $language)
```
Progi1984: I'm not sure about this method : that breaks the principle of single responsability. Could you divide this method in two methods ?

Matt75: Fixed

Progi1984 (approved):

matthieu-rolland (approved):

matks (approved):

PierreRambaud (approved):

khouloudbelguith:
Hi @Matt75,

I installed a shop with the "English" language and Japan country.
In the Fo, the customer language is (Japanese)
- The invoice and delivery slip sent by email are in (Japanese). => OK
image
image

- In the BO => Orders page => pdf files downloaded NOK as reported here: https://github.com/PrestaShop/PrestaShop/issues/9817


Thanks!

Progi1984:
Thanks @Matt75

superskyman100:
This code works only with your configuration. We tried with PS 1.7.6.4 it's not working, the files are different.

Matt75:
@superskyman100 This change is only for 1.7.8 ;)

PR #18393, opened 2020-03-31
Fix product attributes not updated in some themes

### LINKED ISSUE #18509: Product page not updated after changing attributes - BC break
#### Describe the bug

Some parts of the product page must be updated in ajax after changing an attribute, since 1.7.6.3, a too specific selector was introduced in core.js file that break some themes.
This is not a bug per say, as the classic theme still works, but minor updates are not supposed to break third party themes.

#### Expected behavior

Attributes should updates depending on existing combinations.
ex, two combinations:

1. size: S, color black
2. size M, color white

selecting "S" should only display white color and selecting "M" should only display "black" color. This is not the case anymore with some themes who where working well before. 

#### Steps to Reproduce

Can be reproduced with "01 Oslo" theme for exemple.

**Screenshots**

#### Additional information

* PrestaShop version: 1.7.6.3
* PHP version: N/A

glo-www:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | a selector too specific was introduced in core.js in 1.7.6.3, breaking some themes that do not have the "row" class. ".row" is not necessary here. This should be ported to 1.7.7.x too.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18509
| How to test?  |  change attributes on product page, product should be updated, but not associated products/similar products (be carefull about this)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18393)


NeOMakinG (approved):
Hi @glo-www, thanks for your contribution.

It should be fine, hopefully you tested that this is not updating some products which shouldn't be updated (for example associated products, similar products)

Progi1984:
@glo-www  Thank you for your contribution. :hugs: 

Could you explain in an issue your initial bug for this PR (or may be link to an issue) ?

Thanks :+1:

glo-www:
> 
> 
> @glo-www Thank you for your contribution. 🤗
> 
> Could you explain in an issue your initial bug for this PR (or may be link to an issue) ?
> 
> Thanks 👍

Thanks @Progi1984 , I created an issue:
https://github.com/PrestaShop/PrestaShop/issues/18509

Progi1984 (approved):

matks (approved):

PierreRambaud (approved):

NeOMakinG (approved):

SD1982:
18Hello @glo-www I've tried to reproduce the bug you described here : #18509 and even with "01 Bold" theme i can't reproduce the behaviour . 
The test scenario : 
1 - Create a product with combination ( t-shirt => size; S - color; Black / size; M - color; White )
2 - In FO on the previously created product page try to switch between the product combination and see if informations and images are updated .

And as you can see in this screen record the image and information are correctly updated when i switch between combination : 
https://drive.google.com/file/d/1im_B_oLQU8SJ8Y-n5Wrb1bp5X8VnyWg_/view?usp=sharing

glo-www:
Hello @SD1982, 
my mistake, sorry, "01 Bold" is unafected as it has the ".row" class. 
You can reproduce the bug with "01 Oslo" though.
Thank you

NeOMakinG:
@SD1982 test if associated product doesnt moove too, that's important, this .row class has been added by fixing a bug where associated products and similar products were updated too, you should really care on this ! :)

Better taking a product with a label like "New product" or a promotion

glo-www:
@NeOMakinG, arf indeed flags of associated products (ie "same category") are being updated if we remove .row class from the selector.
Sorry :(

NeOMakinG:
@glo-www damn, and when I still fixed this, I couldn't find a better selector, and of course adding a selector wont fix it on every themes since they'll need to update to get the fix and their update wont work anymore :/

NeOMakinG:
I think we can't do more on this PR, or adding a new selector somewhere that themes not using bootstrap can use :/

NeOMakinG (changes requested):
Blocking the merge, this will break because it will update tags on associated products and similar

Need to find a clean workaround to enable it again

rmello2190:
I have same error when install  Prestashop version 1.7.6.4

When I create new product with attributes product page not update changes,

Or when create an copy of product, don't show changes on product page or homepage

Progi1984:
@NeOMakinG What is the solution that you suggest ?

NeOMakinG:
I don't remember precisely what we said on this, but I think the best solution would be to add a class in order to select this precisely and let themes adapt to this (this would be the same role as the .row class, so it looks like the best moove is to just let the .row class and themes should adapt to this by adding this class...)

glo-www:
Hi @NeOMakinG, I think a specific class (ex: ".main-product") should be introduced instead of .row for this specific selector. Indeed the bootstrap class ".row" as two side effects:

1. it has css styles applied to it and can break a theme layout
2. the .row class can be needed as an ancestor for associated products and then the problem remains that the selector is not specific enough.

edit: this solution is even less backward compatible than keeping the .row :/

NeOMakinG:
@glo-www what we decided regarding this issue is that we obviously won't revert that change, but instead, we'll add a new selector not creating side effects. It means that :

- Themes using this .row like the classic theme will continue to work smoothly
- Themes not having the .row class will need to add the new class anyway

Keep in mind that this selector has been used in order to fix a bug on product lists and product page, this was not intended !

NeOMakinG:
Closing this PR in favor of the new class added :

https://github.com/PrestaShop/PrestaShop/pull/19800

Please, refer to my comment on #18509 for more infos !

Thanks for you'r contribution @glo-www , I hope you agree with our choice, even if it's a BC, this was done to resolve some big bugs :/

glo-www:
Thanks @NeOMakinG totally agree

PR #16404, opened 2019-11-13
Improve cache for staticPrice for products with combinations

Progi1984:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Initial PR #8542 from @jbustamanterendon<br>The "id_product_attribute" was not being cached when checking the quantity of each product in the cart, making multiple calls to the "cart_product" table and lowering the store's performance.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9803 & http://forge.prestashop.com/browse/BOOM-4368
| How to test?  | Many products with many combinations in the cart, for example : 77 products in the shopping cart where there were only about 20 products, the other 57 were different combinations of the products.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/16404)

PR #8542, opened 2017-11-24
Fix multiple SQL queries to cart_product table

jbustamanterendon:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | The "id_product_attribute" was not being cached when checking the quantity of each product in the cart, making multiple calls to the "cart_product" table and lowering the store's performance.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | http://forge.prestashop.com/browse/BOOM-4368 & Fixes #9803
| How to test?  | Many products with many combinations in the cart, for example : 77 products in the shopping cart

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/8542)

PR #18427, opened 2020-04-02
Fix issue when there is a quantity discount on combination products

### LINKED ISSUE #18049: Quantity discount based on parameter does not work
#### Describe the bug

When setting the "Quantity discount based on" parameter to "Products", it does not work

#### Expected behavior

When "Quantity discount based on" is set to "Products", if there is a cart rules or a specific price based on quantity, the discount should be applied as soon as we have in cart the required number of products, **even if there are different combinations in cart**.

When "Quantity discount based on" is set to "Combinations", if there is a cart rules or a specific price based on quantity, the discount should be applied only if there is the required number of products **in one combination**, and the discount will be applied only to this combination.

#### Steps to Reproduce

1. Go to Shop Parameters > Products Settings > Products (general)
2. Set "Quantity discounts based on" to "Products
3. Go to Catalog > Discounts > Catalog Price Rules
4. Create a cart rules with a reduction starting from quantity 3 (for example)
5. Go to FO, to a product with multiple combination available
6. Add 3 different combination to cart
7. Go to cart : There is no discount applied, it should be (since we have 3 combination from the same product)

**Screenshots**

https://drive.google.com/file/d/15q6jwbElhMi-PWSfen80WolJbRo5ZAH0/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: N/A

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Quantity discount on same products having different attributes is broken: for example two same sweat with different colors would not be counted as 2 products  
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18049
| How to test?  | See steps in the issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18427)

So, this PR is actually a revert of this pr: #16404, which was a remake of this pr: #8542 

It added an element in the cache key of the **getPriceStatic** method: the id_attribute of the product, supposedly to reduce the number of queries. There was 2 problems with this modification:

**1/** Adding an element to a cache key usually increases the number of queries, to make it diminishing the number of query the PR modified the query by adding the product's **id_product_attribute**. This is a patch, not a fix, the query was modified not for a functional purpose but in order to match the condition just above the query.

**2/** It introduces a bug, as rGaillard and jocel1 warned us:  https://github.com/PrestaShop/PrestaShop/pull/8542#issuecomment-349587244
Here we want the product quantity, including same products having different attributes, otherwise the **priceCalculation** method won't apply the price rules based on quantity.

### DIFF
--- a/tests/Integration/Behaviour/Features/Scenario/Cart/Adding/Product/add_combination.feature
+++ b/tests/Integration/Behaviour/Features/Scenario/Cart/Adding/Product/add_combination.feature
@@ -39,8 +39,11 @@ Feature: Add product combination in cart
     Given there is a product in the catalog named "product7" with a price of 24.324 and 1000 items in stock
     Given product "product7" has a combination named "combi1" with 100 items in stock
     Given product "product7" has a combination named "combi2" with 100 items in stock
-    When I add 2 items of combination "combi1" of product "product7"
+    When I add 1 items of combination "combi1" of product "product7"
     When I add 1 items of combination "combi2" of product "product7"
-    Then I should have 2 items of combination "combi1" of product "product7" in my cart
+    Then I should have 1 items of combination "combi1" of product "product7" in my cart
     Then I should have 1 items of combination "combi2" of product "product7" in my cart
+    Then The price of each product "product7" after reduction should be 24.324
+    When I add 1 items of combination "combi2" of product "product7"
+    Then I should have 2 items of combination "combi2" of product "product7" in my cart
     Then The price of each product "product7" after reduction should be 19.324
--- a/tests/Integration/Behaviour/Features/Scenario/Cart/Adding/Product/add_combination.feature
+++ b/tests/Integration/Behaviour/Features/Scenario/Cart/Adding/Product/add_combination.feature
@@ -3,6 +3,7 @@ Feature: Add product combination in cart
   As a customer
   I must be able to correctly add product combinations in my cart
 
+  @add-combinations-to-cart
   Scenario: Add combination in cart
     Given I have an empty default cart
     Given there is a product in the catalog named "product7" with a price of 24.324 and 1000 items in stock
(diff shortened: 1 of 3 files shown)


tests/Integration/Behaviour/Features/Scenario/Cart/Adding/Product/add_combination.feature
```diff
@@ -28,3 +31,16 @@ Feature: Add product combination in cart
     When I add 600 items of combination "combi1" of product "product7"
     Then I should have 600 items of combination "combi1" of product "product7" in my cart
     Then the remaining available stock for combination "combi1" of product "product7" should be -100
+
+  @add-combinations-to-cart
```
Progi1984, line 35: @matthieu-rolland I would add a behat test with only 2 items (in this case, the specific price doesn't apply)

matks:
@matthieu-rolland you change something into one of the core classes of PrestaShop. Can you explain the exploration process you followed to find that this `$cache_id` and this `$sql` change fixes the issue ?

Without it, it's hard to link the fixed issue and the changed code 😅

tests/Integration/Behaviour/Features/Scenario/Cart/Adding/Product/add_combination.feature
matthieu-rolland: thanx, done 

 I did it in the same scenario so that it follows the real use case:  first I test the price with only 2 products (it must not change), then I add one product, then I check again the price of each product, the quantity price rule must be applied.

Progi1984 (approved):

matks (approved):

SD1982:
Hello @matthieu-rolland It seems that the rules is correctly applied to the product itself and even if there is different combination but the total amount of the cart seems to not take count of it. 

**Without the PR :**
withoutPR

**With PR :**
withPR

matthieu-rolland:
@SD1982 seeing your two captures, I think we are fine !

Before the PR, the result was wrong on the left column, but correct in the total on the right column.

After the PR: the result is correct in the left column, and still correct in the right column. If you add the values on the left, you'll see that the total is good in your second capture  (but maybe I missed something ? 🤔 )

Let me know ;)

SD1982:
Hi @matthieu-rolland  Indeed as you said and as we discuss the total with PR seems to be good I will do a new test on the PR before validate it .

SD1982:
LGTM Thanks @matthieu-rolland

Progi1984:
Thanks @matthieu-rolland

PR #18190, opened 2020-03-17
[WIP] Added a build for PHP 7.4

mickaelandrieu:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | PHP 7.4 is already "mainstream", at least this version should be tested
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Address part of  #16477.
| How to test?  | Nothing to do: I'm pretty sure the build will fail, but it's the first step to provide support of this PHP version

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18190)

<details>
<summary>Got this result on my computer: </summary>

(base) ➜  **PrestaShop git:(feature/php-74) composer install --no-suggest --ansi --prefer-dist**
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them.
Package operations: 195 installs, 0 updates, 0 removals
  - Installing composer/installers (v1.7.0): Downloading (100%)         
  - Installing symfony/polyfill-ctype (v1.13.1): Downloading (100%)         
  - Installing twig/twig (v1.42.4): Downloading (100%)         
  - Installing paragonie/random_compat (v2.0.18): Downloading (100%)         
  - Installing symfony/polyfill-php70 (v1.13.1): Downloading (100%)         
  - Installing symfony/polyfill-util (v1.13.1): Downloading (100%)         
  - Installing symfony/polyfill-php56 (v1.13.1): Downloading (100%)         
  - Installing symfony/polyfill-mbstring (v1.13.1): Downloading (100%)         
  - Installing symfony/symfony (v3.4.37): Downloading (100%)         
  - Installing symfony/polyfill-intl-icu (v1.13.1): Downloading (100%)         
  - Installing symfony/polyfill-apcu (v1.13.1): Downloading (100%)         
  - Installing psr/simple-cache (1.0.1): Downloading (100%)         
  - Installing psr/log (1.1.2): Loading from cache
  - Installing psr/link (1.0.0): Downloading (100%)         
  - Installing psr/container (1.0.0): Loading from cache
  - Installing psr/cache (1.0.1): Loading from cache
  - Installing fig/link-util (1.1.0): Downloading (100%)         
  - Installing doctrine/lexer (1.0.2): Loading from cache
[...]

PR #18434, opened 2020-04-02
Allow Object models to define their own methods of validation

mickaelandrieu:
| Questions     | Answers
| Branch?       | develop
| Description?  | When you need specific validation rules you have to override Validate class, this contribution fixes this issue
| Type?         | new feature
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | This fixes an issue: we can't override the Validate class twice, so...
| How to test?  | By writing tests, obviously

```
<?php

class A extends ObjectModel
{
    public $id;
    public $a;
    public $b;
 
    public static $definition = [
        'table' => 'a',
        'primary' => 'id',
        'fields' => [
            'a' => ['type' => self::TYPE_INT, 'validate' => 'A::isFoo'],
            'b' => ['type' => self::TYPE_INT, 'validate' => 'isInt'],
        ]
    ];

    public static function isFoo($value)
    {
        return false;
    }
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18434)

### DIFF
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -1150,12 +1150,12 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
 
         // Check field validator
         if (!in_array('validate', $skip) && !empty($data['validate'])) {
-            $callable = $data['validate'];
-            if (!is_callable($callable)) {
-                $callable = 'Validate::' . $data['validate'];
+            $functionName = $data['validate'];
+            if (!is_callable($functionName)) {
+                $functionName = 'Validate::' . $data['validate'];
             }
 
-            if (!is_callable($callable)) {
+            if (!is_callable($functionName)) {
                 throw new PrestaShopException(
                     $this->trans(
                         'Validation function not found: %s.',
@@ -1166,20 +1166,20 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
             }
 
             if (!empty($value)) {
-                $is_valid = true;
+                $isValid = true;
                 if (Tools::strtolower($data['validate']) === 'iscleanhtml') {
-                    if (!call_user_func(['Validate', $data['validate']], $value, $ps_allow_html_iframe)) {
-                        $is_valid = false;
+                    if (!call_user_func($functionName, $value, $ps_allow_html_iframe)) {
+                        $isValid = false;
+                    }
+                } else {
+                    if (!call_user_func($functionName, $value)) {
+                        $isValid = false;
                     }
                 }
 
-                if (!call_user_func($callable, $value)) {
-                    $is_valid = false;
-                }
-
-                if (!$is_valid) {
+                if (!$isValid) {
                     if ($human_errors) {
-                        return $this->trans('The %s field is invalid.', [$this->displayFieldName($field, get_class($this))], 'Admin.Notifications.Error');
+                        return $this->trans('The %s field is invalid.', [self::displayFieldName($field, get_class($this))], 'Admin.Notifications.Error');
                     }
 
                     return $this->trans('Property %s is not valid', [get_class($this) . '->' . $field], 'Admin.Notifications.Error');
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -1150,27 +1150,39 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
 
         // Check field validator
         if (!in_array('validate', $skip) && !empty($data['validate'])) {
-            if (!method_exists('Validate', $data['validate'])) {
-                throw new PrestaShopException($this->trans('Validation function not found: %s.', [$data['validate']], 'Admin.Notifications.Error'));
+            $callable = $data['validate'];
+            if (!is_callable($callable)) {
+                $callable = 'Validate::' . $data['validate'];
+            }
+
+            if (!is_callable($callable)) {
+                throw new PrestaShopException(
+                    $this->trans(
+                        'Validation function not found: %s.',
+                        [$data['validate']],
+                        'Admin.Notifications.Error'
+                    )
+                );
             }
 
             if (!empty($value)) {
-                $res = true;
-                if (Tools::strtolower($data['validate']) == 'iscleanhtml') {
+                $is_valid = true;
+                if (Tools::strtolower($data['validate']) === 'iscleanhtml') {
                     if (!call_user_func(['Validate', $data['validate']], $value, $ps_allow_html_iframe)) {
-                        $res = false;
-                    }
-                } else {
-                    if (!call_user_func(['Validate', $data['validate']], $value)) {
-                        $res = false;
+                        $is_valid = false;
                     }
                 }
-                if (!$res) {
+
+                if (!call_user_func($callable, $value)) {
+                    $is_valid = false;
+                }
+
+                if (!$is_valid) {
                     if ($human_errors) {
                         return $this->trans('The %s field is invalid.', [$this->displayFieldName($field, get_class($this))], 'Admin.Notifications.Error');
-                    } else {
-                        return $this->trans('Property %s is not valid', [get_class($this) . '->' . $field], 'Admin.Notifications.Error');
                     }
+
+                    return $this->trans('Property %s is not valid', [get_class($this) . '->' . $field], 'Admin.Notifications.Error');
                 }
             }
         }
--- a/tests-legacy/Unit/Classes/ObjectModelTest.php
+++ b/tests-legacy/Unit/Classes/ObjectModelTest.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace LegacyTests\Unit\Classes;
+
+use LegacyTests\TestCase\UnitTestCase;
+use Alias;
+
+class ObjectModelTest extends UnitTestCase
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function setUp()
+    {
+        Alias::$definition['fields']['fullname'] = [
+        'type' => Alias::TYPE_STRING,
+        'validate' => 'isString'
+        ];
+
+        Alias::$definition['fields']['gender'] = [
+            'type' => Alias::TYPE_STRING,
+            'validate' => 'LegacyTests\Unit\Classes\CustomValidator::isValidGender',
+        ];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function teardown()
+    {
+        unset(
+            Alias::$definition['fields']['fullname'],
+            Alias::$definition['fields']['gender']
+        );
+    }
+
+    public function testValidateField()
+    {
+        $alias = new Alias();
+
+        self::assertTrue($alias->validateField('fullname', 'Mickaël Andrieu'));
+        self::assertTrue($alias->validateField('gender', 'MALE'));
+    }
+}
+
+class CustomValidator {
+    public static function isValidGender($value)
+    {
+        return in_array($value, [
+            'MALE',
+            'FEMALE',
+            'NOT_BINARY',
+        ]);
+    }
+}
--- a/tests-legacy/Unit/Classes/ObjectModelTest.php
+++ b/tests-legacy/Unit/Classes/ObjectModelTest.php
@@ -27,32 +27,21 @@
 namespace LegacyTests\Unit\Classes;
 
 use LegacyTests\TestCase\UnitTestCase;
-use Alias;
+use ObjectModel;
 
 class ObjectModelTest extends UnitTestCase
 {
+    /**
+     * @var ObjectModel
+     */
+    private $dummyObjectModel;
+
     /**
      * {@inheritdoc}
      */
     protected function setUp()
     {
-        # Old retro compatible way
-        Alias::$definition['fields']['fullname'] = [
-        'type' => Alias::TYPE_STRING,
-        'validate' => 'isString'
-        ];
-
-        # What should be documented post 1.7.8
-        Alias::$definition['fields']['whatever'] = [
-            'type' => Alias::TYPE_INT,
-            'validate' => 'Validate::isInt'
-        ];
-
-        # What is also possible post 1.7.8
-        Alias::$definition['fields']['gender'] = [
-            'type' => Alias::TYPE_STRING,
-            'validate' => 'LegacyTests\Unit\Classes\CustomValidator::isValidGender',
-        ];
+        $this->dummyObjectModel = $this->getDummyObject();
     }
 
     /**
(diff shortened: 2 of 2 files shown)


classes/ObjectModel.php
```diff
@@ -1149,27 +1149,39 @@ public function validateField($field, $value, $id_lang = null, $skip = [], $huma
 
         // Check field validator
         if (!in_array('validate', $skip) && !empty($data['validate'])) {
-            if (!method_exists('Validate', $data['validate'])) {
-                throw new PrestaShopException($this->trans('Validation function not found: %s.', [$data['validate']], 'Admin.Notifications.Error'));
+            $callable = $data['validate'];
+            if (!is_callable($callable)) {
```
PululuK, line 1153: I like the ideia @mickaelandrieu  !
Can you remove `!` please ?

PululuK:
Big thanks @mickaelandrieu  !!
I've been waiting for this for a long time

classes/ObjectModel.php
```diff
@@ -1149,27 +1149,39 @@ public function validateField($field, $value, $id_lang = null, $skip = [], $huma
...
+                $callable = 'Validate::'. $data['validate'];
+            }
+
+            if (!is_callable($callable)) {
```
matks: You replaced `method_exists($a, $b)` by `is_callable($a::$b)`
Although it should be equivalent I'm afraid of php crazy behaviors with that kind of dynamic function calls 😅 

Maybe we could restore the `method_exists()` to be 100% sure there is no risk of doing that ?

classes/ObjectModel.php
```diff
@@ -1149,27 +1149,39 @@ public function validateField($field, $value, $id_lang = null, $skip = [], $huma
...
+                        array($data['validate']),
+                        'Admin.Notifications.Error'
+                    )
+                );
             }
 
             if (!empty($value)) {
-                $res = true;
-                if (Tools::strtolower($data['validate']) == 'iscleanhtml') {
+                $is_valid = true;
```
Matt75:
```suggestion
                $isValid = true;
```

classes/ObjectModel.php
```diff
@@ -1149,27 +1149,39 @@ public function validateField($field, $value, $id_lang = null, $skip = [], $huma
...
+                if (Tools::strtolower($data['validate']) === 'iscleanhtml') {
                     if (!call_user_func(['Validate', $data['validate']], $value, $ps_allow_html_iframe)) {
-                        $res = false;
-                    }
-                } else {
-                    if (!call_user_func(['Validate', $data['validate']], $value)) {
-                        $res = false;
+                        $is_valid = false;
```
Matt75:
```suggestion
                        $isValid = false;
```

classes/ObjectModel.php
```diff
@@ -1149,27 +1149,39 @@ public function validateField($field, $value, $id_lang = null, $skip = [], $huma
...
                     }
                 }
-                if (!$res) {
+
+                if (!call_user_func($callable, $value)) {
+                    $is_valid = false;
```
Matt75:
```suggestion
                    $isValid = false;
```

classes/ObjectModel.php
```diff
@@ -1149,27 +1149,39 @@ public function validateField($field, $value, $id_lang = null, $skip = [], $huma
...
+                }
+
+                if (!$is_valid) {
```
Matt75:
```suggestion
                if (!$isValid) {
```

classes/ObjectModel.php
```diff
@@ -1149,27 +1149,39 @@ public function validateField($field, $value, $id_lang = null, $skip = [], $huma
...
+            $callable = $data['validate'];
+            if (!is_callable($callable)) {
+                $callable = 'Validate::'. $data['validate'];
+            }
+
+            if (!is_callable($callable)) {
+                throw new PrestaShopException(
+                    $this->trans(
+                        'Validation function not found: %s.',
+                        array($data['validate']),
```
Matt75:
```suggestion
                       [$data['validate']],
```

classes/ObjectModel.php
```diff
@@ -1149,27 +1149,39 @@ public function validateField($field, $value, $id_lang = null, $skip = [], $huma
 
         // Check field validator
         if (!in_array('validate', $skip) && !empty($data['validate'])) {
-            if (!method_exists('Validate', $data['validate'])) {
-                throw new PrestaShopException($this->trans('Validation function not found: %s.', [$data['validate']], 'Admin.Notifications.Error'));
+            $callable = $data['validate'];
+            if (!is_callable($callable)) {
+                $callable = 'Validate::'. $data['validate'];
```
Matt75:
```suggestion
                $callable = 'Validate::' . $data['validate'];
```

Matt75 (commented):
Make php-cs-fixer happy

classes/ObjectModel.php
```diff
@@ -1149,27 +1149,39 @@ public function validateField($field, $value, $id_lang = null, $skip = [], $huma
...
-                    }
-                } else {
-                    if (!call_user_func(['Validate', $data['validate']], $value)) {
-                        $res = false;
+                        $is_valid = false;
                     }
                 }
-                if (!$res) {
+
+                if (!call_user_func($callable, $value)) {
```
matks: Same here: I'm afraid php is not really clear and strict on what can be called and what cannot be called.
You replaced `call_user_func(['Validate', $data['validate']], $value)` by `call_user_func(Validate::'. $data['validate'], $value)`

It looks equivalent but honestly I'm not a php expert so I cannot say it covers 100% possible php usages and these 2 ways of writing are equivalent. I would feel safer if we keep previous notation to ensure we dont modify the behavior.

matks (commented):
Interesting ! Please see some suggestions but the idea is great

Would you please consider adding some unit tests 😃?

matks:
@mickaelandrieu PHP-CS-Fixer is not happy 😄could you please fix the codestyle of your PR ?

classes/ObjectModel.php
mickaelandrieu: Imho, checking that the method exists proves nothing: what we want is to be able to execute a function.

~~Adding unit tests will ensure the behavior is still right: don't fear PHP, fear your own codebase my friend~~

Edit: what I can do is add the 2 checks, we're both happy, WDYT ?

classes/ObjectModel.php
mickaelandrieu: I disagree with it as this forces me to update all the file: this is not the intent of my contribution, let's focus on the feature and let a bot clean variables, please.

classes/ObjectModel.php
mickaelandrieu: => https://www.php.net/manual/fr/language.types.callable.php

In any case, this will works as the previous check (remember : `is_callable`) is true.

The fear is never a good reason for not improving something: if I redo that change, I'll need at least 3 additional if/else statments in this function: this is not something I want to write  (or test).

I know it because ... it was the first version of my feature :)

mickaelandrieu:
> @mickaelandrieu PHP-CS-Fixer is not happy smilecould you please fix the codestyle of your PR ?

Yes, but how ? :unsure;


clean_cs

classes/ObjectModel.php
Matt75: You have renamed $res to $is_valid yourself ;)

mickaelandrieu: No problem, let's introduce back $res !

mickaelandrieu:
> I've been waiting for this for a long time

You don't need to wait for this => https://github.com/friends-of-presta/super-model

kpodemski:
Small but very useful, thanks!

classes/ObjectModel.php
matks: So you mean calling both `method_exists($a, $b)` and `is_callable($a::$b)` ?

What I'm afraid is a crazy PHP behavior where `method_exists($a, $b)` returns true (and it can be used as a function) but `is_callable($a::$b)` returns false. Some edge case with nasty magic methods.

In this PR was modifying some basic class of prestashop, I would not be this paranoid but this PR modifies something core of PrestaShop 😄extra caution is needed here.

classes/ObjectModel.php
matks: If I read correctly https://www.php.net/manual/en/language.types.callable.php (EN version) then `call_user_func(Validate::'. $data['validate'], $value)` handles less usecases than `call_user_func(['Validate', $data['validate']], $value)`.

See the 1st snippet of code "Example 1 Callback function examples"
```
call_user_func(array('B', 'parent::who')); // A
```
If `$data['validate']` is a string `parent::a_function_name` then your concatenation will convert it to `Validate::parent::a_function_name` which I dont think works 🤔 

Again, I'm sorry for being paranoid this PR modifies something core of PrestaShop 😄extra caution is mandatory here. And since we did not defined [a clear contract](https://build.prestashop.com/news/prestashop-in-2019-and-beyond-part-2-pain-points/#no-clearly-defined-api) for PrestaShop API we have to make sure we preserve all existing behaviors following SemVer.

mickaelandrieu: >  we have to make sure we preserve all existing behaviors following SemVer.

This is pretty simple! the only thing that works now is when you pass the name of the function as value it looks for available methods in `Validate` class.

* Current behavior when using `parent::a_function_name` => throw a PrestaShop Exception
* New behavior when using `parent::a_function_name` => throw a PrestaShop Exception

Why ?

_Because of the method_exists check_: https://3v4l.org/dOskN

classes/ObjectModel.php
mickaelandrieu: > What I'm afraid is a crazy PHP behavior where method_exists($a, $b) returns true (and it can be used as a function) but is_callable($a::$b) returns false.

Exactly my point, if the methods exists but is not callable, the call of `call_user_func` will fail and you won't have any Exception: I'm fixing a bug here !

To confirm that, try to create a private function in Validate class and then there is the current behavior => https://3v4l.org/II9mG

mickaelandrieu:
Long story short: the feature is now available for everyone (PS 1.6.1+) thanks to my module.

If I contribute unit tests, will you accept the contribution or not ?

If you don't, let's close this :+1:

matks:
> Long story short: the feature is now available for everyone (PS 1.6.1+) thanks to my module.
> 
> If I contribute unit tests, will you accept the contribution or not ?

I can approve it but a 2nd approval will be needed and QA validation too (dont know whether this is relevant here). So I cannot guarantee merge, only my review

matks:
@mickaelandrieu OK, after extensive checks it looks the question around method_exists and your implementaton looks good.

However there is another thing I'd like to discuss:

Previous implementation:
```
                $res = true;
                if (Tools::strtolower($data['validate']) == 'iscleanhtml') {
                    if (!call_user_func(['Validate', $data['validate']], $value, $ps_allow_html_iframe)) {
                        $res = false;
                    }
                } else {
                    if (!call_user_func(['Validate', $data['validate']], $value)) {
                        $res = false;
                    }
                }
```
Your implementation:
```
$is_valid = true;
                if (Tools::strtolower($data['validate']) === 'iscleanhtml') {
                    if (!call_user_func(['Validate', $data['validate']], $value, $ps_allow_html_iframe)) {
                        $is_valid = false;
                    }
                }

                if (!call_user_func($callable, $value)) {
                    $is_valid = false;
                }
```

So, if `Tools::strtolower($data['validate'])` is  `'iscleanhtml'` we're going to call call_user_func twice in your implementation while in previous implementation it was only once. So `iscleanhtml` validation will be applied twice, with different parameters. Looks like a bug 🤔wdyt ?

mickaelandrieu:
> Looks like a bug 🤔wdyt ?

You're right, I need time to write tests and as it's an Object Model it's not that easy :/

In the meantime, I need to introduce it back this ``else`` statment, nice catch !

matks:
> You're right, I need time to write tests and as it's an Object Model it's not that easy :/

One possible way to workaround this is to move the logic you created into a dedicated class (ValidateUtils :trollface: ? ValidateExtension ? ValidateTwo ?) which complies to SOLID principles (and should, as a consequence, be very easy to unit test).

Another workaround is to remove from this part of the code all coupling statements (like all static calls to other classes) and put them in a dedicated untested area.
Example:
```
function IAmABadFunction() {
    $data = CallOne::doIt();
    $data = Tools::modifyData($data);
    $data = array_filter('nl2br', $data);
    // some logic modifying $data
    return $data;
}
```
⬇️ 
```
function IAmABadFunction() {
    $data = CallOne::doIt();
    $data = Tools::modifyData($data);
    return $this->IAmAGoodFunctionWithoutStaticCalls($data); // IAmAGoodFunctionWithoutStaticCalls is easier to unit test
}
```

mickaelandrieu:
> $this->IAmAGoodFunctionWithoutStaticCalls

Watch out, static are not " that " evil ! The only evil (static) functions are the one that modify something they are not supposed to.

Ex:

```php
class Foo
{
     private static $c;

     public static function solid($a, $b)
     {
         return $a + $b;
     }

     public static function stupid($a, $b)
     {
         self::$c = $a + $b;

         return self::$c;
     }
}
```

> One possible way to workaround this is to move the logic you created into a dedicated class (ValidateUtils :trollface: ? ValidateExtension ? ValidateTwo ?) which complies to SOLID principles (and should, as a consequence, be very easy to unit test).

I do agree with it, let me refactor this ... maybe by Friday. Still, I'll try to get my functionnal tests work too: we are never enough cautious with this *** ;)

mickaelandrieu:
> I do agree with it, let me refactor this

Thinking about this, it's not a good idea to do a refactoring while providing a new feature.

I want to keep this pull request as short as possible because I don't have time for big ones.

That's why I've decided to close my [another contribution about PHP 7.4](https://github.com/PrestaShop/PrestaShop/pull/18190) to focus on what I can see merged soon.

I've added unit tests as expected, in the "right place" I guess ?

tests-legacy/Unit/Classes/ObjectModelTest.php
```diff
@@ -0,0 +1,92 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function setUp()
+    {
+        # Old retro compatible way
+        Alias::$definition['fields']['fullname'] = [
+        'type' => Alias::TYPE_STRING,
+        'validate' => 'isString'
```
Progi1984:
```suggestion
            'type' => Alias::TYPE_STRING,
            'validate' => 'isString'
```

mickaelandrieu:
Hello @Progi1984 ,

done and you're now the proud co-author of this feature ;)

tests-legacy/Unit/Classes/ObjectModelTest.php
```diff
@@ -0,0 +1,92 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace LegacyTests\Unit\Classes;
+
+use LegacyTests\TestCase\UnitTestCase;
+use Alias;
+
+class ObjectModelTest extends UnitTestCase
```
matks, line 32: I dont feel good having a unit test modifying `Alias::$definition` in `setUp()` and `tearDown()` 🤔.

If multiple unit tests are run at the same time without process isolation, this might break another unit test using `Alias` right ? 🤔 

Can we think about one way to use a inoffensive ObjectModel for this test ? A dummy ObjectModel.

The very best would be to create the dummy class inside `setUp()` and delete it in `tearDown()` but I dont know if we can do this.

PierreRambaud: Agree, please add a DummyObjectModel linked to these tests

mickaelandrieu: > Agree, please add a DummyObjectModel linked to these tests

It wont work or be autoloaded except if I put it on `classes` folder at the root of PrestaShop: you won't agree on this.

I can mock one, but mock a class to test one of its public method... ? Not sure at all it will proves something.

> If multiple unit tests are run at the same time without process isolation, this might break another unit test using Alias right ? :thinking:

Not at all, `setUp` and `tearDown` will be executed only for assertions of this class: what is done in `setUp` is reverted in `tearDown`.

I can eventually extends an existing one instead of using it: what do you think ?

```php
class DummyObjectModel extends ObjectModel
{
      // alter the definition
}
```

matks: > I can eventually extends an existing one instead of using it: what do you think ?
> 
> ```
> class DummyObjectModel extends ObjectModel
> {
>       // alter the definition
> }
> ```
But how will you autoload it if only `classes` folder is being parsed by ObjectModel ?

I agree this is not easy to fix 🤔.

mickaelandrieu: > But how will you autoload it if only classes folder is being parsed by ObjectModel ?

As my DummyObject would be located in a namespaced file, imho Composer will do the job.

Also, another alternative: anonymous class ?

```php
private function dummy()
{
        return new class() extends ObjectModel {
             public $definition = [...]; // what i need to do.
        };
}
```

matks: Anonymous class would be perfect 👍 absolutely no possible side-effect I can think of that could affect other running tests

mickaelandrieu:
> Anonymous class would be perfect +1 absolutely no possible side-effect I can think of that could affect other running tests

Sounds like tests pass AND no side effect AND everyone is finally happy here :trollface:

PrestaEdit:
😢

PrestaEdit:
Hi @PierreRambaud , @Progi1984, @matks and @mickaelandrieu !

Regarding this kind of PR, now that we are focusing on 8.0, could we think about this change inside the core, maybe ?

PR #18445, opened 2020-04-02
Fix tooltip on product buttons of Order View staying on the page after delete and not being visible when adding a product

### LINKED ISSUE #18387: The tooltip of Delete/Edit product from cart ( order view page migrated) -  wrong behaviour
#### Steps to reproduce

- Go to order view page (migrated ) for order with id=4 (`sell/orders/orders/4/view`).
- Add a product to existing cart. Now you have two products but there is  a problem of the tooltip when hovering on "delete icon" and "Edit" icon, it doesn't work for the newly added products.
image
 - Delete the newly added product, now you see a weird behaviour of the tooltip:

image
 - play with page scroll, you will see the tooltip go to the top left of the screen:

image

#### Expected behavior

The tooltip must appear for every product on cart when hovering on icon, in other words, for both producst (those who are loaded on first page load and those who are loaded via ajax).
When the delete icon is clicked, all tooltips must disapear till we hover on one of them.

 - Mozilla Firefox 74.0
- New installation of 1.7.7.0

NeOMakinG:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Tooltips was staying on the page after deleting a product on Order View and were not added when adding a product
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18387.
| How to test?  | The tooltip on products buttons in Order View must appear for every product on cart when hovering on icon, in other words, for both producst (those who are loaded on first page load and those who are loaded via ajax). When the delete icon is clicked, all tooltips must disapear till we hover on one of them.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18445)


NeOMakinG:
@LouiseBonnard I think it has nothing to do with wording, but I let you remove the label as I don't want to miss something :)

admin-dev/themes/new-theme/js/pages/order/view/order-view-page.js
```diff
@@ -138,7 +140,15 @@ export default class OrderViewPage {
...
     $(OrderViewPageMap.productDeleteBtn)
       .off('click')
-      .on('click', event => this.orderProductManager.handleDeleteProductEvent(event));
+      .on('click', event => {
+        this.orderProductManager.handleDeleteProductEvent(event);
+      }
+    );
+  }
+
+  resetToolTips() {
```
arouiadib, line 159: Thank you for this PR :)
I want to ask you 2 questions:
- Isn't it better to put this function in order product renderer and call it `this.orderProductRenderer.resetTooltips()` here?
- What about compiled assets? should I commit them also when I create a front end PR?
Thank you very much for you time and answer.

arouiadib (dismissed):

admin-dev/themes/new-theme/js/pages/order/view/order-view-page.js
NeOMakinG: Hello @arouiadib 

- I still prefered to put this method in the same file than others listeners resets, I was like "Why putting it in another file if every listeners are in this one ?
- Compiled assets needs to be commited on every branch before 1.7.8.x, after 1.7.7, we decided to stop commiting assets as it is an error, cause too much useless rebases and except when you'r building a library, the best practice is to have source files only on a repository.

For the first point, maybe the PrestaShop's team will decline it, let's see what they think about it too :)

You'r welcome! 😄

Progi1984 (dismissed):

matthieu-rolland (dismissed):

matks (dismissed):

SD1982:
Hi @NeOMakinG Can you please resolve the conflicts ? Thanks !

NeOMakinG:
@SD1982 assets, not blocking QA, but doing it anyway..

NeOMakinG:
Need reapprove then..

Progi1984 (dismissed):

jolelievre (commented):
@NeOMakinG this seems really bothersome to fix this behaviour button by button
Could this be fixed in the `pstooltip` component? Like dispose the tooltip on click?

(although it seems pstooltip is actually a copy of jquery tooltip function)

PierreRambaud (dismissed):

NeOMakinG:
> @NeOMakinG this seems really bothersome to fix this behaviour button by button
> Could this be fixed in the `pstooltip` component? Like dispose the tooltip on click?
> 
> (although it seems pstooltip is actually a copy of jquery tooltip function)

Well, would be usefull to add a data-attr to dispose the tooltip on click yes, but it's on 1.7.7.x and that should be done on the UIKit, which can't be updated on 1.7.7.x...

Progi1984:
> > @NeOMakinG this seems really bothersome to fix this behaviour button by button
> > Could this be fixed in the `pstooltip` component? Like dispose the tooltip on click?
> > (although it seems pstooltip is actually a copy of jquery tooltip function)
> 
> Well, would be usefull to add a data-attr to dispose the tooltip on click yes, but it's on 1.7.7.x and that should be done on the UIKit, which can't be updated on 1.7.7.x...

In this case, create an issue for 178x ;)  (with a correct fix for this) (ping @jolelievre)

NeOMakinG:
@Progi1984 I'm not in favor of extending an external library... this is an issue from Bootstrap, they should implement this, look at TinyMCE, we tweaked it (I agree that was a bit dirty anyway) and it was a complete pain to upgrade it.

Progi1984:
Hello @NeOMakinG, thanks for your contribution, would you mind to rebase your pull request? Thanks

Progi1984 (approved):

matthieu-rolland (approved):
lgtm !

Progi1984:
Thanks @NeOMakinG

PR #18451, opened 2020-04-03
Use scandir instead of readdir to get sorted entities

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | Readdir function return an incorrect order depending on the server you're running it. Use scandir instead
| Type?         | bug fix
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Installation must work as before, tests are green. If you want to go further, you can compare the database table `ps_configuration`.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18451)

### DIFF
--- a/src/PrestaShopBundle/Install/XmlLoader.php
+++ b/src/PrestaShopBundle/Install/XmlLoader.php
@@ -177,10 +177,9 @@ class XmlLoader
     public function getSortedEntities()
     {
         // Browse all XML files from data/xml directory
-        $entities = array();
-        $dependencies = array();
-        $fd = opendir($this->data_path);
-        while ($file = readdir($fd)) {
+        $entities = [];
+        $dependencies = [];
+        foreach (scandir($this->data_path) as $file) {
             if (preg_match('#^(.+)\.xml$#', $file, $m)) {
                 $entity = $m[1];
                 $xml = $this->loadEntity($entity);
@@ -199,7 +198,6 @@ class XmlLoader
                 $entities[] = $entity;
             }
         }
-        closedir($fd);
 
         // Sort entities to populate database in good order (E.g. zones before countries)
         do {


matthieu-rolland (approved):

src/PrestaShopBundle/Install/XmlLoader.php
```diff
@@ -177,10 +177,9 @@ public function setIds($ids)
     public function getSortedEntities()
     {
         // Browse all XML files from data/xml directory
-        $entities = array();
-        $dependencies = array();
-        $fd = opendir($this->data_path);
-        while ($file = readdir($fd)) {
+        $entities = [];
+        $dependencies = [];
+        foreach (scandir($this->data_path) as $file) {
```
PululuK, line 182: Hi @PierreRambaud 
Is it possible to use Tools Classe here?
```php
Tools::scandir($this->data_path,'xml')
```
See https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Tools.php#L3768 

Can you allow me to refactor this class please ? Use `Tools` instead of `scandir` + `preg_match`?

PierreRambaud: IMHO, it's not a good idea, this class must be outside the Core because it's only related to the install process. And the `Tools::scandir` doesn't provide great value :/

PululuK: Ok Thanks you @PierreRambaud

matks (approved):

jolelievre (approved):

SD1982:
LGTM Thanks @PierreRambaud

Progi1984:
Thanks @PierreRambaud

PR #18454, opened 2020-04-03
[POC] Product page migration. ProductBaseData query/command

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Prepares query getProductBaseData which provides product name and type. Also adds command UpdateProductName. Initially the purpose was to add command for updating both- name and type(the base data), but the product type seems to be more complex than just setting object model property.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18454)


src/Adapter/Product/QueryHandler/GetProductBaseDataHandler.php
```diff
@@ -0,0 +1,50 @@
...
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Product\QueryHandler;
+
+use PrestaShop\PrestaShop\Adapter\Product\AbstractProductHandler;
+use PrestaShop\PrestaShop\Core\Domain\Product\Query\GetProductBaseData;
+use PrestaShop\PrestaShop\Core\Domain\Product\QueryHandler\GetProductBaseDataHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\QueryResult\ProductBaseData;
+
+final class GetProductBaseDataHandler extends AbstractProductHandler implements GetProductBaseDataHandlerInterface
```
matks, line 36: So we go for "BaseData" name for "name+type" ? 😄

src/Core/Domain/Product/QueryResult/ProductBaseData.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public function getLocalizedNames(): array
+    {
+        return $this->localizedNames;
+    }
+
+    /**
+     * @return ProductType
+     */
+    public function getProductType(): ProductType
```
matks, line 66: Is it necessary ? 🤔
I'm wondering ... since it will always be an integer, do we need a VO ?

src/Adapter/Product/QueryHandler/GetProductBaseDataHandler.php
zuk3975: guess so :sweat_smile:

src/Core/Domain/Product/QueryResult/ProductBaseData.php
zuk3975: idk where to keep those integers constants then and make sure its valid?

PR #18468, opened 2020-04-04
Fix Order.getInvoicesCollection() query

### LINKED ISSUE #12588: Order payment duplicated depending on order state configuration

erouvier29:
| Questions     | Answers
| Branch?       | develop
| Description?  | Avoid "dummy" invoices with number 0 to be returned. It is the case when a delivery slip exists but not an invoice. And it generates a "double" payment under the conditions described below
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Place an order that is paid and invoice is not yet generated. Change order state to Preparation (previously configured with paid = true, delivery = true, invoice = false). Without the fix a "double" payment is automatically created.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18468)

### DIFF
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1976,6 +1976,7 @@ class OrderCore extends ObjectModel
     {
         $order_invoices = new PrestaShopCollection('OrderInvoice');
         $order_invoices->where('id_order', '=', $this->id);
+        $order_invoices->where('number', '!=', 0);
 
         return $order_invoices;
     }


matks:
Nice bug catch, thanks !

matks:
@erouvier29 I created a bug report describing your bug in https://github.com/PrestaShop/PrestaShop/issues/18488 so that this bug is reproduced and tracked 😉please tell me if I wrote it well

Progi1984 (approved):

matthieu-rolland (approved):

matks:
OK I'll also look at https://github.com/PrestaShop/PrestaShop/pull/15417 (author stopped working on the PR) which is supposed to fix the bug too

colinegin:
It seems this PR has been approved by developers, should we put the label "waiting for QA" ?

erouvier29:
Replaced by #19260, as there is a regression with this one:
Delivery slip is not generated while invoice is not generated yet

PR #12968, opened 2019-03-19
Add missing deprecation on previous PerformanceFormHandler

matks:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Add missing deprecation on previous PerformanceFormHandler.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | No need for QA

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/12968)

PR #18470, opened 2020-04-05
Allow developer to customize the Checkout Process

mickaelandrieu:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | This is actually impossible to configure the Checkout Process, there is no Hook so it forces every developper that need to do it to ... override the Core
| Type?         | new feature
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | See [this](https://www.prestashop.com/forums/topic/948530-resolu-supprimer-l%C3%A9tape-moyen-de-livraison-dans-le-tunnel-de-commande/)
| How to test?  | Module on FoP to come as even if you accept it, we won't have this feature before 1.7.8, so 6 months at least. And this is sad because we need an easy way to allow the "Click & Collect" right now.

```php
// in a module...
    /**
     * Remove the delivery part of checkout process (NO Delivery)
     */
    public function hookActionBootstrapCheckoutProcess($hookArgs)
    {
        /** @var CheckoutProcess $checkoutProcess */
        $checkoutProcess = $hookArgs['checkoutProcess'];

        $checkoutSteps = $hookArgs['steps'];

        /** @var CheckoutStepInterface $checkoutStep */
        foreach ($checkoutSteps as $index => $checkoutStep) {
            if ($step->getIdentifier() === 'checkout-delivery-step') {
                unset($checkoutSteps[$index];
            }
        }

        // now available ... but risky
        $checkoutSteps[] = new InstanceOfCheckoutStepInterface();
        
        // safer imho
        $checkoutProcess->add(new InstanceOfCheckoutStepInterface());
    }
```

preview

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18470)

PR #18475, opened 2020-04-06
Enable controllers as a service through autowiring

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Enable controllers as a service through autowiring. See below for details
| Type?         | new feature
| Category?     | CO
| BC breaks?    | no (maybe one, see below)
| Deprecations? | no
| Fixed ticket? | 

## Description

@eternoendless noticed our controllers are not registered as Symfony services, which prevents [decorating](https://symfony.com/doc/3.4/service_container/service_decoration.html) them.

In this PR I enable autowiring for the Controller folder which will register the Controllers inside into the container. I added a custom tag to them as it might be useful in the future, to distinguish them from "regular" services.
I excluded the `Controller\Api` folder because the controllers in this folder are already registered through standard YAML services configuration file.

I also "fixed" a broken class because else autowiring could not work: it's class `src/PrestaShopBundle/Form/Admin/AdvancedParameters/Performance/PerformanceFormHandler.php` . This class extended a class that is not here anymore. I added a [deprecation notice](https://github.com/PrestaShop/PrestaShop/pull/12968) on develop but I needed to syntaxically fix this class to make the autowiring work.

⚠️autowiring has declared these controllers as **private** so they can be used for decoration but cannot be fetched from container using `$container->get()`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18475)

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -19,9 +19,9 @@ parameters:
 # Autowires Core controllers
 services:
     PrestaShopBundle\Controller\:
-        resource: '../../src/PrestaShopBundle/Controller/*'
-        exclude: '../../src/PrestaShopBundle/Controller/Api'
-        tags: ['prestashop_core.controllers']
+        resource: "%kernel.root_dir%/../src/PrestaShopBundle/Controller/*"
+        exclude: "%kernel.root_dir%/../src/PrestaShopBundle/Controller/Api"
+        tags: ['prestashop.core.controllers']
 
 framework:
     assets:
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -16,6 +16,13 @@ parameters:
     mail_themes_uri: "/mails/themes"
     mail_themes_dir: "%kernel.project_dir%%mail_themes_uri%"
 
+# Autowires Core controllers
+services:
+    PrestaShopBundle\Controller\:
+        resource: '../../src/PrestaShopBundle/Controller/*'
+        exclude: '../../src/PrestaShopBundle/Controller/Api'
+        tags: ['prestashop_core.controllers']
+
 framework:
     assets:
         version: !php/const \AppKernel::VERSION
--- a/src/PrestaShopBundle/Form/Admin/AdvancedParameters/Performance/PerformanceFormHandler.php
+++ b/src/PrestaShopBundle/Form/Admin/AdvancedParameters/Performance/PerformanceFormHandler.php
@@ -27,7 +27,6 @@
 namespace PrestaShopBundle\Form\Admin\AdvancedParameters\Performance;
 
 use PrestaShop\PrestaShop\Adapter\Feature\CombinationFeature;
-use PrestaShop\PrestaShop\Core\Form\AbstractFormHandler;
 use PrestaShop\PrestaShop\Core\Form\FormDataProviderInterface;
 use Symfony\Component\Form\FormFactoryInterface;
 
@@ -35,7 +34,7 @@ use Symfony\Component\Form\FormFactoryInterface;
  * This class manages the data manipulated using forms
  * in "Configure > Advanced Parameters > Performance" page.
  */
-final class PerformanceFormHandler extends AbstractFormHandler
+final class PerformanceFormHandler
 {
     /**
      * @var FormFactoryInterface
--- a/src/PrestaShopBundle/Form/Admin/AdvancedParameters/Performance/PerformanceFormHandler.php
+++ b/src/PrestaShopBundle/Form/Admin/AdvancedParameters/Performance/PerformanceFormHandler.php
@@ -61,9 +61,6 @@ final class PerformanceFormHandler
         $this->formDataProvider = $formDataProvider;
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function getForm()
     {
         $formBuilder = $this->formFactory->createBuilder()
@@ -83,9 +80,6 @@ final class PerformanceFormHandler
         return $formBuilder->setData($formBuilder->getData())->getForm();
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function save(array $data)
     {
         $errors = $this->formDataProvider->setData($data);
--- a/src/PrestaShopBundle/Resources/config/services/bundle/controller.yml
+++ b/src/PrestaShopBundle/Resources/config/services/bundle/controller.yml
@@ -2,7 +2,7 @@ services:
     _defaults:
         public: true
 
-    prestashop.core.api.controller:
+    PrestaShopBundle\Controller\Api\ApiController:
         abstract: true
         class: PrestaShopBundle\Controller\Api\ApiController
         calls:
@@ -11,9 +11,12 @@ services:
         tags:
           - { name: prestashop.core.controllers }
 
-    prestashop.core.api.stock.controller:
+    prestashop.core.api.controller:
+        alias: PrestaShopBundle\Controller\Api\ApiController
+
+    PrestaShopBundle\Controller\Api\StockController:
         class: PrestaShopBundle\Controller\Api\StockController
-        parent: prestashop.core.api.controller
+        parent: PrestaShopBundle\Controller\Api\ApiController
         public: true
         properties:
             movements: "@prestashop.core.api.stock.movements_collection"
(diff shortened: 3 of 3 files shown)


app/config/config.yml
```diff
@@ -16,6 +16,13 @@ parameters:
     mail_themes_uri: "/mails/themes"
     mail_themes_dir: "%kernel.project_dir%%mail_themes_uri%"
 
+# Autowires Core controllers
+services:
+    PrestaShopBundle\Controller\:
+        resource: '../../src/PrestaShopBundle/Controller/*'
+        exclude: '../../src/PrestaShopBundle/Controller/Api'
```
matks: This folder contains already-registered controllers as services so I exclude it (see my PR description)

src/PrestaShopBundle/Form/Admin/AdvancedParameters/Performance/PerformanceFormHandler.php
```diff
@@ -27,15 +27,14 @@
...
-use PrestaShop\PrestaShop\Core\Form\AbstractFormHandler;
 use PrestaShop\PrestaShop\Core\Form\FormDataProviderInterface;
 use Symfony\Component\Form\FormFactoryInterface;
 
 /**
  * This class manages the data manipulated using forms
  * in "Configure > Advanced Parameters > Performance" page.
  */
-final class PerformanceFormHandler extends AbstractFormHandler
+final class PerformanceFormHandler
```
matks, line 37: I needed to remove this class (which does not exist anymore) else the autowiring would fail (see my PR description)

src/PrestaShopBundle/Resources/config/services/bundle/controller.yml
```diff
@@ -8,6 +8,8 @@ services:
         calls:
             - [setLogger, ['@logger']]
             - [setContainer, ['@service_container']]
+        tags:
+          - { name: prestashop_core.controllers }
```
matks: Adding the "controller tag" to the controllers from `Controller\API` folder

app/config/config.yml
eternoendless: You can add this as a comment in the yaml file

eternoendless: Can you use `%kernel.root_dir%` instead of relative paths?

src/PrestaShopBundle/Resources/config/services/bundle/controller.yml
eternoendless: Any reason to use `prestashop_core` instead of `prestashop.core`?

app/config/config.yml
eternoendless: We could also declare the previous services as alias?

src/PrestaShopBundle/Resources/config/services/bundle/controller.yml
matks: No, I just wanted to make unambiguous the fact these are the core controllers (to distinguish from module controllers or 3rd party controllers). I chose the name in 2 seconds, it can be improved indeed 👍

app/config/config.yml
matks: > We could also declare the previous services as alias?

It does not seem possible because **these controllers use [property injection](https://symfony.com/doc/current/service_container/injection_types.html#property-injection)** to be constructed.
See for example the [API StockController](https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Resources/config/services/bundle/controller.yml#L12):
```
class StockController extends ApiController
{
    /**
     * @var StockRepository
     */
    public $stockRepository;

    /**
     * @var QueryStockParamsCollection
     */
    public $queryParams;

    /**
     * @var MovementsCollection;
     */
    public $movements;
```

I could replace this with a constructor and allow auto-wiring to do his job (if all necessary objects are correctly type-hinted) but this would be a BC break 🤔

eternoendless: These services could be declared by class name as service name, then aliased to the former names, so that there is coherence in the naming scheme for all controllers.

In all truth, we should have declared service names instead of autowiring them, because I don't think we use that anywhere else.

matks: > These services could be declared by class name as service name, then aliased to the former names, so that there is coherence in the naming scheme for all controllers.

I see what you mean. Modify YAML file to use the classnames are service IDs, then add aliases. I can do that.

> In all truth, we should have declared service names instead of autowiring them, because I don't think we use that anywhere else.

You want me to close this PR and create one with a big YAML file which declares all controllers as services instead ? I can do it but then we'll have to be careful to always think about adding new controllers into it. The autowiring will always be up-to-date whereas the manual declaration will require human review to be always accurate.

eternoendless: I know that autowiring is better, using class names is the de-facto standard in symfony because of autowiring.

> Modify YAML file to use the classnames are service IDs, then add aliases.

That's probably the best solution. Let's keep autowiring and classname services for controllers, and add deprecated aliases for API controllers. Won't be pretty compared with other services, but at least it's a step in the good direction, I guess? Less risky, too.

matks:
Note to myself: need to do "Modify YAML file to use the classnames are service IDs, then add aliases. I can do that."

mickaelandrieu:
> eternoendless added this to the 1.7.7.0 milestone 11 days ago

Any chance to get something else added to 1.7.7.0 ? :angel:

matks:
> > eternoendless added this to the 1.7.7.0 milestone 11 days ago
> 
> Any chance to get something else added to 1.7.7.0 ? 👼

This PR was (exceptionnally) added to 1.7.7.0 scope by @eternoendless in order to allow Controllers decoration for module developers. Without controllers declared as services, they couldn't do it following the standard easy process.

If you have a similar customization need and PR to allow it is really simple, @eternoendless might accept it although it breaks our rules. We ❤️developers so giving them always more ways to customize/extend PrestaShop is a must-have.

mickaelandrieu:
This one is not risky at all => a simple Hook :) https://github.com/PrestaShop/PrestaShop/pull/18470

src/PrestaShopBundle/Resources/config/services/bundle/controller.yml
```diff
@@ -17,6 +19,8 @@ services:
             movements: "@prestashop.core.api.stock.movements_collection"
             stockRepository: "@prestashop.core.api.stock.repository"
             queryParams: "@prestashop.core.api.query_stock_params_collection"
+        tags:
```
jolelievre, line 25: Is it requried to add the tag on each definition? Since it's already defined on the parent?

src/PrestaShopBundle/Form/Admin/AdvancedParameters/Performance/PerformanceFormHandler.php
jolelievre: If I understand this class is actually not used any more? So you just clean it to avoid a bug during container compilation?
The weirdest thing is that I can't fin any version where `AbstractFormHandler` actually exists.. it was introduced during 1.7.4 development, but it seems like it was removed before the release but I can't find when

matks: Correct. Useless class but I dont know what SemVer says about useless classes so I prefer to let it here.

src/PrestaShopBundle/Resources/config/services/bundle/controller.yml
matks: Yes, service inheritance seems not to keep tags (I checked)

matks:
@eternoendless @jolelievre feedback applied, now we have aliases for API controllers

jolelievre (approved):

Progi1984 (approved):

matks:
Waiting for @eternoendless confirmation, then we can merge (I think no QA needed ?)

eternoendless (approved):

eternoendless:
Thank you @matks 

This should be added to the release notes.

mickaelandrieu:
preston_died

"OMG, they have killed ~~Kenny~~ Preston ! :'(

eternoendless:
> "OMG, they have killed Kenny Preston ! :'(

Oh no, it's still alive. It just does other things now.

PR #18491, opened 2020-04-07
Installation under CLI doesn't take BASE_URI and Apache rewrite in consideration

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | File `index_cli.php` ignores the `__PS_BASE_URI__` const when running under a CLI context. The problem is `Datas` class needs `Validate` class, `Validate` class need to be autoloaded, and to make the autoload work, it need `__PS_BASE_URI__`...  <br>To make it work, I add a `class_exist` check in `Datas` class to ignore the CLI validation only to initialize the const value.<br>One more thing is added here, the url rewrite engine is not properly configured during the installation. It should be configurable by a simple `--rewrite` parameter, default value is true.
| Type?         | bug fix
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Execute this pull request with the `PrestaShop/vagrant` and see the blockreassurance module if images are displayed in the Cart process. Moreover, the apache rewrite is activated in the FO.<br>`PR=18491 BRANCH=1.7.6.x vagrant up --provision`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18491)

### DIFF
--- a/install-dev/index_cli.php
+++ b/install-dev/index_cli.php
@@ -33,9 +33,20 @@ if (!defined('PHP_VERSION_ID') || PHP_VERSION_ID < _PS_INSTALL_MINIMUM_PHP_VERSI
 
 /* Redefine REQUEST_URI */
 $_SERVER['REQUEST_URI'] = '/install/index_cli.php';
+require_once __DIR__ . DIRECTORY_SEPARATOR . 'classes/datas.php';
+/**
+ * The autoload needs constant (__PS_BASE_URI__) declared in the init.php
+ * to work properly.
+ * And, this one can have a custom value depending on what the user specify in arguments.
+ *
+ * Using getAndCheckArgs is quite redundant because it's also used in controllerConsole,
+ * but it prevent a duplicate logic and allows the program to retrieve the base_uri
+ * value from the CLI.
+ */
+Datas::getInstance()->getAndCheckArgs($argv);
+
 require_once dirname(__FILE__).'/init.php';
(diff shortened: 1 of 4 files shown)


install-dev/index_cli.php
```diff
@@ -33,9 +33,12 @@
 
 /* Redefine REQUEST_URI */
 $_SERVER['REQUEST_URI'] = '/install/index_cli.php';
+require_once __DIR__ . DIRECTORY_SEPARATOR . 'classes/datas.php';
+// Because some constantes need to be initialized from CLI parameters
```
Quetzacoalt91:
```suggestion
// Because some constants need to be initialized from CLI parameters
```

Progi1984 (dismissed):

install-dev/index_cli.php
matks:
```suggestion
// Becoz some constantes need 2b initialised from CLI parametres
```

Progi1984 (dismissed):

Progi1984 (dismissed):

Progi1984 (approved):

atomiix (approved):

matks (approved):

SD1982:
lgtm The PR correctly fix the missing images for the reassurance block when installing with cli. Thanks @PierreRambaud

Progi1984:
Thanks @PierreRambaud

PR #18493, opened 2020-04-07
Change product redirection rules to redirect to valid attribute url

### LINKED ISSUE #18299: Canonical redirects for product pages do not work, gives duplicate content
#### Describe the bug

Canonical urls do not work on product pages. You can change the link rewrite text to anything. If you use multishop, it will not redirect to www subdomain properly. This will give duplicate content in search engines.

#### Expected behavior

If a link is entered without www, it should redirect to www.
If a product link is entered with the wrong link rewrite text, it should redirect to the correct link.

#### Steps to Reproduce

1. Go to https://typical-wax.demo.prestashop.com/en/women/2-9-brown-bear-ddddd-sweater.html#/1-size-s
2.  See error (it should redirect to 
https://typical-wax.demo.prestashop.com/en/women/2-9-brown-bear-printed-sweater.html#/1-size-s

* PrestaShop version: 1.7.6.1+
* PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | The url of product with combinations was no longer redirected, so invalid urls (from attribute ID, rewriting part or domain error) were no longer redirected This PR now reintroduce the systematic redirection (when necessary) Only valid combination are allowed, and canonical url (without attributes) is used as fallback
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18299
| How to test?  | All these tests should be done with Friendly urls activated. Go to a product page with combination:<br>- change the rewriting url part -> you should be redirect to the same combination with appropriate rewriting<br>- edit the rewriting part in BO and refresh the product page -> you should be redirected to the same combination with new rewriting<br>- use a different (yet accessible of course) domain (like adding/removing www) -> you should be redirected to the same combination with the proper shop domain<br>- enter an invalid combination id (use high value) -> you should be redirected to the canonical url (without attribute) and the default combination is displayed

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18493)

### DIFF
--- a/controllers/front/ProductController.php
+++ b/controllers/front/ProductController.php
@@ -55,18 +55,17 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
     public function canonicalRedirection($canonical_url = '')
     {
         if (Validate::isLoadedObject($this->product)) {
-            if (!$this->product->hasCombinations() ||
-                !$this->isValidCombination(Tools::getValue('id_product_attribute'), $this->product->id)) {
+            $idProductAttribute = Tools::getValue('id_product_attribute', null);
+            if (!$this->product->hasCombinations() || !$this->isValidCombination($idProductAttribute, $this->product->id)) {
                 //Invalid combination we redirect to the canonical url (without attribute id)
                 unset($_GET['id_product_attribute']);
-            } else {
-                //Only redirect to canonical (parent product without combination) when the requested combination is not valid
-                //In this case we are in a valid combination url and we must display it without redirection for SEO purpose
-                return;
+                $idProductAttribute = null;
             }
 
-            //Note: we NEED these 6 arguments to have $ipa=null or else a parameter will be added
-            //id_product_attribute=0 and force the redirection
+            // If the attribute id is present in the url we use it to perform the redirection, this will fix any domain
+            // or rewriting error and redirect to the appropriate url
+            // If the attribute is not present or invalid, we set it to null so that the request is redirected to the
+            // real canonical url (without any attribute)
             parent::canonicalRedirection($this->context->link->getProductLink(
                 $this->product,
                 null,
@@ -74,7 +73,7 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
                 null,
                 null,
                 null,
-                null
+                $idProductAttribute
             ));
         }
     }


atomiix (approved):

Progi1984 (approved):

matks (approved):

SD1982:
Lgtm the only "thing" I have noticed is : 
=> if we go to a product with combination 
=> the default combination is displayed 
=> If we change combination the URL is modified 
=> if we change the combination ID indeed we are well redirected and the default combination is displayed 
=> but the previously modified parameters are still in the URL see screen record.
=> Shouldn't these parameters be replaced by those of the default combination?

**Screen Record**
https://drive.google.com/file/d/1vnvafdkeOHsp7FN4Nz5Tmf0wMJH3BwXL/view?usp=sharing 

What do you think @jolelievre and @colinegin  ?

colinegin:
Hello @SD1982 ,

Thanks for your feedback. 
I had noticed it also, but i'm not sure this is a regression. Do you reproduce it in 1.7.5.2 ? 
Also, i don't think this anchor is taken into account by Google for SEO purposes so I'm not sure it has impact.
What do you think @jolelievre ?

SD1982:
Hello @colinegin  Yes the behaviour was the same in 1.7.5.2.

colinegin:
Alright, so we won't fix this issue for 1.7.6.5. I've created the issue here : https://github.com/PrestaShop/PrestaShop/issues/18546

jolelievre:
Seems like it is the expected behaviour to keep the anchor after redirect, according to this discussion:
https://stackoverflow.com/questions/5283395/url-hash-is-persisting-between-redirects

The 302 redirection indicates that the content has been moved, and a `#resource` is supposed to link to a resource in the initial page, so it makes sense that it should be found in the redirected page So the browser keeps the anchor (if you look at the url we return in the redirection response it is not present)

Now the question is should the parent page (the canonical one) show the combination marked by the anchor after redirection?

And if it does should it be changed:
- via javascript so that the canonical is always showing the default combination for SEO
- via PHP so that the canonical page varies depending on the anchor (which means the anchor has an SEO meaning)

This can be discussed, but anyway it should be dealt in another issue, and in another version

jolelievre:
Thanks @colinegin So if everything else is ok I guess it's validated for QA @SD1982 ?

SD1982:
Hello @jolelievre Yes It was the only point I wanted to clarify for the rest it seems to me that everything is correct. Thanks @jolelievre

Progi1984:
Thanks @jolelievre

PR #18524, opened 2020-04-08
Remove ability to apply discount on all invoices in order view page

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | This PR comes from another pr (#17348) which scope increased too much, so it needed to be separated. Removes ability to apply a discount to all invoices of an order in order view page. From user point of view it removes the checkbox from this (check screenshot) Screenshot from 2020-04-08 15-40-42
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | See description

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18524)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
@@ -878,7 +878,7 @@ class OrderController extends CommonController
                             $data['name'],
                             $data['type'],
                             $data['value'] ?? null,
-                            (int) $data['invoice_id']
+                            empty($data['invoice_id']) ? null : (int) $data['invoice_id']
                         )
                     );
 
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
@@ -871,8 +871,6 @@ class OrderController extends CommonController
             if ($addOrderCartRuleForm->isValid()) {
                 $data = $addOrderCartRuleForm->getData();
 
-                $invoiceId = $data['apply_on_all_invoices'] ? null : (int) $data['invoice_id'];
-
                 try {
                     $this->getCommandBus()->handle(
                         new AddCartRuleToOrderCommand(
@@ -880,7 +878,7 @@ class OrderController extends CommonController
                             $data['name'],
                             $data['type'],
                             $data['value'] ?? null,
-                            $invoiceId
+                            (int) $data['invoice_id']
                         )
                     );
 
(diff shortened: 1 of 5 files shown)


tests/Integration/Behaviour/Features/Scenario/Order/add_cart_rule_to_order.feature
```diff
@@ -0,0 +1,103 @@
...
+    And the module "dummy_payment" is installed
+    And I am logged in as "test@prestashop.com" employee
+    And I add 2 products "Mug The best is yet to come" to the cart "dummy_cart"
+    And I add order "bo_order1" with the following details:
+      | cart                | dummy_cart                 |
+      | message             | test                       |
+      | payment module name | dummy_payment              |
+      | status              | Awaiting bank wire payment |
+    And tax "Sales-taxes US-FL 6%" is applied to order bo_order1
+    And Order "bo_order1" has following prices:
```
jolelievre: That's a great addition! This validate amounts along with currency
@sowbiba @Progi1984 You might be interested in this one in some of your behat tests

jolelievre (commented):
You removed all the javascript part so it would be enough, but I think it would be safer (and cleaner) to also remove the related server side component like:
- `AddOrderCartRuleType->apply_on_all_invoices` field
- the `apply_on_all_invoices` handling in `OrderController::addCartRuleAction` action

zuk3975:
> You removed all the javascript part so it would be enough, but I think it would be safer (and cleaner) to also remove the related server side component like:
> 
> * `AddOrderCartRuleType->apply_on_all_invoices` field
> * the `apply_on_all_invoices` handling in `OrderController::addCartRuleAction` action

okay. Done that. Just didn't touch the AddCartRuleToOrderHandler because its a risky mess and should be done when its being refactored later i think. (here https://github.com/PrestaShop/PrestaShop/issues/18545).

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -871,16 +871,14 @@ public function addCartRuleAction(int $orderId, Request $request): RedirectRespo
...
-
                 try {
                     $this->getCommandBus()->handle(
                         new AddCartRuleToOrderCommand(
                             $orderId,
                             $data['name'],
                             $data['type'],
                             $data['value'] ?? null,
-                            $invoiceId
+                            (int) $data['invoice_id']
```
jolelievre: Be ceraful `$data['invoice_id']` can still be nul when the order has no invoices yet Did ou check if it works fine?
Or should it look like the value field? `(int) $data['invoice_id'] ?? null` .

zuk3975: added a safety check just to make sure.

matks (approved):

Progi1984 (approved):

khouloudbelguith:
Hi @zuk3975,

It is OK :heavy_check_mark: 
image

PS: I found two bugs testing this PR:
https://github.com/PrestaShop/PrestaShop/issues/18807
https://github.com/PrestaShop/PrestaShop/issues/18805 => regression

Thanks!

Progi1984:
Thanks @zuk3975

PR #18532, opened 2020-04-08
Increase precision for DECIMAL fields in database

### LINKED ISSUE #18379: [BO] Can't remove discount added from a cart rule on product selection after adding a product in order
#### Describe the bug

Can't remove discount added from a cart rule on product selection after adding a product in order

#### Expected behavior

Can remove discount added from a cart rule on product selection after adding a product in order

#### Steps to Reproduce

2. Create a cart rule:
  - has no code
  - product selection on previous created product with 100 euro tax included
  - discount amount of 500 EUR tax included
4. The discount has been added to order
5. **ERROR :** The discount can't be removed

Edit by @Robin-Fischer-PS :
6. Go back to the order after trying to delete the discount
7. The voucher is no longer displayed but the discount is still applied in the total of the order

Video from @Robin-Fischer-PS : 
https://drive.google.com/file/d/1yXPaN5rDVTXcLychf1ft3e_3RbblR3QT/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7.x branch

### LINKED ISSUE #18498: BO-Exception when deleting discount in order view
#### Describe the bug

An exception is displayed when adding 2 discount and deleting them and also if we go back to the order view the discount are deleted but one of them are still displayed

#### Expected behavior

no exception should be displayed when deleting a discount

#### Steps to Reproduce

1. Go to BO => orders => orders detail 
2. Add a discount (ex : 5% )
3. Add a second discount ( ex : 5 euros )
4. Delete the amount discount first
5. Delete the percent discount
(also reproducible if both discount are % discount but not if they are both amount discount )
6. And bonus : if we go back to the order view detail the discounts are deleted but the 5% discount is still applied

**Screenshots**
capture d'écran_8
capture d'écran_9

Edit by @Robin-Fischer-PS :

Please check if with there steps, the bug is also fixed : 
https://drive.google.com/file/d/1q7A2ASSAXirr_8R-Q-ixNppRQH2uwzfj/view

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: 7.2

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | There was a difference between the `total_discount` saved in the `order` table and the value of the discount saved in `order_cart_rule` because those fields didn't have the same precision. This PR fixes it.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | /
| How to test?  | 1. Install a 1.7.6.x shop<br>2. Checkout this PR<br>3. Upgrade using `php install-dev/upgrade/upgrade.php`<br>4. The DB structure should contains modified fields as listed below<br>5. Tests should pass

### ⚠️ BC Break
This PR increases precision but decreases integer size on the following table fields:
* `attribute_impact`.`price` (DECIMAL(20,6))
* `cart_rule`
  * `minimum_amount` (DECIMAL(20,6))
  * `reduction_amount` (DECIMAL(20,6))
* `group`.`reduction` (DECIMAL(5,2))
* `order_detail`
  * `reduction_percent` (DECIMAL(5,2))
  * `group_reduction` (DECIMAL(5,2))
  * `ecotax` (DECIMAL(17,6))
* `product`.`additional_shipping_cost` (DECIMAL(20,6))
* `product_shop`.`additional_shipping_cost` (DECIMAL(20,6))
* `order_cart_rule`
  * `value` (DECIMAL(20,6))
  * `value_tax_excl` (DECIMAL(20,6))

## Other modified fields
* `product_group_reduction_cache`.`reduction` (DECIMAL(5,4))
* `order_slip `
  * `amount` (DECIMAL(20,6))
  * `shipping_cost_amount ` (DECIMAL(20,6))
* `order_payment`.`amount` (DECIMAL(20,6))

The upgrade script handles the possibility that a number is bigger than the new column's max size to prevent mysql error but that means the number will not be correct anymore. For a field that changes from `decimal(17,2)` to `decimal(20,6)`, if the number was `123456789012345.12`, the new number will be `12345678901234.512000`.
Keep in mind that this is an extra security and this will probably never happen as it's unlikely that the concerned fields will contain such big numbers.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18532)


Progi1984 (dismissed):

matthieu-rolland (dismissed):

matks (dismissed):

install-dev/data/db_structure.sql
```diff
@@ -1394,8 +1394,8 @@ CREATE TABLE `PREFIX_order_cart_rule` (
   `id_cart_rule` int(10) unsigned NOT NULL,
   `id_order_invoice` int(10) unsigned DEFAULT '0',
   `name` varchar(254) NOT NULL,
-  `value` decimal(17, 2) NOT NULL DEFAULT '0.00',
-  `value_tax_excl` decimal(17, 2) NOT NULL DEFAULT '0.00',
+  `value` decimal(20, 6) NOT NULL DEFAULT '0.000000',
+  `value_tax_excl` decimal(20, 6) NOT NULL DEFAULT '0.000000',
```
PierreRambaud, line 1398: It's a BC if people had 
```
123456789012345.00
```
You will have an error:
```
#1264 - Out of range value for column 'value' at row 1
```

PierreRambaud (changes requested):
BC break

install-dev/data/db_structure.sql
jolelievre: But this column is managed programmatically anyway when a discount is applied on Order, not manually
Besides if you look a the order table in the same file you can see (amongst other columns) ``total_paid_tax_incl` decimal(20, 6) NOT NULL DEFAULT '0.00',`
So the order won't be able to manage this kind of values either, it makes sens that the discount has the same range no?

PierreRambaud: This table can be populate by a module or an external resource.
Even if other fields are already in `DECIMAL(20, 6)` this one can fail if someone use it

install-dev/data/db_structure.sql
```diff
@@ -910,7 +910,7 @@ CREATE TABLE IF NOT EXISTS `PREFIX_gender_lang` (
 /* Group special price rules */
 CREATE TABLE `PREFIX_group` (
   `id_group` int(10) unsigned NOT NULL auto_increment,
-  `reduction` decimal(17, 2) NOT NULL DEFAULT '0.00',
+  `reduction` decimal(5, 2) NOT NULL DEFAULT '0.00',
```
Progi1984, line 913: 17 => 5 ?

atomiix: Yes, because `reduction` is a percentage and you can't add a reduction above 100% so there is no need to have 15 digits for the integer part.

Progi1984: :ok_hand:

Progi1984: But it's a BC ! Ping @jolelievre

jolelievre: Indeed strictly speaking it's a BC and we will have to document and inform about it But in reality it's like the other changed types, although the range is less wide it shouldn't contain such big values anyway So we're safe

Progi1984: :ok_hand:

Progi1984 (dismissed):

matks (commented):
I'm worried that the upgrade script might require huge resources consumption on big tables. Might even timeout connection to mysql.

atomiix:
@matks I agree with you but I think it's either adding that extra security but that might require huge resources or don't add this extra security so it won't need much resources.
IMO we don't need that extra security but I added it just in case. WDYT?

matks:
> @matks I agree with you but I think it's either adding that extra security but that might require huge resources or don't add this extra security so it won't need much resources.
> IMO we don't need that extra security but I added it just in case. WDYT?

Looks like a choice must be made.

@eternoendless will probably have to choose a side. As for me I side with you 😄I think a smooth upgrade process is more important than handling extra usecases. However we can document (and even explain how to apply it manually) the usecase.

eternoendless:
I can't see why performance would be a worse problem in this change compared to any other database change when there are lots of records.

matks:
> I can't see why performance would be a worse problem in this change compared to any other database change when there are lots of records.

That's true 😅but this one in particular. We are modifying `order_detail` which we can guess has huge number of records, and doing heavy operations such as full copy of a column into another or column renaming.

Progi1984:
Hello @atomiix , thanks for your contribution, would you mind to rebase your pull request? Thanks

install-dev/data/db_structure.sql
PierreRambaud: Are you sure reduction is only a percentage? :trollface: 
FYI https://github.com/PrestaShop/PrestaShop/blob/develop/classes/GroupReduction.php#L41

install-dev/upgrade/sql/1.7.7.0.sql
```diff
@@ -553,6 +553,62 @@ INSERT INTO `PREFIX_hook_alias` (`name`, `alias`) VALUES
...
 
+ALTER TABLE `PREFIX_group_reduction` CHANGE `reduction` `reduction` DECIMAL(5, 4) NOT NULL DEFAULT '0.0000';
+ALTER TABLE `PREFIX_product_group_reduction_cache` CHANGE `reduction` `reduction` DECIMAL(5, 4) NOT NULL DEFAULT '0.0000';
+ALTER TABLE `PREFIX_order_slip` CHANGE `amount` `amount` DECIMAL(20, 6) NOT NULL DEFAULT '0.000000';
+ALTER TABLE `PREFIX_order_slip` CHANGE `shipping_cost_amount` `shipping_cost_amount` DECIMAL(20, 6) NOT NULL DEFAULT '0.000000';
+ALTER TABLE `PREFIX_order_payment` CHANGE `amount` `amount` DECIMAL(20, 6) NOT NULL DEFAULT '0.000000';
+
+/* attribute_impact price */
+ALTER TABLE `PREFIX_attribute_impact` ADD `price_tmp` DECIMAL(20, 6) NOT NULL DEFAULT '0.000000' AFTER `price`;
+UPDATE `PREFIX_attribute_impact` SET `price_tmp` = `price`;
```
PierreRambaud: You can't do it:
```
#1264 - Out of range value for column 'price_tmp' at row 1
```

PierreRambaud (changes requested):
The update script is broken

matks:
Confirmed with @eternoendless we can accept this upgrade script (provided it is fixed after @PierreRambaud feedback)

install-dev/upgrade/sql/1.7.7.0.sql
atomiix: @PierreRambaud I don't see why? I've tried it and didn't have any problem 🤔

PierreRambaud: You need to insert the max value in the original price field.
After you can set `price_tmp` with an higher value than it can received.

You need to do something like
```sql
UPDATE `ps_attribute_impact` SET price_tmp = CASE WHEN LENGTH(price) < 18 
    THEN price
    ELSE RIGHT(price, 17)
END
```
Which will have:
image

You don't need to create a temporary column.
If it doesn't fail for you, you need to check your MySQL / MariaDB version :thinking: 

I'm using 
image

and your script is prompt errors.

PierreRambaud (dismissed):
LGTM, all table's fields needs to be tested properly

Progi1984 (dismissed):

SD1982:
Hello @atomiix the #18532 issue is still present but for #18498 it seems to be gone even without PR

SD1982:
Hi @atomiix All the modified database fields are correct except one 

In the table **ps_group** the field **reduction** is set to decimal(17,2) instead of decimal(5,2)
capture d'écran_82

matks (approved):

Progi1984 (approved):

SD1982:
Hi @atomiix It's ok now all the fields have the correct decimal value . Thanks !!

Progi1984:
Thanks @atomiix

PR #18580, opened 2020-04-10
In BO Product page, use ecotax excl. tax instead of tax incl.

### LINKED ISSUE #10026: [BOOM-1506] In BO Product page, ecotax tax excl is used instead of ecotax tax incl
> This issue has been migrated from the Forge. Read the [original ticket here](http://forge.prestashop.com/browse/BOOM-1506).

- _**Reporter:**_ vincentbz
- _**Created at:**_ Tue, 20 Sep 2016 17:39:49 +0200

1) I enable ecotax in international > VAT
2) I define a tax for my ecotax of 20%
Capture du 2018-03-15 16-05-05

3) I set ecotax = 1€ tax incl
The price tax excl is not well recalculated (issue https://github.com/PrestaShop/PrestaShop/issues/9855)
Sans titre2

4) I save my product, reload the page ==> the ecotax = amount Tax Excl. The VAT on the ecotax has disappear in the nature. And the product price tax incl is also wrong (should be 12€)
Capture du 2018-03-15 16-40-38

### LINKED ISSUE #18550: Ecotax value is reduced each time a product is saved
#### Describe the bug

Ecotax value is reduced each time a product is saved.

1. Be sure to have ecotax enabled and set to a specific tax rule (20% in France, for example)
- go to International > Taxes page and enable ecotax
- then refresh page add tax ON ecotax
Capture d’écran 2020-04-10 à 16 54 31
2. Create a new product and set price without tax to 100, click "save" then refresh the page.
3. Set ecotax value to 10, save, refresh the page. You should see that ecotax amount is 8.3 because 20% tax has been applied.
4. ecotax value is now 8,3333 (=> 10/1,20)
5. do not touch anything, **save again**, refresh the page
6. ecotax value is now 6,944444 (8,333/1,20)
7. and so on....

#### Additional information

* PrestaShop version: 1.7.6 and probably lower versions too
* PHP version: 7.2

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Ecotax is stored with tax included but form uses the tax excluded value. However BO Product page would load the form with tax. incl. value, leading to the crazy scenario described in https://github.com/PrestaShop/PrestaShop/issues/18550. I fix that.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/18550 and number 4 of https://github.com/PrestaShop/PrestaShop/issues/10026
| How to test?  | Please test the scenario described in https://github.com/PrestaShop/PrestaShop/issues/18550 . It should not happen anymore. Please check the ecotax amount is correctly being stored and computed for both "with ecotax tax and "without ecotax tax". By this I mean: you can add tax on ecotax 😄See below screenshot. Everything is explained in https://github.com/PrestaShop/PrestaShop/issues/18550

Screenshot

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18580)

### DIFF
--- a/src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
+++ b/src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
@@ -532,9 +532,15 @@ class AdminModelAdapter extends \PrestaShopBundle\Model\AdminModelAdapter
      */
     private function mapStep2FormData(Product $product)
     {
+        // ecotax is stored with tax included but form uses the tax excluded value
+        $ecotax = $this->tools->round(
+            $product->ecotax * (1 + $this->taxRuleDataProvider->getProductEcotaxRate() / 100),
+            2
+        );
+
         return [
             'price' => $product->price,
-            'ecotax' => $product->ecotax,
+            'ecotax' => $ecotax,
             'id_tax_rules_group' => isset($product->id_tax_rules_group)
                 ? (int) $product->id_tax_rules_group
                 : $this->taxRuleDataProvider->getIdTaxRulesGroupMostUsed(),
--- a/src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
+++ b/src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
@@ -533,9 +533,11 @@ class AdminModelAdapter extends \PrestaShopBundle\Model\AdminModelAdapter
     private function mapStep2FormData(Product $product)
     {
         // ecotax is stored with tax included but form uses the tax excluded value
+        // using a precision of 6 digits as `AdminProductsController::_removeTaxFromEcotax()`
+        // which does the opposite uses 6 digits too
         $ecotax = $this->tools->round(
             $product->ecotax * (1 + $this->taxRuleDataProvider->getProductEcotaxRate() / 100),
-            2
+            6
         );
 
         return [


matks:
This is a `Topwatchers` bug and a `Major` bug fixed by only 2 lines of code. I think it could be backported to `1.7.6.x` branch to provide ecotax working field in 1.7.6

src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
```diff
@@ -532,9 +532,12 @@ function ($p) {
      */
     private function mapStep2FormData(Product $product)
     {
+        // ecotax is stored with tax included but form uses the tax excluded value
+        $ecotax = \Tools::ps_round($product->ecotax * (1 + \Tax::getProductEcotaxRate() / 100), 2);
```
Progi1984: Please declare `Tools` and `Tax`.

Progi1984: But it seems magic !

jolelievre: Nice fix! Once it's merged in 176 I think we should adapt it with `ComputingPrecision` for 177
Same as @Progi1984 As soon as the `use` statement I approve right away 😉

matthieu-rolland (dismissed):

src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
matks: PR updated. I found that in this class adapter classes were available so I used already declared adapter classes instead.

Progi1984 (dismissed):

src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
```diff
@@ -532,9 +532,15 @@ function ($p) {
      */
     private function mapStep2FormData(Product $product)
     {
+        // ecotax is stored with tax included but form uses the tax excluded value
+        $ecotax = $this->tools->round(
+            $product->ecotax * (1 + $this->taxRuleDataProvider->getProductEcotaxRate() / 100),
+            2
```
hadjedjvincent: You should use more decimals here, 6 decimals maybe ? as the database `ecotax` column allow 6 decimals.

PierreRambaud: Could you use a constant instead? or explain why you use 2 in a code comment.

matks: Initially I was copying the behavior of `AdminProductsController::_applyTaxToEcotax()` but I see that indeed I should rather pick the same precision that is being used in the function that computes "tax excl." from "tax. incl." price, which is `AdminProductsController::_removeTaxFromEcotax()`

So, as `AdminProductsController::_removeTaxFromEcotax()` uses 6 precision digits, I will use 6 too.

matks: Initially I was copying the behavior of `AdminProductsController::_applyTaxToEcotax()` but I see that indeed I should rather pick the same precision that is being used in the function that computes "tax excl." from "tax. incl." price, which is `AdminProductsController::_removeTaxFromEcotax()`

So, as `AdminProductsController::_removeTaxFromEcotax()` uses 6 precision digits, I will use 6 too.

matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/18580/commits/f3b1d0e650165f0b4b61959deda23a999c346c21

matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/18580/commits/f3b1d0e650165f0b4b61959deda23a999c346c21

Progi1984 (approved):

src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
```diff
@@ -532,9 +532,17 @@ function ($p) {
      */
     private function mapStep2FormData(Product $product)
     {
+        // ecotax is stored with tax included but form uses the tax excluded value
+        // using a precision of 6 digits as `AdminProductsController::_removeTaxFromEcotax()`
+        // which does the opposite uses 6 digits too
+        $ecotax = $this->tools->round(
+            $product->ecotax * (1 + $this->taxRuleDataProvider->getProductEcotaxRate() / 100),
+            6
```
sowbiba, line 540: can we put it in a constant ?

sowbiba (approved):

src/PrestaShopBundle/Model/Product/AdminModelAdapter.php
matks: We could, but it would only be useful if all the places using this value use the constant as well.
I know https://github.com/PrestaShop/PrestaShop/blob/develop/controllers/admin/AdminProductsController.php#L2209 uses it as this function reverts the logic there, and there might be more.

Could it be considered an improvement and I'll do it in another PR ?
I'm afraid of changing other parts of the code and introducing crazy bugs 😅which would then be reported by QA and that would block this PR

sowbiba: Ok with that

SD1982:
LGTM Thanks @matks !

Progi1984:
Thanks @matks

PR #18617, opened 2020-04-14
Make computing precision matching currency's display precision

### LINKED ISSUE #18513: Wrong total value of cart
#### Describe the bug

Total price of the cart is wrongly calculated on demo products.

#### Expected behavior

Well... cart should have a valid total value. This is critical.

#### Steps to Reproduce

1. Install PrestaShop 1.7.7 (1.7.7 branch) in Polish.
2. You'll have few products with PLN currency, one of them, product ID 2, has price equal to: 35.33 PLN
3. Now add two units of this product to cart.
4. Total price of the cart should be 35.33 * 2 => 70.66, but it's 70.65

It was working fine on previous versions with changed precision calculation to `Round on each item` but it's not a case in 1.7.7.

btw. this scenario has been around for years, after 20% discount we have:
35.328

So it's tricky to recalculate final price **but** it was possible by changing type of price calculation, this option from Preferences -> General is not working now... unfortunately. 

**Screenshots**

image

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: 7.2

### LINKED ISSUE #18640: wrong order total in FO
After doing the rollback on computing precision (with this pr: #18617), a bug was revealed in the cart page in FO:

Capture d’écran 2020-04-15 à 15 00 25

The cart order total is wrong, here is how to reproduce this behavior:

### How to reproduce

0- Note that this PR (#18617) should be merged before you can reproduce this issue

1- In BO make a product with price TTC 44.157 and TVA 20%. (you can also change the price of an existing product)

2- Add a 20% discount on the product

3- In fo, order this product with a quantity of 2 (2 items)

4- go to cart and see total price

**PS**: if you have demo products installed, just use the **Hummingbird printed t-shirt** and change its price to TTC 44.157, the 20% reduction should already be applied to this product. Be sure it has a 20% tax.

### Additional information

* PrestaShop version: 1.7.7
* PHP version: All compatible versions

### LINKED ISSUE #18717: Make computing precision matching currency's display precision
This issue aims at rollbacking the previous computing precision to match the display precision.

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Make computing precision matching currency's display precision 
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #18717 
| How to test?  | This PR changes prices everywhere... it should probably not be tested by itself, but it also fixes a bug, and this can be tested replicating the septs given here: #18640

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18617)

### DIFF
--- a/src/Core/Cart/Calculator.php
+++ b/src/Core/Cart/Calculator.php
@@ -186,7 +186,11 @@ class Calculator
     {
         $amount = new AmountImmutable();
         foreach ($this->cartRows as $cartRow) {
-            $amount = $amount->add($cartRow->getInitialTotalPrice());
+            if ($cartRow->getRoundType() !== $cartRow::ROUND_MODE_ITEM) {
+                $amount = $amount->add($cartRow->getInitialTotalPrice());
+            } else {
+                $amount = $amount->add($cartRow->getFinalTotalPrice());
+            }
         }
 
         return $amount;
--- a/src/Core/Cart/Calculator.php
+++ b/src/Core/Cart/Calculator.php
@@ -186,11 +186,7 @@ class Calculator
     {
         $amount = new AmountImmutable();
         foreach ($this->cartRows as $cartRow) {
-            if ($cartRow->getRoundType() !== $cartRow::ROUND_MODE_ITEM) {
-                $amount = $amount->add($cartRow->getInitialTotalPrice());
-            } else {
-                $amount = $amount->add($cartRow->getFinalTotalPrice());
-            }
+            $amount = $amount->add($cartRow->getInitialTotalPrice());
         }
 
         return $amount;
--- a/tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/Currency/currency.feature
+++ b/tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/Currency/currency.feature
@@ -91,7 +91,7 @@ Feature: Cart calculation with currencies
     Given there is a product in the catalog named "product1" with a price of 19.812 and 1000 items in stock
     When I add 3 items of product "product1" in my cart
     Then my cart total should be 83.1 tax included
-    Then my cart total using previous calculation method should be 83.0 tax included
+    Then my cart total using previous calculation method should be 83.1 tax included
 
   Scenario: 3 products in cart, several quantities
     Given I have an empty default cart
--- a/tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/Currency/currency.feature
+++ b/tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/Currency/currency.feature
@@ -90,7 +90,7 @@ Feature: Cart calculation with currencies
     Given currency "currency2" is the current one
     Given there is a product in the catalog named "product1" with a price of 19.812 and 1000 items in stock
     When I add 3 items of product "product1" in my cart
-    Then my cart total should be 83.0 tax included
+    Then my cart total should be 83.1 tax included
     Then my cart total using previous calculation method should be 83.0 tax included
 
   Scenario: 3 products in cart, several quantities
--- a/tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/RoundingType/round_item.feature
+++ b/tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/RoundingType/round_item.feature
@@ -37,10 +37,28 @@ Feature: Cart calculation with rounding type ITEM
     Then my cart total should be precisely 162.4 tax included
     Then my cart total using previous calculation method should be precisely 162.4 tax included
 
-    Scenario: 1 product in cart, with a quantity of 2 and a tricky price
-      Given I have an empty default cart
-      Given specific shop configuration for "rounding type" is set to round each article
-      Given there is a product in the catalog named "product1" with a price of 35.32552 and 1000 items in stock
-      When I add 2 items of product "product1" in my cart
-      Then my cart total should be precisely 77.66 tax included
-      Then my cart total using previous calculation method should be precisely 77.66 tax included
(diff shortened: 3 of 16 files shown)


PierreRambaud (dismissed):

PierreRambaud (dismissed):

tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/Currency/currency.feature
```diff
@@ -90,7 +90,7 @@ Feature: Cart calculation with currencies
     Given currency "currency2" is the current one
     Given there is a product in the catalog named "product1" with a price of 19.812 and 1000 items in stock
     When I add 3 items of product "product1" in my cart
-    Then my cart total should be 83.0 tax included
+    Then my cart total should be 83.1 tax included
```
jolelievre, line 93: Dont' think there's a problem here? 😅
83.1 instead of 83.0 with the previous calculation method
I know it's supposed to be fixed in another PR but still merging invalid tests seems a bit risky

tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/RoundingType/round_total.feature
```diff
@@ -14,15 +14,15 @@ Feature: Cart calculation with rounding type TOTAL
...
+    Then my cart total should be precisely 26.81 tax included
     Then my cart total using previous calculation method should be precisely 26.812 tax included
 
   Scenario: one product in cart, quantity 3
     Given I have an empty default cart
     Given specific shop configuration for "rounding type" is set to round cart total
     Given there is a product in the catalog named "product1" with a price of 19.812 and 1000 items in stock
     When I add 3 items of product "product1" in my cart
-    Then my cart total should be precisely 66.436 tax included
+    Then my cart total should be precisely 66.44 tax included
```
jolelievre, line 25: Also, the legacy method returns a more precise value, because it uses
`$compute_precision            = $this->configuration->get('_PS_PRICE_COMPUTE_PRECISION_');` as a precision
which is defined as 6 in the `config.inc.php` So I'm actually not sure using the same precision for display and computing is actually a good idea, it's not a revert it's drifting even further from the previous behaviour
We should return 6 to stay close to the initial behaviour

tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/Currency/currency.feature
matthieu-rolland: yep, that's why I would like this PR not to be merged too soon, we should wait for the FIX of this issue first

tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/RoundingType/round_total.feature
matthieu-rolland: if we force the computing precision to 6 we won't have a matching display and computing precision, so we'll have the same issues we're trying to fix in the first place...  or I didn't understand what you mean?

matthieu-rolland: Also the `_PS_PRICE_COMPUTE_PRECISION_` has been harcoded to 6 in 1.7.7, to match the results obtained in our behat tests... maybe we could do the same, and setting it to 2 ?

jolelievre (dismissed):

Progi1984 (dismissed):

tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/RoundingType/round_item.feature
```diff
@@ -36,3 +36,11 @@ Feature: Cart calculation with rounding type ITEM
     When I add 1 items of product "product3" in my cart
     Then my cart total should be precisely 162.4 tax included
     Then my cart total using previous calculation method should be precisely 162.4 tax included
+
+    Scenario: 1 product in quart, with a quantity of 2 and a tricky price
```
Progi1984:
```suggestion
    Scenario: 1 product in cart, with a quantity of 2 and a tricky price
```

matthieu-rolland: done !

Progi1984 (dismissed):

src/Core/Cart/Calculator.php
```diff
@@ -186,7 +186,11 @@ public function getRowTotalWithoutDiscount()
     {
         $amount = new AmountImmutable();
         foreach ($this->cartRows as $cartRow) {
-            $amount = $amount->add($cartRow->getInitialTotalPrice());
+            if ($cartRow->getRoundType() !== $cartRow::ROUND_MODE_ITEM) {
```
jolelievre: Is there a test (behat or not) to validate this behaviour and prevent future regression?
This can be done in another issue/PR I think, but it's important nonetheless

jolelievre (dismissed):
Thanks @matthieu-rolland 
I think we should create an issue to add test about this very specific fix (the total bug)

src/Core/Cart/Calculator.php
matthieu-rolland: @jolelievre 

Well my fix only applies to the round mode item, for which I added a scenario replicating the behavior giving us the bug.

In other cases the behavior remains ISO, and is already tested in the behat tests dedicated to rounding types.

Progi1984 (dismissed):

jolelievre (approved):
Nice job @matthieu-rolland

atomiix (approved):

Progi1984 (approved):

PR #18633, opened 2020-04-15
Convert cart rule value when order currency is different

### LINKED ISSUE #18630: BO - Order page - the Value discount is incorrect
#### Describe the bug

In the BO => Order page, the value discount is not correct.

#### Steps to Reproduce

You should have two currencies 
1. € currency by default
2. $ second currency

1. Create a cart rule without code with an amount reduction 10$ tax included
2. Go to FO, select € as a currency
3. Add the product `Hummingbird printed t-shirt Size: S Color: White` to the cart
4. The cart rule is applied
5. process to checkout
6. Go to the BO => Orders page => View last order created
7. See error

1.7.7.7x
image
1.7.6.4
image

1.7.6.0
image

1.7.5.2
image
https://drive.google.com/file/d/1sF-uPWHOBRUs4Vw97ZQL9tK5RFAvFYBQ/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.6.0,1.7.6.4, 1.7.7.x
* PHP version: 7.2

sowbiba:
| Questions     | Answers
| Branch?       | 1.7.6.x
| Description?  | -
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18630 .
| How to test?  | See #18630 description.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18633)

### DIFF
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -1085,9 +1085,16 @@ abstract class PaymentModuleCore extends Module
             //  This is an "amount" reduction, not a reduction in % or a gift
             // THEN
             //  The voucher is cloned with a new value corresponding to the remainder
-            $remainingValue = $cartRule->reduction_amount - $values[$cartRule->reduction_tax ? 'tax_incl' : 'tax_excl'];
+            $cartRuleReductionAmountConverted = $cartRule->reduction_amount;
+            if ((int) $cartRule->reduction_currency !== $cart->id_currency) {
+                $cartRuleReductionAmountConverted = Tools::convertPrice(
+                    $cartRule->reduction_amount,
+                    new \Currency($cart->id_currency)
+                );
+            }
+            $remainingValue = $cartRuleReductionAmountConverted - $values[$cartRule->reduction_tax ? 'tax_incl' : 'tax_excl'];
             $remainingValue = Tools::ps_round($remainingValue, _PS_PRICE_COMPUTE_PRECISION_);
-            if (count($order_list) == 1 && $remainingValue > 0 && $cartRule->partial_use == 1 && $cartRule->reduction_amount > 0) {
+            if (count($order_list) == 1 && $remainingValue > 0 && $cartRule->partial_use == 1 && $cartRuleReductionAmountConverted > 0) {
                 // Create a new voucher from the original
                 $voucher = new CartRule((int) $cartRule->id); // We need to instantiate the CartRule without lang parameter to allow saving it
                 unset($voucher->id);
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -1089,8 +1089,8 @@ abstract class PaymentModuleCore extends Module
             if ((int) $cartRule->reduction_currency !== $cart->id_currency) {
                 $cartRuleReductionAmountConverted = Tools::convertPriceFull(
                     $cartRule->reduction_amount,
-                    new \Currency((int) $cartRule->reduction_currency),
-                    new \Currency($cart->id_currency)
+                    new Currency((int) $cartRule->reduction_currency),
+                    new Currency($cart->id_currency)
                 );
             }
             $remainingValue = $cartRuleReductionAmountConverted - $values[$cartRule->reduction_tax ? 'tax_incl' : 'tax_excl'];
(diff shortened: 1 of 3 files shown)


classes/PaymentModule.php
```diff
@@ -1085,9 +1085,16 @@ protected function createOrderCartRules(
             //  This is an "amount" reduction, not a reduction in % or a gift
             // THEN
             //  The voucher is cloned with a new value corresponding to the remainder
-            $remainingValue = $cartRule->reduction_amount - $values[$cartRule->reduction_tax ? 'tax_incl' : 'tax_excl'];
+            $cartRuleReductionAmountConverted = $cartRule->reduction_amount;
+            if ((int) $cartRule->reduction_currency !== $cart->id_currency) {
```
jolelievre, line 1089: I don't think it's correct as It's the same as a bug you fixed recently The cart currency and/or the cart rule currency can be different from the default one So you need to convert the cart rule price to the default currency, then to the cart currency

Actually as this is not the first time we encounter this kind of behaviour I think we should add a method `Tools::convertPriceToCurrency($price, Currency $fromCurrency, Currency $toCurrency)` (potentially allow `$toCurrency` to be null with fallback to default)
The funny thing is that there is a deprecated note about this `convertPriceToCurrency` method since 1.7.4 but I can't find it anywhere

jolelievre (changes requested):

classes/PaymentModule.php
sowbiba: There is a `Tools::convertPriceFull($amount, Currency $currency_from = null, Currency $currency_to = null)` I will use that one ;)

sowbiba: done

jolelievre: Nice!!
Can you add a `use` statement please? To avoid the `\Currency` notation

travis-scripts/setup-php-fpm.sh
```diff
@@ -19,6 +19,8 @@ echo "cgi.fix_pathinfo = 1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/con
 echo "always_populate_raw_post_data = -1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 echo "error_log = /var/log/php-fpm.log" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 echo "memory_limit = 2G" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+echo "opcache.revalidate_freq=7000" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
```
Progi1984: @sowbiba Why do you change these values ?

sowbiba: I had an error from travis
https://stackoverflow.com/questions/50769286/zend-mm-heap-corrupted-with-standard-php-5-6-locating-the-cause
https://bugs.php.net/bug.php?id=69540

Progi1984: Relative only for PHP 5.6 ?

sowbiba: yes only for 5.6

PierreRambaud: Don't need to add it. Rerun travis, and everything will be ok (sometimes :sweat_smile: )

Progi1984: May be these settings should be conditional to PHP 5.6 ?

PierreRambaud: Since we remove the 5.6 support in 1.7.7.0 and only related to tests, we should maybe ignore it wdyt?

Progi1984: Rerun, I have passed my last PR like it.

Progi1984 (approved):

jolelievre (approved):

khouloudbelguith:
Hi @sowbiba,

It is OK for me :heavy_check_mark: 
image

Ping @SD1982 

Thanks!

SD1982:
Look good to me to ! Thanks @sowbiba !

PR #18666, opened 2020-04-16
Outputting errors for not generated hooks in the console

### LINKED ISSUE #18619: Not all hooks are generated with hooks generation commands
#### Describe the bug

Some hooks are not generated with commands:

- prestashop:update:configuration-file-hooks-listing 

- prestashop:update:sql-upgrade-file-hooks-listing

Instead the errors are catched with PrestaShopLogger and logged.

#### Expected behavior

All hooks generated as expected. If no easy fix possible notify about it the user friendly way.

#### Steps to Reproduce

1. Run Symfony command from terminal: prestashop:update:configuration-file-hooks-listing
2. See the list of hooks added to the configuration file
3. See Logger::addLog logged errors
4. Same for prestashop:update:sql-upgrade-file-hooks-listing

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: 7.2

tdavidsonas88:
| Questions     | Answers
| Branch?       | develop
| Description?  | Outputting errors to the console for non generated hooks
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Partially Fixes #18619 .
| How to test?  | Follow steps in the issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18666)

### DIFF
--- a/src/Core/Hook/Provider/GridDefinitionHookByServiceIdsProvider.php
+++ b/src/Core/Hook/Provider/GridDefinitionHookByServiceIdsProvider.php
@@ -30,6 +30,7 @@ use Exception;
 use Generator;
 use Logger;
 use PrestaShop\PrestaShop\Core\Grid\Definition\Factory\GridDefinitionFactoryInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
@@ -61,10 +62,10 @@ final class GridDefinitionHookByServiceIdsProvider implements HookByServiceIdsPr
     /**
      * {@inheritdoc}
      */
-    public function getHookNames(array $gridDefinitionServiceIds)
+    public function getHookNames(array $gridDefinitionServiceIds, SymfonyStyle $io)
     {
         /** @var Generator $gridDefinitionIds */
-        $gridDefinitionIds = $this->getGridDefinitionIds($gridDefinitionServiceIds);
+        $gridDefinitionIds = $this->getGridDefinitionIds($gridDefinitionServiceIds, $io);
 
         $gridDefinitionHookNames = [];
         $gridQueryBuilderHookNames = [];
@@ -120,7 +121,7 @@ final class GridDefinitionHookByServiceIdsProvider implements HookByServiceIdsPr
      *
      * @return Generator
      */
-    private function getGridDefinitionIds(array $gridDefinitionServiceIds)
+    private function getGridDefinitionIds(array $gridDefinitionServiceIds, SymfonyStyle $io)
     {
         foreach ($gridDefinitionServiceIds as $serviceId) {
             try {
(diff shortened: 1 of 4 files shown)


Progi1984 (dismissed):

matks (dismissed):

src/Core/Hook/Provider/HookByServiceIdsProviderInterface.php
```diff
@@ -36,5 +38,5 @@ interface HookByServiceIdsProviderInterface
      *
      * @return string[]
      */
-    public function getHookNames(array $gridDefinitionServiceIds);
+    public function getHookNames(array $gridDefinitionServiceIds, SymfonyStyle $io);
```
jolelievre, line 41: You are twisting the interface purpose here, and mixing responsibilies
The interface responsibility is just to return a list of hook names, the display or logging of errors has nothing to do with and shouldn't be passed as a parameter to the function

src/Core/Hook/Provider/GridDefinitionHookByServiceIdsProvider.php
```diff
@@ -61,10 +61,10 @@ public function __construct(ContainerInterface $container)
     /**
      * {@inheritdoc}
      */
-    public function getHookNames(array $gridDefinitionServiceIds)
+    public function getHookNames(array $gridDefinitionServiceIds, SymfonyStyle $io)
```
jolelievre, line 64: I'm not sure it's the right approach to couple this service to `SymfonyStyle`
Especially through an argument, at  worst it should be injected in the constructor rather than provided via a parameter
This would avoid breaking/twisting the interface as well (see other comment)

And ideally it would be better to inject a `LoggerInterface` which is more generic, you even use a `NullLogger` and keep this class a bit more independent

jolelievre (changes requested):
Provide the object for rendering via a parameter is not a good idea

It's like transferring a twig engine along multiple function calls, it couples the classes and interfaces to things they are not supposed to deal with

src/Core/Hook/Provider/GridDefinitionHookByServiceIdsProvider.php
PierreRambaud: Agree, don't use SymfonyStyle in services :+1:

src/Core/Hook/Provider/GridDefinitionHookByServiceIdsProvider.php
```diff
@@ -133,7 +133,9 @@ private function getGridDefinitionIds(array $gridDefinitionServiceIds)
                 $camelizedDefinitionId = Container::camelize($definitionId);
                 yield $camelizedDefinitionId;
             } catch (Exception $e) {
-                Logger::addLog(sprintf('Error while loading service: %s . Error: %s', $serviceId, $e));
```
mickaelandrieu, line 136: why not throw the exception again if php is [in cli mode](https://www.php.net/manual/fr/function.php-sapi-name.php) is detected so we can catch the exception later in the command and display it the right way ?

PierreRambaud:
Hi @tdavidsonas88  ,

Thanks!

Issue #14763, opened 2019-07-19
[EPIC] Issue Product page - Basic settings tab 2/9


marionf:
## Issues in **basic settings tab** sorted by priority high to low :

**Must-Have**
- [x] #9945
- [x] #12773 (#9631)
- [x] #30219

**Should-Have**
- [x] #9752
- [x] #9625
- [x] #9626
- [x] #11587
- [x] #9617

**Improvement**
- [x] #20331
- [x] #16306

**Nice to Have**
- [x] #15383
- [x] #9631
- [x] #10459
- [x] #16004 
- [x] #9960
- [x] #13751
- [x] #17822 

**New feature**
- [x] #18823 

**New issues to prioritize** 
- [x] #18636
- [x] #19022
- [x] #25582
- [ ] #26566

## Fixed
- [x] #26519
- [x] #23255
- [x] #9678
- [x] #16308
- [x] #13513
- [x] #17957
- [x] #9668
- [x] #14955
- [x] #16415
- [x] #20035
- [x] #14052
- [x] #15018 (PR: #15035)
- [x] #12061
- [x] #12832
- [x] #14985 (#9668)
- [x] #13698 - #19281
- [x] #22144
- [x] #19524

marionf:
### - Issues in **basic settings tab** sorted by priority high to low :

**Must-Have**
- [ ] #9945
- [x] #12773 - #9631
- [x] #14985  - #9668
- [ ] #16306
- [ ] #13698

**Should-Have**
- [ ] #9752
- [ ] #9625
- [x] #12832 - Multistore
- [ ] #9626
- [ ] #11587
- [ ] #9617
- [x] #15018 PR: https://github.com/PrestaShop/PrestaShop/pull/15035
- [x] #12061

**Improvement**
- [x] #14052
- [ ] #20331
- [x] #14955
- [x] #16415

**Nice to Have**
- [x] #9668
- [ ] #15383 - Multistore
- [x] #17957
- [x] #9631
- [x] #13513
- [x] #10459
- [x] #16004
- [ ] #9960
- [x] #20027
- [x] #9678
- [x] #16308

**New feature**
- [x] #12773
- [x] https://github.com/PrestaShop/PrestaShop/issues/18823

sarjon:
Question https://github.com/PrestaShop/PrestaShop/issues/10632#issuecomment-515981529

MatShir:
[...]

PR #18672, opened 2020-04-16
AddProductCommand

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | AddProductCommand is responsible for adding most basic version of product and it is supposed to be the only command to actually create the product (all upcoming commands will build the product by updating it).
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #14763
| How to test?  | CI pass is enough. No UI to test yet.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18672)


src/Adapter/Product/CommandHandler/AddBasicProductHandler.php
```diff
@@ -0,0 +1,117 @@
...
+     * {@inheritDoc}
+     */
+    public function handle(AddBasicProductCommand $command): ProductId
+    {
+        $product = $this->createProduct($command);
+        $this->addOptionalProperties($product, $command);
+
+        try {
+            //@todo: Check old ProductController::517 for hooks
+            //  will those hooks in new form handler be enough to replace old ones?
```
matks: If we can keep old hooks, we keep them. We try to be retro-compatible.
However keeping them must not make our code unreadable.

So strategy for old hooks is:
- see if we can "just put it back" with very few code
- if it cannot be done easily, remove it (and we must start somewhere a list of all deprecated hooks)

src/Core/Domain/Product/ValueObject/ProductType.php
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * A pack consists multiple units of product.
+     */
+    const TYPE_PACK = 1;
+
+    /**
+     * Items that are not in physical form and can be sold without requiring any shipping
+     * E.g. downloadable photos, videos, software, services etc.
+     */
+    const TYPE_VIRTUAL = 2;
```
matks, line 52: What about "product with combinations" ? Isn't it a type too ?

src/Adapter/Product/CommandHandler/AddBasicProductHandler.php
```diff
@@ -0,0 +1,117 @@
...
+     *
+     * @return Product
+     */
+    private function createProduct(AddBasicProductCommand $command): Product
+    {
+        //@todo: dont forget multishop when specs are prepared.
+        $product = new Product();
+        $product->name = $command->getLocalizedNames();
+        //@todo: check if there is anything more for product type
+        $product->is_virtual = $command->getType() === ProductType::TYPE_VIRTUAL;
```
jolelievre: Isn't it `$command->getType()->getValue()` ?

src/Core/Domain/Product/ValueObject/ProductType.php
zuk3975: https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Product.php#L549

I think only these values are considered in legacy

src/Adapter/Product/CommandHandler/AddProductHandler.php
```diff
@@ -0,0 +1,117 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\AddProductHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductCondition;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductType;
+use PrestaShopException;
+use Product;
+
+final class AddProductHandler implements AddProductHandlerInterface
```
jolelievre, line 44: Please add PHPDoc on all classes and interfaces missing it

src/Adapter/Product/CommandHandler/AddProductHandler.php
```diff
@@ -0,0 +1,117 @@
...
+     *
+     * @return Product
+     */
+    private function createProduct(AddProductCommand $command): Product
+    {
+        //@todo: multistore?
+        $product = new Product();
+        $product->name = $command->getLocalizedNames();
+        $product->active = false;
+        $product->category = $this->defaultCategoryId;
```
jolelievre: This field is supposed to contain the category rewrite-link according to Product class https://github.com/PrestaShop/PrestaShop/blob/e64bc462d3bb275d0ca28efdc9cffd169568af0f/classes/Product.php#L674

Although I'm not sure it's useful for creation as it won't be persisted anyway

src/Adapter/Product/CommandHandler/AddProductHandler.php
```diff
@@ -0,0 +1,117 @@
...
+        $product->id_category_default = $this->defaultCategoryId;
+
+        if ($command->getType() === ProductType::TYPE_VIRTUAL) {
+            $product->is_virtual = true;
+            $product->condition = ProductCondition::NEW;
```
jolelievre: Not sure it's needed since the default value is already set as `new` in the Product class

src/Core/Domain/Product/ValueObject/ProductType.php
jolelievre: Yeah it's a bit weird, "product with combinations" is detected by checking the existence of combinations I believe, but so is the Pack So I don't know why there's a dedicated type for pack but not for combinations

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -55,6 +63,158 @@ public function setProductTaxRulesGroup(string $taxRulesGroupReference, string $
...
+    /**
+     * @When I add product :productReference with following basic information in default language:
+     *
+     * @param string $productReference
+     * @param TableNode $table
+     */
+    public function addProduct(string $productReference, TableNode $table): void
+    {
+        $data = $table->getRowsHash();
+        $defaultLangId = (int) Configuration::get('PS_LANG_DEFAULT');
```
jolelievre: I think instead of only using default language you could use https://github.com/PrestaShop/PrestaShop/blob/95c1d2c7537d71599fa373462bf885b0ebbcaa18/tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php#L136

like in

https://github.com/PrestaShop/PrestaShop/blob/95c1d2c7537d71599fa373462bf885b0ebbcaa18/tests/Integration/Behaviour/Features/Context/Domain/CurrencyFeatureContext.php#L159

It is also usable to test returned localized data like in:

https://github.com/PrestaShop/PrestaShop/blob/95c1d2c7537d71599fa373462bf885b0ebbcaa18/tests/Integration/Behaviour/Features/Context/Domain/CurrencyFeatureContext.php#L231

This would allow you to add data like

```
| name  | en-US:bottle of beer |
```

It adds two advantages:
- you can check that localized values are correctly inserted
- you can test an error scenario when the default language name is not provided

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -55,6 +63,158 @@ public function setProductTaxRulesGroup(string $taxRulesGroupReference, string $
...
+     *
+     * @param string $productReference
+     */
+    public function assertProductAssignedToDefaultCategory(string $productReference)
+    {
+        $defaultCategoryId = (int) Configuration::get('PS_HOME_CATEGORY');
+        $product = $this->getProductByReference($productReference);
+        $productCategories = $product->getCategories();
+
+        foreach ($productCategories as $categoryId) {
```
jolelievre, line 162: This loop validates that the first category is the default one, but if it's the second one it will fail You can use a boolean instead that sotres if the default category was found, and throw the exception when it remains false

tests/Integration/Behaviour/Features/Scenario/Product/add_product.feature
```diff
@@ -0,0 +1,17 @@
...
+  Scenario: I add a product with basic information
+    When I add product "product1" with following basic information in default language:
+      | name | bottle of beer |
+      | type | standard       |
+    Then product "product1" should have following values:
+      | active           | false          |
+      | condition        | new            |
+    And product "product1" type should be standard
+    And product "product1" name should be "bottle of beer" in default language
+    And product "product1" should be assigned to default category
```
jolelievre, line 17: As suggested above, you should add a scenario to validate the the missing default language is not allowed

src/Adapter/Product/CommandHandler/AddProductHandler.php
```diff
@@ -0,0 +1,117 @@
...
+        } else {
+            $product->is_virtual = false;
+        }
+
```
jolelievre, line 102: I think we also need to check that name is in the default language

tests/Integration/Behaviour/Features/Scenario/Product/add_product.feature
jolelievre: And maybe one with an invalid type

tests/Integration/Behaviour/Features/Scenario/Product/add_product.feature
```diff
@@ -0,0 +1,17 @@
...
+
+  @add
+  Scenario: I add a product with basic information
+    When I add product "product1" with following basic information in default language:
+      | name | bottle of beer |
+      | type | standard       |
+    Then product "product1" should have following values:
+      | active           | false          |
+      | condition        | new            |
+    And product "product1" type should be standard
```
jolelievre, line 15: Maybe these two checks could be done along with the previous one, with everything in a table, and only one function needed to check all product data

src/Adapter/Product/CommandHandler/AddProductHandler.php
matks: Do we need to ? It's the beginning of the project, maybe we can do this later ? The code might change a lot in the meantime

src/Adapter/Product/CommandHandler/AddProductHandler.php
matks: Better duplicate it and be sure than rely on legacy code maybe ? 😁

src/Core/Domain/Product/ValueObject/ProductType.php
matks: Inconsistency of legacy code.

If it's not too complicated, I suggest we handle correctly it in our modern code (so I suggest creating TYPE_WITH_COMBINATIONS) and let the handler do the translation between modern code and legacy code.

Advantage: easier and cleaner if you work in the controller layer, closer to domain and user knowledge
Disadvantage: having modern and legacy code use different domain model / structure might create confusion and bugs

matks (dismissed):
I think it's OK.
Only issue I have is about TYPE_COMBINATIONS, should we have one ? Since topic is open we might merge this PR and deal with it later

src/Adapter/Product/CommandHandler/AddProductHandler.php
zuk3975: Its so frustrating to write comments that doesn't add any value :cry: 
I suppose I should write better comments then :sweat:

src/Adapter/Product/CommandHandler/AddProductHandler.php
zuk3975: you are right, not worth duplicating.

tests/Integration/Behaviour/Features/Scenario/Product/add_product.feature
zuk3975: kinda, but it at least type depends from couple product properties its not just "product->type". So  it would still contain a dedicated private method with some statement to identify if its tis field. e.g.
```
if ($field === 'type') {
    $this->assertTypeField();
}
```
I t seemed overhead to try merge it in this case.
So i thought we can make one merged method for  simple properties where we can check simply "if (field === $product->field)", and custom ones for more complicated props.

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
zuk3975: lol really :sweat_smile: :+1:

src/Adapter/Product/CommandHandler/AddProductHandler.php
jolelievre: On the review Pablo made he asked for it, so I guess yes we have to 😅

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -180,13 +180,16 @@ public function assertProductType(string $productReference, string $productTypeN
         $isVirtual = $product->is_virtual;
         $isPack = Pack::isPack($product->id);
         $isStandard = !Pack::isPack($product->id) && !$product->is_virtual;
+        $isCombination = $product->getAttributeCombinations() ? true : false;
```
PierreRambaud:
```suggestion
        $isCombination = !empty($product->getAttributeCombinations());
```
:thinking:

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -180,13 +180,16 @@ public function assertProductType(string $productReference, string $productTypeN
...
 
         if ($isVirtual && $productTypeValue === ProductType::TYPE_VIRTUAL) {
             return;
         } elseif ($isPack && $productTypeValue === ProductType::TYPE_PACK) {
             return;
         } elseif ($isStandard && $productTypeValue === ProductType::TYPE_STANDARD) {
             return;
+        } elseif ($isCombination && $productTypeValue === ProductType::TYPE_COMBINATION) {
+            return;
         }
```
PierreRambaud:
```suggestion
        if ($isVirtual && $productTypeValue === ProductType::TYPE_VIRTUAL) {
            return;
        }

        if ($isPack && $productTypeValue === ProductType::TYPE_PACK) {
            return;
        }

        if ($isStandard && $productTypeValue === ProductType::TYPE_STANDARD) {
            return;
        }

        if ($isCombination && $productTypeValue === ProductType::TYPE_COMBINATION) {
            return;
        }
```

matks (dismissed):

src/Core/Domain/Product/ValueObject/ProductCondition.php
```diff
@@ -0,0 +1,83 @@
...
+        ];
+
+        if (!in_array($value, $allowedConditions, true)) {
+            throw new ProductConstraintException(
+                sprintf(
+                    'Invalid product condition "%s". Allowed conditions are: "%s"',
+                    $value,
+                    implode(',', $allowedConditions)
+                ),
+            ProductConstraintException::INVALID_CONDITION
```
PierreRambaud:
```suggestion
                ProductConstraintException::INVALID_CONDITION
```

PierreRambaud: Weird that cs fixer is not crying :thinking:

src/Core/Domain/Product/ValueObject/ProductType.php
```diff
@@ -0,0 +1,105 @@
...
+     */
+    private function assertProductType(int $value): void
+    {
+        $allowedTypes = [
+            self::TYPE_STANDARD,
+            self::TYPE_PACK,
+            self::TYPE_VIRTUAL,
+            self::TYPE_COMBINATION,
+        ];
+
```
PierreRambaud: Should be a const instead

src/PrestaShopBundle/Resources/config/services/adapter/product.yml
```diff
@@ -65,3 +65,12 @@ services:
         tags:
             - name: tactician.handler
               command: PrestaShop\PrestaShop\Core\Domain\Product\Query\GetProductIsEnabled
+
+    prestashop.adapter.product.command_handler.add_product_handler:
+      class: PrestaShop\PrestaShop\Adapter\Product\CommandHandler\AddProductHandler
+      arguments:
+        - "@=service('prestashop.adapter.legacy.configuration').getInt('PS_LANG_DEFAULT')"
+        - '@=service("prestashop.adapter.legacy.context").getContext().shop.id_category'
```
PierreRambaud, line 73: What if there is no id_category in the shop context? :thinking:

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -55,6 +62,191 @@ public function setProductTaxRulesGroup(string $taxRulesGroupReference, string $
...
+     * @param string $localizedValues
+     */
+    public function assertLocalizedProperty(string $productReference, string $fieldName, string $localizedValues)
+    {
+        $product = $this->getProductByReference($productReference);
+        $expectedLocalizedValues = $this->parseLocalizedArray($localizedValues);
+
+        foreach ($expectedLocalizedValues as $langId => $value) {
+            if ($value !== $product->{$fieldName}[$langId]) {
+                $langIso = \Language::getIsoById($langId);
```
PierreRambaud:
```suggestion
                $langIso = Language::getIsoById($langId);
```

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -55,6 +62,191 @@ public function setProductTaxRulesGroup(string $taxRulesGroupReference, string $
...
+     */
+    private function getProductTypeValueByName(string $typeName): int
+    {
+        $typeValueByName = [
+            'standard' => ProductType::TYPE_STANDARD,
+            'pack' => ProductType::TYPE_PACK,
+            'virtual' => ProductType::TYPE_VIRTUAL,
+            'combination' => ProductType::TYPE_COMBINATION,
+            // undefined value to check the real behavior when type is wrong
+            'undefined' => 500,
```
jolelievre: It's weird to introduce a specific "undefined" use case, I understand why you need it to check invalid values though
I'd suggest only managing valid values, and for other you can change the return the provided string as is:

`return $typeValueByName[$typeName] ?? $typeName` or `return $typeValueByName[$typeName] ?? -1` to match the expected `int` type

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -55,6 +62,191 @@ public function setProductTaxRulesGroup(string $taxRulesGroupReference, string $
...
+    }
+
+    /**
+     * @param string $typeName
+     *
+     * @return int
+     */
+    private function getProductTypeValueByName(string $typeName): int
+    {
+        $typeValueByName = [
```
PierreRambaud, line 239: Should be a const as it never changes. Maybe in ProductType class

src/Core/Domain/Product/Command/AddProductCommand.php
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * @param array $localizedNames
+     * @param int $type
+     */
+    public function __construct(
+        array $localizedNames,
+        int $type
+    ) {
+        $this->localizedNames = $localizedNames;
```
jolelievre, line 56: I think we should also check that the array is not empty, but can be done in issue #19441 like the default language checking

src/Core/Domain/Product/CommandHandler/AddProductHandlerInterface.php
```diff
@@ -0,0 +1,40 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Command\AddProductCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+
+interface AddProductHandlerInterface
```
jolelievre, line 35: Missing PHPDoc

src/Core/Domain/Product/ValueObject/ProductCondition.php
```diff
@@ -0,0 +1,83 @@
...
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+
+class ProductCondition
```
jolelievre, line 36: Missing PHPDoc

src/Core/Domain/Product/ValueObject/ProductCondition.php
jolelievre:
```suggestion
                ProductConstraintException::INVALID_CONDITION
```
No? It's weird CS fixer doesn't bug on this ^^

src/Core/Domain/Product/ValueObject/ProductType.php
```diff
@@ -0,0 +1,105 @@
...
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+
+class ProductType
```
jolelievre, line 36: Missing PHPDoc

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -55,6 +62,191 @@ public function setProductTaxRulesGroup(string $taxRulesGroupReference, string $
...
+            ));
+
+            $this->getSharedStorage()->set($productReference, $productId->getValue());
+        } catch (ProductException $e) {
+            $this->lastException = $e;
+        }
+    }
+
+    /**
+     * @Then /^product "(.+)" "(.+)" should be "(.+)"$/
```
jolelievre: Since this assert function is dedicated to localized fields maybe it should be reflected in the Behat sentence
```suggestion
     * @Then /^product "(.+)" localized "(.+)" should be "(.+)"$/
```
for example

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -55,6 +62,191 @@ public function setProductTaxRulesGroup(string $taxRulesGroupReference, string $
...
+            return;
+        }
+
+        throw new RuntimeException(sprintf('Product type is not as expected. Expected %s', $productTypeName));
+    }
+
+    /**
+     * @Then I should get error that product name is invalid
+     */
+    public function assertLastErrorIsReturnProductDisabled()
```
```suggestion
    public function assertLastErrorIsInvalidProductName()
```

tests/Integration/Behaviour/Features/Scenario/Product/add_product.feature
```diff
@@ -0,0 +1,35 @@
...
+    And product "product1" "name" should be "en-US:bottle of beer"
+    And product "product1" should be assigned to default category
+
+  @add
+  Scenario: I add a product with invalid type
+    When I add product "product2" with following basic information:
+    | name | en-US:empty box |
+    | type | undefined       |
+  Then I should get error that product type is invalid
+
```
jolelievre, line 25: Since it's already managed can you add a test about invalid name (with some invalid characters)

jolelievre (commented):
Some minor comments in the Behat test part

And, sorry to bother you with that, some missing PHPDoc it's important to identify the purpose of each class, especially the ones in the core since they declare the expected contract (for Adapter namespace I agree it's less useful IF the implemented interface is correctly documented)

src/PrestaShopBundle/Resources/config/services/adapter/product.yml
jolelievre: The `id_category` is required in the `Shop` object/table so unless the DB is corrupted in a way it should be hee, I don't think we should overly complicate the service definition for a use case that implies messing with the DB

Although it would be better to have a dedicated service to get these multi shop values, I guess it could be managed properly when the multistore is refactored in 178? But not in this PR

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
zuk3975: okay :+1:

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
zuk3975: but keywords are specific for a test, technically those strings aren't used anywhere.

matks (approved):

jolelievre (approved):

zuk3975:
"#FIRST" :sweat_smile:

PR #18689, opened 2020-04-17
[POC] vue js component for image dropzone

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Just a playground for product page image upload using vue.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of ##14405

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18689)


admin-dev/themes/new-theme/js/vue/pages/product/index.js
```diff
@@ -0,0 +1,9 @@
+import Vue from 'vue';
+import ProductImageUpload from './ProductImageUpload';
+
+export default new Vue({
+  el: '#productPageVue',
```
NeOMakinG, line 7: Would be `.image-drop` as you can have multiple drop component on the same page (not in the product page but just make it much reusable as possible)

admin-dev/themes/new-theme/js/vue/components/DragFileUpload.vue
```diff
@@ -0,0 +1,24 @@
...
+  export default {
+    name: 'DragFileUpload',
+    props: {
+      dropzoneSelector: {
+        type: String,
+        default: '',
+      },
+    },
+    mounted() {
+      document.querySelector(this.dropzoneSelector).addEventListener('dragover', (e) => {
```
PierreRambaud: Better implementation of dragover and drop events.
It's against vuejs rules to use document.querySelector
https://codepen.io/pasoevi/pen/ooOrpo

zuk3975: thanks. I've tried to use `drop` previously, but it wasn't working. Now i see it was because i must prevent `dragover` default even tho i only need the drop.

src/Core/Domain/Product/Image/Command/AddProductImageCommand.php
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * @todo: should i rename it to smth like ..localizedNames?
+     * @var string[]
+     */
+    private $localizedLegends;
+
+    /**
+     * @param int $productId
+     */
+    public function __construct(int $productId)
```
zuk3975: whoops, missing legends in constructor.

src/Adapter/Image/Uploader/ProductImageDestinationFactory.php
```diff
@@ -0,0 +1,51 @@
...
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Image\Uploader;
+
+use Image;
+use PrestaShop\PrestaShop\Core\Domain\Product\Image\ValueObject\ImageId;
+use PrestaShop\PrestaShop\Core\Image\Uploader\Exception\ImageUploadException;
+
+//@todo does it need interface?
```
matks: In theory, if It's in adapter namespace, yes, because one day we will rewrite it with another implementation.

But it's a good opportunity to check this rule still applies. @eternoendless one day you told me you thought we create too much interfaces. Is this a good usecase ?

zuk3975: I've changed the implementation, this class is no longer needed.

src/Adapter/Image/Uploader/ProductImageUploader.php
```diff
@@ -0,0 +1,248 @@
...
+            $temporaryImageName-$this->getDestinationPath($image, false),
+            'products'
+        );
+
+        Hook::exec('actionWatermark', ['id_image' => $image->id, 'id_product' => $productId]);
+        $this->updateCover($image);
+        $image->associateTo($this->contextShopIdsList);
+
+        try {
+            //@todo: this line was originally executed before the hook 'actionWatermark'. does it matter? AdminProductsController::2881
```
matks: I tried to trace it back in history to understand the "necessary to avoid file hacking" comment but it's too old. I gues it's to avoid putting in tmp folder a hacker file that could be executed later.
If it does not hurt, let's keep it.

@PierreRambaud can you tell us if this is a best practice ?

admin-dev/themes/new-theme/js/vue/pages/product/ProductImageUpload.vue
```diff
@@ -79,7 +79,9 @@
   };
 
   async function uploadImages(fileList) {
-    return fetch(router.generate('admin_products_v2_images'), {
+    return fetch(router.generate('admin_products_v2_images', {
+      productId: 1//@todo: where do i store product id?
```
zuk3975: Where should i store productId for js?
it could be data attr, but not sure to what element it could be attached.

NeOMakinG: Meh, if this is a single component, I would use a data attr to pass it as a prop

src/Core/Domain/Product/Image/Command/AddProductImageCommand.php
```diff
@@ -0,0 +1,70 @@
...
+{
+    /**
+     * @var ProductId
+     */
+    private $productId;
+
+    /**
+     * @var string[]
+     */
+    private $localizedLegends;
```
zuk3975: should i rename this to "title" or "caption" ?

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/Product/ImageController.php
```diff
@@ -0,0 +1,95 @@
...
+use PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+
+class ImageController extends FrameworkBundleAdminController
+{
+    public function uploadAction(int $productId, Request $request): JsonResponse
+    {
+        //@todo: handle multiple files
+        $imageFile = $request->files->all()[0];
```
zuk3975: @matks, any suggestion to handle multiple files?
1. I see ability to loop through files and invoke AddProductImageCommand for each of them :/
2. Thought about bulk command, but idk how should I pass the values (which should be collection/array of DTO's) when we only supposed to use scalar values. And anyway it would reuse the same single action handler ?

HELP! :smile:

src/Adapter/File/Uploader/ProductImageUploader.php
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * @var array
+     */
+    private $contextShopIdsList;
+
+    /**
+     * @var int
+     */
+    private $contextShopId;
```
matks, line 65: Do you think we should create some structured data, useful, to carry things contextShopIdsList or $contextShopId ?

Example
```
<?php
class MultistoreContext
{
    private $shopId;
    private $groupdId;
    private $isAllShops;

    public function contextIsSingleShop()
    {
        return ($this->shopId !== null);
    }
}
```
the idea would be to harmonize and provide util methods for everywhere
instead of copy-pasting always the same foreach loops in handler logic

src/PrestaShopBundle/Controller/Admin/Sell/Catalog/Product/ImageController.php
```diff
@@ -0,0 +1,134 @@
...
+            ]);
+        }
+
+        /** @var UploadedFile $imageFile */
+        $imageFile = reset($uploadedFiles);
+        $this->getCommandBus()->handle(new UploadProductImageCommand(
+            $productId,
+            $imageFile->getPathname(),
+            $imageFile->getSize(),
+            //@todo: maybe better using mime type and some lib to guess the extension?
```
matks: That would be a nice-to-have but if it's complex let's not do it. We are in the BO, we can consider the merchant knows what he is uploading.

src/Adapter/Product/CommandHandler/BulkAddProductImageHandler.php
```diff
@@ -0,0 +1,69 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\Image\Command\BulkAddProductImageCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\Image\Command\AddProductImageCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\Image\CommandHandler\BulkAddProductImageHandlerInterface;
+
+final class BulkAddProductImageHandler implements BulkAddProductImageHandlerInterface
+{
+    /**
+     * @var AddProductImageHandler
+     */
+    private $addProductImageHandler;
```
matks, line 42: 👍 nice reuse

src/Adapter/Product/Image/ProductImagePathFactory.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        $image = $this->getImage($imageId);
+
+        if ($this->isLegacyImageMode) {
+            $path = $image->id_product . '-' . $image->id;
+        } else {
+            $path = $image->getImgPath();
+        }
+
+        //@todo: it seems that jpg is hardcoded. AdminProductsController:2836
```
matks, line 68: So only JPG is supported by legacy ?

src/Adapter/File/Uploader/ProductImageUploader.php
zuk3975: Need to check if MultistoreChecker(or smth like that) doesnt handle it already.

src/Adapter/Product/CommandHandler/BulkAddProductImageHandler.php
zuk3975: Yeah, but it will be removed as i suppose to switch to UploadaimageHandler, which adds the entity and uploads the file :D

src/Adapter/Product/Image/ProductImagePathFactory.php
zuk3975: It accepts png, gif and jpg, but saves all of them with 'jpg' extension

src/Adapter/Image/Uploader/ProductImageDestinationFactory.php
eternoendless: In general adapters need an interface because they are meant to be replaced by core reimplementations. Ideally, these interfaces should be well thought and not just mimic the legacy class. You can also rely on interface segregation for that (aka "don't make clients depend on functionality they don't need" aka "make your interfaces simple" aka "implement more than one interface if you need to")

src/Adapter/Image/Uploader/ProductImageUploader.php
PierreRambaud: I don't really understand why they did it :thinking: 
But it's always good to remove temporary files if they are not needed.

jolelievre:
@zuk3975 I think this one can be closed since @NeOMakinG started a new component which has already been partially merged Fell free to reopen it if you think otherwise 😉

PR #18736, opened 2020-04-21
Address State field drop down is ordered by ID, instead of State Name

### LINKED ISSUE #18731: Address State field drop down is ordered by ID, instead of State Name
#### Describe the bug

Anywhere where Address field is displayed (in client are or during checkout), the drop down with States orders values by database ID and not name field.
This makes the newly added States appear last in the list, thus confusing customer who will not see his State unless he will scroll to the bottom of the list.

#### Expected behavior

State field in address forms should be ordered by Name field in ps_state table, or whatever prefix is used.

#### Steps to Reproduce

1. From Client Area, go to Addresses
2. Edit or create a new address
3. State field (not visible by default but it can be added) will have values listed by ID.

**Screenshots**

Untitled4
Untitled5

#### Additional information

* PrestaShop version: Tested with 1.7.6.4 and 1.7.6.5

PululuK:
| Questions     | Answers
| Branch?       | develop 
| Description?  | Anywhere where Address field is displayed (in client are or during checkout), the drop down with States orders values by database ID and not name field.This makes the newly added States appear last in the list, thus confusing customer who will not see his State unless he will scroll to the bottom of the list
| Type?         | refacto 
| Category?     | CO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18731.
| How to test?  | See #18731 by @vitciobanu.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18736)

### DIFF
--- a/classes/State.php
+++ b/classes/State.php
@@ -206,18 +206,22 @@ class StateCore extends ObjectModel
      *
      * @return array|false|mysqli_result|PDOStatement|resource|null
      */
-    public static function getStatesByIdCountry($idCountry, $active = false)
+    public static function getStatesByIdCountry($idCountry, $active = false, $orderBy = null)
     {
         if (empty($idCountry)) {
             die(Tools::displayError());
         }
 
-        return Db::getInstance()->executeS(
-            '
-			SELECT *
-			FROM `' . _DB_PREFIX_ . 'state` s
-			WHERE s.`id_country` = ' . (int) $idCountry . ($active ? ' AND s.active = 1' : '')
-        );
+        $sql = new DbQuery();
+        $sql->select('*');
+        $sql->from('state', 's');
+        $sql->where('s.id_country = ' . (int) $idCountry . ($active ? ' AND s.active = 1' : ''));
+
+        if (array_key_exists($orderBy, self::$definition['fields'])) {
+            $sql->orderBy($orderBy);
+        }
+
+        return Db::getInstance()->executeS($sql);
     }
 
     /**
--- a/classes/State.php
+++ b/classes/State.php
@@ -204,14 +204,17 @@ class StateCore extends ObjectModel
      * @param int $idCountry Country ID
      * @param bool $active true if the state must be active
      * @param string $orderBy order by field
+     * @param string $sort sort key (ASC or DESC)
      *
      * @return array|false|mysqli_result|PDOStatement|resource|null
      */
-    public static function getStatesByIdCountry($idCountry, $active = false, $orderBy = null)
+    public static function getStatesByIdCountry($idCountry, $active = false, $orderBy = null, $sort = 'ASC')
     {
         if (empty($idCountry)) {
             die(Tools::displayError());
         }
+
+        $available_sort = ['DESC', 'ASC', 'asc', 'desc'];
 
         $sql = new DbQuery();
         $sql->select('*');
@@ -219,6 +222,10 @@ class StateCore extends ObjectModel
         $sql->where('s.id_country = ' . (int) $idCountry . ($active ? ' AND s.active = 1' : ''));
 
         if (array_key_exists($orderBy, static::$definition['fields'])) {
+            $sort = trim($sort);
+            if (in_array($sort, $available_sort)) {
+                $orderBy = $orderBy . ' ' . $sort;
+            }
             $sql->orderBy($orderBy);
         }
 
--- a/classes/State.php
+++ b/classes/State.php
@@ -203,6 +203,7 @@ class StateCore extends ObjectModel
      *
      * @param int $idCountry Country ID
      * @param bool $active true if the state must be active
+     * @param string $orderBy order by field
      *
      * @return array|false|mysqli_result|PDOStatement|resource|null
      */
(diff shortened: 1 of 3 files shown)


Progi1984:
@PululuK Could you fix Coding Standards on your PR : `./vendor/bin/php-cs-fixer fix --dry-run --diff` ? Thanks :)

PierreRambaud (dismissed):
You duplicate the getStatesByIdCountry logic. Feel free to improve the original method to allow the  sort by specific key instead.

PululuK:
> You duplicate the getStatesByIdCountry logic. Feel free to improve the original method to allow the sort by specific key instead.

@PierreRambaud I thought about it, but I was afraid of the impact in other places where it is used ... I will do that then!

I like the idea thanks man !

vitciobanu:
Guys, I'm impressed by your quick work on this _minor_ issue! Thank you!

I'm new to PrestaShop and issue resolutions here, so please let me know if you need me to download a new build to retest or something else.

PierreRambaud:
> > You duplicate the getStatesByIdCountry logic. Feel free to improve the original method to allow the sort by specific key instead.
> 
> @PierreRambaud I thought about it, but I was afraid of the impact in other places where it is used ... I will do that then!
> 
> I like the idea thanks man !

Something like this, but I did not test it:
```php
    /**
     * Get states by Country ID.
     *
     * @param int $idCountry Country ID
     * @param bool $active true if the state must be active
     * @param string $orderBy Field to order result
     *
     * @return array|false|mysqli_result|PDOStatement|resource|null
     */
    public static function getStatesByIdCountry($idCountry, $active = false, $orderBy = null)
    {
        if (empty($idCountry)) {
            die(Tools::displayError());
        }

        $sql = 'SELECT * ' .
			'FROM `' . _DB_PREFIX_ . 'state` s ' .
			'WHERE s.`id_country` = ' . (int) $idCountry .
             (
                 $active ? ' AND s.active = 1' : ''
             );

        if ($orderBy !== null) {
            $sql .= 'ORDER BY s.`' . bqSQL($orderBy) . '`';
        }

        return Db::getInstance()->executeS(
            $sql
        );
    }

```

Progi1984:
> > You duplicate the getStatesByIdCountry logic. Feel free to improve the original method to allow the sort by specific key instead.
> 
> @PierreRambaud I thought about it, but I was afraid of the impact in other places where it is used ... I will do that then!
> 
> I like the idea thanks man !

Use parameters for `getStatesByIdCountry` and let the initial behavior as default behavior

classes/State.php
```diff
@@ -206,18 +206,22 @@ public function countUsed()
...
-			SELECT *
-			FROM `' . _DB_PREFIX_ . 'state` s
-			WHERE s.`id_country` = ' . (int) $idCountry . ($active ? ' AND s.active = 1' : '')
-        );
+        $sql = new DbQuery();
+        $sql->select('*');
+        $sql->from('state', 's');
+        $sql->where('s.id_country = ' . (int) $idCountry . ($active ? ' AND s.active = 1' : ''));
+
+        if (array_key_exists($orderBy, self::$definition['fields'])) {
```
PierreRambaud:
```suggestion
        if (array_key_exists($orderBy, static::$definition['fields'])) {
```
If someone already override this class :sweat_smile:

classes/State.php
```diff
@@ -206,18 +206,22 @@ public function countUsed()
      *
```
Progi1984, line 208: Miss only $orderBy and it's :ok_hand: for me.

classes/State.php
```diff
@@ -203,21 +203,26 @@ public function countUsed()
...
      * @param int $idCountry Country ID
      * @param bool $active true if the state must be active
+     * @param string $orderBy order by field
      *
      * @return array|false|mysqli_result|PDOStatement|resource|null
      */
-    public static function getStatesByIdCountry($idCountry, $active = false)
+    public static function getStatesByIdCountry($idCountry, $active = false, $orderBy = null)
```
jolelievre: The only issue is you can't define the order way ASC or DESC?
Should we add another parameter, or use a convention like in the webservice `name_ASC`?
Or do you think it's not useful at all?

PululuK: Very good question. I can add it in my PR if you allow me !
I would like to have the opinion of another developer
Thank you so much @jolelievre 

ping @PierreRambaud  @Progi1984

Progi1984: Good idea @jolelievre

Progi1984 (dismissed):

matks (dismissed):

khouloudbelguith:
Hi @PululuK,

In the FO, it is OK
image

But in the BO, it is NOK
image

Thanks to check and feedback.

PululuK:
> Hi @PululuK,
> 
> In the FO, it is OK
> But in the BO, it is NOK

Thank you very much @khouloudbelguith ! but ... where exactly in BO ?

khouloudbelguith:
@PululuK, in the BO => Customers => Addresses => try to add a new address and select a country which contains states for example Spain
image

Thanks!

PululuK:
> @PululuK, in the BO => Customers => Addresses => try to add a new address and select a country which contains states for example Spain
> Thanks!

Done ! Thank you @khouloudbelguith

Progi1984 (approved):

atomiix (approved):

khouloudbelguith:
Hi @PululuK,

It is OK :heavy_check_mark: 
**BO**
image
**FO**
image

Thanks!

Progi1984:
Thanks @PululuK

PR #18464, opened 2020-04-03
Add modal to customer address listing on bulk action

matthieu-rolland:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | Add modal to customer address listing on bulk action (deletion)
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18414
| How to test?  | 1- Go in BO's customer address listing page<br> 2- select one or more addresses<br> 3- Click on "grouped actions", "delete", a modal should appear

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18464)

Issue #17847, opened 2020-02-26
Add a modal to confirm or cancel delete action for one row [1/1]


marionf:
Add a modal like on product list to confirm or cancel delete for one row

⚠ Add this modal **with the following message** - edited by @LouiseBonnard following #17082:

`title` = _Delete selection_ in _Admin.Actions_
`confirm_message` = _Are you sure you want to delete the selected item(s)?_ in _Admin.Global_

`confirm_button_label` = _Delete_ in _Admin.Actions_
`close_button_label` = _Cancel_ in _Admin.Actions_

---

Already migrated pages for which it must be added:
- [x] Customers > Addresses
- [x] Customers > Customers > View > Addresses
- [x] Customers > Customers > View > Vouchers
- [x] Customer service > Order message
- [x] Catalog > Files
- [x] Catalog > Brands > Brands
- [x] Catalog > Brands > Addresses
- [x] Catalog > Suppliers
- [x] Catalog > Monitoring
- [x] Design > Pages
- [x] International > Localization > Languages
- [x] International > Localization > Currencies
- [x] Appearence > Pages (categories + pages)
- [x] Shop parameters > Traffic & SEO > SEO&URLs
- [x] Advanced parameters > Email  
- [x] Advanced parameters > Team > Employees  
- [x] Advanced parameters > Team > Profiles  
- [x] Advanced parameters > Database > SQL Manager  
- [x] Advanced parameters > Database > DB Backup
- [x] Advanced parameters > Webservice 
- [x] International > Taxes > Taxes
- [x] Shop parameters > Contact > Contacts

jolelievre:
This is such a pain to reopen this one each time 😅

marionf:
Yes, it was to avoid to make an issue by page (such a pain also) :sweat_smile:

Progi1984:
@marionf No more checkbox. ;)

Progi1984:
:muscle: @marionf

PR #18764, opened 2020-04-22
Fix exception when deleting address from customer view

### LINKED ISSUE #18748: BO - Customer View page - An exception is displayed when deleting an address
#### Describe the bug

In the BO => Customers => Customer view page, when we try to delete an address, an exception is displayed.

#### Expected behavior

It should delete the address successfully.

#### Steps to Reproduce

1. Go to BO => Customers => Customer view page or example "John DOE"
2. Try to delete an address
3. See error

**Screenshots**

image

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: 7.2

sowbiba:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fix exception displayed when an address is deleted from customer view
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18748
| How to test?  | 1. Go to BO => Customers => Customer view page or example "John DOE"<br>2. Try to delete an address<br>3. You don't see any error but the delete success alert

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18764)


Progi1984 (dismissed):

Progi1984 (dismissed):

admin-dev/themes/new-theme/js/components/form-submit-button.js
```diff
@@ -60,11 +60,32 @@ export default class FormSubmitButton {
         return;
       }
 
+      let method = 'POST';
+      let addInput = null;
+
+      if ($btn.data('method')) {
+        const btnMethod = $btn.data('method');
+        const isGetOrPostMethod = ['GET', 'POST'].includes(btnMethod);
+        method = isGetOrPostMethod ? method : 'POST';
```
matthieu-rolland, line 69: It seems like the value of the **method** variable can only be 'POST'. Or is there something I don't get ?

src/PrestaShopBundle/Resources/views/Admin/Sell/Customer/Blocks/View/addresses.html.twig
```diff
@@ -92,12 +92,15 @@
                     ></a>
 
                     <div class="dropdown-menu dropdown-menu-right">
-                      <a class="btn tooltip-link dropdown-item"
-                         href="{{ getAdminLink('AdminAddresses', true, {'id_address': address.addressId, 'deleteaddress': 1, 'back': app.request.uri}) }}"
+                      <button type="button"
+                              class="btn tooltip-link dropdown-item js-form-submit-btn"
+                              data-form-confirm-message="{{ 'Are you sure you want to delete the selected item(s)?'|trans({}, 'Admin.Global') }}"
+                              data-form-submit-url="{{ getAdminLink('AdminAddresses', true, {'id_address': address.addressId, 'deleteaddress': 1, 'back': app.request.uri}) }}"
```
jolelievre: You can use the symfony route on this one now `admin_addresses_delete`

admin-dev/themes/new-theme/js/components/form-submit-button.js
matthieu-rolland: It seems like the value of the **method** variable can only be 'POST'. Or is there something I don't get ?
Maybe you wanted to write:
```suggestion
        method = isGetOrPostMethod ? btnMethod : 'POST';
```

matthieu-rolland (commented):
Just one feedback..

admin-dev/themes/new-theme/js/components/form-submit-button.js
```diff
@@ -60,11 +60,32 @@ export default class FormSubmitButton {
         return;
       }
 
+      let method = 'POST';
```
jolelievre, line 65: Great addition to this component!! Can you update the JSDoc on the class and add the new `data-method` parameter please?

admin-dev/themes/new-theme/js/components/form-submit-button.js
jolelievre: No it can be GET or POST, and for other methods (HEAD, DELETE) an hidden input is added and the http method used is POST

matthieu-rolland: yes but the **method** variable is set to 'POST' and never changed after...  I think he meant to use the value of **btnMethod** instead. 🤔

sowbiba: If the method is something else than GET or POST, POST will be set as verb and the hidden input will do the job to make the request with the correct verb

src/PrestaShopBundle/Resources/views/Admin/Sell/Customer/Blocks/View/addresses.html.twig
```diff
@@ -92,12 +92,15 @@
                     ></a>
 
                     <div class="dropdown-menu dropdown-menu-right">
-                      <a class="btn tooltip-link dropdown-item"
-                         href="{{ getAdminLink('AdminAddresses', true, {'id_address': address.addressId, 'deleteaddress': 1, 'back': app.request.uri}) }}"
+                      <button type="button"
+                              class="btn tooltip-link dropdown-item js-form-submit-btn"
+                              data-form-confirm-message="{{ 'Are you sure you want to delete the selected item(s)?'|trans({}, 'Admin.Global') }}"
```
LouiseBonnard: Currently, there are two exceptions on this page: one when deleting one row (_Delete selected item?_) and another when deleting in bulk (_Delete selected items?_). Which exception does this PR fix? Or is it a 2-in-1 exception that would work for both cases?

sowbiba: this confirm message I updated is for single row deletion

LouiseBonnard: Okay, thanks @sowbiba! In this case, I'd keep it consistent with the other exception (the bulk action exception) and write _Delete selected item?_, localized in _Admin.Notifications.Warning_. It is not a deletion modal because the page has not been migrated in Symfony yet, right?

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -144,6 +144,11 @@ public function deleteAction(int $addressId): RedirectResponse
             $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));
         }
 
+        $redirectUrl = $request->query->get('redirectUrl');
```
sowbiba: Maybe I can do it like 

```return $request->query->has('redirectUrl') ? $this->redirect($request->query->get('redirectUrl')) : $this->redirectToRoute('admin_addresses_index');```

WDYT ?

src/PrestaShopBundle/Resources/views/Admin/Sell/Customer/Blocks/View/addresses.html.twig
sowbiba: didn't want to introduice this subject :joy:  but yes, this should be a modal, not a simple confirm box

sowbiba: And in the other modals for deletion we did, the confirm message was "{{ 'Are you sure you want to delete the selected item(s)?'|trans({}, 'Admin.Global') }}"
That's why I choosed this wording

LouiseBonnard: I know, and you're right about it. But since the bug is about a confirm box, we should keep the behavior iso, at least for now, and turn these two confirm boxes into one modal when the page will be migrated.

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
matthieu-rolland: It seems fine to me.

jolelievre: I don't see any problem with that

Progi1984: And me

matthieu-rolland (approved):

src/PrestaShopBundle/Resources/views/Admin/Sell/Customer/Blocks/View/addresses.html.twig
LouiseBonnard: Actually, the page is migrated for 1.7.7 and I've just come across this PR, https://github.com/PrestaShop/PrestaShop/pull/18464, so I guess it should definitely be a modal here. It looks like @marionf is the owner of this topic, can you confirm?

Progi1984 (approved):

src/PrestaShopBundle/Resources/views/Admin/Sell/Customer/Blocks/View/addresses.html.twig
sowbiba: The PR #18464 is for bulk action

https://github.com/PrestaShop/PrestaShop/issues/17847 :thinking:

matks (approved):

src/PrestaShopBundle/Resources/views/Admin/Sell/Customer/Blocks/View/addresses.html.twig
sowbiba: We keep the classic confirm for this PR. The modal will be done in another PR, for 1.7.8

LouiseBonnard (approved):

src/PrestaShopBundle/Resources/views/Admin/Sell/Customer/Blocks/View/addresses.html.twig
Progi1984: :+1:

PR #18787, opened 2020-04-23
Add compatibility to PHP 7.4

### LINKED ISSUE #16477: PrestaShop must be compatible with PHP 7.4
PHP 7.4 is due for release in November 2019.

PrestaShop 1.7.8 should work with PHP 7.4.

PierreRambaud:
| Questions     | Answers
| Branch?       | develop
| Description?  | Make PrestaShop compatible with PHP 7.4
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16477
| How to test?  | Does it requires QA, we certainly find some new issues in the future :thinking: 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18787)


classes/QuickAccess.php
```diff
@@ -125,7 +125,7 @@ public static function getQuickAccessesWithToken($idLang, $idEmployee)
      */
     public function toggleNewWindow()
     {
-        if (!array_key_exists('new_window', $this)) {
+        if (!array_key_exists('new_window', get_object_vars($this))) {
```
Progi1984, line 128: Why not use `property_exists` ?

PierreRambaud: Because property_exists returns true when the property is private or protected

Progi1984: Oh !

PierreRambaud:
Still waiting for modules to be released to continue :sweat_smile:

mvorisek:
> 
> 
> Still waiting for modules to be released to continue 😅

@PierreRambaud what modules need to be addressed?

PierreRambaud:
> > Still waiting for modules to be released to continue sweat_smile
> 
> @PierreRambaud what modules need to be addressed?

Native modules, like psgdpr, will be released soon :)

Progi1984 (dismissed):

Progi1984 (dismissed):

matks:
Behat not happy
```
-- Failed steps:
001 Scenario: Update order shipping details                                                                              # Features/Scenario/Order/order_from_bo.feature:42

      When I update order "bo_order1" Tracking number to "TEST1234" and Carrier to "2 - My carrier (Delivery next day!)" # Features/Scenario/Order/order_from_bo.feature:43
        Notice: Trying to access array offset on value of type bool in /home/travis/build/PrestaShop/PrestaShop/classes/order/OrderCarrier.php line 112

483 scenarios (482 passed, 1 failed)
6525 steps (6522 passed, 1 failed, 2 skipped)
7m46.45s (185.14Mb)
Script @php -d date.timezone=UTC ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml  --format progress --no-snippets --strict handling the integration-behaviour-tests event returned with error code 1
```

PierreRambaud:
> Behat not happy
> 
> ```
> -- Failed steps:
> 001 Scenario: Update order shipping details                                                                              # Features/Scenario/Order/order_from_bo.feature:42
> 
>       When I update order "bo_order1" Tracking number to "TEST1234" and Carrier to "2 - My carrier (Delivery next day!)" # Features/Scenario/Order/order_from_bo.feature:43
>         Notice: Trying to access array offset on value of type bool in /home/travis/build/PrestaShop/PrestaShop/classes/order/OrderCarrier.php line 112
> 
> 483 scenarios (482 passed, 1 failed)
> 6525 steps (6522 passed, 1 failed, 2 skipped)
> 7m46.45s (185.14Mb)
> Script @php -d date.timezone=UTC ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml  --format progress --no-snippets --strict handling the integration-behaviour-tests event returned with error code 1
> ```

Still working on it :sweat_smile: 
I know, but I'm focused on the 1.7.7.x branch right now

matks:
> Still working on it 😅
> I know, but I'm focused on the 1.7.7.x branch right now

WORK FASTER

PierreRambaud:
> > Still working on it sweat_smile
> > I know, but I'm focused on the 1.7.7.x branch right now
> 
> WORK FASTER

:sob:

Progi1984 (dismissed):

classes/Address.php
```diff
@@ -303,7 +303,11 @@ public static function getZoneById($id_address)
 			LEFT JOIN `' . _DB_PREFIX_ . 'state` s ON s.`id_state` = a.`id_state`
 			WHERE a.`id_address` = ' . (int) $id_address);
 
-        self::$_idZones[$id_address] = (int) ((int) $result['id_zone_state'] ? $result['id_zone_state'] : $result['id_zone']);
+        if (empty($result['id_zone_state']) && empty($result['id_zone'])) {
+            return false;
+        }
+
+        self::$_idZones[$id_address] = (int) ((int) $result['id_zone_state'] ?? $result['id_zone']);
```
mvorisek: right side of ?? never reached

```suggestion
        self::$_idZones[$id_address] = (int) ($result['id_zone_state'] ?? $result['id_zone']);
```

PierreRambaud: Thanks, sometimes I forget they didn't learn about null value in the database and stored "0" :facepalm:

matthieu-rolland (dismissed):
lgtm ! Let's see if behat is happy

classes/module/Module.php
```diff
@@ -1380,7 +1380,7 @@ public static function getModulesOnDisk($use_config = false, $logged_on_addons =
                     $file = trim(file_get_contents(_PS_MODULE_DIR_ . $module . '/' . $module . '.php'));
 
                     try {
-                        $parser = (new PhpParser\ParserFactory())->create(PhpParser\ParserFactory::PREFER_PHP7);
+                        $parser = (new PhpParser\ParserFactory())->create(PhpParser\ParserFactory::ONLY_PHP7);
```
jolelievre, line 1383: Does it mean a module not compatible with PHP7 will now be considered invalid?
I'm not sure it's really a BC break, but it looks like it, at the very least it's probably worth mentioning as an important change for this version and needs to be documented no?

controllers/front/ProductController.php
```diff
@@ -789,7 +789,7 @@ protected function assignCategory()
             // various assignements before Hook::exec
             $this->context->smarty->assign([
                 'category' => $this->category,
-                'subCategories' => $sub_categories,
+                'subcategories' => $sub_categories,
```
jolelievre, line 793: If the variable names changes it can result in BC break for themes or modules relying on it no?

src/Adapter/Module/ModuleDataProvider.php
```diff
@@ -218,7 +218,7 @@ public function isModuleMainClassValid($name)
             return false;
         }
 
-        $parser = (new PhpParser\ParserFactory())->create(PhpParser\ParserFactory::PREFER_PHP7);
+        $parser = (new PhpParser\ParserFactory())->create(PhpParser\ParserFactory::ONLY_PHP7);
```
jolelievre, line 221: Same comment as above regarding a potential BC break?

themes/classic/templates/catalog/_partials/category-header.tpl
```diff
@@ -30,7 +30,7 @@
                 {if $category.description}
                     <div id="category-description" class="text-muted">{$category.description nofilter}</div>
                 {/if}
-                {if $category.image.large.url}
+                {if !empty($category.image.large.url)}
```
jolelievre, line 33: Should there me a "no picture" fallback like for product?

jolelievre (dismissed):
@PierreRambaud I globally approve the PR, but I have a few questions that I'd like you to confirm

Especially about potential BC breaks, which are acceptable for this kind of modification but still need to be labelled and documented

classes/Address.php
mvorisek: may be true, but did you tested to remove the inner cast? It comes from join...

classes/module/Module.php
PierreRambaud: We don't support anymore PHP5.6, I don't think it's a BC breaks :thinking:

controllers/front/ProductController.php
PierreRambaud: We use the `$subcategories` in our templates, never the `$subCategories` it was a bug

themes/classic/templates/catalog/_partials/category-header.tpl
PierreRambaud: Since it's a category, I don't think there is :thinking:

tests-legacy/PrestaShopBundle/Utils/DatabaseCreator.php
```diff
@@ -45,6 +46,7 @@ public static function createTestDB()
         require_once __DIR__ . '/../../../install-dev/init.php';
 
         $install = new Install();
+        $install->setTranslator(Context::getContext()->getTranslatorFromLocale('en'));
```
atomiix, line 49: PHP 7.4 make this mandatory?

.travis.yml
```diff
@@ -25,6 +25,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
```
eternoendless, line 27: I think we can remove these two, they are useless

classes/Address.php
```diff
@@ -303,7 +303,15 @@ public static function getZoneById($id_address)
...
-        self::$_idZones[$id_address] = (int) ((int) $result['id_zone_state'] ? $result['id_zone_state'] : $result['id_zone']);
+        if (empty($result['id_zone_state']) && empty($result['id_zone'])) {
+            return false;
+        }
+
+        self::$_idZones[$id_address] = (int) (
+            !empty($result['id_zone_state'])
+            ? $result['id_zone_state']
+            : $result['id_zone']
+        );
```
eternoendless: Can be simplified
```suggestion
        self::$_idZones[$id_address] = !empty($result['id_zone_state'])
            ? (int) $result['id_zone_state']
            : (int) $result['id_zone'];
```

classes/module/Module.php
matks: > Does it mean a module not compatible with PHP7 will now be considered invalid?

I think it's consistent. "My software works with vendor X > v7, then I only accept plugins compatible with X > v7". We have regularly complains about bad modules in the wild, I think we should aim to discard such modules as soon as possible (before install happens).

> at the very least it's probably worth mentioning as an important change for this version and needs to be documented no?

I 👍  . It is better to clear the confusion.

classes/Language.php
```diff
@@ -984,6 +984,10 @@ public static function loadLanguages()
 				LEFT JOIN `' . _DB_PREFIX_ . 'lang_shop` ls ON (l.id_lang = ls.id_lang)';
 
         $result = Db::getInstance()->executeS($sql);
+        if (!is_array($result)) {
+            // executeS method can returns false
```
eternoendless:
```suggestion
            // executeS method can return false
```
That said, I think an exception might be more appropriate, because you should always have at least one language in database right?

controllers/front/ProductController.php
matks: What about 3rd party themes ? If they use `subCategories` ?

Moreover if `$subcategories` is used in Classic Theme, it means that today on `develop` there are 2 smarty variables `subcategories` and `subCategories`. Did you check the content is identical ?

matks (commented):
PR is overall good. I just have 1 warning about `subcategories` variable renaming, see below.

classes/Profile.php
```diff
@@ -128,7 +128,7 @@ public static function getProfileAccess($idProfile, $idTab)
      * @param int $idProfile Profile ID
      * @param string $type Type
      *
-     * @return bool
+     * @return bool|array
```
eternoendless: Would be nice to define the array structure. I like Psalm's [object-like array definitions](https://psalm.dev/docs/annotating_code/type_syntax/array_types/#object-like-arrays).

Otherwise this can be specified like this:
```suggestion
     * @return array|false
```

classes/module/Module.php
eternoendless: This should be noted

controllers/front/CartController.php
```diff
@@ -133,7 +133,7 @@ public function displayAjaxUpdate()
         $productsInCart = $this->context->cart->getProducts();
         $updatedProducts = array_filter($productsInCart, [$this, 'productInCartMatchesCriteria']);
         $updatedProduct = reset($updatedProducts);
-        $productQuantity = $updatedProduct['quantity'];
+        $productQuantity = $updatedProduct['quantity'] ?? false;
```
eternoendless: Shouldn't it be 0? Or null? Why false?

tests/Unit/Core/Webservice/ServerRequirementsCheckerTest.php
```diff
@@ -63,7 +63,10 @@ protected function setUp()
             ->will($this->returnArgument(0));
 
         $this->mockedConfiguration = $this->createMock(Configuration::class);
-        $this->mockedHostingInformation = $this->createMock(HostingInformation::class);
+        $this->mockedHostingInformation = $this->getMockBuilder(HostingInformation::class)
```
eternoendless, line 66: Why was this change needed?

classes/Language.php
PierreRambaud: Unfortunately, during the install process, you don't have any languages in the database. :(

controllers/front/CartController.php
Sinepel: I think 0 would be more appropriate, what do you think?

.travis.yml
PierreRambaud: I'm not sure, since PHP introduces BC breaks on each version, we need to know which one is affected, sometimes. I'm afraid of having something working on 7.2, 7.4 and not working on 7.3 due to a bug or a BC breaks :/ 
Since our QA team is not testing all PHP versions, we should let Travis do it.

classes/module/Module.php
PierreRambaud: It will be in the release note and the devdoc of course

controllers/front/CartController.php
PierreRambaud:
```suggestion
        $productQuantity = $updatedProduct['quantity'] ?? 0;
```

PierreRambaud: Yes, I changed it to 0 :+1:

tests/Unit/Core/Webservice/ServerRequirementsCheckerTest.php
PierreRambaud: To prevent this error 
image

For more information: https://phpunit.de/manual/6.5/en/test-doubles.html

tests-legacy/PrestaShopBundle/Utils/DatabaseCreator.php
PierreRambaud: In fact, not only for PHP7.4, but if something failed during the install process, you don't receive any message in CLI because the translator is not loaded. It helps me to understand what was wrong :sweat_smile:

classes/module/Module.php
jolelievre: I totally agree with the updated constraint, it totally makes sense But then it is a BC break for modules since a code not PHP7 compliant which used to work will be rejected This alone should justify to note this PR as a BC break
Although it's perfectly understandable and acceptable

jolelievre (dismissed):

atomiix (dismissed):

classes/module/Module.php
PierreRambaud: We already did it when we removed the PHP5.6 support. So if a module is not compliant with PHP7, it shouldn't be able to run since the 1.7.6.0 :thinking:

controllers/front/ProductController.php
PierreRambaud: There is only one variable per controller. In ProductController, it's subCategories, in the others it's subcategories. You are right, maybe modules are using it, I'm changing it :)

PierreRambaud, line 793:
```suggestion
                'subCategories' => $sub_categories,
                'subcategories' => $sub_categories,
```

matks (approved):

matthieu-rolland (approved):

atomiix (approved):

jolelievre (approved):

PierreRambaud:
:tada:

PR #18811, opened 2020-04-24
Introduce CompilerPass to perform ControllerAwareTrait injection

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | As explained in https://github.com/symfony/symfony/issues/36567, ControllerAwareTrait requires injection of container and this is performed by Symfony ControllerResolver. If a controller is a decorated, it is not injected the container anymore. I avoid it by manually injecting it into all PrestaShop controllers.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | See below

### How to test

Use attached demo module
[testdecorationcontrollers.zip](https://github.com/PrestaShop/PrestaShop/files/4645455/testdecorationcontrollers.zip)

1. Install module
2. Go on CMS listing page "Improve > Design > Pages"
3. You should see the green message
Capture d’écran 2020-05-18 à 18 44 41
4. You can check on `develop` the same module, you should see the red exception
Capture d’écran 2020-05-18 à 18 44 17

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18811)

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -23,9 +23,7 @@ services:
         exclude: "%kernel.root_dir%/../src/PrestaShopBundle/Controller/Api"
         tags:
           - 'controller.service_arguments'
-          - !php/const PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController::PRESTASHOP_CORE_CONTROLLERS
-
-
+          - !php/const PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController::PRESTASHOP_CORE_CONTROLLERS_TAG
 
 framework:
     assets:
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -21,7 +21,11 @@ services:
     PrestaShopBundle\Controller\:
         resource: "%kernel.root_dir%/../src/PrestaShopBundle/Controller/*"
         exclude: "%kernel.root_dir%/../src/PrestaShopBundle/Controller/Api"
-        tags: ['prestashop.core.controllers', 'controller.service_arguments']
+        tags:
+          - 'controller.service_arguments'
+          - !php/const PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController::PRESTASHOP_CORE_CONTROLLERS
+
+
 
 framework:
     assets:
--- a/src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
+++ b/src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShopBundle\DependencyInjection\Compiler;
+
+use InvalidArgumentException;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Decorated controllers using ControllerAwareTrait
+ * cannot be injected the container as the injection is performed by ControllerResolver.
+ *
+ * This pass injects the container into PrestaShop controllers to overcome this issue.
+ */
+class ContainerAwareControllersManualInjectionPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $controllers = $container->findTaggedServiceIds('prestashop.core.controllers');
+
+        foreach ($controllers as $id => $controller) {
+            $definition = $container->findDefinition($id);
+            $class = $definition->getClass();
+
+            if (!$r = $container->getReflectionClass($class)) {
+                throw new InvalidArgumentException(sprintf(
+                    'Class "%s" used for service "%s" cannot be found.',
+                    $class,
+                    $id
+                ));
+            }
+            $isContainerAware = $r->implementsInterface(ContainerAwareInterface::class)
+                || is_subclass_of($class, AbstractController::class);
+
+            if ($isContainerAware) {
+                $definition->addMethodCall('setContainer', [new Reference('service_container')]);
+            }
+        }
+    }
+}
--- a/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
+++ b/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
@@ -49,7 +49,7 @@ use Symfony\Component\HttpFoundation\Response;
  */
 class FrameworkBundleAdminController extends Controller
 {
-    const PRESTASHOP_CORE_CONTROLLERS = 'prestashop.core.controllers';
+    const PRESTASHOP_CORE_CONTROLLERS_TAG = 'prestashop.core.controllers';
 
     /**
      * @var ConfigurationInterface
--- a/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
+++ b/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
@@ -49,6 +49,8 @@ use Symfony\Component\HttpFoundation\Response;
  */
 class FrameworkBundleAdminController extends Controller
 {
+    const PRESTASHOP_CORE_CONTROLLERS = 'prestashop.core.controllers';
+
     /**
      * @var ConfigurationInterface
      */
--- a/src/PrestaShopBundle/DependencyInjection/Compiler/ContainerInjectionPass.php
+++ b/src/PrestaShopBundle/DependencyInjection/Compiler/ContainerInjectionPass.php
@@ -24,9 +24,12 @@
  * International Registered Trademark & Property of PrestaShop SA
  */
 
+declare(strict_types=1);
+
 namespace PrestaShopBundle\DependencyInjection\Compiler;
 
 use InvalidArgumentException;
+use PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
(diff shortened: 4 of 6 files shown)


app/config/config.yml
```diff
@@ -21,7 +21,7 @@ services:
     PrestaShopBundle\Controller\:
         resource: "%kernel.root_dir%/../src/PrestaShopBundle/Controller/*"
         exclude: "%kernel.root_dir%/../src/PrestaShopBundle/Controller/Api"
-        tags: ['prestashop.core.controllers']
+        tags: ['prestashop.core.controllers', 'controller.service_arguments']
```
matks: Unrelated to this PR, but I saw that Symfony default services configuration has this. So I add it I think it's needed.

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $controllers = $container->findTaggedServiceIds('prestashop.core.controllers');
+
+        foreach ($controllers as $id => $controller) {
+            $definition = $container->findDefinition($id);
+            $class = $definition->getClass();
```
matks: @PierreRambaud I know what you are going to say:

```suggestion
            $class = $container->findDefinition($id)->getClass();
```

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
```diff
@@ -0,0 +1,70 @@
...
+
+            if (!$r = $container->getReflectionClass($class)) {
```
matks: Reflection is slow but since container is compiled, this will not be done at runtime.

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
jolelievre: Auto-review mode on! 😅

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
jolelievre: Yep I had the same process in my mind when I read this 😅

app/config/config.yml
mickaelandrieu: for information => https://symfony.com/doc/3.4/service_container/3.3-di-changes.html#controllers-are-registered-as-services

matks: Thanks, so it's to enable autowiring arguments for controller action methods

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
matks: You can see the psychological damages of hundred of reviews by @PierreRambaud 😱

jolelievre:
@matks Do you think we should change our base class to `AbstractController` as suggested here https://github.com/symfony/symfony/issues/36567#issuecomment-619109593

Compared to the `Controller` class we use the only difference seems to be the `protected function getParameter($name)` function but we could add it into our `FrameworkBundleAdminController`

matks:
> @matks Do you think we should change our base class to `AbstractController` as suggested here [symfony/symfony#36567 (comment)](https://github.com/symfony/symfony/issues/36567#issuecomment-619109593)
> 
> Compared to the `Controller` class we use the only difference seems to be the `protected function getParameter($name)` function but we could add it into our `FrameworkBundleAdminController`

Maybe we could do it by making FrameworkBundleAdminController extend AbstractController. We must keep the inheritance because else it's a BC break.

Previous code
```
$isAController = ($myController instanceof FrameworkBundleAdminController); // returns true
```
After BC break
```
$isAController = ($myController instanceof FrameworkBundleAdminController); // returns false
```

jolelievre:
Yes, that's what I thought of course (sorry it wasn't clear 😅) which is why I compared `AbstractController` and `Controller` because for now `FrameworkBundleAdminController` extends the latter

Si if we make it extends `AbstractController` it will need the protected `getParameter` function

app/config/config.yml
mickaelandrieu: yes, at least according to the documentation: I can't be sure it works in your case, maybe a micro module with autowired args in a controller maybe be proposed ? :)

matks: This PR will be closed as a better solution has be found. Tag will be added in https://github.com/PrestaShop/PrestaShop/pull/18845

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
matks: Ah, in fact I wont do it because I reuse `$definition` later in the code. So code is good.

matks:
@jolelievre So looks like we'll keep this solution. Do you approve it ? If yes, I will provide a demo module for QA to test.

PierreRambaud (dismissed):

matthieu-rolland (dismissed):

matks:
Test case added.

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
```diff
@@ -0,0 +1,70 @@
...
+use InvalidArgumentException;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Decorated controllers using ControllerAwareTrait
+ * cannot be injected the container as the injection is performed by ControllerResolver.
```
jolelievre: This short first line is weird ^^ the line break could have been a bit later for readability

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * This pass injects the container into PrestaShop controllers to overcome this issue.
+ */
+class ContainerAwareControllersManualInjectionPass implements CompilerPassInterface
```
jolelievre: Class name a bit long, besides it's not really "manual" since it's automated via a compiler pass I would suggest something like `ContainerInjectionPass` or `ControllersContainerInjectionPass` 🤔 Although I'm not 100% satisfied about these one either 😅

jolelievre (dismissed):
Only cosmetic comments, but besides that it's great!

matks:
> @jolelievre So looks like we'll keep this solution. Do you approve it ? If yes, I will provide a demo module for QA to test.

# For the record

We considered another solution which would have been `FrameworkBundleAdminController` extends Symfony base `Controller`.

It can be done although a bit complex because there are 2 `ContainerInterface` being used and we want to keep both to avoid a BC break:
- `Symfony\Component\DependencyInjection\ContainerInterface`
- `Psr\Container\ContainerInterface`

It works fine for Core controllers but it does not for Module controllers.
As it's a complex part of Symfony, I could not find the exact reason.

My guess would be that, when I decorate a Core controller with a Module controller, the controller service is somehow "moved" from core namespace, which is autowired, to module namespace, which is not autowired. The consequence is that Symfony cannot inject the container using `setContainer()` and we circle back to the initial bug usecase.

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
```diff
@@ -0,0 +1,70 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
eternoendless: Missing declare strict types

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
eternoendless: Try to avoid assignments in ifs, but if you do it, please try and use meaningful variable names.

In this case, you can use `$reflectedClass`

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
```diff
@@ -0,0 +1,70 @@
...
+            if (!$r = $container->getReflectionClass($class)) {
+                throw new InvalidArgumentException(sprintf(
+                    'Class "%s" used for service "%s" cannot be found.',
+                    $class,
+                    $id
+                ));
+            }
+            $isContainerAware = $r->implementsInterface(ContainerAwareInterface::class)
+                || is_subclass_of($class, AbstractController::class);
```
eternoendless: Add parentheses for clarity:

```suggestion
            $isContainerAware = (
                $r->implementsInterface(ContainerAwareInterface::class)
                || is_subclass_of($class, AbstractController::class)
            );
```

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
```diff
@@ -0,0 +1,70 @@
...
+ * This pass injects the container into PrestaShop controllers to overcome this issue.
+ */
+class ContainerAwareControllersManualInjectionPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $controllers = $container->findTaggedServiceIds('prestashop.core.controllers');
```
eternoendless: Consider extracting `'prestashop.core.controllers'` into a constant.

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
eternoendless: What about `ContainerInjectionPass`?

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
eternoendless: This phrase must be reworded because it makes no sense 😄 

> Decorated controllers using ControllerAwareTrait cannot be injected the container as the injection is performed by ControllerResolver.

src/PrestaShopBundle/DependencyInjection/Compiler/ContainerAwareControllersManualInjectionPass.php
matks: Done into `FrameworkBundleAdminController::PRESTASHOP_CORE_CONTROLLERS`

matks:
@jolelievre @eternoendless I have updated the PR following your feedback

app/config/config.yml
```diff
@@ -21,7 +21,11 @@ services:
     PrestaShopBundle\Controller\:
         resource: "%kernel.root_dir%/../src/PrestaShopBundle/Controller/*"
         exclude: "%kernel.root_dir%/../src/PrestaShopBundle/Controller/Api"
-        tags: ['prestashop.core.controllers']
+        tags:
+          - 'controller.service_arguments'
+          - !php/const PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController::PRESTASHOP_CORE_CONTROLLERS
```
jolelievre:
```suggestion
          - !php/const PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController::PRESTASHOP_CORE_CONTROLLERS_TAG
```
So we understand what it stands for (it's not a service name for example)

jolelievre (commented):
Last comment, but it's great to use this const everywhere!
Great job 😉

app/config/config.yml
jolelievre: Oh and you have conflict on this file by the way, and there's an extra line you can remove maybe

matks: Done in https://github.com/PrestaShop/PrestaShop/pull/18811/commits/e8ed3b04116ea196210b185283e373fba6cc998a

jolelievre (approved):
Thank you @matks

Progi1984 (approved):

matthieu-rolland (approved):

SD1982:
LGTM Thanks @matks !

matks:
Thank you for review and QA !

Issue #14753, opened 2019-07-18
Add internal order note


Hlavtox:
**Is your feature request related to a problem? Please describe.**
Employees often need to write down some information to a single order, for them and other employees to know, what's the status of the order, because order status is not explanatory enough. 

Customer note is not very useful for this. When customer orders something for a second time, you see the previous note and could confuse you a lot. Customer note is only good for entering "good customer"/"asshole customer".

**Describe the solution you'd like**
Add a field with internal note on the order page.

khouloudbelguith:
Hi @Hlavtox,

Thanks for your report.
Ping @colinegin, @marionf what do you think?

Thanks!

MatShir:
@Hlavtox,
This is something to be done! but for now, we will just put a clearer wording for the private message to be used as an internal order note. It will integrate into a 2nd part of the order project after the 177.

kmorgen:
Noticed the new order note field available in PS 1.7.8 which is fine. 
However it is filling up and getting a bit confusing with now 3 internal notes fields?!

If one is for general customer notes (being visible on all orders from this customer), and the new one is a order note (only visible for the current order), then what is the purpose of having a check box in the customer message field too?

Maybe better to remove this check box option, as the purpose of this section is for communicating with the customer. I guess in many cases one might forget to check this box, which mean the customer will never get the message.

![Screenshot](https://user-images.githubusercontent.com/4300688/141971993-ecef06d2-9f6e-4f71-85b2-aa1098b8ecb0.jpg)

MatShir:
It makes sense @kmorgen, if you have the technical knowledge or find someone that could don't hesitate to propose a pull request ([find more information here](https://devdocs.prestashop.com/1.7/contribute/contribute-pull-requests/)). We will be happy to review your contribution to the PrestaShop core.

PR #18825, opened 2020-04-26
Private Note Orders

JavierEB:
Prestashop always, in the private note of customer orders, is not associated with the order itself ... it is associated with the customer. With which it would be a good option to associate only that private note within the orders only to them. For example: a customer has multiple orders, but only one needs xx $ and that note may be only for that order, not for the customer itself ... I don't know if you understand me ... Attached photos of the orders, to all those of the same client are associated with the same "test note"
I think that is a problem. Thank you
Screenshot_2020-04-26 Orders Order FFATNOMMJ from John DOE • Prueba1764
Screenshot_2020-04-26 Orders Order KHWLILZLL from John DOE • Prueba1764
Screenshot_2020-04-26 Information about customer J DOE • Prueba1764

| Questions     | Answers
| Description?  | Please be specific when describing the PR. <br> Every detail helps: versions, browser/server configuration, specific module/theme, etc. Feel free to add more information below this table.
| Type?         | new feature
| Category?     | BO 
| BC breaks?    |  no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/14753
| How to test?  | any version of prestashop

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18825)


Progi1984:
@JavierEB Thank you for your contribution. :hugs: 

Thanks :+1:

matthieu-rolland:
@JavierEB it seems like your PR is merging develop into 1.7.6.x, I'm pretty sure you don't want to do that :D

JavierEB:
> 
> 
> @JavierEB Thank you for your contribution. hugs
> 
> Could you explain in [an issue your initial bug for this PR](https://github.com/PrestaShop/PrestaShop/issues/new/choose) (or may be link to an issue) ?
> Could you give us a scenario which permits to test the bug before and after this PR ?
> 
> Thanks +1

hi, the problem is with all versions of PS. This is a problem that comes from behind I think ...
Because the private note of the orders is actually saved in the table ps_costumers> Column "Note" with which it is associated with all the orders of that customer, not just one.

JavierEB:
> 
> 
> @JavierEB Thank you for your contribution. hugs
> 
> Could you explain in [an issue your initial bug for this PR](https://github.com/PrestaShop/PrestaShop/issues/new/choose) (or may be link to an issue) ?
> Could you give us a scenario which permits to test the bug before and after this PR ?
> 
> Thanks +1

hi, the problem is with all versions of PS. This is a problem that comes from behind I think ...
Because the private note of the orders is actually saved in the table ps_costumers> Column "Note" with which it is associated with all the orders of that customer, not just one.

Progi1984:
@JavierEB Please create an issue which be prioritized thanks our @PrestaShop/prestashop-product-team

JavierEB:
>  Please create an issue which be prioritized thanks our @PrestaShop/prestashop-product-team

ok thanks

MatShir:
Hi @JavierEB,
There are 2 needs here:
- One where we want to have a concise description of the customer (for big client, to save customer behavior/additional information). It is very useful for B2B cases for example.
- One where we want to have a concise description of the order status

The first one is implemented with a private note. 

But the second one is kind of done with the private message which can be only be seen by the merchant in this particular order. 
It is not ideal for the merchant to look to the private messages to process the order or to see quickly notable information.

Instead of editing the feature of customer note, I suggest adding new input for this. I will ask our designer to recommend the ideal implementation for this feature on this order page.
@PrestaShop/prestashop-designers

PierreRambaud:
Sorry @JavierEB , this PR does not include any contribution from your account. This kind of merge should be done manually in order to fix the existing conflicts.
I'm closing it as it can't be accepted as-is.

If you have a feature request or a bug please create an issue https://github.com/PrestaShop/PrestaShop/issues/new/choose

Kind regards

PR #18871, opened 2020-04-28
Functional tests - Update default currency

nesrineabdmouleh:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Update default currency
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  |  TEST_PATH="functional/BO/11_international/01_localization/04_defaultCurrency.js" URL_FO=yourShopURL npm run specific-test

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18871)

### DIFF
--- a/tests/puppeteer/campaigns/functional/BO/11_international/01_localization/04_defaultCurrency.js
+++ b/tests/puppeteer/campaigns/functional/BO/11_international/01_localization/04_defaultCurrency.js
@@ -135,7 +135,7 @@ describe('Update default currency', async () => {
         this.pageObjects = await init();
       });
 
-      if (index === 1) {
+      if (index === (tests.length - 1)) {
         describe('Delete currency added by importing localization pack', async () => {
           it('should go to currencies page', async function () {
             await testContext.addContextItem(this, 'testIdentifier', 'goToCurrenciesPage', baseContext);
--- a/tests/puppeteer/campaigns/functional/BO/11_international/01_localization/04_defaultCurrency.js
+++ b/tests/puppeteer/campaigns/functional/BO/11_international/01_localization/04_defaultCurrency.js
@@ -1,0 +1,126 @@
+require('module-alias/register');
+// Using chai
+const {expect} = require('chai');
+const helper = require('@utils/helpers');
+const loginCommon = require('@commonTests/loginBO');
+// Importing pages
+const BOBasePage = require('@pages/BO/BObasePage');
+const LoginPage = require('@pages/BO/login');
+const DashboardPage = require('@pages/BO/dashboard');
+const LocalizationPage = require('@pages/BO/international/localization');
+const CurrenciesPage = require('@pages/BO/international/currencies');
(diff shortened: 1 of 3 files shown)


tests/puppeteer/campaigns/functional/BO/11_international/01_localization/04_defaultCurrency.js
```diff
@@ -0,0 +1,169 @@
...
+        await testContext.addContextItem(this, 'testIdentifier', `checkCurrencyInFO${index}`, baseContext);
+        page = await this.pageObjects.boBasePage.viewMyShop();
+        this.pageObjects = await init();
+        const defaultCurrency = await this.pageObjects.foBasePage.getDefaultCurrency();
+        await expect(defaultCurrency).to.equal(test.args.currency);
+        page = await this.pageObjects.foBasePage.closePage(browser, 1);
+        this.pageObjects = await init();
+      });
+
+      if (index === 1) {
```
SimonGrn: What were you trying to do here ? Is there a better way to handle this ?

Progi1984 (dismissed):

tests/puppeteer/campaigns/functional/BO/11_international/01_localization/04_defaultCurrency.js
nesrineabdmouleh: The condition is used to do the deletion of the currency during the last iteration, I have no solution because the browser will be closed, so the describe should be inside the forEach method. 
What do you think ?

NeOMakinG (dismissed):

tests/puppeteer/campaigns/functional/BO/11_international/01_localization/04_defaultCurrency.js
SimonGrn: This section must be run only for the *last* currency, or for *every currency but the first* ?

nesrineabdmouleh: The section must be run at the end of the test, it's not related to currency.

SimonGrn:
```suggestion
      if (index === (tests.length - 1)) {
```
Maybe you should use this, to make sure it doesn't break if you add a currency in the `tests` objects.
Please test because I don't remember if `index` starts at 0 or 1 (Javascript....).

nesrineabdmouleh: index starts at 0
I will fix it :) thank you

Progi1984 (approved):

NeOMakinG (approved):

PierreRambaud (approved):

SimonGrn (approved):

PR #18876, opened 2020-04-28
Remove array to string conversion in bulk delete address faulty usecase

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | If bulk delete address operation fails, then BulkDeleteAddressException is thrown but it is badly being handled by controller and it creates an "array to string conversion". I fix it, a `implode()` was missing
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Hard to provide a QA scenario 😄as it requires something to go wrong with delete operation for adresses. Can we say no QA is needed as it's a very simple fix ? A developer can test this by modifying BulkDeleteAddressHandler ligne 62 to force the throwing of BulkDeleteAddressException (and see that, without the PR, the page crashes like screenshot below)

Capture d’écran 2020-04-28 à 16 54 23

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18876)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
@@ -482,7 +482,7 @@ class AddressController extends FrameworkBundleAdminController
                     'An error occurred while deleting this selection.',
                     'Admin.Notifications.Error'
                 ),
-                $e instanceof BulkDeleteAddressException ? $e->getAddressIds() : ''
+                $e instanceof BulkDeleteAddressException ? implode(', ', $e->getAddressIds()) : ''
             ),
             AddressNotFoundException::class => $this->trans(
                 'The object cannot be loaded (or found)',


rokaszygmantas (approved):

Progi1984 (approved):

PierreRambaud (approved):

matks:
@PierreRambaud This error only happens when something wrong happens in ObjectModel (maybe trying to delete something already deleted ?) and it throws exception. I dont know how to provide test case for QA, I tested it by manually modifying code and throwing exception. Can we skip QA for this ?

PierreRambaud:
@matks @PrestaShop/prestashop-qa 
Scenario to test:
- Go to addresses list
- Select all addresses
- Remove all addresses in the database `TRUNCATE ps_address;`
- Go back to browser and execute the bulk action

Is it ok?

Robin-Fischer-PS:
What should happen when executing the bulk action @PierreRambaud ? Nothing, or an error should pop ?

matks:
> What should happen when executing the bulk action @PierreRambaud ? Nothing, or an error should pop ?

The big red error I put in the PR description 😉

SD1982:
LGTM Thanks @matks !

matks:
Thank you for review and QA

PR #18904, opened 2020-04-29
FO: show browser-native error messages during payment selection step in checkout

### LINKED ISSUE #17812: Checkout: Payment Step: Missing error message when terms not accepted
Using a sligtly modified classic theme. 

**To Reproduce**
Put some article in the cart.
Proceed to checkout, finish all steps (persoanl info, adress, shipping) however you want.
Choose a payment method.
Click "Order" Button which is disabled since the "I agree to the terms..:" Checkbox wasn't clicked.

-> There should be a tooltip / error message, informing that the "terms and services" checkbox should be ticked in order to proceed.

**Screenshots**
image

**Additional information**
PrestaShop version: 1.7.6.2
PHP version: 7.2

patrickmyh:
| Questions     | Answers
| Branch?       | develop
| Description?  | Cf **What this change is about**
| Type?         | improvement
| Category?     | FO
| Deprecations? | no
| Fixed ticket? | Fixes #17812
| How to test?  | Cf **How to test**

### What this change is about

During the checkout process, in the payment step (the final step), the confirmation button "place order" was disabled.
It was only enabled when 

- a payment method was chosen, and
- the terms and conditions checkbox were accepted (checkbox checked)

If the user tried to click the button before doing so, nothing happened.
I found that there there should be an error message popping up, informing that some action was still required by the user in order to proceed.
I even could track some users that had trouble finding the missing step, scrolling up and down, clicking the disabled submit button repeatedly.

### How to test

In the Payment-stage of the checkout flow...

Click the submit button with no payment-method selected.  
-> There should be an error message next to the input element, informing that it is needed to chose one to proceed.

Click the submit button without having agreed to the terms and conditions
-> There should be an error message next to the checkbox, informing that it is needed to check the box in order to proceed.

### How it was done

Since other steps in the checkout, e.g. the "personal information" step makes use of the native browser form validation, the same method for displaying errors should be used. 

Since said elements are not in the same form DOM-wise, this was a little tricky.
I traversed the input elements and called the reportValidity() method on them.

Also, the button for submitting the form had the "disabled" attribute, prventing and events to be catched on it. So I used the "disabled" class from bootstrap instead. Still, there is a proper check of the input element's state before the submit is fired.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18904)

### DIFF
--- a/themes/_core/js/checkout-payment.js
+++ b/themes/_core/js/checkout-payment.js
@@ -53,9 +53,7 @@ class Payment {
   }
 
   haveTermsBeenAccepted() {
-    let termsCheckbox = $(this.termsCheckboxSelector);
-
-    return termsCheckbox.prop('checked');
+    return $(this.termsCheckboxSelector).prop('checked');
   }
 
   hideConfirmation() {
@@ -114,11 +112,8 @@ class Payment {
   }
 
   showNativeFormErrors () {
-    let termsCheckbox = $(this.termsCheckboxSelector)
-    termsCheckbox.get(0).reportValidity();
-
-    let paymentOptionsGroup = $('input[name=payment-option]')
-    paymentOptionsGroup.get(0).reportValidity();
+    $(this.termsCheckboxSelector).get(0).reportValidity();
+    $('input[name=payment-option]').get(0).reportValidity();
   }
 
   confirm() {
--- a/themes/_core/js/checkout-payment.js
+++ b/themes/_core/js/checkout-payment.js
@@ -37,15 +37,15 @@ class Payment {
   init() {
     let $body = $('body');
 
-    $body.on('change', this.conditionsSelector + ' input[type="checkbox"]', $.proxy(this.toggleOrderButton, this));
+    $body.on('change', `${this.conditionsSelector} input[type="checkbox"]`, $.proxy(this.toggleOrderButton, this));
     $body.on('change', 'input[name="payment-option"]', $.proxy(this.toggleOrderButton, this));
-    $body.on('click', this.confirmationSelector + ' button', $.proxy(this.confirm, this));
+    $body.on('click', `${this.confirmationSelector} button`, $.proxy(this.confirm, this));
 
     this.collapseOptions();
   }
 
   collapseOptions() {
-    $(this.additionalInformatonSelector + ', ' + this.optionsForm).hide();
+    $(`${this.additionalInformatonSelector}, ${this.optionsForm}`).hide();
   }
 
   getSelectedOption() {
@@ -66,7 +66,7 @@ class Payment {
 
   toggleOrderButton() {
     var show = true;
-    $(this.conditionsSelector + ' input[type="checkbox"]').each((_, checkbox) => {
+    $(`${this.conditionsSelector} input[type="checkbox"]`).each((_, checkbox) => {
       if (!checkbox.checked) {
         show = false;
       }
@@ -79,11 +79,11 @@ class Payment {
       show = false;
     }
 
-    $('#' + selectedOption + '-additional-information').show();
-    $('#pay-with-' + selectedOption + '-form').show();
+    $(`#${selectedOption}-additional-information`).show();
+    $(`#pay-with-${selectedOption}-form`).show();
 
     $('.js-payment-binary').hide();
-    if ($('#' + selectedOption).hasClass('binary')) {
+    if ($(`#${selectedOption}`).hasClass('binary')) {
       var paymentOption = this.getPaymentOptionSelector(selectedOption);
       this.hideConfirmation();
       $(paymentOption).show();
@@ -95,7 +95,7 @@ class Payment {
       }
     } else {
       this.showConfirmation();
-      $(this.confirmationSelector + ' button').toggleClass('disabled', !show);
+      $(`${this.confirmationSelector} button`).toggleClass('disabled', !show);
 
       if (show) {
         $(this.conditionAlertSelector).hide();
@@ -126,8 +126,8 @@ class Payment {
       return;
     }
 
-    $(this.confirmationSelector + ' button').addClass('disabled');
-    $('#pay-with-' + option + '-form form').submit();
+    $(`${this.confirmationSelector} button`).addClass('disabled');
+    $(`#pay-with-${option}-form form`).submit();
   }
 }
 
--- a/themes/_core/js/checkout-payment.js
+++ b/themes/_core/js/checkout-payment.js
@@ -112,13 +112,14 @@ class Payment {
   }
 
   showNativeFormErrors () {
-    $(this.termsCheckboxSelector).get(0).reportValidity();
-    $('input[name=payment-option]').get(0).reportValidity();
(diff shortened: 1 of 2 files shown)


themes/_core/js/checkout-payment.js
```diff
@@ -54,6 +52,12 @@ class Payment {
     return $('input[name="payment-option"]:checked').attr('id');
   }
 
+  haveTermsBeenAccepted() {
+    let termsCheckbox = $(this.termsCheckboxSelector);
+
+    return termsCheckbox.prop('checked');
```
Progi1984:
```suggestion
    return $(this.termsCheckboxSelector).prop('checked');
```

themes/_core/js/checkout-payment.js
```diff
@@ -109,12 +113,26 @@ class Payment {
     return `.js-payment-${moduleName}`;
   }
 
+  showNativeFormErrors () {
+    let termsCheckbox = $(this.termsCheckboxSelector)
+    termsCheckbox.get(0).reportValidity();
+
+    let paymentOptionsGroup = $('input[name=payment-option]')
+    paymentOptionsGroup.get(0).reportValidity();
```
Progi1984:
```suggestion
    $(this.termsCheckboxSelector).get(0).reportValidity();
    $('input[name=payment-option]').get(0).reportValidity();
```

Progi1984 (commented):
Some changes

themes/_core/js/checkout-payment.js
```diff
@@ -109,12 +111,23 @@ class Payment {
...
+    let termsAccepted = this.haveTermsBeenAccepted();
+
+    if (option === undefined || termsAccepted === false) {
+      this.showNativeFormErrors();
+
+      return;
     }
+
+    $(this.confirmationSelector + ' button').addClass('disabled');
+    $('#pay-with-' + option + '-form form').submit();
```
NeOMakinG: This is processed by webpack right ?

Would it be possible to do this at the same time ?

```suggestion
    $(`${this.confirmationSelector} button`).addClass('disabled');
    $(`#pay-with-${option}-form form`).submit();
```

patrickmyh: it's successfully processed by weback.
did so and also changed the other concatenations

Progi1984 (dismissed):

NeOMakinG (dismissed):
Thanks :)

themes/_core/js/checkout-payment.js
```diff
@@ -109,12 +111,23 @@ class Payment {
     return `.js-payment-${moduleName}`;
   }
 
+  showNativeFormErrors () {
+    $(this.termsCheckboxSelector).get(0).reportValidity();
+    $('input[name=payment-option]').get(0).reportValidity();
```
PierreRambaud:
```suggestion
$(`input[name=payment-option], ${this.termsCheckboxSelector}`).each(function() {
  this.reportValidity();
});
```
Prefer this notation please, modules can add inputs

themes/_core/js/checkout-payment.js
```diff
@@ -109,12 +111,23 @@ class Payment {
...
+  }
+
   confirm() {
-    var option = this.getSelectedOption();
-    if (option) {
-      $(this.confirmationSelector + ' button').prop('disabled', true);
-      $('#pay-with-' + option + '-form form').submit();
+    let option = this.getSelectedOption();
+    let termsAccepted = this.haveTermsBeenAccepted();
```
PierreRambaud:
```suggestion
    const option = this.getSelectedOption();
    const termsAccepted = this.haveTermsBeenAccepted();
```

themes/_core/js/checkout-payment.js
patrickmyh: Agree to this notation and will do the change.

About the inputs added by modules: 
The terms checkbox is one single element selected by it's unique name. So later added checkobxes wouldn't be selected anyway.

Additionally added radio buttons for payment methods should belong to the same radio button group, shouldn't they? So calling "reportValidity()" on one group member should apply to the whole group how I understand it.

PierreRambaud: You can chain radio buttons, and it breaks at the first one falling on error when using reportValidity.
Since it's a radio, it only occurs when the radio have the required attribute.

PierreRambaud (approved):

NeOMakinG (approved):

matks:
Thank you @patrickmyh

matks:
> This commit breaks compatibility with themes for Ps < 1.7.8
> Use of "disabled" class instead of "disabled" attribute on the payment button.
> This forces to maintain multiple versions of a theme, any idea how to deal with it? (poke @NeOMakinG)

Hello, as we have no clear definition of the public API of PrestaShop, indeed you could say this breaks compatibility. But this is a very minor break.

Can you provide a snippet of code that does not work with the above changes?

PrestaEdit:
Hi @matks , @Progi1984  and @NeOMakinG ; it's a "minor break" but that break a whole theme (and surely not this one ! :-/) due do the disabled attr changed with a class, here. A real break change.

ghost:
Hello,
It's quite a break as Customers can not place order anymore.
As a workaround, I edited my theme template in order to get the button active again

`<button type="submit" {if !$selected_payment_option} disabled {/if}
                class="btn btn-primary center-block btn-general-bg btn-general-color">
                {l s='Order with an obligation to pay' d='Shop.Theme.Checkout'}
            </button>`

is now :
`<button type="submit"
                class="btn btn-primary center-block btn-general-bg btn-general-color {if !$selected_payment_option}disabled{/if}">
                {l s='Order with an obligation to pay' d='Shop.Theme.Checkout'}
            </button>`

matks:
Hi everyone, thanks for the feedback ! Just a few things to clarify:

> Hi @matks , @Progi1984 and @NeOMakinG ; it's a "minor break" but that break a whole theme (and surely not this one ! :-/) due do the disabled attr changed with a class, here. A real break change.

It does not break the whole theme, it breaks the checkout process.

When I label this a 'minor break', this is because BC breaks can be as big as removing files or classes or as small as adding one parameter. 'Minor' is not for the impact, it's for the size of the change.

Also note that the word 'break' here is not related to something being a broken in the software, but related to the backward compatibility being not possible on this very change. 😉 

Anyway I see an issue was open on https://github.com/PrestaShop/PrestaShop/issues/26274 👍 it will be more useful to discuss how to handle this topic on this issue rather on this PR. Comments on this PR will probably not be read while comments on the issue will be more easily found.

paulmohor:
> Hello, It's quite a break as Customers can not place order anymore. As a workaround, I edited my theme template in order to get the button active again
> 
> `<button type="submit" {if !$selected_payment_option} disabled {/if} class="btn btn-primary center-block btn-general-bg btn-general-color"> {l s='Order with an obligation to pay' d='Shop.Theme.Checkout'} </button>`
> 
> is now : `<button type="submit" class="btn btn-primary center-block btn-general-bg btn-general-color {if !$selected_payment_option}disabled{/if}"> {l s='Order with an obligation to pay' d='Shop.Theme.Checkout'} </button>`

work for me, thank's

PR #18911, opened 2020-04-30
FO: Refresh checkout hook

idnovate:
| Questions     | Answers
| Branch?       | develop
| Description?  | Refresh `displayCheckoutSummaryTop` hook when any action is done in the checkout, like adding/removing a voucher
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Develop a module hooked at displayCheckoutSummaryTop. Display current timestamp, for example. Apply a voucher. This area should be updated with a new timestamp.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18911)


controllers/front/CartController.php
```diff
@@ -182,6 +182,7 @@ public function displayAjaxRefresh()
             'cart_summary_totals' => $this->render('checkout/_partials/cart-summary-totals'),
             'cart_detailed_actions' => $this->render('checkout/_partials/cart-detailed-actions'),
             'cart_voucher' => $this->render('checkout/_partials/cart-voucher'),
+            'hook_checkout_summary_top' => Hook::exec('displayCheckoutSummaryTop')
```
Progi1984:
```suggestion
            'hook_checkout_summary_top' => Hook::exec('displayCheckoutSummaryTop'),
```

Progi1984 (approved):

PierreRambaud (approved):

SD1982:
Hello @idnovate Can you please  provide us  a compatible module so that we can test your PR ?

idnovate:
@SD1982 
[displaycheckoutsummarytop.zip](https://github.com/PrestaShop/PrestaShop/files/4656692/displaycheckoutsummarytop.zip)

SD1982:
Hi @idnovate  I have an error when trying to install the module that you provided me : 
https://drive.google.com/file/d/1yl7-jpy6ljmKu8loULfCeqmf347IVdEN/view?usp=sharing

Can you check this please ?

idnovate:
My apologies, it was a case sensitive problem.

[displaycheckoutsummarytop.zip](https://github.com/PrestaShop/PrestaShop/files/4751091/displaycheckoutsummarytop.zip)

SD1982:
Hi @idnovate  The time stamp is correctly updated for the first applied discount but if we apply a second voucher it's not updated 
https://drive.google.com/file/d/1GfapdvsBWPDKmIfJoSuFmY0JzwAN_8U7/view?usp=sharing

idnovate:
Ok I see! The problem is that `<div class="hook_checkout_summary_top">` https://github.com/PrestaShop/PrestaShop/pull/18911/files#diff-463a9f63ea088495dd06250ef7cbaec6R27 is removed when executing JS https://github.com/PrestaShop/PrestaShop/pull/18911/files#diff-cc1024c22d27236739edaef574de3d9bR61 with the new content. 
Then when a new request is done, JS doesn't find the selector and is not replaced again, only once.

The solution would be to create a new file like https://github.com/PrestaShop/PrestaShop/blob/b2a7e9e448b22ec174fcb0e907b4276b6f9abcf9/themes/classic/templates/checkout/_partials/cart-summary.tpl#L42-L44 with this content:

```
<div class="hook_checkout_summary_top">
    {hook h='displayCheckoutSummaryTop'}
</div>
```

Is it ok if I name this file as `checkout/_partials/cart-hook-checkout-summary-top.tpl`?

SD1982:
> Is it ok if I name this file as `checkout/_partials/cart-hook-checkout-summary-top.tpl`?

@PrestaShop/prestashop-core-developers What do you think ?

matthieu-rolland:
@idnovate I would not necessarily mention "hook" in the template name, `cart-checkout-summary-top.tpl` should be fine, do you agree @Progi1984 , @PierreRambaud  ?

PierreRambaud (changes requested):
Just notice the class name should be `checkout-summary-top`, my bad, could you change it please?

idnovate:
@PierreRambaud 

All partial templates and block names start with `cart` but hook is called `displayCheckoutSummaryTop`.

What do you think about `cart-summary-top`?

PierreRambaud:
> @PierreRambaud
> 
> All partial templates and block names start with `cart` but hook is called `displayCheckoutSummaryTop`.
> 
> What do you think about `cart-summary-top`?

Let's go for `cart-summary-top` :+1:

idnovate:
https://github.com/PrestaShop/PrestaShop/pull/19762

PR #18960, opened 2020-05-04
Wrong shop id at creating new specific price rules in BO

Markus-Gurkcity:
| Questions     | Answers
| Branch?       | develop
| Description?  | At a standard prestashop installation in non-multishop context the shop id is 1. But if you have a non-multishop installation where the shop id is not 1 this will not work. The static shop id has to be replaced to the real shop id from the shop context
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Possible test szenario:<br>let's say we have individual prestashops on individual servers. every shop itself has the shop id 1. we now want to combine the individual shops with a merchandise management system. the merchandise management system should now be able to differentiate between shops based on shop id. So we have to change the shop id for all shops. the first shop has the shop id 1, the second one gets the shop id 2, and so on... You can achieve this by changing all database fields that contain a shop ID.<br><br>Another way to test it:<br>Make a prestashop release where all fixtures for the installation have the shop id 2 (or 3...)<br><br>Here is a example of the shop table with a different shop id. please note that it is not a multishop:<br>image<br><br>all in all, there should be no static shop id in the source code, but the shop id should either come from the context or the configuration PS_SHOP_DEFAULT.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/18960)

### DIFF
--- a/admin-dev/themes/default/template/controllers/specific_price_rule/helpers/form/form.tpl
+++ b/admin-dev/themes/default/template/controllers/specific_price_rule/helpers/form/form.tpl
@@ -139,7 +139,7 @@
 		</div>
 	</div>
 {if !$is_multishop}
-	<input type="hidden" name="id_shop" value="1" />
+	<input type="hidden" name="id_shop" value="{$shop->id}" />
 {/if}
 </div>
 {/block}


PierreRambaud (approved):

Progi1984 (approved):

Progi1984:
@Markus-Gurkcity How can it be possible to test your pull request ? Thanks for the contribution.

Markus-Gurkcity:
Possible test szenario:
let's say we have individual prestashops on individual servers. every shop itself has the shop id 1. we now want to combine the individual shops with a merchandise management system. the merchandise management system should now be able to differentiate between shops based on shop id. So we have to change the shop id for all shops. the first shop has the shop id 1, the second one gets the shop id 2, and so on... You can achieve this by changing all database fields that contain a shop ID.

Another way to test it:
Make a prestashop release where all fixtures for the installation have the shop id 2 (or 3...)

Here is a example of the shop table with a different shop id. please note that it is not a multishop:
image

all in all, there should be no static shop id in the source code, but the shop id should either come from the context or the configuration PS_SHOP_DEFAULT.

SimonGrn:
Hello @Markus-Gurkcity ,

thank you for your contribution. In order for the QA team to review your PR, we need an issue linked or a clear description of the problem: we need to be able to reproduce the original issue you're trying to fix here.
Your "how to test" doesn't really tell us what is wrong and what we should look for to check what was broken and what is now fixed. Could you please elaborate ?

Thanks in advance !

PierreRambaud:
Hi @Markus-Gurkcity  ,

Thanks!

PR #17706, opened 2020-02-18
Improve modern module installation

jolelievre:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | This PR adds few missing features that were lacking during module installation and caused bugs if you tried to use some modern features during the installation process (because they were only enabled on installed modules):<br><br>- include autoload file (if present)<br>- load Doctrine entities from module folder<br>- load services (admin and generic)<br>- access the correct container (Symfony container is always available during an installation)<br>- allows module to use doctrine schema update (for their own entities only)
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14527
| How to test?  | Start by installing this branche, then you can use this productcomments branch, it uses Doctrine entities, module services, and the doctrine updater https://github.com/PrestaShop/productcomments/pull/51<br>Once you have the branch installed in your modules folder you can run these two commands:<br>`php ./bin/console prestashop:module uninstall productcomments`<br>`php ./bin/console prestashop:module install productcomments`<br>You can also check the command with the `ps_linklist` module (no branch needed)<br>`php ./bin/console prestashop:module uninstall ps_linklist`<br>`php ./bin/console prestashop:module install ps_linklist`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/17706)

PR #19016, opened 2020-05-06
Add mapping for Doctrine core entities on front environment

### LINKED ISSUE #19015: Doctrine Entities from Core are inaccessible in front office
#### Describe the bug

When trying to access the core Doctrine entities (like `PrestaShopBundle\Entity\Lang`) in front office an error is thrown:

Capture d’écran 2020-05-06 à 18 36 48

#### Expected behavior

The core entities should be usable in Front office, this would help (amongts other thins) to manage some multi lang Doctrine entities. The problem is that in FrontOffice the container is a light one built by the core it is based on those two classes:

https://github.com/PrestaShop/PrestaShop/blob/1.7.6.5/src/Adapter/ContainerBuilder.php
https://github.com/PrestaShop/PrestaShop/blob/1.7.6.5/src/Adapter/Container/DoctrineBuilderExtension.php

So the extension correctly scans the modules and add their namespaces in Doctrine, but it doesn't do it for core entities from the `PrestaShopBundle\Entity\` namespace. So we need to add a dedicated CompilerPass to add this namespace, inspired from `ModulesDoctrineCompilerPass`.

#### Steps to Reproduce

You will need the example module from this PR https://github.com/PrestaShop/example-modules/pull/20

1. Clone the branch
2. Only copy (or link) the `demodoctrine` folder into your `modules` folder
3. Install the module
4. Configure the module via Module manager
5. Since no data is present you should redirected to the Generate quotes page
6. Click on `Generate quotes` button, the tables and the demo quotes should be installed
7. Go to the front office homepage, you should see the error regarding the `Lang` entity

#### Additional information

* PrestaShop version: 1.7.6
* PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | The Doctrine core entities were not usable in the front container (built by PrestaShop core). This PR fixes this issue by bringing a few chances:<br>- extract doctrine config into dedicated yaml<br>- introduce new YAML config files dedicated to the front<br>- introduce `ConfigYamlLoader` class to load this config in `DoctrineExtension`<br>- integrates a fix originally coming from this PR https://github.com/PrestaShop/PrestaShop/pull/17706 which fixes `Module::get` and new methods like `Module::hasSymfonyContainer` and `Module::hasLegacyContainer` to better detect the environment/container
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19015
| How to test?  | See the issue to reproduce the bug with the example module, with this PR the entities now work correctly and the quotes are displayed on the home page (you should remove your cache after switching to the PR as the container needs to be rebuilt)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19016)

Note: To test this PR you will need this example module https://github.com/PrestaShop/example-modules/pull/20

# Install instruction

- `clone` the demomodule in your `modules` folder (be careful that the repository contains multiple folder, you only need the `demodoctrine` folder)
- `checkout` the appropriate branch for the module `demodoctrine`
- run `composer install` in the module folder
- in the *PrestaShop* root folder make sure you are on the correct branch `doctrine-core-entities`
- run the `rm -fR var/cache/*` command to be sure all cache is empty
- you should now be able to install the module without issue (in the BO or using the CLI command `php bin/console prestashop:module install demodoctrine`


src/PrestaShopBundle/DependencyInjection/Compiler/CoreDoctrineCompilerPass.php
```diff
@@ -0,0 +1,78 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
```
Progi1984: Two things : the copyright date & `declare(strict_types=1);`

Progi1984 (dismissed):

src/PrestaShopBundle/DependencyInjection/Compiler/CoreDoctrineCompilerPass.php
PierreRambaud: It's for the 1.7.6.x we can't use the strict types

src/PrestaShopBundle/DependencyInjection/Compiler/CoreDoctrineCompilerPass.php
```diff
@@ -0,0 +1,78 @@
...
+     * whole process was stopped. So we manually create the DoctrineOrmMappingsPass so that AnnotationDriver ignores
+     * the index.php file.
+     *
+     * @param string $moduleNamespace
+     * @param string $moduleEntityDirectory
+     * @param string $modulePrefix
+     *
+     * @return DoctrineOrmMappingsPass
+     */
+    private function createAnnotationMappingDriver($moduleNamespace, $moduleEntityDirectory, $modulePrefix)
```
PierreRambaud: Do you tried with https://symfony.com/doc/current/reference/configuration/doctrine.html#mapping-entities-outside-of-a-bundle ?

jolelievre: I didn't know about that! Smart!
I used this and removed the compiler pass, I only added it in the `DoctrineExtension` because we only need this in the front container

PierreRambaud: You can do it in a yaml file :) instead of editing the php code

jolelievre: Same difference, and I need to add it to the core config which is already parsed into an array So there would be the few interest two parse two yaml files and merge them partially 🤔 I wonder where I would place this file

PierreRambaud (dismissed):
We should use the config.yml the same place where the `doctrine` component is defined.
It's easier for anyone to improve the configuration if it's stored at the same place.
PIng @eternoendless

jolelievre:
The thing is this is a special case for our light container, only available in legacy environment The generic doctrine config for symfony environment is in `app/config/config.yml`

We could add this no-bundle config in it but it would make it redundant in symfony env because the mapping is automatically managed by the bundle I tried it, it **seems** to work fine but I'm worried about potential side effects, and it forces us to add useless config in the Symfony config just to satisfy our custom front container..

An alternative solution could be to add a `app/config/config_front.yml` config file that would import the config file By the way this makes me realize we never load the config depending on the env, meaning doctrine doesn't use any cache in front for now

eternoendless:
I agree on the settings approach as we discussed, but I think this PR should target 1.7.7.x

src/Adapter/Container/DoctrineBuilderExtension.php
```diff
@@ -40,25 +41,30 @@
...
      * {@inheritdoc}
      */
     public function build(ContainerBuilder $container)
     {
-        $yamlParser = new YamlParser($this->cacheDir);
+        $configDirectories = [_PS_ROOT_DIR_ . '/app/config'];
+        $fileLocator = new FileLocator($configDirectories);
+
+        $configLoader = new ConfigYamlLoader($fileLocator);
+        $configPath = sprintf('config_front_%s.yml', $this->environment->getName());
```
jolelievre: @PierreRambaud @eternoendless I'm not sure about the `_front` prefix because it is used for legacy containers, which is used on legacy context for both front and admin

So I see two solutions:
- use `_legacy` prefix
- add three new files with `_admin` prefix which will be the same as the front ones, but it complicates the config files even more

Note: this legacy admin container is supposed to be removed in 178 because only the Symfony container will be available in BO, so maybe we can accept this for one version Starting 178 the `_front` files and legacy container will only be used for front

PierreRambaud: :+1: for the `_legacy` prefix and the note :)

src/Adapter/Container/DoctrineBuilderExtension.php
```diff
@@ -40,25 +41,30 @@
...
+        $this->environment = $environment;
     }
 
     /**
      * {@inheritdoc}
      */
     public function build(ContainerBuilder $container)
     {
-        $yamlParser = new YamlParser($this->cacheDir);
+        $configDirectories = [_PS_ROOT_DIR_ . '/app/config'];
```
Progi1984:
```suggestion
        $configDirectories = [$container->getParameter('kernel.project_dir') . '/app/config'];
```

Progi1984 (commented):
A small feedback

src/Adapter/Container/DoctrineBuilderExtension.php
jolelievre: Yep you're right!

Progi1984 (dismissed):

app/config/config_legacy.yml
```diff
@@ -0,0 +1,12 @@
+imports:
```
eternoendless, line 3: Consider adding a header comment in this file explaining it

app/config/doctrine.yml
```diff
@@ -0,0 +1,34 @@
...
+        dbname:   "%database_name%"
+        user:     "%database_user%"
+        password: "%database_password%"
+        charset:  utf8mb4
+        # if using pdo_sqlite as your database driver:
+        #   1. add the path in parameters.yml
+        #     e.g. database_path: "%kernel.root_dir%/data/data.db3"
+        #   2. Uncomment database_path in parameters.yml.dist
+        #   3. Uncomment next line:
+        #     path:     "%database_path%"
```
eternoendless: I suppose these lines can go away

app/config/doctrine.yml
```diff
@@ -0,0 +1,34 @@
...
+        mapping_types:
+          enum: string
+
+        options:
+          1002: "SET sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''))"
+
+  orm:
+    auto_generate_proxy_classes: "%kernel.debug%"
+    # naming_strategy: doctrine.orm.naming_strategy.underscore
```
eternoendless: I suppose this line can be deleted

classes/module/Module.php
```diff
@@ -3432,33 +3438,54 @@ public function isAdminLegacyContext()
...
+     */
+    public function getContainer()
+    {
+        if (null === $this->container) {
+            $finder = new ContainerFinder($this->context);
+            try {
+                $this->container = $finder->getContainer();
+            } catch (ContainerNotFoundException $e) {
+                return null;
+            }
```
eternoendless: "A method should do what it's meant to do or throw an exception". Avoid mixed return types, because it's error prone and doesn't add value.

Consider this code:

```php
$this->getContainer()->get('some.service'); // Error: call to member function on null
```

It's best to remove the try and let consumers handle exceptions:
```suggestion
     * @return ContainerInterface
     *
     * @throws ContainerNotFoundException
     */
    public function getContainer(): ContainerInterface
    {
        if (null === $this->container) {
            $finder = new ContainerFinder($this->context);
            $this->container = $finder->getContainer();
```

classes/module/Module.php
```diff
@@ -3432,33 +3438,54 @@ public function isAdminLegacyContext()
      */
     public function isSymfonyContext()
     {
-        return !$this->isAdminLegacyContext() && defined('_PS_ADMIN_DIR_');
+        // Check that a container is available but not generated by PrestaShop
+        $container = $this->getContainer();
+
+        return null !== $container && !($container instanceof LegacyContainerInterface);
```
eternoendless: I think we should keep this as it was. It's the uses that must be updated to perform duck-typing (aka "is the container present?")

src/Adapter/Container/ConfigYamlLoader.php
```diff
@@ -0,0 +1,105 @@
...
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Container;
```
eternoendless: Why is this in Adapter?

src/Adapter/Container/ConfigYamlLoader.php
```diff
@@ -0,0 +1,105 @@
...
+
+use Symfony\Component\Config\Loader\FileLoader;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+use Symfony\Component\Yaml\Yaml;
+
+/**
+ * ConfigYamlLoader is used to load YAML config files and get the config as array.
+ */
+class ConfigYamlLoader extends FileLoader
```
eternoendless: This class looks pretty generic. Doesn't symfony already handle this stuff?

src/Adapter/Container/DoctrineBuilderExtension.php
```diff
@@ -27,8 +27,9 @@
 namespace PrestaShop\PrestaShop\Adapter\Container;
```
eternoendless, line 27: I wonder why this is in Adapter... shouldn't it be in the bundle?

src/Adapter/Container/DoctrineBuilderExtension.php
```diff
@@ -40,25 +41,30 @@
...
+        $this->environment = $environment;
     }
 
     /**
      * {@inheritdoc}
      */
     public function build(ContainerBuilder $container)
     {
-        $yamlParser = new YamlParser($this->cacheDir);
+        $configDirectories = [$container->getParameter('kernel.project_dir') . '/app/config'];
```
eternoendless:
```suggestion
        $configDirectories = [$container->getParameter('kernel.root_dir') . '/config'];
```
Although this path may be better set up as a parameter instead of hard coded?

classes/module/Module.php
jolelievre: But having a container doesn't tell if it's a full filled container from Symfony (with CQRS, bundles, and so on)
Or a legacy one with restricted services and features

So we could leave `isSymfonyContext` untouched, but add a new method to easily test which kind of container is present

Although I think the behaviour of this method has already been modified since 175, because some native modules like ps_linklist or productcomments can't be installed via CLI any more, so something changed

classes/module/Module.php
jolelievre: Ok fine by me, I actually introduced a similar check in the `get` method but had similar trouble so it's logical to remove it here as well

src/Adapter/Container/ConfigYamlLoader.php
jolelievre: Not for config files only, they have `Symfony\Component\DependencyInjection\Loader\YamlFileLoader` but it's meant to loader services and needs a container as input
The `parseImports` from this class is inspired by their `YamlFileLoader` but I needed something more simple that simply loads a config YAML files along with its imported files, and I couldn't find any class that did this (based on a search on the abstract `FileLoader` class)

src/Adapter/Container/DoctrineBuilderExtension.php
jolelievre: Good idea for `kernel.root_dir` however I'm not sure it's worth adding a parameter for the config directory, it will remain here by convention So letting the extension compute the path based on the `kernel.root_dir` seems enough to me

src/Adapter/Container/ConfigYamlLoader.php
jolelievre: Mainly because it's used by `DoctrineBuilderExtension` only, but this one might be moved in a different namespcae as it's quite generic

src/Adapter/Container/DoctrineBuilderExtension.php
jolelievre: Because it is used by our `ContainerBuilder` which namespace is in Adapter

This class is not an extension as in a Symfony extension, it's a custom extension used by our `ContainerBuilder` only, which implements our `PrestaShop\PrestaShop\Adapter\Container\ContainerBuilderExtensionInterface` So it shouldn't be in the bundle as it's not related to Symfony framework and can't be used by it

An extra pre layer was required additionally to the `CompilerPassInterfaces` because some actions need to be done before the container is built and start processing the compiler passes So I introduced this extension notion to manage Doctrine, but from the recent modification from this PR I actually wonder if we could have used the Symfony extension system instead

Also having a centralized/common config parsed by the container builder and passed to extensions might have been nice (that's the Symfony principle), but it represents a more in depth refacto

src/Adapter/Container/ConfigYamlLoader.php
jolelievre: Maybe we could move this in `PrestaShopBundle\DependencyInjection\Config` ?

jolelievre:
@eternoendless I addressed some of your feedback, and commented the others And as the `ps_linklist` install via CLI didn't work any more I proposed a fix on the module https://github.com/PrestaShop/ps_linklist/pull/86

(Actually it was not related to any modif in the module class, it only worked in 175 because it didn't use the container or service back then)

classes/module/Module.php
jolelievre: @eternoendless I reverted the changes on `isSymfonyContext` so no more changes nor BC break
And instead I added two methods `hasSymfonyContainer|hasLegacyContainer` which will allow to detect the context/container more easily

Progi1984 (dismissed):

classes/module/Module.php
```diff
@@ -3436,30 +3442,70 @@ public function isSymfonyContext()
...
         }
 
         return false;
     }
+
+    /**
+     * Return the container depending on the context:
+     *  - Symfony environment: container generated by the kernel
+     *  - Legacy environment admin: container generated by PrestaShop core for admin controllers
+     *  - Legacy environment front: container generated by PrestaShop core for front controllers
```
eternoendless: Phpdoc is meant to explain the _what_, not the _how_. If the _how_ is valuable information, keep it in the delegate service or in the method body.

```suggestion
     * Returns the service container.
```

classes/module/Module.php
```diff
@@ -3436,30 +3442,70 @@ public function isSymfonyContext()
...
+     *
+     * @return ContainerBuilder|ContainerInterface|null
+     */
+    public function getContainer()
```
eternoendless: I don't see  why this should be public

```suggestion
    protected function getContainer()
```

classes/module/Module.php
```diff
@@ -3436,30 +3442,70 @@ public function isSymfonyContext()
...
         return false;
     }
+
+    /**
+     * Return the container depending on the context:
+     *  - Symfony environment: container generated by the kernel
+     *  - Legacy environment admin: container generated by PrestaShop core for admin controllers
+     *  - Legacy environment front: container generated by PrestaShop core for front controllers
+     *
+     * @return ContainerBuilder|ContainerInterface|null
```
eternoendless: I suppose all containers implement `ContainerInterface`? In that case you can just hint that interface. And add it as a return type.

eternoendless: BTW this return type is also wrong in `ContainerFinder` which will never return null.

classes/module/Module.php
eternoendless: Alternatively

```suggestion
    protected function getContainer(): ContainerInterface
```

classes/module/Module.php
```diff
@@ -3436,30 +3442,70 @@ public function isSymfonyContext()
...
+        try {
+            $container = $this->getContainer();
+        } catch (ContainerNotFoundException $e) {
+            $container = null;
         }
-
-        if ($this->context->controller instanceof Controller) {
-            return $this->context->controller->get($serviceName);
+        if (null !== $container) {
+            return $container->get($serviceName);
```
eternoendless: I find this implementation a little awkward. Especially since `getContainer` will never return null.

Here's how it can be simplified:

```php
try {
	$container = $this->getContainer();
}
catch (ContainerNotFoundException $e) {
	return false;
}

return $container->get($serviceName);
```

classes/module/Module.php
eternoendless: I don't think we should add more "maybe useful someday" methods in a class which already does too much.

src/Adapter/Container/DoctrineBuilderExtension.php
eternoendless: Oh I see.

src/Adapter/ContainerBuilder.php
```diff
@@ -197,7 +203,7 @@ private function loadDoctrineAnnotationMetadata()
     /**
      * @param SfContainerBuilder $container
      *
-     * @throws \Exception
+     * @throws Exception
      */
     private function loadServicesFromConfig(SfContainerBuilder $container)
```
eternoendless: Shouldn't all references to `SfContainerBuilder` be  changed as well?

eternoendless (changes requested):

classes/module/Module.php
jolelievre: If we remove the `hasSymfonyContainer` `hasLegacyContainer` then this one is useful as public to check the container type manually

Besides it's also useful to get a parameter: `$this->getContainer()->getParameter()` unless we introduce some new shortcut `Module::getParameter`

classes/module/Module.php
jolelievre: Ok I'll remove this one, and explain this in the PHPDoc of `getContainer`

classes/module/Module.php
jolelievre: You're right I'll also correct ContainerFinder doc

classes/module/Module.php
jolelievre: But here I wanted to explain what can be returned, because depending on which type of container is returned the services available are not the same So maybe it's just about modifying the description of each environment/container

classes/module/Module.php
jolelievre: I removed `ContainerBuilder` indeed, but I put `LegacyContainerInterface` instead to emphasize the fact that two king of containers can be returned (Symfony and legacy)

src/Adapter/ContainerBuilder.php
jolelievre: Yes, you're right

classes/module/Module.php
eternoendless: Ok then

src/Adapter/Container/ConfigYamlLoader.php
eternoendless: Okay

classes/module/Module.php
```diff
@@ -3436,29 +3441,48 @@ public function isSymfonyContext()
...
+
+    /**
+     * Returns the container depending on the environment:
+     *  - Legacy: light container with few services specifically defined for legacy front/admin controllers
+     *  - Symfony: symfony container with all the migrated services (CQRS, ...)
+     *
+     * If you need to detect which kind of container you are using you can check if it is an instance of LegacyContainerInterface,
+     * which means it's a legacy/light container.
+     *
+     * @return LegacyContainerInterface|ContainerInterface
```
eternoendless: No need to specify subclasses in types

```suggestion
     * @return ContainerInterface
```

classes/module/Module.php
```diff
@@ -3436,29 +3441,48 @@ public function isSymfonyContext()
...
+     *
+     * @throws ContainerNotFoundException
+     */
+    public function getContainer()
```
eternoendless: Why not set the return type?

```suggestion
    public function getContainer(): ContainerInterface
```

classes/module/Module.php
eternoendless: They still belong to the same class hierarchy

classes/module/Module.php
jolelievre: It was to emphasize the two different types for type checking, but since I already explained it in the comment I can remove it

classes/module/Module.php
jolelievre: Done

matks (approved):

eternoendless (approved):

matks (approved):

PR #19026, opened 2020-05-07
Enable drag-and-drop for Grid listings in both directions

### LINKED ISSUE #19008: In Symfony listings, drag-and-drop require "sorting" to be ASC
#### Describe the bug

Drag-and-drop on Symfony listings needs the listing to be sorted by ascending positions. Else it is not displayed and not usable (see steps to reproduce).

#### Expected behavior

Drag-and-drop should:
- require position column to be sorted (indeed, if rows are unsorted, "moving a row with drag-and-drop" does not inform the software of the desired ordering) => this should be told to user, right now it's not obvious
- drag-and-drop should work no matter whether column would be sorted ASC or DESC

#### Steps to Reproduce

1. Go to BO, on Design > CMS Page listing
2. Drag-and-drop works: on the left, there is a small blank column where you can grab a handle and move rows
3. Sort listing by descending position: see screenshot below
4. Blank column at the left is not visible anymore. Drag-and-drop does not work

**Screenshots**
Capture d’écran 2020-05-06 à 15 42 17

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Drag-and-drop on grid listings needed the listing to be sorted by ascending positions. I modified the behavior of the Javascript to be able to work whether rows are ordered by ASC or DESC, so it works in both cases. See below for insights
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/19008
| How to test?  | See ticket. The bug should be fixed.

## Insights

Previous implementation was considering that, since rows were ordered following ASC, it could guess the new position (after drag-and-drop was done by user) by counting rows numbers. So it would consider 1st row (from top) would be position 0, next would be position 1, and so on ... It could handle pagination by adding the page offset.

This did not work if rows were ordered DESC.

So I changed the logic. When the page loads, I record, for all rows, in a dedicated jQuery data attribute, its offset in the grid. If the grid has 10 items it will number them 0, 1, 2, etc ...

After drag-and-drop has been done by user, I check all rows and find which rows have moved. Since moved row still has its offset, I can find where they are now and where they used to be.

Now, what about positions ? If I know that row number 3 has moved from position 2 to position 5, I can guess that the new "position" of row number 3 is _the position of the row that was, before it, in position 2_. The idea is to find "what row used to be where I am now" and this is the new "position" of the row.

As it's a complex logic, the code does not look good. I tried to make it more maintainable by adding comments and lot and lot of variable assignments (to use variable names as indicators of what they are).

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19026)

### DIFF
--- a/admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
+++ b/admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
@@ -86,22 +86,21 @@ export default class PositionExtension {
   getRowsPositions(paginationOffset) {
     const tableData = JSON.parse($.tableDnD.jsonize());
     const rowsData = tableData[`${this.grid.getId()}_grid_table`];
-    const regex = /^row_(\d+)_(\d+)$/;
+    const completeRowsData = [];
 
-    const rowsNb = rowsData.length;
-    const positions = [];
-    let rowData; let
-      i;
-    for (i = 0; i < rowsNb; i += 1) {
-      rowData = regex.exec(rowsData[i]);
-      positions.push({
-        rowId: rowData[1],
-        newPosition: paginationOffset + i,
-        oldPosition: parseInt(rowData[2], 10),
-      });
+    let trData;
+
+    for (let i = 0; i < rowsData.length; i++) {
+      trData = this.grid.getContainer()
+        .find('#' + rowsData[i]);
+
+      completeRowsData.push({
+        rowMarker: rowsData[i],
+        offset: trData.data('dragAndDropOffset')
+      })
     }
 
-    return positions;
+    return this.computeMappingBetweenOldAndNewPositions(completeRowsData);
   }
 
   /**
@@ -110,6 +109,9 @@ export default class PositionExtension {
    * @private
    */
   addIdsToGridTableRows() {
+
+    let counter = 0;
+
     this.grid.getContainer()
       .find(`.js-grid-table .js-${this.grid.getId()}-position`)
       .each((index, positionWrapper) => {
@@ -119,6 +121,9 @@ export default class PositionExtension {
         const id = `row_${rowId}_${position}`;
         $positionWrapper.closest('tr').attr('id', id);
         $positionWrapper.closest('td').addClass('js-drag-handle');
+        $positionWrapper.closest('tr').data('dragAndDropOffset', counter);
+
+        counter++;
       });
   }
 
@@ -172,5 +177,62 @@ export default class PositionExtension {
     }
 
     $form.submit();
+  }
+
+  /**
+   * Rows have been reordered. This function
+   * finds, for each row ID: the old position, the new position
+   *
+   * @returns {Array}
+   * @private
+   */
+  computeMappingBetweenOldAndNewPositions(rowsData) {
+    const regex = /^row_(\d+)_(\d+)$/;
+
+    const rowsNb = rowsData.length;
+    const positionsBeforeDragAndDrop = {};
+    const positionsAfterDragAndDrop = {};
+    const mapping = [];
+
+    let rowData;
+    let rowDataParsedData;
+    let i;
+    let rowID;
+    let rowOldPosition;
+    let rowOldOffset;
+    let rowNewOffset;
+
+    for (i = 0; i < rowsNb; i += 1) {
+      rowData = rowsData[i].rowMarker;
+      rowDataParsedData = regex.exec(rowData);
+
+      rowID = rowDataParsedData[1];
+      rowOldPosition = parseInt(rowDataParsedData[2], 10);
+      rowOldOffset = rowsData[i].offset;
+      rowNewOffset = i;
+
+      positionsBeforeDragAndDrop[rowOldOffset] = rowOldPosition;
+      positionsAfterDragAndDrop[rowNewOffset] = rowOldPosition;
+    }
+
+    let previousRowPositionWithThisOffset;
+
+    for (i = 0; i < rowsNb; i += 1) {
+      rowData = rowsData[i].rowMarker;
+      rowDataParsedData = regex.exec(rowData);
+      rowID = rowDataParsedData[1];
+      rowOldPosition = parseInt(rowDataParsedData[2], 10);
+
+      rowNewOffset = i;
+      previousRowPositionWithThisOffset = positionsBeforeDragAndDrop[rowNewOffset];
+
+      mapping.push({
+        rowId: rowID,
+        oldPosition: rowOldPosition,
+        newPosition: previousRowPositionWithThisOffset,
+      });
+    }
+
+    return mapping;
   }
 }
--- a/admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
+++ b/admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
@@ -90,6 +90,8 @@ export default class PositionExtension {
 
     let trData;
 
+    // retrieve dragAndDropOffset offset to have all needed data
+    // for positions mapping evolution over time
     for (let i = 0; i < rowsData.length; i++) {
       trData = this.grid.getContainer()
         .find('#' + rowsData[i]);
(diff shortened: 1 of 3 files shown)


admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
```diff
@@ -119,6 +123,9 @@ export default class PositionExtension {
         const id = `row_${rowId}_${position}`;
         $positionWrapper.closest('tr').attr('id', id);
         $positionWrapper.closest('td').addClass('js-drag-handle');
+        $positionWrapper.closest('tr').data('dragAndDropOffset', counter);
```
matks, line 124: Here I record, when page is loaded, the "initial offset" of the row.

admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
```diff
@@ -173,4 +180,67 @@ export default class PositionExtension {
...
+      rowNewOffset = i;
+
+      positionsBeforeDragAndDrop[rowOldOffset] = rowOldPosition;
+      positionsAfterDragAndDrop[rowNewOffset] = rowOldPosition;
+    }
+
+    let previousRowPositionWithThisOffset;
+
+    // for each row in table, we look at before the drag-and-drop
+    // and find what other rows was there, this is the new position
```
matks:
```suggestion
    // and find what other row was there, this is the new position
```

src/Core/Grid/Presenter/GridPresenter.php
```diff
@@ -137,8 +137,7 @@ public function getOrderingPosition(GridInterface $grid)
         /** @var ColumnInterface $column */
         foreach ($grid->getDefinition()->getColumns() as $column) {
             if ($column instanceof PositionColumn &&
-                strtolower($column->getId()) == strtolower($searchCriteria->getOrderBy()) &&
-                'asc' == strtolower($searchCriteria->getOrderWay())
+                strtolower($column->getId()) == strtolower($searchCriteria->getOrderBy())
```
matks, line 140: I disable the condition that removed the drag-and-drop when position sorting direction was not ASC

src/PrestaShopBundle/Twig/Extension/GridExtension.php
```diff
@@ -168,8 +168,7 @@ public function isOrderingColumn(array $grid)
     {
         if (empty($grid['columns'])
             || empty($grid['sorting']['order_by'])
-            || empty($grid['sorting']['order_way'])
-            || 'asc' != strtolower($grid['sorting']['order_way'])) {
+            || empty($grid['sorting']['order_way'])) {
```
matks, line 171: I disable the condition that removed the drag-and-drop when position sorting direction was not ASC

admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
```diff
@@ -86,22 +86,23 @@ export default class PositionExtension {
   getRowsPositions(paginationOffset) {
     const tableData = JSON.parse($.tableDnD.jsonize());
     const rowsData = tableData[`${this.grid.getId()}_grid_table`];
-    const regex = /^row_(\d+)_(\d+)$/;
```
zuk3975, line 89: nice to get rid of this

Progi1984 (approved):

PierreRambaud (approved):

admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
jolelievre: Then maybe `initialOffset` would be more understandable?

admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
```diff
@@ -173,4 +178,67 @@ export default class PositionExtension {
...
+      rowDataMarker = rowsData[i].rowMarker;
+      rowDataParsedData = regex.exec(rowDataMarker);
+
+      rowID = rowDataParsedData[1];
+      rowOldPosition = parseInt(rowDataParsedData[2], 10);
+      rowOldOffset = rowsData[i].offset;
+      rowNewOffset = i;
+
+      positionsBeforeDragAndDrop[rowOldOffset] = rowOldPosition;
+      positionsAfterDragAndDrop[rowNewOffset] = rowOldPosition;
```
jolelievre, line 218: Seems like you don't use `positionsAfterDragAndDrop` so it could be cleared (`rowNewOffset` can be removed as well in this loop)

This would make the algorithm easier to understand as well, because it's a bit hard to understand right now
First loop would only be `Get positions before drag and drop`

admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
```diff
@@ -110,6 +110,8 @@ export default class PositionExtension {
    * @private
    */
   addIdsToGridTableRows() {
+    let counter = 0;
```
jolelievre, line 113: Be careful you removed the notion of `paginationOffset` and your counter start at 0 here
The offset is used when you drag and drop on second, third, ... pages

Without this I'm afraid the drag and drop on other pages will send position between 0..10 (for 10 elements) which might break the position expected (12, 25, ...) But I may be wrong since I don't understand fully your algorithm

admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
```diff
@@ -83,25 +82,26 @@ export default class PositionExtension {
...
-        newPosition: paginationOffset + i,
-        oldPosition: parseInt(rowData[2], 10),
+    let trData;
+
+    // retrieve dragAndDropOffset offset to have all needed data
+    // for positions mapping evolution over time
+    for (let i = 0; i < rowsData.length; i += 1) {
+      trData = this.grid.getContainer()
+        .find(`#${rowsData[i]}`);
+
```
jolelievre, line 97: I like your idea to format data before computing them, I think you could even perform the regex here, this would lighten the `computeMappingBetweenOldAndNewPositions` algorithm

for example:

```js
    let rowDataParsedData = regex.exec(rowsData[i]);

    completeRowsData.push({
        rowMarker: rowsData[i],
        offset: trData.data('dragAndDropOffset'),
        rowID: rowDataParsedData[1],
        oldPosition: parseInt(rowDataParsedData[2], 10),
    });
```

This way in the compute function you can directly use `rowsData[i].rowId` for example, which would help the readability

admin-dev/themes/new-theme/js/components/grid/extension/position-extension.js
```diff
@@ -173,4 +178,67 @@ export default class PositionExtension {
...
+    // and find what other row was there, this is the new position
+    // of current row
+    for (i = 0; i < rowsNb; i += 1) {
+      rowDataMarker = rowsData[i].rowMarker;
+      rowDataParsedData = regex.exec(rowDataMarker);
+      rowID = rowDataParsedData[1];
+      rowOldPosition = parseInt(rowDataParsedData[2], 10);
+
+      rowNewOffset = i;
+      previousRowPositionWithThisOffset = positionsBeforeDragAndDrop[rowNewOffset];
```
jolelievre, line 233: I'm not sure I get that part, how is the position before drag and drop the new position after 😅

jolelievre (commented):
Hi @matks 

I added a few comments that could help the readability of this new algorithm. Don't take me wrong, it' a huge improvement and the subject is not easy It just would be great if the code can be understood easily because it's a bit "blur" as is (the previous one wasn't any better 😅)

If you feel like working on it as peer programming I'm "hot"

SD1982:
LGTM Thanks @matks !

Progi1984:
Thanks @matks

Issue #14956, opened 2019-08-01
Missing function


idnovate:
**Describe the bug**
Function `Tools::convertPrice()` is deprecated and `convertPriceToCurrency()` should be used instead, but this function does not exist.

```
/**
 * Return price converted.
 *
 * @deprecated since 1.7.4 use convertPriceToCurrency()
 *
 * @param float $price Product price
 * @param object|array $currency Current currency object
 * @param bool $to_currency convert to currency or from currency to default currency
 * @param Context $context
 *
 * @return float Price
 */
public static function convertPrice($price, $currency = null, $to_currency = true, Context $context = null)
{
    $default_currency = (int) Configuration::get('PS_CURRENCY_DEFAULT');

    if (!$context) {
        $context = Context::getContext();
    }
    if ($currency === null) {
        $currency = $context->currency;
    } elseif (is_numeric($currency)) {
        $currency = Currency::getCurrencyInstance($currency);
    }

    $c_id = (is_array($currency) ? $currency['id_currency'] : $currency->id);
    $c_rate = (is_array($currency) ? $currency['conversion_rate'] : $currency->conversion_rate);

    if ($c_id != $default_currency) {
        if ($to_currency) {
            $price *= $c_rate;
        } else {
            $price /= $c_rate;
        }
    }

    return $price;
}
```

**Additional information**
PrestaShop version: 1.7.6

khouloudbelguith:
Hi @idnovate,

Thanks for your report.
https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Tools.php#L853
I’ll add this to the debug roadmap so that it’s fixed. If you have already fixed it on your end or if you think you can do it, please do send us a pull request!
Thanks!

idnovate:
I don't know which is the aim to replace this function. So somebody should decide to remove the `@deprecated` or to write new function `convertPriceToCurrency()`

PierreRambaud:
Found the original commit introducing this wrong deprecated notice: https://github.com/PrestaShop/PrestaShop/pull/8411/files#diff-9079f61ba1c5da7f9f142723d5d42ea0R683

PierreRambaud:
FYI, we remove the deprecation notice because of this commit:
https://github.com/PrestaShop/PrestaShop/pull/8411/commits/bfd633032da529eae9259a47775d39cd029d10b0#diff-0175e8ccd89d320e7f91e6333bc749edb922cd119188d0545f51fa3dd838aa2cL718

PR #19078, opened 2020-05-11
Improve convertPrice function in Tools.php

j0ssGZ:
I've had to resort to this function to be able to format a price and compare it. Unlike other similar functions like displayPrice, here it doesn't  validate if a price is numeric but it comes as a string (something that happens on many occasions). With this little validation we make sure that whenever we enter a valid value, the result will be returned correctly.

On the other hand, I have removed the comment from @deprecated, I understand that until the function is available, this comment should not be displayed because it misleads the developer.

| Questions     | Answers
| Branch?       | develop
| Description?  | Improve the operation of the displayPrice function in Tools.php
| Type?         | improvement
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/14956
| How to test?  | Using Tools :: displayPrice with a numeric value in string.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19078)

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -871,7 +871,7 @@ class ToolsCore
         }
 
         if (gettype($price) === 'string' && is_numeric($price)) {
-            $price = (float)$price;
+            $price = (float) $price;
         } else if (!is_numeric($price)) {
             throw new PrestaShopException('Invalid price');
         }
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -872,8 +872,10 @@ class ToolsCore
 
         if (gettype($price) === 'string' && is_numeric($price)) {
             $price = (float) $price;
-        } elseif (!is_numeric($price)) {
-            throw new PrestaShopException('Invalid price');
+        }
+
+        if (!is_numeric($price)) {
+            return $price;
         }
 
         $c_id = (is_array($currency) ? $currency['id_currency'] : $currency->id);
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -849,8 +849,6 @@ class ToolsCore
     /**
      * Return price converted.
      *
-     * @deprecated since 1.7.4 use convertPriceToCurrency()
-     *
      * @param float $price Product price
      * @param object|array $currency Current currency object
      * @param bool $to_currency convert to currency or from currency to default currency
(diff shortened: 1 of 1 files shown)


PierreRambaud:
Hi @j0ssGZ and thanks for your contribution.

The `@deprecated` annotation is used to document the deprecation version. If the function is still here, it's because we don't release a major version yet and we follow [Semver](https://semver.org/).

This is why you can see many `@deprecated` annotations in the code, it's here to say "Hey, in the next major we will remove it" :wink: 

Because it's a deprecated function, I'm not in favor of touching it without tests. I'm also not in favor of using `float`, we should use the `decimal` library.

PierreRambaud (changes requested):
See comments

classes/Tools.php
```diff
@@ -849,8 +849,6 @@ public static function displayPriceSmarty($params, &$smarty)
     /**
      * Return price converted.
      *
-     * @deprecated since 1.7.4 use convertPriceToCurrency()
```
matks: 👍 nice, thanks

for people who dont understand this: see https://github.com/PrestaShop/PrestaShop/issues/14956

matks:
> Hi @j0ssGZ and thanks for your contribution.
> 
> The `@deprecated` annotation is used to document the deprecation version. If the function is still here, it's because we don't release a major version yet and we follow [Semver](https://semver.org/).
> 
> This is why you can see many `@deprecated` annotations in the code, it's here to say "Hey, in the next major we will remove it" 😉
> 
> Because it's a deprecated function, I'm not in favor of touching it without tests. I'm also not in favor of using `float`, we should use the `decimal` library.

See https://github.com/PrestaShop/PrestaShop/issues/14956 : `convertPriceToCurrency()` does not exist anywhere 😅

classes/Tools.php
PierreRambaud: You're right, I found the original commit: https://github.com/PrestaShop/PrestaShop/pull/8411/files#diff-9079f61ba1c5da7f9f142723d5d42ea0R683

PierreRambaud (dismissed):

j0ssGZ (commented):
Ok, I have replaced the comment and used the Decimal library to get the value

PierreRambaud:
> Ok, I have replaced the comment and used the Decimal library to get the value

I'm really sorry for the delay, seems vacations wasn't the best moment to start reviews :sweat: 
Imho, we shouldn't touch this part of code without tests, and regarding your changes, you only convert the string ($price) into DecimalNumber class and use a default precision without taking in account what has been defined in the database, then directly return it without using the currency. So it seems useless :/

PierreRambaud:
Hi @j0ssGZ ,

Thanks!

PR #19108, opened 2020-05-13
Extend OrderController with FrameworkBundleAdminController instead of CommonController

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | OrderController was extending CommonController which is not designed for that, instead it needs to extend FrameworkBundleAdminController. It didn't use any methods of extended class, so the fix is to simply extend one class instead of another.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | It was more like "technical design failure" which could have caused bugs in future, there is nothing concrete to test. A brief check if actions are working in orders page should be enough.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19108)


Progi1984 (approved):

jolelievre (approved):

PierreRambaud (approved):

SD1982:
LGTM I did not notice any unwanted behavior related to this PR. Thanks @zuk3975

matks:
Thank you @zuk3975

PR #19139, opened 2020-05-14
Do not use cache while checking if an order exists

### LINKED ISSUE #11115: Duplicate order reference for different order 2/2
For the second time we have a duplicate reference for two different orders, the first time both orders were created at the very same time, the second time the orders were created with 1 second difference.

**Describe the bug**

So far it happened twice within a month or so, with hundreds of orders created just fine. We are unable to find out why so there are no steps to reproduce this at the moment, sorry.

Different order, same reference, different customer, different cart, same (or very close) date_add:

"id_order";"reference";"id_customer";"id_cart";"invoice_date";"date_add"
"1125";"C18101001125";"143131";"6367";"2018-10-10 22:06:36";"2018-10-10 22:06:36"
"1126";"C18101001125";"143130";"6372";"2018-10-10 22:06:36";"2018-10-10 22:06:36"
"2745";"C18102102744";"94587";"15912";"2018-10-21 21:01:59";"2018-10-21 21:01:59"
"2746";"C18102102744";"62536";"15904";"2018-10-21 21:02:00";"2018-10-21 21:02:00"

**Additionnal information**
PrestaShop version: 1.7.4.0
PHP version: 7.0

---
> Originally posted by @matks 

@MatShir I confirm that this issue is a global mess 😅 

I think many people observe the same issue (= duplicate order references) but there are actually multiple usecases, some which are faulty, some which are not.

I have noticed:

## Duplicate order references when there is an order split

This usecase is not a bug ✅  . When a cart contains multiple products that cannot be shipped by a single carrier, then the order will be split into sub-orders. Each sub-order is being shipped by a different carrier **but they all share the same order reference**.

## Duplicate order with different shipping adresses

> In my last investigations, it can be related to addresses: sometimes 2 shipping addresses ID are sended in customer cart on PS1.7 but one of this address is deleted=1 in database, so that can create some problems like wrong shipping amount or wrong discounts.

It seems there might be a issue related to customer using a deleted address but this seems to be random so it cannot be reproduced easily.

➡️  possibly a bug but as we dont have a step-by-step 100% reproductible scenario there is nothing we can do

## Duplicate order references with advanced warehouse management

> We had this problem on a mono-shop with advanced stock management activated and it disappeared when advanced stock management has been unabled.

It seems there might be a issue related to using advanced stock management but AFAIK this comes from a module, not the Core then ? 🤔 

## Wrong invoice number assignment

Contributor @Alexandre17220 mentioned some wrong code in the logic of invoice number generation that is prone to race conditions issues.

# Summary

As I said, I think many people observe the same issue (= duplicate order references) but there are actually multiple usecases, some which are faulty, some which are not. This GitHub issue is mixing all of these which is making the situation more ambiguous rather than helping. I suggest closing and locking this issue and opening new issues, one per _reproductible_ usecase to sort it out. Nothing will come out of this issue in its current state, there is too much informations that we cannot act on.

---
https://github.com/PrestaShop/PrestaShop/issues/11115#issuecomment-466052440

the source code is very simple for understand the problem   : 

```
public static function setLastInvoiceNumber($order_invoice_id, $id_shop)
    {
        if (!$order_invoice_id) {
            return false;
        }

        $number = Configuration::get('PS_INVOICE_START_NUMBER', null, null, $id_shop);
        // If invoice start number has been set, you clean the value of this configuration
        if ($number) {
            Configuration::updateValue('PS_INVOICE_START_NUMBER', false, false, null, $id_shop);
        }

        $sql = 'UPDATE `'._DB_PREFIX_.'order_invoice` SET number =';

        if ($number) {
            $sql .= (int)$number;
        } else {
            $getNumberSql = '(SELECT new_number FROM (SELECT (MAX(`number`) + 1) AS new_number
                FROM `'._DB_PREFIX_.'order_invoice`'.(Configuration::get('PS_INVOICE_RESET') ?
                ' WHERE DATE_FORMAT(`date_add`, "%Y") = '.(int)date('Y') : '').') AS result)';
            $getNumberSqlRow = Db::getInstance()->getRow($getNumberSql);
            $newInvoiceNumber = $getNumberSqlRow['new_number'];
            $sql .= $newInvoiceNumber;
        }

        $sql .= ' WHERE `id_order_invoice` = '.(int)$order_invoice_id;

        return Db::getInstance()->execute($sql);
    }
```

**SELECT (MAX(`number`) + 1)**

---

https://github.com/PrestaShop/PrestaShop/issues/11115#issuecomment-466335070

```
SELECT (MAX(`number`) + 1)
```

I see, so this is a race condition where two orders may get the same id. Nice find!

### LINKED ISSUE #17266: Duplicated orders from the same cart

### LINKED ISSUE #20641: Duplicate order reference for different order 1/2
#### Describe the bug

The function orderExists() from Cart.php has been modified in order to use only the cache when the order exists, avoiding the issue with duplicated orders.
**Screenshots**

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.7.x 
| Description?  | The function orderExists() from Cart.php has been modified in order to use only the cache when the order exists, avoiding the issue with duplicated orders.<br>Original pull request #17302
| Type?         | bug fix 
| Category?     | FO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20641 
| How to test?  | See #20641 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19139)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1628,9 +1628,10 @@ class CartCore extends ObjectModel
      */
     public function orderExists()
     {
-        $result = (bool) Db::getInstance()->getValue('SELECT count(*) FROM `' . _DB_PREFIX_ . 'orders` WHERE `id_cart` = ' . (int) $this->id);
-
-        return $result;
+        return (bool) Db::getInstance()->getValue(
+            'SELECT count(*) FROM `' . _DB_PREFIX_ . 'orders` WHERE `id_cart` = ' . (int) $this->id,
+            false
+        );
     }
 
     /**
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1628,15 +1628,9 @@ class CartCore extends ObjectModel
      */
     public function orderExists()
     {
-        $cache_id = 'Cart::orderExists_' . (int) $this->id;
-        if (!Cache::isStored($cache_id)) {
-            $result = (bool) Db::getInstance()->getValue('SELECT count(*) FROM `' . _DB_PREFIX_ . 'orders` WHERE `id_cart` = ' . (int) $this->id);
-            Cache::store($cache_id, $result);
+        $result = (bool) Db::getInstance()->getValue('SELECT count(*) FROM `' . _DB_PREFIX_ . 'orders` WHERE `id_cart` = ' . (int) $this->id);
 
-            return $result;
-        }
-
-        return Cache::retrieve($cache_id);
+        return $result;
     }
 
     /**
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -377,7 +377,7 @@ abstract class PaymentModuleCore extends Module
                     $transaction_id = null;
                 }
 
-                if (!$order->addOrderPayment($amount_paid, null, $transaction_id)) {
+                if (!isset($order) || !$order->addOrderPayment($amount_paid, null, $transaction_id)) {
                     PrestaShopLogger::addLog('PaymentModule::validateOrder - Cannot save Order Payment', 3, null, 'Cart', (int) $id_cart, true);
 
                     throw new PrestaShopException('Can\'t save Order Payment');
@@ -398,7 +398,7 @@ abstract class PaymentModuleCore extends Module
                         $message .= '<br />' . $this->trans('Warning: the secure key is empty, check your payment account before validation', [], 'Admin.Payment.Notification');
                     }
                     // Optional message to attach to this order
-                    if (isset($message) & !empty($message)) {
+                    if (!empty($message)) {
                         $msg = new Message();
                         $message = strip_tags($message, '<br>');
                         if (Validate::isCleanHtml($message)) {
--- a/tests/Integration/Behaviour/Features/Context/CustomerManagerFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/CustomerManagerFeatureContext.php
@@ -32,6 +32,7 @@ use PrestaShop\PrestaShop\Core\Domain\Customer\Command\AddCustomerCommand;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\DeleteCustomerCommand;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\EditCustomerCommand;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Command\TransformGuestToCustomerCommand;
+use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\CustomerNotFoundException;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Exception\DuplicateCustomerEmailException;
 use PrestaShop\PrestaShop\Core\Domain\Customer\Query\GetCustomerForEditing;
 use PrestaShop\PrestaShop\Core\Domain\Customer\QueryResult\EditableCustomer;
@@ -182,6 +183,14 @@ class CustomerManagerFeatureContext extends AbstractPrestaShopFeatureContext
     }
 
     /**
+     * @When I delete customer ":customerReference" with "allow registration after deletion" checked
+     */
+    public function deleteCustomerWithAllowCustomerRegistration($customerReference)
+    {
+        $this->deleteCustomer($customerReference, CustomerDeleteMethod::ALLOW_CUSTOMER_REGISTRATION);
+    }
+
+    /**
      * @When /^I delete customer "(.+)" with method "(.+)"$/
      */
     public function deleteCustomer($customerReference, $methodName)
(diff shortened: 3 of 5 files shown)


classes/Cart.php
```diff
@@ -1628,15 +1628,9 @@ public function _addCustomization($id_product, $id_product_attribute, $index, $t
...
-            $result = (bool) Db::getInstance()->getValue('SELECT count(*) FROM `' . _DB_PREFIX_ . 'orders` WHERE `id_cart` = ' . (int) $this->id);
-            Cache::store($cache_id, $result);
+        $result = (bool) Db::getInstance()->getValue('SELECT count(*) FROM `' . _DB_PREFIX_ . 'orders` WHERE `id_cart` = ' . (int) $this->id);
 
-            return $result;
-        }
-
-        return Cache::retrieve($cache_id);
+        return $result;
     }
```
matthieu-rolland, line 1635: > The function orderExists() from Cart.php has been modified **in order to use only the cache when the order exists**

Here it seems like the cache is never used in the **orderExists** function, is it on purpose?

eternoendless: The cache only makes sense if it's invalidated properly. If an order is checked, then deleted, then checked again it would seem as if it wasn't deleted. So yeah, I think it's correct. However I'm interested in knowing why it was added in the first place.

PierreRambaud: The ugly reason: https://github.com/PrestaShop/PrestaShop/commit/a7869a1f06c32978def39d4ab9ff9b2eebe14344

PierreRambaud:
PR rebased to fix CS

Progi1984 (dismissed):

classes/Cart.php
matthieu-rolland: 👍

matthieu-rolland (dismissed):

prestamodule:
@PierreRambaud does this PR intent that you should use `getValue(<query>, false)` too ? WDYT ?

PierreRambaud:
> @PierreRambaud does this PR intent that you should use `getValue(<query>, false)` too ? WDYT ?

Nice catch, didn't know about this (ugly) SQL "cache" :sweat_smile:

Progi1984 (approved):

colinegin:
Hello guys,

is this PR targeting the develop branch or the 1.7.7.x branch ? 
The milestone indicated is 1770 while the branch seems to be develop.

Thanks !

PierreRambaud:
> Hello guys,
> 
> is this PR targeting the develop branch or the 1.7.7.x branch ?
> The milestone indicated is 1770 while the branch seems to be develop.
> 
> Thanks !

Forget to change the Pull request description, my bad

Robin-Fischer-PS:
Review needed for QA, thanks !

Also, if there is an actual scenario to test, we would really appreciate it, since in the original issue, the bug seems to be "random", and has never been reproduced.

If there is no scenario, would a regression test be sufficient ?

PierreRambaud:
> Review needed for QA, thanks !
> 
> Also, if there is an actual scenario to test, we would really appreciate it, since in the original issue, the bug seems to be "random", and has never been reproduced.
> 
> If there is no scenario, would a regression test be sufficient ?

Yes, only a non-regression test is needed ;)

matthieu-rolland (approved):

Robin-Fischer-PS:
Regression tests done on order and cart, OK for me, thanks !

Progi1984:
Thanks @PierreRambaud

classes/PaymentModule.php
```diff
@@ -377,7 +377,7 @@ public function validateOrder(
                     $transaction_id = null;
                 }
 
-                if (!$order->addOrderPayment($amount_paid, null, $transaction_id)) {
+                if (!isset($order) || !$order->addOrderPayment($amount_paid, null, $transaction_id)) {
```
jf-viguier, line 380: Is there a relation between this isset and the duplicate order issue ?

kpodemski: I don't think so, for the $order to be not defined `$package_list` must not exists which is hardly possible but I guess @PierreRambaud just wanted to make sure ;)

PR #19146, opened 2020-05-14
Improvement - Theme hook only_pages

### LINKED ISSUE #16420: Theme.yml - Hooks configuration, alternative to except_pages
As an example:
```
displayHome:
  - ps_imageslider
```
This will only add the ps_imageslider module to the `displayHome` hook. However, if you go to any other page, category, product, etc. and take a look in the source code you will still see that the CSS/JS for that module is being loaded. Right now we have the option to use the `except_pages` configuration but that would mean that we would have to specify *all* pages. Which means:
```
displayHome:
  - ps_imageslider
      except_pages:
        - category
        - product
        - best-sales
        - prices-drop
        - checkout
        - ....
        - ....
        - ....
        - ....
        - ....
```
As you can imagine, this can get very verbose and difficult to keep track. Sometimes you really need the `except_pages` but I'd wager that in most cases something like `only_on_pages` would make more sense.

 It would a very welcomed addition to have something like `only_on_pages`. That way we could possibly do something like:
```
displayHome:
  - ps_imageslider
    only_on_pages:
      - index
```

lusimeon:
| Questions     | Answers
| Branch?       | develop
| Description?  | In contrast to except_pages sequences in theme.yml who's not very handy, this PR add the posibility to limit hooks to load only on specified pages thanks to an "only_pages" sequence.<br>except_pages continue to work but if only_pages and except_pages are set on the same hook, only_pages has priority.
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16420.
| How to test?  | Run tests, or add only_pages sequence in theme.yml

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19146)

### DIFF
--- a/classes/Dispatcher.php
+++ b/classes/Dispatcher.php
@@ -1184,4 +1184,35 @@ class DispatcherCore
 
         return $controllers;
     }
+
+    /**
+     * Get names of all available FO controllers.
+     *
+     * @var mixed
+     *
+     * @return array
+     */
+    public static function getControllersNames($dirs)
+    {
+        if (!is_array($dirs)) {
+            $dirs = [$dirs];
+        }
+
+        $controllers = Dispatcher::getControllers($dirs);
+
+        $controllersNames = [];
+        foreach ($controllers as $controller) {
+            $reflectionClass = new \ReflectionClass($controller);
+
+            $reflectionClassProperties = $reflectionClass->getDefaultProperties();
+
+            if (empty($reflectionClassProperties['php_self'])) {
+                continue;
+            }
+
+            $controllersNames[] = $reflectionClassProperties['php_self'];
+        }
+
+        return $controllersNames;
+    }
 }
--- a/src/Core/Module/HookRepository.php
+++ b/src/Core/Module/HookRepository.php
@@ -27,6 +27,7 @@
 namespace PrestaShop\PrestaShop\Core\Module;
 
 use Db;
+use Dispatcher;
 use Exception;
 use PrestaShop\PrestaShop\Adapter\Hook\HookInformationProvider;
 use Shop;
@@ -158,28 +159,15 @@ class HookRepository
 
                 if (!empty($extra_data['only_pages'])) {
                     $extra_data['except_pages'] = [];
-                    $pages = [];
 
-                    $controllers = \Dispatcher::getControllers(_PS_FRONT_CONTROLLER_DIR_);
+                    $controllersNames = Dispatcher::getControllersNames(_PS_FRONT_CONTROLLER_DIR_);
 
-                    foreach ($controllers as $controller) {
-                        $reflectionClass = new \ReflectionClass($controller);
-
-                        $reflectionClassProperties = $reflectionClass->getDefaultProperties();
-
-                        if (empty($reflectionClassProperties['php_self'])) {
+                    foreach ($controllersNames as $controllerName) {
+                        if (in_array($controllerName, $extra_data['only_pages'])) {
                             continue;
                         }
 
-                        $pages[] = $reflectionClassProperties['php_self'];
-                    }
-
-                    foreach ($pages as $key => $page) {
-                        if (in_array($page, $extra_data['only_pages'])) {
-                            continue;
-                        }
-
-                        $extra_data['except_pages'][] = $page;
+                        $extra_data['except_pages'][] = $controllerName;
                     }
                 }
 
--- a/src/Core/Module/HookRepository.php
+++ b/src/Core/Module/HookRepository.php
@@ -156,6 +156,33 @@ class HookRepository
 
                 $this->db->insert('hook_module', $row);
 
+                if (!empty($extra_data['only_pages'])) {
+                    $extra_data['except_pages'] = [];
+                    $pages = [];
+
+                    $controllers = \Dispatcher::getControllers(_PS_FRONT_CONTROLLER_DIR_);
+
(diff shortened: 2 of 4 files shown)


Progi1984:
@lusimeon It seems that your PR broke unit tests : https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/334202032#L524

Progi1984 (changes requested):
Unit tests are broken

Progi1984 (approved):

classes/Dispatcher.php
```diff
@@ -1184,4 +1184,35 @@ public static function getControllersInDirectory($dir)
...
+    {
+        if (!is_array($dirs)) {
+            $dirs = [$dirs];
+        }
+
+        $controllers = Dispatcher::getControllers($dirs);
+
+        $controllersNames = [];
+        foreach ($controllers as $controller) {
+            $reflectionClass = new \ReflectionClass($controller);
```
matks, line 1205: Reflection is slow but if `getControllersNames()` is called during hook setup (and not at runtime, when FO page with hooks on it is loaded) it's OK 👍

matks:
@lusimeon Nice work !

So you decided to use `only_pages` data and then compare it to the list of all FO controllers to be able to compute a complete `except_pages` map, is that right ?

I think if we merge this (and it looks like a nice idea) we need to update the documentation

classes/Dispatcher.php
Matt75: https://ocramius.github.io/blog/accessing-private-php-class-members-without-reflection/

src/Core/Module/HookRepository.php
```diff
@@ -156,6 +157,20 @@ public function persistHooksConfiguration(array $hooks)
 
                 $this->db->insert('hook_module', $row);
 
+                if (!empty($extra_data['only_pages'])) {
+                    $extra_data['except_pages'] = [];
+
+                    $controllersNames = Dispatcher::getControllersNames(_PS_FRONT_CONTROLLER_DIR_);
```
Matt75, line 163: Maybe you should add ModuleFrontController too 🤔
Because some module controllers are important like for a blog module or discounted product listing for example

Matt75 (commented):
Some suggestions

src/Core/Module/HookRepository.php
lusimeon: You're right, I will add modules front controllers too.

classes/Dispatcher.php
lusimeon: @Matt75 In my opinion, the way described in the blog post you've shared sounds very difficult to read and so to maintain. Like @matks says, this function is executed only during hook setup, not on runtime. Do we sacrificing readability for a one time performance gain ?

Matt75: Nope, readability is more important ;)

micka-fdz:
I think it would be a good improvement! Good job @lusimeon

sowbiba (commented):
seems good to me, waiting for the modification suggested by @Matt75

matthieu-rolland (commented):
Thank you for your contribution @lusimeon, lgtm !

We'll wait your implementation of @Matt75 's suggestion before sending this PR to QA.

kpodemski:
Great idea!

src/Core/Module/HookRepository.php
lusimeon: After few research, at this time `except_pages` **seems to don't works** with all module controllers because some of thoses don't have `controller_name` or `php_self` (see `classes/Hook.php` **line 858**) attributes set (eg: module `ps_emailsubscription/controllers/front/verification.php` results in `module-ps_emailsubscription-` key).

**I think an issue need to be open** to fix module controller name to have something even (maybe #19226) before we can include module controller in this PR.

@Matt75 **If you have any suggestions** on how to get all modules controllers names? I tried with filenames return by `Dispatcher::getModuleControllers()` but it doesn't works when checking permissions in `Hook::exec()`.

lusimeon:
> @lusimeon Nice work !
> 
> So you decided to use `only_pages` data and then compare it to the list of all FO controllers to be able to compute a complete `except_pages` map, is that right ?
> 
> I think if we merge this (and it looks like a nice idea) we need to update the documentation

@matks You're right, I'll do it when this PR will be merged. Am I right?

lusimeon:
Hi, any news on this one? cc @Matt75 @matks

PierreRambaud:
> Hi, any news on this one? cc @Matt75 @matks

Little reminder guys :)

PierreRambaud:
> > Hi, any news on this one? cc @Matt75 @matks
> 
> Little reminder guys :)

Ping!

Matt75:
Hello @lusimeon 

Good practices for ModuleFrontController is to be declared by developers in their module class : https://github.com/PrestaShop/PrestaShop/blob/1.7.8.1/classes/module/Module.php#L124
At module installation, PrestaShop will automatically register them to Meta table : https://github.com/PrestaShop/PrestaShop/blob/1.7.8.1/classes/module/Module.php#L2970
So controllers will be available in BO >  Shop Parameters > Traffic & SEO to manage URL Rewriting for example.
So you can retrieve list of ModuleFrontController thanks to ps_meta table

About ModuleAdminController, it's required to be registered in ps_tab table

PierreRambaud:
Hi @lusimeon ,

Thanks!

PR #19180, opened 2020-05-18
'Total available for each user' in cart rule not considered

### LINKED ISSUE #15751: 'Total available for each user' in cart rule not considered
**Describe the bug**
When you offer a product with price 0 as a free gift any record is inserted at table `ps_order_cart_rule`, so the customer can use it another time

**To Reproduce**

1. Set a product with price 0
image
1. Create a cart rule limited to 1 use per customer and offer this product as free gift:
image
image

1. Create an order. Cart rule is added correctly.
1. Create a second order. **Cart rule is added but it shouldn't**.

**Additional information**
PrestaShop version: 1.7.6.1

idnovate:
| Questions     | Answers
| Branch?       | develop
| Description?  | Check https://github.com/PrestaShop/PrestaShop/issues/15751
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #15751
| How to test?  | Check https://github.com/PrestaShop/PrestaShop/issues/15751 to reproduce error

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19180)

### DIFF
--- a/src/Adapter/Product/AbstractProductHandler.php
+++ b/src/Adapter/Product/AbstractProductHandler.php
@@ -1,0 +1,73 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Product;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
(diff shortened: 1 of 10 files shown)


matks:
Hi @idnovate ! Thank you for the contribution !

I am sorry, I just read the issue and I see you provided the answer. I think we did not see it (sometimes we get so many issues, we miss some very important ones).

Back to the PR 😄 looks good, I would like to understand a few things.

## Cart rule logic

I think the most important part is

```
if (!$values['tax_excl']) {
    continue;
}
```
⏬ 
```
if ($values['tax_excl'] > 0) {
```

Thanks to this, function `$order->addCartRule($cartRule->id, $cartRule->name, $values, 0, $cartRule->free_shipping);` is called and insert the right record, is that right ?

## Display changes

I see in the PR you modify a few other things, is that some polishing ? It seems not related to the scope of the PR. I'm not saying it's bad I just wish to understand 😄 

## Possible side-effects

Cart rule computations are complex. Because of your change, not only `$order->addCartRule(...)` is executed but also the lines below

```
            $order->addCartRule($cartRule->id, $cartRule->name, $values, 0, $cartRule->free_shipping);

            if ($id_order_state != Configuration::get('PS_OS_ERROR') && $id_order_state != Configuration::get('PS_OS_CANCELED') && !in_array($cartRule->id, $cart_rule_used)) {
                $cart_rule_used[] = $cartRule->id;
                // Create a new instance of Cart Rule without id_lang, in order to update its quantity
                $cart_rule_to_update = new CartRule((int) $cartRule->id);
                $cart_rule_to_update->quantity = max(0, $cart_rule_to_update->quantity - 1);
                $cart_rule_to_update->update();
            }
            $cart_rules_list[] = [
                'voucher_name' => $cartRule->name,
                'voucher_reduction' => ($values['tax_incl'] != 0.00 ? '-' : '') . Tools::getContextLocale($this->context)->formatPrice($values['tax_incl'], $this->context->currency->iso_code),
            ];
```

We will have to ask QA to do deep testing to make sure this part of the code will not modify behavior for other type of cart rules

idnovate:
Hi!

- Display changes
All are related except this (https://github.com/PrestaShop/PrestaShop/pull/19180/files#diff-8ae578b858b70de2f02ae5c46ac5635a)!! I made a mess with commits.

If you don't set the other changes it won't display the `Discounts` line.


- Possible side-effects
This code creates the partial cart rules. If `$values['tax_excl'] > 0` is not true (and if I'm not wrong) it only means that cart rule is a free gift, so no need to create a new partial cart rule.

PR #19188, opened 2020-05-18
Docker-compose is failing due to missing build assets

PierreRambaud:
| Questions     | Answers
| Branch?       | develop
| Description?  | Since we removed assets from the source, we must add `make assets` in the docker_run_git.sh file when using `docker-compose`. Also, remove the check for autoload.php, we need to rerun the `composer install` command each time in case of we change the branch. Otherwise you have to remove manually the `vendor` directory which is unproductive.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Rebuild the container`docker-compose build --no-cache`, remove old volumes `docker-compose down -v` and run `docker-compose up`. 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19188)

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -7,4 +7,7 @@ RUN groupmod -g 1000 www-data \
 COPY /wait-for-it.sh /tmp/
 COPY /docker_run_git.sh /tmp/
 
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+RUN apt install -y nodejs
+
 CMD ["/tmp/docker_run_git.sh"]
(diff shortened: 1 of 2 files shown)


matks:
How could we be automatically notified than docker configuration is wrong ?
- Travis cron every night that attempts to deploy a shop and ping the FO ?
- GitHub Action in CI ?

matks (approved):

PierreRambaud:
@matks This configuration is only for development and not concerned the docker images we are released (based on downloaded zip files)

Progi1984 (approved):

matks:
Thank you @PierreRambaud

Issue #14405, opened 2019-06-28
[SF Migration] Summary of needed tech improvements for rework of Product Page


matks:
We're going to rework the [Product Page](https://github.com/PrestaShop/PrestaShop/issues/10632) in PS 1.7.8 .

Here is the list of technical issues we need to explore to find a suitable solution able to handle properly the usecase.

# Frontend components

### 1. Image dropzone

On 1st tab, there is a complex and powerful image manager component able to upload, edit and delete images and associated data.

Goal: build a stand-alone, configurable and maintainable JS component able to do all these actions
Actions are
- Add/upload image
- Delete image
- List uploaded images (includes drag-and-drop to choose order)
- Togglable edit panel on selected image

### 2. Combinations listing and actions (single action, bulk actions)

On "Combinations tab" there is a listing of all combinations for a given product. Today this listing is displayed whole even if it contains thousand of items. Moreover multiple actions are available on the listing items, some targeting a single row and some having bulk behavior.

Goal: build a stand-alone, configurable and maintainable JS component able to handle this listing.
It must provide
- paginations display (being able to query API endpoints to be given "next page" results)
- search feature (API endpoint being queried)
- checkboxes to be used for bulk actions

### 3.Combinations Generator

There is a component where you select a configuration and it generates combinations for the listing.

Goal: build a stand-alone, configurable and maintainable JS component able to handle this feature.

### 4.Data synchronization and prices computation in the browser 

Some data is duplicated, sometimes directly (same data on 2 tabs), sometimes indirectly on the page (same data but displayed differently on 2 tabs).
So when data 1 is updated, related data 2 must be updated accordingly. Currently there are some issues linked to that.

For example if you modify VAT included price, the VAT excluded must be updated too.

Goal: build a pluggable system that we can use to link fields and be able to describe the relationships between them. 
Also need to use a dedicated mathematical JS library to make sure price computations are accurate.

### 5. Category linking widget

On 1st tab you can link a product to multiple categories and there is a category tree to choose the target categories. It probably breaks when there are too many categories.

Goal: check current categories widget and see how it can scale when given thousand of categories input data.

https://github.com/PrestaShop/PrestaShop/issues/19281

### 6. Options tab multiple dynamic listings

On last "Options" tab there are multiple tabs that have dynamic content:
- add, list and edit custom fields
- add, list and edit attached files
- suppliers listing (multi-level listing)

[...]

PR #19196, opened 2020-05-18
[POC] Vuex state for product page and price/tax calculation in shared inputs

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Introduces Vuex store for product page state and some exemplary inputs that shares a value. Adds mathjs lib to calculate taxes for prices. POC looks good for a single service, but Im afraid that it will grow very big when we start implementing all actions (including auto-save), because every change must be tracked manually ( with vue.js it would be done out of the box using reactive props). Also Im not sure if we are able to fully reuse single store between add/edit page. (even tho they are almost identical, there might be some impactful differences).
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #14405 
| How to test?  |

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19196)


admin-dev/themes/new-theme/js/pages/product/price-refresher.js
```diff
@@ -0,0 +1,139 @@
...
+      el.addEventListener('change', (event) => {
+        if (forTaxIncluded) {
+          this.store.dispatch('updatePriceTaxIncluded', {
+            priceTaxIncluded: event.currentTarget.value,
+            taxRule: this.store.state.taxRule,
+          });
+        } else {
+          this.store.dispatch('updatePriceTaxExcluded', {
+            priceTaxExcluded: event.currentTarget.value,
+            taxRule: this.store.state.taxRule,
```
jolelievre: I don't think you need to provide the tax here, because the state already contains it (and you can see that you get your value from the state ^^) Only provide to the action the modified data

admin-dev/themes/new-theme/js/pages/product/store/actions.js
```diff
@@ -0,0 +1,55 @@
...
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+import * as types from './mutation-types';
+import * as taxCalculations from '../tax-calculations';
+
+/**
+ * Actions responsible for applying domain logic and updating the state after DOM changes.
```
zuk3975, line 30: Im not sure where to put these comments for whole file when its not a class. Looks abit strange just to float here :smile: doesnt it?

admin-dev/themes/new-theme/js/pages/product/edit/index.js
```diff
@@ -0,0 +1,39 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+import PriceRefresher from '../price-refresher';
+import createStore from '../store/index';
+import StateProvider from './state-provider';
+
+const {$} = window;
+const state = new StateProvider().get();
```
jolelievre: You should add this in the `$(() => {});` To make sure the page is loaded and the input are present (just in case)

admin-dev/themes/new-theme/js/pages/product/price-refresher.js
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  /**
+   * Listens for price tax included/excluded inputs changes
+   *
+   * @param selector
+   * @param forTaxIncluded
+   */
+  listenToPriceChangesInDom(selector, forTaxIncluded) {
+    const elements = document.querySelectorAll(selector);
```
jolelievre, line 91: I think this part could be turned into a helper class because we are gonna need this kind of behaviour a lot, something like

```js
export default class StoreBinder {
    constructor(store) {
        this.store = store;
    }

   bindInput(inputSelector, stateField, storeAction) {
        const elements = document.querySelectorAll(inputSelector);

        elements.forEach((el) => {
            el.addEventListener('change', (event) => {
                const payload = {};
                payload[stateField] = event.currentTarget.value;
                this.store.dispatch(storeAction, payload);
            });
        });
    }

   updateInput(inputSelector, stateField, mutationName) {
        this.store.subscribe((mutation) => {
            if (mutation.type === mutationName) {
                return;
            }

            const elements = document.querySelectorAll(this.taxRulesGroupSelections);
            elements.forEach((el) => {
                el.value = this.store.state[stateField];
            });
        });
    }
}
```

jolelievre: Then this allows to do stuff like that:

```js
bindInputs() {
   storeBinder.bindInput(this.taxIncludedInputs, 'priceTaxIncluded', 'updatePriceTaxIncluded');
   storeBinder.updateInput(this.taxIncludedInputs, 'priceTaxIncluded', types.SET_PRICE_TAX_INCLUDED);
}
```
We could even have a two way binding helper that performs these two calls:
```js
storeBinder.twoWayBind(this.taxIncludedInputs, 'priceTaxIncluded', 'updatePriceTaxIncluded', types.SET_PRICE_TAX_INCLUDED)
```

admin-dev/themes/new-theme/js/pages/product/price-refresher.js
```diff
@@ -0,0 +1,153 @@
...
+        }
+      });
+    });
+  }
+
+  /**
+   * Listens to tax rules selection changes
+   */
+  listenToTaxRulesGroupChangesInDom() {
+    const elements = document.querySelectorAll(this.taxRulesGroupSelections);
```
jolelievre, line 112: In this case the store binder could work as well as long as the store contains the list of `taxRulesGroups` The we just need to have a `updateTaxRulesGroupId` that will match the rate thanks to the id

admin-dev/themes/new-theme/js/pages/product/store/actions.js
jolelievre: It doesn't shock, but I'm not even sure it is required Considering its filename it seems obvious that it is defining the actions of the store for product page

admin-dev/themes/new-theme/js/pages/product/price-refresher.js
jolelievre: By the way for the `updateInput` maybe we can try and use the `watch` method from the store, this would avoid us from using the mutation type and we could use the state field name directly It would even allow enriched accessors like `'a.b.c'`, see:

https://vuex.vuejs.org/api/#watch
https://vuejs.org/v2/api/#vm-watch

admin-dev/themes/new-theme/js/pages/product/tax-calculations.js
```diff
@@ -0,0 +1,55 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
```
jolelievre, line 25: I don't know if you saw it but the default theme has this "component" for price computing:
https://github.com/PrestaShop/PrestaShop/blob/develop/admin-dev/themes/default/js/bundle/product/form.js#L1568

It has more rich features, I think it's a good idea to create a new one here But we could inspire from it to manage correctly the eco tax and stuff like that, I don't know if we should export functions like this or make it a class 🤔

Of course it's not to be done in this PR, this should be done in a dedicated PR and integrate automated tests for this new component, that would be great! This implementation is enough for the POC right now.

src/PrestaShopBundle/Form/Admin/Sell/Catalog/Product/ProductPriceType.php
```diff
@@ -0,0 +1,99 @@
...
+            ])
+        ;
+    }
+
+    /**
+     * @param array $taxRuleGroupChoices
+     *
+     * @return array
+     */
+    private function presentTaxRuleGroupChoices(array $taxRuleGroupChoices): array
```
jolelievre, line 89: Is this method necessary, the `ConfigurableTaxRuleGroupChoiceProvider` should return it correctly formatted no?

jolelievre:
@NeOMakinG @PierreRambaud I'd like your opinion on this Vue store, do you think it's a good idea? Is it implemented correctly? This is a POC so it can be improved but at least is it on good tracks?

src/PrestaShopBundle/Form/Admin/Sell/Catalog/Product/ProductPriceType.php
zuk3975: Im getting [name => [id, rate]] because i need to set data attr with tax rate value on every option :/
Not sure if there is a better way.
You can also take a look at modification i've made in legacy class to include the tax-rate.

admin-dev/themes/new-theme/js/pages/product/price-refresher.js
zuk3975: I've tried your suggestion. It seems ok with simple values bind like input.value === state.value.
But once we need something more it tears apart ( like when state contains nested object or when we need to access some data attributes from event).
Overall this "manual binding" looks like we are trying to invent the bike :(

matks (approved):
Aside from @jolelievre, I think this looks good. I'm in favor or merging it so other PRs can use it as a base, and refactor the code from @jolelievre feedbacks in another PR.

About the "we reinvent the bike" feeling: you're 50% right.
**Yes**, we reinvent the bike because other implementations, be it Angular or Vue, have already solved this solution with nice and robust behaviors
**No**, we do not reinvent the bike because we do not bring a new solution to a known problem with known solutions. We reimplement an idea that has been validated.

It's like design patterns. If you see a design pattern implemented in PHP and Java, it looks like PHP reinvents the bike but actually the idea is shared but implemented in 2 different usecases

admin-dev/themes/new-theme/js/pages/product/store/mutations.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+import * as types from './mutation-types';
+
+/**
+ * Mutations for Vuex store
+ */
+export default {
+  [types.SET_PRICE_TAX_EXCLUDED](state, price) {
+    state.priceTaxExcluded = price;
```
PierreRambaud, line 33: I don't understand why prices are in a VueX store, it's a not a state, it's a property / data in components.

PierreRambaud (changes requested):
I'm against using vuex each time you want to store a property / data.
Waiting for @NeOMakinG feedback

admin-dev/themes/new-theme/js/pages/product/store/mutations.js
NeOMakinG: It's more like a property THEN a state inside a component, but if the price is used anywhere outside the component, using it in a store isn't a bad practice I think.

Remember, store should be handling states that are global, like an user account available everywhere in the app, if the price is global and will be manipulated anywhere but not only in one component, that should be fine

PierreRambaud: We must be sure it needs to be used as a state like an user account, and not a way to store everything when properties are enough.

admin-dev/themes/new-theme/js/pages/product/price-refresher.js
NeOMakinG: This is inventing the bike yes, but if we don't use vue, we're kinda forced to do this, and using vue would make it harder for now as we would need modify form_widget in order to add binding to it (v-model, v-prop..) but anyway this looks like this isn't the path we chosen.

NeOMakinG:
I'm kinda blocked, I wanna ask to respect some good practices and say "for the moment it's ok since states in the store looks global" but since we'll almost never use vue components in that page, I'm kinda sure every states will go inside the store which is not designed for...

And the binding isn't shocking as we anyway need bindings but don't want to use vue, we're forced to invent bindings again

admin-dev/themes/new-theme/js/pages/product/store/mutations.js
jolelievre: That's the idea here, the price can be edited in two different tabs (Basic infos and Price tab) But the inputs need to be synced between those tabs So we need to integrate a way in a Symfony page to sync some inputs

Vuex seemed like a good candidate to use as a basis, as it already manages this kind of behaviour inside a Vue app, and since we tend to use Vue in the future at least the store part would already be based on Vuex

Keep in mind that THIS IS NOT a Vue app, it's a Symfony page, with form inputs, most of the logic will be done using jquery like ALL the other migrated pages. Vuex here is only used as a common global state for the page to make sure every input is up to date

PierreRambaud: This is why it's also weird, loading a Vue instance for not using it only to have VueX loaded when you can have a simple `Object` management is not a good idea.

FYI:
Vuex checks for [the existence of Vue](https://github.com/vuejs/vuex/blob/v3.0.1/dist/vuex.js#L308).
Vuex [depends largely on Vue for its reactivity inner workings](https://github.com/vuejs/vuex/blob/v3.0.1/dist/vuex.js#L330).

https://github.com/vuejs/vuex/issues/1172#issuecomment-370692728

eternoendless:
I don't know if Vuex is a good idea if we're not using Vue. 

The way I see it one of the main problem with using Vuex without Vue is that we need to do the two way data binding by hand. Since data is not bound to fields automatically, it kinda loses one of the main points of using it. You could do exactly the same thing using a simpler structure, where setters dispatch events which observers listen to to update their data.

Keep in mind the original problem: there was no centralized state in the page. Compared with an observer pattern, the only advantage Vuex provides us right now is event sourcing, which is nice but adds a lot of complexity for nothing really. An ES log is nice when you intend to implement undo or for debugging when data is being bound automatically (which the case with Vue), none of these cases applies to us.

Consider this example:

```js
class Store {
	constructor(state, eventManager, priceCalculator) {
		this.state = state;
		this.eventManager = eventManager;
		this.priceCalculator;
	}

	subscribe(event, callback) {
		this.eventManager.subscribe(event, callback);
	}

	updatePriceTaxExcluded(newPrice) {
		this._setPriceTaxExcluded(newPrice);
		this._setPriceTaxIncluded(this.priceCalculator.addTax(newPrice));
	}

	updatePriceTaxExcluded(newPrice) {
		this._setPriceTaxIncluded(newPrice);
		this._setPriceTaxExcluded(this.priceCalculator.removeTax(newPrice));
	}

	getPriceTaxExcluded() {
		return this.state.price;
	}

	getPriceTaxIncluded() {
		return this.state.price;
	}

	_setPriceTaxExcluded(newPrice) {
		this.state.price = newPrice;
		this.eventManager.fire('onUpdatePriceTaxExcluded', newPrice);
	}

	_setPriceTaxIncluded(newPrice) {
		this.state.priceTaxIncluded = newPrice;
		this.eventManager.fire('onUpdatePriceTaxIncluded', newPrice);
	}
}


class TaxIncludedSubscriber {
	constructor(selector, store) {
		this.event = store.subscribe('onUpdatePriceTaxExcluded', (newPrice) => {
			const elements = document.querySelectorAll(selector);
            elements[0].value = newPrice;
		});
	}
}
```

In this simplified example, there's a simple store. It's initialized with data, then provides setters and getters to manipulate that data. Setters issue events, to which subscribers can attach in order to be notified by data updates, and update their related components. So when an hypothetical `Tax Excluded Field` is updated by the user, it calls `store.updatePriceTaxExcluded()` which issues `onUpdatePriceTaxExcluded` that `TaxIncludedSubscriber` is listening to. It then updates its view with the new data.

The reverse can be performed by subscribing to the `change` event on a form field and using its value to call the store setter.

NeOMakinG:
@eternoendless classes doesn't have a good reputation in the JS ecosystem, it's purely syntactic sugar that we can avoid using Object methods easily, what about sticking to Vue, React core dev workflow using functional components ?

kpodemski:
I'm a bit confused, do we really have a "state" when it comes to prices on product page?
having different taxes and changing them is not really changing a state of the application? am i right?

imagine that we have dozen of scenarios like this on product page, do we really need to implement different state for disabled product, enabled product, product with redirection, product with combinations, product as a pack etc.? complexity of this would be massive and i don't think about product management as having a real-world SPA application where you could gain from VueX, flux or other state management library

for me - this is just binding proper data based on certain values (properties), not "state" of the product, i have the feeling that most of the time people are using Redux, Vuex and all they have is more complex and harder to manage application, without real perks

@eternoendless
> I don't know if Vuex is a good idea if we're not using Vue.

i thought that 1.7.8 product page would be on Vue like Stock and few other things, plans changed?

matks:
@zuk3975 I guess we might close this PR?

PR #19249, opened 2020-05-21
Fix illogical ternary condition for webservice filtering

### LINKED ISSUE #19128: SQL performance improvement in WebserviceRequest
#### Describe the bug

https://github.com/PrestaShop/PrestaShop/blob/develop/classes/webservice/WebserviceRequest.php#L1734

`(Validate::isFloat(pSQL($filterValue)) ? 'LIKE' : '=')` 
The ternary condition is inverted, it uses a `LIKE` statement instead of a `=` statement for int and float and uses a `=` statement for strings.

By the way, is it useful to use a `LIKE` statement here ?
Since in line [1703](https://github.com/PrestaShop/PrestaShop/blob/develop/classes/webservice/WebserviceRequest.php#L1703) the presence of `%` is checked for the use of a `LIKE` statement. If in `$filterValue` there is no `%`, whether `$filterValue` is an int, float or string, we can use a simple `=`

#### Expected behavior

Use `= "42"` instead of `LIKE "42"` when $filterValue is an int or a float

#### Additional information

* PrestaShop version: 1.7.6.5
* PHP version: 7.2

DelecroixQuentin:
| Questions     | Answers
| Branch?       | develop
| Description?  | #19128 
| Type?         | improvement
| Category?     | WS
| BC breaks?    |no
| Deprecations? | no
| Fixed ticket? | Fixes #19128 .
| How to test?  | Unit tests pass

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19249)


matks:
@DelecroixQuentin thanks for your contribution !

In the issue you said:
> By the way, is it useful to use a LIKE statement here ?
Since in line 1703 the presence of % is checked for the use of a LIKE statement. If in $filterValue there is no %, whether $filterValue is an int, float or string, we can use a simple =

However you keep the ternary condition, but inverted, in this PR. I thought you would simply remove it and put `=` hardcoded 😄 

As this `getSQLRetrieveFilter()` is quite complex, it will be hard to check all possible usecases manually. I think we might be able to unit test it because input and output are quite obvious. Would you be motivated to add these to your PR ? So not only unit tests would validate the behavior is good but also, in the future, if someone modifies this function again, the unit tests will tell him whether he introduces a regression or not.

Progi1984 (dismissed):

DelecroixQuentin:
Hello @matks,

> However you keep the ternary condition, but inverted, in this PR. I thought you would simply remove it and put `=` hardcoded 

Since I didn't really get an answer on this point, I preferred not to remove the condition alone. But I can delete it if you agree with it.

> As this `getSQLRetrieveFilter()` is quite complex, it will be hard to check all possible usecases manually. I think we might be able to unit test it because input and output are quite obvious. Would you be motivated to add these to your PR ? So not only unit tests would validate the behavior is good but also, in the future, if someone modifies this function again, the unit tests will tell him whether he introduces a regression or not.

As you say, this method is quite complex and I'm not sure I can do a full set of tests, especially if you take a look at the `manageFilters()` method to see how the `getSQLRetrieveFilter()` is used.

matks:
> As you say, this method is quite complex and I'm not sure I can do a full set of tests, especially if you take a look at the `manageFilters()` method to see how the `getSQLRetrieveFilter()` is used.

I will try something about it 😉if we have a valid set of tests, we will not be afraid removing totally the LIKE.

My idea is to use Test Driven Development
1. write unit tests for current implementation
2. find the bad usecases where LIKE is not used in right way
3. fix bad usecases => tests go red
4. put your implementation, that should fix the tests

sowbiba (dismissed):
Approved
I think @matks improvement can be done in addition to this PR

Progi1984:
Ping @DelecroixQuentin

matks:
@DelecroixQuentin I tried to initially test the whole WebserviceRequest but it's very hard to isolate.
So I only extracted the logic we look for, see https://github.com/PrestaShop/PrestaShop/pull/19389

Now you can choose
- we merge https://github.com/PrestaShop/PrestaShop/pull/19389 then you update your PR to insert it
- you retrieve my code from https://github.com/PrestaShop/PrestaShop/pull/19389 and it will all get merged together

We should probably add more usecases in the test to be sure it covers everything.

Progi1984:
@matks #19249 merged in `develop` branch

DelecroixQuentin:
Hello @matks @Progi1984,
Sorry I was busy last week, do I have to do anything to close this PR ?

matks:
> Hello @matks @Progi1984,
> Sorry I was busy last week, do I have to do anything to close this PR ?

Well, I added a set of unit tests to this little method to help us validate the PR. The tests are waiting into `develop`

So what you need to do:
- rebase your PR on `develop` to get latest changes (my tests included)
- then when you update your PR, Travis will run the tests and we'll see some tests failing (because you have modified the behavior)
- from the tests which failed we can check if it's a needed failure (= we want to change this behavior) or a bug (= we did not want to modify this behavior)

Are you able to do a rebase ? Else I can do it 😊

DelecroixQuentin:
Ok thanks @matks !
Can you tell me if we keep the `(Validate::isFloat(pSQL($filterValue)) ? 'LIKE' : '=')` or we replace it with a simple `=` ?
wdyt ?

matks:
> Ok thanks @matks !
> Can you tell me if we keep the `(Validate::isFloat(pSQL($filterValue)) ? 'LIKE' : '=')` or we replace it with a simple `=` ?
> wdyt ?

From my tests, we should reaplce it with a simple `=` . Inverting the LIKE introduces weird cases where, again, LIKE is used instead of a =

DelecroixQuentin:
One test failed :/ https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/342069202#L581

But the real question is: why `' AND `price` LIKE "18.2"'` ? This test was supposed to work with the previous code, but with the removal of the ternary condition, this test can't pass, right?

matks:
> One test failed :/ https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/342069202#L581
> 
> But the real question is: why `' AND `price` LIKE "18.2"'` ? This test was supposed to work with the previous code, but with the removal of the ternary condition, this test can't pass, right?

Yes indeed 😄 that's why I said:

"- rebase your PR on develop to get latest changes 
- then when you update your PR, Travis will run the tests and **we'll see some tests failing** (because you have modified the behavior)
- from the tests which failed we can check if it's a needed failure (= we want to change this behavior) or a bug (= we did not want to modify this behavior)"

So we can see that indeed this test looks bad as the outcome seems irrelevant. We can now modify the test to make it pass, and we know what the code used to do, what the code does now, and if it's logical 😄

DelecroixQuentin:
Oh ok, my bad, I didn't get your message like that, but it does seem to make sense.

I'm updating the test !

PierreRambaud (approved):

matks (approved):
Looks great now !

@PierreRambaud Can we consider the unit tests validate this PR and we don't need manual QA ?

PierreRambaud:
> Looks great now !
> 
> @PierreRambaud Can we consider the unit tests validate this PR and we don't need manual QA ?

Yes, and this is why unit tests are important :p

matks:
Then I'm happy to merge this PR 👍 big thanks to @DelecroixQuentin

PR #19277, opened 2020-05-22
Enable addCss and addJs for migrated pages and ControllerStack

### LINKED ISSUE #13943: How to add a JS extension for your modern module
# Summary

1. Enable modules to modify already-bundled JS files

```
import EmailSendingTest from './email-sending-test';
import SmtpConfigurationToggler from './smtp-configuration-toggler';
import Grid from '../../components/grid/grid';
import ReloadListActionExtension from '../../components/grid/extension/reload-list-extension';
import ...

const $ = window.$;

$(() => {
    const emailLogsGrid = new Grid('email_logs');
    emailLogsGrid.addExtension(new ReloadListActionExtension());
    emailLogsGrid.addExtension(new ExportToSqlManagerExtension());
    emailLogsGrid.addExtension(new FiltersResetExtension());
    emailLogsGrid.addExtension(...);
    const selector = '.my-selector';
    new EmailSendingTest(selector);
    new SmtpConfigurationToggler();
});
```

So we want to allow modules to customize this

2. Enable modules to inject JS and CSS files into a page (see https://github.com/PrestaShop/PrestaShop/pull/19277/)

# Context

https://github.com/friends-of-prestashop/demo-cqrs-hooks-usage-module

I have added `ToggleColumn` in Customers list using the module and it works like a charm. `ToggleColumn` required javascript extension  added to the grid `new ColumnTogglingExtension()`. To my luck, Customers list has this extension but developers might face the issues when they add column but the list does not have this extension in javascript compiled.

Possible solutions:
1) After creating sample doc about module creation we should create a sub-doc with advanced use cases such as **adding extension from module** - this will display webpack configuration, using hook to add your new javascript to existing list etc...

2) Provide a mechanism to be used from inside a module to add a specific JS/CSS file to the extended page, like legacy `addCss()`

matks:
| Questions     | Answers
| Branch?       | develop (I need to rebase)
| Description?  | Enable the use of `addCSS()` and `addJS()` in migrated pages, introduces ControllerStack
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes partially https://github.com/PrestaShop/PrestaShop/issues/13943
| How to test?  | See below

# How to test

Install module 
[demotestaddcss.zip](https://github.com/PrestaShop/PrestaShop/files/5232476/demotestaddcss.zip)

Open "customer" listing BO page, a "It works!" alert appears because the module successfully injected `myscript.css` inside the page.

# PR content

## addCss & friends

This PR enables the use of `addCSS()`, `addJS()`, `removeCSS()` and `removeJS()` in Symfony controllers. It was a very useful function available for legacy controllers.

So now you can do `$customerController->addJS($myJavascriptFile);` to inject a JS file into Customer BO page. That is great! 🎉 

## Usage inside hooks

If you are using a Core controller, it's very easy to use `addCSS()` or `addJS()`. Same if you are using a module Symfony controller.

**But a second issues comes:** it is actually hard to use `addCss` or `addJs` **in a hook** ⛵ !

See the following example (from the sample module)
```
public function hookActionCustomerGridDefinitionModifier(array $params)
    {
        
    }
```
You are inside this hook, which goal is to modify the "Customer Grid". Let's say you want to add a new column AND the JS file in charge of managing this column (for example the new column uses a specific ColorPicker). How do you inject the JS file using `addJS()`

What you need to do is:
- explore the code to find out that this hook is fired when user browses "Customer Grid BO page", which is powered by CustomerController
- retrieve the `$customerController` instance

Now that you know the Controller that you want to target is CustomerController, you can do this
```
    public function hookActionCustomerGridDefinitionModifier(array $params)
        $controller = $this->get('PrestaShopBundle\Controller\Admin\Sell\Customer\CustomerController');
        $controller->addJS('myscript.js');
```

## Hooks fired from multiple places

On this hook we are lucky, the hook will be triggered only on Customer Grid page. So we know 100% "the controller that we need to retrieve is CustomerController". **However there are hooks that can be fired from multiple BO pages.**

**So we need a way to know "I am plugged on hook Foo, what is the Controller that handled the HTTP request ?"

In a hook function, the hook params contain PS version, route name, request, but not controller. From request it is possible to find the controller name but it's very dirty.

**Second issue**: even you can find the controller name, but to retrieve the controller _instance_ is even more complex.

➡️ So I did this

### Use controller stack

I introduced a ControllerStack whose behavior is similar to Symfony RequestStack. It listens to Symfony kernel and registers the controllers being matched to HTTP requests.

This allows you to retrieve the "current controller".

Example:
```
    public function hookActionCustomerGridDefinitionModifier(array $params)

        $controller = $this->get('prestashop.core.routing.controller_stack')->getCurrentController();
        $controller->addJS('myscript.js');
```

So this allows module developer to inject his JS from hook without knowing which controller handled the HTTP request.

## Documentation

As this is for module developers, I will need to write a documentation about it when it's merged

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19277)

### DIFF
--- a/classes/controller/Controller.php
+++ b/classes/controller/Controller.php
@@ -404,8 +404,6 @@ abstract class ControllerCore
      * @param string $css_media_type
      * @param int|null $offset
      * @param bool $check_path
-     *
-     * @return true
      */
     public function addCSS($css_uri, $css_media_type = 'all', $offset = null, $check_path = true)
     {
--- a/src/Adapter/LegacyContext.php
+++ b/src/Adapter/LegacyContext.php
@@ -244,7 +244,7 @@ class LegacyContext
         );
 
         if (null !== $previousContextController) {
-            $this->preserveCssAndJSAssets($previousContextController, $originCtrl);
+            $this->transferCssAndJSAssets($previousContextController, $originCtrl);
         }
 
         $originCtrl->run();
@@ -362,7 +362,7 @@ class LegacyContext
      * @param Controller $previousContextController
      * @param AdminLegacyLayoutControllerCore $originCtrl
      */
-    private function preserveCssAndJSAssets($previousContextController, AdminLegacyLayoutControllerCore $originCtrl): void
+    private function transferCssAndJSAssets($previousContextController, AdminLegacyLayoutControllerCore $originCtrl): void
     {
         if (!empty($previousContextController->js_files)) {
             array_map(function ($jsAsset) use ($originCtrl) {
--- a/src/Adapter/LegacyContext.php
+++ b/src/Adapter/LegacyContext.php
@@ -224,6 +224,8 @@ class LegacyContext
         $metaTitle = '',
         $useRegularH1Structure = true
     ) {
+        $previousContextController = $this->getContext()->controller;
+
         $originCtrl = new AdminLegacyLayoutControllerCore(
             $controllerName,
             $title,
@@ -237,6 +239,9 @@ class LegacyContext
             $metaTitle,
             $useRegularH1Structure
         );
+
+        $this->preserveCssAndJSAssets($previousContextController, $originCtrl);
+
         $originCtrl->run();
 
         return $originCtrl->outPutHtml;
@@ -342,5 +347,28 @@ class LegacyContext
     public function getAvailableLanguages()
     {
         return $this->getLanguages(false);
+    }
+
+    /**
+     * This will transfer registered CSS and JS assets
+     * from previous context controller instance
+     * to AdminLegacyLayoutController
+     *
+     * @param Controller $previousContextController
+     * @param AdminLegacyLayoutControllerCore $originCtrl
+     */
+    private function preserveCssAndJSAssets($previousContextController, AdminLegacyLayoutControllerCore $originCtrl): void
+    {
+        if (!empty($previousContextController->js_files)) {
+            array_map(function ($jsAsset) use ($originCtrl) {
+                $originCtrl->addJS($jsAsset);
+            }, $previousContextController->js_files);
+        }
+
+        if (!empty($previousContextController->css_files)) {
+            array_map(function ($cssAsset) use ($originCtrl) {
+                $originCtrl->addCss($cssAsset);
+            }, $previousContextController->css_files);
+        }
     }
 }
--- a/src/Adapter/LegacyContext.php
+++ b/src/Adapter/LegacyContext.php
@@ -224,7 +224,10 @@ class LegacyContext
         $metaTitle = '',
         $useRegularH1Structure = true
     ) {
-        $previousContextController = $this->getContext()->controller;
+        $previousContextController = null;
+        if (null !== $this->getContext()->controller) {
+            $previousContextController = $this->getContext()->controller;
+        }
 
         $originCtrl = new AdminLegacyLayoutControllerCore(
             $controllerName,
@@ -240,7 +243,9 @@ class LegacyContext
             $useRegularH1Structure
         );
 
-        $this->preserveCssAndJSAssets($previousContextController, $originCtrl);
+        if (null !== $previousContextController) {
+            $this->preserveCssAndJSAssets($previousContextController, $originCtrl);
+        }
 
         $originCtrl->run();
 
--- a/src/Core/Routing/ControllerStack.php
+++ b/src/Core/Routing/ControllerStack.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\PrestaShop\Core\Routing;
+
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+
+/**
+ * Inspired by \Symfony\Component\HttpFoundation\RequestStack
+ */
+class ControllerStack
+{
+    /**
+     * @var Controller[]
+     */
+    private $controllers = [];
+
+    /**
+     * @param Controller $controller
+     */
+    public function push(Controller $controller)
+    {
+        $this->controllers[] = $controller;
+    }
+
+    /**
+     * @return Controller|null
+     */
+    public function pop()
+    {
+        if (!$this->controllers) {
+            return null;
+        }
+
+        return array_pop($this->controllers);
+    }
+
+    /**
+     * @return Controller|null
+     */
+    public function getCurrentController()
+    {
+        return end($this->controllers) ?: null;
+    }
+
+    /**
+     * Gets the master Controller.
+     *
+     * @return Controller|null
+     */
+    public function getMasterController()
+    {
+        if (!$this->controllers) {
+            return null;
+        }
+
+        return $this->controllers[0];
+    }
+
+    /**
+     * Returns the parent request of the current.
+     *
+     * If current Controller is the master controller, it returns null.
+     *
+     * @return Controller|null
+     */
+    public function getParentController()
+    {
+        $pos = \count($this->controllers) - 2;
+
+        if (!isset($this->controllers[$pos])) {
+            return null;
+        }
+
+        return $this->controllers[$pos];
+    }
+}
--- a/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
+++ b/src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
@@ -526,4 +526,61 @@ class FrameworkBundleAdminController extends Controller
             $e->getMessage()
         );
     }
+
+    /**
+     * Adds a new stylesheet(s) to the page header.
+     *
+     * @param string|array $cssUri Path to CSS file, or list of css files like this : array(array(uri => media_type), ...)
+     * @param string $cssMediaType
+     * @param int|null $offset
+     * @param bool $checkPath
+     */
+    protected function addCSS(
+        $cssUri,
+        $cssMediaType = 'all',
+        $offset = null,
+        $checkPath = true): void
+    {
+        $this->getContext()->controller->addCSS(
+            $cssUri, $cssMediaType, $offset, $checkPath
+        );
+    }
+
+    /**
+     * Adds a new JavaScript file(s) to the page header.
+     *
+     * @param string|array $jsUri Path to JS file or an array like: array(uri, ...)
+     * @param bool $checkPath
+     */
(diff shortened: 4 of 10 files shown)


classes/controller/Controller.php
```diff
@@ -404,8 +404,6 @@ public function setRedirectAfter($url)
      * @param string $css_media_type
      * @param int|null $offset
      * @param bool $check_path
-     *
-     * @return true
```
matks, line 408: Not in the scope of this PR but I fixed it while working on it

src/Adapter/LegacyContext.php
```diff
@@ -214,6 +214,8 @@ public function getLegacyLayout(
         $metaTitle = '',
         $useRegularH1Structure = true
     ) {
+        $previousContextController = $this->getContext()->controller;
```
PierreRambaud: You maybe need to check there is a controller in the context? 
I can't trust the `getContext` method :sweat_smile:

matks: Trust me 🤩

PierreRambaud (dismissed):

Progi1984 (dismissed):

src/Adapter/LegacyContext.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/19277/commits/a4cd78e376c097f7b5c407e4934d715432c7e2b6

Progi1984 (dismissed):

PierreRambaud (dismissed):

src/Adapter/LegacyContext.php
```diff
@@ -333,4 +343,27 @@ public function getAvailableLanguages()
...
+
+    /**
+     * This will transfer registered CSS and JS assets
+     * from previous context controller instance
+     * to AdminLegacyLayoutController
+     *
+     * @param Controller $previousContextController
+     * @param AdminLegacyLayoutControllerCore $originCtrl
+     */
+    private function preserveCssAndJSAssets($previousContextController, AdminLegacyLayoutControllerCore $originCtrl): void
```
jolelievre:
```suggestion
    private function transferCssAndJSAssets($previousContextController, AdminLegacyLayoutControllerCore $originCtrl): void
```

or `copyCssAndJSAssets` would be easier to understand

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
```diff
@@ -526,4 +526,61 @@ protected function getErrorMessageForException(Exception $e, array $messages)
...
+     * @param int|null $offset
+     * @param bool $checkPath
+     */
+    protected function addCSS(
+        $cssUri,
+        $cssMediaType = 'all',
+        $offset = null,
+        $checkPath = true): void
+    {
+        $this->getContext()->controller->addCSS(
```
jolelievre, line 555: No need to check if `controller` is present here? If I remember correctly a legacy controller is always created along with the context, but it's a behaviour we aim to remove in time, so it might no always be the case
I also remember of cases when this `controller` field was not properly set (CLI or Installer environment I'm not sure any more) so even as it is now maybe it's better to check it

jolelievre (commented):
Just a few comments but the idea seems good to me

One thing I'm not sure to understand though: we add js/css fiels references to a **legacy** controller so how can it work for a **symfony** controller? Is it thanks/because of the legacy layout that integrates the files from the controller automatically?

If so I guess when we remove the legacy layout we will have to store those references somewhere else, and integrate them in the twig layout automatically as well? Then maybe a comment to remind us to do so could be added in the new `FrameworkBundleAdminController` functions?

PS: I just need answers to these questions and then I can approve this with more peace of mind ^^

matks:
> Is it thanks/because of the legacy layout that integrates the files from the controller automatically?

Yes

> If so I guess when we remove the legacy layout we will have to store those references somewhere else, and integrate them in the twig layout automatically as well

Yes

SD1982:
Hi @matks Don't hesitate to ping the QA team when the test module is ready ! Thanks !

src/Core/Routing/ControllerStack.php
```diff
@@ -0,0 +1,100 @@
+<?php
```
matks: @Progi1984 You will tell me to add declare strict type
I trust you 🙏

src/Core/Routing/ControllerStack.php
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+namespace PrestaShop\PrestaShop\Core\Routing;
+
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+
+/**
+ * Inspired by \Symfony\Component\HttpFoundation\RequestStack
+ */
+class ControllerStack
```
matks: If PR intent is approved, I will add unit tests

src/PrestaShopBundle/Resources/config/services/core/routing.yml
```diff
@@ -0,0 +1,6 @@
+services:
+  _defaults:
+    public: true
+
+  prestashop.core.routing.controller_stack:
```
matks: I'm a bit sad I had to create a new YAML file to add this service but I could not find a relevant file already existing 😞

src/Adapter/LegacyContext.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/19277/commits/db3bcbb7490699d6071d2094d1ab1f708bb3220c

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/19277/commits/db3bcbb7490699d6071d2094d1ab1f708bb3220c

matks:
> Then maybe a comment to remind us to do so could be added in the new `FrameworkBundleAdminController` functions?

Fixed in https://github.com/PrestaShop/PrestaShop/pull/19277/commits/db3bcbb7490699d6071d2094d1ab1f708bb3220c

src/Core/Routing/ControllerStack.php
Progi1984: It's the rule for new files ;) Thanks :)

Progi1984: https://devdocs.prestashop.com/1.7/development/coding-standards/#strict-typing

src/Core/Routing/ControllerStack.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        return end($this->controllers) ?: null;
+    }
+
+    /**
+     * Gets the master Controller.
+     *
+     * @return Controller|null
+     */
+    public function getMasterController()
```
jolelievre:
```suggestion
    public function getRootController()
```

Suggestion

src/PrestaShopBundle/Controller/Admin/FrameworkBundleAdminController.php
```diff
@@ -526,4 +528,93 @@ protected function getErrorMessageForException(Exception $e, array $messages)
...
+     * @param bool $checkPath
+     */
+    public function addCSS(
+        $cssUri,
+        $cssMediaType = 'all',
+        $offset = null,
+        $checkPath = true): void
+    {
+        if (null === $this->getContext()->controller) {
+            throw new \RuntimeException('No context controller available');
```
jolelievre:
```suggestion
            throw new RuntimeException('No context controller available');
```

Suggestion: With appropriate `use` statement

src/Core/Routing/ControllerStack.php
jolelievre: I wonder if it should be in the core Namespace or rather in the Bundle one, since it's titghtly related to routing, event listeners, ... Which are themselves tightly related to the Symfony framework rather than the Core business/domain logic
Maybe it also explains why you couldn't find a yaml related to routing in the core folder (there is one in the bundle folder)

src/PrestaShopBundle/Resources/config/services/bundle/event_listener.yml
```diff
@@ -101,3 +101,10 @@ services:
           - '@logger'
         tags:
           - { name: kernel.event_subscriber }
+
+    prestashop.bundle.controller_stack:
+        class: PrestaShopBundle\EventListener\ControllerListener
+        arguments:
+          - '@prestashop.core.routing.controller_stack'
+        tags:
+          - { name: kernel.event_listener, event: kernel.controller, method: onKernelController }
```
jolelievre, line 110: Maybe you will need a priority parameter or something similar (if possible on this event) To make sure this listener is called early in the controller workflow

jolelievre (commented):
A few comments, mainly suggestion, questons or discussions
I like the idea of this controller stack (although I'd say it's more to manage in bundle than core domain - see comment)

But you don't seem to use it all?! I guess it's a work in progress and you intend to use it in `FrameworkAdminController` if the principle is validated? Is it required since I thought the idea was to transfer legacy assets (from modules) to modern ones? But you are watching for Symfony controller creations (which are not legacy ones) what is the purpose then?

matks:
> But you don't seem to use it all?! I guess it's a work in progress and you intend to use it in `FrameworkAdminController` if the principle is validated? Is it required since I thought the idea was to transfer legacy assets (from modules) to modern ones? But you are watching for Symfony controller creations (which are not legacy ones) what is the purpose then?

The ControllerStack is not needed for the logic in this PR but for modules.
The main issue it solves is this one: "I am writing a module that uses actionFooBar(), how do I call `addCss()` on the controller being used through this hook to add my CSS, within my hook ?"

With the controller stack, modules will be able to do
```
public function hookActionFooBar($parameters)
{
    $stack = $this->getControllerStack();
    $controller = $stack->getRootController();

    $controller->addCss(__DIR__.'/css/main.css');
}
```

Without the controller stack, modules cannot know "what is the controller that is handling the HTTP request".

EDIT: it will be easy to understand when I have written the Test Module for QA 😊

src/PrestaShopBundle/Controller/Api/ApiController.php
```diff
@@ -29,13 +29,14 @@
...
 use Psr\Log\LoggerInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 
-abstract class ApiController
+abstract class ApiController extends Controller
```
matks, line 39: ApiController should be Symfony controller IMO 🤔  this provides the ability for module and shop developers to benefit from Symfony built-in feature

src/PrestaShopBundle/Controller/Api/ApiController.php
```diff
@@ -52,7 +53,7 @@ public function setLogger(LoggerInterface $logger)
      */
     protected $container;
 
-    public function setContainer(ContainerInterface $container)
+    public function setContainer(ContainerInterface $container = null)
```
matks, line 56: Needed to comply with Symfony controller contract

src/PrestaShopBundle/Controller/Api/ApiController.php
```diff
@@ -194,16 +195,16 @@ protected function jsonResponse(
...
      *
-     * @param string $controller name of the controller
-     * @param array $accessLevel
+     * @param array $attributes
+     * @param string $subject name of the controller
      *
      * @return bool
      */
-    protected function isGranted(array $accessLevel, $controller)
+    protected function isGranted($attributes, $subject = null)
```
matks, line 203: Needed to comply with Symfony controller contract

src/PrestaShopBundle/Resources/config/services/bundle/routing.yml
```diff
@@ -53,3 +53,9 @@ services:
         class: 'PrestaShopBundle\Routing\Linter\AdminRouteProvider'
         arguments:
             - '@router'
+
+    prestashop.bundle.routing.controller_stack:
+        class: 'PrestaShopBundle\Routing\ControllerStack'
+
+    controller_stack:
```
matks, line 60: A useful alias 😁

matks:
Note to myself: need to rebase for 178

matks:
Closed in favor of https://github.com/PrestaShop/PrestaShop/pull/21016

PR #19400, opened 2020-05-27
Fix logic for adding anchor to URL

kpodemski:
| Questions     | Answers
| Branch?       | develop
| Description?  | as you can see in code, we couldn't remove anchor from URL at all, this PR fixes this issue and make sure that param $addAnchor is indeed used, to prevent this PR to be a "breaking change" i set "addAnchor" to true and added new new param to be provided for getAnchor method
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | try to get URL for a product with "addAnchor" set to `false` before and after this change

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19400)

### DIFF
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -127,8 +127,9 @@ class LinkCore
      * @param int|null $ipa ID product attribute
      * @param bool $force_routes
      * @param bool $relativeProtocol
-     * @param bool $addAnchor
+     * @param bool $addAnchor Either add or not # with combination informations after an URL
      * @param array $extraParams
+     * @param bool $withIdInAnchor Add or not id_attribute in combination anchor
      *
      * @return string
      */
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -142,8 +142,9 @@ class LinkCore
         $ipa = null,
         $force_routes = false,
         $relativeProtocol = false,
-        $addAnchor = false,
-        $extraParams = []
+        $addAnchor = true,
+        $extraParams = [],
+        $withIdInAnchor = false
     ) {
         $dispatcher = Dispatcher::getInstance();
 
@@ -237,7 +238,11 @@ class LinkCore
         if ($ipa) {
             $product = $this->getProductObject($product, $idLang, $idShop);
         }
-        $anchor = $ipa ? $product->getAnchor((int) $ipa, (bool) $addAnchor) : '';
+        
+        $anchor = '';
+        if ($addAnchor) {
+             $anchor = $ipa ? $product->getAnchor((int) $ipa, (bool) $withIdInAnchor) : '';
+        }
 
         return $url . $dispatcher->createUrl('product_rule', $idLang, array_merge($params, $extraParams), $force_routes, $anchor, $idShop);
     }
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -241,7 +241,7 @@ class LinkCore
         
         $anchor = '';
         if ($addAnchor) {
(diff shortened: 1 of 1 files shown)


Progi1984 (dismissed):

classes/Link.php
```diff
@@ -237,7 +238,11 @@ public function getProductLink(
         if ($ipa) {
             $product = $this->getProductObject($product, $idLang, $idShop);
         }
-        $anchor = $ipa ? $product->getAnchor((int) $ipa, (bool) $addAnchor) : '';
+
```
Progi1984:
```suggestion

```

Progi1984 (dismissed):

matks:
Hi @kpodemski ! Nice contribution !

3 things worry me:

1) Could you please also update the function phpDoc to explain the new argument ?

2) It is a bit confusing to have 2 parameters `$addAnchor` and `$withIdInAnchor`. Could we try  to find better names to make sure people understand exactly what this is doing ?

matks:
3) I am worried about upgrade.

## Currently

If someone uses this
```
getProductLink(
  $product,
  $alias,
  $category,
  ...
  $addAnchor,
  $extraParams
);
```
- with `$addAnchor` to true, he gets a link with an anchor and an ID inside
- with `$addAnchor` to false, he gets a link with an anchor without ID inside (so we can see the parameter has currently a bad name)


## If we merge this PR (so it will be 1.7.8)

If someone uses this
```
getProductLink(
  $product,
  $alias,
  $category,
  ...
  $addAnchor,
  $extraParams
);
```
- with `$addAnchor` to true, he gets a link with an anchor without ID inside => so developer did not modify its code but behavior has changed ⚠️ 
- with `$addAnchor` to false, he gets a link without anchor => so developer did not modify its code but behavior has changed ⚠️ 

This is my biggest worry. Even if previous code was wrong, some people have discovered it and started using the faulty code for good purposes. This is one of the horror of backward compatibility: sometimes we cannot fix a bug because people have built code (modules, themes, shops, integrations ...) that rely on this bad behavior.

Could we try to find a way to make sure that code in 1.7.8 behaves like before => so it does not create issues (people will tell us we have broken something 😭 ) **AND** it allows developers who look at it to obtain the right behavior ?

kpodemski:
@matks good point, but at this moment i'm not sure how it would be best to handle this, we could treat "addAnchor" as before and add an extra param but how do we name this? ;) "removeAnchor"? :D

hard to say, maybe others @PrestaShop/prestashop-core-developers have some thoughts about this?

matthieu-rolland (commented):
>
> hard to say, maybe others @PrestaShop/prestashop-core-developers have some thoughts about this?

Honestly I'm not sure, @matks point is correct, and as far as "**removeAnchor**" is concerned, adding yet another parameter to this already crowded method seems like a bad idea, BUT... I don't have anything better to propose 😓   (I mean without causing any BC breaks)

kpodemski:
ok guys, i'm closing this PR, for those who wants to have URL without anchor for product with combinations here's a little override for this:

```
class Link extends LinkCore
{
    public function getProductLink(
        $product,
        $alias = null,
        $category = null,
        $ean13 = null,
        $idLang = null,
        $idShop = null,
        $ipa = null,
        $force_routes = false,
        $relativeProtocol = false,
        $addAnchor = false,
        $extraParams = array()
    ) {
        $originalUrl = parent::getProductLink(...func_get_args());

        if (!$ipa) {
            return $originalUrl;
        }
        
        return explode('#', $originalUrl)[0];
    }
}
```

thanks 👍

PR #19402, opened 2020-05-27
Rename and deprecate confusing method in classes/Configuration.php

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | classes/Configuration.php had a method `getInt()` which was returning `array`. As the name suggests developer might expect it to return int and it is easy to confuse with adapter/Configuration::getInt(). Deprecated the method and added a new one with a more suitable name to avoid confusion and mistakes in future.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | yes
| Fixed ticket? | 
| How to test?  | nothing to test. If CI passes, we are good.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19402)

### DIFF
--- a/classes/Configuration.php
+++ b/classes/Configuration.php
@@ -255,6 +255,14 @@ class ConfigurationCore extends ObjectModel
     }
 
     /**
+     * @deprecated use Configuration::getConfigInMultipleLangs() instead.
+     */
+    public static function getInt($key, $idShopGroup = null, $idShop = null)
+    {
+        return self::getConfigInMultipleLangs($key, $idShopGroup, $idShop);
+    }
+
+    /**
      * Get a single configuration value (in multiple languages).
      *
      * @param string $key Configuration Key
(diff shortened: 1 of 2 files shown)


Progi1984 (dismissed):

classes/Configuration.php
```diff
@@ -254,6 +254,14 @@ public static function getGlobalValue($key, $idLang = null)
         return Configuration::get($key, $idLang, 0, 0);
     }
 
+    /**
+     * @deprecated use Configuration::getConfigInMultipleLangs() instead.
```
matks:
```suggestion
     * @deprecated use Configuration::getConfigInMultipleLangs() instead
```
PHP CS Fixer wants this

matks (approved):

PierreRambaud (approved):

classes/Configuration.php
```diff
@@ -255,7 +255,7 @@ public static function getGlobalValue($key, $idLang = null)
     }
 
     /**
-     * @deprecated use Configuration::getConfigInMultipleLangs() instead.
+     * @deprecated use Configuration::getConfigInMultipleLangs() instead
```
zuk3975, line 258: is there a known reason why cs fixer doesn't allow a dot at the end of comment?

matks: See `phpdoc_annotation_without_dot` https://github.com/FriendsOfPHP/PHP-CS-Fixer
"PHPDoc annotation descriptions should not be a sentence."

Dont know why but it's documented

matks:
Thank you @zuk3975

matks:
Just fixed PR body: actually there is deprecations

I dont know this will do 🤔

Progi1984:
@matks I add the label in this case ;)

PR #19407, opened 2020-05-27
UpdateProductDescriptionCommand

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | command for updating product description/description_short
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #14763
| How to test?  | Behat tests should cover basic use case. Not testable in Ui until integrated with form.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19407)

### DIFF
--- a/src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
+++ b/src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
@@ -31,13 +31,15 @@ namespace PrestaShop\PrestaShop\Adapter\Product\CommandHandler;
 use PrestaShop\PrestaShop\Adapter\Product\AbstractProductHandler;
 use PrestaShop\PrestaShop\Core\Domain\Product\Command\UpdateProductDescriptionCommand;
 use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\UpdateProductDescriptionHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\CannotUpdateProductException;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use PrestaShopException;
 use Product;
 
 final class UpdateProductDescriptionHandler extends AbstractProductHandler implements UpdateProductDescriptionHandlerInterface
 {
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
      */
     public function handle(UpdateProductDescriptionCommand $command): void
     {
@@ -55,6 +57,24 @@ final class UpdateProductDescriptionHandler extends AbstractProductHandler imple
         if (null !== $localizedShortDescriptions) {
             $this->validateDescriptions($product, $localizedShortDescriptions, true);
             $product->description_short = $localizedShortDescriptions;
+        }
+
+        try {
+            if (false === $product->update()) {
+                throw new CannotUpdateProductException(sprintf(
+                    'Error occurred when trying to update product #%s description',
+                    $product->id
+                ));
+            }
+        } catch (PrestaShopException $e) {
+            throw new CannotUpdateProductException(
+                sprintf(
+                    'Error occurred when trying to update product #%s description',
+                    $product->id
+                ),
+                0,
+                $e
+            );
         }
     }
 
--- a/src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
+++ b/src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
@@ -36,6 +36,9 @@ use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintExcepti
 use PrestaShopException;
 use Product;
 
+/**
+ * Handles UpdateProductDescriptionCommand using legacy object model
+ */
 final class UpdateProductDescriptionHandler extends AbstractProductHandler implements UpdateProductDescriptionHandlerInterface
 {
     /**
--- a/src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
+++ b/src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
@@ -41,6 +41,12 @@ use Product;
  */
 final class UpdateProductDescriptionHandler extends AbstractProductHandler implements UpdateProductDescriptionHandlerInterface
 {
+    /** @var string Product description property name */
+    private const DESCRIPTION = 'description';
+
+    /** @var string Product short description property name */
+    private const DESCRIPTION_SHORT = 'description_short';
+
     /**
      * {@inheritdoc}
      */
@@ -53,12 +59,12 @@ final class UpdateProductDescriptionHandler extends AbstractProductHandler imple
         $localizedShortDescriptions = $command->getLocalizedShortDescriptions();
 
         if (null !== $localizedDescriptions) {
-            $this->validateDescriptions($product, $localizedDescriptions, false);
+            $this->validate($product, $localizedDescriptions, self::DESCRIPTION);
             $product->description = $localizedDescriptions;
         }
 
         if (null !== $localizedShortDescriptions) {
-            $this->validateDescriptions($product, $localizedShortDescriptions, true);
+            $this->validate($product, $localizedShortDescriptions, self::DESCRIPTION_SHORT);
             $product->description_short = $localizedShortDescriptions;
         }
 
@@ -83,29 +89,27 @@ final class UpdateProductDescriptionHandler extends AbstractProductHandler imple
 
     /**
      * @param Product $product
-     * @param array $localizedDescriptions
-     * @param bool $forShortDescription
+     * @param array $localizedValues
+     * @param string $propertyName
      *
      * @throws ProductConstraintException
-     * @throws \PrestaShopException
+     * @throws PrestaShopException
      */
-    private function validateDescriptions(Product $product, array $localizedDescriptions, bool $forShortDescription): void
+    private function validate(Product $product, array $localizedValues, string $propertyName): void
     {
-        $descriptionType = 'description';
-        $errorCode = ProductConstraintException::INVALID_DESCRIPTION;
-
-        if ($forShortDescription) {
-            $descriptionType = 'short description';
+        if (self::DESCRIPTION === $propertyName) {
+            $errorCode = ProductConstraintException::INVALID_DESCRIPTION;
+        } else {
             $errorCode = ProductConstraintException::INVALID_SHORT_DESCRIPTION;
         }
 
-        foreach ($localizedDescriptions as $langId => $localizedDescription) {
-            if (true !== $product->validateField('description', $localizedDescription, $langId)) {
+        foreach ($localizedValues as $langId => $localizedValue) {
+            if (true !== $product->validateField($propertyName, $localizedValue, $langId)) {
                 throw new ProductConstraintException(
                     sprintf(
                         'Invalid %s "%s" in language with id "%s"',
-                        $descriptionType,
-                        $localizedDescription,
+                        $propertyName,
+                        $localizedValue,
                         $langId
                     ),
                     $errorCode
--- a/src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
+++ b/src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
@@ -70,10 +70,13 @@ final class UpdateProductDescriptionHandler extends AbstractProductHandler imple
 
         try {
             if (false === $product->update()) {
-                throw new CannotUpdateProductException(sprintf(
-                    'Error occurred when trying to update product #%s description',
-                    $product->id
-                ));
+                throw new CannotUpdateProductException(
+                    sprintf(
+                        'Error occurred when trying to update product #%s description',
+                        $product->id
+                    ),
+                    CannotUpdateProductException::FAILED_UPDATE_DESCRIPTIONS
+                );
             }
         } catch (PrestaShopException $e) {
             throw new CannotUpdateProductException(
@@ -81,7 +84,7 @@ final class UpdateProductDescriptionHandler extends AbstractProductHandler imple
                     'Error occurred when trying to update product #%s description',
                     $product->id
                 ),
-                0,
+                CannotUpdateProductException::FAILED_UPDATE_DESCRIPTIONS,
                 $e
             );
         }
--- a/src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
+++ b/src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
@@ -1,0 +1,93 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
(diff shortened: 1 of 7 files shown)


src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
```diff
@@ -0,0 +1,113 @@
...
+
+use PrestaShop\PrestaShop\Adapter\Product\AbstractProductHandler;
+use PrestaShop\PrestaShop\Core\Domain\Product\Command\UpdateProductDescriptionCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\UpdateProductDescriptionHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\CannotUpdateProductException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use PrestaShopException;
+use Product;
+
+final class UpdateProductDescriptionHandler extends AbstractProductHandler implements UpdateProductDescriptionHandlerInterface
```
jolelievre, line 42: I wonder if we should use `final` on these in case a module wants to extend and override it? 🤔

src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
```diff
@@ -0,0 +1,113 @@
...
+                throw new CannotUpdateProductException(sprintf(
+                    'Error occurred when trying to update product #%s description',
+                    $product->id
+                ));
+            }
+        } catch (PrestaShopException $e) {
+            throw new CannotUpdateProductException(
+                sprintf(
+                    'Error occurred when trying to update product #%s description',
+                    $product->id
```
jolelievre, line 84:
```suggestion
                    $command->getProductId()
```

I think it's preferable to use the command dat rather than the "unstable legacy ones when possible

src/Core/Domain/Product/Command/UpdateProductDescriptionCommand.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        return $this->localizedDescriptions;
+    }
+
+    /**
+     * @param string[]|null $localizedDescriptions
+     *
+     * @return UpdateProductDescriptionCommand
+     */
+    public function setLocalizedDescriptions(?array $localizedDescriptions): UpdateProductDescriptionCommand
```
jolelievre: Maybe we should keep the setters strict? They are only called when an array is available

(The getter keeps the double type since it's null by default)

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -94,11 +95,16 @@ public function addProduct(string $productReference, TableNode $table): void
      */
     public function assertLocalizedProperty(string $productReference, string $fieldName, string $localizedValues)
     {
+        $propertyName = $this->getPropertyByFieldName($fieldName);
         $product = $this->getProductByReference($productReference);
         $expectedLocalizedValues = $this->parseLocalizedArray($localizedValues);
 
         foreach ($expectedLocalizedValues as $langId => $value) {
-            if ($value !== $product->{$fieldName}[$langId]) {
+            if ('empty' === $value) {
```
jolelievre: I'm not at ease with this it includes some logic (special wording) in the fixture, and in the (unlikely) case we do want to insert `empty` as a value we can't

Maybe it would work without this just by using `"en-US:"` in the scenario? If not maybe a quick fix on `parseLocalizedArray` would allow it

src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
zuk3975: It has an interface, so i it seems ok to me.

src/Core/Domain/Product/Command/UpdateProductDescriptionCommand.php
zuk3975: Kind of. I had this thought. Yet it didn't seem to hurt and was generated like this with autocomplete, so i left it :D but yea, the only case i can think of where its usefull is that sometimes you dont need extra check if var is null before setting it, but its kinda rare and still doesn't make much sense :D

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
zuk3975: I agree it is strange and dull (it was the easiest thing to do though :D). sure ill check how to make something better

src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
zuk3975: Im not sure about this :thinking: 
Ofc its such a minor thing that it won't ever possibly make any difference, but IMO the product is the source of truth and its ok to use its properties after its loaded instead of using values which was used to load it.
Idk if you got me and if it makes sense? :sweat_smile:

src/Adapter/Product/AbstractProductHandler.php
```diff
@@ -0,0 +1,72 @@
...
+        try {
+            $product = new Product($productIdValue);
+
+            if ((int) $product->id !== $productIdValue) {
+                throw new ProductNotFoundException(sprintf(
+                    'Product #%s was not found',
+                    $productIdValue
+                ));
+            }
+        } catch (\PrestaShopException $e) {
```
jolelievre: add `use PrestaShopException;` statement

src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
jolelievre: Yeah ok, especially since `getProduct` checks that they are equal anyway

src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
```diff
@@ -0,0 +1,116 @@
...
+        $descriptionType = 'description';
+        $errorCode = ProductConstraintException::INVALID_DESCRIPTION;
+
+        if ($forShortDescription) {
+            $descriptionType = 'short description';
+            $errorCode = ProductConstraintException::INVALID_SHORT_DESCRIPTION;
+        }
+
+        foreach ($localizedDescriptions as $langId => $localizedDescription) {
+            if (true !== $product->validateField('description', $localizedDescription, $langId)) {
```
jolelievre:
```suggestion
            if (true !== $product->validateField($descriptionType, $localizedDescription, $langId)) {
```

And `$descriptionType` should be `description_short` for short descriptions

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -229,6 +261,48 @@ public function assertLastErrorIsInvalidTypeConstraint()
...
+    }
+
+    /**
+     * Map a technical property name with a more user oriented name from scenario.
+     *
+     * @param string $fieldName
+     *
+     * @return string
+     */
+    private function getPropertyByFieldName(string $fieldName): string
```
jolelievre: This seems overly complex to me, it would be easier to simply use `description_short` in the sentence I get the human readable idea but I'm not sure it has such a good complexity/value ratio

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -107,13 +110,42 @@ public function assertLocalizedProperty(string $productReference, string $fieldN
...
+     * @Then I update product :productReference descriptions with following information:
+     *
+     * @param string $productReference
+     * @param TableNode $table
+     */
+    public function updateLocalizedDescriptions(string $productReference, TableNode $table)
+    {
+        $data = $table->getRowsHash();
+        $descriptions = isset($data['description']) ? $this->parseLocalizedArray($data['description']) : null;
+        $shortDescriptions = isset($data['short description']) ? $this->parseLocalizedArray($data['short description']) : null;
```
```suggestion
        $shortDescriptions = isset($data['description_short']) ? $this->parseLocalizedArray($data['description_short']) : null;
```

src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
jolelievre: I wondered how the Behat test could work, but it's actually because both fields have the same validation method ^^ Lucky match 😛

src/Core/Domain/Product/Command/UpdateProductDescriptionCommand.php
```diff
@@ -0,0 +1,108 @@
...
+ */
+class UpdateProductDescriptionCommand
+{
+    /**
+     * @var ProductId
+     */
+    private $productId;
+
+    /**
+     * @var string[]|null key value pairs where key is the id of language
```
jolelievre: We should add a comment for nullable field like `if null no update is applied` To clarify the behaviour regarding the partial updates
This way just reading the command PHPDoc is self explanatory "ok i can update JUST the description with this"

jolelievre (changes requested):
One bug to fix in validation, the rest is more about simplifying and use the field name `description_short` everywhere instead of `short description`

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
zuk3975: it might have more value on other fields (like if its integer but needs to reflect some word), but ok, lets stick with simple solution while we can.

matks (dismissed):

src/Adapter/Product/AbstractProductHandler.php
```diff
@@ -0,0 +1,73 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+use PrestaShopException;
+use Product;
+
+/**
+ * Provides reusable methods for product handlers
+ */
+abstract class AbstractProductHandler
```
sowbiba: duplicated ? https://github.com/PrestaShop/PrestaShop/pull/19529

src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
```diff
@@ -0,0 +1,123 @@
...
+
+    /** @var string Product short description property name */
+    private const DESCRIPTION_SHORT = 'description_short';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(UpdateProductDescriptionCommand $command): void
+    {
+        $productId = $command->getProductId();
```
sowbiba: variable not necessary

src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
```diff
@@ -0,0 +1,123 @@
...
+     * @param Product $product
+     * @param array $localizedValues
+     * @param string $propertyName
+     *
+     * @throws ProductConstraintException
+     * @throws PrestaShopException
+     */
+    private function validate(Product $product, array $localizedValues, string $propertyName): void
+    {
+        if (self::DESCRIPTION === $propertyName) {
```
sowbiba, line 102: $errorCode = (self::DESCRIPTION === $propertyName) ? ProductConstraintException::INVALID_DESCRIPTION : ProductConstraintException::INVALID_SHORT_DESCRIPTION;

src/Core/Domain/Product/Exception/CannotUpdateProductException.php
```diff
@@ -31,4 +31,8 @@
  */
 class CannotUpdateProductException extends ProductException
 {
+    /**
+     * When updating product description or description_short fails
+     */
+    const FAILED_UPDATE_DESCRIPTIONS = 1;
```
sowbiba: is it useful to add errorCode when we have only one ? It's a real question :D @PrestaShop/prestashop-core-developers

sowbiba: and why do we have "1" when we have multiples of 10 in the exception just above ?

zuk3975: > is it useful to add errorCode when we have only one ? It's a real question :D

The number of codes will increase with each update command

copy-paste from chat:
> one more thing I noticed.
 Up until now we didn't use any codes for CannotUpdate{resource}Exception, because we used to update all the obj fields with one command.
The case with product is different, we could add at least one err code for each command. e.g. CannotUpdateProductException::FAILED_UPDATE_DESCRIPTIONS.
Else, we might have problems identifying the err source in controller later. (unless we try some new way of error handling).
Not sure tho, wdyt? (edited) 

this might just allow to identify what operation failed, becaues we will have tons of update commands.

src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
zuk3975: I would say its a personal preference, I imagine shorthands are loosing its purpose if it exceeds one line.
And Im not sure if its easier to read this way :man_shrugging:, is it?
A small thing, but I might know what to do next time with such case, wdyt @matks, @jolelievre ?

src/Adapter/Product/AbstractProductHandler.php
zuk3975: yup, will rebase it.

src/Core/Domain/Product/Exception/CannotUpdateProductException.php
zuk3975: > and why do we have "1" when we have multiples of 10 in the exception just above ?

because there is no convention, there are some places to have 10 another have 1, so idk, which is better approach. I agree I should stay with one of them.

src/Adapter/Product/CommandHandler/UpdateProductDescriptionHandler.php
jolelievre: I prefer it for short single lines as well, but as you said it's a matter of preference Sometimes I even change my preference regarding this 😅

You could also init it with `INVALID_DESCRIPTION` and only have one if statement, just to make the code a bit lighter

I honestly can't take a firm position on this ^^but I'm sure it's not enough to block the PR

tests/Integration/Behaviour/Features/Scenario/Product/update_description.feature
```diff
@@ -0,0 +1,40 @@
...
+      | name       | en-US:potato |
+      | is_virtual | false        |
+    And product "product1" localized "description" should be "en-US:"
+    And product "product1" localized "description_short" should be "en-US:"
+    When I update product "product1" descriptions with following information:
+      | description       | en-US:sweet potato        |
+      | description_short | en-US:Just a sweet potato |
+    Then product "product1" localized "description" should be "en-US:sweet potato"
+    Then product "product1" localized "description_short" should be "en-US:Just a sweet potato"
+
```
jolelievre, line 18: Could you please add a "Partial update" scenario, to validate that we can udpate only `description` and `short_description` isn't modified (and vice versa)

This is specifically to avoid regression on this partial feature (which can happen quickly with a quick modif without paying attention)

zuk3975: modified one of scenarios to cover this.

src/Core/Domain/Product/Command/UpdateProductDescriptionCommand.php
zuk3975: I've added this, but im not sure its worth. It is not exceptional behavior, every other domain works the same, so we should add it in docs instead, or go through each domain and comment in each updateCommand.

jolelievre (approved):
Thank you @zuk3975

matks (approved):

matthieu-rolland (approved):

PR #19431, opened 2020-05-27
WIP: Use OrderProductRemover service in BO order detail product deletion

### LINKED ISSUE #18432: Refactor DeleteProductFromOrderHandler
- [x] #18564 : Refactor two handlers : 
- src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
- src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php

because updateOrder are nearly the same.

- [ ] Refactor handler src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php for using src/Adapter/Order/Refund/OrderProductRemover.php as service.

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Use the OrderProductRemover service in DeleteProductFromOrderHandler
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18432
| How to test?  | In BO's order detail page, check that product deletion still works properly.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19431)


matthieu-rolland:
About this "refacto":  

The `DeleteProductFromOrderHandler` does a lot of things that are not done in the `OrderProductRemover` service, at the beginning I started by implementing its added behaviors in the service, but I realised that was probably outside of its scope, and since it is used elsewhere, I should not change its behavior, it would add a lot of things, that's too risky (and it breaks the single responsibility principle that services above all should respect) 

So in the end I think there is not much to remove from the `DeleteProductFromOrderHandler` class: it deals with cartRules, OrderInvoices, Order weight, Shipping costs, many things that are not taken into account by the `OrderProductRemover`.

@Progi1984 and @jolelievre , what do you think?

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
```diff
@@ -116,29 +127,6 @@ public function handle(DeleteProductFromOrderCommand $command)
...
-            false,
-            'down'
-        );
-        // Remove & Add CartRules applied to cart
-        CartRule::autoRemoveFromCart();
-        CartRule::autoAddToCart();
-
-        return $cart->update();
-    }
-
```
matthieu-rolland, line 141: I removed the `updateCart` method, which is the only thing it had in common with the service.

matthieu-rolland:
I put this PR in WIP, it doesn't pass tests

Progi1984:
The problem is that the naming indicates they do the same thing : 
* OrderProductRemover
* DeleteProductFromOrderHandler

Else DeleteProductFromOrderHandler may not do all things it should

matthieu-rolland:
this PR is in progress

matthieu-rolland:
I close this PR, a more accurate one has been done here:  #19981

PR #19437, opened 2020-05-28
Make product image in list overridable

### LINKED ISSUE #19029: Regression - Product thumbnails always uses cover and doesn’t allow customization anymore
#### Describe the bug

Introduced into https://github.com/PrestaShop/PrestaShop/commit/79753f45c60ee641763c476ef34d5d61d61ae8a6 in branch 1.7.7.x and develop

#### Expected behavior

We expect to be able to change the `id_image` on a product array after defining a specific id_product_attribute (covered by `ImageRetriever::getProductImages`).

For an unknown reason, this behaviour has been changed and static call to Product::getCover have been introduced.
`Product::getCover` method does not take any combination into consideration or does not allow any override of it using hooks, like it was possible using `actionGetProductProperties`* hooks.

#### Steps to Reproduce

Use the classic theme using branch 1.7.7 develop
BTW, this bug does not exists into classic-rocket project (ping @prestarocket)

#### Additional information

* PrestaShop version: 1.7.7 and more

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | A top watcher bug has been fixed in 177 that allowed to display the Product cover in the list instead of the combination images Though it didn't allow to customize this image any more This PR changes the way the cover is selected, so the default behaviour is still to use the cover but now it is possible to override the image used as cover thanks to the `actionGetProductPropertiesAfter` A documentation is being written to complete this PR
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19029
| How to test?  | On a product with combination add a new image and set it as cover (one that is not used by any combination so that it's easier to see the difference) In the different product list the cover is used, but when you are on the product page and select a combination the images are those for the selected combination (the first one displayed is still the cover though)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19437)

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -5028,9 +5028,11 @@ class ProductCore extends ObjectModel
 
         $row['ecotax_rate'] = (float) Tax::getProductEcotaxRate($context->cart->{Configuration::get('PS_TAX_ADDRESS_TYPE')});
 
-        $cover = static::getCover($row['id_product']);
-        if (isset($cover['id_image'])) {
-            $row['cover_image_id'] = $cover['id_image'];
+        if (!isset($row['cover_image_id'])) {
+            $cover = static::getCover($row['id_product']);
+            if (isset($cover['id_image'])) {
+                $row['cover_image_id'] = $cover['id_image'];
+            }
         }
 
         Hook::exec('actionGetProductPropertiesAfter', [
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -5028,6 +5028,11 @@ class ProductCore extends ObjectModel
 
         $row['ecotax_rate'] = (float) Tax::getProductEcotaxRate($context->cart->{Configuration::get('PS_TAX_ADDRESS_TYPE')});
 
+        $cover = static::getCover($row['id_product']);
+        if (isset($cover['id_image'])) {
+            $row['cover_image_id'] = $cover['id_image'];
+        }
+
         Hook::exec('actionGetProductPropertiesAfter', [
             'id_lang' => $id_lang,
             'product' => &$row,
--- a/classes/Search.php
+++ b/classes/Search.php
@@ -23,6 +23,12 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  * International Registered Trademark & Property of PrestaShop SA
  */
+
+/**
+ * @deprecated This const is deprecated you should use the configuration variable instead Configuration::get('PS_SEARCH_MAX_WORD_LENGTH')
+ */
+define('PS_SEARCH_MAX_WORD_LENGTH', 30);
+
 /* Copied from Drupal search module, except for \x{0}-\x{2f} that has been replaced by \x{0}-\x{2c}\x{2e}-\x{2f} in order to keep the char '-' */
 define(
     'PREG_CLASS_SEARCH_EXCLUDE',
--- a/src/Adapter/Image/ImageRetriever.php
+++ b/src/Adapter/Image/ImageRetriever.php
@@ -199,6 +199,7 @@ class ImageRetriever
             'medium' => $medium,
             'large' => $large,
             'legend' => isset($object->meta_title) ? $object->meta_title : $object->name,
+            'id_image' => $id_image,
         ];
     }
 
--- a/src/Adapter/Image/ImageRetriever.php
+++ b/src/Adapter/Image/ImageRetriever.php
@@ -54,9 +54,8 @@ class ImageRetriever
      *
      * @return array
      */
-    public function getProductImages(array $product, Language $language)
+    public function getAllProductImages(array $product, Language $language)
     {
-        $productAttributeId = $product['id_product_attribute'];
         $productInstance = new Product(
             $product['id_product'],
             false,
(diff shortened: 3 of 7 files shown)


jolelievre:
A doc is being written to explain how to override the cover, but here is a cod example:

```php
    public function hookActionGetProductPropertiesAfter($params) {
        $product = $params['product'];
        $productInstance = new Product($product['id_product']);
        $productAttributeId = $product['id_product_attribute'];
        $combinationImages = $productInstance->getCombinationImages($params['id_lang']);
        if (empty($combinationImages)) {
            return;
        }

        // Update cover image ID to use the first image of the product combination
        foreach ($combinationImages as $imgs) {
            foreach ($imgs as $img) {
                if ($productAttributeId == $img['id_product_attribute']) {
                    $params['product']['cover_image_id'] = $img['id_image'];
                    return;
                }
            }
        }
    }
```

Matt75 (dismissed):

PululuK:
>         // Update cover image ID to use the first image of the product combination
>         foreach ($combinationImages as $imgs) {
>             foreach ($imgs as $img) {
>                 if ($productAttributeId == $img['id_product_attribute']) {
>                     $params['product']['cover_image_id'] = $img['id_image'];
>                     return;
>                 }
>             }
>          }

Why double foreach then if you are just looking for the first element ? You can just use index 0 right ? 🤔

Something like :

```php
  $fist_element = $combinationImages[array_keys($combinationImages)[0]];
```

matthieu-rolland (dismissed):

Progi1984 (dismissed):

jolelievre:
Hi @PululuK you're right this can be simplified, not exactly the code you proposed because we still need to match the appropriate combination not just the first one

But I inspired from your suggestion to update the doc:
https://github.com/PrestaShop/docs/pull/550

kpodemski:
@jolelievre i'm wondering if having to initiate another instance of Product will impact performance?

jolelievre:
@kpodemski It might have a short drawback compared to before, although there are many places where product instance are created on the fly so I don't think it will be noticeable

An alternative I thought about was to get all the product images from the `ImageRetriever` by unset the product attribute This would allow to get the cover for the image list, and then the attribute filtering would be performed in `ProductLazyArray` instead of `ImageRetriever:: getProductImages`

The only thing is that then the filtering in `ImageRetriever:: getProductImages` would become useless since never used elsewhere (at least in the core)

jolelievre:
@kpodemski I performed a little refacto to optimize this modification a little

All the product images are fetched by `ProductLazyArray` which allows it to filter them for the combination AND to get the cover IF it is among the product images So an additional `Product` instance will only be created if the cover image is an image that has no relation with the product 

This was not possible until then so I think the additional memory usage in this edge case is acceptable

matks:
Unit tests broken
```
) LegacyTests\Unit\Core\Product\ProductPresenterTest::testPriceShouldNotBeShownInCatalogMode
TypeError: Argument 1 passed to PrestaShop\PrestaShop\Adapter\Presenter\Product\ProductLazyArray::filterImagesForCombination() must be of the type array, null given, called in /home/travis/build/PrestaShop/PrestaShop/src/Adapter/Presenter/Product/ProductLazyArray.php on line 595
```

Matt75 (dismissed):

PierreRambaud (dismissed):

Progi1984 (dismissed):

PululuK (dismissed):

prestamodule:
@jolelievre thank you for your work on this, however, as we've discussed on Slack, we aren't able to have the same "combination cover" when coming on the product page.
For example, if we have a link to a product with a specific combination, the main product image will always be the product cover, and not the combinations images.

If i keep my example of the red skirt, you'll have this image into the product list but not anymore on the product page (even if the combination is defined into the URL).

Regarding the changes on our side (for product list), we can keep the previous behaviour by defining `cover_image_id` before calling `Product::getProductProperties`, so that's fine for us.

jolelievre:
Hi @prestamodule 

Don't worry I didn't forget about the issue on product page, I just didn't have pushed it yet This should be all good now Don't hesitate to tell us if you still spot some problems

prestamodule:
Hi @jolelievre, looks fine now ;)
Theme contributors will have to follow your modifications into `product-cover-thumbnails.tpl` file in order to have the correct behaviour.

PierreRambaud (approved):

jolelievre:
@prestamodule 

Indeed you're right, I tried to keep templates modifications at the minimum, but this one was unfortunately required. There are other templates where `product.cover` is used, but since `getProductProperties` is not called, the cover ID is not set and the `default_image` is automatically used as a fallback (like cart detail, product modal, ...)

It's a bit "out of luck" that it works, so I think I'll provide another PR that changes these templates as well to `default_image` out of consistency This PR only focuses on the *required* modifications

I'm also wondering if we should have a similar behaviour as `cover_image_id` in `getProductProperties` to allow overriding it? 🤔

I also updated the doc PR https://github.com/PrestaShop/docs/pull/550 to explain this new `default_image` a little

PululuK (approved):

prestamodule:
> I'm also wondering if we should have a similar behaviour as `cover_image_id` in `getProductProperties` to allow overriding it? 🤔

Indeed, that is a good idea and not hard to do 👍

matthieu-rolland (approved):

matthieu-rolland:
@jolelievre I didn't assign the "waiting for QA", it seems like you have an improvement suggestion, will you do it in another PR or in this one ?

jolelievre:
@matthieu-rolland The improvement will be done in another PR since it's a bit out of this scope Let's focus on fixing the existing problem for now before improving this new solution

Progi1984:
Thanks @jolelievre

PR #19448, opened 2020-05-28
Fix AdminModulesController fatal error

### LINKED ISSUE #19447: AdminModules : Call to a member function isUsingNewTranslationSystem()
### /var/log/dev.log
[2020-05-28 23:02:58] php.CRITICAL: Call to a member function isUsingNewTranslationSystem() on boolean {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Call to a member function isUsingNewTranslationSystem() on boolean at /var/www/html/controllers/admin/AdminModulesController.php:1419)"} []
[2020-05-28 23:02:58] request.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\FatalThrowableError: "Call to a member function isUsingNewTranslationSystem() on boolean" at /var/www/html/controllers/admin/AdminModulesController.php line 1419 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Call to a member function isUsingNewTranslationSystem() on boolean at /var/www/html/controllers/admin/AdminModulesController.php:1419)"} []

### Fatal error

```
Call to a member function isUsingNewTranslationSystem() on boolean

Symfony\Component\Debug\Exception\
FatalThrowableError
in controllers/admin/AdminModulesController.php (line 1419)
AdminModulesControllerCore->initModal()
in classes/controller/AdminController.php (line 2914)
AdminControllerCore->init()
in classes/controller/Controller.php (line 273)
ControllerCore->run()
in classes/Dispatcher.php (line 515)
DispatcherCore->dispatch()
in admin-dev/index.php (line 93)
```

#### Expected behavior

PrestaShop should works and notify smoothly to Merchant the requested Module configuration page is not available.

#### How to test

1. Go to Module Manager
2. Click on uninstall ps_checkout (PrestaShop Checkout) if installed
3. Go to /admin-dev/index.php?controller=AdminModules&configure=ps_checkout&token= (with our token btw)

**Screenshots**

ServiceNotFoundException

#### Additional information

* PrestaShop version: 1.7.7.0
* PHP version: 7.2

Matt75:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | If requested Module is not found when trying to access module configuration page
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/19447
| How to test?  | Follow "How to test" described in related issue.

You should see
fix-AdminModulesController

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19448)

### DIFF
--- a/controllers/admin/AdminModulesController.php
+++ b/controllers/admin/AdminModulesController.php
@@ -1415,7 +1415,13 @@ class AdminModulesControllerCore extends AdminController
         $translateLinks = [];
 
         if (Tools::getIsset('configure')) {
+            /** @var Module|false $module */
             $module = Module::getInstanceByName(Tools::getValue('configure'));
+
+            if (false === $module) {
+                return;
+            }
+
             $isNewTranslateSystem = $module->isUsingNewTranslationSystem();
             $link = Context::getContext()->link;
             foreach ($languages as $lang) {


arouiadib (approved):

controllers/admin/AdminModulesController.php
```diff
@@ -1415,7 +1415,13 @@ public function initModal()
         $translateLinks = [];
 
         if (Tools::getIsset('configure')) {
+            /** @var Module|false $module */
             $module = Module::getInstanceByName(Tools::getValue('configure'));
+
+            if (false === $module) {
```
arouiadib, line 1421: maybe `if (!$module instanceof Module)`  is more generic  :thinking:  but not that important. Thanks for fixing this 500.

Progi1984 (approved):

matks (approved):

PierreRambaud (approved):

controllers/admin/AdminModulesController.php
PululuK, line 1421: @Matt75  Sorry but, this case needs to be handled more cleanly
This should throw an Exception if the module is not valid, not silently return.
I really don't understand why you ignored this comment https://github.com/PrestaShop/PrestaShop/pull/19448#discussion_r432150973 by @arouiadib 🤔 ?

PululuK (changes requested):

controllers/admin/AdminModulesController.php
matks: @PululuK I might side with @Matt75 on this one 😄 

I think we should fail silently here because modules not being found can happen in multiple real usecases. For example people who want to uninstall a module will sometimes `rm -rf` the folder without triggering the uninstall 😅 .

So even though in theory an exception would be better, I think it's safe here to choose code robustness against code integrity.

The idea is "even if some modules are not uninstalled properly, the merchant should be able to see his page because he needs it to manage other modules".

If we wanted to do it a lot cleaner, what we could do:
- move this logic outside the controller because parsing modules should not happen in a controller, for example in a class `ModuleParser.php`
- in `ModuleParser.php` we would raise meaningful exceptions for such usecases
- in controller, we decide how to deal with such exceptions with a `try/catch` and here we would collect those exceptions, display a warning message about it BUT still allow the page to render

matks: (also, @Matt75 provided a quick fix for this critical issue ; so we might merge the quick fix to "bring back" the page ; but also work on a better fix later)

PululuK: Precisely, as this is a critical case it will have to be managed more properly.
What will happen if $module is not false and is not a instance of Module class ?

PululuK: Critical bugs should be considered with serenity !

SimonGrn:
Is there a reason there is a "Critical" label on this PR ?

PierreRambaud:
> Is there a reason there is a "Critical" label on this PR ?

Because the first type was `Critical` prestonbot added it automatically. I change it for you :heart:

controllers/admin/AdminModulesController.php
PierreRambaud: It has been defined as critical by the author, not by the team :sweat_smile: 
The check about defined if it's an instance of Module class or not must be in `getInstanceByName`.
I agree with @matks here (and he knows it doesn't happen often :trollface: )

matks: SCREENSHOT

SimonGrn:
> > Is there a reason there is a "Critical" label on this PR ?
> 
> Because the first type was `Critical` prestonbot added it automatically. I change it for you

Are you sure ? Because I see PrestonBot in the history... damn bot :robot: !!!!!

controllers/admin/AdminModulesController.php
arouiadib: I am a bit late for this, but just want to note that @matt75 didn't ignore my comment, it is me that  approved PR before commenting (Maybe github considers such comments as stale). Thanks @PululuK, and all of you guys, for the usual informative PRs and discussions :+1:

matks: > Precisely, as this is a critical case it will have to be managed more properly.

> Critical bugs should be considered with serenity !

Wise words indeed 👍 with the benefit of hindsight I acted too fast. I was blinded by the CRITICAL label and I wanted to fix it quickly

Thanks for the warning @PululuK 😉

PululuK: Don't worry @matks  !

Progi1984:
Thanks @matks

matks:
> Thanks @matks

Matthias ... Mathieu ... Matthieu ... I guess we all look the same to you :trollface:

PR #19450, opened 2020-05-29
Add phpdoc for throws - Link

Matt75:
| Questions     | Answers
| Branch?       | develop
| Description?  | Add phpdoc throws tag to help developer to identify potential error cases
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShopCorp/ps_mbo/pull/96
| How to test?  | Everything should works as before

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19450)

### DIFF
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -86,8 +86,6 @@ class LinkCore
      * @param int $idPicture ID of the picture to delete
      *
      * @return string
-     *
-     * @throws PrestaShopException
      */
     public function getProductDeletePictureLink($product, $idPicture)
     {
@@ -259,8 +257,6 @@ class LinkCore
      * @param int|null $idCustomization
      *
      * @return string
-     *
-     * @throws PrestaShopException
      */
     public function getRemoveFromCartURL(
         $idProduct,
@@ -294,8 +290,6 @@ class LinkCore
      * @param int|null $idCustomization
      *
      * @return string
-     *
-     * @throws PrestaShopException
      */
     public function getUpQuantityCartURL(
         $idProduct,
@@ -313,8 +307,6 @@ class LinkCore
      * @param int|null $idCustomization
      *
      * @return string
-     *
-     * @throws PrestaShopException
      */
     public function getDownQuantityCartURL(
         $idProduct,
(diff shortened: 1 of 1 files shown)


Progi1984 (dismissed):

classes/Link.php
```diff
@@ -81,6 +86,8 @@ public function __construct($protocolLink = null, $protocolContent = null)
      * @param int $idPicture ID of the picture to delete
      *
      * @return string
+     *
+     * @throws PrestaShopException
```
PierreRambaud: Why adding this, this method doesn't throw PrestaShopException :thinking:

PierreRambaud (changes requested):
I'm against adding `@throws` when the method itself is not throwing exceptions, otherwise, we must change ALL methods to say it throws `Exception`, `DbExpection`, `PDOException`, etc. Which is illogic :/

classes/Link.php
Matt75: PHPStorm inspection ;)

Matt75: Due to usage of Configuration::get() can return PrestaShopException on call Db::executeS

Matt75: I will update Pull Request according your feedback

PierreRambaud: Phpstorm :vomiting_face:

Progi1984 (dismissed):

PierreRambaud (changes requested):
Still against this kind of change, sorry, methods are not throwing exceptions.
Structural elements are not properly defined, many exceptions are missing and if we follow phpdoc
> This tag is used to present in your documentation which error COULD occur and under which circumstances. It is RECOMMENDED to provide a description that describes the reason an exception is thrown.

Otherwise, this annotation has no aim

Matt75:
> Still against this kind of change, sorry, methods are not throwing exceptions.
> Structural elements are not properly defined, many exceptions are missing and if we follow phpdoc
> 
> > This tag is used to present in your documentation which error COULD occur and under which circumstances. It is RECOMMENDED to provide a description that describes the reason an exception is thrown.
> 
> Otherwise, this annotation has no aim

Pull request updated to only use @throws when the method itself throws an exception
Add warning on phpdoc for other cases about Symfony Router

matks (approved):

PierreRambaud (approved):

PierreRambaud:
Thanks @Matt75

PR #19453, opened 2020-05-29
Help on textarea (twig) show 2 times

clotairer:
| Questions     | Answers
| Branch?       | develop
| Description?  | {{ block('form_help') }} is already in {{- parent() -}} 
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes ...
| How to test?  | image<br>Create a Symfony Form Type and add a TextareaType::class on build form method for instance to manage a configuration parameter. Don't forget 

Sample code:
```php
    {
        $builder
            ->add('Myfield', TextareaType::class, [
                'label' => $this->trans('My label', 'Module.__Module__.Admin'),
                'help' => $this->trans('My help', 'Module.__Module__.Admin'),
            ]);
    }
```

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19453)

### DIFF
--- a/src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
@@ -39,7 +39,6 @@
     {% set attr = attr|merge({class: (attr.class|default('') ~ ' form-control')|trim}) %}
     {{- parent() -}}
 
-    {{ block('form_help') }}
 {%- endblock textarea_widget %}
 
 {% block button_widget -%}
--- a/src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
@@ -38,7 +38,6 @@
 {% block textarea_widget -%}
     {% set attr = attr|merge({class: (attr.class|default('') ~ ' form-control')|trim}) %}
     {{- parent() -}}
-
 {%- endblock textarea_widget %}
 
 {% block button_widget -%}


src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
```diff
@@ -39,7 +39,6 @@
     {% set attr = attr|merge({class: (attr.class|default('') ~ ' form-control')|trim}) %}
     {{- parent() -}}
```
Progi1984:
```suggestion
```

Progi1984 (commented):

Progi1984 (approved):

PierreRambaud (approved):

matthieu-rolland (approved):

SimonGrn:
Hello @clotaire202 ,

thank you for your contribution. In order for the QA team to review your PR, we need an issue linked or a clear description of the problem.
We also cannot do some advanced technical stuff such as creating forms or modules, so if we need a specific piece of code would you be so kind as to give us one to use ?
Thanks in advance !

clotairer:
@SimonGrn 

This PR is a part of EPIC #19537 based on a [sample module](https://github.com/202ecommerce/checkboform) with all From Type from PrestaShop Bundle or Symfony.
Is this EPIC issue with PR explanation OK, or do you need a specific ticket associated to this PR ?

SimonGrn:
Let me see if I understand correctly:
when you create a module that have a textarea in the BO and you use a `help` string, currently it shows twice because in the twig template there is a call to `{{ block('form_help') }}`. BUT the `{{- parent() -}}` tag already calls this `form_help` so your string appears twice ?
Your correction remove the `{{ block('form_help') }}` call to let the `{{- parent() -}}` call display the string.
Is that correct ?

clotairer:
>  Is that correct ?

Yep that's simply as that !

SimonGrn:
Thanks @clotaire202 !
Very nice module you have here, we'll definitely use it in the future !

PierreRambaud:
Thanks @clotaire202

PR #19506, opened 2020-06-01
Handle AbstractGridDefinitionFactory compatibility

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Because of https://github.com/PrestaShop/PrestaShop/pull/18228, if you use CommonController::searchGridAction with a AbstractGridDefinitionFactory it does not work.<br> This is because https://github.com/PrestaShop/PrestaShop/pull/18228 only handles 2 cases: AbstractFilterableGridDefinitionFactory and controller/action filter keys
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  |

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19506)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/CommonController.php
+++ b/src/PrestaShopBundle/Controller/Admin/CommonController.php
@@ -33,6 +33,7 @@ use PrestaShop\PrestaShop\Core\Addon\Module\ModuleManagerBuilder;
 use PrestaShop\PrestaShop\Core\Domain\Notification\Command\UpdateEmployeeNotificationLastElementCommand;
 use PrestaShop\PrestaShop\Core\Domain\Notification\Query\GetNotificationLastElements;
 use PrestaShop\PrestaShop\Core\Domain\Notification\QueryResult\NotificationsResults;
+use PrestaShop\PrestaShop\Core\Grid\Definition\Factory\AbstractGridDefinitionFactory;
 use PrestaShop\PrestaShop\Core\Grid\Definition\Factory\FilterableGridDefinitionFactoryInterface;
 use PrestaShop\PrestaShop\Core\Grid\Definition\Factory\GridDefinitionFactoryInterface;
 use PrestaShop\PrestaShop\Core\Kpi\Row\KpiRowInterface;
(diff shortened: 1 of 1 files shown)


matks:
## How I found this issue

I started using ` CommonController::searchGridAction` everywhere but all my modified listings would not be able to search because of I did not use AbstractFilterableGridDefinitionFactory instead of AbstractGridDefinitionFactory.

I will use AbstractFilterableGridDefinitionFactory as AbstractGridDefinitionFactory is deprecated however this small compatibility layer can be nice for module developers

src/PrestaShopBundle/Controller/Admin/CommonController.php
```diff
@@ -358,30 +359,42 @@ public function searchGridAction(
         $gridDefinitionFactoryServiceId,
         $redirectRoute,
         array $redirectQueryParamsToKeep = []
-    ) {
+    )
+    {
+        /** @var ResponseBuilder $responseBuilder */
+        $responseBuilder = $this->get('prestashop.bundle.grid.response_builder');
```
jolelievre: It would be better to check for the legacy case first by checking the factory is NOT `FilterableGridDefinitionFactoryInterface` and NOT `AbstractGridDefinitionFactory`
Then you can use `controller_response_builder` and return the response

The idea is just not to get the `prestashop.bundle.grid.response_builder` if it's not needed (therefore not instanciate it and save memory)

jolelievre (changes requested):

src/PrestaShopBundle/Controller/Admin/CommonController.php
matks: See https://github.com/PrestaShop/PrestaShop/pull/19506/commits/c47c04caf682a0b64f66d8ed73278ec4641b5185 it should be better now

jolelievre (approved):
Thank @matks

Progi1984 (approved):

PierreRambaud (approved):

matthieu-rolland (approved):

Progi1984:
@PrestaShop/qa-functional Does it need to be tested ?

matks:
@Progi1984 It's a refacto that cannot be seen by QA team 🤔

SimonGrn:
> @Progi1984 It's a refacto that cannot be seen by QA team

That means only exploratory tests in some pages that implements this grid, a minima. Could you list a few pages impacted by this change, please ?

SimonGrn:
> @Progi1984 It's a refacto that cannot be seen by QA team

If QA cannot test it, we need one or two Core devs taking the responsibility for this merge :stuck_out_tongue:

PierreRambaud (approved):

matks:
Close and reopen to reset Travis

matks:
Thank you for review !

PR #19529, opened 2020-06-02
Refacto AddProductCommand and introduce GetEditableProduct query

jolelievre:
| Questions     | Answers
| Branch?       | develop
| Description?  | The `AddProductCommand` command contract changes: it only needs a boolean `isVirtual` as input The product type interpreation has been moved into the `GetEditableProduct` query so the behat tests use this query
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | Technical PR, only green tests on Travis are enough

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19529)

### DIFF
--- a/src/Core/Domain/Product/QueryResult/EditableProduct.php
+++ b/src/Core/Domain/Product/QueryResult/EditableProduct.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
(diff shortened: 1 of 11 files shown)


src/Core/Domain/Product/QueryResult/EditableProduct.php
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public function getProductId(): int
+    {
+        return $this->productId;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getLocalizedNames(): array
```
jolelievre, line 70: This part is not used/tested yet, but it should be used in the behat tests when we check for localized fields (to be done in future PRs)

src/Core/Domain/Product/QueryResult/ProductType.php
```diff
@@ -67,35 +67,35 @@ class ProductType
...
     }
 
     /**
-     * @param int $value
+     * @param string $value
      *
      * @throws ProductConstraintException
      */
-    private function assertProductType(int $value): void
+    private function assertProductType(string $value): void
```
jolelievre, line 98: I believe this checking part and especially the exception throwing should be removed
In query result I don't think we throw exceptions because we are supposed to control the data provided so it shouldn't be needed (not like in ValueObject which contains input from user)

src/Core/Domain/Product/QueryResult/ProductType.php
```diff
@@ -38,23 +38,23 @@ class ProductType
...
      * Standard product
      */
-    const TYPE_STANDARD = 0;
+    const TYPE_STANDARD = 'standard';
 
     /**
      * A pack consists multiple units of product.
      */
-    const TYPE_PACK = 1;
+    const TYPE_PACK = 'pack';
```
zuk3975, line 46: I wonder if its ok with legacy types being int https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Product.php#L631

zuk3975:
looking good :+1:

matks (approved):

PierreRambaud (approved):

src/Core/Domain/Product/QueryResult/ProductType.php
sowbiba: I get the point. But you cannot avoid anyone to instantiate an object from anywhere. I guess this is a minimal security to at least check that expected types are used

sowbiba (approved):

Issue #14775, opened 2019-07-19
[EPIC] Product page - Prices tab 7/9


marionf:
### - As a merchant I want to be able to manage the prices of my products

[https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#pricing-tab-epic](https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#pricing-tab-epic) 

- [x] #19671
- [x] #19672
- [x] #19673

### [- As a merchant I want to know if there are errors when I complete a field in the product page](https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#pricing-tab-)

- [ ]  #18861

### Issues in PRICE TAB sorted by priority high to low :
**Regression on develop**

**Must-Have**
- [ ] #9855 - **`ecotax`**
- [ ] #15044 - **`ecotax`**
- [ ] #10030 - **`ecotax`**
- [x] #14093
- [ ] #12363

**New feature**
- [x] #11329

**wording**

**Should-Have**
- [x] #22552

**Nice to Have**
- [x] #15785
- [x] #12911

### Issues in PRICE TAB without priroty
- [ ] #18656 
- [x] #22552
- [x] #24860 
- [x] #25045
- [x] #28430
- [x] #29296
- [ ] https://github.com/PrestaShop/PrestaShop/issues/31830

## Fixed
- [x] #27051
- [x]  #20223
- [x] #12475
- [x] #9897
- [x] #18701
- [x] #14052
- [x] #12078
- [x] #10792
- [x] #10026 - **`ecotax`**
- [x] https://github.com/PrestaShop/PrestaShop/issues/18751#event-3336713500
- [x] #19674
- [x] #15038 - **`ecotax`**

marionf:
- Use a constant for decimal precision for ecotax computation #18861

### - Issues in **price tab** sorted by priority high to low :
**Regression on develop**

**Must-Have**
- [ ] #9855 ecotax
- [ ] #15044 ecotax
- [ ] #10030 ecotax
- [ ] #14093 
- [ ] #12363

**New feature**
- [ ] #11329

**wording**

**Should-Have**
- [ ] #22552 - Try to add a specific price to a product, then add another one, you will see that the selected user for the previous specific price is still selected

**Nice to Have**
- [ ] #15785
- [ ] #12911

## Fixed
[...]

PR #15611, opened 2019-09-19
[WIP] Migrate "International > Taxes > Tax Rules > Add new / edit tax rule group" page

RaimondasSapola:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Migrate "International > Taxes > Tax Rules > Add new / edit tax rule group" page and tax rule grid with its forms and creation and editing handling
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | https://github.com/PrestaShop/PrestaShop/issues/10610
| How to test?  | For tax rule group creation go to: `/admin-dev/index.php/improve/international/tax_rules_groups/new`. For editing: `/admin-dev/index.php/improve/international/tax_rules_groups/{taxRulesGroupId}/edit`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/15611)

PR #19542, opened 2020-06-03
UpdateProductPricesCommand

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Command for updating product prices information. price, price tax excluded, wholesale price, tax rule id, unit price etc.
| Type?         | refacto
| Category?     | BO
| BC breaks?    |  no
| Deprecations? | no
| Fixed ticket? | part of #14775
| How to test?  | Behat tests covers basic usecases. Nothing to test until its integrated with form.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19542)

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -30,6 +30,7 @@ define('_CUSTOMIZE_FILE_', 0);
 // Deprecated since 1.5.0.1 use Product::CUSTOMIZE_TEXTFIELD
 define('_CUSTOMIZE_TEXTFIELD_', 1);
 
+use PrestaShop\Decimal\Number;
 use PrestaShop\PrestaShop\Adapter\ServiceLocator;
 use PrestaShop\PrestaShop\Core\Product\ProductInterface;
 
@@ -642,6 +643,14 @@ class ProductCore extends ObjectModel
     public function __construct($id_product = null, $full = false, $id_lang = null, $id_shop = null, Context $context = null)
     {
         parent::__construct($id_product, $id_lang, $id_shop);
+
+        $unitPriceRatio = new Number((string) $this->unit_price_ratio);
+        $price = new Number((string) $this->price);
+
+        if ($unitPriceRatio->isGreaterThan(new Number('0'))) {
+            $this->unit_price = (float) (string) $price->dividedBy($unitPriceRatio);
+        }
+
         if ($full && $this->id) {
             if (!$context) {
                 $context = Context::getContext();
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -112,7 +112,7 @@ class ProductCore extends ObjectModel
     public $unity = null;
 
     /** @var float price for product's unity */
-    public $unit_price;
+    public $unit_price = 0;
 
     /** @var float price for product's unity ratio */
     public $unit_price_ratio = 0;
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -112,7 +112,7 @@ class ProductCore extends ObjectModel
     public $unity = null;
 
     /** @var float price for product's unity */
-    public $unit_price = 0;
+    public $unit_price;
 
     /** @var float price for product's unity ratio */
     public $unit_price_ratio = 0;
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -647,7 +647,7 @@ class ProductCore extends ObjectModel
         $unitPriceRatio = new Number((string) $this->unit_price_ratio);
         $price = new Number((string) $this->price);
 
-        if ($unitPriceRatio->isGreaterThan(new Number('0'))) {
+        if ($unitPriceRatio->isGreaterThanZero()) {
             $this->unit_price = (float) (string) $price->dividedBy($unitPriceRatio);
         }
 
--- a/src/Adapter/Product/AbstractProductHandler.php
+++ b/src/Adapter/Product/AbstractProductHandler.php
@@ -28,6 +28,7 @@ declare(strict_types=1);
 
 namespace PrestaShop\PrestaShop\Adapter\Product;
 
+use PrestaShop\Decimal\Number;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductNotFoundException;
@@ -61,6 +62,8 @@ abstract class AbstractProductHandler
                     $productIdValue
                 ));
             }
+
+            $this->setUnitPrice($product);
         } catch (PrestaShopException $e) {
             throw new ProductException(
                 sprintf('Error occurred when trying to get product #%s', $productId),
@@ -119,6 +122,19 @@ abstract class AbstractProductHandler
                 ),
                 $errorCode
             );
+        }
+    }
+
+    /**
+     * @param Product $product
+     */
+    private function setUnitPrice(Product $product)
+    {
+        $price = new Number((string) $product->price);
+        $unitPriceRatio = new Number((string) $product->unit_price_ratio);
+
+        if (!$unitPriceRatio->equals(new Number('0'))) {
+            $product->unit_price = (float) (string) $price->dividedBy($unitPriceRatio);
         }
     }
 }
--- a/src/Adapter/Product/AbstractProductHandler.php
+++ b/src/Adapter/Product/AbstractProductHandler.php
@@ -76,6 +76,37 @@ abstract class AbstractProductHandler
     }
 
     /**
+     * Provides product field as Number instead of float.
+     *
+     * @param Product $product
+     * @param string $field
+     *
+     * @return Number
+     */
+    protected function getPropertyAsNumber(Product $product, string $property): Number
+    {
+        $numericProperties = [
+            'price',
+            'ecotax',
+            'wholesale_price',
+            'unit_price',
+            'unit_price_ratio',
+        ];
+
+        if (!in_array($property, $numericProperties, true)) {
+            throw new ProductException(sprintf('Product property "%s" does\'t exist or is not numeric', $property));
+        }
+
+        // To make sure all values are safely converted to Number.
+        // Because casting null to string results in empty string which isn't valid to create Number and throws error.
+        if (null === $product->{$property}) {
+            $product->{$property} = 0;
+        }
+
+        return new Number((string) $product->{$property});
+    }
+
+    /**
      * @todo: product name is not required. Fix that? issue: #19441 for discussion
      *
      * Validates Product object model multilingual property using legacy validation
@@ -130,8 +161,8 @@ abstract class AbstractProductHandler
      */
     private function setUnitPrice(Product $product)
     {
-        $price = new Number((string) $product->price);
-        $unitPriceRatio = new Number((string) $product->unit_price_ratio);
+        $price = $this->getPropertyAsNumber($product, 'price');
+        $unitPriceRatio = $this->getPropertyAsNumber($product, 'unit_price_ratio');
 
         if (!$unitPriceRatio->equals(new Number('0'))) {
             $product->unit_price = (float) (string) $price->dividedBy($unitPriceRatio);
--- a/src/Adapter/Product/AbstractProductHandler.php
+++ b/src/Adapter/Product/AbstractProductHandler.php
@@ -41,6 +41,14 @@ use Product;
 abstract class AbstractProductHandler
 {
     /**
+     * @var array specific product fields which needs to be updated.
+     *
+     * This is necessary because product is not fully loaded from database by default
+     * So during partial update we don't want to accidentally reset some fields
+     */
+    protected $fieldsToUpdate = [];
+
+    /**
      * @param ProductId $productId
      *
      * @return Product
--- a/src/Adapter/Product/AbstractProductHandler.php
+++ b/src/Adapter/Product/AbstractProductHandler.php
@@ -62,8 +62,6 @@ abstract class AbstractProductHandler
                     $productIdValue
                 ));
             }
-
-            $this->setUnitPrice($product);
         } catch (PrestaShopException $e) {
             throw new ProductException(
                 sprintf('Error occurred when trying to get product #%s', $productId),
@@ -153,19 +151,6 @@ abstract class AbstractProductHandler
                 ),
                 $errorCode
             );
-        }
-    }
-
-    /**
-     * @param Product $product
-     */
-    private function setUnitPrice(Product $product)
-    {
-        $price = $this->getPropertyAsNumber($product, 'price');
-        $unitPriceRatio = $this->getPropertyAsNumber($product, 'unit_price_ratio');
-
-        if (!$unitPriceRatio->equals(new Number('0'))) {
-            $product->unit_price = (float) (string) $price->dividedBy($unitPriceRatio);
         }
     }
 }
--- a/src/Adapter/Product/AbstractProductHandler.php
+++ b/src/Adapter/Product/AbstractProductHandler.php
@@ -28,7 +28,6 @@ declare(strict_types=1);
 
 namespace PrestaShop\PrestaShop\Adapter\Product;
 
-use PrestaShop\Decimal\Number;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductNotFoundException;
@@ -71,37 +70,6 @@ abstract class AbstractProductHandler
         }
 
         return $product;
-    }
-
-    /**
-     * Provides product field as Number instead of float.
-     *
-     * @param Product $product
-     * @param string $field
-     *
-     * @return Number
-     */
-    protected function getPropertyAsNumber(Product $product, string $property): Number
-    {
-        $numericProperties = [
-            'price',
-            'ecotax',
-            'wholesale_price',
-            'unit_price',
-            'unit_price_ratio',
-        ];
-
-        if (!in_array($property, $numericProperties, true)) {
-            throw new ProductException(sprintf('Product property "%s" does\'t exist or is not numeric', $property));
-        }
-
-        // To make sure all values are safely converted to Number.
-        // Because casting null to string results in empty string which isn't valid to create Number and throws error.
-        if (null === $product->{$property}) {
-            $product->{$property} = 0;
-        }
-
-        return new Number((string) $product->{$property});
     }
 
     /**
--- a/src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
+++ b/src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
@@ -52,6 +52,11 @@ final class UpdateProductPricesHandler extends AbstractProductHandler implements
     private $numberExtractor;
 
     /**
+     * @var Product the product being updated
+     */
+    private $product;
+
+    /**
      * @param NumberExtractor $numberExtractor
      */
     public function __construct(
@@ -65,7 +70,11 @@ final class UpdateProductPricesHandler extends AbstractProductHandler implements
      */
     public function handle(UpdateProductPricesCommand $command): void
     {
+        //@todo: update the Number lib dependency. It should have methods to compare to 0 already
+        //  Check whole class to refactor occurrences of `new Number('0')`
+
         $product = $this->getProduct($command->getProductId());
+        $this->product = $product;
         $this->fillUpdatableFieldsWithCommandData($product, $command);
         $product->setFieldsToUpdate($this->fieldsToUpdate);
 
@@ -84,18 +93,13 @@ final class UpdateProductPricesHandler extends AbstractProductHandler implements
      */
     private function fillUpdatableFieldsWithCommandData(Product $product, UpdateProductPricesCommand $command): void
     {
-        $price = $command->getPrice();
-        $unitPrice = $command->getUnitPrice();
-
-        if (null !== $price) {
-            $product->price = (float) (string) $price;
+        if (null !== $command->getPrice()) {
+            $product->price = (float) (string) $command->getPrice();
             $this->validateField($product, 'price', ProductConstraintException::INVALID_PRICE);
             $this->fieldsToUpdate['price'] = true;
         }
 
-        if (null !== $unitPrice) {
-            $this->setUnitPriceInfo($product, $unitPrice, $price);
-        }
+        $this->handleUnitPriceInfo($command);
 
         if (null !== $command->getUnity()) {
             $product->unity = $command->getUnity();
@@ -127,6 +131,40 @@ final class UpdateProductPricesHandler extends AbstractProductHandler implements
             $this->validateField($product, 'wholesale_price', ProductConstraintException::INVALID_WHOLESALE_PRICE);
             $this->fieldsToUpdate['wholesale_price'] = true;
         }
+    }
+
+    /**
+     * Update unit_price and unit_price ration depending on price
+     *
+     * @param UpdateProductPricesCommand $command
+     *
+     * @throws ProductConstraintException
+     */
+    private function handleUnitPriceInfo(UpdateProductPricesCommand $command): void
+    {
+        $price = $command->getPrice();
+        $unitPrice = $command->getUnitPrice();
+        $zero = new Number('0');
+
+        // if price was reset then also reset unit_price and unit_price_ratio
+        $priceWasReset = null !== $price && $price->equals($zero);
+        if ($priceWasReset) {
+            $this->resetUnitPriceInfo();
+        }
+
+        //if price was not reset then allow setting new unit_price and unit_price_ratio
+        if (!$priceWasReset && null !== $unitPrice) {
+            $this->setUnitPriceInfo($this->product, $unitPrice, $price);
+        }
+    }
+
+    /**
+     * @throws ProductConstraintException
+     */
+    private function resetUnitPriceInfo(): void
+    {
+        $zero = new Number('0');
+        $this->setUnitPriceInfo($this->product, $zero, $zero);
     }
 
     /**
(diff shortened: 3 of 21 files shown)


zuk3975:
`Bug` label should be removed. (Added wrong type in description at first, sorry)

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /**
+     * @param Product $product
+     * @param string $fieldName
+     *
+     * @throws ProductConstraintException
+     * @throws \PrestaShopException
+     */
+    private function validateField(Product $product, string $fieldName): void
```
jolelievre: I believe you used this in other handlers as well (description if I'm right) maybe this should be in the AbstractProductHandler, or even in a dedicated service

zuk3975:
blocked by https://github.com/PrestaShop/PrestaShop/pull/19577

src/Adapter/Product/AbstractProductHandler.php
```diff
@@ -61,6 +62,8 @@ protected function getProduct(ProductId $productId): Product
                     $productIdValue
                 ));
             }
+
+            $this->setUnitPrice($product);
```
zuk3975: I need to set unit_price manually to be able to use Number for calculations. And the unit_price is not calculated when simply creating Product object (it requires passing additional argument to load $full object and possibly something more)

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,221 @@
...
+            $product->ecotax = (float) (string) $command->getEcotax();
+            $this->validateField($product, 'ecotax', ProductConstraintException::INVALID_ECOTAX);
+        }
+
+        $taxRulesGroupId = $command->getTaxRulesGroupId();
+
+        if (null !== $taxRulesGroupId) {
+            $product->id_tax_rules_group = $taxRulesGroupId;
+            $this->validateField($product, 'id_tax_rules_group', ProductConstraintException::INVALID_TAX_RULES_GROUP_ID);
+            $this->assertTaxRulesGroupExists($taxRulesGroupId);
```
zuk3975, line 117: legacy didn't care if tax rules group exists, but i do :sweat_smile:

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,221 @@
...
+     */
+    private function setUnitPriceInfo(Product $product, Number $unitPrice, ?Number $price): void
+    {
+        $this->validateUnitPrice($unitPrice);
+
+        if ($unitPrice->equals(new Number('0'))) {
+            return;
+        }
+
+        if (null === $price) {
```
zuk3975, line 212: if its a partial update and price is not provided, we fallback to Product price from db. If its also 0 we throw exception, because in this case the unit_price and ratio cannot be set to any number except 0

src/Core/Domain/Product/QueryResult/ProductForEditing.php
```diff
@@ -51,22 +51,37 @@ class ProductForEditing
...
+     * @param int $taxRulesGroupId
      */
     public function __construct(
         int $productId,
         bool $active,
         ProductBasicInformation $basicInformation,
-        ProductCategoriesInformation $categoriesInformation
+        ProductCategoriesInformation $categoriesInformation,
+        ProductPricesInformation $pricesInformation,
+        int $taxRulesGroupId
```
zuk3975: this could be improved later depending on needs. For example we can have a DTO if we need a name or something more from taxRulesGroup.

tests/Integration/Behaviour/Features/Context/Domain/TaxRulesGroupFeatureContext.php
```diff
@@ -0,0 +1,62 @@
...
+use TaxRulesGroup;
+
+class TaxRulesGroupFeatureContext extends AbstractDomainFeatureContext
+{
+    /**
+     * @param string $name
+     *
+     * @return int
+     */
+    public static function getTaxRulesGroupByName(string $name): TaxRulesGroup
```
zuk3975, line 41: not sure about this.

tests/Integration/Behaviour/Features/Context/Domain/TaxRulesGroupFeatureContext.php
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Tests\Integration\Behaviour\Features\Context\Domain;
+
+use RuntimeException;
+use TaxRulesGroup;
+
+class TaxRulesGroupFeatureContext extends AbstractDomainFeatureContext
```
zuk3975, line 34: Not sure about this class. There is a TaxFeatureContext already, but it did not seem to fit and TaxRulesGroup is a different context IMO

tests/Integration/Behaviour/Features/Context/Domain/TaxRulesGroupFeatureContext.php
zuk3975, line 41: is it ok to have static method in featureContext ?

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,221 @@
...
+     *
+     * @throws ProductConstraintException
+     */
+    private function fillProductWithCommandData(Product $product, UpdateProductPricesCommand $command): void
+    {
+        $price = $command->getPrice();
+        $unitPrice = $command->getUnitPrice();
+
+        if (null !== $price) {
+            $product->price = (float) (string) $price;
```
matks: 💪 :trollface:

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
matks: 👍

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,221 @@
...
+    /**
+     * @param int $taxRulesGroupId
+     *
+     * @throws ProductConstraintException
+     * @throws ProductException
+     */
+    private function assertTaxRulesGroupExists(int $taxRulesGroupId): void
+    {
+        // 0 is valid value meaning no tax rules group is applied.
+        if (0 === $taxRulesGroupId) {
```
matks: we might need a constant here

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
matks: can you add this information into the code as comment ?

src/Adapter/Product/QueryHandler/GetProductForEditingHandler.php
```diff
@@ -85,6 +89,41 @@ private function getCategoriesInformation(Product $product): ProductCategoriesIn
...
+        $priceNumberFields = [
+            'price',
+            'ecotax',
+            'wholesale_price',
+            'unit_price',
+            'unit_price_ratio',
+        ];
+
+        // To make sure all values are safely converted to Number.
+        // Because casting null to string results in empty string which isn't valid to create Number and throws error.
```
matks: So this is to ensure that, when data is fetched from database, it is valid when put in Numbers ?

So we'll need this logic in multiple places

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -165,14 +170,179 @@ public function assertProductFields(string $productReference, TableNode $table)
...
+     * @When I update product :productReference prices and apply non-existing tax rules group
+     *
+     * @param string $productReference
+     */
+    public function updateTaxRulesGroupWithNonExistingGroup(string $productReference): void
+    {
+        $productId = $this->getSharedStorage()->get($productReference);
+
+        $command = new UpdateProductPricesCommand($productId);
+        $command->setTaxRulesGroupId(50000000);
```
matks, line 238:
```suggestion
        $command->setTaxRulesGroupId(50000000); // this ID does not exist, we use it on purpose
```

matks: so people after you will not wonder "why did he do that ?"

tests/Integration/Behaviour/Features/Context/Domain/TaxRulesGroupFeatureContext.php
matks: indeed 2 different contexts

tests/Integration/Behaviour/Features/Context/Domain/TaxRulesGroupFeatureContext.php
matks: yes, static methods are not evil

if there is not `$this` inside it and no need to have 2 instances of it, it's OK

matks (dismissed):

src/Adapter/Product/AbstractProductHandler.php
jolelievre: Should we change this behaviour in the `Product` class itself? It can be done every time this won't drop performance (we just need to move the line outside of the "full" block

jolelievre: An alternative would be to use `full = true` in the constructor.. As we will perform many change on Product some other handlers will probably need the full data And for back office actions it's not really a problem to get extra data

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
jolelievre: Free party is over!!!

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,221 @@
...
+            return;
+        }
+
+        $code = ProductConstraintException::INVALID_TAX_RULES_GROUP_ID;
```
jolelievre: This variable seems useless

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,221 @@
...
+     * @param Number $unitPrice
+     * @param Number $price
+     *
+     * @throws ProductConstraintException
+     */
+    private function setUnitPriceInfo(Product $product, Number $unitPrice, ?Number $price): void
+    {
+        $this->validateUnitPrice($unitPrice);
+
+        if ($unitPrice->equals(new Number('0'))) {
```
jolelievre: I'm not sure you need to return directly, if a previous unit price was set, and then you want to set it back to zero The modification will be ignored no? I believe you should handle a special case when value is zero and set `unit_price_ratio` and `unit_price` to zero?

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
jolelievre: Then is it considered an error or should we indeed set everything to zero? Or maybe use unit price as price?

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,221 @@
...
+    /**
+     * Unit price validation is not involved in legacy validation, so it is checked manually to have unsigned int value
+     *
+     * @param Number $unitPrice
+     *
+     * @throws ProductConstraintException
+     */
+    private function validateUnitPrice(Number $unitPrice): void
+    {
+        if ($unitPrice->isLowerThan(new Number('0'))) {
```
jolelievre: Damn but Pablo had added some extra methods to deal with this comparison I believe we need to update the lib dependency in 177, this will allow us to simplify all these calls (not in this PR of course)

src/Adapter/Product/QueryHandler/GetProductForEditingHandler.php
jolelievre: Maybe we can add a method `getProductFieldAsNumber(Product $product, string $fieldName)` in the abstract handler? This would simplify the code, there won't be any more need to update the object fields and we can use the same 0 Number instance: triple win!!!

src/Adapter/Product/QueryHandler/GetProductForEditingHandler.php
```diff
@@ -85,6 +89,41 @@ private function getCategoriesInformation(Product $product): ProductCategoriesIn
...
+        foreach ($priceNumberFields as $field) {
+            if (null === $product->{$field}) {
+                $product->{$field} = 0;
+            }
+        }
+
+        return new ProductPricesInformation(
+            new Number((string) $product->price),
+            new Number((string) $product->ecotax),
+            (int) $product->id_tax_rules_group,
```
jolelievre, line 115: So `id_tax_rules_group` is present in `ProductPricesInformation` and in `ProductForEditing` ?

src/Core/Domain/Product/QueryResult/ProductForEditing.php
```diff
@@ -51,22 +51,37 @@ class ProductForEditing
      */
     private $categoriesInformation;
 
+    /**
+     * @var ProductPricesInformation
+     */
+    private $pricesInformation;
```
jolelievre, line 57:
```suggestion
    private $pricesInformation;

```

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -165,14 +170,179 @@ public function assertProductFields(string $productReference, TableNode $table)
...
+
+    /**
+     * @param array $data
+     * @param ProductPricesInformation $pricesInfo
+     */
+    private function assertPriceFields(array $data, ProductPricesInformation $pricesInfo): void
+    {
+        if (isset($data['on_sale'])) {
+            $expectedOnSale = PrimitiveUtils::castStringBooleanIntoBoolean($data['on_sale']);
+            $onSaleInWords = $expectedOnSale ? 'to be on sale' : 'to be not on sale';
```
```suggestion
            $onSaleInWords = $expectedOnSale ? 'to be on sale' : 'not to be on sale';
```

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
jolelievre: this is an outdated comment from an old review, ignore it

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
zuk3975: Not sure. It doesn't look right if we allow resetting price when setting unit_price to 0 (it could be vice versa tho). So I think error is fine, everything else we can do in form validation/js side.

src/Adapter/Product/QueryHandler/GetProductForEditingHandler.php
zuk3975: Oh right, thats my mistake :D ill remove the additional one

src/Core/Domain/Product/ProductTaxRulesGroupSettings.php
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Some static settings of tax rules group inside Product context
+ */
+class ProductTaxRulesGroupSettings
+{
+    /**
+     * Value of tax rules group which reflects that product has no tax rules group applied
+     * Null value doesn't fit because it is used to identify partial updates
+     */
+    const NONE_APPLIED = 0;
```
zuk3975, line 40: not sure about this one. 
there was similar discussion here https://github.com/PrestaShop/PrestaShop/pull/15611/files#r327987588

matks: it works for me

matks (dismissed):

matks:
Reset travis

classes/Product.php
```diff
@@ -642,6 +643,14 @@ class ProductCore extends ObjectModel
     public function __construct($id_product = null, $full = false, $id_lang = null, $id_shop = null, Context $context = null)
     {
         parent::__construct($id_product, $id_lang, $id_shop);
+
+        $unitPriceRatio = new Number((string) $this->unit_price_ratio);
+        $price = new Number((string) $this->price);
+
+        if ($unitPriceRatio->isGreaterThan(new Number('0'))) {
```
jolelievre: We really need to udpate the decimal library 😅  The last version includes method to perform this kind of check I'll propose a PR for 177, and then merge 177 into develop

src/Adapter/Product/AbstractProductHandler.php
```diff
@@ -72,6 +73,37 @@ protected function getProduct(ProductId $productId): Product
...
+     * Provides product field as Number instead of float.
+     *
+     * @param Product $product
+     * @param string $field
+     *
+     * @return Number
+     */
+    protected function getPropertyAsNumber(Product $product, string $property): Number
+    {
+        $numericProperties = [
```
jolelievre: You should use `NumberExtractor` here, as it was designed for this (the PR has been merged already)

jolelievre: Or, as you seem to be doing, if you inject the extractor in handlers and don't use this method any more please remove it
An alternative to check would be to inject and use `NumberExtractor` in this abstract class, but it's only valid if ALL the handlers that extend this abstract class use this method

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,259 @@
...
+ */
+final class UpdateProductPricesHandler extends AbstractProductHandler implements UpdateProductPricesHandlerInterface
+{
+    /**
+     * @var array specific product fields which needs to be updated.
+     *
+     * This is necessary because product is not fully loaded from database by default
+     * So during partial update we don't want to accidentally reset some fields
+     */
+    private $fieldsToUpdate = [];
```
jolelievre: Maybe since this fields to update behaviour will likely be common to all product handlers it should be handled by the abstract class?

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,259 @@
...
+        if ($unitPrice->equals(new Number('0'))) {
+            return;
+        }
+
+        if (null === $price) {
+            $price = $this->numberExtractor->extract($product, 'price');
+        }
+
+        //@todo: update the Number lib dependency. It should have methods to compare to 0 already.
+        if ($price->equals(new Number('0'))) {
```
jolelievre: Let's agree on this rule "No unit price if price is 0" which makes sense This needs to be commented in this method and we need some behat tests to validate this behaviour and avoid regression

But as I said in https://github.com/PrestaShop/PrestaShop/pull/19542/files#r436901928 there is still the case where you wan to reset unit price to zero this should be allowed not ignored (here you simply return so any previous value will be kept and never changed)

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
jolelievre: https://github.com/PrestaShop/PrestaShop/pull/19748 new library is coming soon 😉

src/Core/Util/Number/NumberExtractor.php
```diff
@@ -98,6 +114,35 @@ public function extract($resource, string $propertyPath): Number
...
+    {
+        if (!is_object($resource)) {
+            return null;
+        }
+
+        if (!property_exists($resource, $property)) {
+            return null;
+        }
+
+        $reflectedObj = new ReflectionClass($resource);
```
jolelievre, line 134: Please a local cache system here to avoid creating too many `ReflectionClass`
Just a class field with array mapping via the classname (using `get_class`)

src/Core/Util/Number/NumberExtractor.php
```diff
@@ -64,13 +65,27 @@ public function __construct(PropertyAccessorInterface $propertyAccessor)
...
+     *                                       else it will search for getters.
+     *                                       Note: this will only work when providing exact property name,
+     *                                       but not path selector for inner objects
      *
      * @return Number
+     *
+     * @throws NumberExtractorException
      */
-    public function extract($resource, string $propertyPath): Number
+    public function extract($resource, string $propertyPath, bool $prioritizePublicProperty = true): Number
```
jolelievre: Not sure about this additional field I would say either it's a configuration for the service (from the constructor or a setter)
Or we provide an additional method `extractFromPublicProperty` (that would be called in `extract` by default)

tests/Integration/Behaviour/Features/Scenario/Product/update_prices.feature
```diff
@@ -0,0 +1,141 @@
...
+  As a BO user I want to be able to update product fields associated with price.
+
+  Scenario: I update product prices
+    Given I add product "product1" with following information:
+      | name       | en-US:magic staff   |
+      | is_virtual | false               |
+    And product "product1" should have following values:
+      | price              | 0           |
+      | ecotax             | 0           |
+      | tax rules group    |             |
```
jolelievre, line 14: Not `0` ?

classes/Product.php
```diff
@@ -642,6 +643,14 @@ class ProductCore extends ObjectModel
...
+            $this->unit_price = (float) (string) $price->dividedBy($unitPriceRatio);
```
jolelievre, line 651: Since you perform the computation here shouldn't it be removed in the "full block" ? (L676)

tests/Integration/Behaviour/Features/Scenario/Product/update_prices.feature
```diff
@@ -0,0 +1,141 @@
...
+      Scenario: I update product unit price without when product price is 0
+        Given I add product "product3" with following information:
+          | name       | en-US: black hat  |
+          | is_virtual | false             |
+        And product "product3" has following values:
+          | price           | 0            |
+          | unit_price      | 0            |
+        When I update product "product3" prices with following information:
+          | unit_price      | 300          |
+        Then I should get error that product "unit price" is invalid
```
jolelievre: It's actually weird to have the same exception for this use case as when you try to set a negative unit price

Maybe this kind of error would deserve a dedicated error code, especially because it's weird to have an invalid error code matching `unit_price` whereas it's also related to `price` field (changing the input for unit_price only will never be enough)

tests/Integration/Behaviour/Features/Scenario/Product/update_prices.feature
```diff
@@ -0,0 +1,141 @@
...
+          | unit_price       | 0            |
+        When I update product "product4" prices with following information:
+          | price            | 20           |
+          | unit_price       | 500          |
+        Then product "product4" should have following values:
+          | price            | 20           |
+          | unit_price       | 500          |
+          | unit_price_ratio | 0.04         |
+        When I update product "product4" prices with following information:
+          | price            | 0            |
```
jolelievre, line 129: So the 0 value is purely ignored.. this seems weird not to be able to set 0 if you want to

zuk3975: indeed it seems wrong here.
I think price should be reset to 0 in this case along with the unit_price and_unit_price ratio :thinking: 
Or an exception should be thrown.
maybe its possible to create VO for those 3 properties, that should be validated together?

classes/Product.php
zuk3975: in that block its calculated using Product::getStaticPrice, It is different price, I'll better allow overriding it to avoid bc breaks :smile: 
(it is the price from Product::getPriceStatic).

src/Adapter/Product/AbstractProductHandler.php
zuk3975: not used anymore, just forgot to delete :+1:

src/Core/Util/Number/NumberExtractor.php
zuk3975: removed it entirely.

tests/Integration/Behaviour/Features/Scenario/Product/update_prices.feature
zuk3975: its not id anymore, so its empty/none/not assigned. ( can write smth like that if its better?)
`if ($data['tax rules group'] === 'none') { assert as 0 }`

But it was same situation for some localized properties, we decided to use empty string there instead.

tests/Integration/Behaviour/Features/Scenario/Product/update_prices.feature
zuk3975: removed this exception at all for simplicity. Now if the price is 0 then unit_price is always reset to 0 too.

tests/Integration/Behaviour/Features/Scenario/Product/update_prices.feature
zuk3975: edited - the price has priority now, if its 0 then it resets unit_price and ratio to 0 aswell.

classes/Product.php
jolelievre: But maybe we should use `Product::getStaticPrice` here as well then?

zuk3975: I only want to check the plain value from db. I dont see any other method to do that.
The Product::getPriceStatic is calculating the price which depends from customer, cart, address etc..

classes/Product.php
jolelievre:
```suggestion
        if ($unitPriceRatio->isGreaterThanZero()) {
```

classes/Product.php
jolelievre: Yeah you're right let's stick with the database value

src/Adapter/Product/CommandHandler/UpdateProductPricesHandler.php
```diff
@@ -0,0 +1,279 @@
...
+        $product = $this->getProduct($command->getProductId());
+        $this->product = $product;
+        $this->fillUpdatableFieldsWithCommandData($product, $command);
+        $product->setFieldsToUpdate($this->fieldsToUpdate);
+
+        if (empty($this->fieldsToUpdate)) {
+            return;
+        }
+
+        $this->performUpdate($product);
```
jolelievre, line 82: Since `$this->fieldsToUpdate` is now in the abstract handler, and we will ned the same principle in other handlers, maybe all this could be moved into `performUpdate` and the method can be put in the abstract handler as well

And to finalize the feature add a `markUpdatedField` to avoid modifying the array directly and have a more semantic function

Of course it's just a suggestion, not blocking for this PR and can be done in another PR, along with the harmonization on other handlers

jolelievre (approved):
Thank you @zuk3975

matks (approved):

PR #19577, opened 2020-06-04
Refactor EditableProduct to ProductForEditing

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Renamed EditableProduct and its query/handlers to ProductForEditing (for  consistency). Also added some more structure to this DTO similarly as with OrderForViewing. keep in mind that It will evolve during the product migration. Also adjusted some behats to use ProductForEditing instead of legacy object model for assertions. Finally, removed assertions for product condition, because it doesn't give any value now and it should be done when time comes for Options tab/commands.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Not testable. CI must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19577)

### DIFF
--- a/tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
@@ -309,18 +309,6 @@ class ProductFeatureContext extends AbstractDomainFeatureContext
     /**
      * @param string $reference
      *
-     * @return Product
-     */
-    private function getProductByReference(string $reference): Product
-    {
-        $productId = $this->getSharedStorage()->get($reference);
-
-        return $this->getProductById($productId);
-    }
-
-    /**
-     * @param string $reference
-     *
      * @return ProductForEditing
      */
     private function getProductForEditing(string $reference): ProductForEditing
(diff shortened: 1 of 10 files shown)


tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -274,6 +267,25 @@ public function assertLastErrorIsInvalidShortDescriptionConstraint()
...
+     */
+    private function extractValueFromProductForEditing(ProductForEditing $productForEditing, string $propertyName)
+    {
+        $pathsByNames = [
+            'name' => 'basicInformation.localizedNames',
+            'description' => 'basicInformation.localizedDescriptions',
+            'description_short' => 'basicInformation.localizedShortDescriptions',
+        ];
+
+        $propertyAccessor = PropertyAccess::createPropertyAccessor();
```
matks, line 285: looks like it works 🎉

matks (dismissed):

matks (dismissed):

zuk3975:
rr travis

matks (approved):

tests/Integration/Behaviour/Features/Scenario/Product/add_product.feature
```diff
@@ -11,7 +11,6 @@ Feature: Add basic product from Back Office (BO)
       | is_virtual | false                |
     Then product "product1" should have following values:
       | active           | false          |
-      | condition        | new            |
```
jolelievre, line 14: We will have to reintroduce this test, but let's merge this for now 😉

jolelievre (approved):

PR #19581, opened 2020-06-04
Fix bugs occuring when an order state is deleted but still assigned to orders

### LINKED ISSUE #19098: Undefined index Notice error in debug mode when deleting a custom order status
#### Describe the bug

This is issue #16657, which has already been raised, and closed, though I think there is an issue here.
The problem comes if an order status has been removed (or as in the original bug report the mapping between order status has not been completed correctly).

I think you can also recreate this by creating an order using paypal for example, and then uninstalling the paypal module.  The order remains, but the status does not, this then causes the error.

As explained in the original bug report, in these circumstances, $val['id_order_state'] below is not guaranteed to have a value, can be null, and then this code generates the Notice error in debug mode, and it's not possible to load the page even to cancel the order.

`
foreach ($res as $key => $val) {
    $res[$key]['order_state'] = 
    $indexedOrderStates[$val['id_order_state']]['name'];            
    $res[$key]['invoice'] = $indexedOrderStates[$val['id_order_state']]['invoice'];
     $res[$key]['order_state_color'] = $indexedOrderStates[$val['id_order_state']]['color'];
}
`
#### Expected behavior

Customer page should open without error.

#### Steps to Reproduce

1. Enable debug mode
2. Create a new order status
3. Raise a new order, and assign the order the status you created.
4. Delete the order status
5. Try to load the customer screen - Error
6. Assign the order a different order status
7. Load the customer screen again - Now it loads.

**Screenshots**
image

image

image

#### Additional information

* PrestaShop version: 1.7.6.5
* PHP version: 7.2

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Some pages would throw exceptions when a deleted order state was still assigned to an order. This pr transforms the order state hard delete into a soft delete, also it keeps order state selectors from displaying deleted states, except if the current state is deleted. 
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19098
| How to test?  | **1-** In FO, create an order<br>**2-** In BO, create a new order state<br>**3-** In BO, assign this state to the order previously created<br>**4-** In BO, delete the order state previously created<br>**5-** In BO, go to the order list page => the order state of the last created order should still be displayed (with the mention "(deleted)".<br>**6-** In BO, go the detail order page of this order, the status should be selected with the mention "(selected)"<br>**7-** In BO, still in the order detail page, if you click on customer detail, the page should display (no exception)<br>**8-** In FO, in the order history page, the status should display correctly (without any mention)<br>**9-** In BO, both in order list page and in order detail page, if you select another state, the old state should disappear from the selector. <br>**10-** All deleted states should not appear in selectors, except if they are currently selected

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19581)

### DIFF
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -888,6 +888,29 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
     }
 
     /**
+     * Does a soft delete on current object, using the "deleted" field in DB
+     * If the model object has no "deleted" property or no "deleted" definition field it will throw an exception
+     *
+     * @return bool
+     * @throws PrestaShopDatabaseException
+     * @throws PrestaShopException
+     */
+    public function softDelete()
+    {
+        $definitions = ObjectModel::getDefinition($this);
+
+        if (empty($definitions['fields']['deleted'])) {
+            throw new PrestaShopException('Field "deleted" is missing from definition in object model ' . get_class($this));
+        }
+        if (!array_key_exists('deleted', $this)) {
+            throw new PrestaShopException('Property "deleted" is missing in object model ' . get_class($this));
+        }
+        $this->deleted = 1;
+
+        return $this->update();
+    }
+
+    /**
      * Toggles object status in database.
      *
      * @return bool Update result
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -903,7 +903,7 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
         if (empty($definitions['fields']['deleted'])) {
             throw new PrestaShopException('Field "deleted" is missing from definition in object model ' . get_class($this));
         }
-        if (!property_exists($this, 'deleted')) {
+        if (!array_key_exists('deleted', get_object_vars($this))) {
             throw new PrestaShopException('Property "deleted" is missing in object model ' . get_class($this));
         }
         $this->deleted = 1;
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -902,7 +902,7 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
         if (empty($definitions['fields']['deleted'])) {
             throw new PrestaShopException('Field "deleted" is missing from definition in object model ' . get_class($this));
         }
-        if (!array_key_exists('deleted', $this)) {
+        if (!property_exists($this, 'deleted')) {
             throw new PrestaShopException('Property "deleted" is missing in object model ' . get_class($this));
         }
         $this->deleted = 1;
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -892,6 +892,7 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
      * If the model object has no "deleted" property or no "deleted" definition field it will throw an exception
      *
      * @return bool
+     *
      * @throws PrestaShopDatabaseException
      * @throws PrestaShopException
      */
--- a/classes/order/OrderState.php
+++ b/classes/order/OrderState.php
@@ -118,10 +118,12 @@ class OrderStateCore extends ObjectModel
      *
      * @return array Order statuses
      */
-    public static function getOrderStates($id_lang, $getDeletedStates = false)
+    public static function getOrderStates($id_lang, $filterDeleted = true)
     {
-        $deletedStates = $getDeletedStates ? '' : 'WHERE deleted = 0';
+        $deletedStates = $filterDeleted ? 'WHERE deleted = 0' : '';
         $cache_id = 'OrderState::getOrderStates_' . (int) $id_lang;
+        $cache_id .= $filterDeleted ? '_filterDeleted' : '';
+
         if (!Cache::isStored($cache_id)) {
             $result = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS('
             SELECT *
--- a/classes/order/OrderState.php
+++ b/classes/order/OrderState.php
@@ -114,19 +114,20 @@ class OrderStateCore extends ObjectModel
      * Get all available order statuses.
      *
      * @param int $id_lang Language id for status name
+     * @param bool $getDeletedStates
      *
      * @return array Order statuses
      */
-    public static function getOrderStates($id_lang)
+    public static function getOrderStates($id_lang, $getDeletedStates = false)
     {
+        $deletedStates = $getDeletedStates ? '' : 'WHERE deleted = 0';
         $cache_id = 'OrderState::getOrderStates_' . (int) $id_lang;
         if (!Cache::isStored($cache_id)) {
             $result = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS('
             SELECT *
             FROM `' . _DB_PREFIX_ . 'order_state` os
-            LEFT JOIN `' . _DB_PREFIX_ . 'order_state_lang` osl ON (os.`id_order_state` = osl.`id_order_state` AND osl.`id_lang` = ' . (int) $id_lang . ')
-            WHERE deleted = 0
-            ORDER BY `name` ASC');
+            LEFT JOIN `' . _DB_PREFIX_ . 'order_state_lang` osl ON (os.`id_order_state` = osl.`id_order_state` AND osl.`id_lang` = ' . (int) $id_lang . ')'
+            . $deletedStates . ' ORDER BY `name` ASC');
             Cache::store($cache_id, $result);
 
             return $result;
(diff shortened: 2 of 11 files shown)


src/Core/Form/ChoiceProvider/OrderStateByIdChoiceProvider.php
```diff
@@ -51,31 +53,46 @@ final class OrderStateByIdChoiceProvider implements FormChoiceProviderInterface,
...
     {
         $orderStates = $this->orderStateDataProvider->getOrderStates($this->languageId);
         $choices = [];
 
         foreach ($orderStates as $orderState) {
+            if ($orderState['deleted'] == 1 && (empty($options['current_state']) || $options['current_state'] != $orderState['id_order_state'])) {
+                continue;
+            }
+            $orderState['name'] .= $orderState['deleted'] == 1 ? ' ' . $this->translator->trans('(deleted)', [], '') : '';
             $choices[$orderState['name']] = $orderState['id_order_state'];
```
matthieu-rolland, line 96: @LouiseBonnard here we will display the mention `(deleted)` with parenthesis, following the order status name, if the status has been deleted. Do you have any idea about what domain we should use ?

LouiseBonnard: @Junebyun is to one to help you get the proper domain. ;-)

Junebyun: @matthieu-rolland Hi! If I got it right the _(deleted)_ message would appear in the BO, correct?

matthieu-rolland: yes 👍 we have a selector with a list of order statuses (ex: "waiting for payment"), and we will add `(deleted)` when the currently selected status is deleted so that the user knows that once he selects another order status this one will disappear.

Junebyun: @matthieu-rolland Hi! You can put it in the _Admin.Global_ domain :)

classes/order/OrderState.php
```diff
@@ -114,19 +114,20 @@ class OrderStateCore extends ObjectModel
      * Get all available order statuses.
      *
      * @param int $id_lang Language id for status name
+     * @param bool $getDeletedStates
      *
      * @return array Order statuses
      */
-    public static function getOrderStates($id_lang)
+    public static function getOrderStates($id_lang, $getDeletedStates = false)
```
jolelievre: The parameter seems a bit confusing (double negative) Maybe use `$filterDeleted = true` instead?

src/Core/Form/ChoiceProvider/OrderStateByIdChoiceProvider.php
matthieu-rolland: @Junebyun thanx 👍 done

controllers/admin/AdminStatusesController.php
```diff
@@ -622,7 +622,10 @@ public function postProcess()
             if (!$order_state->isRemovable()) {
                 $this->errors[] = $this->trans('For security reasons, you cannot delete default order statuses.', [], 'Admin.Shopparameters.Notification');
             } else {
-                return parent::postProcess();
+                $order_state->deleted = 1;
```
jolelievre: So it's always a soft delete even if the state has never been used?

matthieu-rolland: Yes

Not that it would be difficult to implement, but I thought it would add overall complexity to have sometimes a soft delete and sometimes a hard delete, do you think it's worth it?

jolelievre: It's not mandatory for sure, but once we validate and release this behaviour we won't be able to change it back without a BC Maybe this should be discussed with @PrestaShop/prestashop-product-team

matthieu-rolland: @colinegin @MatShir hello guys, do you see a possible use case where doing a hard delete would be better than doing a soft delete?

matks:
Re-running Travis

controllers/admin/AdminStatusesController.php
colinegin: i see 2 potential problems/risks : 
- can it cause a performance problem if there are thousands / hundreds of thousands order statuses in the database with soft delete status ?
- what happens if i try to create an order status that i had decided to delete in the past ? Will i get an error or is it possible to "undelete" an order that was soft deleted ?

jolelievre:
@matthieu-rolland travis sanity tests kept failing so I just rebase the branch ;)

controllers/admin/AdminStatusesController.php
jolelievre: 1- I think there can only be drawbacks in BO as in any other parts the soft deleted states are ignored
2- if you deleted one and recreate it this will be a new record

jolelievre: So if we always soft delete them regardless of their use they will keep piling up But then maybe this should be check directly in `OrderState::delete` method, so the behaviour will remain the same no matter where it's called (a module, the legacy controller, the migrated page)

classes/order/OrderState.php
```diff
@@ -114,19 +114,22 @@ class OrderStateCore extends ObjectModel
...
      *
      * @param int $id_lang Language id for status name
+     * @param bool $getDeletedStates
      *
      * @return array Order statuses
      */
-    public static function getOrderStates($id_lang)
+    public static function getOrderStates($id_lang, $filterDeleted = true)
     {
+        $deletedStates = $filterDeleted ? 'WHERE deleted = 0' : '';
```
jolelievre:
```suggestion
        $deletedStates = $filterDeleted ? ' WHERE deleted = 0' : '';
```

controllers/admin/AdminStatusesController.php
kpodemski: @jolelievre @colinegin 

order states are so rarely deleted that it shouldn't be an issue here, if we have `softDelete` method available from 1.7.7 we can - as a developers - think about proper indexes if we need more performance :) but it's not something what should be done in core IMO

classes/order/OrderState.php
matthieu-rolland: @jolelievre done ! thanx

classes/ObjectModel.php
```diff
@@ -887,6 +887,29 @@ public function deleteSelection($ids)
...
+     * @throws PrestaShopException
+     */
+    public function softDelete()
+    {
+        $definitions = ObjectModel::getDefinition($this);
+
+        if (empty($definitions['fields']['deleted'])) {
+            throw new PrestaShopException('Field "deleted" is missing from definition in object model ' . get_class($this));
+        }
+        if (!array_key_exists('deleted', $this)) {
```
PierreRambaud: Please don't test with array_key_exists for object :) It's not compatible with PHP7.4

PierreRambaud:
```suggestion
        if (!array_key_exists('deleted', get_object_vars($this))) {
```

PierreRambaud (changes requested):
Just a little comment about PHP 7.4 compatibility

classes/ObjectModel.php
jolelievre: What about this? Does it still work in 7.4?

```suggestion
        if (!property_exists($this, 'deleted')) {
```

matthieu-rolland: thanx ! done (yes it should work in 7.4)

matthieu-rolland: thanx, done!

controllers/admin/AdminStatusesController.php
matthieu-rolland: thank you for your input @kpodemski ! you're right, and it has been decided that it will be soft delete only.

jolelievre (dismissed):

classes/ObjectModel.php
PierreRambaud: Nope nope nope, the `property_exists` is also checking for private and protected property :scream:

PierreRambaud: Since it's related to ObjectModel, maybe people are using `private $deleted;` :sweat_smile:

PierreRambaud: image

PierreRambaud (approved):

jolelievre (approved):

classes/ObjectModel.php
jolelievre: Ok, good to know indeed!

Junebyun (approved):

SD1982:

Progi1984:
Thanks @matthieu-rolland

matks:
I mark this PR as key feature, it needs to be quoted into the Release Note (or the devdocs)

PR #19582, opened 2020-06-04
Order products pagination

### LINKED ISSUE #19486: Droplist to change the number of products in the list
**User story**
As a merchant, when I have more than 50 products on the order page I would like to check them quickly to edit one without having to go through all the pages.

**Solution**
The best solution would have been a search but because we are stabilizing at the beta phase. The simpler solution, we will be to add a drop list to change the number of products displayed by pages.

**Detailed specs**

On the left-bottom of the product list, a droplist can change the number of products on the product list. The possible value is 8, 20, 50, or 100.

Capture d’écran 2020-06-03 à 17 18 43
Invision link: https://invis.io/YKTGGFZAMCB#/392128609_Products_-_Pagination

The selected value will be saved and shared to all the orders
If a merchant selects 20 on one order then, on the next order, the product list will display the 20 products per page.

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Add selector to control the number of products displayed in Order view page, save this in a global shop configuration
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19486
| How to test?  | Create an order with a lot of products (at least 20 different ones). Go to the order view page and have fun with the number of element displayed: change the number, check that it displays the right number of product, go to different pages Refresh the page to check that the number has been saved (even on other orders) Try to perform edition, when you cancel the pagination must be correctly displayed, ...

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19582)


src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
```diff
@@ -110,28 +109,41 @@
...
+                  <option value="{{ numPageOption }}"{% if numPageOption == paginationNum %} selected{% endif %}>{{ numPageOption }}</option>
+                {% endfor %}
+              </select>
+            </div>
+          </div>
+          {% set numPages = (orderForViewing.products.products|length / paginationNum)|round(0, 'ceil') %}
+          <nav aria-label="Products Navigation"{% if orderForViewing.products.products|length <= paginationNum %} class="d-none"{% endif %} id="orderProductsNavPagination">
+            <ul class="pagination" id="orderProductsTablePagination" data-num-per-page="{{ paginationNum }}" data-num-pages="{{ numPages }}">
+              <li class="page-item disabled" id="orderProductsTablePaginationPrev">
+                <a class="page-link" href="javascript:void(0);" aria-label="Previous">
```
matthieu-rolland, line 128: I've rarely seen this, `javascript:void(0);`, is it really useful here? 🤔 

(But you're not the one who wrote it so...)

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -458,6 +458,13 @@ public function viewAction(int $orderId, Request $request): Response
         /** @var OrderSiblingProviderInterface $orderSiblingProvider */
         $orderSiblingProvider = $this->get('prestashop.adapter.order.order_sibling_provider');
 
+        $paginationNum = (int) $this->configuration->get('PS_ORDER_PRODUCTS_NB_PER_PAGE', 8);
+        $paginationNumOptions = [8, 20, 50, 100];
```
matthieu-rolland: Maybe this hardcoded array should be set as a constant? (it's possible to have a const containing an array as of PHP 5.6)

(idem for the value `8` in the line just above)

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1415,6 +1424,32 @@ public function cancellationAction(int $orderId, Request $request)
...
+     *
+     * @param Request $request
+     *
+     * @return JsonResponse
+     */
+    public function configureProductPaginationAction(Request $request): JsonResponse
+    {
+        $numPerPage = (int) $request->request->get('numPerPage');
+        if ($numPerPage < 1) {
+            $numPerPage = 10;
```
matthieu-rolland: Idem here (hardcoded value), it would be nice if those values were defined somewhere (as a const in this class, or somewhere else?)

matthieu-rolland (commented):
Small feedback, but I'm not blocking the PR though, your choice 👍

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/products.html.twig
NeOMakinG: This is to avoid scroll moove or refresh by clicking on the link, btw you could also just kick the href off and add a `cursor: pointer;` or maybe using a button !

jolelievre: It's probably ideal, this is the existing pagination I prefer not to modify it since it works for now 😅

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: Yep, good idea Done

admin-dev/themes/new-theme/js/pages/order/view/order-product-renderer.js
```diff
@@ -195,6 +196,29 @@ export default class OrderProductRenderer {
     this.togglePaginationControls();
   }
 
+  updateNumPerPage(numPerPage) {
+    const $rows = $(OrderViewPageMap.productsTable).find('tr[id^="orderProduct_"]');
+    const $tablePagination = $(OrderViewPageMap.productsTablePagination);
+    const numPages = Math.ceil($rows.length / numPerPage);
```
Progi1984, line 205: And if you have 0 rows ? or if numPerPage = 0 ?

jolelievre: `numPerPage` is bounded on the controller side, and I didn't think about empty products But you're right I should add some safeguard in the js code as well

jolelievre: Done!

jolelievre:
@Progi1984 @matthieu-rolland All your feedback have been addressed 😉

Progi1984 (approved):

NeOMakinG (approved):

Junebyun (approved):

LouiseBonnard (approved):

PierreRambaud (approved):

Progi1984:
Thanks @jolelievre

PR #19631, opened 2020-06-08
FO: fix cumulative percentage reduction amount value in cart 

### LINKED ISSUE #18810: Cumulative cart Rules with %age discount don't display the correct discount amount
#### Describe the bug

If tax rule are cumulated the total discount is correctly calculated but the display of individual discount is wrong

#### Expected behavior

The value displayed for each discount should be correct the second cart rules value has to be calculated on the total cart amount with the first cart rule included

#### Steps to Reproduce

1. Go to BO => catalog => discount => cart rules and create 2 new cart rule :
- first cart rule => 10% 
- second cart rule => 10% 
2. GO to FO and create cart with 5 "pack mug + framed poster" (just to have a round price )
3. Apply one cart rule the value for this cart rule is in this case exemple 21€ (screen 1)
4.Apply the second tax rules the total is 39.90€ but individually the tax rules value is still 21€ ( screen 2 )
5. In this case we should have the first cart rules = 21€ and the second one 18,90€

**Screenshots**
**Screen 1**
capture d'écran_5

**Screen 2**
capture d'écran_6

#### Additional information

* PrestaShop version: 1.7.7.x  and 1.7.6.x
* PHP version: 7.2

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | The amount corresponding to a percentage reduction was always calculated on the base total cart amount, even if other reductions were applied before. (the bug only impacted the reduction value displayed, not the actual total to be paid)
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18810
| How to test?  | See the issue #18810

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19631)

### DIFF
--- a/tests/Integration/Behaviour/Features/Context/CartRuleFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/CartRuleFeatureContext.php
@@ -391,4 +391,26 @@ class CartRuleFeatureContext extends AbstractPrestaShopFeatureContext
             throw new \RuntimeException(sprintf('Expects %s, got %s instead', $expectedValue, $cartRule->reduction_amount));
         }
     }
+
+    /**
+     * @Then /^cart rule "(.+)" has a contextual reduction value of (\d+.\d+)$/
+     *
+     * @param $cartRuleName
+     * @param $expectedValue
+     */
+    public function checkCartRuleContextualValue(string $cartRuleName, float $expectedValue)
+    {
+        $cartRules = $this->getCurrentCart()->getCartRules();
+        $cartRuleFound = false;
+        foreach ($cartRules as $currentCartRule) {
+            if ($currentCartRule['description'] === $cartRuleName && round($currentCartRule['value_real'], 6) != round($expectedValue, 6)) {
+                throw new \RuntimeException(sprintf('Expects %s, got %s instead', $expectedValue, $currentCartRule['value_real']));
+            }
+            if ($currentCartRule['description'] === $cartRuleName) $cartRuleFound = true;
+        }
+
+        if (!$cartRuleFound) {
+            throw new \RuntimeException(sprintf('The cart rule "%s" was not found', $cartRuleName));
+        }
+    }
 }
--- a/tests/Integration/Behaviour/Features/Context/CartRuleFeatureContext.php
+++ b/tests/Integration/Behaviour/Features/Context/CartRuleFeatureContext.php
@@ -403,7 +403,8 @@ class CartRuleFeatureContext extends AbstractPrestaShopFeatureContext
         $cartRules = $this->getCurrentCart()->getCartRules();
         $cartRuleFound = false;
         foreach ($cartRules as $currentCartRule) {
-            if ($currentCartRule['description'] === $cartRuleName && round($currentCartRule['value_real'], 6) != round($expectedValue, 6)) {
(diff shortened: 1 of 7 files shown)


jolelievre (dismissed):

Progi1984:
@matthieu-rolland Two points :
 - Travis CI is :red_circle: 
 - No way to add Behat tests like you add code in Cart.php & CartRule.php ?

matthieu-rolland:
> @matthieu-rolland Two points :
> 
> * Travis CI is 🔴
> * No way to add Behat tests like you add code in Cart.php & CartRule.php ?

@Progi1984  done 👍

tests/Integration/Behaviour/Features/Context/CartRuleFeatureContext.php
```diff
@@ -371,4 +371,26 @@ public function checkCartRuleValueForCustomer($customerName, $position, $expecte
...
+     */
+    public function checkCartRuleContextualValue(string $cartRuleName, float $expectedValue)
+    {
+        $cartRules = $this->getCurrentCart()->getCartRules();
+        $cartRuleFound = false;
+        foreach ($cartRules as $currentCartRule) {
+            if ($currentCartRule['description'] === $cartRuleName && round($currentCartRule['value_real'], 6) != round($expectedValue, 6)) {
+                throw new \RuntimeException(sprintf('Expects %s, got %s instead', $expectedValue, $currentCartRule['value_real']));
+            }
+            if ($currentCartRule['description'] === $cartRuleName) $cartRuleFound = true;
```
Progi1984:
```suggestion
            if ($currentCartRule['description'] === $cartRuleName) {
                $cartRuleFound = true;
            }
```

matks (dismissed):
I have no idea if this code works 😄 because cart rules are so complex.

However I trust our behat tests. And if we do a mistake undetected by our behat tests, we'll find out and improve the tests

matthieu-rolland:
> I have no idea if this code works 😄 because cart rules are so complex.
> 
> However I trust our behat tests. And if we do a mistake undetected by our behat tests, we'll find out and improve the tests

The good news is that this PR "should not" impact the order total value, it only fixes how the "contextual reduction value" is calculated in case of cumulative percentage reductions, whose result is used to display reduction values in FO, in the cart.

(we talk about contextual values because a percentage reduction will not have the same value depending on wether there were other reductions previously applied or not)

jolelievre (dismissed):

matks (dismissed):

PierreRambaud (dismissed):

sarahdib:
@matthieu-rolland 
there is a bug in create an order with this PR 
https://drive.google.com/file/d/1AqWR24QwsUGPxT4dEcjjX3Tv3oUpbDQj/view

Without
Capture d’écran 2020-06-19 à 12 00 00

matthieu-rolland:
> https://drive.google.com/file/d/1AqWR24QwsUGPxT4dEcjjX3Tv3oUpbDQj/view

thanx @sarahdib !    In your example, the version without my PR is wrong too, in your video test, it's the last reduction that has its value wrong.

I'm working on it 👍

atomiix (approved):

matks (approved):

Progi1984:
Thanks @matthieu-rolland & @sarahdib

PR #19646, opened 2020-06-09
Synchronize cart_product and order_detail when quantity is changed

### LINKED ISSUE #19592: Update product's quantity in the order does not update the discount

#### Describe the bug

I create a cart rule on the product Brown bear at 9 euro with an automated discount of 50% on the total of the order.
On the back-office, when I update the quantity on the order the discount is not updated

#### Expected behavior

The discount should apply the 50% reduction on all the product amount even if after they are updated 

#### Steps to Reproduce

1. Create a cart rule of 50% reduction on the total of the order with **no code** specifically applied to a product
2. Add the product to any order
3. edit the product quantity

**Screenshots**

After editing the quantity, the discount is not updated
image

#### Additional information

* PrestaShop version: N/A
* PHP version: N/A

### LINKED ISSUE #19607: The minimum quantity is not respected when I add a product
Be aware of the EPIC - Errors when adding a product with minimum quantity in order page #19595

#### Describe the bug

When an edit the quantity of product with min quantity required then delete in the order and I add the product again, it does not throw an error.

#### Expected behavior

Display the error message: "Error! Minimum quantity of "X" must be added when I add the product again.

#### Steps to Reproduce

1. Order a product with a min quantity=3
2. Edit the product quantity on BO in the order page quantity=2
3. Delete the product 
4. Add the product again with the quantity =2
5. Error, it should display an error message: Error! Minimum quantity of "3" must be added

**Screenshots**

#### Additional information

* PrestaShop version: N/A
* PHP version: N/A

sowbiba:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | 2 assertions : <br>When a product is deleted, the existing cart_product is decreased by the quantity setted on order_detail.<br>When a cart_product already exist and the quantity is increased minimum quantity is not checked<br><br>And On an order, when quantity is updated, cart_product is not updated aswell<br><br>... led to this issue<br>The fix is to update quantity on cart_product when the order's product quantity is updated 
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19607 and fixes #19592 
| How to test?  | See the before and after videos<br>*Before* : https://drive.google.com/file/d/1qTSkNfmmrjPhAkY9VDPy4sbhcWFdY6Ui/view<br>*After* : https://drive.google.com/file/d/1pRolBU_UH_xDRl8BcR2Yv7XMddPKtSyj/view

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19646)

### DIFF
--- a/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
@@ -28,6 +28,7 @@ namespace PrestaShop\PrestaShop\Adapter\Order\CommandHandler;
 
 use Address;
 use Attribute;
+use Cache;
 use Carrier;
 use Cart;
 use CartRule;
@@ -221,6 +222,9 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
      */
     private function updateOrderCartRules(Order $order, ?OrderInvoice $invoice, array $cartCartRules)
     {
+        CartRule::resetStaticCache();
+        Cache::clean('getContextualValue_*');
+
         foreach ($cartCartRules as $cartCartRule) {
             $rule = new CartRule($cartCartRule['id_cart_rule']);
 
--- a/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
@@ -26,13 +26,9 @@
 
 namespace PrestaShop\PrestaShop\Adapter\Order\CommandHandler;
 
-use Cart;
-use CartRule;
-use Context;
 use Customization;
 use Hook;
 use Order;
-use OrderCartRule;
 use OrderDetail;
 use OrderInvoice;
 use PrestaShop\PrestaShop\Adapter\Order\AbstractOrderHandler;
@@ -245,93 +241,5 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements 
                 throw new ProductOutOfStockException('Not enough products in stock');
             }
         }
-    }
-
-    /**
-     * @param Cart $cart
-     * @param Order $order
-     * @param OrderInvoice|null $invoice
-     *
-     * @throws \PrestaShopDatabaseException
-     * @throws \PrestaShopException
-     */
-    private function updateCartRules(Cart $cart, Order $order, ?OrderInvoice $invoice = null)
-    {
-        $computingPrecision = Context::getContext()->getComputingPrecision();
-
-        $cart = $this->syncCartOrderCartRules($order, $cart);
-        Context::getContext()->cart = $cart;
-        CartRule::autoAddToCart();
-        CartRule::autoRemoveFromCart();
-
-        $orderCartRulesData = $order->getCartRules();
-        foreach ($orderCartRulesData as $orderCartRuleData) {
-            $orderCartRule = new OrderCartRule((int) $orderCartRuleData['id_order_cart_rule']);
-            $idCartRule = (int) $orderCartRule->id_cart_rule;
-            $cartRule = new CartRule($idCartRule);
-            $values = [
-                'tax_incl' => \Tools::ps_round($cartRule->getContextualValue(true), $computingPrecision),
-                'tax_excl' => \Tools::ps_round($cartRule->getContextualValue(false), $computingPrecision),
-            ];
-
-            if (
-                ($values['tax_incl'] !== \Tools::ps_round($orderCartRule->value, $computingPrecision)) ||
-                ($values['tax_excl'] !== \Tools::ps_round($orderCartRule->value_tax_excl, $computingPrecision))
-            ) {
-                $orderCartRule->value = $values['tax_incl'];
-                $orderCartRule->value_tax_excl = $values['tax_excl'];
-
-                $orderCartRule->update();
-            }
-        }
-
-        $order = $this->orderAmountUpdater->update($order, $cart, $invoice && !empty($invoice->id));
-        $order->update();
-    }
-
-    /**
-     * @param Order $order
-     * @param Cart $cart
-     *
-     * @return Cart
-     */
-    private function syncCartOrderCartRules(Order $order, Cart $cart): Cart
-    {
-        $orderCartRulesData = $order->getCartRules();
-        $orderCartRulesIds = array_map(
-            function (array $orderCartRuleData) { return (int) $orderCartRuleData['id_cart_rule']; },
-            $orderCartRulesData
-        );
-
-        $cartRules = $cart->getCartRules();
-        $cartRulesIds = array_map(
-            function (array $cartRule) { return (int) $cartRule['id_cart_rule']; },
-            $cartRules
-        );
-
-        sort($orderCartRulesIds);
-        sort($cartRulesIds);
-
-        $difference = array_diff($orderCartRulesIds, $cartRulesIds);
-
-        if (count($orderCartRulesIds) > count($cartRulesIds)) {
-            foreach ($difference as $cartRuleId) {
-                $cart->addCartRule($cartRuleId);
-            }
-        } else {
-            foreach ($difference as $cartRuleId) {
-                $cartRule = $cartRules[$cartRuleId];
-                $order->addCartRule(
-                    $cartRuleId,
-                    $cartRule->name[$order->id_lang],
-                    [
-                        'tax_incl' => $cartRule['value_real'],
-                        'tax_excl' => $cartRule['value_tax_exc'],
-                    ]
-                );
-            }
-        }
-
-        return $cart;
     }
 }
--- a/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
@@ -160,6 +160,7 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements 
             $orderDetail,
             $old_quantity,
             $command->getQuantity(),
+            false,
             ($orderInvoice && !empty($orderInvoice->id))
         );
 
--- a/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
@@ -109,25 +109,8 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements 
             $res &= $order->update();
         }
 
-        $old_quantity = $orderDetail->product_quantity;
-
-        $orderDetail->product_quantity = $product_quantity;
-        $orderDetail->reduction_percent = 0;
-
-        // update taxes
-        $res &= $orderDetail->updateTaxAmount($order);
-
-        // Save order detail
-        $res &= $orderDetail->update();
-
         // Update quantity and amounts
-        $order = $this->orderProductUpdater->update(
-            $order,
-            $orderDetail,
-            $old_quantity,
-            $command->getQuantity(),
-            $orderInvoice
-        );
+        $order = $this->orderProductUpdater->update($order, $orderDetail, $product_quantity, $orderInvoice);
 
         if (!$res) {
             throw new OrderException('An error occurred while editing the product line.');
--- a/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
@@ -164,7 +164,7 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements 
 
         // Apply the changes on the cart
         $cart->updateQty(
-            $old_quantity - $orderDetail->product_quantity,
+            abs($old_quantity - $orderDetail->product_quantity),
             $orderDetail->product_id,
             $orderDetail->product_attribute_id,
             false,
@@ -276,6 +276,14 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements 
         }
     }
 
+    /**
+     * @param Cart $cart
+     * @param Order $order
+     * @param OrderInvoice|null $invoice
+     *
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
     private function updateCartRules(Cart $cart, Order $order, ?OrderInvoice $invoice = null)
     {
         $computingPrecision = Context::getContext()->getComputingPrecision();
@@ -286,8 +294,7 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements 
         CartRule::autoRemoveFromCart();
 
         $orderCartRulesData = $order->getCartRules();
-        /** @var OrderCartRule $orderCartRule */
-        foreach($orderCartRulesData as $orderCartRuleData) {
+        foreach ($orderCartRulesData as $orderCartRuleData) {
             $orderCartRule = new OrderCartRule((int) $orderCartRuleData['id_order_cart_rule']);
             $idCartRule = (int) $orderCartRule->id_cart_rule;
             $cartRule = new CartRule($idCartRule);
@@ -311,17 +318,23 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements 
         $order->update();
     }
 
+    /**
+     * @param Order $order
+     * @param Cart $cart
+     *
+     * @return Cart
+     */
     private function syncCartOrderCartRules(Order $order, Cart $cart): Cart
     {
         $orderCartRulesData = $order->getCartRules();
         $orderCartRulesIds = array_map(
-            function(array $orderCartRuleData) { return (int)$orderCartRuleData['id_cart_rule']; },
+            function (array $orderCartRuleData) { return (int) $orderCartRuleData['id_cart_rule']; },
             $orderCartRulesData
         );
 
         $cartRules = $cart->getCartRules();
         $cartRulesIds = array_map(
-            function(array $cartRule) { return (int)$cartRule['id_cart_rule']; },
+            function (array $cartRule) { return (int) $cartRule['id_cart_rule']; },
             $cartRules
         );
 
--- a/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
@@ -32,7 +32,7 @@ use Order;
 use OrderDetail;
 use OrderInvoice;
 use PrestaShop\PrestaShop\Adapter\Order\AbstractOrderHandler;
-use PrestaShop\PrestaShop\Adapter\Order\OrderProductUpdater;
+use PrestaShop\PrestaShop\Adapter\Order\OrderProductQuantityUpdater;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\CannotEditDeliveredOrderProductException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Product\Command\UpdateProductInOrderCommand;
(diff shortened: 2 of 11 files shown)


atomiix (dismissed):

Progi1984 (dismissed):

PierreRambaud (dismissed):

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -249,4 +275,92 @@ private function assertProductCanBeUpdated(
...
+        sort($cartRulesIds);
+
+        $difference = array_diff($orderCartRulesIds, $cartRulesIds);
+
+        if (count($orderCartRulesIds) > count($cartRulesIds)) {
+            foreach ($difference as $cartRuleId) {
+                $cart->addCartRule($cartRuleId);
+            }
+        } else {
+            foreach ($difference as $cartRuleId) {
```
sowbiba: I do not have the case where some rules are on the cart but not on the order. Maybe I need to remove this until we need it ?

matks: I dont know either if it could happen.

3 possible solutions
- ignore it, we dont handle it because we think it never happens
- ignore it, we dont handle it because we have no idea what should happen and we hope "do nothing" is the right thing to do
- throw a RuntimeException "should never happen" here

sowbiba: Solution 3 is brutal but I like it. We can't miss it if it happens

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -249,4 +275,92 @@ private function assertProductCanBeUpdated(
...
+     * @param Cart $cart
+     * @param Order $order
+     * @param OrderInvoice|null $invoice
+     *
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
+    private function updateCartRules(Cart $cart, Order $order, ?OrderInvoice $invoice = null)
+    {
+        $computingPrecision = Context::getContext()->getComputingPrecision();
```
jolelievre: Since the precision depends on the currency, so depends on the cart's currency in BO side you should get precision from it Like in https://github.com/PrestaShop/PrestaShop/blob/develop/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php#L697

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -249,4 +275,92 @@ private function assertProductCanBeUpdated(
...
+
+        $cart = $this->syncCartOrderCartRules($order, $cart);
+        Context::getContext()->cart = $cart;
```
jolelievre: It would be better to integrate the `ContextStateManager` for this

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -249,4 +275,92 @@ private function assertProductCanBeUpdated(
...
+        CartRule::autoAddToCart();
+        CartRule::autoRemoveFromCart();
+
+        $orderCartRulesData = $order->getCartRules();
+        foreach ($orderCartRulesData as $orderCartRuleData) {
+            $orderCartRule = new OrderCartRule((int) $orderCartRuleData['id_order_cart_rule']);
+            $idCartRule = (int) $orderCartRule->id_cart_rule;
+            $cartRule = new CartRule($idCartRule);
+            $values = [
+                'tax_incl' => \Tools::ps_round($cartRule->getContextualValue(true), $computingPrecision),
```
jolelievre: (optional) can you add `use Tools;` to remove the `\`

jolelievre (commented):
Some improvement suggested

jolelievre:
To be checked but this fix might also fix this issuse https://github.com/PrestaShop/PrestaShop/issues/19592 since it's related to updating the product quantity

atomiix (dismissed):

Progi1984:
@sowbiba Travis is :red_circle: 

https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/349465874#L5300

matthieu-rolland (commented):
It seems like the update of quantity in the table ps_customization in case this is a customized product is missing.

what do you think @sowbiba ? see `src/Adapter/Order/Refund/OrderProductRemover.php:246`

src/Adapter/Order/OrderProductUpdater.php
```diff
@@ -0,0 +1,519 @@
...
+            $orderCartRulesData
+        );
+
+        $cartRules = $cart->getCartRules();
+        $cartRulesIds = array_map(
+            function (array $cartRule) { return (int) $cartRule['id_cart_rule']; },
+            $cartRules
+        );
+
+        if (count($orderCartRulesIds) === count($cartRulesIds)) {
```
jolelievre: You shouldn't rely on `count` Let's say a cart rule was removed and another one was added the count will be equals but the cart rules are not in sync any more
You can only trust the cart rule id

src/Adapter/Order/OrderProductUpdater.php
```diff
@@ -0,0 +1,519 @@
...
+            return $cart;
+        }
+
+        if (count($orderCartRulesIds) > count($cartRulesIds)) {
+            if ($delete) { // Product deletion, the extra order cart rules have to be deleted
```
jolelievre: For the same reason the `delete` feature shouldn't depend on the changed quantity There can be cart rules removed EVEN when a product is added (let's image a cartrule that is only valid when the product is between 5 and 10 instances)

src/Adapter/Order/OrderProductUpdater.php
```diff
@@ -0,0 +1,519 @@
...
+                foreach ($order->getCartRules() as $orderCartRule) {
+                    foreach (array_diff($orderCartRulesIds, $cartRulesIds) as $idCartRule) {
+                        if ($idCartRule == $orderCartRule['id_cart_rule']) {
+                            $orderCartRule = new OrderCartRule($orderCartRule['id_order_cart_rule']);
+                            $orderCartRule->delete();
+                        }
+                    }
+                }
+            } else { // desynchronization of cartRules, need to be fixed
+                sort($orderCartRulesIds);
```
jolelievre: All the sort and array_diff algorithms that have been used so far have had issues There's always a use case that doesn't work with this implementation Besides it's not "developer friendly" as the algorithm is not understandable straight away

I would advise a clearer implementation (from a human perspective) like what has been done in this PR https://github.com/PrestaShop/PrestaShop/pull/19698 in the separated handlers and merge the principle into a generic in this (awesome) service

Basically, a double loop through order cart rules + new order cart rules (I don't think we need the previous ones since they are in order already), from which you populate two arrays `$orderCartRulesToBeRemoved` `$missingOrderCartRules` And then perform the delete/create accordingly

src/Adapter/Order/OrderProductUpdater.php
```diff
@@ -0,0 +1,519 @@
...
+            $orderCartRule = new OrderCartRule((int) $orderCartRuleData['id_order_cart_rule']);
+            $idCartRule = (int) $orderCartRule->id_cart_rule;
+            $cartRule = new CartRule($idCartRule);
+            $values = [
+                'tax_incl' => Tools::ps_round($cartRule->getContextualValue(true), $computingPrecision),
+                'tax_excl' => Tools::ps_round($cartRule->getContextualValue(false), $computingPrecision),
+            ];
+
+            if (
+                ($values['tax_incl'] !== Tools::ps_round($orderCartRule->value, $computingPrecision)) ||
```
jolelievre: Not sure if this check is necessary, we could update the order all the time (just one additional request is acceptable if it ensure stability) Besides the `$values` could be removed with simpler code:
```php
$orderCartRule->value = Tools::ps_round($cartRule->getContextualValue(true), $computingPrecision);
$orderCartRule->value_tax_excl = Tools::ps_round($cartRule->getContextualValue(false), $computingPrecision);
```

Progi1984 (dismissed):

matks (dismissed):

matks:
Travis fails
```
    Features/Scenario/Order/order_from_bo.feature:97
```

matthieu-rolland (dismissed):

src/Adapter/Order/OrderProductUpdater.php
```diff
@@ -0,0 +1,545 @@
...
+     * @return Order
+     *
+     * @throws OrderException
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
+    public function update(
+        Order $order,
+        OrderDetail $orderDetail,
+        int $oldQuantity,
```
jolelievre: I believe this parameter is useless since it's always equal to `$orderDetail->product_quantity` ?

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -146,35 +154,21 @@ public function handle(UpdateProductInOrderCommand $command)
...
-        $order_carrier = new OrderCarrier((int) $order->getIdOrderCarrier());
-        if (Validate::isLoadedObject($order_carrier)) {
-            $order_carrier->weight = (float) $order->getTotalWeight();
-            $res &= $order_carrier->update();
-            if ($res) {
-                $order->weight = sprintf('%.3f ' . Configuration::get('PS_WEIGHT_UNIT'), $order_carrier->weight);
-            }
-        }
+        // Update quantity and amounts
+        $order = $this->orderProductUpdater->update(
```
jolelievre: Shouldn't the OrderDetail updates (performed on L140) be done in the OrderProductUpdater as well? Along with Customization updates?

Also I might be wrong but it would seem more logical to make the order details updates before updating the order no? And we could could maybe use the OrderAmountUpdater service as well here to clean a big piece of code? (total computing and all?)

jolelievre: Oh actually I just realized that `OrderProductUpdater` already uses `OrderAmountUpdater` which is even better!

I also understood that the total udpates in this class are related to the invoice not the order, since it's the invoice related to the OrderDetail maybe it should be done in the service as well?

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -1063,6 +1064,12 @@ private function getOrderDiscounts(int $orderId): array
      */
     private function getOrderDiscountByName(int $orderId, string $cartRuleName): ?OrderDiscountForViewing
     {
+        $cartRuleName = CommonFeatureContext::getContainer()->get('translator')->trans(
```
jolelievre, line 1067: Are you sure of this one? Since OrderCartRule only have one name (lot localized), I'm afraid it won't match any more with the english name in database?

src/Adapter/Order/OrderProductUpdater.php
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    /**
+     * @param Cart $cart
+     * @param Order $order
+     *
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
+    private function recalculateCartRules(Cart $cart, Order $order): void
```
jolelievre: I believe this one  is redundant with `updateOrderCartRules` where we already loop through cart rules and update OrderCartRule when they are still present

src/Adapter/Order/OrderProductUpdater.php
```diff
@@ -0,0 +1,545 @@
...
+     *
+     * @throws OrderException
+     */
+    private function updateCustomizationOnProductDelete(Order $order, OrderDetail $orderDetail, int $oldQuantity): void
+    {
+        if (!(int) $order->getCurrentState()) {
+            throw new OrderException('Could not get a valid Order state before deletion');
+        }
+
+        if ($order->hasBeenDelivered()) {
```
jolelievre: Are you this use case should happen? I thought updates were blocked on delievered/paid orders? 🤔

This kinf of updates only happen when refunding products but it's another feature/handler

jolelievre: So after a quick check in the handler L216, udpates are blocked for **delivered** orders, same for deleting products L279, but I guess it can happen for a paid order not yet delivered So only one use case to manage here
Unless this product updater could be used in another handler?

MatShir:
@jolelievre Be aware the problem appears also when the price of the product is edited for the cart rule wrong calculation. Check the issue #19890 caught by @sowbiba

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
sowbiba: In https://github.com/PrestaShop/PrestaShop/blob/develop/src/Adapter/Cart/CommandHandler/SetFreeShippingToCartHandler.php#L118 the cart rule name is translated. That's why I had 'Livraison gratuite' and not 'Free shipping'

src/Adapter/Order/OrderProductUpdater.php
sowbiba: done

src/Adapter/Order/OrderProductUpdater.php
sowbiba: done

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
sowbiba: done

src/Adapter/Order/OrderProductUpdater.php
sowbiba: In UpdateProductInOrderHandler we need to update the orderDetail quantity AND keep the old value to update the cart and stock

jolelievre: It's not used any more in `UpdateProductInOrderHandler` (only to pass it to this method), so it's not needed to ave this extra parameter

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -146,35 +120,15 @@ public function handle(UpdateProductInOrderCommand $command)
...
-        StockAvailable::updateQuantity(
-            $orderDetail->product_id,
-            $orderDetail->product_attribute_id,
-            $old_quantity - $orderDetail->product_quantity,
-            $order->id_shop
+        // Update quantity and amounts
+        $order = $this->orderProductUpdater->update(
+            $order,
+            $orderDetail,
+            $old_quantity,
```
jolelievre: My bad `$old_quantity` is till used to save the former quantity of OrderDetail
But it's only because it is updated right before this call, if the OrderDetail update is made in this function (which makes sense), then there is no more need to keep the old quantity or to pass it as a parameter to this function

Do you thin the `orderProductUpdater` should also be responsible for:
- updating the prices?
- updating the customization quantity?

If the prices is too much and would be better staying here, then maybe the `update` method should be called `updateQuantity` instead to have a clear view of its responsibility

src/Adapter/Order/OrderProductUpdater.php
sowbiba: My point is that we need in OrderProductUpdater to know what was the original value of the quantity. I can do the `set` in the service 
$oldQuantity = $orderDetail->product_quantity
$orderDetail->product_quantity = $newQuantity

sowbiba: done

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
sowbiba: I just read this message. It's done. The update of product_quantity and taxes is now in this service

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
jolelievre: Oh ok, what worries me is that it might be the only one 😅 
Besides the default language should be english, but nevermind, as long as the test suite is happy we can always fix it later if necessary

src/Adapter/Order/OrderProductUpdater.php
```diff
@@ -0,0 +1,561 @@
...
+use StockMvt;
+use Tools;
+use Validate;
+use Warehouse;
+
+/**
+ * Increase or decrease quantity of an order's product.
+ * Recalculate cart rules, order's prices and shipping infos.
+ */
+class OrderProductUpdater
```
jolelievre: Reading this description and the content of the method, it seems more logical to call the method `updateQuantity`

jolelievre (commented):
Last comment for me before approval 😉

src/Adapter/Order/OrderProductUpdater.php
sowbiba: I renamed the service to OrderProductQuantityUpdater ;)

matks (dismissed):

matthieu-rolland (dismissed):

tests/Integration/Behaviour/Features/Scenario/Order/order_cart_rules.feature
```diff
@@ -80,6 +80,66 @@ Feature: Order from Back Office (BO)
...
+      | total_discounts_tax_incl | 15.900 |
+      | total_paid_tax_excl      | 30.8   |
+      | total_paid_tax_incl      | 32.650 |
+      | total_paid               | 32.650 |
+      | total_paid_real          | 0.0    |
+      | total_shipping_tax_excl  | 7.0    |
+      | total_shipping_tax_incl  | 7.42   |
+    When I edit product "Test Product Cart Rule On Select Product" to order "bo_order1" with following products details:
+      | amount        | 3                       |
+      | price         | 15                      |
```
jolelievre, line 126: Can you use a different amount? This would allow to check that the discount is correctly updated

tests/Integration/Behaviour/Features/Scenario/Order/order_cart_rules.feature
```diff
@@ -80,6 +80,66 @@ Feature: Order from Back Office (BO)
...
+    Then order "bo_order1" should have 5 products in total
+    Then order "bo_order1" should contain 3 products "Test Product Cart Rule On Select Product"
+    Then order "bo_order1" should have 1 cart rule
+    Then order "bo_order1" should have cart rule "CartRuleAmountOnSelectedProduct" with amount "$15.00"
```
jolelievre: Oh actually it seems there's a problem here, the amount should be `$45` no?

jolelievre (changes requested):
I block this one because of the invalid amount in behat test

jolelievre:
@sowbiba I believe this should fix this issue as well no? https://github.com/PrestaShop/PrestaShop/issues/19890

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -221,6 +222,9 @@ public function handle(AddProductToOrderCommand $command)
      */
     private function updateOrderCartRules(Order $order, ?OrderInvoice $invoice, array $cartCartRules)
     {
+        CartRule::resetStaticCache();
```
jolelievre, line 225: I'm surprised that you need to fix this cache issue in the Add handler for a bug that happens in the Update tests 😅
But since you said the bug only occured in the behat tests I guess it's because the cache is set at first when the product is added, still I wonder if this should be applied in the update handlers as well to be sure all the data are always valid

jolelievre (approved):
I have doubt about the place where the cache is cleared, but we can improve this later if needed

matthieu-rolland (approved):

SD1982:
LGTM the linked issue seems to be correctly fixed !! Thanks @sowbiba !

jolelievre:
Seems like issue bot is trolling us ^^

PR #19648, opened 2020-06-09
Number extractor service

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | A service to extract Number() object from any given object or array. It uses symfony property accessor inside to reach the properties by passing path as string. We need to use Number in many calculations, this should save some time on the convertion of ObjectModel floats to Number().
| Type?         |  refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | There is nothing to test in UI. Added couple unit tests that should pass with CI

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19648)

### DIFF
--- a/src/Core/Util/Number/NumberExtractor.php
+++ b/src/Core/Util/Number/NumberExtractor.php
@@ -1,0 +1,112 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+declare(strict_types=1);
+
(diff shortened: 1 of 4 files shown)


src/Core/Util/Number/NumberExtractor.php
```diff
@@ -0,0 +1,109 @@
...
+namespace PrestaShop\PrestaShop\Core\Util\Number;
+
+use PrestaShop\Decimal\Number;
+use Symfony\Component\PropertyAccess\Exception\AccessException;
+use Symfony\Component\PropertyAccess\Exception\InvalidArgumentException;
+use Symfony\Component\PropertyAccess\Exception\UnexpectedTypeException;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+
+/**
+ * Extracts numeric value as @var Number from given resource
```
matks, line 39: A little example would be great inside this phpDoc 😄

src/Core/Util/Number/NumberExtractor.php
```diff
@@ -0,0 +1,123 @@
...
+     */
+    private $propertyAccessor;
+
+    public function __construct(PropertyAccessorInterface $propertyAccessor)
+    {
+        $this->propertyAccessor = $propertyAccessor;
+    }
+
+    /**
+     * If provided resource is array, access its values using brackets e.g. '[one_property][another_level_property]'
```
jolelievre, line 54: I thought the property accessor can use both syntax indiferently? Like `one_property.another_level_property` can work on an array no?

zuk3975: I thought so too, but tests shown that it must be accessed with brackets when using array :man_shrugging:

jolelievre (approved):
PHPDoc might be improved, but can be done later ;)

src/PrestaShopBundle/Resources/config/services/core/util/number.yml
```diff
@@ -0,0 +1,8 @@
+services:
+  _defaults:
+    public: true
+
+  prestashop.core.util.number.number_extractor:
+    class: 'PrestaShop\PrestaShop\Core\Util\Number\NumberExtractor'
+    arguments:
+      '@property_accessor'
```
jolelievre: Oh that's why travis fails!
```suggestion
      - '@property_accessor'
```

jolelievre (changes requested):

jolelievre (approved):

zuk3975:
rr travis

PierreRambaud (approved):

matks:
Thank you @zuk3975

PR #19653, opened 2020-06-09
Functional tests - Switch tests to playwright (instead of puppeteer)

boubkerbribri:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Switching node library that controls browser to playwright instead of pupeteer
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | `BROWSER='chromium' URL_FO=shopUrl npm run functional tests`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19653)


tests/UI/.docker/Dockerfile
```diff
@@ -0,0 +1,51 @@
+# A minimal Docker image with Node and playwright
```
boubkerbribri, line 1: @PierreRambaud need your review here

.travis.yml
```diff
@@ -72,6 +72,7 @@ before_install:
   # Apache & php-fpm configuration
   - bash travis-scripts/setup-php-fpm.sh
   - bash travis-scripts/setup-apache.sh
+  - bash travis-scripts/setup-node-js.sh
```
boubkerbribri, line 75: Need to install nodejs 10 to run with playwright, I know we already did that for develop, but we need it now for 1.7.7.x.
@PierreRambaud @NeOMakinG WDYT ?

NeOMakinG: I think this is fine because 1.7.7.x already support node 10, but what about documenting it? does it support node 12-13?

boubkerbribri: yes it support node 10 and greater.

tests/UI/.docker/Dockerfile
PierreRambaud: LGTM I don't see something strange :sweat_smile:

PierreRambaud: Maybe the Playwright version can be a parameter

boubkerbribri: I hope that for next playwright version, that we don't need to force install browsers and no need for version to be set

tests/UI/pages/BO/orders/creditSlips/index.js
```diff
@@ -99,22 +99,58 @@ module.exports = class CreditSlips extends BOBasePage {
...
+    const [download] = await Promise.all([
+      this.page.waitForEvent('download'), // wait for download to start
+      this.page.click(this.creditSlipDownloadButton(lineNumber)),
+    ]);
+    return download.path();
   }
 
   /**
-   * Generate PDF by date
+   * Generate PDF by date and download
```
PierreRambaud:
```suggestion
   * Generate PDF by date and download it
```

tests/UI/campaigns/tools/linkchecker.js
```diff
@@ -0,0 +1,143 @@
+require('module-alias/register');
+
+const {expect} = require('chai');
+
+// Import utils
+const helper = require('@utils/helpers');
+const files = require('@utils/files');
+const urlsList = require('@tools/urls.js');
+
+const LOG_PASSED = process.env.LOG_PASSED || false;
```
PierreRambaud: Why not this for a boolean value??
```suggestion
const LOG_PASSED = !!process.env.LOG_PASSED;
```

tests/UI/campaigns/utils/helpers.js
```diff
@@ -0,0 +1,84 @@
...
+      return (await playwright[global.BROWSER.name].launch(browserConfig));
+    } catch (e) {
+      if (attempt <= 3) {
+        await (new Promise(resolve => setTimeout(resolve, 5000)));
+        return this.createBrowser(attempt + 1);
+      }
+      throw new Error(e);
+    }
+  },
+  /**
```
PierreRambaud, line 36:
```suggestion

  /**
```

PierreRambaud (approved):

Progi1984 (approved):

SimonGrn:
Yay, thanks guys !

PR #19657, opened 2020-06-09
Deprecate hookcount and add missing hook in hook.xml

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | The function `hookcount` returns 1 even when no module subscribed to the hook when the hook is present in `hook.xml`. It also does not take into account programmatically subscribed hooks. Therefore, this function should be deprecated and we should look for the content returned by `renderhook` or `renderhookarray` instead.<br>Also, the `displayAdminProductsExtra` hook was missing form `hook.xml`.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | yes
| Fixed ticket? | N/A
| How to test?  | 1. Go on the order view page, no empty rectangle below the tabs (status, transport...) should be displayed.<br>2. Go on the product view page, no "Module" tab should be displayed.<br>3. Install @Matt75's [module](https://github.com/Matt75/testhookdisplayadminproductsextra/releases), go back on the product view page, you should see the "Module" tab.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19657)

### DIFF
--- a/src/PrestaShopBundle/Resources/views/Admin/Product/ProductPage/Blocks/tabs.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Product/ProductPage/Blocks/tabs.html.twig
@@ -34,7 +34,7 @@
     <li id="tab_step2" class="nav-item"><a href="#step2" role="tab" data-toggle="tab" class="nav-link">{{ 'Pricing'|trans({}, 'Admin.Catalog.Feature') }}</a></li>
     <li id="tab_step5" class="nav-item"><a href="#step5" role="tab" data-toggle="tab" class="nav-link">{{ 'SEO'|trans({}, 'Admin.Catalog.Feature') }}</a></li>
     <li id="tab_step6" class="nav-item"><a href="#step6" role="tab" data-toggle="tab" class="nav-link">{{ 'Options'|trans({}, 'Admin.Global') }}</a></li>
-    {% if hookcount('displayAdminProductsExtra') > 0 %}
+    {% if hooksarraycontent(hooks) is not empty %}
       <li id="tab_hooks" class="nav-item"><a href="#hooks" role="tab" data-toggle="tab" class="nav-link">{{ 'Modules'|trans({}, 'Admin.Catalog.Feature') }}</a></li>
     {% endif %}
   </ul>
--- a/src/PrestaShopBundle/Resources/views/Admin/Product/ProductPage/product.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Product/ProductPage/product.html.twig
@@ -24,9 +24,9 @@
  *#}
 {% extends '@PrestaShop/Admin/layout.html.twig' %}
 
-{% set hooks = renderhooksarray('displayAdminProductsExtra', { 'id_product': id_product }) %}
-
 {% block content %}
+
+  {% set hooks = renderhooksarray('displayAdminProductsExtra', { 'id_product': id_product }) %}
 
   <form name="form" id="form" method="post" class="form-horizontal product-page row justify-content-md-center" novalidate="novalidate">
 
@@ -53,7 +53,7 @@
     <div id="form-loading" class="col-xxl-10">
       {# FORM TABS CONTAINER #}
       {% block product_tabs_container %}
-        {{ include('@Product/ProductPage/Blocks/tabs.html.twig') }}
+        {{ include('@Product/ProductPage/Blocks/tabs.html.twig', { 'hooks': hooks }) }}
       {% endblock %}
       <div id="form_content" class="tab-content">
 
--- a/src/PrestaShopBundle/Resources/views/Admin/Product/ProductPage/product.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Product/ProductPage/product.html.twig
@@ -24,6 +24,8 @@
  *#}
 {% extends '@PrestaShop/Admin/layout.html.twig' %}
 
+{% set hooks = renderhooksarray('displayAdminProductsExtra', { 'id_product': id_product }) %}
+
 {% block content %}
 
   <form name="form" id="form" method="post" class="form-horizontal product-page row justify-content-md-center" novalidate="novalidate">
@@ -150,7 +152,7 @@
 
         {# PANEL HOOKED MODULES #}
         {% block product_panel_modules %}
-          {% if hookcount('displayAdminProductsExtra') > 0 %}
+          {% if hooksarraycontent(hooks) is not empty %}
             <div role="tabpanel" class="form-contenttab tab-pane" id="hooks">
               <div class="row">
                 <div class="col-md-12">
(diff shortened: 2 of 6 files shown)


src/PrestaShopBundle/Twig/HookExtension.php
```diff
@@ -165,16 +167,43 @@ public function renderHook($hookName, array $hookParameters = [])
...
      *
+     * @deprecated since 1.7.7.0
+     *
      * @param string $hookName
      *
      * @return number the listeners count that will respond to the hook name
      */
     public function hookCount($hookName)
     {
+        @trigger_error('The ' . __METHOD__ . ' method is deprecated since version 1.7.7.0.', E_USER_DEPRECATED);
```
Progi1984, line 204:
```suggestion
        @trigger_error('The ' . __METHOD__ . ' method is deprecated since version 1.7.7.0. Please use hooksArrayContent.', E_USER_DEPRECATED);
```

src/PrestaShopBundle/Twig/HookExtension.php
```diff
@@ -165,16 +167,43 @@ public function renderHook($hookName, array $hookParameters = [])
...
+     *
+     * @return string
+     */
+    public function hooksArrayContent($hooksArray)
+    {
+        $content = '';
+        if (!is_array($hooksArray)) {
+            return $content;
+        }
+
```
Progi1984:
```suggestion
        if (!is_array($hooksArray)) {
            return '';
        }

        $content = '';
```

src/PrestaShopBundle/Twig/HookExtension.php
```diff
@@ -165,16 +167,43 @@ public function renderHook($hookName, array $hookParameters = [])
...
+        foreach ($hooksArray as $hook) {
+            if (isset($hook['content'])) {
+                $content .= $hook['content'];
+            }
```
Progi1984:
```suggestion
            $content .= $hook['content'] ?? '';
```

Progi1984 (commented):
Added a small optimization

src/PrestaShopBundle/Twig/HookExtension.php
atomiix: I thought about it but `hooksArrayContent` is not aim to be a replacement of the `hookCount` function. In our case we can use `hooksArrayContent` for the same result but it's probably not always the case. WDYT?

src/PrestaShopBundle/Resources/views/Admin/Product/ProductPage/product.html.twig
```diff
@@ -24,6 +24,8 @@
  *#}
 {% extends '@PrestaShop/Admin/layout.html.twig' %}
 
+{% set hooks = renderhooksarray('displayAdminProductsExtra', { 'id_product': id_product }) %}
```
PierreRambaud: Should be inside the content block in case someone already overrides this template.

src/PrestaShopBundle/Resources/views/Admin/Product/ProductPage/Blocks/tabs.html.twig
```diff
@@ -34,7 +34,7 @@
     <li id="tab_step2" class="nav-item"><a href="#step2" role="tab" data-toggle="tab" class="nav-link">{{ 'Pricing'|trans({}, 'Admin.Catalog.Feature') }}</a></li>
     <li id="tab_step5" class="nav-item"><a href="#step5" role="tab" data-toggle="tab" class="nav-link">{{ 'SEO'|trans({}, 'Admin.Catalog.Feature') }}</a></li>
     <li id="tab_step6" class="nav-item"><a href="#step6" role="tab" data-toggle="tab" class="nav-link">{{ 'Options'|trans({}, 'Admin.Global') }}</a></li>
-    {% if hookcount('displayAdminProductsExtra') > 0 %}
+    {% if hooksarraycontent(hooks) is not empty %}
```
PierreRambaud, line 37: What if this template was used before, `hooks` is not defined in this one :thinking: 
Be sure the variable is properly passed to this template

src/PrestaShopBundle/Twig/HookExtension.php
```diff
@@ -165,16 +167,43 @@ public function renderHook($hookName, array $hookParameters = [])
...
+     * Return the concatenated content of a renderHooksArray response
+     *
+     * @param array $hooksArray the array returned by the renderHooksArray function
+     *
+     * @return string
+     */
+    public function hooksArrayContent($hooksArray)
+    {
+        $content = '';
+        if (!is_array($hooksArray)) {
```
PierreRambaud, line 179: Are you sure this variable can not be a string with HTML content?

atomiix: @PierreRambaud this function aims to be used with the result of `renderhooksarray`, not `renderhook` (which, as you said, contains HTML content).

Progi1984 (approved):

PierreRambaud (approved):

Progi1984:
Thanks @atomiix

matks:
Hi @atomiix ! I'm sorry I'm late on this topic.

I'm afraid we might have done something wrong here 🤔 

The aim of function `hookCount()` is to provide the information "how many modules are plugged on this hook ?" **WITHOUT** triggering the hook itself.

So you could do this:
```
if (hookCount($myHook) > 0) {
    $result = Hook::exec($myHook);
}
```

Now that you have deprecated it, we have no other ways to retrieve this information without triggering the hook. If we do
```
if (count(renderhookarray($myHook)) > 0) {
    $result = Hook::exec($myHook);
}
```
then the hook `$myHook` will be triggered 2 times, which is a big change in behavior.

I'm afraid we need to provide an alternative way to retrieve this information.

Can you tell me if I'm right ?

atomiix:
Hey @matks!
>The aim of function hookCount() is to provide the information "how many modules are plugged on this hook ?" WITHOUT triggering the hook itself.

You are right! **BUT** you don't have necessary to trigger 2 times hooks you can store the result of the hooks, check if they are empty and if not, display the content. You can see what I did [here](https://github.com/PrestaShop/PrestaShop/pull/19657/files#diff-1a42492ba7a3b1fc9477d3ad77b8edb5R114-R125) and [here](https://github.com/PrestaShop/PrestaShop/pull/19657/files#diff-6e62c7f1156567412fdf8a98e7976598R29-R155).

matks:
Be careful about that 😉 it introduced a small regression https://github.com/PrestaShop/PrestaShop/pull/19820

So indeed it is avoidable to call the hooks 2 times ... but we still cannot know whether there's modules plugged on the hook without firing the event at least once ?

atomiix:
>but we still cannot know whether there's modules plugged on the hook without firing the event at least once ?

That's right. But as this function wasn't working properly in the first place and as we can have a workaround, I think we can deprecate it. (Also, I'm pretty sure that's what @eternoendless thought too)

Issue #19595, opened 2020-06-05
[EPIC] Errors when adding a product with minimum quantity in order page


MatShir:
The issues below are grouped in an epic because it seems they have to be solved together.

## BUGS

- [x] #18568 - BO - Order page - An exception is displayed when adding a product with a min quantity
- [x] #18695 - A product allowed to be ordered out of stock with a minimum quantity cannot be added to the BO order.
- [x] #19607 - The minimum quantity is not respected when I add a product 
- [x] #21525 - BO - Order page - An error 500 is displayed when adding a product with a min quantity

zuk3975:
There was a discussion about how to handle scenario when merchant adds product with a custom price to order in which  identical product already exists.
So:
1. Our main goal is to merge identical product rows (maybe in 1.7.8 or later, but not for 177)
2. That means we cannot handle different prices for same product

The solution for 1.7.7:
We check if order already has identical product. If it contains identical product, we don't allow to change product price anymore, but leave the existing product price instead. (in future we will want only to increase the quantity of existing product)
And if merchant updates the product price in the list, all of identical product prices should change.

@MatShir please correct me if I'm wrong

marionf:
@MatShir The 3 issues are fixed, we can close this epic :tada:

khouloudbelguith:
Hi @MatShir,

we have an issue related to this Epic reported here: https://github.com/PrestaShop/PrestaShop/issues/21525.
Can we re-open this Epic?

Thanks!

MatShir:
No need, the issue will be quickly closed for the 1.7.7 build.

Issue #19490, opened 2020-06-01
Wrong price displayed before adding a product on a order


MatShir:
### Describe the bug

When I add the same product in the order that was previously discounted, the price of the product on the add interface displays the new price, but when it is added, it displays the old price. 

![image](https://user-images.githubusercontent.com/43613217/83390576-60d1d980-a3f2-11ea-8779-bf5114209aef.png)

**Displays price on the add interface:**
35.9 € tax excl 
43.08 € tax incl

**Price displayed after the add:**
28.72 € tax excl 
34.46 € tax incl

### Expected behavior

When I add the same product in the order that was previously discounted, the price of the product on the add interface and after the add action should remains the same (the discounted one) 
The user can edit the price while adding the product, but in this case, a warning modal will be displayed the previous prices will be updated

Displays price on the add interface:
28.72 € tax excl 
34.46 € tax incl

The price displayed after the add:
28.72 € tax excl 
34.46 € tax incl

#### Steps to Reproduce

1. Product A at 20 Euro
2. Add the product to an order
3. edit the price to 10 Euro
4. Add the same product again 

price display add interface =/= price edited in the order

See the add interface and the product price before and after

**Screenshots**

If applicable, add screenshots or screen recordings to help explain your problem.

#### Additional information

* PrestaShop version: N/A
* PHP version: N/A

colinegin:
Calculation is correct, the display is correct after you save the product.
The display is incorrect in the form.

colinegin:
Hello,

what's the status of this issue ?

colinegin:
Specs to be done today by @MatShir

MatShir:
This issue should be fixed by https://github.com/PrestaShop/PrestaShop/issues/20010.
To be reproduced

marionf:
I removed this issue of the kanban to keep only the EPIC: https://github.com/PrestaShop/PrestaShop/issues/20161 as the same PR fixes all the issues, this way we keep a better readability of the kanban

PR #19698, opened 2020-06-10
Fix order view total discounts

### LINKED ISSUE #18695: A product allowed to be ordered out of stock with a minimum quantity cannot be added to the BO order.
Be aware of the EPIC - Errors when adding a product with minimum quantity in order page #19595

**Describe the bug**
A product allowed be order when it is out of stock and has required minimum quantity can't be added to order even when the required minimum is respected

**Expected behavior**

We have an incorrect alert when the product is out of stock Following these steps:

**Steps to reproduced**

1. Create a productA with Quantity=0 and minimal quantity=10 and Allow orders
2. Go to `admin-dev/sell/orders/orders/{orderId}/view
3. Make sure order status is not "delivered" and you can add products.
4. Click Add product, search and select the last product created productA with quantity=1
5. check alert => NOK (Error! Product with id "22" is out of stock.)
 => it is not out of stock => it should be displayed `Error! Minimum quantity of "10" must be added`
6. Click Add product, search and select the last product created productA with quantity=10
it displays:"
Error!
An unexpected error occurred. [PrestaShopDatabaseException code 0]: Duplicate entry '19-0-2-0-0000-00-00 00:00:00-0000-00-00 00:00:00-0-0-1-0-0-1-0' for key 'id_product_2'", instead of just adding the product in the order

https://drive.google.com/file/d/1B-vnVzqsNDxPNRTYuTvgl3DLjl7r44Ii/view?usp=sharing

Thanks!

_Originally posted by @khouloudbelguith in https://github.com/PrestaShop/PrestaShop/pull/18614#issuecomment-613537868_

### LINKED ISSUE #19184: BO - Order detail - Voucher is displayed multiple time

#### Describe the bug

When creating an order from the BO if a cart rule authorizing several uses per user is added and then we add products to this order from the BO a new line of vouchers is added for each product added

#### Expected behavior

For one voucher we should have only one display

#### Steps to Reproduce

1. Go to BO => Orders => Orders
2. Click on "Add new order"
3. Select a customer for the order
4. Add a product to the order
5. Click on "add new voucher"
6. In the voucher creation modal create a voucher with the following parameter : 
capture d'écran_13
capture d'écran_14
capture d'écran_15
7. When the voucher is created add it to the order
8. Then click on "Create the order"
9. In the Order detail click on the "add a product" and add a product
10. See error
capture d'écran_16

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: 7.2

### LINKED ISSUE #19591: Adding a second product in the order does not update correctly the cart list

#### Describe the bug

I create a cart rule on the product Brown bear at 9 euro with an automated discount of 50% on the total of the order.

On the back-office, when I add product the product to the order it apply a 50% reduction in the total of the order. But when I add again the same product,  the discount is updated but it keeps the old discount line on the cart rule list.

#### Expected behavior

The cart rule line should be updated

#### Steps to Reproduce

1. Create a cart rule of 50% reduction on the total of the order with **no code** specifically applied to a product
2. Add the product to any order
3. Add the product a second time

**Screenshots**

2 discount same product

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: N/A

### LINKED ISSUE #19609: Discount total not display on tax excl order
followed the steps and I see different behavior.

#### Describe the bug

When I create an order with a customer that displays the price as tax excluded and I add a discount of 10% on the total of the order, the discount total in summary disappears. The discount also does not update, the value displayed on the cart rule is incorrect.

#### Expected behavior
The discount should be updated when the product is added with the right display on the total summary.

#### Steps to Reproduce

1. Edit customer group, to displays price with tax excluded
2. Click on "Add new order"
3. Select a customer for the order from the customer group
4. Add a product to the order
5. Click on "add new voucher"
6. In the voucher creation modal create a voucher with the following parameter : 
capture d'écran_13
capture d'écran_14
capture d'écran_15
7. When the voucher is created add it to the order
8. Then click on "Create the order"
9. In the Order detail click on the "add a product" and add a product
10. See error
Screenshot from 2020-05-21 14-47-04

_Originally posted by @zuk3975 in https://github.com/PrestaShop/PrestaShop/issues/19184#issuecomment-632043396_

zuk3975:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Following changes were applied: Update existing vouchers instead of creating a new one each time new product is added. From technical point of view - removed differentiating $oldCartRules from $newCartRules and instead of creating new cart rule for each difference, I update existing OrderCartRule with a $newCartRule info. Also deleted restoring whole order specific prices, which I hope is not necessary for further order calculations. Need to be tested well.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #19184 #19591 #19609 #18695
| How to test?  | Create order from back office with at least one product and using some voucher, then add one more product to that order in order view page. Discount should be recalculated instead of duplicating. See more in issue #19184. If it's possible, please check following issues (I couldn't reproduce them anymore, this PR might have fixed them too, even tho it wasn't the purpose): #19591, #19609<br>Also try to remove (NOT CANCEL) some product associated to discount => the discount associated to the removed product should disappear BUT the general discounts (not restricted to product but the whole shop or discount code) should still be present and well computed

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19698)

### DIFF
--- a/admin-dev/themes/new-theme/js/pages/order/view/order-prices-refresher.js
+++ b/admin-dev/themes/new-theme/js/pages/order/view/order-prices-refresher.js
@@ -37,7 +37,7 @@ export default class OrderPricesRefresher {
     $.ajax(this.router.generate('admin_orders_get_prices', {orderId}))
       .then((response) => {
         $(OrderViewPageMap.orderTotal).text(response.orderTotalFormatted);
-        $(OrderViewPageMap.orderDiscountsTotal).text(response.discountsAmountFormatted);
+        $(OrderViewPageMap.orderDiscountsTotal).text(`-${response.discountsAmountFormatted}`);
         $(OrderViewPageMap.orderDiscountsTotalContainer).toggleClass('d-none', !response.discountsAmountDisplayed);
         $(OrderViewPageMap.orderProductsTotal).text(response.productsTotalFormatted);
         $(OrderViewPageMap.orderShippingTotal).text(response.shippingTotalFormatted);
--- a/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
@@ -126,21 +126,20 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
 
             $cart = $this->createNewOrEditExistingCart($order);
 
-            $specificPrice = $this->createSpecificPriceIfNeeded(
-                new AmountImmutable(
-                    (float) (string) $command->getProductPriceTaxIncluded(),
-                    (float) (string) $command->getProductPriceTaxExcluded()
-                ),
-                $order,
-                $cart,
-                $product,
-                $combination
-            );
-
             // Restore any specific prices for the products in the order
             $restoredSpecificPrices = $this->restoreOrderProductsSpecificPrices(
                 $order,
                 $cart
+            );
+
+            // Add specific price for the product being added
+            $specificPrice = $this->createSpecificPriceIfNeeded(
+                $command->getProductPriceTaxIncluded(),
+                $command->getProductPriceTaxExcluded(),
+                $order,
+                $cart,
+                $product,
+                $combination
             );
 
             $this->addProductToCart($cart, $product, $combination, $command->getProductQuantity());
@@ -364,7 +363,8 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
             $product = new Product((int) $orderDetail->product_id);
 
             $restoredSpecificPrices[] = $this->createSpecificPriceIfNeeded(
-                new AmountImmutable($orderDetail->unit_price_tax_incl, $orderDetail->unit_price_tax_excl),
+                new Number((string) $orderDetail->unit_price_tax_incl),
+                new Number((string) $orderDetail->unit_price_tax_excl),
                 $order,
                 $cart,
                 $product,
@@ -376,7 +376,8 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
     }
 
     /**
-     * @param AmountImmutable $amount
+     * @param Number $priceTaxIncluded
+     * @param Number $priceTaxExcluded
      * @param Order $order
      * @param Cart $cart
      * @param Product $product
@@ -385,27 +386,15 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
      * @return SpecificPrice|null
      */
     private function createSpecificPriceIfNeeded(
-        AmountImmutable $amount,
+        Number $priceTaxIncluded,
+        Number $priceTaxExcluded,
         Order $order,
         Cart $cart,
         Product $product,
         $combination
     ) {
-        $initialProductPriceTaxIncl = Product::getPriceStatic(
-            $product->id,
-            true,
-            $combination ? $combination->id : null,
-            2,
-            null,
-            false,
-            true,
-            1,
-            false,
-            $order->id_customer,
-            $cart->id,
-            $order->{Configuration::get('PS_TAX_ADDRESS_TYPE', null, null, $order->id_shop)}
-        );
-
+        // Check it the SpecificPrice has already been added by restoreOrderProductsSpecificPrices, if yes ignore new
+        // price because the first one is kept
         if (SpecificPrice::exists(
             $product->id,
             $combination ? $combination->id : 0,
@@ -421,8 +410,22 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
             return null;
         }
 
-        if (!(new Number((string) $amount->getTaxIncluded()))->equals(new Number((string) $initialProductPriceTaxIncl))) {
-            // @todo: use private method to create specific price object
+        $initialProductPriceTaxIncl = Product::getPriceStatic(
+            $product->id,
+            true,
+            $combination ? $combination->id : null,
+            2,
+            null,
+            false,
+            true,
+            1,
+            false,
+            $order->id_customer,
+            $cart->id,
+            $order->{Configuration::get('PS_TAX_ADDRESS_TYPE', null, null, $order->id_shop)}
+        );
+
+        if (!$priceTaxIncluded->equals(new Number((string) $initialProductPriceTaxIncl))) {
             $specificPrice = new SpecificPrice();
             $specificPrice->id_shop = 0;
             $specificPrice->id_cart = 0;
@@ -433,11 +436,11 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
             $specificPrice->id_customer = $order->id_customer;
             $specificPrice->id_product = $product->id;
             $specificPrice->id_product_attribute = $combination ? $combination->id : 0;
-            $specificPrice->price = new Number((string) $amount->getTaxExcluded());
+            $specificPrice->price = (float) (string) $priceTaxExcluded;
             $specificPrice->from_quantity = 1;
             $specificPrice->reduction = 0;
             $specificPrice->reduction_type = 'amount';
-            $specificPrice->reduction_tax = 0;
+            $specificPrice->reduction_tax = !$priceTaxIncluded->equals($priceTaxExcluded);
             $specificPrice->from = '0000-00-00 00:00:00';
             $specificPrice->to = '0000-00-00 00:00:00';
             $specificPrice->add();
--- a/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
@@ -52,7 +52,6 @@ use PrestaShop\PrestaShop\Core\Domain\Order\Product\Command\AddProductToOrderCom
 use PrestaShop\PrestaShop\Core\Domain\Order\Product\CommandHandler\AddProductToOrderHandlerInterface;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductOutOfStockException;
 use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
-use PrestaShop\PrestaShop\Core\Localization\CLDR\ComputingPrecision;
 use PrestaShop\PrestaShop\Core\Util\DateTime\DateTime;
 use Product;
 use Shop;
@@ -736,18 +735,5 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
         if (!$product->checkQty($command->getProductQuantity())) {
             throw new ProductOutOfStockException(sprintf('Product with id "%s" is out of stock, thus cannot be added to cart', $product->id));
         }
-    }
-
-    /**
-     * @param Cart $cart
-     *
-     * @return int
-     */
-    private function getPrecisionFromCart(Cart $cart): int
-    {
-        $computingPrecision = new ComputingPrecision();
-        $currency = new Currency((int) $cart->id_currency);
-
-        return $computingPrecision->getPrecision($currency->precision);
     }
 }
--- a/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
@@ -54,6 +54,7 @@ use PrestaShop\PrestaShop\Core\Domain\Order\Product\CommandHandler\AddProductToO
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductOutOfStockException;
 use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
 use PrestaShop\PrestaShop\Core\Localization\CLDR\ComputingPrecision;
+use PrestaShop\PrestaShop\Core\Util\DateTime\DateTime;
 use Product;
 use Shop;
 use SpecificPrice;
@@ -136,6 +137,12 @@ final class AddProductToOrderHandler extends AbstractOrderHandler implements Add
                 $combination
             );
 
+            // Restore any specific prices for the products in the order
+            $restoredSpecificPrices = $this->restoreOrderProductsSpecificPrices(
+                $order,
+                $cart
+            );
+
             $this->addProductToCart($cart, $product, $combination, $command->getProductQuantity());
 
             // Fetch Cart Product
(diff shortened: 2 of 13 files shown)


src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -138,6 +136,7 @@ public function handle(AddProductToOrderCommand $command)
                 $combination
             );
 
+            //@TODO: what is this? it causes duplicate entry because $specificPrice above is created identical.
```
zuk3975: @sowbiba I thought you should know the answer as you've spent some time with this, could you help?
I've done the following- after each product addition, it updates existing OrdercartRules instead of piling up new ones, which caused the duplication.

And after that (Im not sure if it wasn't happening before) I get sql error of duplicate entry if i try to add same product once again, which is caused precisely by that $restoredSpecificPrices, because $specificPrice above is already created identical. :exploding_head:

zuk3975: I've simply added a check if identical specific price doesn't  exist before creating new one, please tell me you have any concerns about that.

admin-dev/themes/new-theme/js/pages/order/view/order-prices-refresher.js
```diff
@@ -37,7 +37,7 @@ export default class OrderPricesRefresher {
     $.ajax(this.router.generate('admin_orders_get_prices', {orderId}))
       .then((response) => {
         $(OrderViewPageMap.orderTotal).text(response.orderTotalFormatted);
-        $(OrderViewPageMap.orderDiscountsTotal).text(response.discountsAmountFormatted);
+        $(OrderViewPageMap.orderDiscountsTotal).text(`-${response.discountsAmountFormatted}`);
```
jolelievre, line 40: This should be managed in the query that returns `discountsAmountFormatted` because CLDR manages some negative pattern for some languages
This could be done on the front side but since it's already a "formatted" value it should be easier to manage this on server side

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -182,33 +181,29 @@ public function handle(AddProductToOrderCommand $command)
...
-                $orderCartRule->id_order_invoice = !empty($invoice->id) ? $invoice->id : 0;
-                $orderCartRule->name = $cartRule['name'];
-                $orderCartRule->value = $values['tax_incl'];
-                $orderCartRule->value_tax_excl = $values['tax_excl'];
-                $orderCartRule->add();
+                    continue;
+                }
+
+                foreach ($order->getCartRules() as $orderCartRule) {
+                    if ($orderCartRule['id_cart_rule'] !== $cartRuleInArray['id_cart_rule']) {
```
jolelievre: This double loop feels weird, I have the feeling it could create too many OrderCartRule, especially if they are not in the same order. Example:

```php
$cartRuleInArray = [
  ['id_cart_rule' => 1],
  ['id_cart_rule' => 2],
];

$order->getCartRules() = [
  ['id_cart_rule' => 2],
  ['id_cart_rule' => 1],
];
```

With the current implementation it will believe there is no match because it doesn't loop through the whole list to find the matching id.

I trie something similar but slightly different (I also find it easier to read/understand):

```php
        foreach ($cartRules as $cartRule) {
                $rule = new CartRule($cartRule['id_cart_rule']);

                // Search order cart rule
                $orderCartRule = null;
                foreach ($order->getCartRules() as $orderCartRule) {
                    if ($orderCartRule['id_cart_rule'] === $cartRule['id_cart_rule']) {
                        $orderCartRule = new OrderCartRule($orderCartRule['id_order_cart_rule']);

                        break;
                    }
                }

                if (null === $orderCartRule) {
                    $orderCartRule = new OrderCartRule();
                }
                $orderCartRule->id_order = $order->id;
                $orderCartRule->id_cart_rule = $cartRule['id_cart_rule'];
                $orderCartRule->id_order_invoice = !empty($invoice->id) ? $invoice->id : 0;
                $orderCartRule->name = $cartRule['name'];
                $orderCartRule->value = $rule->getContextualValue(true);
                $orderCartRule->value_tax_excl = $rule->getContextualValue(false);
                $orderCartRule->save();
            }
```

admin-dev/themes/new-theme/js/pages/order/view/order-prices-refresher.js
zuk3975: Ok I can change it. But idk if its really worth it. What if we want to use the formatted value without '-' ? Its possible usecase, it might be even used somewhere already :man_shrugging:

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
zuk3975: agree :+1:

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -398,6 +406,21 @@ private function createSpecificPriceIfNeeded(
             $order->{Configuration::get('PS_TAX_ADDRESS_TYPE', null, null, $order->id_shop)}
         );
 
+        if (SpecificPrice::exists(
```
jolelievre, line 414: What if specific price already existed but we want to change it? 🤔

zuk3975: @sowbiba maybe you can help with this topic?

zuk3975: with https://github.com/PrestaShop/PrestaShop/pull/19698/commits/7147d973b2ad07e075bd1e2ee2406f31e9490b7f I've removed "restoring whole order specific prices". It looks like its working well, but as @jolelievre said, it might be necessary for some upcoming order computations.
So idk, @sowbiba you are the one to tell us what could go wrong :sweat_smile:

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
zuk3975: https://github.com/PrestaShop/PrestaShop/pull/19698#discussion_r438746132

jolelievre:
Note: after different tests this PR fixes several issues: #19184 #19591 #19609

Only one remaining bug:
With a cart rule on whole order (10% for example), when I remove one product from the order the discounts summary is still correct (and updated) but the list of discounts is now empty (idea: the cart rule is still applied but the order cart rule has been removed, or vice versa)

This bug should be fixed in this PR.


Regarding the #19490 it only fixes partially (or at least force a behaviour): when a product with a specific price is added to an order and I add a second product the second product uses the same price as the first one (regardless of what has been input)

Only two remaining drawbacks:
- the default values in the add product row don't use the specific prices
- when a product was already in the order (without specific price) and I the same product with a specific price the second one is changed but the first one keeps its original price (@MatShir to be defined: should we update the previous prices in this case?)

The missing part for 19490 shouldn't be done in this PR

matks:
@jolelievre 

> This bug should be fixed in this PR.

Do you mean "in another PR" ?

jolelievre:
Hi @matks 

For the bug of "remove a product" I meant **this PR** I just pushed the fix, this allowed to have a more precise behat test

I added the same kind of test for "cancel product" but I had to comment many assertions in the gerkin file because it doesn't manage discounts recomputing **AT ALL** But the tests are here waiting to be uncommented Since there are lots of thing to add though it's better to fix them in another PR especially since there is already an issue related to cancel product issues https://github.com/PrestaShop/PrestaShop/issues/19717

And for specific prices remaining issues they should be managed in a different issue and PR

admin-dev/themes/new-theme/js/pages/order/view/order-prices-refresher.js
PierreRambaud: We also have a cldr js library in the BO to do the job :wink:

matks (dismissed):
Code seems correct. As for the logic, the code reflects the high complexity of both the computations being in place and the way it is stored in the database 😢 

I approve this PR but I'd like us at some point in the future to improve all these crazy computations we do (see comment below)

matks:
With the migration of the "Create an order" page and the "View an order" page we have discovered a HUGE quantity of code doing complex computations for Orders and Carts. We have discovered crazy logic related to discounts, cart rules, taxes, shipping costs, taxes, rounding ...

This part was painful and hurt our brain but now we understand it better and we have a ever-growing set of Behat tests to help us make sure we don't have any more regressions.

I think we are ready for the next step, to be delivered in 1.7.8 : the refactoring. 🎉 

We will have to discuss how this refactoring takes place. Here is my opinion on the topic, other opinions must be heard as well:

# Refactoring

## Refactor this into DDD-styled statements

I suggest we rewrite all of these code following a DDD approach, and by this I mean that _all the computations and logic should be abstracted_.

All the complexity should happen inside php classes that do not care about ObjectModel, do not care about SQL queries, do not care bout how the data is persisted. It should be "pure" php code, only moving and computing primitives.

This part could be easily unit-tested by the way.

Obviously at some point we need to fetch and persist data. All of these parts should be put into dedicated classes (DataProviders and DataPersistors ?) that do this and only this. This code will interact with either raw SQL statements or ObjectModels.

In theory, if we can decouple the "code that does the computations" enough from the "code that fetch and persist data" ; we could tomorrow rewrite the "code that fetch and persist data" to use an API instead of MySQL.

## Refactor this into stateless and stupid code

I see quite a lot of our code is stateful. In this PR and in other PRs I have seen bugfixes that were done by moving a function call 20 lines above or 20 lines below. The bug was fixed by "calling addSpecificPrices() before addProduct() instead of after" or "refreshing shipping costs before computing the tax"

Too much of all of this code relies on "calling the right function at the right moment". So I suggest we aim for another situation, a situation a lot easier but less optimized.

I'm not sure about this idea because it has a huge CPU cost though 😅 . But at least I can present it.

This strategy consists of **re-computing everything every time**.

What I mean by that is that, _any time we modify something into the Cart, like adding a discount, adding a product, modifying the tax, we should call a function_ `$cart->recomputeEverything()`. So we know that, at any moment, all prices, numbers, whatever, are up-to-date !

It has huge of benefits because you never need to worry whether, after modifying something, you need to "refresh" or "re-compute" something. It's always up-to-date !

The idea comes from JS frameworks like Angular. You modify one model, then the whole page is rendered again, modifications happen in cascade. You dont need to re-render this block, update this widget ... everything that needs to be updated is updated without you needing to choose what. It's very comfortable for the developer.

## Test as deep as possible

We have now a big suite of Behat tests to cover ourselves when modifying deep logic. It's great, but when one of those tests fail, sometimes we need hours to find exactly what is wrong ! 😢 

We should try to test as deep as possible our logic, so that when tests fail, we know with accuracy what is wrong and can debug/fix very fast ! 😄

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
sowbiba: We already talk about it on Slack I think. We need to restore the specific prices. They are deleted after order amounts calculation
We have 
productA which cost 10€
Let's edit that price to put 5€. A specific price is added, the order amounts are computed, and the specific price deleted

Now let's add productB
To compute order amounts, we dont have anymore the specific price for productA.
Before "restoring specific prices", the original price for productA was taken. That was a bug

But if we do that, if you add another line of productA, the price of 5€ will be taken.
If you try to change the specific price, you will have an exception because of the UniqueKey on specific price table.
I explained this to @PrestaShop/prestashop-product-team and they validated it

For the @jolelievre question about "what if we want to change that specific price?"
Today, we can't do that. Maybe @MatShir can help

matthieu-rolland (dismissed):

sowbiba (dismissed):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
colinegin: @MatShir is off this week so I'll try to help. 
I don't know much the topic, so let me know if you want to discuss more in depth about it.
But from what i can read, I would rather choose not to add new features as we are close to the final release. Can you confirm that not being able to change the specific price is not a regression ?

sowbiba: This can be useful
https://github.com/PrestaShop/PrestaShop/issues/19595#issuecomment-642730057

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
```diff
@@ -174,6 +174,28 @@ private function updateOrderCartRules(Order $order, Cart $cart)
...
+            $cartRule = new CartRule($newCartRule['id_cart_rule']);
+
+            $orderCartRule = new OrderCartRule();
+            $orderCartRule->id_order = $order->id;
+            $orderCartRule->id_cart_rule = $newCartRule['id_cart_rule'];
+            $orderCartRule->id_order_invoice = $order->getInvoicesCollection()->getLast();
+            $orderCartRule->name = $newCartRule['name'];
+            $orderCartRule->value = Tools::ps_round($cartRule->getContextualValue(true), $computingPrecision);
+            $orderCartRule->value_tax_excl = Tools::ps_round($cartRule->getContextualValue(false), $computingPrecision);
+            $orderCartRule->save();
```
sowbiba, line 196: I did something similar here https://github.com/PrestaShop/PrestaShop/pull/19646/files#diff-5d60a3a71cc2e386cad2cf1bd26deeb1R430

sowbiba (approved):

matthieu-rolland (approved):

src/Adapter/Order/CommandHandler/DeleteProductFromOrderHandler.php
jolelievre: Yeah I was waiting for the QA approval to warn you I didn't plan to do as much but I ended doing almost the same thing 😅 

Although your implementation is better via the service, so the refacto from your PR is meaningful The good point is that I added behat tests for all those use cases (only for delete not cancel) So you can have fun break and refacto everything with peace of mind 👍

matks:
Thank you @zuk3975

zuk3975:
> Thank you @zuk3975

Thanks to @jolelievre :stuck_out_tongue:

jolelievre:
> > Thank you @zuk3975
> 
> Thanks to @jolelievre 😛

Thank you @matks

PR #19706, opened 2020-06-10
Don't try to translate using domain unless the "d" parameter is present

eternoendless:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | The Smarty translation was trying to translate using the context translator even when the `d` parameter of the `{l}` function wasn't provided. This should provide a small performance improvement when translating third party modules using the legacy translation system and when a wording is not translated. This change only applies to FO.
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | N/A
| How to test?  | Install a module that uses the legacy translation system, like https://github.com/PrestaShopCorp/psgdpr. See that its wordings in FO appear translated as before in any page. See that the rest of the page appears translated as before in any page.

Note: this change makes the behavior of FO closer to BO, where we don't try other translation methods as soon as the `d` parameter is provided: https://github.com/PrestaShop/PrestaShop/blob/1.7.7.x/config/smartyadmin.config.inc.php#L78-L114

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19706)

### DIFF
--- a/config/smartyfront.config.inc.php
+++ b/config/smartyfront.config.inc.php
@@ -175,7 +175,7 @@ function smartyTranslate($params, $smarty)
         $params['d'] = null;
     }
 
-    if (null !== $params['d']) {
+    if (!empty($params['d'])) {
         if (isset($params['tags'])) {
             $backTrace = debug_backtrace();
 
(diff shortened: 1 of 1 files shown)


matks (approved):

jolelievre (approved):

Progi1984:
Thanks @eternoendless

PR #19721, opened 2020-06-11
Add composer update --dry-run roave/security-advisories into CI

ksaandev:
| Questions     | Answers
| Branch?       | develop
| Description?  | `composer update --dry-run roave/security-advisories` is a nice check to make sure no dependencies with known vulnerabilities are used
| Type?         | new feature
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | CI should pass, excluding usecase where not secure dependency is used

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19721)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,6 +87,8 @@ install:
     fi
 
 script:
+  - composer update --dry-run roave/security-advisories
+
   - if [ "$PRESTASHOP_TEST_TYPE" = "unit" ]; then
         bash tests/check_file_syntax.sh;
     fi


ksaandev:
I'm not sure I'm doing it right but here is the idea https://ocramius.github.io/blog/roave-security-advisories-protect-against-composer-packages-with-security-issues/

matks:
Hi ! Nice idea ! But I think you need to first embed the dependency into composer.json THEN you can use the dry-run update

wdyt @PierreRambaud ?

PierreRambaud:
I'm not sure about this one, it needs to be updated each time, can't be in the composer.lock, and GitHub add a beta checker (we will add a GitHub action really soon) to do the same job (and we also have the Symfony dependency checker).

This dependency is also too restrictive and don't forget:
> Please just note that this only works when adding new dependencies or when running composer update: security issues in your composer.lock cannot be checked with this technique

matks:
> Symfony dependency checker

Do we ? I searched for it 🤔 but could not find it
Or maybe not a CI ?

PierreRambaud:
> > Symfony dependency checker
> 
> Do we ? I searched for it  but could not find it
> Or maybe not a CI ?

It's not in the CI, but it's mandatory to release a new version of PrestaShop :)

PierreRambaud:
@matks IMHO, it's not a good idea to block a pull request if dependencies change and give a report telling there are security issues :/

Progi1984:
@PierreRambaud It must be done in the nightly and put in the nightly board as said in retro.

ksaandev:
I am not sure what you want me to do

PierreRambaud:
> I am not sure what you want me to do

Unfortunately nothing, we are not in favor of adding the dependency security check in the CI, it should be done by our nightly checks.
A pull request shouldn't be blocked by sometimes unrelated to the changes. :/

I'm closing your pull request since we don't plan to add this kind of checks in Travis

ksaandev:
Alright, thanks for answers

PR #19730, opened 2020-06-12
Enable fancybox for edit an address in View an Order page

### LINKED ISSUE #19499: Address form should be opened in modal
By the way, the Address form should be opened in modal, not a new page. (same behavior as in order create page), isn't it? (because now it seems to be forgotten. It opens up in a new page)

_Originally posted by @zuk3975 in https://github.com/PrestaShop/PrestaShop/issues/19341#issuecomment-634475159_

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Enable fancybox for edit an address in View an Order page
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/19499
| How to test?  | See ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19730)


src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -397,6 +397,7 @@ public function editAction(int $addressId, Request $request): Response
     public function editOrderAction(int $orderId, string $addressType, Request $request): Response
     {
         try {
+            // @todo: don't rely on Order ObjectModel, use a Adapter DataProvider
```
matks, line 400: Out of the scope of this PR, just a side-note

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -394,9 +394,10 @@ public function editAction(int $addressId, Request $request): Response
      *
      * @return Response
      */
-    public function editOrderAction(int $orderId, string $addressType, Request $request): Response
+    public function editOrderAddressAction(int $orderId, string $addressType, Request $request): Response
```
PululuK, line 397: But, changing the name of this method can introduce BC ? 🤔

PierreRambaud: No, because this controller is not released yet ;)

PierreRambaud (approved):

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
PululuK: Sorry I'm in the future  🚀

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -441,6 +442,13 @@ public function editOrderAction(int $orderId, string $addressType, Request $requ
             if ($result->isSubmitted() && $result->isValid()) {
                 $this->addFlash('success', $this->trans('Update successful', 'Admin.Notifications.Success'));
 
+                if ($request->query->has('submitFormAjax')) {
```
PululuK, line 445: And you need to check if `submitFormAjax` is true  ?

matks: I have not seen a usecase where it was passed as false 🤔 just "it's in the query"

PululuK: What's the use then `submitFormAjax` ?

matks: It's an old mechanism that indicates "hi, this form must be displayed inside a fancybox". This modifies the behavior of the successfull submit: instead of returning HTTP 302 to GET "view form" ; it returns a small javascript that closes the fancybox and do additional operations like passing IDs to window.parent if needed

It's quite dirty but it's how it was done before. Keeping it working allows modules to continue embedding our forms inside fancybox using this `submitFormAjax` parameter in their modules if they want to.

PululuK: Ok ... Thanks for the explanation !

PululuK (approved):

Progi1984 (approved):

Progi1984:
Thanks @matks

Issue #14777, opened 2019-07-19
[EPIC] Product page - Options tab 9/9


marionf:
### -As a merchant I want to be able to manage the options of my products

https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#options-tab-epic

- [x] #19679
- [x] #19680
- [x] #19681
- [x] #19682
- [x] #19683

### [- As a merchant I want to know if there are errors when I complete a field in the product page](https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#options-tab-)

---

### - Issues in **options tab** sorted by priority high to low :

**Nice to Have**
- [x] #9580
- [x] #15553
- [x] #28430

**Feature**
- [ ] #19179
- [ ] #17362 - **[PR]**
- [x] #14135
- [x] #11795
- [ ] #12095 - **`TBS`**

## Fixed
- [x] #27659
- [x] #18919
- [x] #21633
- [x] #22123
- [x] #9790
- [x] #16272

marionf:
### - Issues in **options tab** sorted by priority high to low :

**Nice to Have**
- [ ] #9580
- [ ] #15553

**Improvement**
- [ ] https://github.com/PrestaShop/PrestaShop/issues/19179 - Module
- [ ] #17362
- [ ] #14135
- [ ] #11795

## Fixed
- [x] #18919 - Multistore
- [x] #21633 [BO][Products][Product customization] checkbox replace switch only after saving
- [x] #9790
- [x] #16272

PR #19732, opened 2020-06-12
UpdateProductOptions Command

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | UpdateProductOptionsCommand and handler
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | no UI to test. CI must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19732)


tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -223,47 +382,42 @@ public function assertProductType(string $productReference, string $productTypeN
     }
 
     /**
-     * @Then I should get error that product name is invalid
+     * @Then I should get error that product field :fieldName is invalid
```
PierreRambaud: You should keep the original statement, we already had a discussion with @eternoendless, Gherkin language should not be specific nor contains technical information.
```suggestion
     * @Then I should get error that product :fieldName is invalid
```

zuk3975: OrderFeature context has the same scenario and it becomes ambigous if i leave it. This is what I decided to do. Other fix would be to rename the OrderFeature scenario to following :
```...i should get error that product for order "[field]" is invalid```

Comment (matks) [reply tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php]: @PierreRambaud This behat test is describing a real end-user situation 😄 it's the usage of BO Product page. User is using a form, the UX is is front of us.

"When I enter a weak password, field 'password' is colored in red and I am prompted an error message about the password field" is valid when talking about a login process isn't it ?

**Comment by zuk3975** [comment]
rr travis :sweat:

File: tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php (hunk 1 above)
Comment (zuk3975) [reply tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php]: the main problem was that Order and Product context contained scenarios that matched same string. 
Finally I reverted the changes in Product context and modified scenarios in Order context.
Instead of ```Then I should get error that quantity is invalid``` I changed to ```Then I should get error that quantity is invalid for order``` , so that Gherkin understands from which Context these assertions comes from.

File: classes/Tag.php [hunk 2]
```diff
@@ -303,13 +303,26 @@ public function setProducts($array)
...
      *
      * @param int $idProduct Product ID
+     * @param int|null $langId provide lang id to delete tags only in specific language
      *
      * @return bool
+     *
+     * @throws PrestaShopDatabaseException
      */
-    public static function deleteTagsForProduct($idProduct)
+    public static function deleteTagsForProduct($idProduct, int $langId = null)
```
jolelievre: To avoid changing the legacy code too much it would be better to have a new `deleteTagsForProductByLang` (but you can use the same private method behind the scene to avoid duplicate code)

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+            $product->available_for_order = $command->isAvailableForOrder();
+            $this->fieldsToUpdate['available_for_order'] = true;
+        }
+
+        if (null !== $command->isOnlineOnly()) {
+            $product->online_only = $command->isOnlineOnly();
+            $this->fieldsToUpdate['online_only'] = true;
+        }
+
+        if (null !== $command->toShowPrice()) {
```
jolelievre: I'm puzzled about this one Why not just `showPrice` ? You think it looks better with the `to` ?

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+                throw new CannotUpdateProductException(
+                    sprintf(
+                        'Failed to update product #%s options',
+                        $product->id
+                    ),
+                    CannotUpdateProductException::FAILED_UPDATE_OPTIONS
+                );
+            }
+
+            $this->updateTags($product, $localizedTags);
```
jolelievre: I would separate the `updateTags` from the `performUpdate` This way `performUpdate` will look like other handlers and allow us to refactor it in the common abstract handler

It would also avoid having a nullable  parameter`?array $localizedTags` in `updateTags`, the null checking can be done before calling the method this way it would have a stricter prototype `array $localizedTags`

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+        // delete all tags for product if array is empty
+        if (empty($localizedTags)) {
+            $this->deleteAllTagsForProduct($productId);
+
+            return;
+        }
+
+        foreach ($localizedTags as $langId => $tags) {
+            // don't do anything with this language tags if its null (equivalent to other partial updates)
+            if (null === $tags) {
```
jolelievre: Nice, I get the idea Just wondering why you would need to add `['fr' => null]` (for example), when not putting any info about this lang would be easier 🤔
But maybe some use cases require to deal with this?

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+            // delete all this product tags for this lang
+            if (false === Tag::deleteTagsForProduct($productId, $langId)) {
+                throw new CannotUpdateProductException(
+                    sprintf('Failed to delete product #%s previous tags in lang #%s', $productId, $langId),
+                    CannotUpdateProductException::FAILED_UPDATE_OPTIONS
+                );
+            }
+
+            // validate each tag and remove empty values
+            $this->validateTags($tags, $langId);
```
jolelievre: It would be better to validate before removing all tags

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+     * Validate each tag in provided language and remove empty values
+     *
+     * @param array &$tags
+     * @param int $langId
+     *
+     * @return void
+     *
+     * @throws ProductConstraintException
+     */
+    private function validateTags(array &$tags, int $langId): void
```
jolelievre: Not a big fan about referenced arrays when it can be avoided, maybe return the filtered array and call `$tags = $this->validateTags($tags, $langId);` instead?

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /**
+     * @param Product $product
+     * @param UpdateProductOptionsCommand $command
+     */
+    private function fillUpdatableFieldsWithCommandData(Product $product, UpdateProductOptionsCommand $command): void
+    {
+        if (null !== $command->getVisibility()) {
+            $product->visibility = $command->getVisibility()->getValue();
```
jolelievre, line 71: All the validation for all fields are missing: `$this->validateLocalizedField($product, 'name', ProductConstraintException::INVALID_NAME);`

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+            // don't do anything with this language tags if its null (equivalent to other partial updates)
+            if (null === $tags) {
+                continue;
+            }
+
+            // delete all this product tags for this lang
+            if (false === Tag::deleteTagsForProduct($productId, $langId)) {
+                throw new CannotUpdateProductException(
+                    sprintf('Failed to delete product #%s previous tags in lang #%s', $productId, $langId),
+                    CannotUpdateProductException::FAILED_UPDATE_OPTIONS
```
jolelievre, line 185: Maybe a dedicated error code here would be useful to tell the difference with product update
```suggestion
                    CannotUpdateProductException::FAILED_UPDATE_TAGS
```

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+            //remove empty value
+            if (empty($tag)) {
+                unset($tags[$key]);
+
+                continue;
+            }
+
+            //validate tag
+            if (false === Validate::isGenericName($tag)) {
+                throw new ProductConstraintException(
```
jolelievre: Missing dedicated error code here for invalid tag

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * @throws CannotUpdateProductException
+     * @throws PrestaShopException
+     */
+    private function deleteAllTagsForProduct(int $productId): void
+    {
+        if (false === Tag::deleteTagsForProduct($productId)) {
+            throw new CannotUpdateProductException(
+                sprintf('Failed to delete all tags for product #%s', $productId),
+                CannotUpdateProductException::FAILED_UPDATE_OPTIONS
```
jolelievre, line 215:
```suggestion
                CannotUpdateProductException::FAILED_UPDATE_TAGS
```

src/Core/Domain/Product/CommandHandler/UpdateProductOptionsHandlerInterface.php
```diff
@@ -26,15 +26,15 @@
...
 
-use PrestaShop\PrestaShop\Core\Domain\Product\Command\UpdateProductDescriptionCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\Command\UpdateProductOptionsCommand;
 
 /**
- * Defines contract for UpdateProductDescriptionHandler
+ * Defines contract for UpdateProductOptionsHandler
  */
-interface UpdateProductDescriptionHandlerInterface
+interface UpdateProductOptionsHandlerInterface
```
jolelievre, line 34: Weird 😅 
A forgotten file I guess?

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
jolelievre: Or do we consider they are useless because the ValueObjects handle this? Though if someone overrides Product class (or we update it ourselves) the new validation rules will be ignored
I wonder if it's better to have a double check 🤔

src/Core/Domain/Product/ValueObject/Isbn.php
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Holds product ISBN code value
+ */
+class Isbn
+{
+    /**
+     * Valid ISBN regex pattern
+     */
+    const VALID_PATTERN = '/^[0-9-]{0,32}$/';
```
jolelievre, line 41: I wonder if those const should be in the ValueObject or in a dedicated interface/class for ISBN, or in a dedicated validation class that contains those constants 🤔

Because if we need them somewhere else it feels weird to have a relation with the ValueObject that is supposed to be used for command only

src/Core/Domain/Product/ValueObject/ProductVisibility.php
```diff
@@ -0,0 +1,90 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+
+/**
+ * Holds value for product visibility setting
+ */
+class ProductVisibility
+{
+    const VISIBLE_IN_CATALOG = 'catalog';
+    const VISIBLE_IN_SEARCH = 'search';
+    const VISIBLE_EVERYWHERE = 'both';
```
jolelievre, line 40: Much better! 👍

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -139,25 +160,98 @@ public function assertLocalizedProperty(string $productReference, string $fieldN
...
+
+            return;
+        }
+
         foreach ($expectedLocalizedValues as $langId => $expectedValue) {
-            $actualValue = $this->extractValueFromProductForEditing($productForEditing, $fieldName)[$langId];
+            $actualValues = $this->extractValueFromProductForEditing($productForEditing, $fieldName);
+            $langIso = Language::getIsoById($langId);
+
+            if (!isset($actualValues[$langId])) {
```
jolelievre, line 175: Nice!

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+declare(strict_types=1);
```
matks, line 27: Progi1984 will like that 😛

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -139,25 +160,98 @@ public function assertLocalizedProperty(string $productReference, string $fieldN
...
         }
     }
 
+    /**
+     * Product tags differs from other localized properties, because each locale can have an array of tags
+     * (whereas common property will have one value per language)
+     * This is why it needs some additional parsing
+     *
+     * @param array $localizedTagStrings key value pairs where key is language id and value is string representation of array separated by comma
+     *                                   e.g. [1 => 'hello, goodbye', 2 => 'bonjour,Au revoir']
```
jolelievre: In this case the tag is ` goodbye` or `goodbye` ? We keep the space?

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * @param Product $product
+     * @param array $localizedTags
+     *
+     * @throws CannotUpdateProductException
+     * @throws PrestaShopException
+     */
+    private function updateTags(Product $product, ?array $localizedTags)
+    {
+        // don't do anything with tags if its null. (It means it is a partial update and tags aren't changed)
```
matks: 👍 clear and useful comment => great !

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -139,25 +160,98 @@ public function assertLocalizedProperty(string $productReference, string $fieldN
...
+            if (!isset($actualLocalizedTags[$langId])) {
+                throw new RuntimeException(sprintf(
+                    'Expected localized tags value "%s" is not set in %s language',
+                    var_export($expectedTags, true),
+                    $langIso
+                ));
+            }
+
+            // finally assert if both non-empty values are equal
+            Assert::assertEquals(
```
jolelievre: What if they are not in the same order? I guess it shouldn't happen?

src/Core/Domain/Product/Command/UpdateProductOptionsCommand.php
```diff
@@ -0,0 +1,337 @@
...
+    /**
+     * @return string[]|null
+     */
+    public function getLocalizedTags(): ?array
+    {
+        return $this->localizedTags;
+    }
+
+    /**
+     * @param string[] $localizedTags
```
jolelievre: We should add a comment about the expected format `array of tags in an array indexed by language ID` And maybe add some validation to assert it's a two dimensional array And not just an array of strings unlike other localized fields

matks (dismissed):

src/Core/Domain/Product/Command/UpdateProductOptionsCommand.php
jolelievre: Unless we make it simpler and use an array of string and the handler is the one that parses the comma string?

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -394,72 +575,47 @@ public function assertProductType(string $productReference, string $productTypeN
...
-     * @Then I should get error that product name is invalid
+     * @Then I should get error that product :fieldName is invalid
      */
-    public function assertLastErrorIsInvalidNameConstraint()
+    public function assertConstraintError(string $fieldName): void
     {
         $this->assertLastErrorIs(
             ProductConstraintException::class,
-            ProductConstraintException::INVALID_NAME
+            $this->getConstraintErrorCode($fieldName)
```
jolelievre, line 602: Nice improvement here as well!!

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
zuk3975: you are right indeed. Overengineering happens :hankey:

classes/Tag.php
zuk3975: If I use same private method to avoid duplication, i will still have the same risk, because old code will be using the same modified code anyway. So i prefer to stay with this, as its already done, is that ok?

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
zuk3975: you are right, this is unnecessary :thinking:

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
zuk3975: well i hope It doesn't happen, but even if it does, we'll have a failing test and then we'll fix it, I don't think its worth investigating it now :thinking:

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
zuk3975: yes, I've removed it because it does absolutely the same.
And It is our domain VO it cannot be overriden by something else, so we shouldn't care about legacy validation anymore.

src/Adapter/Product/CommandHandler/UpdateProductOptionsHandler.php
zuk3975: Then it tells us, that maybe we should separate Tags from options command ?

src/Core/Domain/Product/CommandHandler/UpdateProductOptionsHandlerInterface.php
zuk3975: yes, the one I merged into "basicInformation" command. Forgot to delete one apparently :sweat_smile:

src/Core/Domain/Product/ValueObject/Isbn.php
zuk3975: yes you are right, this is some kind of global standard, idk how should we approach it ?
Even Symfony has such constraint for form validation, so in case we want to use that one, it might introduce some strange behavior if it validates differently

src/Core/Domain/Product/Command/UpdateProductOptionsCommand.php
zuk3975: for now I added a comment, if we see in future that its really worth it, we can change it

matks: > And maybe add some validation to assert it's a two dimensional array And not just an array of strings unlike other localized fields

Hey, good idea ! I always wanted some custom assertions for it.
There is a standard way to describe it: <key, value>

So here it would be <key, <string, string>>

classes/Tag.php
jolelievre: Yes but it clarifies the "API" we did the same thing with `TaxRulesGroup::getTaxRulesGroupsForOptions` recently I don't doubt that the code itself works so the private method is not an issue

But at least we have two distinct functions that can be used for distinct needs, instead of an optional parameter

src/Core/Domain/Product/ValueObject/Isbn.php
jolelievre: yep we can merge this like this for now, but we need to think about it

tests/Integration/Behaviour/Features/Scenario/Product/update_options.feature
```diff
@@ -0,0 +1,123 @@
...
+      And product "product3" localized "tags" should be "en-US:mechanic,watch;fr-FR:"
+      When I update product "product3" options with following values:
+        | tags       | fr-FR:montre |
+      And product "product3" localized "tags" should be "en-US:mechanic,watch;fr-FR:montre"
+
+    Scenario: Update product tags with invalid values
+      Given product "product3" localized "tags" should be "en-US:mechanic,watch;fr-FR:montre"
+      When I update product "product3" options with following values:
+        | tags       | en-US:#<{ |
+      Then I should get error that product tag is invalid
```
jolelievre, line 123: Can you add a test here to validate that the previous values are still present (not removed before the validation)

classes/Tag.php
```diff
@@ -231,7 +231,7 @@ public static function getProductTags($idProduct)
         }
         $result = [];
         foreach ($tmp as $tag) {
-            $result[(int) $tag['id_lang']][] = $tag['name'];
+            $result[$tag['id_lang']][] = $tag['name'];
```
zuk3975: reverted int casting to avoid changing legacy classes. (as now i still need to loop through when building DTO, so I cast there instead)

jolelievre (approved):

matks (approved):

matks:
Thank you @zuk3975

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -332,13 +333,14 @@ private function setUpdateOptionsCommandData(array $data, UpdateProductOptionsCo
 
         if (isset($data['tags'])) {
             $localizedTagStrings = $this->parseLocalizedArray($data['tags']);
-            $localizedTags = [];
+            $localizedTagsList = [];
 
             foreach ($localizedTagStrings as $langId => $localizedTagString) {
-                $localizedTags[$langId] = explode(',', $localizedTagString);
+                $tags = explode(',', $localizedTagString);
+                $localizedTagsList[] = new LocalizedTags($langId, $tags);
```
matks, line 352: Shall we `trim` stuff here, to remove empty spaces ?

PR #19778, opened 2020-06-16
Cast changelogs to array for twig

### LINKED ISSUE #18450: All modules Changelog are not displayed in the backoffice
#### Describe the bug

In the modules updates screen admin/index.php/improve/modules/updates changelog are not displayed.

changelog

#### Expected behavior

When there is a new version of a module the changelog MUST BE provided before the merchant decide to upgrade or not.

#### Additional information

Could be that this information is only available if you have authenticated the shop with the PrestaShop addons platform. This can't be the expected behavior.

Sample addons API call with a changelog inside:
https://api-addons.prestashop.com/?method=listing&action=module&id_module=296&format=json

The addons API is documented here:
https://docs.google.com/document/d/1gEoTVJQXmNZHZM-pGTx0ztAP-578Xybezx-EALGthCA/edit
(private PrestaShop document)

* PrestaShop version: at least all 1.7.6.x

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Even if the `changeLog` key of the api-addons was send, the actual implementation would be unable to show it. This PR fixes it.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18450
| How to test?  | 1. Clear cache with `rm -rf var/cache/*`<br>2. Go to `Improve > Modules > Module manager`<br>3. Uninstall the MBO module<br>4. Upload an [outdated version of the MBO module](https://github.com/PrestaShopCorp/ps_mbo/releases/download/v2.0.0/ps_mbo.zip)<br>5. In the `Updates` tab, you should see the changelog for a new MBO module version available

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19778)

### DIFF
--- a/src/PrestaShopBundle/Resources/views/Admin/Module/Includes/card_notification_update.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Module/Includes/card_notification_update.html.twig
@@ -35,7 +35,7 @@
 {% block addon_description %}
   {% if module.attributes.changeLog is defined %}
     <ul>
-      {% for version, lines in module.attributes.changeLog %}
+      {% for version, lines in module.attributes.changeLog|arrayCast %}
         <li><b>{{ version }}:</b>
           {% for line in lines %}
             {{ line }}<br/>


src/PrestaShopBundle/Resources/views/Admin/Module/Includes/card_notification_update.html.twig
```diff
@@ -35,7 +35,7 @@
 {% block addon_description %}
   {% if module.attributes.changeLog is defined %}
     <ul>
-      {% for version, lines in module.attributes.changeLog %}
+      {% for version, lines in module.attributes.changeLog|arrayCast %}
```
matks, line 38: Suggestion / crazy idea: be able to handle potential future API changes by being able to parse `whatever the API sends to us`

How to do it ?
Parse the API data received into `module.attributes.changeLog` and have something like this:
```
if ($isJson) {
    return json_decode($data);
elseif ($isArray) {
    return arrayCast($data);
elseif ($isStdClass) {
    return objectCast($data);
}
```

This will make sure that, even if Addons API changes in 2 or 3 years, people using PS 1.7.8 will still be able to parse it.

The above ⏫ is a suggestion. Now I have a **mandatory** review feedback 😄 we must make sure that `even if Addons API is broken or changes we must now crash the page`
So can you test what happens if
- fetched data is `null`
- fetched data is bad JSON
- fetched data is JSON / XML / StdClass / array

atomiix: @matks I think everything is already handled by `AddonsDataProvider` which takes care of fetching data from addons but as you can see [here](https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Service/DataProvider/Marketplace/ApiClient.php#L109), `json_decode` returns objects, not arrays.

matks (approved):

src/PrestaShopBundle/Resources/views/Admin/Module/Includes/card_notification_update.html.twig
matks: Gotcha 👍

Progi1984 (approved):

PierreRambaud (approved):

Progi1984:
@atomiix Your fix is on the bad branch. Please target `develop` branch.

atomiix:
@Progi1984, @toutantic wanted it for 1.7.7 ;)

Progi1984:
@atomiix But @eternoendless has defined the milestone to 1.7.8.x

matks:
> @atomiix But @eternoendless has defined the milestone to 1.7.8.x

I just changed the milestone 😄 
I think the issue scope was initially for 178 because we thought 177 would be released before Addons API was ready.

Anyway Aurélien is the boss of @eternoendless so I know who I side with :trollface:

matthieu-rolland (approved):

matthieu-rolland:
@atomiix you removed the waiting for qa label, isn't the PR ready to be tested on 1.7.7.x ?

atomiix:
@matthieu-rolland It wasn't ready until now, I updated the "How to test" section. Everything's good now! 🙂

Progi1984:
Thanks @atomiix & @sarahdib

PR #19800, opened 2020-06-16
Add a new selector in order to select the product page more precisely

### LINKED ISSUE #18509: Product page not updated after changing attributes - BC break

NeOMakinG:
| Questions     | Answers
| Branch?       |  1.7.6.x
| Description?  | The bug fix we did on FO selectors for attributes refresh when changing size, color or quantity of product is a BC, and we force people to use a .row but they don't want to be forced to use a .row class, so I added a selector and refresh the attr only on the first occurence.
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18509.
| How to test?  | Use a product with a promo code, with "Related Product" and "Product in the same category" plugins, change attribute, the price/pictures/other stuff should update fine on the product page but shouldnt update related products and more, also test with "Quick View"

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19800)


Progi1984:
@NeOMakinG Could you target the `1.7.6.x` Branch ?

themes/_core/js/product.js
```diff
@@ -39,21 +39,22 @@ var currentRequestDelayedId = null;
...
-      && prestashop.urls !== null
-      && prestashop.urls.pages !== null
-      && prestashop.urls.pages.product !== ''
-      && prestashop.urls.pages.product !== null
+  if (
+    prestashop !== null &&
+    prestashop.urls !== null &&
+    prestashop.urls.pages !== null &&
+    prestashop.urls.pages.product !== '' &&
+    prestashop.urls.pages.product !== null
```
eternoendless: Please put the boolean operators at the start of the line

Progi1984 (approved):

atomiix (approved):

PierreRambaud (approved):

Progi1984:
Thanks @NeOMakinG & @Robin-Fischer-PS

PR #19820, opened 2020-06-17
Hooks displayAdminOrderTabLink or displayAdminOrderTabContent displayed not escaped to allow raw HTML

### LINKED ISSUE #19822: Hooks displayAdminOrderTabLink or displayAdminOrderTabContent are displayed with raw HTML
#### Describe the bug

When you use displayAdminOrderTabLink or displayAdminOrderTabContent, you'll see HTML content.
image

#### Expected behavior

Content should be not escape.

#### Steps to Reproduce

1. Install https://github.com/PrestaShop/PrestaShop/files/4695351/tottesthooks.zip
2. See error on order page

#### Additional information

* PrestaShop version: develop / 1.7.7.x
* PHP version: N/A

clotairer:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When you want to use dispalyAdminOrderTabLink or dispalyAdminOrderTabContent, HTML content is escape.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19822
| How to test?  | You can use https://github.com/PrestaShop/PrestaShop/files/4695351/tottesthooks.zip you reproduce the bug image

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19820)

### DIFF
--- a/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
@@ -117,12 +117,12 @@
   <div class="mt-2" id="order_hook_tabs">
     <ul class="nav nav nav-tabs" role="tablist">
       {# Rendering of hook displayAdminOrderTabLink, we expect tab links #}
-      {{ displayAdminOrderTabLink }}
+      {{ displayAdminOrderTabLink|raw }}
     </ul>
 
     <div class="tab-content">
       {# Rendering of hook displayAdminOrderTabContent, we expect tab contents #}
-      {{ displayAdminOrderTabContent }}
+      {{ displayAdminOrderTabContent|raw }}
     </div>
   </div>
 {% endif %}


matks:
@PierreRambaud It looks legitimate to allow modules to return raw HTML I believe. But this can be source of XSS.

Is it better to use twig `escape(html)` here ?

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
```diff
@@ -117,12 +117,12 @@
   <div class="mt-2" id="order_hook_tabs">
     <ul class="nav nav nav-tabs" role="tablist">
       {# Rendering of hook displayAdminOrderTabLink, we expect tab links #}
-      {{ displayAdminOrderTabLink }}
+      {{ displayAdminOrderTabLink|raw }}
```
clotairer, line 120: We propose this fix but it could be better to update renderhook twig function.

clotairer (commented):
@matks or manage security restriction inside renderhook twig extension...

PierreRambaud (approved):
There is no other way since hooks content are stored in variables. And because hooks are rendering links & HTML content we don't care about the raw, it's not coming from a user input :)

matks:
All right ! so `|raw` it is 😊 

And I'll check if I can find more hooks which need this `|raw` too

matks (approved):

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/details.html.twig
matks: We got @PierreRambaud approval !
It looks like it can be done by passing `is_safe` option to the `renderHook()` TwigFunction

So we shall pass `is_safe` option to all TwigFunctions in src/PrestaShopBundle/Twig/HookExtension.php I think

matks: So actually these functions _already_ have the `is_safe` option but because I put the result in a twig variable, it was escaped again.
So the `|raw` here is good 👍 and no other modification is needed

matks:
I checked this PR
- with provided demo module
- with official demo module https://github.com/PrestaShop/example-modules/tree/master/demovieworderhooks

All is OK for me ✅ QA is OK

matks:
Thank you @clotaire202 !

Codencode:
In Prestashop 1.7.7.0 Beta displayAdminOrderTabLink and displayAdminOrderTabLink don't work because **hookcount** function don't work

matks:
> In Prestashop 1.7.7.0 Beta displayAdminOrderTabLink and displayAdminOrderTabLink don't work because **hookcount** function don't work

You're right, there was a 2nd issue but it was solved as well 😉 see https://github.com/PrestaShop/PrestaShop/pull/19657

Issue #19523, opened 2020-06-02
[Basic settings tab] - Pack 13/14


MatShir:
As a merchant, I want to be able to complete the basic information of my pack products.

- [ ] **Pack search for a product**: Select Pack of products as product type in the drop-down. A search field to search a product is displayed. You can search by product name or product reference.       
When you start typing, if there are results, they are displayed and highlighted in the dropdown when using a keyboard and hover with mouse.
If you try to search for another product that is already a pack, it shouldn't be displayed in the dropdown because you can't add a pack in a pack.
When you select a result, then the product name is inserted in the search bar.       
You shouldn't be able to add the same product several times to the pack. If the product is already in the pack, you shouldn't see it in the results \([issue 9617](https://github.com/PrestaShop/PrestaShop/issues/9617)\)

- [ ] **Pack add a product**: Button to add the product in the quantity specified to the pack.

- [ ] **Pack delete a product**: When you hover one of the pack products, a trash appears to remove the product from the pack. When you click on it, you have a modal for cancel or confirm the action.

---
List of issues related to Pack:
- [x] #11587
- [ ] #20443
- [x] #25582

MatShir:
Issue:
- [ ] https://github.com/PrestaShop/PrestaShop/issues/11587
- [ ] https://github.com/PrestaShop/PrestaShop/issues/20443

PR #19827, opened 2020-06-17
UpdateProductPackCommand

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | 
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #19523
| How to test?  | non testable in UI. CI pases -> :ok_hand: 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19827)


src/Adapter/Product/CommandHandler/AddProductToPackHandler.php
```diff
@@ -0,0 +1,117 @@
...
+
+final class AddProductToPackHandler extends AbstractProductHandler implements AddProductToPackHandlerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(AddProductToPackCommand $command): void
+    {
+        // Pack will have the id of product which is being edited
+        $pack = $this->getProduct($command->getPackId());
```
matks: At first I thought it was a mistake `$pack = $this->getPack()` ... then I understood 😅

src/Adapter/Product/CommandHandler/AddProductToPackHandler.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Builds string with ids from command, that will help to identify objects that was being updated in case of error
+     *
+     * @param string $messageBody
+     * @param AddProductToPackCommand $command
+     *
+     * @return string
+     */
+    private function buildMessageWithCommandInputs(string $messageBody, AddProductToPackCommand $command): string
```
matks: 👍

matks (dismissed):
Flawless again

src/Adapter/Product/CommandHandler/AddProductToPackHandler.php
jolelievre: You should that `$pack` is a pack itself no?
Or maybe the first time you adda product it's not a Pack yet?

src/Adapter/Product/CommandHandler/AddProductToPackHandler.php
```diff
@@ -0,0 +1,117 @@
...
+        $combinationId = $command->getCombinationId();
+
+        if (null === $combinationId) {
+            $combinationId = CombinationId::NO_COMBINATION;
+        }
+
+        try {
+            $this->assertProductIsAvailableForPacking($productId);
+
+            if (false === Pack::deleteItems($pack->id)) {
```
jolelievre: Why do you delete all items each time? Then it can only contain one product no?

src/Adapter/Product/CommandHandler/AddProductToPackHandler.php
```diff
@@ -0,0 +1,117 @@
...
+use Pack;
+use PrestaShop\PrestaShop\Adapter\Product\AbstractProductHandler;
+use PrestaShop\PrestaShop\Core\Domain\Product\Combination\ValueObject\CombinationId;
+use PrestaShop\PrestaShop\Core\Domain\Product\Command\AddProductToPackCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\AddProductToPackHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductPackingException;
+use PrestaShopException;
+
+final class AddProductToPackHandler extends AbstractProductHandler implements AddProductToPackHandlerInterface
```
jolelievre: I'm not sure any more about what we decided This is related to your comment here https://github.com/PrestaShop/docs/pull/558/files#r442763264

At first we talked about the trio `AddProductToPackCommand` `RemoveProductFromPackCommand` (and additional one `UpdatePackedProductQuantityCommand` as you said)

But I believe after discussion with @matks we decided that it would be simpler to have a single `UpdateProductPackCommand` which contains all the items of the pack (along with their quantities I guess)

It also matches the note from the PR https://github.com/PrestaShop/docs/pull/558/files#diff-c85d131d0c114e4a1d61d32b2f9cccfbR70

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -94,6 +97,38 @@ public function addProduct(string $productReference, TableNode $table): void
...
 
+    /**
+     * @When I pack :quantity :type product :productReference to a pack of :packReference
+     *
+     * @param string $type
+     * @param int $quantity
+     * @param string $productReference
+     * @param string $packReference
+     */
+    public function addProductToPack(string $type, int $quantity, string $productReference, string $packReference)
```
jolelievre: I don't understand why you need this type? Maybe there's a particular way to handle combination (will it be a different command?) But if so you can use a different function later Keep this one simple for now without the type

tests/Integration/Behaviour/Features/Scenario/Product/add_to_pack.feature
```diff
@@ -0,0 +1,39 @@
...
+  Scenario: I add standard product to a pack
+    Given I add product "productPack1" with following information:
+      | name       | en-US: weird sunglasses box |
+      | is_virtual | false                       |
+    And product "productPack1" type should be standard
+    And I add product "product2" with following information:
+      | name       | en-US: shady sunglasses     |
+      | is_virtual | false                       |
+    And product "product2" type should be standard
+    When I pack 5 standard product "product2" to a pack of "productPack1"
```
jolelievre: This is not easy to understand I would suggest:
```suggestion
    When I add 5 products "product2" to a pack of "productPack1"
```

(also remove the type as suggested above because it's confusing especially you add virtual products.. ^^)

tests/Integration/Behaviour/Features/Scenario/Product/add_to_pack.feature
```diff
@@ -0,0 +1,39 @@
...
+    Given I add product "productPack2" with following information:
+      | name       | en-US: street photos        |
+      | is_virtual | false                       |
+    And product "productPack2" type should be standard
+    And I add product "product3" with following information:
+      | name       | en-US: summerstreet         |
+      | is_virtual | true                        |
+    And product "product3" type should be virtual
+    When I pack 3 standard product "product3" to a pack of "productPack2"
+    Then product "productPack2" type should be pack
```
jolelievre: Oh I forgot, some test to validate what the pack contains will be needed
And some tests that update/delete products as well

src/Adapter/Product/CommandHandler/AddProductToPackHandler.php
zuk3975: > You should that `$pack` is a pack itself no?
> Or maybe the first time you adda product it's not a Pack yet?

Any product can become a pack. There is an assertion later to make sure we are not adding a pack into a pack, but nothing more..
Legacy used to assert if product type is a pack, but that was only a form assertion, command doesn't have anything to do with that. (as we've discussed-technically we can just decide if its a pack in read model by checking if it has pack ralations in db.

zuk3975: > You should that `$pack` is a pack itself no?
> Or maybe the first time you adda product it's not a Pack yet?

Any product can become a pack. There is an assertion later to make sure we are not adding a pack into a pack, but nothing more..
Legacy used to assert if product type is a pack, but that was only a form assertion, command doesn't have anything to do with that. (as we've discussed- technically we can just decide if its a pack in read model by checking if it has pack relations in db)

src/Adapter/Product/CommandHandler/AddProductToPackHandler.php
zuk3975: good catch Ill check that.
Based on your suggestion to use UpdateProductPack instead, after refactoring ill provide a list of products instead of one so that each time it will override existing relations with new ones (like legacy did).

tests/Integration/Behaviour/Features/Scenario/Product/add_to_pack.feature
zuk3975: okay.
I thought we'll need different string for combinational product, but now I think that we can use the same string as you suggest, without mentioned type and we will get the combination id from the reference.. So later ill need to refactor to save product_id and combination_id inside product reference instead of id only.

src/Adapter/Product/CommandHandler/AddProductToPackHandler.php
zuk3975: Ok, Ill check what I can do

src/Adapter/Product/CommandHandler/UpdateProductPackHandler.php
```diff
@@ -0,0 +1,134 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\QuantifiedProduct;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+use PrestaShopException;
+
+final class UpdateProductPackHandler extends AbstractProductHandler implements UpdateProductPackHandlerInterface
+{
+    /**
+     * @var int product id which represents the pack
+     */
+    private $packId;
```
matks: Is the purpose of this variable to be used in `buildMessageWithCommandInputs()` ?

I am usually against such variables because they make the code stateful. If `UpdateProductPackHandler` handles 2 commands, one after the other, it could have strange behaviors.

So if we can avoid having this, I will be happy.

However if the sole purpose of this is to make `buildMessageWithCommandInputs()` easier to use ... well, it's just log message. It's not a big deal 😄 . But in that case at least check `$this->packId` exists before using it.

src/Adapter/Product/CommandHandler/UpdateProductPackHandler.php
```diff
@@ -0,0 +1,134 @@
...
+                    );
+                }
+            } catch (PrestaShopException $e) {
+                throw new ProductException(
+                    $this->buildMessageWithCommandInputs('Error occurred when trying to add product to pack.', $productForPacking),
+                    0,
+                    $e
+                );
+            } finally {
+                Pack::resetStaticCache();
```
matks, line 93: Lovely statements 😄

src/Core/Domain/Product/Command/UpdateProductPackCommand.php
```diff
@@ -0,0 +1,90 @@
...
+        return $this->products;
+    }
+
+    /**
+     * @param array $products
+     */
+    private function setProducts(array $products): void
+    {
+        foreach ($products as $product) {
+            //@todo: shouldn't i validate the quantity, at least if its greater than zero?
```
matks: It would be nice, but if we don't do it, it's nos big deal. If merchant wants to create a pack with a "zero product A" in it, well, he does whatever he wants, right ? 😄

zuk3975: I wonder if he makes it negative if it won't break anything further :man_shrugging:

tests/Integration/Behaviour/Features/Context/CommonFeatureContext.php
```diff
@@ -95,6 +97,15 @@ public static function getContainer()
         return static::$kernel->getContainer();
     }
 
+    /**
+     * @AfterFeature @clear-cache-after-feature
+     */
+    public static function clearCache()
```
matks, line 103: 👍

matks (dismissed):

src/Core/Domain/Product/Command/UpdateProductPackCommand.php
matks: Oh, yeah, you're right. Let's check it's `> 0` then

src/Adapter/Product/CommandHandler/UpdateProductPackHandler.php
```diff
@@ -0,0 +1,134 @@
...
+                ProductPackingException::FAILED_DELETING_PREVIOUS_PACKS
+            );
+        }
+
+        //reset cache_default_attribute
+        $pack->setDefaultAttribute(0);
+
+        foreach ($productsForPacking as $productForPacking) {
+            $productId = $productForPacking->getProductId()->getValue();
+            $combinationId = $productForPacking->getCombinationId();
```
jolelievre: I believe `getCombinationId` is a `CombinationId` so it should be something like:
```php
if (null === $productForPacking->getCombinationId()) {
    $combinationId = CombinationId::NO_COMBINATION;
} else {
    $combinationId = $productForPacking->getCombinationId()->getValue();
}
```

jolelievre: You also should add a behat test with a combination to validate this

src/Adapter/Product/QueryHandler/GetPackedProductsHandler.php
```diff
@@ -0,0 +1,62 @@
...
+        $langId = $query->getLanguageId()->getValue();
+
+        $packedItems = Pack::getItems($packId, $langId);
+
+        $packedProducts = [];
+        foreach ($packedItems as $packedItem) {
+            $packedProducts[] = new PackedProduct(
+                (int) $packedItem->id,
+                (int) $packedItem->pack_quantity,
+                $packedItem->id_pack_product_attribute ? (int) $packedItem->id_pack_product_attribute : null
```
jolelievre: Why not use `CombinationId::NO_COMBINATION` here as well?

matks (dismissed):

src/Core/Domain/Product/Exception/ProductPackingException.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Exception;
+
+/**
+ * Thrown when some product packing actions fails
+ */
+class ProductPackingException extends ProductException
```
jolelievre:
```suggestion
class ProductPackException extends ProductException
```
We use the word pack everywhere else, better be consistent no?

src/Adapter/Product/CommandHandler/UpdateProductPackHandler.php
```diff
@@ -0,0 +1,130 @@
...
+     *
+     * @return string
+     */
+    private function appendIdsToMessage(string $messageBody, QuantifiedProduct $product, int $packId): string
+    {
+        return sprintf(
+            "$messageBody. [packId #%s; productId #%s; combinationId #%s]",
+            $packId,
+            $product->getProductId()->getValue(),
+            $product->getCombinationId()->getValue()
```
jolelievre: You need to check if `getCombinationId` is not null

src/Core/Domain/Product/Query/GetPackedProducts.php
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public function getPackId(): ProductId
+    {
+        return $this->packId;
+    }
+
+    /**
+     * @return LanguageId
+     */
+    public function getLanguageId(): LanguageId
```
jolelievre: I'm not sure this is really useful to set the language id, especially since `PackedProduct` doesn't contain any localized value You could simply use the default language id for `Pack:getItems`
Or maybe it would be useful to have more info about the product from the pack, like their name and images Do you think it's something that will be added later?

tests/Integration/Behaviour/Features/Scenario/Product/update_pack.feature
```diff
@@ -0,0 +1,65 @@
...
+    Given product "product2" type should be standard
+    And product "product3" type should be virtual
+    When I update pack productPack4 with following product quantities:
+      | product2 | 2 |
+      | product3 | 3 |
+    Then product "productPack4" type should be pack
+    And pack productPack4 should contain products with following quantities:
+      | product2   | 2                          |
+      | product3   | 3                          |
+
```
jolelievre, line 64: Maybe a test with empty products and check that the product's type is not Pack any more

Unless you think a dedicated command would be better? `CleanPackCommand` or something similar?

zuk3975: lets do separate command for deletion. Ill refactor that UpdateProductPack throws exception if empty array is provided.

jolelievre (approved):

matks (approved):

matks:
Thank you @zuk3975

PR #19828, opened 2020-06-17
Check for ::GRID_ID when definition is AbstractGridDefinitionFactory

### LINKED ISSUE #19812: [BO - employees page] - Filter are not working
#### Describe the bug
Filter are not working on Employees page on BO
#### Expected behavior
Filter should work

#### Steps to Reproduce

1. Go to 'BO - employees Page'
2. Filter by any column of the grid (id = 1 for example)
3. Click on search
4. See error : Page is refreshed, list of employees is the same, filter column is empty (not filter is applied)

**Screen record of the bug :**  
[Bug employees page.zip](https://github.com/PrestaShop/PrestaShop/files/4791151/Bug.employees.page.zip)

#### Additional information

* PrestaShop version: branch 1.7.7.x
* PHP version: 7.2

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Filters from the employees page weren't working anymore. This PR fixes it.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19812
| How to test?  | 1. Go to Configure > Advanced Parameters > Team<br>2.Filter by any column of the grid (id = 1 for example)<br>3. Click on search<br>4. The filter should work as expected

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19828)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/CommonController.php
+++ b/src/PrestaShopBundle/Controller/Admin/CommonController.php
@@ -41,6 +41,7 @@ use PrestaShopBundle\Security\Annotation\AdminSecurity;
 use PrestaShopBundle\Service\DataProvider\Admin\RecommendedModules;
 use PrestaShopBundle\Service\Grid\ControllerResponseBuilder;
 use PrestaShopBundle\Service\Grid\ResponseBuilder;
+use ReflectionClass;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\RedirectResponse;
@@ -363,29 +364,27 @@ class CommonController extends FrameworkBundleAdminController
         /** @var GridDefinitionFactoryInterface $definitionFactory */
         $definitionFactory = $this->get($gridDefinitionFactoryServiceId);
 
-        if ($definitionFactory instanceof FilterableGridDefinitionFactoryInterface) {
-            /** @var ResponseBuilder $responseBuilder */
-            $responseBuilder = $this->get('prestashop.bundle.grid.response_builder');
+        $filterId = null;
 
-            return $responseBuilder->buildSearchResponse(
-                $definitionFactory,
-                $request,
-                $definitionFactory->getFilterId(),
-                $redirectRoute,
-                $redirectQueryParamsToKeep
-            );
+        if ($definitionFactory instanceof FilterableGridDefinitionFactoryInterface) {
+            $filterId = $definitionFactory->getFilterId();
+        } elseif ($definitionFactory instanceof AbstractGridDefinitionFactory) {
+            // for backward compatibility for AbstractGridDefinitionFactory
+            // using ::GRID_ID (that has been replaced by AbstractFilterableGridDefinitionFactory)
+            $reflect = new ReflectionClass($definitionFactory);
+            if (array_key_exists('GRID_ID', $reflect->getConstants())) {
+                $filterId = $definitionFactory::GRID_ID;
+            }
         }
 
-        // for backward compatibility with AbstractGridDefinitionFactory
-        // replaced by AbstractFilterableGridDefinitionFactory
-        if ($definitionFactory instanceof AbstractGridDefinitionFactory) {
+        if (null !== $filterId) {
             /** @var ResponseBuilder $responseBuilder */
             $responseBuilder = $this->get('prestashop.bundle.grid.response_builder');
 
             return $responseBuilder->buildSearchResponse(
                 $definitionFactory,
                 $request,
-                $definitionFactory->getDefinition()->getId(),
+                $filterId,
                 $redirectRoute,
                 $redirectQueryParamsToKeep
             );


Progi1984 (dismissed):

matks (dismissed):
Nice catch !
Then ... do you think @atomiix that we should check _all_ similar filters to add the Grid ID ? because it might mean they don't work ?

Progi1984:
@matks It's not in the scope of the issue ;)

matks:
> @matks It's not in the scope of the issue ;)

Yeah, I know, but I'm asking @atomiix whether he thinks another issue should be created 😉 to handle the issue globally

atomiix:
@matks yep, I checked every other filters and even if some of them don't have this `protected $filterId` defined, they are all still working because they are handled differently :)

matks:
> @matks yep, I checked every other filters and even if some of them don't have this `protected $filterId` defined, they are all still working because they are handled differently :)

So what was wrong with this grid 🤔 ? What is different about it ? Should we change that, so this grid behaves like the others ?

I'm asking because we want, as much as possible, to have all grids using similar behavior. This makes module developers task easier.

atomiix:
>So what was wrong with this grid 🤔 ? What is different about it ? Should we change that, so this grid behaves like the others ?

This grid extends from `AbstractGridDefinitionFactory` and is handled by `CommonController::searchGridAction` which led to use `prestashop.bundle.grid.response_builder` to build the search response. In this configuration, the `$filterId` property of the filter should be set.

The other filters that don't have this property set are all using `prestashop.bundle.grid.controller_response_builder` to build the search response (through their own action in their controller, or because they don't extend from `FilterableGridDefinitionFactoryInterface` or `AbstractGridDefinitionFactory`)

Now, if you think we should try to make them all use `prestashop.bundle.grid.response_builder` we definitely have to create a new issue!

matks:
> > So what was wrong with this grid 🤔 ? What is different about it ? Should we change that, so this grid behaves like the others ?
> 
> This grid extends from `AbstractGridDefinitionFactory` and is handled by `CommonController::searchGridAction` which led to use `prestashop.bundle.grid.response_builder` to build the search response. In this configuration, the `$filterId` property of the filter should be set.
> 
> The other filters that don't have this property set are all using `prestashop.bundle.grid.controller_response_builder` to build the search response (through their own action in their controller, or because they don't extend from `FilterableGridDefinitionFactoryInterface` or `AbstractGridDefinitionFactory`)
> 
> Now, if you think we should try to make them all use `prestashop.bundle.grid.response_builder` we definitely have to create a new issue!

Is it the issue I have fixed in https://github.com/PrestaShop/PrestaShop/pull/19506 ?

atomiix:
>Is it the issue I have fixed in #19506 ?

Yep, that specific filter was working before that PR but from what I saw, it was the only one not working after.

PierreRambaud (dismissed):

src/Core/Grid/Definition/Factory/FilterableEmployeeGridDefinitionFactory.php
```diff
@@ -0,0 +1,300 @@
...
+                        $this->trans('Yes', [], 'Admin.Global') => 1,
+                        $this->trans('No', [], 'Admin.Global') => 0,
+                    ],
+                    'required' => false,
+                    'choice_translation_domain' => false,
+                ])
+            )
+            ->add((new Filter('active', YesAndNoChoiceType::class))
+                ->setAssociatedColumn('active')
+            )
```
PululuK: Double active ? 🤔

atomiix: Nice catch @PululuK, this file is mainly just a copy/paste from [`EmployeeGridDefinitionFactory`](https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Grid/Definition/Factory/EmployeeGridDefinitionFactory.php) but indeed, I'll have to change it ;)

PululuK: @atomiix  see https://github.com/PrestaShop/PrestaShop/pull/19873

matks (commented):
I reviewed the PR, everything is OK excluding @PululuK comment https://github.com/PrestaShop/PrestaShop/pull/19828/files#r442855002

If you fix it, I approve the PR :)

src/Core/Grid/Definition/Factory/FilterableEmployeeGridDefinitionFactory.php
atomiix: @PululuK I removed this duplicated filter on the new `FilterableEmployeeGridDefinitionFactory` class and cherry-picked your commit https://github.com/PululuK/PrestaShop/commit/234d87e83633bc158ae1dffb69e45463dfbe2e8e We should be all good!

PululuK: Thank you @atomiix 

@matks  I'm very curious to see what will happen on my [PR](https://github.com/PrestaShop/PrestaShop/pull/19873) if you merge this one before 🤔

IMHO  :  

- We shouldn't fix `EmployeeGridDefinitionFactory.php` here. It's not in the scope of the issue. In this PR just `FilterableEmployeeGridDefinitionFactory.php` should be fixed
-  Or, you can close this [PR](https://github.com/PrestaShop/PrestaShop/pull/19873), but in this case QA team need to check if  `BO > Advanced params > teams` filter by active/disabled work too.

PululuK (dismissed):

matks (dismissed):
I approve this PR as it is valid.

I have an idea though. A suggestion:

Would it be interesting to have `FilterableEmployeeGridDefinitionFactory extends EmployeeGridDefinitionFactory` to remove a lot of copy-paste ?

src/Core/Grid/Definition/Factory/FilterableEmployeeGridDefinitionFactory.php
```diff
@@ -0,0 +1,288 @@
...
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
```
Progi1984:
```suggestion
 */
 declare(strict_types=1);
```

Progi1984 (dismissed):

PierreRambaud (dismissed):

matks:
@atomiix I compared the code of CommonController before (= what I implemented) and after your PR

In my code, I did this:
```
if ($definitionFactory instanceof AbstractGridDefinitionFactory) {
            /** @var ResponseBuilder $responseBuilder */
            $responseBuilder = $this->get('prestashop.bundle.grid.response_builder');

            return $responseBuilder->buildSearchResponse(
                $definitionFactory,
                $request,
                $definitionFactory->getDefinition()->getId(),
                $redirectRoute,
                $redirectQueryParamsToKeep
            );
        }
```

So I thought at this time that, if you have a `AbstractGridDefinitionFactory`, then you can fetch the filter ID thanks to `$definitionFactory->getDefinition()->getId();`

But you fetch it by doing
```
$reflect = new ReflectionClass($definitionFactory);
            if (array_key_exists('GRID_ID', $reflect->getConstants())) {
                $filterId = $definitionFactory::GRID_ID;
            }
```

So it means `$definitionFactory->getDefinition()->getId()` is not reliable ?

atomiix:
@matks exactly! If you check https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/Grid/Definition/Factory/EmployeeGridDefinitionFactory.php for instance, It extends `AbstractGridDefinitionFactory` without using `GRID_ID`. That means that the filter isn't using the Definition's `GRID_ID` either and so it cannot work with filterId.

jolelievre:
@matks no `$definitionFactory->getDefinition()->getId()` is not reliable in that case because all grids have this data, wether they save their Filter via a `filterId` or via a `controller/action`

So the presence of the idea doesn't allow to differenciate those tow use cases, which is why the `FilterableGridDefinitionFactoryInterface` was introduced, to identify clearly which grid use a filterId but some grid might use it without the interface and the only way to detect those is via the `GRID_ID` const

It's not ideal and we should update all those grids with the correct interface, but this check is still necessary for backward compatibility

jolelievre (approved):

matks (approved):
Answers are legit 👍 it makes sense

LGTM !

Progi1984 (approved):

SD1982:
LGTM Thanks @atomiix

eternoendless:
Thank you @atomiix

PR #19866, opened 2020-06-19
Simplify Taxes and Tax options forms

JevgenijVisockij:
| Questions     | Answers
| Branch?       | develop
| Description?  | Simplify forms for Taxes and Tax options, introduces BC breaks for both forms by extending TranslatorAwareType. Now any forms that have extended TaxType or TaxOptionType will crash.
| Type?         | refacto
| Category?     | BO
| BC breaks?    |  yes
| Deprecations? | no
| Fixed ticket? | Fixes #16482
| How to test?  | Tax page and tax options pages must work exactly like before. Also they must look mostly the same.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19866)

### DIFF
--- a/src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
+++ b/src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
@@ -113,6 +113,8 @@ class TaxOptionsType extends TranslatorAwareType
             ])
             ->add('tax_address_type', ChoiceType::class, [
                 'label' => $this->trans('Based on', 'Admin.International.Feature'),
+                'required' => false,
+                'placeholder' => false,
                 'choices' => $this->taxAddressTypeChoiceProvider->getChoices(),
             ])
             ->add('use_eco_tax', SwitchType::class, [
--- a/src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
+++ b/src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
@@ -56,7 +56,7 @@ class TaxOptionsType extends TranslatorAwareType
     /**
      * TaxOptionsType constructor.
      *
-     * Backwards compatibility break due to addition of translator via TranslatorAwareTrait
+     * Backwards compatibility break introduced in 1.7.8.0 due to extension of TranslatorAwareType
      *
      * @param TranslatorInterface $translator
      * @param array $locales
--- a/src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
+++ b/src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
@@ -28,16 +28,19 @@ namespace PrestaShopBundle\Form\Admin\Improve\International\Tax;
 
 use PrestaShop\PrestaShop\Core\Form\FormChoiceProviderInterface;
 use PrestaShopBundle\Form\Admin\Type\SwitchType;
+use PrestaShopBundle\Translation\TranslatorAwareTrait;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\FormBuilderInterface;
-use Symfony\Component\Translation\TranslatorInterface;
+use RuntimeException;
 
 /**
  * Defines "Improve > International > Taxes" options form
  */
 class TaxOptionsType extends AbstractType
 {
+    use TranslatorAwareTrait;
+
     /**
      * @var bool
      */
@@ -54,28 +57,22 @@ class TaxOptionsType extends AbstractType
     private $taxRuleGroupChoiceProvider;
 
     /**
-     * @var TranslatorInterface
-     */
-    private $translator;
-
-    /**
      * TaxOptionsType constructor.
+     *
+     * Backwards compatibility break due to addition of translator via TranslatorAwareTrait
      *
      * @param bool $ecoTaxEnabled
      * @param FormChoiceProviderInterface $taxAddressTypeChoiceProvider
      * @param FormChoiceProviderInterface $taxRuleGroupChoiceProvider
-     * @param TranslatorInterface $translator
      */
     public function __construct(
         $ecoTaxEnabled,
         FormChoiceProviderInterface $taxAddressTypeChoiceProvider,
-        FormChoiceProviderInterface $taxRuleGroupChoiceProvider,
-        TranslatorInterface $translator
+        FormChoiceProviderInterface $taxRuleGroupChoiceProvider
     ) {
         $this->ecoTaxEnabled = $ecoTaxEnabled;
         $this->taxAddressTypeChoiceProvider = $taxAddressTypeChoiceProvider;
         $this->taxRuleGroupChoiceProvider = $taxRuleGroupChoiceProvider;
-        $this->translator = $translator;
     }
 
     /**
@@ -83,12 +80,16 @@ class TaxOptionsType extends AbstractType
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        if ($this->translator ===null) {
+            throw new RuntimeException('Translator variable must not be null');
+        }
+
         $builder->add(
             'enable_tax',
             SwitchType::class,
             [
-                'label' => $this->translator->trans('Enable tax', [], 'Admin.International.Feature'),
-                'help' => $this->translator->trans(
+                'label' => $this->trans('Enable tax', [], 'Admin.International.Feature'),
+                'help' => $this->trans(
                     'Select whether or not to include tax on purchases.',
                     [],
                     'Admin.International.Help'
@@ -100,12 +101,12 @@ class TaxOptionsType extends AbstractType
             ]
         )
             ->add('display_tax_in_cart', SwitchType::class, [
-                'label' => $this->translator->trans(
+                'label' => $this->trans(
                     'Display tax in the shopping cart',
                     [],
                     'Admin.International.Feature'
                 ),
-                'help' => $this->translator->trans(
+                'help' => $this->trans(
                     'Select whether or not to display tax on a distinct line in the cart.',
                     [],
                     'Admin.International.Help'
@@ -116,14 +117,14 @@ class TaxOptionsType extends AbstractType
                 ],
             ])
             ->add('tax_address_type', ChoiceType::class, [
-                'label' => $this->translator->trans('Based on', [], 'Admin.International.Feature'),
+                'label' => $this->trans('Based on', [], 'Admin.International.Feature'),
                 'required' => false,
                 'choices' => $this->taxAddressTypeChoiceProvider->getChoices(),
             ])
             ->add('use_eco_tax', SwitchType::class, [
-                'label' => $this->translator->trans('Use ecotax', [], 'Admin.International.Feature'),
+                'label' => $this->trans('Use ecotax', [], 'Admin.International.Feature'),
                 'required' => false,
-                'help' => $this->translator->trans(
+                'help' => $this->trans(
                     'If you disable the ecotax, the ecotax for all your products will be set to 0.',
                     [],
                     'Admin.International.Help'),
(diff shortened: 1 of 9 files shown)


Progi1984 (dismissed):

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
```diff
@@ -52,38 +53,89 @@ class TaxOptionsType extends AbstractType
...
      * @param FormChoiceProviderInterface $taxAddressTypeChoiceProvider
      * @param FormChoiceProviderInterface $taxRuleGroupChoiceProvider
+     * @param TranslatorInterface $translator
      */
     public function __construct(
         $ecoTaxEnabled,
         FormChoiceProviderInterface $taxAddressTypeChoiceProvider,
-        FormChoiceProviderInterface $taxRuleGroupChoiceProvider
+        FormChoiceProviderInterface $taxRuleGroupChoiceProvider,
+        TranslatorInterface $translator
```
PierreRambaud: BC breaks :thinking:

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
```diff
@@ -52,38 +53,89 @@ class TaxOptionsType extends AbstractType
...
     }
 
     /**
      * {@inheritdoc}
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder->add('enable_tax', SwitchType::class)
-            ->add('display_tax_in_cart', SwitchType::class)
+        $builder->add('enable_tax', SwitchType::class, [
```
PierreRambaud:
```suggestion
        $builder->add(
            'enable_tax', 
            SwitchType::class, 
            [
```

If you want a better indentation (look at line 97, it's really weird)

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
matks: Indeed.

@JevgenijVisockij you need to make this parameter optional `TranslatorInterface $translator = null` to make sure that, if someone has extended this class, this will not break its code.

So
1. make it an optional new argument `,TranslatorInterface $translator = null)`
2. in all the places where you use it, check whether it's null `if ($this->translator === null)`

If however this adds a lot of `if ($this->translator === null)` everywhere, another solution is to deprecate `TaxOptionsType` and create a new class instead. We use the new class but we leave old TaxOptionsType to avoid breaking code that would rely on it.

New class could be `TwigOptimizedTaxOptionsType` for example 😄

src/PrestaShopBundle/Resources/views/Admin/Improve/International/Tax/Blocks/form.html.twig
```diff
@@ -34,32 +34,8 @@
...
-      }) }}
-
-      {{ ps.form_group_row(taxForm.is_enabled, {}, {
-        'label': 'Enable'|trans({}, 'Admin.Actions')
-      }) }}
-
-      {% block tax_form_rest %}
-        {{ form_rest(taxForm) }}
+      {% block tax_form_widget %}
+        {{ form_widget(taxForm) }}
```
matks, line 38: Looks amazing 🤩

matks (commented):
Apart from the BC break; everything is good 😄 

Great work !

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
JevgenijVisockij: So I wondered about first option. Translator is only really being used in one function, buildForm. However if add this at the start of the function
```
      if ($this->translator === null) {
            throw new Exception('Fail');
        }
```
It's still a BC break right? If somebody extends TaxOptionType and relies on buildForm now it will throw an exception.
So in that case I should use alternative to $this->translator instead?

matks: Yes, you're right.
So I look deeper at the PR, well, without the translator nothing really works.

I see 2 possibilities:
1. use another class, as mentioned above => we let this class here for backward compatibility and use a new one
2. introduce "soft" BC break by using optional arg but add

```
if ($this->translator === null) {
            throw new Exception('This class requires the translator to be injected to work.');
}
```

it's a BC break a little more friendly as it does not make module code crash but it tells developer what it needs

wdty @jolelievre @matthieu-rolland ?

jolelievre: Throwing an exception or forcing the parameter is a bit the same to me 🤔. Either way it's a BC break that would cause an error at some point

If we absolutely want to avoid BC break then adding a new class is the only solution, however in this case I would choose the BC break to avoid duplicate code and dead code Especially because we are talking a rather specific form type that has very few changes of being extended

A last alternative would be to introduce a trait or interface to inject the translator via DI automatically, this would avid the BC break from the constructor and if the form type is extended as long as it uses the DI and is not used directly it should be correctly injected as well

JevgenijVisockij: Can't we use already existing TranslatorAwareTrait for that? 

I don't understand traits that well yet, but you only need to add for it to work, and this won't cause BC break right?

```
        calls:
            - { method: setTranslator, arguments: ['@translator'] }
```

PierreRambaud: If the user needs to change something to make it works, it's a BC break.

matks: > Can't we use already existing TranslatorAwareTrait for that?
> 
> I don't understand traits that well yet, but you only need to add for it to work, and this won't cause BC break right?
> 
> ```
>         calls:
>             - { method: setTranslator, arguments: ['@translator'] }
> ```

The idea is the following: let's say someone wants to create a module that provides advanced capabilities for taxes. He sees most of the fields he looks for are available in TaxOptionsType. So he says: "cool ! I will reuse this class" and he creates a `MyModuleTaxType extends TaxOptionsType" and he declares it like this:

```yaml
      class: 'MyModule\MyModuleTaxType'
      arguments:
        - '@=service("prestashop.adapter.legacy.configuration").getBoolean("PS_USE_ECOTAX")'
        - '@=service("prestashop.core.form.choice_provider.tax_address_type_choice_provider")'
        - '@=service("prestashop.core.form.choice_provider.tax_rule_group_choice_provider")'
      tags:
        - { name: form.type }
```

No we modify TaxOptionsType to require the translator for 1.7.8 .

The shop owner, who uses the module, updates prestashop to 1.7.8. He loads the page where MyModuleTaxType is used. The page crashes because the translator needs to be injected.

The module _needs to be fixed_ by adding
```yml
- { method: setTranslator, arguments: ['@translator'] }
```

**If the module developer needs to modify his implementation to "just make it work" it means it was a BC break.**

matks: So if we agree, I suggest we follow this:

> I would choose the BC break to avoid duplicate code and dead code Especially because we are talking a rather specific form type that has very few changes of being extended

To do this is pretty simple:
- in the code we do the BC break (maybe identify explicitely with a comment
- we mark this PR as "BC break"

matks (approved):

matks (changes requested):
Requesting one last change:
- use TranslatorAwareTrait to introduce translation capabilities into the FormType
- in the code, when using the translator, perform a `if ($this->translator !==null)` else throw a RuntimeException
- add a comment into the constructor to explicitely acknowledge the BC break

And I add the "BC break" label to the PR

matks (approved):
Approved ✅

Progi1984 (dismissed):

matks (dismissed):

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxType.php
```diff
@@ -61,8 +62,24 @@ public function __construct(TranslatorInterface $translator)
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        $invalidCharsText = sprintf(
+            '%s ' . TypedRegexValidator::CATALOG_CHARS,
+            $this->translator->trans('Invalid characters:', [], 'Admin.Notifications.Info')
```
eternoendless: FYI
This form type could be simplified by making it extend `TranslatorAwareType` instead of `AbstractType`. That way you could simply use `$this->trans`

src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/bootstrap_4_layout.html.twig
```diff
@@ -800,7 +800,7 @@
 
 {% block form_help %}
   {% if help %}
-    <small class="form-text">{{ help }}</small>
+    <small class="form-text">{{ help|nl2br }}</small>
```
eternoendless: Be aware that this should have no effect on your PR since you're not using this file in the forms you edited (this file is the old one which will be deleted once we switch to the new form theme).

The `form_help` block is declared here: https://github.com/PrestaShop/PrestaShop/blob/6ca37a71cc51668f45732d3616f2e6ab71875ff2/src/PrestaShopBundle/Resources/views/Admin/TwigTemplateForm/prestashop_ui_kit_base.html.twig#L821-L828

I know this may be confusing, but the `bootstrap_4_layout.html` file referenced in `prestashop_ui_kit_base.html.twig` actually points to `vendor/symfony/symfony/src/Symfony/Bridge/Twig/Resources/views/Form/bootstrap_4_layout.html.twig`, not to this file.

khouloudbelguith:
Hi @JevgenijVisockij,

After a fresh install of your PR, I found these issues:
1. Go to BO => International => Taxes => Taxes Rules => Edit the first tax rule "FR Taux standard (20%)"  => there is no tax 
image
Before it was ok
image

2. This issue is detected by automatic tests launched by @nesrineabdmouleh 
Go to BO => International => Taxes => Tax options => Enable "Use ecotax" and save
In the Ecotax options, the "FR Taux standard (20%)"  is not displayed
image

Before was ok
image

We are still testing this PR, if we found some other issues we will keep you updated.

Thanks!

SD1982:
Hello @JevgenijVisockij  After more investigation we no longer detect the problem reported by @khouloudbelguith  the only point that could be corrected is this one ( it s not a big deal but it maybe can have an  impact our automated test )

With your PR there is now a blank in the "based on" dropdown selector
Screenshot at 11-58-27


There is no blank in this dropdown without the PR
Screenshot at 11-54-15

matks (dismissed):

matks (approved):
Approved ✅

matthieu-rolland (approved):

LouiseBonnard (approved):

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
```diff
@@ -55,15 +56,22 @@ class TaxOptionsType extends AbstractType
...
+     *
+     * @param TranslatorInterface $translator
+     * @param array $locales
      * @param bool $ecoTaxEnabled
      * @param FormChoiceProviderInterface $taxAddressTypeChoiceProvider
      * @param FormChoiceProviderInterface $taxRuleGroupChoiceProvider
      */
     public function __construct(
+        TranslatorInterface $translator,
+        array $locales,
```
PierreRambaud, line 69: It's a BC :)

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
```diff
@@ -74,16 +82,58 @@ public function __construct(
...
+                    'Select whether or not to include tax on purchases.',
+                    'Admin.International.Help'
+                ),
+                'required' => false,
+                'attr' => [
+                    'class' => 'js-enable-tax',
+                ],
+            ]
+        )
+            ->add('display_tax_in_cart', SwitchType::class, [
```
PierreRambaud, line 100: You should write as you did for 'enable_tax', the indentation is weird.

PierreRambaud: Sorry, I didn't notice the description hasn't been changed, but the label was added :wink:

src/PrestaShopBundle/Form/Admin/Improve/International/Tax/TaxOptionsType.php
PierreRambaud: Sorry, I didn't notice the description hasn't been changed, but the label was added wink

khouloudbelguith:
Hi @JevgenijVisockij,
After more investigation and tests with @SD1982.
Your PR is ok :heavy_check_mark: 

Thanks!

Progi1984:
Thanks @JevgenijVisockij @SD1982 & @khouloudbelguith

Issue #19518, opened 2020-06-02
[Basic settings tab] - Associated categories 8/14


MatShir:
As a merchant I want to be able to complete the basic information of my products.

- [ ] **Select category**: Allows to associate categories to the product.        
A category is necessarily selected, it is impossible not to have one.       
When you create a new product, “Home” is selected.       
When you search a category elements are highlighted in dropdown when using keyboard and hover with mouse \([issue 20331](https://github.com/PrestaShop/PrestaShop/issues/20331)\).       
When you click on the result, the category appears in “associated category” and is checked in the list. You can also select a category in the tree by clicking in the checkbox, then the category appears in “associated category”.
- [ ]  **Unselect category**: You can unselect a category by unchecking the checkbox in the tree. Then, the category is removed in “associated category”.       
You can also click on the cross in “associated category”. Then, the checkbox for the corresponding category is unchecked.       
If you try to unselect all categories, you can’t do it for the last one.
- [x]  **Category tree**: When you create a new product, the tree is open for the first 2 levels of categories \([issue 14955](https://github.com/PrestaShop/PrestaShop/issues/14955)\).       
When I click on "collapse" it folds the whole tree. I can also click on each parent to folds / unfolds its subcategories. When I click on "expend" the whole tree is opened.       
Categories are sorted by level AND position:       
Categ 1 level 1 position 1       
Categ 3 level 1 position 2       
Categ 2 level 1 position 3       
Categ 2-2 level 2 position 1       
Categ 2-1 level 2 position 2       
- [ ]  **Main category**: Radio button to select the main category of the product. There can only be one.           
The main category is the one that will be in the product URL, before the product name and in the breadcrumb in front-office you have all parent categories up to the main one.
- [ ]  **Parent of the category**: Drop-down with all categories enabled, sorted by level and position:       
  Categ 1 level 1 position 1           
  Categ 3 level 1 position 2          
  Categ 2 level 1 position 3             
  Categ 2-2 level 2 position 1          
  Categ 2-1 level 2 position 2

MatShir:
Issues:

- [ ] Product page - tech task - Improve Category tree widget to handle huge number of categories #19281
- [ ] https://github.com/PrestaShop/PrestaShop/issues/9626
- [ ] https://github.com/PrestaShop/PrestaShop/issues/16415
- [ ] https://github.com/PrestaShop/PrestaShop/issues/20331

zuk3975:
[...]

PR #19871, opened 2020-06-19
Add UpdateProductCategoriesCommand

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | AssignProductToCategoriesCommand
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #19518
| How to test?  | non testable UI. CI pass -> :ok_hand: 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19871)

### DIFF
--- a/classes/Category.php
+++ b/classes/Category.php
@@ -499,7 +499,9 @@ class CategoryCore extends ObjectModel
             // update by batch of 5000 categories
             $chunks = array_chunk($queries, 5000);
             foreach ($chunks as $chunk) {
-                $sqlChunk = array_map(function ($value) { return '(' . rtrim(implode(',', $value)) . ')'; }, $chunk);
+                $sqlChunk = array_map(function ($value) {
+                    return '(' . rtrim(implode(',', $value)) . ')';
+                }, $chunk);
                 Db::getInstance()->execute('INSERT INTO `' . _DB_PREFIX_ . 'category` (id_category, nleft, nright)
                 VALUES ' . rtrim(implode(',', $sqlChunk), ',') . '
                 ON DUPLICATE KEY UPDATE nleft=VALUES(nleft), nright=VALUES(nright)');
@@ -690,15 +692,15 @@ class CategoryCore extends ObjectModel
         }
 
         $cacheId = 'Category::getAllCategoriesName_' . md5(
-            (int) $idRootCategory .
-            (int) $idLang .
-            (int) $active .
-            (int) $useShopRestriction .
-            (isset($groups) && Group::isFeatureActive() ? implode('', $groups) : '') .
-            (isset($sqlFilter) ? $sqlFilter : '') .
-            (isset($orderBy) ? $orderBy : '') .
-            (isset($limit) ? $limit : '')
-        );
+                (int) $idRootCategory .
+                (int) $idLang .
+                (int) $active .
+                (int) $useShopRestriction .
+                (isset($groups) && Group::isFeatureActive() ? implode('', $groups) : '') .
+                (isset($sqlFilter) ? $sqlFilter : '') .
+                (isset($orderBy) ? $orderBy : '') .
+                (isset($limit) ? $limit : '')
+            );
 
         if (!Cache::isStored($cacheId)) {
             $result = Db::getInstance()->executeS(
@@ -1025,7 +1027,7 @@ class CategoryCore extends ObjectModel
 				LEFT JOIN `' . _DB_PREFIX_ . 'product` p
 					ON p.`id_product` = cp.`id_product`
 				' . Shop::addSqlAssociation('product', 'p') .
-                (Combination::isFeatureActive() ? ' LEFT JOIN `' . _DB_PREFIX_ . 'product_attribute_shop` product_attribute_shop
+            (Combination::isFeatureActive() ? ' LEFT JOIN `' . _DB_PREFIX_ . 'product_attribute_shop` product_attribute_shop
 				ON (p.`id_product` = product_attribute_shop.`id_product` AND product_attribute_shop.`default_on` = 1 AND product_attribute_shop.id_shop=' . (int) $context->shop->id . ')' : '') . '
 				' . Product::sqlStock('p', 0) . '
 				LEFT JOIN `' . _DB_PREFIX_ . 'category_lang` cl
@@ -1043,9 +1045,9 @@ class CategoryCore extends ObjectModel
 					ON m.`id_manufacturer` = p.`id_manufacturer`
 				WHERE product_shop.`id_shop` = ' . (int) $context->shop->id . '
 					AND cp.`id_category` = ' . (int) $this->id
-                    . ($active ? ' AND product_shop.`active` = 1' : '')
-                    . ($front ? ' AND product_shop.`visibility` IN ("both", "catalog")' : '')
-                    . ($idSupplier ? ' AND p.id_supplier = ' . (int) $idSupplier : '');
+            . ($active ? ' AND product_shop.`active` = 1' : '')
+            . ($front ? ' AND product_shop.`visibility` IN ("both", "catalog")' : '')
+            . ($idSupplier ? ' AND p.id_supplier = ' . (int) $idSupplier : '');
 
         if ($random === true) {
             $sql .= ' ORDER BY RAND() LIMIT ' . (int) $randomNumberProducts;
@@ -1604,6 +1606,38 @@ class CategoryCore extends ObjectModel
 		WHERE c.`id_category` = ' . (int) $idCategory);
 
         return isset($row['id_category']);
+    }
+
+    /**
+     * Check if all categories by provided ids are present in database.
+     * If at least one is missing return false
+     *
+     * @param int[] $categoryIds
+     *
+     * @return bool
+     *
+     * @throws PrestaShopDatabaseException
+     */
+    public static function categoriesExists(array $categoryIds): bool
+    {
+        if (empty($categoryIds)) {
+            return false;
+        }
+
+        $categoryIds = array_map('intval', $categoryIds);
+        $categoryIdsFormatted = implode(',', $categoryIds);
+
+        $result = Db::getInstance()->query('
+            SELECT COUNT(c.id_category) as categories_found
+            FROM ' . _DB_PREFIX_ . 'category c
+            WHERE c.id_category IN (' . $categoryIdsFormatted . ')
+        ')->fetch();
+
+        if (count($categoryIds) === (int) $result['categories_found']) {
+            return true;
+        }
+
+        return false;
     }
 
     /**
--- a/classes/Category.php
+++ b/classes/Category.php
@@ -499,9 +499,7 @@ class CategoryCore extends ObjectModel
             // update by batch of 5000 categories
             $chunks = array_chunk($queries, 5000);
             foreach ($chunks as $chunk) {
-                $sqlChunk = array_map(function ($value) {
-                    return '(' . rtrim(implode(',', $value)) . ')';
-                }, $chunk);
+                $sqlChunk = array_map(function ($value) { return '(' . rtrim(implode(',', $value)) . ')'; }, $chunk);
                 Db::getInstance()->execute('INSERT INTO `' . _DB_PREFIX_ . 'category` (id_category, nleft, nright)
                 VALUES ' . rtrim(implode(',', $sqlChunk), ',') . '
                 ON DUPLICATE KEY UPDATE nleft=VALUES(nleft), nright=VALUES(nright)');
@@ -692,15 +690,15 @@ class CategoryCore extends ObjectModel
         }
 
         $cacheId = 'Category::getAllCategoriesName_' . md5(
-                (int) $idRootCategory .
-                (int) $idLang .
-                (int) $active .
-                (int) $useShopRestriction .
-                (isset($groups) && Group::isFeatureActive() ? implode('', $groups) : '') .
-                (isset($sqlFilter) ? $sqlFilter : '') .
-                (isset($orderBy) ? $orderBy : '') .
-                (isset($limit) ? $limit : '')
-            );
+            (int) $idRootCategory .
+            (int) $idLang .
+            (int) $active .
+            (int) $useShopRestriction .
+            (isset($groups) && Group::isFeatureActive() ? implode('', $groups) : '') .
+            (isset($sqlFilter) ? $sqlFilter : '') .
+            (isset($orderBy) ? $orderBy : '') .
+            (isset($limit) ? $limit : '')
+        );
 
         if (!Cache::isStored($cacheId)) {
             $result = Db::getInstance()->executeS(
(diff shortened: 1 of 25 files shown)


matks (dismissed):

classes/Category.php
```diff
@@ -1606,6 +1606,38 @@ public static function categoryExists($idCategory)
...
+            SELECT COUNT(c.id_category) as categories_found
+            FROM ' . _DB_PREFIX_ . 'category c
+            WHERE c.id_category IN (' . $categoryIdsFormatted . ')
+        ')->fetch();
+
+        if (count($categoryIds) === (int) $result['categories_found']) {
+            return true;
+        }
+
+        return false;
```
PierreRambaud:
```suggestion
        return count($categoryIds) === (int) $result['categories_found']);
```

src/Adapter/Product/AbstractProductHandler.php
```diff
@@ -129,4 +130,29 @@ protected function validateField(Product $product, string $field, int $errorCode
...
+        try {
+            if (false === $product->update()) {
+                throw new CannotUpdateProductException(
+                    sprintf('Failed to update product #%s', $product->id),
+                    $errorCode
+                );
+            }
+        } catch (PrestaShopException $e) {
+            throw new ProductException(
+                sprintf('Error occurred when trying to update product #%s', $product->id),
```
PierreRambaud:
```suggestion
                sprintf('Error occurred when trying to update product #%d', $product->id),
```

src/Adapter/Product/AbstractProductHandler.php
```diff
@@ -129,4 +130,29 @@ protected function validateField(Product $product, string $field, int $errorCode
...
+     *
+     * @throws CannotUpdateProductException
+     * @throws ProductException
+     */
+    protected function performUpdate(Product $product, int $errorCode): void
+    {
+        try {
+            if (false === $product->update()) {
+                throw new CannotUpdateProductException(
+                    sprintf('Failed to update product #%s', $product->id),
```
PierreRambaud:
```suggestion
                    sprintf('Failed to update product #%d', $product->id),
```

src/Adapter/Product/CommandHandler/AssignProductToCategoriesHandler.php
```diff
@@ -0,0 +1,113 @@
...
+     *
+     * @throws CannotUpdateProductException
+     * @throws ProductException
+     */
+    private function updateCategories(Product $product, array $categoryIds): void
+    {
+        try {
+            if (false === Category::categoriesExists($categoryIds)) {
+                throw new CannotUpdateProductException(
+                    sprintf('Failed to update product #%s categories. Some of categories doesn\'t exist.', $product->id),
```
PierreRambaud:
```suggestion
                    sprintf('Failed to update product #%d categories. Some of categories doesn\'t exist.', $product->id),
```

src/Adapter/Product/CommandHandler/AssignProductToCategoriesHandler.php
```diff
@@ -0,0 +1,113 @@
...
+     * @return int[]
+     */
+    private function formatCategoryIdsList(AssignProductToCategoriesCommand $command): array
+    {
+        $categoryIds = array_map(function (CategoryId $categoryId) {
+            return $categoryId->getValue();
+        }, $command->getCategoryIds());
+
+        $categoryIds[] = $command->getDefaultCategoryId()->getValue();
+        $categoryIds = array_unique($categoryIds, SORT_NUMERIC);
```
PierreRambaud: Not mandatory but no reason to use SORT_NUMERIC since `getValue` for `CategoryId` should return int. This parameter is only here to change the type :thinking:

src/Adapter/Product/CommandHandler/AssignProductToCategoriesHandler.php
```diff
@@ -0,0 +1,113 @@
...
+            if (false === Category::categoriesExists($categoryIds)) {
+                throw new CannotUpdateProductException(
+                    sprintf('Failed to update product #%s categories. Some of categories doesn\'t exist.', $product->id),
+                    CannotUpdateProductException::FAILED_ASSIGN_TO_CATEGORIES
+                );
+            }
+
+            if (false === $product->updateCategories($categoryIds)) {
+                throw new CannotUpdateProductException(
+                    sprintf('Failed to update product #%s categories', $product->id),
```
PierreRambaud:
```suggestion
                    sprintf('Failed to update product #%d categories', $product->id),
```

src/Adapter/Product/CommandHandler/AssignProductToCategoriesHandler.php
```diff
@@ -0,0 +1,113 @@
...
+                    CannotUpdateProductException::FAILED_ASSIGN_TO_CATEGORIES
+                );
+            }
+        } catch (PrestaShopException $e) {
+            throw new ProductException(
+                sprintf('Error occurred when trying to update product #%s categories', $product->id),
```
PierreRambaud:
```suggestion
                sprintf('Error occurred when trying to update product #%d categories', $product->id),
```

tests/Integration/Behaviour/Features/Context/Domain/CategoryFeatureContext.php
```diff
@@ -456,6 +457,27 @@ public function bulkDisableCategories(string $categoriesReferences)
...
+     * @Given category :categoryReference in default language named :categoryName exists
+     *
+     * @param string $categoryReference
+     * @param string $categoryName
+     */
+    public function assertCategoryExistsByName(string $categoryReference, string $categoryName)
+    {
+        $foundCategory = Category::searchByName($this->defaultLanguageId, $categoryName, true);
+
+        if (!$foundCategory || $foundCategory['name'] !== $categoryName) {
```
PierreRambaud:
```suggestion
        if (!isset($foundCategory['name']) || $foundCategory['name'] !== $categoryName) {
```

You should check for `$foundCategory['name']` which is the key you're using in the condition.

src/Adapter/Product/AbstractProductHandler.php
zuk3975: It doesn't make any difference here, but i guess its semantically correct :thinking:

src/Adapter/Product/CommandHandler/AssignProductToCategoriesHandler.php
zuk3975: by default its SORT_STRING.
I guess best is to provide SORT_REGULAR to avoid redundant conversion behind the scenes ?

src/Adapter/Product/AbstractProductHandler.php
PierreRambaud: We must use `%f`, `%s`, `%d` depending on the type otherwise the sprint has no aim and a simple concatenation is enough :sweat_smile:

src/Adapter/Product/CommandHandler/AssignProductToCategoriesHandler.php
PierreRambaud: Types are already an integer, the `SORT_REGULAR` is the default option so
```
array_unique($categoryIds)
```
is enough

zuk3975: default option is SORT_STRING

zuk3975:
@PierreRambaud resolved feedback :heavy_check_mark:

src/Adapter/Product/CommandHandler/AssignProductToCategoriesHandler.php
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Handles AssignProductToCategoriesCommand using legacy object model
+ */
+final class AssignProductToCategoriesHandler extends AbstractProductHandler implements AssignProductToCategoriesHandlerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(AssignProductToCategoriesCommand $command): void
```
jolelievre: Even though this naming is more correct, maybe we could keep something close to all other commands for consistency Like `UpdateProductCategoriesCommand` ? Especially since it seems to erase the previous one so it does more than just assign, it also un-assign?

src/Adapter/Product/CommandHandler/AssignProductToCategoriesHandler.php
```diff
@@ -0,0 +1,113 @@
...
+    {
+        $product = $this->getProduct($command->getProductId());
+        $defaultCategoryId = $command->getDefaultCategoryId()->getValue();
+        $categoryIds = $this->formatCategoryIdsList($command);
+
+        $this->updateCategories($product, $categoryIds);
+
+        $product->id_category_default = $defaultCategoryId;
```
jolelievre: No need for partial update?

src/Adapter/Product/AbstractProductHandler.php
```diff
@@ -129,4 +130,29 @@ protected function validateField(Product $product, string $field, int $errorCode
...
+    {
+        try {
+            if (false === $product->update()) {
+                throw new CannotUpdateProductException(
+                    sprintf('Failed to update product #%d', $product->id),
+                    $errorCode
+                );
+            }
+        } catch (PrestaShopException $e) {
+            throw new ProductException(
```
jolelievre, line 151: Haha, I did something similar in a branch of mine ^^

But why not a `CannotUpdateProductException` here? As it was in the `private` functions

src/Adapter/Product/CommandHandler/AddProductHandler.php
```diff
@@ -74,7 +74,7 @@ public function handle(AddProductCommand $command): ProductId
             if (!$product->add()) {
                 throw new ProductException('Failed to add new basic product');
             }
-            $product->updateCategories([$this->defaultCategoryId]);
+            $product->addToCategories($this->defaultCategoryId);
```
jolelievre: I know it accepts `int` or `int[]` but it feels more logical with an array

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -296,6 +299,166 @@ public function assertLocalizedProperty(string $productReference, string $fieldN
...
+    /**
+     * Product tags differs from other localized properties, because each locale can have an array of tags
+     * (whereas common property will have one value per language)
+     * This is why it needs some additional parsing
+     *
+     * @param array $localizedTagStrings key value pairs where key is language id and value is string representation of array separated by comma
+     *                                   e.g. [1 => 'hello,goodbye', 2 => 'bonjour,Au revoir']
+     * @param ProductForEditing $productForEditing
+     */
+    private function assertLocalizedTags(array $localizedTagStrings, ProductForEditing $productForEditing)
```
jolelievre, line 384: How come this one always moves PR after PR 😅

tests/Integration/Behaviour/Features/Scenario/Product/add_product.feature
```diff
@@ -1,5 +1,6 @@
 # ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml -s product --tags add
 @reset-database-before-feature
+@clear-cache-after-feature
```
jolelievre, line 3: I don't know if it's related but Behat tests seem to fail on this feature file

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -296,6 +299,166 @@ public function assertLocalizedProperty(string $productReference, string $fieldN
...
+     * @Then product :productReference should be assigned to following categories:
+     *
+     * @param string $productReference
+     * @param TableNode $table
+     */
+    public function assertProductCategories(string $productReference, TableNode $table)
+    {
+        $data = $table->getRowsHash();
+        $productId = $this->getSharedStorage()->get($productReference);
+        $actualCategoryIds = array_map('intval', Product::getProductCategories($productId));
```
jolelievre: You use legacy classes to perform the check, you don't plan to make the Query part for ProductForEditing in this PR?

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -296,6 +299,166 @@ public function assertLocalizedProperty(string $productReference, string $fieldN
...
+        Assert::assertEquals($actualCategoryIds, $expectedCategoryIds, 'Unexpected categories assigned to product');
+    }
+
+    /**
+     * @When I assign product :productReference to following categories including non-existing ones:
+     *
+     * @param string $productReference
+     * @param TableNode $table
+     */
+    public function assignToCategoriesIncludingNonExistingOnes(string $productReference, TableNode $table)
```
jolelievre, line 334: Maybe you could have only one statement function for this, an remove `assignToCategories` ?

jolelievre (commented):
A few comments, and there is an error in behat tests Don't know if it's related to the annotation that was added in `add_product.feature`

src/Adapter/Product/AbstractProductHandler.php
zuk3975: As i remember we used to throw generic exception in this case so it shows that it wasn't really the update that caused it, but rather some connection problems. idk really I don't mind putting `..Update..Exception`

zuk3975: there is no point allowing partially set default category when we are reassigning whole category tree each time :thinking:

src/Adapter/Product/CommandHandler/AssignProductToCategoriesHandler.php
zuk3975: there is no point allowing partially set default category when we are reassigning whole category tree each time :thinking:

src/Adapter/Product/AbstractProductHandler.php
jolelievre: Ok good for me, I was just surprised because the previous private method used the `CannotUpdateProductException` It makes sense to use a different Exception but then maybe it shouldn't be a `ProductException` either but an even more generic

Although when I check at all other handlers they use the same convention so `ProductException` is the best choice

src/Adapter/Product/CommandHandler/UpdateProductCategoriesHandler.php
```diff
@@ -0,0 +1,113 @@
...
+     */
+    public function handle(UpdateProductCategoriesCommand $command): void
+    {
+        $product = $this->getProduct($command->getProductId());
+        $defaultCategoryId = $command->getDefaultCategoryId()->getValue();
+        $categoryIds = $this->formatCategoryIdsList($command);
+
+        $this->updateCategories($product, $categoryIds);
+
+        $product->id_category_default = $defaultCategoryId;
```
jolelievre, line 57: Why not use the partial update here? With `$product->setFieldsToUpdate` like in other handlers?

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
jolelievre: You didn't answer to this question, do you think it's worth having a single function for all use cases, as it seems this one would work for the simple use case?
Or better keep a simple one, and a more complex one?

src/Core/Domain/Product/Exception/CannotUpdateProductException.php
```diff
@@ -50,4 +50,9 @@ class CannotUpdateProductException extends ProductException
      * When product tags update fails
      */
     const FAILED_UPDATE_TAGS = 40;
+
+    /**
+     * When failed to assign product categories
+     */
+    const FAILED_ASSIGN_TO_CATEGORIES = 50;
```
jolelievre:
```suggestion
    const FAILED_UPDATE_CATEGORIES = 50;
```
This should be renamed later for consistency, it can be done in another PR

jolelievre (dismissed):

src/Core/Domain/Product/Exception/CannotUpdateProductException.php
zuk3975: :+1: Ill better do it now :p

jolelievre (approved):

matks (approved):

src/Adapter/Product/CommandHandler/AssignProductToCategoriesHandler.php
PierreRambaud: Ohh, I really needed vacations :facepalm:

matks:
Thank you @zuk3975

PR #19903, opened 2020-06-23
Auto create symfony roles

### LINKED ISSUE #19857: Symfony controller from module without a tab has no authorization role created
#### Describe the bug

When a module is installed `ModuleTabRegister` is in charge of creating the associated Tab, which creates the appropriate authorization roles

If the module creates its tab automatically (via `$tabs` property) or manually everything is fine, but if there is a controller (with permission via annotation) without tab no roles will be created thus an Exception will be thrown when trying to access it

For legacy controllers the service automatically scans the controller classes present in `controllers/admin` and adds undeclared tabs (tab not visible in the menu but used for security)

#### Expected behavior

The `ModuleTabRegister` service must be able to scan the `routing` file from module and detect the routes that have a `_legacy_controller` option These undeclared controllers must be automatically added in invisible tabs

#### Steps to Reproduce

A test module should be provided to validate this behaviour, it must contain:
- a Symfony controller with routing using `_legacy_controller`
- no legacy controller classes in the `controllers/admin` folder
- the `configure` method of the module must redirect to the Symfony route

Without the fix an exception should be thrown with missing autorization role, when the behaviour is fiex roles are automatically added and the controller is accessible

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Symfony controllers that do not create Tab had no authorization roles created, with this update the `ModuleTabRegister` is able to scan routing file from module and automatically creates hidden Tab which creates the necessary roles (the route needs to have a `_legacy_controller` option defined)
| Type?         | improvement
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | Fixes #19857
| How to test?  | To test this new feature you need to use the new demo module from this PR https://github.com/PrestaShop/example-modules/pull/21 (an archive is attached to this issue if you prefer) Once the module is installed go to the configuration of the module and try to access the "Secured controller" page Without this PR you are redirected to the Dashboard because of a permission denied error With this PR reinstall the module so that the roles are correctly created, the "Secured controller" page is now accessible

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19903)

Use this module to test this PR:

[democontrollertabs.zip](https://github.com/PrestaShop/PrestaShop/files/4816408/democontrollertabs.zip)

## Breaking Change

The `ModuleTabRegister` constructor needs a new parameter, however since this class is meant for internal use and in the Adapter namespace the impact should be limited.

### DIFF
--- a/src/Adapter/Module/Tab/ModuleTabRegister.php
+++ b/src/Adapter/Module/Tab/ModuleTabRegister.php
@@ -162,16 +162,7 @@ class ModuleTabRegister
             }
         }, $tabs);
 
-        $legacyControllersFilenames = $this->getModuleAdminControllersFilename($moduleName);
-        $legacyControllers = array_map(function ($legacyControllersFilename) {
-            return str_replace('Controller.php', '', $legacyControllersFilename);
-        }, $legacyControllersFilenames);
-        $legacyControllers = $legacyControllers ?? [];
-
-        $routingControllers = $this->getModuleControllersFromRouting($moduleName);
-        $routingControllers = $routingControllers ?? [];
-
-        $detectedControllers = array_merge($legacyControllers, $routingControllers);
+        $detectedControllers = $this->getDetectedModuleControllers($moduleName);
         foreach ($detectedControllers as $adminControllerName) {
             if (in_array($adminControllerName, $tabsNames)) {
                 continue;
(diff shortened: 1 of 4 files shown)


src/Adapter/Module/Tab/ModuleTabRegister.php
```diff
@@ -80,22 +81,36 @@ class ModuleTabRegister
...
     /**
      * @param TabRepository $tabRepository
      * @param LangRepository $langRepository
      * @param LoggerInterface $logger
      * @param TranslatorInterface $translator
      * @param Filesystem $filesystem
      * @param array $languages
+     * @param LoaderInterface $routingConfigLoader
      */
-    public function __construct(TabRepository $tabRepository, LangRepository $langRepository, LoggerInterface $logger, TranslatorInterface $translator, Filesystem $filesystem, array $languages)
```
matks, line 91: BC break.
But it's a class highly related to internal behavior so I think it's all right.

However we should acknowledge it:
- mark "yes" to "BC breaks ?" in PR body
- add "BC break" label

matks (approved):
PR is good 👍

PierreRambaud (approved):

SD1982:
LGTM Thanks @jolelievre !!

Progi1984:
Thanks @jolelievre  & @SD1982

PR #19922, opened 2020-06-24
[CO] Mail variable are not replaced

SebBareyre:
| Questions     | Answers
| Branch?       | develop
| Description?  | There is a bug when customer has uppercase character in the mail domain. Swift convert mail domain automaticaly to lower case. The replacement of variable doesn't work in this case..
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Just create (or modify) a customer account with an email containing uppercase in domain. Try to receive transactional email like order confirmation. The variable in email are not replaced.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19922)

### DIFF
--- a/classes/Mail.php
+++ b/classes/Mail.php
@@ -316,10 +316,8 @@ class MailCore extends ObjectModel
                           self::mimeEncode($addrName);
                 $message->addTo(self::toPunycode($addr), $addrName);
             }
-            $toPlugin = $to[0];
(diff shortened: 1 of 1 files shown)


SebBareyre:
Ok, it's already fixed in 1.7.7...sorry :)

PR #19938, opened 2020-06-25
Adjust UpdatePack command to be able to remove items

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | implement static factories in UpdatePackCommand. cleanPack() and upsertPack(). Adjust behat tests and add some missing cases. All in all the UpdatePackCommand will handle all cases - it will add/update products list in pack or delete them. Basically it always replace old association with new one. We can separate commands if we really need it later, but for now its not worth the excessive work until we are sure.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | not testable

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19938)

### DIFF
--- a/src/Core/Domain/Product/Command/UpdateProductPackCommand.php
+++ b/src/Core/Domain/Product/Command/UpdateProductPackCommand.php
@@ -52,7 +52,7 @@ class UpdateProductPackCommand
      *
      * @param int $packId the id of product which becomes the pack after it contains packed items
      *
-     * @return static
+     * @return UpdateProductPackCommand
      */
     public static function cleanPack(int $packId): self
     {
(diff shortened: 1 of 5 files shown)


matthieu-rolland (dismissed):

matks (approved):

src/Core/Domain/Product/Command/UpdateProductPackCommand.php
```diff
@@ -45,17 +45,39 @@ class UpdateProductPackCommand
...
+     *
+     * @return UpdateProductPackCommand
+     */
+    public static function cleanPack(int $packId): self
+    {
+        return new self($packId, []);
+    }
+
+    /**
+     * Builds command to upsert pack with new products list.
```
jolelievre, line 63: upsert? 🤔
Like update/insert? Then I would propose `indate` ! It's easier to pronounce! 😂

jolelievre: Ok after a quick google search I just learnt it's actually a thing 😅 
In PostgreSQL for instance

jolelievre (approved):
I thought of only one statif catory But the second one which checks that the array is not empty stabilizes the contract even more
Nice touch!

PR #19968, opened 2020-06-26
Fixed specific price list in the product editor

### LINKED ISSUE #20021: Catalog Price Rule without condition group doesn't appear in the product editor
#### Describe the bug

When a Catalog Price Rule is created without condition group, this rule is not listed in the Price tab in the product editor.

#### Expected behavior

Display the Catalog Price Rule created.

#### Steps to Reproduce

1. Create a Catalog Price Rule without any condition group
2. Go to the product list and edit some product
3. The Catalog Price Rule created is not listed

**Screenshots**

https://streamable.com/jc22a9

#### Additional information

* PrestaShop version: 1.7.8 / 1.7.7
* PHP version: 7.2

ghost:
| Questions     | Answers
| Branch?       | develop
| Description?  | This fix show the hidden specific price created from the Catalog Price Rule when no condition group is applied, and this also fix another error that duplicated the row of the specific price in the database when saving changes to the product.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | #20021
| How to test?  | Create a Catalog Price Rule without any condition group. Error video: https://streamable.com/jc22a9

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19968)


Progi1984:
@Rolige Could you rebase your PR for removing two first commits "Merge" ? Thanks

classes/SpecificPriceRule.php
```diff
@@ -275,14 +275,16 @@ public function getAffectedProducts($products = false)
...
+                if (!SpecificPrice::getByProductId(0, false, false, (int) $this->id)) {
+                    $query = new DbQuery();
+                    $query->select('p.`id_product`')
+                        ->select('NULL as `id_product_attribute`')
+                        ->from('product', 'p')
+                        ->leftJoin('product_shop', 'ps', 'p.`id_product` = ps.`id_product`')
+                        ->where('ps.id_shop = ' . (int) $current_shop_id);
+                    $query->where('p.`id_product` IN (' . implode(', ', array_map('intval', $products)) . ')');
+                    $result = Db::getInstance()->executeS($query);
+                }
```
Progi1984, line 287:
```suggestion
                } else {
                    $result = [['id_product' => 0, 'id_product_attribute' => null]];
                }
```

Progi1984:
@ Thank you for your contribution. :hugs: 

Thanks :+1:

classes/SpecificPriceRule.php
Rolige: @Progi1984 This is not necessary. The `$products` param is used (according to the method description) to filter the products affected by the `SpecificPriceRule`. So, unless an array(0) is passed as value of the `$products` params, doesn't have sense to return an affected product (id_product => 0) this value is not present on the `$products` params.
Additionally, this method `getAffectedProducts` is only called specifying params on the apply function of the same `SpecificPriceRule` class. It is call previous to creating of specific price to the products list returned. If an specific price is already created with `id_product = 0`, then it is not necessary to create again any other rule for a real product (because an specific price with `id_product = 0` is already affecting all products) or for `id_product = 0` (because is already created). This is the main issue that this PR fixs. In ALL the remaining PrestaShop code, the method `getAffectedProducts` is called without params.

Rolige:
> @Rolige Could you rebase your PR for removing two first commits "Merge" ? Thanks

This is absolutely necessary? Because we are using a unique forked project maintained with upstream. If you have a step by step guide to do this without complications please send us.

Rolige:
> @ Thank you for your contribution. 🤗
> 
> Could you explain in [an issue your initial bug for this PR](https://github.com/PrestaShop/PrestaShop/issues/new/choose) (or may be link to an issue) ?
> 
> Thanks 👍

Dona, but I still don't understand why is this necessary, we are repeating the same info.

matks (dismissed):
This PR seems good and the explanations provided by @Rolige are very nice to help.
It's a ✅  for me

classes/SpecificPrice.php
```diff
@@ -188,14 +188,15 @@ public function delete()
         return false;
     }
 
-    public static function getByProductId($id_product, $id_product_attribute = false, $id_cart = false)
+    public static function getByProductId($id_product, $id_product_attribute = false, $id_cart = false, $id_price_rule = false)
```
PierreRambaud:
```suggestion
    public static function getByProductId($id_product, $id_product_attribute = false, $id_cart = false, $id_price_rule = null)
```

classes/SpecificPrice.php
```diff
@@ -188,14 +188,15 @@ public function delete()
...
     {
         return Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS('
 			SELECT *
 			FROM `' . _DB_PREFIX_ . 'specific_price`
 			WHERE `id_product` = ' . (int) $id_product .
             ($id_product_attribute ? ' AND id_product_attribute = ' . (int) $id_product_attribute : '') . '
-			AND id_cart = ' . (int) $id_cart);
+			AND id_cart = ' . (int) $id_cart .
+            ($id_price_rule ? ' AND id_specific_price_rule = ' . (int) $id_price_rule : ''));
```
PierreRambaud:
```suggestion
            ($id_price_rule !== null ? ' AND id_specific_price_rule = ' . (int) $id_price_rule : ''));
```

classes/SpecificPrice.php
PierreRambaud: For a better type hint. Never mix boolean and integer. Prefer a null value instead

PierreRambaud (changes requested):
A little comment about the null value :wink:

Rolige:
> A little comment about the null value 😉

Ok, commit suggestions applied!

PierreRambaud (approved):

atomiix (approved):

SD1982:
LGTM Thanks @Rolige !

matks:
Thank you @Rolige ! 6th PR merged 😉  nice record !

stickyrst:
Just pointing out another issues that this solves:
If you have a specific price rule that applies on all products, and then an algorithm calls

SpecificPriceRule::applyAllRules(array((int) $id_product));

Then the algorithm (as it was before) was deleting the line in specific_price with id_product=0 (so it disabled all the reductions on all products, basically, only leaving the reduction on $id_product).

Thanks!

Issue #18809, opened 2020-04-24
Use placeholders in strings to improve the translation consistency


LouiseBonnard:
#### Is your feature request related to a problem?

Some strings currently contain other strings in it, like _Your data is cached in order to sort it and filter it. You can refresh the cache by clicking on the "Refresh cache" button._. It is not ideal because:

1/ the string in the wording might change,
2/ it is therefore subject to inconsistency, especially in the translation.

So, for the 1.7, here are the non-exhaustive list of wordings I've found.

#### Describe the solution you'd like:

- [x] _There is a huge quantity of data, so each connection corresponding to a referrer is indexed. You can also refresh this index by clicking the "[1]" button. This process may take a while, and it's only needed if you modified or added a referrer, or if you want changes to be retroactive._ in _Admin.Shopparameters.Help_

[1] = _Refresh index_ in _Admin.Shopparameters.Feature_
 

- [x] _Your data is cached in order to sort it and filter it. You can refresh the cache by clicking on the "[1]" button._ in Admin.Shopparameters.Help

[1] = _Refresh cache_ in _Admin.Shopparameters.Feature_

- [x] _You can also decide to move your images to the new storage system. In this case, click on the "[1]" button below. Please be patient. This can take several minutes._ in Admin.Design.Notification
 
[1] = _Move images_ in _Admin.Design.Feature_

- [x] _Step 1: On the left side, select the attributes you want to use (Hold down the "Ctrl" key on your keyboard and validate by clicking on "[1]")_ in _messages_

[1] = _Add_ in _messages_? _Admin.Actions_?

- [x] _If there is no existing module for your carrier, then you can register that carrier by hand using the information that it can provide you: shipping rates, regional zones, size and weight limits, etc. Click on the "[1]" button below to open the Carrier Wizard, which will help you register a new carrier in a few steps._ in _Admin.Shipping.Help_

[1] = _Add new carrier_ in _Admin.Shipping.Feature_

- [x] _This will copy the images too. If you wish to proceed, click "[1]". If not, click "[2]"._ in _Admin.Catalog.Notification_

[1] = _Yes_ in _Admin.Global_?
[2] = _No_ in _Admin.Global_?
 

- [x] _This is the code users should enter to apply the voucher to a cart. Either create your own code or generate one by clicking on "[1]"._ in _Admin.Catalog.Help_

[1] = _Generate_ in _Admin.Actions_?

 
- [x] _Please click on "[1]"._ in _messages_

[...]

PR #19981, opened 2020-06-29
Refactor cancel product and bug fix

### LINKED ISSUE #18432: Refactor DeleteProductFromOrderHandler

### LINKED ISSUE #19717: Wrong order total when cancelling a product on an order having cart rules
#### Describe the bug
There is a bug when canceling a product from an order having cart rules, order total is wrong.

Also this block should not be displayed when canceling a product:
Capture d’écran 2020-06-11 à 15 20 18

And the reduction values table is not updated.

#### Expected behavior
Canceling a product (quantity: one) should have the same impact as deleting it. Cart rules reduction values should be updated and the order total to be paid should be the same.

#### Steps to Reproduce

1. In FO, create an order with 2 products (quantity one for each product), go through completion of the order
2. In FO, create a second order, the exact same order: with the 2 same products and quantity of one for each product.
3. In BO, go the order detail page of the first order you created, and delete one product.
4. In BO, go to the second order you created and cancel one product.

**There are 3 problems:**
- The two orders end up with two different total order values.
- In both cases, the reduction amount table is not updated
- When canceling a product, this block appears (it should only appear for refunds)
Capture d’écran 2020-06-11 à 15 20 18

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Refactor Cancel Product feature to use the OrderProductQuantityUpdater, it also fixes the three bugs described in this PR: #19717
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19717
| How to test?  | See issue #19717

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19981)


atomiix (commented):
I'm not sure about the file `package-lock.json` being changed 🤔

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
```diff
@@ -114,29 +110,17 @@ public function handle(CancelOrderProductCommand $command)
...
-                    if ($order_carrier->update()) {
-                        $order->weight = sprintf('%.3f ' . Configuration::get('PS_WEIGHT_UNIT'), $order_carrier->weight);
-                    }
-                }
-
-                if (Configuration::get('PS_ADVANCED_STOCK_MANAGEMENT') && StockAvailable::dependsOnStock($orderDetail->product_id)) {
-                    StockAvailable::synchronize($orderDetail->product_id);
-                }
+                $newQuantity = (int) $orderDetail->product_quantity - (int) $qty_cancel_product;
+                $this->orderProductQuantityUpdater->update($order, $orderDetail, $newQuantity, null);
```
jolelievre: You use null as the last parameter, but the OrderDetail can have an invoice associated

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
```diff
@@ -114,29 +110,17 @@ public function handle(CancelOrderProductCommand $command)
...
                 Hook::exec('actionProductCancel', ['order' => $order, 'id_order_detail' => (int) $orderDetail->id_order_detail], null, false, true, false, $order->id_shop);
             }
         }
         if (!empty($orderDetails['customizedProductsOrderDetail'])) {
             foreach ($orderDetails['customizedProductsOrderDetail'] as $orderDetail) {
                 $qtyCancelProduct = abs($orderDetails['customizedCancelQuantity'][$orderDetail->id_customization]);
-                $this->orderProductRemover->deleteProductFromOrder($order, $orderDetail, $qtyCancelProduct);
+                $newQuantity = (int) $orderDetail->product_quantity - (int) $qtyCancelProduct;
+
+                $this->orderProductQuantityUpdater->update($order, $orderDetail, $newQuantity, null);
```
jolelievre: Same here, you should provide the invoice if present (it can be generated manually even if the order has not been paid)

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
matthieu-rolland: If the order has an invoice then the user should not be able to perform a "cancel product", this case should never happen.

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
matthieu-rolland: same as above

admin-dev/themes/new-theme/package-lock.json
```diff
@@ -4161,8 +4161,7 @@
                 "ansi-regex": {
```
matthieu-rolland, line 4161: @NeOMakinG when I did my npm install, "optional: true" was added in many places in the package-lock.json, do you think it's ok ?

(thanks @atomiix )

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
```diff
@@ -114,31 +130,23 @@ public function handle(CancelOrderProductCommand $command)
...
-                    $order_carrier->weight = (float) $order->getTotalWeight();
-                    if ($order_carrier->update()) {
-                        $order->weight = sprintf('%.3f ' . Configuration::get('PS_WEIGHT_UNIT'), $order_carrier->weight);
-                    }
-                }
-
-                if (Configuration::get('PS_ADVANCED_STOCK_MANAGEMENT') && StockAvailable::dependsOnStock($orderDetail->product_id)) {
-                    StockAvailable::synchronize($orderDetail->product_id);
-                }
+                $newQuantity = (int) $orderDetail->product_quantity - (int) $qty_cancel_product;
```
jolelievre: Even tough it is checked in OrderProductRemover it might be safer to set a `max(,0)` here as well, to be sure the quantity is never negative

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
```diff
@@ -114,31 +130,23 @@ public function handle(CancelOrderProductCommand $command)
...
+                $orderInvoice = $orderDetail->id_order_invoice != 0 ? new OrderInvoice($orderDetail->id_order_invoice) : null;
+                $this->orderProductQuantityUpdater->update($order, $orderDetail, $newQuantity, $orderInvoice);
                 Hook::exec('actionProductCancel', ['order' => $order, 'id_order_detail' => (int) $orderDetail->id_order_detail], null, false, true, false, $order->id_shop);
             }
         }
         if (!empty($orderDetails['customizedProductsOrderDetail'])) {
             foreach ($orderDetails['customizedProductsOrderDetail'] as $orderDetail) {
                 $qtyCancelProduct = abs($orderDetails['customizedCancelQuantity'][$orderDetail->id_customization]);
-                $this->orderProductRemover->deleteProductFromOrder($order, $orderDetail, $qtyCancelProduct);
+                $newQuantity = (int) $orderDetail->product_quantity - (int) $qtyCancelProduct;
```
jolelievre: Same here `max(,0)`

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
```diff
@@ -31,35 +31,51 @@
...
      *
-     * @param OrderProductRemover $orderProductRemover
+     * @param OrderProductQuantityUpdater $orderProductQuantityUpdater
+     * @param LoggerInterface $logger
+     * @param TranslatorInterface $translator
      */
     public function __construct(
-        OrderProductRemover $orderProductRemover
+        OrderProductQuantityUpdater $orderProductQuantityUpdater,
+        LoggerInterface $logger,
```
Progi1984, line 73: It's a BC with you add new parameters to the construct

Progi1984 (commented):
:warning: It's a BC.

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
matthieu-rolland: But this handler doesn't exist in previous versions :D

admin-dev/themes/new-theme/package-lock.json
NeOMakinG: It should not do anything dangerous, I think it's not a problem at all

matthieu-rolland: thanks for the deedback :)

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
jolelievre: In the adapter namescape BC are usually accepted although they need to be documented
But in this case indeed it's a new class for 177 so no BC here

jolelievre (approved):

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
Progi1984: Ok. Bad one for me. Sorry.

Progi1984 (approved):

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
```diff
@@ -191,4 +199,24 @@ private function checkOrderState(Order $order)
...
+    private function cancelOrder(Order $order)
+    {
+        $history = new OrderHistory();
+        $history->id_order = (int) $order->id;
+        $history->changeIdOrderState(Configuration::get('PS_OS_CANCELED'), $order);
+        if (!$history->addWithemail()) {
+            // email failure must not block order update process
+            $this->logger->warning(
+                $this->translator->trans(
+                    'Order history email could not be sent, test your email configuration in the Advanced Parameters > E-mail section of your back office.',
```
LouiseBonnard, line 215: Hi there, should the 'Advanced Parameters > E-mail' part use placeholders instead? Like we did with issue #18809 with @jolelievre, what do you think?

matthieu-rolland: @LouiseBonnard 
Actually I didn't add a new message, I used the same message that was used in the `OrderProductRemover.php` file.

@SD1982  is currently testing this PR so I can't push code right now, but if you want I can change the message in another PR once this one is merged 👍

LouiseBonnard: Okay, no worries, it can be done later with, perhaps, other wordings of that kind, it was more a question to Jonathan. But please keep in mind that QA is not supposed to test pull requests when the _waiting for wording_ label is still visible or else it complicates the wording review. :-)

PierreRambaud:
I mark the PR a BC breaks

PierreRambaud:
Woot, just notices the class is final. Did it to fast. @Progi1984 the class is final, this is not a BC breaks :)

SD1982:

Progi1984:
Thanks @matthieu-rolland & @SD1982

src/Adapter/Order/CommandHandler/CancelOrderProductHandler.php
matthieu-rolland: yep, I removed the label thinking that since I was moving existing code the wording didn't need to be checked again, I'll keep the waiting for wording label to be sure next time 👍

PR #20016, opened 2020-07-01
Simplify Address form

JevgenijVisockij:
| Questions     | Answers
| Branch?       | develop
| Description?  | Simplify address form
| Type?         | refacto
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | no
| How to test?  | Address form must look and behave exactly as before. There is one exception to this, I changed some of 'Invalid characters' because they weren't accurate before. Also this form impacts not only form in Customers->Address. There is also possibility to add and edit address when creating order in BO. 

:notebook: **BC Break**
Backwards compatibility break introduced due to addition of Router as mandatory constructor argument
as well as extension of TranslationAwareType instead of using translator as dependency. This means if any module extends CustomerAddressType file they will get an exception. 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20016)


src/PrestaShopBundle/Form/Admin/Type/VisibilityChoiceType.php
```diff
@@ -0,0 +1,70 @@
...
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Class CountryChoiceType is responsible for providing country choices with -- symbol in front of array.
+ */
+class VisibilityChoiceType extends AbstractType
+{
```
JevgenijVisockij: This probably needs a better name. So far it only adds visibility param to choice type but may be it will be alerted in the future, but I am not sure what to call it in that case . CustomChoiceType, ImprovedChoiceType?  Neither seems to be an optimal name.

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -106,47 +123,41 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
             $builder->add('id_customer', HiddenType::class);
         }
 
-        $builder->add('phone_mobile', TextType::class, [
+        $builder->add('dni', TextType::class, [
+            'label' => $this->translator->trans('Identification number', [], 'Admin.Orderscustomers.Feature'),
+            'help' => $this->translator->trans(
+                'The national ID card number of this person, or a unique tax identification number.',
+                [],
+                'Admin.Orderscustomers.Help'
```
Junebyun:
```suggestion
            'help' => $this->translator->trans(
                'The national ID card number of this person, or a unique tax identification number.',
                [],
                'Admin.Orderscustomers.Feature'
```

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -87,12 +98,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $data = $builder->getData();
         $countryId = 0 !== $data['id_country'] ? $data['id_country'] : $this->contextCountryId;
+        $genericInvalidCharsMessage = $this->translator->trans(
+            'Invalid characters: %s',
```
LouiseBonnard: Why not keep the string as written elsewhere in the software: _Invalid characters:_ in _Admin.Notifications.Info_?

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -106,47 +123,41 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
             $builder->add('id_customer', HiddenType::class);
         }
 
-        $builder->add('phone_mobile', TextType::class, [
+        $builder->add('dni', TextType::class, [
+            'label' => $this->translator->trans('Identification number', [], 'Admin.Orderscustomers.Feature'),
+            'help' => $this->translator->trans(
+                'The national ID card number of this person, or a unique tax identification number.',
+                [],
+                'Admin.Orderscustomers.Feature'
```
LouiseBonnard, line 138: Also localized in _Admin.Orderscustomers.Help_, I let @Junebyun decide what domain she wants to keep. ;-)

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -288,51 +316,54 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+            ->add('city', TextType::class, [
+                'label' => $this->translator->trans('City', [], 'Admin.Global'),
+                'required' => true,
                 'constraints' => [
-                    new AddressZipCode([
-                        'id_country' => $countryId,
-                        'required' => false,
+                    new NotBlank([
+                        'message' => $this->translator->trans(
+                            'This field cannot be empty', [], 'Admin.Notifications.Error'
```
LouiseBonnard: Or _This field is required_ which might be less negative to read for the user.

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
LouiseBonnard: Following what decision is made, it will have to be harmonized in all occurrences in the PR.

src/Core/ConstraintValidator/TypedRegexValidator.php
```diff
@@ -41,6 +41,10 @@
  */
 class TypedRegexValidator extends ConstraintValidator
 {
+    const GENERIC_NAME_CHARS = '<>={}';
+    const MESSAGE_CHARS = '<>{}';
+    const NAME_CHARS = '0-9!<>,;?=+()@#"�{}_$%:';
```
matks, line 47: I think placeholders in this class are served by `getPattern()` function. See `TypedRegexValidator::getPattern()`

src/PrestaShopBundle/Form/Admin/Type/VisibilityChoiceType.php
```diff
@@ -0,0 +1,70 @@
...
+namespace PrestaShopBundle\Form\Admin\Type;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Class CountryChoiceType is responsible for providing country choices with -- symbol in front of array.
```
matks: Copy-paste comment ?

Looks like something to delete 🤔 😉

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -106,47 +123,41 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                     ]),
                     new ExistingCustomerEmail(),
                 ],
+                'attr' => [
+                    'data-customer-information-url' => $this->router->generate('admin_customer_for_address_information'),
```
matks, line 127: So this was hardcoded into Twig before 😮 ?

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -66,18 +68,27 @@ class CustomerAddressType extends AbstractType
...
      * @param ConfigurableFormChoiceProviderInterface $stateChoiceProvider
+     * @param $contextCountryId
+     * @param RouterInterface $router
      */
     public function __construct(
         TranslatorInterface $translator,
         ConfigurableFormChoiceProviderInterface $stateChoiceProvider,
-        $contextCountryId
+        $contextCountryId,
+        RouterInterface $router
```
matks, line 87: Just like in https://github.com/PrestaShop/PrestaShop/pull/19866 this is a BC break

So we have to check
- can we easily avoid the BC break
=> I dont think so 😅 
- then either we acknowledge the BC break or we find a workaround like creating a new class

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
JevgenijVisockij: Well kinda, not sure if you mean this by hardcoded

 ```
{{ ps.form_group_row(addressForm.customer_email, {'attr': {
            'data-customer-information-url': path('admin_customer_for_address_information') }}, {
            'label': 'Customer email'|trans({}, 'Admin.Orderscustomers.Feature'),
 }) }}

```

File: src/PrestaShopBundle/Form/Admin/Type/VisibilityChoiceType.php (hunk 1 above)
Comment (matks) [reply src/PrestaShopBundle/Form/Admin/Type/VisibilityChoiceType.php]: So this is a form input that can be visible OR invisible due to other parts of the form, is that right ? Like "France" has no state but "USA" has states so if I choose France of USA in the "country" dropdown, this toggles the "states" form.

All of this "toggle input on country input change" behavior is already handled by JS I think ?

If yes, I see 2 options to address your question/issue:
- either we consider it's some extra data but it's not worth it creating a new php class, so we pass it as extra data or as attribute
- either we choose to create a php class ; then maybe ToggleableChoiceType ? it says what it is: "this choice type can be toggled"

File: src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig [hunk 10]
```diff
@@ -53,95 +53,16 @@
...
   </div>
 
   <div class="card-footer">
     <div class="d-inline-flex">
-      {% if (displayInIframe is defined and displayInIframe == true) %}
-        <a href="javascript:window.parent.$.fancybox.close();" class="btn btn-outline-secondary">
-      {% else %}
-        <a href="{{ cancelPath|default(path('admin_addresses_index')) }}" class="btn btn-outline-secondary">
-      {% endif %}
+      <a href="{{ path('admin_addresses_index') }}" class="btn btn-outline-secondary">
```
JevgenijVisockij: @todo need to make sure I didn't break something here

matks (commented):
Reviewed : see feedbacks 😊

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
Junebyun:
```suggestion
                            'This field is required', [], 'Admin.Notifications.Error'
```

I agree!

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
Junebyun: I'd rather keep _Admin.Orderscustomers.Feature_ because most strings of this PR are located in this domain!

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
Junebyun: @matks is the "_%s_" of this string necessary (for listing the characters for instance)? If not, we could keep "_Invalid characters:_", which already exists, like Louise suggested!

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -87,12 +98,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $data = $builder->getData();
         $countryId = 0 !== $data['id_country'] ? $data['id_country'] : $this->contextCountryId;
+        $genericInvalidCharsMessage = $this->translator->trans(
+            'Invalid characters: %s',
+            [TypedRegexValidator::GENERIC_NAME_CHARS],
+            'Admin.Notifications.Info'
```
Junebyun:
```suggestion
            'Invalid characters:',
            [TypedRegexValidator::GENERIC_NAME_CHARS],
            'Admin.Notifications.Info'
```

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -169,6 +180,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 ],
             ])
             ->add('first_name', TextType::class, [
+                'label' => $this->translator->trans('First name', [], 'Admin.Global'),
+                'help' => $this->translator->trans(
+                    'Invalid characters: %s',
+                    [TypedRegexValidator::NAME_CHARS],
+                    'Admin.Notifications.Info'
```
Junebyun:
```suggestion
                    'Invalid characters:',
                    [TypedRegexValidator::NAME_CHARS],
                    'Admin.Notifications.Info'
```

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -191,6 +208,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 ],
             ])
             ->add('last_name', TextType::class, [
+                'label' => $this->translator->trans('Last name', [], 'Admin.Global'),
+                'help' => $this->translator->trans(
+                    'Invalid characters: %s',
+                    [TypedRegexValidator::NAME_CHARS],
+                    'Admin.Notifications.Info'
```
Junebyun:
```suggestion
                    'Invalid characters:',
                    [TypedRegexValidator::NAME_CHARS],
                    'Admin.Notifications.Info'
```

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -371,8 +409,32 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                    ]),
+                ],
+            ])
             ->add('other', TextareaType::class, [
                 'required' => false,
+                'label' => $this->translator->trans('Other', [], 'Admin.Global'),
+                'help' => $this->translator->trans(
+                    'Invalid characters: %s',
+                    [TypedRegexValidator::MESSAGE_CHARS],
+                    'Admin.Notifications.Info'
```
Junebyun:
```suggestion
                    'Invalid characters:',
                    [TypedRegexValidator::MESSAGE_CHARS],
                    'Admin.Notifications.Info'
```

src/Core/ConstraintValidator/TypedRegexValidator.php
JevgenijVisockij: Indeed it seems they do. Problem is that its a private function. Would it make sense to make it public and add any non-existing patterns there?

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
JevgenijVisockij: So following the same principle as in the other PR I am gonna leave the comment about BC break

src/PrestaShopBundle/Form/Admin/Type/VisibilityChoiceType.php
JevgenijVisockij: I added it as attribute as I think it's cleaner this way. But there is one reason I dislike it: it's not clear that you can just add visible => false to attributes and make whole block invisible. However there is no way to know this other then look up the the prestashop_ui_kit.html.twig file or find example where it has been used. 

Any way we can indicate what options are available in attr parameter?

Junebyun (dismissed):

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
matks: I meant the URL was generated into Twig, directly, using `path`.
So yes it was 😄

src/Core/ConstraintValidator/TypedRegexValidator.php
matks: Yes it makes sense to make it public. It could then be used by external classes to fetch a pattern they need 👍

src/PrestaShopBundle/Form/Admin/Type/VisibilityChoiceType.php
matks: > Any way we can indicate what options are available in attr parameter?

None than I know. But I think code is straightforward here, it's not an issue.

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
matks: Was it for you or for me 😄  ?

src/PrestaShopBundle/Form/Admin/Sell/Address/CustomerAddressType.php
```diff
@@ -67,17 +69,30 @@ class CustomerAddressType extends AbstractType
...
 
     /**
+     * @var RouterInterface
+     */
+    private $router;
+
+    /**
+     * CustomerAddressType constructor.
+     *
+     * Backwards compatibility break due to addition of Router
```
matks:
```suggestion
     * Backwards compatibility break introduced in 1.7.8.0 due to addition of Router as mandatory constructor argument
```

matks (commented):
New review, almost perfect !
Need 2 things to be fixed:
- https://github.com/PrestaShop/PrestaShop/pull/20016/files#r452803653
- https://github.com/PrestaShop/PrestaShop/pull/20016#discussion_r452802425

src/PrestaShopBundle/Resources/views/Admin/Sell/Address/Blocks/form.html.twig
JevgenijVisockij: For me, don't worry about it :D

admin-dev/themes/new-theme/js/pages/address/form.js
```diff
@@ -40,11 +40,15 @@ $(document).ready(() => {
       company: addressFormMap.addressCompanyInput,
     },
   );
+
+  const addressStateBlock = $(addressFormMap.addressStateBlock).closest('.form-group.row');
```
PierreRambaud: Not a good idea. If someone changes the parent, it will break the js.
Change the addressStateBlock selector instead.

src/Core/ConstraintValidator/TypedRegexValidator.php
JevgenijVisockij: I investigated this further and there are problems with using getPattern function. For example TYPE_GENERIC_NAME returns pattern of /^[^<>={}]*$/u of course i only need <>={} part for description. 

Now I see three ways of solving this:

1. Create a function getPatternDescription which would remove all the regex stuff and leave pattern chars only. It would be pretty convenient but I couldn't find a nice way to do it, writing a big regex to consider all possible cases could make it pretty error-prone. 
2. Use the same function but hardcode those patterns instead I.E 
```
$typePatterns = [
       TypedRegex::TYPE_GENERIC_NAME => '<>={}'
  ]
  if (isset($typePatterns[$type])) {
       return $typePatterns[$type];
  }
  throw new InvalidArgumentException();
]
```

 Main disadvantage here is that if I need a pattern I need to have TypedRegexValidator as dependancy

3. Leave it the way it's done above the comment. Disadvantage is that constants are not that as intuitive as having a function based on the constants like TypedRegex::TYPE_GENERIC_NAME (which are already used when adding regex constraint), however we also don't need extra dependency if we want to get patterns. 

I think 1 would be the best way to do it if there was an easy way to remove all the regex stuff and leave only a pattern and getPatternDescription would be very simple `return $this->cleanRegex($this->getPattern($type)) `. If not I would go with 3 as it avoids an unnecessary dependency.

What do you think? @matks

matks: Let's go for 3. because we are not needing regexps here.

In future, it would be nice to implement 1. and make it a static function (and maybe make `getPattern($type)` static too because we dont need it to be an instance so, as you said, it's a shame to need a dependency injection just for that.

matks:
Code is good for me ✅  but you need to apply @PierreRambaud feedback

Progi1984 (dismissed):

PierreRambaud (dismissed):

Progi1984:
@JevgenijVisockij Could you explain in the PR the BC you have introduced for the release note ? Thanks

Progi1984:
I thought here : https://github.com/PrestaShop/PrestaShop/pull/20016#issue-442703929

admin-dev/themes/new-theme/js/pages/address/form.js
JevgenijVisockij: Fixed by adding class to the whole row via 
```
 'row_attr' => [
      'class' => 'js-address-state-select',
],
```

And using that class as selector.

JevgenijVisockij:
> I thought here : [#20016 (comment)](https://github.com/PrestaShop/PrestaShop/pull/20016#issue-442703929)

Added it at the bottom of the comment. Tried adding it to "BC breaks?" column but it kept breaking design.

Progi1984:
@JevgenijVisockij Perfect for me.

Progi1984 (approved):

matks (approved):

PierreRambaud (approved):

khouloudbelguith:
Hi @JevgenijVisockij,

It is ok :heavy_check_mark: 
But we need a confirmation from the Product Team about the New design

**Without PR**
image

**With PR**
image

Ping @PrestaShop/prestashop-product-team what do you think?

Thanks!

matks:
@TristanLDD @prestascott do you agree on this little red border ?

(I find it a bit heavy we have to validate each little detail like this 🙄  )

matks:
@marionf @khouloudbelguith Hi ! I understand you want to have a double check with UX team but while we wait for it, the chances that this PR [enters in a conflict state](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-merge-conflicts) that will need a rebase. A rebase means it will wipe out the approvals so it will need, at least, 2 new code review approvals.

Moreover if there is a need for a UX change, we will need to implement it, passe the double code review phase, pass again the QA phase (maybe the wording validation again ? 🙏 ) all of this for ... a small red border 😅  .

This scenario does not seem like a good investment of our time and energy so I take the (foolish) liberty to merge this PR as  Khouloud validated the behavior.

But even if I merge the PR, it does not mean we will not implement possible UX feedbacks ⚠️  . If there is a need for a small UX change it will be done, just not in the scope of this PR 😉 . I promise 😄

matks:
Thank you @JevgenijVisockij

matks (approved):

TristanLDD:
@matks The only thing that is missing is a 8px spacing between the alert and the input field.

matks:
> @matks The only thing that is missing is a 8px spacing between the alert and the input field.

Got it ! it will be done

TristanLDD:
@matks erratum!! I was looking a bit too fast sorry, this is the style of alerts but contextual input alerts are actually different, they must look like Khouloud's screenshot: no borders and only icon and text. Sorry for that!

matks:
@TristanLDD should be fixed by https://github.com/PrestaShop/PrestaShop/pull/20305 😉

PR #20083, opened 2020-07-07
Refactoring of the Order Updater and multiple related bugfixes

### LINKED ISSUE #19890: Update product's price in the order does not update the discount

#### Describe the bug

On the back-office, when I create a percentage discount and then change the product's price on the order, the discount is not updated.

#### Expected behavior

The discount should apply the 50% reduction on all the product amount even if after they are updated 

#### Steps to Reproduce

1. Add the product to any order
2. Create a cart rule of 50% in the order page
3. edit the product price
4. See error - wrong discount total

**Screenshots**

#### Additional information

* PrestaShop version: N/A
* PHP version: N/A

### LINKED ISSUE #19949: BO - order detail / Discount is deleted when deleting another discount
#### Describe the bug

In an order we have a product with discount of 50% on the total of the order cumulated with another discount . If we delete the product the 2 discount are deleted

#### Expected behavior

The discount not related to the deleted product should not be deleted

#### Steps to Reproduce

1. Create a cart rule with 50% reduction on the total of the order with no code specifically and applied to a product
2. In FO create an order with any product and checkout
3. In Bo go to the previously created order and add the product with the rule => the 50 % discount is applied 
4. Add another discount ( for exemple 5% ) => the 5 % discount is applied
5. Delete the product related to the 50% discount
6. See error => the 2 discount are deleted

**Screenshots**

https://drive.google.com/file/d/1rVEr6K9cAJqfv2K-kSTsK5Jjkhd3eppW/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: 7.2

### LINKED ISSUE #19960: [1.7.7] Order BO discount removing/changing not working properly
Removing discounts or changing product quantity is not working.

#### Video
https://drive.google.com/file/d/1e2-2veDRzWT45niwQJLlwOpc7waDWF5Y/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: 7.2

### LINKED ISSUE #20162: [EPIC] Wrong discount on the order page
To facilitate the tracking in the kanban, its issues are grouped in an epic.

- [x] #19960 - Order BO discount removing/changing not working properly 
- [x] #19949 - BO - order detail / Discount is deleted when deleting another discount
- [x] #19890 - Update product's price in the order does not update the discount

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | This PR originally intended to fix a few bugs, but I realized a few things needed some refacto so it includes quite a lot of modifications:<br>- **CartRule:checkValidity**: some rules were considered invalid when auto removed because `$alreadyInCart` was ignored in most cases<br>- CartRule addition/removal on order was performed **manually**, so any future computations led to invalid amounts, now it insert/delete a CartRule instance in DB and then the amounts are re-computed (which fixes a lot of issues when performing multiple edition in the BO)<br>- the Order discount feature was false, it computed the total amount (so shipping fees were included in the computations)<br>- same problem happened with Invoices udpates that were updated manually<br>- the OrderAmountUpdater didn't use the carrier in its computation, it now handles weight and shipping cost updates correctly<br>- all the modifs are now performed thanks to OrderAmountUpdater which is used in all the handlers, it prevents from having different behaviour and allows to fix the potential problems in ONE place
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19949 #19960 #19890 (EPIC - Wrong discount on the order page #20162)
| How to test?  | You can look into the issues for more details (and sample videos), but basically try and perform multiple edition in an Order, by adding various discounts and products with associated discounts, try and remove discounts separately, try and remove products separately IMPORTANT NOTE: when a global discount is applied on all order you can not remove it any more until it's activated (it is actually removed and added again) Which makes sense because it respects the Shop cart rules, however from a merchant point of view this needs to be validated by the @PrestaShop/prestashop-product-team 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20083)

### DIFF
--- a/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
@@ -67,7 +67,7 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
      */
     public function handle(AddCartRuleToOrderCommand $command): void
     {
-        $this->validatePercentCartRule($command);
+        $this->assertPercentCartRule($command);
         $order = $this->getOrder($command->getOrderId());
 
         $computingPrecision = new ComputingPrecision();
@@ -82,8 +82,8 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
                 throw new OrderException('Can\'t load Order Invoice object');
             }
         }
-        $this->validateAmountCartRule($command, $order, $orderInvoice);
-        $this->validateFreeShippingCartRule($command, $order, $orderInvoice);
+        $this->assertAmountCartRule($command, $order, $orderInvoice);
+        $this->assertFreeShippingCartRule($command, $order, $orderInvoice);
 
         $cart = Cart::getCartByOrderId($order->id);
         $cartRuleObj = new CartRule();
@@ -135,7 +135,7 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
      *
      * @throws InvalidCartRuleDiscountValueException
      */
-    private function validatePercentCartRule(AddCartRuleToOrderCommand $command): void
+    private function assertPercentCartRule(AddCartRuleToOrderCommand $command): void
     {
         if (OrderDiscountType::DISCOUNT_PERCENT !== $command->getCartRuleType()) {
             return;
@@ -162,7 +162,7 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
      *
      * @throws InvalidCartRuleDiscountValueException
      */
-    private function validateAmountCartRule(AddCartRuleToOrderCommand $command, Order $order, ?OrderInvoice $orderInvoice): void
+    private function assertAmountCartRule(AddCartRuleToOrderCommand $command, Order $order, ?OrderInvoice $orderInvoice): void
     {
         if (OrderDiscountType::DISCOUNT_AMOUNT !== $command->getCartRuleType()) {
             return;
@@ -207,7 +207,7 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
      *
      * @throws InvalidCartRuleDiscountValueException
      */
-    private function validateFreeShippingCartRule(AddCartRuleToOrderCommand $command, Order $order, ?OrderInvoice $orderInvoice): void
+    private function assertFreeShippingCartRule(AddCartRuleToOrderCommand $command, Order $order, ?OrderInvoice $orderInvoice): void
     {
         if (OrderDiscountType::FREE_SHIPPING !== $command->getCartRuleType()) {
             return;
--- a/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
@@ -30,11 +30,9 @@ use Cart;
 use CartRule;
 use Configuration;
 use Currency;
-use OrderCartRule;
 use OrderInvoice;
 use PrestaShop\PrestaShop\Adapter\Order\AbstractOrderHandler;
 use PrestaShop\PrestaShop\Adapter\Order\OrderAmountUpdater;
-use PrestaShop\PrestaShop\Core\Domain\CartRule\Exception\InvalidCartRuleDiscountValueException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Command\AddCartRuleToOrderCommand;
 use PrestaShop\PrestaShop\Core\Domain\Order\CommandHandler\AddCartRuleToOrderHandlerInterface;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
@@ -124,6 +122,6 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
             throw new OrderException('An error occurred while adding CartRule to cart', 0, $e);
         }
 
-        $this->orderAmountUpdater->update($order, $cart, null !== $orderInvoice);
+        $this->orderAmountUpdater->update($order, $cart, null !== $orderInvoice ? (int) $orderInvoice->id : null);
     }
 }
--- a/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
@@ -26,6 +26,7 @@
 
 namespace PrestaShop\PrestaShop\Adapter\Order\CommandHandler;
 
+use Cart;
 use CartRule;
 use Configuration;
 use Currency;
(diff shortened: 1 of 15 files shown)


src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -28,49 +28,60 @@
...
+     * @throws PrestaShopDatabaseException
+     * @throws PrestaShopException
      */
     public function update(
         Order $order,
         Cart $cart,
-        bool $hasInvoice
-    ): Order {
+        ?int $orderInvoiceId
+    ): void {
```
PululuK, line 77: @jolelievre  it is a BC ... right ?

matthieu-rolland: It would indeed have been a BC if the class `OrderAmountUpdater` already existed in previous prestashop versions, but this class was introduced in 1.7.7, which is not released yet, so we're fine 👍

jolelievre: Yep, it's the last occasion we have to make it right ^^

matthieu-rolland (dismissed):

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -46,269 +49,191 @@
...
-                        $cartRules[$orderInvoice->id]['value_tax_excl'] = Tools::ps_round(
-                            $discountValue / (1 + ($order->getTaxesAverageUsed() / 100)),
-                            $precision
-                        );
+    /**
+     * @param AddCartRuleToOrderCommand $command
+     *
+     * @throws InvalidCartRuleDiscountValueException
+     */
+    private function validatePercentCartRule(AddCartRuleToOrderCommand $command): void
```
Progi1984: Please use naming `assert`.

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -46,269 +49,191 @@
...
-                    );
-                }
+    /**
+     * @param AddCartRuleToOrderCommand $command
+     * @param Order $order
+     * @param OrderInvoice|null $orderInvoice
+     *
+     * @throws InvalidCartRuleDiscountValueException
+     */
+    private function validateAmountCartRule(AddCartRuleToOrderCommand $command, Order $order, ?OrderInvoice $orderInvoice): void
```
Progi1984: Please use naming `assertAmountCartRule`

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -28,62 +28,274 @@
...
+        if (empty($invoiceProducts)) {
+            return;
+        }
+
+        $firstInvoice = $order->getInvoicesCollection()->getFirst();
+        $invoice = new OrderInvoice($orderInvoiceId);
+
+        // Shipping are computed on first invoice only
+        $carrierId = $order->id_carrier;
+        $totalMethod = $firstInvoice === null || $firstInvoice->id == $invoice->id ? Cart::BOTH : Cart::BOTH_WITHOUT_SHIPPING;
```
Progi1984:
```suggestion
        $totalMethod = ($firstInvoice === null || $firstInvoice->id == $invoice->id) ? Cart::BOTH : Cart::BOTH_WITHOUT_SHIPPING;
```

Progi1984 (commented):
Some comments

jolelievre:
@Progi1984 thanks for the comments, all addressed 😉

Progi1984 (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1515,13 +1515,28 @@ private function getErrorMessages(Exception $e)
...
+                    'Admin.Orderscustomers.Notification'
+                ),
+                InvalidCartRuleDiscountValueException::INVALID_MAX_PERCENT => $this->trans(
+                    'Percent value cannot exceed 100.',
+                    'Admin.Orderscustomers.Notification'
+                ),
+                InvalidCartRuleDiscountValueException::INVALID_MIN_AMOUNT => $this->trans(
+                    'Amount value must be greater than 0.',
+                    'Admin.Orderscustomers.Notification'
+                ),
```
Junebyun, line 1530: Hi @jolelievre ! Is it necessary to have _"Percent value must be greater than 0."_ and _"Amount value must be greater than 0."_ both texts separated? 

If not, we can replace them by the string **_"Percent or amount value must be greater than 0."_** in _Admin.Orderscustomers.Notification_

jolelievre: That was the initial text but it was used in a list because we had no way to know what was the cause of the error. We now have a much more precise exception which indicates precisely if the problem comes from a value or a percent.

I can use the same error message for both errors though if you prefer, I just thought more accurate error messages were better for the user

Junebyun: @jolelievre indeed having accurate error messages are better for users! Let's keep both messages separated then. Thanks for explaining the context :)

LouiseBonnard (approved):

Progi1984 (approved):

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -28,62 +28,274 @@
...
 {
+    /**
+     * @var ConfigurationInterface
+     */
+    private $configuration;
+
+    /**
+     * @param ConfigurationInterface $configuration
+     */
+    public function __construct(ConfigurationInterface $configuration)
```
PierreRambaud, line 59: Isn't a BC, the constructor is now mandatory :thinking:

PierreRambaud (approved):
Everything seems ok for me, just a question about a BC breaks

Progi1984 (approved):

src/Adapter/Order/OrderAmountUpdater.php
jolelievre: The class is new in 177 so no BC break here

matthieu-rolland (approved):

khouloudbelguith:
Hi @jolelievre,

In my case, I checked the 3 issues, they are fixed.
But, I detected that this [issue](https://github.com/PrestaShop/PrestaShop/issues/18379) is fixed previously but with this PR, the problem gets back.
So we have an issue following these steps:

2. Create a cart rule:
2.1. has no code
2.2. product selection on the previous created product with 100 euro tax included
2.3. the discount amount of 500 EUR tax included
4. The discount has been added to the order
5. try to remove the discount => impossible

https://drive.google.com/file/d/1EWqM6DyjdPlp1-0XuKnmaQPZrXSLPpGg/view?usp=sharing
Thanks!

matks:
@khouloudbelguith as this PR is complex, solves 3 issues, and bring multiple changes into `1.7.7.x`, we merge it but we will immediately create a new PR to fix your feedback

matks:
Thank you @jolelievre

PR #20096, opened 2020-07-08
Show last visible order state in customer order history page

### LINKED ISSUE #18892: Show last visible status order on order history page
**EDIT**
On front office > customer order history page, show the last visible order status.
At some point of the order flow, it might get a specific order status that the merchant decides not to display to the customer (option available in the status configuration). When the order gets this status, nothing is displayed in the customer account, which can create confusion for the customer.
With this improvement, it will show in the customer account the last status the order had before getting this "invisible" status.

Edge case : if the first status an order gets is "invisible" for the customer, then show nothing.

**Initial description :**
We received some complains from customer that can't see the order-status on the order history page. Our reply was always its there, or click on details. I never understood why they where not seeing anything until I realised that we use statuses that are invisible for the customers.  Once an order is shipped, we change status for administrations reasons. The customer does not need to see this status. 

When the status is not there, customers think that their order was not processed  and think they have to reorder. It causes a bit of confusion.

I attached a screenshot with two orders. One order is cancelled, the other is put in to the next status phase. As you can see its empty ( status will show up once you click on details ).  

Request: show on the order history page the last visible status for customer. Just like in the order detail page.

Screenshot 2020-04-29 at 12 59 04

Prestashop 1.7.6.5

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | When an order has a state that is invisible to customer, in the customer order history page, we should display the last visible order state instead of displaying an empty field.
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18892
| How to test?  | **1/** In BO > Shop Parameters > Orders > Order States, create a new state, and check "hide this state for customers"<br>**2/** In FO, do a complete order.<br>**3/** In BO, go to the created order and change its state, use the one you just created<br>**4/** Go in FO > Order History, check the current state of the order: it should not be empty, it should not display the state you just set (because it's invisible to customers), instead it should display the last visible order state.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20096)

### DIFF
--- a/src/Adapter/Presenter/Order/OrderLazyArray.php
+++ b/src/Adapter/Presenter/Order/OrderLazyArray.php
@@ -298,6 +298,8 @@ class OrderLazyArray extends AbstractLazyArray
         $historyList = $order->getHistory($context->language->id, false, true);
 
         foreach ($historyList as $historyId => $history) {
+            // HistoryList only contains order states that are not hidden to customers, the last visible order state,
+            // that is to say the one we get in the first iteration
             if ($historyId === array_key_first($historyList)) {
                 $historyId = 'current';
             }
--- a/src/Adapter/Presenter/Order/OrderLazyArray.php
+++ b/src/Adapter/Presenter/Order/OrderLazyArray.php
@@ -298,7 +298,7 @@ class OrderLazyArray extends AbstractLazyArray
         $historyList = $order->getHistory($context->language->id, false, true);
 
         foreach ($historyList as $historyId => $history) {
-            if ($history['id_order_state'] == $order->current_state) {
+            if ($historyId === array_key_first($historyList)) {
                 $historyId = 'current';
             }
             $orderHistory[$historyId] = $history;


Progi1984 (approved):

matks (approved):

khouloudbelguith:
Hi @matthieu-rolland,

If you follow these steps, the current state of the order is empty:
1. Go to BO > Shop Parameters > Orders > Order States
2. Edit the first status "Awaiting check payment" and check "hide this state for customers"
image
3. In FO, do a complete order with "Pay by Check"
4. Go in FO > Order History, check the current state of the order: it is empty
image

Thanks!

matthieu-rolland:
thank you @khouloudbelguith for your feedback ! 

I'm not sure what should be done here, the goal of this PR is to display the last visible order state, if the first order state assigned is invisible, then it makes sense to not display it. I think it's the merchant's responsibility to not hide order states that are the first ones to be assigned to an order.

I don't know what would be the alternative, define a default order state for when no visible order state has been assigned yet?

@PrestaShop/prestashop-product-team what do you guys think ?

colinegin:
I agree with you @matthieu-rolland it makes no sense to hide the first order status. Usually merchants need to hide status that help them prepare and send the order, not the one related to payment.

My recommendation is to leave the field empty if the merchant chooses to hide the first order status.
On all other cases, the last visible order statuses will be displayed.

Is that ok for you @matthieu-rolland @khouloudbelguith ?

matthieu-rolland:
@khouloudbelguith except this matter, was the PR ok for you?

khouloudbelguith:
Hi @matthieu-rolland,

It is ok :heavy_check_mark: 

Thanks!

01park:
Hi Guys,

I am just reading this topic. 

The reason why I created this post is because once the customer clicks on details behind the order, then the full order history is visible. The question was, can from this list the last visible status be displayed in the order history list. Maybe mirror it?

Because now, lets say the status is put to shipped and this is visible. The customer can see its shipped. But then if we change it to an invisible status, the customer wil see nothing, while in my option  if status shipped stayed, it would cause less confusion. 

This is feedback btw I got from my customers.

colinegin:
@khouloudbelguith did you test also the order details page ?

khouloudbelguith:
Hi @colinegin,

Yes!
In the BO
image
In the FO => Order details page
image

In this case, the "test" status is Hidden and the "Awaiting check payment" is Hidden also.

Thanks!

matthieu-rolland:
> Hi Guys,
> 
> I am just reading this topic.
> 
> The reason why I created this post is because once the customer clicks on details behind the order, then the full order history is visible. The question was, can from this list the last visible status be displayed in the order history list. Maybe mirror it?
> 
> Because now, lets say the status is put to shipped and this is visible. The customer can see its shipped. But then if we change it to an invisible status, the customer wil see nothing, while in my option if status shipped stayed, it would cause less confusion.
> 
> This is feedback btw I got from my customers.

this is exactly what we understood, and what this PR does, no worries 👍

PR #20104, opened 2020-07-08
[WIP] Simplified credit slip options

JevgenijVisockij:
| Questions     | Answers
| Branch?       | develop
| Description?  | Simplify credit slip options
| Type?         | refacto
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | no
| How to test?  | Form must work and look exactly the same as before

:notebook: **BC Break**
Backwards compatibility break introduced due to extension of TranslationAwareType instead of using translator as dependency. This means if any module extends CreditSlipsOptionsType or GeneratePdfByDateType files they will get an exception.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20104)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Order/CreditSlipController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Order/CreditSlipController.php
@@ -84,7 +84,6 @@ class CreditSlipController extends FrameworkBundleAdminController
         $pdfByDateForm = $this->createForm(GeneratePdfByDateType::class, [], [
             'method' => Request::METHOD_GET,
         ]);
-        $pdfByDateForm->handleRequest($request);
 
         return $this->render('@PrestaShop/Admin/Sell/Order/CreditSlip/index.html.twig', [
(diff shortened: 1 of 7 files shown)


JevgenijVisockij:
It's still WIP because I get an error 

"The CSRF token is invalid. Please try to resubmit the form." when trying to generate pdf but there are no credit slips in date range.

src/PrestaShopBundle/Controller/Admin/Sell/Order/CreditSlipController.php
```diff
@@ -84,7 +84,6 @@ public function indexAction(
         $pdfByDateForm = $this->createForm(GeneratePdfByDateType::class, [], [
             'method' => Request::METHOD_GET,
         ]);
-        $pdfByDateForm->handleRequest($request);
```
JevgenijVisockij, line 87: Removing because it tried to handle generatePpdfByDate form in index page, when it should only do in it's own action.

JevgenijVisockij:
Closed in favor of #20204

Issue #14774, opened 2019-07-19
[EPIC] Product page - Shipping tab 6/9


marionf:
### - As a merchant I want to be able to manage carriers information for my products

[https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md##shipping-tab-epic](https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#shipping-tab-epic) 

**This tab is not displayed for a virtual product.**

- [ ] #19666 - [Shipping tab] - Package dimension 1/4
- [ ] #19667 - [Shipping tab] - Delivery Time 2/4
- [ ] #19668 - [Shipping tab] - Shipping fees 3/4
- [ ] #19669 - [Shipping tab ] - Available carriers 4/4

### [- As a merchant I want to know if there are errors when I complete a field in the product page](https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#shipping-tab-)

---

### - Issues in **shipping tab** sorted by priority high to low :

**Should-Have**

**Improvement**
- [x] #15377 - Available carriers for a product
- [x] #10759 - Delivery Time, confusing UI in back-office

**Bugs**
- [ ] #24497 - List of available carriers in product edit page contain disabled carriers - product v1

## Fixed
- [x] #24687 - Precise product weight is not correctly displayed in back-office
- [x] #11379 - Multiple carriers with same name & delay key will not be display PR: #12379

marionf:
### - Issues in **shipping tab** sorted by priority high to low :

**Should-Have**

**Improvement**
- [ ] #15377

## Fixed
- [x] #11379 
Work already started: https://github.com/PrestaShop/PrestaShop/pull/12379

PR #20107, opened 2020-07-08
Add UpdateProductShipping command

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Implement command to update product shipping options. Cover basic use-cases with behat scenarios
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | yes
| Fixed ticket? | part of #14774
| How to test?  | none

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20107)

### DIFF
--- a/src/Adapter/Product/CommandHandler/UpdateProductShippingHandler.php
+++ b/src/Adapter/Product/CommandHandler/UpdateProductShippingHandler.php
@@ -32,6 +32,7 @@ use PrestaShop\PrestaShop\Adapter\Product\AbstractProductHandler;
 use PrestaShop\PrestaShop\Core\Domain\Product\Command\UpdateProductShippingCommand;
 use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\UpdateProductShippingHandlerInterface;
 use PrestaShop\PrestaShop\Core\Domain\Product\Exception\CannotUpdateProductException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
 use Product;
 
 /**
@@ -58,26 +59,31 @@ final class UpdateProductShippingHandler extends AbstractProductHandler implemen
     {
         if (null !== $command->getWidth()) {
             $product->width = (string) $command->getWidth();
+            $this->validateField($product, 'width', ProductConstraintException::INVALID_WIDTH);
             $this->fieldsToUpdate['width'] = true;
         }
 
         if (null !== $command->getHeight()) {
             $product->height = (string) $command->getHeight();
+            $this->validateField($product, 'height', ProductConstraintException::INVALID_HEIGHT);
             $this->fieldsToUpdate['height'] = true;
         }
 
         if (null !== $command->getDepth()) {
             $product->depth = (string) $command->getDepth();
+            $this->validateField($product, 'depth', ProductConstraintException::INVALID_DEPTH);
             $this->fieldsToUpdate['depth'] = true;
         }
 
         if (null !== $command->getWeight()) {
             $product->weight = (string) $command->getWeight();
+            $this->validateField($product, 'weight', ProductConstraintException::INVALID_WEIGHT);
             $this->fieldsToUpdate['weight'] = true;
         }
 
         if (null !== $command->getAdditionalShippingCost()) {
             $product->additional_shipping_cost = (string) $command->getAdditionalShippingCost();
+            $this->validateField($product, 'additional_shipping_cost', ProductConstraintException::INVALID_ADDITIONAL_SHIPPING_COST);
             $this->fieldsToUpdate['additional_shipping_cost'] = true;
         }
 
--- a/src/Adapter/Product/CommandHandler/UpdateProductShippingHandler.php
+++ b/src/Adapter/Product/CommandHandler/UpdateProductShippingHandler.php
@@ -31,6 +31,7 @@ namespace PrestaShop\PrestaShop\Adapter\Product\CommandHandler;
 use PrestaShop\PrestaShop\Adapter\Product\AbstractProductHandler;
 use PrestaShop\PrestaShop\Core\Domain\Product\Command\UpdateProductShippingCommand;
 use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\UpdateProductShippingHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\CannotUpdateProductException;
 use Product;
 
 /**
@@ -46,7 +47,7 @@ final class UpdateProductShippingHandler extends AbstractProductHandler implemen
         $product = $this->getProduct($command->getProductId());
         $this->fillUpdatableFieldsWithCommandData($product, $command);
 
-        //@todo: performUpdate after rebase.
+        $this->performUpdate($product, CannotUpdateProductException::FAILED_UPDATE_SHIPPING_OPTIONS);
     }
 
     /**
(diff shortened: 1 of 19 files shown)


tests/Integration/Behaviour/Features/Context/Domain/CarrierFeatureContext.php
```diff
@@ -0,0 +1,72 @@
...
+use Carrier;
+use Configuration;
+use Context;
+use Group;
+use PrestaShopException;
+
+class CarrierFeatureContext extends AbstractDomainFeatureContext
+{
+    /**
+     * @todo: It is a temporary method to use sharedStorage and should be improved once Carrier creation is migrated.
```
matks, line 40: 👍

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -467,6 +559,12 @@ public function assertProductFields(string $productReference, TableNode $table)
 
         $this->assertTaxRulesGroup($data, $productForEditing);
         $this->assertPriceFields($data, $productForEditing->getPricesInformation());
+        $this->assertShippingInformation($data, $productForEditing->getShippingInformation());
+
+        // Assertions checking isset() can hide some errors if it doesn't find array key,
```
matks, line 569: 👍

matks (dismissed):
Code is good. But I discover the concept of "delivery time note". What is this thing ? 😅

zuk3975:
> Code is good. But I discover the concept of "delivery time note". What is this thing ?

Basically i just used another name for legacy properties to reflect the use-case.

There are following properties in Product obj model (and database):
1. $additional_delivery_times
2. $delivery_in_stock
3. $delivery_out_stock

By the names you couldn't guess what are their purpose.
As I found out the `2` and `3` are some kind of notes, which appears in front office product details and should tell delivery time depending if product is in stock or out of stock (eg: Delivered within 3-4 days).
The `1` one turns out to be a setting, which tells if those above-mentioned notes should be specific for this product (which are the `2` and `3`, or taken from default settings (`admin-dev/configure/shop/product-preferences/` -> `Product stock`), or turned off.

So in domain i used abit tweaked names to better reflect their use-case
I called them accordingly:
1. $deliveryTimeNotesType
2. $deliveryTimeInStockNotes
3. $deliveryTimeOutOfStockNotes

EDIT: @matks, wdyt ? (You offered some similar renaming regarding "notes" in slack chat, do you still think I could rename it to something more relevant?)

src/Adapter/Product/CommandHandler/UpdateProductShippingHandler.php
```diff
@@ -0,0 +1,119 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
```
Progi1984, line 3: Please use new license headers

Progi1984 (commented):
Check comments

src/Adapter/Product/CommandHandler/UpdateProductShippingHandler.php
zuk3975: I've updated headers in my editor. 
Not gonna update them manually now, as Im aware there is a command for that, which will be run at some point later to update all headers isn't it?

matks: Indeed. @Progi1984 I regularly run the command to update the codebase and make sure it stays fine 😉

matks (approved):

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -257,6 +261,94 @@ public function updateProductOptions(string $productReference, TableNode $table)
...
+     *
+     * @return array values that was provided, but wasn't handled
+     */
+    private function setUpdateShippingCommandData(array $data, UpdateProductShippingCommand $command): array
+    {
+        $unhandledValues = $data;
+
+        if (isset($data['width'])) {
+            $command->setWidth($data['width']);
+            unset($unhandledValues['width']);
```
jolelievre, line 302: Oohh, that's a really nice improvement We should add this kind of checks in other tests 👍

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
```diff
@@ -472,6 +564,12 @@ public function assertProductFields(string $productReference, TableNode $table)
 
         $this->assertTaxRulesGroup($data, $productForEditing);
         $this->assertPriceFields($data, $productForEditing->getPricesInformation());
+        $this->assertShippingInformation($data, $productForEditing->getShippingInformation());
+
+        // Assertions checking isset() can hide some errors if it doesn't find array key,
+        // to make sure all provided fields were checked we need to unset every asserted field
+        // and finally, if provided data is not empty, it means there are some unnasserted values left
+        Assert::assertEmpty($data, sprintf('Some provided fields haven\'t been asserted: %s', implode(',', $data)));
```
jolelievre, line 572: Nice one
I'm only afraid of one thing, in case you need to perform two checks on one field (or use it for anything else) it will have been cleared from the `$data` This might lead to unexpected behaviour for someone who is not aware of this cleaning (like "why did my data disappear?")

But for now, as long as we don't have issues this is perfect for the job

tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature
```diff
@@ -1,5 +1,6 @@
 # ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml -s cldr
 @reset-database-before-feature
+@clear-cache-after-feature
```
jolelievre, line 3: Why did you need to change this? Did it impact the product tests?
I'm always very careful with tests related to CLDR.. I don't know why 😅

tests/Integration/Behaviour/Features/Scenario/Product/update_shipping.feature
```diff
@@ -0,0 +1,127 @@
...
+      | height                           | 5                          |
+      | depth                            | 4                          |
+      | weight                           | 2                          |
+
+  Scenario: Provide negative additional shipping cost
+    Given product product1 should have following values:
+      | additional_shipping_cost         | 12                         |
+    When I update product product1 shipping information with following values:
+      | additional_shipping_cost         | -12                        |
+    Then I should get error that product additional_shipping_cost is invalid
```
jolelievre, line 100: You can add a check that `additional_shipping_cost` is still 12 like you did for other fields

jolelievre (approved):

jolelievre:
Thank you @zuk3975

tests/Integration/Behaviour/Features/Scenario/CLDR/cldr.feature
zuk3975: for a record, 
It used to cache languages in memory, so my product multilang scenarios failed. 
e.g. I was expecting only EN and FR langs, but there was multiple other languages created, so values didn't match.

tests/Integration/Behaviour/Features/Context/Domain/ProductFeatureContext.php
zuk3975: Yea, we'll figure something out when time comes :sweat_smile:

PR #20111, opened 2020-07-08
Fix order max quantity

### LINKED ISSUE #19855: Invalid check for max quantity when editing product quantity in Order
#### Describe the bug

When I edit a product quantity in an order I can't use the max available quantity, I get an error that there are not enough products in stock.

Capture d’écran 2020-06-18 à 15 46 50

When I already added a big quantity of the product I can also get the same error when I try to reduce the amount (see detailed scenario).

#### Expected behavior

The problem comes from the `UpdateProductInOrderHandler` that compares the remaining stock with the new quantity from the command. The problem is that it doesn't take into account the quantity **already** in the Order.

So if you have 300 products in the order, and 1 is left When you edit the quantity to 301 it will be higher than 1 and so it is blocked. Same if you try to reduce to 200, it is higher than 1 so it is blocked as well..

#### Steps to Reproduce

1. Go to an order (the ones from fixtures can work), pick a product let's say with quantity 1 and 300 available
Capture d’écran 2020-06-18 à 16 32 43

2. Update its quantity to the available amount 300 and save
Capture d’écran 2020-06-18 à 16 33 01

3. This works you have 300 products in your order, and 1 left available
4. Now try to edit to 301, this will fail
5. Try to edit to 200, this will fail as well

**Note**

I didn't detail all the test cases but this behaviour should be checked AND fixed if needed with:
- adding a product
- editing a combination (as each combination has its stock) to make sure they are correctly handled as well
- adding a combination for the same reason

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Max allowed quantity was wrongly calculated for update actions (the amount in the order was not taken into account), this is now possible to update the product quantity with all the remaining stock (and to reduce it without any bug) Also homogenize the stock comparison in AddProductToOrderHandler and simplify the update handler a bit
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19855
| How to test?  | The scenario is detailed in the issue (with screenshots and all 😉 )

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20111)


matthieu-rolland (dismissed):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -387,14 +389,23 @@ private function getCartProductData(Cart $cart, Product $product, int $quantity)
      */
     private function addProductToCart(Cart $cart, Product $product, $combination, $quantity)
     {
+        /**
+         * Update product and customization in the cart, last argument skip quantity check
```
matks: It's a bit hard to understand when you are not "in the context" 😅 

May I suggest:
```
         * Here we update product and customization in the cart.
         *
         * The last argument "skip quantity check" is set to true because
         * 1) the quantity has already been checked,
         * 2) (main reason) when the cart checks the availability ; it substracts
         * its own quantity from available stock.
         *
         * This is because a product in a cart is not really out of the stock, because it is not checked out yet.
         *
         * Here we are editing an order, not a cart, so what has been ordered
         * has already been substracted from the stock.
```

src/Adapter/Order/OrderProductQuantityUpdater.php
```diff
@@ -183,13 +193,23 @@ private function updateProductQuantity(
 
             $this->updateCustomizationOnProductDelete($order, $orderDetail, $oldQuantity);
         } else {
-            // Update product and customization in the cart
+            /**
+             * Update product and customization in the cart, last argument skip quantity check
```
matks: Same complex explanation (it's complex because it explains a complex issue), I suggest same change

matks (dismissed):
One suggestion but code is good anyway

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
jolelievre: Yep much easier to read 😉

matthieu-rolland (dismissed):

matks (dismissed):

Robin-Fischer-PS:
Hi @jolelievre !

I have this issue : 
- Have an order with product A
- Add product A in the order with all stock available
- Without refreshing, now edit the first product line and also add all stock available
- Save : No error, there should be one since the stock is no longer available

See screenrecord : https://drive.google.com/file/d/11AgVEOFW7-AttKNDkvBm8j5meCOMtw4o/view?usp=sharing

jolelievre:
Thanks @Robin-Fischer-PS 
I fixed the calculation problem, I summed up all the OrderDetail's quantities but only the one being edited needs to be used

I didn't udpate the totals in the front so on the second edition you get an Out of stock error, I think it's enough to avoid a forbidden actions, you can always refresh the page to have updated totals Besides if the stock was changed in another order by someone else you wouldn't see it either

This will need to be approved by the @PrestaShop/prestashop-product-team but I think it's enough

matks (dismissed):

Robin-Fischer-PS:
@jolelievre One new issue : When a product is set to "Allow out-of-stock order", it is not possible to edit with more product than available.
When adding a product, it is possible to do so.

See screenrecord : 

https://drive.google.com/file/d/1LwyidPmWqf8h0cup2wusZXZLaACsDe1q/view?usp=sharing

PierreRambaud (dismissed):

matks (dismissed):

matks (dismissed):

PierreRambaud (dismissed):

colinegin:
ping @MatShir

khouloudbelguith:
Hi @jolelievre,

A new issue Following these steps:


1. Create a productA with 2 combinations _Size - S, Color - Black_ Quantity=20 and _Size - S, Color - Red_ Quantity=20 
2. The combination by default is _Size - S, Color - Black_ 
2. Go to the FO and create an order with this product _Size - S, Color - Black_ with quantity=20
3. Go to `admin-dev/sell/orders/orders/{orderId}/view and View the last order created
8. Click Add product, search and select the last product created productA  _Size - S, Color - Red_  with quantity=5
4. Check the product added is not a good combination => the product added is the combination by default and the quantity now <0 

https://drive.google.com/file/d/1RMgAYL0VyK4xKlb83adG-LCc34JYoeED/view?usp=sharing

Thanks!

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -329,18 +331,22 @@ private function restoreOrderProductsSpecificPrices(Order $order, Cart $cart): a
      *
      * @param Cart $cart
      * @param Product $product
+     * @param Combination|null $combination
      * @param int $quantity
      *
      * @return array
      */
-    private function getCartProductData(Cart $cart, Product $product, int $quantity): array
+    private function getCartProductData(Cart $cart, Product $product, $combination, int $quantity): array
```
Progi1984:
```suggestion
    private function getCartProductData(Cart $cart, Product $product, ?Combination $combination, int $quantity): array
```

matks (dismissed):
This PR is good for me 👍

matthieu-rolland (dismissed):

tests/Integration/Behaviour/Features/Context/CommonFeatureContext.php
```diff
@@ -106,6 +107,16 @@ public function clearEntityManager()
...
     }
 
+    /**
+     * @BeforeStep
+     *
+     * Clear Product prices cache at each step in order to get fresh data
+     */
+    public function clearProductPrices()
+    {
+        Product::flushPriceCache();
```
jolelievre: Maybe it's a bit overkill to perform this clear at each step? (and in Common context which is used by many tests contexts) This could lead to unexpected issues, like the cache is cleared at each step in behat tests, but it's not the case in the real BO or FO which leads to different behaviour that we can't detect in the tests

atomiix: I agree but not doing it lead to unexpected issues too 😅. I think that clearing it between each step is closer to what really happens between each http request than not clearing.

matks (dismissed):

tests/Integration/Behaviour/Features/Scenario/Order/order_from_bo.feature
```diff
@@ -263,7 +263,6 @@ Feature: Order from Back Office (BO)
       | name          | Mug Today is a good day |
       | amount        | 2                       |
       | price         | 16                      |
-      | free_shipping | true                    |
```
Progi1984, line 266: Why remove this line ?

tests/Integration/Behaviour/Features/Scenario/Order/order_from_bo.feature
```diff
@@ -350,6 +349,27 @@ Feature: Order from Back Office (BO)
...
+      | combination   | combination1  |
+      | amount        | 1             |
+      | price         | 10            |
+    And I add products to order "bo_order1" with new invoice and the following products details:
+      | name          | My Product    |
+      | combination   | combination2  |
+      | amount        | 1             |
+      | price         | 10            |
+    Then order "bo_order1" should contain 1 products "My Product - Size : L"
+    And order "bo_order1" should contain 1 products "My Product - Size : M"
```
Progi1984:
```suggestion
    Then order "bo_order1" should contain 1 product "My Product - Size : L"
    And order "bo_order1" should contain 1 product "My Product - Size : M"
```

tests/Integration/Behaviour/Features/Scenario/Order/order_from_bo.feature
atomiix: There has been a change in the function and this key is not mandatory anymore

Progi1984 (commented):
One question and one comment

khouloudbelguith:
Hi @jolelievre,

Could you please resolve conflicts?

Thanks!

tests/Integration/Behaviour/Features/Scenario/Order/order_from_bo.feature
Progi1984: :ok_hand:

Progi1984 (dismissed):

src/Adapter/Order/OrderProductQuantityUpdater.php
```diff
@@ -429,4 +455,44 @@ private function getTaxCountry(Cart $cart): Country
...
+     * @param int $newQuantity
+     *
+     * @throws ProductOutOfStockException
+     */
+    private function assertValidProductQuantity(Order $order, OrderDetail $orderDetail, int $newQuantity)
+    {
+        //check if product is available in stock
+        if (!Product::isAvailableWhenOutOfStock(StockAvailable::outOfStock($orderDetail->product_id))) {
+            $availableQuantity = StockAvailable::getQuantityAvailableByProduct($orderDetail->product_id, $orderDetail->product_attribute_id);
+            $productQuantity = $this->getProductQuantityInOrder($order, (int) $orderDetail->product_id, (int) $orderDetail->product_attribute_id);
```
jolelievre: Woups, seems like I forgot to fix this one This was originally done in the handler as well (same code) But it's actually not correct We shouldn't use the quantity in the whole order, but only the one from OrderDetail It's actually weird that this doesn't make tests crash.. This makes me fear that this check is not effective 🤔

src/Adapter/Order/OrderProductQuantityUpdater.php
```diff
@@ -429,4 +455,43 @@ private function getTaxCountry(Cart $cart): Country
...
+    }
+
+    /**
+     * @param Order $order
+     * @param int $productId
+     * @param int $productAttributeId
+     *
+     * @return int
+     */
+    private function getProductQuantityInOrder(Order $order, int $productId, int $productAttributeId): int
```
jolelievre: this method can be removed since it's not used any more

jolelievre:
I can't approve because I'm the original author, but everything seems good to me 👍

Progi1984 (approved):

matthieu-rolland (approved):

khouloudbelguith:
Hi @jolelievre,

We have an incorrect behavior when the product is out of stock Following these steps:

1. Create a productA with Quantity=20 and minimal quantity=10
2. Go to `admin-dev/sell/orders/orders/{orderId}/view
3. Make sure order status is not "delivered" and you can add products.
4. Click Add product, search and select the last product created productA with quantity=1
5. it is ok => check the alert
6. Set the quantity to 20 and add this product productA to the order => ok
7. Go back to the Catalog => Product page
8. Edit productA
9. Set the quantity of this productA to 0 and Allow orders
10. Go back to the previous Order
11. Click Add product, search and select the last product created productA with quantity=1

=> the product well added => NOK
=> it should be displayed `Error! Minimum quantity of "10" must be added`
https://drive.google.com/file/d/1ioFi2mV_L-HNKmR0pfNjCUB5XXLsh-nm/view?usp=sharing

I tried with 1.7.7.x, without this PR, at the `step11`, the product can't be added but with an **incorrect alert**
image

I have an incorrect behavior following these steps:
1. Create a productA with Quantity=20 and minimal quantity=10
2. Go to `admin-dev/sell/orders/orders/{orderId}/view
3. Make sure order status is not "delivered" and you can add products.
4. Click Add product, search and select the last product created productA with quantity=1
5. it is ok => check the alert
6. Set the quantity to 20 and add this product productA to the order => ok
7. Edit this product and update the quantity with 9 (**< minimal quantity**) 
8. Save => the quantity is well updated
=> NOK => it should be displayed `Error! Minimum quantity of "10" must be added`

https://drive.google.com/file/d/13VlhSvvIJha_vWcN7pqVzGLo647ADq8O/view?usp=sharing

I tried with 1.7.7.x, without this PR, at step 8, we are blocked by the original issue as reported here: https://github.com/PrestaShop/PrestaShop/issues/19855


Thanks!

atomiix:
Hi @khouloudbelguith,

I'm maybe wrong, but I think that it is the expected behavior from what I can read in the [Order specs](https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/orders/orders/Order%20Page%20View.md#iv-products-panel) (section *IV. Products panel -> 2. Add new products*)

khouloudbelguith:
Hi @atomiix,

Thanks for these clarifications.
So as described in the specs we have 3 cases:
1. Add product to order for the **first time** with `quantity < minimum quantity` => an error is launched: "Error! Minimum quantity of X must be added"
2. Add a product to order **exists before** with `quantity < minimum quantity `=> is authorized, the responsibility belongs to the merchant.
3. **Update** a product's quantity with `quantity < minimum quantity` => is authorized, the responsibility belongs to the merchant.

In this case, the PR is ok :heavy_check_mark: 
https://drive.google.com/file/d/1aVinOMocMEximDg21pOyzqGKoSod1vV8/view?usp=sharing

Thanks @jolelievre @atomiix

jolelievre:
Thanks @khouloudbelguith 

Indeed the specs define the following behaviour:

```
The minimum quantity of a product is determined by all the products of the order regardless of whether the products are separated between invoices. It allow a merchant to add a product on another invoice because the minimum quantity is respected on in the order. Edting the quantity of the product to a quantity less than the minimum quantity required is authorised, the responsibility belongs to the merchant.

For example, when I add a product A with a minimum quantity, if the quantity of the new product A and the current quantity are equal to or greater than the product is added. Otherwise, an error is launched: "Error! Minimum quantity of X must be added", X quantity missing to be equal to the required minimum quantity.
```

So for your point 2. actually it is possible if the total of the product (among multi invoices) is greater than the minimal quantity Maybe we should test this behaviour, the easier way would be to change the minimal quantity after the product is added For example:

1. Add 10 product with minimal quantity 10 -> OK
2. Change minimal product quantity to 15
3. Add 3 product to new invoice -> Error shown (because 13 < 15)
4. Add 5 products to new invoice -> OK

jolelievre:
Maybe we should add some behat tests to ensure this behaviour is ok and avoid future regressions ^^
In another PR, so we can merge this one for now ;)

jolelievre:
Thanks @khouloudbelguith and @atomiix

PR #20118, opened 2020-07-09
PHPStan (Level 2)

### LINKED ISSUE #16471: Use static analyser to improve code quality
Let's integrate [PHPStan](https://github.com/phpstan/phpstan) or [Psalm](https://github.com/vimeo/psalm) into our CI to improve code quality.

[EDIT by @Progi1984]
:notebook: Applied
- [x] on `admin-dev/`
  - #24678
- [x]  on `classes/` (445)
  - #25533
  - #25916
  - #26164
  - #26478
  - #26727
  - #26913
- [x]  on `config/` (25)
  - #26479
- [x]  on `controllers/` (799)
  - #26518
  - #26770
  - #26864
  - #27004
  - #27102
  - #27417
- [x]  on `install-dev/` (497)
  - #26819
- [x]  on `js/` (2)
  - #26744
- [x] on `override/`
  - #24466
- [x] on `src/` (above)
  - #19898
  - #20118 
  - #21241 
  - #21515 
  - #22547 
- [x] on `tools/`
  - #24466
- [x]  on `tests/` (312)
  - #26592
  - #27107
  - #27125
  - #27141
  - #27287
- [x]  on `tests-legacy/`
  - #27247
- [x] on `webservice/`
  - #23960

Final 
- [x] Update to latest version of PhpStan : #27723
- [x] Use `dynamicConstantNames` (https://phpstan.org/config-reference#constants) and remove `@phpstan-ignore-next-line` : #28060

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | PHPStan (Level 2)
| Type?         | improvement
| Category?     | PM
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | Fixes #16471
| How to test?  | PHPStan is 💚  & Travis CI is 💚 <br>Exploratory analysis (add a check in the BO > Products/Supplier/Warehouse (mainly forms)) 

## :notebook: BC Breaks 
* `src/Adapter/Module/Tab/ModuleTabRegister.php` : The parameter `$routingConfigLoader` uses `Symfony\Component\Config\Loader\Loader` as parameter instead of `Symfony\Component\Config\Loader\LoaderInterface` because we need to use the `import` method
* `src/PrestaShopBundle/EventListener/ModuleActivatedListener.php` : The parameter `$session` is now a `Symfony\Component\HttpFoundation\Session\Session` object, and was a `Symfony\Component\HttpFoundation\Session\SessionInterface` object, because we need to use the `getFlashBag()` method
* `src/PrestaShopBundle/EventListener/AccessDeniedListener.php` : The parameter `$session` is now a `Symfony\Component\HttpFoundation\Session\Session` object, and was a `Symfony\Component\HttpFoundation\Session\SessionInterface` object, because we need to use the `getFlashBag()` method

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20118)


matks (dismissed):
I know this is WIP but the work done so far is good 👍

matthieu-rolland (dismissed):

.github/workflows/phpstan/autoload.php
```diff
@@ -25,6 +25,17 @@ function requireFileIfItExists(string $filepath) : bool {
 require_once _PS_ROOT_DIR_ . '/install-dev/classes/exception.php';
 require_once _PS_ROOT_DIR_ . '/install-dev/classes/session.php';
 require_once _PS_ROOT_DIR_ . '/var/SymfonyRequirements.php';
+$moduleCronJobs = _PS_ROOT_DIR_ . '/modules/cronjobs/cronjobs.php';
+if (file_exists($moduleCronJobs)) {
+    require_once $moduleCronJobs;
+} else {
+    class CronJobs extends Module
+    {
+        public static function addOneShotTask($task, $description, $execution = array())
```
matks: From the outside it seems like a dirty fix for a bad code design 😅  . Is there no way we rather fix the bad code that needed you to do this instead of adding this trick ?

If you dont want to burden the scope of this PR, we can add a `@todo` for later 😉

src/Adapter/Admin/AbstractAdminQueryBuilder.php
```diff
@@ -117,11 +117,11 @@ private function compileSqlWhere(array $whereArray)
      * Format example for $order:
      * $order = array('name ASC', 'id_product DESC');
      *
-     * @param array[array[mixed]] $select
-     * @param array[mixed] $table
-     * @param array[mixed] $where
-     * @param array[string] $groupBy
-     * @param array[string] $order
+     * @param array<array<mixed>> $select
```
matks, line 120: ❤️

src/Adapter/Cache/MemcacheServerManager.php
```diff
@@ -100,7 +100,7 @@ public function testConfiguration($serverIp, $serverPort)
     /**
      * Delete a memcache server (a deletion returns the number of rows deleted).
      *
-     * @param int $serverId_server id (in database)
```
matks, line 103: 🤦‍♂️

src/Adapter/Import/Handler/ImportHandlerFinder.php
```diff
@@ -41,7 +41,7 @@ final class ImportHandlerFinder implements ImportHandlerFinderInterface
     private $importHandlers;
 
     /**
-     * @param ImportHandlerInterface[] $importHandlers
+     * @param ImportHandlerInterface $importHandlers
```
matks: But it's an array 😮  it's the splat operator https://lornajane.net/posts/2014/php-5-6-and-the-splat-operator

Maybe PHPstan does not support the understanding  of the splat operator 🤔  ?

src/Adapter/Import/ImportDataFormatter.php
```diff
@@ -41,7 +41,7 @@ final class ImportDataFormatter
     private $configuration;
 
     /**
-     * @var Tools
+     * @vaPrestaShopBundle/Install/Upgrade.phpr Tools
```
matks: I'm quite sure than this is wrong 😄

src/Adapter/Language/CommandHandler/EditLanguageHandler.php
```diff
@@ -229,11 +229,13 @@ private function assertLanguageWithIsoCodeDoesNotExist(Language $language, EditL
         if (null !== $command->getIsoCode()) {
             return;
         }
-
+        //@todo Fix l.229 $command->getIsoCode() === null
```
matks: We cannot merge this ⚠️

src/Adapter/Module/Tab/ModuleTabUnregister.php
```diff
@@ -72,7 +72,7 @@ public function __construct(TabRepository $tabRepository, LangRepository $langRe
      *
      * This is done automatically as part of the module uninstallation.
      *
-     * @return bool returns true if the module tabs were successfully uninstalled, false if any of them failed to do so
+     * @return void
```
matks, line 75: Can we check usages of `Module::unregisterTabs()`

I think the phpDoc is right and the code is wrong 😅  instead of the opposite

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -585,7 +585,6 @@ private function createNewOrEditExistingInvoice(
      * @param Cart $cart
      * @param bool $isFreeShipping
      * @param array $newProducts
-     * @param
```
matks, line 364: 🤦

matks: (I feel bad)

src/Core/Domain/Address/Command/AbstractEditAddressCommand.php
```diff
@@ -238,7 +238,7 @@ public function getCountryId(): ?CountryId
     }
 
     /**
-     * @param int|null $countryId
+     * @param int $countryId
```
matks, line 241: But it can be `null` sometimes 😮 
Maybe PHPStan has another way to define nullable properties, but writing `@param int $countryId` would be lying. This field can be null in some expected usecases.

src/Core/Domain/Attachment/Command/EditAttachmentCommand.php
```diff
@@ -85,15 +85,15 @@ public function getAttachmentId(): AttachmentId
...
+     * @return string
      */
     public function getPathName(): ?string
     {
         return $this->pathName;
     }
 
     /**
-     * @param string|null $pathName
+     * @param string $pathName
```
matks, line 96: Same here, all of these fields are null when used in a partial edit usecase. The phpDoc is right. If PHPStan marks these statements as wrong, something is wrong either in our PHPStan configuration or in PHPStan internal logic.

src/Core/Domain/Manufacturer/Command/BulkToggleManufacturerStatusCommand.php
```diff
@@ -89,7 +89,7 @@ private function setManufacturerIds(array $manufacturerIds)
     /**
      * Validates that value is of type boolean
      *
-     * @param $value
+     * @param bool $value
```
matks: `$value` real type is `mixed`. We _want_ it to be a boolean 😄  but we are not sure which is why we assert it. We expect sometimes people pass something else than a boolean here.

src/Core/Domain/Customer/ValueObject/CustomerDeleteMethod.php
```diff
@@ -44,12 +44,14 @@ class CustomerDeleteMethod
     const DENY_CUSTOMER_REGISTRATION = 'deny_registration_after';
 
     /**
-     * @var int
+     * @var string
```
matks, line 47: Nice catch

src/Core/Domain/Customer/ValueObject/ApeCode.php
```diff
@@ -40,7 +40,7 @@ class ApeCode
     private $code;
 
     /**
-     * @param $code
+     * @param string $code
```
matks: `$code` real type is `mixed`. We want it to be a string 😄  but we are not sure which is why we assert it. We expect sometimes people pass something else than a string here.

src/Core/Domain/CmsPage/Command/EditCmsPageCommand.php
```diff
@@ -300,7 +300,7 @@ public function getShopAssociation()
     }
 
     /**
-     * @param array|null $shopAssociation
+     * @param array $shopAssociation
```
matks: Nullable property => need to be solved like the others

src/Core/Hook/HookDispatcher.php
```diff
@@ -77,7 +78,7 @@ public function dispatchRendering(HookInterface $hook)
             $hook->getParameters()
         );
 
-        $content = $event->getContent();
+        $content = $event instanceof RenderingHookEvent ? $event->getContent() : [];
```
matks: I'm not 100% sure this is not changing the behavior of the HookDispatcher

src/PrestaShopBundle/Controller/Admin/ProductController.php
```diff
@@ -1263,7 +1267,7 @@ public function renderFieldAction($productId, $step, $fieldName)
             $this->get('prestashop.adapter.data_provider.tax'),
             $this->get('router')
         );
-        $form = $this->createFormBuilder($modelMapper->getFormData());
+        $form = $this->createFormBuilder($modelMapper->getFormData($product));
```
matks, line 1270: We need to check this with QA 😮 I would be surprised / afraid it does not modify the behavior

src/PrestaShopBundle/EventListener/ModuleActivatedListener.php
```diff
@@ -57,7 +58,7 @@ class ModuleActivatedListener
     private $translator;
 
     /**
-     * @var SessionInterface
+     * @var Session
```
matks, line 60: 🤔  sure about that ? usually we prefer to type-hint the Interface rather than the implementation

src/PrestaShopBundle/EventListener/TokenizedUrlsListener.php
```diff
@@ -108,7 +109,7 @@ public function onKernelRequest(KernelEvent $event)
             $token = new CsrfToken('form', $request->query->get('form')['_token']);
         }
 
-        if (false === $token || !$this->tokenManager->isTokenValid($token)) {
+        if ((false === $token || !$this->tokenManager->isTokenValid($token)) && $event instanceof GetResponseEvent) {
```
matks, line 112: Need to be checked with @PierreRambaud

src/PrestaShopBundle/Security/Role/DynamicRoleHierarchy.php
```diff
@@ -27,12 +27,18 @@
...
 
 /**
  * Class DynamicRoleHierarchy is used for Symfony role hierarchy voter to load roles from database.
  */
 class DynamicRoleHierarchy implements RoleHierarchyInterface
 {
+    /**
+     * @param array<RoleInterface> $roles An array of directly assigned roles
+     *
+     * @return void An array of all reachable roles
```
matks: It's strange to return void but describe its content 😅

matks (changes requested):
Hi @Progi1984 ! What a huge work 😄 

I am sorry I am going to be again the guy who complains but I saw a few things that I cannot approve/merge. I think we need to examine it deeper to see why PHPStan mark it as issues 😉  please see my feedbacks

src/Adapter/Import/Handler/ImportHandlerFinder.php
Progi1984: Fixed.

src/Adapter/Import/ImportDataFormatter.php
Progi1984: Fixed

src/Adapter/Language/CommandHandler/EditLanguageHandler.php
Progi1984: Fixed

src/Adapter/Module/Tab/ModuleTabUnregister.php
Progi1984: :warning: it's not the scope of the PR ;)

Progi1984: But the new PHPDoc is right ;)

src/Core/Domain/Address/Command/AbstractEditAddressCommand.php
Progi1984: No it can't be null

`public function setCountryId(int $countryId)` And the file is in strict type mode.

src/PrestaShopBundle/Security/Role/DynamicRoleHierarchy.php
Progi1984: Fixed

src/PrestaShopBundle/EventListener/ModuleActivatedListener.php
Progi1984: You're right. I checked the `src/PrestaShopBundle/Resources/config/services/bundle/event_listener.yml` and it's well `Session`. I change the __construct. Is it a BC ?

src/PrestaShopBundle/Controller/Admin/ProductController.php
Progi1984: I don't understand why `getFormData` don't return any error as it lacks a parameter

src/Core/Hook/HookDispatcher.php
Progi1984: Fixed. The PhpDoc in `Adapter/Hook/HookDispatcher.php` was updated at an another moment.

src/Core/Domain/CmsPage/Command/EditCmsPageCommand.php
Progi1984: Fixed

src/Core/Domain/Customer/ValueObject/ApeCode.php
Progi1984: Wrong. The class `ApeCode` is used twice :
- src/Core/Domain/Customer/Command/AddCustomerCommand.php
- src/Core/Domain/Customer/Command/AddCustomerCommand.php

In each case, the parameter is in string. And for each use of setApeCode, data are casted in string.

src/Core/Domain/Manufacturer/Command/BulkToggleManufacturerStatusCommand.php
Progi1984: The `BulkToggleManufacturerStatusCommand` is used twice in `src/PrestaShopBundle/Controller/Admin/Sell/Catalog/ManufacturerController.php` and each time, the `expectedStatus` is a boolean.

src/Core/Domain/Attachment/Command/EditAttachmentCommand.php
Progi1984: Wrong ! The data passed here come from a SplFileInfo object and `getPathname` method return always a string.

phpstan.neon.dist
```diff
@@ -1,8 +1,15 @@
 parameters:
-	level: 1
+	level: 2
 	bootstrapFiles:
 		- .github/workflows/phpstan/autoload.php
 	paths:
 		- src
 	ignoreErrors:
 		- '#Unsafe usage of new static().#'
```
PierreRambaud:
```suggestion
		- '~^Unsafe usage of new static\(\)\.$~'
```
These are regex ;)

.github/workflows/phpstan/autoload.php
PierreRambaud: I agree

src/Adapter/Currency/CommandHandler/ToggleExchangeRateAutomatizationHandler.php
```diff
@@ -180,10 +180,13 @@ private function getCronUrl()
      */
     private function createCronJob($cronUrl)
     {
+        if (!class_exists(CronJobs::class)) {
+            return false;
+        }
         /** @var CronJobs $cronJobsModule */
         $cronJobsModule = Module::getInstanceByName('cronjobs');
```
PierreRambaud:
```suggestion
        /** @var CronJobs $cronJobsModule */
        $cronJobsModule = Module::getInstanceByName('cronjobs');
        if (empty($cronJobsModule)) {
            return false;
        }
```
Just test if the module is correctly loaded.

phpstan.neon.dist
Progi1984: Fixed

src/Adapter/Currency/CommandHandler/ToggleExchangeRateAutomatizationHandler.php
Progi1984: Fixed

.github/workflows/phpstan/autoload.php
Progi1984: Fixed

src/PrestaShopBundle/EventListener/ModuleActivatedListener.php
Progi1984: Ping @matks

matks: Yes it's the BC (if someone else used another class implementing SessionInterface, with this code change it will fail for him).

And we _want_ to require the interface, not the implementation. That is good practice 🤔

Progi1984: The problem is that SessionInterface doesn't declare `getFlashBag` method used at l.143.

matks: Could we use `public function getBag($name);` instead, which belongs to SessionInterface method

src/PrestaShopBundle/Controller/Admin/ProductController.php
matks: Magic !

src/PrestaShopBundle/EventListener/ModuleActivatedListener.php
Progi1984: @matks I fixed without BC ;)

src/PrestaShopBundle/Controller/Admin/ProductController.php
Progi1984: No problem. After validation, I ask a QA with check on BO & Product ;)

matks (dismissed):
Nice work 🎉  I approve the PR

src/Adapter/Module/Module.php
```diff
@@ -349,7 +351,7 @@ public function onReset()
             return false;
         }
 
-        return $this->instance->reset();
+        return method_exists($this->instance, 'reset') ? $this->instance->reset() : true;
```
PierreRambaud: Watch out https://www.php.net/manual/en/function.method-exists.php#124462

src/Core/ConstraintValidator/AddressDniRequiredValidator.php
```diff
@@ -56,6 +57,9 @@ public function __construct(CountryRequiredFieldsProviderInterface $countryRequi
      */
     public function validate($value, Constraint $constraint)
     {
+        if (!($constraint instanceof AddressDniRequired)) {
+            return;
+        }
```
PierreRambaud, line 62:
```suggestion
    public function validate($value, AddressDniRequired $constraint)
    {
```
Why not something like this? :thinking:

src/Core/ConstraintValidator/AddressZipCodeValidator.php
```diff
@@ -70,9 +70,11 @@ public function __construct(
      */
     public function validate($value, Constraint $constraint)
     {
+        if (!($constraint instanceof AddressZipCode)) {
```
PierreRambaud, line 73: same here

src/Core/ConstraintValidator/NotBlankWhenRequiredValidator.php
```diff
@@ -26,14 +26,15 @@
...
+use PrestaShop\PrestaShop\Core\ConstraintValidator\Constraints\NotBlankWhenRequired;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\Constraints\NotBlankValidator;
 
 class NotBlankWhenRequiredValidator extends NotBlankValidator
 {
     public function validate($value, Constraint $constraint)
     {
-        if (true === $constraint->required) {
+        if ($constraint instanceof NotBlankWhenRequired && true === $constraint->required) {
```
PierreRambaud, line 37: same

src/Core/ConstraintValidator/AddressStateRequiredValidator.php
```diff
@@ -58,12 +60,17 @@ public function __construct(CountryRequiredFieldsProviderInterface $countryRequi
      */
     public function validate($value, Constraint $constraint)
     {
+        if (!($constraint instanceof AddressStateRequired)) {
```
PierreRambaud, line 63: same

src/Adapter/Module/Tab/ModuleTabRegister.php
```diff
@@ -288,11 +289,13 @@ protected function getModuleControllersFromRouting(string $moduleName): array
         }
 
         $routingControllers = [];
-        $moduleRoutes = $this->routingConfigLoader->import($routingFile, 'yaml');
-        foreach ($moduleRoutes->getIterator() as $route) {
-            $legacyController = $route->getDefault('_legacy_controller');
-            if (!empty($legacyController)) {
-                $routingControllers[] = $legacyController;
+        if ($this->routingConfigLoader instanceof Loader) {
```
PierreRambaud: routingConfigLoader is already a LoaderInterface and is a mandatory parameter :thinking:

src/Adapter/Module/Module.php
Progi1984: Interesting and fixed ;) Thanks :heart_eyes:

src/Core/Domain/Notification/Command/UpdateEmployeeNotificationLastElementCommand.php
```diff
@@ -40,11 +40,9 @@ class UpdateEmployeeNotificationLastElementCommand
     private $type;
 
     /**
-     * UpdateEmployeeNotificationLastElementCommand constructor.
-     *
      * @param string $type
      *
-     * @throws TypeException
+     * @throws NotificationException
```
PierreRambaud, line 45: it throws nothing :thinking:

src/PrestaShopBundle/EventListener/TokenizedUrlsListener.php
PierreRambaud: I approve this check :+1:

src/Core/ConstraintValidator/AddressDniRequiredValidator.php
Progi1984: Because that brings a BC !

src/Adapter/Module/Tab/ModuleTabRegister.php
Progi1984: But the method `import` is only declared on abstract class `Loader`.

src/PrestaShopBundle/EventListener/ModuleActivatedListener.php
```diff
@@ -140,6 +141,9 @@ public function onKernelController(FilterControllerEvent $event)
      */
     private function showNotificationMessage(ModuleActivated $moduleActivated)
     {
+        if (!($this->session instanceof Session)) {
```
PierreRambaud: already a SessionInterface and it's mandatory. If something is failing, the interface needs to be updated :thinking:

src/PrestaShopBundle/EventListener/AccessDeniedListener.php
```diff
@@ -53,7 +54,7 @@ class AccessDeniedListener
     private $translator;
 
     /**
-     * @var SessionInterface
+     * @var Session
```
PierreRambaud, line 56: Always prefer SessionInterface

src/Core/Domain/Notification/Command/UpdateEmployeeNotificationLastElementCommand.php
Progi1984: The `new Type($type)` : yes

src/PrestaShopBundle/EventListener/ModuleActivatedListener.php
Progi1984: But the method `getFlashBag` is only declared in `Session` class and not in the interface.

src/PrestaShopBundle/Controller/Admin/WarehouseController.php
```diff
@@ -70,7 +70,7 @@ public function refreshProductWarehouseCombinationFormAction($idProduct)
             $this->get('prestashop.adapter.data_provider.tax'),
             $this->get('router')
         );
-        $allFormData = $modelMapper->getFormData();
+        $allFormData = $modelMapper->getFormData($product);
```
PierreRambaud, line 73: :scream:

src/PrestaShopBundle/EventListener/AccessDeniedListener.php
Progi1984: Fixed

src/PrestaShopBundle/Controller/Admin/WarehouseController.php
Progi1984: :warning: The method`getFormData` of modelMapper has a missing parameter :(

I asked a exploratory analysis from QA

matks:
Travis not happy:
```
1) LegacyTests\Integration\PrestaShopBundle\Controller\Admin\SurvivalTest::testPagesAreAvailable with data set "admin_categories_index" ('Categories', 'admin_categories_index')

Error: Cannot use object of type PrestaShop\PrestaShop\Core\Grid\Filter\FilterCollection as array

/home/travis/build/PrestaShop/PrestaShop/src/Core/Grid/Factory/CategoryGridFactoryDecorator.php:135
/home/travis/build/PrestaShop/PrestaShop/src/Core/Grid/Factory/CategoryGridFactoryDecorator.php:81
/home/travis/build/PrestaShop/PrestaShop/src/PrestaShopBundle/Controller/Admin/Sell/Catalog/CategoryController.php:98
/home/travis/build/PrestaShop/PrestaShop/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php:151
/home/travis/build/PrestaShop/PrestaShop/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php:68
/home/travis/build/PrestaShop/PrestaShop/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:200
/home/travis/build/PrestaShop/PrestaShop/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Client.php:68
/home/travis/build/PrestaShop/PrestaShop/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Client.php:131
/home/travis/build/PrestaShop/PrestaShop/vendor/symfony/symfony/src/Symfony/Component/BrowserKit/Client.php:318
/home/travis/build/PrestaShop/PrestaShop/tests-legacy/Integration/PrestaShopBundle/Controller/Admin/SurvivalTest.php:86
```

src/PrestaShopBundle/EventListener/ModuleActivatedListener.php
PierreRambaud: So, we must change the signature of the method if we use undefined methods :thinking:

src/Adapter/Module/Tab/ModuleTabRegister.php
PierreRambaud: So we need a Loader, and not a LoaderInterface :/

src/Core/ConstraintValidator/AddressDniRequiredValidator.php
PierreRambaud: I know, but not we maybe need to accept this kind of BC.

ping @matks

src/Core/Domain/Notification/Command/UpdateEmployeeNotificationLastElementCommand.php
PierreRambaud: I hate the `@throws`  docblock :sweat:

src/Core/ConstraintValidator/AddressDniRequiredValidator.php
matks: IMO the "cost" to avoid the BC is very low here to pay.

As we disagree on it maybe better ask the big boss @eternoendless 😉

PierreRambaud: Don't forget, currently if we use a different class, an exception is raised :sweat_smile:

matks: Oh 😮 then it's a bugfix not a BC 😄

src/Core/Domain/Cart/Command/RemoveCartRuleFromCartCommand.php
```diff
@@ -45,8 +45,10 @@ class RemoveCartRuleFromCartCommand
     private $cartRuleId;
 
     /**
-     * @param $cartId
-     * @param $cartRuleId
+     * @param int $cartId
+     * @param int $cartRuleId
+     *
+     * @throws \PrestaShop\PrestaShop\Core\Domain\Cart\Exception\CartConstraintException
```
atomiix: Is there any reason to not import the exception with a `use` here?

src/Core/Kpi/Row/HookableKpiRowFactory.php
```diff
@@ -73,7 +73,9 @@ public function __construct(
     }
 
     /**
-     * {@inheritdoc}
+     * @return KpiRow|KpiRowInterface|void
```
atomiix: `KpiRowInterface|void` is not enough?

src/PrestaShopBundle/Event/Dispatcher/NullDispatcher.php
```diff
@@ -41,14 +41,30 @@ public function addSubscriber(EventSubscriberInterface $subscriber)
     {
     }
 
+    /**
+     * @param string $eventName
+     * @param \Symfony\Component\EventDispatcher\Event|null $event
+     *
+     * @return \Symfony\Component\EventDispatcher\Event|void
+     */
     public function dispatch($eventName, \Symfony\Component\EventDispatcher\Event $event = null)
```
atomiix: Same here, why not importing with `use`?

src/PrestaShopBundle/Event/Dispatcher/NullDispatcher.php
```diff
@@ -73,10 +95,21 @@ public function dispatchWithParameters($hookName, array $hookParameters = [])
     {
     }
 
+    /**
+     * @param HookInterface $hook
+     *
+     * @return \PrestaShop\PrestaShop\Core\Hook\RenderedHookInterface|void
```
atomiix: Same here

src/PrestaShopBundle/Event/Dispatcher/NullDispatcher.php
```diff
@@ -73,10 +95,21 @@ public function dispatchWithParameters($hookName, array $hookParameters = [])
...
+     */
     public function dispatchRendering(HookInterface $hook)
     {
     }
 
+    /**
+     * @param string $hookName
+     * @param array $hookParameters
+     *
+     * @return \PrestaShop\PrestaShop\Core\Hook\RenderedHookInterface|void
```
atomiix: Same here

src/Core/Domain/Cart/Command/RemoveCartRuleFromCartCommand.php
Progi1984: Fixed

src/Core/Kpi/Row/HookableKpiRowFactory.php
Progi1984: Fixed

src/PrestaShopBundle/Event/Dispatcher/NullDispatcher.php
Progi1984: Fixed

src/PrestaShopBundle/Event/Dispatcher/NullDispatcher.php
Progi1984: Fixed

src/PrestaShopBundle/Event/Dispatcher/NullDispatcher.php
Progi1984: Fixed

phpstan.neon.dist
```diff
@@ -1,8 +1,15 @@
 parameters:
-	level: 1
+	level: 2
 	bootstrapFiles:
 		- .github/workflows/phpstan/autoload.php
 	paths:
 		- src
 	ignoreErrors:
 		- '#^Unsafe usage of new static\(\)\.$#'
+		- '#Call to an undefined method Symfony\\Component\\Config\\Definition\\Builder\\NodeDefinition\:\:#'
```
PierreRambaud:
```suggestion
		- '#^Call to an undefined method Symfony\\Component\\Config\\Definition\\Builder\\NodeDefinition\:\:#'
```

src/PrestaShopBundle/EventListener/AccessDeniedListener.php
```diff
@@ -104,7 +105,9 @@ private function getAccessDeniedResponse(Request $request, AdminSecurity $adminS
             ], Response::HTTP_FORBIDDEN);
         }
 
-        $this->session->getFlashBag()->add('error', $this->getErrorMessage($adminSecurity));
+        if ($this->session instanceof Session) {
```
PierreRambaud: I'm still not sure we should have this kind of check, if the SessionInterface is not the right class to use we should change it for Session instead :thinking:

src/Adapter/Module/Tab/ModuleTabRegister.php
```diff
@@ -288,11 +288,13 @@ protected function getModuleControllersFromRouting(string $moduleName): array
         }
 
         $routingControllers = [];
-        $moduleRoutes = $this->routingConfigLoader->import($routingFile, 'yaml');
-        foreach ($moduleRoutes->getIterator() as $route) {
-            $legacyController = $route->getDefault('_legacy_controller');
-            if (!empty($legacyController)) {
-                $routingControllers[] = $legacyController;
+        if (is_callable([$this->routingConfigLoader, 'import'])) {
```
PierreRambaud: maybe `routingConfigLoader` should be a Loader instead of LoaderInterface :thinking:

phpstan.neon.dist
Progi1984: Fixed

src/PrestaShopBundle/EventListener/AccessDeniedListener.php
Progi1984: image

src/Core/ConstraintValidator/AddressDniRequiredValidator.php
Progi1984: It's not possible : that breaks the interface `vendor/symfony/symfony/src/Symfony/Component/Validator/ConstraintValidatorInterface.php`
```php
public function validate($value, Constraint $constraint);
```

src/Adapter/Module/Tab/ModuleTabRegister.php
Progi1984: It adds a BC but it's logical.

Progi1984: I documented it ;)

src/Adapter/Module/Tab/ModuleTabRegister.php
Progi1984: Oops. I forgot it. I fixed it and documented the BC ;)

src/Core/ConstraintValidator/AddressDniRequiredValidator.php
matks: Let's go for @Progi1984 solution. We can always improve it later. The PR should not be blocked by this minor detail.

src/Core/Domain/Address/Command/AbstractEditAddressCommand.php
matks: Euh yes it can be null 😛 look
Capture d’écran 2020-08-03 à 16 32 31

Nobody obliges you to **call** `setCountryId(int $countryId)`. It's used when there is data 😄 . No data, no setter, it's null. If `EditCustomerAddressCommand` is built without calling the setter, `$countryId` is null. Please trust me 😄 

```
        $editAddressCommand = new EditCustomerAddressCommand($addressId);
        dump($editAddressCommand);
```

src/Core/Domain/Address/Command/AbstractEditAddressCommand.php
```diff
@@ -362,7 +362,7 @@ public function getHomePhone(): ?string
     }
 
     /**
-     * @param string|null $homePhone
+     * @param string $homePhone
```
matks, line 365: Same here. We made these fields nullable to allow people to perform partial updates. Nullable fields are valid for Edit Commands (but NOT for Create Commands)

src/Core/Domain/Attachment/Command/EditAttachmentCommand.php
matks: Nothing into the implementation of `EditAttachmentCommand` obliges the user to get this data from a SplFileInfo object. Our usage of `EditAttachmentCommand` class inside `AttachmentFormDataHandler` uses SplFileInfo data to fill it, but we could fetch the data from another source later.

Moreover we must let modules and shop developers free to choose where they pick the data to fill in `EditAttachmentCommand`. This code must be built to allow extension of the behavior by module and shop developers.

matks: > Wrong !

Please dont use exclamation marks in your comments 😅 
You probably dont want it to sound that way but when I read it, in my mind I think of this:


It feels a bit aggressive 😅

src/Core/Domain/Attachment/QueryResult/EditableAttachment.php
```diff
@@ -44,7 +44,7 @@ class EditableAttachment
     private $name;
 
     /**
-     * @var string[]|null
+     * @var string[]
```
matks, line 47: Same here, same logic: Edit Command have nullable fields and it's done on purpose 😉

src/Core/Domain/Customer/ValueObject/ApeCode.php
matks: It is not enough to check that `ApeCode` _usage_ is always strings, we must also expect that `ApeCode` will be used
- in code that will be written by outside people using our code as API (module and shop developers)
- in code that will be written in the future by us (although we could consider we will be careful)

The class `ApeCode` must behave like a stand-alone class that might be used outside of our context. It's built to be used and extended by module and shop developers. It must contain the necessary checks to ensure its own integrity because we cannot rely on the discipline of our users 😅 .

src/Core/Domain/Manufacturer/Command/BulkToggleManufacturerStatusCommand.php
matks: As explained in https://github.com/PrestaShop/PrestaShop/pull/20118/files#r464460113, we cannot rely on our usage of the code. This code has a life of its own outside of PrestaShop main usage because we expect our code to be extended by module and shop developers, which is why it must be adapted to this specific usage.

src/Core/Domain/Tax/Command/BulkToggleTaxStatusCommand.php
```diff
@@ -89,7 +89,7 @@ private function setTaxIds(array $taxIds)
     /**
      * Validates that value is of type boolean
      *
-     * @param $value
+     * @param bool $value
```
matks: Same logic here. This method `assertIsBool` very purpose is to check that given variable is a boolean. It makes no sense to validate it if we assume it's a boolean 😉

src/PrestaShopBundle/DataCollector/HookRegistry.php
```diff
@@ -93,8 +94,8 @@ public function hookedByModule(ModuleInterface $module)
     /**
      * A callback have been executed by the module during the Hook dispatch.
      *
-     * @param ModuleInterface $module
-     * @param $args array All arguments passed to the Module callback
+     * @param ModuleCore $module
```
matks, line 97: Is it mandatory to force-type the implementation rather than the interface ? This reduces the customization/extension capabilities of module and shop developers

src/PrestaShopBundle/EventListener/AccessDeniedListener.php
matks: This looks good enough to me 😊

src/PrestaShopBundle/EventListener/DemoModeEnabledListener.php
```diff
@@ -54,7 +55,7 @@ class DemoModeEnabledListener
     private $translator;
 
     /**
-     * @var SessionInterface
```
matks, line 57: Could we have SessionInterface ?
If you tell me we cannot, we keep Session 😉

src/PrestaShopBundle/Translation/Provider/AbstractProvider.php
```diff
@@ -208,7 +208,7 @@ public function getDatabaseCatalogue($theme = null)
         $databaseCatalogue = new MessageCatalogue($this->locale);
 
         foreach ($this->getTranslationDomains() as $translationDomain) {
-            $domainCatalogue = $this->getDatabaseLoader()->load(null, $this->locale, $translationDomain, $theme);
+            $domainCatalogue = $this->getDatabaseLoader()->load(null, $this->locale, $translationDomain);
```
matks: ⚠️ DatabaseTranslationLoader **has** a `$theme` argument
```
public function load($resource, $locale, $domain = 'messages', $theme = null)
```

I think phpstan complains because the interface does not have this argument. Once again, you and PHPStan have found a usecase where implementation and interface do not have the same signature 😭 

We cannot remove `$theme` I think it would introduce bugs.

tests-legacy/Unit/Adapter/Module/Tab/ModuleTabRegisterTest.php
```diff
@@ -192,7 +192,7 @@ protected function buildFilesystemMock()
 
     protected function buildRoutingConfigLoaderMock()
     {
-        $moduleRoutingLoader = $this->getMockBuilder(LoaderInterface::class)
+        $moduleRoutingLoader = $this->getMockBuilder(Loader::class)
```
matks, line 195: We should aim for interface when possible. Is this possible here ?

matks (changes requested):
Hi ! I reviewed all the PR

Fixed:
- https://github.com/PrestaShop/PrestaShop/pull/20118/files#r462859378
=> I think your last answer is good, we need to comply with ConstraintValidator interface

Blockers:
- https://github.com/PrestaShop/PrestaShop/pull/20118/files#r464453875
- https://github.com/PrestaShop/PrestaShop/pull/20118/files#r464460113
- https://github.com/PrestaShop/PrestaShop/pull/20118/files#r464468310
- https://github.com/PrestaShop/PrestaShop/pull/20118/files#r464472692
- https://github.com/PrestaShop/PrestaShop/pull/20118/files#r464474513

To fix
- https://github.com/PrestaShop/PrestaShop/pull/20118/files#r455651933

matks:
@Progi1984 This PR is complex because it is very wide and we have to deal with bad old code.
Maybe you could split this PR in two ?
- all the work already validated in 1 PR
- the last blocker items in another PR

This would allow us to merge the 1st PR and the 2nd PR would be much smaller, so easier to handle.

In order for php-stan to accept the 1st PR, we would need to add some temporary `ignore` flags

src/Core/Domain/Address/Command/AbstractEditAddressCommand.php
Progi1984: @matks Sorry ! As `declare(strict_types=1);` is defined, a nullable value in parameter will trigger a fatal error :

image

src/Core/Domain/Attachment/QueryResult/EditableAttachment.php
Progi1984: A null parameter will trigger a fatal error : 
image

src/Core/Domain/Address/Command/AbstractEditAddressCommand.php
matks: @Progi1984 We dont call the setter so the value stays null because it was initialized `null` when we call `$command = new AddressCommand()`. So there is no fatal error 😉

Progi1984: So the setter must be `string`. The getter and the property must be `string|null`. Are we agree ?

matks: I'm not sure 🤔  .
Imagine the following scenario:
- merchant inputs a `homePhone` data
- so form is populated with this data, and EditAddressCommand is populated with this data
- however merchant has installed a module that handles phone numbers in a specific ERP using an API !
- so the module wants to empty the Core field and put the data in its own database using an API call

If you put `string` mandatory in the setter, we cannot empty a field after it has been set 🤔  consequently limiting the ability of users to manipulate it.

A workaround is to create a ModuleExtendedEditAddressCommand to be used 😄  but it's a little bit more work

Progi1984: But if we accept `null` in parameter, we need to change the signature. It become a BC. So it must be done in another PR because it's not the scope of the PR.

matks: Do you mean **this** is a BC ?
```
    public function setHomePhone(string $homePhone): self
```
⬇️ 
```
    public function setHomePhone(string $homePhone = null): self
```

Since that change is backward-compatible with previous version it's not a BC break IMO 🤔  . BC Break stands for "Backward Compatibility Change". Signature changes are allowed provided they are backward compatible. That is how it sounds to me.

I however agree this PR is getting out-of-scope 👍  so I suggest to make phpstan ignore this AbstractEditAddressCommand file for now, and then in another PR, we can sort it out by finding the best solution 🎖️

tests-legacy/Unit/Adapter/Module/Tab/ModuleTabRegisterTest.php
PierreRambaud: We are using methods not declared in the interface, here, we depend on the Loader type, not the interface.

src/PrestaShopBundle/EventListener/DemoModeEnabledListener.php
PierreRambaud: We can't because we use methods not in the interface, but only in the Session class

src/PrestaShopBundle/EventListener/AccessDeniedListener.php
PierreRambaud: @Progi1984 :sweat_smile: We should always prefer interface when it's possible.
Here, we are using methods not register in the Interface, and the whole code used by the `$session` is in a condition to check if the class is an instance of Session. The interface has no aim here :thinking:

src/Core/Domain/Address/Command/AbstractEditAddressCommand.php
PierreRambaud: @matks changing the signature is a BC. Since PHP can be used as strict mode, and be able to catch `Error`, you create a new behavior here.

src/Core/Domain/Tax/Command/BulkToggleTaxStatusCommand.php
Progi1984: I agree. I put a `mixed` type.

src/PrestaShopBundle/DataCollector/HookRegistry.php
Progi1984: @matks Yes, it's mandatory because the property name is defined in `ModuleCore` and not in the interface.

src/PrestaShopBundle/Translation/Provider/AbstractProvider.php
Progi1984: I fixed the `$theme` and the return value for `getDatabaseLoader`.

src/PrestaShopBundle/EventListener/ModuleActivatedListener.php
Progi1984: Ping @matks 

I changed the signature of __construct.
I accept a `Session` object and not anymore a `SessionInterface`.
I documented the BC Break.

Ok with that ?

Progi1984:
@matks I want to fix all in this PR.

src/PrestaShopBundle/DataCollector/HookRegistry.php
Progi1984: @matks Should I modify the interface or the parameter of the method ?

src/Core/Domain/Tax/Command/BulkToggleTaxStatusCommand.php
Progi1984: :heavy_check_mark:

src/Core/Domain/Address/Command/AbstractEditAddressCommand.php
Progi1984: I think we should change the signature when the need will be there.

src/Adapter/Module/Tab/ModuleTabRegister.php
Progi1984: :heavy_check_mark:

src/PrestaShopBundle/EventListener/AccessDeniedListener.php
Progi1984: Changed the signature. Documented the BC.

PierreRambaud (dismissed):

Progi1984:
Sorry :pray: @PierreRambaud, I rebase :)

PierreRambaud (dismissed):

src/PrestaShopBundle/EventListener/ModuleActivatedListener.php
matks: I approve this change ✅ . It's a "BC break introduced because the initial code was flawed" so it's all right

src/Core/Domain/Address/Command/AbstractEditAddressCommand.php
matks: I think this must be decided by @eternoendless . Personally, I am 50% against ; 50% for.

I'm 50% for because it's clearer 😄 
I'm 50% against because of the "X wants to empty the field" usecase" described here: https://github.com/PrestaShop/PrestaShop/pull/20118#discussion_r464846373

src/PrestaShopBundle/DataCollector/HookRegistry.php
matks: In this PR scope, you are right, you are fixing some ambiguous code. Maybe in a future PR we could fix it in a better way by using an updated interface. However I don't want to burden the scope of this PR more.

matks (dismissed):

matks:
@Progi1984 I approve the PR as I believe the last things I am worried about are not worthy enough to block it.



Here is however the list of items I believe we might need to fix in a future PR:

1. In  `ModuleTabUnregister::unregisterTabs()`, you changed return type to `void` to match with implementation. However I think maybe the initial PHPDoc was what was expected, and the implementation is faulty 😅 . Maybe we should check how `ModuleTabUnregister::unregisterTabs()` is used to find out ?

2. There are 2 places where you replaced a `mixed` input arg by a `bool` or `string:
- `BulkToggleManufacturerStatusCommand.php`
- `ApeCode` 

For both these classes, we used to manually check the type because we did not have php7 type-checking. So we expected a `mixed` and then we asserted it to be another type.
I think the right fix is:
- either keep `mixed`
- or fix the code to _use_ type-checking which allows us to remove the assertion logic

3. We still have the debate about nullable setters or not (https://github.com/PrestaShop/PrestaShop/pull/20118#discussion_r495776646). I think this needs @eternoendless insights as this is a "what if modules used our code in this way" usecase and these are very exploratory ; so it's easy to over-engineer it.

Progi1984:
@matks Thanks for your review

1. It's only PHPDoc. I don't change signature or behavior. I just applied the PHPDoc to the current behavior.
For uses don't be a problem. No use for a possible return value : 
image

2. Correct. I change to mixed. (and pushed ;))

3. I sent the PR in QA waiting for the @eternoendless answer.

PierreRambaud (dismissed):

matks (dismissed):

PierreRambaud (approved):

boubkerbribri:
Tested with the UI Campaign (sanity + functional) 
=> All good

matks (approved):

PR #20126, opened 2020-07-09
Correctly uncheck previous default combination

Amazzing:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | `removeAttr('checked')` doesn't actually uncheck the checkbox, so if you change default combination multiple times, then many inputs`$('.attribute_default_checkbox')` will remain checked, and only the last in DOM tree will be saved as default combination. 
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | --
| How to test?  | Open product sheet with 5-10 combinations. Most probably 1st combination is default. Set 5th combination as default and save the product. There will be **Settings updated** notification, and combination will be updated correctly.<br>After this don't reload page, and set 3rd combination as default and save product again. There will be **Settings updated** notification but default combination will not be updated correctly, you can check that after reloading page, or checking database table. 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20126)

### DIFF
--- a/admin-dev/themes/default/js/bundle/product/product-combinations.js
+++ b/admin-dev/themes/default/js/bundle/product/product-combinations.js
@@ -160,7 +160,7 @@ var combinations = (function() {
             }
           });
 
-          $('.attribute_default_checkbox').removeAttr('checked');
+          $('.attribute_default_checkbox').prop('checked', false);
           getCombinationForm(attributeId)
             .find('input[id^="combination"][id$="_attribute_default"]')
             .prop("checked", true);


Progi1984 (approved):

PierreRambaud (approved):

khouloudbelguith:
Hi @Amazzing,

It is ok :heavy_check_mark: 
https://drive.google.com/file/d/1LBMXbUnMMAVM5KsaVZ2e2Hu1S876YisN/view?usp=sharing

Thanks!

PierreRambaud:
Thanks @Amazzing

PR #20137, opened 2020-07-10
Computing precision: set minimum computing precision value to zero

### LINKED ISSUE #17729: [1.7.7] Order totals not rounded to 0 decimals
**Describe the bug**
In 1.7.6, I have **currency->precision set to 0** and **Store settings > General > Decimal places set to 0 in PS config**. The order gets rounded fine.

In 1.7.7 - and after removal of decimal places from PS config, you set **currency->precision to 0** and the order does not get rounded at all!

In @khouloudbelguith 's video, the order total displayed is 735, but in database it's not rounded and it's 734,57. That's wrong, 734,57 will be sent to payment gateway instead of 735.
https://drive.google.com/file/d/1BrY4FXVVpkx4N9A3TzICuedxSRMXHeh3/view

**Why is it bad**
The customer pays 734.57 instead of 735.00 as it should be and it immediately creates differences between the payment and the invoice from the financial software. More manual labour, additional rouding and complications on invoices, VAT calculation differences etc. etc.

**To Reproduce**
PRESTASHOP VERSION 1.7.6
1. Go to Store settings > General > Number of decimal places and set it to 0.
2. Go to database administration, open ps_currency table, set 'precision' to 0.
3. Order something.
4. Totals are rounded.

PRESTASHOP VERSION 1.7.7
1. Go to International > Localization > Currency. Open up your currency. Set decimal places ('precision' in DB) to 0.
2. Order something.
3. Totals are not rounded.

**Additional information**
PrestaShop version: 1.7.6 -> 1.7.7

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | On prestashop 1.7.7, the computing precision and the display precision (that is to say the precision of the currency) always have the same value, except that there is a minimum computing precision of 2. So if a currency has a precision of ZERO then the computing precision and display precision don't match anymore. This PR fixes this by setting the minimum precision to ZERO, so that it matches all the time. <br><br>In 1.7.8, we would like to make the computing precision always higher than the display precision, this would fix some minor but nonetheless important bugs we have on subtotal calculations. This won't be done in 1.7.7.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #17729
| How to test?  | See issue #17729, I'm not sure there is much to test. Maybe setting the currency's precision to 0 in database, and then test that prices are well rounded, even after percent reductions?

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20137)

### DIFF
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -948,6 +948,8 @@ abstract class PaymentModuleCore extends Module
         $order = new Order();
         $order->product_list = $productList;
 
+        $computingPrecision = Context::getContext()->getComputingPrecision();
+
         if (Configuration::get('PS_TAX_ADDRESS_TYPE') == 'id_address_delivery') {
             $address = new Address((int) $addressId);
             $context->country = new Country((int) $address->id_country, (int) $cart->id_lang);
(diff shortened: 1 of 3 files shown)


PierreRambaud (dismissed):

matks:
@Hlavtox happy ? 😄

Hlavtox:
@matks @matthieu-rolland https://www.youtube.com/watch?v=GC5E8ie2pdM

matks:
> https://www.youtube.com/watch?v=GC5E8ie2pdM

I thought you would send this https://www.youtube.com/watch?v=y6Sxv-sUYtM

matks (dismissed):

Hlavtox:
@matks I wanted to rickroll you, but that would be ungrateful 😄

khouloudbelguith:
Hi @matthieu-rolland,

Steps to reproduce the issue: The EUR currency precision = 0
1. Go to BO => Catalog => discounts => create a new cart rule 15%
2. Go to the FO => Add a product to the cart for example `demo_6`
2.1 Price displayed Tax_included= €35
3. Apply the voucher =15%
3.1 Total (tax incl.) = €30
3.2 Discount = €5
4. Complete the order (same results in the BO => Orders details page)

5. Check the Database 	`Select * from ps_orders where id_order=6`

The `total_shipping` = `total_shipping_tax_incl` =8.4 not rounded

image

In the FO/BO => OK
image

Thanks!

matthieu-rolland:
Thank you for the feedback @khouloudbelguith !  

There must be a piece of code where the computing precision is still hardcoded...  I'll check this out 👍

classes/PaymentModule.php
```diff
@@ -986,29 +988,59 @@ protected function createOrderFromCart(
...
 
-        $order->total_products = (float) $cart->getOrderTotal(false, Cart::ONLY_PRODUCTS, $order->product_list, $carrierId);
-        $order->total_products_wt = (float) $cart->getOrderTotal(true, Cart::ONLY_PRODUCTS, $order->product_list, $carrierId);
-        $order->total_discounts_tax_excl = (float) abs($cart->getOrderTotal(false, Cart::ONLY_DISCOUNTS, $order->product_list, $carrierId));
-        $order->total_discounts_tax_incl = (float) abs($cart->getOrderTotal(true, Cart::ONLY_DISCOUNTS, $order->product_list, $carrierId));
+        $order->total_products = (float) Tools::ps_round(
+            (float) $cart->getOrderTotal(false, Cart::ONLY_PRODUCTS, $order->product_list, $carrierId),
+            $computingPrecision
+        );
+        $order->total_products_wt = (float) Tools::ps_round(
```
sowbiba: why do we need to cast, as ps_round already returns a float ?

sowbiba (commented):
Just 1 question

classes/PaymentModule.php
matks: Initially I did it because I saw in other parts of the code there was a cast.

However indeed `Tools::ps_round()` phpDoc says it returns a float.

When I look at the `ps_round()` code, it seems to rely on ceilf, floorf and `Tools::math_round`
There might be 1 tiny little path in `Tools::math_round` that might return an integer (not sure), all other paths rely on php functions returning floats.

So it seems safe to assert the cast is, indeed, useless 👍

matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/20137/commits/ca6528b2975e891e481b8ee39131171679926d94

jolelievre (approved):

sowbiba (approved):

khouloudbelguith:
Hi @matthieu-rolland,

It is ok :heavy_check_mark: 
image

Thanks!

matks:
Thank you @matthieu-rolland

PR #20143, opened 2020-07-10
Add actionGetProductPropertiesAfterUnitPrice hook

olecorre:
| Questions     | Answers
| Branch?       | develop
| Description?  | Move the hook actionGetProductPropertiesAfter to allow the unit price to be changed if necessary. Basically, the hook is called before the calculation
| Type?         | new feature 
| Category?     | FO 
| BC breaks?    | no
| Deprecations? | no
| How to test? | Use module https://github.com/PrestaShop/ps_qualityassurance. This PR adds a new hook `actionGetProductPropertiesAfterUnitPrice` that is called in the same situation as `actionGetProductPropertiesAfter` but a little bit later so more data are accessible.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20143)

### DIFF
--- a/classes/Hook.php
+++ b/classes/Hook.php
@@ -104,6 +104,7 @@ class HookCore extends ObjectModel
 
         // Controller
         'actionAjaxDieBefore' => ['from' => '1.6.1.1'],
+        'actionGetProductPropertiesAfter' => ['from' => '1.7.8.0'],
     ];
 
     const MODULE_LIST_BY_HOOK_KEY = 'hook_module_exec_list_';
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -5539,6 +5539,12 @@ class ProductCore extends ObjectModel
             }
         }
 
+        Hook::exec('actionGetProductPropertiesAfter', [
+            'id_lang' => $id_lang,
+            'product' => &$row,
+            'context' => $context,
+        ]);
+
         $combination = new Combination($id_product_attribute);
 
         if (0 != $combination->unit_price_impact && 0 != $row['unit_price_ratio']) {
@@ -5552,7 +5558,7 @@ class ProductCore extends ObjectModel
             $row['unit_price'] = 0.0;
         }
 
-         Hook::exec('actionGetProductPropertiesAfter', [
+        Hook::exec('actionGetProductPropertiesAfterUnitPrice', [
             'id_lang' => $id_lang,
             'product' => &$row,
             'context' => $context,
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -5539,12 +5539,6 @@ class ProductCore extends ObjectModel
             }
         }
 
-        Hook::exec('actionGetProductPropertiesAfter', [
-            'id_lang' => $id_lang,
-            'product' => &$row,
-            'context' => $context,
-        ]);
-
         $combination = new Combination($id_product_attribute);
 
         if (0 != $combination->unit_price_impact && 0 != $row['unit_price_ratio']) {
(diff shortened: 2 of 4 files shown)


matks (commented):
Hi @olecorre so I see the idea of this PR and it's a good idea.

I wish I could approve the PR but I see one issue blocking: backward compatibility

## Before this PR

You have
1. `Hook::exec('actionGetProductPropertiesAfter');` is called allowing a module to modify `$row` data.
2. `$row['unit_price_ratio']` is computed following a mathematical operation using $row['price_tax_exc'] , $row['unit_price_ratio']) and $combination->unit_price_impact;
3. `$row['unit_price']` is computed using following a mathematical operation using $row['price'] and $row['unit_price_ratio']

## After the PR

You have
1. `$row['unit_price_ratio']` is computed following a mathematical operation using $row['price_tax_exc'] , $row['unit_price_ratio']) and $combination->unit_price_impact;
2. `$row['unit_price']` is computed using following a mathematical operation using $row['price'] and $row['unit_price_ratio']
3. `Hook::exec('actionGetProductPropertiesAfter');` is called allowing a module to modify `$row` data.

## So there is an issue

If we merge this PR for 1.7.8.0, it means any module using hook `actionGetProductPropertiesAfter` will observe a different behavior between 1.7.7 and 1.7.8 which might result in values `$row['unit_price_ratio']` and `$row['unit_price']` have different results. This can create trouble for these modules.

Could we find another option to improve this code, because your idea is good, without doing it in a backward incompatible manner ?

olecorre:
the easiest way would be to add a hook actionGetProductPropertiesAfterUnitPrice

matks:
> the easiest way would be to add a hook actionGetProductPropertiesAfterUnitPrice

Indeed I was thinking about the same thing 🤔 

And even if we must keep the old hook `actionGetProductPropertiesAfter` to preserve backward compatibility, maybe we can deprecate it so developers know that we expect them to use `actionGetProductPropertiesAfterUnitPrice`. And in PrestaShop 8 we remove `actionGetProductPropertiesAfter`

@PrestaShop/prestashop-core-developers did we deprecate hooks in the past ? 😄 
I think I've seen some.

olecorre:
There's never enough hook 😁

matks:
> There's never enough hook 😁

Don't say that 😄 hooks have a performance impact. Each time a hook is dispatched there is a small overhead of "searching what modules are plugged on it" then every module plugged on it is loaded (if it was not loaded before) and called.

matks:
Found it !

Here is an example of how to deprecate hooks => https://github.com/PrestaShop/PrestaShop/pull/12841/

olecorre:
Good !

matks:
So to go forward, in this PR it is needed:
- to restore the hook like it was before
- deprecate it (see previous message)
- introduce new hook
- register this hook (see how to do it here https://devdocs.prestashop.com/1.7/development/register-new-hook/)

matks:
Also PHP CS Fixer is not happy 😄 https://github.com/PrestaShop/PrestaShop/pull/20143/checks?check_run_id=858514591

olecorre:
I change the PR as soon as I have 5 minutes.

olecorre:
I changed the PR,
Add the new hook
Add the old hook to the deprecated ones

matks (approved):
This PR is perfect 👍   nothing to add

olecorre:
thank @matks

PierreRambaud (approved):

Progi1984 (approved):

SD1982:
LGTM Thanks @olecorre !!

Progi1984:
Thanks @olecorre & @SD1982

PR #20145, opened 2020-07-10
Product management and specific prices new rules in Order

### LINKED ISSUE #18237: Possibility to add multiple time the same product in an order view page
#### Describe the bug

On a order view it is possible to add multiple time the same product and it's not cumulated to the first added a new (same) product is added

#### Expected behavior

When adding again a product in a order the only the quantity should be updated

#### Steps to Reproduce

1. Go to BO => sell => orders
2. Select any order and click on view
3. On the view order page add a new product
4. Now add the same product added before

**Screenshots**
multiple time added product

**Screenrecord**
https://drive.google.com/file/d/1xJ5hflon-UT9KaAgNfgb2D3OnBo-x-r1/view?usp=sharing

#### Additional information
* PrestaShop version: 1.7.7.x
* PHP version: 7.2

### LINKED ISSUE #19490: Wrong price displayed before adding a product on a order

### LINKED ISSUE #20010: When I add an existing product in the order's BO
### Wanted behaviors

1) When I add a product already present in the order from the back office, the addition is blocked by displaying the following error message (EM): "Your product is already present in your order.".

If Product A at 10 euro
And I add product A again
Then blocks the addition and displays the error message

2) When the order has several invoices, it is possible to add an already existing product in the order if they are related to a different invoice. 
There is only one possible price per product even if they are from different invoices.  Therefore, if a product is added with a different price in another invoice, the price will be updated on the first invoice. To avoid confusion, a pop-up window will be displayed before the addition to confirm that the action may have an impact on the other order. 

If Product A at 10 euro in invoice 1
And I add product A in a second invoice with the specific price of 5 euro in the invoice 2
Then Product A of invoice 1 will be at 5 euro after validation of the popup.

If Product A at 10 euro in invoice 1 and 2
And I edit the product A in the invoice with the specific of 5 euro in the invoice 1
Then Product A of the invoice 2 will be at 5 euro after validation of the popup

If Product A at 10 euro in invoice 1 and 2
And I edit the product A with the specific of 5 euro in the invoice 2
Then Product A of invoice 1 will be at 5 euro after validation of the popup

If Product A at 10 euro in invoice 1
And I add Product A in the invoice 1
Then the addition is blocked by displaying the corresponding error message (EM) describe in the first case above.

### Wording

**First error message:**
- one when the user add when there is no invoice: _This product is already in your order, please edit the quantity instead._
- one when the user add when there are some invoices: _This product is already in this invoice #number, please edit the quantity instead._

**Second modal message:**
title = _Edit the price_ in _Admin.Actions_  
confirm_message = _Are you sure you want to edit this product price? It will be applied to all invoices of this order._ in _Admin.Orderscustomers.Notification_
confirm_button_label = _Continue_ in _Admin.Actions_
close_button_label = _Cancel_ in _Admin.Actions_

### LINKED ISSUE #20161: [EPIC] Wrong product prices on adding/editing on the order's BO
To facilitate the tracking in the kanban, its issues are grouped in an epic.

- [x] #20010 - When I add an existing product in the order's BO
- [x] #19490 - Wrong price displayed before adding a product on an order
- [x] #18237 - Possibility to add multiple time the same product
- [x] #20533 - Order / Product price is incorrect after adding a cart rule 
- [x] #20542 - Total products price is not correctly calculated in specifics conditions

Both issues should be fixed by https://github.com/PrestaShop/PrestaShop/pull/20145

### LINKED ISSUE #20533: BO - Order / Product price is incorrect after adding a cart rule
#### Describe the bug

In BO after a product price edit if we add a discount the product price is incorrect

#### Expected behavior

The product price should be correct

#### Steps to Reproduce

1. Create a product with price => 14.30 € for example
2. Go to FO and create an order with the previously created product
3. Go to BO and go to the previously created order detail
4. Edit the produt price => The product price is correcly updated in the order detail
5. Add a discount => 10% for exemple
6. See error the product price is the price before the edit in step 4

**Screenshots**

https://drive.google.com/file/d/1iv8Qkd0SF7olOCXoz17H39Hoa--rEifH/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7
**PS => for 176 it is not possible to add a discount after a product price edit**
* PHP version: 7.1

### LINKED ISSUE #20542: BO - Order detail / Total products price is not correctly calculated in specifics conditions
#### Describe the bug

In BO order detail the total product price of an order is not correctly recalculated after adding a discount

#### Expected behavior

The total should be correctly recalculated

#### Steps to Reproduce

1. Create 2 products with the following parameters : 
                  product n°1 => 10€ taxes included
                  product n°2 => 15€ taxes included
2. Go to FO and create an order with only the product n°1
3. Go to BO in the order detail of the previously created order => the total product price is correct
4. Add the second product => the total product price is correct
5. Add a discount ( 5€ ) => the total product price is correct
6. Edit the price of one of the product => the total product price is correct
7 Add a second discount ( 10%) => **the total product price is not correct**

**Screenshots**

https://drive.google.com/file/d/1819bZo0X_-n5G3DYgwfbIbsez_A9oZsG/view?usp=sharing

**Other steps to reproduce the issue:**

1. Create an order with a product which has a price of 10 euro for example
2. Go to BO > Orders > Order details page > edit the product price with 100 euro
3. Add a discount
4. see error => total products incorrect / Total incorrect

image

If you delete the discount, the total products incorrect / Total incorrect 
image

#### Additional information

* PrestaShop version: 177
* PHP version: 7.1

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | This PR has the purpose of bringin more rules to the product management in order, to sum it up:<br>- duplicate products are not allowed any more in an Order **unless** they are assigned to different invoices<br>- each product has **one single** price even if it has duplicates among invoices (all invoices are updated)<br>- some additional popup are displayed to prevent the user when he is about to change several products prices at once<br>- when adding a product the preset values for prices must match the specific price if present (#19490)
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20010 and fixes #19490 and fixes #18237 (EPIC #20161) and #20533 and #20542
| How to test?  | You can follow the instructions and use cases from the issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20145)


Progi1984:
Travis is :red_circle: for integration tests : 
```
    Features/Scenario/Order/order_cart_rules.feature:161
    Features/Scenario/Order/order_from_bo.feature:80
    Features/Scenario/Order/order_from_bo.feature:97
    Features/Scenario/Order/order_from_bo.feature:149
```

atomiix (commented):
LGTM, Maybe add a test for adding the same product but a different combination?

matthieu-rolland:
@atomiix sorry this PR is a work in progress, I should have added the label 😅 
There is still work to do:

- Edit invoices having the same product (with same combination if any)
- Edit (or refresh?) all same products line in the frontend after adding the product (today it adds a line for the new product, but former products are not edited, now it's needed)
- Update behat tests, add behat tests regarding the effects of creating a new specific price rules

classes/order/OrderInvoice.php
```diff
@@ -161,6 +161,30 @@ public static function getInvoiceByNumber($id_invoice)
...
 
+    /**
+     * Get multiple invoices for a single order (useful for multi-invoices orders)
+     *
+     * @param int $orderId
+     * @return array
+     * @throws PrestaShopDatabaseException
+     * @throws PrestaShopException
+     */
+    public static function getInvoicesByOrderId(int $orderId)
```
jolelievre: Are your you need this? since Order class has a `getInvoicesCollection` method

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -275,4 +275,68 @@ protected function clearTemporarySpecificPrices(array $temporarySpecificPrices):
...
+
+            $specificPrice->price = $priceTaxExcluded;
+            $specificPrice->update();
+
+            return $specificPrice;
+        }
+
+        $specificPrice = new SpecificPrice();
+        $specificPrice->id_shop = 0;
+        $specificPrice->id_cart = 0;
```
jolelievre: I think we should define the cart id as well, or we risk having duplicate entries error in case a customer has specific prices on different orders/carts

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -275,4 +275,68 @@ protected function clearTemporarySpecificPrices(array $temporarySpecificPrices):
...
+        if (!empty($existingSpecificPrice)) {
+
+            $specificPrice = new SpecificPrice($existingSpecificPrice['id_specific_price']);
+
+            $specificPrice->price = $priceTaxExcluded;
+            $specificPrice->update();
+
+            return $specificPrice;
+        }
+
```
jolelievre, line 225: The product price needs to be created ONLY if the price is different from the catalog price So you need to add a check on the price value (it has been done previously in the handlers, I believe we use the tax excluded value as a reference since it's the one stored in database)
Then you have two possible cases:
- prices match but there is an existing SpecificPrice -> remove it from DB
- prices don't match -> create/update the SpecificPrice

In which case the method can probably be renamed as a simple `updateSpecificPrice` as it handles multiple modifications (like `updateCartRules` and such)

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -591,4 +591,86 @@ private function checkProductInStock(Product $product, AddProductToOrderCommand
...
+            $orderDetail = new OrderDetail($row['id_order_detail']);
+            if ($orderDetail->product_id !== (int) $product->id) {
+                continue;
+            }
+            if (!empty($combinationId) && $combinationId !== (int) $orderDetail['product_attribute_id']) {
+                continue;
+            }
+            $orderDetail->unit_price_tax_excl = (float) (string) $priceTaxExcluded;
+            $orderDetail->unit_price_tax_incl = (float) (string) $priceTaxIncluded;
+            $orderDetail->total_price_tax_excl = Tools::ps_round((float) (string) $priceTaxExcluded * $productQuantity, $this->computingPrecision);
```
jolelievre: Careful you mustn't  use the new `$productQuantity` since it's the quantity of the OrderDetail being updated Since the purpose here is to is to updater **other** OrderDetails you must use the `$orderDetail->quantity` instead Or all lines will have the same quantity

And since you only edit the other rows I think you can add a check `if ($row['id_order_detail'] == $orderDetail->id) { continue; }`

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -591,4 +591,86 @@ private function checkProductInStock(Product $product, AddProductToOrderCommand
...
+     *
+     * @param Order $order
+     * @param Product $product
+     * @param int $productQuantity
+     * @param Number $priceTaxIncluded
+     * @param Number $priceTaxExcluded
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
+    private function updateOrderDetails(
```
jolelievre: Maybe this update (along with the OrderInvoice update) should be handled in the service `OrderAmountUpdater` ?

admin-dev/themes/new-theme/js/pages/order/view/order-product-add.js
```diff
@@ -189,14 +189,25 @@ export default class OrderProductAdd {
...
       const modal = new ConfirmModal({
         id: 'modal-confirm-new-invoice',
         confirmTitle: this.invoiceSelect.data('modal-title'),
         confirmMessage: this.invoiceSelect.data('modal-body'),
         confirmButtonLabel: this.invoiceSelect.data('modal-apply'),
         closeButtonLabel: this.invoiceSelect.data('modal-cancel'),
-      }, () => this.addProduct(orderId));
-
+      }, () => {
+        modalEditPrice.show();
```
jolelievre: You need to check if the price is different from other products' price before displaying the warning modal
Else you can add the product directly

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -199,10 +201,8 @@ public function handle(AddProductToOrderCommand $command)
             // Update totals amount of order
             $this->orderAmountUpdater->update($order, $cart, (int) $orderDetail->id_order_invoice);
 
-            // Delete temporary specific prices
-            $this->clearTemporarySpecificPrices($temporarySpecificPrices);
+            $order->update();
```
jolelievre: No need for `$order->update();` it is already done by `orderAmountUpdater`

jolelievre (commented):
I believe it's still in WIP but I commented a few things
Also you managed the js for add product, but don't forget to do the same (confirm modal + udpate prices) for product edition 😉

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -202,6 +202,8 @@ public function handle(AddProductToOrderCommand $command)
             $this->orderAmountUpdater->update($order, $cart, (int) $orderDetail->id_order_invoice);
 
             $order->update();
+
+            $specificPrice->delete();
```
jolelievre: Mmm, the purpose of this PR is to never delete the specific price any more They must stay in database and be updated as long as the price is different from catalog
The only case where you delete them is if the price is changed to match the catalog one

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
matthieu-rolland: You're totally right, but strangely I tested and the quantity and total remained correct even with this code 🤔 

This is really strange, I fixed it as you said anyway.

matthieu-rolland: About the condition, I can't use it because we're looping on orderDetails, `$row['id_order_detail']` and `$orderDetail->id` are always the same.

What I can do is pass the currently updated OrderDetail as a parameter to do this check 👍  (not doing the check is not necessarily bad though, we're just doing a recalculation on this OrderDetail, the values remain the same)

admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
```diff
@@ -157,6 +157,31 @@ export default class OrderProductEdit {
     this.setupListener();
   }
 
+  handleEditProductWithConfirmationModal(event) {
+    const currentPriceTaxExclRaw = $(`#orderProduct_${this.orderDetailId} ${OrderViewPageMap.productEditBtn}`).data('product-price-tax-excl');
```
jolelievre: Seems like you are comparing the edited price from the input with the product price that was stored, so you compare it to itself
But the idea is not to check if the row price was modified, it's to see if there are other rows with the same product but different prices

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -298,4 +301,77 @@ private function getPrecisionFromCart(Cart $cart): int
...
+     *
+     * @param Order $order
+     * @param int $precision
+     *
+     * @throws PrestaShopDatabaseException
+     * @throws PrestaShopException
+     */
+    public function updateOrderInvoices(Order $order, int $precision): void
+    {
+        $orderInvoices = OrderInvoice::getInvoicesByOrderId($order->id);
```
jolelievre: Why not use `$order->getInvoicesCollection` here? instead of creating a new method in the legacy class

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -298,4 +301,77 @@ private function getPrecisionFromCart(Cart $cart): int
...
+
+        foreach ($orderInvoices as $invoice) {
```
jolelievre: You should never compute order totals yourself There are too many business rules (cart rules especially) that can be applied The only one that can do this correct is the Cart itself (or indirectly the ) by using `$cart->getOrderTotal`

Like it's been done in the `updateOrderInvoice` function, it's actually this method that you should modify, so that it doesn't update only one invoice any more but all of them Be careful as well about shipping, it should only be computed for the first invoice

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -298,4 +301,77 @@ private function getPrecisionFromCart(Cart $cart): int
...
+     * @param Product $product
+     * @param int|null $combinationId
+     * @param int $productQuantity
+     * @param Number $priceTaxIncluded
+     * @param Number $priceTaxExcluded
+     *
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
+    public function updateOrderDetailsWithSameProduct(
```
jolelievre: This method along with `updateOrderInvoices` must not be public It's a private method that is called by the `update` method
This way the service only has one end point, and it manages the complexity internally

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -298,4 +301,77 @@ private function getPrecisionFromCart(Cart $cart): int
...
+        Order $order,
+        OrderDetail $updatedOrderDetail,
+        Product $product,
```
jolelievre: You don't need the Product nor the Combination as an input here The OrderDetail already has these informations through its fields `product_id` and `product_attribute_id`

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -275,4 +275,80 @@ protected function clearTemporarySpecificPrices(array $temporarySpecificPrices):
...
+            $order->id_currency,
+            0,
+            0,
+            $productQuantity,
+            $combination ? $combination->id : 0,
+            $order->id_customer,
+            $order->id_cart
+        );
+
+        $productOriginalPrice = New Number((string) $product->price);
```
jolelievre:
```suggestion
        $productOriginalPrice = new Number((string) $product->price);
```

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -275,4 +275,80 @@ protected function clearTemporarySpecificPrices(array $temporarySpecificPrices):
...
+     * @throws \PrestaShopDatabaseException
+     */
+    protected function updateSpecificPrice(
+        Number $priceTaxIncluded,
+        Number $priceTaxExcluded,
+        int $productQuantity,
+        Order $order,
+        Product $product,
+        $combination
+    ): ?SpecificPrice {
```
jolelievre: I don't think you need to return the object, it's never used As long as it's correctly updated in database everything is fine

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -275,4 +275,80 @@ protected function clearTemporarySpecificPrices(array $temporarySpecificPrices):
...
+        $existingSpecificPrice = SpecificPrice::getSpecificPrice(
+            $product->id,
+            0,
+            $order->id_currency,
+            0,
+            0,
+            $productQuantity,
```
jolelievre: I believe you don't need this `$productQuantity` parameter, it is useful when we set SpecificPrice that depends on quantity used (in the Product page). But in this case it is only used to override the catalog price, so you can use 0 as this value this should be compatible with the query in

https://github.com/PrestaShop/PrestaShop/blob/develop/classes/SpecificPrice.php#L554-L556

Because this restriction is only useful when `$specificPrice->from_quantity` is `> 1`

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -56,9 +58,29 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements
...
+     * UpdateProductInOrderHandler constructor.
+     *
+     * @param OrderProductQuantityUpdater $orderProductQuantityUpdater
+     * @param OrderAmountUpdater $orderAmountUpdater
+     */
+    public function __construct(
+        OrderProductQuantityUpdater $orderProductQuantityUpdater,
+        OrderAmountUpdater $orderAmountUpdater
+    ) {
+        $this->computingPrecision = Context::getContext()->getComputingPrecision();
```
jolelievre: Computing precision doesn't depend on the context, it must be computed based on the Order's currency (besides using `Context` class is ugly 😁)

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -198,11 +201,7 @@ public function handle(AddProductToOrderCommand $command)
 
             // Update totals amount of order
             $this->orderAmountUpdater->update($order, $cart, (int) $orderDetail->id_order_invoice);
```
jolelievre, line 192: I believe we made a mistake here, it would be more logic to call the hook after the update (like in the updater handler) or the module won't receive the correct total in the order

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -694,6 +695,44 @@ public function addProductAction(int $orderId, Request $request): Response
...
 
+    /**
+     * @AdminSecurity("is_granted(['read'], request.get('_legacy_controller'))", redirectRoute="admin_orders_index")
+     *
+     * @param int $orderId
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function refreshProductPricesAction(int $orderId, Request $request): Response
```
```suggestion
    public function productPricesAction(int $orderId, Request $request): Response
```

jolelievre: Just a suggestion to have a more generic name, because the controller itself doesn't refresh anything it only returns the price list It's the js that uses this data to refresh

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/orders.yml
```diff
@@ -352,3 +352,14 @@ admin_orders_display_customization_image:
     type: \d+
     name: \w+
     value: \w+
+
+admin_orders_refresh_product_prices:
```
jolelievre:
```suggestion
admin_orders_product_prices:
```

jolelievre: Same suggestion as above for the controller

src/Adapter/Order/OrderAmountUpdater.php
matthieu-rolland: I think it's needed for comparison purpose, in order to check if the current orderDetail has the product we're looking for:

`
if ((int) $orderDetail->product_id !== (int) $product->id) {
   continue;
}
`

admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
matthieu-rolland: Actually from this PR, the same products in an order will always have the same price, since all similar products are systematically updated with the data of the one being updated, so this should not be problematic.

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -275,4 +275,78 @@ protected function clearTemporarySpecificPrices(array $temporarySpecificPrices):
...
+        $specificPrice->price = (float) (string) $priceTaxExcluded;
+        $specificPrice->from_quantity = 1;
+        $specificPrice->reduction = 0;
+        $specificPrice->reduction_type = 'amount';
+        $specificPrice->reduction_tax = !$priceTaxIncluded->equals($priceTaxExcluded);
+        $specificPrice->from = '0000-00-00 00:00:00';
+        $specificPrice->to = '0000-00-00 00:00:00';
+        $specificPrice->add();
+
+        return;
```
jolelievre: This one is useless (the return)

admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
jolelievre: Oh right I didn't understand this, so indeed it works but it's not very clear at first so maybe a comment would be good
Alternatively you will need to make this check on other order details for the "add product" feature because you don't have an existing product row to rely on
You could create a `checkOtherProductPriceMatch` function that loops through the rows and returns false when a matching product has a different price, this could be used for both add/edit product and the function name is self explanatory so no need for a comment to understand the logic behind this modal message

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -275,4 +277,119 @@ protected function clearTemporarySpecificPrices(array $temporarySpecificPrices):
...
+     * @param Number $priceTaxIncluded
+     * @param Number $priceTaxExcluded
+     *
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
+    protected function updateOrderDetailsWithSameProduct(
+        Order $order,
+        OrderDetail $updatedOrderDetail,
+        ?int $combinationId,
```
jolelievre: No need for combinationId either, it can be retrieved via `$updatedOrderDetail->product_attribute_id`

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -275,4 +277,119 @@ protected function clearTemporarySpecificPrices(array $temporarySpecificPrices):
...
+        Number $priceTaxIncluded,
```
jolelievre: It seems like you don't update the `$updatedOrderDetail` so it means its own prices have already been updated before this method is called This also means that `$priceTaxIncluded` and `$priceTaxExcluded` are redundant since you can retieve them via `$updatedOrderDetail->unit_price_tax_excl` and `$updatedOrderDetail->unit_price_tax_incl`
If we can minimize the number of parameters it's always easier to use the method

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -56,9 +57,28 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements
...
+
+    /**
+     * UpdateProductInOrderHandler constructor.
+     *
+     * @param OrderProductQuantityUpdater $orderProductQuantityUpdater
+     * @param OrderAmountUpdater $orderAmountUpdater
+     */
+    public function __construct(
+        OrderProductQuantityUpdater $orderProductQuantityUpdater,
+        OrderAmountUpdater $orderAmountUpdater
```
jolelievre: You no longer need `OrderAmountUpdater` since `updateOrderDetailsWithSameProduct` is a method from the abstract class

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -72,6 +92,9 @@ public function handle(UpdateProductInOrderCommand $command)
 
         try {
             $order = $this->getOrder($command->getOrderId());
+            $cart = Cart::getCartByOrderId($order->id);
+            $this->computingPrecision = $this->getPrecisionFromCart($cart);
```
jolelievre: That's a good idea to use the computing precisions here, but you can use a local variable since it's not used in private functions (it avoids having a "stateful" class if it's not required)
Besides I notices that many computations in this class still use the hard coded `2` precision or even don't round the computations at all (like `$orderDetail->total_price_tax_incl` computation) It would be great to use this `$computingPrecision` everywhere it needs to

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -95,36 +118,35 @@ public function handle(UpdateProductInOrderCommand $command)
...
                 $command->getPriceTaxIncluded(),
                 $command->getPriceTaxExcluded(),
                 $order,
-                $cart,
                 $product,
                 $combination
             );
 
-            if (null !== $specificPrice) {
-                $temporarySpecificPrices[] = $specificPrice;
```
jolelievre, line 116: You can also remove the `$temporarySpecificPrices` variable which is not used any more

jolelievre (commented):
Some new comments for things to be finished ^^

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/add_product_row.html.twig
```diff
@@ -64,7 +64,11 @@
           'data-modal-title': 'Create a new invoice'|trans({}, 'Admin.Orderscustomers.Feature'),
           'data-modal-apply': 'Create'|trans({}, 'Admin.Actions'),
           'data-modal-cancel': 'Cancel'|trans({}, 'Admin.Actions'),
-          'data-modal-body': 'Are you sure you want to create a new invoice?'|trans({}, 'Admin.Orderscustomers.Notification')
+          'data-modal-body': 'Are you sure you want to create a new invoice?'|trans({}, 'Admin.Orderscustomers.Notification'),
+          'data-modal-edit-price-title': 'Edit the price'|trans({}, 'Admin.Actions'),
```
LouiseBonnard:
```suggestion
          'data-modal-edit-price-title': 'Edit the price'|trans({}, 'Admin.Orderscustomers.Feature'),
```

LouiseBonnard: If it is the modal's title and not a CTA, it's best to localize it in _Admin.Orderscustomers.Feature_.

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/add_product_row.html.twig
```diff
@@ -64,7 +64,11 @@
...
+          'data-modal-edit-price-apply': 'Continue'|trans({}, 'Admin.Actions'),
```
LouiseBonnard:
```suggestion
          'data-modal-edit-price-apply': 'Edit'|trans({}, 'Admin.Actions'),
```

LouiseBonnard: For consistency, thanks.

matks:
Failing Behat scenarios
```
--- Failed scenarios:
    Features/Scenario/Order/order_cancel_product.feature:194
    Features/Scenario/Order/order_cart_rules.feature:26
    Features/Scenario/Order/order_from_bo.feature:80
    Features/Scenario/Order/order_from_bo.feature:97
    Features/Scenario/Order/order_from_bo.feature:149
    Features/Scenario/Order/order_out_of_stock.feature:118
    Features/Scenario/Order/order_out_of_stock.feature:145
    Features/Scenario/Order/order_out_of_stock.feature:163
    Features/Scenario/Order/order_out_of_stock.feature:205
    Features/Scenario/Order/order_out_of_stock.feature:235
    Features/Scenario/Order/order_out_of_stock.feature:261
```

Robin-Fischer-PS:
Hi @jolelievre ! I think your PR will also fix this issue : https://github.com/PrestaShop/PrestaShop/issues/18237

Could you confirm that ? Thanks

admin-dev/themes/new-theme/js/pages/order/view/order-prices-refresher.js
```diff
@@ -44,4 +44,42 @@ export default class OrderPricesRefresher {
...
+        });
+      });
+  }
+
+  checkOtherProductPricesMatch(givenPrice, productId, combinationId) {
+    const productRows = document.querySelectorAll('tr.cellProduct');
+    let unmatchingPriceExists = false;
+
+    productRows.forEach((product) => {
+      const productEditBtn = document.querySelector(`#${product.id} ${OrderViewPageMap.productEditBtn}`);
```
jolelievre: This should be in the `OrderViewPageMap` you can define some dynamic selectors with arrow functions, like this
```
productsTableRow: (productId) => `#orderProduct_${productId}`,
```

This would improve the readability a lot, besides you could use jquery like in the rest of the code `const productEditBtn = $(OrderViewPageMap.productEditBtn(product.id))` (maybe the current selector `productEditBtn` should be renamed into `productEditButtons` since it selects multiple buttons, or we just add a new one for the dynamic selector)

Which ultimately allows to get the data as usual `productEditBtn.data('combinationId')` instead of using the dataset Don't get me wrong, everything you did work! It's just to (try ^^) remain consistent with the rest of the code

matthieu-rolland:
> Hi @jolelievre ! I think your PR will also fix this issue : #18237
> 
> Could you confirm that ? Thanks

@Robin-Fischer-PS  yes it will (adding an already existing product will pop an error message asking you to increase the quantity instead)

admin-dev/themes/new-theme/js/pages/order/view/order-product-add.js
```diff
@@ -189,14 +191,32 @@ export default class OrderProductAdd {
     const orderId = $(event.currentTarget).data('orderId');
 
     if (invoiceId === 0) {
+      const combinationId = typeof $(':selected', this.combinationsSelect).val() === 'undefined' ? 0 : $(':selected', this.combinationsSelect).val();
+      const productPriceMatch = this.orderPricesRefresher.checkOtherProductPricesMatch(this.priceTaxIncludedInput.val(), this.productIdInput.val(), combinationId);
+
+      const modalEditPrice = new ConfirmModal({
```
jolelievre: Maybe this modal should be created only in the `if (!productPriceMatch)` block, since it's not used otherwise

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -56,8 +58,15 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements
      */
     private $orderProductQuantityUpdater;
 
-    public function __construct(OrderProductQuantityUpdater $orderProductQuantityUpdater)
-    {
+    /**
+     * UpdateProductInOrderHandler constructor.
+     *
+     * @param OrderProductQuantityUpdater $orderProductQuantityUpdater
+     * @param OrderAmountUpdater $orderAmountUpdater
```
jolelievre: This parameter no longer exists

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/add_product_row.html.twig
jolelievre: It feels weird to click on an 'Edit' button to validate an action, from a UX perspective I would expect a form or an input to appear, this could be confusing But the consistency is important though like you said, maybe something in between 'Validate edition', 'Confirm edition' or 'Apply modification' ?

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -275,4 +277,111 @@ protected function clearTemporarySpecificPrices(array $temporarySpecificPrices):
             $specificPrice->delete();
         }
     }
+
```
jolelievre: You can remove the now useless `clearTemporarySpecificPrices` function

admin-dev/themes/new-theme/js/pages/order/view/order-product-add.js
jolelievre: This can still be optimized, a bit like you did in the edition, the modal doesn't need to be created unless the prices don't match

classes/Currency.php
```diff
@@ -863,13 +863,17 @@ public static function getIdByIsoCode($isoCode, $idShop = 0, $forceRefreshCache
             $query = Currency::getIdByQuery($idShop, $includeDeleted);
             $query->where('iso_code = \'' . pSQL($isoCode) . '\'');
 
-            $result = (int) Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue($query->build());
+            $result = Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue($query->build());
```
jolelievre: This is weird.. I thought this change was from a PR made by @Progi1984 
How come it's in this PR, bad rebase?
(Same for Language modif)

jolelievre: Initial PR https://github.com/PrestaShop/PrestaShop/pull/20398

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -559,4 +550,39 @@ private function checkProductInStock(Product $product, AddProductToOrderCommand
...
+            }
+            $invoicesContainingProduct[] = (int) $orderDetail['id_order_invoice'];
+        }
+
+        if (empty($invoicesContainingProduct)) {
+            return;
+        }
+
+        // If it's a new invoice (or no invoice), the ID is null, so we check if the Order has invoice (in which case
+        // a new one is going to be created) If it doesn't have invoices we do'nt allow adding duplicate OrderDetail
```
jolelievre:
```suggestion
        // a new one is going to be created) If it doesn't have invoices we don't allow adding duplicate OrderDetail
```

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -559,4 +550,39 @@ private function checkProductInStock(Product $product, AddProductToOrderCommand
...
+        if (empty($command->getOrderInvoiceId()) && !$order->hasInvoice()) {
+            throw new DuplicateProductInOrderException('You cannot add this product in the order has it is already present');
```
jolelievre:
```suggestion
            throw new DuplicateProductInOrderException('You cannot add this product in the order as it is already present');
```

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -559,4 +550,39 @@ private function checkProductInStock(Product $product, AddProductToOrderCommand
...
+        }
+
+        // If we are targeting a specific invoice check that the ID has not been found in the OrderDetail list
+        if (!empty($command->getOrderInvoiceId()) && in_array((int) $command->getOrderInvoiceId(), $invoicesContainingProduct)) {
+            throw new DuplicateProductInOrderException('You cannot add this product in the order has it is already present');
```
jolelievre:
```suggestion
            throw new DuplicateProductInOrderException('You cannot add this product in the order as it is already present');
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/View/add_product_row.html.twig
Junebyun: Hi @jolelievre I see your point! But I think we should keep a one-verb label for the confirmation button to stay consistent with the "_**cancel**_" button. 

I looked in the BO and there seems to be the verb "**_update_**"as a confirmation button when you edit an information. So I suggest "**_Update_**" in _**Admin.Actions**_. Wdyt?

image

jolelievre: Seems perfect ;)

matthieu-rolland (dismissed):

NeOMakinG (dismissed):
Front-end part looks fine for me, even if you could improve readability by adding some spaces between conditions and some parts, I don't want to block it for it ^^

classes/Product.php
```diff
@@ -962,6 +962,22 @@ public static function updateIsVirtual($id_product, $is_virtual = true)
         ], 'id_product = ' . (int) $id_product);
     }
 
+    /**
+     * @see ObjectModel::validateField()
+     *
+     * reset static cache (eg unit testing purpose).
+     */
+    public static function resetStaticCache()
```
matks, line 970: 😱

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
matks: In theory we should not follow logic but rather attempt to mimic previous position of hook. So even if previous hook dispatch was not logically placed we must try to be as close as possible to it 😄  else modules behavior will be impacted

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -694,6 +695,43 @@ public function addProductAction(int $orderId, Request $request): Response
...
 
+    /**
+     * @AdminSecurity("is_granted(['read'], request.get('_legacy_controller'))", redirectRoute="admin_orders_index")
+     *
+     * @param int $orderId
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function productPricesAction(int $orderId, Request $request): Response
```
matks: I cannot guess what this method is for 🤔  could we improve the naming ? it seems to return ... some of the informations of the product lists of an order ?

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/orders.yml
```diff
@@ -352,3 +352,14 @@ admin_orders_display_customization_image:
     type: \d+
     name: \w+
     value: \w+
+
+admin_orders_product_prices:
+  path: /{orderId}/products/refresh-prices
```
matks: This route is not refreshing prices, it's only fetching data from the database 🤔   if I am not wrong

```suggestion
  path: /{orderId}/products/get-prices
```

matks (commented):
Code is overall good 👍  added a few suggestions

Only one important is name of controller action and route path that I find misleading

classes/Product.php
jolelievre: Yes indeed ^^ There were some problems with behat suites, the price of a product was static cached (based on its id) So it used the old price when another suite started (DB reset, so same id, the price came from the cache..)
This is really a pain 😅

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: It returns the product prices from an order indeed, it is used to refresh the product prices in the order view page

I think the naming is appropriate (OrderController::productPricesAction matches with its action/behaviour) But I'm probably too familiar with to miss the confusion Do you gave a suggestion for better naming?

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/orders.yml
jolelievre: Indeed it had already been renamed in that sense (the controller function was also `refreshProductPricesAction`), but there must have been a rebase mistake or something

matks (dismissed):

NeOMakinG (dismissed):

khouloudbelguith:
Hi @jolelievre @matthieu-rolland,

After some tests done by Me, @SD1982 and @Robin-Fischer-PS,
We found those issues:

Issue #19490 not fixed

Following the Expected behavior => Issue not fixed

1. The "add interface":
 - displays the new price => NOK
 => it should display the old price

image


2. The price displayed "after the add":
 - The second product displays the new price => NOK (it was ok in the 177x branch) 
 - the price of the first product is updated also with the New price => NOK(it was ok in the 177x branch) 

image

Issue #20010: pop up not displayed

Case: _If Product A at 10 euro in invoice 1
And I add product A in a second invoice with the specific price of 5 euro in the invoice 2
Then Product A of invoice 1 will be at 5 euro after validation of the popup._

Is **KO**: There is only a popup for the creation of a new order, but no popup to confirm that the action may have an impact on the other order.
image

https://drive.google.com/file/d/1Eu_cHNjwdSwJocW-fA_Zex2PuY_vQUMr/view?usp=sharing

Thanks!

jolelievre:
Thanks @khouloudbelguith 

I'm not sure we should follow the description of #19490 scrupulously because it's an old one still assigned to the Epic, but as stated in this issue https://github.com/PrestaShop/PrestaShop/issues/20010 the main purpose of this PR is that we now have ONE possible price for all the products in the order (even when they are in multiple invoices)

Previously we stated that once a price is assigned to a product all subsequent additions will keep the same price, but it prevented the merchant from editing the prices once a product is present, hence the new rule is that you can now edit the price but it also updates the previous prices that's why we display a warning modal

So for these points:

1. once a specific price is used I indeed use it as the default value in the input field, it makes more sense since we encourage the merchant to use the same price as the already present products
2. this is also the expected behaviour, we can edit the price of the new added product but it updates the first one

I would, however, need confirmation from the @PrestaShop/prestashop-product-team to validate this behaviour

As for the last point regarding the modal, this one is indeed a problem If the new price is different from the first one it should display the confirm modal before adding the new product and updating both prices

marionf:
> you can now edit the price but it also updates the previous prices that's why we display a warning modal

This seems logical for me, the main point is to alert the user of his/her change

> 1. Once a specific price is used I indeed use it as the default value in the input field, it makes more sense since we encourage the merchant to use the same price as the already present products
> 2. this is also the expected behaviour, we can edit the price of the new added product but it updates the first one

I am ok with that as it seems to match with the bahavior described by @MatShir in the issue #20010 :

> There is only one possible price per product even if they are from different invoices. Therefore, if a product is added with a different price in another invoice, the price will be updated on the first invoice. To avoid confusion, a pop-up window will be displayed before the addition to confirm that the action may have an impact on the other order.

I am going to update the expected behavior in the issue #19490 to match with the behavior described in #20010

NeOMakinG (dismissed):

classes/Product.php
PierreRambaud: This kind of method is used in many other objectmodel classes (`resetStaticCache` or `resetCache`)

src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -188,4 +190,35 @@ private function deleteOrderCartRule(
...
+
+    /**
+     * @param Order $order
+     * @param OrderDetail $orderDetail
+     * @param Cart $cart
+     */
+    private function deleteSpecificPrice(Order $order,
+                                         OrderDetail $orderDetail,
+                                         Cart $cart
+    ): void {
```
PierreRambaud:
```suggestion
    private function deleteSpecificPrice(
        Order $order,
        OrderDetail $orderDetail,
        Cart $cart
    ): void {
```

src/PrestaShopBundle/Controller/Admin/ProductController.php
```diff
@@ -1311,9 +1311,13 @@ public function searchProductsAction(Request $request): JsonResponse
             $currencyIsoCode = $currencyId !== null
                 ? Currency::getIsoCodeById((int) $currencyId)
                 : Currency::getIsoCodeById((int) Configuration::get('PS_CURRENCY_DEFAULT'));
+            $orderId = null;
+            if ($request->query->has('order_id')) {
+                $orderId = (int) $request->query->get('order_id');
+            }
```
PierreRambaud, line 1317:
```suggestion
            $orderId = (int) $request->query->get('order_id', null);
```
Use default value instead

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -694,6 +695,43 @@ public function addProductAction(int $orderId, Request $request): Response
...
+                    'unitPriceTaxInclRaw' => $product->getUnitPriceTaxInclRaw(),
+                    'quantity' => $product->getQuantity(),
+                    'availableQuantity' => $product->getAvailableQuantity(),
+                    'totalPrice' => $product->getTotalPrice(),
+                ];
+            }
+        } catch (Exception $e) {
+            return $this->json(
+                ['message' => $this->getErrorMessageForException($e, $this->getErrorMessages($e))],
+                Response::HTTP_BAD_REQUEST
```
PierreRambaud: Why using bad request?

src/PrestaShopBundle/Controller/Admin/ProductController.php
jolelievre: I thought about it but it wouldn't work because null would be cast into int `0`

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: Mmm, good question @matthieu-rolland added this one
But why would you suggest instead? `Response::HTTP_INTERNAL_SERVER_ERROR` ?

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -225,66 +224,70 @@ private function updateOrderCartRules(
...
+            );
+            $invoice->total_products_wt = (float) $cart->getOrderTotal(
+                true,
+                Cart::ONLY_PRODUCTS,
+                $currentInvoiceProducts,
+                $carrierId
+            );
+
+            $invoice->total_discount_tax_excl = $invoice->total_discount_tax_incl = 0;
+            foreach ($order->getCartRules() as $orderCartRuleData) {
```
atomiix: Why not using `$cart->getOrderTotal(true, Cart::ONLY_DISCOUNTS, $currentInvoiceProducts, $carrierId)` here?

src/PrestaShopBundle/Controller/Admin/ProductController.php
PierreRambaud: Ohhh you're right :sweat_smile:

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
PierreRambaud: Depending on the part, if inputs from users are bad => bad request, if something went wrong from the server => internal server

matthieu-rolland: Hm, yes actually, we should return a 500 error, `Response::HTTP_INTERNAL_SERVER_ERROR` seems good.

src/Adapter/Order/OrderAmountUpdater.php
matthieu-rolland: This code existed prior to this PR (but I'm not sure to understand what you want to replace with this call to $cart->getOrderTotal ?)

atomiix: It looks like you're manually iterating through all cart rules to get the total discount for the invoice, I was just suggesting using `getOrderTotal()` with `Cart::ONLY_DISCOUNTS` to simplify but maybe I'm missing something?

classes/Product.php
jolelievre: Yep it actually does the same thing as `ObjectModel::resetStaticCache` but with more clearing related to Product cache only
Which makes me realize I need to fix the PHPDoc here

src/Adapter/Order/OrderAmountUpdater.php
jolelievre: No I think you're right I didn't think about it but it's worth trying it! It tests are still green it's a go! 👍

matthieu-rolland: damn my message was erased, yes I just understood you were talking about the code below your comment.
Hm, maybe @jolelievre knows if there's a limitation keeping us from using `getOrderTotal()` with `Cart::ONLY_DISCOUNTS` ?

(it only iterates through cart rules applied to this order hopefully :D )

matthieu-rolland: Thinking about it, I'm not sure `getOrderTotal()` can calculate discounts only for a specific invoice inside an order, this is probably why @jolelievre did it this way, can you confirm ?

**Edit:** ah it takes a list of products, that are related to this invoice, and this list is always sent to the calculator, so it should be fine...  I guess. 🤔

matthieu-rolland: damn my github is so buggy, messages disappear, and I see new messages much later even after refresh 😓

atomiix (dismissed):

NeOMakinG (dismissed):

PierreRambaud (dismissed):

khouloudbelguith:
Hi @jolelievre,

As reported [here](https://github.com/PrestaShop/PrestaShop/issues/19490): 
_When I add the same product in the order that was previously discounted, the price of the product on the add interface and after the add action should remains the same (the **discounted one**)_

In my case, No it shows always the new price and not the discounted one.
I attached a screen record
https://drive.google.com/file/d/16jCNJ2HWJkto0XIA66YlrelaS_E1wXNg/view?usp=sharing
image


Also, following this [comment](https://github.com/PrestaShop/PrestaShop/issues/20010#issuecomment-653095522)
_one when the user add when there are some invoices (This product is already in the invoice #number, please edit the quantity instead.)_

In my case, the message error is incorrect
image
image


Thanks!

jolelievre:
Hi @khouloudbelguith 

I'm not sure I understand what you mean by the "new price", do you mean you changed the price in the catalog after adding it?
Also what do you mean by the "discounted price", did you use the input in the order view page when adding or editing the product? or is a discount that was related to a CartRule or maybe a Specific price defined in the product page?

It's just to know if we are talking about the same thing, because Specific price is used in two ways:
- via Specific price rules in product pages (e.g: set a price for a country, a group of customers, ...)
- used in Order view page edition to override the default price (from catalog)

This can be confusing so I want to be sure we're on the same page

jolelievre:
From your video, it seems like you set a discount in the product page using the Specific price rules, which is not the scope of this PR Although I'm gonna check at the behaviour, it seems it creates unexpected behaviour

khouloudbelguith:
Hi @jolelievre,

I'm following these steps as described [here:](https://github.com/PrestaShop/PrestaShop/issues/19490)

1. Product demo_3: 34.46 Euro reduce by 20% by a specific Price
2. Add the product to an order
3. Remove the specific price (delete the specific price from Product page)
4. Add the same product again

**Expected behavior**
Displays price on the add interface:

- 28.72 € tax excl
- 34.46 € tax incl

=> In my case => **NOK**
Displays price on the add interface:

- 35.9 € tax excl
- 43.08 € tax incl

image


@MatShir thanks to correct me if I'm wrong.

Thanks!

MatShir:
@khouloudbelguith I'm fine with the error message (for the time we have left, we don't need to complexify the behaviors for a small added value)
And I have updated the scenario because actually deleting the specific price is not on this PR scope. We need just to check if the edited price on the order is displayed as well on the add interface.

jolelievre:
I created this issue to manage the improvement of the error message later https://github.com/PrestaShop/PrestaShop/issues/20612

Feel free to edit it if something is missing

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -243,15 +234,81 @@ protected function updateSpecificPrice(
...
+            SpecificPrice::ORDER_DEFAULT_DATE,
+            false,
+            $order->id_cart
+        );
+
+        if (!empty($existingSpecificPriceId)) {
+            return new SpecificPrice($existingSpecificPriceId);
+        }
+
+        return null;
```
PierreRambaud:
```suggestion
        if (empty($existingSpecificPriceId)) {
            return null;
        }

        return new SpecificPrice($existingSpecificPriceId);
```

Fail fast instead, at least in the future if we add a condition for failing it will be easier :)

jolelievre: Seriously there is no gain in this one 😅. But just to avoid losing time on this +1month long PR I'll go for it 😉

PierreRambaud: It was just a suggestion this is why it's just a comment and not a "Request change" :sweat_smile:

PierreRambaud (dismissed):

NeOMakinG (dismissed):

NeOMakinG (dismissed):

PierreRambaud (dismissed):

khouloudbelguith:
Hi @jolelievre,

I found this issue related to this ticket: https://github.com/PrestaShop/PrestaShop/issues/20010
The pop up of validation is not displayed following these steps:
1. You have already an order with multi-invoices (Product1 => IN00018) and (Product2 => IN00019)
image
2. Try to add Product1 in the second invoice (IN00019) with editing the price
image
3. Click on Add
4. See error => there is no popup validation

https://drive.google.com/file/d/1nRpUgKUlceLjTSBre49af-RZvUyS31G1/view?usp=sharing
Thanks!

jolelievre:
Hi @khouloudbelguith 
Thank you I'll look into it

jolelievre:
Hi @khouloudbelguith,

I just fixed the bug it's ok now, actually the price was only checked when adding a product to a new invoice not to an existing one..

PierreRambaud (dismissed):

NeOMakinG (dismissed):

khouloudbelguith:
Hi @jolelievre,

I found this issue following these steps:
1. Create an order with a Product1
2. Change the status of the Order to payment accepted
3. Try to add the same product to the order with editing the price
4. Click Add
image
5. See error => the pop up "`Edit the price`" is displayed instead of this alert "`Error! This product is already in your order, please edit the quantity instead.`"
image

At this time, if you click **Update** => the right alert is displayed.

A related issue following these steps:

1. Create an order with a Product1
2. Try to add the same product to the order with editing the price
image
3. see error => an empty pop up is displayed
image


Thanks!

matks (approved):

PierreRambaud (approved):

PR #20186, opened 2020-07-15
Take product already in cart into account when editing/adding product

### LINKED ISSUE #19855: Invalid check for max quantity when editing product quantity in Order

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Already in cart products weren't take into account when checking available quantity and as during the `updateQty` process, the second quantity check should be skipped because it doesn't take products in an existing order into account. This PR make those corrections.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19855
| How to test?  | See #19855 steps to reproduce, no more fail should happen (even with different combinations when adding/editing). You should also see the quantity and price live update when changing quantity of a product.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20186)


admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
```diff
@@ -45,13 +45,13 @@ export default class OrderProductEdit {
 
   setupListener() {
     this.quantityInput.on('change keyup', (event) => {
-      const quantity = Number(event.target.value);
+      this.quantity = Number(event.target.value);
```
arouiadib, line 48: Thanks for the PR,
I am not sure if it is needed to add  `quantity` as attribute in this class, but only keep the constant for this temporary use. If else, I think this change should also be done in `order-product-add.js`

Progi1984, line 48: Why changing in constant ?

Progi1984 (commented):
A small comment and a question : Can you add a behat test ?

admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
atomiix: @arouiadib it is because of https://github.com/PrestaShop/PrestaShop/blob/1.7.7.x/admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js#L98, that's mandatory to update the price accordingly.
On `order-product-add.js` on the other hand, it's handled differently so it's not needed.

atomiix: See my answer just above 😉 : https://github.com/PrestaShop/PrestaShop/pull/20186#discussion_r455610730

admin-dev/themes/new-theme/js/pages/order/view/order-product-add.js
```diff
@@ -80,7 +80,7 @@ export default class OrderProductAdd {
         const availableOutOfStock = this.availableText.data('availableOutOfStock');
         this.availableText.text(available);
         this.availableText.toggleClass('text-danger font-weight-bold', available < 0);
-        const disableAddActionBtn = quantity <= 0 || (available <= 0 && !availableOutOfStock) ? true : false;
+        const disableAddActionBtn = quantity <= 0 || (available < 0 && !availableOutOfStock) ? true : false;
```
PierreRambaud:
```suggestion
        const disableAddActionBtn = quantity <= 0 || (available < 0 && !availableOutOfStock);
```

Progi1984 (approved):

PierreRambaud (approved):

SD1982:
Hi @atomiix I have noticed there is an unwanted behaviour and it's related to another issue not yet fixed ( this issue : #18237 )

As you can this in the screen record https://drive.google.com/file/d/1mfrfs-2pLo7hzoGBoKlP2FOQA4TsxoMM/view?usp=sharing 

I you have for example a product with 100 units in stock if you create an order with 99 units of this product and if you add again this product you have a second line created for this product and il you edit the line with 99 units it is possible to have a cart with 101 units even with the "Allow ordering of out-of-stock products" set to NO

Screenshot at 11-41-35

atomiix:
This PR is a duplicate of #20111.

PR #20207, opened 2020-07-17
Added hook & gravatar management for avatar employee 

### LINKED ISSUE #18653: Allow users to better handle avatars
### Is your feature request related to a problem?

Currently, as a PrestaShop user, it's not very easy to manage my avatar since it is fetched from the related forum profile (if there is any) and this call slows down the back office.

_Originally posted by @LouiseBonnard in https://github.com/PrestaShop/PrestaShop/pull/18346#issuecomment-611000014_

### Describe the solution you'd like:

Selecting an avatar is quite a basic feature so the idea here is to first offer users a simple way to define their avatar (points 1, 2, 6). Next, it is to make it flexible enough to have a great base for users willing to go further to be able to extend its usage or management (points 3, 4, 5). It includes:

1. :heavy_check_mark: **Cut off the call** to the PrestaShop service to gain performance, fixed by #18622, thanks @matks!

⚠️ do not forget to delete the information notification, cf. https://github.com/PrestaShop/PrestaShop/issues/18653#issuecomment-621243509

2.  :heavy_check_mark: #20114 **Add an image upload field** to the avatar in order to allow users to define their own profile picture - could be the same characteristics as the image upload fields of the Design > Theme & Logo page with the following allowed formats: .gif, .jpg, .png., and .svg

3. :heavy_check_mark:  #20207 **Create a hook** so that this space can also be used by modules or any other extensions requested by users.

4. :heavy_check_mark: #20207 **Manage its calls** in case several images use this hook: the last item using this hook will be the one displayed.

5. :heavy_check_mark: #20207 **Allow a Gravatar option field** for users who want to use it.

6. :heavy_check_mark: #20430 **Extend the avatar usage to profiles** and allow the SuperAdmin profile (only) to assign a default avatar to an employee category in the Team > Profiles section of the back office. 

:notebook: Avatar are displayed in Logs Page

### Alternatives you've considered

#18346

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  |  Added hook & gravatar management for avatar employee 
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18653
| How to test?  | **GRAVATAR** :<br> * Install PrestaShop<br>* For current employee (or new), Gravatar is disabled<br>* Add an avatar<br>* The avatar is updated<br>* Enable Gravatar on profile<br>* If you have Gravatar enabled for your image, that takes your gravatar image, else you have the preston<br>* Disable Gravatar<br>* Your avatar is your image<br><br>**HOOK**<br>* Install module [`ps_qualityassurance`](https://github.com/PrestaShop/ps_qualityassurance)<br>* Register a new hook with `actionOverrideEmployeeImage`<br>* Use this code `$params['imageUrl'] = 'https://api.adorable.io/avatars/285/' .urlencode($params['employee']->email).'.png';`<br>And the hook will override the avatar or gravatar : <br>image

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20207)

### DIFF
--- a/classes/Employee.php
+++ b/classes/Employee.php
@@ -102,6 +102,11 @@ class EmployeeCore extends ObjectModel
     public $reset_password_validity;
 
     /**
+     * @var bool
+     */
+    public $has_enabled_gravatar = false;
+
+    /**
      * @see ObjectModel::$definition
      */
     public static $definition = [
@@ -133,6 +138,7 @@ class EmployeeCore extends ObjectModel
             'id_last_customer' => ['type' => self::TYPE_INT, 'validate' => 'isUnsignedInt'],
             'reset_password_token' => ['type' => self::TYPE_STRING, 'validate' => 'isSha1', 'size' => 40, 'copy_post' => false],
             'reset_password_validity' => ['type' => self::TYPE_DATE, 'validate' => 'isDateOrNull', 'copy_post' => false],
+            'has_enabled_gravatar' => ['type' => self::TYPE_BOOL, 'validate' => 'isBool'],
         ],
     ];
 
@@ -606,6 +612,11 @@ class EmployeeCore extends ObjectModel
 
         // Default Image
         $imageUrl = $imageUrl ?? $default;
+
+        // Gravatar
+        if ($this->has_enabled_gravatar) {
+            $imageUrl = 'https://www.gravatar.com/avatar/' . md5(strtolower(trim($this->email))) . '?d=' . urlencode($default);
+        }
 
         // Hooks
         Hook::exec(
--- a/classes/Employee.php
+++ b/classes/Employee.php
@@ -620,7 +620,7 @@ class EmployeeCore extends ObjectModel
 
         // Hooks
         Hook::exec(
-            'actionGetEmployeeImage',
+            'actionOverrideEmployeeImage',
             [
                 'employee' => $this,
                 'imageUrl' => &$imageUrl,
--- a/classes/Employee.php
+++ b/classes/Employee.php
@@ -593,14 +593,30 @@ class EmployeeCore extends ObjectModel
      */
     public function getImage()
     {
+        $default = Tools::getAdminImageUrl('prestashop-avatar.png');
+        $imageUrl = '';
+
+        // Local Image
         $imagePath = $this->image_dir . $this->id . '.jpg';
         if (file_exists($imagePath)) {
-            return Context::getContext()->link->getMediaLink(
+            $imageUrl = Context::getContext()->link->getMediaLink(
                 str_replace($this->image_dir, _THEME_EMPLOYEE_DIR_, $imagePath)
             );
         }
 
-        return Tools::getAdminImageUrl('prestashop-avatar.png');
+        // Default Image
+        $imageUrl = $imageUrl ?? $default;
+
+        // Hooks
+        Hook::exec(
+            'actionGetEmployeeImage',
+            [
+                'employee' => $this,
+                'imageUrl' => &$imageUrl,
+            ]
+        );
+
+        return $imageUrl;
     }
 
     /**
--- a/install-dev/upgrade/sql/1.7.8.0.sql
+++ b/install-dev/upgrade/sql/1.7.8.0.sql
@@ -12,6 +12,6 @@ INSERT IGNORE INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `po
   (NULL, 'actionCheckoutRender', 'Checkout process render', 'This hook is called when checkout process is constructed', '1'),
   (NULL, 'actionPresentProductListing', 'Product Listing Presenter', 'This hook is called before a product listing is presented', '1'),
   (NULL, 'actionGetProductPropertiesAfterUnitPrice', 'Product Properties', 'This hook is called after defining the properties of a product', '1'),
-  (NULL, 'actionGetEmployeeImage', 'Get Employee Image', 'This hook is used to get the employee image', '1')
+  (NULL, 'actionOverrideEmployeeImage', 'Override Employee Image', 'This hook is used to override the employee image', '1')
 ;
 ALTER TABLE `PREFIX_employee` ADD `has_enabled_gravatar` TINYINT UNSIGNED DEFAULT 0 NOT NULL;
--- a/install-dev/upgrade/sql/1.7.8.0.sql
+++ b/install-dev/upgrade/sql/1.7.8.0.sql
@@ -10,6 +10,7 @@ INSERT IGNORE INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `po
   (NULL, 'actionModuleUninstallBefore', 'Module uninstall before', 'This hook is called before module uninstall process', '1'),
   (NULL, 'actionModuleUninstallAfter', 'Module uninstall after', 'This hook is called at the end of module uninstall process', '1'),
   (NULL, 'actionCheckoutRender', 'Checkout process render', 'This hook is called when checkout process is constructed', '1'),
-  (NULL, 'actionPresentProductListing', 'Product Listing Presenter', 'This hook is called before a product listing is presented', '1')
-  (NULL, 'actionGetProductPropertiesAfterUnitPrice', 'Product Properties', 'This hook is called after defining the properties of a product', '1')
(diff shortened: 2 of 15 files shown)


classes/Employee.php
```diff
@@ -593,14 +599,35 @@ public function isSuperAdmin()
...
+        // Default Image
+        $imageUrl = $imageUrl ?? $default;
+
+        // Gravatar
+        if ($this->has_enabled_gravatar) {
+            $imageUrl = 'https://www.gravatar.com/avatar/' . md5(strtolower(trim($this->email))) . '?d=' . urlencode($default);
+        }
+
+        // Hooks
+        Hook::exec(
```
matks, line 622: Nice 👍

classes/Employee.php
```diff
@@ -593,14 +599,35 @@ public function isSuperAdmin()
...
+            'actionGetEmployeeImage',
```
matks:
```suggestion
            'actionOverrideEmployeeImage',
```

install-dev/upgrade/sql/1.7.8.0.sql
```diff
@@ -10,4 +10,6 @@ INSERT IGNORE INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `po
   (NULL, 'actionModuleUninstallBefore', 'Module uninstall before', 'This hook is called before module uninstall process', '1'),
   (NULL, 'actionModuleUninstallAfter', 'Module uninstall after', 'This hook is called at the end of module uninstall process', '1'),
   (NULL, 'actionPresentProductListing', 'Product Listing Presenter', 'This hook is called before a product listing is presented', '1')
```
matks:
```suggestion
  (NULL, 'actionPresentProductListing', 'Product Listing Presenter', 'This hook is called before a product listing is presented', '1'),
```
Missing comma

src/Core/Domain/Employee/Command/AddEmployeeCommand.php
```diff
@@ -101,7 +107,8 @@ public function __construct(
         $languageId,
         $active,
         $profileId,
-        array $shopAssociation
+        array $shopAssociation,
+        bool $hasEnabledGravatar
```
matks: FInally we reach the moment where we want to update a Command 😄 . And in theory this is a BC break.

@eternoendless finally we have a real world usecase for this issue we knew would come

So ... nullable new argument or new Command ?

src/Core/Domain/Employee/QueryResult/EditableEmployee.php
```diff
@@ -108,7 +114,8 @@ public function __construct(
         $languageId,
         $active,
         $profileId,
-        array $shopAssociation
+        array $shopAssociation,
+        bool $hasEnabledGravatar
```
matks: Another BC break

matks (commented):
Code looks good 👍 only real issue I have is about the BC break. Also I suggest another name for hook

Progi1984:
> Code looks good +1 only real issue I have is about the BC break. Also I suggest another name for hook

@matks You're right for the hook name. And I removed BC Break.

matks (dismissed):

matks:
I approve this PR :)

PierreRambaud (dismissed):

SD1982:
LGTM Everything seems correct except for the replacement image ( preston ) which does not appear because of my local test environment 
Screenshot at 14-20-07

(seen with @Progi1984 ) 
Everything else is working correctly. 
So @LouiseBonnard  and @Junebyun  if the wording is correct it is QA approved !

src/PrestaShopBundle/Resources/views/Admin/Configure/AdvancedParameters/Employee/Blocks/form.html.twig
```diff
@@ -59,6 +59,10 @@
           </div>
         </div>
 
+        {{ ps.form_group_row(employeeForm.has_enabled_gravatar, {}, {
+          'label': 'Use gravatar'|trans({}, 'Admin.Global')
```
LouiseBonnard: What terminology best suits the field: 'Enable Gravatar', 'Connect to Gravatar' or 'Use Gravatar'?

Progi1984: I think `Enable Gravatar` is the best terminology.

`Connect to Gravatar` is not the case, because there is not auth.
`Use Gravatar` could be a second terminology, but I think Enable is better

LouiseBonnard:
```suggestion
          'label': 'Enable Gravatar'|trans({}, 'Admin.Global')
```

Progi1984: @LouiseBonnard Wording applied

LouiseBonnard (dismissed):

matks (dismissed):

src/Core/Form/IdentifiableObject/DataHandler/EmployeeFormDataHandler.php
```diff
@@ -124,7 +124,8 @@ public function create(array $data)
             $data['language'],
             $data['active'],
             $data['profile'],
-            isset($data['shop_association']) ? $data['shop_association'] : $this->defaultShopAssociation
+            isset($data['shop_association']) ? $data['shop_association'] : $this->defaultShopAssociation,
+            $data['has_enabled_gravatar'] ?? false
```
jolelievre: Is the `??` check necessary? You don't use it in the update function
And seems like both provider and Form always include this field, so it should always be present

src/PrestaShopBundle/Resources/config/services/adapter/image.yml
```diff
@@ -17,6 +17,7 @@ services:
       class: 'PrestaShop\PrestaShop\Adapter\Image\Uploader\EmployeeImageUploader'
       arguments:
         - !php/const _PS_EMPLOYEE_IMG_DIR_
+        - !php/const _PS_TMP_IMG_DIR_
```
jolelievre, line 20: No modif on EmployeeImageUploader , the argument was missing?

jolelievre (dismissed):
Small questions, but seems ok for me!

matks (dismissed):

src/PrestaShopBundle/Resources/config/services/adapter/image.yml
Progi1984: Yes, I forgot it in the previous PR.

src/Core/Form/IdentifiableObject/DataHandler/EmployeeFormDataHandler.php
Progi1984: @jolelievre You are right.

matks (approved):

jolelievre (approved):

ksaandev:
Oh this is good to know we can now use gravatar thanks!

install-dev/upgrade/sql/1.7.8.0.sql
```diff
@@ -10,6 +10,8 @@ INSERT IGNORE INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `po
   (NULL, 'actionModuleUninstallBefore', 'Module uninstall before', 'This hook is called before module uninstall process', '1'),
   (NULL, 'actionModuleUninstallAfter', 'Module uninstall after', 'This hook is called at the end of module uninstall process', '1'),
   (NULL, 'actionCheckoutRender', 'Checkout process render', 'This hook is called when checkout process is constructed', '1'),
-  (NULL, 'actionPresentProductListing', 'Product Listing Presenter', 'This hook is called before a product listing is presented', '1')
-  (NULL, 'actionGetProductPropertiesAfterUnitPrice', 'Product Properties', 'This hook is called after defining the properties of a product', '1')
+  (NULL, 'actionPresentProductListing', 'Product Listing Presenter', 'This hook is called before a product listing is presented', '1'),
```
ksaandev, line 13: Why is it `'1'` and not `1` ? `ps_hook` table column `position` is tinyint

PR #20211, opened 2020-07-17
Fix missing pagination for alternative lang url

### LINKED ISSUE #17122: Missing pagination for hreflang
When a page has pagination, the pagination in the hreflang does not appear.
No link to the current page itself is present.

Example :
https://equi-clic.com/gb/10-competition-and-performance?page=2

We have the alternate pages :
`<link rel="alternate" href="https://equi-clic.com/fr/10-performance" hreflang="fr-fr">`
`<link rel="alternate" href="https://equi-clic.com/gb/10-competition-and-performance" hreflang="en-gb">`
`<link rel="alternate" href="https://equi-clic.com/es/10-para-el-concurso" hreflang="es-es">`

We should have the alternate pages :
`<link rel="alternate" href="https://equi-clic.com/fr/10-performance?page=2" hreflang="fr-fr">`
`<link rel="alternate" href="https://equi-clic.com/gb/10-competition-and-performance?page=2" hreflang="en-gb">`
`<link rel="alternate" href="https://equi-clic.com/es/10-para-el-concurso?page=2" hreflang="es-es">`

Guideline : "Each language version must list itself as well as all other language versions.".

PrestaShop version: 1.7.6.2

Sinepel:
| Questions     | Answers
| Branch?       | develop 
| Description?  | This PR makes it possible to correct the absence of the "page" parameter on the `<link rel=" alternate ">` tags.  This modification can also work for other parameters and pages other than the category page.
| Type?         | bug fix 
| Category?     | FO 
| BC breaks?    | yes / no
| Deprecations? | yes / no
| Fixed ticket? | Fixes #17122
| How to test?  | Have a Prestashop with active multilingual. Go to a category page, on page 2 and check if the different meta files related to the alternative versions in other languages contain the "page" parameter. 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20211)

### DIFF
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -1986,7 +1986,7 @@ class FrontControllerCore extends Controller
         foreach ($languages as $lang) {
             $langUrl = $this->context->link->getLanguageLink($lang['id_lang']);
 
-            $params = array();
+            $params = [];
             $url_details = parse_url($langUrl);
 
             if (!empty($url_details['query'])) {
@@ -1996,7 +1996,7 @@ class FrontControllerCore extends Controller
                 }
             }
 
-            $excluded_key = array('isolang', 'id_lang', 'controller', 'fc', 'id_product', 'id_category', 'id_manufacturer', 'id_supplier', 'id_cms');
+            $excluded_key = ['isolang', 'id_lang', 'controller', 'fc', 'id_product', 'id_category', 'id_manufacturer', 'id_supplier', 'id_cms'];
             $excluded_key = array_merge($excluded_key, $this->redirectionExtraExcludedKeys);
             foreach ($_GET as $key => $value) {
                 if (!in_array($key, $excluded_key) && Validate::isUrl($key) && Validate::isUrl($value)) {
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -797,29 +797,8 @@ class FrontControllerCore extends Controller
 
         $match_url = rawurldecode(Tools::getCurrentUrlProtocolPrefix() . $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI']);
         if (!preg_match('/^' . Tools::pRegexp(rawurldecode($canonical_url), '/') . '([&?].*)?$/', $match_url)) {
-            $params = [];
-            $url_details = parse_url($canonical_url);
-
-            if (!empty($url_details['query'])) {
-                parse_str($url_details['query'], $query);
-                foreach ($query as $key => $value) {
-                    $params[Tools::safeOutput($key)] = Tools::safeOutput($value);
-                }
-            }
-            $excluded_key = ['isolang', 'id_lang', 'controller', 'fc', 'id_product', 'id_category', 'id_manufacturer', 'id_supplier', 'id_cms'];
-            $excluded_key = array_merge($excluded_key, $this->redirectionExtraExcludedKeys);
-            foreach ($_GET as $key => $value) {
-                if (!in_array($key, $excluded_key) && Validate::isUrl($key) && Validate::isUrl($value)) {
-                    $params[Tools::safeOutput($key)] = Tools::safeOutput($value);
-                }
-            }
-
-            $str_params = http_build_query($params, '', '&');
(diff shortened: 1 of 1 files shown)


Progi1984 (dismissed):

classes/controller/FrontController.php
```diff
@@ -1984,7 +1984,34 @@ protected function getAlternativeLangsUrl()
...
+                    $params[Tools::safeOutput($key)] = Tools::safeOutput($value);
+                }
+            }
+
+            $str_params = http_build_query($params, '', '&');
+            if (!empty($str_params)) {
+                $langUrl = preg_replace('/^([^?]*)?.*$/', '$1', $langUrl) . '?' . $str_params;
+            } else {
+                $langUrl = preg_replace('/^([^?]*)?.*$/', '$1', $langUrl);
+            }
```
atomiix: I think those lines are the same as [these lines](https://github.com/PrestaShop/PrestaShop/blob/develop/classes/controller/FrontController.php#L800-L822), you can maybe extract that logic to reuse it in both methods?

Sinepel: Yes, I think so, too, but I just don't know how to call it.

A `protected` in this controller sounds good to you? Who'd get called twice, once for canonical, once here.

Progi1984: Ping @atomiix

atomiix: @Sinepel I think you can go `private` with this new method. And yes, It will be called twice like you said ;)

Sinepel: @atomiix I just pushed the fix :)

matthieu-rolland: thanks @Sinepel ! you just need to run cs fixer on your PHP files so that the CI is happy and you'll be all set ;)

classes/controller/FrontController.php
```diff
@@ -1984,9 +1963,47 @@ protected function getAlternativeLangsUrl()
...
         return $alternativeLangs;
     }
+
+    /**
+     * Sanitize / Clean params of an URL
+     *
+     * @param string $url URL to clean
+     * @return string cleaned URL
+     */
+    private function sanitizeUrl($url)
```
PierreRambaud:
```suggestion
    private function sanitizeUrl(string $url): string
```
Because it's develop :heart:

classes/controller/FrontController.php
```diff
@@ -1984,9 +1963,47 @@ protected function getAlternativeLangsUrl()
...
+     * Sanitize / Clean params of an URL
+     *
+     * @param string $url URL to clean
+     * @return string cleaned URL
+     */
+    private function sanitizeUrl($url) 
+    {
+        $params = [];
+        $url_details = parse_url($url);
+        $sanitizedUrl = null;
```
PierreRambaud:
```suggestion
```

PierreRambaud (approved):

atomiix (approved):

matthieu-rolland (approved):

ngodefroy:
@matthieu-rolland  I can see the page number, however the mention "page" doesn't appears as stated in the https://github.com/PrestaShop/PrestaShop/issues/17122.
If the mention page isn't mandatory then its OK for me.
Capture d’écran 2020-08-17 à 14 41 51

atomiix:
@ngodefroy Your screenshot shows that you're not on the second page of a category. Maybe that's why?

ngodefroy:
> @ngodefroy Your screenshot shows that you're not on the second page of a category. Maybe that's why?

Forgive me for that ^^
Here is the right screenshot where you can see it lack the mention of the page 2

Capture d’écran 2020-08-17 à 15 17 39

Sinepel:
@ngodefroy Hi, can you check from the source display (Cmd+U on Firefox Mac for example) and not from the inspector? It doesn't update this data when the page is modified in AJAX

ngodefroy:
Hello Guys !

Just tested, it's finally working for me.


Capture d’écran 2020-08-19 à 15 19 14

Thank you @Sinepel :)

atomiix:
Thanks @Sinepel & @ngodefroy

PR #20235, opened 2020-07-20
Remove useless ini_set for Windows

### LINKED ISSUE #20206: Options that can't be changed via ini_set()
#### Describe the bug

PHP ini settings: `realpath_cache_size` and `realpath_cache_ttl` can only be changed via `php.ini` file or `httpd.conf` but not via `ini_set()`.
Because these options has mode "PHP_INI_SYSTEM"

[These options in php manual](https://www.php.net/manual/en/ini.core.php#ini.sect.performance)
> **Changeable** PHP_INI_SYSTEM

[More about PHP_INI_SYSTEM](https://www.php.net/manual/en/configuration.changes.modes.php)
> PHP_INI_SYSTEM Entry can be set in php.ini or httpd.conf

#### Expected behavior
N/A
#### Steps to Reproduce

```php
if (ini_set('realpath_cache_size', '1M') === false) {
    echo 'Cannot set "realpath_cache_size" ini option', PHP_EOL;
}
if (ini_set('realpath_cache_ttl', '300') === false) {
    echo 'Cannot set "realpath_cache_ttl" ini option';
}
```

**Screenshots**

#### Additional information

* PrestaShop version: N/A
* PHP version: N/A
* File where bug occured: https://github.com/PrestaShop/PrestaShop/blob/42833c4097071b46f37b76f6c347fb579447e3e1/classes/Windows.php#L35-L36

PierreRambaud:
| Questions     | Answers
| Branch?       | develop
| Description?  | These configuration variables can only be set by php.ini or httpd.conf and have no effect here. I keep the class in case of someone override the Windows class.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20206
| How to test?  | No need QA.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20235)

### DIFF
--- a/classes/Windows.php
+++ b/classes/Windows.php
@@ -32,7 +32,5 @@ class WindowsCore
 {
     public static function improveFilesytemPerformances()
     {
-        ini_set('realpath_cache_size', '1M');
-        ini_set('realpath_cache_ttl', '300');
     }
 }
--- a/classes/Windows.php
+++ b/classes/Windows.php
@@ -27,6 +27,8 @@
 /**
  * Class Windows, used to improve the experience
  * when using PrestaShop on Windows operating system.
+ *
+ * @deprecated since 1.7.8.0
  */
 class WindowsCore
 {
--- a/config/config.inc.php
+++ b/config/config.inc.php
@@ -55,7 +55,11 @@ if (!file_exists(_PS_ROOT_DIR_ . '/app/config/parameters.yml') && !file_exists(_
 
 require_once $currentDir . DIRECTORY_SEPARATOR . 'bootstrap.php';
 
-/* Improve PHP configuration on Windows */
+/*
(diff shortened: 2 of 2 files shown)


Progi1984 (dismissed):

classes/Windows.php
```diff
@@ -32,7 +32,5 @@ class WindowsCore
 {
     public static function improveFilesytemPerformances()
```
PululuK, line 35: Maybe add deprecated in this case

PierreRambaud: You're right, good idea!

Progi1984 (approved):

atomiix (approved):

PR #20263, opened 2020-07-21
Fix bad redirection after error in address controller

### LINKED ISSUE #20170: BO - Adresses page - Incorrect behavior after creating a new address with an invalid field Address1
#### Describe the bug

In the BO => Addresses page, when we try to create a new address with invalid field Address1 (too long > 128 chars), there is a redirection to the Address list page with an unclear alert
image

#### Expected behavior

After clicking save => no redirection to the Address list page and display a clear alert as the PS1.7.6
image

#### Steps to Reproduce

1. Go to the BO => Addresses page
2. Create a new address
3. Set an invalid field Address 1 ( too long > 128 chars )
4. Click save
5. See error

**Screenshots**

https://drive.google.com/file/d/1uz469CCWwT5tAuPAaS0qqDrTZ7WKbtvV/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7.x, 1770beta1
* PHP version: 7.1

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fix bad redirection after error in address controller
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/20170
| How to test?  | See ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20263)

### DIFF
--- a/src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
@@ -296,8 +296,6 @@ class AddressController extends FrameworkBundleAdminController
             }
         } catch (Exception $e) {
             $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));
-
-            return $this->redirectToRoute('admin_addresses_index');
         }
 
         return $this->render('@PrestaShop/Admin/Sell/Address/add.html.twig', [


Progi1984 (approved):

matthieu-rolland (approved):

PierreRambaud (approved):

khouloudbelguith:
Hi @matks,


The redirection is ok.
But as always we have an unclear alert comparing 1.7.6.
The Address page is migrated in the PS1.7.7.

In the PS1.7.6., we have this
image

The alert is clear

In the PS1.7.7., we have this alert `An error occurred when attempting to update the required fields.`
image

In the same page, if we change the `Alias` or `Address 2` with invalid expression, the message is clear
image


Thanks!

matks:
> But as always we have an unclear alert comparing 1.7.6.
> The Address page is migrated in the PS1.7.7.
> 
> The alert is clear
> 
> In the PS1.7.7., we have this alert `An error occurred when attempting to update the required fields.`

Hi @khouloudbelguith I agree with you that the previous error message was better. But the reason why "address1" field has an unclear error message is because he has a specific status of "required field" that "alias" or "address2" do not have. Required fields can be configured by merchants which is why they have a specific behavior.

I can hack this PR to obtain the same error message as before, however the issue will still be there for other required fields. Actually this is only a symptom of a bigger issue which is "how to handle/display error messages for required fields".

I suggest we only handle the redirection issue in this PR and solve the issue "how to handle/display error messages for required fields" on a global scale in another PR

khouloudbelguith:
@matks, thanks for your feedback.

>I suggest we only handle the redirection issue in this PR and solve the issue "how to handle/display error messages for required fields" on a global scale in another PR

Great idea!
Issue created here: https://github.com/PrestaShop/PrestaShop/issues/20279

It is ok :heavy_check_mark: 

Thanks!

matks:
Thank you @khouloudbelguith 😉

PR #20270, opened 2020-07-21
Customer names must not begin with special characters (including dots)

### LINKED ISSUE #20269: [FO & BO ] You can create account with an invalid customer name
#### Describe the bug

You can create account with an invalid customer name. For example : `. ` or `. i .`

image

#### Expected behavior

Customer names must not begin with special characters (including dots)

#### Steps to Reproduce

1. Go to Create account page (FO or BO)
2. Set Customer Firstname or Lastname = `. ` (dot + space) or `. i .` (dot + space + exclamation + space + dot)
3. Welcome! You have successfully registered despite your rather special name !

**Screenshots**
https://drive.google.com/file/d/1vUoLYw_4YVH5U7SXTOkEMq_7q7kcIlbr/view
#### Additional information

* PrestaShop version: 1.7.7.x & develop
* PHP version: N/A

PululuK:
| Questions     | Answers
| Branch?       | develop
| Description?  | You can create account with an invalid customer name. For example : `. ` or `. i .` .
| Type?         | bug fix
| Category?     | FO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/20269.
| How to test?  | See https://github.com/PrestaShop/PrestaShop/issues/20269.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20270)

### DIFF
--- a/src/Core/ConstraintValidator/CustomerNameValidator.php
+++ b/src/Core/ConstraintValidator/CustomerNameValidator.php
@@ -38,7 +38,7 @@ use Symfony\Component\Validator\Exception\UnexpectedTypeException;
 class CustomerNameValidator extends ConstraintValidator
 {
     const PATTERN_NAME = '/^(?!\s*$)(?:[^0-9!<>,;?=+()\/\\\\@#"°*`{}_^$%:¤\[\]|\.。]|[。\.](?:\s|$))*$/u';
-    const PATTERN_DOT_SPACED = '/[\.。](\s{1}[^\ ]|$)/';
+    const PATTERN_DOT_SPACED = '/(?!^)[\.。](\s{1}[^\ ]|$)/';
 
     /**
      * @var CharacterCleaner
@@ -99,11 +99,6 @@ class CustomerNameValidator extends ConstraintValidator
         if (mb_strpos($name, '.') === false && mb_strpos($name, '。') === false) {
             return true;
         }
-
-        if (mb_strpos($name, '.') === 0 || mb_strpos($name, '。') === 0) {
-            return false;
-        }
-
         $pattern = $this->characterCleaner->cleanNonUnicodeSupport(self::PATTERN_DOT_SPACED);
 
         return (bool) preg_match($pattern, $name);
--- a/src/Core/ConstraintValidator/CustomerNameValidator.php
+++ b/src/Core/ConstraintValidator/CustomerNameValidator.php
@@ -99,6 +99,11 @@ class CustomerNameValidator extends ConstraintValidator
         if (mb_strpos($name, '.') === false && mb_strpos($name, '。') === false) {
             return true;
         }
+
+        if (mb_strpos($name, '.') === 0 || mb_strpos($name, '。') === 0) {
+            return false;
+        }
+
         $pattern = $this->characterCleaner->cleanNonUnicodeSupport(self::PATTERN_DOT_SPACED);
 
         return (bool) preg_match($pattern, $name);


Progi1984:
@LouiseBonnard I ping you because we work together on #1415. Are you ok with that : "No customer name can begin with a dot" ?

src/Core/ConstraintValidator/CustomerNameValidator.php
```diff
@@ -99,6 +99,11 @@ private function isPointSpacedValid($name)
         if (mb_strpos($name, '.') === false && mb_strpos($name, '。') === false) {
             return true;
         }
+
+        if (mb_strpos($name, '.') === 0 || mb_strpos($name, '。') === 0) {
+            return false;
+        }
+
         $pattern = $this->characterCleaner->cleanNonUnicodeSupport(self::PATTERN_DOT_SPACED);
```
PierreRambaud: Only update `self::PATTERN_DOT_SPACES` by 
`(?!^)[\.。](\s{1}[^\ ]|$)`

PierreRambaud (changes requested):
Improve regex instead :wink:

src/Core/ConstraintValidator/CustomerNameValidator.php
PululuK: @PierreRambaud 

I thought about this but I think it is less efficient (in terms of performance) :
The minimal cases is to  : 
- Check if the name does not contain dots [see here](https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/ConstraintValidator/CustomerNameValidator.php#L99)
- Check if the name contains at least a point at the beginning.

In these two cases it is useless to use `self::PATTERN_DOT_SPACES`

See https://we-love-php.blogspot.com/2014/03/strpos-vs-pregmatch-vs-stripos.html#:~:text=strpos()%20is%203%2D16,than%20using%20a%20long%20string

According to the official php docs for [preg_match](https://www.php.net/manual/fr/function.preg-match.php):

image

PierreRambaud: I agree with preg_match, but here the regex is not only for checking if a char exists or not, it's searching for a position.
Using one preg_match is faster than two strpos + one preg_match :wink:

PululuK: See https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/ConstraintValidator/CustomerNameValidator.php#L102 and  https://github.com/PrestaShop/PrestaShop/blob/develop/src/Core/String/CharacterCleaner.php#L44

In the end you have well : `preg_replace` + `preg_match`

PierreRambaud: I think the call of `$this->characterCleaner->cleanNonUnicodeSupport(` is also useless here, we control the content of the regex, it's not coming from a user :thinking: 
I think using one regex to do the job instead of multiple conditions to do one job should be better, but I'm not the only one who decide :sweat_smile:

LouiseBonnard:
Hi @Progi1984, thanks, suits me fine! Is there any error notification displayed? At least a small _Invalid name_ (localized in _Shop.Forms.Errors_) would be nice for the user.

src/Core/ConstraintValidator/CustomerNameValidator.php
PululuK: @PierreRambaud  done ! Thanks

src/Core/ConstraintValidator/CustomerNameValidator.php
```diff
@@ -38,7 +38,7 @@
 class CustomerNameValidator extends ConstraintValidator
 {
     const PATTERN_NAME = '/^(?!\s*$)(?:[^0-9!<>,;?=+()\/\\\\@#"°*`{}_^$%:¤\[\]|\.。]|[。\.](?:\s|$))*$/u';
-    const PATTERN_DOT_SPACED = '/[\.。](\s{1}[^\ ]|$)/';
+    const PATTERN_DOT_SPACED = '/(?!^)[\.。](\s{1}[^\ ]|$)/';
```
PierreRambaud, line 41: Can we have unit test for this class.
I'm afraid it has never been tested correctly :sweat_smile: 
```
F  . . . 。 Lefevre
```
Look valid

matks:
Hi @PululuK ,

It looks like this PR is not moving anymore. As mentioned in our [Maintainer Guide](https://devdocs.prestashop.com/1.7/project/maintainers-guide/reviewing-pull-requests/#stale-pull-requests), "Pull Requests may be closed after 30 days of inactivity following a request for modifications".

I close this pull request as it seems stale and you might dont want to continue on it. If however you want to continue it, you can re-open 😉 .

Thank you!

PR #20284, opened 2020-07-23
Order address country

### LINKED ISSUE #20217: BO - Orders page - Strange behavior after editing the country
#### Describe the bug

In the BO => Orders => New Order page, it is impossible to edit the country of an address.
When we try to change the country and save, all the addresses do not appear anymore 

#### Expected behavior

The address should be updated

#### Steps to Reproduce

1. Go to BO => Orders 
2. Click on "Add new Order"
3. Create a new customer or choose an old one for example `pub@prestashop.com`
4. add a product to cart
5. try to edit the address (Delivery OR Invoice)
6. change the country
7. save
8. see the error, the block address is empty, there is no more addresses displayed
9. if you try to create a new order for this customer, all previous addresses are not displayed

**Screenshots**

https://drive.google.com/file/d/1uKSvz3VRz2NiTRPOkfN9X7vXN4KB_jV8/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7.x, develop
* PHP version: 7.2

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | After changing the country of a customer's adress it was no longer selectable in the Order create page, this is because disabled countries were filtered Even though they may be used in cart/order history So the filtering has been disabled<br>Additionally the behat tests have been improved, and the state is automatically reset when the country is updated and has no states
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20217
| How to test?  | See the issue for test scenario in create Order page Also you can test to edit an address and set its country to one that has no state, when you check the database you can now notice that the `id_state` column is now equals to `0`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20284)


src/Adapter/Cart/QueryHandler/GetCartInformationHandler.php
```diff
@@ -158,13 +158,6 @@ private function getAddresses(Cart $cart): array
 
         foreach ($customer->getAddresses($cart->id_lang) as $data) {
             $addressId = (int) $data['id_address'];
-            $countryIsEnabled = (bool) Address::isCountryActiveById($addressId);
```
jolelievre, line 161: This is the actual fix for the initial issue, only removing this filtering which prevented selecting these addresses even though they are used in orders/carts

src/Adapter/Address/AbstractAddressHandler.php
```diff
@@ -77,4 +78,20 @@ protected function deleteAddress(Address $address): bool
...
+     */
+    protected function validateAddress(Address $address): void
+    {
+        if (true !== ($validateResult = $address->validateFields(false, true))) {
+            throw new InvalidAddressFieldException(sprintf('Address contains invalid field values: %s', $validateResult));
+        }
+        if (true !== ($validateResult = $address->validateFieldsLang(false, true))) {
+            throw new InvalidAddressFieldException(sprintf('Address contains invalid field values: %s', $validateResult));
+        }
+    }
```
PierreRambaud:
```suggestion
        if (true !== ($validateResult = $address->validateFields(false, true)) 
            || true !== ($validateResult = $address->validateFieldsLang(false, true))
        ) {
            throw new InvalidAddressFieldException(sprintf('Address contains invalid field values: %s', $validateResult));
        }
    }
```

Compute both, do not duplicate throw

tests/Integration/Behaviour/Features/Context/Domain/AddressFeatureContext.php
```diff
@@ -211,14 +225,28 @@ private function updateEditCommandFields(AbstractEditAddressCommand $editAddress
...
             $editAddressCommand->setCountryId($countryId);
 
             /* @var CountryStateByIdChoiceProvider $countryStateChoiceProvider */
-            if (isset($testCaseData['State'])) {
+            if (!empty($testCaseData['State'])) {
                 $countryStateChoiceProvider = $this->getContainer()->get('prestashop.adapter.form.choice_provider.country_state_by_id');
-                $countryStateId = $countryStateChoiceProvider->getChoices(['id_country' => $countryId])[$testCaseData['State']];
-                $editAddressCommand->setStateId($countryStateId);
+                $countryStateList = $countryStateChoiceProvider->getChoices(['id_country' => $countryId]);
+                if (!isset($countryStateList[$testCaseData['State']])) {
```
PierreRambaud: duplicate from line 356, maybe create a method to test if the State exists? :thinking:

PierreRambaud: and line 444

Progi1984 (dismissed):

Progi1984 (dismissed):

tests/Integration/Behaviour/Features/Context/Domain/AddressFeatureContext.php
```diff
@@ -533,4 +508,32 @@ public function editBrandAddressWithFollowingDetails($manufacturerAddressReferen
...
+                throw new RuntimeException(sprintf(
+                    'Cannot find state %s for country %s',
+                    $testCaseData['State'],
+                    $testCaseData['Country']
+                ));
+            }
+            $stateId = (int) $countryStateList[$testCaseData['State']];
+        }
+
+        return $stateId;
```
PierreRambaud:
```suggestion
        if (empty($testCaseData['State'])) {
            return null;
        }
        
        /* @var CountryStateByIdChoiceProvider $countryStateChoiceProvider */
        $countryStateChoiceProvider = $this->getContainer()->get('prestashop.adapter.form.choice_provider.country_state_by_id');
        $countryStateList = $countryStateChoiceProvider->getChoices(['id_country' => $countryId]);
        if (!isset($countryStateList[$testCaseData['State']])) {
            throw new RuntimeException(sprintf(
                'Cannot find state %s for country %s',
                $testCaseData['State'],
                $testCaseData['Country']
            ));
        }
        
        return (int) $countryStateList[$testCaseData['State']];
```

Progi1984 (dismissed):

PierreRambaud (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -519,11 +520,15 @@ private function getErrorMessages(Exception $e): array
...
                 ),
-                AddressConstraintException::INVALID_FIELDS => $this->trans(
+                AddressConstraintException::INVALID_REQUIRED_FIELDS => $this->trans(
                     'An error occurred when attempting to update the required fields.',
                     'Admin.Notifications.Error'
                 ),
             ],
+            InvalidAddressFieldException::class => $this->trans(
+                'Address contains invalid field values',
+                'Admin.Notifications.Error'
```
Junebyun: @jolelievre is this string gonna be a generic one or does it have to be precise? If it is the latter case I would suggest _'Address contains invalid values'_

jolelievre: It's specific for the address domain, but it's kind of generic because it doesn't specify which field You only know "some" fields are invalid
I guess your suggestion is applicable then? `'Address fields contains invalid values`

Junebyun:
```suggestion
                'Address contains invalid values.',
                'Admin.Notifications.Error'
```

jolelievre: @Junebyun done I just corrected the sentence with "Address fields contain~s~ invalid values" (since there are potentially several fields)

jolelievre: Ok I just fixed it again it should be good now ^^

Junebyun: Yes if there can be several fields '**_Address fields contain invalid values._**' is better! Thanks @jolelievre :)

jolelievre: Alright I reverted it back 😉

Progi1984 (dismissed):

PierreRambaud (dismissed):

matks (dismissed):

SD1982:
Hi @jolelievre I have noticed something while testing your PR 

**1-** If while creating an order in BO we edit the delivery address ( country ) , the order is not created fully ( just a cart ) and there is an error message ( it was the same behaviour without the PR ) 

**2-** And if we want to create a another new order ( still from BO ) with the same customer there is a warning message ( "no carrier available" ) even if we set the delvery address country to the initial state 

More details in the screenrecord:
https://drive.google.com/file/d/1uKSvz3VRz2NiTRPOkfN9X7vXN4KB_jV8/view?usp=sharing

SD1982:
Hi @jolelievre after some few more investigation i also found this :
https://drive.google.com/file/d/11-ZoQ_Sq_bisXeBnMh7kh1GGWxIx4zlA/view?usp=sharing

As you can see in the screenrecord I have created a new customer from the FO and if i try to create an order from BO with this freshly created customer and trying to edit the country address the address just disappears

jolelievre:
Thanks @SD1982,

I had trouble reproducing the bug at first, I created a new user in front as you said.
Actually it's not related to a new user, the bug happens when you edit an address that was already assigned to an order.

There were several problems that needed to be fixed:
- the new address was not assigned to the cart, so it was not correctly updated
- when the cart was updated in the create order page, the deleted addresses were ignored even if they were used by the cart

atomiix (dismissed):

src/Adapter/Address/CommandHandler/EditCartAddressHandler.php
```diff
@@ -0,0 +1,172 @@
...
+{
+    /**
+     * @var CommandBusInterface
+     */
+    private $commandBus;
+
+    /**
+     * @param CommandBusInterface $commandBus
+     */
+    public function __construct(CommandBusInterface $commandBus)
```
matks: A handler should not have access to CommandBus. Else we have a recursive dependency loop.

matks: We should rather inject the service `EditCustomerAddressCommandHandler` directly.

src/Adapter/Address/CommandHandler/EditCartAddressHandler.php
```diff
@@ -0,0 +1,172 @@
...
+    {
+        try {
+            $cart = new Cart($command->getCartId()->getValue());
+            if (!Validate::isLoadedObject($cart) || $command->getCartId()->getValue() !== (int) $cart->id) {
+                throw new CartNotFoundException(sprintf('Cart with id "%s" was not found', $command->getCartId()->getValue()));
+            }
+
+            $addressCommand = $this->createEditAddressCommand($command, $cart);
+            /** @var AddressId $addressId */
+            $addressId = $this->commandBus->handle($addressCommand);
```
matks:
```suggestion
            $addressId = $this->editCustomerAddressCommandhandler->handle($addressCommand);
```
We should have handlers call other handlers

src/Core/Form/IdentifiableObject/DataHandler/CartAddressFormDataHandler.php
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function create(array $data)
+    {
+        // Not used for creation, only edition
+    }
+
+    /**
+     * {@inheritdoc}
```
matks, line 61: Probably copy-paste from below

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -473,6 +476,99 @@ public function editOrderAddressAction(int $orderId, string $addressType, Reques
         ]);
     }
 
+    /**
+     * Handles edit form rendering and submission for cart address
+     *
+     * @AdminSecurity(
+     *     "is_granted('update', request.get('_legacy_controller'))",
```
matks, line 496: If I am not wrong, it means `admin_cart_address_edit` route can only be used in "Create an order" page if BO user is granted "Adresses" rights 🤔  isn't it a bit strange ? Maybe we should rather require Order rights here

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
```diff
@@ -473,6 +476,99 @@ public function editOrderAddressAction(int $orderId, string $addressType, Reques
...
+                        ['refreshCartAddresses' => 'false']
+                    );
+                }
+
+                return $this->redirectToRoute('admin_carts_view', ['cartId' => $cartId]);
+            }
+        } catch (Exception $e) {
+            $this->addFlash('error', $this->getErrorMessageForException($e, $this->getErrorMessages($e)));
+
+            return $this->redirectToRoute('admin_carts_view', ['cartId' => $cartId]);
```
matks: Why redirect to cart listing pages if an exception is thrown ?
I suggest removing this `redirectToRoute()`, let's show the merchant the nice error message added in flash bag

tests/Integration/Behaviour/Features/Context/Domain/AddressFeatureContext.php
```diff
@@ -471,6 +587,35 @@ public function editBrandAddressWithFollowingDetails($manufacturerAddressReferen
...
+     * @throws RuntimeException
+     */
+    private function getStateId(int $countryId, array $testCaseData): ?int
+    {
+        if (empty($testCaseData['State'])) {
+            return null;
+        }
+        $stateId = null;
+        /* @var CountryStateByIdChoiceProvider $countryStateChoiceProvider */
+        $countryStateChoiceProvider = $this->getContainer()->get('prestashop.adapter.form.choice_provider.country_state_by_id');
```
matks, line 609: Funny way to retrieve state ID 😄  I think it would be accepted here to use legacy class
Just suggestion, not blocking

tests/Integration/Behaviour/Features/Scenario/Address/customer_address.feature
```diff
@@ -0,0 +1,265 @@
+# ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml -s address --tags customer-address
+@reset-database-before-feature
+@customer-address
+Feature: Address
+  PrestaShop allows BO users to manage customer addresses
+  As a BO user
+  I should be able to customize customer addresses
+
+  # NOTE: these scenario cannot be run independently you need to run them all in the right order
```
matks, line 9: Then maybe it should be one big scenario ?

matks (changes requested):
I am sorry 😞  I see you did a huge work but there are some issues in the code I believe will bring bugs in the future

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
jolelievre: In a way it makes sense.. You can select the address in the order based on the existing ones But if you want to edit the addresses you must have the permission for it No?

tests/Integration/Behaviour/Features/Context/Domain/AddressFeatureContext.php
jolelievre: I admit it's not orthodox ^^ I only used the existing code and refacto it a little actually But it has the advantage to make sure you use a state that belongs to the country (which means our behat tests are actually smarter than our handlers.. 😅)

tests/Integration/Behaviour/Features/Scenario/Address/customer_address.feature
jolelievre: It could but it would become a bit hard to read.. Here each scenario is a different use case, the different scenarii are only dependent because of the total number of addresses in database

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
jolelievre: I was about to do it but I realized it might a bit trickier than I thought 😅 Because the try/catch block includes too much code, so it can fail when we get the data, when the form is created, ...

The risk is that some variable may not be initialized when we try to render the page This is the same problem we had recently with PHPStan, I don't remember how we solved this? Multiple try/catch blocks no?

matks: Yes, multiple try/catch

src/PrestaShopBundle/Controller/Admin/Sell/Address/AddressController.php
matks: Yeah, indeed with this point of view it could make sense.

Let's keep it like this for now and ask @PrestaShop/prestashop-product-team for the long term 👍

src/Core/Form/IdentifiableObject/DataHandler/CartAddressFormDataHandler.php
matks: Still there 😄

src/Adapter/Address/CommandHandler/EditCartAddressHandler.php
matks: Solved and approved 👍

src/Adapter/Address/CommandHandler/EditCartAddressHandler.php
```diff
@@ -0,0 +1,172 @@
...
+     * @throws CannotUpdateCartAddressException
+     * @throws CountryConstraintException
+     * @throws StateConstraintException
+     */
+    public function handle(EditCartAddressCommand $command): AddressId
+    {
+        try {
+            $cart = new Cart($command->getCartId()->getValue());
+            if (!Validate::isLoadedObject($cart) || $command->getCartId()->getValue() !== (int) $cart->id) {
+                throw new CartNotFoundException(sprintf('Cart with id "%s" was not found', $command->getCartId()->getValue()));
```
PierreRambaud:
```suggestion
                throw new CartNotFoundException(sprintf('Cart with id "%d" was not found', $command->getCartId()->getValue()));
```

src/Adapter/Address/CommandHandler/EditCartAddressHandler.php
```diff
@@ -0,0 +1,172 @@
...
+            switch ($command->getAddressType()) {
+                case CartAddressType::DELIVERY_ADDRESS_TYPE:
+                    $cart->id_address_delivery = $addressId->getValue();
+                    break;
+                case CartAddressType::INVOICE_ADDRESS_TYPE:
+                    $cart->id_address_invoice = $addressId->getValue();
+                    break;
+            }
+            if (!$cart->update()) {
+                throw new CannotUpdateCartAddressException(sprintf('An error occurred when updating address for cart "%s"', $command->getCartId()->getValue()));
```
PierreRambaud:
```suggestion
                throw new CannotUpdateCartAddressException(sprintf('An error occurred when updating address for cart "%d"', $command->getCartId()->getValue()));
```

src/Adapter/Address/CommandHandler/EditCartAddressHandler.php
```diff
@@ -0,0 +1,172 @@
...
+            }
+        } catch (PrestaShopException $e) {
+            throw new CannotUpdateCartAddressException(sprintf('An error occurred when updating address for cart "%s"', $command->getCartId()->getValue()));
```
PierreRambaud:
```suggestion
            throw new CannotUpdateCartAddressException(sprintf('An error occurred when updating address for cart "%d"', $command->getCartId()->getValue()));
```

tests/Integration/Behaviour/Features/Context/Domain/AddressFeatureContext.php
```diff
@@ -293,6 +342,55 @@ public function checkOrderAddress(string $orderReference, string $addressReferen
...
+            case 'delivery':
+                $cartAddressId = (int) $cart->id_address_delivery;
+                break;
+        }
+        $expectedAddressId = (int) SharedStorage::getStorage()->get($addressReference);
+
+        Assert::assertEquals(
+            $expectedAddressId,
+            $cartAddressId,
+            sprintf('Invalid cart %s address, expected %s but found %s', $addressType, $expectedAddressId, $cartAddressId)
```
PierreRambaud:
```suggestion
            sprintf('Invalid cart %s address, expected %d but found %d', $addressType, $expectedAddressId, $cartAddressId)
```

matks (dismissed):

PierreRambaud (dismissed):

matks:
@jolelievre broke Linux so I restart the build
```
it:2 https://downloads.apache.org/cassandra/debian 39x InRelease
Fetched 28.7 kB in 2s (10.6 kB/s)
Reading package lists... Done
W: http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/Release.gpg: Signature by key 15866BAFD9BCC4F3C1E0DFC7D69548E1C17EAB57 uses weak digest algorithm (SHA1)
W: GPG error: https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6B05F25D762E3157
W: The repository 'https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease' is not signed.
N: Data from such a repository can't be authenticated and is therefore potentially dangerous to use.
N: See apt-secure(8) manpage for repository creation and user configuration details.
W: There is no public key available for the following key IDs:
6B05F25D762E3157  
```

I thought you were a good guy 😢

PierreRambaud:
> @jolelievre broke Linux so I restart the build
> 
> ```
> it:2 https://downloads.apache.org/cassandra/debian 39x InRelease
> Fetched 28.7 kB in 2s (10.6 kB/s)
> Reading package lists... Done
> W: http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/Release.gpg: Signature by key 15866BAFD9BCC4F3C1E0DFC7D69548E1C17EAB57 uses weak digest algorithm (SHA1)
> W: GPG error: https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6B05F25D762E3157
> W: The repository 'https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease' is not signed.
> N: Data from such a repository can't be authenticated and is therefore potentially dangerous to use.
> N: See apt-secure(8) manpage for repository creation and user configuration details.
> W: There is no public key available for the following key IDs:
> 6B05F25D762E3157  
> ```
> 
> I thought you were a good guy

He's just telling to Linux he can do whatever he wants because he's the best :muscle:

matks (dismissed):

matks (approved):

Progi1984 (approved):

SD1982:
LGTM this PR seems to correctly fix #20217  Thanks @jolelievre !!

matks (approved):

matks:
> QA OK without required approvals !? :trollface:

YOLO

PR #20311, opened 2020-07-24
Mark deprecated order hooks for 1.7.7 as such

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | I deprecated the hooks deprecated in 1.7.7 according to [this devdocs page](https://devdocs.prestashop.com/1.7/modules/core-updates/1.7.7/) and I add the missing alias
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | yes, hooks
| Fixed ticket? | 
| How to test?  | I dont know 😅 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20311)

### DIFF
--- a/classes/Hook.php
+++ b/classes/Hook.php
@@ -102,6 +102,14 @@ class HookCore extends ObjectModel
         'displayProductTabContent' => ['from' => '1.7.0.0'],
         'displayProductTab' => ['from' => '1.7.0.0'],
 
+        // Order page
+        'displayAdminOrderRight' => ['from' => '1.7.7.0'],
+        'displayAdminOrderLeft' => ['from' => '1.7.7.0'],
+        'displayAdminOrderTabOrder' => ['from' => '1.7.7.0'],
+        'displayAdminOrderTabShip' => ['from' => '1.7.7.0'],
+        'displayAdminOrderContentOrder' => ['from' => '1.7.7.0'],
+        'displayAdminOrderContentShip' => ['from' => '1.7.7.0'],
+
         // Controller
         'actionAjaxDieBefore' => ['from' => '1.6.1.1'],
     ];
--- a/install-dev/data/xml/hook_alias.xml
+++ b/install-dev/data/xml/hook_alias.xml
@@ -337,10 +337,13 @@
       <alias>actionAfterDeleteProductInCart</alias>
       <name>actionDeleteProductInCartAfter</name>
     </hook_alias>
-
     <hook_alias id="displayInvoice">
       <alias>displayInvoice</alias>
       <name>displayAdminOrderTop</name>
+    </hook_alias>
+    <hook_alias id="displayBackOfficeOrderActions">
+      <alias>displayBackOfficeOrderActions</alias>
+      <name>displayAdminOrderSide</name>
     </hook_alias>
   </entities>
 </entity_hook_alias>


Progi1984 (approved):

classes/Hook.php
```diff
@@ -102,6 +102,14 @@ class HookCore extends ObjectModel
...
         'displayProductTab' => ['from' => '1.7.0.0'],
 
+        // Order page
+        'displayAdminOrderRight' => ['from' => '1.7.7.0'],
+        'displayAdminOrderLeft' => ['from' => '1.7.7.0'],
+        'displayAdminOrderTabOrder' => ['from' => '1.7.7.0'],
+        'displayAdminOrderTabShip' => ['from' => '1.7.7.0'],
+        'displayAdminOrderContentOrder' => ['from' => '1.7.7.0'],
+        'displayAdminOrderContentShip' => ['from' => '1.7.7.0'],
+
```
PululuK, line 112: @matks We cannot use ` AppKernel::VERSION` 🤔 ?
See https://github.com/PrestaShop/PrestaShop/blob/1.7.7.x/app/AppKernel.php#L33

jolelievre: No because in the next versions `AppKernel::VERSION` will change (1.7.8, 1.7.9, ...) So this information needs to be hard coded so that it never changes

PululuK: @jolelievre  but the next version (1.7.8.x for example) is from the develop branch and not 1.7.7.x ?

jolelievre (approved):

PululuK (approved):

classes/Hook.php
jolelievre: Yes so when we will merge this branch into develop, the const value will have changed But this class will still need to know which exact version deprecated these hooks

PululuK: You are right ! Thank you

eternoendless:
To test hook deprecation:

1. Hook on to the deprecated hooks
2. When the hooks are launched there should be an `E_USER_DEPRECATED` error (should be visible on the debug console if issued from a Symfony page, or by setting up this error as visible on local PHP settings) + a log entry (that you can see in the Logs page)

See `Hook::exec()` and `Tools::throwDeprecated()`.

matks:
> To test hook deprecation:
> 
> 1. Hook on to the deprecated hooks
> 2. When the hooks are launched there should be an `E_USER_DEPRECATED` error (should be visible on the debug console if issued from a Symfony page, or by setting up this error as visible on local PHP settings) + a log entry (that you can see in the Logs page)
> 
> See `Hook::exec()` and `Tools::throwDeprecated()`.

The issue, as noticed by @jolelievre, is that these hooks are removed in 1.7.7 due to the migration of the page. So `Hook::exec()` will not be called.

@jolelievre created an issue to work around that https://github.com/PrestaShop/PrestaShop/issues/20314

jolelievre:
I tested this PR everything is ok the deprecation logs are correctly called and saved

https://drive.google.com/file/d/10ScEWNiiW09h3mCPKjzgUnF5-vkFm0Ga/view

Progi1984:
Perfect @jolelievre & Thanks again to @PierreRambaud for ps_qualityassurance ;)

PR #20320, opened 2020-07-24
Fixed alert system failing when the $error variable is malformed

jimmydupre:
| Questions     | Answers
| Branch?       | develop
| Description?  | this notification is visible on back even is empty with array(null).
| Type?         | bug fix 
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? |
| How to test?  | this notification is visible before modification. I just change one line. <br>I test this bug with any simaltion {$error = array('test')} before line 34.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20320)

### DIFF
--- a/admin-dev/themes/default/template/layout.tpl
+++ b/admin-dev/themes/default/template/layout.tpl
@@ -31,7 +31,7 @@
 		</div>
 	</div>
 {/if}
-{if isset($error)}
+{if count($error)}
   <div class="bootstrap">
     <div class="alert alert-danger">
       <button type="button" class="close" data-dismiss="alert">&times;</button>
--- a/admin-dev/themes/default/template/layout.tpl
+++ b/admin-dev/themes/default/template/layout.tpl
@@ -31,7 +31,7 @@
 		</div>
 	</div>
 {/if}
-{if count($error)}
+{if !empty($error)}
   <div class="bootstrap">
     <div class="alert alert-danger">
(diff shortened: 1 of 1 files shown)


admin-dev/themes/default/template/layout.tpl
```diff
@@ -31,7 +31,7 @@
 		</div>
 	</div>
 {/if}
-{if isset($error)}
+{if count($error)}
```
PierreRambaud, line 34:
```suggestion
{if !empty($error)}
```

NeOMakinG (dismissed):
Should be fine I think

admin-dev/themes/default/template/layout.tpl
davidglezz: I think it's better to use `empty` function. If `$error` was not defined or it is not array, it could give error.
```suggestion
{if !empty($error)}
```

davidglezz (changes requested):

davidglezz (approved):

Progi1984 (approved):

atomiix (approved):

matthieu-rolland (approved):

matthieu-rolland:
Hi @jimmydupre, I tested your PR, and indeed before that we would display an empty error message if for some reason the `$error` variable had let's say `false` as a value. 

Thank you for your contribution :+1:

PR #20335, opened 2020-07-24
Remove discount from order (persist state)

### LINKED ISSUE #20246: [BO] Removing a cart rule associated to a product, or a generic shop cart rule is not possible
#### Describe the bug

This bug was reported on this PR https://github.com/PrestaShop/PrestaShop/pull/20083#issuecomment-659477917
The PR reverts a previously fixed bug https://github.com/PrestaShop/PrestaShop/issues/18379

But the reason behind this is probably very different, it was actually intended to have this behaviour during the PR development. The problem comes from the `CartRule::autoAddToCart` and `CartRule::autoRemove` methods that automatically reassign the discounts after they have been removed.

Before the PR #20083 we created and assigned the cart rules manually, which resulted in many bugs, unexpected behaviour or different behaviour from the FO which relies on legacy logic to auto add/remove cart rules based on their criteria. Now the BO matches this behaviour so it's much more stable, but it removes the possibility to fine tune an order and remove some discounts.

#### Expected behavior

Explain what you expected to happen instead.

#### Steps to Reproduce

2. Create a cart rule:
  2.1. has no code
  2.2. product selection on the previous created product with 100 euro tax included
  2.3. the discount amount of 500 EUR tax included
4. The discount has been added to the order
5. try to remove the discount => impossible

**Screenrecording**

https://drive.google.com/file/d/1EWqM6DyjdPlp1-0XuKnmaQPZrXSLPpGg/view?usp=sharing

**Possible solution**

We could remove the `autoAdd` feature from the BO services/handlers, but it's a bad solution because it could result in unexpected/different behaviour from the FO. Besides even if we don't use `CartRule::autoAddToCart` in the handlers this bethod is called (by default) in other places, like `Cart::getCartRules` `FrontController::init` so it could be called afterwards in many other unexpected situations.

So both BO and FO need to rely on the same logic, and we should update the behaviour of `CartRule::autoAddToCart`. For now the cart rules don't have the possibility to detect that they have been removed from an order or a cart. We need to add this kind of feature, this probably implies changing the DB structure, like adding a soft delete on `OrderCartRule` or adding a additional field in the `cart_cart_rule` to table to perform soft delete.

## Product decision

Before we start digging into this problem, the expected feature and behaviour needs to be discussed with the @PrestaShop/prestashop-product-team This unwanted behaviour was already anticipated for generic Shop rules that apply on all Orders (without code), and they are also automatically added after removal. This was discussed and validated with @MatShir but maybe we under estimated the impact especially for Product cart rules.

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Manage soft delete on OrderCartRule this allows to persist this state and avoid to auto add the cart rule after it has been removed
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20246
| How to test?  | See issue for detailed scenario, but basically create a cart rule associated to a product (without code), when the product is added the discount is automatically added, you should be able to remove the discount while keeping the product You can perform the same test for a cart rule applied on all orders (without code)<br>To make sure the discount is not permanently disabled try the same scenario on several orders, on second order the discount must be applied even if it was removed in the first one

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20335)

### DIFF
--- a/classes/CartRule.php
+++ b/classes/CartRule.php
@@ -526,7 +526,7 @@ class CartRuleCore extends ObjectModel
 		SELECT id_cart_rule
 		FROM `' . _DB_PREFIX_ . 'order_cart_rule` ocr
 		LEFT JOIN `' . _DB_PREFIX_ . 'orders` o ON ocr.`id_order` = o.`id_order`
-		WHERE ocr.`id_cart_rule` = ' . (int) $this->id . '
+		WHERE ocr.`deleted` = 0 AND ocr.`id_cart_rule` = ' . (int) $this->id . '
 		AND o.`id_customer` = ' . (int) $id_customer);
     }
 
@@ -631,7 +631,7 @@ class CartRuleCore extends ObjectModel
             return false;
         }
 
-        // All these checks are necessary when you add the cart rule the first, so when it's not in cart yet
+        // All these checks are necessary when you add the cart rule the first time, so when it's not in cart yet
         // However when it's in the cart and you are checking if the cart rule is still valid (when performing auto remove)
         // these rules are outdated For example:
         //  - the cart rule can now be disabled but it was at the time it was applied, so it doesn't need to be removed
@@ -655,11 +655,12 @@ class CartRuleCore extends ObjectModel
         if ($context->cart->id_customer) {
             $quantityUsed = Db::getInstance()->getValue('
 			SELECT count(*)
-			FROM ' . _DB_PREFIX_ . 'orders o
-			LEFT JOIN ' . _DB_PREFIX_ . 'order_cart_rule od ON o.id_order = od.id_order
-			WHERE o.id_customer = ' . $context->cart->id_customer . '
-			AND od.id_cart_rule = ' . (int) $this->id . '
-			AND ' . (int) Configuration::get('PS_OS_ERROR') . ' != o.current_state
+			FROM `' . _DB_PREFIX_ . 'orders` o
+			LEFT JOIN `' . _DB_PREFIX_ . 'order_cart_rule` ocr ON o.`id_order` = ocr.`id_order`
+			WHERE o.`id_customer` = ' . $context->cart->id_customer . '
+			AND ocr.`deleted` = 0
+			AND ocr.`id_cart_rule` = ' . (int) $this->id . '
+			AND ' . (int) Configuration::get('PS_OS_ERROR') . ' != o.`current_state`
 			');
             // When checking the cart rules present in that cart the request result is accurate
             // When we check if using the cart rule one more time is valid then we increment this value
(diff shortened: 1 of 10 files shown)


classes/CartRule.php
```diff
@@ -835,6 +836,18 @@ public function checkValidity(Context $context, $alreadyInCart = false, $display
...
+        // Check if order cart rule was removed from back office
+        $removed_order_cartRule_id = (int) Db::getInstance()->getValue('
+			SELECT ocr.`id_order_cart_rule`
+			FROM `' . _DB_PREFIX_ . 'order_cart_rule` ocr
+			LEFT JOIN `' . _DB_PREFIX_ . 'orders` o ON ocr.`id_order` = o.`id_order`
+			WHERE ocr.`id_cart_rule` = ' . (int) $this->id . '
+			AND ocr.`deleted` = 1
+			AND o.`id_cart` = ' . $context->cart->id);
+        if ($removed_order_cartRule_id) {
+            return (!$display_error) ? false : $this->trans('You cannot use this voucher because it was manually removed.', [], 'Shop.Notifications.Error');
```
LouiseBonnard:
```suggestion
            return (!$display_error) ? false : $this->trans('You cannot use this voucher because it has manually been removed.', [], 'Shop.Notifications.Error');
```

Progi1984 (dismissed):

matks (dismissed):
PR seems good but we need QA team to perform extensive regression tests against known scenarios using deleted cart rules

atomiix (dismissed):

jolelievre:
Additional scenario: (seen with @MatShir )
- I add a product with associated discount to the order
- The product and its discount are correctly added to the order
- I remove the discount, the product is still present and its full price is applied
- I remove the product from the order
- I add the same product again
- The product is added and the (deleted) discount is reapplied at this moment

From a technical point of view, this means we have to remove the OrderCartRule from the database when the product is removed, so that the persisted delete state doesn't exist any more

PierreRambaud (dismissed):

matks (approved):
The code seems OK. What is scary is not the code inside this PR but the code outside of this PR that might behave in weird ways because it does not ackowledge the new attribute `deleted` on this ObjectModel.

I trust our strong QA, be it analysts or automated tests, to find the issues 👍

matks:
I add the "key feature" label because this new SQL property must go into "notable changes" for the release note

tests/Integration/Behaviour/Features/Scenario/Order/order_cart_rules.feature
```diff
@@ -319,26 +319,9 @@ Feature: Order from Back Office (BO)
...
-      | total_shipping_tax_excl  | 7.0    |
-      | total_shipping_tax_incl  | 7.42   |
-    # Now that the cart rule is disabled it can be removed
-    When cart rule "CartRuleAmountOnEveryOrder" is disabled
-    And I remove cart rule "CartRuleAmountOnEveryOrder" from order "bo_order1"
     Then order "bo_order1" should have 0 cart rule
     And order "bo_order1" should not have cart rule "CartRuleAmountOnEveryOrder"
-    And order "bo_order1" should have 3 products in total
-    And order "bo_order1" should have following details:
+    Then order "bo_order1" should have following details:
```
Progi1984, line 324:
```suggestion
    And order "bo_order1" should have following details:
```

Progi1984 (approved):

SD1982:
LGTM Thanks @jolelievre !!

matks:
Thank you @jolelievre @Progi1984 @SD1982 @PierreRambaud @atomiix @LouiseBonnard

matks:
Note: added "key feature" because this might need to be explained in devdocs / release note

PR #20378, opened 2020-07-29
Use Computing precision instead of hardcoded 2

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Use Computing precision instead of hardcoded 2
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Behat tests should be green

## Context

See initial issue https://github.com/PrestaShop/PrestaShop/issues/17729

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20378)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -431,7 +431,7 @@ class CartCore extends ObjectModel
         if ($cart_vat_amount == 0 || $cartAmountTaxExcluded == 0) {
             return 0;
         } else {
-            return Tools::ps_round($cart_vat_amount / $cartAmountTaxExcluded, Context::getContext()->getComputingPrecision());
+            return Tools::ps_round($cart_vat_amount / $cartAmountTaxExcluded, Tax::TAX_DEFAULT_PRECISION);
         }
     }
 
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -431,7 +431,7 @@ class CartCore extends ObjectModel
         if ($cart_vat_amount == 0 || $cartAmountTaxExcluded == 0) {
             return 0;
         } else {
-            return Tools::ps_round($cart_vat_amount / $cartAmountTaxExcluded, 3);
+            return Tools::ps_round($cart_vat_amount / $cartAmountTaxExcluded, Context::getContext()->getComputingPrecision());
         }
     }
 
@@ -3628,7 +3628,7 @@ class CartCore extends ObjectModel
             }
         }
 
-        $shipping_cost = (float) Tools::ps_round((float) $shipping_cost, 2);
+        $shipping_cost = (float) Tools::ps_round((float) $shipping_cost, Context::getContext()->getComputingPrecision());
         Cache::store($cache_id, $shipping_cost);
 
         return $shipping_cost;
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -4877,7 +4877,7 @@ class ProductCore extends ObjectModel
         );
 
         if (self::$_taxCalculationMethod == PS_TAX_EXC) {
-            $row['price_tax_exc'] = Tools::ps_round($row['price_tax_exc'], 2);
+            $row['price_tax_exc'] = Tools::ps_round($row['price_tax_exc'], Context::getContext()->getComputingPrecision());
             $row['price'] = Product::getPriceStatic(
                 (int) $row['id_product'],
                 true,
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -601,8 +601,8 @@ class OrderCore extends ObjectModel
         $row['tax_calculator'] = $tax_calculator;
         $row['tax_rate'] = $tax_calculator->getTotalRate();
 
-        $row['product_price'] = Tools::ps_round($row['unit_price_tax_excl'], 2);
-        $row['product_price_wt'] = Tools::ps_round($row['unit_price_tax_incl'], 2);
+        $row['product_price'] = Tools::ps_round($row['unit_price_tax_excl'], Context::getContext()->getComputingPrecision());
+        $row['product_price_wt'] = Tools::ps_round($row['unit_price_tax_incl'], Context::getContext()->getComputingPrecision());
 
         $group_reduction = 1;
         if ($row['group_reduction'] > 0) {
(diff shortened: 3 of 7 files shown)


classes/Cart.php
```diff
@@ -431,7 +431,7 @@ public function getAverageProductsTaxRate(&$cartAmountTaxExcluded = null, &$cart
         if ($cart_vat_amount == 0 || $cartAmountTaxExcluded == 0) {
             return 0;
         } else {
-            return Tools::ps_round($cart_vat_amount / $cartAmountTaxExcluded, 3);
```
matks, line 434: This was 3 instead of 2 😛

jolelievre: Mmm, probably because it's a tax rate, si it represents a percentage Maybe in this specific case the precision has to be different

For example if you have a 20.5% tax rate -> 0.205 you indeed need three digit precision The computing precision must be used for prices computation but not in this case

We can create a const to store this value though, like `Tax::DEFAULT_PRECISION` or something similar

jolelievre (commented):
Just a comment regarding the tax rate, I think it's a special case and shouldn't be linked to computing precision nor depend on the currency precision

classes/Cart.php
matks: Feedback valid and applied in https://github.com/PrestaShop/PrestaShop/pull/20378/commits/da189d1ec824ed02d41f3aa4c475c6c187daccb4 👍

jolelievre (dismissed):

atomiix (dismissed):

sowbiba (dismissed):

PierreRambaud (dismissed):

PierreRambaud:
@matks QA is needed here, not sure all behat tests are covering everything :sweat_smile:

Progi1984 (commented):
Why a file @php ?

matks:
> Why a file @php ?

Probably a bad manipulation 😆  but now you have pinged the official PHP team

Nicely catched ! the other ones and me we did not spot it

matks:
> Why a file @php ?

Fixed in https://github.com/PrestaShop/PrestaShop/pull/20378/commits/e9da2b605469d7dae0b37bb2fd0f6b20e8056edc

Progi1984 (approved):

Robin-Fischer-PS:
Hi @PierreRambaud  ! Could you give me some details about the "how to test" ?

Thanks :)

PierreRambaud:
@Robin-Fischer-PS Nope :sweat_smile: 
But since it touches `Cart::getAverageProductsTaxRate`, `Cart::getPackageShippingCost`, `Product::getProductProperties`, `Order::setProductPrices`, `Order::addOrderPayment`, `OrderSlip::getEcoTaxTaxesBreakdown`, and the order history, maybe create cart / order / slip with a custom precision should be enough (@matks can confirm)

matks:
> @Robin-Fischer-PS Nope 😅
> But since it touches `Cart::getAverageProductsTaxRate`, `Cart::getPackageShippingCost`, `Product::getProductProperties`, `Order::setProductPrices`, `Order::addOrderPayment`, `OrderSlip::getEcoTaxTaxesBreakdown`, and the order history, maybe create cart / order / slip with a custom precision should be enough (@matks can confirm)

Indeed, the "spec" of this ticket could be "Check that nothing is wrong or badly rounded:
- on cart computations (= cart total, cart subtotals, discounts ...)
- when order is created by customer (= order total, order subtotals, discounts, invoices ...)

PierreRambaud:
> > @Robin-Fischer-PS Nope sweat_smile
> > But since it touches `Cart::getAverageProductsTaxRate`, `Cart::getPackageShippingCost`, `Product::getProductProperties`, `Order::setProductPrices`, `Order::addOrderPayment`, `OrderSlip::getEcoTaxTaxesBreakdown`, and the order history, maybe create cart / order / slip with a custom precision should be enough (@matks can confirm)
> 
> Indeed, the "spec" of this ticket could be "Check that nothing is wrong or badly rounded:
> 
> * on cart computations (= cart total, cart subtotals, discounts ...)
> * when order is created by customer (= order total, order subtotals, discounts, invoices ...)

And do it with love @Robin-Fischer-PS :heart:

matthieu-rolland (approved):
thank you @matks !

jolelievre (approved):

HartLarsson:
this round calculation that affect also 1.7.6.7 version is terrible.
Every order made with a voucher or  more discounts create a problem also on paypal payment:

image

plus a customer that saw this result on the cart... well... is not really nice:
image

must be 145,91€

matks:
> this round calculation that affect also 1.7.6.7 version is terrible.
> Every order made with a voucher or more discounts create a problem also on paypal payment:
> 
> image
> 
> plus a customer that saw this result on the cart... well... is not really nice:
> image
> 
> must be 145,91€

Hi ! Could you please open a dedicated issue with a scenario we can reproduce ? Without it, it's hard to analyze that 😅 

Without more items, I cant be sure, but my guess would be that these prices are rounded to 2 digits when displayed but computations actually use more than 2 digits. This is why it looks wrong. I think the total is right, what is wrong is the display. But only way to be sure is to submit a bug report 😉

Robin-Fischer-PS:
Thanks @matks ! It's OK for us, QA :heavy_check_mark: :)

matks:
Thank you for review and QA !

HartLarsson:
> > this round calculation that affect also 1.7.6.7 version is terrible.
> > Every order made with a voucher or more discounts create a problem also on paypal payment:
> > image
> > plus a customer that saw this result on the cart... well... is not really nice:
> > image
> > must be 145,91€
> 
> Hi ! Could you please open a dedicated issue with a scenario we can reproduce ? Without it, it's hard to analyze that 😅
> 
> Without more items, I cant be sure, but my guess would be that these prices are rounded to 2 digits when displayed but computations actually use more than 2 digits. This is why it looks wrong. I think the total is right, what is wrong is the display. But only way to be sure is to submit a bug report 😉

hi Matks,

i've already open: https://github.com/PrestaShop/PrestaShop/issues/21071
and https://github.com/PrestaShop/PrestaShop/issues/20889 
Because the problem is even with the right ammount in the database PS is not able to make a simple + and -

PR #19818, opened 2020-06-17
Added Countries Translations (Installation & LocalizationPack)

Progi1984:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Added Countries Translations in Installation Process
| Type?         | improvement
| Category?     | IN
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #12168 & Fixes #18929
| How to test? | * Start the installation process<br>* Choose a language (Arabic Language)<br>* **BEFORE:** The countries list must be in English<br>* **AFTER:** The countries list must be translated (Fixes #12168)<br>* Choose the United States country for example<br>* Go to BO > Catalog > Brands page > Addresses section<br>* The country is displayed in Arabic<br>* Change the employee language to "English"<br>* Go to BO > Catalog > Brands page > Addresses section<br>* **The country is displayed in English** (Fixes https://github.com/PrestaShop/PrestaShop/pull/19818#issuecomment-704930659)<br>* Check for another pages : Orders page, Addresses page, Countries page<br>* In DB, the table ps_country_lang has countries in English and Arabic<br>* In BO, add the localization pack for Spain<br>* **In DB, the table ps_country_lang has countries in Spanish, English and Arabic** (Fixes #18929)<br>

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19818)

PR #20379, opened 2020-07-29
Translate Country in ps_country_lang

### LINKED ISSUE #18929: Wrong country names ps_country_lang
After adding diffirent languages all Country names are in English.

In DB table ps_country_lang I see this
image

- Installed PS 1.7.6.4 English version with Russian localization
- later languages were added in this sequance as in picture.  
image

#### Additional information

* PrestaShop version: 1.7.6.4 English version with Russian localization
* PHP version: Apache/2.4.33 (Win32) OpenSSL/1.1.0h PHP/7.2.6

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | Translate Country in ps_country_lang
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18929
| How to test?  | 0. Build assets<br>1. Install PrestaShop in French<br>2. In DB, the table `ps_country_lang` has countries in French<br>3. In BO, add the localization pack for Algeria<br>4. In DB, the table `ps_country_lang` has countries in French and in Arabic<br>5. In BO, add the localization pack for England<br>6. In DB, the table `ps_country_lang` has countries in French, English and Arabic<br>

:notebook: **BC Break :**
* Added a method `getAllTerritories` to the interface `PrestaShop\PrestaShop\Core\Localization\CLDR\LocaleInterface`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20379)

### DIFF
--- a/classes/Language.php
+++ b/classes/Language.php
@@ -1456,7 +1456,7 @@ class LanguageCore extends ObjectModel implements LanguageInterface
             }
             $territoryData = $territories[$country['iso_code']];
             $sql = 'UPDATE `' . _DB_PREFIX_ . 'country_lang`
-                    SET `name` = "' . pSQL(ucwords($territoryData->getName())) . '"
+                    SET `name` = "' . pSQL(mb_ucwords($territoryData->getName())) . '"
                     WHERE `id_country` = "' . (int) $country['id_country'] . '"
                     AND `id_lang` = "' . (int) $lang->id . '" LIMIT 1;';
             Db::getInstance()->execute($sql);
--- a/classes/Language.php
+++ b/classes/Language.php
@@ -1436,32 +1436,30 @@ class LanguageCore extends ObjectModel implements LanguageInterface
 
     public static function updateMultilangFromCldr($lang)
     {
-        $cldrLocale = $lang->getLocale();
-        $cldrFile = _PS_TRANSLATIONS_DIR_ . 'cldr/datas/main/' . $cldrLocale . '/territories.json';
+        $translatableCountries = Db::getInstance()->executeS('SELECT c.`iso_code`, cl.* FROM `' . _DB_PREFIX_ . 'country` c
+            INNER JOIN `' . _DB_PREFIX_ . 'country_lang` cl ON c.`id_country` = cl.`id_country`
+            WHERE cl.`id_lang` = "' . (int) $lang->id . '" ', true, false);
 
-        if (file_exists($cldrFile)) {
-            $cldrContent = json_decode(file_get_contents($cldrFile), true);
+        if (empty($translatableCountries)) {
+            return;
+        }
 
-            if (!empty($cldrContent)) {
-                $translatableCountries = Db::getInstance()->executeS('SELECT c.`iso_code`, cl.* FROM `' . _DB_PREFIX_ . 'country` c
-                    INNER JOIN `' . _DB_PREFIX_ . 'country_lang` cl ON c.`id_country` = cl.`id_country`
-                    WHERE cl.`id_lang` = "' . (int) $lang->id . '" ', true, false);
+        $container = SymfonyContainer::getInstance();
+        /** @var LocaleRepository $localeRepoCLDR */
+        $localeRepoCLDR = $container->get('prestashop.core.localization.cldr.locale_repository');
+        $localeCLDR = $localeRepoCLDR->getLocale($lang->locale);
+        $territories = $localeCLDR->getAllTerritories();
 
-                if (!empty($translatableCountries)) {
-                    $cldrLanguages = $cldrContent['main'][$cldrLocale]['localeDisplayNames']['territories'];
-
-                    foreach ($translatableCountries as $country) {
-                        if (isset($cldrLanguages[$country['iso_code']]) &&
-                            !empty($cldrLanguages[$country['iso_code']])
-                        ) {
-                            $sql = 'UPDATE `' . _DB_PREFIX_ . 'country_lang`
-                                SET `name` = "' . pSQL(ucwords($cldrLanguages[$country['iso_code']])) . '"
-                                WHERE `id_country` = "' . (int) $country['id_country'] . '" AND `id_lang` = "' . (int) $lang->id . '" LIMIT 1;';
-                            Db::getInstance()->execute($sql);
-                        }
-                    }
-                }
+        foreach ($translatableCountries as $country) {
+            if (empty($territories[$country['iso_code']])) {
+                continue;
             }
+            $territoryData = $territories[$country['iso_code']];
+            $sql = 'UPDATE `' . _DB_PREFIX_ . 'country_lang`
+                    SET `name` = "' . pSQL(ucwords($territoryData->getName())) . '"
+                    WHERE `id_country` = "' . (int) $country['id_country'] . '"
+                    AND `id_lang` = "' . (int) $lang->id . '" LIMIT 1;';
+            Db::getInstance()->execute($sql);
         }
     }
 
--- a/classes/LocalizationPack.php
+++ b/classes/LocalizationPack.php
@@ -402,13 +402,14 @@ class LocalizationPackCore
             foreach ($xml->languages->language as $data) {
                 /** @var SimpleXMLElement $data */
                 $attributes = $data->attributes();
+                $isoCode = (string) $attributes['iso_code'];
                 // if we are not in an installation context or if the pack is not available in the local directory
-                if (Language::getIdByIso($attributes['iso_code']) && !$install_mode) {
+                if (Language::getIdByIso($isoCode) && !$install_mode) {
                     continue;
                 }
 
-                $freshInstall = empty(Language::getIdByIso($attributes['iso_code']));
-                $errors = Language::downloadAndInstallLanguagePack($attributes['iso_code'], $attributes['version'], $attributes, $freshInstall);
+                $freshInstall = empty(Language::getIdByIso($isoCode));
+                $errors = Language::downloadAndInstallLanguagePack($isoCode, $attributes['version'], $attributes, $freshInstall);
                 if ($errors !== true && is_array($errors)) {
                     $this->_errors = array_merge($this->_errors, $errors);
                 }
--- a/classes/LocalizationPack.php
+++ b/classes/LocalizationPack.php
@@ -403,7 +403,7 @@ class LocalizationPackCore
                 /** @var SimpleXMLElement $data */
                 $attributes = $data->attributes();
                 $isoCode = (string) $attributes['iso_code'];
-                // if we are not in an installation context or if the pack is not available in the local directory
+                // if we are not in an installation context or if the pack is available in the local directory
                 if (Language::getIdByIso($isoCode) && !$install_mode) {
                     continue;
                 }
--- a/src/Core/Localization/CLDR/Locale.php
+++ b/src/Core/Localization/CLDR/Locale.php
@@ -107,6 +107,13 @@ final class Locale implements LocaleInterface
      */
     protected $currencies;
 
+    /**
+     * All territories, by ISO code.
+     *
+     * @var TerritoryData[]
+     */
+    protected $territories;
+
     public function __construct(LocaleData $localeData)
     {
         $this->localeCode = $localeData->getLocaleCode();
@@ -118,6 +125,7 @@ final class Locale implements LocaleInterface
         $this->percentPatterns = $localeData->getPercentPatterns();
         $this->currencyPatterns = $localeData->getCurrencyPatterns();
         $this->currencies = $localeData->getCurrencies();
+        $this->territories = $localeData->getTerritories();
     }
 
     /**
@@ -300,5 +308,15 @@ final class Locale implements LocaleInterface
     public function getAllCurrencies()
     {
         return $this->currencies;
+    }
+
+    /**
+     * Get CLDR data of all territories.
+     *
+     * @return TerritoryData[] Data of all territories
+     */
+    public function getAllTerritories()
+    {
+        return $this->territories;
     }
 }
--- a/src/Core/Localization/CLDR/Locale.php
+++ b/src/Core/Localization/CLDR/Locale.php
@@ -125,7 +125,7 @@ final class Locale implements LocaleInterface
         $this->percentPatterns = $localeData->getPercentPatterns();
         $this->currencyPatterns = $localeData->getCurrencyPatterns();
         $this->currencies = $localeData->getCurrencies();
-        $this->territories = $localeData->getTerritories();
+        $this->territories = $localeData->getTerritories() ?: [];
     }
 
     /**
--- a/src/Core/Localization/CLDR/Locale.php
+++ b/src/Core/Localization/CLDR/Locale.php
@@ -315,7 +315,7 @@ final class Locale implements LocaleInterface
      *
      * @return TerritoryData[] Data of all territories
      */
-    public function getAllTerritories()
+    public function getAllTerritories(): array
     {
         return $this->territories;
     }
--- a/src/Core/Localization/CLDR/LocaleData.php
+++ b/src/Core/Localization/CLDR/LocaleData.php
@@ -109,9 +109,9 @@ class LocaleData
     /**
      * All territories, by ISO code.
      *
-     * @var TerritoryData[]
+     * @var TerritoryData[]|null
      */
-    protected $territories = [];
+    protected $territories;
 
     /**
      * Override this object's data with another LocaleData object.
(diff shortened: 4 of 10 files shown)


src/Core/Localization/CLDR/LocaleData.php
```diff
@@ -106,6 +106,13 @@ class LocaleData
      */
     protected $currencies;
 
+    /**
+     * All territories, by ISO code.
+     *
+     * @var TerritoryData[]
+     */
+    protected $territories;
```
PierreRambaud, line 114:
```suggestion
    protected $territories = [];
```
Why not this to remove the null condition?

src/Core/Localization/CLDR/LocaleData.php
```diff
@@ -368,4 +382,24 @@ public function setCurrencies($currencies)
 
         return $this;
     }
+
+    /**
+     * @return TerritoryData[]
+     */
+    public function getTerritories()
```
PierreRambaud:
```suggestion
    public function getTerritories(): array
```
opr `?array` if you keep the null

src/Core/Localization/CLDR/LocaleData.php
Progi1984: :+1:

src/Core/Localization/CLDR/LocaleData.php
Progi1984: Fixed

src/Core/Localization/CLDR/Locale.php
```diff
@@ -310,11 +318,21 @@ public function getCurrencyData($currencyCode)
...
     {
         return $this->currencies;
     }
+
+    /**
+     * Get CLDR data of all territories.
+     *
+     * @return TerritoryData[] Data of all territories
+     */
+    public function getAllTerritories()
```
PierreRambaud:
```suggestion
    public function getAllTerritories(): array
```

src/Core/Localization/CLDR/Reader.php
```diff
@@ -592,6 +592,24 @@ protected function mapLocaleData(SimpleXMLElement $xmlLocaleData, $supplementalD
...
         }
 
+        // Territories
+        $localeDisplayNamesData = $xmlLocaleData->localeDisplayNames;
+        if (isset($localeDisplayNamesData->territories)) {
+            $territories = [];
+            foreach ($localeDisplayNamesData->territories->children() as $territoryNode) {
+                $territoryCode = (string) $territoryNode->attributes()->type;
+                $variant = (string) $territoryNode->attributes()->variant;
+                if (!empty($territories[$territoryCode]) && empty($variant)) {
```
PierreRambaud: You should add a comment explaining why $variant can't be empty :thinking:

src/Core/Localization/CLDR/TerritoryData.php
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
```
PierreRambaud, line 27:
```suggestion

declare(strict_types=1);

```

src/Core/Localization/CLDR/TerritoryData.php
```diff
@@ -0,0 +1,108 @@
...
+            $this->setName($territoryData->getName());
+        }
+
+        return $this;
+    }
+
+    /**
+     * @return string
+     */
+    public function getIsoCode()
```
PierreRambaud:
```suggestion
    public function getIsoCode(): string
```

src/Core/Localization/CLDR/TerritoryData.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        return $this->isoCode;
+    }
+
+    /**
+     * @param string $isoCode
+     *
+     * @return TerritoryData
+     */
+    public function setIsoCode($isoCode)
```
PierreRambaud:
```suggestion
    public function setIsoCode(string $isoCode): self
```

src/Core/Localization/CLDR/TerritoryData.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        $this->isoCode = $isoCode;
+
+        return $this;
+    }
+
+    /**
+     * @return string
+     */
+    public function getName()
```
PierreRambaud:
```suggestion
    public function getName(): string
```

src/Core/Localization/CLDR/TerritoryData.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        return $this->name;
+    }
+
+    /**
+     * @param string $name
+     *
+     * @return TerritoryData
+     */
+    public function setName($name)
```
PierreRambaud:
```suggestion
    public function setName(string $name): self
```

PierreRambaud (dismissed):

src/Core/Localization/CLDR/Reader.php
Progi1984: Fixed

PierreRambaud (dismissed):

classes/LocalizationPack.php
```diff
@@ -402,13 +402,14 @@ protected function _installLanguages($xml, $install_mode = false)
             foreach ($xml->languages->language as $data) {
                 /** @var SimpleXMLElement $data */
                 $attributes = $data->attributes();
+                $isoCode = (string) $attributes['iso_code'];
                 // if we are not in an installation context or if the pack is not available in the local directory
```
jolelievre:
```suggestion
                // if we are not in an installation context or if the pack is available in the local directory
```

The comment seems wrong compared to what checks are performed no?

src/Core/Localization/CLDR/Locale.php
```diff
@@ -310,11 +318,21 @@ public function getCurrencyData($currencyCode)
...
     {
         return $this->currencies;
     }
+
+    /**
+     * Get CLDR data of all territories.
+     *
+     * @return TerritoryData[] Data of all territories
+     */
+    public function getAllTerritories(): array
```
jolelievre, line 318: Shouldn't the interface be updated as well Also I think this can be considered a BC break since the core now needs CLDR Locale to expose `getAllTerritories`

By the way why `getAllTerritories` and not simply `getTerritories` ?

jolelievre (commented):
Small comments, and question Did you test both types of install (using CLI and browser) ?

classes/LocalizationPack.php
Progi1984: Approved.

src/Core/Localization/CLDR/Locale.php
Progi1984: You're right for adding the method in interface. :heavy_check_mark: But I'm not agree with changing the name method, it's more coherent relative to `getAllNumberSymbols` or `getAllCurrencies`.

Progi1984: @jolelievre I added the label "BC Break" and a note with the BC Break explained.

src/Core/Localization/CLDR/TerritoryData.php
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @return $this Fluent interface
+     */
+    public function overrideWith(TerritoryData $territoryData)
+    {
+        if (null !== $territoryData->getIsoCode()) {
+            $this->setIsoCode($territoryData->getIsoCode());
+        }
+
+        if (null !== $territoryData->getName()) {
```
sowbiba, line 64: it can happen to have in parameter a territoryData with empty name. In this case, the name of the current TerritoryData will be keep (kept ?), with the isoCode of the given TerritoryData

Progi1984: As you can see in `src/Core/Localization/CLDR/Reader.php`, I check that the name can't be empty. But the interface accepting null values, I must check that null values are not replacing existing values like isocode.

sowbiba: I think the method must check that the object given is complete to prevent weird behaviours

classes/Language.php
```diff
@@ -1415,32 +1415,30 @@ public static function updateMultilangTables(Language $language, array $tablesTo
...
+        $localeCLDR = $localeRepoCLDR->getLocale($lang->locale);
+        $territories = $localeCLDR->getAllTerritories();
+
+        foreach ($translatableCountries as $country) {
+            if (empty($territories[$country['iso_code']])) {
+                continue;
             }
+            $territoryData = $territories[$country['iso_code']];
+            $sql = 'UPDATE `' . _DB_PREFIX_ . 'country_lang`
+                    SET `name` = "' . pSQL(ucwords($territoryData->getName())) . '"
```
jolelievre: Careful with `ucword` I would advise using `mb_ucwords`

I had issues in https://github.com/PrestaShop/PrestaShop/blob/c7af233fb637eeb6ab732cd908d9c19ec4843333/src/Core/Currency/CurrencyGridDataFactory.php#L106 when using japanese because there were ideograms characters that did not like `ucfirst` very much and I ended up with little squares displayed I think it's the same with `ucwords`

src/Core/Localization/CLDR/TerritoryData.php
jolelievre: The different layers are supposed to handle this correctly If it doesn't want to eras a value with empty string then it doesn't set anything This way the field is null and doesn't override anything

However some layer might legitimately wants an empty value in which case it should be able to override previous layers

When I look at LocaleData it doesn't check empty strings and it works fine for now so I think we can rely on the same architecture

jolelievre (changes requested):
Careful with japanese or arabic characters (basically utf8 characters I guess)

classes/Language.php
Progi1984: Thanks. I fixed.

src/Core/Localization/CLDR/LocaleData.php
```diff
@@ -170,6 +177,10 @@ public function overrideWith(LocaleData $localeData)
             }
         }
 
+        if (null !== $localeData->getTerritories()) {
+            $this->territories = array_merge($this->territories, $localeData->getTerritories());
```
jolelievre:
```suggestion
            $this->setTerritories(array_merge($this->territories, $localeData->getTerritories()));
```

jolelievre: To be consistent with https://github.com/PrestaShop/PrestaShop/pull/21680

src/Core/Localization/CLDR/LocaleData.php
```diff
@@ -106,6 +106,13 @@ class LocaleData
      */
     protected $currencies;
 
+    /**
+     * All territories, by ISO code.
+     *
+     * @var TerritoryData[]
+     */
+    protected $territories = [];
```
jolelievre:
```suggestion
     * @var TerritoryData[]|null
     */
    protected $territories;
```

jolelievre: To be consistent with other fields and https://github.com/PrestaShop/PrestaShop/pull/21680

src/Core/Localization/CLDR/TerritoryData.php
```diff
@@ -0,0 +1,110 @@
...
+ * The TerritoryData class is the exact representation of Territory's data structure inside CLDR xml data files.
+ */
+class TerritoryData
+{
+    /**
+     * Alphabetic ISO 4217 territory code.
+     *
+     * @var string
+     */
+    protected $isoCode;
```
jolelievre, line 42: All these fields are nullable since in the override function you check if they are null
So it needs to be refelcted in all the PHPDoc and in the return types `?string`

tests/Unit/Core/Localization/CLDR/LocaleTest.php
```diff
@@ -0,0 +1,63 @@
...
+            [
+                (new LocaleData())->setTerritories(
+                    [
+                        (new TerritoryData())->setIsoCode('DK')->setName('Denmark'),
+                        (new TerritoryData())->setIsoCode('EC')->setName('Ecuador'),
+                        (new TerritoryData())->setIsoCode('FR')->setName('France'),
+                    ]
+                ),
+            ],
+        ];
```
jolelievre:
```suggestion
        yield [
            new LocaleData(),
        ];

        yield [
            (new LocaleData())->setTerritories(
                [
                    (new TerritoryData())->setIsoCode('DK')->setName('Denmark'),
                    (new TerritoryData())->setIsoCode('EC')->setName('Ecuador'),
                    (new TerritoryData())->setIsoCode('FR')->setName('France'),
                ]
            ),
        ];
```

jolelievre: Just a suggestion but I recently tested the `yield` in data providers, I find this notation easier to read as it clearly separates each individual tests
Besides `yield` is kinda cool 😛

jolelievre: (this could be done in you other tests dataProvider

tests/Unit/Core/Localization/CLDR/TerritoryDataTest.php
```diff
@@ -0,0 +1,131 @@
...
+
+        // Result
+        $result = $this->territoryData->overrideWith($territoryData);
+        $this->assertInstanceOf(TerritoryData::class, $result);
+        $this->assertEquals($expected, $result->getIsoCode());
+    }
+
+    public function providerIsoCode(): array
+    {
+        return [
```
jolelievre: You could add use cases with `null` values to assert nothing is modified
And one with empty value to assert it DOES modify

tests/Unit/Core/Localization/CLDR/TerritoryDataTest.php
```diff
@@ -0,0 +1,131 @@
...
+            [
+                'France',
+                'France',
+                'France',
+            ],
+            [
+                'France',
+                'Denmark',
+                'Denmark',
```
jolelievre: Same here you could test `null` and `''`

jolelievre (commented):
Some are just suggestions, but some are really necessary improvement

tests/Unit/Core/Localization/CLDR/LocaleTest.php
Progi1984: Done.

src/Core/Localization/CLDR/LocaleData.php
Progi1984: Fixed.

src/Core/Localization/CLDR/LocaleData.php
Progi1984: Fixed.

src/Core/Localization/CLDR/TerritoryData.php
Progi1984: Fixed

tests/Unit/Core/Localization/CLDR/TerritoryDataTest.php
Progi1984: Fixed

tests/Unit/Core/Localization/CLDR/TerritoryDataTest.php
Progi1984: Fixed

src/Core/Localization/CLDR/Locale.php
```diff
@@ -118,6 +125,7 @@ public function __construct(LocaleData $localeData)
         $this->percentPatterns = $localeData->getPercentPatterns();
         $this->currencyPatterns = $localeData->getCurrencyPatterns();
         $this->currencies = $localeData->getCurrencies();
+        $this->territories = $localeData->getTerritories() ?: [];
```
jolelievre, line 128: Why does this one need a special case? (compared to all other data) Can't it remain `null` ?

classes/Language.php
```diff
@@ -1436,32 +1436,30 @@ public static function updateMultilangTables(Language $language, array $tablesTo
...
+        $localeCLDR = $localeRepoCLDR->getLocale($lang->locale);
+        $territories = $localeCLDR->getAllTerritories();
+
+        foreach ($translatableCountries as $country) {
+            if (empty($territories[$country['iso_code']])) {
+                continue;
             }
+            $territoryData = $territories[$country['iso_code']];
+            $sql = 'UPDATE `' . _DB_PREFIX_ . 'country_lang`
+                    SET `name` = "' . pSQL(mb_ucwords($territoryData->getName())) . '"
```
jolelievre, line 1459: Maybe the `mb_ucwords` shoud be done in the `Reader` instead of here?

Progi1984:
Closed in favor of use of Intl => #19818

PR #20383, opened 2020-07-29
Clean a bit PaymentModule class

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Clean a bit PaymentModule class. I fix bad english, fix some phpDoc and remove a deprecated constant. And add a return early
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20383)

### DIFF
--- a/classes/PaymentModule.php
+++ b/classes/PaymentModule.php
@@ -320,7 +320,7 @@ abstract class PaymentModuleCore extends Module
             // We don't use the following condition to avoid the float precision issues : http://www.php.net/manual/en/language.types.float.php
             // if ($order->total_paid != $order->total_paid_real)
             // We use number_format in order to compare two string
-            if ($order_status->logable && number_format($cart_total_paid, _PS_PRICE_COMPUTE_PRECISION_) != number_format($amount_paid, _PS_PRICE_COMPUTE_PRECISION_)) {
+            if ($order_status->logable && number_format($cart_total_paid, Context::getContext()->getComputingPrecision()) != number_format($amount_paid, _PS_PRICE_COMPUTE_PRECISION_)) {
                 $id_order_state = Configuration::get('PS_OS_ERROR');
             }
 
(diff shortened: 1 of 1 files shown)


Progi1984 (dismissed):

classes/PaymentModule.php
```diff
@@ -817,9 +819,8 @@ public function getCurrency($current_id_currency = null)
         if (!isset($id_currency) || empty($id_currency)) {
             return false;
         }
-        $currency = new Currency((int) $id_currency);
 
-        return $currency;
+        return Currency((int) $id_currency);
```
Matt75:
```suggestion
        return Currency::getCurrencyInstance((int) $id_currency);
```

Matt75:
@matks Can you remove all `die(Tools::displayError());` in this class ?
Because `Tools::displayError()` throw a PrestaShopException only when `_PS_MODE_DEV_` is to true so in production it die...
Currently PaymentModule developer cannot catch it to fallback on degraded or error response properly... So it can be a great improvement 👍

matks:
> @matks Can you remove all `die(Tools::displayError());` in this class ?
> Because `Tools::displayError()` throw a PrestaShopException only when `_PS_MODE_DEV_` is to true so in production it die...
> Currently PaymentModule developer cannot catch it to fallback on degraded or error response properly... So it can be a great improvement 👍

I can, but then in another PR.
Because the current PR brings no changes to the behavior (so it can be tested without QA). Removing `die(Tools::displayError());` cannot be done without thorough testing from both me and the QA team

Matt75:
Ok, i can open an issue for that if needed

PierreRambaud (approved):

Progi1984 (approved):

Progi1984:
Thanks @matks . No need QA

PR #19404, opened 2020-05-27
Put back prefix for filter name inside CustomerQueryBuilder

juliendombret:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | The filter name must to have a table prefix<br/>in my case, i add a filter phone on customer grid. the lastname field exist on customer and address table.<br/>I have a sql error
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Customer listing page should work as before

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19404)

PR #20389, opened 2020-07-29
Fix integrity constraint violation on admin customers list when joins address table

### LINKED ISSUE #20390: Integrity constraint violation on admin customers list when joins address table
#### Describe the bug

When a module joins the `ps_address` table to the customer list retrieval request, when you try to search for a customer by his first or last name, you get the error below because the `lastname`, `firstname` and `company` fields are common to both the `ps_customer` and `ps_address` tables

`email` can also cause an error if a table with an `email` field is join to the `ps_customer` table, so I also added the alias for this field.

88841129-c7e6f100-d1dd-11ea-91b1-b377028fbbba

#### Steps to Reproduce

Module to test: [rockettest.zip](https://github.com/PrestaShop/PrestaShop/files/4999980/rockettest.zip)

- install the provided module.
- in the admin, go to Customers > Customers
- try to search for a customer by first or last name (company should also cause an error)
- you should have an error 500
- add the fix
- refresh the customers admin, it should work again

Fix: #20389 

#### Additional information

* PrestaShop version: 1.7.6
* PHP version: 7.2

DelecroixQuentin:
| Questions     | Answers
| Branch?       | develop
| Description?  | Description below
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | No ticket
| How to test?  | See below

## Description
When a module joins the `ps_address` table to the customer list retrieval request, when you try to search for a customer by his first or last name, you get the error below because the `lastname`, `firstname` and `company` fields are common to both the `ps_customer` and `ps_address` tables

`email` can also cause an error if a table with an `email` field is join to the `ps_customer` table, so I also added the alias for this field.

Capture d’écran 2020-07-29 à 17 43 57

## How to test
Module to test: [rockettest.zip](https://github.com/PrestaShop/PrestaShop/files/4996883/rockettest.zip)
- install the provided module.
- in the admin, go to Customers > Customers
- try to search for a customer by first or last name (company should also cause an error)
- you should have an error 500
- add the fix
- refresh the customers admin, it should work again

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20389)

### DIFF
--- a/src/Core/Grid/Query/CustomerQueryBuilder.php
+++ b/src/Core/Grid/Query/CustomerQueryBuilder.php
@@ -222,7 +222,9 @@ final class CustomerQueryBuilder extends AbstractDoctrineQueryBuilder
                 continue;
             }
 
-            $qb->andWhere('c.`' . $filterName . '` LIKE :' . $filterName);
+            $customerTableAlias = in_array($filterName, ['firstname', 'lastname', 'company', 'email']);
+
+            $qb->andWhere(($customerTableAlias ? 'c.' : '') . '`' . $filterName . '` LIKE :' . $filterName);
             $qb->setParameter($filterName, '%' . $filterValue . '%');
         }
     }


matks (dismissed):
Nice idea and nice PR 👍 

I approve this PR ✅  .

Moreover ... I'm wondering: adding aliases to ALL filters for ALL QueryBuilders would bring only benefits, right ? 🤔  maybe this is something we could do everywhere ?

src/Core/Grid/Query/CustomerQueryBuilder.php
```diff
@@ -222,7 +222,12 @@ private function applyFilters(array $filters, QueryBuilder $qb)
                 continue;
             }
 
-            $qb->andWhere('`' . $filterName . '` LIKE :' . $filterName);
+            $customerTableAlias = false;
+            if (in_array($filterName, ['firstname', 'lastname', 'company', 'email'])) {
+                $customerTableAlias = true;
+            }
```
Progi1984:
```suggestion
            $customerTableAlias = in_array($filterName, ['firstname', 'lastname', 'company', 'email']);
```

DelecroixQuentin: Done

DelecroixQuentin:
Thanks @matks

In some QueryBuilder, alias is already use like in `TaxQueryBuilder` or `TaxRulesGroupQueryBuilder`
If you take a look in `SupplierQueryBuilder`, alias is handle differently. 

In most cases, this is handled and here I think it's mainly a concern because both tables have 3 fields in common (other than `date_add|upd` , `active` etc... which are usually filtered with the alias).

Progi1984 (dismissed):

matks (dismissed):

PierreRambaud (dismissed):

SD1982:
Hi @DelecroixQuentin Everything looks good to me excepted a little point :

With your PR after adding the provided module i have a duplicated customer in the customers list
duplicate customer

It's ok without the PR so i guess the problem is not from the module

@PrestaShop/prestashop-core-developers wdyt ?

DelecroixQuentin:
Hi @SD1982,

It is indeed the module that generates the duplication of customers, the purpose of the module was the possibility to search for customers via their phone number (to facilitate the after-sales service). So in the module, I modify a little bit the query of the customers list to add `left join` on the table `ps_address` (since the phone numbers are in this table). So, for each customer address, there will be a customer line in the list.

SD1982:
Hi @DelecroixQuentin 
Thank you for the information i will make a second check on the PR just to be sure and if i found nothing eles it will be Ok for me :+1:

SD1982:
Hi @DelecroixQuentin can you check to resolve the conficting files ?

DelecroixQuentin:
Hi @SD1982,

I solved the conflicts but some github actions failed without any information and idk why, if someone can check this out it would be nice

SD1982:
Hi @PrestaShop/prestashop-core-developers can you check this please ?

> I solved the conflicts but some github actions failed without any information and idk why, if someone can check this out it would be nice

Progi1984:
@SD1982 I fixed Github Actions ;)

SD1982:
LGTM thanks @DelecroixQuentin !!

Progi1984:
@DelecroixQuentin Congratulations :tada: Your PR can be merged. But in your commits, you have a merge commit. Could you rebase your PR (and squash this merge commit) ? Thanks

DelecroixQuentin:
Hi @Progi1984,

Is it ok for you ?

Progi1984 (approved):

matthieu-rolland (approved):

src/Core/Grid/Query/CustomerQueryBuilder.php
```diff
@@ -222,7 +222,9 @@ private function applyFilters(array $filters, QueryBuilder $qb)
                 continue;
             }
 
-            $qb->andWhere('c.`' . $filterName . '` LIKE :' . $filterName);
+            $customerTableAlias = in_array($filterName, ['firstname', 'lastname', 'company', 'email']);
+
+            $qb->andWhere(($customerTableAlias ? 'c.' : '') . '`' . $filterName . '` LIKE :' . $filterName);
```
jolelievre, line 227: Mmmm, I know I arrive probably late in this conversation But I don't understand why we remove the alias from the core query so that it works correctly with the module override?
SHouldn't it be the module which needs to adapt? Besides removing an alias can only lead to even more ambiguity, how can it lessen it?

jolelievre (changes requested):
Sorry, but after reflexion I really don't understand why we should remove the alias from the core table I prefer to block this one until I understand the valid reason that I may have missed

But the general rule is simple the core should not be adapted FOR a module, the module adapts to the core In cases where the Core is too strict of course we can change it to improve its extensability But in this case I don't see why we need to remove the alias `c` which is correctly associated to the customer table

If the address has identical fields then you need to have an alias for the address table and maybe change the way the `where` conditions are created so that it favors the address one

PierreRambaud (changes requested):
I didn't not manage to reproduce the issue. There is nothing related to firstname / lastname in your module.
Then, I agree with @jolelievre, removing the alias is not a good idea.
My bad, I validated it too fast

DelecroixQuentin:
Hi @jolelievre, @PierreRambaud,

The use of the alias was added about two months ago, after the PR was validated.

Cf: https://github.com/PrestaShop/PrestaShop/blame/develop/src/Core/Grid/Query/CustomerQueryBuilder.php#L225

The purpose of PR is not to adapt the Core to the module but to fix the Core which, before, did not use aliases when filtering the customer list and which could cause an integrity error if a module added columns in the customer list based on a table containing some common columns with the `customer` table (ie `address` table). 

With the introduction of the alias, this PR has no more value indeed.

See #19404

jolelievre:
Oh my bad @DelecroixQuentin Back in the day your request was indeed relevant 😉
But this is an interesting issue, we should check that all the Grid builder correctly use an alias!

PR #20398, opened 2020-07-30
Fixed Cache Problem in Language & Currency

Progi1984:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fixed Cache Problem<br>Detected problem in #20379 and Backported after @jolelievre ask it ;) 
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | N/A
| How to test?  | N/A

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20398)

### DIFF
--- a/classes/Currency.php
+++ b/classes/Currency.php
@@ -863,13 +863,17 @@ class CurrencyCore extends ObjectModel
             $query = Currency::getIdByQuery($idShop, $includeDeleted);
             $query->where('iso_code = \'' . pSQL($isoCode) . '\'');
 
-            $result = (int) Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue($query->build());
+            $result = Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue($query->build());
+            if (empty($result)) {
+                return 0;
+            }
+
             Cache::store($cacheId, $result);
 
-            return $result;
+            return (int) $result;
         }
 
-        return Cache::retrieve($cacheId);
+        return (int) Cache::retrieve($cacheId);
     }
 
     /**
--- a/tests-legacy/Integration/classes/LanguageTest.php
+++ b/tests-legacy/Integration/classes/LanguageTest.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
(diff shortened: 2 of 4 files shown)


matks (dismissed):

jolelievre (dismissed):

matthieu-rolland (dismissed):

tests-legacy/Integration/classes/LanguageTest.php
```diff
@@ -0,0 +1,66 @@
...
+        parent::setUp();
+
+        Database::restoreTestDB();
+    }
+
+    public function testGetIdByIsoWithCache()
+    {
+        $this->assertNull(Language::getIdByIso('zz'));
+        $this->addDummyLanguage();
+        $this->assertIsString(Language::getIdByIso('zz'));
```
jolelievre: It returns a string? Not an `int` ?
Besides even if it's not strictly typed, it would be better to add a test to ensure it's not 0 like in currency test

Progi1984: Fixed ;)

classes/Currency.php
```diff
@@ -863,8 +863,12 @@ public static function getIdByIsoCode($isoCode, $idShop = 0, $forceRefreshCache
...
             $query->where('iso_code = \'' . pSQL($isoCode) . '\'');
 
-            $result = (int) Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue($query->build());
-            Cache::store($cacheId, $result);
+            $result = Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue($query->build());
+            if (empty($result)) {
+                return 0;
+            }
+
+            Cache::store($cacheId, (int) $result);
```
jolelievre: Better cast the return instead of the cached value (or both) Or else the first call will return a string and the next ones an int

Progi1984: @jolelievre Fixed

jolelievre (approved):

jolelievre (changes requested):
Oups tests are failing because of returned type, they should be adapted

jolelievre (approved):
Thanks @Progi1984

matks (approved):

Progi1984:
Thanks @jolelievre & @matks

PR #20403, opened 2020-07-31
Remove useless assignment of logout_url in MyAccountController

Sinepel:
| Questions     | Answers
| Branch?       | develop 
| Description?  | Remove useless assignment of `logout_url` in MyAccountController to use `$urls.actions.logout `
| Type?         | improvement
| Category?     | FO 
| BC breaks?    | yes 
| Deprecations? | yes
| Fixed ticket? | Fixes 
| How to test?  | Go to My Account page and test to logout

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20403)

### DIFF
--- a/controllers/front/MyAccountController.php
+++ b/controllers/front/MyAccountController.php
@@ -37,6 +37,13 @@ class MyAccountControllerCore extends FrontController
      */
     public function initContent()
     {
+        /*
+        * @deprecated since 1.7.8
+        */
+        $this->context->smarty->assign([
+            'logout_url' => $this->context->link->getPageLink('index', true, null, 'mylogout'),
+        ]);
+
         parent::initContent();
         $this->setTemplate('customer/my-account');
     }
--- a/controllers/front/MyAccountController.php
+++ b/controllers/front/MyAccountController.php
@@ -37,10 +37,6 @@ class MyAccountControllerCore extends FrontController
      */
     public function initContent()
     {
-        $this->context->smarty->assign([
-            'logout_url' => $this->context->link->getPageLink('index', true, null, 'mylogout'),
-        ]);
-
         parent::initContent();
         $this->setTemplate('customer/my-account');
     }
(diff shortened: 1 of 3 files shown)


PierreRambaud (commented):
As far as I know, it's also used in ps_customersignin.tpl (and in some modules). We can keep the logout_rul assignment, mark it as deprecated, and change values in templates, wdyt? :)

Sinepel:
@PierreRambaud 
Hey!
Can you tell me the benefit of marking it down as depreciated if we can replace it? It's not a criticism, it's a real question, it's for my personal culture.

matks:
> @PierreRambaud
> Hey!
> Can you tell me the benefit of marking it down as depreciated if we can replace it? It's not a criticism, it's a real question, it's for my personal culture.

Marking something as deprecated is an indication and a preparation for future changes.
- maintainers (us) know that when we work on next major PS version (= BC breaks accepted) we need to remove this
- users (module developers) of this library know that this piece of code is working _now_ but will be removed in the future so they should start using the replacement

So when you see a piece of code marked as deprecated, the message behind is "Hello ! So you are using this piece of code. We want you to know that we have planned to remove it in next major version (because we have built something better/faster or something that we believe is more useful/appropriate). You must use the time remaining before next version to stop using this piece of code and perform the necessary changes into your module/theme/whatever to use the replacement which will stay here".

You can read more about it on Wikipedia https://en.wikipedia.org/wiki/Deprecation

Sinepel:
Can you tell me if it's okay for you like this? 
I'm not used to depreciating variable assignments, usually it's a full function ^^

NeOMakinG (approved):
@matks I think this is fine ? We should take care of this PR as @PierreRambaud can't handle it in next days :)

Progi1984 (approved):

PierreRambaud (approved):

PierreRambaud:
Thank you @Sinepel

PR #20408, opened 2020-07-31
Fix bug on Invoices tax detail 

### LINKED ISSUE #20393: Wrong Tax Detail in PDF invoice when there are multiple invoices
### Describe the bug

When an order has more than one Invoice, the invoice's "tax detail" is wrong: it takes all the order's products into account.

### Expected behavior
The invoice's "tax detail" should only take into account the products related to this invoice, not all the products in the order.

### Steps to Reproduce

1 - In FO, add a product to a cart and complete the order.
2 - In BO, go to the order you just made, and click on "generate invoice"
3 - Still in BO Order page, add a product **with a new invoice**

See the two invoices, the Tax Detail takes into account the total of the order's products and not only products in the invoice 

### Screenshots
**The order contains 2 products, on two distinct invoices:**
Capture d’écran 2020-07-30 à 11 34 18

**The first invoice, see tax details, it takes all the products into account instead of just this invoice's product:**
Capture d’écran 2020-07-30 à 11 34 34

**The second invoice, same problem:**
Capture d’écran 2020-07-30 à 11 34 46

### Additional information

* PrestaShop version: 1.7.7.x / 1.7.6.x

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | The Tax Detail frame inside the invoice PDF should only contain tax details for the current invoice, not tax details of all products of all invoices in this order.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20393
| How to test?  | See issue #20393 for steps and screen captures

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20408)

### DIFF
--- a/classes/order/OrderInvoice.php
+++ b/classes/order/OrderInvoice.php
@@ -359,6 +359,9 @@ class OrderInvoiceCore extends ObjectModel
         if ($sum_composite_taxes) {
             $grouped_details = [];
             foreach ($details as $row) {
+                if ($this->id !== (int) $row['id_order_invoice']) {
+                    continue;
+                }
                 if (!isset($grouped_details[$row['id_order_detail']])) {
                     $grouped_details[$row['id_order_detail']] = [
                         'tax_rate' => 0,
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -2586,6 +2586,7 @@ class OrderCore extends ObjectModel
                     'total_tax_base' => $total_tax_base,
                     'unit_amount' => $unit_amount,
                     'total_amount' => $total_amount,
+                    'id_order_invoice' => $order_detail['id_order_invoice'],
                 ];
             }
         }


classes/order/OrderInvoice.php
```diff
@@ -359,6 +359,9 @@ public function getProductTaxesBreakdown($order = null)
         if ($sum_composite_taxes) {
             $grouped_details = [];
             foreach ($details as $row) {
+                if ($this->id !== (int) $row['id_order_invoice']) {
+                    continue;
+                }
```
matthieu-rolland, line 364: I wonder if this is a breaking change or not 🤔 

With this PR I'm changing the behavior of the method (which was wrong), so if let's say a module was using this method expecting the previous wrong behavior, it could be perceived as a breaking change.

jolelievre: I think the fix is correct, this method didn't take into account the possibility of multi invoices per Order So this fixes the (incorrect) behaviour for multi invoices, but it doesn't change it for single invoice case
Hence it shouldn't break any behaviour

jolelievre (approved):

Progi1984 (approved):

SD1982:

Progi1984:
Thanks @matthieu-rolland & @SD1982

PR #20413, opened 2020-07-31
Use SymfonyContainer::getInstance() instead of ContainerBuilder::getContainer('admin')

### LINKED ISSUE #14995: Two service containers coexist in back office - there should be only one
In back office, when requesting a page that has been migrated in symfony, when asking for a container from Context, we get an instance of **AdminContainer**, a service container generated by symfony, whose name is probably based on **config/services/admin**.

So it seems like we have the **SymfonyContainer** and the so called **AdminContainer** living in parallel when we are in a migrated page in back office.

atomiix:
| Questions     | Answers
| Branch?       | develop
| Description?  | In the BO, there were 2 containers. 1 `SymfonyContainer` and 1 `AdminContainer` living in parallel when we are in a migrated page in back office. This PR aims to delete the use of `AdminContainer` as `SymfonyContainer` is available everywhere in the BO. More details on #14995
| Type?         | improvement
| Category?     | BO
| BC breaks?    | yes - An exception is now thrown if you try to use `ContainerBuilder::getContainer()` with `$containerName = 'admin'`
| Deprecations? | no
| Fixed ticket? | Fixes #14995
| How to test?  | It's kinda hard to say how to test but the BO should be working exactly like before

As @jolelievre mentioned [here](https://github.com/PrestaShop/PrestaShop/issues/14995#issuecomment-518309252), some services available in `AdminContainer` were not available in `SymfonyContainer`. Here are the list : 
- context.static
- finder
- hashing
- hook_configurator
- hook_provider
- hook_repository
- theme_manager
- theme_validator

Most of them have been added by an alias or a service declaration except for `theme_validator` as it was already unavailable since [this commit](https://github.com/PrestaShop/PrestaShop/commit/c31300e9cf32de246eb6ed43626d69ed57fc9cc6?branch=c31300e9cf32de246eb6ed43626d69ed57fc9cc6&diff=unified#diff-cda73c4a10a5a7c3778451b0401f5d9dR43) as the service declaration hasn't been updated.

:notebook: **BC Break**
- An exception is now thrown if you try to use ContainerBuilder::getContainer() with $containerName = 'admin'

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20413)

### DIFF
--- a/src/Adapter/ContainerBuilder.php
+++ b/src/Adapter/ContainerBuilder.php
@@ -37,6 +37,7 @@ use PrestaShop\PrestaShop\Adapter\Container\LegacyContainer;
 use PrestaShop\PrestaShop\Adapter\Container\LegacyContainerBuilder;
 use PrestaShop\PrestaShop\Core\EnvironmentInterface;
 use PrestaShopBundle\DependencyInjection\Compiler\LoadServicesFromModulesPass;
+use PrestaShopBundle\Exception\ServiceContainerException;
 use Symfony\Component\Config\ConfigCache;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Config\Resource\FileResource;
@@ -90,6 +91,11 @@ class ContainerBuilder
      */
     public static function getContainer($containerName, $isDebug)
     {
+        if ($containerName === 'admin') {
+            throw new ServiceContainerException(
+                'You should use `SymfonyContainer::getInstance()` instead of `ContainerBuilder::getContainer(\'admin\')`'
+            );
+        }
         if (!isset(self::$containers[$containerName])) {
             $builder = new ContainerBuilder(new Environment($isDebug));
             self::$containers[$containerName] = $builder->buildContainer($containerName);
--- a/src/PrestaShopBundle/Resources/config/services/alias.yml
+++ b/src/PrestaShopBundle/Resources/config/services/alias.yml
@@ -11,3 +11,12 @@ services:
 
     prestashop.twig.extension.dataFormatter:
         alias: prestashop.twig.extension.data_formatter
+
+    hashing:
+      alias: prestashop.core.crypto.hashing
+
+    hook_provider:
+      alias: prestashop.adapter.legacy.hook
+
+    theme_manager:
+      alias: prestashop.core.addon.theme.theme_manager
--- a/src/PrestaShopBundle/Resources/config/services/alias.yml
+++ b/src/PrestaShopBundle/Resources/config/services/alias.yml
@@ -20,3 +20,6 @@ services:
 
     theme_manager:
       alias: prestashop.core.addon.theme.theme_manager
+
+    theme_validator:
+      alias: prestashop.core.addon.theme.theme_validator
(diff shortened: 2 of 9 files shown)


src/Adapter/ContainerBuilder.php
```diff
@@ -90,6 +91,11 @@ class ContainerBuilder
      */
     public static function getContainer($containerName, $isDebug)
     {
+        if ($containerName === 'admin') {
+            throw new ServiceContainerException(
```
PierreRambaud, line 95: It should be considered as a BC since the method now throw an exception

src/PrestaShopBundle/Resources/config/services/alias.yml
```diff
@@ -11,3 +11,12 @@ services:
 
     prestashop.twig.extension.dataFormatter:
         alias: prestashop.twig.extension.data_formatter
+
+    hashing:
```
PierreRambaud, line 15: Could you add comments about aliases?
Why are they needed?

src/Adapter/ContainerBuilder.php
atomiix: @PierreRambaud What do you think it's best then, remove the exception or set it as a BC Break?

src/PrestaShopBundle/Resources/config/services/alias.yml
atomiix: @PierreRambaud there's already [a comment](https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Resources/config/services/alias.yml#L5) above saying it's for backward compatibility purpose.

PierreRambaud: Ohhh, thanks

src/Adapter/ContainerBuilder.php
PierreRambaud: Set it as BC break :sweat_smile:

atomiix: @PierreRambaud I updated the PR description ;)

PierreRambaud (dismissed):

tests/Integration/Adapter/ContainerBuilderTest.php
```diff
@@ -52,21 +51,6 @@ public function testFrontContainerContainsAnEntityManager()
         $this->assertInstanceOf(EntityManagerInterface::class, $entityManager);
     }
 
-    public function testGetAdminContainer()
```
jolelievre, line 55: You could add a test here asserting that the `ServiceContainerException` is thrown when trying to access the AdminContainer

jolelievre (dismissed):
Small comment for improvement in tests

jolelievre:
About the `theme_validator` invalid declaration maybe it's the occasion to fix it?

tests/Integration/Adapter/ContainerBuilderTest.php
atomiix: Done!

PierreRambaud (approved):

jolelievre (approved):

Progi1984:
@atomiix Could you add a note in the PR for the BC Break for release notes, please ?

atomiix:
@Progi1984 I added a note in the PR description in the section `BC Breaks?`. Isn't it enough?

Progi1984:
Ok ! sorry ;)

Progi1984:
Thanks @atomiix & @ngodefroy

PR #20430, opened 2020-08-03
Allows profile to override the avatar of employee

### LINKED ISSUE #18653: Allow users to better handle avatars

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | Allows profile to override the avatar of employee
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #18653
| How to test?  | 0. Build assets<br>1. Install PrestaShop<br>2. Add an avatar to group Translator<br>3. Add an employee "A" with an avatar<br>4. Add an employee "B" with an avatar and add it to Translator<br>5. Add an employee "C" without an avatar and add it to Translator<br>6. Add an employee "D" without an avatar<br>7. Add an employee "E" without an avatar with Gravatar<br>8. Add an employee "F" without an avatar with Gravatar and add it to Translator<br>9. Add an employee "G" with an avatar with Gravatar and add it to Translator<br>10. Connect as A<br>11. :heavy_check_mark: The avatar of the employee A is displayed<br>12. Connect as B<br>13. :heavy_check_mark: The avatar of the group Translator is displayed<br>14. Connect as C<br>15. :heavy_check_mark: The avatar of the group Translator is displayed<br>16. Connect as D<br>17. :heavy_check_mark: The avatar "White man on black background" is displayed<br>18. Connect as E<br>19. :heavy_check_mark: The gravatar is displayed<br>20. Connect as F<br>21. :heavy_check_mark: The avatar of the group Translator is displayed <br>22. Connect as superadmin<br>23. Install module [`ps_qualityassurance`](https://github.com/PrestaShop/ps_qualityassurance)<br>23 bis. Give permissions "View" on module `ps_qualityassurance` for the group "Translator"<br>24. Register a new hook with `actionOverrideEmployeeImage`<br>25. Use this code `$params['imageUrl'] = 'https://avatars.dicebear.com/api/male/' .urlencode($params['employee']->email).'.svg';`<br>26. Connect as A, B, C, D, E, F, G<br>27. And the hook will override the avatar or gravatar : <br>image

:notebook: **BC Break**
* The final class `PrestaShop\PrestaShop\Core\Form\IdentifiableObject\DataHandler\ProfileFormDataHandler` has a new second parameter : a parameter of type `PrestaShop\PrestaShop\Core\Image\Uploader\ImageUploaderInterface`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20430)


matks (dismissed):

classes/Employee.php
```diff
@@ -599,8 +599,23 @@ public function isSuperAdmin()
...
     public function getImage()
     {
-        $default = Tools::getAdminImageUrl('prestashop-avatar.png');
-        $imageUrl = '';
+        $defaultSystem = Tools::getAdminImageUrl('pr/default.jpg');
+        $imageUrl = null;
+
+        // Default from Profile
+        $profile = new Profile($this->id_profile);
+        $profilePath = $profile ? ($profile->image_dir . $profile->id . '.jpg') : '';
```
jolelievre: The test with just `$profile ?` seems useless, it will always be true no? Maybe a more relevant one would be `$profile->id == $this->id_profile` to ensure the Object does exist in DB and matches the employee settings

classes/Employee.php
```diff
@@ -599,8 +599,23 @@ public function isSuperAdmin()
...
+        $defaultProfile = file_exists($profilePath)
```
jolelievre: Maybe all this code to retrieve the profile picture could be moved in the Profile class in a `getProfileImage` method

classes/Employee.php
```diff
@@ -599,8 +599,23 @@ public function isSuperAdmin()
...
+            : null;
+        $imageUrl = $imageUrl ?? $defaultProfile;
+
+        // Gravatar
+        if ($this->has_enabled_gravatar) {
+            $imageUrl = $imageUrl ?? 'https://www.gravatar.com/avatar/' . md5(strtolower(trim($this->email))) . '?d=' . urlencode($defaultSystem);
+        }
 
         // Local Image
         $imagePath = $this->image_dir . $this->id . '.jpg';
```
jolelievre, line 616: This means the local image always has priority even though `has_enabled_gravatar` is true? So you need to enable this field AND remove the local image for it to work?

src/Adapter/Profile/QueryHandler/GetProfileForEditingHandler.php
```diff
@@ -68,4 +88,25 @@ private function getProfile(ProfileId $profileId)
...
+        $imagePath = _PS_PROFILE_IMG_DIR_ . $imageId . '.jpg';
+        $imageTag = $this->getTmpImageTag($imagePath, $imageId, 'profile');
+        $imageSize = $this->getImageSize($imagePath);
+
+        if (empty($imageTag) || null === $imageSize) {
+            return null;
+        }
+
+        return [
+            'size' => sprintf('%skB', $imageSize),
```
jolelievre, line 119: So this `size` field is never used? Why not return the url directly then?

src/Core/Form/IdentifiableObject/DataHandler/ProfileFormDataHandler.php
```diff
@@ -41,12 +44,21 @@ final class ProfileFormDataHandler implements FormDataHandlerInterface
...
+
     /**
      * @param CommandBusInterface $bus
+     * @param ImageUploaderInterface|null $imageUploader
      */
-    public function __construct(CommandBusInterface $bus)
-    {
+    public function __construct(
+        CommandBusInterface $bus,
+        ImageUploaderInterface $imageUploader = null
```
jolelievre: Since this class is `final` only used for internal use, and the service is built via DI I would go for the BC here 😜
And force the second parameter

src/PrestaShopBundle/Resources/views/Admin/Configure/AdvancedParameters/Profiles/create.html.twig
```diff
@@ -28,7 +28,9 @@
 {% block content %}
   <div class="row">
     <div class="col">
-      {% include '@PrestaShop/Admin/Configure/AdvancedParameters/Profiles/Blocks/form.html.twig' %}
+      {% include '@PrestaShop/Admin/Configure/AdvancedParameters/Profiles/Blocks/form.html.twig' with {
```
jolelievre: Maybe you can get rid of this one if you use `{% if avatarUrl|default('') is not empty %}` in the form twig?

src/PrestaShopBundle/Resources/views/Admin/Configure/AdvancedParameters/Profiles/edit.html.twig
```diff
@@ -28,7 +28,9 @@
 {% block content %}
   <div class="row">
     <div class="col">
-      {% include '@PrestaShop/Admin/Configure/AdvancedParameters/Profiles/Blocks/form.html.twig' %}
+      {% include '@PrestaShop/Admin/Configure/AdvancedParameters/Profiles/Blocks/form.html.twig' with {
```
jolelievre, line 31: This one makes more sense!

jolelievre (commented):
Mostly improvement comments ^^
Seems like a great feature otherwise!

classes/Profile.php
```diff
@@ -52,6 +52,16 @@ class ProfileCore extends ObjectModel
...
     protected static $_cache_accesses = [];
 
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct($id = null, $idLang = null, $idShop = null, $translator = null)
+    {
+        parent::__construct($id, $idLang, $idShop, $translator);
+
+        $this->image_dir = _PS_PROFILE_IMG_DIR_;
```
PierreRambaud, line 62: Why adding this property? :thinking:

classes/Employee.php
Progi1984: Fixed. Thanks

classes/Employee.php
Progi1984: Fixed.

classes/Employee.php
Progi1984: Fixed. Hooks > Profile > Gravatar > Local Image > Default from System

src/Adapter/Profile/QueryHandler/GetProfileForEditingHandler.php
Progi1984: It's for keeping the same array like in `src/Adapter/Profile/Employee/QueryHandler/GetEmployeeForEditingHandler.php`.

src/Core/Form/IdentifiableObject/DataHandler/ProfileFormDataHandler.php
Progi1984: Fixed. Added a note on BC Break. Added label "BC Break".

src/PrestaShopBundle/Resources/views/Admin/Configure/AdvancedParameters/Profiles/create.html.twig
Progi1984: Good idea. Fixed

src/PrestaShopBundle/Resources/views/Admin/Configure/AdvancedParameters/Profiles/edit.html.twig
Progi1984: :heart: on you

classes/Profile.php
Progi1984: For keeping the same coherence on each legacy class which has some images : 
image

PierreRambaud: I love legacy parts...

src/Adapter/Profile/QueryHandler/GetProfileForEditingHandler.php
```diff
@@ -68,4 +88,25 @@ private function getProfile(ProfileId $profileId)
...
     }
+
+    /**
+     * @param int $imageId
+     *
+     * @return array|null
+     */
+    private function getAvatarUrl(int $imageId): ?array
+    {
+        $imagePath = _PS_PROFILE_IMG_DIR_ . $imageId . '.jpg';
```
PierreRambaud: put _PS_PROFILE_IMG_DIR_ in the constructoir please

Progi1984: Done

classes/Profile.php
```diff
@@ -52,6 +52,30 @@ class ProfileCore extends ObjectModel
 
     protected static $_cache_accesses = [];
 
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct($id = null, $idLang = null, $idShop = null, $translator = null)
```
PierreRambaud, line 58:
```suggestion
    public function __construct(?int $id = null, ?int $idLang = null, ?int $idShop = null, ?Translator $translator = null)
```

src/Adapter/Image/Uploader/ProfileImageUploader.php
```diff
@@ -0,0 +1,90 @@
...
+        string $tmpImageDir = _PS_TMP_IMG_DIR_
+    ) {
+        $this->profileImageDir = $profileImageDir;
+        $this->tmpImageDir = $tmpImageDir;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function upload($profileId, UploadedFile $image)
```
PierreRambaud, line 64:
```suggestion
    public function upload(int $profileId, UploadedFile $image): void
```

src/Adapter/Image/Uploader/ProfileImageUploader.php
```diff
@@ -0,0 +1,90 @@
...
+        $destination = $this->profileImageDir . $profileId . '.jpg';
+        $this->uploadFromTemp($tempImageName, $destination);
+    }
+
+    /**
+     * Deletes old image
+     *
+     * @param $id
+     */
+    private function deleteOldImage($id)
```
PierreRambaud:
```suggestion
    private function deleteOldImage($id): void
```

src/Core/Domain/Profile/QueryResult/EditableProfile.php
```diff
@@ -68,4 +78,12 @@ public function getLocalizedNames()
     {
         return $this->localizedNames;
     }
+
+    /**
+     * @return string|null
+     */
+    public function getAvatarUrl()
```
PierreRambaud:
```suggestion
    public function getAvatarUrl(): ?string
```

src/Core/Form/IdentifiableObject/DataHandler/ProfileFormDataHandler.php
```diff
@@ -57,6 +68,12 @@ public function create(array $data)
         /** @var ProfileId $profileId */
         $profileId = $this->bus->handle(new AddProfileCommand($data['name']));
 
+        /** @var UploadedFile $uploadedAvatar */
+        $uploadedAvatar = $data['avatarUrl'] ?? null;
+        if (!empty($uploadedAvatar) && $uploadedAvatar instanceof UploadedFile) {
```
PierreRambaud:
```suggestion
        if ($uploadedAvatar instanceof UploadedFile) {
```

src/Adapter/Profile/QueryHandler/GetProfileForEditingHandler.php
```diff
@@ -26,28 +26,56 @@
...
+
+    /**
+     * @param ImageTagSourceParserInterface|null $imageTagSourceParser
+     * @param string $imgDir
+     */
+    public function __construct(
+        ImageTagSourceParserInterface $imageTagSourceParser = null,
+        string $imgDir = _PS_PROFILE_IMG_DIR_
+    ) {
+        $this->imageTagSourceParser = $imageTagSourceParser ?? new ImageTagSourceParser(__PS_BASE_URI__);
```
jolelievre: Maybe we should add a deprecate notice that `null` won't be allowed in next major?

jolelievre (dismissed):
One comment, but it's more a suggestion
Not blocking for the PR

src/Adapter/Profile/QueryHandler/GetProfileForEditingHandler.php
Progi1984: @jolelievre Have you got an example ?

jolelievre: For once I do ^^
https://github.com/PrestaShop/PrestaShop/blob/42b365326376591dd471359497b5d41fe1942279/src/Core/Grid/GridFactory.php#L78

jolelievre: I admit we don't do it everytime we should, but it's a good practice to help us identify the pieces we will have to remove in the next major

Progi1984: Done :heavy_check_mark:

jolelievre (dismissed):

jolelievre (dismissed):

PierreRambaud (dismissed):

SD1982:
Hi @Progi1984 All the steps from 1 to 21 are OK but after the step 21 it seems that the hook not override correctly and i still have the "default' avatars or the translator avatar created in the previous steps.

The QA module is installed and configured with the script given in the step 25 and I still have the translator avatar:
image

SD1982:
LGTM The employee avatar is correctly overrided by the `actionOverrideEmployeeImage` hook
image

Thanks @Progi1984 !!

atomiix (approved):

matks (approved):

Progi1984:
Thanks @SD1982

Issue #10621, opened 2018-09-21
Migrate "Shop parameters > Order settings > Statuses > Add new / edit order status" page


matks:
## Status

The "Order Statuses" page has been migrated to Symfony but is hidden because there is a few remaining issues. I mistakenly tried to fix all the remaining issues in a single PR https://github.com/PrestaShop/PrestaShop/pull/20442 because each issue needs only some lines of code, so I wanted to submit one PR with medium size rather than PRs with each 10 lines of code.

**It was a mistake** because one issue was eventually found to be much complicated than expected and the PR is blocked so none of the issues are fixed. I will probably have to circle back to submit 1 PR per issue.

So this issue is blocked because of this. Work has been done in https://github.com/PrestaShop/PrestaShop/pull/20442 to fix it but QA has found issues.

<hr>

Part of Symfony migration project
Specs:
https://github.com/PrestaShop/prestashop-specs/pull/83

Todo:
- [x] Migrate page https://github.com/PrestaShop/PrestaShop/pull/17215
- [x] Removes the Icon input (as it is not used anymore)
- [x] @LouiseBonnard  rewording to do, especially on "delivery"
- [x] #19496 Add information message for multistore

matks:
@MatShir Warning: while poking around with @sowbiba we found that enabling one checkbox shows an additional input in the form at the end, with a "preview" button. But we did not understand what it does.

MatShir:
Hi, @matks and @sowbiba new specs of the page have been provided check it on https://github.com/PrestaShop/prestashop-specs/pull/83. If you find missing or unclear behaviors don't hesitate to comment on the PR or DM me.

matks:
Will be done by https://github.com/PrestaShop/PrestaShop/pull/17215

Issue #10619, opened 2018-09-21
Migrate "Shop parameters > Order settings > Statuses" page


matks:
## Status

The "Order Statuses" page has been migrated to Symfony but is hidden because there is a few remaining issues. I mistakenly tried to fix all the remaining issues in a single PR https://github.com/PrestaShop/PrestaShop/pull/20442 because each issue needs only some lines of code, so I wanted to submit one PR with medium size rather than PRs with each 10 lines of code.

**It was a mistake** because one issue was eventually found to be much complicated than expected and the PR is blocked so none of the issues are fixed. I will probably have to circle back to submit 1 PR per issue.

So this issue is blocked because of this. Work has been done in https://github.com/PrestaShop/PrestaShop/pull/20442 to fix it but QA has found issues.

<hr>

Part of Symfony migration project
Specs:
https://github.com/PrestaShop/prestashop-specs/pull/83

Todo:
- [x] Migrate page https://github.com/PrestaShop/PrestaShop/pull/17215
- [x] Removes the column Icon (as it is not used anymore)
- [x] @LouiseBonnard  rewording to do, especially on "delivery"
- [x] #19495 Add information message for multistore
- [x] [Add a modal to confirm or cancel delete in bulk action](https://github.com/PrestaShop/PrestaShop/issues/14462)
  - #27606
- [x] [Add a modal to confirm or cancel delete action for one row](https://github.com/PrestaShop/PrestaShop/issues/17082)
  - #27606

marionf:
To be done during the migration of the page : 

**Features:**
- [x] #15228

**Issue for this page:**
- [x] #12105
- [x] #13838

LouiseBonnard:
Sooo, I've just checked the wording with @MatShir and I have two recommendations regarding the 'Delivery' column and the (wrongly) related tickbox 'Show delivery PDF' - undoubtedly an inheritance of the 1.5 version. Here it is:

- ~keep the 'Delivery' column and name the tickbox with the following wording, _Consider the associated order in delivery._~

- **rename the column 'In transit' and name the tickbox with the following wording, _Set the order as in transit._**

Both solutions suit me fine, I then let @MatShir pick the one he prefers!

LouiseBonnard:
Hi @sowbiba, we've just made up our minds with @MatShir, we will keep the second option. Could you please apply it to the related PR, #17215? Many thanks!

matks:
[...]

PR #20442, opened 2020-08-04
Finish order statuses

### LINKED ISSUE #19495: Multistore - Shop parameters > Order settings > Statuses - Add information message
This page is available only in all shops context, other contexts are not displayed in the header.

We should add an information message: "Note that this page is available in all shops context only, this is why your context has just switched."

The All Shops context should be forced.

[Specs](https://github.com/PrestaShop/prestashop-specs/blob/master/broader-topics/multistorespecialsspecs.md#list-of-pages-for-specific-case-3-the-page-can-be-displayed-only-in-all-shops-context)

### LINKED ISSUE #19496: Multistore - Shop parameters > Order settings > Statuses > Add new / edit order status - Add information message
Currently, this page don’t have shop association block.
In this case, the element is added for all shops, no matter the context selected

To avoid frustration, we should inform the merchants of this behavior:
Display an information message on this page: "Note that this feature is available in all shops context only. It will be added to all your stores."

The All Shops context should be forced.

This is specified in the [multistore specials cases specs](https://github.com/PrestaShop/prestashop-specs/blob/master/broader-topics/multistorespecialsspecs.md#list-of-pages-for-specific-case-2-add-new-pages-without-shop-association-block)

### LINKED ISSUE #19497: Multistore - Shop parameters > Order settings > Statuses > Add new / edit order return status - Add information message
Currently, this page don’t have shop association block.
In this case, the element is added for all shops, no matter the context selected

To avoid frustration, we should inform the merchants of this behavior:
Display an information message on this page: "Note that this feature is available in all shops context only. It will be added to all your stores."

The All Shops context should be forced.

This is specified in the [multistore specials cases specs](https://github.com/PrestaShop/prestashop-specs/blob/master/broader-topics/multistorespecialsspecs.md#list-of-pages-for-specific-case-2-add-new-pages-without-shop-association-block)

### LINKED ISSUE #20441: Create a new Order State or Create a new Order Return State cannot work with empty color
#### Describe the bug

On migrated Symfony pages "Create a new Order State" or "Create a new Order Return State" the "color" field can be empty bu the page crashes.

If you create a new "order state" or "order return state" and you do not fill the "color" field, leaving it empty, the page crashes. This is because color is actually needed 😅 

#### Expected behavior

You should not be able to submit the form without providing a valid value for the "color" field.

#### Steps to Reproduce

1. Go to migrated Order State page `/admin-dev/index.php/configure/shop/order-states` then "Add new order status" button
2. Fill in a name but leave "color" field empty
3. Submit the form

5. Go to migrated Order State page `/admin-dev/index.php/configure/shop/order-states` then "Add new order return status" button
6. Fill in a name but leave "color" field empty
7. Submit the form
8. See error

#### Additional information

* PrestaShop version: develop

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | This PR brings in multiple small things, see list below
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/19495 https://github.com/PrestaShop/PrestaShop/issues/19496 https://github.com/PrestaShop/PrestaShop/issues/20441 https://github.com/PrestaShop/PrestaShop/issues/19497
| How to test?  | See all tickets. This PR also "enables" the Symfony pages for "Order States" pages* . This means I remove old legacy files that should now be useless and all links inside the BO that target these pages should now go to the migrated page

:warning: After merge, ping @boubkerbribri for automatisation ;) 

## Description

- add multistore messages for the 3 new migrated Order State pages (see screenshots below)
- enable "info" flash messages
- enable the legacy_link for Order State pages
- remove legacy files
- fix one bug in these pages https://github.com/PrestaShop/PrestaShop/issues/20441 

## Scope

When this PR is merged, we can close the following "migration" issues:
- https://github.com/PrestaShop/PrestaShop/issues/10619
- https://github.com/PrestaShop/PrestaShop/issues/10620
- https://github.com/PrestaShop/PrestaShop/issues/10621

## Screenshots

Capture d’écran 2020-08-04 à 12 01 01
Capture d’écran 2020-08-04 à 12 03 17
Capture d’écran 2020-08-04 à 12 03 26

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20442)


Progi1984 (dismissed):

Progi1984:
@matks Please fix Travis CI : https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/368202266#L681

Progi1984 (dismissed):

LouiseBonnard:
Hi, thanks for the screenshots! It looks like an information notification rather than a success notification (the green one, with a check) so my recommendation would be to localize them in _Admin.Notifications.Info_ instead.

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
```diff
@@ -113,6 +121,14 @@ public function searchGridAction(Request $request)
      */
     public function createAction(Request $request)
     {
+        if ($this->get('prestashop.adapter.multistore_feature')->isActive()) {
+            $this->addFlash(
+                'info',
+                $this->trans('Note that this feature is available in all shops context only. It will be added to all your stores.',
+                    'Admin.Notifications.Success')
```
matks:
```suggestion
                    'Admin.Notifications.Info')
```

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
```diff
@@ -153,6 +169,14 @@ function (array $language) {
      */
     public function editAction(int $orderStateId, Request $request)
     {
+        if ($this->get('prestashop.adapter.multistore_feature')->isActive()) {
+            $this->addFlash(
+                'info',
+                $this->trans('Note that this feature is available in all shops context only. It will be added to all your stores.',
+                    'Admin.Notifications.Success')
```
matks:
```suggestion
                    'Admin.Notifications.Info')
```

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
```diff
@@ -198,6 +222,14 @@ function (array $language) {
      */
     public function createOrderReturnStateAction(Request $request)
     {
+        if ($this->get('prestashop.adapter.multistore_feature')->isActive()) {
+            $this->addFlash(
+                'info',
+                $this->trans('Note that this feature is available in all shops context only. It will be added to all your stores.',
+                    'Admin.Notifications.Success')
```
matks:
```suggestion
                    'Admin.Notifications.Info')
```

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
```diff
@@ -230,6 +262,14 @@ public function createOrderReturnStateAction(Request $request)
      */
     public function editOrderReturnStateAction(int $orderReturnStateId, Request $request)
     {
+        if ($this->get('prestashop.adapter.multistore_feature')->isActive()) {
+            $this->addFlash(
+                'info',
+                $this->trans('Note that this feature is available in all shops context only. It will be added to all your stores.',
+                    'Admin.Notifications.Success')
```
matks:
```suggestion
                    'Admin.Notifications.Info')
```

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
```diff
@@ -70,6 +70,14 @@ public function indexAction(
             $this->buildFiltersParamsByRequest($request, $orderReturnStatesFilters)
         );
 
+        if ($this->get('prestashop.adapter.multistore_feature')->isActive()) {
+            $this->addFlash(
+                'info',
+                $this->trans('Note that this page is available in all shops context only, this is why your context has just switched.',
+                    'Admin.Notifications.Success')
```
matks:
```suggestion
                    'Admin.Notifications.Info')
```

matks:
> Hi, thanks for the screenshots! It looks like an information notification rather than a success notification (the green one, with a check) so my recommendation would be to localize them in _Admin.Notifications.Info_ instead.

Done in https://github.com/PrestaShop/PrestaShop/pull/20442/commits/26e8bc7602dbf2d623c811aef4ae63cf6c6a74e4 😄

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
```diff
@@ -70,6 +70,14 @@ public function indexAction(
             $this->buildFiltersParamsByRequest($request, $orderReturnStatesFilters)
         );
 
+        if ($this->get('prestashop.adapter.multistore_feature')->isActive()) {
+            $this->addFlash(
+                'info',
+                $this->trans('Note that this page is available in all shops context only, this is why your context has just switched.',
+                    'Admin.Notifications.Info')
+            );
```
PierreRambaud:
```suggestion
                $this->trans(
                    'Note that this page is available in all shops context only, this is why your context has just switched.',
                    'Admin.Notifications.Info'
                )
            );
```
Better indentation, thinking something was missing :sweat_smile:

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
```diff
@@ -113,6 +121,14 @@ public function searchGridAction(Request $request)
      */
     public function createAction(Request $request)
     {
+        if ($this->get('prestashop.adapter.multistore_feature')->isActive()) {
+            $this->addFlash(
+                'info',
+                $this->trans('Note that this feature is available in all shops context only. It will be added to all your stores.',
+                    'Admin.Notifications.Info')
```
PierreRambaud:
```suggestion
                $this->trans(
                    'Note that this feature is available in all shops context only. It will be added to all your stores.',
                    'Admin.Notifications.Info'
                )
```

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
```diff
@@ -153,6 +169,14 @@ function (array $language) {
      */
     public function editAction(int $orderStateId, Request $request)
     {
+        if ($this->get('prestashop.adapter.multistore_feature')->isActive()) {
+            $this->addFlash(
+                'info',
+                $this->trans('Note that this feature is available in all shops context only. It will be added to all your stores.',
+                    'Admin.Notifications.Info')
```
PierreRambaud:
```suggestion
                $this->trans(
                    'Note that this feature is available in all shops context only. It will be added to all your stores.',
                    'Admin.Notifications.Info'
                )
```

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
```diff
@@ -230,6 +262,14 @@ public function createOrderReturnStateAction(Request $request)
...
     public function editOrderReturnStateAction(int $orderReturnStateId, Request $request)
     {
+        if ($this->get('prestashop.adapter.multistore_feature')->isActive()) {
+            $this->addFlash(
+                'info',
+                $this->trans('Note that this feature is available in all shops context only. It will be added to all your stores.',
+                    'Admin.Notifications.Info')
+            );
+        }
+
```
PierreRambaud: Can you create a method to do the job, you repeat the same condition everytime

matks: I was not sure about that (I was afraid of doing premature abstraction) but if you believe it's better, then let's do it 👍

matks: Done in https://github.com/PrestaShop/PrestaShop/pull/20442/commits/4153ab72cfb6a663bf7c67452c710bb6fd4f6feb

PierreRambaud (dismissed):

eternoendless (dismissed):
LGTM

Robin-Fischer-PS:
Hi @matks !

I have two issues with your PR : 

* When you access the Status page (and its sub-pages), there is the info message stating the context has been switched to "All shops", but the context displayed in the header hasn't change (it's still "Youpila" in my screenrecord instead of "All shops"). However, the context selected for actions is indeed "All shops".

* Color field cannot be filled in manually, it's always replaced by "#000000".

See screenrecord below for theses issues : 

https://drive.google.com/file/d/10hLsI67wMuLeuaiw6RLAXYDKboRDnMxX/view?usp=sharing

Thanks !

matks:
> Hi @matks !
> 
> I have two issues with your PR :
> 
> * When you access the Status page (and its sub-pages), there is the info message stating the context has been switched to "All shops", but the context displayed in the header hasn't change (it's still "Youpila" in my screenrecord instead of "All shops"). However, the context selected for actions is indeed "All shops".

Indeed.

In order to fix that, I created a service able to force the "all shops" context.

> * Color field cannot be filled in manually, it's always replaced by "#000000".

We use [Bootstrap ColorPicker](https://itsjavi.com/bootstrap-colorpicker/). There is an option to avoid forcing "only valid values": it's `autoInputFallback` (see [the documentation](https://itsjavi.com/bootstrap-colorpicker/tutorial-Basics.html) > Disable invalid color auto replacement)

I enabled it on the color picker.
It's not perfect but at least it's not trying to always auto-complete the input so it seems a good compromise to me.

If you want something better we'll need to hack how Bootstrap ColorPicker works 😞 

**Now waiting for code review**

src/PrestaShopBundle/Service/MultishopContextSwitcher.php
```diff
@@ -23,13 +23,31 @@
...
 
-header('Cache-Control: no-store, no-cache, must-revalidate');
-header('Cache-Control: post-check=0, pre-check=0', false);
-header('Pragma: no-cache');
+use PrestaShop\PrestaShop\Adapter\LegacyContext;
+use Shop;
 
-header('Location: ../../../../../../../../');
-exit;
+class MultishopContextSwitcher
```
matks, line 33: I put this class in `src/PrestaShopBundle/Service/` because it modifies prestashop business logic through the use of `Shop::setContext()` AND cookie through `context->cookie`. So it's an hybrid, it does belong to both application layer and domain layer. :/

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
Progi1984: @matks Suggestion not implemented

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/OrderStateController.php
```diff
@@ -398,4 +415,15 @@ private function getErrorMessages(\Exception $e)
...
     }
+
+    protected function addMultishopWarningFlashMessage(): void
+    {
+        if ($this->get('prestashop.adapter.multistore_feature')->isActive()) {
+            $this->addFlash(
+                'info',
+                $this->trans('Note that this feature is available in all shops context only. It will be added to all your stores.',
+                    'Admin.Notifications.Info')
+            );
```
Progi1984:
```suggestion
            $this->addFlash(
                'info',
                $this->trans(
                    'Note that this feature is available in all shops context only. It will be added to all your stores.',
                    'Admin.Notifications.Info'
                )
            );
```

Progi1984 (commented):
Small changes, but else I :+1:

matks:
> Small changes, but else I 👍

Applied, thanks 😉

Progi1984:
@matks It lacks @PierreRambaud's one.

matks:
> @matks It lacks @PierreRambaud's one.

Done

Progi1984 (approved):

PierreRambaud (approved):

Robin-Fischer-PS:
Hi @matks !

Two issues/questions with your fix : 

* The context is well changed when going on status page, but, if we want the same behaviour as in legacy, we should be switched back to the original context when going on another page (Dashboard for example). Do we want this behaviour here as well ?
* The color picker is still automatically filled, preventing the user from entering the RGB code directly in the field. He must select it in color picker or paste it in the field. But if there is no possibility to disable this auto-complete behaviour without also disabling the format verification on this field, I guess it's OK for me :)

https://drive.google.com/file/d/1ZTzdFSnJoRyOksMm3HGJLH2QQOLKoIEl/view?usp=sharing

matks:
> * The color picker is still automatically filled, preventing the user from entering the RGB code directly in the field.

By "entering", you mean "input the code, one character after one character" ?

Robin-Fischer-PS:
> > * The color picker is still automatically filled, preventing the user from entering the RGB code directly in the field.
> 
> By "entering", you mean "input the code, one character after one character" ?

Exactly :wink:

matks:
> * The color picker is still automatically filled, preventing the user from entering the RGB code directly in the field. He must select it in color picker or paste it in the field. But if there is no possibility to disable this auto-complete behaviour without also disabling the format verification on this field, I guess it's OK for me :)

So, according to [Bootstrap ColorPicker documentation](https://itsjavi.com/bootstrap-colorpicker/tutorial-Basics.html) we can disable the auto-complete BUT then we cannot force the user to use a specific color format.

This means that user can fill in
- `hsl(56, 93%, 63%)`
- `rgb(241, 138, 49)`
- `#16813D`

All of them are valid.

If we want user to use hexadecimal format (`#16813D`)  we have to enable the format validation ... which creates the auto-complete behavior you have seen 😞 

In my opinion, "autocomplete + right format" is the best answer but if you (or PM) want to change it, I can implement the change

Robin-Fischer-PS:
@PrestaShop/product-team What's your opinion on this ?

LouiseBonnard:
In my opinion, the color picker should work the same as on the Multistore and Attributes & Features page sof the back office. We are supposed to support only HTML colors. Auto-complete is fine for me as long as this behavior stays consistent. :-)

MatShir:
HTML colors corresponds to hexadecimal ? @LouiseBonnard

Also, it seems the typo size looks small compared to the labels in the order list (might a problem on my shop)? @Robin-Fischer-PS @matks

LouiseBonnard:
From what I know it can be the color name, the hex, and the RGB color codes: https://htmlcolorcodes.com. If using Bootstrap here modifies the formats allowed, alright for me but the tooltip needs to be reworded. :-)

Capture d’écran 2020-09-28 à 11 48 19

MatShir:
Thanks! I agree to have all the same behavior for now hex and color name seems to work. BUT the only thing that bothers me is the color preview on the picker is in RGB but we only allow hex and HTML color name. It might be confusing for the merchant. 
image

WDYT? @LouiseBonnard @matks

Robin-Fischer-PS:
@MatShir I also have a small font size : 

pr20442 small text

@matks Could you fix that ?

matks:
@MatShir @LouiseBonnard @Robin-Fischer-PS Hi ! Finally I have some time to come back on this PR.

I can fix the size of the order statuses.

Regarding the colorpicker, you have to understand that in this PR I did not "created" the colorpicker used on this page. This colorpicker is the one that comes with Bootstrap. This is the one that is available on all migrated pages (although I dont know if there are migrated pages that use it). The latest statements I see on this PR are talking about _modifying the behavior of this colorpicker_.

So this PR was supposed to be "just finish and clean Order Status BO pages to release it" and now the scope has evolved into "modify the colorpicker users in migrated Back Office Theme to behave like old Back Office Theme" 😅  .

@Robin-Fischer-PS can I modify this PR and make sure that it does not remove legacy files so this BO page stays "hidden" even if we merge this PR, then we merge this PR, and then there will be another PR to deal with the colorpicker stuff?

LouiseBonnard:
@matks, I'm catching up with what we said three months ago (😅) but from what I can read: there is no change, we keep the same color picker component as on the _Multistore_ and _Attributes & Features_ pages, we only need to rephrase the help text since the formats mentioned in it are not up-to-date.

matks:
> @matks, I'm catching up with what we said three months ago (😅) but from what I can read: there is no change, we keep the same color picker component as on the _Multistore_ and _Attributes & Features_ pages, we only need to rephrase the help text since the formats mentioned in it are not up-to-date.

The colorpicker component on pages is the colorpicker of old Back Office Theme, used on legacy BO Pages.

The "Order Statuses" page has been migrated to Symfony and new Back Office Theme, so it uses the colorpicker component of the new Back Office Theme.

So the 2 colorpickers are _not the same_. This is the key issue.

So if you say "we need to use the same colorpicker" it means
- I must, in the migrated Order Status BO page, _remove_ the colorpicker of the new Back Office Theme
- then I must _re-initialize_ and enable the colorpicker of the old Back Office Theme in this BO page (I have no idea how)
- and finally I must make the colorpicker of the old Back Office Theme work with the form (filling the input and so on)

That's a lot more work 😅  basically it's "pick a frontend component of old theme and make it work in new theme" (instead of using the component already enabled and configured in new theme)

LouiseBonnard:
Right, then I suggest 1/ we use 'the component already enabled and configured in the new theme' for this page (part of the scope, not difficult, everybody is happy) and 2/ create an issue to standardize its use by replacing the old color pickers in the back office whenever it will be possible. :-)

matks:
I close this PR as I will have to split it into small bits to make each of them validated and merged.

Issue #14776, opened 2019-07-19
[EPIC] Product page - SEO tab 8/9


marionf:
### - Here is the list of the elements of SEO tab:
[https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#seo-tab-epic](https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#seo-tab-epic) 

- [x] #19677 - [SEO tab] - Search Engine Optimization 1/2
- [x] #19678 - [SEO tab] - Redirection page 2/2

### [- As a merchant I want to know if there are errors when I complete a field in the product page](https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#seo-tab-)

---
## Issues in **SEO tab** sorted by priority high to low :

**Should-Have**
- [x] #14984 - SEO preview on product page
- [ ] #15936 - Friendly url not create in some case.
- [ ] #17078 - Product Meta description not filled by default in the SEO preview

## Improvement
- [ ] #14317 - Merchant should be able to choose to index/noindex each prestashop url

## Fixed
- [x] #14054 - URL in SEO preview is not correct when the product is not active
- [x] #21495 - Bug on "reset url" button of the product pages when "Accented URLs" is enabled and there are special characters in the product name

marionf:
### - Issues in **SEO tab** sorted by priority high to low :

**Should-Have**
- [ ] #14984
- [ ] #15936
- [ ] #17078

**Improvement**
- [ ] #14317

## Fixed
- [x] #14054
- [x] #21495 Bug on "reset url" button of the product pages when "Accented URLs" is enabled and there are special characters in the product name

PR #20473, opened 2020-08-05
Add UpdateProductSeoCommand

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | UpdateProductSeo command and handler covered with basic behat scenarios
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #14776

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20473)


classes/Product.php
```diff
@@ -193,7 +195,7 @@ class ProductCore extends ObjectModel
      *
      * @see ProductInterface
      */
-    public $redirect_type = '';
+    public $redirect_type = ProductInterface::REDIRECT_TYPE_NOT_FOUND;
```
matks: Are you sure this is good ? "Empty string" seems to be a valid value for this variable (see `Product::toggleStatus()`)

src/Core/Domain/Product/QueryResult/ProductForEditing.php
```diff
@@ -24,10 +24,15 @@
...
  */
 
+declare(strict_types=1);
+
 namespace PrestaShop\PrestaShop\Core\Domain\Product\QueryResult;
 
 /**
  * Product information for editing
+ *
+ * @todo: should we follow {...}Information or {..}Options concept? do we need those suffixes at all?
```
matks: I think these names are valid when used in the right place.

1. For a product, you can enable multiple shipping options. Whether it's the weight, the carriers, the shipping fee ... it's settings ("options" and "settings" seem close) that merchant modifies to configure its sales flow.

2. However there are also "Informations" that merchant just puts there to be displayed to customer. Product description, product creation time, product expiration date ... it's "informations". They dont modify how the shop behaves they are just shown to customer.

In my opinion, suffixes are good because they carry an additional layer of informations to the developer. It reminds me of the debate "should we put `Interface` suffix for all interfaces or not ?

`ProductCategories` => what can I get from this name ? well, this is a variable that contain ... stuff about product and categories. Is it a relationship ? Is it references ? Is it informations ?

`ProductCategoriesAssociation` => aaaah a bit better

I'm from the "Java" school 😄  I like overdescriptive variable names. Better too much infos than not enough

src/Core/Domain/Product/ValueObject/ProductRedirectOption.php
```diff
@@ -0,0 +1,227 @@
...
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\ValueObject;
+
+use LogicException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use PrestaShop\PrestaShop\Core\Product\ProductInterface;
+
+/**
+ * Holds values for product redirect option.
+ * todo: there are type constants declared already in @var ProductInterface. Should I reuse them? deprecate them?
```
matks: I dont know 😅  i'm fifty-fifty on this, between "reuse it" or "deprecate it"

@PrestaShop/prestashop-core-developers help 😅

src/Core/Domain/Product/ValueObject/ProductRedirectOption.php
```diff
@@ -0,0 +1,227 @@
...
+     * @var int
+     */
+    private $redirectTargetId;
+
+    /**
+     * Builds self with a type of no_redirect, automatically filling the only available target id value for this type
+     *
+     * @return static
+     */
+    public static function buildNoRedirect(): self
```
matks: I see you're starting to put these static constructors everywhere 😄 
Nice ! it's indeed a very good practice

classes/Product.php
zuk3975: No im not sure :D but the empty string is equal to 404(at least its handled equally), so idk how to make it right :thinking:

src/Core/Domain/Product/ValueObject/ProductRedirectOption.php
zuk3975: still relevant

classes/Product.php
zuk3975: still relevant

jolelievre: When you look a the front productController indeed the default behaviour is to match `REDIRECT_TYPE_NOT_FOUND` So I guess it would make sense to force a real default value instead of an empty one
Although if we do that we should probably update `toggleStatus` as well so that it correctky set the default value, and maybe introduce a new const `ProductInterface::DEFAULT_REDIRECT_TYPE` that would contain `REDIRECT_TYPE_NOT_FOUND` this would allow to have a single place to modify if we want to change this behaviour someday

classes/Product.php
```diff
@@ -164,7 +164,9 @@ class ProductCore extends ObjectModel
     /** @var string|array Meta description or array of meta description by id_lang */
     public $meta_description;
 
-    /** @var string|array Meta keywords or array of meta keywords by id_lang */
+    /**
+     * @deprecated
+     */
     public $meta_keywords;
```
jolelievre, line 171: I believe meta_keywords is still used vie `FrontController::getTemplateVarPage` which automatically sets them depending on the page type, `Meta::getMetaTags` is able to retrieve the meta keywords from a Product object

So I think we should handle them as well, even though it seems to be useless for SEO today we still handle them in the migration of CMS or Category pages so out of consistency we should also handle them here The day we remove the meta tag from classic theme we will be able to remove all these

So maybe we can leave a deprecated message indicating that it's useless for SEO now and will likely be removed in 1.8 version, but for now it's safer to manage its saving

src/Core/Domain/Product/Command/UpdateProductSeoCommand.php
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * @param string $redirectType
+     * @param int $redirectTargetId
+     *
+     * @return UpdateProductSeoCommand
+     */
+    public function setRedirectOption(string $redirectType, int $redirectTargetId): self
+    {
+        if ($redirectType === ProductRedirectOption::TYPE_NO_REDIRECT) {
+            $this->redirectOption = ProductRedirectOption::buildNoRedirect();
```
jolelievre: Since @eternoendless recently advised not to use this kind of factory methods I would suggest an alternative that is easy to use as well Create a `ProductNoRedirectOption` class that extends `ProductRedirectOption` and forces the preset values for no redirect

src/Core/Domain/Product/ValueObject/ProductRedirectOption.php
jolelievre: I would be in favor to keep a single place to store these, maybe not necessarily in `ProductInterface` if you think it's not the appropriate place
But I don't think these should be define in a ValueObject either, usually in this kind of situation we create a dedicated class like `PrestaShop\PrestaShop\Core\Domain\Product\ProductRedirectionSettings` (so ProductInterface would use it to populate its const and could be deprecated)

src/Core/Domain/Product/ValueObject/ProductRedirectOption.php
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    /**
+     * @param int $id
+     *
+     * @throws ProductConstraintException
+     */
+    private function assertCategoryId(int $id): void
+    {
+        if ($id !== self::NO_TARGET_VALUE && $id <= 0) {
```
jolelievre: You shouldn't check `self::NO_TARGET_VALUE` when category redirection is set, it's only used for `TYPE_NO_REDIRECT` so this special use case feels weird it's equivalent to `$id !== 0 && $id <= 0` which is like `$id < 0`
It would be more logical remove this check and call `assertCategoryId` only when redirection type is `TYPE_CATEGORY_PERMANENT` or `TYPE_CATEGORY_TEMPORARY`

jolelievre: Unless `NO_TARGET_VALUE` is used to redirect to parent category? then maybe it's worth adding a dedicated const (event if it also contains 0) so that it's self explanatory

classes/Product.php
zuk3975: @matks should have something to say about this

classes/Product.php
matks: Then I'm Ok with this change 👍  thanks for the explanations

src/Core/Domain/Product/ValueObject/ProductRedirectOption.php
matks: > But I don't think these should be define in a ValueObject either, usually in this kind of situation we create a dedicated class like PrestaShop\PrestaShop\Core\Domain\Product\ProductRedirectionSettings (so ProductInterface would use it to populate its const and could be deprecated)

Agreed 👍

classes/Product.php
matks: @jolelievre When building this Command, Julius found out that **in current product page (177) there is no way to manage meta tags**. So the fields are in still in ObjectModel but they cannot be reached from BO.

We explored this and found out that meta tags are now 100% useless for SEO so we believe
- they were removed from BO on purpose
- they were left in ObjectModel, undeprecated, either for backward compatibility or because it was forgotten

This is why we agreed to "finish the job" by deprecating them. But this PR does not remove any feature, it only marks the field as deprecated.

src/Core/Domain/Product/Command/UpdateProductSeoCommand.php
matks: Indeed. @zuk3975 please see https://github.com/PrestaShop/PrestaShop/pull/20554 that introduces a `NoManufacturerId` . This is something we heavily discussed and agreed on: no more "zero value" or "no value" but rather "zero class" or "no class".

src/Core/Domain/Product/ValueObject/ProductRedirectOption.php
matks: The latest concept from https://github.com/PrestaShop/PrestaShop/pull/20554 should solve this. @jolelievre if I am not wrong here is the right place to introduce a `NoCategoryId` class ?

zuk3975: yup its the case `If no category is selected, the Main Category is used`.

zuk3975: So we have these cases:
1. No redirect -> we create new VO ProductNoRedirectOption which extends ProductRedirectOption and avoids parent constraints with forced values (as in #20554).
2. Category redirect (when category id is 0) -> whats here? (it is still a RedirectOption, just indicating that it uses some default category for redirection) - should we introduce something like "ProductCategoryDefaultRedirectOption"? Isn't it overengineering? its almost if we soon gonna have 1class=1option. 
3. all other types -> ProductRedirectOption($type, $targetId);
In this case there is no way to typehint it in command.
Then in SetRedirectOption (when building command) decide which object should be initiated. And all of that just to ensure command validity before reaching handler?

zuk3975: Its hard to come by a good solution. If we create 3 separate objects(e.g NoRedirectOption, RedirectToCategoryOption, RedirectToProductOption) we cannot typehint it in command, which seems even worse than having previous implementation (Having abstract class for typehint or an interface for VO also doesn't sound good). So I wait for @eternoendless decision on this.

zuk3975: We've decided to use simpliest solution with @matks. I wont use the VO at all, pass the redirectType and targetId to command and simply validate it inside the handler.

zuk3975: so finally, I've came up to a solution which you could see in this commit https://github.com/PrestaShop/PrestaShop/pull/20473/commits/de28e0b8c6f0496517a718023809f0fca2087698.
TLDR: added RedirectType and RedirectTarget VO's and wrapped them in RedirectOption which ensures their integrity before reaching handler. By far it seems the best solution for me

matks (dismissed):
I quite find RedirectType + RedirectTarget elegant 👍 LGTM

matks:
> I quite find RedirectType + RedirectTarget elegant 👍 LGTM

You have a conflict though 😢

zuk3975:
> You have a conflict though 😢

Not surprised :D

matks (dismissed):

src/Core/Domain/Product/ValueObject/RedirectType.php
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Holds valid value of product redirect type
+ */
+class RedirectType
+{
+    /**
+     * Represents value of no redirection. Page not found (404) will be displayed.
+     */
+    const TYPE_NO_REDIRECT = '404';
```
jolelievre:
```suggestion
    const TYPE_NOT_FOUND = '404';
```

I think it's more accurate, 404 has a specific meaning in HTTP so better not hide behind a word that loses its meaning.

tests/Integration/Behaviour/Features/Context/Domain/Product/CommonProductFeatureContext.php
```diff
@@ -129,6 +129,14 @@ public function assertProductFields(string $productReference, TableNode $table)
         $data = $table->getRowsHash();
 
         $this->assertBoolProperty($productForEditing, $data, 'active');
+        $this->assertStringProperty($productForEditing, $data, 'visibility');
```
jolelievre: Mmm, it's weird I believe these checks were removed from the Common feature to the `UpdateOptionsFeatureContext` in this PR https://github.com/PrestaShop/PrestaShop/pull/20554
Probably a rebase that reintroduced them

The idea that you started with other sub domains was to have a context dedicated to each product sub domains: SEO, Options, Shipping, ... Because the common context was getting bigger and bigger, so it should only deal with the "general" fields (only active seemed independent 😅 )

So in this idea `redirect_type` should be checked in the new `UpdateSeoFeatureContext` you introduced

tests/Integration/Behaviour/Features/Scenario/Product/update_seo.feature
```diff
@@ -0,0 +1,168 @@
...
+    And product product2 redirect target should be product1
+    When I update product product2 SEO information with following values:
+      | redirect_type    | 302-product               |
+      | redirect_target  |                           |
+    Then I should get error that product redirect_target is invalid
+    And product "product2" localized "meta_title" should be "en-US:product2 meta title"
+    And product "product2" localized "meta_description" should be "en-US:product2 meta description"
+    And product "product2" localized "link_rewrite" should be "en-US:waterproof-boots"
+    And product product2 should have following values:
+      | redirect_type    | 301-product                |
```
jolelievre, line 50:
```suggestion
      | redirect_type    | 302-product                |
```
It would seems more logical to change this value since the redirect_type is already `301-product`

tests/Integration/Behaviour/Features/Scenario/Product/update_seo.feature
```diff
@@ -0,0 +1,168 @@
...
+    And product product2 redirect target should be product1
+    When I update product product2 SEO information with following values:
```
jolelievre, line 52: I believe you didn't test any scenario where you set a category ID

This makes me think about other more complex/advanced scenario and use cases:

I have a product redirected to a specified product, I change it to redirect category without specifying a category ID:
- should we reset the redirect target as it was targeting a product?
- should we keep the initial ID even if it was previously a product ID and now points to a category ID?

More generically, should the handler check that the provided IDs (product or category), are valid ones (existing and active)?

jolelievre: Mmm, actually I believe the fact that we have a "double setter" manages the first question: `setRedirectOption` forces you to specify a target ID, so it will necessarily be overridden

The question about checking if IDs are valid in the handler remains though

tests/Integration/Behaviour/Features/Scenario/Product/update_seo.feature
zuk3975: I think you missed something. It is the outcome (the assertion), not the event. Im asserting that the value haven't changed after exception was thrown

tests/Integration/Behaviour/Features/Scenario/Product/update_seo.feature
zuk3975: To update the option it is required to provide both values, which was the purpose of this VO (ensure that it is the expected intention to change both values because they are related).
I've added asserts to check if the entity exists depending on the redirect type :+1:

src/Adapter/Product/CommandHandler/UpdateProductSeoHandler.php
```diff
@@ -0,0 +1,145 @@
...
+     * @throws ProductNotFoundException
+     */
+    private function fillRedirectOptionValues(Product $product, RedirectOption $redirectOption): void
+    {
+        $redirectType = $redirectOption->getRedirectType();
+        $redirectTarget = $redirectOption->getRedirectTarget();
+
+        if ($redirectType->isProductType()) {
+            $this->assertProductExists($redirectTarget->getValue());
+        } elseif (!$redirectType->isTypeNotFound() && !$redirectTarget->isNoTarget()) {
```
jolelievre, line 105: I wonder if it would be useful to have `isCategoryType()` maybe worth adding it before feature freeze since after the class will be "locked" (suggestion for future PR)

tests/Integration/Behaviour/Features/Scenario/Product/update_seo.feature
jolelievre: Oh right my bad 😅

jolelievre (approved):

src/Adapter/Product/CommandHandler/UpdateProductSeoHandler.php
zuk3975: As it was said multiple times, we shouldn't implement "Future methods" expecting that they might be useful for someone some day, so I guess not :man_shrugging:

matks (approved):

src/Adapter/Product/CommandHandler/UpdateProductSeoHandler.php
zuk3975: I've changed my mind, yes you are right. We should improve it in some other pr, it will become much cleaner to use `$redirectType->isCategoryType()` instead of `!$redirectType->isTypeNotFound()`

PR #20482, opened 2020-08-06
Set CartRule amount with tax in BO instead of without tax

### LINKED ISSUE #20462: Wrong rounding of the cart rule
If a product at 20EUR is ordered without a discount.
And In the BO order page, you add 10 EUR discount

Then you will see the error, the 10 EUR discount becomes a 9.99 EUR discount

screenshot_at_10-52-59

Case: 
If the product is at 14.30 EUR, the discount is at 9.99.
BUT you edit the product at 15 EUR, it goes at 10.00

If the product is at 17.00 EUR, the discount is at 9.99.
BUT you edit the product at 18.00 EUR, it goes at 10.00

_Originally posted by @MatShir in https://github.com/PrestaShop/PrestaShop/issues/20397#issuecomment-668510305_

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Since we compute all the cart rules when editing/adding/deleting products in the BO, because of the way cart rules amount added in the BO was saved we miss precision and that led to incorrect discount amount. This PR fixes it.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | yes (function parameter ignored)
| Deprecations? | no
| Fixed ticket? | Fixes #20462
| How to test?  | 1. Create a product with Price tax included= 14.30, Tax rule =20%<br>2. Go to FO, add the last product created to the cart => complete the order<br>3. Go to the BO => Orders => View the last Order created<br>4. Add a new discount 10 euro => You should see the total discount = 10€ (same thing on the invoice)<br>5. Try to update the product's price with 15€, 17€ & 20€ (TTC) and check that the discount value is still 10€

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20482)

### DIFF
--- a/src/Core/Cart/Calculator.php
+++ b/src/Core/Cart/Calculator.php
@@ -27,6 +27,8 @@
 namespace PrestaShop\PrestaShop\Core\Cart;
 
 use Cart;
+use Context;
+use Tools;
 
 /**
  * provides methods to process cart calculation.
@@ -70,7 +72,12 @@ class Calculator
      */
     protected $isProcessed = false;
 
-    public function __construct(Cart $cart, $carrierId)
+    /**
+     * @var int|null
+     */
+    protected $computePrecision;
+
+    public function __construct(Cart $cart, $carrierId, ?int $computePrecision = null)
     {
         $this->setCart($cart);
         $this->setCarrierId($carrierId);
@@ -78,6 +85,11 @@ class Calculator
         $this->fees = new Fees();
         $this->cartRules = new CartRuleCollection();
         $this->cartRuleCalculator = new CartRuleCalculator();
+
+        if (null === $computePrecision) {
+            $computePrecision = Context::getContext()->getComputingPrecision();
+        }
+        $this->computePrecision = $computePrecision;
     }
 
     /**
@@ -117,16 +129,16 @@ class Calculator
     /**
      * run the whole calculation process: calculate rows, discounts, fees.
      *
-     * @param int $computePrecision
+     * @param int $computePrecision Not used since 1.7.7.0, kept for backward compatibility
      *
      * @return $this
      */
-    public function processCalculation($computePrecision)
+    public function processCalculation($computePrecision = null)
     {
         // calculate product rows
         $this->calculateRows();
         // calculate fees
-        $this->calculateFees($computePrecision);
+        $this->calculateFees();
         // calculate discounts
         $this->calculateCartRules();
         // store state
@@ -149,14 +161,14 @@ class Calculator
         }
 
         $amount = $this->getRowTotalWithoutDiscount();
-        $amount = $amount->sub($this->getDiscountTotal());
+        $amount = $amount->sub($this->rounded($this->getDiscountTotal(), $this->computePrecision));
         $shippingFees = $this->fees->getInitialShippingFees();
         if (null !== $shippingFees) {
-            $amount = $amount->add($shippingFees);
+            $amount = $amount->add($this->rounded($shippingFees, $this->computePrecision));
         }
         $wrappingFees = $this->fees->getFinalWrappingFees();
         if (null !== $wrappingFees) {
-            $amount = $amount->add($wrappingFees);
+            $amount = $amount->add($this->rounded($wrappingFees, $this->computePrecision));
         }
 
         return $amount;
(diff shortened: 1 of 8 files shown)


matks:
Behat tests failing
```
 Failed scenarios:
    Features/Scenario/Order/add_cart_rule_to_order.feature:98
    Features/Scenario/Order/add_cart_rule_to_order.feature:123
459 scenarios (457 passed, 2 failed)
6753 steps (6749 passed, 2 failed, 2 skipped)
```

matks:
I put this PR back in "In progress" column in Kanban because the work needed on tests might take long :/

tests/Integration/Behaviour/Features/Scenario/Cart/Calculation/CartRule/percent_multiple.feature
```diff
@@ -51,9 +51,7 @@ Feature: Cart rule (percent) calculation with multiple cart rules
     Then cart rule "cartrule2" has a contextual reduction value of 29.72
     When I use the discount "cartrule3"
     Then cart rule "cartrule3" has a contextual reduction value of 5.944
-    Then my cart total should be 33.7462 tax included
-    #known to fail on previous
-    #Then my cart total using previous calculation method should be 33.75 tax included
+    Then my cart total should be precisely 33.75 tax included
```
matks, line 54: ❤️

matks (dismissed):
The code seems OK.

You have modified values inside `add_cart_rule_to_order.feature`, the new values make sense ? Nothing to be scared of ?

atomiix:
> You have modified values inside `add_cart_rule_to_order.feature`, the new values make sense ? Nothing to be scared of ?

@matks yep, If you do the calculation manually it make more sense now than before!

Progi1984 (dismissed):

src/Core/Cart/Calculator.php
```diff
@@ -70,14 +72,24 @@ class Calculator
...
     {
         $this->setCart($cart);
         $this->setCarrierId($carrierId);
         $this->cartRows = new CartRowCollection();
         $this->fees = new Fees();
         $this->cartRules = new CartRuleCollection();
         $this->cartRuleCalculator = new CartRuleCalculator();
+
+        if (null === $computePrecision) {
+            $computePrecision = Context::getContext()->getComputingPrecision();
```
jolelievre: Since we have the cart it would be better to get the precision from the cart's currency directly instead of the context Even though in FO context they are supposed to be the same it would be safer and it would avoid coupling this class with the Context

src/Core/Cart/Calculator.php
```diff
@@ -117,16 +129,16 @@ public function addCartRule(CartRuleData $cartRule)
     /**
      * run the whole calculation process: calculate rows, discounts, fees.
      *
-     * @param int $computePrecision
+     * @param int $computePrecision Not used since 1.7.7.0, kept for backward compatibility
```
jolelievre, line 134: Add `@deprecated` maybe?

jolelievre: Besides you don't use the parameter any more at all, so it's completely ignored this looks like a BC break since you can't force the precisions via this parameter any more
You should check if the value is null, and use the class field as the default fallback in that case

src/Core/Cart/Calculator.php
atomiix: I used Context because that's what we used 2 of the 3 times we use `Calculator` but it make sense to use the cart. I'll do the change!

src/Core/Cart/Calculator.php
atomiix: @jolelievre I don't think we can add `@deprecated` to a function's param 🤔.

I don't think it's a BC break, the computing precision was used in a sub calculation of fees that we never return as it. Plus, we should use the same precision for all calculations if we want to have coherent results.

SD1982:
Hi @atomiix 

Everithing is good until we edit the product price ... After editing the product price the invoice  is incorrect

**Invoice before product price edit**
add-discount-invoice

**Invoice after product price edit**
edit-product-price-invoice


**_EDIT_**
This behaviour seems to be present also without the PR

atomiix:
@SD1982 Yep, this behavior is fixed in #20492 😉

SD1982:
Hi @atomiix so as the problem that I found yesterday is in fact not related to this PR I think this PR can be validated
If the review is Ok it's Ok for me !
Thanks @atomiix !

jolelievre (approved):

matthieu-rolland (approved):

jolelievre:
Thank you @atomiix

PR #20491, opened 2020-08-07
Fix UpdateCustomizationFieldsHandler todo

### LINKED ISSUE #19681: [Options tab] - Customization 3/5
As a merchant, I want to be able to manage the options of my products.

**Add a customization field**: You can click several times on this button to add multiple customization fields. Fields are displayed on the product page in front-office. When you click on it that opens a form with the following fields:
  * **Label**: Enter the name of the field which is displayed on the product page in front-office. A drop-down next to the field is displayed when there are several languages installed \(disabled languages are diplayed in the drop-down\). It allows to choose in which language the field is displayed. Also, in the case of multilanguage, if an error occurs, then I must be warned in the message which language is problematic \([improvement 16299](https://github.com/PrestaShop/PrestaShop/issues/16299)\)
  * **Type**: Drop-down with 3 options: text, file & numeric. By default text is selected. If you choose “text” in front-office you can complete the field with alphanumeric chars. If you choose “file”, you can upload a file. If you choose “numeric” you can enter only numbers \([improvement 11795](https://github.com/PrestaShop/PrestaShop/issues/11795)\)
  * **Delete**: Button to delete the field. When you click on it, you have a modal to cancel or confirm the action.
  * **Required**: Checkbox not checked by default. When checked the field is required in front-office and you can’t add the product to the cart while it’s not completed.

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fixes todo. See below
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #19681

(Fixes #19681)

## Description

**1.** To be closer to current convention (`Add` commands used to always return new resource id, so such new type of Bulk update command should do the same), instead of returning array of DTO's(CustomizationField) , the UpdateCustomizationFieldsHandler now returns array of CustomizationFieldIds.
**2.** requires all properties except id in CustomizationField DTO which is used in UpdateCustomizationFieldsHandler. The handler is complex and it decides weather to add, update or delete resource, so it is less confusing to just let the id be optional, but require all other fields.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20491)

### DIFF
--- a/src/Adapter/Product/CommandHandler/UpdateProductCustomizationFieldsHandler.php
+++ b/src/Adapter/Product/CommandHandler/UpdateProductCustomizationFieldsHandler.php
@@ -38,8 +38,8 @@ use PrestaShop\PrestaShop\Core\Domain\Product\Customization\CommandHandler\Delet
 use PrestaShop\PrestaShop\Core\Domain\Product\Customization\CommandHandler\UpdateCustomizationFieldHandlerInterface;
 use PrestaShop\PrestaShop\Core\Domain\Product\Customization\CommandHandler\UpdateProductCustomizationFieldsHandlerInterface;
 use PrestaShop\PrestaShop\Core\Domain\Product\Customization\CustomizationField;
-use PrestaShop\PrestaShop\Core\Domain\Product\Customization\Query\GetProductCustomizationFields;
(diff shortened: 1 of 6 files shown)


matks (approved):

matthieu-rolland (approved):

matthieu-rolland:
thanks @zuk3975 !

PR #20499, opened 2020-08-07
Ecotax is displayed tax excl instead of tax incl in FO

### LINKED ISSUE #18835: Ecotax is displayed tax excl instead of tax incl in FO
#### Describe the bug

A clear and concise description of what went wrong.

#### Expected behavior

Explain what you expected to happen instead.

#### Steps to Reproduce

1) I enable ecotax in international > VAT
2) I define a tax for my ecotax of 20%
Capture du 2018-03-15 16-05-05

3) I set ecotax = 1€ tax incl
The price tax excl is not well recalculated (issue https://github.com/PrestaShop/PrestaShop/issues/9855)
Sans titre2

4) I save my product, reload the page ==> the ecotax = amount Tax Excl. The VAT on the ecotax has disappear in the nature. And the product price tax incl is also wrong (should be 12€): issue [#10026](https://github.com/PrestaShop/PrestaShop/issues/10026)
Capture du 2018-03-15 16-40-38

5) The amount displayed on FO is not correct. It displays the Ecotax Tax Excl (0.83) whereas it should display Tax Incl (1€)
 Capture du 2018-03-15 16-43-03

#### Additional information

* PrestaShop version: 1.7.5, 1.7.6, 1.7.7
* PHP version: N/A

PululuK:
| Questions     | Answers
| Branch?       | develop
| Description?  |Ecotax is displayed tax excl instead of tax incl in FO. More details here https://github.com/PrestaShop/PrestaShop/issues/18835 by @marionf 
| Type?         | bug fix
| Category?     | FO 
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/18835.
| How to test?  | See https://github.com/PrestaShop/PrestaShop/issues/18835 by @marionf 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20499)


controllers/front/ProductController.php
```diff
@@ -1027,7 +1027,8 @@ public function getTemplateVarProduct()
         $product['minimal_quantity'] = $this->getProductMinimalQuantity($product);
         $product['quantity_wanted'] = $this->getRequiredQuantity($product);
         $product['extraContent'] = $extraContentFinder->addParams(['product' => $this->product])->present();
-        $product['ecotax'] = Tools::convertPrice((float) $product['ecotax'], $this->context->currency, true, $this->context);
+        $ecotax_rate = (float) Tax::getProductEcotaxRate();
+        $product['ecotax'] = Tools::ps_round((float) $product['ecotax'] * (1 + $ecotax_rate / 100), 6, null);
```
Progi1984:
```suggestion
        $product['ecotax'] = Tools::ps_round((float) $product['ecotax'] * (1 + $ecotax_rate / 100), $this->context->currency->precision, null);
```

Progi1984 (commented):
A small feedback about the precision of `ps_round`.

matks:
Hi, the PR seems nice however I got [some feedback](https://github.com/PrestaShop/PrestaShop/issues/18835#issuecomment-626588648) from a module developer that told me this issue is already solved on some Themes where they convert the price.

So if we change `$product['ecotax']` to put the tax incl. value in it, these themes will apply taxes on a data that has tax already included 😅 

He suggested to
- create a new property inside `$product` (`$product = $this->objectPresenter->present($this->product);`)
- put the correct value into this new property
- modify Classic Theme to use the new value

This way we keep the old value as it was, preserving compatibility for the themes which already handled the conversion.

PululuK:
> Hi, the PR seems nice however I got [some feedback](https://github.com/PrestaShop/PrestaShop/issues/18835#issuecomment-626588648) from a module developer that told me this issue is already solved on some Themes where they convert the price.
> 
> So if we change `$product['ecotax']` to put the tax incl. value in it, these themes will apply taxes on a data that has tax already included sweat_smile
> 
> He suggested to
> 
>     * create a new property inside `$product` (`$product = $this->objectPresenter->present($this->product);`)
> 
>     * put the correct value into this new property
> 
>     * modify Classic Theme to use the new value
> 
> 
> This way we keep the old value as it was, preserving compatibility for the themes which already handled the conversion.

Big thanks @matks !

In my opinion, we must :

- Create a static function that returns the ecotax based on the product id (_I don't know where to put it yet, but I think in the product class that would make sense_) + your suggestions ! This function will be reusable because ecotax is everywhere (BO, FO, PDF, ...)

I will check.

controllers/front/ProductController.php
PululuK: I will change the approach of my PR ... Thank you  @Progi1984!

src/Adapter/Presenter/Object/ObjectPresenter.php
```diff
@@ -52,6 +52,11 @@ public function present($object)
         foreach ($fields as $fieldName => $null) {
             $presentedObject[$fieldName] = $object->{$fieldName};
         }
+
+        if (get_class($object) === 'Product') {
```
matks: If merchant uses an override that extends `Product` class, this might not work ? 🤔 

Could we use something that works with children classes, like `instanceof` ?

matks (commented):
PR is good for me 👍 I added a suggestion to handle inheritance

I know however that another dev was wondering if we should stick to `float` or use [Number](https://github.com/prestashop/Decimal) to ensure we have accurate operation for `$this->ecotax * (1 + $ecotax_rate / 100)` 🤔

src/Adapter/Presenter/Object/ObjectPresenter.php
PululuK: Bien vu ! Thanks @matks

PululuK:
> PR is good for me +1 I added a suggestion to handle inheritance
> 
> I know however that another dev was wondering if we should stick to `float` or use [Number](https://github.com/prestashop/Decimal) to ensure we have accurate operation for `$this->ecotax * (1 + $ecotax_rate / 100)` thinking

@matks  I can open another PR, but here I try to stay consistent with the BO
See https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Model/Product/AdminModelAdapter.php#L538

classes/Product.php
```diff
@@ -8034,4 +8034,21 @@ public function updateDefaultSupplierData(int $idSupplier, string $supplierRefer
...
     }
+
+    /**
+     * Get Product ecotax
+     *
+     * @return ecotax
+     */
+    public function getEcotax($precision = null, $include_tax = true)
+    {
+        $precision = $precision ? $precision : Context::getContext()->currency->precision;
```
PierreRambaud:
```suggestion
        $precision = $precision ?? Context::getContext()->currency->precision;
```
or 
```suggestion
        $precision = $precision ?: Context::getContext()->currency->precision;
```

classes/Product.php
```diff
@@ -8034,4 +8034,21 @@ public function updateDefaultSupplierData(int $idSupplier, string $supplierRefer
...
+        $ecotax_rate = $include_tax ? (float) Tax::getProductEcotaxRate() : 0;
+
+        return  Tools::ps_round(
+            (float) $this->ecotax * (1 + $ecotax_rate / 100),
```
PierreRambaud, line 8054:
```suggestion
            (float) $this->ecotax * ((1 + $ecotax_rate) / 100),
```
You mean this?

PululuK: J'ai commité par erreur ! J'ai suis pas d'accord

PululuK: @PierreRambaud  it's correct
```php
$this->ecotax * (1 + $ecotax_rate / 100)
```

PierreRambaud (dismissed):

matks (dismissed):

ngodefroy:
Hello @PululuK 

I Have Tested the PR, everything is OK for me except the currency sign that disappeared. 

Thank you.

Capture d’écran 2020-08-14 à 17 03 27

themes/classic/templates/catalog/_partials/product-prices.tpl
```diff
@@ -79,7 +79,7 @@
 
     {block name='product_ecotax'}
       {if $product.ecotax.amount > 0}
-        <p class="price-ecotax">{l s='Including %amount% for ecotax' d='Shop.Theme.Catalog' sprintf=['%amount%' => $product.ecotax_tax_inc]}
+        <p class="price-ecotax">{l s='Including %amount% for ecotax' d='Shop.Theme.Catalog' sprintf=['%amount%' => $product.ecotax_tax_inc]} {$currency.sign}
```
matks: I think we should add the currency sign following how other variables carry it (like `$product.price_tax_exc`). I think these other variables use something like `formatPrice()` ? 🤔

PierreRambaud (approved):

matks (approved):

ngodefroy:
TEST OK:
BO:
https://drive.google.com/file/d/12MvPc-Svhx78Kl_ehSnyuTggehxwlpIQ/view
FO:
Capture d’écran 2020-09-04 à 11 08 23

matks:
Thank you @PululuK and @ngodefroy 

And @prestonBot 😄

PR #20518, opened 2020-08-10
Add GenerateProductCombinationsCommand

### LINKED ISSUE #14773: [EPIC] Product page - Combinations tab 5/9
### - As a merchant, I want to be able to manage my products with combinations

https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#combinations-tab-epic

- [ ] #20648 - [Combinations tab] - Combinaison List 1/2
- [ ] #20649 - [Combinations tab] - Combinaison Edit 2/2

### [- As a merchant I want to know if there are errors when I complete a field in the product page](https://github.com/PrestaShop/prestashop-specs/blob/master/back-office/catalog/catalog-products-add-edit.md#combinations-tab-)

---
### - Issues in **combinations tab** sorted by priority high to low :

**Must-Have**
- [ ] #16102 - As a merchant when I change a value for a combination, I want it to be automatically saved without clicking on Save
- [x] #9809 - [BOOM-4298] Large number of product combinations block browser
- [ ] #9826 - [BOOM-4202] Generate combinations
- [x] #14910 - Error when save an existing product before it is fully loaded PR: #15034
- [ ] #13479 - Automatically assign a "default" image to a product combination
- [ ] #9936 - [BOOM-3544] Product backoffice image hover alt text
- [ ]  #9972 - [BOOM-3165] Price calculation product combination HT / TTC

**Improvement**
- [ ] #11305 - Shipping fees per combination
- [ ] Product: allow up to 14 chars in EAN field PR #17362
- [x] #20536 - Merchant should be able to set delivery information for each combination
- [ ] #24778 - Be able to specify 'label when not in stock and backorders disabled' for each product
- [ ] #22165 - Saving product with combination without value in quantity field

**Should-Have**
- [ ] #9686 - [BOOM-5237] BO: Multistore | List of the attributes, with the association only to the second shop, doesn't appear in the product page PR #13217
- [x] #15405 - In product combinations the "Final retail price (tax excl.)" doesn't change setting "Impact on price (tax incl.)"
- [ ] #14913 - Variation product negative quantity bug
- [ ] #12814 - Wrong quantities showing in BO for products with combinations

**Nice to Have**
- [ ] #14372 - Delete product inventory at editing time (Persian language)
- [ ] #12244 - Multishop Default combinations not reassigned if you delete it
- [ ] #9737 - [BOOM-4827] Changing default combination
- [ ] #9601 - [BOOM-5849] All the attribute Size group are not selected, if we select the attribute Size and value ALL in the combination bar
- [ ] #10021 - Display the specific price in the combinations tab respecting the CLDR 

**New feature**
- [ ] http://forge.prestashop.com/browse/FF-74 - Work already started: #7890

## Bugs
- [x] #26611
- [x] #28430
- [ ] #32125 

## Fixed

- [x] #16103 - Add pagination for combinations listing
- [x] #16108 - As a merchant I want to be able to search and sort my combinations
- [x] #22788 - Add custom module fields in ProductCombination form 
- [x] #23922
- [x] #26424
- [x] #12440 - BO - Product page - combinations creation
- [x] #9910 - [BOOM-3704] Decimal display problems in combination page
- [x] #9896 - [BOOM-3825] Invalid price in combination
- [x] #15059 - When editing a cmbination, the ecotax field is tax incl
- [x] #16223 - As a merchant I need to know if the prices are displayed tax incl or tax excl
- [x] #18949 - [BO][IMPROVIMENT] Allow check all in product combinations
- [x] #14052 - Highlight element in dropdown menus of the product page when using keyboard
- [x] #16101 - As a merchant I want to edit my combination reference without editing it
- [x] #15416 - The final product price for the combination sould be displayed also tax included
- [x] #16100 - Collapse right column if combinations have already been generated
- [x] #9939 - [BOOM-3475] Seulement 5 attributs proposés / Only five attribtues shown

### LINKED ISSUE #21788: Consider configuration PS_ATTRIBUTE_ANCHOR_SEPARATOR in BO when building combination name
This is a reminder issue for Product page migration issued here https://github.com/PrestaShop/PrestaShop/pull/20518#discussion_r517874837

When generating BO combination names the `PS_ATTRIBUTE_ANCHOR_SEPARATOR` config was not taken into account up until now. Should we add it?
Discussed in slack with @MatShir @matks @jolelievre

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | GenerateProductCombinationsCommand saves new generated combinations to database.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #14773
| How to test?  | Automatic tests must pass

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20518)


src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
```diff
@@ -0,0 +1,153 @@
...
+        $this->saveProductAttributeAssociation($combinationId, $generatedCombination);
+
+        return new CombinationId($combinationId);
+    }
+
+    /**
+     * @param int $combinationId
+     * @param array $attributeIds
+     */
+    private function saveProductAttributeAssociation(int $combinationId, array $attributeIds): void
```
PululuK: To avoid having orphan combinations, `saveProductAttributeAssociation` should return a boolean and you should check that in `addCombination` and throw an exception if something bad happens (and delete the combination maybe ?)

zuk3975: Guess you are right, or ill just add db transaction and roll back in case of error.

eternoendless (changes requested):
> Reviewed 5 of 12 files at r1, 2 of 8 files at r2, 9 of 11 files at r3.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20518)* status: 16 of 19 files reviewed, 6 unresolved discussions (waiting on @PululuK and @zuk3975)

___
*[src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php, line 182 at r3](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEWtxuU2-3QzQ1XuFln:-MEWtxuU2-3QzQ1XuFlo:b2tqnip) ([raw file](https://github.com/prestashop/prestashop/blob/b7f1c55a37ca8d475adf46252afc1d1589078b9b/src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php#L182)):*
> ```
> Db::getInstance()
> ```
If you put this in a class attribute then you'll avoid N calls to Db::getInstance() (where N is the number of combinations to add)
___
*[src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php, line 183 at r3](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEWtEwgAcNK744ic-Bp:-MEWtEwgAcNK744ic-Bq:b1u0v2o) ([raw file](https://github.com/prestashop/prestashop/blob/b7f1c55a37ca8d475adf46252afc1d1589078b9b/src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php#L183)):*
> ```PHP
>         try {
>             if (!Db::getInstance()->insert('product_attribute_combination', $attributesList)) {
>                 $this->connection->rollBack();
> ```

You should rollback in the same block that started the transaction, trust me
Catch all exceptions, rollback, then throw them again.
___
*[src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php, line 48 at r3](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEWuX1R26SO1S2MC-dR:-MEWuX1R26SO1S2MC-dS:bnufumf) ([raw file](https://github.com/prestashop/prestashop/blob/b7f1c55a37ca8d475adf46252afc1d1589078b9b/src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php#L48)):*
> ```PHP
>     {
>         $product = $this->getProduct($query->getProductId());
>         //@todo: allow pagination?
> ```

Yes this will have to be paginated
___
*[src/Core/Domain/Product/Combination/Command/GenerateProductCombinationsCommand.php, line 50 at r3](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEWvF8Z5RJ_FGzjibNK:-MEWvF8Z5RJ_FGzjibNL:b-9hssfc) ([raw file](https://github.com/prestashop/prestashop/blob/b7f1c55a37ca8d475adf46252afc1d1589078b9b/src/Core/Domain/Product/Combination/Command/GenerateProductCombinationsCommand.php#L50)):*
> ```PHP
>     /**
>      * @param int $productId
>      * @param array $groupedAttributeIds key-value pairs where key is the attribute group id and value is the list of that group attribute ids
> ```

Consider using [psalm's array annotation style](https://psalm.dev/docs/annotating_code/type_syntax/array_types/): `@param array<int, array<int>>` 

(I'm guessing from your description that the array is indexed by int and values are arrays of int)
___
*[src/Core/Domain/Product/Combination/QueryResult/ProductCombinationForEditing.php, line 78 at r3](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEWwQ70CEBlyhAlZmh1:-MEWwQ70CEBlyhAlZmh2:bkd4027) ([raw file](https://github.com/prestashop/prestashop/blob/b7f1c55a37ca8d475adf46252afc1d1589078b9b/src/Core/Domain/Product/Combination/QueryResult/ProductCombinationForEditing.php#L78)):*
> ```PHP
>      * @return string
>      */
>     public function buildCombinationName(): string
> ```

Sounds like `getCombinationName`
If this one is expected to be called more than once you may consider doing it in the constructor then storing it as an attribute.

eternoendless (changes requested):
> Reviewed 5 of 6 files at r4.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20518)* status: 16 of 19 files reviewed, 3 unresolved discussions (waiting on @PululuK and @zuk3975)

___
*[src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php, line 161 at r4](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEYGy9B8lR6M4Hx5q8h:-MEYGy9B8lR6M4Hx5q8i:bqu2f3k) ([raw file](https://github.com/prestashop/prestashop/blob/ccf7b4aeff47b50f69c067a25e928c0577af82bc/src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php#L161)):*
> ```PHP
>         try {
>             $this->saveProductAttributeAssociation($combinationId, $generatedCombination);
>         } catch (CannotAddCombinationException $e) {
> ```

Maybe it's better to catch all exceptions? in what case would you get and exception and not rollback?
___
*[src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php, line 176 at r4](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEYI6p65CajSG25Pkui:-MEYI6p65CajSG25Pkuj:b-5dl6ht) ([raw file](https://github.com/prestashop/prestashop/blob/ccf7b4aeff47b50f69c067a25e928c0577af82bc/src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php#L176)):*
> ```PHP
>         $qb = $this->connection->createQueryBuilder();
>         $qb->select('pa.id_product_attribute')
>             ->addSelect('pa.id_product')
> ```

This field is removed by your array_map function below. You end up returning `int[]` of `id_product_attribute`

eternoendless (dismissed):
> Reviewed 3 of 4 files at r5.
> Dismissed @PululuK and @zuk3975 from a discussion.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20518)* status: 16 of 19 files reviewed, 2 unresolved discussions (waiting on @zuk3975)

___
*[src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php, line 186 at r5](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEb7E476bi72DWS7CC7:-MEb7E476bi72DWS7CC8:bjrqcbr) ([raw file](https://github.com/prestashop/prestashop/blob/13710662745a4b9bd58a36f82fba3294b387c9ce/src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php#L186)):*
> ```PHP
>      * @param int $limit
>      *
>      * @return array
> ```

I think the returned array looks like this

```
array<array{id_product_attribute: string, id_product: string}>
```

Again, I don't know why do you keep the `addSelect('pa.id_product')` if you're not using it
___
*[src/Core/Domain/Product/Combination/Query/GetProductCombinationsForEditing.php, line 68 at r5](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEb8MYn3--57T9BAlTs:-MEb8MYn3--57T9BAlTt:bqij7yn) ([raw file](https://github.com/prestashop/prestashop/blob/13710662745a4b9bd58a36f82fba3294b387c9ce/src/Core/Domain/Product/Combination/Query/GetProductCombinationsForEditing.php#L68)):*
> ```PHP
>         int $productId,
>         int $languageId,
>         //@todo: where do we put default limit? Or leave it required here and take care of it in presentation? (like in some grid filters)
> ```

A class constant I guess?

src/Adapter/Product/CombinationDataProvider.php
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * @param int $productId
+     * @param int $limit
+     * @param int $offset
+     * @param array $filters
+     *
+     * @return array
+     */
+    public function getCombinationsInfo(int $productId, int $limit, int $offset = 0, array $filters = []): array
```
jolelievre: Maybe allow `$limit = 0` as default value which would disable pagination (in case we need to get all the product combinations for another use case)

src/Adapter/Product/CombinationDataProvider.php
```diff
@@ -0,0 +1,178 @@
...
+        //@todo: filters not handled.
+        $qb = $this->getCombinationsQueryBuilder($productId)
+            ->select('pa.*')
+            ->setParameter('productId', $productId)
+            ->setFirstResult($offset)
+            ->setMaxResults($limit)
+        ;
+
+        return [
+            'combinations' => $qb->execute()->fetchAll(),
```
jolelievre: I think you could separate the pagination call from the count call (which would become public), this way each method has a distinct identified purpose and you have the possibility to get only the count separately
I you split them then this method should maybe be renamed into `getProductCombinations` or `listProductCombinations`

src/Adapter/Product/CombinationDataProvider.php
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * @param int[] $combinationIds
+     * @param LanguageId $langId
+     *
+     * @return array
+     */
+    public function getAttributesInfoByCombinationIds(array $combinationIds, LanguageId $langId): array
+    {
+        $qb = $this->connection->createQueryBuilder();
```
jolelievre: maybe we could separate these query calls into private functions, this would make it easier to read and understand: `getAssociatedAttributes`, `getAttributesInfos`, ...

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
```diff
@@ -0,0 +1,194 @@
...
+     * {@inheritdoc}
+     *
+     * @todo: multistore
+     */
+    public function handle(GenerateProductCombinationsCommand $command): array
+    {
+        $product = $this->getProduct($command->getProductId());
+        $generatedCombinations = $this->combinationGenerator->generate($command->getGroupedAttributeIds());
+
+        SpecificPriceRule::disableAnyApplication();
```
jolelievre: Could you explain with a comment why this is necessary please? I get that it's to improve performance but some details might help to understand what and when it disables exactly
If I get it right the rules are applied at each product/combination addition?

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
```diff
@@ -0,0 +1,194 @@
...
+     *
+     * @return CombinationId
+     */
+    private function addCombination(Product $product, array $generatedCombination): CombinationId
+    {
+        $newCombination = new Combination();
+        $newCombination->id_product = $product->id;
+        $newCombination->default_on = 0;
+
+        $this->connection->beginTransaction();
```
jolelievre: Using a transaction is a great idea, however two things are bothering me:
- to have a better impact on performance shouldn't the transaction begin at the start of the loop and commit at the end, here it is only used for One combination addition so it's not such a huge query
- I'm surprised that Doctrine connection and the legacy Db class or ObjectModel add/save methods are related? To really profit from the transaction I think the insertion request need to be performed via this Doctrine connection no?

src/Adapter/Product/CombinationDataProvider.php
zuk3975: :heavy_check_mark: I suppose null would be better then.

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
zuk3975: Im not really sure, why, but it was like that in legacy. Ill dig deeper and try to explain it.

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
zuk3975: * each combination is saved saparately, so we dont want to revert all combinations if only one of them fails :thinking: (although maybe its worth discussing if we could insert them by chunks?)
* I suppose transactions are handled in sql, it doesn't care about php part (who/how executed those queries) :thinking:

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
```diff
@@ -0,0 +1,136 @@
...
+
+        return new ProductCombinationsForEditing($totalCombinationsCount, $combinationsForEditing);
+    }
+
+    /**
+     * @param CombinationAttributeInformation[] $attributesInformation
+     *
+     * @return string
+     */
+    private function buildCombinationName(array $attributesInformation): string
```
jolelievre, line 136: We don't need the product name before the attribute list? Or is it concatenated after?

src/Core/Domain/Product/Combination/Command/GenerateProductCombinationsCommand.php
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * @param int $productId
+     * @param array<int, array<int>> $groupedAttributeIds key-value pairs where key is the attribute group id and value is the list of that group attribute ids
+     */
+    public function __construct(
+        int $productId,
+        array $groupedAttributeIds
+    ) {
+        $this->productId = new ProductId($productId);
+        //@todo: should I loop here and put to objects of AttributeGroupId and AttributeId?
```
jolelievre: Maybe it's not needed to have a VO here, but just some checking that the input value is valid (two dimensional arrays and positive integers) ?

src/Core/Domain/Product/Combination/Query/GetProductCombinationsForEditing.php
```diff
@@ -0,0 +1,109 @@
...
+     * @param int $productId
+     * @param int $languageId
+     * @param int $limit
+     * @param int $offset
+     */
+    public function __construct(
+        int $productId,
+        int $languageId,
+        //@todo: where do we put default limit? Or leave it required here and take care of it in presentation? (like in some grid filters)
+        int $limit,
```
jolelievre: If we set a default limit here, as I suggested earlier, it would be 0 or null to allow disabling pagination Other than that the default pagination number shouldn't be handled here but rather in the presentation code as you suggested

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
```diff
@@ -176,4 +177,12 @@ protected function parseLocalizedArray(string $parsedArray): array
 
         return $localizedArray;
     }
+
+    /**
+     * @return int
+     */
+    protected function getDefaultLangId(): int
+    {
+        return Configuration::get('PS_LANG_DEFAULT');
```
jolelievre:
```suggestion
        return (int) Configuration::get('PS_LANG_DEFAULT');
```

tests/Integration/Behaviour/Features/Context/Domain/Product/ProductCombinationFeatureContext.php
```diff
@@ -0,0 +1,136 @@
...
+     * @param string $groupsAttributesReferencesPairs
+     *
+     * @return array
+     */
+    private function parseAttributes(string $groupsAttributesReferencesPairs): array
+    {
+        $attributeByGroupIds = [];
+        $groupAttributeReferencesPairs = explode(';', $groupsAttributesReferencesPairs);
+        foreach ($groupAttributeReferencesPairs as $groupAttributeReferencesPair) {
+            $groupAttributePair = explode(':', $groupAttributeReferencesPair);
```
jolelievre:
```suggestion
            list($groupReference, $attributeReference) = explode(':', $groupAttributeReferencesPair);
            $groupId = $this->getSharedStorage()->get($groupReference);
            $attributeId = $this->getSharedStorage()->get($attributeReference);
```
Just a suggestion, I find `list` pretty convenient when dealing with explode

tests/Integration/Behaviour/Features/Context/Domain/Product/ProductCombinationFeatureContext.php
```diff
@@ -0,0 +1,136 @@
...
+            );
+        }
+    }
+
+    /**
+     * @param string $groupsAttributesReferencesPairs
+     *
+     * @return array
+     */
+    private function parseAttributes(string $groupsAttributesReferencesPairs): array
```
jolelievre: Seems like this method is not used, was it designed to check the Combination attributes after generation?
Although this check might be more accurate, the name comparison seems fine for now You can delete this one if you don't use it

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
zuk3975: hmm, its only the combination name like in Bo product page:
Screenshot from 2020-08-14 11-43-32

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
zuk3975: I've tested it, indeed it didn't work. Changed to use Dbinstance transaction, works as expected now.

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
```diff
@@ -0,0 +1,185 @@
...
+     *
+     * @return CombinationId
+     */
+    private function addCombination(Product $product, array $generatedCombination): CombinationId
+    {
+        $newCombination = new Combination();
+        $newCombination->id_product = $product->id;
+        $newCombination->default_on = 0;
+
+        $this->dbInstance->execute('START TRANSACTION');
```
jolelievre: Maybe we should add methods into the DB object instead, it would be cleaner

Also instead of directly executing the transaction it would be better to use the driver sub methods (for PDO, mysqli, ...) Because they also include some test about nested requests and so on, let's take profit of the expertise of existing code :)

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
```diff
@@ -0,0 +1,185 @@
...
+                $product->id,
+                var_export($generatedCombination, true)
+            ));
+        }
+        $combinationId = (int) $newCombination->id;
+
+        try {
+            $this->saveProductAttributeAssociation($combinationId, $generatedCombination);
+        } catch (Exception $e) {
+            $this->dbInstance->rollback();
```
jolelievre: This try/catch for rollback should englobe the whole transaction (start right before beginTransaction, and and right after commit) So that any exception will trigger the rollback

tests/Integration/Behaviour/Features/Context/Domain/Product/ProductCombinationFeatureContext.php
zuk3975: i tried to assert attribute ids, but then dropped (because actually the nmae seems to be enough). Forgot to delete it.

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
zuk3975: :heavy_check_mark:

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
zuk3975: I've put $product->setAvailableDate outside the loop, so there is no more sql actions inside, meaining I dont need to wrap a transaction anywhere further

zuk3975:
note - I've implemented completely new methods to retrieve product combinations and have put them in Adapter/CombinationDataProvider.
Although the same class contains similar methods and it might be possible to reuse some of them, not yet sure if its worth it, maybe that would clear up in the process (when building UpdateCombinationsCommand)?

jolelievre:
> note - I've implemented completely new methods to retrieve product combinations and have put them in Adapter/CombinationDataProvider.
> Although the same class contains similar methods and it might be possible to reuse some of them, not yet sure if its worth it, maybe that would clear up in the process (when building UpdateCombinationsCommand)?

I guess we'll see later if we need to clear it, and if it's acceptable regarding the BC breaks

jolelievre (changes requested):
> Reviewed 2 of 12 files at r1, 3 of 8 files at r6, 6 of 7 files at r9, 1 of 3 files at r10, 1 of 1 files at r11, 1 of 1 files at r12, 6 of 6 files at r13.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20518)* status: all files reviewed, 9 unresolved discussions (waiting on @zuk3975)

___
*[src/Adapter/CombinationDataProvider.php, line 235 at r13](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEhQ3mi8_Y3to2qbQqw:-MEhQ3mi8_Y3to2qbQqx:b-kat4hh) ([raw file](https://github.com/prestashop/prestashop/blob/92dff53ea7cf42df05d30713a1c1d7b58e16bfe3/src/Adapter/CombinationDataProvider.php#L235)):*
> ```PHP
>      * @return int
>      */
>     public function getTotalCombinationsCount(int $productId): int
> ```

We will need the `array $filters = []` here as well
___
*[src/Adapter/CombinationDataProvider.php, line 258 at r13](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEhQKig6GiymLiDbYUN:-MEhQKig6GiymLiDbYUO:b3p6llh) ([raw file](https://github.com/prestashop/prestashop/blob/92dff53ea7cf42df05d30713a1c1d7b58e16bfe3/src/Adapter/CombinationDataProvider.php#L258)):*
> ```PHP
>         $attributesInfo = $this->getAttributesInformation($attributeIds, $langId->getValue());
> 
>         $attributesInfoByAttributeId = [];
> ```

This loop could be done in the `getAttributesInformation` method
___
*[src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php, line 154 at r9](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEMphNFiNQZuPB3R8ls-r9-154:-MEhRFgT4xErr-x0JoFb:b-mq2nyd) ([raw file](https://github.com/prestashop/prestashop/blob/8bc617f4eb45dd684b1ce0132ba0d48eba27db4b/src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php#L154)):*
<details><summary><i>Previously, zuk3975 (Julius Žukauskas) wrote…</i></summary><blockquote>

I've put $product->setAvailableDate outside the loop, so there is no more sql actions inside, meaining I dont need to wrap a transaction anywhere further
</blockquote></details>

There is still `$newCombination->add()` which performs an insert Besides the commit should be in the try/catch block as well
___
*[src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php, line 114 at r13](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEhQtKYAMTLWpHe4lRg:-MEhQtKYAMTLWpHe4lRh:b-hgbaaf) ([raw file](https://github.com/prestashop/prestashop/blob/92dff53ea7cf42df05d30713a1c1d7b58e16bfe3/src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php#L114)):*
> ```PHP
> 
>             try {
>                 $product->setAvailableDate();
> ```

As we discussed this could be done outside of the loop
___
*[src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php, line 101 at r13](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEva7vXDpTe2Dlu-Ui_:-MEva7vXDpTe2Dlu-Uia:b-d8bhia) ([raw file](https://github.com/prestashop/prestashop/blob/92dff53ea7cf42df05d30713a1c1d7b58e16bfe3/src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php#L101)):*
> ```PHP
> 
>             foreach ($attributesInformationByCombinationId[$combinationId] as $attributesInfo) {
>                 $combinationAttributesInformation[] = new CombinationAttributeInformation(
> ```

I believe with the code moving into the provider, we go back to the previous "issue" that too many `CombinationAttributeInformation` objects are built, it is not necessary for now but maybe this could be optimized
___
*[src/Core/Domain/Product/Combination/QueryResult/ProductCombinationsForEditing.php, line 34 at r13](https://reviewable.io/reviews/prestashop/prestashop/20518#-MEvcE-8CiaKp5e3Oj6n:-MEvcE-8CiaKp5e3Oj6o:bgjzuu6) ([raw file](https://github.com/prestashop/prestashop/blob/92dff53ea7cf42df05d30713a1c1d7b58e16bfe3/src/Core/Domain/Product/Combination/QueryResult/ProductCombinationsForEditing.php#L34)):*
> ```PHP
>  * Transfers product combinations data
>  */
> class ProductCombinationsForEditing
> ```

Suggestion, it seems redundant to have Product as a prefix here, whereas `CombinationForEditing` doesn't have one, it' shorter and still easy to understand Maybe the idea is mainly not to confuse both with just adding an `s` ? If so I would suggest `CombinationListForEditing` or something similar which would be less confusing

jolelievre:
Note: most of my comments are just suggestions Only remaining point is about the try/catch block that manages commit/rollback

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
zuk3975: As you explained yourself in slack, Combination->add() calls SpecificPriceRulle->applyAllRules(). When $rules_application_enable = false it avoids performing specificPriceRule application on every new combination.

eternoendless (commented):
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20518)* status: 17 of 26 files reviewed, 10 unresolved discussions (waiting on @eternoendless, @jolelievre, and @zuk3975)

___
*[src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php, line 187 at r14](https://reviewable.io/reviews/prestashop/prestashop/20518#-MLDcdi38AhvyDm1eL7Z:-MLDcdi38AhvyDm1eL7_:bcdlnqd) ([raw file](https://github.com/prestashop/prestashop/blob/9e97603ea28bc5b5e3ed019fac331cf5e33ad710/src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php#L187)):*
> ```PHP
> 
>         try {
>             if (!$this->dbInstance->insert('product_attribute_combination', $attributesList)) {
> ```

Isn't this supposed to be done through `Combination->setAttributes()`?

zuk3975:
@eternoendless 
> Isn't this supposed to be done through Combination->setAttributes()?

Nope. During generation `Product->generateMultipleCombinations` is used which performs straight db insert :thinking:

src/Adapter/CombinationDataProvider.php
```diff
@@ -180,6 +200,71 @@ public function completeCombination($attributesCombinations, $product)
...
+            ->setParameter('productId', $productId)
+        ;
+
+        if ($offset) {
+            $qb->setFirstResult($offset);
+        }
+
+        if ($limit) {
+            $qb->setMaxResults($limit);
+        }
```
Progi1984:
```suggestion
        $qb = $this->getCombinationsQueryBuilder($productId, $filters)
            ->select('pa.*')
            ->setParameter('productId', $productId)
            ->setFirstResult($offset)
            ->setMaxResults($limit)
        ;
```

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
```diff
@@ -0,0 +1,195 @@
...
+        $productId = (int) $product->id;
+        $product->setAvailableDate();
+
+        $addedCombinationIds = [];
+        foreach ($generatedCombinations as $generatedCombination) {
+            $combinationExists = $product->productAttributeExists($generatedCombination, false, null, true);
+
+            if ($combinationExists) {
+                continue;
+            }
```
Progi1984:
```suggestion
            if ($product->productAttributeExists($generatedCombination, false, null, true)) {
                continue;
            }
```

src/Core/Domain/Product/AttributeGroup/Attribute/ValueObject/AttributeId.php
```diff
@@ -57,17 +57,32 @@ public function getValue()
         return $this->attributeId;
     }
 
+    /**
+     * @param $value
```
Progi1984:
```suggestion
     * @param mixed $value
```

src/Core/Domain/Product/AttributeGroup/Attribute/ValueObject/AttributeId.php
```diff
@@ -57,17 +57,32 @@ public function getValue()
...
+     *
+     * @return bool
+     */
+    public static function isValid($value): bool
```
Progi1984, line 65: Why a static ?

src/Core/Domain/Product/Combination/Command/GenerateProductCombinationsCommand.php
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * @param array $groupedAttributeIds
+     *
+     * @throws AttributeConstraintException
+     * @throws AttributeGroupConstraintException
+     */
+    private function assertAttributesListIsValid(array $groupedAttributeIds): void
+    {
+        foreach ($groupedAttributeIds as $groupId => $attributeIds) {
+            if (!AttributeGroupId::isValid($groupId)) {
```
Progi1984: Why not creator a specific validator rather than a static call ? Ping @PrestaShop/prestashop-core-developers

Progi1984 (commented):
Please define BC Breaks in Db.php

& Some feedbacks

src/Core/Domain/Product/Combination/Command/GenerateProductCombinationsCommand.php
zuk3975: refactored this at all and removed static methods.
Now it builds Value objects inside loop. (and the VO's are validating them selves)

src/Adapter/CombinationDataProvider.php
zuk3975: these are optional params, I cannot set `null` neither for FirstResult nor MaxResults

zuk3975:
> Please define BC Breaks in Db.php

Moved Db changes to separate PR

src/Adapter/CombinationDataProvider.php
Progi1984: Default values for FirstResult & MaxResults are null so you can set to `null`.

src/Adapter/Product/CommandHandler/GenerateProductCombinationsHandler.php
zuk3975: @matks, now that I moved Db transactions & combinations wrapping in separate PR this place is unstable I guess.

src/Adapter/Product/Create/CombinationCreator.php
```diff
@@ -0,0 +1,161 @@
...
+     * @return CombinationId
+     */
+    private function addCombination(int $productId, array $generatedCombination): CombinationId
+    {
+        $newCombination = new Combination();
+        $newCombination->id_product = $productId;
+        $newCombination->default_on = 0;
+
+        $combinationId = $this->combinationRepository->add($newCombination);
+        $this->combinationRepository->saveProductAttributeAssociation($combinationId, $generatedCombination);
```
jolelievre: As a safety, we could watch for `CannotAddCombinationException` when adding attributes to the combinations And if it is thrown the freshly created combination should be removed So that we don't end up with a half created Combination

src/Adapter/Product/Repository/CombinationRepository.php
```diff
@@ -0,0 +1,253 @@
...
+     * @param int $productId
+     * @param int $limit
+     * @param int $offset
+     * @param array $filters
+     *
+     * @return array<int, array<string, mixed>>
+     */
+    public function getProductCombinations(int $productId, ?int $limit = null, ?int $offset = null, array $filters = []): array
+    {
+        $qb = $this->getCombinationsQueryBuilder($productId, $filters)
```
jolelievre, line 120: As commented in the `GetProductCombinationsForEditingHandler`, this should check for product existence with `ProductRepository:assertProductExists`
Also I wonder if it should use `ProductId` as an input (since the handler uses a VO) or an `int` 🤷‍♂️

src/Adapter/Product/Repository/CombinationRepository.php
```diff
@@ -0,0 +1,253 @@
...
+            return $attributeByCombination['id_attribute'];
+        }, $attributeCombinationAssociations));
+
+        $attributesInfoByAttributeId = $this->getAttributesInformation($attributeIds, $langId->getValue());
+
+        $attributesInfoByCombinationId = [];
+        foreach ($attributeCombinationAssociations as $attributeCombinationAssociation) {
+            $combinationId = (int) $attributeCombinationAssociation['id_product_attribute'];
+            $attributeId = (int) $attributeCombinationAssociation['id_attribute'];
+            $attributesInfoByCombinationId[$combinationId] = $attributesInfoByAttributeId[$attributeId];
```
jolelievre:
```suggestion
            $attributesInfoByCombinationId[$combinationId][] = $attributesInfoByAttributeId[$attributeId];
```
Shouldn't it be a list, since a combination can have multiple attributes?

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
```diff
@@ -0,0 +1,138 @@
...
+        int $totalCombinationsCount
+    ): CombinationListForEditing {
+        $combinationsForEditing = [];
+
+        foreach ($combinations as $combination) {
+            $combinationId = (int) $combination['id_product_attribute'];
+            $combinationAttributesInformation = [];
+
+            foreach ($attributesInformationByCombinationId[$combinationId] as $attributesInfo) {
+                $combinationAttributesInformation[] = new CombinationAttributeInformation(
```
jolelievre: We already discussed about how this duplicates the `CombinationAttributeInformation` but it can be optimized later ^^

src/Adapter/Product/Repository/CombinationRepository.php
```diff
@@ -0,0 +1,253 @@
...
+        $attributesList = [];
+        foreach ($attributeIds as $attributeId) {
+            $attributesList[] = [
+                'id_product_attribute' => $combinationId->getValue(),
+                'id_attribute' => $attributeId,
+            ];
+        }
+
+        try {
+            if (!Db::getInstance()->insert('product_attribute_combination', $attributesList)) {
```
jolelievre, line 202: Maybe you should check if it exists before?

tests/Integration/Behaviour/Features/Context/Domain/Product/ProductCombinationFeatureContext.php
```diff
@@ -0,0 +1,116 @@
...
+        ));
+
+        Assert::assertEquals(
+            count($combinationsForEditing->getCombinations()),
+            $expectedCombinationsCount,
+            'Unexpected combinations count'
+        );
+
+        /** @var CombinationForEditing $combinationForEditing */
+        foreach ($combinationsForEditing as $key => $combinationForEditing) {
```
jolelievre:
```suggestion
        foreach ($combinationsForEditing->getCombinations() as $key => $combinationForEditing) {
```

tests/Integration/Behaviour/Features/Scenario/Product/generate_combinations.feature
```diff
@@ -0,0 +1,36 @@
...
+      | size       | [s,m]               |
+      | color      | [white,black,blue]  |
+    Then product product1 should have following combinations:
+      | combination name                 |
+      | Size - S, Color - White          |
+      | Size - S, Color - Black          |
+      | Size - S, Color - Blue           |
+      | Size - M, Color - White          |
+      | Size - M, Color - Black          |
+      | Size - M, Color - Blue           |
```
jolelievre: Could you add a test for pagination please? In another scenario maybe where you would get Page 1 of 3 and Page 2 of 3

jolelievre: The total combinations should also be tested

jolelievre: And the list of attributes is not tested either Maybe you could add a behat function to check:

```
And (first|second) combination of product product1 should have the following attributes:
    | size   | S        |
    | color | white |
```

For the offset either use plain words like what was done for invoice behat tests:
https://github.com/PrestaShop/PrestaShop/blob/c020552ca5d94aa1637fb8af6ccb37340c2c95be/tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php#L569

It's just a matching between word and index, it doesn't allow to check all positions but for the limited test cases it was enough

If you really want to be able to check any index just use an int with sentence `combination at position 12` or we could add a method in `PrimitiveUtils` to handle this

jolelievre: I actually discovered `PrimitiveUtils::castStringIntegerIntoInteger` existed so you can use this one, and maybe it could be improved to allow a little parsing of  `11th` `18th` `21st` `22nd` `23rd` (I believe there are only those four suffixes)

jolelievre (changes requested):
I'm not sure but I believe I spotted a bug in the `getAttributesInfoByCombinationIds` function And I wondered why tests didn't detect it but, again I'm not sure, it appears the combination list is not asserted correctly

matks (commented):
Nothing to add, @jolelievre said it all 👍

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
zuk3975: I don't recall what are you talking about, sorry (this PR is ~3 months old already and we've talked alot) :sweat_smile: 
But we can come back to it later if you say its not so important now

tests/Integration/Behaviour/Features/Context/Domain/Product/ProductCombinationFeatureContext.php
zuk3975: not sure how tests are working when there is a critical error like this. I guess the coverage is way too low

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
```diff
@@ -0,0 +1,149 @@
...
+    ): CombinationListForEditing {
+        $combinationsForEditing = [];
+
+        foreach ($combinations as $combination) {
+            $combinationId = (int) $combination['id_product_attribute'];
+            $combinationAttributesInformation = [];
+
+            foreach ($attributesInformationByCombinationId[$combinationId] as $attributesInfo) {
+                foreach ($attributesInfo as $attributeInfo) {
+                    $combinationAttributesInformation[] = new CombinationAttributeInformation(
```
jolelievre, line 112: In response to previous comment: To optimize the object creation (memory usage mainly) we could have only one instance of `CombinationAttributeInformation` per attribute_id and they would assigned/shared to the various `CombinationForEditing` that have those attributes in common

This would make sense when we have dozens of combinations to handle as it prevents duplicating this object for each combination that have it in common

But it's not mandatory for now

tests/Integration/Behaviour/Features/Context/Domain/Product/ProductCombinationFeatureContext.php
jolelievre: Well the object is not Traversable so I guess the loop was simply ignored 🤷‍♂️

src/Adapter/Product/Repository/CombinationRepository.php
jolelievre: You unresolved this one so I guess it doesn't seem relevant to you? ^^

I mean this to avoid doublons or insertion errors in case the association already exist, then the present associations should be ignored to avoid making all the other insertions fail

jolelievre: Oh my bad, actually you made a modif to check that all attributes exist I didn't think of that one but it's even safer indeed I still worry about doublons in relationship rows though That's what my initial comment was about but i understand now it was not clear enough ^^

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
```diff
@@ -0,0 +1,149 @@
...
+        $this->combinationRepository = $combinationRepository;
+        $this->productRepository = $productRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(GetProductCombinationsForEditing $query): CombinationListForEditing
+    {
+        $productId = $query->getProductId();
```
jolelievre, line 72: You think it's better to do the check in the handler rather than in the `CombinationRepository` in the end?

src/Adapter/Product/Repository/CombinationRepository.php
```diff
@@ -0,0 +1,294 @@
...
+     * @param Combination $combination
+     *
+     * @return CombinationId
+     *
+     * @throws CombinationConstraintException
+     * @throws CoreException
+     */
+    public function add(Combination $combination): CombinationId
+    {
+        //@todo: validation?
```
jolelievre: Maybe not necessary now since Combination are generated and have little data (and generated data is supposed to be valid) But when we will add combination manually some validation will probably be necessary (for prices, minimal quantity, ...)

tests/Integration/Behaviour/Features/Scenario/Product/generate_combinations.feature
```diff
@@ -0,0 +1,37 @@
...
+  Scenario: Generate product combinations
+    When I add product "product1" with following information:
+      | name       | en-US:universal T-shirt |
+      | is_virtual | false                   |
+    Then product product1 type should be standard
+    When I generate combinations for product product1 using following attributes:
+      | Size  | [S,M]              |
+      | Color | [White,Black,Blue] |
+    Then product product1 should have following combinations:
+      | combination name        | attributes           |
```
jolelievre, line 30: Nice format! 👍
It makes assertions data easy to write (and read)

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
zuk3975: I don't really now.
Im wondering if/when repository methods should even hold those assertions, or should it be left to some upper layer. (is it really concern of repository to check if some id exist? what if app allows inserting non-existing ids - then it would seem like core/domain logic, so it wouldn't go to repository)
Im afraid we will end up checking if product (or any other entity exists) multiple times per request if in each repository method we add assertion.
E.g. apply same principle in another PR (product duplicator) we are doing ~10 inserts to database with same product, so if we have a repository methods for each insertion & each of them would assert the same product over and over again it doesn't look right. :thinking: 
@matks maybe you could enlighten?

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
```diff
@@ -0,0 +1,149 @@
...
+     * @param CombinationAttributeInformation[] $attributesInformation
+     *
+     * @return string
+     */
+    private function buildCombinationName(array $attributesInformation): string
+    {
+        $combinedNameParts = [];
+        foreach ($attributesInformation as $combinationAttributeInformation) {
+            $combinedNameParts[] = sprintf(
+                '%s - %s',
```
Progi1984, line 141:
```suggestion
                '%s - %s',
```

Please change the ` - ` by the configuration option : `PS_ATTRIBUTE_ANCHOR_SEPARATOR`

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
```diff
@@ -0,0 +1,149 @@
...
+                $combinationAttributeInformation->getAttributeGroupName(),
+                $combinationAttributeInformation->getAttributeName()
+            );
+        }
+
+        return implode(', ', $combinedNameParts);
```
Progi1984, line 147: Please change the `,` by the configuration option : PS_ATTRIBUTE_ANCHOR_SEPARATOR

src/Adapter/Product/Repository/CombinationRepository.php
```diff
@@ -0,0 +1,294 @@
...
+            ->select('pa.*')
+        ;
+
+        if ($offset) {
+            $qb->setFirstResult($offset);
+        }
+
+        if ($limit) {
+            $qb->setMaxResults($limit);
+        }
```
Progi1984:
```suggestion
        $qb = $this->getCombinationsQueryBuilder($productId, $filters)
            ->select('pa.*')
            ->setFirstResult($offset)
            ->setMaxResults($limit)
        ;
```

tests/Integration/Behaviour/Features/Scenario/Product/generate_combinations.feature
```diff
@@ -0,0 +1,37 @@
...
+      | Size - S, Color - White | [Size:S,Color:White] |
+      | Size - S, Color - Black | [Size:S,Color:Black] |
+      | Size - S, Color - Blue  | [Size:S,Color:Blue]  |
+      | Size - M, Color - White | [Size:M,Color:White] |
+      | Size - M, Color - Black | [Size:M,Color:Black] |
+      | Size - M, Color - Blue  | [Size:M,Color:Blue]  |
+#@todo: test pagination. comment: https://github.com/PrestaShop/PrestaShop/pull/20518#discussion_r517391210
```
Progi1984: Need a scenario  with `PS_ATTRIBUTE_ANCHOR_SEPARATOR` = `-`

Progi1984 (commented):
Some feedback.

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
zuk3975: nice catch, thanks

src/Adapter/Product/Repository/CombinationRepository.php
zuk3975: I thought I answered this already :thinking: 
Anyway- $offset and $limit are optional params, I cannot set `null` neither as FirstResult nor as MaxResults

tests/Integration/Behaviour/Features/Scenario/Product/generate_combinations.feature
zuk3975: although it has one drawback, that reference should be identical to name. (can't do any better though)

src/Adapter/Product/Repository/CombinationRepository.php
Progi1984: `setFirstResult` & `setMaxResults` accept null or integer in parameter (:warning: it's not defined in PHPDoc :stuck_out_tongue:)

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
zuk3975: I checked.
So previous page didn't use this option when building name
CombinationDataProvider L188:
```
 private function getCombinationName($attributesCombinations)
    {
        $name = [];

        foreach ($attributesCombinations as $attribute) {
            $name[] = $attribute['group_name'] . ' - ' . $attribute['attribute_name'];
        }

        return implode(', ', $name);
    }
```
It seems that one might only be used in FO.
Btw it would change the `-` between attribute group and attribute ( but not the comma).

I wonder should we start using that option in BO as well?
maybe @MatShir could answer?
btw @Progi1984 , @matks would it be a BC break to if we start generating name according to that option in BO? (it doesn't look like one, but maybe someone is using the name to parse attributes or so..)

Progi1984: Wow : it's weird to have a separator in front office and a different one in back office. But I let @MatShir decide.

src/Adapter/Product/Repository/CombinationRepository.php
zuk3975: hmm well yeah, because its not typed, but in theory its not what method expects :thinking:
Anyway, ok lets use this :smile:

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
jolelievre: IMHO this is a parameter used for format display, it shouldn't be a BC break if we start using it in both BO and FO, there might be someone that parses this but it's not a public contract we ensure that `-` would ALWAYS and STRICTLY be used so that people can parse this data So not a BC break in my opinion, and definitely a godd thing to uniformise both BO and FO display

Progi1984: @jolelievre In addition to this, that should fix some bugs relative to a configuration not applied ;)

zuk3975: created reminder issue to avoid blocking PR for this.
https://github.com/PrestaShop/PrestaShop/issues/21788

tests/Integration/Behaviour/Features/Scenario/Product/generate_combinations.feature
zuk3975: created reminder issue to avoid blocking pr
https://github.com/PrestaShop/PrestaShop/issues/21788

src/Adapter/Product/QueryHandler/GetProductCombinationsForEditingHandler.php
matks: Haha 😅  this is a gray area, it's hard to define the border between Handler and Repository.

So if we come back to the comparison for web standard MVC
- Handler is like a Web Controller
- Repository is like the Model in a MVC web app

If tomorrow I send a HTTP `GET /app/products/17`request to get product data, I expect a HTTP 404 response if product 17 does not exist. How will I do that ?
Well I believe we will have
- at controller level a check "does product 17 exists ?" and if it does not we return HTTP 404
- at model level when data is being fetched from database we should either check if product exists to avoid MySQL error or we should simply check if SQL query works correctly

So my answer would be
- yes, we need to check if product exists in the Handler, and if it does not exist we can exit early
- depending on the usecase, sometimes it's good to perform the check AGAIN in the repository (because method could be called from another place) following defensive programming strategy ; but not always

src/Adapter/Product/Repository/CombinationRepository.php
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    /**
+     * @param int[] $combinationIds
+     *
+     * @return array<int, array<string, mixed>>
+     */
+    private function getAttributeCombinationAssociations(array $combinationIds): array
+    {
+        $qb = $this->connection->createQueryBuilder();
```
matks, line 217: I just realized this class uses a mix of ObjectModel / Db and Doctrine 🤔

matks (dismissed):
Looks good enough for me 👍

src/Adapter/Product/Repository/CombinationRepository.php
zuk3975: as mentioned in slack, its only due simplicity (so I don't need to write my own sql query when there is one written and working already). Legacy already provided method to insert multiple product attributes, so I just reused it

tests/Integration/Behaviour/Features/Scenario/Product/generate_combinations.feature
```diff
@@ -0,0 +1,59 @@
...
+      | Size - S, Color - White | [Size:S,Color:White] |
+      | Size - S, Color - Black | [Size:S,Color:Black] |
+    And I should see following combinations of product product1 in page 2:
+      | combination name        | attributes           |
+      | Size - S, Color - Blue  | [Size:S,Color:Blue]  |
+      | Size - M, Color - White | [Size:M,Color:White] |
+    And I should see following combinations of product product1 in page 3:
+      | combination name        | attributes           |
+      | Size - M, Color - Black | [Size:M,Color:Black] |
+      | Size - M, Color - Blue  | [Size:M,Color:Blue]  |
```
Progi1984, line 59: And what you see in page 4 ?

Progi1984 (commented):
A small feedback and it's ok for me

src/Adapter/Product/Repository/CombinationRepository.php
```diff
@@ -0,0 +1,289 @@
...
+     *
+     * @return Combination
+     *
+     * @throws CoreException
+     */
+    public function create(ProductId $productId): Combination
+    {
+        $combination = new Combination();
+        $combination->id_product = $productId->getValue();
+        $combination->default_on = 0;
```
Progi1984:
```suggestion
        $combination->default_on = false;
```

Progi1984 (commented):
A small feedback for Combination

matks (approved):

Progi1984 (approved):

src/Core/Domain/Product/Combination/QueryResult/CombinationListForEditing.php
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Combination\QueryResult;
+
+/**
+ * Transfers product combinations data
+ */
+class CombinationListForEditing
```
jolelievre, line 34: Since it will the result object for pagination I wonder if in the future we should add the offset and limit associated (so that you know this is page 1 over 3 for example)
But not needed for now

tests/Integration/Behaviour/Features/Scenario/Product/generate_combinations.feature
```diff
@@ -0,0 +1,60 @@
...
+  Scenario: Product combinations pagination returns correct results
+    Given product product1 should have following combinations:
+      | combination name        | attributes           |
+      | Size - S, Color - White | [Size:S,Color:White] |
+      | Size - S, Color - Black | [Size:S,Color:Black] |
+      | Size - S, Color - Blue  | [Size:S,Color:Blue]  |
+      | Size - M, Color - White | [Size:M,Color:White] |
+      | Size - M, Color - Black | [Size:M,Color:Black] |
+      | Size - M, Color - Blue  | [Size:M,Color:Blue]  |
+    When I limit product product1 combinations per page to 2
```
jolelievre, line 47: This is kind of a "stateful" way to handle pagination I'm not sure we should store this info but rather it should always be specified as a parameter of the step `Then I should see following combinations of product product1 in page 1:`

jolelievre: Because it increases the risk of different tests messing with each other data, or to forget a limit was set an not understand the result is not as expected

jolelievre (approved):
I commented about the behat test, the storage of the limit seems risky to me as it makes the tests stateful

We already have many issues related to the legacy cache which saves/stores some contextual and which ruin our life 😅  So we shouldn't add similar problems ourselves

However the rest of the PR is ok so I don't think it deserves to be blocked just for this, so maybe we should fix this in another PR if the team agrees with the modification?

matks:
Thank you @zuk3975 @jolelievre @Progi1984 !

PR #20501, opened 2020-08-07
Use static factories for bulk product update commands

zuk3975:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Static factories improves code readability. Change for UpdateProductSuppliersCommand, UpdateProductTagsCommand, UpdateProductPack, UpdateProductCategoriesCommand. (UpdateCustomizationFieldsCommand changed using same convention in another pr). In addition this also provides partial update endpoint for categories
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | none. Behats must pass

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20501)

PR #20532, opened 2020-08-11
Add RemoveAllAssociatedProductCategoriesCommand

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | There was  UpdateProductCategoriesCommand which handled the deletion of all categories too. Now it throws LogicException if empty categoryIds are provided. Added RemoveAllAssociatedProductCategoriesCommand for such usecase.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | raised from #20501

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20532)

### DIFF
--- a/src/Adapter/Product/AbstractProductCategoriesAssociationHandler.php
+++ b/src/Adapter/Product/AbstractProductCategoriesAssociationHandler.php
@@ -43,7 +43,7 @@ abstract class AbstractProductCategoriesAssociationHandler extends AbstractProdu
      * @param Product $product
      * @param int[] $categoryIds
      *
-     * Warning: $categoryIds will replace current categories, erasing previous data.
+     * Warning: $categoryIds will replace current categories, erasing previous data
      *
      * @throws CannotUpdateProductException
      * @throws ProductException
--- a/src/Adapter/Product/AbstractProductCategoriesAssociationHandler.php
+++ b/src/Adapter/Product/AbstractProductCategoriesAssociationHandler.php
@@ -37,7 +37,7 @@ use Product;
 /**
  * Provides reusable methods for product category handlers
  */
-abstract class AbstractProductCategoryHandler extends AbstractProductHandler
+abstract class AbstractProductCategoriesAssociationHandler extends AbstractProductHandler
 {
     /**
      * @param Product $product
--- a/src/Adapter/Product/AbstractProductCategoriesAssociationHandler.php
+++ b/src/Adapter/Product/AbstractProductCategoriesAssociationHandler.php
@@ -41,7 +41,7 @@ abstract class AbstractProductCategoriesAssociationHandler extends AbstractProdu
 {
     /**
      * @param Product $product
-     * @param array $categoryIds
+     * @param int[] $categoryIds
      *
      * @throws CannotUpdateProductException
      * @throws ProductException
@@ -72,7 +72,6 @@ abstract class AbstractProductCategoriesAssociationHandler extends AbstractProdu
      *
      * @throws CannotUpdateProductException
      * @throws ProductException
-     * @throws \PrestaShopDatabaseException
      */
     protected function updateDefaultCategory(Product $product, int $categoryId): void
     {
@@ -93,10 +92,9 @@ abstract class AbstractProductCategoriesAssociationHandler extends AbstractProdu
     }
 
     /**
-     * @param array $categoryIds
+     * @param int[] $categoryIds
      *
      * @throws CannotUpdateProductException
-     * @throws \PrestaShopDatabaseException
      */
     protected function assertCategoriesExists(array $categoryIds): void
     {
--- a/src/Adapter/Product/AbstractProductCategoriesAssociationHandler.php
+++ b/src/Adapter/Product/AbstractProductCategoriesAssociationHandler.php
@@ -43,6 +43,8 @@ abstract class AbstractProductCategoriesAssociationHandler extends AbstractProdu
      * @param Product $product
      * @param int[] $categoryIds
      *
+     * Warning: $categoryIds will replace current categories, erasing previous data.
+     *
      * @throws CannotUpdateProductException
      * @throws ProductException
      */
--- a/src/Adapter/Product/AbstractProductCategoryHandler.php
+++ b/src/Adapter/Product/AbstractProductCategoryHandler.php
@@ -1,0 +1,110 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Product;
+
+use Category;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\CannotUpdateProductException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
+use PrestaShopException;
+use Product;
+
+/**
+ * Provides reusable methods for product category handlers
+ */
+abstract class AbstractProductCategoryHandler extends AbstractProductHandler
+{
+    /**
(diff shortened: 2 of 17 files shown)


src/Adapter/Product/AbstractProductCategoryHandler.php
```diff
@@ -0,0 +1,110 @@
...
+use Product;
+
+/**
+ * Provides reusable methods for product category handlers
+ */
+abstract class AbstractProductCategoryHandler extends AbstractProductHandler
+{
+    /**
+     * @param Product $product
+     * @param array $categoryIds
```
eternoendless: Please use the phpdoc notation (`int[]`) or phpstan notation (`array<int>`)

src/Adapter/Product/CommandHandler/DeleteAllProductCategoriesHandler.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(DeleteAllProductCategoriesCommand $command): void
+    {
+        $product = $this->getProduct($command->getProductId());
+
+        $this->updateCategories($product, [$this->homeCategoryId]);
```
eternoendless: Does a product always need to be associated with a category?

src/Adapter/Product/AbstractProductCategoryHandler.php
```diff
@@ -0,0 +1,110 @@
...
+use Category;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\CannotUpdateProductException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
+use PrestaShopException;
+use Product;
+
+/**
+ * Provides reusable methods for product category handlers
+ */
+abstract class AbstractProductCategoryHandler extends AbstractProductHandler
```
eternoendless: I'm not a fan of abstract classes with utility methods. If you intend to keep this class abstract, all of its methods must be used in concrete classes (IS-A relationship), otherwise it becomes a utility service that should be injected (HAS-A relationship).

```suggestion
abstract class AbstractUpdateProductCategoriesAssociationHandler extends AbstractProductHandler
```

src/Core/Domain/Product/Command/DeleteAllProductCategoriesCommand.php
```diff
@@ -0,0 +1,58 @@
...
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Command;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+
+/**
+ * Deletes all product-category associations
+ */
+class DeleteAllProductCategoriesCommand
```
eternoendless: Avoid using "delete" when talking about relationships

```suggestion
class RemoveAllAssociatedProductCategoriesCommand
```

src/Core/Domain/Product/Command/UpdateProductCategoriesCommand.php
```diff
@@ -28,9 +28,13 @@
...
 namespace PrestaShop\PrestaShop\Core\Domain\Product\Command;
 
+use LogicException;
 use PrestaShop\PrestaShop\Core\Domain\Category\ValueObject\CategoryId;
 use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
 
+/**
+ * Updates product-category associations
+ */
 class UpdateProductCategoriesCommand
```
eternoendless: Avoid generic terms like "update"

```suggestion
class SetAssociatedProductCategoriesCommand
```

src/Core/Domain/Product/Command/UpdateProductCategoriesCommand.php
```diff
@@ -88,4 +87,32 @@ public function getCategoryIds(): array
     {
         return $this->categoryIds;
     }
+
+    /**
+     * @param int[] $categoryIds
+     */
+    private function setCategoryIds(array $categoryIds): void
+    {
+        $this->assertCategoryIdsAreNotEmpty($categoryIds);
```
eternoendless, line 96: I think this assertion is not needed, because `new CategoryId($id)` should issue an exception if `$id` is empty.

You could catch the exception and re issue another one if needed

src/Core/Domain/Product/Command/UpdateProductCategoriesCommand.php
```diff
@@ -88,4 +87,32 @@ public function getCategoryIds(): array
...
+        );
+    }
+
+    /**
+     * @param int[] $categoryIds
+     */
+    private function assertCategoryIdsAreNotEmpty(array $categoryIds): void
+    {
+        if (empty($categoryIds)) {
+            throw new LogicException(sprintf(
```
eternoendless: I'm not a fan of generic exceptions, I prefer using domain exceptions because they can be catched independently. If you choose to keep a generic exception, you're choosing to let it bubble up. In that case, I think the more descriptive exception would be `InvalidArgumentException`.

eternoendless (changes requested):

src/Core/Domain/Product/Command/UpdateProductCategoriesCommand.php
zuk3975: If its empty array, the loop inside array_map wont start and no exception will be thrown.

src/Core/Domain/Product/Command/UpdateProductCategoriesCommand.php
zuk3975: Indeed, I let it bubble up, because IMO it is purely technical error that does not involve any domain rules and the LogicException seems to fit this case.
According to PHP docs LogicException - `Exception that represents error in the program logic. This kind of exception should lead directly to a fix in your code.` (it does point directly to the command which is supposed to be used in this case)
Whereas InvalidArgumentException - `Exception thrown if an argument is not of the expected type.`

src/Adapter/Product/CommandHandler/DeleteAllProductCategoriesHandler.php
zuk3975: yes, according to products specification `A category is necessarily selected, it is impossible not to have one`. And current product page behavior - does not allow to unselect last category.

src/Core/Domain/Product/Command/UpdateProductCategoriesCommand.php
eternoendless: My mistake, I read too fast, for some reason I thought you were checking each element for non emptiness.

src/Core/Domain/Product/Command/UpdateProductCategoriesCommand.php
eternoendless: Yep, as I said in my other comment I thought you were verifying each category id independently for some reason, my bad.

That said, this sounds more like a Runtime than Logic error because the size of the array cannot be inspected statically.

eternoendless (commented):
Looks good!

eternoendless (dismissed):

src/Adapter/Product/AbstractProductCategoriesAssociationHandler.php
```diff
@@ -0,0 +1,108 @@
...
+use Product;
+
+/**
+ * Provides reusable methods for product category handlers
+ */
+abstract class AbstractProductCategoriesAssociationHandler extends AbstractProductHandler
+{
+    /**
+     * @param Product $product
+     * @param int[] $categoryIds
```
matks, line 44:
```suggestion
     * @param int[] $categoryIds
     *
     * Warning: $categoryIds will replace current categories, erasing previous data.
```

src/Adapter/Product/CommandHandler/RemoveAllAssociatedProductCategoriesHandler.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(RemoveAllAssociatedProductCategoriesCommand $command): void
+    {
+        $product = $this->getProduct($command->getProductId());
+
+        $this->updateCategories($product, [$this->homeCategoryId]);
```
matks, line 62:
```suggestion
        // remove all categories associated with this product, keep only home Category
        $this->updateCategories($product, [$this->homeCategoryId]);
```

By the way this will fail in multishop, right ? 😄  maybe we should put a `@todo`

matks (commented):
Small suggestions

src/Adapter/Product/CommandHandler/RemoveAllAssociatedProductCategoriesHandler.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(RemoveAllAssociatedProductCategoriesCommand $command): void
+    {
+        $product = $this->getProduct($command->getProductId());
+
+        // remove all categories associated with this product, keep only home Category
+        $this->updateCategories($product, [$this->homeCategoryId]);
```
matks, line 62:
```suggestion
        // @todo: this is likely to break with multishop
        $this->updateCategories($product, [$this->homeCategoryId]);
```

matks (dismissed):

eternoendless (dismissed):

src/Adapter/Product/AbstractProductCategoriesAssociationHandler.php
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * Warning: $categoryIds will replace current categories, erasing previous data.
```
matks:
```suggestion
     * Warning: $categoryIds will replace current categories, erasing previous data
```

eternoendless (approved):

matks (approved):

matks:
Thank you @zuk3975

PR #20537, opened 2020-08-11
Add $useCache argument to getModuleTranslation() to be able to work around getting wrong translations

### LINKED ISSUE #13024: `Translate::getModuleTranslation()` with locale returns string in wrong language if there is none in provided locale
In a cron task, we need to generate a french and an english version. As we can't rely on context language, we use `Translate::getModuleTranslation()` and pass the `Locale` as a parameter.

We found that in `getModuleTranslation` it will populate `$_MODULES` array with the translation. Problem is that it doesn't care of the locale and the context and so it will mix french and english in `$_MODULES`. As we don't have english translation in our translation file, we end up with french text even if we specify english in our locale if the current context in french and if `$_MODULES` is already used,

---

#### Describe the bug

Using $this->l() with locale provided uses last language instead of falling back to english/base string/empty string/throwing exception/whatever.
I added pl, de, fr to default english, and for example, when I have translation in polish, but doesn't have german and french, instead of english/whatever I get polish.

#### Expected behavior

I would like to get english/base string/empty string/exception thrown/something that would indicate that there is no such string in given language.

#### Steps to Reproduce

My module is working on certain products, so I thought I'll add them in code and wrote function that uses translations to insert needed products and categories to database. I run this function is run inside modules getContent() and dump its return for testing purposes. If some string doesn't have translation in given locale I get one from previous language.

```
    public function generateLanguageArrayForString($string, $module) : array
    {
        $langarr = [];
        $activeLangs = Language::getLanguages();
        foreach ($activeLangs as $lang) {
            $langarr[(int)$lang['id_lang']] = $this->l($string, $module, $lang['locale']);
        }
        return $langarr;
    }
```
For example, when ran on "Category name" which has only polish and english translations, I get this:
```
array(4) {
  [1]=>
  string(13) "Category name"
  [6]=>
  string(15) "Nazwa kategorii"
  [7]=>
  string(15) "Nazwa kategorii"
  [8]=>
  string(15) "Nazwa kategorii"
}
```

#### Additional information
When `global $_MODULES, $_MODULE, $_LANGADM;` inside _Translate.php_'s _getModuleTranslation()_
 is changed into `$_MODULES = array(); $_MODULE = array(); $_LANGADM = array();` I get translations only for languages that have them, otherwise I get base string (but when I add english translation that is different from base string, I don't get this one, which is pretty weird). But apart from weird issue with english, the _global_ must be there for a reason, and I don't want to change things outside of module's code for this to work.

### LINKED ISSUE #20495: Translate::getModuleTranslation() with locale returns string in wrong language if there is none in provided locale
#### Describe the bug

Using $this->l() with locale provided uses last language instead of falling back to english/base string/empty string/throwing exception/whatever.
I added pl, de, fr to default english, and for example, when I have translation in polish, but doesn't have german and french, instead of english/whatever I get polish.

#### Expected behavior

I would like to get english/base string/empty string/exception thrown/something that would indicate that there is no such string in given language.

#### Steps to Reproduce

My module is working on certain products, so I thought I'll add them in code and wrote function that uses translations to insert needed products and categories to database. I run this function is run inside modules getContent() and dump its return for testing purposes. If some string doesn't have translation in given locale I get one from previous language.

```
    public function generateLanguageArrayForString($string, $module) : array
    {
        $langarr = [];
        $activeLangs = Language::getLanguages();
        foreach ($activeLangs as $lang) {
            $langarr[(int)$lang['id_lang']] = $this->l($string, $module, $lang['locale']);
        }
        return $langarr;
    }
```
For example, when ran on "Category name" which has only polish and english translations, I get this:
```
array(4) {
  [1]=>
  string(13) "Category name"
  [6]=>
  string(15) "Nazwa kategorii"
  [7]=>
  string(15) "Nazwa kategorii"
  [8]=>
  string(15) "Nazwa kategorii"
}
```

#### Additional information
When `global $_MODULES, $_MODULE, $_LANGADM;` inside _Translate.php_'s _getModuleTranslation()_
 is changed into `$_MODULES = array(); $_MODULE = array(); $_LANGADM = array();` I get translations only for languages that have them, otherwise I get base string (but when I add english translation that is different from base string, I don't get this one, which is pretty weird). But apart from weird issue with english, the _global_ must be there for a reason, and I don't want to change things outside of module's code for this to work.

marekjedrzejewski:
Prevents getting translation in wrong language.

| Questions     | Answers
| Branch?       | develop 
| Description?  | When locale was specified when invoking getModuleTranslation one could get string from other language (not original string, or default lang) if there was no string in specified locale.
| Type?         | bug fix 
| Category?     | CO 
| BC breaks?    | maybe
| Deprecations? | no
| Fixed ticket? | Fixes #13024
| How to test?  | Inside issue #20495 there is simple module uploaded, which can be used for testing this change.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20537)


classes/Translate.php
```diff
@@ -194,6 +194,8 @@ public static function getModuleTranslation(
         }
 
         if (empty($iso)) {
+            // using global $_MODULES only when $locale is not set to prevent getting translation in wrong language
```
matks: Hi ! I am sorry I dont think this bugfix can be merged. It might solve your issue but it impacts the whole Translate class and I am quite sure there will be unwanted side-effects.

Can you explain the logic behind this PR ? We might find another way that solves your issue without introducing others.

Do you only set `$_MODULES` here because `getIsoByLocale` uses it ?

matks (changes requested):
See comment

classes/Translate.php
marekjedrzejewski: As I've written in the Issue mentioned in this PR, when I use `$this->l()` with locale specified, and there's no string for this locale, I get string in previous language that had translation for this string.
Because `$_MODULES` variable stores it without any mention of language the string is in, I can't know if that's proper translation, or something carried over from previous function call, so only thing that seemed to fix it for me was not using global `$_MODULES` when locale was provided. There might be better way to do this, and the way I did it might introduce bugs elsewhere, but I made this PR anyway, to get the ball rolling.

marekjedrzejewski: Other thing I thought about is adding another argument, for example `useCache`, that is true by default, but this seemed more like a workaround than real fix.

marekjedrzejewski:
My previous attempt was very naive, and not working after I added more strings to the equation. I decided to go with extending getModuleTranslation with useCache argument and it works with my test module. Doesn't seem to break anything, but previous one also didn't when testing really small case.

Here's updated version of module that can be used for testing:
[langtest.zip](https://github.com/PrestaShop/PrestaShop/files/5068931/langtest.zip)

marekjedrzejewski (commented):
Should be 'Add $useCache to getModuleTranslation'. Bash betrayed me :(

hibatallahAouadni:
@marekjedrzejewski why did you close your PR?

marekjedrzejewski:
It might be automatically closed after merge to my fork, but it might be me not being sure it doesn't break anything, without anyone more versed in presta than me to look at it. Or both. I don't remember why exactly it is closed, but I ended up using workaround, which I wrote about [here](https://github.com/PrestaShop/PrestaShop/issues/13024#issuecomment-694131637).

PR #20538, opened 2020-08-11
Fix order cart rule deletion

### LINKED ISSUE #20534: BO - Order page - Invalid cart rule provided.
#### Describe the bug

In the BO => Order details page => When we try to remove a discount, an exception is dispalyed
image

#### Expected behavior

The discount should be well removed

#### Steps to Reproduce

1. Create a cart rule (Any type)
2. Go to FO => add some product to the cart => apply the cart rule created previously
3. Proceed to checkout
4. Go to BO => Catalog => Cart rules page
5. Delete the cart rule just applied
6. Go to BO => Order details page => try to remove the discount
7. See error

**Screenshots**

https://drive.google.com/file/d/1O4XGPOy4u8p5GhO1wH32S-pzy5m7PO-N/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: 7.2

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | An OrderCartRule should be deletable even if it's related CartRule doesn't exist anymore.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20534
| How to test?  | 1. Create a cart rule (Any type)<br>2. Go to FO => add some product to the cart => apply the cart rule created previously<br>3. Proceed to checkout<br>4. Go to BO => Catalog => Cart rules page<br>5. Delete the cart rule just applied<br>6. Go to BO => Order details page => try to remove the discount<br>7. There should be no error

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20538)

### DIFF
--- a/src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
@@ -70,11 +70,6 @@ final class DeleteCartRuleFromOrderHandler extends AbstractOrderHandler implemen
             throw new OrderException('Invalid cart provided.');
         }
 
-        $cartRule = new CartRule($orderCartRule->id_cart_rule);
-        if (!Validate::isLoadedObject($cartRule)) {
-            throw new OrderException('Invalid cart rule provided.');
-        }
-
         // Delete Order Cart Rule and update Order
         $orderCartRule->softDelete();
         $cart->removeCartRule($orderCartRule->id_cart_rule);
--- a/src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
@@ -27,7 +27,6 @@
 namespace PrestaShop\PrestaShop\Adapter\Order\CommandHandler;
 
 use Cart;
-use CartRule;
 use OrderCartRule;
 use PrestaShop\PrestaShop\Adapter\Order\AbstractOrderHandler;
 use PrestaShop\PrestaShop\Adapter\Order\OrderAmountUpdater;


src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
```diff
@@ -70,11 +69,6 @@ public function handle(DeleteCartRuleFromOrderCommand $command)
             throw new OrderException('Invalid cart provided.');
         }
 
-        $cartRule = new CartRule($orderCartRule->id_cart_rule);
-        if (!Validate::isLoadedObject($cartRule)) {
-            throw new OrderException('Invalid cart rule provided.');
-        }
-
```
matthieu-rolland, line 77: I removed this check, because the deletion of an OrderCartRule should not be depending on the fact that its related CartRule still exists.

Another solution would have been to implement a soft delete mechanism on the cart rule deletion feature (when we delete a CartRule in the BO). 

This solution seems less invasive, and avoids having many remaining unused cart rules in the database, but I'm ready to discuss it :+1:

matks (approved):
It makes sense 👍

jolelievre (approved):

jolelievre:
Nice catch!

Robin-Fischer-PS:
@matthieu-rolland I still have an exception, but with a scenario a little bit far-fetched.

If I update the URL of the deletion of a voucher on an order, by changing the ID of the cart rule to a cart rule not applied to the order, for example : 
"http://localhost/ps177/admin-dev/index.php/sell/orders/orders/7/cart-rules/6/delete?_token=Ru-2T8UT1_bZghokRjBPnZNT81VKsqEuQua70Jy3Qw0" changed to "http://localhost/ps177/admin-dev/index.php/sell/orders/orders/7/cart-rules/666/delete?_token=Ru-2T8UT1_bZghokRjBPnZNT81VKsqEuQua70Jy3Qw0"

I have the same exception : 
pr20538

But, IMO, if a merchant try to modify the URL... It's his problem if he has an exception. What do you think ? (Maybe there is another case where we could have this situation without having to change the URL)

matthieu-rolland:
> @matthieu-rolland I still have an exception, but with a scenario a little bit far-fetched.
> 
> If I update the URL of the deletion of a voucher on an order, by changing the ID of the cart rule to a cart rule not applied to the order, for example :

Yes, if the URL is wrong, then the exception is doing its job: keeping someone from doing something unwanted with a wrong URL.

ngodefroy:
Thank you @matthieu-rolland I have tested the PR and did not found any bug.
Scenario 1

I created different types of cart rule in BO (Amount/percentage/free shipping/ gift)
Deleted them
No error encountered==> No error encountered, I get the "successful update" message.

Scenario 2

I went into the order detail to manually add a discount from this screen and repeated the steps above==> No error encountered, I get the "successful update" message.


1  Added cart rule
2  confirmation_delete

jolelievre:
Thank you @matthieu-rolland @ngodefroy and @Robin-Fischer-PS

PR #20299, opened 2020-07-23
Add UpdateProductSuppliers command

zuk3975:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop 
| Description?  | `UpdateProductSuppliersHandler` contains three handlers inside: `AddProductSupplierHandler`, `UpdateProductSupplierHandler` and `DeleteProductSupplierHandler`. It decides what to add/update/delete depending on provided suppliers list. Also added `ProductSupplierOptions` in ProductForEditing query result and covered basic usecases with behats (including some combination asserts - had to move combination creation(in tests) to `CombinationFactory` for reusability)
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Partially fixes #19683
| How to test?  | none

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20299)

PR #20198, opened 2020-07-16
Product supplier_reference is not updated when saving a product in the BO

PierreRambaud:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | Mark `wholesale_price` and `supplier_reference` as deprecated. These values are already register with `ProductSupplier` ObjectModel, it's useless to duplicate values. I also restore the registration of these fields when there is no combination even if values can be fetched by a request into `ProductSupplier` table.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | yes
| Fixed ticket? | Fixes #20094
| How to test?  | Follow ticket instruction. Make sure the issue #16353 is still resolved.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20198)

PR #20544, opened 2020-08-11
Add RemoveAllAssociatedProductSuppliersCommand

### LINKED ISSUE #19683: [Options tab] - Suppliers 5/5
As a merchant, I want to be able to manage the options of my products.

**Suppliers**: This part is displayed only if you have created and enabled at least one supplier. Indicating the product's supplier is not really important for customers, but it may turn out to be an essential part for the merchant’s internal management, not least when managing stock: the merchant simply needs to know who he bought the product from.
  * **Information message about suppliers**: “This interface allows you to specify the suppliers of the current product and its combinations if any. You can specify supplier references according to previously associated suppliers.” Click on read more link to display the following information: “When using the advanced stock management tool \(see Shop Parameters &gt; Products settings\), the values you define \(price, references\) will be used in supply orders.”
  * **Choose the suppliers associated with this product**: All enabled suppliers are displayed with a checkbox before their names. If it’s checked, the product is associated to the supplier and displayed on the supplier page in front-office.
  * **Default supplier**: When you check a supplier, a radio button allowing to choose the default supplier appears. When there is only one supplier checked, the default supplier radio button is automatically checked. The first supplier checked is the default supplier. When there are many suppliers checked, you can change the default supplier.
  * **Supplier reference\(s\)**: This part is displayed only if you have associated at least one supplier to the product. The supplier's section also features a table that enables you to set the precise reference and unit price/currency for each product combination, per supplier. If the product has more than one supplier, the table will display each supplier one after the other.
  * **Information message about supplier reference\(s\)**: “You can specify product reference\(s\) for each associated supplier. Click "Save" after changing selected suppliers to display the associated product references.”
  * **Table of suppliers**:
    * **The product name**: Display the product name
    * **Supplier reference**: Field to specify the product reference for each supplier. 
    * **Price \(tax excl.\)**: Field to specify the product price for each supplier. (check wording https://github.com/PrestaShop/PrestaShop/issues/18701)
    * **Currency**: Dropdown with all installed and enabled currencies. If there are many currencies, the default one is selected by default. Allow to specify the product currency for each supplier.

---

Edited by Louise B. on September 21st:

'Disassociate one supplier should delete only this supplier's data and not the other suppliers' data. Same for the deletion of a supplier in Catalog > Brands & Suppliers > Suppliers.' (https://github.com/PrestaShop/prestashop-specs/pull/132)

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | UpdateProductSuppliersCommand used to also delete all the suppliers. Now it has its own dedicated command. beforementioned command was renamed to SetProductSuppliersCommand. Also instead of injecting handlers, we start using Services. There is ProductSupplierPersister to handle the add,update,delete actions using legacy object models. As the SetProductSuppliers and RemoveAllProductSuppliers covers add/edit/delete cases I removed the redundant Add/Update/DeleteProductSupplierCommands|Handlers which is no more injected into other services
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | related #20299 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20544)


src/Adapter/Product/CommandHandler/DeleteProductSupplierHandler.php
```diff
@@ -87,9 +87,7 @@ private function refreshProductDefaultSupplier(ProductSupplier $productSupplier)
         }
 
         $product->id_supplier = 0;
-        $product->supplier_reference = '';
```
matks, line 90: Oh ? I think previous code makes sense. Why delete ? Did I miss something ?

src/Adapter/Product/CommandHandler/SetProductSuppliersHandler.php
```diff
@@ -89,65 +88,41 @@ public function __construct(
...
 
     /**
-     * @param UpdateProductSuppliersCommand $command
+     * @param SetProductSuppliersCommand $command
      *
      * @throws CannotUpdateProductException
-     * @throws ProductException
      */
-    private function handleDefaultSupplierUpdate(UpdateProductSuppliersCommand $command): void
+    private function assertDefaultSupplierIsOneOfProvidedSuppliers(SetProductSuppliersCommand $command): void
```
matks: 👍

matks (dismissed):
I approve however I got one question (see comment)

I think I'm missing something 😅

src/Adapter/Product/CommandHandler/DeleteProductSupplierHandler.php
zuk3975: because @PierreRambaud deprecated this field in some of his recent pr's

matks: Yes it was deprecated but until it is removed, we keep it functional.
See https://github.com/PrestaShop/PrestaShop/pull/20198/

We especially made sure to keep filling it with relevant data to preserve backward compatibility. Deprecating something does not mean "we break it" it means "it will be removed soon, but until it's removed we keep it working" (well ... in theory)

PierreRambaud: Yes unfortunately we need to keep it working until the next major version.

PierreRambaud (dismissed):

src/Adapter/Product/CommandHandler/SetProductSuppliersHandler.php
```diff
@@ -89,65 +88,41 @@ public function __construct(
...
-            $this->updateProductSuppliers($productId, $command->getProductSuppliers());
-        }
-
-        $deletedAllProductSuppliers = is_array($command->getProductSuppliers()) && empty($command->getProductSuppliers());
-
-        if (!$deletedAllProductSuppliers) {
-            $this->handleDefaultSupplierUpdate($command);
-        }
+        $this->setProductSuppliers($productId, $command->getProductSuppliers());
+        $this->updateProductDefaultSupplier($productId, $defaultSupplierIdValue);
```
PierreRambaud: Watch out, when you update supplier_id and supplier_reference you also need to update the `wholesale_price` in the product table :)

zuk3975: Done. Can you check if its alright? Following commits:
- https://github.com/PrestaShop/PrestaShop/pull/20544/commits/3f68de43bf0709790e361e3e6963869519e325cd
- https://github.com/PrestaShop/PrestaShop/pull/20544/commits/f94efb41e63dbd8ac01e95c4288ee8d5ca9b25d3
For the history, could you briefly explain how is the wholesale_price related to supplier?

PierreRambaud: When you select a default supplier without having a combination, it takes the supplier reference and the wholesale price.

More information here https://github.com/PrestaShop/PrestaShop/pull/20198#issuecomment-665150376

src/Adapter/Product/CommandHandler/DeleteProductSupplierHandler.php
jolelievre: But what happens when a supplier is removed then? Should we reste the data as if no default supplier is assigned? Or should we pick one among the remaining ones? (In which case the reset will still be required when there's no more suppliers associated)

jolelievre (commented):
Expect from the conversation regading the default values when a supplier is deleted, everything seems good But currently we need to define the default supplier behaviour, either:
- allow deleting any supplier, if it's the default one then:
   - a new default one is picked among the remaining ones
   - the product default supplier values are reset
- we forbid to remove the association on the default supplier

src/Adapter/Product/CommandHandler/DeleteProductSupplierHandler.php
zuk3975: I've reverted the changes. If default supplier is deleted then product->supplier_reference is reset to empty string and product->id_supplier is reset to 0.
Although I don't know if its ok. 
In case the default supplier is deleted, but there are still other suppliers then product will be left without a default supplier. Anyway, Im not sure if its worth to spend lots of time figuring this as it is  deprecated.

zuk3975:
> Expect from the conversation regading the default values when a supplier is deleted, everything seems good But currently we need to define the default supplier behaviour, either:
> 
> * allow deleting any supplier, if it's the default one then:
>   
>   * a new default one is picked among the remaining ones
>   * the product default supplier values are reset
> * we forbid to remove the association on the default supplier

Agree, both behavior looks valid. Could @MatShir provide us the answer?

matks:
@MatShir is off, @marionf is off

@LouiseBonnard will save us :)

LouiseBonnard:
> Expect from the conversation regading the default values when a supplier is deleted, everything seems good But currently we need to define the default supplier behaviour, either:
> 
> * allow deleting any supplier, if it's the default one then:
>   
>   * a new default one is picked among the remaining ones
>   * the product default supplier values are reset
> * we forbid to remove the association on the default supplier

In my opinion, it raises the question of having, necessarily, a default supplier when one or several suppliers are associated with a product:

1. if no, then deleting a supplier which is set as the default supplier for one or several products should not be an issue as any product can be associated with suppliers without having a default supplier - it is how it works right now

Capture d’écran 2020-09-04 à 17 07 05

2. if yes, then deleting a supplier should lead to the same kind of popup window when deleting a category, see screenshot below, and that requires an action from the user. It could be something like 'You are about the delete a supplier set as the default supplier for one or several items, what do you want to do?' with the following options: 'I want to pick a new default one among the remaining associated ones' and 'I want to reset all product values related to this default supplier'.

Capture d’écran 2020-09-04 à 17 08 11

What do you think?

zuk3975:
> 1. if no, then deleting a supplier which is set as the default supplier for one or several products should not be an issue as any product can be associated with suppliers without having a default supplier - it is how it works right now

I would say we leave the same behavior then. If it didn't raise any issues until now, there is no point bothering. (and it is the easiest solution from technical point of view)

LouiseBonnard:
> > 1. if no, then deleting a supplier which is set as the default supplier for one or several products should not be an issue as any product can be associated with suppliers without having a default supplier - it is how it works right now
> 
> I would say we leave the same behavior then. If it didn't raise any issues until now, there is no point bothering. (and it is the easiest solution from technical point of view)

I'm okay with that, thanks @zuk3975! @jolelievre, @matks, does this solution suit you as well?

jolelievre:
OK let's keep the current behaviour, just so we're sure we're on the same page:
- we can disassociate any supplier from the product (even the default one)
- if no default supplier is defined for the product its default values are reset

Is it ok?

src/Adapter/Product/CommandHandler/SetProductSuppliersHandler.php
```diff
@@ -71,83 +71,76 @@ final class UpdateProductSuppliersHandler extends AbstractProductHandler impleme
...
      * @param DeleteProductSupplierHandlerInterface $deleteProductSupplierHandler
+     * @param ProductUpdater $productUpdater
+     * @param ProductProvider $productProvider
      */
     public function __construct(
         AddProductSupplierHandlerInterface $addProductSupplierHandler,
         UpdateProductSupplierHandlerInterface $updateProductSupplierHandler,
-        DeleteProductSupplierHandlerInterface $deleteProductSupplierHandler
+        DeleteProductSupplierHandlerInterface $deleteProductSupplierHandler,
+        ProductUpdater $productUpdater,
```
Progi1984, line 95: It's a BC Break @PrestaShop/prestashop-core-developers ?

src/Adapter/Product/CommandHandler/SetProductSuppliersHandler.php
```diff
@@ -52,9 +52,9 @@
 use ProductSupplier as ProductSupplierEntity;
 
 /**
- * Handles @var UpdateProductSuppliersCommand using legacy object model
+ * Handles @var SetProductSuppliersCommand using legacy object model
  */
-final class UpdateProductSuppliersHandler extends AbstractProductHandler implements UpdateProductSuppliersHandlerInterface
+final class SetProductSuppliersHandler extends AbstractProductHandler implements SetProductSuppliersHandlerInterface
```
Progi1984: BC Break @PrestaShop/prestashop-core-developers

src/Core/Domain/Product/QueryResult/ProductSupplierOptions.php
```diff
@@ -50,17 +50,17 @@ class ProductSupplierOptions
 
     /**
      * @param int $defaultSupplierId
-     * @param string $defaultSupplierReference
      * @param ProductSupplierOption[] $optionsBySupplier
+     * @param string $defaultSupplierReference
```
Progi1984: Please revert changes in this file. No value in these.

src/Core/Domain/Product/Supplier/Command/SetProductSuppliersCommand.php
```diff
@@ -30,35 +30,42 @@
...
 use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
-use PrestaShop\PrestaShop\Core\Domain\Supplier\Exception\SupplierException;
 use PrestaShop\PrestaShop\Core\Domain\Supplier\ValueObject\SupplierId;
+use RuntimeException;
 
 /**
  * Updates product suppliers
  */
-class UpdateProductSuppliersCommand
+class SetProductSuppliersCommand
```
Progi1984, line 40: BC Break

src/Core/Domain/Product/Supplier/CommandHandler/SetProductSuppliersHandlerInterface.php
```diff
@@ -26,18 +26,18 @@
...
-use PrestaShop\PrestaShop\Core\Domain\Product\Supplier\QueryResult\ProductSupplier;
+use PrestaShop\PrestaShop\Core\Domain\Product\Supplier\Command\SetProductSuppliersCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\Supplier\ValueObject\ProductSupplierId;
 
 /**
- * Defines contract to handle @var UpdateProductSuppliersCommand
+ * Defines contract to handle @var SetProductSuppliersCommand
  */
-interface UpdateProductSuppliersHandlerInterface
+interface SetProductSuppliersHandlerInterface
```
Progi1984, line 35: BC Break

Progi1984 (changes requested):
Please define BC Break in a note in your Pull Request. Thanks

matks:
> Please define BC Break in a note in your Pull Request. Thanks

If I am not wrong, these files have not been yet released (they were introduced in `develop` very recently) so there is no BC Break. Until they are released with PrestaShop 1.7.8.0, we can change them as many times as we want in backward incompatible manner 😉

Progi1984:
> > Please define BC Break in a note in your Pull Request. Thanks
> 
> If I am not wrong, these files have not been yet released (they were introduced in `develop` very recently) so there is no BC Break. Until they are released with PrestaShop 1.7.8.0, we can change them as many times as we want in backward incompatible manner wink

:hankey: It's not enough clear in review. Sorry @zuk3975

Progi1984:
But we need rebase and fix the :red_circle: TravisCI.

zuk3975:
>  It's not enough clear in review. Sorry @zuk3975

yep matks is right. No problem :sun_with_face: 

Btw, Im implementing recent architecture changes (with the additional layer of services) in this PR.
Im adding a WIP label once again to make it clear

src/Adapter/AbstractObjectModelProvider.php
```diff
@@ -64,4 +64,27 @@ protected function getObjectModel(int $id, string $objectModelClass, string $exc
...
+
+    /**
+     * @param int $id
+     * @param string $objectModelName
+     * @param string $exceptionClass
+     * @param int $errorCode
+     *
+     * @throws CoreException
+     */
+    protected function assertObjectModelExists(int $id, string $objectModelName, string $exceptionClass, int $errorCode = 0): void
```
jolelievre:
```suggestion
    protected function assertObjectModelExists(int $id, string $objectTableName, string $exceptionClass, int $errorCode = 0): void
```

src/Adapter/Currency/CurrencyProvider.php
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @throws CoreException
+     * @throws CurrencyNotFoundException
+     */
+    public function get(CurrencyId $currencyId): Currency
+    {
+        /** @var Currency $currency */
+        $currency = $this->getObjectModel(
+            $currencyId->getValue(),
+            'currency',
```
jolelievre:
```suggestion
            Currency::class,
```

src/Adapter/Product/CommandHandler/DeleteProductSupplierHandler.php
```diff
@@ -29,68 +29,69 @@
...
-     * @throws ProductNotFoundException
+     * {@inheritdoc}
      */
-    private function refreshProductDefaultSupplier(ProductSupplier $productSupplier): void
+    public function handle(DeleteProductSupplierCommand $command): void
     {
-        $productId = (int) $productSupplier->id_product;
-        $product = $this->getProduct(new ProductId($productId));
+        $this->productSupplierDeleter->delete($command->getProductSupplierId());
+        $productSupplier = $this->productSupplierProvider->get($command->getProductSupplierId());
```
jolelievre: I don't understand you get the ProductSupplier after it's been removed, shouldn't it be the opposite Fetch before it's been removed or else you can't fetch it any more logically

src/Adapter/Product/CommandHandler/SetProductSuppliersHandler.php
```diff
@@ -0,0 +1,256 @@
...
+
+        return $this->getProductSupplierIds($productId);
+    }
+
+    /**
+     * @param SetProductSuppliersCommand $command
+     *
+     * @throws CannotUpdateProductException
+     */
+    private function assertDefaultSupplierIsOneOfProvidedSuppliers(SetProductSuppliersCommand $command): void
```
jolelievre: I wonder if this should be in the Handler or in a dedicated service so that the domain logic is managed by the handler any more We talked about dedicated services per domain like a `ProductSupplierUpdater` in this case which can handle product relationship and sub domain constraints like these ones

jolelievre: Or it could be managed by `ProductUpdater::updateProductDefaultSupplier` which might make more sense, it would just need to be called after all the update/creation/removals have been made (which is already the case)

src/Adapter/Product/ProductUpdater.php
```diff
@@ -61,7 +62,7 @@ public function __construct(
      *
      * @throws CoreException
      */
-    public function update(Product $product, array $propertiesToUpdate, int $errorCode): void
+    public function update(Product $product, array $propertiesToUpdate, int $errorCode = 0): void
```
jolelievre: Is the optional parameter necessary for Product, I though we had a dedicated error code for each sub domain so we might enforce that the error code should be provided in all cases?

src/Adapter/Product/ProductUpdater.php
```diff
@@ -91,15 +92,58 @@ public function refreshProductCustomizabilityProperties(Product $product): void
...
+     * @param Product $product
+     * @param int $defaultSupplierId
+     */
+    public function updateProductDefaultSupplier(Product $product, int $defaultSupplierId): void
+    {
+        if ($product->hasCombinations() || !$defaultSupplierId) {
+            $fieldsToUpdate['supplier_reference'] = '';
+            $fieldsToUpdate['wholesale_price'] = 0;
+        } elseif ($defaultSupplierId && !$product->hasCombinations()) {
+            $fieldsToUpdate['supplier_reference'] = ProductSupplier::getProductSupplierReference(
```
jolelievre: Should we check that the default supplier exists? And we could also check here that the default supplier is one the suppliers associated to this product?

src/Core/Domain/Product/Supplier/Command/SetProductSuppliersCommand.php
jolelievre: No this command was introduced in 178, so it's a simple refacto

src/Core/Domain/Product/Supplier/CommandHandler/SetProductSuppliersHandlerInterface.php
jolelievre: Nope refacto as well

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateProductSuppliersFeatureContext.php
```diff
@@ -256,7 +239,7 @@ public function assertDefaultSupplier(string $productReference, TableNode $table
             unset($data['default supplier reference']);
         }
 
-        Assert::assertEmpty($data, sprintf('Some provided product supplier fields haven\'t been asserted: %s', var_export($data, true)));
+        Assert::assertEmpty($data, sprintf('Some provided product supplier fields haven\'t been asserted: %s', implode(',', $data)));
```
jolelievre: The implode output doesn't allow to understand what is left since it doesn't display the array keys That's why I replaced it with a `var_export` in most Context classes, it helps debugging more efficiently

src/Adapter/Product/CommandHandler/DeleteProductSupplierHandler.php
jolelievre: This made me realize that this command is not behat tested by the way

zuk3975: indeed this command is full of crap :sweat:

matks: The crap of a developer is the treasure of another (they say)

src/Adapter/Currency/CurrencyProvider.php
zuk3975: not sure if its ok to provide the class namespace. As you mentioned above, It expects database table name not the class name :thinking: It will work now, but only by a coincidence. For example it would not work with CustomizationField I suppose. IMO we shouldn't provide such example just because it works now and looks better in this case, but won't work in another :thinking:

zuk3975: btw in ObjModelProvider::get() method I used the objModel::class, because it initializes that class, its not the same here

src/Adapter/Product/CommandHandler/SetProductSuppliersHandler.php
zuk3975: then we introduce a possibility for a non existing default supplier :thinking: (when updatable product suppliers list is missing current default supplier, we will update product suppliers, but default supplier will not be updated and will contain of supplier id which doesnt exist in product_supplier table anymore)).
To avoid this we could reset the default supplier to 0 if the default supplier update fails ? :thinking:

src/Adapter/AbstractObjectModelPersister.php
```diff
@@ -91,6 +91,31 @@ protected function updateObjectModel(ObjectModel $objectModel, string $exception
...
+     * @throws CoreException
+     */
+    protected function deleteObjectModel(ObjectModel $objectModel, bool $soft = false): bool
+    {
+        try {
+            if ($soft) {
+                return (bool) $objectModel->softDelete();
+            }
+
+            return (bool) $objectModel->delete();
```
Progi1984:
```suggestion
            return (bool) ($soft ? $objectModel->softDelete() : $objectModel->delete());
```

src/Adapter/Product/CommandHandler/AddProductSupplierHandler.php
```diff
@@ -28,63 +28,52 @@
...
-     *
-     * @throws CurrencyNotFoundException
-     * @throws ProductNotFoundException
-     * @throws SupplierNotFoundException
      */
     private function fillEntityWithCommandData(ProductSupplier $productSupplier, AddProductSupplierCommand $command): void
     {
         $productIdValue = $command->getProductId()->getValue();
         $supplierIdValue = $command->getSupplierId()->getValue();
         $currencyIdValue = $command->getCurrencyId()->getValue();
```
Progi1984: Remove these unuseful variables.

src/Adapter/Currency/CurrencyProvider.php
jolelievre: Mmm, I'm not sure the `assertObjectModelExists` uses the table name so it's logical to provide it via a string

But here you are using `getObjectModel` which purpose is to instanciate an object model based on the provided class, see:

https://github.com/PrestaShop/PrestaShop/blob/develop/src/Adapter/AbstractObjectModelProvider.php#L52

So it needs the **class** name not the **table** name, using `Currency::class` is totally appropriate And that's what you've done in all other providers by the way

src/Adapter/Product/ProductUpdater.php
```diff
@@ -91,15 +113,94 @@ public function refreshProductCustomizabilityProperties(Product $product): void
...
+        try {
+            $this->supplierProvider->assertSupplierExists($supplierId);
+            $this->productSupplierProvider->assertProductSupplierExists(new ProductSupplierId(
+                (int) ProductSupplier::getIdByProductAndSupplier(
+                    $product->id,
+                    0,
+                    $supplierId->getValue())
+            ));
+        } catch (SupplierNotFoundException | ProductSupplierNotFoundException $e) {
+            $this->resetDefaultSupplier($product);
```
jolelievre: I don't think this is a good idea, because it will erase the previous default supplier So if a product has a valid default supplier, and you intend to set one that doesn't exists It will not only fail the update but remove the previous default supplier
It should only checks if the supplier exists, if it doesn't it simply throws an error and nothing is changed

One thing we could do however is check if the supplier is correctly associated to the Product, and if not associate it when one wants to set it as default (this would avoid the problem you mentioned earlier: what happens if you try to use a default supplier that was not previously associated)

src/Adapter/Currency/CurrencyProvider.php
zuk3975: indeed you are right, I was sure that this is the `assertCurrencyExists` method, sorry.

src/Adapter/Product/CommandHandler/SetProductSuppliersHandler.php
```diff
@@ -0,0 +1,256 @@
...
+    private function updateProductSupplier(ProductSupplierDTO $productSupplierDTO): void
+    {
+        $productSupplier = $this->productSupplierProvider->get(new ProductSupplierId($productSupplierDTO->getProductSupplierId()));
+        $this->productSupplierPersister->update($productSupplier, $this->formatUpdatableProperties($productSupplierDTO));
+    }
+
+    /**
+     * @param ProductSupplierDTO $productSupplierDTO
+     *
+     * @return array<string, mixed>
```
matks: 👍

matks (commented):
Code seems OK for me. I'll let you and @jolelievre complete how to handle product supplier management 😉 then I'll approve the PR

Progi1984 (dismissed):

Progi1984 (dismissed):

matks (approved):

src/Adapter/Currency/CurrencyRepository.php
```diff
@@ -0,0 +1,72 @@
...
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Currency;
```
jolelievre, line 29:
```suggestion
namespace PrestaShop\PrestaShop\Adapter\Currency\Repository;
```
Even if there is only one in this case if some ObjectModel have a sub namespace and others don't it's gonna make it more complicated to search for the repos
This should be done in the next PR when you clean the namespaces though

src/Adapter/Product/ProductSupplierUpdater.php
```diff
@@ -0,0 +1,186 @@
...
+            ));
+        }
+
+        $product->supplier_reference = ProductSupplier::getProductSupplierReference($productIdValue, 0, $supplierIdValue);
+        $product->wholesale_price = ProductSupplier::getProductSupplierPrice($productIdValue, 0, $supplierIdValue);
+        $product->id_supplier = $supplierIdValue;
+
+        $this->productRepository->partialUpdate(
+            $product,
+            ['supplier_reference' => true, 'wholesale_price' => true, 'id_supplier' => true],
```
jolelievre, line 154: Sorry I didn't realize in previous PR that the partial update behaved this way now (although you warned me but I didn't get it ^^) So we don't use an array with the values to fill but only the list of updated fields
I'm fine with it but couldn't it be made even easier by using a simple array? Like this:
```suggestion
            ['supplier_reference', 'wholesale_price', 'id_supplier'],
```

Or is it because of the localized fields?

jolelievre (approved):

src/Adapter/Product/ProductSupplierUpdater.php
zuk3975: yes. Localized fields might introduce a problem :thinking: , but I think we can still get rid of those bools

PR #19964, opened 2020-06-26
Add UpdateProductCustomizationFields command

zuk3975:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Adds command to Update product customization fields. Covers basic usecases with behat scenarios.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #19681
| How to test?  | none

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19964)

PR #20549, opened 2020-08-12
Add RemoveAllCustomizationFieldsFromProductCommand And Introduce AbstractObjectModelPersister

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Refactors UpdateProductCustomizationFields. Rename it to SetProductCustomizationFields and add a separate command to handle RemoveAll scenario - RemoveAllCustomizationFieldsFromProductCommand
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | related #19964 
| How to test?  | behats must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20549)


matks (dismissed):

src/Adapter/Product/CommandHandler/RemoveAllCustomizationFieldsFromProductHandler.php
```diff
@@ -0,0 +1,67 @@
...
+        $product = $this->getProduct($command->getProductId());
+
+        $customizationFieldIds = array_map(function ($field) {
+            return (int) $field['id_customization_field'];
+        }, $product->getCustomizationFieldIds());
+
+        foreach ($customizationFieldIds as $customizationFieldId) {
+            $this->deleteCustomizationFieldHandler->handle(new DeleteCustomizationFieldCommand($customizationFieldId));
+        }
+    }
```
eternoendless: Are you sure this is the best way?

zuk3975: Im sure its the worst :sweat_smile: 
It preserves the object model hooks tho :man_shrugging:

matks: @eternoendless Correct. Doing it this way ensures that if a developer extends the `DeleteCustomizationFieldHandler` to enrich the behavior (exemple: decorate it to add a call to an API at the end in order to implement a synchronization between PrestaShop and an ERP) the behavior is kept consistent between the "single delete" and the "bulk delete"

This is **not** the optimized way to do it but I believe optimization should not happen at this level. PrestaShop should rather aim for consistency and robustness.

This however is my opinion 🤔   please tell me if I got it wrong

src/Core/Domain/Product/Customization/Command/SetProductCustomizationFieldsCommand.php
```diff
@@ -77,6 +78,14 @@ public function getCustomizationFields(): array
      */
     private function setCustomizationFields(array $customizationFields): void
     {
+        if (empty($customizationFields)) {
+            throw new RuntimeException(sprintf(
+                'Empty customization fields array provided in %s. To remove customization fields use %s',
+                self::class,
+                    RemoveAllCustomizationFieldsFromProductCommand::class
```
jolelievre:
```suggestion
                RemoveAllCustomizationFieldsFromProductCommand::class
```

src/Adapter/Product/CommandHandler/RemoveAllCustomizationFieldsFromProductHandler.php
jolelievre: I agree it feels heavy, but it's consistent with the `SetProductCustomizationFieldsHandler` which also uses the removal of customization fields with this handler At least if you need to hook, or override this behaviour it can be done in a single place instead of three

eternoendless: Removing one and removing all is not the same action, therefore to me it makes more sense to leave it to the developer to choose to enrich/replace one, the other, or both.

zuk3975: need a decision :sun_with_face:

eternoendless: I also stand my case regarding ERP sync. Imagine the sync in single delete is performed naively through a synchronous HTTP service call. If you do it this way for bulk delete then you'll have a lot of subsequent HTTP calls instead of a single bulk one. Two use cases, two implementations.

matks:
So, for this PR, we need to change how we have been doing Bulk actions. Rather than always reusing the same code logic we must think about optimization.

However, we're not going to rework all the BulkDeleteHandlers we have built 😛 .

I think we should apply the following strategy:
- there are some entities in PrestaShop that are simple and/or do not need advanced optimization (example: Employees or CMS Pages, I dont expect a shop to have more than a few hundred of them in the most exotic usecases)
- other entities however might require a different implementation of the SingleDeleteHandler and the BulkDeleteHandler, for performance purposes

## If the entity is simple and/or does not need advanced optimization

### For handlers already implemented

We change nothing.

### For new handlers

We are going to use the logic of ObjectModel. We are going to use `ObjectModel::deleteSelections($ids)`.

➡️ we might need to override it sometimes to handle a specific object model data structure
➡️ since all object models `extend` the generic class, the override is simple inheritance override

The SingleDeleteHandler uses the ObjectModel:
```
$customer = new Customer($myIdentifier);
$customer->delete();
```

The BulkDeleteHandler uses the ObjectModel by providing the array of IDs to delete.
```
$this->objectModel->deleteSelections($ids);
```

## If the entity is complex or there is a lot of them (example: products, customers, adresses...)

### For handlers already implemented

We might consider reworking them to fit with the need for optimization.

### For new handlers

SingleDeleteHandler uses the legacy ObjectModel logic. Example
```
$customer = new Customer($myIdentifier);
$customer->delete();
```

BulkDeleteHandler however, when relevant, should use a smarter way to delete the entity. 

If `deleteSelections` is written in a smart and optimized way we can use it.

If however an extra level of optimization is needed, or the "delete" operation is complex (we need to manage soft delete or clean/check the references) we might use a dedicated logic. Possibly a SQL query (ping @PierreRambaud you'll love that !). Depending on the usecase, it might be relevant to put this logic into the ObjectModel or in a dedicated service.

src/Adapter/Product/ProductUpdater.php
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Performs update of provided product fields
+ */
+class ProductUpdater
+{
+    /**
+     * @var array<string, bool|array<int, bool>>
+     */
+    private $fieldsToUpdate = [];
```
jolelievre: One thing that worries me is that this service can update any Product but only has one array to store the fields to update As long as only one instance of Product is used everything works fine But if multiple products are updated in the same process then the updated fields will start to merge with each others

So to avoid this and keep this service state-less maybe we should store the fields to update by product ID.

src/Adapter/Product/CommandHandler/RemoveAllCustomizationFieldsFromProductHandler.php
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(RemoveAllCustomizationFieldsFromProductCommand $command): void
+    {
+        $product = $this->productProvider->get($command->getProductId());
+
+        $customizationFieldIds = array_map(function ($field): CustomizationFieldId {
```
jolelievre:
```suggestion
        $customizationFieldIds = array_map(function (array $field): CustomizationFieldId {
```

Is it possible in anonymous functions?

src/Adapter/Product/CustomizationFieldDeleter.php
```diff
@@ -0,0 +1,141 @@
...
+     * @return void
+     *
+     * @throws CannotDeleteCustomizationFieldException
+     * @throws CustomizationFieldException
+     * @throws ProductException
+     * @throws ProductNotFoundException
+     */
+    private function performDeletion(CustomizationField $customizationField, int $errorCode): void
+    {
+        $product = $this->productProvider->get(new ProductId((int) $customizationField->id_product));
```
jolelievre: I'm afraid this is not very optimized, most of the time all customization fields have the same product, so we could fetch it only once.

Not of course nothing ensures that they do have the same product, but then maybe we could add a private `getProduct` method that would store the product by ID thus avoiding creating new instances of `ProductId` and `Product` for each deleted customization field

src/Adapter/Product/CustomizationFieldValidator.php
```diff
@@ -0,0 +1,72 @@
...
+
+use CustomizationField;
+use PrestaShop\PrestaShop\Core\Domain\Product\Customization\Exception\CustomizationException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Customization\Exception\CustomizationFieldConstraintException;
+use PrestaShopException;
+
+/**
+ * Validates CustomizationField field using legacy object model
+ */
+class CustomizationFieldValidator
```
jolelievre: I'm afraid most of the validator will all have the exact same code, except for the Exception class and the ObjectModel type, maybe we could have an abstract class `AbstractObjectModelFieldValidator` that implements `validateField` and `validatedLocalizedField`  (as protected functions).

The code would be the same except for an additional exception class, and ObjectModel as an input? It could be used by both `CustomizationFieldValidator` and `ProductFieldValidator` (for now but probably others later).

src/Adapter/Product/ProductFieldValidator.php
```diff
@@ -0,0 +1,72 @@
...
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductConstraintException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
+use PrestaShopException;
+use Product;
+
+/**
+ * Validates product field using legacy object model
+ */
+class ProductFieldValidator
```
jolelievre:
```suggestion
class ProductValidator
```

The `Field` suffix si not necessary, maybe you named it this way because of `CustomizationFieldValidator` except it is used for the `CustomizationField` object so it's not a suffix it's part of the Object name.

src/Adapter/Product/ProductFieldValidator.php
```diff
@@ -0,0 +1,72 @@
...
+     * Validates Product object model properties using legacy validation
+     *
+     * @param Product $product
+     * @param string $field
+     * @param int $errorCode
+     *
+     * @throws ProductConstraintException
+     * @throws ProductException
+     */
+    private function validate(Product $product, string $field, int $errorCode): void
```
jolelievre: I guess this will become a public function?

src/Adapter/Product/ProductProvider.php
```diff
@@ -0,0 +1,73 @@
...
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+use PrestaShopException;
+use Product;
+
+/**
+ * Provides requested product if it exists
+ */
+class ProductProvider
```
jolelievre: Same suggestion as for the validator, we could probably have an abstract class `AbstractObjectModelProvider` since the code is almost always the same, it would implement a method:

```php
protected function getObjectModel(int $id, class $objectModelClass, class $notFoundClass): ObjectModel
```

src/Adapter/Product/ProductUpdater.php
jolelievre: Alternatively we could add a `refresh` method to clean this array, but I'm not sure it would be handy And most likely error prone instead.

src/PrestaShopBundle/Resources/config/services/adapter/product.yml
```diff
@@ -230,3 +249,25 @@ services:
         tags:
             - name: tactician.handler
               command: PrestaShop\PrestaShop\Core\Domain\Product\Command\UpdateProductSeoCommand
+
```
jolelievre, line 262: I wonder if we should separate handlers services from other services in two yml files?

jolelievre (commented):
I validate the general idea of the PR, nice work!!
My comments are more about suggestions of improvement and small worries

src/Adapter/Product/CommandHandler/RemoveAllCustomizationFieldsFromProductHandler.php
zuk3975: hmm I think so

src/Adapter/Product/CustomizationFieldValidator.php
zuk3975: But then what exception to throw when PrestashopException is caught? (we don't want it to bubble up to the surface). And I think its not good to throw same NotFound exception in this case.

What if we just throw CoreException in this case?

src/Adapter/Product/ProductFieldValidator.php
zuk3975: I think I can make it ProductValidator and add methods like "validateProperty" and "validateLocalizedProperty" ?
As its in adapter, more methods can be added later depending on usecase :thinking: (something like validateAll which maybe would validate all the fields). Also using `Property` suffix will not interfare with Field(CustomizationField)

src/PrestaShopBundle/Resources/config/services/adapter/product.yml
zuk3975: Maybe. But its already separated by domain (product) and it's in adapter ns, not sure if its worth it. :man_shrugging:

jolelievre: Yeah probably, it was just a suggestion to avoid having too big yml files, but the important thing is that we can find the services definition easily anyway

src/Adapter/AbstractObjectModelValidator.php
```diff
@@ -0,0 +1,103 @@
...
+     * @throws CoreException
+     */
+    protected function validateObjectModelProperty(ObjectModel $objectModel, string $propertyName, string $exceptionClass, int $errorCode = 0): void
+    {
+        try {
+            if (true !== $objectModel->validateField($propertyName, $objectModel->{$propertyName})) {
+                throw new $exceptionClass(
+                    sprintf(
+                        'Invalid %s %s. Got "%s"',
+                        get_class($objectModel),
```
jolelievre, line 55: Smart!

src/Adapter/Product/ProductUpdater.php
```diff
@@ -0,0 +1,109 @@
...
+            );
+        } finally {
+            $this->fieldsToUpdate = [];
+        }
+    }
+
+    /**
+     * @param Product $product
+     */
+    public function refreshProductCustomizabilityFields(Product $product): void
```
jolelievre:
```suggestion
    public function refreshProductCustomizabilityProperties(Product $product): void
```

To be consistent with the abstract class functions

src/Adapter/AbstractObjectModelUpdater.php
```diff
@@ -0,0 +1,119 @@
...
+                );
+            }
+        } catch (PrestaShopException $e) {
+            throw new CoreException(
+                sprintf('Error occurred when trying to update %s #%d', get_class($objectModel), $objectModel->id),
+                0,
+                $e
+            );
+        } finally {
+            $objectModel->setFieldsToUpdate(null);
```
jolelievre: Carful the method expects an array, so either we modify it to accept null, or we add a dedicated function `cleanFieldsToUpdate`

src/Adapter/Product/CommandHandler/UpdateCustomizationFieldHandler.php
```diff
@@ -28,70 +28,88 @@
...
     }
 
     /**
      * @param UpdateCustomizationFieldCommand $command
-     * @param CustomizationField $field
+     *
+     * @return array
      */
-    private function fillEntityWithCommandData(UpdateCustomizationFieldCommand $command, CustomizationField $field): void
+    private function formatPropertiesForUpdate(UpdateCustomizationFieldCommand $command): array
```
jolelievre, line 99: I get that you didn't like the array formatting that much, but when you see the diff here (and most handlers will look this way I think) it's actually not that different from populating individual fields one by one

src/Adapter/Product/CustomizationFieldDeleter.php
```diff
@@ -0,0 +1,164 @@
...
+     * @throws ProductException
+     * @throws ProductNotFoundException
+     */
+    private function getProduct(int $productId): Product
+    {
+        if (isset($this->productsById[$productId])) {
+            return $this->productsById[$productId];
+        }
+
+        return $this->productProvider->get(new ProductId($productId));
```
jolelievre:
```suggestion
        if (!isset($this->productsById[$productId])) {
            $this->productsById[$productId] = $this->productProvider->get(new ProductId($productId));
        }

        return $this->productsById[$productId];
```

jolelievre: Or the cache field always remains empty ans is useless :)

src/Adapter/Product/CustomizationFieldValidator.php
```diff
@@ -0,0 +1,55 @@
...
+class CustomizationFieldValidator extends AbstractObjectModelValidator
+{
+    /**
+     * @param CustomizationField $customizationField
+     *
+     * @throws CoreException
+     */
+    public function validate(CustomizationField $customizationField): void
+    {
+        parent::validateObjectModelProperty($customizationField, 'type', CustomizationFieldConstraintException::class, CustomizationFieldConstraintException::INVALID_TYPE);
```
jolelievre:
```suggestion
        $this->validateObjectModelProperty($customizationField, 'type', CustomizationFieldConstraintException::class, CustomizationFieldConstraintException::INVALID_TYPE);
```

jolelievre: Maybe using `$this` will look better? Wdyt?

src/Adapter/Product/ProductProvider.php
```diff
@@ -0,0 +1,60 @@
...
+use PrestaShop\PrestaShop\Adapter\AbstractObjectModelProvider;
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductNotFoundException;
+use PrestaShop\PrestaShop\Core\Domain\Product\ValueObject\ProductId;
+use PrestaShop\PrestaShop\Core\Exception\CoreException;
+use Product;
+
+/**
+ * Provides existing Product
+ */
+class ProductProvider extends AbstractObjectModelProvider
```
jolelievre, line 40: I think Provider classes look pretty simple and clean this way 🤔

src/Adapter/Product/ProductUpdater.php
```diff
@@ -68,20 +84,22 @@ public function refreshProductCustomizabilityFields(Product $product): void
...
         ]);
     }
 
     /**
-     * @param array<string, bool|array<int, bool>> $fieldsToUpdate
+     * @param Product $customizationField
+     * @param array $propertiesToUpdate
      */
-    public function addPropertiesToUpdate(array $fieldsToUpdate): void
+    private function fillProperties(Product $customizationField, array $propertiesToUpdate): void
```
jolelievre:
```suggestion
    private function fillProperties(Product $product, array $propertiesToUpdate): void
```

src/Adapter/AbstractObjectModelUpdater.php
zuk3975: right it is typed. Ignored that as it was in ObjModel :face_with_head_bandage:

src/Adapter/Product/CustomizationFieldDeleter.php
zuk3975: sure, nice catch :smile:

classes/ObjectModel.php
```diff
@@ -2147,9 +2147,9 @@ public function getFieldByLang($field_name, $id_lang = null)
      *
      * @since 1.5.0.1
      *
-     * @param array $fields
+     * @param array<string, bool|array<int, bool>>|null $fields
      */
-    public function setFieldsToUpdate(array $fields)
+    public function setFieldsToUpdate(?array $fields)
```
PierreRambaud, line 2152: It's a BC. What is someone use `setFieldsToUpdate` without calling `addFieldsToUpdate`? 
You must keep the original behaviour and don't call `setFieldsToUpdate` when there is nothing to add
```suggestion
    public function setFieldsToUpdate(array $fields)
```

classes/ObjectModel.php
```diff
@@ -2196,4 +2196,18 @@ public function getHtmlFields()
...
+
+    /**
+     * @param array $fields
+     */
+    public function addFieldsToUpdate(array $fields): void
+    {
+        if (null === $this->update_fields) {
+            $this->update_fields = [];
+        }
+
```
PierreRambaud, line 2208:
```suggestion
```

classes/ObjectModel.php
zuk3975: as discussed in slack- its OK, because initial update_fields value is NULL.

classes/ObjectModel.php
zuk3975: update_fields is null|array type, so we need to initiate the array before trying to append any item to it.

classes/ObjectModel.php
jolelievre: Yep it's not a BC since the default value is `null` And it doesn't prevent from using the method with an array

src/Adapter/AbstractObjectModelPersister.php
```diff
@@ -0,0 +1,152 @@
...
+     * @param array $properties
+     */
+    protected function fillLocalizedProperty(ObjectModel $objectModel, string $propertyName, array $properties)
+    {
+        if (!array_key_exists($propertyName, $properties)) {
+            return;
+        }
+
+        if (!is_array($properties[$propertyName])) {
+            throw new RuntimeException(sprintf(
```
jolelievre:
```suggestion
            throw new CoreException(sprintf(
```

src/Adapter/Product/CustomizationFieldValidator.php
zuk3975: :+1: it doesn't change anything now, but to provide more flexibility lets use $this.

src/Adapter/Product/CommandHandler/AddCustomizationFieldHandler.php
```diff
@@ -28,56 +28,76 @@
 
 namespace PrestaShop\PrestaShop\Adapter\Product\CommandHandler;
 
-use CustomizationField as CustomizationFieldEntity;
+use CustomizationField;
```
jolelievre, line 31: Just a side note: I wonder if we should use something like `use CustomizationField as LegacyCustomizationField` everytime we use legacy objectmodels This would clarify which objects we are using especially since we have some VOs which names match with the legacy objects

src/Adapter/Product/CommandHandler/AddCustomizationFieldHandler.php
```diff
@@ -28,56 +28,76 @@
...
+     */
+    public function __construct(
+        ProductProvider $productProvider,
+        ProductUpdater $productUpdater,
+        CustomizationFieldPersister $customizationFieldPersister,
+        CustomizationFieldValidator $customizationFieldValidator
+    ) {
+        $this->productProvider = $productProvider;
+        $this->customizationFieldPersister = $customizationFieldPersister;
+        $this->customizationFieldValidator = $customizationFieldValidator;
```
jolelievre: You don't need to inject this one anymore, since it's already injected and managed by the persister

src/Adapter/Product/CommandHandler/SetProductCustomizationFieldsHandler.php
```diff
@@ -0,0 +1,190 @@
...
+     * @param ProductUpdater $productUpdater
+     */
+    public function __construct(
+        AddCustomizationFieldHandlerInterface $addCustomizationFieldHandler,
+        CustomizationFieldPersister $customizationFieldPersister,
+        CustomizationFieldDeleterInterface $customizationFieldDeleter,
+        CustomizationFieldProvider $customizationFieldProvider,
+        ProductUpdater $productUpdater
+    ) {
+        $this->addCustomizationFieldHandler = $addCustomizationFieldHandler;
```
jolelievre: This one is not needed any more, the persister manages it all It's great that we can remove these sub handlers dependencies

src/Adapter/Product/CommandHandler/SetProductCustomizationFieldsHandler.php
```diff
@@ -0,0 +1,190 @@
...
+
+use CustomizationField;
+use PrestaShop\PrestaShop\Adapter\Product\AbstractCustomizationFieldHandler;
+use PrestaShop\PrestaShop\Adapter\Product\CustomizationFieldPersister;
+use PrestaShop\PrestaShop\Adapter\Product\CustomizationFieldProvider;
+use PrestaShop\PrestaShop\Adapter\Product\ProductUpdater;
+use PrestaShop\PrestaShop\Core\Domain\Product\Customization\Command\SetProductCustomizationFieldsCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\Customization\CommandHandler\AddCustomizationFieldHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\Customization\CommandHandler\SetProductCustomizationFieldsHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\Customization\CustomizationField as CustomizationFieldDTO;
```
jolelievre, line 38: Nice touch! It clarifies reading This echoes with my other comment about `LegacyCustomizationField` it's even more valuable in cases like this one where DTO or VO are used with legacy ObjectModels

src/Adapter/Product/CommandHandler/SetProductCustomizationFieldsHandler.php
```diff
@@ -0,0 +1,190 @@
...
+        foreach ($command->getCustomizationFields() as $customizationFieldDTO) {
+            if ($customizationFieldDTO->getCustomizationFieldId()) {
+                $this->handleUpdate($customizationFieldDTO);
+            } else {
+                $this->handleCreation($command->getProductId(), $customizationFieldDTO);
+            }
+        }
+
+        $this->deleteCustomizationFields($deletableFieldIds);
+        $product = $this->getProduct($command->getProductId());
```
jolelievre: Maybe the product could be fetched only once (you also call `$this->getProduct` in `deleteCustomizationFields`) and passed as an argument

Or is it mandatory to get a "fresh" Product after the customization fields have been deleted? In that case maybe you should add a comment here `// Fetch fresh product` or something so that no one thinks of refactoring this part and mess with it

src/Adapter/Product/CommandHandler/SetProductCustomizationFieldsHandler.php
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * @param CustomizationFieldDTO $customizationFieldDTO
+     *
+     * @throws CoreException
+     */
+    private function handleUpdate(CustomizationFieldDTO $customizationFieldDTO): void
+    {
+        $fieldId = $customizationFieldDTO->getCustomizationFieldId();
```
jolelievre, line 146: It's too bad we need to instanciate the VO from the DTO just to pass it as an argument to the persister?

Maybe it's a dumb question but can't DTO use VOs? like `$customizationFieldDTO->getCustomizationFieldId()` could return a `CustomizationFieldId` directly?

src/Adapter/Product/ProductProvider.php
```diff
@@ -0,0 +1,60 @@
...
+     * @param ProductId $productId
+     *
+     * @return Product
+     *
+     * @throws CoreException
+     */
+    public function get(ProductId $productId): Product
+    {
+        /** @var Product $product */
+        $product = $this->getObjectModel(
```
jolelievre, line 52: Oh I just thought of something, it doesn't need to be managed here but the problem will probably arise at some point

The Product object has an additional `full` parameter in its constructor, and we will definitely need it sometimes (I don't know if it should be optional or always full just to be sure) To my knowledge it's the only ObjectModel with such additional constructor field but me might have to handle a special case for it at some point

Maybe allow optional additional parameters in the `AbstractObjectModelProvider` later or something.. we'll see ^^

src/Adapter/Product/CustomizationFieldPersister.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * @param CustomizationField $customizationField
+     * @param array<string, mixed> $propertiesToUpdate
+     * @param int $errorCode
+     *
+     * @throws CoreException
+     */
+    public function update(CustomizationField $customizationField, array $propertiesToUpdate, int $errorCode = 0): void
```
jolelievre, line 80: I don't believe we have special error codes for CustomizationField like we do for Product sub domains
But I guess you prefer to leave this parameter available just in case?

src/PrestaShopBundle/Resources/config/services/adapter/product.yml
```diff
@@ -167,15 +181,27 @@ services:
             - name: tactician.handler
               command: PrestaShop\PrestaShop\Core\Domain\Product\Customization\Query\GetProductCustomizationFields
 
-    prestashop.adapter.product.command_handler.update_product_customization_fields_handler:
-        class: PrestaShop\PrestaShop\Adapter\Product\CommandHandler\UpdateProductCustomizationFieldsHandler
+    prestashop.adapter.product.command_handler.set_product_customization_fields_handler:
+        class: PrestaShop\PrestaShop\Adapter\Product\CommandHandler\SetProductCustomizationFieldsHandler
         arguments:
             - '@prestashop.adapter.product.command_handler.add_customization_field_handler'
```
jolelievre: This one can be removed

src/PrestaShopBundle/Resources/config/services/adapter/product.yml
```diff
@@ -145,18 +145,32 @@ services:
 
     prestashop.adapter.product.command_handler.add_customization_field_handler:
         class: PrestaShop\PrestaShop\Adapter\Product\CommandHandler\AddCustomizationFieldHandler
+        arguments:
+            - '@prestashop.adapter.product.product_provider'
+            - '@prestashop.adapter.product.product_updater'
+            - '@prestashop.adapter.product.customization_field_persister'
+            - '@prestashop.adapter.product.customization_field_validator'
```
jolelievre: This one can be removed

jolelievre (commented):
A few comments, but only details

src/Adapter/Product/CommandHandler/AddCustomizationFieldHandler.php
zuk3975: in some other places I used CustomizationFieldDTO to mark it. (same idea, just reversed, instead of marking ObjectModel I marked the DTO)

src/Adapter/Product/CommandHandler/SetProductCustomizationFieldsHandler.php
zuk3975: I don't think so.
It suppose to just transfer data without doing anything else.
In other words it prevents us from using array, nothing more.

src/Adapter/Product/ProductProvider.php
zuk3975: btw it also accepts `$id_lang = null, $id_shop = null, Context $context = null`, but we haven't used any of this so far :smile:

src/Adapter/Product/CommandHandler/SetProductCustomizationFieldsHandler.php
jolelievre: Mmm, then maybe DTO could be used in place of the array in our Persister/Updater services 🤔 Assuming they would have a common interface, like allowing to detect if such field has been set or not for partial update (maybe based on Symfony ParameterBag or something)
This is just a thought of improvement not to be handled right now of course ^^

src/Adapter/Product/ProductProvider.php
jolelievre: No indeed, I think `$id_lang` must remain null, when it's set the localized fields are transformed into strings automatically We don't want this for edition
The shop might be useful though when we will deal with mutlishop

src/Adapter/Product/CommandHandler/SetProductCustomizationFieldsHandler.php
zuk3975: It would mean separate DTO for each ObjectModel and thats already confusing enough :smile:

jolelievre (approved):
Great work @zuk3975 !! This makes the handlers so much cleaner

matks (approved):

PR #19530, opened 2020-06-02
UpdateProductBasicInformationCommand

zuk3975:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | A command covered with some basic tests, which is responsible for updating product basic information (name and isVirtual). **EDIT:** After some discussions, we've decided to include descriptions in this command, so this pr also removes previously merged UpdateDescriptionCommand and instead handles descriptions itself.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #14763
| How to test?  | no UI to test until its integrated with form. Behat tests will cover basic usecase.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/19530)

PR #20554, opened 2020-08-12
Add manufacturer to UpdateProductBasicInformationCommand

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Allow updating product manufacturer through UpdateProductBasicInformationCommand. Adjust behats to support it. **NOTE** this might still worth a discussion if manufacturer belongs in basicInformationCommand. Apparently, we haven't yet agreed on this
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | related #19530
| How to test?  | Behats must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20554)

### DIFF
--- a/src/Core/Domain/Manufacturer/ValueObject/ManufacturerId.php
+++ b/src/Core/Domain/Manufacturer/ValueObject/ManufacturerId.php
@@ -66,8 +66,11 @@ class ManufacturerId
      */
     private function assertIsIntegerGreaterThanZero($value)
     {
-        if (!is_int($value) || 0 >= $value) {
-            throw new ManufacturerConstraintException(sprintf('Invalid manufacturer id "%s".', var_export($value, true)));
+        if (!is_int($value) || 0 > $value) {
+            throw new ManufacturerConstraintException(
+                sprintf('Invalid manufacturer id "%s".', var_export($value, true)),
+                ManufacturerConstraintException::INVALID_ID
+            );
         }
     }
 }
--- a/src/Core/Domain/Manufacturer/ValueObject/NoManufacturerId.php
+++ b/src/Core/Domain/Manufacturer/ValueObject/NoManufacturerId.php
@@ -24,53 +24,22 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  */
 
+declare(strict_types=1);
+
 namespace PrestaShop\PrestaShop\Core\Domain\Manufacturer\ValueObject;
 
-use PrestaShop\PrestaShop\Core\Domain\Manufacturer\Exception\ManufacturerConstraintException;
-
 /**
- * Provides manufacturer id
+ * This class is used to remove association with manufacturer
  */
-class ManufacturerId
+class NoManufacturerId extends ManufacturerId
 {
-    /**
-     * @var int
-     */
-    protected $id;
+    const NO_MANUFACTURER_ID = 0;
 
     /**
-     * @param int $id
-     *
-     * @throws ManufacturerConstraintException
+     * This class is used to remove association with manufacturer
      */
-    public function __construct($id)
+    public function __construct()
     {
-        $this->assertIsIntegerGreaterThanZero($id);
-        $this->id = $id;
-    }
-
-    /**
-     * @return int
-     */
-    public function getValue()
-    {
-        return $this->id;
-    }
-
-    /**
-     * Validates that the value is integer and is greater than zero
-     *
-     * @param $value
-     *
-     * @throws ManufacturerConstraintException
-     */
-    private function assertIsIntegerGreaterThanZero($value)
-    {
-        if (!is_int($value) || 0 >= $value) {
-            throw new ManufacturerConstraintException(
-                sprintf('Invalid manufacturer id "%s".', var_export($value, true)),
-                ManufacturerConstraintException::INVALID_ID
-            );
-        }
+        $this->id = static::NO_MANUFACTURER_ID;
     }
 }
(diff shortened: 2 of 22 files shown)


matks (commented):
@jolelievre I'm feeling a bit weird about putting manufacturer data into something with Basic in its name

manufacturer is 12 letters !
12 letters is not basic 😛 

it would "feel" better to put it into AddProductOptionsCommand
but I am not sure my "feelings" are a relevant indicator for whether a PR is good or not

wdyt ?

tests/Integration/Behaviour/Features/Scenario/Product/update_basic_information.feature
```diff
@@ -28,16 +33,23 @@ Feature: Update product basic information from Back Office (BO)
...
+    And manufacturer "studioDesign" should be assigned to product product1
     When I update product "product1" basic information with following values:
-      | is_virtual       | false        |
+      | is_virtual       | false         |
+      | manufacturer     | graphicCorner |
     Then product "product1" type should be standard
     And product "product1" localized "name" should be "en-US:photo of funny mug"
     And product "product1" localized "description" should be "en-US:nice mug"
     And product "product1" localized "description_short" should be "en-US:Just a nice mug"
+    And product "product1" type should be standard
```
jolelievre: These duplicate tests are useless (copied from above I guess)

jolelievre (dismissed):
@matks Wether it's in Basic or Options is fine for me 🤔
But since it's handled in the first tab with the other basic informations it doesn't shock me to have it in the same command 

Rather if we start having too many data in this command maybe we should rename it into something different from Basic

(Side note: the only thing bugging me is the duplicate lines in the tests but I don't consider this blocking, this should be cleaned later however)

matthieu-rolland (dismissed):

src/Core/Domain/Manufacturer/ValueObject/ManufacturerId.php
```diff
@@ -66,8 +66,11 @@ public function getValue()
      */
     private function assertIsIntegerGreaterThanZero($value)
     {
-        if (!is_int($value) || 0 >= $value) {
-            throw new ManufacturerConstraintException(sprintf('Invalid manufacturer id "%s".', var_export($value, true)));
+        if (!is_int($value) || 0 > $value) {
```
jolelievre: This one I'm not sure of, I allowed the 0 value so that we can un-assign the manufacturer from the Product
However `ManufacturerId` is also used in `ManufacturerController` and related commands (although there are no tests nor check for this 0 value elsewhere)

So either we accept this modification, or I though of an alternative to create a class that would extend `ManufacturerId` and would be dedicated to store this 0 value, we could name it `EmptyManufacturerId` for instance and it would be used by `UpdateProductOptionsCommand::setManufacturerId` when 0 value is used

jolelievre: @matks I would like your opinion on this one ^^

src/Core/Domain/Manufacturer/ValueObject/NoManufacturerId.php
```diff
@@ -0,0 +1,45 @@
...
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Manufacturer\ValueObject;
+
+/**
+ * This class is used to remove association with manufacturer
```
matks:
```suggestion
 * Manufacturer Identifier is a ValueObject which represents a valid identifier of a manufacturer.
 * It is being used in every class that must refer to a Manufacturer object or is linked to a Manufacturer object.
 *
 * However it is possible to decide to un-link the class from a Manufacturer. For example a product can be linked to a Manufacturer now and later this relationship is removed.
 *
 * This class NoManufacturerId carriers this intent, instead of using `null` which has another meaning.
 *
 * This picture might help understanding the situation: https://pbs.twimg.com/media/DusCOfyXcAA9_F7.jpg
```

matks (dismissed):
I approve the model. Only added a suggestion

jolelievre (approved):

matks (approved):

jolelievre:
Thank you @zuk3975 and @matks

src/Core/Domain/Manufacturer/ValueObject/NoManufacturerId.php
zuk3975: picture :smile:

src/Core/Domain/Manufacturer/ValueObject/NoManufacturerId.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+class NoManufacturerId extends ManufacturerId
+{
+    const NO_MANUFACTURER_ID = 0;
+
+    /**
+     * This class is used to remove association with manufacturer
+     */
+    public function __construct()
```
zuk3975, line 48: Is it surely a good idea ? It ignores parent constructor with all the constraints. It looks strange :thinking:

jolelievre: It is the very purpose of this class, to ignore the parent constraint (0 forbidden)

zuk3975: dirty ^^

PR #20567, opened 2020-08-13
Compile assets for 177x

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Just compiling assets as a preparation step for 177 beta 2
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20567)


atomiix (approved):
LGTM. Those are the only 2 files not up-to-date?

matks:
> LGTM. Those are the only 2 files not up-to-date?

Yes

PierreRambaud (changes requested):
Which version of node are you using, look like your compiled assets removed an escape 
image

PierreRambaud:
Ping @NeOMakinG 
There are differences if we build with node 12 (two files changes), and node 10 (nothing change). Can you have a check? I'm not sure all our 1.7.7.x npm dependencies are compatible with node 12 :thinking:

matks:
IMO
- if Node version is not right, the `make assets` command should return an error
- we should investigate what the differences are because contributors might build assets unaware of this issue

PierreRambaud:
> IMO
> 
> * if Node version is not right, the `make assets` command should return an error
> * we should investigate what the differences are because contributors might build assets unaware of this issue

The make assets is only available in the develop branch, and the node 12 version is compatible now :) 

We will investigate to be sure there are no changes between node 10 & 12

PR #20591, opened 2020-08-14
Introduce Javascript component loading through window.prestashop (part1)

JevgenijVisockij:
| Questions     | Answers
| Branch?       | develop
| Description?  | Makes javascript components available in modules without needing to implement whole webpack. 
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Category page should work exactly the same as before

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20591)

### DIFF
--- a/admin-dev/themes/new-theme/js/app/utils/init-components.js
+++ b/admin-dev/themes/new-theme/js/app/utils/init-components.js
@@ -33,7 +33,7 @@ const initPrestashopComponents = () => {
   window.prestashop.instance = {...window.prestashop.instance};
 
   window.prestashop.component = {
-    initPrestashopComponents(components) {
+    initComponents(components) {
       components.forEach((component) => {
         if (window.prestashop.component[component] === undefined) {
           console.error(`Failed to initialize PrestaShop component "${component}". This component doesn't exist.`);
--- a/admin-dev/themes/new-theme/js/app/utils/init-components.js
+++ b/admin-dev/themes/new-theme/js/app/utils/init-components.js
@@ -28,31 +28,36 @@ import TranslatableInput from '@js/components/translatable-input.js';
 import TinyMCEEditor from '@js/components/tinymce-editor.js';
 import TaggableField from '@js/components/taggable-field.js';
 
-const initComponents = () => {
+const initPrestashopComponents = () => {
   window.prestashop = {...window.prestashop};
+  window.prestashop.instance = {...window.prestashop.instance};
+
   window.prestashop.component = {
-    initComponents(components) {
-      window.prestashop.instance = {...window.prestashop.instance};
+    initPrestashopComponents(components) {
       components.forEach((component) => {
         if (window.prestashop.component[component] === undefined) {
           console.error(`Failed to initialize PrestaShop component "${component}". This component doesn't exist.`);
           return;
         }
 
-        if (window.prestashop.instance[component] !== undefined) {
+        const componentInstanceName = component.charAt(0).toLowerCase() + component.slice(1);
+
+        if (window.prestashop.instance[componentInstanceName] !== undefined) {
           console.warn(
             `Failed to initialize PrestaShop component "${component}". This component is already initialized.`,
           );
           return;
         }
 
-        window.prestashop.instance[component] = new window.prestashop.component[component]();
+
+        window.prestashop.instance[componentInstanceName] = new window.prestashop.component[component]();
       });
     },
+    // @todo: add all standard components in this list
     TranslatableField,
     TinyMCEEditor,
     TranslatableInput,
     TaggableField,
   };
 };
-export default initComponents;
+export default initPrestashopComponents;
--- a/admin-dev/themes/new-theme/js/app/utils/init-components.js
+++ b/admin-dev/themes/new-theme/js/app/utils/init-components.js
@@ -28,44 +28,36 @@ import TranslatableInput from '@js/components/translatable-input.js';
 import TinyMCEEditor from '@js/components/tinymce-editor.js';
 
 const initComponents = () => {
+  let translatableInput = null;
+  let tinyMCEEditor = null;
+  let translatableField = null;
   // Initialize the prestashop global object
   window.prestashop = {
     components: {
       translatableInput: {
-        enabled: false,
-        enable() {
-          // Already enabled
-          if (window.prestashop.components.translatableInput.enabled === true) {
-            return;
+        init() {
+          if (translatableInput === null) {
+            translatableInput = new TranslatableInput();
           }
-          window.prestashop.components.translatableInput.instance = new TranslatableInput();
-          window.prestashop.components.translatableInput.enabled = true;
+          return translatableInput;
         },
-        instance: null,
       },
       tinyMCEEditor: {
-        enabled: false,
-        enable() {
-          // Already enabled
-          if (window.prestashop.components.tinyMCEEditor.enabled === true) {
-            return;
+        init() {
+          if (tinyMCEEditor === null) {
+            tinyMCEEditor = new TinyMCEEditor();
           }
-          window.prestashop.components.tinyMCEEditor.instance = new TinyMCEEditor();
-          window.prestashop.components.tinyMCEEditor.enabled = true;
+          return tinyMCEEditor;
         },
-        instance: null,
+
       },
       translatableField: {
-        enabled: false,
-        enable() {
-          // Already enabled
-          if (window.prestashop.components.translatableField.enabled === true) {
-            return;
+        init() {
+          if (translatableField === null) {
+            translatableField = new TranslatableField();
           }
-          window.prestashop.components.translatableField.instance = new TranslatableField();
-          window.prestashop.components.translatableField.enabled = true;
+          return translatableField;
         },
-        instance: null,
       },
     },
   };
--- a/admin-dev/themes/new-theme/js/app/utils/init-components.js
+++ b/admin-dev/themes/new-theme/js/app/utils/init-components.js
@@ -29,17 +29,17 @@ const initComponents = () => {
   // Initialize the prestashop global object
   window.prestashop = {
     components: {
-      enableTranslatableInput,
+      EnableTranslatableInput,
     },
   };
 };
 
 export default initComponents;
 
-const enableTranslatableInput = function () {
-  if (window.prestashop.components.enableTranslatableInput.enabled) {
+const EnableTranslatableInput = function () {
+  if (window.prestashop.components.EnableTranslatableInput.enabled) {
     return;
   }
   new TranslatableInput();
-  window.prestashop.components.enableTranslatableInput.enabled = true;
+  window.prestashop.components.EnableTranslatableInput.enabled = true;
 };
--- a/admin-dev/themes/new-theme/js/app/utils/init-components.js
+++ b/admin-dev/themes/new-theme/js/app/utils/init-components.js
@@ -39,7 +39,7 @@ const initComponents = () => {
         window.prestashop.instance = {};
       }
       components.forEach((component) => {
-        if (typeof window.prestashop.instance[component] === 'undefined') {
+        if (window.prestashop.instance[component] === undefined) {
           window.prestashop.instance[component] = new window.prestashop.component[component]();
         }
       });
--- a/admin-dev/themes/new-theme/js/app/utils/init-components.js
+++ b/admin-dev/themes/new-theme/js/app/utils/init-components.js
@@ -35,7 +35,7 @@ const initComponents = () => {
 
   window.prestashop.component = {
     initComponents(components) {
-      if (typeof window.prestashop.instance === 'undefined') {
+      if (window.prestashop.instance === undefined) {
         window.prestashop.instance = {};
       }
       components.forEach((component) => {
--- a/admin-dev/themes/new-theme/js/app/utils/init-components.js
+++ b/admin-dev/themes/new-theme/js/app/utils/init-components.js
@@ -26,11 +26,17 @@
 import TranslatableField from '@js/components/translatable-field.js';
 import TranslatableInput from '@js/components/translatable-input.js';
 import TinyMCEEditor from '@js/components/tinymce-editor.js';
+import TaggableField from '@js/components/taggable-field.js';
 
 const initComponents = () => {
   let translatableInput = null;
   let tinyMCEEditor = null;
   let translatableField = null;
+  let taggableField = null;
+
+  if (typeof window.prestashop === 'undefined') {
+    window.prestashop = {};
+  }
   // Initialize the prestashop global object
   window.prestashop.components = {
     translatableInput: {
@@ -40,6 +46,9 @@ const initComponents = () => {
         }
         return translatableInput;
       },
+      component(options) {
+        return new TranslatableInput(options);
+      },
     },
     tinyMCEEditor: {
       init() {
@@ -47,6 +56,9 @@ const initComponents = () => {
           tinyMCEEditor = new TinyMCEEditor();
         }
         return tinyMCEEditor;
+      },
+      component(options) {
+        return new TinyMCEEditor(options);
       },
 
     },
@@ -56,6 +68,20 @@ const initComponents = () => {
           translatableField = new TranslatableField();
         }
         return translatableField;
+      },
+      component(options) {
+        return new TranslatableField(options);
+      },
+    },
+    taggableField: {
+      init() {
+        if (taggableField === null) {
+          taggableField = new TaggableField();
+        }
+        return taggableField;
+      },
+      component(options) {
+        return new TaggableField(options);
       },
     },
   };
--- a/admin-dev/themes/new-theme/js/app/utils/init-components.js
+++ b/admin-dev/themes/new-theme/js/app/utils/init-components.js
@@ -29,17 +29,17 @@ const initComponents = () => {
   // Initialize the prestashop global object
   window.prestashop = {
     components: {
-      EnableTranslatableInput,
+      enableTranslatableInput,
     },
   };
 };
 
 export default initComponents;
 
-const EnableTranslatableInput = function () {
-  if (window.prestashop.components.EnableTranslatableInput.enabled) {
+const enableTranslatableInput = function () {
+  if (window.prestashop.components.enableTranslatableInput.enabled) {
     return;
   }
   new TranslatableInput();
-  window.prestashop.components.EnableTranslatableInput.enabled = true;
+  window.prestashop.components.enableTranslatableInput.enabled = true;
 };
--- a/admin-dev/themes/new-theme/js/app/utils/init-components.js
+++ b/admin-dev/themes/new-theme/js/app/utils/init-components.js
@@ -43,27 +43,27 @@ const initComponents = () => {
         },
         instance: null,
       },
-      TinyMCEEditor: {
+      tinyMCEEditor: {
         enabled: false,
         enable() {
           // Already enabled
-          if (window.prestashop.components.TinyMCEEditor.enabled === true) {
+          if (window.prestashop.components.tinyMCEEditor.enabled === true) {
             return;
           }
-          window.prestashop.components.TinyMCEEditor.instance = new TinyMCEEditor();
-          window.prestashop.components.TinyMCEEditor.enabled = true;
+          window.prestashop.components.tinyMCEEditor.instance = new TinyMCEEditor();
+          window.prestashop.components.tinyMCEEditor.enabled = true;
         },
         instance: null,
       },
-      TranslatableField: {
+      translatableField: {
         enabled: false,
         enable() {
           // Already enabled
-          if (window.prestashop.TranslatableField.enabled === true) {
+          if (window.prestashop.components.translatableField.enabled === true) {
             return;
           }
-          window.prestashop.TranslatableField.instance = new TranslatableField();
-          window.prestashop.TranslatableField.enabled = true;
+          window.prestashop.components.translatableField.instance = new TranslatableField();
+          window.prestashop.components.translatableField.enabled = true;
         },
         instance: null,
       },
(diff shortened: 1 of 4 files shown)


matks (dismissed):
Approving the WIP

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,45 @@
...
+
+export default initComponents;
+
+const enableTranslatableInput = function () {
+  if (window.prestashop.components.enableTranslatableInput.enabled) {
+    return;
+  }
+  new TranslatableInput();
+  window.prestashop.components.enableTranslatableInput.enabled = true;
+};
```
PierreRambaud:
```suggestion
const initComponents = () => {
  // Initialize the prestashop global object
  window.prestashop = {
    components: {
      translatableInput: {
        enabled: false,
        enable() {
          // Already enabled
          if (window.prestashop.translatableInput.enabled === true) {
            return;
          }

          window.prestashop.translatableInput.instance = new TranslatableInput();
          window.prestashop.translatableInput.enabled = true;
        },
        instance: null,
      }
      enableTranslatableInput,
    },
  };
};

export default initComponents;
```

Not sure your code will work if you call `window.prestashop.components.enableTranslatableInput()` twice, you change the object definition when calling this method

admin-dev/themes/new-theme/js/components/translatable-input.js
```diff
@@ -40,11 +40,7 @@ class TranslatableInput {
     this.localeButtonSelector = opts.localeButtonSelector || '.js-locale-btn';
     this.localeInputSelector = opts.localeInputSelector || '.js-locale-input';
 
-    $('body').on(
-      'click',
-      this.localeItemSelector,
-      this.toggleLanguage.bind(this),
-    );
+    $('body').on('click', this.localeItemSelector, this.toggleLanguage.bind(this));
```
PierreRambaud: No reason to change the indentation :thinking:

admin-dev/themes/new-theme/js/app/utils/init-components.js
PierreRambaud: You also forget to return the instance of TranslatableInput component :thinking:

JevgenijVisockij: Your suggestion looks great, but I don't quite understand one thing: why "enableTranslatableInput," is still there if the function was deleted and enabling works diffrently now?

admin-dev/themes/new-theme/js/components/translatable-input.js
NeOMakinG: I think it's an autofix from his IDE, as there are some ESLint errors

JevgenijVisockij: Yeah I think it is, will fix it in next commit.

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,73 @@
...
+
+const initComponents = () => {
+  // Initialize the prestashop global object
+  window.prestashop = {
+    components: {
+      translatableInput: {
+        enabled: false,
+        enable() {
+          // Already enabled
+          if (window.prestashop.components.translatableInput.enabled === true) {
```
jolelievre: I'm not sure you need this `enabled` field you could directly check if `instance` is `null`

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,73 @@
...
+            return;
+          }
+          window.prestashop.components.translatableInput.instance = new TranslatableInput();
+          window.prestashop.components.translatableInput.enabled = true;
+        },
+        instance: null,
```
jolelievre: What troubles me with this way of creating the component is that all the internal variables are directly accessible via `window.prestashop.components.translatableInput.enabled` and could be messed with

Ideally they should be private, but javascript doesn't manage this but at least if we had some modules that only expose the required api it would make it more reliable

admin-dev/themes/new-theme/js/pages/category/index.js
```diff
@@ -75,9 +72,10 @@ $(() => {
   const showcaseCard = new ShowcaseCard('categoriesShowcaseCard');
   showcaseCard.addExtension(new ShowcaseCardCloseExtension());
 
-  new TranslatableField();
-  new TinyMCEEditor();
-  const translatorInput = new TranslatableInput();
+  window.prestashop.components.translatableField.enable();
+  window.prestashop.components.tinyMCEEditor.enable();
+  window.prestashop.components.translatableInput.enable();
```
jolelievre: This double call seems redundant, it would be nice to enable an get the component in one call:
```
const translatorInput = window.prestashop.components.translatableInput.enable();
```

But the function name would not we adapted any more 🤔  maybe we should use `getInstance` instead like in a singleton

admin-dev/themes/new-theme/js/app/utils/init-components.js
matks: > and could be messed with

What are you afraid of ? 🤔  modules messing with each other, or deeper, security issues ?

admin-dev/themes/new-theme/js/pages/category/index.js
JevgenijVisockij: I like the idea to change it to getInstance. The only thing I am not sure about is what if I don't need the instance, I only need to init it. just leave window.prestashop.components.translatableInput.getInstance() without assigning  anything to it? 

It should work but I wonder if this is considered a clean implementation.

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,63 @@
...
+  // Initialize the prestashop global object
+  window.prestashop.components = {
+    translatableInput: {
+      init() {
+        if (translatableInput === null) {
+          translatableInput = new TranslatableInput();
+        }
+        return translatableInput;
+      },
+    },
```
jolelievre:
```suggestion
        component: TranslatableInput
    },
```

We discussed about letting the choice to the developer to instanciate the component himself, so maybe we should let the class itself accessible so that we could do:

```js
    new window.prestashop.components.translatableInput.component({localeItemSelector: '.my-selector'});
```

I'm not sure if `component` is the right naming, it's just a suggestion

PierreRambaud: Agreed, it's better to return only the component instead of instantiation.

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,89 @@
...
+import TranslatableField from '@js/components/translatable-field.js';
+import TranslatableInput from '@js/components/translatable-input.js';
+import TinyMCEEditor from '@js/components/tinymce-editor.js';
+import TaggableField from '@js/components/taggable-field.js';
+
+const initComponents = () => {
+  let translatableInput = null;
+  let tinyMCEEditor = null;
+  let translatableField = null;
+  let taggableField = null;
```
PierreRambaud: I'm really not in favor of having a bunch of variables here. In a few days, we will have 200 variables :/
Why not doing this:
```js
  window.prestashop.components = {
    translatableInput: {
      instance: null, 
      init: function initialize() {
        if (this.instance === null) {
          this.instance = new TranslatableInput();
        }
      },
      component(options) {
        return new TranslatableInput(options);
      },
    },
```
And play with the instance. An `init` method must initialize something, not returning an instance :thinking:

admin-dev/themes/new-theme/js/app/utils/init-components.js
PierreRambaud: It's for the eslint, you can't have anonymous function without name :sweat_smile:

PierreRambaud: It was a residue from the git change :sweat_smile: 
The correct implementation: https://github.com/PrestaShop/PrestaShop/pull/20591#discussion_r475110356

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,53 @@
...
+  if (typeof window.prestashop === 'undefined') {
+    window.prestashop = {};
+  }
+
+  window.prestashop.component = {
+    initComponents(components) {
+      if (typeof window.prestashop.instance === 'undefined') {
+        window.prestashop.instance = {};
+      }
+      $.each(components, (key, component) => {
```
PierreRambaud:
```suggestion
      components.forEach((component) => {
```
jQuery is not needed

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,53 @@
...
+import TaggableField from '@js/components/taggable-field.js';
+
+const initComponents = () => {
+  if (typeof window.prestashop === 'undefined') {
+    window.prestashop = {};
+  }
+
+  window.prestashop.component = {
+    initComponents(components) {
+      if (typeof window.prestashop.instance === 'undefined') {
```
PierreRambaud:
```suggestion
      if (window.prestashop.instance === undefined) {
```

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,53 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+import TranslatableField from '@js/components/translatable-field.js';
+import TranslatableInput from '@js/components/translatable-input.js';
+import TinyMCEEditor from '@js/components/tinymce-editor.js';
+import TaggableField from '@js/components/taggable-field.js';
+
+const initComponents = () => {
+  if (typeof window.prestashop === 'undefined') {
```
PierreRambaud:
```suggestion
  if (window.prestashop === undefined) {
```

JevgenijVisockij:
No idea why PHP CS fixer has failed. It says it failed in Tree.php, I am not modifying any PHP files.

admin-dev/themes/new-theme/js/app/utils/init-components.js
JevgenijVisockij: Thanks! Commited those changes.

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  window.prestashop.component = {
+    initComponents(components) {
+      if (window.prestashop.instance === undefined) {
+        window.prestashop.instance = {};
+      }
+      components.forEach((component) => {
+        if (window.prestashop.instance[component] === undefined) {
+          window.prestashop.instance[component] = new window.prestashop.component[component]();
```
jolelievre: You should probably add a check to see if `window.prestashop.component[component]` exists

JevgenijVisockij: Good point!

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,54 @@
...
+import TranslatableInput from '@js/components/translatable-input.js';
+import TinyMCEEditor from '@js/components/tinymce-editor.js';
+import TaggableField from '@js/components/taggable-field.js';
+
+const initComponents = () => {
+  if (window.prestashop === undefined) {
+    window.prestashop = {};
+  }
+
+  window.prestashop.component = {
```
jolelievre, line 35: Should we perform the same kind of check as the prestashop object here? IE merge this with a potential already existing `window.prestashop.component` object, it would happen if a module wants to add its own component, but added it before calling the `initComponents` function

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,54 @@
...
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+import TranslatableField from '@js/components/translatable-field.js';
+import TranslatableInput from '@js/components/translatable-input.js';
+import TinyMCEEditor from '@js/components/tinymce-editor.js';
+import TaggableField from '@js/components/taggable-field.js';
+
+const initComponents = () => {
```
jolelievre:
```suggestion
const initPrestashopComponents = () => {
```

I suggest a different naming than the `window.prestashop.component.initComponents` function just to make sure there is no confusion

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,54 @@
...
+
+  window.prestashop.component = {
+    initComponents(components) {
+      if (window.prestashop.instance === undefined) {
+        window.prestashop.instance = {};
+      }
+      components.forEach((component) => {
+        if (window.prestashop.instance[component] === undefined
+          && window.prestashop.component[component] !== undefined) {
+          window.prestashop.instance[component] = new window.prestashop.component[component]();
```
jolelievre: Maybe we should add some `console.error` or `console.warning` when the component is not managed?

jolelievre (commented):
Just some suggestions of improvement to discuss
But other than that the PR seems legit for me, it could be merged as is and improved later

jolelievre (dismissed):
Since my suggestion are not blocking I approve the PR ^^
Still curious for opinions regarding my comments though

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,54 @@
...
+
+import TranslatableField from '@js/components/translatable-field.js';
+import TranslatableInput from '@js/components/translatable-input.js';
+import TinyMCEEditor from '@js/components/tinymce-editor.js';
+import TaggableField from '@js/components/taggable-field.js';
+
+const initComponents = () => {
+  if (window.prestashop === undefined) {
+    window.prestashop = {};
+  }
```
PierreRambaud:
```suggestion
  window.prestashop = Object.assign({}, window.prestashop};
```

JevgenijVisockij: Implemented this slightly differently (  window.prestashop = {...window.prestashop}; ) thanks for suggestion!

admin-dev/themes/new-theme/js/app/utils/init-components.js
JevgenijVisockij: Great suggestion! I think error makes more sense, at least for me I constantly get a couple of warnings so I think a lot of people wouldn't notice it, while the error is clearly noticeable.

admin-dev/themes/new-theme/js/app/utils/init-components.js
JevgenijVisockij: Good point! I implemented 
      window.prestashop.instance = {...window.prestashop.instance};
I think it should cover this.

admin-dev/themes/new-theme/js/app/utils/init-components.js
JevgenijVisockij: Makes sense to me, before I implement this I want to ask does ADR need to be modified as well? Or ADR is more like first draft and implementation can differ?

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,58 @@
...
+import TranslatableField from '@js/components/translatable-field.js';
+import TranslatableInput from '@js/components/translatable-input.js';
+import TinyMCEEditor from '@js/components/tinymce-editor.js';
+import TaggableField from '@js/components/taggable-field.js';
+
+const initComponents = () => {
+  window.prestashop = {...window.prestashop};
+  window.prestashop.component = {
+    initComponents(components) {
+      window.prestashop.instance = {...window.prestashop.instance};
```
PierreRambaud: You need to initialize it outside this method. The instance part doesn't need to be initialized each time you initialize a component.

matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/20591/commits/2ba96e3f7bab064ce76c9c6a7fa11b99b1786fb4

admin-dev/themes/new-theme/js/app/utils/init-components.js
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/20591/commits/2ba96e3f7bab064ce76c9c6a7fa11b99b1786fb4

matks:
I pushed a commit. Inside this commit
- I performed the renaming required by code reviews
- I implemented the camelCase component conversion: so while we have a `window.prestashop.component.TranslatableField` component, we have a `window.prestashop.instance.translatableField` instance 

**I suggest we merge this as a Version 0.1 of this new system.**

In v0.2 we'll have to address:
- Advanced protection as suggested by @jolelievre : how to make sure module developer cannot mess with already instantiated instances?
- Register all available components (currently there's only TranslatableField, TinyMCEEditor, TranslatableInput and TaggableField), maybe provide a versioning for them to help people know which ones are available in 1.7.8 or 1.7.9.
- Decide whether we perform a "default initialization". With current implementation, module developer must explicitely ask which Components he wants to use:
```
$(document).ready(function () {
  window.prestashop.component.initComponents(
    [
      'TranslatableField',
      'TinyMCEEditor',
      'TranslatableInput',
    ],
  );
});
```

Should we provide a default initialization function which provide instances for all components ?

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,63 @@
...
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+import TranslatableField from '@js/components/translatable-field.js';
+import TranslatableInput from '@js/components/translatable-input.js';
+import TinyMCEEditor from '@js/components/tinymce-editor.js';
+import TaggableField from '@js/components/taggable-field.js';
+
+const initPrestashopComponents = () => {
```
jolelievre, line 31: Woups I believe you missed the point of my previous comment here ^^
The idea was to rename this method name ONLY, and keep `window.prestashop.component.initComponents` (L36)

The purpose is to avoid confusing this method that initialize the prestashop component APIs, from the public API method itself.

Thus `initPrestashopComponents` (internal API/function) would only be called once in `admin-dev/themes/new-theme/js/theme.js` but `window.prestashop.component.initComponents` (public API/function) could be called in many modules, and we avoid people confusing them.

admin-dev/themes/new-theme/js/app/utils/init-components.js
```diff
@@ -0,0 +1,63 @@
...
+  window.prestashop = {...window.prestashop};
+  window.prestashop.instance = {...window.prestashop.instance};
+
+  window.prestashop.component = {
+    initPrestashopComponents(components) {
```
jolelievre:
```suggestion
    initComponents(components) {
```

This way the "public" API respects the ADR naming https://github.com/PrestaShop/ADR/blob/master/0009-expose-js-components-using-window-variable.md#decision

jolelievre (changes requested):
Sorry my previous comment was probably confusing 😅  But only one of the two methods need to be renamed, the purpose is not to confuse them any more

jolelievre:
> * I performed the renaming required by code reviews
> * I implemented the camelCase component conversion: so while we have a `window.prestashop.component.TranslatableField` component, we have a `window.prestashop.component.translatableField` instance

Little typo here I think it's `window.prestashop.instance.translatableField` that contains the component, as stated in the ADR and performed in the code you provided (just wanted to avoid confusion here)

> 
> **I suggest we merge this as a Version 0.1 of this new system.**

👍

> In v0.2 we'll have to address:
> 
> * Advanced protection as suggested by @jolelievre : how to make sure module developer cannot mess with already instantiated instances?

As the `window.prestashop.component` is accessible by anyone I'm not sure we can really protect it, we could always provide some `registerComponent` methods or such, but it won't force to use them nor prevent to modify the internal variables.

> * Register all available components (currently there's only TranslatableField, TinyMCEEditor, TranslatableInput and TaggableField), maybe provide a versioning for them to help people know which ones are available in 1.7.8 or 1.7.9.

Maybe instead of versioning we could provide a `hasComponent` method? and/or `listComponents()`? This would have the advantage to check for components added by modules as well.

> * Decide whether we perform a "default initialization". With current implementation, module developer must explicitely ask which Components he wants to use:
> 
> ```
> $(document).ready(function () {
>   window.prestashop.component.initPrestashopComponents(
>     [
>       'TranslatableField',
>       'TinyMCEEditor',
>       'TranslatableInput',
>     ],
>   );
> });
> ```
> 
> Should we provide a default initialization function which provide instances for all components ?

I'm not sure it's needed, it seems better, from a performance point of view, to init the components required on the page only, so only the page itself is aware of its needs. The core pages already init the components they need, so the module that plugs in it should know which one are already there and which ones it needs to additionally activate.

admin-dev/themes/new-theme/js/app/utils/init-components.js
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/20591/commits/4a1efe9ddf00bcff6588b8c06560ec3a1dc08164 and https://github.com/PrestaShop/PrestaShop/pull/20591/commits/458ca9b6edb4361db4cac8fed10bc3c96f155209

matks (approved):

jolelievre (approved):

PierreRambaud (approved):

PR #20598, opened 2020-08-14
Do not reconvert discounts

### LINKED ISSUE #20594: Wrong cart rules discount on FO when using alternative currency
#### Describe the bug

On FO, when adding a cart rule on an order placed in a currency other than the default currency, the sum displayed in the cart is wrong. I think it's converted two times instead of one : For example, a discount of 1€ should be displayed as "151,69 DZD" (1x151,69), but it's displayed "23.009,86 DZD" (1x151,69x151,69).

It was OK in 1.7.6.7.

The invoice is however OK.

#### Expected behavior

No double conversion...

#### Steps to Reproduce

1. Go to BO > International > Localization > Currency > Add a new currency (Algerian Dinar is a good choice, since is easier to see the bug with this one)
2. Go to Catalog > Discount > Add a new cart rule (10% discount for example)
3. Go to FO, add a product to the cart, and then, add the cart rule
4. Check the cart rule value : It's wrong (it is converted two times), but the "Discount(s)" total is OK.
5. Complete the order, go to BO, change the status as paid, check Invoice : Invoice is OK.

wrong cart rule total in FO

**Screenshots**

https://drive.google.com/file/d/1D6PKVstPw7YgGdLRI32STWmh0LcMUR8p/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: N/A

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | In the FO, when adding a discount created with a different currency than the one used in the FO, the amount of the reduction could be wrong. It seems that we were converting an already converted amount to the current currency. This PR fixes it.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20594
| How to test?  | See #20594, Also, be sure that #9914 is still fixed

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20598)

### DIFF
--- a/src/Adapter/Presenter/Cart/CartPresenter.php
+++ b/src/Adapter/Presenter/Cart/CartPresenter.php
@@ -570,19 +570,6 @@ class CartPresenter implements PresenterInterface
             } else {
                 $freeShippingOnly = false;
                 $totalCartVoucherReduction = $this->includeTaxes() ? $cartVoucher['value_real'] : $cartVoucher['value_tax_exc'];
-                $currencyFrom = new \Currency($cartVoucher['reduction_currency']);
-                $currencyTo = new \Currency($cart->id_currency);
-                if ($currencyFrom->conversion_rate == 0) {
-                    $totalCartVoucherReduction = 0;
-                } else {
-                    // convert to default currency
-                    $defaultCurrencyId = (int) Configuration::get('PS_CURRENCY_DEFAULT');
-                    $totalCartVoucherReduction /= $currencyFrom->conversion_rate;
-                    if ($defaultCurrencyId == $currencyTo->id) {
-                        // convert to destination currency
-                        $totalCartVoucherReduction *= $currencyTo->conversion_rate;
-                    }
-                }
             }
 
             // when a voucher has only a shipping reduction, the value displayed must be "Free Shipping"
@@ -593,7 +580,7 @@ class CartPresenter implements PresenterInterface
                     'Admin.Shipping.Feature'
                 );
             } else {
-                $cartVoucher['reduction_formatted'] = '-' . $this->priceFormatter->convertAndFormat($totalCartVoucherReduction);
+                $cartVoucher['reduction_formatted'] = '-' . $this->priceFormatter->format($totalCartVoucherReduction);
             }
             $vouchers[$cartVoucher['id_cart_rule']]['reduction_formatted'] = $cartVoucher['reduction_formatted'];
             $vouchers[$cartVoucher['id_cart_rule']]['delete_url'] = $this->link->getPageLink(


matks (approved):
I love a PR that removes more code than it adds ❤️

src/Adapter/Presenter/Cart/CartPresenter.php
```diff
@@ -570,19 +570,6 @@ private function getTemplateVarVouchers(Cart $cart)
             } else {
                 $freeShippingOnly = false;
                 $totalCartVoucherReduction = $this->includeTaxes() ? $cartVoucher['value_real'] : $cartVoucher['value_tax_exc'];
-                $currencyFrom = new \Currency($cartVoucher['reduction_currency']);
```
jolelievre, line 573: I'm not sure you should remove it all 🤔  But maybe rather check if the currencies ids match, if they don't then you need to perform the conversion You can also use `Tools::convertPriceFull` which (I believe) manages all these checks and conversion

atomiix: I've tried many combinations with different currencies, different discount currencies, different kind of discounts and each time it worked well without performing any other conversion.

jolelievre: So does it mean `$cart->getCartRules()` already performs the conversion?
If this is something we changed we might need to mark this as a BC (if developers expect it to return not converted values)

jolelievre: Seems like it's a recent change (introduced in 177) by this PR https://github.com/PrestaShop/PrestaShop/pull/14263 So I guess we just need to ensure that it doesn't create a regression related to this issue https://github.com/PrestaShop/PrestaShop/issues/9914

atomiix: I'll add it in the *how to test* section 😉

jolelievre (approved):

PierreRambaud (approved):

ngodefroy:
@atomiix I created a 20AED amount cart rule here:
Capture d’écran 2020-08-20 à 14 09 00


I then applied this cart rule on the product below but the price of the discount should be 4,60 € instead of 5,52 €
Capture d’écran 2020-08-20 à 14 06 14


Currency rate AED
Capture d’écran 2020-08-20 à 14 05 47

atomiix:
@ngodefroy I think you forgot the tax: `20 / 4.35 = 4.60`, `4.60 * 1.20 = 5.52`

ngodefroy:
@atomiix  OK for me, I misunderstood the "tax excluded" box, thank you.

PR #20601, opened 2020-08-14
Allow admin user to configure the SameSite cookie attribute

PierreRambaud:
| Questions     | Answers
| Branch?       | develop
| Description?  | Regarding this article https://web.dev/samesite-cookies-explained/ we must provide a way to allow admin user to configure the SameSite cookie attribute. There is a BC about the `GeneralConfiguration` which is now require `Cookie` in construct parameter.
| Type?         | improvement
| Category?     | CO
| BC breaks?    | yes
| Deprecations? | no
| How to test?  | Make a fresh install, the `PS_COOKIE_SAMESITE` must have the `Lax` value. Then, when a user is connected, in FO or BO you should be able to check your cookie and see something like that: image<br> Must be tested with PHP 7.1 & PHP 7.3

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20601)

### DIFF
--- a/classes/Cookie.php
+++ b/classes/Cookie.php
@@ -29,6 +29,16 @@ use PrestaShop\PrestaShop\Core\Session\SessionInterface;
 
 class CookieCore
 {
+    const SAMESITE_NONE = 'None';
+    const SAMESITE_LAX = 'Lax';
+    const SAMESITE_STRICT = 'Strict';
+
+    const SAMESITE_AVAILABLE_VALUES = [
+        self::SAMESITE_NONE => self::SAMESITE_NONE,
+        self::SAMESITE_LAX => self::SAMESITE_LAX,
+        self::SAMESITE_STRICT => self::SAMESITE_STRICT,
+    ];
+
     /** @var array Contain cookie content in a key => value format */
     protected $_content = [];
 
@@ -374,7 +384,40 @@ class CookieCore
             $time = 1;
         }
 
-        return setcookie($this->_name, $content, $time, $this->_path, $this->_domain, $this->_secure, true);
+        $sameSite = Configuration::get('PS_COOKIE_SAMESITE');
+        // SameSite=None is only available when Secure is set
+        if (empty($sameSite) || ($sameSite === Cookie::SAMESITE_NONE && !$this->_secure)) {
+            $sameSite = Cookie::SAMESITE_LAX;
+        }
+
+        /*
+         * The alternative signature supporting an options array is only available since
+         * PHP 7.3.0, before there is no support for SameSite attribute.
+         */
+        if (PHP_VERSION_ID < 70300) {
+            return setcookie(
+                $this->_name,
+                $content,
+                $time,
+                $this->_path,
+                $this->_domain . '; SameSite=' . $sameSite,
+                $this->_secure,
+                true
+            );
+        }
+
+        return setcookie(
+            $this->_name,
+            $content,
+            [
+                'expires' => $time,
+                'path' => $this->_path,
+                'domain' => $this->_domain,
+                'secure' => $this->_secure,
+                'httponly' => true,
+                'samesite' => $sameSite,
+            ]
+        );
     }
 
     public function __destruct()
--- a/classes/Cookie.php
+++ b/classes/Cookie.php
@@ -385,10 +385,6 @@ class CookieCore
         }
 
         $sameSite = Configuration::get('PS_COOKIE_SAMESITE');
-        // SameSite=None is only available when Secure is set
-        if (empty($sameSite) || ($sameSite === Cookie::SAMESITE_NONE && !$this->_secure)) {
-            $sameSite = Cookie::SAMESITE_LAX;
-        }
 
         /*
          * The alternative signature supporting an options array is only available since
--- a/src/Adapter/GeneralConfiguration.php
+++ b/src/Adapter/GeneralConfiguration.php
@@ -26,6 +26,7 @@
 
 namespace PrestaShop\PrestaShop\Adapter;
 
+use Cookie;
 use PrestaShop\PrestaShop\Core\Configuration\DataConfigurationInterface;
 
 /**
@@ -38,9 +39,15 @@ class GeneralConfiguration implements DataConfigurationInterface
      */
     private $configuration;
 
-    public function __construct(Configuration $configuration)
+    /**
+     * @var Cookie
+     */
+    private $cookie;
+
+    public function __construct(Configuration $configuration, Cookie $cookie)
     {
         $this->configuration = $configuration;
+        $this->cookie = $cookie;
     }
 
     /**
@@ -53,6 +60,7 @@ class GeneralConfiguration implements DataConfigurationInterface
             'check_ip_address' => $this->configuration->getBoolean('PS_COOKIE_CHECKIP'),
             'front_cookie_lifetime' => $this->configuration->get('PS_COOKIE_LIFETIME_FO'),
             'back_cookie_lifetime' => $this->configuration->get('PS_COOKIE_LIFETIME_BO'),
+            'cookie_samesite' => $this->configuration->get('PS_COOKIE_SAMESITE'),
         ];
     }
 
@@ -64,10 +72,22 @@ class GeneralConfiguration implements DataConfigurationInterface
         $errors = [];
 
         if ($this->validateConfiguration($configuration)) {
-            $this->configuration->set('PRESTASTORE_LIVE', (bool) $configuration['check_modules_update']);
-            $this->configuration->set('PS_COOKIE_CHECKIP', (bool) $configuration['check_ip_address']);
-            $this->configuration->set('PS_COOKIE_LIFETIME_FO', (int) $configuration['front_cookie_lifetime']);
-            $this->configuration->set('PS_COOKIE_LIFETIME_BO', (int) $configuration['back_cookie_lifetime']);
+            if (!$this->validateSameSite($configuration['cookie_samesite'])) {
+                $errors[] = [
+                    'key' => 'The SameSite=None is only available in Secure mode.',
+                    'domain' => 'Admin.Advparameters.Notification',
+                    'parameters' => [],
+                ];
+            } else {
+                $this->configuration->set('PRESTASTORE_LIVE', (bool) $configuration['check_modules_update']);
+                $this->configuration->set('PS_COOKIE_CHECKIP', (bool) $configuration['check_ip_address']);
+                $this->configuration->set('PS_COOKIE_LIFETIME_FO', (int) $configuration['front_cookie_lifetime']);
+                $this->configuration->set('PS_COOKIE_LIFETIME_BO', (int) $configuration['back_cookie_lifetime']);
+                $this->configuration->set('PS_COOKIE_SAMESITE', $configuration['cookie_samesite']);
+                // Clear checksum to force the refresh
+                $this->cookie->checksum = '';
+                $this->cookie->write();
+            }
         }
 
         return $errors;
@@ -83,6 +103,27 @@ class GeneralConfiguration implements DataConfigurationInterface
             $configuration['check_ip_address'],
             $configuration['front_cookie_lifetime'],
             $configuration['back_cookie_lifetime']
+        ) && in_array(
+            $configuration['cookie_samesite'],
+            Cookie::SAMESITE_AVAILABLE_VALUES
         );
+    }
+
+    /**
+     * Validate SameSite.
+     * The SameSite=None is only working when Secure is settle
+     *
+     * @param string $sameSite
+     *
+     * @return bool
+     */
+    protected function validateSameSite(string $sameSite): bool
+    {
+        $forceSsl = Configuration::get('PS_SSL_ENABLED') && Configuration::get('PS_SSL_ENABLED_EVERYWHERE');
+        if ($sameSite === Cookie::SAMESITE_NONE) {
+            return $forceSsl;
+        }
+
+        return true;
     }
 }
--- a/src/Adapter/GeneralConfiguration.php
+++ b/src/Adapter/GeneralConfiguration.php
@@ -111,7 +111,7 @@ class GeneralConfiguration implements DataConfigurationInterface
 
     /**
      * Validate SameSite.
-     * The SameSite=None is only working when Secure is settle
+     * The SameSite=None is only working when Secure is settled
      *
      * @param string $sameSite
      *
--- a/src/Adapter/GeneralConfiguration.php
+++ b/src/Adapter/GeneralConfiguration.php
@@ -119,7 +119,7 @@ class GeneralConfiguration implements DataConfigurationInterface
      */
     protected function validateSameSite(string $sameSite): bool
     {
-        $forceSsl = Configuration::get('PS_SSL_ENABLED') && Configuration::get('PS_SSL_ENABLED_EVERYWHERE');
+        $forceSsl = $this->configuration->get('PS_SSL_ENABLED') && $this->configuration->get('PS_SSL_ENABLED_EVERYWHERE');
         if ($sameSite === Cookie::SAMESITE_NONE) {
             return $forceSsl;
         }
--- a/install-dev/data/xml/configuration.xml
+++ b/install-dev/data/xml/configuration.xml
@@ -332,6 +332,9 @@
   <configuration id="PS_COOKIE_CHECKIP" name="PS_COOKIE_CHECKIP">
     <value>1</value>
   </configuration>
+  <configuration id="PS_COOKIE_SAMESITE" name="PS_COOKIE_SAMESITE">
+    <value>Lax</value>
+  </configuration>
   <configuration id="PS_USE_ECOTAX" name="PS_USE_ECOTAX">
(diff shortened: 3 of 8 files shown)


src/Adapter/GeneralConfiguration.php
```diff
@@ -38,9 +39,15 @@ class GeneralConfiguration implements DataConfigurationInterface
      */
     private $configuration;
 
-    public function __construct(Configuration $configuration)
+    /**
+     * @var Cookie
+     */
+    private $cookie;
+
+    public function __construct(Configuration $configuration, Cookie $cookie)
```
matks, line 47:
```suggestion
    public function __construct(Configuration $configuration, Cookie $cookie = null)
```

Else it's a BC break ?

src/Adapter/GeneralConfiguration.php
```diff
@@ -83,6 +103,27 @@ public function validateConfiguration(array $configuration)
...
             $configuration['back_cookie_lifetime']
+        ) && in_array(
+            $configuration['cookie_samesite'],
+            Cookie::SAMESITE_AVAILABLE_VALUES
         );
     }
+
+    /**
+     * Validate SameSite.
+     * The SameSite=None is only working when Secure is settle
```
matks:
```suggestion
     * The SameSite=None is only working when Secure is settled
```

matks (commented):
I provided a few suggestions.

Shall we consider a backport for 1.7.7 ? It's a security feature, how important is it ?

src/Adapter/GeneralConfiguration.php
PierreRambaud: Yes it's a BC, forget to edit the PR description. We can't have $cookie = null, otherwise the SameSite configuration will not work as expected.

PierreRambaud:
> I provided a few suggestions.
> 
> Shall we consider a backport for 1.7.7 ? It's a security feature, how important is it ?

AFAIK, no new feature in the 1.7.7.x, ping @eternoendless

matks:
> > I provided a few suggestions.
> > Shall we consider a backport for 1.7.7 ? It's a security feature, how important is it ?
> 
> AFAIK, no new feature in the 1.7.7.x, ping @eternoendless

No feature indeed. Question is whether this is a feature or not. Chrome popping up errors in the Console (and maybe in future versions something worth than errors) can be considered a bug 😄  but the decision shall be his

matks (approved):

PierreRambaud:
> > > I provided a few suggestions.
> > > Shall we consider a backport for 1.7.7 ? It's a security feature, how important is it ?
> > 
> > 
> > AFAIK, no new feature in the 1.7.7.x, ping @eternoendless
> 
> No feature indeed. Question is whether this is a feature or not. Chrome popping up errors in the Console (and maybe in future versions something worth than errors) can be considered a bug  but the decision shall be his

Google Chrome is popping errors / warnings only when the SameSite option is not set, by default the value is Lax.

matthieu-rolland (approved):

msaustral:
Hi one question, to get the cookie samesite=none working i have to modify all 8 files?

classes/Cookie.php
```diff
@@ -374,7 +384,36 @@ protected function encryptAndSetCookie($cookie = null)
             $time = 1;
         }
 
-        return setcookie($this->_name, $content, $time, $this->_path, $this->_domain, $this->_secure, true);
+        $sameSite = Configuration::get('PS_COOKIE_SAMESITE');
```
msaustral, line 387: if (Configuration::get('PS_SSL_ENABLED') == 0){
            return setcookie($this->_name, $content, $time, $this->_path, $this->_domain, $this->_secure, true);
        }else{

classes/Cookie.php
```diff
@@ -374,7 +384,36 @@ protected function encryptAndSetCookie($cookie = null)
...
+            [
+                'expires' => $time,
+                'path' => $this->_path,
+                'domain' => $this->_domain,
+                'secure' => $this->_secure,
+                'httponly' => true,
+                'samesite' => $sameSite,
+            ]
+        );
     }
```
msaustral, line 417: }

msaustral:
Hi i suggest this change to the Cookie class as this:

`

if (Configuration::get('PS_SSL_ENABLED') == 0){

            return setcookie($this->_name, $content, $time, $this->_path, $this->_domain, $this->_secure, true);

        }else{

           if (PHP_VERSION_ID < 70300) {
            return setcookie(
                $this->_name,
                $content,
                $time,
                $this->_path,
                $this->_domain . '; SameSite=' . $sameSite,
                $this->_secure,
                true
            );
        }

        return setcookie(
            $this->_name,
            $content,
            [
                'expires' => $time,
                'path' => $this->_path,
                'domain' => $this->_domain,
                'secure' => $this->_secure,
                'httponly' => true,
                'samesite' => $sameSite,
            ]
        );
}
`

when you work on localhost without SSL, Chrome prevent the issue of the cookie as it can not be secure and the car does not work.

PierreRambaud:
> Hi one question, to get the cookie samesite=none working i have to modify all 8 files?

Hi, the SameSite cookie settled as None requires to have the SSL activated everywhere as you can see here https://github.com/PrestaShop/PrestaShop/pull/20601/files#diff-75f4381b94496cb1e5b63be089ed55f7R122

msaustral:
Hi PierreRambaud, when we are developping a new e-commerce and working on localhost we do not use SSL, if you do not validate that on the Cookie class, Chrome does not allow to isue the cookie and the store does not work. Have a good day.

msaustral (commented):
> > Hi one question, to get the cookie samesite=none working i have to modify all 8 files?
> 
> Hi, the SameSite cookie settled as None requires to have the SSL activated everywhere as you can see here https://github.com/PrestaShop/PrestaShop/pull/20601/files#diff-75f4381b94496cb1e5b63be089ed55f7R122

Hi PierreRambaud, when we are developping a new e-commerce and working on localhost we do not use SSL, if you do not validate that on the Cookie class, Chrome does not allow to isue the cookie and the store does not work. Have a good day.

PierreRambaud:
> Hi PierreRambaud, when we are developping a new e-commerce and working on localhost we do not use SSL, if you do not validate that on the Cookie class, Chrome does not allow to isue the cookie and the store does not work. Have a good day.

This is something I don't understand, why setting it to None, when you can use Lax if you don't use SSL, Lax is the default value when the Cookie doesn't have SameSite property, and with your [commit](https://github.com/PrestaShop/PrestaShop/pull/20601#discussion_r474962944) you're disabling the SameSite behavior :thinking:

msaustral:
> > Hi PierreRambaud, when we are developping a new e-commerce and working on localhost we do not use SSL, if you do not validate that on the Cookie class, Chrome does not allow to isue the cookie and the store does not work. Have a good day.
> 
> This is something I don't understand, why setting it to None, when you can use Lax if you don't use SSL, Lax is the default value when the Cookie doesn't have SameSite property, and with your [commit](https://github.com/PrestaShop/PrestaShop/pull/20601#discussion_r474962944) you're disabling the SameSite behavior 🤔

 Hi PierreRambaud, we set it this way so when we send it for production will work with the payment module, if cookie is lax it does not work on production.

msaustral:
> > > Hi PierreRambaud, when we are developping a new e-commerce and working on localhost we do not use SSL, if you do not validate that on the Cookie class, Chrome does not allow to isue the cookie and the store does not work. Have a good day.
> > 
> > 
> > This is something I don't understand, why setting it to None, when you can use Lax if you don't use SSL, Lax is the default value when the Cookie doesn't have SameSite property, and with your [commit](https://github.com/PrestaShop/PrestaShop/pull/20601#discussion_r474962944) you're disabling the SameSite behavior 🤔
> 
> Hi PierreRambaud, we set it this way so when we send it for production will work with the payment module, if cookie is lax it does not work on production.

Hi Pierre, we are havin problem with the backoffice when we set the cookie to samesite=none, it not allowing access to it without any error, the cookie is not set, any idea?

PierreRambaud:
> > > > Hi PierreRambaud, when we are developping a new e-commerce and working on localhost we do not use SSL, if you do not validate that on the Cookie class, Chrome does not allow to isue the cookie and the store does not work. Have a good day.
> > > 
> > > 
> > > This is something I don't understand, why setting it to None, when you can use Lax if you don't use SSL, Lax is the default value when the Cookie doesn't have SameSite property, and with your [commit](https://github.com/PrestaShop/PrestaShop/pull/20601#discussion_r474962944) you're disabling the SameSite behavior thinking
> > 
> > 
> > Hi PierreRambaud, we set it this way so when we send it for production will work with the payment module, if cookie is lax it does not work on production.
> 
> Hi Pierre, we are havin problem with the backoffice when we set the cookie to samesite=none, it not allowing access to it without any error, the cookie is not set, any idea?

image

It's because you *must* use SSL when using the SameSite cookie at None.
With what you suggest, you totally remove the usage of SameSite :thinking:

msaustral:
> > > > > Hi PierreRambaud, when we are developping a new e-commerce and working on localhost we do not use SSL, if you do not validate that on the Cookie class, Chrome does not allow to isue the cookie and the store does not work. Have a good day.
> > > > 
> > > > 
> > > > This is something I don't understand, why setting it to None, when you can use Lax if you don't use SSL, Lax is the default value when the Cookie doesn't have SameSite property, and with your [commit](https://github.com/PrestaShop/PrestaShop/pull/20601#discussion_r474962944) you're disabling the SameSite behavior thinking
> > > 
> > > 
> > > Hi PierreRambaud, we set it this way so when we send it for production will work with the payment module, if cookie is lax it does not work on production.
> > 
> > 
> > Hi Pierre, we are havin problem with the backoffice when we set the cookie to samesite=none, it not allowing access to it without any error, the cookie is not set, any idea?
> 
> image
> 
> It's because you _must_ use SSL when using the SameSite cookie at None.
> With what you suggest, you totally remove the usage of SameSite 🤔

Hi Pierre, thank you for your answer, we use the samesite and secure, there are 2 things, 

1.- the samesite and secure is needed for production, not for developing, if you do not use SSL on developing and you do not validate it, the store wont work.

2.- when you use samesite=none secure on cookie class (production site), the front page work perfect but backoffice does not enter, if you check there is no cookie when you try to access.

Any idea??

Progi1984:
@PierreRambaud Could you add a note in the PR :notebook: for the BC, please ? (for release notes)

PierreRambaud:
> > > > > > Hi PierreRambaud, when we are developping a new e-commerce and working on localhost we do not use SSL, if you do not validate that on the Cookie class, Chrome does not allow to isue the cookie and the store does not work. Have a good day.
> > > > > 
> > > > > 
> > > > > This is something I don't understand, why setting it to None, when you can use Lax if you don't use SSL, Lax is the default value when the Cookie doesn't have SameSite property, and with your [commit](https://github.com/PrestaShop/PrestaShop/pull/20601#discussion_r474962944) you're disabling the SameSite behavior thinking
> > > > 
> > > > 
> > > > Hi PierreRambaud, we set it this way so when we send it for production will work with the payment module, if cookie is lax it does not work on production.
> > > 
> > > 
> > > Hi Pierre, we are havin problem with the backoffice when we set the cookie to samesite=none, it not allowing access to it without any error, the cookie is not set, any idea?
> > 
> > 
> > image
> > It's because you _must_ use SSL when using the SameSite cookie at None.
> > With what you suggest, you totally remove the usage of SameSite thinking
> 
> Hi Pierre, thank you for your answer, we use the samesite and secure, there are 2 things,
> 
> 1.- the samesite and secure is needed for production, not for developing, if you do not use SSL on developing and you do not validate it, the store wont work.
> 
> 2.- when you use samesite=none secure on cookie class (production site), the front page work perfect but backoffice does not enter, if you check there is no cookie when you try to access.
> 
> Any idea??

If you're not in production you must use SameSite=Lax, as it's said in the documentation, this is the default value.
You need to find a way to do it, because if a few days/weeks, Google Chrome will automatically enable this feature and will consider all cookies without any SameSite to Lax.

SD1982:
LGTM the "lax" value is correctly added to the cookies 
Tested with php 7.1 / 7.2 / 7.3
Thanks @PierreRambaud !!

Progi1984:
Thanks @PierreRambaud & @SD1982

ffiebig:
Sorry to revive this. @PierreRambaud thanks a lot for this contribution. 

When trying out this fix in order to address some problems with Payment Gateway redirections I used a portion of your code, the `encryptAndSetCookie` return to be exact, and I forced the `SameSite` attribute to be `None` at all times, also using SSL so it would be also marked `Secure`.

I checked that `PS_SSL_ENABLED_EVERYWHERE` and `PS_SSL_ENABLED` where both enabled 

Long story short cookies get the `SameSite=None` and `Secure` set, this fixes the gateway issue, but it prevents admins from logging into the back office.  `SameSite=Lax` and `Secure` brings back the back office but it breaks the payment gateway
Did you stumble with something like this ? 

I got this behaviour using Prestashop 1.7.6.7. Maybe I'm getting this behaviour because I'm not on the development branch ?

msaustral:
Hi Felipe, we have the code that solve the issue https://github.com/msaustral/Prestashop-cookies

fkomaralp:
You need to override Prestashop's Cookie.php file. **Update :** This method is not working on AdminLogin page. So, we need to update code.

modules/module_name/overrides/classes/Cookie.php

```
<?php

class Cookie extends CookieCore
{
    protected function encryptAndSetCookie($cookie = null)
    {
        // Check if the content fits in the Cookie
        $length = (ini_get('mbstring.func_overload') & 2) ? mb_strlen($cookie, ini_get('default_charset')) : strlen($cookie);
        if ($length >= 1048576) {
            return false;
        }
        if ($cookie) {
            $content = $this->cipherTool->encrypt($cookie);
            $time = $this->_expire;
        } else {
            $content = 0;
            $time = 1;
        }

        if(Tools::getValue("controller") !== "AdminLogin"){
            if (PHP_VERSION_ID < 70300) {
                return setcookie($this->_name, $content, $time, $this->_path, $this->_domain . '; SameSite=None;', $this->_secure, true);
            }
            else {
                return setcookie($this->_name, $content, [
                    'expires' => $time,
                    'path' => $this->_path,
                    'domain' => $this->_domain,
                    'samesite' => 'None',
                    'secure' => $this->_secure,
                    'httponly' => true,
                ]);
            }
        } else {
            return setcookie($this->_name, $content, $time, $this->_path, $this->_domain, $this->_secure, true);
        }
    }
}
```

Thats worked for me.

ffiebig:
I'm assuming that setting the `SameSite` attribute to `None` is what breaks the back office. I don't understand the attribute behaviour enough (nor PrestaShop) to know if that is supposed to happen or not. 

If it's ok for it to break then it might be good to have that condition @fkomaralp, if it shouldn't break then there must be an underlying reason to fix instead

PierreRambaud:
Hi guys.
Regarding the RFC, you can't have `SameSite` attribute to `None` without activating the `Secure` mode.
This mean you must activate SSL and it's recommended by Google, future version will mark websites without SSL as not secure.
https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-06
https://www.chromestatus.com/feature/5633521622188032
https://web.dev/samesite-cookies-explained/
This is why we add this feature, and other browsers are going to follow it too.

> Enabled by default (tracking bug) in:
Chrome for desktop release 85

msaustral:
Hi Pierre, we have moths saying the same, we do know that we need SameSite=none and SSL, what we are reporting is that when you tag the cookie that way, **the BACKOFFICE STOP working, it does not open.**

PierreRambaud:
> Hi Pierre, we have moths saying the same, we do know that we need SameSite=none and SSL, what we are reporting is that when you tag the cookie that way, **the BACKOFFICE STOP working, it does not open.**

Can you please create an issue it will be easier to follow :)

msaustral:
> > Hi Pierre, we have moths saying the same, we do know that we need SameSite=none and SSL, what we are reporting is that when you tag the cookie that way, **the BACKOFFICE STOP working, it does not open.**
> 
> Can you please create an issue it will be easier to follow :)

Hi Pierre, the issue is caused by the same topic that we are discussing here. 

Once you that the cookie SameSite=none the backoffice stop working .

If we open a new ticket it will be mark as a duplicated and close.

If you are improving the way as PS is taging the cookies as the title say "Allow admin user to configure the SameSite cookie attribute", you have to consider that we need the backoffice working to do it, don't you think?

PierreRambaud:
> > > Hi Pierre, we have moths saying the same, we do know that we need SameSite=none and SSL, what we are reporting is that when you tag the cookie that way, **the BACKOFFICE STOP working, it does not open.**
> > 
> > 
> > Can you please create an issue it will be easier to follow :)
> 
> Hi Pierre, the issue is caused by the same topic that we are discussing here.
> 
> Once you that the cookie SameSite=none the backoffice stop working .
> 
> If we open a new ticket it will be mark as a duplicated and close.
> 
> If you are improving the way as PS is taging the cookies as the title say "Allow admin user to configure the SameSite cookie attribute", you have to consider that we need the backoffice working to do it, don't you think?

My bad, I'm really tired for weeks (vacations in a few days :crossed_fingers: )

Here's the fix #21249, Thank you @msaustral

PR #20203, opened 2020-07-17
Adding a link to the general terms and conditions in the checkout footer (Opquast n°45)

Sinepel:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop 
| Description?  | Currently the general terms and conditions of sale are not accessible in the order tunnel except at step 4 / the payment step. In order to comply with the OPQUAST checklist and recommendation number 45, the general terms and conditions of sale should be accessible from all pages of the site. I have therefore added a link to this page in the footer of the order tunnel. This link opens a modal in order not to leave the tunnel.
| Type?         | improvement 
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | Add a product to your cart & go to the checkout process

Capture d’écran 2020-08-21 à 15 13 16

:notebook: BC Break
* The modal `<div class="modal fade" id="modal">` has been moved from `themes/classic/templates/checkout/_partials/steps/payment.tpl` (block `step_content`) to `themes/classic/templates/checkout/_partials/footer.tpl` (block `footer`).

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20203)

PR #20606, opened 2020-08-17
Fix invoice address selection at checkout

### LINKED ISSUE #11595: Checkout: unchecking "Use this address for invoice too" when having more than one address
**Describe the bug**
At checkout, when you have more than one address and you edit the delivery address, if you uncheck "Use this address for invoice too" and click "Save", after the page refreshes you are in the delivery step, instead of being presented with a different invoicing address to choose from. 

**To Reproduce**
See attached screen recording

**Screenshots**
https://drive.google.com/open?id=1fZ71XYZMW21rcB0p-GyzjKD-fUMoDAoG

**Wanted behavior**
We should be redirected on address step with no address selected for invoice and continue button disabled

**Additionnal information**
PrestaShop version: 1.7.4.3
PHP version: N/A

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | At checkout (FO), when you have more than one address and you edit the delivery address, if you uncheck "Use this address for invoice too" and click "Save", after the page refreshes you are in the delivery step, instead of being presented with a different invoicing address to choose from.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | yes, we do not perform the automatic address assignment in OrderController
| Deprecations? | no
| Fixed ticket? | Fixes #11595
| How to test?  | See #11595, but we should also check<br/>- using/setting adresses behaviors as a whole (checkout, customer addresses in FO, in BO...)<br/>- if possible test with modules that interact with Checkout Process and might alter adresses data (auto-set a data ? use a map to allow user to pick its address ?)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20606)

:notebook: BC Break
* In the FrontController, the forceful assignment of the invoice address id is now disabled for OrderController.
Before:
`if (!isset($cart->id_address_invoice) || $cart->id_address_invoice == 0) {`
After:
`if ($this->automaticallyAllocateInvoiceAddress && (!isset($cart->id_address_invoice) || $cart->id_address_invoice == 0)) {`

### DIFF
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -398,7 +398,6 @@ class FrontControllerCore extends Controller
                 $cart->update();
             }
             /* Select an address if not set */
-
         }
 
         if (!isset($cart) || !$cart->id) {
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -172,7 +172,7 @@ class FrontControllerCore extends Controller
 
     /**
      * Set this parameter to false if you don't want cart's invoice address
-     * to be set automatically (this bad behavior is kept for legacy and BC purpose)
+     * to be set automatically (this behavior is kept for legacy and BC purpose)
      *
      * @var bool automaticallyAllocateInvoiceAddress
      */
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -171,6 +171,14 @@ class FrontControllerCore extends Controller
     protected $cccReducer;
 
     /**
+     * Set this parameter to false if you don't want cart's invoice address
+     * to be set automatically (this bad behavior is kept for legacy and BC purpose)
+     *
+     * @var bool automaticallyAllocateInvoiceAddress
+     */
+    protected $automaticallyAllocateInvoiceAddress = true;
+
+    /**
      * Controller constructor.
      *
      * @global bool $useSSL SSL connection flag
@@ -398,6 +406,21 @@ class FrontControllerCore extends Controller
                 $cart->update();
             }
             /* Select an address if not set */
+            if (isset($cart) && (!isset($cart->id_address_delivery) || $cart->id_address_delivery == 0 ||
+                    !isset($cart->id_address_invoice) || $cart->id_address_invoice == 0) && $this->context->cookie->id_customer) {
+                $to_update = false;
+                if (!isset($cart->id_address_delivery) || $cart->id_address_delivery == 0) {
+                    $to_update = true;
+                    $cart->id_address_delivery = (int) Address::getFirstCustomerAddressId($cart->id_customer);
+                }
+                if ($this->automaticallyAllocateInvoiceAddress && (!isset($cart->id_address_invoice) || $cart->id_address_invoice == 0)) {
+                    $to_update = true;
+                    $cart->id_address_invoice = (int) Address::getFirstCustomerAddressId($cart->id_customer);
+                }
+                if ($to_update) {
+                    $cart->update();
+                }
+            }
         }
 
         if (!isset($cart) || !$cart->id) {
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -179,6 +179,14 @@ class FrontControllerCore extends Controller
     protected $automaticallyAllocateInvoiceAddress = true;
 
     /**
+     * Set this parameter to false if you don't want cart's delivery address
+     * to be set automatically (this behavior is kept for legacy and BC purpose)
+     *
+     * @var bool automaticallyAllocateDeliveryAddress
+     */
+    protected $automaticallyAllocateDeliveryAddress = true;
+
+    /**
      * Controller constructor.
      *
      * @global bool $useSSL SSL connection flag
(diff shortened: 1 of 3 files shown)


classes/controller/FrontController.php
```diff
@@ -391,21 +391,7 @@ public function init()
                 $cart->update();
             }
             /* Select an address if not set */
-            if (isset($cart) && (!isset($cart->id_address_delivery) || $cart->id_address_delivery == 0 ||
```
matthieu-rolland, line 401: I'm aware this is violent...  do you guys think it could be considered as a BC ? Imho this code was very problematic and was partly the cause of the bug this PR is fixing. 

This is old code from 2011, I just hope we don't have features relying on cart's delivery and invoice addresses being updated in `FrontController::init()` 😓

matks: > This is old code from 2011, I just hope we don't have features relying on cart's delivery and invoice addresses being updated in FrontController::init() 😓

I expect that we have some 😓  😭  although the best people to tell us this would be people relying on such mechanisms to build their FO modules.

Would it be possible to get some insights from @v4lux @Quetzacoalt91 @PululuK @kpodemski @jf-viguier @prestarocket @prestamodule ? Do you think this little bit of logic "if there is a cart but no address, I pick up the first address I can use" is used and necessary ?

matks: If (as I am afraid) this bit of logic must be kept, maybe we can disable it for your usecase.

I'm thinking of introducing a new property `$automaticallyAllocateAnAddress = true` inside FrontController
and your usecase we must modify it to `false` in `init()` to avoid the logic being triggered

But I agree with you this little "magic" trick seems to be a potential cause for lot of weird behaviors 😞

Quetzacoalt91: Yeah be careful with that, I can easily imagine with this code, we may consider that a cart will always have a delivery & invoice address a long as at least one is defined in the account.

For instance, this could be used to estimate early (= before the checkout) how much the shipping costs.

matthieu-rolland: @Quetzacoalt91 So, in the end, I only disabled the forceful assignment of the invoice address id (not delivery), and only for OrderController. It should have minimum impact, I don't think invoice address weights on the shipping cost (or any cost).

classes/controller/FrontController.php
```diff
@@ -170,6 +170,14 @@ class FrontControllerCore extends Controller
      */
     protected $cccReducer;
 
+    /**
+     * Set this parameter to false if you don't want cart's invoice address
+     * to be set automatically (this bad behavior is kept for legacy and BC purpose)
```
matks:
```suggestion
     * to be set automatically (this behavior is kept for legacy and BC purpose)
```
We are not sure this is bad because we dont have the history 😅

matks (dismissed):
Approving the PR (added a suggestion)

matks:
I enhanced the "How to test" because I'm afraid 😅

classes/controller/FrontController.php
matthieu-rolland: well, I assume any code updating a specific field in the DB in `FrontController::init()` is bad 🤔 

But you're right, it's not a great idea to describe it like this in a comment, I'll update it right now 👍

matks:
Also added "BC Break" as it might be worth mentioning in Release Note this change

matks (dismissed):

uruk-hai-21 (dismissed):

atomiix (dismissed):

PierreRambaud (dismissed):

Progi1984:
@matthieu-rolland Please add a note :notebook: about the BC Break in the pull request for the release note (Sample #20203). Thanks

SD1982:
Hi @matthieu-rolland 
The redirection after a delivery address edit is correct now so for me the PR is OK !! :+1: 
But we also have the same wrong redirection after editing an "invoice address" That was not the purpose of this PR but maybe this behaviour can be corrected in the same time . WDYT ?
https://drive.google.com/file/d/1s4uNCkIgaz6vXbd1WGPjeLLmd04E4YZa/view?usp=sharing
@PrestaShop/prestashop-product-team WDYT ?

LouiseBonnard:
I agree with @SD1982, let's fix it for this behavior too, thanks!

jolelievre:
@matthieu-rolland seems like you need to update the behaviour for both addresses

matthieu-rolland:
> Hi @matthieu-rolland
> The redirection after a delivery address edit is correct now so for me the PR is OK !! 👍
> But we also have the same wrong redirection after editing an "invoice address" That was not the purpose of this PR but maybe this behaviour can be corrected in the same time . WDYT ?
> https://drive.google.com/file/d/1s4uNCkIgaz6vXbd1WGPjeLLmd04E4YZa/view?usp=sharing
> @PrestaShop/prestashop-product-team WDYT ?

nice catch !  done 👍

Progi1984 (dismissed):

atomiix (dismissed):

SD1982:
Hi @matthieu-rolland I still have the same wrong redirection ( to the delivery step ) than before when editing the invoice address ...

matthieu-rolland:
> Hi @matthieu-rolland I still have the same wrong redirection ( to the delivery step ) than before when editing the invoice address ...

thanks for your feedback @SD1982 !

I fixed this, so now if you edit an invoice address, you are not redirected to next step. Actually this bug was already existing in previous versions.

There is still a slight bug, that already existed in previous versions: when we edit an address (delivery or invoice), the address selectors are reversed, it's not a big deal but it's strange. Unfortunately there is no simple or clean way to fix this in this PR, I made an issue about this, it's related to the way addresses are stored in database: https://github.com/PrestaShop/PrestaShop/issues/21437

Progi1984 (dismissed):

Progi1984 (dismissed):

atomiix (dismissed):

SD1982:
LGTM tested from FO and from BO and with the checkout module 
Thanks @matthieu-rolland !!

Progi1984 (approved):

atomiix (approved):

atomiix:
Thanks @matthieu-rolland & @SD1982 !

PR #20622, opened 2020-08-18
Do not proceed setCurrentState if order already has the right state

### LINKED ISSUE #20623: setCurrentState to the same id_order_state
#### Describe the bug

It happens at every webservice PUT request for updating any order field. The order_state is reprocessed without checking if the order already has the state, sending the state's email and triggering all related hooks.

#### Expected behavior

If the order already has the requested state, avoid to execute the setCurrentState's body

#### Steps to Reproduce

Create an order state with the email sending option set to on, put an order to this state and then, send again the same put request using the WS. You will receive two emails because the order has changed twice to the same state.

#### Additional information

Fix related: https://github.com/PrestaShop/PrestaShop/pull/20622

bertoferrero:
This fix avoids executing the setCurrentState body if the order already has that id_order_state

| Questions     | Answers
| Branch?       | develop
| Description?  | It happens at every webservice PUT request for updating any order field. The order_state is reprocessed without checking if the order already has the state, sending the state's email and triggering all related hooks.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20623
| How to test?  | Create an order state with the email sending option set to on, put an order to this state and then, send again the same put request using the WS. You will receive two emails because the order has changed twice to the same state.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20622)

### DIFF
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1673,8 +1673,11 @@ class OrderCore extends ObjectModel
      */
     public function setCurrentState($id_order_state, $id_employee = 0)
     {
-        if (empty($id_order_state)) {
+        if (empty($id_order_state) || $id_order_state == $this->current_state) {
             return false;
+        }
+        if ($id_order_state == $this->current_state) {
+            return;
         }
         $history = new OrderHistory();
         $history->id_order = (int) $this->id;
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1676,9 +1676,6 @@ class OrderCore extends ObjectModel
         if (empty($id_order_state) || $id_order_state == $this->current_state) {
             return false;
         }
-        if ($id_order_state == $this->current_state) {
-            return;
-        }
         $history = new OrderHistory();
         $history->id_order = (int) $this->id;
         $history->id_employee = (int) $id_employee;
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1673,7 +1673,7 @@ class OrderCore extends ObjectModel
      */
     public function setCurrentState($id_order_state, $id_employee = 0)
     {
-        if (empty($id_order_state) || $id_order_state == $this->current_state) {
+        if (empty($id_order_state) || (int) $id_order_state === (int) $this->current_state) {
             return false;
         }
         $history = new OrderHistory();


classes/order/Order.php
```diff
@@ -1673,9 +1673,12 @@ public function getWsOrderRows()
      */
     public function setCurrentState($id_order_state, $id_employee = 0)
     {
-        if (empty($id_order_state)) {
+        if (empty($id_order_state) || $id_order_state == $this->current_state) {
             return false;
         }
+        if ($id_order_state == $this->current_state) {
```
PululuK: unnecessary 🤔 ?
Already checked just before

PierreRambaud (changes requested):
Duplicated condition as @PululuK said :)

classes/order/Order.php
bertoferrero: Yes, you are right, a mistake pasting the code, let me fix it.

bertoferrero:
> 
> 
> Duplicated condition as @PululuK said :)

Fixed, sorry

classes/order/Order.php
```diff
@@ -1673,7 +1673,7 @@ public function getWsOrderRows()
      */
     public function setCurrentState($id_order_state, $id_employee = 0)
     {
-        if (empty($id_order_state)) {
+        if (empty($id_order_state) || $id_order_state == $this->current_state) {
```
PierreRambaud:
```suggestion
        if (empty($id_order_state) || (int) $id_order_state === (int) $this->current_state) {
```
Let's have a strict comparison :)

PierreRambaud (approved):

matks (approved):

matks:
Nice catch :) thanks, moving to QA now

khouloudbelguith:
Hi @prestaquality,

It is ok :heavy_check_mark: 
https://drive.google.com/file/d/1sfGY4HXpt18g-mULlMmKcLcpAXXfjJcO/view?usp=sharing

Thanks!

Progi1984:
Thanks @prestaquality & @khouloudbelguith

PR #20626, opened 2020-08-18
Fix duplicate ids in order page

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Make sure html are not duplicated.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Part in screenshots are working as expected.<br>image<br>image<br>image

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20626)


admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
```diff
@@ -105,6 +105,9 @@ export default class OrderProductEdit {
   displayProduct(product) {
     this.productRowEdit = $(OrderViewPageMap.productEditRowTemplate).clone(true);
     this.productRowEdit.attr('id', `editOrderProduct_${this.orderDetailId}`);
+    this.productRowEdit.find('*[id]').each(function removeAllIds() {
```
matks, line 108: 😆 
"A deleted ID is a good ID that does not mess with my code"

matks (dismissed):
I approve the simple bugfix for 1.7.7 but I would like that we create an issue to fix this in a better way in 1.7.8 . 😉

NeOMakinG:
> I approve the simple bugfix for 1.7.7 but I would like that we create an issue to fix this in a better way in 1.7.8 . 😉

Is there any better way ? I mean, removing every ids from form will break the whole BO, don't you think ? ^^

NeOMakinG (dismissed):

admin-dev/themes/new-theme/js/pages/order/view/order-product-edit.js
PierreRambaud: I want to die

PierreRambaud:
> > I approve the simple bugfix for 1.7.7 but I would like that we create an issue to fix this in a better way in 1.7.8 . wink
> 
> Is there any better way ? I mean, removing every ids from form will break the whole BO, don't you think ? ^^

It removes ids from the template because it's not really a template :sweat_smile: 
For the 1.7.8.x it must be replaced by a CollectionType of forms and play with the prototype attribute.

NeOMakinG (dismissed):

matks (dismissed):

NeOMakinG:
It's WIP, I'm dumb 🤣

matthieu-rolland (dismissed):
**LGTM**
blindbart

Robin-Fischer-PS:
Hi @PierreRambaud !

Changing status from the Status block does not work with your PR. Changing it from the header of the order is OK.

pr20626 changing status KO

NeOMakinG (approved):

matthieu-rolland (approved):

matks (approved):

PR #20661, opened 2020-08-20
Add some missing throws in ObjectModelCore

ksaandev:
| Questions     | Answers
| Branch?       | develop
| Description?  | My IDE started crying when I opened this class. I though I could give it some love.
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | Open your IDE, open this class, see the tears. This PR solves some tiny things.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20661)

### DIFF
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -1521,7 +1521,6 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
      * @param bool $all if true, caches required fields of all object classes
      *
      * @throws PrestaShopDatabaseException
-     *
      */
     public function cacheFieldsRequiredDatabase($all = true)
     {
@@ -1882,7 +1881,6 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
      * @return bool
      *
      * @throws PrestaShopDatabaseException
-     *
      */
     public static function existsInDatabase($id_entity, $table)
     {
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -1235,6 +1235,8 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
      * @param bool $htmlentities if true, uses htmlentities() for field name translations in errors
      *
      * @return array
+     *
+     * @throws \PrestaShop\PrestaShop\Adapter\CoreException
      */
     public function validateController($htmlentities = true)
     {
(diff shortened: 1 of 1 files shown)


matthieu-rolland (commented):
Thank you @ksaandev !

Seems like php cs fixer is not happy  :) Could you just do this, from the root of your prestashop project ?

`./vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix classes/ObjectModel.php`

(of course `composer install` must have been done before)

And then commit the change? Then your PR can be merged, thanx !

ksaandev:
> Thank you @ksaandev !
> 
> Seems like php cs fixer is not happy :) Could you just do this, from the root of your prestashop project ?
> 
> `./vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix classes/ObjectModel.php`
> 
> (of course `composer install` must have been done before)
> 
> And then commit the change? Then your PR can be merged, thanx !

Done ! thx

classes/ObjectModel.php
```diff
@@ -1235,6 +1235,8 @@ public static function displayFieldName($field, $class = __CLASS__, $htmlentitie
      * @param bool $htmlentities if true, uses htmlentities() for field name translations in errors
      *
      * @return array
+     *
+     * @throws \PrestaShop\PrestaShop\Adapter\CoreException
```
PierreRambaud, line 1239: This method throws nothing here :thinking: 
If you start to use the `@throws` you need to specify the reason as it's described in the PHPdoc documentation
> This tag is used to present in your documentation which error COULD occur and under which circumstances. It is RECOMMENDED to provide a description that describes the reason an exception is thrown.

Otherwise adding a `@throws` here is in my own opinion useless

PierreRambaud (changes requested):
Methods are throwing nothing. I add a comment to explain why :)

classes/ObjectModel.php
ksaandev: @PierreRambaud To be honest, I did not add these `@throws`. My IDE did 😄 when I was working on these functions I just committed the result.

What you say makes sense 👍 but my IDE tells me these `throws` are missing, not the context 😆 . I will close my PR then.

ksaandev:
Closing the PR as I get relevant feedback from @PierreRambaud 

I will come back if I can find a better context 😉

PR #20684, opened 2020-08-24
Fix empty optional customization exception when creating BO order

### LINKED ISSUE #20671: BO - Orders page - An exception is displayed when we try to add a customized product
#### Describe the bug

In the BO => Orders => New Order page => an exception is displayed when we try a customized product 
image

#### Steps to Reproduce

1. Go to BO => Catalog => Products page
2. Create a product with **minimal quantity > 1** and with a **customized text** not required
image

3. Go to BO => Orders page => Add a new Order page
4. Search for the last product created
4.1 => if the customization is empty => an exception is displayed in the console
(**this exception is displayed also when minimal quantity =1**)
4.2 => if the customization is not empty => an alert is displayed with wrong alert `Minimum quantity of 0 must be added to cart.`

#### Additional information

* PrestaShop version: 1.7.7.x, 1770beta2
* PHP version: 7.2

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When creating an order in the BO, an exception was thrown when adding a product with an empty **optional** customization field. This PR fixes it.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | yes
| Fixed ticket? | Fixes #20671
| How to test?  | See #20671

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20684)

### DIFF
--- a/src/Adapter/Cart/CommandHandler/AddProductToCartHandler.php
+++ b/src/Adapter/Cart/CommandHandler/AddProductToCartHandler.php
@@ -79,7 +79,7 @@ final class AddProductToCartHandler extends AbstractCartHandler implements AddPr
                 $command->getCustomizationsByFieldIds()
             ));
             if (null !== $customizationIdVO) {
-                $combinationId = $customizationIdVO->getValue();
+                $customizationId = $customizationIdVO->getValue();
             }
         }
 
--- a/src/Adapter/Cart/CommandHandler/AddProductToCartHandler.php
+++ b/src/Adapter/Cart/CommandHandler/AddProductToCartHandler.php
@@ -73,11 +73,14 @@ final class AddProductToCartHandler extends AbstractCartHandler implements AddPr
         $customizationId = null;
 
         if (!empty($command->getCustomizationsByFieldIds())) {
-            $customizationId = $this->addCustomizationHandler->handle(new AddCustomizationCommand(
+            $customizationIdVO = $this->addCustomizationHandler->handle(new AddCustomizationCommand(
                 $cartIdValue,
                 $command->getProductId()->getValue(),
                 $command->getCustomizationsByFieldIds()
-            ))->getValue();
+            ));
+            if (null !== $customizationIdVO) {
+                $combinationId = $customizationIdVO->getValue();
+            }
         }
 
         $cart = $this->getCart($command->getCartId());


PierreRambaud (dismissed):

matthieu-rolland (dismissed):

src/Adapter/Cart/CommandHandler/AddProductToCartHandler.php
```diff
@@ -73,11 +73,14 @@ public function handle(AddProductToCartCommand $command): void
         $customizationId = null;
```
matthieu-rolland, line 73: Now the value of `$customizationId` remains null, yet this variable is passed in `$cart->getProductQuantity` and `$this->updateProductQuantityInCartHandler->handle()` just below.

Is it ok that this value will be passed as null all the time ? (in this case we could just pass a  null value instead of using this variable)

atomiix: @matthieu-rolland yes, an optional customization not filled is the same as no customization at all ;)

matthieu-rolland: alright ! in that case maybe we can drop the `$customizationId` variable, since its value is always null and never changes ? (and pass null as a parameter instead below)

matthieu-rolland: Or maybe, if `$customizationIdVO` is set, we should put `$customizationIdVO->getValue()` as a parameter instead of `$customizationId` that is always null (in the two functions below) ?

atomiix: Oh! Nice catch! I did something stupid here 😅 I used `$combinationId` as `$customizationId`!

matthieu-rolland (approved):

PierreRambaud (approved):

ngodefroy:
TEST OK:
With customization=> No Alert displayed
https://drive.google.com/file/d/1_XUnl3BglT-N_UWi8JCmz9kbx0Mo1LrE/view

Without Customization=> No error in console 
custom_empty

matks:
Thank you @atomiix

PR #20707, opened 2020-08-25
Fix ReflexionHelper class name

mvorisek:
| Questions     | Answers
| Branch?       | develop
| Description?  | see title
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | CI must pass

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20707)

### DIFF
--- a/tests-legacy/TestCase/ReflectionHelper.php
+++ b/tests-legacy/TestCase/ReflectionHelper.php
@@ -30,7 +30,7 @@ use PHPUnit\Framework\TestCase;
 use ReflectionClass;
 
 /**
- * Class ReflexionHelper
+ * Class ReflectionHelper
  * @package LegacyTests\TestCase
  *
  * Provides utilities to access private or protected properties inside classes.
@@ -41,7 +41,7 @@ use ReflectionClass;
  *
  * In the end, this kind of features is here just for convenience to be able to test quickly dirty legacy code.
  */
-class ReflexionHelper extends TestCase
+class ReflectionHelper extends TestCase
 {
     public static function invoke($object, $method)
     {


mvorisek:
@PierreRambaud can we merge this small fix incl. the updated `.gitignore`?

.gitignore
```diff
@@ -280,6 +280,7 @@ sftp-config.json
 
 ## Directory-based project format:
 .idea/
+nbproject/
```
mvorisek: for Netbeans please :)

PierreRambaud: It's not needed, with git you can create a global gitignore file in your home directory
`~/.gitignore` we should only use what we have in this project :thinking:

mvorisek: you are right, but as there was `nbproject/private` ignore already, I simplified it and both usecases should be handled

.gitignore
```diff
@@ -302,7 +302,7 @@ crashlytics.properties
 crashlytics-build.properties
 
 ### NetBeans ###
-nbproject/private/
+nbproject/
```
mvorisek: we should ignore all projects, not only private files, like for `.idea`

PierreRambaud: You must do it in another commit / pull request, it has no aim regarding the title and the description :/

mvorisek: PR here - https://github.com/PrestaShop/PrestaShop/pull/21076, this PR also updated to contain only the described change

PierreRambaud (approved):

Progi1984 (approved):

matks (approved):

matks:
Nice catch @mvorisek thank you

Issue #18547, opened 2020-04-09
PrestaShop must be compatible with PHP 8.0


mvorisek:
PHP 8.0 is out and PrestaShop must be compatible

**Related issues:**
- [x] https://github.com/PrestaShop/PrestaShop/issues/17300
- [x] https://github.com/PrestaShop/PrestaShop/issues/20739
- [x] https://github.com/PrestaShop/PrestaShop/issues/17351
- [x] #25921
- [x] #25922

Gipielle:
:-) mvorisek you are a dreamer !

eternoendless:
I think it will be harder than with PHP 7.4 though, since it's a major version they may remove deprecated code and introduce a lot of breaking changes.

But of course we'll keep a close eye as they release more information about what it will change.

kpodemski:
PHP 8...... .......PrestasHoP 8

davidglezz:
PHP 8.0.0 Alpha 1 available for testing!
https://www.php.net/index.php#id2020-06-25-1

lil5:
Any update on the matter?

PHP 8.0.0 has been released 

https://www.php.net/archive/2020.php#2020-11-26-3

matks:
Hi, our current php compatibility roadmap is

| PHP version     | will be compatible with | status
| ------------- | ---------------------------------|--------------------
| php7.3 | PrestaShop 1.7.7.0 | done ✅ 
| php7.4 | PrestaShop 1.7.8.0 | done ✅ 
| php8.0 | PrestaShop 1.7.9.0 | to do

mariochacondev:
Hi guys, any chance of this happening anytime soon?

matks:
@mariochacondev Not soon, no.

PrestaShop 1.7.8 will bring php7.4 compatibility and be delivered in summer 2021. PrestaShop 1.7.9 will bring php8.0 compatibility and be delivered somewhere around end of 2021 or beginning of 2022.

Sorry about that, but PrestaShop is a big software, we try to maintain a range of compatibility* and php8 is a very big jump (see https://24daysindecember.net/2020/12/21/a-perfect-storm/ for example).

*It's easy to be compatible with php8 only. It's harder to become compatible with php8 _and_ keep old php7 versions running 😄

mariochacondev:
Alright, I can understand. I just want to be prepared for a project that is due very soon. Thanks for the info

Cheers Mario

mvorisek:
related issues:
[...]

PR #20710, opened 2020-08-25
Rename non-namespaced Attribute class used internally by PHP 8

mvorisek:
| Questions     | Answers
| Branch?       | develop
| Description?  | Starting with PHP 8.0 non-namespaced `Attribute` class is used internally, see https://3v4l.org/9KiqJ
| Type?         | refacto
| Category?     | CO
| BC breaks?    | BC class is provided, but there is a break - overrided Attribute will not override the AttributePs
| Deprecations? | no
| Fixed ticket? | related with https://github.com/PrestaShop/PrestaShop/issues/18547
| How to test?  | CI tests must pass

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20710)


mvorisek:
Please advise better name.

Table name etc. can stay.

For PHP lower than 8.0 we can provide a BC class, thus there wll be no BC break for current versions.

mvorisek:
can someone what else needs to be renamed, error from install:
```
Class '\Attribute' not foundInstalled failed, displaying errors from logs:
```
probably some fixtures data... BC class will stay for users, but this PR itself should fix all related issues in PrestaShop and work without that BC class

PierreRambaud (changes requested):
Thanks for your contribution, but it's too early to consider using PHP8, we are working to make it available with PHP7.4 on the develop branch, and we already know there is too many BC breaks in PHP8 :disappointed:

mvorisek:
> 
> 
> Thanks for your contribution, but it's too early to consider using PHP8, we are working to make it available with PHP7.4 on the develop branch, and we already know there is too many BC breaks in PHP8 😞

I already have install fully working for PHP 8: https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/377695757

please merge this change asap, yesterday, I checked all new coliding classes in PHP 8 and this is the only one - we should merge to `develop` now, I was able to provide full BC

PierreRambaud:
> > Thanks for your contribution, but it's too early to consider using PHP8, we are working to make it available with PHP7.4 on the develop branch, and we already know there is too many BC breaks in PHP8 disappointed
> 
> I already have install fully working for PHP 8: https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/377695757
> 
> please merge this change asap, yesterday, I checked all new coliding classes in PHP 8 and this is the only one - we should merge to `develop` now, I was able to provide full BC

It's a huge BC breaks since the `Attribute` class is no longer working as expected. People that are doing override (because it's still allowed), will not call the right class.

matks:
> > Thanks for your contribution, but it's too early to consider using PHP8, we are working to make it available with PHP7.4 on the develop branch, and we already know there is too many BC breaks in PHP8 😞
> 
> I already have install fully working for PHP 8: https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/377695757
> 
> please merge this change asap, yesterday, I checked all new coliding classes in PHP 8 and this is the only one - we should merge to `develop` now, I was able to provide full BC

Hi, I think there is a misunderstanding.

BC break stands for backward compatible break. To make it short, this means "a change introduced in a codebase that provides previously compatible usages to work anymore following upgrade". I started an article about it here https://github.com/PrestaShop/prestashop.github.io/pull/730 (still draft).

I think you mention "BC" as in "Breaking Change" because "`develop` is not php8 compatible yet".

That being said, back to current topic.

We can see that you are working on "compatibility between `develop` branch and php8.0" and that is amazing 😄  thanks for the hard work !

However this is a complex work. Because, as you have found out, in order to allow this compatibility to happen, multiple changes into PrestaShop `develop` will be needed.

Some of these changes might be BC breaks. In theory, since we follow SemVer, **we should aim for as few as possible BC breaks** however we expect there will be some necessary breaks. Necessary = there is no backward compatible way to do it (or the backward compatible way is very expensive).

**But** we dont aim for `develop` to be php8 compliant **yet**. `develop` will give birth to `1.7.8.x` branch which will bring php7.4 compability thanks to @PierreRambaud work. Later, there will probably be a `1.7.9.x` branch and I guess at this moment we will introduce php8 compatibility. One step at a time.

So **if we merge your PR now** it will introduce a big BC break (renaming the class) although it is not needed yet, because we do not target php8 compatibility for `develop` now.

However, if we target php8 compatibility for `1.7.9.x` branch, then __as soon as 1.7.8.x branch is out develop will become the "future 1.7.9" and we will accept this PR__.

So my answer, to make it short, is "your work is nice however it is too early to be merged because what we merge now will be shipped in 1.7.9".

Obviously we are not against PRs that help php8 compatibility without introducing BC breaks. The issue here "users will pay the cost of a BC break without the benefits"

mvorisek:
> So if we merge your PR now it will introduce a big BC break (renaming the class)

@matks I do not think so - please describe what will be broken

matks:
> > So if we merge your PR now it will introduce a big BC break (renaming the class)
> 
> @matks I do not think so - please describe what will be broken

```
$attributes = Attribute::getAttributes(Context::getContext()->language->id, true);
```
⬇️ 
```
$attributes = AttributePs::getAttributes(Context::getContext()->language->id, true);
```

If a PrestaShop user has built a module for its shop by using [raw override](https://devdocs.prestashop.com/1.7/modules/concepts/overrides/), upon upgrade, the changes brought by your PR will disable his override. Since the class **being used** changes, the override file will modify the behavior of `AttributeCore`, not `AttributePsCore` and then the shop will ignore what used to work.

For example, quite a lot of people use ObjectModel override in order to alter the SQL schema. If you wish to have a new SQL column into Attribute object, it is easy to use an override and ObjectModel will out-of-the-box acknowledge the updated data structure, including inside PrestaShop Webservice. This is especially useful when building a synchronization data module between your shop and an ERP.

So there is a BC break. And we will probably have to merge it indeed, but let's do it at the right moment in the right place 😉  when the code we deliver is aiming for a php8 compatibility.

mvorisek:
@matks https://github.com/PrestaShop/PrestaShop/pull/20710/files#diff-f13cec78081c7bf8b66d98d53d2ea63dR32 really?

matks:
> @matks https://github.com/PrestaShop/PrestaShop/pull/20710/files#diff-f13cec78081c7bf8b66d98d53d2ea63dR32 really?

Yes, really 😄 .

I might be wrong - and I will check that because you make me doubt 😛  - but

## Currently in the code, let's say it's PS 1.7.7.0

I have `AttributeCore` class.

I can modify the behavior of `AttributeCore` using an override.

When I use an override, **everywhere** in the PrestaShop software, __the code from my override is being used, instead of the original file__.

## With your PR, let's say it's PS 1.7.8.0

I have `AttributeCore` class AND `AttributePsCore` classes.

The original software uses `AttributePsCore` now. But `AttributeCore` class is still here, and it extends `AttributePsCore` (and that is good !).

## If I upgrade from 1.7.7.0 to 1.7.8.0

If I modify nothing, my override keeps modifying `AttributeCore` NOT `AttributePsCore`. But modifying `AttributeCore` is not useful anymore because it's not used anymore.

mvorisek:
I updated the description to reflect this, yes, override of Attribute will not affect the new AttributePs

as the nonCore classes are generated, we should update the generator to apply overrides of AttributeCore to AttributePsCore instead, right? This way, we can merge immediatelly, right?

matks:
> as the nonCore classes are generated, we should update the generator to apply overrides of AttributeCore to AttributePsCore instead, right? This way, we can merge immediatelly, right?

Haha 😄  nice idea ! Indeed this seems to work as a plan

What do you think @PierreRambaud ? Would it be possible to trick the generator to avoid the BC break ? 😛

PierreRambaud:
> > as the nonCore classes are generated, we should update the generator to apply overrides of AttributeCore to AttributePsCore instead, right? This way, we can merge immediatelly, right?
> 
> Haha  nice idea ! Indeed this seems to work as a plan
> 
> What do you think @PierreRambaud ? Would it be possible to trick the generator to avoid the BC break ?

It will be still a BC break since the class name change.
Also, the `AttributePsCore` is not a good class name, the Ps have nothing to do in the class name. It should be something like `ProductAttribute` and will be also a BC break.

The PHP8.0 is still in beta, so everything can still change, the Attribute protected class will maybe be changed, we don't know and we should not waste our time debate on something which is not released yet for something which is not started (the 1.7.9.0).
I will reconsider it when the PHP8.0 will be officially released.

mvorisek:
Attribute class will stay because of PHP 8.0 feature freeze, RC expected in September.

I renamed the class from `Attribute` to `AttributePs` as there are other classes like `AttributeGroup` which will be named inconsistently then.

Let's wait until PHP 8.0 RC1 is out and then find a solution which we can merge immediatelly.

PierreRambaud:
> Attribute class will stay because of PHP 8.0 feature freeze, RC expected in September.
> 
> I renamed the class from `Attribute` to `AttributePs` as there are other classes like `AttributeGroup` which will be named inconsistently then.
> 
> Let's wait until PHP 8.0 RC1 is out and then find a solution which we can merge immediatelly.

Exactly, if we rename the Attribute class, we will have to do it for AttributeGroup to be consistent. Let's wait for a 8.0 release and also for the 1.7.9.0 to start. We can close this pull request.

mvorisek:
no, we will keep it, this can be solved for 1.7.8.0 too, as the change will need to be done anyway

PierreRambaud:
> no, we will keep it, this can be solved for 1.7.8.0 too, as the change will need to be done anyway

I'm sorry but I'm against adding breaking changes not needed for the 1.7.8.0. Let's wait for @PrestaShop/prestashop-core-developers feedbacks too

kpodemski:
Hello y'all, is it even possible to have such a change in 1.7.x? Tbh i thought that php 8 is plan for v8 :) I'm not sure if it would be possible to make PrestaShop compatible with php 8 without having more BC breaks. Lets remember about all dependencies, right now people are instantly moving to => () and I'm sure with `match` will be no different 😂

Maybe solution would be to have 1.7 as LTS and 1.8 before v...9? ;) 

PHP 8 is coming this year, it would be good to know what's the plan for this y'all have :)

#notifyme

PierreRambaud:
> Hello y'all, is it even possible to have such a change in 1.7.x? Tbh i thought that php 8 is plan for v8 :) I'm not sure if it would be possible to make PrestaShop compatible with php 8 without having more BC breaks. Lets remember about all dependencies, right now people are instantly moving to => () and I'm sure with `match` will be no different 
> 
> Maybe solution would be to have 1.7 as LTS and 1.8 before v...9? ;)
> 
> PHP 8 is coming this year, it would be good to know what's the plan for this y'all have :)
> 
> #notifyme


AFAIK, if PHP8.0 is released this year, we will support it on the next release. This means, currently the 1.7.8.0 will support PHP 7.4, but, if the 1.7.6.8 is not released yet when the PHP8.0 will be available, we will do our best to support it. 
But (again :sweat_smile:), it's still in beta and many dependencies are (maybe) not compliant with PHP8, we should take in consideration all our dependencies before doing the job and add many unneeded BC breaks.

kpodemski:
@PierreRambaud maybe having PrestaShop compatible wouldn't be so hard, but i'm afraid of minimum PHP version needed ;)

PierreRambaud:
> @PierreRambaud maybe having PrestaShop compatible wouldn't be so hard, but i'm afraid of minimum PHP version needed ;)

This is also a possible problem :sweat_smile:

jderusse:
For the record, this issue prevents users running php 7 to require dependencies (or their own code) using polyfill (list of packages impacted: https://packagist.org/packages/symfony/polyfill-php80/dependents?order_by=downloads)

I sugesst to add "conflict" rule in your `composer.json`

PierreRambaud:
Closed in favor of #24609

mvorisek:
I wonder why a separate PR with original authorship is not preferred.

PR #20720, opened 2020-08-26
Refresh when change combination product.js 

### LINKED ISSUE #14744: "There are not enough products in stock" when change combination with minimal quantity>1
**Describe the bug**
Product with 3 combinations
combination 1 : quantity 24 , minimal quantity 1, default combination
combination 2 : quantity 12 , minimal quantity 6
combination 3 : quantity 4 , minimal quantity 1, 

**To Reproduce**
1. Go to product page
2. Click on combination 2
3.Click on combination 3
4. See error "There are not enough products in stock."

**Screenshots**

**Additional information**
PrestaShop version: 1.7.6.0
PHP version: 7.1

guil182:
| Questions     | Answers
| Branch?       | develop 
| Description?  | On product page, when you change combination from a combination with a minimum quantity > minimum quantity of the new combination, the message "not enough product in stock appear". Solve it.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14744
| How to test?  | Change combination from a combination with a minimum quantity > minimum quantity of the other combination.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20720)

### DIFF
--- a/themes/_core/js/product.js
+++ b/themes/_core/js/product.js
@@ -172,7 +172,7 @@ function updateProduct(event, eventType, updateUrl) {
         ajax: 1,
         action: 'refresh',
         quantity_wanted:
-          eventType === 'updatedProductCombination' ? $quantityWantedInput.attr('min') : $quantityWantedInput.val(),
+          eventType === 'updatedProductCombination' ? 1 : $quantityWantedInput.val(),
       },
       dataType: 'json',
       beforeSend() {


NeOMakinG:
Hello @guil182 ,

Can you complete the description ? It's important for many reasons : QA Testing, assigning an issue to it, knowing what sort of bugs it's been fixing..

Thanks for your contribution.

PierreRambaud (changes requested):
Hi and thanks for your contribution. You need to provide us a complete description and a proper title to understand what we should review. Please could you also create an issue?

guil182:
Hi,
I modify question /answer.
I hope it's ok.
Regards.

PierreRambaud:
> Hi,
> I modify question /answer.
> I hope it's ok.
> Regards.

I'm sorry, you still need to complete the description and others fields :sweat_smile: 
image

davidglezz (approved):

guil182:
Excuse me, I hope it's ok now.

Regards

themes/_core/js/product.js
```diff
@@ -172,7 +172,7 @@ function updateProduct(event, eventType, updateUrl) {
         ajax: 1,
         action: 'refresh',
         quantity_wanted:
-          eventType === 'updatedProductCombination' ? $quantityWantedInput.attr('min') : $quantityWantedInput.val(),
+          eventType === 'updatedProductCombination' ? 1 : $quantityWantedInput.val(),
```
NeOMakinG, line 175: What happen if the combination has a minimum to order different than 1 ?

guil182: L175 is only when you click to a new combination on the product page. The success data give the good minimum quantity of the combination.
And for keyup and keydown, it's not l175.

Where is the bug ? see #14744

davidglezz:
From what I can see, that ajax call request for the new combination with the minimum amount of the current selected combination, so the patch is correct in my opinion. Another option would be to request quantityWanted or better: 
```js
quantity_wanted:
          eventType === 'updatedProductCombination' && $quantityWantedInput.attr('min') === $quantityWantedInput.val() ? 1 : $quantityWantedInput.val(),
```

I think it really is a bug on the PHP side.
If the quantity required is greater than what is available in stock, the generated html is: quantity input=1 but the buy button disabled because the requested quantity is not available. This is unexpected behavior.

themes/_core/js/product.js
davidglezz, line 175:
```suggestion
          eventType === 'updatedProductCombination' && $quantityWantedInput.attr('min') === $quantityWantedInput.val() ? 1 : $quantityWantedInput.val(),
```
I think it is better to keep the amount that the user has selected, except in the case that it is equal to the minimum amount, in which case it is understood that the user wanted the minimum amount of next combination, since we don't know, 1 is the best.

davidglezz (changes requested):

themes/_core/js/product.js
guil182: hum... I'm not agree. For exemple. On a combination with minimum quantity=3 and input = 4. I change to a combinaison with minimum quantity=2 and stock=2. With your change, I have the message "not enough product in stock instead to have the new combinaison with input=2. On combination change, nothing must to be link with the input value I think because it's a kind of new product with it's own minimum quantity and stock.

NeOMakinG: But tbh, the input should be updated to the new combination minimum to order, maybe I'm wrong, but it looks like every time you'll change the combination, it will be updated to 1 even if the combination has a minimum order of 3

guil182: No, we are speaking about the ajax request. The response always have the good minimum quantity of the new combination.

NeOMakinG: Ah yeah, my bad just understood it

I think it's better to reset the amount in the input instead of keeping the actual amount chosen by the user, because it would lead to misunderstanding on some cases, but we can ask our @PrestaShop/prestashop-product-team :)

davidglezz: @guil182  The message "not enough product in stock" is due to a bug on the PHP side.

NeOMakinG: I think there is a PR related to this, from @atomiix

guil182: @davidglezz I'm not sure. For example. I am on a combination with a minimum quantity of 12. I select a combination with a minimum quantity of 2 and a stock of 8.
I will have this request: ajax = 1 & action = refresh & quantity_wanted = 12

the quantity wanted sent is greater than the stock therefore the error message is "normal"

davidglezz: The problem is that there are 2 errors:
1º We agree that request the minimum quantity of the previous combination as the quantity required for the new one is not correct (These are corrected here, your patch is correct in my opinion)
2º The message may be fine, but the response html is not consistent: if there is not enough quantity, it shows a message, sets the quantity input to 1 and the buy button is disabled, but it should be enabled. This error must be fixed in some PHP file

marionf: When changing combination, the input quantity should be = to the minimal quantity defined for the combination selected.
The message "There are not enough products in stock" should be displayed when the quantity input is > to the stock, unless if ordering of out of stock product is allowed

I hope that answers your question, otherwise let me know :)

guil182:
Hi Progi1984, 

Waiting for author. What does that mean ?

Regards

themes/_core/js/product.js
guil182: Hi @davidglezz,

I'm not agree with your suggestion because it's not behaviour we want ( @marionf) ->When changing combination, the input quantity should be = to the minimal quantity defined for the combination selected.

regards

davidglezz (approved):

MatShir:
Waiting for the author is used by Prestashop Team to ping the contributor to let him know that there is still some work to do on the PR.
Maybe it is ready now? @PrestaShop/prestashop-core-developers

guil182:
hum...always the same bug on 1.7.7.1...:/

matthieu-rolland (approved):
LGTM

matthieu-rolland (commented):
The PR needs to be rebased (there's a conflict on `themes/_core/js/product.js` )

PierreRambaud:
Hi @guil182 ,

Thanks!

Issue #20731, opened 2020-08-26
Product_suppliers aren't filled throw webservice


Max84:
#### Describe the bug
Add or update a product throw webservice with id_supplier and supplier_reference filled.
Only product table is filled and product_supplier table isn't filled. This is a problem because database schema isn't repected like a product's update into backoffice.

#### Expected behavior
product_supplier table must be filled too.

#### Additional information
PrestaShop version: 1.7.7
I already did the patch, I'll commit it asap.

matks:
Hi ! this might be related/duplicate of https://github.com/PrestaShop/PrestaShop/pull/20198 and https://github.com/PrestaShop/PrestaShop/issues/20094

Max84:
Hi @matks 
I'm not sure. These issues tell the problem is product table isn't updated. Here, product table is updated and the product supplier isn't updated.

I suggest to update Product addWs() and updateWs() in order to use addSupplierReference() in order to update too product supplier table.

I see it only with webservice, I don't get this issue with BO.

What do you think about?

marionf:
As it concerns the 1.7.7, we need to ensure if it's not a regression compared to 1.7.6
@khouloudbelguith @sarahdib

khouloudbelguith:
Hi @marionf,

I manage to reproduce the issue with PS1.7.6.7.
I tried with this script
```
<prestashop>
<product>
  <id/>
  <id_manufacturer>2</id_manufacturer>
  <id_supplier xlink:href="http://prestashop.local/1767/api/suppliers/1">1</id_supplier>
  <supplier_reference>testing</supplier_reference>
  <state>1</state>
  <id_category_default>2</id_category_default>
  <active>1</active>
  <advanced_stock_management>1</advanced_stock_management>
  <associations>
    <categories>
      <category><id>2</id></category>
    </categories>
  </associations>
  <reference>boom</reference>
  <name>
  <language id="1">product_success_add</language></name>
  <price>100</price>
  <id_shop_default>1</id_shop_default>
  <low_stock_alert>0</low_stock_alert>
  <link_rewrite>abc</link_rewrite>
  <show_price>1</show_price>
  <minimal_quantity>1</minimal_quantity>
  <available_for_order>1</available_for_order>
</product>
</prestashop>
```

In the Table `ps_product` => ok
![image](https://user-images.githubusercontent.com/16067358/91413224-17efbc80-e84b-11ea-9304-fca622f41d25.png)
[...]

Issue #16086, opened 2019-10-23
Sending Supplier ID - Supplier_refrence  with Webservice not record while success results


8ternity:
**Describe the bug**
I have an integration with a product datas import and dropship and they have an issue with sending via Webservices the Supplier ID (already created in the store) and the webservice sent succesfully the post, but in Prestashop the Supplier are not present.

All the other informations pushed worked (Product Title, summary, description, features, size, weight, etc). Only Supplier ID are not recorded.

They send the HTTP request has follow :
`<product>
<id />
<id_supplier >2</id_supplier>
.....

</product>
`
The answer are "Success".

--------------------------------------------------
>Edited by Khouloud

I tried with this script
```
<prestashop>
<product>
  <id/>
  <id_manufacturer>2</id_manufacturer>
  <id_supplier xlink:href="http://prestashop.local/1767/api/suppliers/1">1</id_supplier>
  <supplier_reference>testing</supplier_reference>
  <state>1</state>
  <id_category_default>2</id_category_default>
  <active>1</active>
  <advanced_stock_management>1</advanced_stock_management>
  <associations>
    <categories>
      <category><id>2</id></category>
    </categories>
  </associations>
  <reference>boom</reference>
  <name>
  <language id="1">product_success_add</language></name>
  <price>100</price>
  <id_shop_default>1</id_shop_default>
  <low_stock_alert>0</low_stock_alert>
  <link_rewrite>abc</link_rewrite>
  <show_price>1</show_price>
  <minimal_quantity>1</minimal_quantity>
  <available_for_order>1</available_for_order>
</product>
</prestashop>
```

In the Table `ps_product` => ok
![image](https://user-images.githubusercontent.com/16067358/91413224-17efbc80-e84b-11ea-9304-fca622f41d25.png)
The table `ps_product_supplier` is empty
![image](https://user-images.githubusercontent.com/16067358/91413300-33f35e00-e84b-11ea-8a1b-e9fa950eca55.png)

In the BO => Catalog => Product details page, the supplier is not setected
![image](https://user-images.githubusercontent.com/16067358/91413642-b714b400-e84b-11ea-817d-c0ca200396f8.png)

-----------------------------------------------

**Screenshots**
You can see the table of supplier, they are configured.
![2019-10-23_09h02_15](https://user-images.githubusercontent.com/16780421/67395419-f6131280-f573-11e9-8ced-503234896273.png)

**Additional information**
PrestaShop version: 1.7.5.1
[...]

PR #20736, opened 2020-08-26
Allow Prestashop to fill product_supplier table onto webservice product add or update

Max84:
| Questions     | Answers
| Branch?       | develop
| Description?  | This fix allow an add or an update product's action throw webservice to fill product_supplier table
| Type?         | bug fix
| Category?     | WS
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? |  #20731, #16086
| How to test?  | Try to add or update a product with a supplier throw webservice. Product_supplier table must be filled.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20736)

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -7340,8 +7340,11 @@ class ProductCore extends ObjectModel
     public function addWs($autodate = true, $null_values = false)
     {
         $success = $this->add($autodate, $null_values);
-        if ($success && Configuration::get('PS_SEARCH_INDEXATION')) {
-            Search::indexation(false, $this->id);
+        if ($success) {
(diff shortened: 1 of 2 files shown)


PierreRambaud (changes requested):
I'm really not sure it's the right way to do it.
You're adding a supplier reference only if you add one by default, and you can't add more than one here :thinking: , it should be a dedicated endpoint or add entries to add supplier references. Then add one by default.

Max84:
Thanks for your feedback. I understand the spirit about it, I'll check it.

matks:
Hi, it seems this PR went stale so I close it. If you want to continue the topic, please tell us, we can reopen.

PR #20757, opened 2020-08-27
Save customer message on order creation from BO

### LINKED ISSUE #20686: BO - Orders page - Order Message is not saved after creating a new order from BO
#### Describe the bug

In the BO => Orders => New Order page => when we add an order message => in the order page there is no message displayed
image
image

#### Expected behavior

It was ok with PS1.7.6.7
image

#### Specs
https://github.com/PrestaShop/prestashop-specs/commit/002dd12effcceede6d725d9fa01c5dc75587f279

#### Steps to Reproduce

1. Go to BO => Orders => Add new Order page
2. Add some products to the cart
3. IN the Summary section => add an order message
4. Create the Order
5. see error => no message displayed

#### Additional information

* PrestaShop version: 1.7.7.0beta2build1, 177x
* PHP version: 7.2

sowbiba:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Save customer message when creating an order from the Backoffice
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20686
| How to test?  | 1. Go to BO => Orders => Add new Order page<br>2. Add some products to the cart<br>3. IN the Summary section => add an order message<br>4. Create the Order<br>5. The message given is displayed in the order messages list

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20757)

### DIFF
--- a/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
@@ -144,7 +144,7 @@ final class AddOrderFromBackOfficeHandler implements AddOrderFromBackOfficeHandl
     private function addOrderMessage(Cart $cart, string $orderMessage): void
     {
         if (!Validate::isMessage($orderMessage)) {
-            throw new OrderConstraintException('', OrderConstraintException::INVALID_CUSTOMER_MESSAGE);
+            throw new OrderConstraintException('The order message is invalid', OrderConstraintException::INVALID_CUSTOMER_MESSAGE);
         }
 
         $messageId = null;
--- a/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
@@ -37,12 +37,14 @@ use Customer;
 use Employee;
 use Exception;
 use Language;
+use Message;
 use Module;
 use PrestaShop\PrestaShop\Adapter\ContextStateManager;
 use PrestaShop\PrestaShop\Core\Domain\Order\Command\AddOrderFromBackOfficeCommand;
 use PrestaShop\PrestaShop\Core\Domain\Order\CommandHandler\AddOrderFromBackOfficeHandlerInterface;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
 use PrestaShop\PrestaShop\Core\Domain\Order\ValueObject\OrderId;
+use Validate;
 
 /**
  * @internal
@@ -99,6 +101,8 @@ final class AddOrderFromBackOfficeHandler implements AddOrderFromBackOfficeHandl
         );
 
         try {
+            $this->addOrderMessage($cart, $command->getOrderMessage());
+
             $paymentModule->validateOrder(
                 (int) $cart->id,
                 $command->getOrderStateId(),
@@ -121,6 +125,30 @@ final class AddOrderFromBackOfficeHandler implements AddOrderFromBackOfficeHandl
         }
 
         return new OrderId((int) $paymentModule->currentOrder);
+    }
+
+    /**
+     * Saves customer message and link it to the cart.
+     *
+     * @param Cart $cart
+     * @param string $orderMessage
+     *
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
+    private function addOrderMessage(Cart $cart, string $orderMessage): void
+    {
+        $messageId = null;
+        if ($oldMessage = Message::getMessageByCartId((int) $cart->id)) {
+            $messageId = $oldMessage['id_message'];
+        }
+        $message = new Message((int) $messageId);
+        if (Validate::isMessage($orderMessage)) {
+            $message->message = $orderMessage;
+            $message->id_cart = (int) $cart->id;
+            $message->id_customer = (int) $cart->id_customer;
+            $message->save();
+        }
     }
 
     /**
--- a/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
@@ -42,6 +42,7 @@ use Module;
 use PrestaShop\PrestaShop\Adapter\ContextStateManager;
 use PrestaShop\PrestaShop\Core\Domain\Order\Command\AddOrderFromBackOfficeCommand;
 use PrestaShop\PrestaShop\Core\Domain\Order\CommandHandler\AddOrderFromBackOfficeHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderConstraintException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
 use PrestaShop\PrestaShop\Core\Domain\Order\ValueObject\OrderId;
 use Validate;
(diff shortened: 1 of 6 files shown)


src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
```diff
@@ -123,6 +127,30 @@ public function handle(AddOrderFromBackOfficeCommand $command)
...
+     * @throws \PrestaShopException
+     */
+    private function addOrderMessage(Cart $cart, string $orderMessage): void
+    {
+        $messageId = null;
+        if ($oldMessage = Message::getMessageByCartId((int) $cart->id)) {
+            $messageId = $oldMessage['id_message'];
+        }
+        $message = new Message((int) $messageId);
+        if (Validate::isMessage($orderMessage)) {
```
jolelievre: If the message is not valid then the save will be ignored silently, you should throw some `OrderConstraintException` here with a specific code related to INVALID_MESSAGE so that we can assign it a dedicated translated error message in the controller

jolelievre (commented):
A dedicated exception needs to be thrown when the message is not valid
Also could you add some behat tests:
- one to check if the message is correctly saved
- one to check that exception with the appropriate error code is thrown when using a invalid message

Both tests can be added in `order_from_bo.feature` since the command is already tested in this feature

src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
sowbiba: Done

PierreRambaud: This is the first thing you need to do before continue in this method.
You don't need to waste memory & requests if the message is not valid

```
private function addOrderMessage(Cart $cart, string $orderMessage): void
{
    if (!Validate::isMessage($orderMessage)) {
        throw new OrderContraintException(...);
    }

    //  then you can load the message
}

File: src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php [hunk 2]
```diff
@@ -123,6 +131,33 @@ public function handle(AddOrderFromBackOfficeCommand $command)
...
+     * @throws OrderConstraintException
+     */
+    private function addOrderMessage(Cart $cart, string $orderMessage): void
+    {
+        $messageId = null;
+        if ($oldMessage = Message::getMessageByCartId((int) $cart->id)) {
+            $messageId = $oldMessage['id_message'];
+        }
+        $message = new Message((int) $messageId);
+        if (!Validate::isMessage($orderMessage)) {
```
PierreRambaud: FYI: https://github.com/PrestaShop/PrestaShop/pull/20757#discussion_r478536340

src/Core/Domain/Order/Exception/OrderConstraintException.php
```diff
@@ -31,4 +31,18 @@
...
+     * Used in create order from BO when the customer message is invalid.
+     */
+    const INVALID_CUSTOMER_MESSAGE = 1;
+
+    /**
+     * @param int $code
+     * @param string $message
+     * @param Throwable|null $previous
+     */
+    public function __construct($code = 0, $message = '', Throwable $previous = null)
```
jolelievre: Don't change the regular constructor, since this exception is also used elsewhere
When you throw it you can use a message that describes the problem along with the error code as a second argument

src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
sowbiba: You're right. Fixing it

tests/Integration/Behaviour/Features/Context/Domain/OrderMessageContext.php
```diff
@@ -66,4 +70,50 @@ public function addWithSpecifiedProperties(string $reference)
...
+        $orderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId));
+
+        $messageFound = false;
+        foreach ($orderForViewing->getMessages()->getMessages() as $orderMessageForViewing) {
+            if (0 === strcmp($orderMessageForViewing->getMessage(), $messageContent)) {
+                $messageFound = true;
+            }
+        }
+        if (!$messageFound) {
+            var_dump($orderForViewing->getMessages()->getMessages());
```
jolelievre: Don't forget to clean this `var_dump`

jolelievre (changes requested):

src/Adapter/Order/CommandHandler/AddOrderFromBackOfficeHandler.php
```diff
@@ -135,20 +139,23 @@ public function handle(AddOrderFromBackOfficeCommand $command)
      *
      * @throws \PrestaShopDatabaseException
      * @throws \PrestaShopException
+     * @throws OrderConstraintException
      */
     private function addOrderMessage(Cart $cart, string $orderMessage): void
     {
+        if (!Validate::isMessage($orderMessage)) {
+            throw new OrderConstraintException('', OrderConstraintException::INVALID_CUSTOMER_MESSAGE);
```
jolelievre:
```suggestion
            throw new OrderConstraintException('The order message is invalid', OrderConstraintException::INVALID_CUSTOMER_MESSAGE);
```

jolelievre (approved):
Thank you @sowbiba

PierreRambaud (approved):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1661,6 +1662,13 @@ private function getErrorMessages(Exception $e)
                     'Admin.Orderscustomers.Notification'
                 ),
             ],
+
+            OrderConstraintException::class => [
+                OrderConstraintException::INVALID_CUSTOMER_MESSAGE => $this->trans(
+                    'The order message given is invalid',
```
atomiix, line 1668: @LouiseBonnard There's a bug in PrestonBot right now preventing the detection of new wordings but I think this one is new.

LouiseBonnard: Noted, thanks @atomiix! @sowbiba, why could the order message be invalid? Are there special characters forbidden or length restriction?

LouiseBonnard: I'll update the [specs](https://github.com/PrestaShop/prestashop-specs/commit/002dd12effcceede6d725d9fa01c5dc75587f279) accordingly. ;-)

SD1982:
LGTM Thanks @sowbiba !!

Progi1984:
Thanks @sowbiba & @SD1982

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
sowbiba: Yes there are forbidden special characters like <> and {}

LouiseBonnard: Oops, too bad it was merged, I should have labeled the PR as _waiting for wording_ since the bot is broken. It could have been nice to add a more specified notification and a final dot to the sentence.

LouiseBonnard: @sowbiba, thanks for the reply!

@marionf, I've just updated the specs, you can find it right here: https://github.com/PrestaShop/prestashop-specs/commit/1ea87ae7d70c4f01870e741c93049c37259b8906. ;-)

marionf: Yeah, thank you :kissing_cat:

PR #20775, opened 2020-08-28
Improve classic theme colors, font size and spaces and readability

NeOMakinG:
| Questions     | Answers
| Branch?       | develop
| Description?  | Classic theme needs a bit of improves regarding spacing and colors
| Type?         | improvement
| Category?     | FO
| BC breaks?    | yes
| Deprecations? | no
| How to test?  | Check classic theme colors, spacing, typo (contact me in order to get a complete list of things to be done)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20775)

⚠️ On the paper, it looks like I didn't introduced any BC Breaks because:
- No class have been removed
- No css have been really removed

But changing a lot of styles like this will possibly break some child themes, here is a list of "risky" changes I've done:
- I fixed a bug about aligns in the header, by removing floats and clearfix styles, and replacing them by using flexboxes, the responsive alignments has been modified too to make it way cleaner.
- I added some style to the newsletter block, which means that child themes adding their own styles will probably need to adapt their needs (as our block was pretty not stylised at all)
- General checkout/card paddings of informations card has been modified, it could lead to changes on child layouts
- Button disabled style changed, it's now really less visible (as it's a disabled button)
- Checkout page really changed, paddings, font-sizes, background behind steps numbers, forms, checkbox/radios border width and colors
- Checkboxes style on checkout has been refactored into a position absolute instead of using a display: inline-block;

These changes are overall needed to fix alignments bugs or strange behaviors, this is a choice

These looks like not being REAL breaking changes, but it could be impacting child themes, so please, keep the BC label! :warning:

### Before

image

image

### After

Header aligns are now well done

image

Font weight, sizes, spacing and newsletter block have been improved 

image

### Before

image

### After

image

### Before

image

image

### After

General font sizes and weight has been improved

image

image

### Before 

image

### After

image


NeOMakinG:
- Slider title => fw-700
- Slider et categories title > subtitle 1 => 20px & fw-600 (Regarder si ca correspond)
- Slider text => 16px
- Labels product => row
- Content bottom => change default style 
- Checkout => leave separators and change total fw-700
- Blockreassurance kick fill of svgs and check separators
- checkout payment checkbox alignment
- Change disable gray (background: #F6F6F6, text: #DDDDDD)
- Radio => border-width 1px, checkbox => same color as radio border one (color: #979797)

themes/classic/_dev/package.json
```diff
@@ -34,6 +34,7 @@
     "jquery": "^3.4.1",
     "jquery-touchswipe": "^1.6",
     "jquery.browser": "^0.1.0",
+    "manrope": "^2.0.0",
     "material-design-icons-iconfont": "^5.0.1",
     "node-sass": "^4.13.1",
     "notosans-fontface": "~1.0.1",
```
Progi1984: We keep the old font ?

NeOMakinG: I think we can remove it

Progi1984: Do it ;)

Progi1984 (approved):

PierreRambaud (approved):

SimonGrn:
That looks pretty slick ! Congrats

MatShir:
What a great improvement!
I have a question for you guys @NeOMakinG @prestascott, why the checkout has a flat design and the rest has kinda 3D design with shadows?
IMO I prefer flat design :sweat_smile:

NeOMakinG:
> What a great improvement!
> I have a question for you guys @NeOMakinG @prestascott, why the checkout has a flat design and the rest has kind a 3D design with shadows ?

@prestascott did a new template, to avoid too much breaking changes, we decided to implement what we could implement first, maybe this is why it looks a bit strange to you

prestascott:
Thanks for your feedback @MatShir!

After this PR I will update the Theme Classic design with shadows harmonizations so it will looks more consistent.

atomiix:
Thanks @NeOMakinG

Issue #19935, opened 2020-06-25
Allow a cart rule not to be compatible with all other cart rules (old and new)


superskyman100:
The problem is that when someone creates a new cart rule, by default, without ticking the compatibility check box, all the cart rules you have previously created are compatible with this new cart rule you are setting up.
So every time you create a new cart rule you have to manually check the compatibility box, and then remove all the previous cart rules.

This is extremely frustrating and time wasting for shops that have the need to create new discount codes on a weekly or daily basis. It’s a lot of work.

The default behaviour, in my view, should be that if you create a cart rule and you don’t check the compatibility box, it is not compatible with all the others you have previously created. And if you check it, then you should be obliged to manually add the ones you want to be compatible.

florine2623:
Hi @superskyman100 ,
You need to delete the content of ps_cart_rule_combination and none of the cart rule will be compatible with each other. i just did the test and it works correctly.
Please check and let me know.

superskyman100:
Have you tested what happened in case a new cart rule is created ?

We are not speaking about static situation, but dynamic one. We need newly
created discounts (we will have hundreds per days due to reminder cart
modules) to be incompatible by default.

BUT we also need some existing cart rule to stay compatible with others....

Deleting all ps_cart_rule combinaison won't solve the (dynamic) issue....
Right ?

florine2623:
Then you must add the restriction directly manually from the BO
If you have specific cases in your shop, you will need to contact an agency to develop a special tool for you.
Or ou need to contact the developer of the tool you are using directly from your Addons account to request a new feature. The GitHub space is for PrestaShop core software, not external modules.
> Deleting all ps_cart_rule combinaison won't solve the (dynamic) issue....

No it won't.

superskyman100:
Well we are speaking about BUG of the core software : if we uncheck all
cart rules compatibility (as shown up in several screenshots), they are
still compatible >>  there is obviously something wrong.

This BUG creates problem when we use Prestashop, with or without modules.
Modules that create discounts are not the problem. The problem is that we
cannot use the core prestahop functionalities as they supposed to work,
cause it's not working.

Just look the forums, there is plenty of reports about this problem.....

florine2623:
I can't reproduce your issue. When the cart rules are uncombinable this is the issue I have in the FO :
[...]

PR #20776, opened 2020-08-28
WIP: Improve multi invoices shipping, discounts and computing

### LINKED ISSUE #20409: Order Multi Invoice Shipping amounts
#### Describe the bug

In order with multi invoices, the shipping total takes into account only the first invoice and does not take into account the other ones.

2 shipping cost

Here both products have a shipping cost of 8.40

Resulting with a wrong total and a wrong shipping cost in the total summary.

#### Expected behavior

When the merchant adds a new product into a new invoice from the order back-office page with the shipping cost, the shipping total displays the sum of shipping costs from the different invoices.

A new line in the list of carriers is added with the corresponding shipping cost and carrier.
If the carrier is updated it updates all carrier in the carrier tab and recalculate the shipping cost (an order can technically have only one carrier).

#### Steps to Reproduce

1. Create an order
2. Generate an Invoice
3. Add a product into a new invoice
4. See the total summary

**Screenshots**

..

#### Additional information

* PrestaShop version: N/A
* PHP version: N/A

### LINKED ISSUE #20778: BO - Orders page - Multi-invoice issue - Discount is applied in all invoices
#### Describe the bug

In the BO > Order details page, when we have a multi-invoices, adding a discount to a selected invoice, it will be applied in all invoices.

#### Steps to Reproduce

1. You have already an order with multi-invoices (Product1 => IN0007) and (Product2 => IN0008)
2. Add a discount 100% for IN0007 =100%
3. The discount is applied in all invoices

**Screenshots**
After
image
Before
image

#### Additional information

* PrestaShop version: 177x
* PHP version: 7.2

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When an order had several invoices, the shipping used to be handled only on first invoice The next ones had necessarily 0 this was hard coded but resulted in many inconsistent behaviours Now each invoice has independent shipping and discount management and Order is the addition of all sub invoices This leads to more consistent values, but there is still the possibility to have new invoices with shipping discount (which works more logically now as well) A few behat scenarii have been added to validate all these behaviours<br />It also fixes the edition of an OrderDetail's invoice, and then updates all invoices accordingly, along with checking that it doesn't create duplicate entries in an invoice (error message has been improved when dealing with invoices)
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20409 and #20778.
| How to test?  | Modify an order so that it contains several invoices, each invoice should have its own shipping cost and the Order reflects it in its summary (addition of all subtotals are consistent). Also you can add discount to specific invoice, and modify an OrderDetail's invoice

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20776)

NOTE: this PR is blocked for now, as suggested by @eternoendless I will try to implement multi-cart to manage orders with multi-invoices

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -2228,10 +2228,10 @@ class CartCore extends ObjectModel
                 }
             }
         }
-        if (!empty($cartRules) && null !== $id_order_invoice) {
+        if (!empty($cartRules) && !empty($id_order_invoice)) {
             foreach ($cartRules as $key => $value) {
                 $cartRule = $cartRules[$key];
-                if (isset($cartRule['id_order_invoice']) && (int) $id_order_invoice !== (int) $cartRule['id_order_invoice']) {
+                if (!empty($cartRule['id_order_invoice']) && (int) $id_order_invoice !== (int) $cartRule['id_order_invoice']) {
                     unset($cartRules[$key]);
                 }
             }
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1892,6 +1892,7 @@ class CartCore extends ObjectModel
      * @param array $products
      * @param int $id_carrier
      * @param bool $use_cache @deprecated
+     * @param int|null $id_order_invoice
      *
      * @return float Order total
      *
@@ -1902,7 +1903,8 @@ class CartCore extends ObjectModel
         $type = Cart::BOTH,
         $products = null,
         $id_carrier = null,
-        $use_cache = false
+        $use_cache = false,
+        $id_order_invoice = null
     ) {
         if ((int) $id_carrier <= 0) {
             $id_carrier = null;
@@ -1964,7 +1966,7 @@ class CartCore extends ObjectModel
         // CART CALCULATION
         $cartRules = [];
         if (in_array($type, [Cart::BOTH, Cart::BOTH_WITHOUT_SHIPPING, Cart::ONLY_DISCOUNTS])) {
-            $cartRules = $this->getTotalCalculationCartRules($type, $type == Cart::BOTH);
+            $cartRules = $this->getTotalCalculationCartRules($type, $type == Cart::BOTH, $id_order_invoice);
         }
 
         $computePrecision = Context::getContext()->getComputingPrecision();
@@ -2127,10 +2129,11 @@ class CartCore extends ObjectModel
     /**
      * @param $type
      * @param $withShipping
+     * @param int|null $id_order_invoice
      *
      * @return array
      */
-    protected function getTotalCalculationCartRules($type, $withShipping)
+    protected function getTotalCalculationCartRules($type, $withShipping, $id_order_invoice = null)
     {
         if ($withShipping || $type == Cart::ONLY_DISCOUNTS) {
             $cartRules = $this->getCartRules(CartRule::FILTER_ACTION_ALL);
@@ -2147,6 +2150,14 @@ class CartCore extends ObjectModel
 
                 if (!$alreadyAddedCartRule) {
                     $cartRules[] = $cartRuleCandidate;
+                }
+            }
+        }
+        if (!empty($cartRules) && null !== $id_order_invoice) {
+            foreach ($cartRules as $key => $value) {
+                $cartRule = $cartRules[$key];
+                if (isset($cartRule['id_order_invoice']) && (int) $id_order_invoice !== (int) $cartRule['id_order_invoice']) {
+                    unset($cartRules[$key]);
                 }
             }
         }
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1945,17 +1945,17 @@ class CartCore extends ObjectModel
             $type = Cart::BOTH_WITHOUT_SHIPPING;
         }
 
-        // Manage order total computation (ignore if we are already computing a specific invoice)
-        if (empty($id_order_invoice)) {
-            // When cart is associated to order the computation may be different, especially for multi invoices orders
-            $orderId = Order::getIdByCartId((int) $this->id);
-            if (!empty($orderId)) {
-                $order = new Order($orderId);
-                // If cart is already assigned to an order we get product from the order as it contains more details
-                if (null === $products) {
-                    $products = $order->getCartProducts();
-                }
+        // When cart is associated to order the computation may be different, especially for multi invoices orders
+        $orderId = Order::getIdByCartId((int) $this->id);
+        if (!empty($orderId)) {
+            $order = new Order($orderId);
+            // If cart is already assigned to an order we get product from the order as it contains more details
+            if (null === $products) {
+                $products = $order->getCartProducts();
+            }
 
+            // If we are already computing an invoice total we don't check for multi invoice
+            if (empty($id_order_invoice)) {
                 $orderInvoices = $order->getInvoicesCollection();
                 // If order has multi invoices we compute each one individually and add the results
                 if ($orderInvoices->count() > 1) {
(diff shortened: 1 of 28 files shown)


classes/Cart.php
```diff
@@ -1943,11 +1945,45 @@ public function getOrderTotal(
             $type = Cart::BOTH_WITHOUT_SHIPPING;
         }
 
-        // filter products
+        // When cart is associated to order the computation may be different, especially for multi invoices orders
```
jolelievre, line 1948: I decided to modify the `Cart` class directly because too much of the logic was implemented inside the handler The problem being that some modules, and especially Payment modules use the `Cart::getOrderTotal` function to compute order amounts and compare them to validate their payments

So if the whole multi-invoice logic was absent from this legacy class, it wouldn't be able to compute correctly the values wouldn't match and therefore the payment would not be validated

Though it might seem scary to modify this method, it is supported by a few arguments:
- it only modifies the behaviour when cart has an associated Order, or an invoice id is provided (which is used internally by the handler)
- it doesn't change the behaviour of a regular cart in FO
- both cart computations and order computations are supported by more and more behat tests (which are all green)
- it is anyway necessary 😅

jolelievre:
This PR will likely have known issues regarding how we handle OrderCarrier:
- when we have shipping discount should the OrderCarrier have the real total as a value or the discounted value? this needs to be specified by the @PrestaShop/prestashop-product-team (for now it uses the real total, if it were to change it can be handled in a separate PR)
- this PR doesn't handle the carrier changing in the Order page: when the carrier is changed so are the shipping fees, the OrderAmountUpdater should then be called to update all (this should be handled in a separate PR once this one is merged)

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -344,79 +345,15 @@ private function addProductToCart(Cart $cart, Product $product, $combination, $q
...
 
-    /**
-     * @param AddProductToOrderCommand $command
-     * @param Order $order
-     * @param Cart $cart
-     * @param array $products
-     *
-     * @return OrderInvoice|null
-     */
-    private function createNewOrEditExistingInvoice(
```
jolelievre, line 355: The update of both created and updated invoice was redundant, since the OrderAmountUpdater manages it all much better So a lot of code can be cleared here:
- no more computations for new invoice
- no more `updateExistingInvoice` method
- therefore even `createNewOrEditExistingInvoice` becomes useless and is managed directly in the `handle` method which makes it simpler to understand

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -429,72 +366,64 @@ private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping, arr
...
      *
-     * @return OrderInvoice
+     * @throws OrderException
      */
-    private function updateExistingInvoice($orderInvoiceId, Cart $cart, array $newProducts)
+    private function addFreeShippingDiscount(Order $order, Cart $cart, OrderInvoice $orderInvoice): void
     {
-        $invoice = new OrderInvoice($orderInvoiceId);
+        $defaultLangId = (int) Configuration::get('PS_LANG_DEFAULT');
+        $cartRuleObj = new CartRule();
```
matks, line 392: Maybe we could use a factory here, so we can reuse this logic. I have already seen it multiple times in PS codebase.
```suggestion
        $cartRuleObj = CartRule::createFreeShippingCartRule();
```

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -429,72 +366,64 @@ private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping, arr
...
+            if (!$cartRuleObj->add()) {
+                throw new OrderException('An error occurred during the CartRule creation');
+            }
+        } catch (PrestaShopException $e) {
+            throw new OrderException('An error occurred during the CartRule creation', 0, $e);
+        }
 
-        return $invoice;
+        try {
+            // It's important to add the cart rule to the cart Or it will be ignored when cart performs AutoRemove AddAdd
```
matks, line 420: `AutoRemove AddAdd` ? 🤔

classes/CartRule.php
```diff
@@ -96,6 +97,7 @@ class CartRuleCore extends ObjectModel
         'multilang' => true,
         'fields' => [
             'id_customer' => ['type' => self::TYPE_INT, 'validate' => 'isUnsignedId'],
+            'id_order_invoice' => ['type' => self::TYPE_INT, 'validate' => 'isUnsignedId'],
```
matks, line 100: This is frightening me. Introducing a new field and SQL column into the amazingly complex data model of Cart Rules. We know there are some complex SQL queries doing massive checks (see https://github.com/PrestaShop/PrestaShop/issues/19935) impacted by number of rows.

This change will not be beta-tested (unless there is a beta 3). I think it's too risky to be releases with PS 1.7.7.0 .

matks (changes requested):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -344,79 +345,15 @@ private function addProductToCart(Cart $cart, Product $product, $combination, $q
...
     /**
      * @param Order $order
      * @param Cart $cart
-     * @param bool $isFreeShipping
      * @param array $newProducts
-     * @param
+     * @param bool|null $withFreeShipping
      */
-    private function createNewInvoice(Order $order, Cart $cart, $isFreeShipping, array $newProducts)
+    private function createNewInvoice(Order $order, Cart $cart, array $newProducts, ?bool $withFreeShipping)
```
atomiix, line 354:
```suggestion
    private function createNewInvoice(Order $order, Cart $cart, array $newProducts, bool $withFreeShipping = false)
```

classes/CartRule.php
jolelievre: I get why you are not reassured ^^ However the risk here is really contained because this extra field is only used internally for cart rules that are created programmatically by the core handlers. And it is only used in the multi-invoice "mode", so 95% of the time it won't even be used by merchants.

I didn't change any of the existing queries so it can't impact them, this field is only used to perform a filtering in the PHP side (which prevented from modifying the queries).

However, currently we have a regression because it's not possible any more to create discounts on a specific invoice (see this issue https://github.com/PrestaShop/PrestaShop/issues/20778 )

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
jolelievre: I had similar questions indeed, but not specifically on the Free shipping cart rule because I'm not sure there are that many places where we specifically need this (usually we create a cart rule based on a switch type: percent, amount, free shipping)

However I though of having a protected function to easily create a CartRule that is specifically related to an order/invoice, as it needs to have some specific settings (related to customer, limited number of uses, not visible, ...) and could be common with the `AddCartRuleToOrderHandler`

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
jolelievre: Yes this a legacy mechanism that is internal to the Cart, by default when you call `$cart->getCartRules()` it performs an automatic add of CartRules that may be associated to it, as long as they meet the conditions

This is this mechanism that will automatically add a discount or a free shipping when you reach a minimum amount in your cart for example. Similarly there is an autoRemove feature which is called when you remove a product from the cart or update its quantity.

I didn't know of these two mechanisms before either, but they are deeply connected to the cart rules behaviour, and can not be easily disabled. The autoAdd can be ignored if you force to false when calling `getCartRules` but it doesn't prevent any other module or hook to call the same method which will automatically add them later. And the `autoRemove` is simply just not possible to be disabled.

So instead of fighting against these two mechanisms it's better to go with them, this is what has been done in the recent fixes to Order cart rules, we don't compute them manually (which resulted in many computations errors), instead we rely on the cart intelligence to add the good cart rules, compute them correctly as it would do in the front and only then we synchronize OrderCartRules based on the computed cart rules.

So we "just" need to correctly initialize the CartRules and add them to the cart to profit from this mechanism.

jolelievre:
Closed because issues need to be specified, we need to clarify the expected behaviour for multi invoices regarding shipping fees management and discounts The main problem being that we only have ONE Cart for multiple OrderInvoice so we have to manage weird computations to split the total among several invoices

One possible (and clean) solution would be to have one dedicated cart per invoice although it requires quite some changes in the database and model But we need to decide:
- does each invoice have its dedicated shipping cost
- can each invoice have a different Carrier
- this means each invoice must have its Delivery slip
- ... probably other issues ^^

PR #20791, opened 2020-08-31
upload function classes/Uploader.php

Dardena7:
It looks like setting a custom destination is impossible because the file will have no name. And if I put the filename in the $dest variable, function is_dir will be false.
Maybe I'm using it wrong ??

| Questions     | Answers
| Description?  | Please be specific when describing the PR. <br> Every detail helps: versions, browser/server configuration, specific module/theme, etc. Feel free to add more information below this table.
| Category?     | FO / BO / CO / IN / WS / TE / LO / ME / PM
| BC breaks?    | yes / no
| Deprecations? | yes / no

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20791)

### DIFF
--- a/classes/Uploader.php
+++ b/classes/Uploader.php
@@ -265,7 +265,7 @@ class UploaderCore
     {
         if ($this->validate($file)) {
             if (isset($dest) && is_dir($dest)) {
-                $filePath = $dest;
+                $filePath = $dest . $file['name'];
             } else {
                 $filePath = $this->getFilePath(isset($dest) ? $dest : $file['name']);
             }


classes/Uploader.php
```diff
@@ -265,7 +265,7 @@ public function upload($file, $dest = null)
     {
         if ($this->validate($file)) {
             if (isset($dest) && is_dir($dest)) {
-                $filePath = $dest;
+                $filePath = $dest . $file['name'];
```
matks, line 268:
```suggestion
                $filePath = $dest . $file['name'];
```

Progi1984:
@Dardena7  Thank you for your contribution. :hugs: 

Could you fill the PR table ?

Thanks :+1:

PR #20828, opened 2020-09-02
Introduce ManufacturerIdInterface for ValueObjects

jolelievre:
| Questions     | Answers
| Branch?       | develop
| Description?  | Introduce `ManufacturerIdInterface ` interface to be used as a transverse type instead of `ManufacturerId` which is restricted to more accurate cases
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | Travis tests green

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20828)

### DIFF
--- a/src/Core/Domain/Manufacturer/ValueObject/ManufacturerId.php
+++ b/src/Core/Domain/Manufacturer/ValueObject/ManufacturerId.php
@@ -31,7 +31,7 @@ use PrestaShop\PrestaShop\Core\Domain\Manufacturer\Exception\ManufacturerConstra
 /**
  * Provides manufacturer id
  */
-class ManufacturerId
+class ManufacturerId extends ManufacturerRelation
 {
     /**
      * @var int
@@ -46,15 +46,7 @@ class ManufacturerId
     public function __construct($id)
     {
         $this->assertIsIntegerGreaterThanZero($id);
-        $this->id = $id;
-    }
-
-    /**
-     * @return int
-     */
-    public function getValue()
-    {
-        return $this->id;
+        parent::__construct((int) $id);
     }
 
     /**
--- a/src/Core/Domain/Manufacturer/ValueObject/ManufacturerId.php
+++ b/src/Core/Domain/Manufacturer/ValueObject/ManufacturerId.php
@@ -50,7 +50,7 @@ class ManufacturerId implements ManufacturerIdInterface
     }
 
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
      */
     public function getValue()
     {
--- a/src/Core/Domain/Manufacturer/ValueObject/ManufacturerId.php
+++ b/src/Core/Domain/Manufacturer/ValueObject/ManufacturerId.php
@@ -31,7 +31,7 @@ use PrestaShop\PrestaShop\Core\Domain\Manufacturer\Exception\ManufacturerConstra
 /**
  * Provides manufacturer id
  */
-class ManufacturerId extends ManufacturerRelation
+class ManufacturerId implements ManufacturerIdInterface
 {
     /**
      * @var int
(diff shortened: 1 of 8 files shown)


Progi1984 (dismissed):

PierreRambaud (dismissed):

Progi1984 (dismissed):

src/Core/Domain/Manufacturer/ValueObject/ManufacturerRelation.php
```diff
@@ -0,0 +1,60 @@
...
+namespace PrestaShop\PrestaShop\Core\Domain\Manufacturer\ValueObject;
+
+/**
+ * Class ManufacturerRelation represents a relation with a Manufacturer, it is not usable as
+ * a stand alone class though, you need to to use one of its children.
+ *
+ * @see ManufacturerId
+ * @see NoManufacturer
+ */
+abstract class ManufacturerRelation
```
eternoendless: I still think this should be called `OptionalManufacturerId`. An ID is not a relation but an identification.

src/Core/Domain/Manufacturer/ValueObject/ManufacturerRelation.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    /**
+     * @var int
+     */
+    protected $id;
+
+    /**
+     * @param int $id
+     */
+    public function __construct(int $id)
```
eternoendless: This one should only exist in `ManufacturerId`

src/Core/Domain/Manufacturer/ValueObject/ManufacturerRelation.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $this->id = $id;
+    }
+
+    /**
+     * @return int
+     */
+    public function getValue()
```
eternoendless: This one should be abstract

src/Core/Domain/Manufacturer/ValueObject/ManufacturerRelation.php
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @see ManufacturerId
+ * @see NoManufacturer
+ */
+abstract class ManufacturerRelation
+{
+    /**
+     * @var int
+     */
+    protected $id;
```
eternoendless: IMO this should be removed

src/Core/Domain/Manufacturer/ValueObject/ManufacturerRelation.php
eternoendless: This _could_ become an interface as per my comments below

src/Core/Domain/Manufacturer/ValueObject/NoManufacturer.php
```diff
@@ -47,6 +47,6 @@ class NoManufacturerId extends ManufacturerId
      */
     public function __construct()
     {
-        $this->id = static::NO_MANUFACTURER_ID;
+        parent::__construct(static::NO_MANUFACTURER_ID);
```
eternoendless: You should implement `getValue()` instead

src/Core/Domain/Manufacturer/ValueObject/ManufacturerId.php
```diff
@@ -46,15 +46,7 @@ class ManufacturerId
...
-        $this->id = $id;
-    }
-
-    /**
-     * @return int
-     */
-    public function getValue()
-    {
-        return $this->id;
+        parent::__construct((int) $id);
```
zuk3975: you should not cast `int` here I suppose.
VO should not allow invalid data, but it shouldn't be fixing it.

atomiix (approved):

zuk3975:
soo, we have an interface for a value object now :thinking:

matks:
> soo, we have an interface for a value object now 🤔

I'm puzzle too 😄  but let's give it a chance!

matthieu-rolland (approved):

Progi1984 (approved):

matks:
Would be awesome to have also @eternoendless to confirm we are following the right track 😄

matks:
Since @eternoendless is off for 1 week, I believe we can safely merge this PR as it's small and change our course later if we found out an issue

PR #20830, opened 2020-09-02
Introduce SetAssociatedProductAttachmentsCommand and AssociateProductAttachmentCommand

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | It is a part of product page migration. Introduces SetAssociatedProductAttachmentsCommand and AssociateProductAttachmentCommand.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | travis :heavy_check_mark: 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20830)


src/Core/Domain/Attachment/Command/AddAttachmentCommand.php
```diff
@@ -34,32 +34,32 @@ class AddAttachmentCommand
     /**
      * @var string|null
      */
-    private $pathName;
+    protected $pathName;
```
Progi1984: Isn't a BC to change the visibility of a variable ?!

Progi1984: Ping @PrestaShop/prestashop-core-developers

zuk3975: It would be indeed if I make it private from protected (vice versa).
Now I doubt it.
BUT, I am not sure if its okay to extend the command with another one, maybe I should duplicate all the properties instead?
ping @matks, @jolelievre

jolelievre: Regardless of the authorized extend or not, I don't think you need to change the visibility of these fields Since you can modify them via the public method setFileInformation anyway

src/Adapter/AbstractObjectModelProvider.php
```diff
@@ -64,4 +64,27 @@ protected function getObjectModel(int $id, string $objectModelClass, string $exc
...
+
+    /**
+     * @param int $id
+     * @param string $objectModelName
+     * @param string $exceptionClass
+     * @param int $errorCode
+     *
+     * @throws CoreException
+     */
+    protected function assertObjectModelExists(int $id, string $objectModelName, string $exceptionClass, int $errorCode = 0): void
```
jolelievre:
```suggestion
    protected function assertObjectModelExists(int $id, string $objectTableName, string $exceptionClass, int $errorCode = 0): void
```

We need the table name here, so it's easier to understand if the parameter expresses this

src/Adapter/Product/ProductAttachmentUpdater.php
```diff
@@ -0,0 +1,134 @@
...
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Product;
```
jolelievre, line 29: Since we're likely to have more and more updater classes maybe we should have dedicated namespace to store them into, or it will quickly become messy (and move ProductAdapter into it as well)

jolelievre (commented):
Two small comments and the code will be good
But some behat tests should be added though

src/Adapter/Product/ProductAttachmentUpdater.php
zuk3975: any suggestions? :P

src/Adapter/AbstractObjectModelProvider.php
zuk3975: > But some behat tests should be added though

I've tried introducing behats for attachments, but failed.
Add/Update Attachment handler uses move_uploaded_file inside, which must be uploaded by HTTP POST, so there is no way (at least I don't know any) to test it

zuk3975:
> But some behat tests should be added though

I've tried introducing behats for attachments, but failed.
Add/Update Attachment handler uses move_uploaded_file inside, which must be uploaded by HTTP POST, so there is no way (at least I don't know any) to test it

src/Adapter/Attachment/AttachmentProvider.php
```diff
@@ -0,0 +1,71 @@
...
+use PrestaShop\PrestaShop\Core\Exception\CoreException;
+
+/**
+ * Reusable methods to get Attachment object model
+ */
+class AttachmentProvider extends AbstractObjectModelProvider
+{
+    /**
+     * @param AttachmentId $attachmentId
+     * @param string $exceptionClass
```
Progi1984:
```suggestion
```

src/PrestaShopBundle/Resources/config/services/adapter/attachment.yml
```diff
@@ -57,3 +57,12 @@ services:
...
       - { name: 'tactician.handler', command: 'PrestaShop\PrestaShop\Core\Domain\Attachment\Command\AddAttachmentCommand' }
+
+  prestashop.adapter.attachment.attachment_provider:
+    class: PrestaShop\PrestaShop\Adapter\Attachment\AttachmentProvider
+
+  prestashop.adapter.attachment.query_handler.get_attachment_information_list_handler:
+    arguments:
+      - '@prestashop.adapter.data_provider.attachment'
+    tags:
+      - { name: 'tactician.handler', command: 'PrestaShop\PrestaShop\Core\Domain\Attachment\Query\GetAttachmentInformationList' }
```
Progi1984:
```suggestion
      - name: 'tactician.handler'
        command: 'PrestaShop\PrestaShop\Core\Domain\Attachment\Query\GetAttachmentInformationList'
```

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateProductAttachmentsFeatureContext.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * @When I add new attachment :path to product :productReference with following details:
+     *
+     * @param string $path
+     * @param string $productReference
+     * @param TableNode $table
+     */
+    public function addProductAttachment(string $path, string $productReference, TableNode $table): void
+    {
+        //@todo: finish up
```
Progi1984: To finish !

tests/Integration/Behaviour/Features/Scenario/Product/update_attachments.feature
```diff
@@ -0,0 +1,25 @@
...
+      | description | en-US:puffin photo nr1 |
+      | name        | en-US:puffin           |
+      | file_name   | app_icon.png           |
+    Then attachment "att1" should have following properties:
+      | description | en-US:puffin photo nr1 |
+      | name        | en-US:puffin           |
+      | file_name   | app_icon.png           |
+      | mime        | image/png              |
+      | size        | 19187                  |
+    When I associate attachment "att1" with product product1
```
Progi1984, line 35: No `Then` ?

classes/Attachment.php
```diff
@@ -173,13 +173,24 @@ public static function deleteProductAttachments($idProduct)
...
+        return static::associateProductAttachment((int) $idProduct, (int) $this->id);
+    }
+
+    /**
+     * @param int $productId
+     * @param int $attachmentId
+     *
+     * @return bool true if success
+     */
+    public static function associateProductAttachment(int $productId, int $attachmentId): bool
```
Progi1984, line 186: A `static` :scream:  ?

zuk3975: yeah, I don't want to load full entity for association. At least if i make it public, IMO I should move this method to Product, because it makes much more sense to have Product->addAttachment, than Attachment->addProduct. I choose to leave this here as Product has too many methods already :smile:

tests/Integration/Behaviour/Features/Scenario/Product/update_attachments.feature
zuk3975: on the way :smile:

classes/Attachment.php
Progi1984: :ok_hand:

src/Core/Domain/Attachment/Command/AddAttachmentCommand.php
PierreRambaud: It's not a BC, it won't break anything.

classes/Attachment.php
```diff
@@ -173,13 +173,24 @@ public static function deleteProductAttachments($idProduct)
...
     {
         $res = Db::getInstance()->execute('
 			INSERT INTO ' . _DB_PREFIX_ . 'product_attachment
 				(id_attachment, id_product) VALUES
-				(' . (int) $this->id . ', ' . (int) $idProduct . ')');
+				(' . $attachmentId . ', ' . $productId . ')');
```
matks, line 191: I think there is no `declare strict types` in this class (because adding it would be a BC break).

So is `(int $productId, int $attachmentId): bool` enough to ensure these are integers and consequently secure the query against SQL injection ?

zuk3975: A string will be converted to integer. (If it is a text php will throw TypeError). So I suppose it is enough

tests/Integration/Behaviour/Features/Context/CommonFeatureContext.php
```diff
@@ -100,6 +100,18 @@ public static function getContainer()
         return static::$kernel->getContainer();
     }
 
+    /**
+     * @AfterFeature @clear-downloads-after-feature
+     */
+    public static function clearDownloads(): void
```
matks, line 106: Woah this thing can be dangerous 😄 

Is `_PS_DOWNLOAD_DIR_` used for Behat tests different from regular `_PS_DOWNLOAD_DIR_` used for running the shop in localhost ? (for example is there a `test/download` for tests and `download` for prod)

tests/Integration/Behaviour/Features/Context/Domain/AttachmentFeatureContext.php
```diff
@@ -0,0 +1,107 @@
...
+
+class AttachmentFeatureContext extends AbstractDomainFeatureContext
+{
+    /**
+     * @When I add new attachment :reference with following properties:
+     *
+     * @param string $reference
+     * @param TableNode $tableNode
+     */
+    public function addAttachment(string $reference, TableNode $tableNode): void
```
matks, line 45: Suggestion: 
```
public function addAttachment(...)
{
    $this->createDummyFileInDownloadDirectory(...);

    $attachment = new Attachment();
    ...
}
```

Basically it would be nice to wrap this "create dummy download file" into a reusable function 😉

matks (dismissed):

tests/Integration/Behaviour/Features/Context/CommonFeatureContext.php
zuk3975: yes its the same. I could try redefining it in case its _PS_IN_TEST_ mode :thinking:

src/Adapter/AbstractObjectModelProvider.php
jolelievre: I believe you can drop this one, since it's in the Repository abstract class and it was not used anywhere else

src/Adapter/AbstractObjectModelRepository.php
```diff
@@ -34,6 +34,29 @@
...
 {
+    /**
+     * @param int $id
+     * @param string $objectModelName
+     * @param string $exceptionClass
+     * @param int $errorCode
+     *
+     * @throws CoreException
+     */
+    protected function assertObjectModelExists(int $id, string $objectModelName, string $exceptionClass, int $errorCode = 0): void
```
jolelievre:
```suggestion
    protected function assertObjectModelExists(int $id, string $objectTableName, string $exceptionClass, int $errorCode = 0): void
```

matks (approved):

tests/Integration/Behaviour/Features/Context/Domain/AttachmentFeatureContext.php
```diff
@@ -0,0 +1,120 @@
...
+    {
+        $source = _PS_ROOT_DIR_ . '/tests/Resources/dummyFile/' . $fileName;
+
+        if (!is_file($source)) {
+            throw new RuntimeException('%s is not a file', $source);
+        }
+
+        $destination = _PS_DOWNLOAD_DIR_ . $fileName;
+        copy($source, $destination);
+        chmod($destination, 7550);
```
jolelievre, line 116: Why not `6440` ? Does it need to be executable?

jolelievre (approved):

tests/Integration/Behaviour/Features/Context/Domain/AttachmentFeatureContext.php
zuk3975: Thanks for not blocking the PR for this. 
opened another one, that isn't in a hurry & won't block anything
https://github.com/PrestaShop/PrestaShop/pull/21209

PR #20882, opened 2020-09-07
Recompute specific price tax excluded when mismatch with price tax included

### LINKED ISSUE #20573: BO - Orders page - Wrong rounding price displayed in the Editing product interface
#### Describe the bug

In the BO => Order page => a wrong rounding price displayed in the Editing product interface

#### Steps to Reproduce

1. Create a `ProductA` with 10 euro tax included
2. Create an order with the last product created and generate an invoice `invoice1`
3. In the BO => Order details page => Edit the `ProductA`'s price = 100 euro tax included
4. I add product A in a second invoice `invoice2` with price =100 tax included
5. Edit the `ProductA`'s price
image

https://drive.google.com/file/d/1GWR8s9WBBVXgW32D783QFG3z-ULvoLQF/view?usp=sharing

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: 7.2

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | When a product is added or edited in the Order page the price excluded and included might differ (eg: the computed price_tax_excl + applied taxes != price_tax_incl) This is usually due to an insufficient precision in the values In which case we use the price tax included as a reference and recompute the price tax excluded Both values are saved in SpecificPrice and in OrderDetail without rounding (at least their unit price, unless the config says otherwise) This allows to perform more precise computations in total and keep consistent values<br />Also this PR include a little improvement in OrderAmountUpdater it now updates all OrderDetails with the correct prices based on the values returned by the cart This allows to keep all data in sync and reduces the action performed in the handlers
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20573
| How to test?  | See steps described in issue #20573

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20882)


src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -235,8 +240,13 @@ private function createOrderDetail(Order $order, ?OrderInvoice $invoice, Cart $c
      *
      * @return array
      */
-    private function getCartProductData(Cart $cart, Product $product, ?Combination $combination, int $quantity): array
-    {
+    private function getCartProductData(
+        Cart $cart,
+        Product $product,
+        ?Combination $combination,
```
sowbiba, line 240: combination can be null ? you can put it at the end of the parameters

Progi1984: Why at the end ?

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -249,10 +259,14 @@ private function getCartProductData(Cart $cart, Product $product, ?Combination $
...
 
+        $productItem['price_with_reduction'] = (float) (string) $command->getProductPriceTaxIncluded();
+        $productItem['price_without_reduction'] = (float) (string) $command->getProductPriceTaxIncluded();
+        $productItem['price_wt'] = (float) (string) $command->getProductPriceTaxIncluded();
+
         switch (Configuration::get('PS_ROUND_TYPE')) {
             case Order::ROUND_TOTAL:
                 $productItem['total'] = $productItem['price_with_reduction_without_tax'] * $quantity;
-                $productItem['total_wt'] = $productItem['price_with_reduction'] * $quantity;
+                $productItem['total_wt'] = (float) (string) $command->getProductPriceTaxIncluded() * $quantity;
```
Progi1984, line 269:
```suggestion
                $productItem['total_wt'] = $productItem['price_without_reduction'] * $quantity;
```

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
jolelievre: It's not optionnable, just nullable But the parameter is always provided

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
jolelievre: You should be able to use `$productItem['price_without_reduction']` instead since you already replaced it correctly, this would we more consistent because you already use `$productItem['price_with_reduction_without_tax']`

And it avoids performing the type casting each time

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -169,9 +170,14 @@ public function addProductsToOrderWithNewInvoiceAndTheFollowingDetails(string $o
     {
         $orderId = SharedStorage::getStorage()->get($orderReference);
 
+        $taxCalculator = $this->getOrderTaxCalculator($orderId);
+
         $data = $table->getRowsHash();
+        $data['price_tax_incl'] = $data['price_tax_incl'] ?? (string) $taxCalculator->addTaxes($data['price']);
```
jolelievre: Does it mean you can force the value by using `price_tax_incl` in the behat data? Then in that case you don't need the `$taxCalculator` Maybe you should handle this separately in an if statement rather than a ternary

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -1356,4 +1367,25 @@ private function getOrderDiscountByName(int $orderId, string $cartRuleName): ?Or
...
+     *
+     * @return TaxCalculator|null
+     *
+     * @throws \PrestaShopDatabaseException
+     * @throws \PrestaShopException
+     */
+    private function getOrderTaxCalculator(int $orderId)
+    {
+        $order = new Order($orderId);
+        $taxes = $order->getOrderDetailTaxes();
```
jolelievre: I didn't know this one! Seems handy!
But it means each product can have its dedicated tax group?

tests/Integration/Behaviour/Features/Scenario/Order/order_from_bo.feature
```diff
@@ -229,9 +229,9 @@ Feature: Order from Back Office (BO)
     And product "Mug The best is yet to come" in order "bo_order1" has following details:
       | product_quantity            | 3  |
       | product_price               | 12 |
-      | unit_price_tax_incl         | 12 |
+      | unit_price_tax_incl         | 12.72 |
```
jolelievre, line 223: It's surprising that it's the only place you needed to fix this This indicates that we probably didn't test the OrderDetails values enough

tests/Integration/Behaviour/Features/Scenario/Order/order_from_bo.feature
```diff
@@ -333,3 +333,21 @@ Feature: Order from Back Office (BO)
...
+      | price_tax_incl | 100                     |
+      | free_shipping  | false                   |
+    When I add products to order "bo_order1" with new invoice and the following products details:
+      | name           | Mug The best is yet to come |
+      | amount         | 1                           |
+      | price          | 94.33                       |
+      | price_tax_incl | 100                         |
+      | free_shipping  | false                       |
+    And product "Mug The best is yet to come" in order "bo_order1" has following details:
+      | unit_price_tax_incl         | 100   |
```
jolelievre: You should add tests on other values (like in edition), since it's the only test that checks OrderDetail values on addition (`product_quantity`,  `product_price`,  `total_price_tax_incl`, `total_price_tax_excl`

jolelievre (commented):
Nothin mandatory to fix, but some suggestions for improvement

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
matthieu-rolland: Given the way it works it seems to behave as if it was a possibility, but I'm not sure in what context can this happen 🤔

tests/Integration/Behaviour/Features/Scenario/Order/order_from_bo.feature
matthieu-rolland: well I only had to fix it when there was an "add product" or "edit product" without error before checking the price

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
Progi1984: In this case, the position should not change

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -458,12 +465,18 @@ public function editCombinationToOrderWithFollowingDetails(string $combinationNa
      */
     private function updateProductInOrder(int $orderId, array $productOrderDetail, array $data)
     {
+        // if tax included price is not give, it is calculated
```
Progi1984:
```suggestion
        // if tax included price is not given, it is calculated
```

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -458,12 +465,18 @@ public function editCombinationToOrderWithFollowingDetails(string $combinationNa
      */
     private function updateProductInOrder(int $orderId, array $productOrderDetail, array $data)
     {
+        // if tax included price is not given, it is calculated
+        if (!isset($data['price_tax_incl'])) {
+            $taxCalculator = $this->getOrderTaxCalculator($orderId);
```
sowbiba: taxCalculator can be null and it is not handled here

tests/Integration/Behaviour/Features/Scenario/Order/order_from_bo.feature
```diff
@@ -324,3 +324,25 @@ Feature: Order from Back Office (BO)
       | Postal code      | 12345                              |
     When I change order "bo_order1" shipping address to "test-address"
     Then order "bo_order1" shipping address should be "test-address"
+
+  Scenario: Generate order then modify product price then add same product on another invoice and check the price
+      Given order "bo_order1" does not have any invoices
+      When I generate invoice for "bo_order1" order
```
atomiix:
```suggestion
    Given order "bo_order1" does not have any invoices
    When I generate invoice for "bo_order1" order
```

sowbiba (dismissed):

Progi1984 (dismissed):

matks (dismissed):

SD1982:
Hi @matthieu-rolland Can you please resolve the conflicting files ?

matthieu-rolland:
rebased and green 👍

atomiix (dismissed):

PierreRambaud (dismissed):

SD1982:
Hi @matthieu-rolland the initial issue seems to be correctly fixed but in the order detail and in the order list the total is not correct 

**Currency set to 2 decimals**
order list ( the total should be 200 )
Error order list
order detail ( the total should be 200 )
Error order detail

**Currency set to 3 decimals**
order list ( the total should be 400 )
error order list 3 decimal
order detail ( the total should be 400 )
error order detail 3 decimal

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -243,6 +245,35 @@ protected function updateSpecificPrice(
         $specificPrice->add();
     }
 
+    /**
+     * @param Number $priceTaxIncluded
+     * @param Number $priceTaxExcluded
```
Progi1984:
```suggestion
     * @param \PrestaShop\Decimal\Number $priceTaxIncluded
     * @param \PrestaShop\Decimal\Number $priceTaxExcluded
```

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -243,6 +245,35 @@ protected function updateSpecificPrice(
...
+     * @param Order $order
+     * @param Product $product
+     *
+     * @return Number
```
Progi1984:
```suggestion
     * @return \PrestaShop\Decimal\Number
```

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -243,6 +245,35 @@ protected function updateSpecificPrice(
...
+     */
+    private function getPrecisePriceTaxExcluded(
+        Number $priceTaxIncluded,
+        Number $priceTaxExcluded,
+        Order $order,
+        Product $product
+    ): Number {
+        $taxAddress = new Address($order->{Configuration::get('PS_TAX_ADDRESS_TYPE', null, null, $order->id_shop)});
```
Progi1984, line 262:
```suggestion
        $taxAddress = new Address($order->{Configuration::get('PS_TAX_ADDRESS_TYPE', null, null, $order->id_shop)});
```

You don't check the return of `Configuration::get` exists in $order ?

tests/Integration/Behaviour/Features/Context/CarrierFeatureContext.php
```diff
@@ -411,4 +411,18 @@ public function enableCarrier(string $carrierReference)
         // Reset cache so that the carrier becomes selectable
         Carrier::resetStaticCache();
     }
+
+    /**
+     * @Then I associate tax rule group :taxRulesGroupReference to carrier :carrierReference
```
Progi1984:
```suggestion
     * @Then I associate the tax rule group :taxRulesGroupReference to the carrier :carrierReference
```

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -273,6 +284,11 @@ public function addProductsToOrderWithExistingInvoiceAndTheFollowingDetails(stri
             $combinationId = 0;
         }
 
+        if (empty($data['price_tax_incl'])) {
+            $taxCalculator = $this->getProductTaxCalculator((int) $orderId, $product->getProductId());
+            $data['price_tax_incl'] = !empty($taxCalculator) ? (string) $taxCalculator->addTaxes($data['price']) : $data['price'];
+        }
```
Progi1984, line 290: Duplicated lines ?

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -943,6 +965,35 @@ public function checkProductDetailsWithReference(string $orderReference, string
         }
     }
 
+    /**
+     * @Then product :productName in :invoicePosition invoice from order :orderReference should have following details:
```
Progi1984:
```suggestion
     * @Then the product :productName in the :invoicePosition invoice from the order :orderReference should have these following details:
```

tests/Integration/Behaviour/Features/Context/Domain/TaxFeatureContext.php
```diff
@@ -259,9 +259,9 @@ private function createTaxUsingCommand($taxReference, array $data)
     }
 
     /**
-     * @Then I add tax rule group for tax :taxReference with following conditions:
+     * @Then I add tax rule group :taxGroupReference for tax :taxReference with following conditions:
```
Progi1984:
```suggestion
     * @Then I add the tax rule group :taxGroupReference for the tax :taxReference with these following conditions:
```

Progi1984 (commented):
Some comments

src/Adapter/Number/RoundModeConverter.php
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * @param int $legacyRoundMode
+     *
+     * @return string
+     *
+     * @throws CoreException
+     */
+    public static function getNumberRoundMode(int $legacyRoundMode): string
+    {
+        if (!isset(self::MODE_MAP[$legacyRoundMode])) {
```
PierreRambaud:
```suggestion
        if (!isset(static::MODE_MAP[$legacyRoundMode])) {
```

src/Adapter/Number/RoundModeConverter.php
```diff
@@ -0,0 +1,59 @@
...
+            throw new CoreException(sprintf('Cannot map round mode %d', $legacyRoundMode));
+        }
+
+        return self::MODE_MAP[$legacyRoundMode];
```
PierreRambaud:
```suggestion
        return static::MODE_MAP[$legacyRoundMode];
```

src/Adapter/Order/AbstractOrderHandler.php
jolelievre: Nope, this is a mandatory configuration that is installed with the Shop If it doesn't exist the whole checkout will be broken so this part won't be the most important ^^

And I couldn't use the `ConfigurationInterface` because it's an abstract class so it would force all classes that use this method to be injected the service even if they don't use it nor this method

src/Adapter/Order/AbstractOrderHandler.php
jolelievre: All for the sake of PHPStan that is only enabled on the develop branch 😅

jolelievre: I'm kidding, I'm gonna fix it to ease the merge
But seriously it's a HUGE pain that we need to use the fully qualified class name, this is just gonna make our code uglier and uglier, reduce the motivation to use this library, and prevent us from using our IDE instinctively..
In my opinion it's PHPStan that should adapt to this use case, with a configuration or something, not us..

jolelievre:
@PierreRambaud @Progi1984 I addresses all your feedback

src/Adapter/Order/AbstractOrderHandler.php
Progi1984: Open an issue on PHPStan repository :p

src/Adapter/Order/AbstractOrderHandler.php
Progi1984: Ok for me.

Progi1984 (approved):

PierreRambaud (approved):

SD1982:

Progi1984:
Thanks @matthieu-rolland, @jolelievre & @SD1982

PR #20902, opened 2020-09-08
Keep order detail from being updated in BO > Order detail page, when product doesn't exist anymore

### LINKED ISSUE #20818: [1.7.7 beta 2] Order view - Product ID is invalid error message
#### Steps to Reproduce

1. Install 1.7.7 beta 2.
2. Delete all products.
3. Go edit KHWLILZLL.
4. Change price of the product.
5. Price will change, but error message will be displayed that the product ID is invalid.

#### Additional information

* PrestaShop version: 1.7.7 beta 2
* PHP version: 7.2

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | In BO > Order detail, when updating the price of a deleted product, we would have a bad error message and the order detail would still be updated. This PR improves the error message and do not update the price.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20818
| How to test?  | See steps described in #20818

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20902)

### DIFF
--- a/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
@@ -39,6 +39,7 @@ use PrestaShop\PrestaShop\Adapter\Order\AbstractOrderHandler;
 use PrestaShop\PrestaShop\Adapter\Order\OrderProductQuantityUpdater;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\CannotEditDeliveredOrderProductException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\DuplicateProductInOrderInvoiceException;
+use PrestaShop\PrestaShop\Core\Domain\Order\Exception\CannotUpdateProductInOrderException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Exception\OrderException;
 use PrestaShop\PrestaShop\Core\Domain\Order\Product\Command\UpdateProductInOrderCommand;
 use PrestaShop\PrestaShop\Core\Domain\Order\Product\CommandHandler\UpdateProductInOrderHandlerInterface;
@@ -151,6 +152,12 @@ final class UpdateProductInOrderHandler extends AbstractOrderHandler implements 
         Order $order,
         OrderInvoice $orderInvoice = null
     ) {
+        // assert product exists
+        $product = new Product($orderDetail->product_id);
+        if ($product->id !== (int) $orderDetail->product_id) {
+            throw new CannotUpdateProductInOrderException('You cannot edit the price of a product that no longer exists in your catalog.');
+        }
+
         if (!Validate::isLoadedObject($orderDetail)) {
             throw new OrderException('The Order Detail object could not be loaded.');
         }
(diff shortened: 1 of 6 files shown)


Progi1984 (dismissed):

matks (dismissed):

sowbiba (dismissed):

src/Core/Domain/Order/Exception/CannotUpdateProductInOrder.php
```diff
@@ -0,0 +1,35 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Order\Exception;
+
+/**
+ * Thrown on failure to update product in an order
+ */
+class CannotUpdateProductInOrderException extends OrderException
```
jolelievre: The filename is incorrect -> `CannotUpdateProductInOrderException.php`

src/Adapter/Order/CommandHandler/UpdateProductInOrderHandler.php
```diff
@@ -145,6 +146,12 @@ private function assertProductCanBeUpdated(
         Order $order,
         OrderInvoice $orderInvoice = null
     ) {
+        // assert product exists
+        $product = new Product($orderDetail->product_id);
+        if ($product->id !== (int) $orderDetail->product_id) {
+            throw new CannotUpdateProductInOrderException('You cannot edit the price of a product that no longer exists in your catalog.');
```
jolelievre: The exception name seems too generic as there are many reasons why the product can not be updated (see all the rest of the function). This exception should reflect this particular reason, something like `ProductInOrderNotFoundException` `CannotFindProductInOrderException` or similar

jolelievre (changes requested):
The name and especially filename of the exception need to be improved Also could you add a behat test for this use case please? 🙂

src/Core/Domain/Order/Exception/CannotUpdateProductInOrder.php
matthieu-rolland: damn yes, I broke my branch and reproduced this code on a new one, and didn't name this file right 🤦

jolelievre (dismissed):

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -1494,4 +1495,67 @@ private function getOrderDiscountByName(int $orderId, string $cartRuleName): ?Or
...
+        $product->delete();
+    }
+
+    /**
+     * @When I update deleted product :productReference in order :orderReference
+     *
+     * @param string $productReference
+     * @param string $orderReference
+     */
+    public function tryUpdatingProductDeletedFromCatalogue(string $productReference, string $orderReference)
```
atomiix, line 1519: Why not updating `updateProductInOrder()` directly?

matthieu-rolland: Indeed ! So I called `updateProductInOrder()` in this method, thanx for the catch. 

This method is still necessary, we are in a special case: updating a product that was deleted from the catalogue, going through `editProductsToOrderWithFollowingDetails()` method would trigger errors at several points before updating because the product doesn't exist anymore.

Progi1984 (dismissed):

atomiix (dismissed):

PierreRambaud (dismissed):

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -1494,4 +1497,54 @@ private function getOrderDiscountByName(int $orderId, string $cartRuleName): ?Or
 
         return null;
     }
+
+    /**
+     * @When I delete product :productReference from catalogue
+     *
+     * @param string $productReference
+     */
+    public function removeProductFromCatalogue(string $productReference)
```
jolelievre, line 1506: This should be in `ProductFeatureContext`

matks (approved):

jolelievre (approved):

PR #20903, opened 2020-09-08
Update Order amount when its addresses are modified

### LINKED ISSUE #21099: Update delivery cost and taxe prices when changing the address
#### Describe the bug

When an address is edited in an Order, it won't update the shipping cost, taxes rules and the cart rules

#### Expected behavior

When an address is edited, the shipping cost, taxes rules and the cart rules are updated in an Order

#### Steps

I found a new issue following these steps: (All carriers are enabled & the United States is enabled)
1. In the FO you are logged with "John Doe"
2. Add the product demo_6 "The best is yet to come' Framed poster" to the cart, 
3. In the Address Step select "My address (Miami)"
4. In the Shipping Adress, select My light carrier = €2.00 tax incl. 
5. Continue checkout
6. In the BO > View the last order created
7. Add a new discount Free shipping
image
8. For shipping Address => select "Mon addressed(France)"
9. For invoice Address => select "Mon address(France)"

The Shipping well updated but `Discounts`, `Value discount`, and `Total` are incorrect
See error
image

@jolelievre @PrestaShop/product-team this issue is reproduced with PS1767
image

>However maybe it is not exactly expected and we want the tax rate of the products to be updated when we change the address

I have a strange behavior about Product prices following these steps:
1. In the FO you are logged with "John Doe"
2. Add the product demo_6 "The best is yet to come' Framed poster" to the cart, 
3. In the Address Step select "My address (Miami)"
4. In the Shipping Adress, select My light carrier = €2.00 tax incl. 
5. Continue checkout
6. In the BO > View the last order created
7. Add a new discount Free shipping
8. For shipping Address => select "Mon addressed(France)"
9. For invoice Address => select "Mon address(France)"
image
10. **Delete the voucher**
image
@jolelievre this issue not reproduced with PS1767
image

Thanks!

_Originally posted by @khouloudbelguith in https://github.com/PrestaShop/PrestaShop/pull/20903#issuecomment-691182842_

### LINKED ISSUE #21254: BO Free Shipping Discount wrongly calculated
#### Describe the bug

In BO, adding a "free shipping" voucher on an order whose carrier has been changed in BO (thus changing the shipping costs), the voucher value is set at the previous shipping costs, not the new one.

#### Expected behavior

Free shipping voucher value should be calculated with current carrier cost, and not the original one.

#### Steps to Reproduce

0. Go to FO, make an order with a free carrier
1. Go to BO > Orders > Open order > Change carrier for a carrier with a fee
2. Add a "Free Shipping" voucher : It's value is at 0€ instead of being at the actual shipping cost

**Screenshots**

https://drive.google.com/file/d/1-zd3aFeN56iz_dlJkwLyeTRVf9hwUDpM/view?usp=sharing

#### Additional information

* PrestaShop version: 1770rc1build1
* PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Add behat tests regarding the order's addresses modification Updated a few handler related to order/cart addresses and carrier along the way to make sure modifications are consistent The automated tests did highlight that the cart was not correctly updated however I'm not sure if it had consequences of shipping calculations
| Type?         | improvement
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | yes
| Fixed ticket? |  Fixes #21099 and #21254
| How to test?  | Travis tests green, not sure how QA can test it except by modifying addresses in order creation and edition and checking that shipping fees are correctly updated (this can be done thanks to the new carriers pre installed in the fixtures)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20903)

## :notebook: BC break

* A new parameter was introduced in this method: `PrestaShop\PrestaShop\Adapter::set($key, $value, ShopConstraint $shopConstraint = null, array $options = [])`
  * Thus the `$options` becomes the fourth parameter, we didn't find any use in the core code, but some modules might use it.
* Some other public methods have new (optional) parameters:
  * `PrestaShop\PrestaShop\Adapter::get($key, $default = null, ShopConstraint $shopConstraint = null)`
  * `PrestaShop\PrestaShop\Adapter::has($key, ShopConstraint $shopConstraint = null)`

## Deprecation

Nothing really changed but I reallized some fields in OrderDetail are not used any more, they are probably here for backward compatibility but the tax info is now stored in `order_detail_tax`

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -333,6 +333,36 @@ class CartCore extends ObjectModel
     }
 
     /**
+     * Update the Delivery Address ID of the Cart.
+     *
+     * @param int $id_address Current Address ID to change
+     * @param int $id_address_new New Address ID
+     */
+    public function updateDeliveryAddressId($id_address, $id_address_new)
+    {
+        $to_update = false;
+        if (!isset($this->id_address_delivery) || $this->id_address_delivery == $id_address) {
+            $to_update = true;
+            $this->id_address_delivery = $id_address_new;
+        }
+        if ($to_update) {
+            $this->update();
+        }
+
+        $sql = 'UPDATE `' . _DB_PREFIX_ . 'cart_product`
+        SET `id_address_delivery` = ' . (int) $id_address_new . '
+        WHERE  `id_cart` = ' . (int) $this->id . '
+            AND `id_address_delivery` = ' . (int) $id_address;
+        Db::getInstance()->execute($sql);
+
+        $sql = 'UPDATE `' . _DB_PREFIX_ . 'customization`
+            SET `id_address_delivery` = ' . (int) $id_address_new . '
+            WHERE  `id_cart` = ' . (int) $this->id . '
+                AND `id_address_delivery` = ' . (int) $id_address;
+        Db::getInstance()->execute($sql);
+    }
+
+    /**
      * Deletes current Cart from the database.
      *
      * @return bool True if delete was successful
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -340,12 +340,8 @@ class CartCore extends ObjectModel
      */
     public function updateDeliveryAddressId(int $currentAddressId, int $newAddressId)
     {
-        $needsUpdate = false;
         if (!isset($this->id_address_delivery) || (int) $this->id_address_delivery === $currentAddressId) {
-            $needsUpdate = true;
             $this->id_address_delivery = $newAddressId;
-        }
-        if ($needsUpdate) {
             $this->update();
         }
 
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -335,30 +335,30 @@ class CartCore extends ObjectModel
     /**
      * Update the Delivery Address ID of the Cart.
      *
-     * @param int $id_address Current Address ID to change
-     * @param int $id_address_new New Address ID
+     * @param int $currentAddressId Current Address ID to change
+     * @param int $newAddressId New Address ID
      */
-    public function updateDeliveryAddressId($id_address, $id_address_new)
+    public function updateDeliveryAddressId(int $currentAddressId, int $newAddressId)
     {
-        $to_update = false;
-        if (!isset($this->id_address_delivery) || $this->id_address_delivery == $id_address) {
-            $to_update = true;
-            $this->id_address_delivery = $id_address_new;
+        $needsUpdate = false;
+        if (!isset($this->id_address_delivery) || (int) $this->id_address_delivery === $currentAddressId) {
+            $needsUpdate = true;
+            $this->id_address_delivery = $newAddressId;
         }
-        if ($to_update) {
+        if ($needsUpdate) {
             $this->update();
         }
 
         $sql = 'UPDATE `' . _DB_PREFIX_ . 'cart_product`
-        SET `id_address_delivery` = ' . (int) $id_address_new . '
+        SET `id_address_delivery` = ' . $newAddressId . '
         WHERE  `id_cart` = ' . (int) $this->id . '
-            AND `id_address_delivery` = ' . (int) $id_address;
+            AND `id_address_delivery` = ' . $currentAddressId;
         Db::getInstance()->execute($sql);
 
         $sql = 'UPDATE `' . _DB_PREFIX_ . 'customization`
-            SET `id_address_delivery` = ' . (int) $id_address_new . '
+            SET `id_address_delivery` = ' . $newAddressId . '
             WHERE  `id_cart` = ' . (int) $this->id . '
-                AND `id_address_delivery` = ' . (int) $id_address;
+                AND `id_address_delivery` = ' . $currentAddressId;
         Db::getInstance()->execute($sql);
     }
 
--- a/classes/order/OrderDetail.php
+++ b/classes/order/OrderDetail.php
@@ -456,12 +456,29 @@ class OrderDetailCore extends ObjectModel
         return $this->saveTaxCalculator($order, true);
     }
 
+    /**
+     * Get a TaxCalculator adapted for the OrderDetail's product and the specified address
+     *
+     * @param Address $address
+     *
+     * @return TaxCalculator
+     */
     public function getTaxCalculatorByAddress(Address $address)
     {
         $this->setContext((int) $this->id_shop);
-        $tax_manager = TaxManagerFactory::getManager($address, (int) Product::getIdTaxRulesGroupByIdProduct((int) $this->product_id, $this->context));
+        $tax_manager = TaxManagerFactory::getManager($address, $this->getTaxRulesGroupId());
 
         return $tax_manager->getTaxCalculator();
+    }
+
+    /**
+     * Dynamically get the taxRulesGroupId instead of relying one the one saved in database
+     *
+     * @return int
+     */
+    public function getTaxRulesGroupId(): int
+    {
+        return (int) Product::getIdTaxRulesGroupByIdProduct((int) $this->product_id, $this->context);
     }
 
     /**
--- a/classes/order/OrderDetail.php
+++ b/classes/order/OrderDetail.php
@@ -392,10 +392,6 @@ class OrderDetailCore extends ObjectModel
             return false;
         }
 
-        if (count($this->tax_calculator->taxes) == 0) {
-            return true;
-        }
-
         if ($order->total_products <= 0) {
             return true;
         }
@@ -441,11 +437,15 @@ class OrderDetailCore extends ObjectModel
             Db::getInstance()->execute('DELETE FROM `' . _DB_PREFIX_ . 'order_detail_tax` WHERE id_order_detail=' . (int) $this->id);
         }
 
-        $values = rtrim($values, ',');
-        $sql = 'INSERT INTO `' . _DB_PREFIX_ . 'order_detail_tax` (id_order_detail, id_tax, unit_amount, total_amount)
+        if (!empty($values)) {
+            $values = rtrim($values, ',');
+            $sql = 'INSERT INTO `' . _DB_PREFIX_ . 'order_detail_tax` (id_order_detail, id_tax, unit_amount, total_amount)
                 VALUES ' . $values;
 
-        return Db::getInstance()->execute($sql);
+            return Db::getInstance()->execute($sql);
+        }
+
+        return true;
     }
 
     public function updateTaxAmount($order)
--- a/classes/order/OrderDetail.php
+++ b/classes/order/OrderDetail.php
@@ -463,7 +463,7 @@ class OrderDetailCore extends ObjectModel
      *
      * @return TaxCalculator
      */
-    public function getTaxCalculatorByAddress(Address $address)
+    public function getTaxCalculatorByAddress(Address $address): TaxCalculator
     {
         $this->setContext((int) $this->id_shop);
         $tax_manager = TaxManagerFactory::getManager($address, $this->getTaxRulesGroupId());
--- a/classes/order/OrderDetail.php
+++ b/classes/order/OrderDetail.php
@@ -139,10 +139,14 @@ class OrderDetailCore extends ObjectModel
     /** @var datetime */
     public $download_deadline;
 
-    /** @var string $tax_name * */
+    /**
+     * @var string @deprecated Order Detail Tax is saved in order_detail_tax table now
+     */
     public $tax_name;
 
-    /** @var float $tax_rate * */
+    /**
+     * @var float @deprecated Order Detail Tax is saved in order_detail_tax table now
+     */
     public $tax_rate;
 
     /** @var float $tax_computation_method * */
@@ -357,9 +361,8 @@ class OrderDetailCore extends ObjectModel
         if ($results = Db::getInstance()->executeS($sql)) {
             foreach ($results as $result) {
                 $taxes[] = new Tax((int) $result['id_tax']);
+                $computation_method = $result['tax_computation_method'];
             }
-
-            $computation_method = $result['tax_computation_method'];
         }
 
         return new TaxCalculator($taxes, $computation_method);
@@ -372,6 +375,9 @@ class OrderDetailCore extends ObjectModel
      * @deprecated Functionality moved to Order::updateOrderDetailTax
      *             because we need the full order object to do a good job here.
      *             Will no longer be supported after 1.6.1
+     *             (Note: this one is not that deprecated because Order::updateOrderDetailTax
+     *             performs no update unless order_detail_tax is filled. So we rely on updateTaxAmount
+     *             which correctly builds the TaxCalculator with up to date taxes unlike getTaxCalculatorStatic)
      *
      * @return bool
      */
@@ -444,12 +450,18 @@ class OrderDetailCore extends ObjectModel
 
     public function updateTaxAmount($order)
     {
-        $this->setContext((int) $this->id_shop);
         $address = new Address((int) $order->{Configuration::get('PS_TAX_ADDRESS_TYPE')});
-        $tax_manager = TaxManagerFactory::getManager($address, (int) Product::getIdTaxRulesGroupByIdProduct((int) $this->product_id, $this->context));
-        $this->tax_calculator = $tax_manager->getTaxCalculator();
+        $this->tax_calculator = $this->getTaxCalculatorByAddress($address);
 
         return $this->saveTaxCalculator($order, true);
+    }
+
+    public function getTaxCalculatorByAddress(Address $address)
+    {
+        $this->setContext((int) $this->id_shop);
+        $tax_manager = TaxManagerFactory::getManager($address, (int) Product::getIdTaxRulesGroupByIdProduct((int) $this->product_id, $this->context));
+
+        return $tax_manager->getTaxCalculator();
     }
 
     /**
--- a/src/Adapter/Configuration.php
+++ b/src/Adapter/Configuration.php
@@ -139,8 +139,17 @@ class Configuration extends ParameterBag implements ConfigurationInterface, Shop
             return $this->getLocalized($key, $shopId, $shopGroupId);
         }
 
-        if (ConfigurationLegacy::hasKey($key, null, $shopGroupId, $shopId)) {
-            return ConfigurationLegacy::get($key, null, $shopGroupId, $shopId);
+        // Since has key doesn't check manage the fallback shop > shop group > global, we handle it manually
+        if (ConfigurationLegacy::hasKey($key, null, null, $shopId)) {
+            return ConfigurationLegacy::get($key, null, null, $shopId);
+        }
+
+        if (ConfigurationLegacy::hasKey($key, null, $shopGroupId)) {
+            return ConfigurationLegacy::get($key, null, $shopGroupId);
+        }
+
+        if (ConfigurationLegacy::hasKey($key)) {
+            return ConfigurationLegacy::get($key);
         }
 
         return $default;
--- a/src/Adapter/Configuration.php
+++ b/src/Adapter/Configuration.php
@@ -139,7 +139,7 @@ class Configuration extends ParameterBag implements ConfigurationInterface, Shop
             return $this->getLocalized($key, $shopId, $shopGroupId);
         }
 
-        // Since has key doesn't check manage the fallback shop > shop group > global, we handle it manually
+        // Since hasKey doesn't check manage the fallback shop > shop group > global, we handle it manually
         if (ConfigurationLegacy::hasKey($key, null, null, $shopId)) {
             return ConfigurationLegacy::get($key, null, null, $shopId);
         }
@@ -231,7 +231,11 @@ class Configuration extends ParameterBag implements ConfigurationInterface, Shop
         $shopId = null !== $shopConstraint->getShopId() ? $shopConstraint->getShopId()->getValue() : null;
         $shopGroupId = null !== $shopConstraint->getShopGroupId() ? $shopConstraint->getShopGroupId()->getValue() : null;
 
-        return ConfigurationLegacy::hasKey($key, null, $shopGroupId, $shopId);
+        return
+            ConfigurationLegacy::hasKey($key, null, null, $shopId)
+            || ConfigurationLegacy::hasKey($key, null, $shopGroupId)
+            || ConfigurationLegacy::hasKey($key)
+        ;
     }
 
     /**
--- a/src/Adapter/Configuration.php
+++ b/src/Adapter/Configuration.php
@@ -30,10 +30,8 @@ use Combination;
 use Configuration as ConfigurationLegacy;
 use Feature;
 use Language;
-use PrestaShop\PrestaShop\Core\ConfigurationInterface;
 use PrestaShop\PrestaShop\Core\Domain\Configuration\ShopConfigurationInterface;
 use PrestaShop\PrestaShop\Core\Domain\Shop\ValueObject\ShopConstraint;
-use PrestaShop\PrestaShop\Core\Domain\Shop\ValueObject\ShopId;
 use PrestaShopBundle\Exception\NotImplementedException;
 use Shop;
 use Symfony\Component\HttpFoundation\ParameterBag;
@@ -41,7 +39,7 @@ use Symfony\Component\HttpFoundation\ParameterBag;
 /**
  * Adapter of Configuration ObjectModel.
  */
-class Configuration extends ParameterBag implements ConfigurationInterface, ShopConfigurationInterface
+class Configuration extends ParameterBag implements ShopConfigurationInterface
 {
     /**
      * @var Shop
@@ -96,39 +94,19 @@ class Configuration extends ParameterBag implements ConfigurationInterface, Shop
      *
      * @param string $key
      * @param mixed $default The default value if the parameter key does not exist
+     * @param ShopConstraint|null $shopConstraint
      *
      * @return mixed
      */
-    public function get($key, $default = null)
+    public function get($key, $default = null, ShopConstraint $shopConstraint = null)
     {
         if (defined($key)) {
             return constant($key);
         }
 
-        return $this->doGet($key, $default);
-    }
+        $shopId = $this->getShopId($shopConstraint);
+        $shopGroupId = $this->getShopGroupId($shopConstraint);
 
-    /**
-     * {@inheritdoc}
-     */
-    public function getForShop(string $key, ShopConstraint $shopConstraint)
-    {
-        $shopId = null !== $shopConstraint->getShopId() ? $shopConstraint->getShopId()->getValue() : null;
-        $shopGroupId = null !== $shopConstraint->getShopGroupId() ? $shopConstraint->getShopGroupId()->getValue() : null;
-
-        return $this->doGet($key, null, $shopId, $shopGroupId);
-    }
-
-    /**
-     * @param $key
-     * @param null $default
-     * @param int|null $shopId
-     * @param int|null $shopGroupId
-     *
-     * @return array|false|mixed|string|null
-     */
-    private function doGet($key, $default = null, ?int $shopId = null, ?int $shopGroupId = null)
-    {
         //If configuration has never been accessed it is still empty and hasKey/isLangKey will always return false
         if (!ConfigurationLegacy::configurationIsLoaded()) {
             ConfigurationLegacy::loadConfiguration();
@@ -140,16 +118,18 @@ class Configuration extends ParameterBag implements ConfigurationInterface, Shop
         }
 
         // Since hasKey doesn't check manage the fallback shop > shop group > global, we handle it manually
-        if (ConfigurationLegacy::hasKey($key, null, null, $shopId)) {
-            return ConfigurationLegacy::get($key, null, null, $shopId);
+        $hasKey = ConfigurationLegacy::hasKey($key, null, null, $shopId);
+        if ($hasKey || $shopConstraint->isStrict()) {
+            return $hasKey ? ConfigurationLegacy::get($key, null, null, $shopId) : null;
         }
 
-        if (ConfigurationLegacy::hasKey($key, null, $shopGroupId)) {
-            return ConfigurationLegacy::get($key, null, $shopGroupId);
+        $hasKey = ConfigurationLegacy::hasKey($key, null, $shopGroupId);
+        if ($hasKey || $shopConstraint->isStrict()) {
+            return $hasKey ? ConfigurationLegacy::get($key, null, $shopGroupId) : null;
         }
 
-        if (ConfigurationLegacy::hasKey($key)) {
-            return ConfigurationLegacy::get($key);
+        if ($hasKey = ConfigurationLegacy::hasKey($key) || $shopConstraint->isStrict()) {
+            return $hasKey ? ConfigurationLegacy::get($key) : null;
         }
 
         return $default;
@@ -160,21 +140,25 @@ class Configuration extends ParameterBag implements ConfigurationInterface, Shop
      *
      * @param string $key
      * @param mixed $value
+     * @param ShopConstraint|null $shopConstraint
      * @param array $options Options
      *
      * @return $this
      *
      * @throws \Exception
      */
-    public function set($key, $value, array $options = [])
+    public function set($key, $value, ShopConstraint $shopConstraint = null, array $options = [])
     {
         // By default, set a piece of configuration for all available shops and shop groups
         $shopGroupId = null;
         $shopId = null;
 
-        if ($this->shop instanceof Shop) {
+        if ($this->shop instanceof Shop && null === $shopConstraint) {
             $shopGroupId = $this->shop->id_shop_group;
             $shopId = $this->shop->id;
+        } else {
+            $shopId = $this->getShopId($shopConstraint);
+            $shopGroupId = $this->getShopGroupId($shopConstraint);
         }
 
         $html = isset($options['html']) ? (bool) $options['html'] : false;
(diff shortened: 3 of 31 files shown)


classes/Cart.php
```diff
@@ -332,6 +332,36 @@ public function updateAddressId($id_address, $id_address_new)
...
 
+    /**
+     * Update the Delivery Address ID of the Cart.
+     *
+     * @param int $id_address Current Address ID to change
+     * @param int $id_address_new New Address ID
+     */
+    public function updateDeliveryAddressId($id_address, $id_address_new)
+    {
+        $to_update = false;
```
Progi1984:
```suggestion
        $needsUpdate = false;
```

classes/Cart.php
```diff
@@ -332,6 +332,36 @@ public function updateAddressId($id_address, $id_address_new)
...
+        if (!isset($this->id_address_delivery) || $this->id_address_delivery == $id_address) {
+            $to_update = true;
```
Progi1984:
```suggestion
            $needsUpdate = true;
```

classes/Cart.php
```diff
@@ -332,6 +332,36 @@ public function updateAddressId($id_address, $id_address_new)
...
+            $this->id_address_delivery = $id_address_new;
+        }
+        if ($to_update) {
```
Progi1984:
```suggestion
        if ($needsUpdate) {
```

classes/Cart.php
```diff
@@ -332,6 +332,36 @@ public function updateAddressId($id_address, $id_address_new)
         Db::getInstance()->execute($sql);
     }
 
+    /**
+     * Update the Delivery Address ID of the Cart.
+     *
+     * @param int $id_address Current Address ID to change
+     * @param int $id_address_new New Address ID
+     */
+    public function updateDeliveryAddressId($id_address, $id_address_new)
```
Progi1984:
```suggestion
    public function updateDeliveryAddressId(int $id_address, int $id_address_new)
```

src/Adapter/Order/CommandHandler/ChangeOrderDeliveryAddressHandler.php
```diff
@@ -53,11 +53,33 @@ public function handle(ChangeOrderDeliveryAddressCommand $command)
             throw new OrderException('New delivery address is not valid');
         }
 
-        $order->id_address_delivery = $address->id;
-        $cart->id_address_delivery = $address->id;
+        $cart->updateDeliveryAddressId((int) $cart->id_address_delivery, (int) $address->id);
+        $cart->setDeliveryOption([
+            $cart->id_address_delivery => $this->formatLegacyDeliveryOptionFromCarrierId($order->id_carrier),
```
Progi1984:
```suggestion
            (int) $cart->id_address_delivery => $this->formatLegacyDeliveryOptionFromCarrierId($order->id_carrier),
```

tests/Integration/Behaviour/Features/Context/CommonFeatureContext.php
```diff
@@ -142,9 +146,13 @@ public function clearEntityManager()
...
+        Address::resetStaticCache();
         Cache::clear();
-        Pack::resetStaticCache();
+        Carrier::resetStaticCache();
+        Cart::resetStaticCache();
+        CartRule::resetStaticCache();
         Category::resetStaticCache();
+        Pack::resetStaticCache();
         Product::resetStaticCache();
         Language::resetCache();
```
Progi1984, line 166: My god !

tests/Integration/Behaviour/Features/Context/Domain/OrderShippingFeatureContext.php
```diff
@@ -165,6 +166,19 @@ public function changeOrderShippingAddressTo(string $orderReference, string $ord
...
 
+    /**
+     * @When I change order :orderReference invoice address to :orderInvoiceAddress
+     *
+     * @param string $orderReference
+     * @param string $orderInvoiceAddress
+     */
+    public function changeOrderInvoiceAddressTo(string $orderReference, string $orderInvoiceAddress)
+    {
+        $orderId = SharedStorage::getStorage()->get($orderReference);
```
Progi1984, line 177:
```suggestion
        $orderId = (int) SharedStorage::getStorage()->get($orderReference);
```

tests/Integration/Behaviour/Features/Context/Domain/OrderShippingFeatureContext.php
```diff
@@ -181,4 +195,21 @@ public function orderShippingAddressShouldBe(string $orderReference, string $ord
...
+
+    /**
+     * @Then order :orderReference invoice address should be :orderInvoiceAddress
+     *
+     * @param string $orderReference
+     * @param string $orderInvoiceAddress
+     */
+    public function orderInvoiceAddressShouldBe(string $orderReference, string $orderInvoiceAddress)
+    {
+        $orderId = SharedStorage::getStorage()->get($orderReference);
```
Progi1984, line 207:
```suggestion
        $orderId = (int) SharedStorage::getStorage()->get($orderReference);
```

tests/Integration/Behaviour/Features/Scenario/Order/order_shipping.feature
```diff
@@ -390,3 +392,155 @@ Feature: Order from Back Office (BO)
...
+    And I add order "bo_order1" with the following details:
+      | cart                | dummy_cart                 |
+      | message             | test                       |
+      | payment module name | dummy_payment              |
+      | status              | Awaiting bank wire payment |
+    And order "bo_order1" should have 2 products in total
+    And order "bo_order1" should have 0 invoices
+    And order "bo_order1" should have 0 cart rule
+    And order "bo_order1" should have "price_carrier" as a carrier
+    And order "bo_order1" should have 2 products in total
```
Progi1984:
```suggestion
```

tests/Integration/Behaviour/Features/Scenario/Order/order_shipping.feature
```diff
@@ -390,3 +392,155 @@ Feature: Order from Back Office (BO)
...
```
Progi1984:
```suggestion
```

Progi1984 (commented):
Some comments

classes/Cart.php
jolelievre: I just copied the function above but alright 😉

tests/Integration/Behaviour/Features/Context/Domain/OrderShippingFeatureContext.php
jolelievre: Not required since the stored value is already an int This same code is used for ALL the order tests so it would imply modifying them all with no added value

Progi1984 (dismissed):

matthieu-rolland (dismissed):

Progi1984:
@PrestaShop/qa-functional Do you need to test this PR ?

khouloudbelguith:
Hi @jolelievre,

I found an issue following these steps:

1. In the FO you are logged with "John Doe"
2. Add the product demo_6 "The best is yet to come' Framed poster" to the cart, 
3. In the Address Step select "My address (Miami)"
4. In the Shipping Adress, select My light carrier
5. Continue checkout
6. In the BO > View the last order created
7. For shipping Address => select "Mon addressed(France)"
8. For invoice Address => select "Mon address(France)"

The Shipping well updated, but the products => are still displayed with tax excluded => NOK
They should be updated? 
Is this related to this PR or I need to create an issue separately?
image

Thanks!

jolelievre:
Hi @khouloudbelguith 

I'm not 100% sure but I believe this behaviour is wanted, because the principle of OrderDetail is to "save" the info at the moment the Order is placed. In this case the product had no taxes when it was ordered (because US taxes are not installed), so even if we change the address it doesn't change previous order details.

Besides this seems ok when you try to add new products:
- add another product it will have the appropriate the new tax rate (20%)
- even when you ad the same product it does propose to add it with the new tax rate

This behaviour seems consistent to me. However maybe it is not exactly expected and we want the tax rate of the products to be updated when we change the address. This should be decided by the @PrestaShop/product-team and managed in a separate issue/PR (after all this PR scope is about shipping costs not taxes).

If you agree with this, and if it's the only bug, I think we can validate the rest of this PR and merge it :)

jolelievre:
By the way I checked on 176, it is the exact same behaviour

And just for notice, you can see in your screenshot that for shipping costs the tax rate is correctly applied! The sipping cost was modified from 2€ to 2.40€ ^^ Only the product prices are not modified which makes sense

khouloudbelguith:
Hi @jolelievre,

I found a new issue following these steps: (All carriers are enabled & the United States is enabled)
1. In the FO you are logged with "John Doe"
2. Add the product demo_6 "The best is yet to come' Framed poster" to the cart, 
3. In the Address Step select "My address (Miami)"
4. In the Shipping Adress, select My light carrier = €2.00 tax incl. 
5. Continue checkout
6. In the BO > View the last order created
7. Add a new discount Free shipping
image
8. For shipping Address => select "Mon addressed(France)"
9. For invoice Address => select "Mon address(France)"

The Shipping well updated but `Discounts`, `Value discount`, and `Total` are incorrect
See error
image

@jolelievre @PrestaShop/product-team this issue is reproduced with PS1767
image

>However maybe it is not exactly expected and we want the tax rate of the products to be updated when we change the address

I have a strange behavior about Product prices following these steps:
1. In the FO you are logged with "John Doe"
2. Add the product demo_6 "The best is yet to come' Framed poster" to the cart, 
3. In the Address Step select "My address (Miami)"
4. In the Shipping Adress, select My light carrier = €2.00 tax incl. 
5. Continue checkout
6. In the BO > View the last order created
7. Add a new discount Free shipping
8. For shipping Address => select "Mon addressed(France)"
9. For invoice Address => select "Mon address(France)"
image
10. **Delete the voucher**
image
@jolelievre this issue not reproduced with PS1767
image

Thanks!

jolelievre:
Hi @khouloudbelguith,

1. so for the first bug, the issue is that discounts are not updated when Shipping costs are updated. To update it we should update all the Order's amount to sync them
2. second bug is a bit related, I think what happens is that when we remove the discount the whole order is recomputed and the correct tax is applied on the initial product, it was not visible when just switched the address because the Order was not recomputed

The real question behind these two behaviours is: should we update the whole order amount, along with shipping costs, discount amounts, tax values at every operation (even just switching address). This will need to be discussed with @MatShir next week I guess to validate the expected behaviour.

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -343,4 +346,40 @@ protected function updateOrderDetailsWithSameProduct(
...
+            'total_price_tax',
+            'total_shipping_price_tax',
+        ];
+        $taxAddressId = $order->{Configuration::get('PS_TAX_ADDRESS_TYPE', null, null, $order->id_shop)};
+        $taxAddress = new Address($taxAddressId);
+        $computingPrecision = $this->getPrecisionFromCart($cart);
+
+        foreach ($order->getOrderDetailList() as $row) {
+            $orderDetail = new OrderDetail($row['id_order_detail']);
+            $orderDetail->id_tax_rules_group = (int) Product::getIdTaxRulesGroupByIdProduct($orderDetail->product_id, Context::getContext());
```
Progi1984:
```suggestion
        $context = Context::getContext();

        foreach ($order->getOrderDetailList() as $row) {
            $orderDetail = new OrderDetail($row['id_order_detail']);
            $orderDetail->id_tax_rules_group = (int) Product::getIdTaxRulesGroupByIdProduct($orderDetail->product_id, $context);
```

src/Adapter/Order/CommandHandler/ChangeOrderInvoiceAddressHandler.php
```diff
@@ -39,6 +41,19 @@
...
 {
+    /**
+     * @var OrderAmountUpdater
+     */
+    private $orderAmountUpdater;
+
+    /**
+     * @param OrderAmountUpdater $orderAmountUpdater
+     */
+    public function __construct(OrderAmountUpdater $orderAmountUpdater)
```
Progi1984, line 51: use `PrestaShop\PrestaShop\Adapter\Configuration`;

src/Adapter/Order/CommandHandler/ChangeOrderDeliveryAddressHandler.php
```diff
@@ -39,6 +41,19 @@
...
 {
+    /**
+     * @var OrderAmountUpdater
+     */
+    private $orderAmountUpdater;
+
+    /**
+     * @param OrderAmountUpdater $orderAmountUpdater
+     */
+    public function __construct(OrderAmountUpdater $orderAmountUpdater)
```
Progi1984, line 51: use `PrestaShop\PrestaShop\Adapter\Configuration`;

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -343,4 +346,40 @@ protected function updateOrderDetailsWithSameProduct(
...
+    protected function updateOrderDetailsTax(Order $order, Cart $cart): void
+    {
+        // Fields that are updated (refunds are not modified, even though the tax changed we need to keep track
+        // of the actual refunded amount).
+        $updatedFields = [
+            'unit_price_tax',
+            'total_price_tax',
+            'total_shipping_price_tax',
+        ];
+        $taxAddressId = $order->{Configuration::get('PS_TAX_ADDRESS_TYPE', null, null, $order->id_shop)};
```
Progi1984: use `PrestaShop\PrestaShop\Adapter\Configuration`;

Progi1984 (commented):
Some feedbacks

src/Adapter/Order/CommandHandler/ChangeOrderDeliveryAddressHandler.php
jolelievre: I thought about it but the adapter service doesn't allow to specify the shop id..

src/Adapter/Order/AbstractOrderHandler.php
jolelievre: It's not more performant since it's a singleton but ok 😉

src/Adapter/Configuration.php
```diff
@@ -102,14 +105,47 @@ public function get($key, $default = null)
             return constant($key);
         }
 
+        return $this->doGet($key, $default);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getForShop(string $key, ShopConstraint $shopConstraint)
```
jolelievre: Side note: why introduce a `ShopConstraint` object? Because just providing shopID or just shopGroupId separately won't do the job. There is a fallback behaviour when dealing with multishop configurations because they are not necessarily overridden so the priority is:

1. shopId
2. groupShopId
3. global

Which is why I introduced this object that include both `ShopId` and `ShopGroupId` for conveniency (one parameter)

It could be reshaped as `public function getForShop(string $key, ShopId $shopId, ShopGroupId $shopGroupId)` but then we would need to make some parameters optional..

jolelievre: Also one drawback in this behaviour is that the fallback system is automatic, meaning you can't get the value for this specific Shop and get null or an exception when it doesn't exist, you will get the fallback value but you won't know if it comes from the Shop, its Group or the Global...

jolelievre: I started introducing a `ConfigurationValue` VO but stopped to avoid over-engineering this PR, the idea was to store the key and the value and add a few getters (getInt, getBool, getByLang, ...) for conveniency.

It could be used both for input and output, and it could also store the "source" shopId or shopGroupID, but this is far beyond the scope of this PR imo, the problem being that we start integrating interfaces here that are not mature enough, because of lack of time and discussion..

src/Core/Domain/Shop/ValueObject/ShopConstraint.php
```diff
@@ -0,0 +1,88 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Shop\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Shop\Exception\ShopException;
+
+class ShopConstraint
```
eternoendless, line 33: VOs should be immutable

src/Adapter/Configuration.php
eternoendless: `ShopConstraint` makes sense. 

To control fallback you can add an optional parameter `$fallback = true`.

That said, I think we need a better `Configuration` class because multishop is a core feature which is incompatible with a key value storage like ParameterBag (which this class extends). So trying to put it inside this class is basically creating a Frankenstein for the sake of not introducing a new class.

src/Core/Domain/Shop/ValueObject/ShopConstraint.php
jolelievre: Yes I though so as well at first, but I wanted to keep the possibility to define only shopId or only shopGroupId on this VO. So I guess I can make a constructor that accepts both but null can be provided?

src/Adapter/Configuration.php
jolelievre: Yep and there are even some methods of the ParameterBag that are overridden to thrown exception because they are not handled.. So I agree that this Adapter class is only temporary and should be completely rebuilt, but not in this PR
The fallback principle makes sense even as a core feature I guess 🤔. So I can still implement it now even with the Adapter implementation

jolelievre: Although it won't be used right now, so it might be safer to wait for next version to add this feature To make sure we make it right

src/Adapter/Configuration.php
```diff
@@ -124,21 +140,25 @@ public function get($key, $default = null)
...
      * @param mixed $value
+     * @param ShopConstraint|null $shopConstraint
      * @param array $options Options
      *
      * @return $this
      *
      * @throws \Exception
      */
-    public function set($key, $value, array $options = [])
+    public function set($key, $value, ShopConstraint $shopConstraint = null, array $options = [])
```
jolelievre, line 151: This is a potential BC break, I think we can assume it as I didn't any use of this `$options` parameter in the core code And it's present in none of our interfaces

classes/Cart.php
```diff
@@ -332,6 +332,36 @@ public function updateAddressId($id_address, $id_address_new)
...
+    public function updateDeliveryAddressId(int $currentAddressId, int $newAddressId)
+    {
+        $needsUpdate = false;
+        if (!isset($this->id_address_delivery) || (int) $this->id_address_delivery === $currentAddressId) {
+            $needsUpdate = true;
+            $this->id_address_delivery = $newAddressId;
+        }
+        if ($needsUpdate) {
+            $this->update();
+        }
```
PierreRambaud:
```suggestion
        if (!isset($this->id_address_delivery) || (int) $this->id_address_delivery === $currentAddressId) {
            $this->id_address_delivery = $newAddressId;
            $this->update();
        }
```

jolelievre:
@PierreRambaud Thanks, I just made the optim ;)

src/Adapter/Configuration.php
```diff
@@ -124,21 +141,25 @@ public function get($key, $default = null)
      *
      * @param string $key
      * @param mixed $value
+     * @param ShopConstraint|null $shopConstraint
      * @param array $options Options
```
eternoendless: Please deprecate this parameter

src/Adapter/Configuration.php
Progi1984: It's a BC Break (you change signature of a public method). Please document it in PR.

src/Core/Domain/Configuration/ShopConfigurationInterface.php
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public function set($key, $value, ShopConstraint $shopConstraint = null);
+
+    /**
+     * @param string $key
+     * @param ShopConstraint|null $shopConstraint
+     *
+     * @return mixed
+     */
+    public function has($key, ShopConstraint $shopConstraint = null);
```
Progi1984:
```suggestion
     * @return bool
     */
    public function has($key, ShopConstraint $shopConstraint = null): bool;
```

src/Core/Domain/Configuration/ShopConfigurationInterface.php
```diff
@@ -0,0 +1,61 @@
...
+    public function get($key, $default = null, ShopConstraint $shopConstraint = null);
+
+    /**
+     * @param string $key
+     * @param mixed $value
+     * @param ShopConstraint|null $shopConstraint
+     *
+     * @return mixed
+     */
+    public function set($key, $value, ShopConstraint $shopConstraint = null);
```
Progi1984:
```suggestion
     * @return ShopConfigurationInterface
     */
    public function set($key, $value, ShopConstraint $shopConstraint = null): ShopConfigurationInterface;
```

src/Core/Domain/Shop/ValueObject/ShopId.php
```diff
@@ -40,10 +40,12 @@ class ShopId
 
     /**
      * @param int $shopId
+     *
+     * @throws ShopException
      */
-    public function __construct($shopId)
+    public function __construct(int $shopId)
```
Progi1984, line 46: @jolelievre BC Break

Progi1984 (commented):
Small comments

src/Adapter/Configuration.php
jolelievre: It is documented

src/Adapter/Configuration.php
jolelievre: Which one? `$options` ?

jolelievre: It is still used for now.. Does it mean we keep using it for now but it will disappear (along with its feature) in the next version?

src/Core/Domain/Configuration/ShopConfigurationInterface.php
jolelievre: I can't add any strict typing because it wouldn't respect the parent interface `ConfigurationInterface`
Already tried.. got kicked!

src/Core/Domain/Shop/ValueObject/ShopId.php
jolelievre: Mmm, I have doubts on this one, because this VO was introduced in 177 but it's not used anywhere
It was probably used in a previous implementation and not cleaned after a refacto..
The thing is it was shipped in the RC1 though..

src/Core/Domain/Configuration/ShopConfigurationInterface.php
jolelievre: I can only improve the PHPDoc

src/Adapter/Configuration.php
Progi1984: Ok. I close

src/Core/Domain/Configuration/ShopConfigurationInterface.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @param string $key
+     * @param ShopConstraint|null $shopConstraint
+     *
+     * @return bool
+     */
+    public function has($key, ShopConstraint $shopConstraint = null);
```
Progi1984, line 60:
```suggestion
    public function get(string $key, $default = null, ShopConstraint $shopConstraint = null);

    /**
     * @param string $key
     * @param mixed $value
     * @param ShopConstraint|null $shopConstraint
     *
     * @return ShopConfigurationInterface
     */
    public function set(string $key, $value, ShopConstraint $shopConstraint = null): ShopConfigurationInterface;

    /**
     * @param string $key
     * @param ShopConstraint|null $shopConstraint
     *
     * @return bool
     */
    public function has(string $key, ShopConstraint $shopConstraint = null): bool;
```

src/Core/Domain/Shop/ValueObject/ShopId.php
Progi1984: If it's 177, it's ok.

Progi1984 (commented):
A last comment for strict typing the new interface.

src/Core/Domain/Configuration/ShopConfigurationInterface.php
jolelievre: Nope, for the second time ^^ it's not possible that's the one I wanted to add (especially `string` type for `$key`)
But it's not possible because `ConfigurationInterface` (the interface we are extending) is type less, so adding strict type results in an error 

```php
Fatal error: Declaration of PrestaShop\PrestaShop\Core\Domain\Configuration\ShopConfigurationInterface::get(string $key, $default = NULL, ?PrestaShop\PrestaShop\Core\Domain\Shop\ValueObject\ShopConstraint $shopConstraint = NULL) must be compatible with PrestaShop\PrestaShop\Core\ConfigurationInterface::get($key) in /Users/jo/www/prestashop-177/src/Core/Domain/Configuration/ShopConfigurationInterface.php on line 34
```

jolelievre: The best we can do is having correct PHPDoc and wait for the next major to add appropriate strict type
Besides for the return type, if we wanted to allow extending the interface (but maybe we won't need any more) we'd better return `self` instead of a strict type (same in the parent interface)

Progi1984 (dismissed):

Progi1984 (dismissed):

src/Core/Domain/Shop/ValueObject/ShopGroupId.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        $this->assertIsGreaterThanZero($shopGroupId);
+
+        $this->shopGroupId = $shopGroupId;
+    }
+
+    /**
+     * @return int
+     */
+    public function getValue()
```
sowbiba:
```suggestion
    public function getValue(): int
```

classes/Cart.php
```diff
@@ -332,6 +332,32 @@ public function updateAddressId($id_address, $id_address_new)
...
+        if (!isset($this->id_address_delivery) || (int) $this->id_address_delivery === $currentAddressId) {
+            $this->id_address_delivery = $newAddressId;
+            $this->update();
+        }
+
+        $sql = 'UPDATE `' . _DB_PREFIX_ . 'cart_product`
+        SET `id_address_delivery` = ' . $newAddressId . '
+        WHERE  `id_cart` = ' . (int) $this->id . '
+            AND `id_address_delivery` = ' . $currentAddressId;
+        Db::getInstance()->execute($sql);
```
sowbiba, line 352: Do we need a variable ? I would use sprintf here :D

src/Core/Domain/Shop/ValueObject/ShopGroupId.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        return $this->shopGroupId;
+    }
+
+    /**
+     * @param int $shopGroupId
+     *
+     * @throws ShopException
+     */
+    private function assertIsGreaterThanZero(int $shopGroupId)
```
sowbiba:
```suggestion
    private function assertIsGreaterThanZero(int $shopGroupId): void
```

classes/Cart.php
jolelievre: I just copied the initial function `Cart:updateAddressId` I don't particulary try to improve legacy code ^^

Progi1984 (dismissed):

PierreRambaud (dismissed):

khouloudbelguith:
Hi @jolelievre,

I found a new issue following these steps: (All carriers are enabled & the United States is enabled)
1. In the FO you are logged with "John Doe"
2. Add the product demo_6 "The best is yet to come' Framed poster" to the cart, 
3. In the Address Step select "My address (Miami)"
4. In the Shipping Adress, select My light carrier = €2.00 tax incl. 
5. Continue checkout
6. In the BO > View the last order created
7. For shipping Address => select "Mon addressed(France)"
8. For invoice Address => select "Mon address(France)"
9. Generate an invoice (Change Order's status to Payment accepted")
10. see error "Tax Block" => **No taxes**
NOK
image
It should be
image

PS: 
https://github.com/PrestaShop/PrestaShop/issues/21254 => Fixed
https://drive.google.com/file/d/1HDInO7eDTc1sGrgHNUdUTq3uH6WhZal6/view?usp=sharing
https://github.com/PrestaShop/PrestaShop/issues/21099 => Fixed
https://drive.google.com/file/d/1N4hVSl71SW3LcJjqJl72Y9PD-FRT7Y1Y/view?usp=sharing
Some other issues detected **by not related to this PR**:
1. https://github.com/PrestaShop/PrestaShop/issues/21267
2. https://github.com/PrestaShop/PrestaShop/issues/20409

Thanks to check and feedback.

Progi1984 (dismissed):

classes/order/OrderDetail.php
```diff
@@ -444,14 +450,37 @@ public function saveTaxCalculator(Order $order, $replace = false)
...
     }
 
+    /**
+     * Get a TaxCalculator adapted for the OrderDetail's product and the specified address
+     *
+     * @param Address $address
+     *
+     * @return TaxCalculator
+     */
+    public function getTaxCalculatorByAddress(Address $address)
```
Progi1984:
```suggestion
    public function getTaxCalculatorByAddress(Address $address): TaxCalculator
```

Progi1984 (dismissed):

Progi1984 (dismissed):

src/Adapter/Order/CommandHandler/ChangeOrderDeliveryAddressHandler.php
```diff
@@ -53,11 +67,31 @@ public function handle(ChangeOrderDeliveryAddressCommand $command)
...
+     * However the structure of deliveryOptions is still used with comma in legacy, so
+     * this method provides assurance for deliveryOption structure until major refactoring
+     *
+     * @param int $carrierId
+     *
+     * @return string
+     */
+    private function formatLegacyDeliveryOptionFromCarrierId(int $carrierId): string
+    {
+        return sprintf('%s,', $carrierId);
```
PierreRambaud:
```suggestion
        return sprintf('%d,', $carrierId);
```

src/Adapter/Order/CommandHandler/UpdateOrderShippingDetailsHandler.php
```diff
@@ -101,4 +123,22 @@ public function handle(UpdateOrderShippingDetailsCommand $command)
...
+     * However the structure of deliveryOptions is still used with comma in legacy, so
+     * this method provides assurance for deliveryOption structure until major refactoring
+     *
+     * @param int $carrierId
+     *
+     * @return string
+     */
+    private function formatLegacyDeliveryOptionFromCarrierId(int $carrierId): string
+    {
+        return sprintf('%s,', $carrierId);
```
PierreRambaud:
```suggestion
        return sprintf('%d,', $carrierId);
```

It sounds duplicate :thinking:

src/Core/Domain/Shop/ValueObject/ShopConstraint.php
```diff
@@ -0,0 +1,127 @@
...
+     *
+     * @param int $shopId
+     *
+     * @return static
+     *
+     * @throws ShopException
+     */
+    public static function shop(int $shopId): self
+    {
+        return new self($shopId, null, false);
```
PierreRambaud:
```suggestion
        return new static($shopId, null, false);
```

src/Core/Domain/Shop/ValueObject/ShopConstraint.php
```diff
@@ -0,0 +1,127 @@
...
+     *
+     * @param int $shopGroupId
+     *
+     * @return static
+     *
+     * @throws ShopException
+     */
+    public static function shopGroup(int $shopGroupId): self
+    {
+        return new self(null, $shopGroupId, false);
```
PierreRambaud:
```suggestion
        return new static(null, $shopGroupId, false);
```

src/Core/Domain/Shop/ValueObject/ShopConstraint.php
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Constraint to get configuration for all shops (the global value)
+     *
+     * @return static
+     */
+    public static function allShops(): self
+    {
+        return new self(null, null, false);
```
PierreRambaud:
```suggestion
        return new static(null, null, false);
```

src/Core/Domain/Shop/ValueObject/ShopGroupId.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @param int $shopGroupId
+     *
+     * @throws ShopException
+     */
+    private function assertIsGreaterThanZero(int $shopGroupId): void
+    {
+        if (0 >= $shopGroupId) {
+            throw new ShopException(sprintf('Shop id %s is invalid. Shop id must be number that is greater than zero.', var_export($shopGroupId, true)));
```
PierreRambaud:
```suggestion
            throw new ShopException(
                sprintf(
                    'Shop id %d is invalid. Shop id must be number that is greater than zero.', 
                    var_export($shopGroupId, true)
                )
            );
```

src/Core/Domain/Shop/ValueObject/ShopId.php
```diff
@@ -40,29 +40,33 @@ class ShopId
...
+     *
+     * @throws ShopException
      */
-    private function assertIntegerIsGreaterThanZero($shopId)
+    private function assertIsGreaterThanZero(int $shopId): void
     {
-        if (!is_int($shopId) || 0 > $shopId) {
-            throw new OrderException(sprintf('Shop id %s is invalid. Shop id must be number that is greater than zero.', var_export($shopId, true)));
+        if (0 >= $shopId) {
+            throw new ShopException(sprintf('Shop id %s is invalid. Shop id must be number that is greater than zero.', var_export($shopId, true)));
```
PierreRambaud:
```suggestion
            throw new ShopException(
                sprintf(
                    'Shop id %d is invalid. Shop id must be number that is greater than zero.', 
                    var_export($shopId, true)
                )
            );
```

src/Adapter/Order/CommandHandler/UpdateOrderShippingDetailsHandler.php
jolelievre: Yep I realized too, I moved it into the Abstract class

src/Core/Domain/Shop/ValueObject/ShopGroupId.php
```diff
@@ -0,0 +1,77 @@
...
+     * @throws ShopException
+     */
+    private function assertIsGreaterThanZero(int $shopGroupId): void
+    {
+        if (0 >= $shopGroupId) {
+            throw new ShopException(
+                sprintf(
+                    'Shop id %s is invalid. Shop id must be number that is greater than zero.',
+                    var_export($shopGroupId, true
+                    )
```
PierreRambaud:
```suggestion
                    var_export($shopGroupId, true)
```

PierreRambaud (dismissed):

Progi1984 (dismissed):

jolelievre:
@khouloudbelguith It's alright I fixed this new bug and the PR is approved once again 😉

khouloudbelguith:
Hi @jolelievre,

As described [here](https://github.com/PrestaShop/PrestaShop/pull/20903#issuecomment-704851456)
The invoice generated **after** changing the address => OK (Fixed)
But, If we generate the invoice **before** changing => NOK

Two ways to reproduce the issue:

1. In the FO you are logged with "John Doe"
2. Add the product demo_6 "The best is yet to come' Framed poster" to the cart,
3. In the Address Step select "My address (Miami)"
4. In the Shipping Adress, select My light carrier = €2.00 tax incl.
5. Continue checkout
6. In the BO > View the last order created
7. Change Status to payment accepted(Invoice generated)
8. For shipping Address => select "Mon addressed(France)"
9. For invoice Address => select "Mon address(France)"
10. Download invoice
11. see error
image


Also, you can follow these steps:

1. In the FO you are logged with "John Doe"
2. Add the product demo_6 "The best is yet to come' Framed poster" to the cart,
3. In the Address Step select "My address (Miami)"
4. In the Shipping Adress, select My light carrier = €2.00 tax incl.
5. Continue checkout
6. In the BO > View the last order created
7. Change Status to payment accepted(Invoice generated)
8.  Change The carrier for a carrier "My carrier" = €7
9. Download the invoice again
10. see error
image
11. For shipping Address => select "Mon addressed(France)"
12. For invoice Address => select "Mon address(France)"
13. Download invoice
14. see error

image


Thanks!

Progi1984 (approved):

PierreRambaud (approved):

jolelievre:
@khouloudbelguith thanks for finding this bug again! There was a big part missing, the shipping was never updated on OrderInvoice

atomiix (approved):

Robin-Fischer-PS:
Hi @jolelievre !

After a double check by @khouloudbelguith and I, it's QA :heavy_check_mark: !

Thanks :)

Progi1984:
Congratulations @jolelievre and Thanks @Robin-Fischer-PS & @khouloudbelguith

PR #20905, opened 2020-09-08
The correct instance type of the created Order

Seleda:
| Questions     | Answers
| Branch?       | develop
| Description?  | "Self" in this case in the "OrderCore" class, and the instance type will be "OrderCore". The correct type must be "Order" to use in Prestashop.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20905)

### DIFF
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1173,7 +1173,7 @@ class OrderCore extends ObjectModel
     {
         $id_order = (int) self::getIdByCartId((int) $id_cart);
 
-        return ($id_order > 0) ? new Order($id_order) : null;
+        return ($id_order > 0) ? new static($id_order) : null;
     }
 
     /**
--- a/classes/order/Order.php
+++ b/classes/order/Order.php
@@ -1173,7 +1173,7 @@ class OrderCore extends ObjectModel
     {
         $id_order = (int) self::getIdByCartId((int) $id_cart);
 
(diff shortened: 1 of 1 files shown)


Progi1984:
@Seleda Thank you for your contribution. :hugs: 

Thanks :+1:

classes/order/Order.php
```diff
@@ -1173,7 +1173,7 @@ public static function getByCartId($id_cart)
     {
         $id_order = (int) self::getIdByCartId((int) $id_cart);
 
-        return ($id_order > 0) ? new self($id_order) : null;
+        return ($id_order > 0) ? new Order($id_order) : null;
```
PierreRambaud:
```suggestion
        return ($id_order > 0) ? new static($id_order) : null;
```

If you want to return the same instance type, use `static` instead

Seleda: For example: OrderSlip::create() requires an instance of "Order". And many other places.

Seleda: new static($id_order) well

atomiix (approved):

PierreRambaud (approved):

PierreRambaud:
As QA is not needed here I merge it!

Thanks @Seleda

PR #20920, opened 2020-09-09
Fix jquery migrate being loaded two times in FO

### LINKED ISSUE #20904: jquery-migrate is loaded two times in FO
#### Describe the bug

It seems like jquery-migrate is being loaded two times in FO.

#### Expected behavior

Jquery-migrate should be loaded only once.

#### Steps to Reproduce

1/ Go to FO's home page
2 Check the console

**Screenshots**
Capture d’écran 2020-09-08 à 18 14 28

#### Additional information

* PrestaShop version: N/A
* PHP version: N/A

matthieu-rolland:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Jquery migrate was loaded two times in FO.
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20904
| How to test?  | Open any FO page, see the console, there should be no message indicating that jquery-migrate was loading two times (see capture in issue)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20920)


Progi1984 (dismissed):

themes/_core/js/theme.js
```diff
@@ -4,7 +4,6 @@ window.$ = $;
 window.jQuery = $;
 
 import './migrate-mute'
-import 'jquery-migrate';
```
jolelievre: Then when is `jquery-migrate` included I found an other reference in classic theme
If so maybe it's better to remove it in the theme, this way the core dependency will still include it even if one use another theme

matthieu-rolland: yep you're right actually, just fixed it 👍

jolelievre (approved):

Progi1984 (approved):

Sinepel:
It is a BC to rather remove its load in `Media::getJqueryPath` & rather keep its import in the new assets system?

khouloudbelguith:
Hi @matthieu-rolland,

It is ok :heavy_check_mark: 
**Before**
image

**After**
image

Thanks!

Progi1984:
Thanks @matthieu-rolland & @khouloudbelguith

PR #20929, opened 2020-09-10
Avoid loading all combinations in memory when only one is needed

### LINKED ISSUE #20928: Improve product combination / variants loading

Some products can easily have thousands of combinations / variants. The current implementation is not optimal for such products and performance takes a big hit.
Improvements are already in progress for the admin product page and some improvements could also be implemented for the front end.

I currently have a product with over 2000 combinations. Changing between variants in the product page takes almost 3 seconds.

At the moment, all combinations are queried from the database and loaded into memory in order to find the one with the correct id.  
Querying directly the desired id will improve performance.

PR to follow.

jbenezech:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fixes slow loading of product page when the product has a lot of combination. Filters the desired combination in SQL instead of in-memory
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20928
| How to test?  | composer run phpunit-legacy

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20929)

### DIFF
--- a/controllers/front/ProductController.php
+++ b/controllers/front/ProductController.php
@@ -1120,7 +1120,7 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
     {
         $combinations = $this->product->getAttributesGroups($this->context->language->id, $combinationId);
 
-        if (empty($combinations)) {
+        if ($combinations === false || !is_array($combinations) || empty($combinations)) {
             return null;
         }
 
--- a/controllers/front/ProductController.php
+++ b/controllers/front/ProductController.php
@@ -1124,7 +1124,7 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
             return null;
         }
 
-        return $combinations[0];
+        return reset($combinations);
     }
 
     /**
--- a/controllers/front/ProductController.php
+++ b/controllers/front/ProductController.php
@@ -1119,7 +1119,7 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
     public function findProductCombinationById($combinationId)
     {
         $combinations = $this->product->getAttributesGroups($this->context->language->id, $combinationId);
-        
+
         if (empty($combinations)) {
             return null;
         }
--- a/controllers/front/ProductController.php
+++ b/controllers/front/ProductController.php
@@ -1118,17 +1118,13 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
      */
     public function findProductCombinationById($combinationId)
     {
-        $foundCombination = null;
-        $combinations = $this->product->getAttributesGroups($this->context->language->id);
-        foreach ($combinations as $combination) {
-            if ((int) ($combination['id_product_attribute']) === $combinationId) {
-                $foundCombination = $combination;
-
-                break;
-            }
+        $combinations = $this->product->getAttributesGroups($this->context->language->id, $combinationId);
+        
+        if (empty($combinations)) {
+            return null;
         }
 
-        return $foundCombination;
+        return $combinations[0];
     }
 
     /**
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -4385,7 +4385,7 @@ class ProductCore extends ObjectModel
                 ';
 
         if ($id_product_attribute !== null) {
-            $sql .= ' AND `product_attribute_shop.id_product_attribute` = ' . (int) $id_product_attribute . ' ';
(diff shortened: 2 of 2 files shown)


jbenezech:
This is in WIP for 2 reasons:

1. I cannot get the tests to run locally. Seems there are issues with loading fixture, but don't have more time to investigate. Letting travis do the job for me ;)

2. There is an existing method `getAttributeCombinationsById` (https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Product.php#L2719) which would be more logical to reuse but it doesn't seem to return exactly the same data as `getAttributesGroups` so I have no idea if that would break things down the line. I chose what seemed the safest option but definitely needs review by somebody with more insights into Prestashop internals.

jbenezech:
Removed WIP status since tests have passed. My second comment still stands though

controllers/front/ProductController.php
```diff
@@ -1100,17 +1100,13 @@ protected function getProductMinimalQuantity($product)
...
+        $combinations = $this->product->getAttributesGroups($this->context->language->id, $combinationId);
 
-                break;
-            }
+        if (empty($combinations)) {
+            return null;
         }
 
-        return $foundCombination;
+        return $combinations[0];
```
Progi1984:
```suggestion
        return reset($combinations);
```

Progi1984 (commented):
Good PR. But a small change

jbenezech:
Thanks @Progi1984 . Fixed

Progi1984 (dismissed):

Progi1984:
@jbenezech Approved for me. Have you seen a performance improvement ? How ? Where ?

jbenezech:
> @jbenezech Approved for me. Have you seen a performance improvement ? How ? Where ?

Manual checks only but time spent on the server side for my sample product (2560 combinations) went from 2.8s to 1.3s (on a dual CPU/4GB DO droplet). That `findProductCombinationById` being called twice.

The rest of the time is mostly spent here, for the exact same reason: https://github.com/PrestaShop/PrestaShop/blob/19dc654b76f7e45a465958d5308ef28c2dd5dac0/controllers/front/ProductController.php#L545

I'll open another issue when this is merged as the code in there is a bit more obscure and I don't really have time to spend on it right now.

atomiix (approved):

matthieu-rolland (dismissed):

ngodefroy:
@PrestaShop/prestashop-core-developers

controllers/front/ProductController.php
```diff
@@ -1100,17 +1100,13 @@ protected function getProductMinimalQuantity($product)
...
+        $combinations = $this->product->getAttributesGroups($this->context->language->id, $combinationId);
 
-                break;
-            }
+        if (empty($combinations)) {
+            return null;
         }
 
-        return $foundCombination;
+        return reset($combinations);
```
PierreRambaud, line 1127: The executeS method can return false. be sure it's an array :thinking:

matthieu-rolland: yep, that would trigger a warning

PierreRambaud: Yes, and Warning are catch in goes to exception :)

PierreRambaud (dismissed):

atomiix (dismissed):

jbenezech:
Apologies for the delay.  
Any idea where this Travis error comes from ?  
`Could not scan for classes inside "modules/ps_themecusto//ps_themecusto.php" which does not appear to be a file nor a folder`

Progi1984:
@jbenezech I restared all failing jobs.

atomiix (approved):

PierreRambaud (approved):

SD1982:
@PrestaShop/prestashop-core-developers can we let you handle this one ?

PierreRambaud:
> @PrestaShop/prestashop-core-developers can we let you handle this one ?

Of course!

atomiix:
@jbenezech I just did some tests with and without your PR, there is no significant memory consumption improvement because the function is still called with `$id_product_attribute = null` at some point to show every possible combination on the FO product page and then as the request is cached, if you call it again, it will reuse the cached results.
I'm still validating this PR as it avoid an (potentially big) array iteration.

It's QA ✔️

PierreRambaud:
Thank you @jbenezech

PR #20943, opened 2020-09-10
We improved the code readability and added a hook

xiropot:
1. The version has to be updated only if the upgrade procedure has been completed without errors.
2. It is useful for a hook to exist after a module has been upgraded. Maybe another module needs this kind of information in order to execute some query.

| Questions     | Answers
| Type?         | improvement / new feature / refacto
| Category?     | FO / BO
| BC breaks?    | no
| Deprecations? | no

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20943)

### DIFF
--- a/src/Adapter/Module/ModuleDataUpdater.php
+++ b/src/Adapter/Module/ModuleDataUpdater.php
@@ -106,10 +106,14 @@ class ModuleDataUpdater
             if (LegacyModule::initUpgradeModule($module)) {
                 $legacy_instance = LegacyModule::getInstanceByName($name);
                 $legacy_instance->runUpgradeModule();
+                
+                if (empty($legacy_instance->getErrors()){
+                    LegacyModule::upgradeModuleVersion($name, $module->version);
+                    Hook::exec('actionModuleUpgradeAfter', array('object' => $module));
+                    return true;
+                }
 
-                LegacyModule::upgradeModuleVersion($name, $module->version);
-
-                return !count($legacy_instance->getErrors());
+                return false;
             } elseif (LegacyModule::getUpgradeStatus($name)) {
                 return true;
             }
--- a/src/Adapter/Module/ModuleDataUpdater.php
+++ b/src/Adapter/Module/ModuleDataUpdater.php
@@ -109,7 +109,7 @@ class ModuleDataUpdater
                 
                 if (empty($legacy_instance->getErrors()){
                     LegacyModule::upgradeModuleVersion($name, $module->version);
-                    Hook::exec('actionModuleUpgradeAfter', array('object' => $module));
+                    Hook::exec('actionModuleUpgradeAfter', array('module' => $module));
                     return true;
                 }
 


src/Adapter/Module/ModuleDataUpdater.php
```diff
@@ -106,10 +106,14 @@ public function upgrade($name)
...
                 $legacy_instance = LegacyModule::getInstanceByName($name);
                 $legacy_instance->runUpgradeModule();
-
-                LegacyModule::upgradeModuleVersion($name, $module->version);
-
-                return !count($legacy_instance->getErrors());
+                
+                if (empty($legacy_instance->getErrors()){
+                    LegacyModule::upgradeModuleVersion($name, $module->version);
+                    Hook::exec('actionModuleUpgradeAfter', array('object' => $module));
```
matks:
```suggestion
                    Hook::exec('actionModuleUpgradeAfter', array('module' => $module));
```
Suggestion: maybe simply provide a key that describes the content of this parameter ? :)

xiropot: Sure why not? 2 (or more) brains are better than one 💯

matks (commented):
Hi ! that is an interesting idea :)

The PR is valid for me 👍 although I provided a suggestion.

I see you removed the "How to test" from the PR description. Can you please put it back, filled with a scenario to test ? Every PR needs to be tested by our QA team (it's the only way to detect unexpected regressions) and they need to have a base scenario they can work on (see https://devdocs.prestashop.com/1.7/contribute/contribution-guidelines/pull-requests/ for details)

The very best would be to have module samples to use for this scenario. One with errors, so we can acknowledge the version number is not updated ; and one which uses the new hook.

matks:
Ah, almost forgot: in order for your new hook to be fully acknowledged we need you to follow those little steps https://devdocs.prestashop.com/1.7/development/components/hook/register-new-hook/ 😉

matks:
Oh, I see you got some feedbacks from Continous Integration 😉 
https://github.com/PrestaShop/PrestaShop/pull/20943/checks?check_run_id=1098706405
https://github.com/PrestaShop/PrestaShop/pull/20943/checks?check_run_id=1098706406

kpodemski:
Good idea! Few suggestions tho!

IMO:
- you should run cs fixer, you don't have space before bracket in your IF
- maybe you could use count instead of empty
- if this is new hook @matks gave you an instruction to follow

For the testing I suggest to create two modules, one with an original version and an upgraded one, second module hooked into this first one and doing something after an upgrade.

You should test scenario where this second module Hook would fail too.

It seems much to do but you're touching significant part of Core :)

xiropot:
@kpodemski the cs fixer seems broken in the vs code. But I will definitely prepare a proper scenario as mentioned by both of you. Thanx in advance. You have inspired me a lot and have got useful knowledge by your code (especially the simple blog)

PierreRambaud:
Hi @tapandagr ,

Thanks!

PR #20951, opened 2020-09-11
Fixed Refunds in Total Sales Calculation

### LINKED ISSUE #16770: A partial refund does not reduce the sales for the day
Bonjour,

Un de mes clients se plaint car il effectue des remboursements partiels sur certaines commandes qui génèrent des avoirs, mais le CA ne baisse pas ni dans le dashboard ni dans les statistiques.

---
#### Describe the bug

One of my clients complains because he makes partial refunds on certain orders which generate credits, but the sales does not drop either in the dashboard or in the statistics.

#### Expected behavior

Refunds should be taken into account in both modules.

#### Steps to Reproduce

1. Create an Order
2. Make a partial refund
3. Go to Dashboard
4. See error > the refund is not taken into account

**Screenshots**

image
image
image

#### Additional information

* PrestaShop version: 1.7.6.2
* PHP version: 7.1

Progi1984:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fixed Refunds in Total Sales Calculation
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #16770
| How to test?  | Cf https://github.com/PrestaShop/PrestaShop/issues/16770#issuecomment-564550504 from @khouloudbelguith<br >Don't forget to pull the PrestaShop/dashtrends#34  

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20951)

### DIFF
--- a/controllers/admin/AdminStatsController.php
+++ b/controllers/admin/AdminStatsController.php
@@ -259,49 +259,78 @@ class AdminStatsControllerCore extends AdminStatsTabController
         );
     }
 
-    public static function getTotalSales($date_from, $date_to, $granularity = false)
+    /**
+     * @param string $date_from
+     * @param string $date_to
+     * @param bool $granularity
+     * @param bool $removeRefunds
+     *
+     * @return array|false|string
+     *
+     * @throws PrestaShopDatabaseException
+     */
+    public static function getTotalSales($date_from, $date_to, $granularity = false, $removeRefunds = false)
     {
+        $sqlSales = 'SELECT'
+            . (($granularity == 'day' || $granularity == 'month') ? ' LEFT(o.invoice_date, ' . ($granularity == 'day' ? 10 : 7) . ') AS date,' : '')
+            . ' SUM((o.total_paid_tax_excl - o.total_shipping_tax_excl) / o.conversion_rate) AS sales'
+            . ' FROM `' . _DB_PREFIX_ . 'orders` o'
+            . ' LEFT JOIN `' . _DB_PREFIX_ . 'order_state` os ON o.current_state = os.id_order_state'
+            . ' WHERE o.invoice_date BETWEEN "' . pSQL($date_from) . ' 00:00:00" AND "' . pSQL($date_to) . ' 23:59:59" AND os.logable = 1'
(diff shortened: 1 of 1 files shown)


matks (approved):
Code style seems correct and I trust you and the QA analyst who will validate the PR for the SQL logic 👍

PierreRambaud (approved):

controllers/admin/AdminStatsController.php
```diff
@@ -259,49 +259,78 @@ public static function getTotalProducts()
...
+     * @param string $date_from
+     * @param string $date_to
+     * @param bool $granularity
+     * @param bool $removeRefunds
+     *
+     * @return array|false|string
+     *
+     * @throws PrestaShopDatabaseException
+     */
+    public static function getTotalSales($date_from, $date_to, $granularity = false, $removeRefunds = false)
```
jolelievre, line 272: Maybe silly question, but does this behaviour need to be optional? Or could it be the default behaviour?

I get that without this parameter it changes the behaviour and can be considered as a BC, but maybe every use cases actually need this change/fix So if all use cases need the change you can simply update the method and force to remove funds no?

It would also avoid updating the module to use this new extra parameter.

jolelievre: So the question is: does any other module (or core code) will need to use:

`AdminStatsController::getTotalSales($date_from, $date_to, $granularity, true)`
AND
`AdminStatsController::getTotalSales($date_from, $date_to, $granularity, false)`

jolelievre: (Also don't get me started with static function in a controller, that's another topic 😅  )

Progi1984: I think it's optional because when you want sales, you may want remove refunds or not. 

For static : I agree. May be in v2 for `dashtrends`.

jolelievre (approved):

PierreRambaud (approved):

sowbiba (approved):

ngodefroy:
see comments in https://github.com/PrestaShop/dashtrends/pull/34

Issue #18901, opened 2020-04-29
Change visibility of function "tryToGetAvailableIdProductAttribute" in ProductController


cdcvince:
Hello,

I would like to override the behavior of the private function in **ProductController.php** :
```
private function tryToGetAvailableIdProductAttribute($checkedIdProductAttribute)
```
Can you change the visibility to 'protected' ?

This function is used for :

```
    /**
     * If the PS_DISP_UNAVAILABLE_ATTR functionality is enabled, this method check
     * if $checkedIdProductAttribute is available.
     * If not try to return the first available attribute, if none are available
     * simply returns the input.
     *
     * @param int $checkedIdProductAttribute
     *
     * @return int
     */
```

The use case of this override is pretty straightforward, it returns the next available combination when the selected combination is not available. But maybe we don't want the default combination returned by Prestashop, but a smart mechanism to return the best combination (based on external data for example).

florine2623:
Hi @cdcvince ,

You can override this file by recreating it if the` /override` folder and putting your method in the **ProductController.php** file.

If you need customized development, you can contact a PrestaShop Agency : 
https://www.prestashop.com/en/experts?iso=&budgetMin=0&budgetMax=50000

Thanks!

cdcvince:
Thank you @florine2623 for your answer but we cannot override a **private** function. That's exactly for what private is made for.
We can only override function with **public** or **protected** visibility. Which is exactly what I asked.

eternoendless:
You're right @cdcvince. There's a problem because you can't customize that part.

This is symptomatic of a design problem in that controller, because it does too much. However, I'm not a fan of making private methods protected, firstly because it violates the open/closed principle of SOLID design, and secondly (and most importantly) because it would promote that method as part of the API, rendering it impossible to refactor without introducing a breaking change.

My first reaction would be to suggest adding a hook there. But I don't think it's the right choice. Hooks are generally meant to allow enriched behavior, not _replace_ core behavior.

The best solution in my opinion would be to refactor these methods by extracting that logic into a service. That way, the service could be decorated/replaced as needed by modules.

PR #21016, opened 2020-09-16
Enable add css and js for migrated pages

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Enable the use of `addCSS()` and `addJS()` in migrated pages, introduces ControllerStack
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes partially https://github.com/PrestaShop/PrestaShop/issues/13943
| How to test?  | See below

# How to test

Install module 
[demotestaddcss.zip](https://github.com/PrestaShop/PrestaShop/files/5232476/demotestaddcss.zip)

Open "customer" listing BO page, a "It works!" alert appears because the module successfully injected `myscript.css` inside the page.

# PR content

## addCss & friends

This PR enables the use of `addCSS()`, `addJS()`, `removeCSS()` and `removeJS()` in Symfony controllers. It was a very useful function available for legacy controllers.

So now you can do `$customerController->addJS($myJavascriptFile);` to inject a JS file into Customer BO page. That is great! 🎉 

## Usage inside hooks

If you are using a Core controller, it's very easy to use `addCSS()` or `addJS()`. Same if you are using a module Symfony controller.

**But a second issues comes:** it is actually hard to use `addCss` or `addJs` **in a hook** ⛵ !

See the following example (from the sample module)
```php
    public function hookActionCustomerGridDefinitionModifier(array $params)
    {
        
    }
```
You are inside this hook, which goal is to modify the "Customer Grid". Let's say you want to add a new column AND the JS file in charge of managing this column (for example the new column uses a specific ColorPicker). How do you inject the JS file using `addJS()`

What you need to do is:
- explore the code to find out that this hook is fired when user browses "Customer Grid BO page", which is powered by CustomerController
- retrieve the `$customerController` instance

Now that you know the Controller that you want to target is CustomerController, you can retrieve is because it is a service! So you can do this
```php
    public function hookActionCustomerGridDefinitionModifier(array $params)
    {
        $controller = $this->get('PrestaShopBundle\Controller\Admin\Sell\Customer\CustomerController');
        $controller->addJS('myscript.js');
    }
```

## Hooks fired from multiple places

On this hook we are lucky, the hook will be triggered only on Customer Grid page. So we know 100% "the controller that we need to retrieve is CustomerController". **However there are hooks that can be fired from multiple BO pages.**

➡️  _So we need a way to know "I am plugged on hook Foo, what is the Controller that handled the HTTP request ?"_

In a hook function, the hook params contain PS version, route name, request, but not controller. From request it is possible to find the controller name but it's very dirty.

**Second issue**: even you can find the controller name, but to retrieve the controller _instance_ is even more complex.

➡️ So I did this

### Use controller stack

I introduced a ControllerStack whose behavior is similar to Symfony RequestStack. It listens to Symfony kernel and registers the controllers being matched to HTTP requests.

This allows you to retrieve the "current controller".

Example:
```php
    public function hookActionCustomerGridDefinitionModifier(array $params)
    {
        $controller = $this->get('controller_stack')->getCurrentController();
        $controller->addJS('myscript.js');
    }
```

So this allows module developer to inject his JS from hook without knowing which controller handled the HTTP request.

## Documentation

As this is for module developers, I will need to write a documentation about it when it's merged

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21016)

### DIFF
--- a/classes/controller/Controller.php
+++ b/classes/controller/Controller.php
@@ -404,8 +404,6 @@ abstract class ControllerCore
      * @param string $css_media_type
      * @param int|null $offset
      * @param bool $check_path
-     *
-     * @return true
      */
     public function addCSS($css_uri, $css_media_type = 'all', $offset = null, $check_path = true)
     {
--- a/src/Adapter/LegacyContext.php
+++ b/src/Adapter/LegacyContext.php
@@ -244,7 +244,7 @@ class LegacyContext
         );
 
         if (null !== $previousContextController) {
-            $this->preserveCssAndJSAssets($previousContextController, $originCtrl);
+            $this->transferCssAndJSAssets($previousContextController, $originCtrl);
         }
 
         $originCtrl->run();
@@ -362,7 +362,7 @@ class LegacyContext
      * @param Controller $previousContextController
      * @param AdminLegacyLayoutControllerCore $originCtrl
      */
-    private function preserveCssAndJSAssets($previousContextController, AdminLegacyLayoutControllerCore $originCtrl): void
+    private function transferCssAndJSAssets($previousContextController, AdminLegacyLayoutControllerCore $originCtrl): void
     {
         if (!empty($previousContextController->js_files)) {
             array_map(function ($jsAsset) use ($originCtrl) {
--- a/src/Adapter/LegacyContext.php
+++ b/src/Adapter/LegacyContext.php
@@ -224,6 +224,11 @@ class LegacyContext
         $metaTitle = '',
         $useRegularH1Structure = true
     ) {
+        $previousContextController = null;
+        if (null !== $this->getContext()->controller) {
+            $previousContextController = $this->getContext()->controller;
+        }
+
         $originCtrl = new AdminLegacyLayoutControllerCore(
             $controllerName,
             $title,
@@ -237,6 +242,11 @@ class LegacyContext
             $metaTitle,
             $useRegularH1Structure
         );
+
+        if (null !== $previousContextController) {
+            $this->preserveCssAndJSAssets($previousContextController, $originCtrl);
+        }
+
         $originCtrl->run();
 
         return $originCtrl->outPutHtml;
@@ -342,5 +352,28 @@ class LegacyContext
     public function getAvailableLanguages()
     {
         return $this->getLanguages(false);
+    }
+
+    /**
+     * This will transfer registered CSS and JS assets
+     * from previous context controller instance
+     * to AdminLegacyLayoutController
+     *
+     * @param Controller $previousContextController
+     * @param AdminLegacyLayoutControllerCore $originCtrl
+     */
+    private function preserveCssAndJSAssets($previousContextController, AdminLegacyLayoutControllerCore $originCtrl): void
+    {
+        if (!empty($previousContextController->js_files)) {
+            array_map(function ($jsAsset) use ($originCtrl) {
+                $originCtrl->addJS($jsAsset);
+            }, $previousContextController->js_files);
+        }
+
+        if (!empty($previousContextController->css_files)) {
+            array_map(function ($cssAsset) use ($originCtrl) {
+                $originCtrl->addCss($cssAsset);
+            }, $previousContextController->css_files);
+        }
     }
 }
--- a/src/PrestaShopBundle/Controller/Api/ApiController.php
+++ b/src/PrestaShopBundle/Controller/Api/ApiController.php
@@ -29,13 +29,14 @@ namespace PrestaShopBundle\Controller\Api;
 use Exception;
 use PrestaShopBundle\Api\QueryParamsCollection;
 use Psr\Log\LoggerInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 
-abstract class ApiController
+abstract class ApiController extends Controller
 {
     /**
      * @var LoggerInterface
@@ -55,7 +56,7 @@ abstract class ApiController
      */
     protected $container;
 
-    public function setContainer(ContainerInterface $container)
+    public function setContainer(ContainerInterface $container = null)
     {
         $this->container = $container;
     }
@@ -197,16 +198,16 @@ abstract class ApiController
     /**
      * Checks if access is granted.
      *
-     * @param array $accessLevel
-     * @param string $controller name of the controller
+     * @param array $attributes
+     * @param string $subject name of the controller
      *
      * @return bool
      */
-    protected function isGranted(array $accessLevel, $controller)
+    protected function isGranted($attributes, $subject = null)
     {
         return $this->container->get('security.authorization_checker')->isGranted(
-            $accessLevel,
-            $controller . '_'
+            $attributes,
+            $subject . '_'
         );
     }
 }
--- a/src/PrestaShopBundle/EventListener/ControllerListener.php
+++ b/src/PrestaShopBundle/EventListener/ControllerListener.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * 2007-2020 PrestaShop SA and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.prestashop.com for more information.
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShopBundle\EventListener;
+
+use PrestaShopBundle\Routing\ControllerStack;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
+
+/**
+ * This class listens to Symfony kernel
+ * and register what controller is being matched with incoming requests
+ */
+class ControllerListener
+{
+    /**
+     * @var ControllerStack
+     */
+    private $controllerStack;
+
+    /**
+     * @param ControllerStack $controllerStack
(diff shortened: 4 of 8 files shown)


classes/controller/Controller.php
```diff
@@ -404,8 +404,6 @@ public function setRedirectAfter($url)
      * @param string $css_media_type
      * @param int|null $offset
      * @param bool $check_path
-     *
-     * @return true
```
matks, line 408: Not in the scope of this PR but I fixed it while working on it

src/PrestaShopBundle/Controller/Api/ApiController.php
```diff
@@ -29,13 +29,14 @@
...
 use Psr\Log\LoggerInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 
-abstract class ApiController
+abstract class ApiController extends Controller
```
matks, line 39: ApiController should be Symfony controller IMO 🤔 this provides the ability for module and shop developers to benefit from Symfony built-in feature

src/PrestaShopBundle/Controller/Api/ApiController.php
```diff
@@ -55,7 +56,7 @@ public function setLogger(LoggerInterface $logger)
      */
     protected $container;
 
-    public function setContainer(ContainerInterface $container)
+    public function setContainer(ContainerInterface $container = null)
```
matks, line 59: This change is needed to comply with Symfony controller contract

src/PrestaShopBundle/Controller/Api/ApiController.php
```diff
@@ -197,16 +198,16 @@ protected function jsonResponse(
...
      *
-     * @param array $accessLevel
-     * @param string $controller name of the controller
+     * @param array $attributes
+     * @param string $subject name of the controller
      *
      * @return bool
      */
-    protected function isGranted(array $accessLevel, $controller)
+    protected function isGranted($attributes, $subject = null)
```
matks, line 206: Needed to comply with Symfony controller contract

src/PrestaShopBundle/Resources/config/services/bundle/routing.yml
```diff
@@ -53,3 +53,9 @@ services:
         class: 'PrestaShopBundle\Routing\Linter\AdminRouteProvider'
         arguments:
             - '@router'
+
+    prestashop.bundle.routing.controller_stack:
+        class: 'PrestaShopBundle\Routing\ControllerStack'
+
+    controller_stack:
```
matks, line 60: This alias allows module developers to simply retrieve this doing `$this->getContainer()->get('controller_stack');` (there are similar Symfony aliases : `router`,  `request_stack` ...)

src/PrestaShopBundle/Controller/Api/ApiController.php
PierreRambaud: Since we change the contract, it's a BC, you can't anymore pass null in public methods.

src/PrestaShopBundle/EventListener/ControllerListener.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        $this->controllerStack = $controllerStack;
+    }
+
+    /**
+     * @param FilterControllerEvent $event
+     */
+    public function onKernelController(FilterControllerEvent $event)
+    {
+        if ($event->getController()[0] instanceof Controller) {
```
PierreRambaud, line 59:
```suggestion
        if (isset($event->getController()[0]) && $event->getController()[0] instanceof Controller) {
```

PierreRambaud: Never trust an event

src/PrestaShopBundle/Routing/ControllerStack.php
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Inspired by \Symfony\Component\HttpFoundation\RequestStack
+ */
+class ControllerStack
+{
+    /**
+     * @var Controller[]
+     */
+    private $controllers = [];
```
PierreRambaud, line 41: The class is not final, don't put this accessor private, otherwise, someone creating his stack must recreate properties.
```suggestion
    protected $controllers = [];
```

src/PrestaShopBundle/Routing/ControllerStack.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        $this->controllers[] = $controller;
+    }
+
+    /**
+     * @return Controller|null
+     */
+    public function pop()
+    {
+        if (!$this->controllers) {
```
PierreRambaud, line 56:
```suggestion
        if (empty($this->controllers)) {
```
or `!count()` since it's an array

src/PrestaShopBundle/Routing/ControllerStack.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Gets the master Controller.
+     *
+     * @return Controller|null
+     */
+    public function getRootController()
+    {
+        if (!$this->controllers) {
```
PierreRambaud, line 78: same as previous

src/PrestaShopBundle/Routing/ControllerStack.php
matks: 🤔 Do you mean we should always introduce `protected` properties for non final classes, never `private` unless wanting to ?

src/PrestaShopBundle/Routing/ControllerStack.php
matks: My code was inspired of Symfony 😄 
https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/RequestStack.php

src/PrestaShopBundle/Routing/ControllerStack.php
PierreRambaud: Yes, otherwise you must rewrite everything due to lack of visibility :/

src/PrestaShopBundle/Routing/ControllerStack.php
PierreRambaud: :sob: Why they did this...

src/Adapter/LegacyContext.php
```diff
@@ -224,6 +224,11 @@ public function getLegacyLayout(
         $metaTitle = '',
         $useRegularH1Structure = true
     ) {
+        $previousContextController = null;
+        if (null !== $this->getContext()->controller) {
+            $previousContextController = $this->getContext()->controller;
+        }
```
Progi1984, line 230:
```suggestion
        $previousContextController = $this->getContext()->controller;
```

src/PrestaShopBundle/Controller/Api/ApiController.php
Progi1984: Added the label "BC Break". @matks Could you document in PR the BC Break, plz ? Thanks

src/PrestaShopBundle/Controller/Api/ApiController.php
Progi1984: BC Break

src/PrestaShopBundle/Controller/Api/ApiController.php
Progi1984: BC Break

src/PrestaShopBundle/Resources/config/services/bundle/event_listener.yml
```diff
@@ -101,3 +101,10 @@ services:
           - '@logger'
         tags:
           - { name: kernel.event_subscriber }
+
+    prestashop.bundle.controller_stack:
+        class: PrestaShopBundle\EventListener\ControllerListener
+        arguments:
+          - '@prestashop.bundle.routing.controller_stack'
+        tags:
+          - { name: kernel.event_listener, event: kernel.controller, method: onKernelController }
```
Progi1984, line 110:
```suggestion
          - name: kernel.event_listener
            event: kernel.controller
            method: onKernelController
```

src/PrestaShopBundle/Controller/Api/ApiController.php
matks: If I am not wrong, this change is backward compatible 🤔 ?

Progi1984 (changes requested):
Need some changes requested by @PierreRambaud and me

src/PrestaShopBundle/Controller/Api/ApiController.php
matks: If I am not wrong, this change is backward compatible 🤔 ?

src/PrestaShopBundle/Controller/Api/ApiController.php
Progi1984: Oops. You're right.

src/PrestaShopBundle/Routing/ControllerStack.php
matks: I'm actually OK with that 😄 but I'm afraid The Big Boss will be mad against me because [he said he disliked making private methods protected](https://github.com/PrestaShop/PrestaShop/issues/18901#issuecomment-623335766).

I feel like we have 2 schools on this debate: "use protected, be extensible" to allow better customization versus "use final and private, promote composition over inheritance".

@eternoendless Could you please tell us your vision on this topic ?

src/PrestaShopBundle/Routing/ControllerStack.php
matks: I will fix this one and submit a PR on Symfony side :trollface:

src/PrestaShopBundle/Controller/Api/ApiController.php
matks: If I am not wrong, this change is backward compatible 🤔 ? The class `Symfony\Bundle\FrameworkBundle\Controller\Controller` contains very few logic, and all of them are compatible with this class.

src/PrestaShopBundle/Routing/ControllerStack.php
eternoendless: By the open/closed principle, classes should be open to extension and closed to modification.
If you make a property protected, it becomes part of your contract. Are you willing to make that commitment?

src/PrestaShopBundle/Controller/Api/ApiController.php
PierreRambaud: You're right, sometimes I read the red line as a new line -_-

src/PrestaShopBundle/Routing/ControllerStack.php
PierreRambaud: If you use private properties, you must provide accessors in order to be able to use others methods like before.
In this case, if `$controllers` is private, add `getControllers()` and use `$this->getControllers()` instead of `$this->controllers`.
Otherwise, you can extend the class, but you're not able to change anything because you can't call `$this->controllers`

matks: @PierreRambaud It makes sense. Nice feedback, I will apply it !

matks: => next part of this discussion https://github.com/PrestaShop/PrestaShop/issues/21383

matks:
I wont be able to finish this PR 😞 . I close it. if anybody wants to follow, be my guest and cherry-pick my commits.

matks:
Related https://github.com/PrestaShop/PrestaShop/pull/36433

PR #21018, opened 2020-09-16
Throw exception if bad JS path provided to addJS

### LINKED ISSUE #21019: Controller::addJS fails silently if bad input is provided and path check is required
#### Describe the bug

Today I lost around 2 hours because `Controller::addJS()` fails silently if bad input is provided and path check is required.

See the implementation:
```php
    public function addJS($js_uri, $check_path = true)
    {
        if (!is_array($js_uri)) {
            $js_uri = [$js_uri];
        }

        foreach ($js_uri as $js_file) {
            $js_file = explode('?', $js_file);
            $version = '';
            if (isset($js_file[1]) && $js_file[1]) {
                $version = $js_file[1];
            }
            $js_path = $js_file = $js_file[0];
            if ($check_path) {
                $js_path = Media::getJSPath($js_file);
            }

            if ($js_path && !in_array($js_path, $this->js_files)) {
                $this->js_files[] = $js_path . ($version ? '?' . $version : '');
            }
        }
    }
```

If you use `check_path` default value (true) and provide a `$js_uri` that does not refer to a real JS file ; then `Media::getJSPath($js_file)` will return `false`.
➡️ Then `$js_path` will be false
➡️ The next if `if ($js_path && !in_array($js_path, $this->js_files))` will fail so `$js_uri` will not be added to `$this->js_files[]`.

What happened is that, at some moment, by mistake, I modified the `$js_uri` I was passing to `addJS()` and then my JS file was not included anymore. I passed `script.js` instead of `myscript.js` and this was enough to create the issue.

I thought the issue would come from an other area of code (the one I was working on) so I searched 2 hours until I used the debugger to finally see how my file would be ignored. 😞 

#### Expected behavior

I wish that, instead of not adding my JS file silently, PrestaShop had thrown at me a big red error message telling me "Mathieu ! You are so lame ! You are a failure as a developer and a human being 😠 ! Look at your mess ! You told me to add the JS file myscript.js and it does not exist !" instead of failing silently and letting me search through the issue for hours.

I submit a PR to implement this: https://github.com/PrestaShop/PrestaShop/pull/21018

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | Throw exception if bad JS path provided to addJS, instead of failing silently
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/21019
| How to test?  | See issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21018)

### DIFF
--- a/classes/controller/Controller.php
+++ b/classes/controller/Controller.php
@@ -500,7 +500,7 @@ abstract class ControllerCore
             if ($check_path) {
                 $js_path = Media::getJSPath($js_file);
                 if ($js_path === false) {
-                    throw new PrestaShopException(sprintf('Cannot compute the path for JS file %s', $js_uri));
+                    throw new PrestaShopException(sprintf('Cannot compute the path for JS file %s', $js_file));
                 }
             }
 
--- a/classes/controller/Controller.php
+++ b/classes/controller/Controller.php
@@ -499,6 +499,9 @@ abstract class ControllerCore
             $js_path = $js_file = $js_file[0];
             if ($check_path) {
                 $js_path = Media::getJSPath($js_file);
+                if ($js_path === false) {
+                    throw new PrestaShopException(sprintf('Cannot compute the path for JS file %s', $js_uri));
+                }
             }
 
             if ($js_path && !in_array($js_path, $this->js_files)) {


classes/controller/Controller.php
```diff
@@ -499,6 +499,9 @@ public function addJS($js_uri, $check_path = true)
             $js_path = $js_file = $js_file[0];
             if ($check_path) {
                 $js_path = Media::getJSPath($js_file);
+                if ($js_path === false) {
+                    throw new PrestaShopException(sprintf('Cannot compute the path for JS file %s', $js_uri));
```
PierreRambaud:
```suggestion
                    throw new PrestaShopException(sprintf('Cannot compute the path for JS file %s', $js_file));
```

classes/controller/Controller.php
```diff
@@ -499,6 +499,9 @@ public function addJS($js_uri, $check_path = true)
             $js_path = $js_file = $js_file[0];
             if ($check_path) {
                 $js_path = Media::getJSPath($js_file);
+                if ($js_path === false) {
```
PierreRambaud, line 502: It's a BC break, the old method wasn't throwing anything.

You have to change the PHPDoc for this method and add ` * @throw PrestaShopException`

matks:
I wont have the time to continue this topic

PR #21062, opened 2020-09-18
Make locale errors from translatable type work with locale

JevgenijVisockij:
…e FormattableTextArea compatible with translatableType

| Questions     | Answers
| Branch?       | develop
| Description?  | This PR is meant to make locale errors from translatable type work with simplified forms as well as make FormattedTextareaType work with TranslatableType
| Type?         | refacto
| Category?     | BO
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | 
| How to test?  | This PR best to be tested with pr to example-modules(https://github.com/PrestaShop/example-modules/pull/28/).That PR contains simplified form with translatable fields and errors. Please test the form behaves in a relevant way.<br/><br/>Also some of the already existing translatable fields need to be tested to make sure they were not affected. Please test 1 or 2 BO page with fields that can be translated to check the form errors are working as expected.

BC break because TranslatableType now extends TranslatorAwareType so any module that extended TranslatableType before might run into problems. 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21062)

### DIFF
--- a/src/PrestaShopBundle/Form/Admin/Type/TranslatableType.php
+++ b/src/PrestaShopBundle/Form/Admin/Type/TranslatableType.php
@@ -26,21 +26,22 @@
 
 namespace PrestaShopBundle\Form\Admin\Type;
 
-use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormError;
 use Symfony\Component\Form\FormErrorIterator;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
 use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Translation\TranslatorInterface;
 
 /**
  * Class TranslatableType adds translatable inputs with custom inner type to forms.
  * Language selection uses a dropdown.
  */
-class TranslatableType extends AbstractType
+class TranslatableType extends TranslatorAwareType
 {
     /**
      * @var array List of enabled locales
@@ -73,6 +74,8 @@ class TranslatableType extends AbstractType
     private $defaultShopLanguageId;
 
     /**
+     * @param TranslatorInterface $translator
+     * @param array $locales
      * @param array $availableLocales
      * @param UrlGeneratorInterface $urlGenerator
      * @param bool $saveFormLocaleChoice
@@ -80,12 +83,15 @@ class TranslatableType extends AbstractType
      * @param int $defaultShopLanguageId
      */
     public function __construct(
+        TranslatorInterface $translator,
+        array $locales,
         array $availableLocales,
         UrlGeneratorInterface $urlGenerator,
         $saveFormLocaleChoice,
         $defaultFormLanguageId,
         $defaultShopLanguageId
     ) {
+        parent::__construct($translator, $locales);
         $this->enabledLocales = $this->filterEnableLocales($availableLocales);
         $this->availableLocales = $availableLocales;
         $this->urlGenerator = $urlGenerator;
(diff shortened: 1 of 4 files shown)


matks (dismissed):
Code is OK but we need to look at one or two migrated pages using TranslatableType to be sure that these changes do not affect badly how errors are displayed today.

We might find that we need to update a little every page already using TranslatableType

src/PrestaShopBundle/Form/Admin/Type/TranslatableType.php
```diff
@@ -116,6 +122,31 @@ public function buildForm(FormBuilderInterface $builder, array $options)
      */
     public function buildView(FormView $view, FormInterface $form, array $options)
     {
+        $errors = [];
+
+        foreach ($view->vars['errors'] as $existingError) {
+            $errors[] = $existingError;
```
sowbiba: based on https://github.com/symfony/symfony/blob/v3.4.43/src/Symfony/Component/Form/FormView.php
index `errors` can be undefined in `$view->vars` array

If it's defined and is an array, why not `$errors = $view->vars['errors']`

So can you try `$errors = $view->vars['errors'] ?? [];`

JevgenijVisockij: This makes sense, thanks!

JevgenijVisockij: Ahh nevermind. $view->vars['errors'] is an iterator not an array. And it won't allow me to edit it, thats why in line after it there is foreach trough that itterator to get errors one by one.

src/PrestaShopBundle/Resources/views/Admin/macros.html.twig
```diff
@@ -142,11 +142,6 @@
     {% endfor %}
 
     {#iterating over child errors because errors can be nested#}
```
JevgenijVisockij, line 144: Reasoning for that is that now I am collecting all the errors from children in TranslatableType so this is no longer needed. As far as I tested translatable types still work and according due commits and comment this was added specifically for translatable type, but I am not 100% sure this won't affect something else.

matks: Well, QA will be able to help us asserting the 100% 😉

matks (dismissed):

src/PrestaShopBundle/Form/Admin/Type/TranslatableType.php
PierreRambaud: you can use ̀iterator_to_array`or try to cast it to array ?

src/PrestaShopBundle/Form/Admin/Type/TranslatableType.php
```diff
@@ -116,6 +122,31 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+        }
+
+        $errorsByLocale = $this->getErrorsByLocale($view, $form, $options['locales']);
+
+        if (!empty($errorsByLocale)) {
```
PierreRambaud:
```suggestion
        if ($errorsByLocale !== null) {
```

src/PrestaShopBundle/Form/Admin/Type/TranslatableType.php
JevgenijVisockij: iterator_to_array to array worked perfectly, thanks!

matks (approved):

PierreRambaud (approved):

Robin-Fischer-PS:
Hi @JevgenijVisockij !

I have an error on the "demosymfonyform" module : When I fill the fields with valid values, if I save, there is an exception : 

pr21062 exception module

However, I think this is only related to the module. Otherwise the PR seems OK for me, so if you can confirm that the error comes from the module and not the PR, I will validate your PR :)

JevgenijVisockij:
> Hi @JevgenijVisockij !
> 
> I have an error on the "demosymfonyform" module : When I fill the fields with valid values, if I save, there is an exception :
> 
> pr21062 exception module
> 
> However, I think this is only related to the module. Otherwise the PR seems OK for me, so if you can confirm that the error comes from the module and not the PR, I will validate your PR :)

Hey thanks for testing!

Yes can confirm that error is strictly module related.

Robin-Fischer-PS:
Thanks @JevgenijVisockij , then it's QA :heavy_check_mark: :)

Progi1984:
Thanks @JevgenijVisockij & @Robin-Fischer-PS

PR #21093, opened 2020-09-21
Fix function typo when changing address on create order

### LINKED ISSUE #20238: BO - Orders page - Cannot change the address selected in the New Order page - The address-details is not updated
#### Describe the bug

In the BO => Orders => New Order page, it is impossible to change the address selected
With  branch 1.7.7.x => OK
#### Expected behavior

The address should be changed

#### Steps to Reproduce

1. Go to BO => Orders
2. Click on "Add new Order"
3. Create a new customer or choose an old one for example `pub@prestashop.com`
PS: this customer has more than one address
4. add a product to cart
5. Try to change the address of the customer, The address-details is not updated (Delivery OR Invoice)
 
**Screenshots**

https://drive.google.com/file/d/1N0Yh-c5-Y75rtlXnWu8z8ZTQ70a0nAyl/view?usp=sharing

#### Additional information

* PrestaShop version: develop
* PHP version: 7.2

sowbiba:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fix a function call typo when changing customer address on create order page
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #20238 
| How to test?  | 1. Go to BO => Orders<br>2. Click on "Add new Order"<br>3. Create a new customer or choose an old one for example `pub@prestashop.com`<br>   PS: this customer has more than one address<br>4. add a product to cart<br>5. Try to change the address of the customer on delivery or invoice, The address-details must be updated
 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21093)

### DIFF
--- a/admin-dev/themes/new-theme/js/pages/order/create/create-order-page.js
+++ b/admin-dev/themes/new-theme/js/pages/order/create/create-order-page.js
@@ -240,7 +240,7 @@ export default class CreateOrderPage {
 
     this.$container.on('change', createOrderMap.listedProductUnitPriceInput, (e) => this.initProductChangePrice(e));
     this.$container.on('change', createOrderMap.listedProductQtyInput, (e) => this.initProductChangeQty(e));
-    this.$container.on('change', createOrderMap.addressSelect, () => this.nchangeCartAddresses());
+    this.$container.on('change', createOrderMap.addressSelect, () => this.changeCartAddresses());
     this.$container.on('click', createOrderMap.productRemoveBtn, (e) => this.initProductRemoveFromCart(e));
   }
 


matks (approved):

atomiix (approved):

NeOMakinG (approved):

ngodefroy:
Hello @sowbiba 

It seems that the delivery  address & invoice is changed at the same time when I update one of each: https://drive.google.com/file/d/1lezNE4_9_AutjciogVsFtzGJSHtgZVJx/view

Also, I don't have the previous address listed in the dropdown list, it's not saved.

@PrestaShop/product-team  can you confirm this is the right behavior to follow?

Thank you

matks:
> Hello @sowbiba
> 
> It seems that the delivery address & invoice is changed at the same time when I update one of each: 

~This is wanted, it allows you to choose a different address for shipping and delivery.
If, when adresses are identical, when user modifies one address we update also the other address ; user will never be able to have 2 different adresses 😅 .~

~Note that merchant can however go into the BO page dedicated to adresses and modify one address ; modifications will be applied everywhere.~

EDIT: my bad it's the other way

> Also, I don't have the previous address listed in the dropdown list, it's not saved.

I think this is because, when you edit an address, it actually does not "edit" it but rather create a new address and deletes the previous. So the previous is deleted, which is why it does not appear in the dropdown.

**However** if I am not wrong, this works:

1. Make one order, use address X for shipping and delivery
2. Make a 2nd order, use again address X
3. Modify 2nd order shipping address. Now the dropdown shows both the old and the new

This is because we only delete an address once it's not linked/used by any cart/order.

sowbiba (commented):
Hi @ngodefroy 
In my point of view, this is normal. It's the address object which is modified. As it's the same one on delivery and invoice, it change on both blocks. Maybe the 2 edit buttons is a bit confusing, it's like you can edit one of them and not the other

@PrestaShop/product-team wdyt ?

PierreRambaud (approved):

Progi1984 (approved):

jolelievre:
@MatShir I saw your comment in the issue https://github.com/PrestaShop/PrestaShop/issues/20238#issuecomment-698250628

That it is blocked because we need to re-think this feature. I don't see any reason why since this issue has been raised lots of time and the duplication/soft-delete behaviour is not instinctive, so it probably deserves some improvement on UX.

However we can keep the issue opened for discussion but this particular PR should be merged, it only fixes a typo that was very likely introduce during a merge from 177. `_changeCartAddresses` was renamed into `changeCartAddresses` but was falsely rewritten as `nchangeCartAddresses` (starts with an `n`) This is a human error that should be fixed regardless of the product decision (everywhere else the method was correctly renamed).

jolelievre (approved):

MatShir:
You are right @jolelievre! We could work on the UX matter on another PR/issue. I thought the PR would change some behaviors. I didn't mean to block the issue for this matter.

ngodefroy:
Hello @sowbiba & @MatShir 

following the test steps:
1. Go to BO => Orders
2. Click on "Add new Order"
3. Create a new customer or choose an old one for example pub@prestashop.com
PS: this customer has more than one address
4. add a product to cart
5. Try to change the address of the customer on delivery or invoice, The address-details must be updated

=> As mentioned above, the delivery and invoice address gets updated at the same time, if you can confirm me this is the expected behavior, then QA is OK.

Thank you

MatShir:
@ngodefroy The bahavior is OK for this PR. Thank you all!

NeOMakinG:
Thanks @sowbiba

PR #21094, opened 2020-09-21
Fixed array parameter processing in Link::getCategoryObject method

### LINKED ISSUE #21136: Link::getCategoryLink doesn't handle array input correctly
#### Describe the bug

When I use an array input to create a category link like this:

`Context::getContext()->link->getCategoryLink(array('id_category' => Category::getRootCategory()->id));`

I don't get the expected result, because internally `Link::getCategoryObject` is buggy and returns the root category so I end up with a false url:

`http://local.prestashop-develop/en/2-root` (correct ID wrong rewrite)
Whereas the expected url is: `http://local.prestashop-develop/en/2-home` (all is good)

#### Expected behavior

I can use `Link::getCategoryLink` with an array input and the url is correctly generated

#### Steps to Reproduce

1. Install `ps_qualityassurance` module
2. Register to a generic hook like `displayAdminAfterHeader`
3. Use this code `return Context::getContext()->link->getCategoryLink(array('id_category' => Category::getRootCategory()->id));`
4. Go in a page (let's say Category edition for example)
5. You should see the correct URL for Home category (not Root category)

**Screenshots**

Wrong behaviour:

Capture d’écran 2020-09-24 à 12 22 37

Correct behaviour:

Capture d’écran 2020-09-24 à 12 21 59

#### Additional information

* PrestaShop version: 1.7.8 (at least)
* PHP version: N/A

gfilippakis:
| Questions     | Answers
| Branch?       | develop
| Description?  | Fixed array parameter processing in Link::getCategoryObject method.
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #21136
| How to test?  | See issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21094)

### DIFF
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -394,7 +394,7 @@ class LinkCore
     public function getCategoryObject($category, $idLang)
     {
         if (!is_object($category)) {
-            if (is_array($category) && isset($category['id_category'])) {
+            if (isset($category['id_category'])) {
                 $category = new Category($category['id_category'], $idLang);
             } elseif ((int) $category) {
                 $category = new Category((int) $category, $idLang);
@@ -438,11 +438,9 @@ class LinkCore
         $params = [];
         if (Validate::isLoadedObject($category)) {
             $params['id'] = $category->id;
-        } elseif (is_array($category) && isset($category['id_category'])) {
+        } elseif (isset($category['id_category'])) {
             $params['id'] = $category['id_category'];
-        } elseif (is_int($category)) {
-            $params['id'] = $category;
-        } elseif (ctype_digit($category)) {
+        } elseif (is_int($category) or ctype_digit($category)) {
             $params['id'] = (int) $category;
         } else {
             throw new \InvalidArgumentException('Invalid category parameter');
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -436,11 +436,12 @@ class LinkCore
 
         // Set available keywords
         $params = [];
-
         if (Validate::isLoadedObject($category)) {
             $params['id'] = $category->id;
         } elseif (is_array($category) && isset($category['id_category'])) {
             $params['id'] = $category['id_category'];
+        } elseif (is_int($category)) {
+            $params['id'] = $category;
         } elseif (ctype_digit($category)) {
             $params['id'] = (int) $category;
         } else {
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -437,10 +437,14 @@ class LinkCore
         // Set available keywords
         $params = [];
 
-        if (!is_object($category)) {
-            $params['id'] = $category;
-        } else {
+        if (Validate::isLoadedObject($category)) {
             $params['id'] = $category->id;
+        } elseif (is_array($category) && isset($category['id_category'])) {
+            $params['id'] = $category['id_category'];
+        } elseif (ctype_digit($category)) {
+            $params['id'] = (int) $category;
+        } else {
+            throw new \InvalidArgumentException('Invalid category parameter');
         }
 
         // Selected filters is used by the module ps_facetedsearch
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -395,7 +395,7 @@ class LinkCore
     {
         if (!is_object($category)) {
             if (is_array($category) && isset($category['id_category'])) {
-                $category = new Category($category, $idLang);
+                $category = new Category($category['id_category'], $idLang);
             } elseif ((int) $category) {
                 $category = new Category((int) $category, $idLang);
             } else {


Progi1984 (dismissed):

PierreRambaud (dismissed):

PierreRambaud:
Not sure QA is needed here, it can be done by dev only.

Progi1984:
@PrestaShop/qa-functional Ok for the QA by a dev ?

SimonGrn:
There is no linked issue.
We are okay for a dev only QA on this one.

Edit by @Robin-Fischer-PS : Please make sure you did some regression test for this PR (since it's technical I cannot suggest what to test...).

jolelievre:
Hi @gfilippakis,

I tested your PR, it works fine to get back the correct object and its fields (rewrite, meta fields and so on) because as you spotted the array input wasn't handled correctly. However to be consistent I think you should also fix the `Category::getCategoryLink` code, the part where it gets the ID from the input:

```php
        if (!is_object($category)) {
            $params['id'] = $category;
        } else {
            $params['id'] = $category->id;
        }
```

It should handle the array just like it does in `getCategoryObject`, because with the example you provided we end up with:

```php
$params = [
    'id' => [
        'id_category' => 2,
    ]
]
```

when we would expect `$params = ['id' => 2]`

So it does work in the because the Dispatcher::createLink is able to fallback and takes the first value from the array, but it's almost out of luck. It would be more stable and more consistent to correctly retrieve the ID from the array input.

Other than that I test your fix and it works correctly:
Before url was `http://local.prestashop-develop/en/2-root` (correct ID wrong rewrite)
With you fix: `http://local.prestashop-develop/en/2-home` (all is good)

jolelievre (changes requested):
Please also fix the `Link::getCategoryLink` function on Line 440, as explained in my comment.

jolelievre:
I also created the issue regarding this bug with steps to test and reproduce the bug 😉

gfilippakis:
> 
> 
> Hi @gfilippakis,
> 
> I tested your PR, it works fine to get back the correct object and its fields (rewrite, meta fields and so on) because as you spotted the array input wasn't handled correctly. However to be consistent I think you should also fix the `Category::getCategoryLink` code, the part where it gets the ID from the input:
> 
> ```
>         if (!is_object($category)) {
>             $params['id'] = $category;
>         } else {
>             $params['id'] = $category->id;
>         }
> ```
> 
> It should handle the array just like it does in `getCategoryObject`, because with the example you provided we end up with:
> 
> ```
> $params = [
>     'id' => [
>         'id_category' => 2,
>     ]
> ]
> ```
> 
> when we would expect `$params = ['id' => 2]`
> 
> So it does work in the because the Dispatcher::createLink is able to fallback and takes the first value from the array, but it's almost out of luck. It would be more stable and more consistent to correctly retrieve the ID from the array input.
> 
> Other than that I test your fix and it works correctly:
> Before url was `http://local.prestashop-develop/en/2-root` (correct ID wrong rewrite)
> With you fix: `http://local.prestashop-develop/en/2-home` (all is good)


Feel free to modify other methods as you deem appropriate. The bug was just too obvious to let it be.

Link class is still a mess in 1.7, so there are many ways to improve it. 
Since we're allowing different types for the $category parameter in Category::getCategoryLink, I would prefer:
```
if (Validate::isLoadedObject($category)) {
    $params['id'] = $category->id;
} elseif (is_array($category) && isset($category['id_category'])) {
    $params['id'] = $category['id_category'];
} elseif (is_integer($category)) {
    $params['id'] = $category;
}
```
for more strict type and value checking.

jolelievre:
I just pushed the improvement for `Link::getCategoryLink` This PR will need to be re-approved and then tested by a dev from @PrestaShop/prestashop-core-developers

classes/Link.php
```diff
@@ -437,10 +437,14 @@ public function getCategoryLink(
         // Set available keywords
         $params = [];
 
-        if (!is_object($category)) {
-            $params['id'] = $category;
-        } else {
+        if (Validate::isLoadedObject($category)) {
             $params['id'] = $category->id;
+        } elseif (is_array($category) && isset($category['id_category'])) {
```
PierreRambaud:
```suggestion
        } elseif (isset($category['id_category'])) {
```

classes/Link.php
```diff
@@ -395,7 +395,7 @@ public function getCategoryObject($category, $idLang)
     {
         if (!is_object($category)) {
             if (is_array($category) && isset($category['id_category'])) {
```
PierreRambaud:
```suggestion
            if (isset($category['id_category'])) {
```

PierreRambaud (dismissed):

sowbiba (dismissed):

Progi1984:
@boubkerbribri Have you got an idea why sanity tests crash : https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/398861047#L1013 ?

boubkerbribri:
> @boubkerbribri Have you got an idea why sanity tests crash : https://travis-ci.com/github/PrestaShop/PrestaShop/jobs/398861047#L1013 ?

I will check this but we have the same error in all sanity tests and it means that something has changed (or gone wrong)

boubkerbribri:
@Progi1984 I found the error Why the sanity tests failed.

We have an exception when accessing the front office.

image


@gfilippakis Can you check the FO and fix the error please

Progi1984:
Thanks @boubkerbribri :)

gfilippakis:
> 
> 
> @Progi1984 I found the error Why the sanity tests failed.
> 
> We have an exception when accessing the front office.
> 
> image
> 
> @gfilippakis Can you check the FO and fix the error please

You should ask @jolelievre to fix his code since he added ctype_digit without really paying attention to the relevant note in [ctype_digit](https://www.php.net/manual/en/function.ctype-digit.php) .

jolelievre:
My bad I didn't check the FO indeed, I'm gonna look at it although I'm surprised why `ctype_digit` would be a bad choice it's supposed to tell if a string is composed of integers no?

jolelievre:
Hi @gfilippakis 

I just fixed it, I didn't expect ctype_digit (or even PHP) to be this strict for once ^^
So if the input was of type `int` it simply returned `false` I'll remember this one!

Progi1984 (dismissed):

sowbiba (dismissed):

Progi1984:
@PrestaShop/qa-functional OK for QA Dev ?

khouloudbelguith:
Hi @Progi1984,

>@PrestaShop/qa-functional OK for QA Dev ?

Yes, Thank you!

PierreRambaud (approved):

jolelievre (approved):

matosagencia (commented):
`// Set available keywords
        $params = [];
if ($category instanceof Category) {
    $params['id'] = $category->id;
} elseif (is_array($category) && isset($category['id_category'])) {
    $params['id'] = $category['id_category'];
} elseif (is_int($category) || ctype_digit($category)) {
    $params['id'] = (int) $category;
} else {
    throw new \InvalidArgumentException('Invalid category parameter');
}`

i change this because in my case i have the module pssitemap that send the values but the code link.php cannot identify if is an array correctly.

PR #21103, opened 2020-09-22
Introduce UpdateProductStockCommand

jolelievre:
| Questions     | Answers
| Branch?       | develop
| Description?  | Introduce UpdateProductStockCommand, it is able to update `Product` fields, along with `StockAvailable` field and update `StockMovements`
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | Travis green

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21103)


src/Adapter/Product/Provider/StockAvailableProvider.php
```diff
@@ -0,0 +1,114 @@
...
+class StockAvailableProvider extends AbstractObjectModelProvider
+{
+    /**
+     * @param ProductId $productId
+     *
+     * @return StockAvailable
+     *
+     * @throws CoreException
+     */
+    public function getOrCreate(ProductId $productId): StockAvailable
```
zuk3975: This could bring some discussions, but
personally I would avoid method which defines multiple repsonsibility in its name.
Idk if it is necessary, e.g. if its used in single place only, maybe this method is not needed at all ( you could check for get/create in that specific place only)? Or maybe we could use something like `provide` for this case?

src/Core/Domain/Product/Pack/ValueObject/OutOfStockType.php
```diff
@@ -0,0 +1,99 @@
...
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Pack\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductStockException;
+
+class OutOfStockType
+{
+    /**
+     * Product is available for order even when out of stock.
+     */
+    const OUT_OF_STOCK_AVAILABLE = 'out_of_stock_available';
```
zuk3975: Im afraid this might get confusing. 
Wouldn't it be better just to leave these constants as integers and use it as the only source of truth (instead of creating more of them in Product obj model and then doing additional mapping)?
Using integers is not too painful when we have constants

src/Core/Domain/Warehouse/ValueObject/WarehouseId.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public function getValue()
+    {
+        return $this->warehouseId;
+    }
+
+    /**
+     * @param int $warehouseId
+     */
+    private function assertIntegerIsGreaterThanZero($warehouseId)
```
zuk3975: missing property type

src/Core/Domain/Warehouse/ValueObject/WarehouseId.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        if (!is_int($warehouseId) || 0 > $warehouseId) {
```
zuk3975: is_int can be thrown away when property is typed

classes/Product.php
```diff
@@ -4221,7 +4225,7 @@ public static function isAvailableWhenOutOfStock($out_of_stock)
 
         $ps_order_out_of_stock = Configuration::get('PS_ORDER_OUT_OF_STOCK');
 
-        return (int) $out_of_stock == 2 ? (int) $ps_order_out_of_stock : (int) $out_of_stock;
+        return (int) $out_of_stock == static::DEFAULT_OUT_OF_STOCK_TYPE ? (int) $ps_order_out_of_stock : (int) $out_of_stock;
```
Progi1984:
```suggestion
        return (int) $out_of_stock === static::DEFAULT_OUT_OF_STOCK_TYPE ? (int) $ps_order_out_of_stock : (int) $out_of_stock;
```

src/Adapter/AbstractObjectModelFiller.php
```diff
@@ -0,0 +1,108 @@
...
+use ObjectModel;
+use PrestaShop\PrestaShop\Core\Exception\CoreException;
+
+class AbstractObjectModelFiller
+{
+    /**
+     * @param ObjectModel $objectModel
+     * @param string $propertyName
+     * @param array $properties
+     */
```
Progi1984:
```suggestion
     * @param array $properties
     *
     * @return void
     */
```

src/Adapter/AbstractObjectModelFiller.php
```diff
@@ -0,0 +1,108 @@
...
+            $objectModel->addFieldsToUpdate([$propertyName => true]);
+        }
+    }
+
+    /**
+     * @param ObjectModel $objectModel
+     * @param string $propertyName
+     * @param array $properties
+     */
+    protected function fillLocalizedProperty(ObjectModel $objectModel, string $propertyName, array $properties)
```
Progi1984:
```suggestion
    protected function fillLocalizedProperty(ObjectModel $objectModel, string $propertyName, array $properties): void
```

src/Adapter/AbstractObjectModelFiller.php
```diff
@@ -0,0 +1,108 @@
...
+        if ($objectModel->id) {
+            $objectModel->addFieldsToUpdate([$propertyName => true]);
+        }
+    }
+
+    /**
+     * @param ObjectModel $objectModel
+     * @param string $propertyName
+     * @param array $properties
+     */
```
Progi1984:
```suggestion
     * @param array $properties
     *
     * @return void
     */
```

src/Adapter/AbstractObjectModelFiller.php
```diff
@@ -0,0 +1,108 @@
...
+        if ($objectModel->id) {
+            $this->addLocalizedPropertyToUpdate($objectModel, $propertyName, $properties[$propertyName]);
+        }
+    }
+
+    /**
+     * @param ObjectModel $objectModel
+     * @param string $propertyName
+     * @param array<int, string> $values
+     */
```
Progi1984:
```suggestion
     * @param array<int, string> $values
     *
     * @return void
     */
```

src/Adapter/Product/CommandHandler/UpdateProductStockHandler.php
```diff
@@ -0,0 +1,134 @@
...
+use PrestaShop\PrestaShop\Adapter\Product\Converter\OutOfStockTypeConverter;
+use PrestaShop\PrestaShop\Adapter\Product\Converter\PackStockTypeConverter;
+use PrestaShop\PrestaShop\Adapter\Product\Repository\ProductRepository;
+use PrestaShop\PrestaShop\Adapter\Product\Repository\StockAvailableRepository;
+use PrestaShop\PrestaShop\Adapter\Product\Update\ProductStockUpdater;
+use PrestaShop\PrestaShop\Core\Domain\Product\Command\UpdateProductStockCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\UpdateProductStockHandlerInterface;
+
+/**
+ * @internal
```
Progi1984:
```suggestion
```

src/Adapter/Product/Converter/OutOfStockTypeConverter.php
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Product\Converter;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Stock\ValueObject\OutOfStockType;
+use Product as LegacyProduct;
+
+class OutOfStockTypeConverter
```
Progi1984: Please add unit tests

src/Adapter/Product/Converter/PackStockTypeConverter.php
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Product\Converter;
+
+use Pack as LegacyPack;
+use PrestaShop\PrestaShop\Core\Domain\Product\Pack\ValueObject\PackStockType;
+
+class PackStockTypeConverter
```
Progi1984: Please add unit tests

src/Adapter/Product/QueryHandler/GetProductForEditingHandler.php
```diff
@@ -58,13 +65,21 @@ final class GetProductForEditingHandler extends AbstractProductHandler implement
...
+    private $stockAvailableRepository;
+
     /**
      * @param NumberExtractor $numberExtractor
+     * @param StockAvailableRepository $stockAvailableRepository
      */
     public function __construct(
-        NumberExtractor $numberExtractor
+        NumberExtractor $numberExtractor,
+        StockAvailableRepository $stockAvailableRepository
```
Progi1984, line 79: BC Break

src/Adapter/Product/Update/ProductStockUpdater.php
```diff
@@ -0,0 +1,261 @@
...
+     * @param StockAvailable $stockAvailable
+     * @param array $propertiesToUpdate
+     * @param bool $addMovement
+     *
+     * @throws CoreException
+     * @throws ProductStockConstraintException
+     */
+    public function update(Product $product, StockAvailable $stockAvailable, array $propertiesToUpdate, bool $addMovement = true): void
+    {
+        $advancedStockEnabled = (bool) $this->configuration->get('PS_ADVANCED_STOCK_MANAGEMENT');
```
Progi1984, line 94: Pass this value as argument ;)

src/Adapter/Product/Update/ProductStockUpdater.php
```diff
@@ -0,0 +1,261 @@
...
+     * @param StockAvailable $stockAvailable
+     * @param array $propertiesToUpdate
+     * @param bool $addMovement
+     *
+     * @throws CoreException
+     * @throws ProductStockConstraintException
+     */
+    private function updateAdvancedStock(Product $product, StockAvailable $stockAvailable, array $propertiesToUpdate, bool $addMovement): void
+    {
+        $productHasAdvancedStock = isset($propertiesToUpdate['advanced_stock_management']) ? $propertiesToUpdate['advanced_stock_management'] : $product->advanced_stock_management;
```
Progi1984:
```suggestion
        $productHasAdvancedStock = $propertiesToUpdate['advanced_stock_management'] ?: $product->advanced_stock_management;
```

src/Adapter/Product/QueryHandler/GetProductForEditingHandler.php
zuk3975: ProductForEditingHandler not released yet, no bc break :stuck_out_tongue:

Progi1984: Correct ;)

src/Adapter/AbstractObjectModelFiller.php
jolelievre: I don't see the interest of PHPDoc `@return void` Because to me `void` is not even a type, it returns nothing So strictly typing a method as void makes sense But the `@return` in this case has no interest and is even more confusing
Of course it's just my point of view ^^

src/Adapter/Product/Provider/StockAvailableProvider.php
jolelievre: You're right I split both methods and they are used by the ProductStockUpdater

classes/Pack.php
```diff
@@ -463,7 +463,7 @@ public static function deleteItems($id_product)
      */
     public static function addItem($id_product, $id_item, $qty, $id_attribute_item = 0)
     {
-        $id_attribute_item = (int) $id_attribute_item ? (int) $id_attribute_item : Product::getDefaultAttribute((int) $id_item);
+        $id_attribute_item = (int) $id_attribute_item ? (int) $id_attribute_item : Product::getDefaultAttribute((int) $id_item, 0, true);
```
jolelievre: Refresh the pack default attribute, or in behat tests Some previous combination values were cached and reused in later tests Which made them fail

src/Adapter/AbstractObjectModelFiller.php
zuk3975: >  strictly typing a method as void makes sense But the @return in this case has no interest

I agree

src/PrestaShopBundle/Resources/config/services/core/stock.yml
```diff
@@ -0,0 +1,6 @@
+services:
+    _defaults:
+        public: true
+
+    prestashop.core.stock.stock_manager:
+        class: 'PrestaShop\PrestaShop\Core\Stock\StockManager'
```
jolelievre, line 6: I integrated this one so that it can be easily injected in handers/services
However it as built a long time ago and relies on `PrestaShop\PrestaShop\Adapter\ServiceLocator` and `PrestaShop\PrestaShop\Adapter\SymfonyContainer`

I think it would be worth having a clean refacto so that it is fully built by Symfony framework and DI, however it will probably be a BC since it might be used in core or in modules by just doing `new StockManager ()` At the very least it is used in `StockAvailable` like this `$stockManager = ServiceLocator::get('\\PrestaShop\\PrestaShop\\Core\\Stock\\StockManager');`

So we maybe need to create a duplicate class with proper constructor and keep this one as deprecated

tests/Integration/Behaviour/Features/Context/Domain/AbstractDomainFeatureContext.php
```diff
@@ -64,9 +64,10 @@ public static function prepare(BeforeSuiteScope $scope)
     public function checkLastException(AfterScenarioScope $scope)
     {
         $e = $this->getLastException();
+        $this->cleanLastException();
 
         if (TestResult::FAILED === $scope->getTestResult()->getResultCode() && null !== $e) {
-            throw new RuntimeException(sprintf('Might be related to the last exception: %s %s', get_class($e), $e->getTraceAsString()));
+            throw new RuntimeException(sprintf('Might be related to the last exception: %s: %s Use -vvv for additional stack trace info', get_class($e), $e->getMessage()), 0, $e);
```
jolelievre, line 70: @zuk3975 we talked about this one, I improved it a bit The duplicate messages were because the exception was never cleaned, so it repeated this display for each following scenarios

I also reduced the message so that it focuses on last error, the stack trace can already be displayed use verbosity parameter so I'm not sure we should display it all the time

src/Adapter/Product/Converter/OutOfStockTypeConverter.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * @param int $legacyPackStockType
+     *
+     * @return string
+     */
+    public static function convertToValueObject(int $legacyPackStockType): string
+    {
+        switch ($legacyPackStockType) {
```
zuk3975: I still don't like these. I think the VO could simply contain integers defined as constants

src/Core/Domain/Product/QueryResult/ProductStock.php
```diff
@@ -0,0 +1,218 @@
...
+        string $packStockType,
+        string $outOfStockType,
+        int $quantity,
+        int $minimalQuantity,
+        string $location,
+        int $lowStockThreshold,
+        bool $lowStockAlert,
+        array $localizedAvailableNowLabels,
+        array $localizedAvailableLaterLabels,
+        DateTime $availableDate
```
zuk3975: shouldn't we use DateTimeInterface instead of DateTime everywhere?

src/Core/Domain/Product/Stock/Exception/CannotAddStockAvailableException.php
```diff
@@ -0,0 +1,33 @@
...
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\Stock\Exception;
+
+class CannotAddStockAvailableException extends ProductStockException
```
zuk3975, line 34: i know php doc would be redundant here, but there are some guys that always asks them :smile: 
There are alot of classes without docblock above class name in this PR. I wonder if we really need to add them everywhere?

src/Adapter/Product/Update/ProductStockUpdater.php
```diff
@@ -0,0 +1,185 @@
...
+     *
+     * @return bool
+     */
+    private function updateQuantity(Product $product, StockAvailable $stockAvailable, array $propertiesToUpdate, bool $addMovement): bool
+    {
+        if (!in_array('quantity', $propertiesToUpdate)) {
+            return false;
+        }
+
+        $deltaQuantity = (int) (int) $product->quantity - $stockAvailable->quantity;
```
Progi1984:
```suggestion
        $deltaQuantity = (int) $product->quantity - $stockAvailable->quantity;
```

Progi1984 (commented):
Some comments always in to do ;)

src/Adapter/Product/Validate/ProductValidator.php
```diff
@@ -133,6 +149,101 @@ private function validateShipping(Product $product): void
...
+    /**
+     * @param Product $product
+     *
+     * @throws ProductConstraintException
+     * @throws ProductPackConstraintException
+     * @throws ProductStockConstraintException
+     */
+    private function validateStock(Product $product): void
+    {
+        $advancedStockEnabled = (bool) $this->configuration->get('PS_ADVANCED_STOCK_MANAGEMENT');
```
matks: Maybe we could extract this property in constructor to store only the property, not the Configuration service ?

Progi1984:
```suggestion
        $advancedStockEnabled = $this->configuration->getBoolean('PS_ADVANCED_STOCK_MANAGEMENT');
```

src/Core/Domain/Product/Command/UpdateProductStockCommand.php
```diff
@@ -0,0 +1,392 @@
...
+    {
+        return $this->location;
+    }
+
+    /**
+     * @param string $location
+     *
+     * @return UpdateProductStockCommand
+     */
+    public function setLocation(string $location): UpdateProductStockCommand
```
matks: Should we rename Command to UpdateProductStockSettingsCommand ?

When I see UpdateProductStockCommand, I think "oh, this is used to update the stock quantity !" 😄

src/Core/Domain/Product/QueryResult/ProductStock.php
```diff
@@ -0,0 +1,218 @@
...
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\QueryResult;
+
+use DateTime;
+
+class ProductStock
```
matks: ProductStockSettings ? ProductStockInformations ?

src/Core/Domain/Product/Stock/Exception/CannotAddStockAvailableException.php
matks: > but there are some guys that always asks them 
👀 

> I wonder if we really need to add them everywhere

Only if they bring relevant information (for example if name is ambiguous, it can help clearing the confusion). The phpDoc must add documentation value, else it can go away.

src/PrestaShopBundle/Resources/config/services/core/stock.yml
matks: >So we maybe need to create a duplicate class with proper constructor and keep this one as deprecated

I'm in favor of that 👍

tests/Integration/Behaviour/Features/Context/Domain/Product/AbstractProductFeatureContext.php
```diff
@@ -116,6 +117,48 @@ protected function assertStringProperty(ProductForEditing $productForEditing, ar
         }
     }
 
+    /**
+     * @param ProductForEditing $productForEditing
+     * @param array $data
+     * @param string $propertyName
+     */
+    protected function assertNumberProperty(ProductForEditing $productForEditing, array &$data, string $propertyName): void
```
matks, line 147: We might need to move all these `assert...Property()` helpers into a dedicated Util/Helper class one day

tests/Integration/Behaviour/Features/Context/Domain/Product/AbstractProductFeatureContext.php
```diff
@@ -116,6 +117,48 @@ protected function assertStringProperty(ProductForEditing $productForEditing, ar
...
+            $actualValue = $this->extractValueFromProductForEditing($productForEditing, $propertyName);
+            $formattedExpectedDate = $expectedValue->format('Y-m-d');
+            $formattedActualDate = $actualValue->format('Y-m-d');
+            Assert::assertEquals(
+                $formattedExpectedDate,
+                $formattedActualDate,
+                sprintf('Expected %s "%s". Got "%s".', $propertyName, $formattedExpectedDate, $formattedActualDate)
+            );
+
+            unset($data[$propertyName]);
```
matks, line 211: I understand this is done because of
```
        // Assertions checking isset() can hide some errors if it doesn't find array key,
        // to make sure all provided fields were checked we need to unset every asserted field
        // and finally, if provided data is not empty, it means there are some unnasserted values left
```

The 1st time I checked this function I was like "wtf ? 🤔 " when I saw the unset. Now I understand.

Could we make it easier for the reader to understand this ? My suggestions:
- store this statement/explanation somewhere we can lead people to (phpDoc of this class ? devdocs ?) and refer to it
- add into `assertDateProperty()` phpDoc the explanation about the isset (can be a simple link)
- modify `assertDateProperty()` into something more meaningful `assertDatePropertyAndUnsetData()` but it's very verbose, I dont like it

src/Core/Domain/Product/QueryResult/ProductStock.php
jolelievre: I guess so

src/Core/Domain/Product/Command/UpdateProductStockCommand.php
jolelievre: It does update the quantity Among other things

tests/Integration/Behaviour/Features/Context/Domain/Product/AbstractProductFeatureContext.php
jolelievre: yep but in another PR as these methods, for now, are tightly related to `ProductForEditing `

tests/Integration/Behaviour/Features/Context/Domain/Product/AbstractProductFeatureContext.php
jolelievre: All the assert methods from this class use the same principle of unsetting checked value I was surprised first time I saw that myself so I understand your trouble

Though since they all work the same I wouldn't go for the `assertDatePropertyAndUnsetData` which can become unreadable if they are all renamed like this

I'd rather go for a simple comment on top of each `unset`

src/Adapter/Product/Validate/ProductValidator.php
jolelievre: Ok I did it (along with the default pack stock type) However this forces us to reboot the kernel for each feature because the configuration is no longer dynamic on runtime It is injected once when the the service is created and can not change later

For most use cases it's acceptable, but it also means that switching this parameter in the BO requires to clean the cache We must keep this in mind for other cases with configuration that are more likely to be changed by the merchant

jolelievre: Actually it doesn't require to clean the cache, my bad Since the DI gets the value from database at each new process it's ok But we still should think a bit about this when we use configuration in Dependency Injection

src/Core/Domain/Product/Command/UpdateProductStockCommand.php
jolelievre: I understand why you suggest a more detailed name since it's more than updating one value However all other commands reflect this idea:
- `UpdateProductSeoCommand` (not `UpdateProductSeoSettingsCommand`)
- `UpdateProductShippingCommand` (not `UpdateProductShippingInformationsCommand`)

And so on, they each relate to a subdomain so once you get the big picture it actually makes sense and it allows to keep short understandable names.

I almost decided to follow your lead because I though maybe we will need a command to change the stock quantity only and the name would already be taken! But then I figured then we would call such command `UpdateProductQuantityCommand` because Stock is actually a wider domain than just quantity So I think the naming suits the need in this case 🤷‍♂️

src/Core/Domain/Product/QueryResult/ProductStock.php
jolelievre: Ok for this one I admit I'm a bit more lost Because the other query results don't have a clear convention:
- `ProductSeoOptions`
- `ProductPricesInformation`
- `ProductSupplierOptions` / `ProductSupplierOption`
- `ProductCategoriesInformation`

It's kind of a mess and would deserve to clarify the naming, either use Options or Informations or nothing But we should do the same everywhere So I suggest we clarify and modify this in a later PR 😅

src/Core/Domain/Product/Stock/Exception/CannotAddStockAvailableException.php
jolelievre: I added every missing PHPDoc ^^ Even if the name seem clear an additional detail can't hurt
And mostly I bother you so much @zuk3975 with those PHPDocs than I can't look the other way when I'm reminded this 😅

src/PrestaShopBundle/Resources/config/services/core/stock.yml
jolelievre: In **much** later PR then ^^

src/Core/Domain/Product/QueryResult/ProductStock.php
zuk3975: I have already issued this some time ago, we decided just to call as it fits. 
as (if Im not wrong) @matks argumented - Options fits more to something we can adjust. Information is more suitable for (well I don't remember :trollface: ) anyway I agree it is still too much space left for interpretation. 
But it doesnt sound right if e.g. you call ProductCategoriesOptions does it? Information sounds better this way.
Although I would vote that we find one suffix +-fitting all cases

classes/ObjectModel.php
```diff
@@ -2155,7 +2155,7 @@ public function setFieldsToUpdate(?array $fields)
     }
 
     /**
-     * @return array|null
+     * @return array<string, bool|array<int, bool>>|null
      */
     public function getFieldsToUpdate(): ?array
```
zuk3975, line 2160: Is it still used?
You might have added it in early stages (when I added it in another PR and you might have duplicated this to avoid blocking the PR). but I think we don't use it anymore

classes/Product.php
```diff
@@ -273,7 +274,7 @@ class ProductCore extends ObjectModel
      *          - 1 Allow orders
      *          - 2 Use global setting
      */
-    public $out_of_stock;
+    public $out_of_stock = OutOfStockType::OUT_OF_STOCK_DEFAULT;
```
zuk3975, line 278: doesn't it count as a BC break?

src/Adapter/Product/CommandHandler/UpdateProductStockHandler.php
```diff
@@ -0,0 +1,148 @@
...
+        if (null !== $command->getPackStockType()) {
+            $product->pack_stock_type = $command->getPackStockType()->getValue();
+            $updatableProperties[] = 'pack_stock_type';
+        }
+        if (null !== $command->getQuantity()) {
+            $product->quantity = $command->getQuantity();
+            $updatableProperties[] = 'quantity';
+        }
+        if (null !== $command->getAvailableDate()) {
+            $product->available_date = $command->getAvailableDate()->format('Y-m-d');
```
zuk3975: in one of PR's (https://github.com/PrestaShop/PrestaShop/pull/21804/files#diff-9d738966e547d8db477408207db40533ec9cf8b178d4803c2952e91b318616dfR39)
I've added Util/DateTime::DEFAULT_FORMAT.
If it gets merged we could use it here

src/Adapter/Product/CommandHandler/UpdateProductStockHandler.php
```diff
@@ -0,0 +1,148 @@
...
+use PrestaShop\PrestaShop\Adapter\Product\Repository\ProductRepository;
+use PrestaShop\PrestaShop\Adapter\Product\Update\ProductStockUpdater;
+use PrestaShop\PrestaShop\Core\Domain\Product\Command\UpdateProductStockCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\UpdateProductStockHandlerInterface;
+use Product;
+
+/**
+ * Handles @see UpdateProductStockCommand using legacy object model
+ */
+final class UpdateProductStockHandler extends AbstractProductHandler implements UpdateProductStockHandlerInterface
```
zuk3975: `AbstractProductHandler` could be thrown away I believe

src/Adapter/Product/CommandHandler/UpdateProductStockHandler.php
```diff
@@ -0,0 +1,148 @@
...
+
+        $propertiesToUpdate = $this->fillUpdatableProperties($product, $command);
+        $this->productStockUpdater->update($product, $propertiesToUpdate, $command->addMovement());
+    }
+
+    /**
+     * @param Product $product
+     * @param UpdateProductStockCommand $command
+     *
+     * @return array
```
zuk3975: could be improved by using psalm annotation

src/Adapter/Product/Repository/ProductPackRepository.php
```diff
@@ -30,7 +30,7 @@
 
 use Pack;
 use PrestaShop\PrestaShop\Core\Domain\Product\Combination\ValueObject\CombinationId;
-use PrestaShop\PrestaShop\Core\Domain\Product\Exception\ProductPackException;
+use PrestaShop\PrestaShop\Core\Domain\Product\Pack\Exception\ProductPackException;
```
zuk3975, line 33: As you introduced new dir for Pack, we should move in commands here too.
Could you add a reminder somewhere? (or move them in this pr)

src/Adapter/Product/Repository/ProductRepository.php
```diff
@@ -168,6 +171,7 @@ public function get(ProductId $productId): Product
             Product::class,
             ProductNotFoundException::class
         );
+        $product->loadStockData();
```
zuk3975: should we wrap this in try-catch? (replacing the PrestaShopException to CoreException)?
I know we have not covered this 100%, but Its either we do it everywhere or don't bother at all :smile: (I would go for not bothering, but I've been told to do it since first days of my work in migration :sweat:)

src/Adapter/Product/Repository/StockAvailableRepository.php
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @throws CoreException
+     * @throws StockAvailableNotFoundException
+     */
+    public function create(ProductId $productId): StockAvailable
+    {
+        $stockAvailable = new StockAvailable();
+        $stockAvailable->id_product = $productId->getValue();
+        $shopParams = [];
+        StockAvailable::addSqlShopParams($shopParams);
```
zuk3975: same here, should we wrap in try-catch for PrestaShop exception?

src/Adapter/Product/Update/ProductStockUpdater.php
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * @param Product $product
+     * @param StockAvailable $stockAvailable
+     * @param array $propertiesToUpdate
+     * @param bool $addMovement
+     *
+     * @return bool
+     */
+    private function updateQuantity(Product $product, StockAvailable $stockAvailable, array $propertiesToUpdate, bool $addMovement): bool
```
zuk3975: method name suits more for some `write` action rather than `read`.
maybe something like this? :
```suggestion
    private function isQtyUpdateRequired(Product $product, StockAvailable $stockAvailable, array $propertiesToUpdate, bool $addMovement): bool
```

src/Adapter/Product/Validate/ProductValidator.php
```diff
@@ -133,6 +160,100 @@ private function validateShipping(Product $product): void
...
+     */
+    private function validateStock(Product $product): void
+    {
+        if ($this->advancedStockEnabled) {
+            $this->validateAdvancedStock($product);
+        } else {
+            $this->validateClassicStock($product);
+        }
+
+        $this->validateProductProperty($product, 'low_stock_threshold');
```
zuk3975: Missing error codes.
according to https://github.com/PrestaShop/PrestaShop/pull/21110#discussion_r519180880
I think we came to conclusion that every prop should be validated and have dedicated error code.

src/Core/Domain/Product/QueryResult/ProductStock.php
```diff
@@ -0,0 +1,221 @@
...
+     * @var string[] key value pairs where key is the id of language
+     */
+    private $localizedAvailableLaterLabels;
+
+    /**
+     * @var DateTimeInterface
+     */
+    private $availableDate;
+
+    public function __construct(
```
zuk3975: missing docblock

src/PrestaShopBundle/Api/QueryParamsCollection.php
```diff
@@ -114,6 +114,26 @@ public function fromRequest(Request $request)
         return $this;
     }
 
+    /**
+     * @param array $queryParams
+     * @param array $allParams
+     *
+     * @return $this
+     */
+    public function fromArray(array $queryParams, array $allParams = []): QueryParamsCollection
```
zuk3975, line 123: I have no idea what is this (all this API thing), but I'm not gonna dig into it, I believe you :smile: 
I just hope I didn't need to migrate some API related things for other product parts? :thinking:  (cuz I didn't)

tests/Integration/Behaviour/Features/Context/Domain/Product/AbstractProductFeatureContext.php
```diff
@@ -84,13 +86,17 @@ protected function assertBoolProperty(ProductForEditing $productForEditing, arra
     {
         if (isset($data[$propertyName])) {
             $expectedValue = PrimitiveUtils::castStringBooleanIntoBoolean($data[$propertyName]);
+            // Don't cast on purpose, the value should already be typed as bool
             $actualValue = $this->extractValueFromProductForEditing($productForEditing, $propertyName);
-            Assert::assertEquals(
+
+            // Use assertSame (not assertEquals) to check value AND type
+            Assert::assertSame(
```
zuk3975, line 107: :+1:

tests/Integration/Behaviour/Features/Context/Domain/Product/AbstractProductFeatureContext.php
```diff
@@ -84,13 +86,17 @@ protected function assertBoolProperty(ProductForEditing $productForEditing, arra
...
                 $expectedValue,
                 $actualValue,
                 sprintf('Expected %s "%s". Got "%s".', $propertyName, $expectedValue, $actualValue)
             );
 
+            // Unset the checked field from array so we can validate they havel all been asserted
             unset($data[$propertyName]);
```
zuk3975, line 114: I wonder if we could somehow implement some general assert at the end like `assertAllPropertiesHaveBeenAsserted(array $data)`.
Anyway its for some time later.

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateStockFeatureContext.php
```diff
@@ -0,0 +1,325 @@
...
+        }
+    }
+
+    /**
+     * @Then product :productReference should have following stock information:
+     *
+     * @param string $productReference
+     * @param TableNode $table
+     */
+    public function assertStockInformation(string $productReference, TableNode $table)
```
zuk3975:
```suggestion
    public function assertStockInformation(string $productReference, TableNode $table): void
```

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateStockFeatureContext.php
```diff
@@ -0,0 +1,325 @@
...
+        // Assertions checking isset() can hide some errors if it doesn't find array key,
+        // to make sure all provided fields were checked we need to unset every asserted field
+        // and finally, if provided data is not empty, it means there are some unnasserted values left
+        Assert::assertEmpty($data, sprintf('Some provided product stock fields haven\'t been asserted: %s', var_export($data, true)));
+    }
+
+    /**
+     * @Then product :productReference last stock movement has following details:
+     */
+    public function assertProductLastStockMovement(string $productReference, TableNode $table)
```
zuk3975: again no big deal, but IMO void lets us now that type is not forgotten and its not intended to return anything on purpose.

```suggestion
    public function assertProductLastStockMovement(string $productReference, TableNode $table): void
```

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateStockFeatureContext.php
```diff
@@ -0,0 +1,325 @@
...
+                    $lastMovement[$movementField]
+                )
+            );
+        }
+    }
+
+    /**
+     * @Then product :productReference has no stock movements
+     */
+    public function assertProductHasNoStockMovement(string $productReference)
```
zuk3975:
```suggestion
    public function assertProductHasNoStockMovement(string $productReference): void
```

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateStockFeatureContext.php
```diff
@@ -0,0 +1,325 @@
...
+     * @return int
+     */
+    private function convertPackStockTypeToInt(string $outOfStock): int
+    {
+        $intValues = [
+            'default' => PackStockType::STOCK_TYPE_DEFAULT,
+            'products_only' => PackStockType::STOCK_TYPE_PRODUCTS_ONLY,
+            'pack_only' => PackStockType::STOCK_TYPE_PACK_ONLY,
+            'both' => PackStockType::STOCK_TYPE_BOTH,
+            'invalid' => 42,
```
zuk3975: little comment explaining why its 42 and not 420 :smoking: :trollface: ?
I imagine smth like:
`This random number is hardcoded intentionally to reflect invalid stock type`

tests/Integration/Behaviour/Features/Scenario/Product/update_stock_advanced.feature
```diff
@@ -0,0 +1,292 @@
...
+      | is_virtual | false               |
+    And product "product1" should have following stock information:
+      | use_advanced_stock_management | false       |
+    When I update product "product1" stock with following information:
+      | use_advanced_stock_management | true        |
+    Then product "product1" should have following stock information:
+      | use_advanced_stock_management | true        |
+
+  Scenario: I update product depends on stock (also check automatic update when disabling advanced stock on product)
+    Given I add product "product1" with following information:
```
zuk3975, line 61: it could be product2  or if its intentionally replaced before each scenario, then it could go to background, no?

tests/Integration/Behaviour/Features/Scenario/Product/update_stock_advanced.feature
```diff
@@ -0,0 +1,292 @@
...
+      | is_virtual | false               |
+    And product "product1" should have following stock information:
+      | quantity | 0 |
+    When I update product "product1" stock with following information:
+      | quantity | 51 |
+    Then product "product1" should have following stock information:
+      | quantity | 51 |
+    And product "product1" last stock movement has following details:
+      | physical_quantity | 51 |
+      | sign              | 1  |
```
zuk3975: what `sign` stands for? (I know its data column of stock movement), but what does that word mean in stock context?

tests/Integration/Behaviour/Features/Scenario/Product/update_stock_advanced.feature
```diff
@@ -0,0 +1,292 @@
...
+      | low_stock_alert        | false      |
+      | available_date         | 0000-00-00 |
+    And product "product1" localized "available_now_labels" should be "en-US:"
+    And product "product1" localized "available_later_labels" should be "en-US:"
+    When I update product "product1" stock with following information:
+      | minimal_quantity       | 12                 |
+      | location               | dtc                |
+      | low_stock_threshold    | 42                 |
+      | low_stock_alert        | true               |
+      | available_now_labels   | en-US:get it now   |
```
zuk3975, line 283: you might want to check multiple langs at some point :smile: 
(at least a scenario to check if required localized fields are filled with default lang value when you aren't providing them)

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateStockFeatureContext.php
```diff
@@ -0,0 +1,325 @@
...
+        $this->assertDateProperty($productForEditing, $data, 'available_date');
+
+        // Assertions checking isset() can hide some errors if it doesn't find array key,
+        // to make sure all provided fields were checked we need to unset every asserted field
+        // and finally, if provided data is not empty, it means there are some unnasserted values left
+        Assert::assertEmpty($data, sprintf('Some provided product stock fields haven\'t been asserted: %s', var_export($data, true)));
+    }
+
+    /**
+     * @Then product :productReference last stock movement has following details:
```
zuk3975: dockblock missing. (I suppose we should add it in tests also?)
`@param string $productReference`
`@param TableNode $table`

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateStockFeatureContext.php
```diff
@@ -0,0 +1,325 @@
...
+
+    /**
+     * @Then product :productReference has no stock movements
+     */
+    public function assertProductHasNoStockMovement(string $productReference)
+    {
+        $productId = $this->getSharedStorage()->get($productReference);
+
+        /** @var StockMovementRepository $stockMovementRepository */
+        $stockMovementRepository = $this->getContainer()->get('prestashop.core.api.stock_movement.repository');
```
zuk3975, line 163: I think you can initialize this in constructor

classes/ObjectModel.php
jolelievre: I think you meant the former `addFieldsToUpdate` which has been removed?

classes/Product.php
jolelievre: No because this is the default value that is set in the `add` function anyway So I just emphasized its default value in the class And I added additional behat tests to ensure these default values at creation

classes/ObjectModel.php
zuk3975: oh you might be right, sorry

src/Adapter/Product/CommandHandler/UpdateProductStockHandler.php
jolelievre: Oh indeed nice, but it's not merged yet so I copied it into my PR

jolelievre: I actually added two consts because we need one for DateTime and one for Date only

jolelievre: I only added DEFAULT_DATE_FORMAT in the end to avoid conflicts and unused const, but I think you should rename yours DEFAULT_DATETIME_FORMAT in your PR to be consistent

src/Adapter/Product/Repository/ProductRepository.php
jolelievre: Yeah you're right, nice catch!

src/Adapter/Product/CommandHandler/UpdateProductStockHandler.php
zuk3975: done :heavy_check_mark:

src/Adapter/Product/Repository/ProductPackRepository.php
zuk3975: maybe its even worth to move those pack changes to separate PR, because this one is already HEAVY (btw im in need of those Pack changes in another PR already :smile: )

src/Adapter/Product/Repository/StockAvailableRepository.php
jolelievre: It's just supposed to add sql params, but since it needs to fetch ObjectModel behind the scene you're right it's safer!

src/Adapter/Product/Update/ProductStockUpdater.php
jolelievre: No because the method actually performs some updates (on the StockAvailable object, and in the StockMovement)
The result returned indicates if a field has been modified and if the object needs to be updated

jolelievre: So it prevents from updating StockAvailable if no updated fields are related it (as the updater also handles configuration fields which are only related to Product and not StockAvailable) I'll add a bit more documentation though

jolelievre: Ok I ended up modifying this function so that it doesn't return anything any more, the need for update is all handled in `updateStockAvailable`

src/Adapter/Product/Validate/ProductValidator.php
jolelievre: Yeah we decided this after I coded this, I'm gonna add them

src/Core/Domain/Product/QueryResult/ProductStock.php
jolelievre: Ok you convinced me, I went with `ProductStockInformation` I'm not sure the information suffix is the best choice for all other use cases but in this case it felt better as it doesn't contain only configurations or options but actual dynamic values (like quantity)

src/PrestaShopBundle/Api/QueryParamsCollection.php
jolelievre: No it's only to get StockMovement so it's only related to the Product stock The existent implementation was very coupled with query so it didn't allow to perform the same query in test environment with simple array input

tests/Integration/Behaviour/Features/Context/Domain/Product/AbstractProductFeatureContext.php
jolelievre: Since it's used in many different contexts it would probably be useful yes!

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateStockFeatureContext.php
jolelievre: I prefer to call it "on demand" as this context might be used in other features but without using this assessment so no need to get the service then
And getting a service multiple times has no performance drawbacks

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateStockFeatureContext.php
jolelievre: Alright :trollface:
I see you are getting your revenge here 😛

tests/Integration/Behaviour/Features/Scenario/Product/update_stock_advanced.feature
jolelievre: It could be product2 but then each new scenario will need to always rename the product which is a bit of pain 😅  (and it can lead to mistakes if you forget to replace one) Always using the same reference means it will override the reference to the ID anyway

And I do want to create a new product at each scenario but it's not always "product1" in the pack related scenarios I use "productPack1" for instance So if I always add "product1" in the background in some cases it wouldn't be used

tests/Integration/Behaviour/Features/Scenario/Product/update_stock_advanced.feature
jolelievre: It's used in StockMovement to indicate if the stock was increased (`1`) or decreased (`-1`) It would be more accurate to use `+` or `-` but I went for the simplicity here and just checked the actual value in DB
Do you think I should add some transformations in the Behat context so that we use `+-` in the feature file?

tests/Integration/Behaviour/Features/Scenario/Product/update_stock_advanced.feature
jolelievre: Yep nice catch!

jolelievre: I added checked with french language it works correctly when it is provided However there is no fallback system on the default language I'm not sure we have implemented this earlier because I believe all other tests always provide the two values And I'm not even sure it has ever been handled in our commands 🤔  Isn't it handled by the form constraint usually? or the front js which automatically populates missing fields?

Do you have any example of fallback?

classes/ObjectModel.php
jolelievre: No it's ok, I just think this is an old comment that stayed even though the code has been removed (it was indeed present in previous versions of the PR)

src/Adapter/Product/Repository/ProductPackRepository.php
jolelievre: I added the todo for the commands to avoid making this one even bigger However since the exceptions are already moved and behat test already rely on them I prefer to keep them as is in this PR

tests/Integration/Behaviour/Features/Scenario/Product/update_stock_advanced.feature
zuk3975: yeah sign is very confusing, I would go for `+-`, its much clearer.

src/Adapter/Product/QueryHandler/GetProductForEditingHandler.php
```diff
@@ -269,4 +285,51 @@ private function getSeoOptions(Product $product): ProductSeoOptions
...
+     *
+     * @param Product $product
+     *
+     * @return ProductStockInformation
+     */
+    private function getProductStockInformation(Product $product): ProductStockInformation
+    {
+        try {
+            // Theoretically StockAvailable is created for each product when Product::add is called
+            $stockAvailable = $this->stockAvailableRepository->get(new ProductId($product->id));
```
matks, line 300: What happens when stock management is disabled in shop ?

src/Adapter/Product/Repository/ProductRepository.php
```diff
@@ -225,6 +227,16 @@ public function get(ProductId $productId): Product
             ProductNotFoundException::class
         );
 
+        try {
+            $product->loadStockData();
```
matks, line 231: Same worry: what happens when stock management is disabled in shop ?

src/Adapter/Product/Update/ProductStockUpdater.php
```diff
@@ -0,0 +1,183 @@
...
+     */
+    public function update(Product $product, array $propertiesToUpdate, bool $addMovement = true): void
+    {
+        $stockAvailable = $this->getStockAvailable($product);
+        $this->productRepository->partialUpdate($product, $propertiesToUpdate, CannotUpdateProductException::FAILED_UPDATE_STOCK);
+
+        // It is very important to update StockAvailable after product, because the validation is performed in ProductRepository::partialUpdate
+        $this->updateStockAvailable($product, $stockAvailable, $propertiesToUpdate, $addMovement);
+
+        $advancedStockEnabled = (bool) $this->configuration->get('PS_ADVANCED_STOCK_MANAGEMENT');
```
matks: Should we pass to constructor the whole `$configuration` service or just this boolean?

src/Core/Domain/Product/Command/UpdateProductStockCommand.php
matks: > I almost decided to follow your lead

So close! 💪 

But OK for keeping current name 👍

src/Core/Util/DateTime/DateTime.php
```diff
@@ -33,10 +33,15 @@
  */
 final class DateTime
 {
+    /**
+     * Default format for date time string
+     */
+    public const DEFAULT_DATE_FORMAT = 'Y-m-d';
```
matks, line 39: Should'nt this depend on locale and CLDR ? 🤔

jolelievre: No because it's the format used to store in database

src/Adapter/Product/Update/ProductStockUpdater.php
jolelievre: Yep I'll try that!

src/Core/Domain/Product/Command/UpdateProductStockCommand.php
jolelievre: You'll be happy to know I'm gonna modify this class name so that it matches with the query result ^^

src/Adapter/Product/QueryHandler/GetProductForEditingHandler.php
jolelievre: So from what I see in the current code the `PS_STOCK_MANAGEMENT` is mainly used to:
- hide elements in the BO or FO (like product quantities labels or inputs)
- ignore out_of_stock behaviour

But it doesn't block quantity update, or stock movements so i don't see any reason why the CQRS command should handle this for now

src/Adapter/Product/Validate/ProductValidator.php
```diff
@@ -133,6 +160,100 @@ private function validateShipping(Product $product): void
...
+            $packStockType = $this->defaultPackStockType;
+        }
+
+        // Either the pack has its own stock, or else ALL products from the pack must depend on the stock as well
+        if ($packStockType === Pack::STOCK_TYPE_PACK_ONLY || Pack::allUsesAdvancedStockManagement($product->id)) {
+            return;
+        }
+
+        throw new ProductPackConstraintException(
+            'You cannot link your pack to product stock because one of them has no advanced stock enabled',
```
zuk3975, line 252: not a blocker, just a suggestion.
I think its better to format messages in different manner, like:
`Pack cannot be linked to product stock because one of them has no advanced stock enabled`.
The idea is to remove the "addressing" to user, because its not important, whether its "**you**" or "**me**" trying to link the pack :stuck_out_tongue:

src/Core/Util/DateTime/DateTime.php
```diff
@@ -33,10 +33,15 @@
...
+
     /**
      * DateTime value which should be considered same as null
      */
-    const NULL_VALUE = '0000-00-00 00:00:00';
+    public const NULL_VALUE = '0000-00-00 00:00:00';
```
zuk3975, line 44: :+1:

tests/Integration/Behaviour/Features/Scenario/Product/update_stock_advanced.feature
zuk3975: I think as we discussed its already resolved?
for the record - only required properties are filled with default value, all other are filled with empty string if not provided at all.

zuk3975 (dismissed):

tests/Integration/Behaviour/Features/Scenario/Product/update_stock_advanced.feature
```diff
@@ -240,14 +240,14 @@ Feature: Update product stock from Back Office (BO)
       | quantity | 51 |
     And product "product1" last stock movement has following details:
       | physical_quantity | 51 |
```
zuk3975: i thought it would become something like this:
```
      | physical_quantity | +51 |
      | physical_quantity | -51 |
```

jolelievre: ooohhh

tests/Integration/Behaviour/Features/Scenario/Product/update_stock_advanced.feature
```diff
@@ -238,16 +238,12 @@ Feature: Update product stock from Back Office (BO)
...
-      | sign              | +  |
+    And product "product1" last stock movement increased by 51
     When I update product "product1" stock with following information:
       | quantity | 42 |
     Then product "product1" should have following stock information:
       | quantity | 42 |
-    And product "product1" last stock movement has following details:
-      | physical_quantity | 9 |
-      | sign              | - |
+    And product "product1" last stock movement decreased by 9
```
zuk3975, line 246: :+1:

zuk3975 (dismissed):

Progi1984 (changes requested):
Some questions need answers

src/Core/Domain/Product/CommandHandler/UpdateProductStockInformationHandlerInterface.php
```diff
@@ -0,0 +1,42 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
```
Progi1984:
```suggestion
```

tests/Integration/Behaviour/Features/Context/Domain/Product/AbstractProductFeatureContext.php
Progi1984: Please add a todo :)

Progi1984 (approved):

matks (approved):

zuk3975 (approved):

matks:
Thank you @jolelievre and everybody that helped making this happen, especially @zuk3975

PR #21110, opened 2020-09-23
Use ProductRepository in AddProductHandler

### LINKED ISSUE #14763: [EPIC] Issue Product page - Basic settings tab 2/9

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | refactors AddProductHandler to use ProductRepository service inside. (Following changes are not BC break, all of refactored classes haven't been released yet and is part of Product migration)
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | update of #18672
| How to test?  | travis :heavy_check_mark: 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21110)


zuk3975:
rr travis

src/Core/Domain/Product/Customization/Command/AddCustomizationFieldCommand.php
```diff
@@ -33,6 +33,8 @@
 
 /**
  * Adds customization field to a product
+ *
+ * @todo: this command along with update/deleteCustimizationField can be removed, because we only have a usecase for Set & RemoveAll
```
jolelievre, line 37: Since https://github.com/PrestaShop/PrestaShop/pull/21260 has been merged I think you can rebase this one

src/Adapter/AbstractObjectModelValidator.php
```diff
@@ -78,8 +78,14 @@ protected function validateObjectModelProperty(ObjectModel $objectModel, string
      */
     protected function validateObjectModelLocalizedProperty(ObjectModel $objectModel, string $propertyName, string $exceptionClass, int $errorCode = 0)
     {
+        $localizedValues = $objectModel->{$propertyName};
+
+        if (null === $localizedValues) {
```
jolelievre: I don't understand this check.. Why would the field be null, at worst it should be an empty array no?
And why ignore when it's null isn't it an error worth indicating?

src/Adapter/Product/Repository/ProductRepository.php
```diff
@@ -83,6 +84,22 @@ public function get(ProductId $productId): Product
...
     }
 
+    /**
+     * @param Product $product
+     *
+     * @return ProductId
+     *
+     * @throws CannotAddProductException
+     */
+    public function add(Product $product): ProductId
```
jolelievre: I hesitate, I wonder if you should instead have a `create` method with only the needed parameters
Instead of prefilling the object in the handler, it would give a single easy endpoint to create Product (usable somewhere else)
I'm just worried if it's not already too much domain logic for a Repository, but we're only talking about one default value (`active`) the other three would be parameters: `$isVirtual` `$localizedNames` `$defaultCategoryId`

src/Adapter/AbstractObjectModelValidator.php
zuk3975: yeah so it happens because it collided with another command. 
Im not sure its safe to have this check.
Basic example: we are updating name, but description is null (because it will be updated with other command), so validator still checks every property and then tries to loop through product->description which is null.

jolelievre: Then maybe it's preferable not to ignore this, because I don't think it's an expected behaviour to have fields turned into `null` by other commands 🤔
Better have an error indicating something is wrong in that case

zuk3975: fields are not TURNED into null, they are simply empty, because we fill the object only partially, but validate it whole. This what happens:
1. Add product command passed tests before, because there was only manual validation implemented in handler (which checked only those fields that we fill)
2. I rebased the PR so now it uses ProductValidator, which checks whole object validity inluding multilang fields like description.
3. Now when I `AddProduct`, it only adds `name`, but validator also asserts the `description, meta_title etc.`, those fields are null as they haven't yet reached the `UpdateBasicInfo` phase to be inserted.

So Its rather implementation error than "something went wrong", because it is logical & it will happen all the time :(

src/Adapter/Product/Repository/ProductRepository.php
matks: According to the [Repository Design Pattern](https://medium.com/@pererikbergman/repository-design-pattern-e28c0f3e4a30) the constructed object must be given to the repository to be saved 🤔

src/Adapter/AbstractObjectModelValidator.php
matks: So I can see 3 paths ahead of us:

1. Current implementation: validate whole object even though not all fields have been hydrated. This means that we need, indeed, to ignore `null` fields.
2. Validate only hydrated fields by providing the exact list of what fields to validate.
3. Validate whole object after having applied modifications to it. This means a) fetch Product object b) modify fields provided by Command c) now we have a Product object with updated data and we can validate it whole. There is no `null` fields.

If I am not wrong,

Solution 1 can introduce issues because `null` is a very generic value and could be mistakenly set and this validator, however it will be catched eventually later when persisting the ObjectModel I think 🤔 ?

Solution 2 requires lot of boilerplate code to pass around list of updated fields like `['name', 'description', 'seo', ...]` => very unsatisfying solution

Solution 3 is interesting 🤔 but maybe had to implement as it means we load Product "just to validate" it ; it would actually change the whole workflow. Right now we validate first then we fetch+update the ObjectModel.

So if  `it will be catched eventually later when persisting the ObjectModel I think` is confirmed I think solution 1 seems OK

src/Adapter/Product/Repository/ProductRepository.php
matks: I've seen other Repositories where there is such a thing (like `ProductTagRepository::add()` I believe.

I think in "pure" DDD the Entity constructor contains the creation logic. As we dont have an equivalent, it's fine to have it in the Repository.

matks (dismissed):

src/Adapter/AbstractObjectModelValidator.php
jolelievre: Ok I see the problem of object partially filled, but validation as a whole

Then I would suggest relying on `ObjectModel::$update_fields` We add a getter `getFieldsToUpdate()` and we check if fieldName is inside it if not return It's a more accurate check than relying on `null` value => we check only what we modified

src/Core/Domain/Product/Customization/Command/AddCustomizationFieldCommand.php
jolelievre: This one can be removed no?

jolelievre (dismissed):

matks (dismissed):

zuk3975:
The idea to check validation fields based on update_fields not gonna work (thats why behat tests are failing).
Reason: update_fields are set **after** validation. So localized fields validation will always be skipped on partial update :(

classes/ObjectModel.php
```diff
@@ -2154,6 +2154,14 @@ public function setFieldsToUpdate(?array $fields)
         $this->update_fields = $fields;
     }
 
+    /**
+     * @return array|null
+     */
+    public function getFieldsToUpdate(): ?array
```
jolelievre, line 2160: Not useful any more in this one but I'm sure we'll need it somewhere else 👍

src/Adapter/Product/AbstractProductHandler.php
```diff
@@ -84,7 +86,7 @@ protected function getProduct(ProductId $productId): Product
 
     /**
      * @todo: product name is not required. Fix that? issue: #19441 for discussion
-     *
+     * @todo: to be removed when all usecases are replaced by ObjectModelPersister & validator
```
jolelievre, line 89: I guess the whole class will be removed then 😅

src/Adapter/Product/Validate/ProductValidator.php
```diff
@@ -39,6 +39,22 @@
...
+    /**
+     * This method is specific for product creation only.
+     *
+     * @param Product $product
+     *
+     * @throws CoreException
+     */
+    public function validateCreation(Product $product): void
+    {
+        $this->validateProductLocalizedProperty(
```
jolelievre, line 51: Just one question, `is_virtual` doesn't need validation?

jolelievre (approved):

matks (approved):

src/Adapter/Product/Validate/ProductValidator.php
zuk3975: late answer (answered in slack). But for the record, it is a boolean & we make sure it is using php types, nothing more to validate here

zuk3975: @matks, @jolelievre Im coming back to this question :smile: 

At first I thought we don't need to validate those fields which in modern php is supposed to be handled by strict typing (bool, int validations).
**BUT**
validators accepts legacyObject model which doesn't have strict types, so in theory anyone using validator (without using our commands) will be able to pass invalid values
**SO**
maybe it is right to validate all the properties after all? :thinking: 
In this case, maybe it's not worth adding constants (error codes) in constraint exception for such properties ? 
Or keep it simple and add error code for each obj model property?

jolelievre: I guess you're right it would be safer to validate all the fields

matks: Always better to double check!

PR #21117, opened 2020-09-23
Use ProductCustomizationFieldUpdater service in CustomizationFieldHandlers

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | implement a service which handles relations of Product and Customization field update and reuse that service inside CustomizationField handlers
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | refacto of #20549
| How to test?  | Travis :heavy_check_mark: 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21117)


src/Adapter/Product/CommandHandler/RemoveAllCustomizationFieldsFromProductHandler.php
```diff
@@ -51,23 +52,23 @@ final class RemoveAllCustomizationFieldsFromProductHandler implements RemoveAllC
     private $productProvider;
 
     /**
-     * @var ProductUpdater
+     * @var ProductPersister
```
jolelievre:
```suggestion
     * @var ProductCustomizationFieldUpdater
```

And you can remove the useless `use` statement for `ProductPersister`

src/Adapter/Product/ProductPersister.php
```diff
@@ -54,43 +52,22 @@ public function __construct(
...
-                'customizable' => $customizable,
-                'text_fields' => $product->countCustomizationFields(CustomizationFieldType::TYPE_TEXT),
-                'uploadable_files' => $product->countCustomizationFields(CustomizationFieldType::TYPE_FILE),
-            ], CannotUpdateProductException::FAILED_UPDATE_CUSTOMIZATION_FIELDS
-        );
-    }
-
     /**
      * @param Product $product
      * @param array $propertiesToUpdate
```
jolelievre: While you're cleaning this one, can you create the sub function `fillCustomizationFields` please? which will be called by `fillProperties`

src/Adapter/Product/ProductCustomizationFieldUpdater.php
```diff
@@ -0,0 +1,154 @@
...
+        $this->customizationFieldDeleter = $customizationFieldDeleter;
+        $this->productPersister = $productPersister;
+        $this->productProvider = $productProvider;
+    }
+
+    /**
+     * @param ProductId $productId
+     * @param CustomizationField[] $customizationFields
+     */
+    public function setProductCustomizationFields(ProductId $productId, array $customizationFields): void
```
jolelievre, line 80: I think you could `Product $product` as the parameter here, since you already fetch the product in the handler, and it would remove on dependency with `ProductProvider` in this service

I guess using ValueObject IDs parameters are useful when dealing with "pure" relationship that we can manage with inserting directly in the table

jolelievre (commented):
Seems good this way, responsibilities are split without over engineering I'd say 🤔

However while we are cleaning I think keeping everything in the `PrestaShop\PrestaShop\Adapter\Product` will quickly become a mess, as we are adding Persister Updater and Deleter classes for each sub domain

I can see two solutions:
- create a namespace for each sub domain (CustomizationField, Attachment, ProductSupplier, ...)
- or create a namespace for each responsibility (Persister, Updater, Deleter)

What do you think?

jolelievre:
So after the discussion during standup seems like you can merge all Provider, Persister, Deleter classes into a single Repository class, therefore I think we can introduce the `Repository` sub namespace.

I don't know if we should have an inter-domain common one: `PrestaShop\PrestaShop\Adapter\Repository`
Or a domain specific one: `PrestaShop\PrestaShop\Adapter\Product\Repository`

I guess since some domains may use transversal ObjectModel, therefore transversal Repositories will be needed as well, and having them all in the same place will make them easier to find, and less likely to duplicate.

Now the domain logic doesn't move, it still needs to be handled in dedicated services (the so-called "Updater") although I still think we should search for a better naming (and namespace) - `Manager` is a no go 😅

src/Adapter/Product/ProductCustomizationFieldUpdater.php
jolelievre: So now that the Updater is in charge of update the product it makes sense that it has the `productRepository` since it needs it in `refreshProductCustomizability`

But now it's the opposite, the `SetProductCustomizationFieldsHandler` maybe doesn't need it any more, because it doesn't use the Product instance as an input nor anywhere else it only uses it to create the array it returns.

Should this array_map be moved into this Updater class? Or does it make more sense for the Handler to manage itself the format of its return?

jolelievre (approved):

matks (approved):

matks:
Thank you @zuk3975 and @jolelievre

PR #21125, opened 2020-09-23
Introduce LegacyContextLoader to load legacy context in Commands

### LINKED ISSUE #14173: Impossible to load Legacy classes in Console commands

matks:
| Questions     | Answers
| Branch?       | develop
| Description?  | See below
| Type?         | new feature
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/14173
| How to test?  | There is a test that checks a Command using legacy Product ObjectModel works. I will expand the test. When the coverage is good enough I believe there is no need to QA this PR, Travis should be green.

## Description

[When using legacy classes in Symfony Commands, multiple issues arise](https://github.com/PrestaShop/PrestaShop/issues/14173). This is because these classes need specific settings to be initialized inside `Context` class (currency, employee, shop ...).

We initially explored the possibility of loading all of these informations for every PrestaShop command, however @PierreRambaud suggested that we'd rather let the user load the context he needs to run the Command. No need to load the whole Earth for every Commands, else we'd be reimplementing the [Banana Monkey Jungle problem](https://softwareengineering.stackexchange.com/questions/368797/sample-code-to-explain-banana-monkey-jungle-problem-by-joe-armstrong).

So I introduce a `LegacyContextLoader` (name can be improved) that provides helper functions to load contexts, with the ability to target a specific a context or just a default ; and a `loadGenericContext()` to "load all of this" as a helper shortcut.

Naming can probably be improved as well as the content of the functions so it's not top-notch: I expect code review to bring in nice improvements and ideas. This is not a ready-to-be-merged piece of work.

I also provided an integration test powered by PHPUnit Integration Setup to verify the helper does its job. It was quite hard because I used [anonymous Command classes](https://www.php.net/manual/en/language.oop5.anonymous.php) so I had some issues with autoloading (this is why sometimes I do not use `use` but rather fully qualified class names) but ... it works! Please be kind with this part, it was hard to make it right 😅 

There are some more usecases to test 🤔 , I will add them but I am waiting for opinions on this part before going further 😊 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21125)


classes/controller/DummyController.php
```diff
@@ -0,0 +1,128 @@
...
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+use PrestaShop\PrestaShop\Adapter\SymfonyContainer;
+
+class DummyControllerCore extends Controller
```
matks: This dummy controller (naming can be improved) is necessary as some legacy classes expect `$context->controller` to be not null

src/Adapter/ContextLoadHelper.php
```diff
@@ -0,0 +1,140 @@
...
+use Context;
+use Currency;
+use DummyController;
+use Employee;
+use Shop;
+
+/**
+ * Helps loading specific context
+ */
+class ContextLoadHelper
```
matks: Naming can be improved

src/Adapter/ContextLoadHelper.php
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * @param LegacyContext $context
+     */
+    public function __construct(LegacyContext $context)
+    {
+        $this->context = $context->getContext();
+    }
+
+    public function loadGenericContext(
```
matks: I'm wondering if `loadGenericContext()` should accept optional parameters. Maybe users will use each function separately if they want to use accurate identifiers ?

tests/Integration/PrestaShopBundle/Command/LoadLegacyClassesinCommandTest.php
```diff
@@ -0,0 +1,102 @@
...
+
+class LoadLegacyClassesinCommandTest extends KernelTestCase
+{
+    private $previousErrorReportingLevel;
+
+    public function setUp()
+    {
+        parent::setUp();
+        self::bootKernel();
+        $this->previousErrorReportingLevel = error_reporting(2);
```
matks: This is needed for PHPUnit to convert errors to assertable exceptions.
See https://phpunit.readthedocs.io/en/7.4/writing-tests-for-phpunit.html?highlight=exception#testing-php-errors

tests/Integration/PrestaShopBundle/Command/LoadLegacyClassesinCommandTest.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public function testLoadLegacyCommandWithoutContextFails()
+    {
+        $this->expectException(\TypeError::class, 'Not enough arguments (missing: "theme, locale").');
+
+        $application = new Application(static::$kernel);
+        $application->add(new class() extends ContainerAwareCommand {
```
matks, line 56: I load an anonymous Command class to test it.

src/Adapter/ContextLoadHelper.php
```diff
@@ -0,0 +1,140 @@
...
+    {
+        if (null === $currencyId) {
+            $currency = new Currency();
+            $currency->precision = 2;
+            $this->context->currency = $currency;
+
+            return;
+        }
+
+        $this->context->currency = new Currency($currencyId);
```
PierreRambaud:
```suggestion
        $currency = new Currency($currencyId);
        if (null === $currencyId) {
            $currency->precision = 2;
        }

        $this->context->currency = $currency;
```

src/Adapter/ContextLoadHelper.php
```diff
@@ -0,0 +1,140 @@
...
+
+    public function loadEmployeeContext(?int $employeeId = null)
+    {
+        if (null === $employeeId) {
+            $this->context->employee = new Employee();
+
+            return;
+        }
+
+        $this->context->employee = new Employee($employeeId);
```
PierreRambaud:
```suggestion
        $this->context->employee = new Employee($employeeId);
```

tests/Integration/PrestaShopBundle/Command/LoadLegacyClassesinCommandTest.php
PierreRambaud:
```suggestion
        $this->previousErrorReportingLevel = error_reporting(E_WARNING);
```
Otherwise we don't understand why 2 :sweat_smile:

src/Adapter/ContextLoadHelper.php
```diff
@@ -0,0 +1,140 @@
...
+        if (null === $shopId) {
+            $this->context->shop = new Shop();
+            Shop::setContext(Shop::CONTEXT_SHOP, 1);
+
+            return;
+        }
+
+        $this->context->shop = new Shop($shopId);
+        Shop::setContext(Shop::CONTEXT_SHOP, $shopId);
+    }
```
PierreRambaud:
```suggestion
    public function loadShopContext(int $shopId = 1)
    {
        $this->context->shop = new Shop($shopId);
        Shop::setContext(Shop::CONTEXT_SHOP, $shopId);
    }
```

Regarding install.php, and other things like that, we never use empty shop :thinking:

src/Adapter/ContextLoadHelper.php
```diff
@@ -0,0 +1,140 @@
...
+
+    public function loadShopGroupId(int $shopGroupId)
+    {
+        Shop::setContext(Shop::CONTEXT_GROUP, $shopGroupId);
+    }
+
+    public function loadAdminDirectoryContext(string $directory)
```
PierreRambaud: Not sure it's needed, it's hard for commands to define the admin directory, no?

src/Adapter/ContextLoadHelper.php
```diff
@@ -0,0 +1,124 @@
...
+        $this->context = $context->getContext();
+    }
+
+    public function loadGenericContext(
+        ?string $controllerClassName = null,
+        ?int $currencyId = null,
+        ?int $employeeId = null,
+        ?int $shopId = null,
+        ?int $shopGroupId = null)
+    {
```
PierreRambaud:
```suggestion
        ?int $shopGroupId = null
    ) {
```

src/Adapter/ContextLoadHelper.php
```diff
@@ -0,0 +1,141 @@
...
+        }
+        if (null !== $shopGroupId) {
+            $this->loadShopGroupId($shopGroupId);
+        }
+    }
+
+    /**
+     * @param string|null $controllerClassName
+     */
+    public function loadControllerContext(?string $controllerClassName = null)
```
jolelievre: You should return `$this` in all these load functions, this would allow to chain the calls:

```php
$this
    ->loadCurrencyContext($currencyId)
    ->loadControllerContext($controllerClassName)
    ->loadEmployeeContext($employeeId)
;
```

classes/controller/DummyController.php
jolelievre: At first I thought it was used for testing so I was about to suggest moving this into the tests folder But then I saw you use it as a default value when no specific controller is specified

So I understand why you need this, however I find it a bit dangerous to expose a useless controller class like this (because it will be accessible just by its presence)

An alternative could be to use bracket namespace inside your ContextLoadHelper.php file (which @PierreRambaud reminded me of recently), something like this:

```php
namespace {
   class DummyController  extends Controller {
       // ...
   }
}
```

The advantage, if I'm not mistaken, is that with this the DummyController is only accessible by your `ContextLoadHelper` class. To be checked though maybe importing the `ContextLoadHelper` would make it visible but at least it reduces the risk.

Also note that I dropped the `Core` suffix because I don't think this controller should be overridden by anyone.

By the way the dynamic class that you used for test would also do the job (that's what gave me the idea), it's just that it would make the code not very readable.

jolelievre: If the previous mentioned solutions are not acceptable for you, at the very least we should heavily document this controller to explain that it's not meant to be used and is only here for internal purposes related with `ContextLoadHelper`

classes/controller/DummyController.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        return '';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function buildContainer()
+    {
+        return SymfonyContainer::getInstance();
```
jolelievre: Hhmmmm, smart 😅

src/Adapter/ContextLoadHelper.php
```diff
@@ -0,0 +1,141 @@
...
+{
+    /**
+     * @var Context
+     */
+    private $context;
+
+    /**
+     * @param LegacyContext $context
+     */
+    public function __construct(LegacyContext $context)
```
jolelievre: I'm wondering if we should rely on `LegacyContext` or inject `Context` (this can be done in the service definition anyway)
It's just that since `LegacyContext` is not really useful anyway besides getting access to the original `Context` we might as well let the class only dependable on what it needs It would be more flexible as it could be used in other use cases by injecting a pre-created context:

```php
$context = new Context();
// ... something happens
$loader = new ContextLoadHelper($context);
// ... load what you need
```

src/Adapter/ContextLoadHelper.php
jolelievre: So `loadGenericContext()` would have no parameter at all but simply call all the load methods with no parameters, which results in using the default fallback?

That's what I imagined at first indeed, some method you can call out of the box without needing to think of what happens behind

src/Adapter/ContextLoadHelper.php
jolelievre: I think so too, maybe the Load and Helper concepts are a bit redundant 🤔 
Why not `LegacyContextBuilder` ? Or `LegacyContextLoader` ? But I'm not sure the `Helper` suffix brings much information

jolelievre (commented):
A few comments, also I wonder if we should integrate this helper in our existing commands?
Or at least one of them as a practical example?

classes/controller/DummyController.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/21125/commits/8129ee72c6ccbf4e6cee5b7122f9450491c71af3 I used bracket namespace it works good 👍

src/Adapter/ContextLoadHelper.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/21125/commits/2b89cac34e6ebcf56b999f38faf4277a7b485d0f

src/Adapter/ContextLoadHelper.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/21125/commits/2b89cac34e6ebcf56b999f38faf4277a7b485d0f

src/Adapter/ContextLoadHelper.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/21125/commits/2b89cac34e6ebcf56b999f38faf4277a7b485d0f

matks:
@jolelievre All feedbacks fixed 😄

Progi1984:
Please fix PHPStan.

src/Adapter/LegacyContextLoader.php
```diff
@@ -0,0 +1,239 @@
...
+
+            return $this;
+        }
+
+        /**
+         * @param int $shopGroupId
+         *
+         * @return self
+         */
+        public function loadShopGroupId(int $shopGroupId)
```
Progi1984:
```suggestion
        public function loadShopGroupId(int $shopGroupId): self
```

src/Adapter/LegacyContextLoader.php
```diff
@@ -0,0 +1,239 @@
...
+
+            return $this;
+        }
+
+        /**
+         * @param int $shopId
+         *
+         * @return self
+         */
+        public function loadShopContext(int $shopId = 1)
```
Progi1984:
```suggestion
        public function loadShopContext(int $shopId = 1): self
```

src/Adapter/LegacyContextLoader.php
```diff
@@ -0,0 +1,239 @@
...
+
+            return $this;
+        }
+
+        /**
+         * @param int|null $employeeId
+         *
+         * @return self
+         */
+        public function loadEmployeeContext(?int $employeeId = null)
```
Progi1984:
```suggestion
        public function loadEmployeeContext(?int $employeeId = null): self
```

src/Adapter/LegacyContextLoader.php
```diff
@@ -0,0 +1,239 @@
...
+
+            return $this;
+        }
+
+        /**
+         * @param int|null $currencyId
+         *
+         * @return self
+         */
+        public function loadCurrencyContext(?int $currencyId = null)
```
Progi1984:
```suggestion
        public function loadCurrencyContext(?int $currencyId = null): self
```

src/Adapter/LegacyContextLoader.php
```diff
@@ -0,0 +1,239 @@
...
+
+            return $this;
+        }
+
+        /**
+         * @param string|null $controllerClassName
+         *
+         * @return self
+         */
+        public function loadControllerContext(?string $controllerClassName = null)
```
Progi1984:
```suggestion
        public function loadControllerContext(?string $controllerClassName = null): self
```

src/Adapter/LegacyContextLoader.php
```diff
@@ -0,0 +1,239 @@
...
+         *
+         * @return self
+         */
+        public function loadGenericContext(
+            ?string $controllerClassName = null,
+            ?int $currencyId = null,
+            ?int $employeeId = null,
+            ?int $shopId = null,
+            ?int $shopGroupId = null
+        ) {
```
Progi1984:
```suggestion
        ): self {
```

tests/Integration/PrestaShopBundle/Command/LoadLegacyClassesinCommandTest.php
```diff
@@ -0,0 +1,102 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+namespace Tests\Integration\PrestaShopBundle\Command;
```
Progi1984:
```suggestion
 */

declare(strict_types=1);

namespace Tests\Integration\PrestaShopBundle\Command;
```

tests/Integration/PrestaShopBundle/Command/LoadLegacyClassesinCommandTest.php
```diff
@@ -0,0 +1,102 @@
...
+    private $previousErrorReportingLevel;
+
+    public function setUp()
+    {
+        parent::setUp();
+        self::bootKernel();
+        $this->previousErrorReportingLevel = error_reporting(E_WARNING);
+    }
+
+    public function testLoadLegacyCommandWithoutContextFails()
```
Progi1984, line 51:
```suggestion
    public function testLoadLegacyCommandWithoutContextFails(): void
```

tests/Integration/PrestaShopBundle/Command/LoadLegacyClassesinCommandTest.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        $this->expectException(\TypeError::class, 'Not enough arguments (missing: "theme, locale").');
```
Progi1984:
```suggestion
        $this->expectException(\TypeError::class, 'Not enough arguments (missing: "theme, locale").');d
```

define `TypeError` in use.

tests/Integration/PrestaShopBundle/Command/LoadLegacyClassesinCommandTest.php
```diff
@@ -0,0 +1,102 @@
...
+        $application = new Application(static::$kernel);
+        $application->add(new class() extends ContainerAwareCommand {
+            protected function configure()
+            {
+                $this->setName('prestashop-tests:load-legacy-classes');
+            }
+
+            protected function execute(InputInterface $input, OutputInterface $output)
+            {
+                $products = \Product::getNewProducts(1);
```
Progi1984:
```suggestion
                $products = \Product::getNewProducts(1);
```
define Product in use.

Progi1984 (commented):
Fixed some comments

tests/Integration/PrestaShopBundle/Command/LoadLegacyClassesinCommandTest.php
matks: I will add them but I'm having difficulties understanding why it's useful to have strict return type in unit test actions 🤔 . This code is not meant to be called by other functions?

src/Adapter/LegacyContextLoader.php
```diff
@@ -0,0 +1,241 @@
...
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace CommandLineUtils\Controller {
```
PierreRambaud: I'm not sure it's a good practice two namespaces in the same file.
Not sure PSR4 allowing that without a warning :/

matks: It was an idea of https://github.com/PrestaShop/PrestaShop/pull/21125#discussion_r499722110 . It has pros and cons :/ I'm open to alternatives but this DummyController, wherever it is, will be a workaround.

PierreRambaud: I'm more in favor of having a dedicated file per class if it's not related to tests. It will be easier to update, manage, move if needed but if everyone agree, I'm not against :sweat_smile:

matks: Fixed by https://github.com/PrestaShop/PrestaShop/pull/21125/commits/c64bb28fba3725e14699262556984c5605017b85

src/Adapter/LegacyContextLoader.php
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * @param int|null $currencyId
+     *
+     * @return self
+     */
+    public function loadCurrencyContext(?int $currencyId = null): self
+    {
+        $currency = new Currency($currencyId);
+        if (null === $currencyId) {
+            $currency->precision = 2;
```
PierreRambaud:
```suggestion
            $currency->precision = Precision::DEFAULT_PRECISION;
```
Stored in `PrestaShop\PrestaShop\Core\Domain\Currency\ValueObject\ Precision`

matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/21125/commits/98576208b2c085ca211f197109bb94781ef34679

PierreRambaud (approved):

Progi1984 (approved):

SD1982:
Hi @PrestaShop/prestashop-core-developers we let you handle the QA for this one
Thanks guys !!

matks:
Hi @PrestaShop/prestashop-core-developers 

I initially planned the following:

> There is a test that checks a Command using legacy Product ObjectModel works. I will expand the test. When the coverage is good enough I believe there is no need to QA this PR, Travis should be green.

But thanks to @Seb33300 [issue](https://github.com/PrestaShop/PrestaShop/issues/23370) I finally found some more "Impossible to load legacy stuff in Console Commands" usecase, and when I look at how to fix it, I see it's going to make this PR grow very big. Probably 1 class per "legacy stuff" to handle. I see also some places where `$this->context->employee` is likely overriden, so my dummy employee trick is not enough.

So I would like that we validate this PR only, which fixes https://github.com/PrestaShop/PrestaShop/issues/14173, and continue fixing "Impossible to load legacy stuff in Console Commands"  such as https://github.com/PrestaShop/PrestaShop/issues/23370 in other PRs, to keep each PR size manageable?

PierreRambaud:
Thank you @matks

matks:
I can also add a test case where `StockManagementRepository` is used

This is a good example of a class that need Context to be properly initialized
```
trace: {
      ./src/PrestaShopBundle/Entity/Repository/StockManagementRepository.php:130 {
        PrestaShopBundle\Entity\Repository\StockManagementRepository->__construct(ContainerInterface $container, Connection $connection, EntityManager $entityManager, LegacyContext $contextAdapter, ImageManager $imageManager, $tablePrefix)^
        › if (!$this->context->employee instanceof Employee) {
        ›     throw new RuntimeException('Determining the active language requires a contextual employee instance.');
        › }
      }
```

matks:
ContextLoaderHelper should replace https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Command/ModuleCommand.php#L83

PR #21184, opened 2020-09-28
Avoid creating new state with existing iso code

boubkerbribri:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Avoid creating new state with existing iso code in UI tests
| Type?         | refacto
| Category?     | TE
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | `TEST_PATH="functional/BO/11_international/02_locations/03_states/*" URL_FO=shopUrl/ npm run specific-test`

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21184)

### DIFF
--- a/tests/UI/campaigns/functional/BO/11_international/02_locations/03_states/03_bulkActionsStates.js
+++ b/tests/UI/campaigns/functional/BO/11_international/02_locations/03_states/03_bulkActionsStates.js
@@ -26,8 +26,8 @@ let page;
 let numberOfStates = 0;
 
 const statesToCreate = [
-  new StateFaker({name: 'todelete1'}),
-  new StateFaker({name: 'todelete2'}),
+  new StateFaker({name: 'todelete1', isoCode: 'HM'}),
+  new StateFaker({name: 'todelete2', isoCode: 'BV'}),
 ];
 
 describe('Create 2 states then enable, disable and delete by bulk actions', async () => {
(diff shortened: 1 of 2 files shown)


tests/UI/campaigns/functional/BO/11_international/02_locations/03_states/03_bulkActionsStates.js
```diff
@@ -26,8 +26,8 @@ let page;
 let numberOfStates = 0;
 
 const statesToCreate = [
-  new StateFaker({name: 'todelete1'}),
-  new StateFaker({name: 'todelete2'}),
+  new StateFaker({name: 'todelete1', isoCode: 'HM'}),
+  new StateFaker({name: 'todelete2', isoCode: 'BV'}),
```
SimonGrn, line 30: Why did you hardcode these values when faker is taking random values from the list you provided ?

boubkerbribri: The array `statesIsoCodes` is used in CRUDState test : there will be no problem because we create 1 state and we update it
=> So even if State 1 iso code is equal to State 2 iso code, there will be no problem.

For the bulk actions test, we create 2 state at the same time.
=> So using the array, there is `1/64` chance that State 1 iso code is equal to State 2 iso code

SimonGrn (approved):

PierreRambaud (approved):

PR #21219, opened 2020-09-30
Simplify shop parameters/general

JevgenijVisockij:
| Questions     | Answers
| Branch?       | develop
| Description?  | Simplify preferences form
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no 
| Deprecations? | no
| How to test?  |  Simplifying preferences form. Everything must look the same with one exception: blue help box replaced with help text under input. Also some options now are marked as required as they always were.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/20243)

### DIFF
--- a/src/Core/Domain/Tab/CommandHandler/UpdateTabStatusByClassNameHandlerInterface.php
+++ b/src/Core/Domain/Tab/CommandHandler/UpdateTabStatusByClassNameHandlerInterface.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Tab\CommandHandler;
+
+use PrestaShop\PrestaShop\Core\Domain\Tab\Command\UpdateTabStatusByClassNameCommand;
+
+interface UpdateTabStatusByClassNameHandlerInterface
+{
+    public function handle(UpdateTabStatusByClassNameCommand $command): void;
+}
--- a/src/Core/Domain/Tab/CommandHandler/UpdateTabStatusByClassNameHandlerInterface.php
+++ b/src/Core/Domain/Tab/CommandHandler/UpdateTabStatusByClassNameHandlerInterface.php
@@ -24,8 +24,6 @@
  * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
  */
 
-declare(strict_types=1);
-
 namespace PrestaShop\PrestaShop\Core\Domain\Tab\CommandHandler;
 
 use PrestaShop\PrestaShop\Core\Domain\Tab\Command\UpdateTabStatusByClassNameCommand;
--- a/src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/PreferencesController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/PreferencesController.php
@@ -50,7 +50,7 @@ class PreferencesController extends FrameworkBundleAdminController
      * @throws \Exception
      * @AdminSecurity("is_granted('read', request.get('_legacy_controller'))")
      */
-    public function indexAction(Request $request)
+    public function indexAction(Request $request, FormInterface $form = null)
     {
         $form = $this->get('prestashop.adapter.preferences.form_handler')->getForm();
 
--- a/src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/PreferencesController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/PreferencesController.php
@@ -26,11 +26,9 @@
 
 namespace PrestaShopBundle\Controller\Admin\Configure\ShopParameters;
 
-use Doctrine\ORM\EntityManager;
 use PrestaShop\PrestaShop\Adapter\Tools;
+use PrestaShop\PrestaShop\Core\Domain\Tab\Command\UpdateTabStatusByClassNameCommand;
 use PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController;
-use PrestaShopBundle\Entity\Repository\TabRepository;
-use PrestaShopBundle\Entity\Tab;
 use PrestaShopBundle\Security\Annotation\AdminSecurity;
 use PrestaShopBundle\Security\Annotation\DemoRestricted;
 use Symfony\Component\Form\FormInterface;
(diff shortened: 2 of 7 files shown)


atomiix (dismissed):

matks (dismissed):

src/PrestaShopBundle/Form/Admin/Configure/ShopParameters/General/PreferencesType.php
```diff
@@ -90,42 +90,121 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                ),
+            ]);
         }
 
         $builder
             ->add('enable_ssl_everywhere', SwitchType::class, [
                 'disabled' => !$isSslEnabled,
+                'label' => $this->trans(
+                    'Enable SSL on all pages',
+                    'Admin.Shopparameters.Help'
```
Julievrz: Shouldn't it be located in Admin.Shopparameters.Feature?

src/PrestaShopBundle/Resources/views/Admin/Configure/ShopParameters/preferences.html.twig
```diff
@@ -37,144 +37,31 @@
...
+                        </label>
+                      <div class="col-sm">
+                        <div class="input-group">
+                          <div class="form-control-plaintext">
+                            <a class="d-block" href="{{ sslUri }}">
+                              {{ 'Please click here to check if your shop supports HTTPS.'|trans({}, 'Admin.Shopparameters.Feature') }}
+                            </a>
+                          </div>
+                          <small class="form-text">
+                            {{ 'If you want to enable SSL on all the pages of your shop, activate the "Enable on all the pages" option below.'|trans({}, 'Admin.Shopparameters.Feature') }}
```
Julievrz: I think it should be located in Admin.Shopparameters.Help and not in Admin.Shopparameters.Feature.

ngodefroy:
Hello @JevgenijVisockij 

After I installed the PR, the save button has disappeared:

Capture d’écran 2020-10-16 à 15 00 27

Thank you

src/PrestaShopBundle/Resources/views/Admin/Configure/ShopParameters/preferences.html.twig
JevgenijVisockij: Yeah you are right!

JevgenijVisockij:
Fixed the issue with missing save button and fixed 2 translation domains.

src/PrestaShopBundle/Resources/views/Admin/Configure/ShopParameters/preferences.html.twig
```diff
@@ -37,144 +37,37 @@
...
-                {% block shop_preferences_form_rest %}
-                  {{ form_rest(generalForm) }}
-                {% endblock %}
+                {{ form_widget(generalForm) }}
               </div>
             </div>
             <div class="card-footer">
               <div class="d-flex justify-content-end">
                 <button class="btn btn-primary" id="form-preferences-save-button">{{ 'Save'|trans({}, 'Admin.Actions') }}</button>
+                {{ form_widget(generalForm) }}
```
atomiix: Is it intended to have `form_widget(generalForm)` twice? 🤔

JevgenijVisockij: Not it wasn't. Fixed it!

atomiix (dismissed):

PierreRambaud (dismissed):

src/PrestaShopBundle/Form/Admin/Configure/ShopParameters/General/PreferencesType.php
```diff
@@ -90,42 +90,121 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                    'Clean the HTML content on text fields. We recommend that you leave this option enabled.',
+                    'Admin.Shopparameters.Help'
+                ),
             ])
+            ->add(
+                'price_round_mode', ChoiceType::class, [
+                    'placeholder' => false,
+                    'choices_as_values' => true,
+                    'choices' => [
+                        'Round up away from zero, when it is half way there (recommended)' => $configuration->get('PS_ROUND_HALF_UP'),
```
jolelievre, line 150: One question about those choice fields They are not translated here But I believe Symfony has automatic translation for choices fields and you can even define a translation domain

So I guess we should use `Admin.Shopparameters.Feature` as `choice_translation_domain` option (considering these wordings are present in this domain)

jolelievre: Ok so it seems it actually works because the template has this `{% trans_default_domain "Admin.Shopparameters.Feature" %}` So the correct domain will be used anyway, so it's not blocking but still maybe it would be better to explicitely set the choice domain in the form?
This is not blocking for the PR anyway and can be improved later, alongs with other similar cases I guess

Progi1984 (dismissed):

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/PreferencesController.php
```diff
@@ -102,32 +80,53 @@ public function processFormAction(Request $request)
...
 
-        if (0 === count($saveErrors)) {
-            /** @var EntityManager $em */
-            $em = $this->get('doctrine.orm.entity_manager');
+                /** @var TabRepository $tabRepository */
+                $tabRepository = $em->getRepository(Tab::class);
 
-            /** @var TabRepository $tabRepository */
-            $tabRepository = $em->getRepository(Tab::class);
+                $tabRepository->changeStatusByClassName(
```
atomiix: I'm wondering if it shouldn't be handled by a CQRS command? But that's probably out of scope 🤔

JevgenijVisockij: Good point. IMO since I already started improving controllers this can be done as well.

JevgenijVisockij: I updated the PR to use the command

src/Core/Domain/Tab/CommandHandler/UpdateTabStatusByClassNameHandlerInterface.php
```diff
@@ -0,0 +1,36 @@
...
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
```
Progi1984:
```suggestion
```

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/PreferencesController.php
```diff
@@ -102,32 +78,49 @@ public function processFormAction(Request $request)
...
+            $this->flashErrors($saveErrors);
+        }
 
-            $this->addFlash('success', $this->trans('Successful update.', 'Admin.Notifications.Success'));
+        return $this->renderForm($request, $form);
+    }
 
-            return $this->redirectToRoute('admin_preferences');
-        }
+    private function renderForm(Request $request, FormInterface $form)
```
Progi1984:
```suggestion
    private function renderForm(Request $request, FormInterface $form): Response
```

src/PrestaShopBundle/Controller/Admin/Configure/ShopParameters/PreferencesController.php
```diff
@@ -46,39 +44,17 @@ class PreferencesController extends FrameworkBundleAdminController
...
-     * @AdminSecurity("is_granted('read', request.get('_legacy_controller'))")
      *
      * @return Response
      *
-     * @throws \LogicException
+     * @throws \Exception
+     * @AdminSecurity("is_granted('read', request.get('_legacy_controller'))")
      */
-    public function indexAction(Request $request, FormInterface $form = null)
+    public function indexAction(Request $request)
```
Progi1984: BC Break, you change the signature

Progi1984 (commented):
Small feedbacks

Progi1984 (approved):

matks (approved):

src/PrestaShopBundle/Form/Admin/Configure/ShopParameters/General/PreferencesType.php
matks: Now I think I remember Mickael saying something like "we cannot use some Symfony Form features like Constraints because Symfony expects only 1 domain" 🤔

Roomsmaniia:
LGTM,

Thanks @JevgenijVisockij

Progi1984:
Thanks @JevgenijVisockij & @RomainBocheux

PR #21227, opened 2020-09-30
Refacto ObjectModel::existsInDatabase

### LINKED ISSUE #21210: function "existsInDatabase" in "ObjectModel" discrepancy with "orders" table
#### Describe the bug

Function `existsInDatabase` inside `ObjectModel.php` use `$id_entity` and `$table` as parameters.
In case of `orders` this call will throw a **PrestaShopDatabaseException** because the query won't find any fields called **id_orders**.
In Prestashop current design it has been chosen **orders** (plural version) instead of **order** (singular version) which means the query will become something like this

```
SELECT `id_orders` as id
FROM `prefix_orders` e
WHERE e.`id_orders` = $id LIMIT 1
```

#### Expected behavior

As you can understand from above explanation this might be either a database design error or a missed update on the function `existsInDatabase`. So there are 2 potential fixes:

The first one will necessarily face changing the name table from **orders** to **order** in order to have a match with its auto incremental id.

The second case will see an update of the function itself from:

```
public static function existsInDatabase($id_entity, $table)
    {
        $row = Db::getInstance()->getRow('
			SELECT `id_' . bqSQL($table) . '` as id
			FROM `' . _DB_PREFIX_ . bqSQL($table) . '` e
			WHERE e.`id_' . bqSQL($table) . '` = ' . (int) $id_entity, false
        );

        return isset($row['id']);
    }
```
to something like

```
public static function existsInDatabase($id_entity, $table, $field)
    {
        $row = Db::getInstance()->getRow('
			SELECT `id_' . bqSQL($field) . '` as id
			FROM `' . _DB_PREFIX_ . bqSQL($table) . '` e
			WHERE e.`id_' . bqSQL($field) . '` = ' . (int) $id_entity, false
        );

        return isset($row['id']);
    }
```
This means an extra parameter where you can define the field of the table of interest. I feel like it is more a database design error so the first solution would be the best one, looking at the current database design it looks like it has been chosen the _singular version naming rule_ for the majority of the tables.

### Note

I didn't check other broken case scenarios, with other kind of entities.

#### Steps to Reproduce

Working example:
`Product::existsInDatabase(1, 'product');`

Broken example:
`Order::existsInDatabase(1, 'orders');`

#### Additional information

* PrestaShop version: 1.7.5.2
* PHP version: 7.2.22
* MySql: 5.7

PululuK:
| Questions     | Answers
| Branch?       | develop 
| Description?  | See https://github.com/PrestaShop/PrestaShop/issues/21210 by @roncarino.
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | yes
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/21210
| How to test?  | See https://github.com/PrestaShop/PrestaShop/issues/21210 by @roncarino

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21227)

### DIFF
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -1890,19 +1890,32 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
      * Checks if an object exists in database.
      *
      * @param int $id_entity
-     * @param string $table
+     * @param string $table Deprecated since 1.7.8.x
      *
      * @return bool
      */
-    public static function existsInDatabase($id_entity, $table)
+    public static function existsInDatabase($id_entity, $table = null)
     {
+        if ($table !== null && static::class == 'ObjectModel') {
+            $primary = 'id_' . bqSQL($table);
+        } else {
+            $object_def = static::$definition;
+
+            if (empty($object_def['table']) || empty($object_def['primary'])) {
+                return false;
+            }
+
+            $table = $object_def['table'];
+            $primary = $object_def['primary'];
+        }
+
         $row = Db::getInstance()->getRow(
-            '
-			SELECT `id_' . bqSQL($table) . '` as id
-			FROM `' . _DB_PREFIX_ . bqSQL($table) . '` e
-			WHERE e.`id_' . bqSQL($table) . '` = ' . (int) $id_entity,
-            false
-        );
+                (new DbQuery())
+                    ->select('e.`' . $primary . '` as id')
+                    ->from($table, 'e')
+                    ->where('e.`' . $primary . '` = ' . (int) $id_entity),
+                false
+            );
 
         return isset($row['id']);
     }
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -1896,6 +1896,10 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
      */
     public static function existsInDatabase($id_entity, $table = null)
     {
+        if ($table !== null) {
+            Tools::displayParameterAsDeprecated('table');
+        }
+
         if ($table !== null && static::class == 'ObjectModel') {
             $primary = 'id_' . bqSQL($table);
         } else {


classes/ObjectModel.php
```diff
@@ -1876,15 +1876,28 @@ public function deleteImage($force_delete = false)
...
+        if ($table === null) {
+            $object_def = static::$definition;
+
+            if (!array_key_exists('table', $object_def) || !array_key_exists('primary', $object_def)) {
+                return false;
+            }
+
+            $table = $object_def['table'];
+            $primary = $object_def['primary'];
+        }
```
PierreRambaud:
```suggestion
        if ($table !== null) {
            $primary = 'id_' . bqSQL($table);
        } else {
            $object_def = static::$definition;

            if (!isset($object_ref['table']) || !isset($object_ref['primary'])) {
                return false;
            }

            $table = $object_def['table'];
            $primary = $object_def['primary'];
        }
```

Do not declare something you will not use if $table is null.
Always prefer isset instead of array_key_exists if you don't care null value, it's 2.5 faster.

PierreRambaud (changes requested):
Nice PR, just a littme comment :)

classes/ObjectModel.php
PululuK: Merci !

PierreRambaud (dismissed):

classes/ObjectModel.php
```diff
@@ -1876,15 +1876,28 @@ public function deleteImage($force_delete = false)
...
+
+            if (!isset($object_ref['table']) || !isset($object_ref['primary'])) {
+                return false;
+            }
+
+            $table = $object_def['table'];
+            $primary = $object_def['primary'];
+        }
+
         $row = Db::getInstance()->getRow(
```
Sinepel, line 1916:
```suggestion
        $exists = Db::getInstance()->getValue(
```

classes/ObjectModel.php
```diff
@@ -1876,15 +1876,28 @@ public function deleteImage($force_delete = false)
...
-            false
-        );
+                (new DbQuery())
+                    ->select('e.`' . $primary . '` as id')
+                    ->from($table, 'e')
+                    ->where('e.`' . $primary . '` = ' . (int) $id_entity),
+                true
+            );
 
         return isset($row['id']);
```
Sinepel, line 1924:
```suggestion
        return (bool) $exists;
```

Sinepel (changes requested):
It's just a suggestion, tell me what you think.

I think it can improve and simplify the code a little bit.

PululuK:
> It's just a suggestion, tell me what you think.
> 
> I think it can improve and simplify the code a little bit.

Thank  you @Sinepel 
But ... I think it's not better in terms of performance!

`getValue = getRow + array_shift` . See [`Db::getValue`](https://github.com/PrestaShop/PrestaShop/blob/develop/classes/db/Db.php#L691).

Sinepel:
Hi @PululuK ! 

Yes i know :D 

Simplifying the code sometimes has some disadvantages!  

It depends on what you prefer and what the core team prefers.  

We should also be able to test what that represents as differences in performance.

No problem if it's not accepted 😃👍.

PululuK:
Is it really necessary in this case 🤔?

> It depends on what you prefer and what the core team prefers.
> We should also be able to test what that represents as differences in performance.

But ... I'm not a Prestashop maintainer, so I can't decide anything 😀
> No problem if it's not accepted .

Sinepel:
We could test but I don't think it is necessary for this case indeed :)

I can't decide either 😃 !

sowbiba (dismissed):

classes/ObjectModel.php
```diff
@@ -1876,15 +1876,28 @@ public function deleteImage($force_delete = false)
...
-			SELECT `id_' . bqSQL($table) . '` as id
-			FROM `' . _DB_PREFIX_ . bqSQL($table) . '` e
-			WHERE e.`id_' . bqSQL($table) . '` = ' . (int) $id_entity,
-            false
-        );
+                (new DbQuery())
+                    ->select('e.`' . $primary . '` as id')
+                    ->from($table, 'e')
+                    ->where('e.`' . $primary . '` = ' . (int) $id_entity),
+                true
```
PierreRambaud:
```suggestion
                false
```

PululuK: @PierreRambaud  Why are you disabling the cache? 🤔

jolelievre: I agree with @PierreRambaud using cache is often prone to mistake (especially in PrestaShop where it's used badly quite often ^^)
For a critical method like this one that checks if an object exists we shouldn't cache the result If the object has been deleted during the process, the method will keep returning that it exists And we have no easy way to clear this cache on deletion either so we can't keep it consistent

Sinepel: I agree as well. I didn't pay attention to this point in my previous suggestion.

PululuK: Thanks guys !

PierreRambaud (dismissed):

sowbiba (dismissed):

ngodefroy:
@PrestaShop/prestashop-core-developers

PululuK:
> @PrestaShop/prestashop-core-developers


Hi @ngodefroy 
You need help ? I'm here 😊

classes/ObjectModel.php
```diff
@@ -1876,15 +1876,28 @@ public function deleteImage($force_delete = false)
      *
      * @return bool
      */
-    public static function existsInDatabase($id_entity, $table)
+    public static function existsInDatabase($id_entity, $table = null)
     {
+        if ($table !== null) {
```
jolelievre, line 1899: Hi @PululuK based on the initial issue, this will not fix the problem when calling `Order::existsInDatabase(1, 'orders');` because it will still fail with the `orders` parameter since `id_orders` does not exist

I wonder if we could not use the definition at all times and make `$table` deprecated? Or the table could be used in case the definition isn't correctly set like if you try to use `ObjectModel::existsInDatabase(1, 'product');` then you can only rely on the table parameter.

ngodefroy:
> > @PrestaShop/prestashop-core-developers
> 
> Hi @ngodefroy
> You need help ? I'm here 😊

@PululuK After concerting with the QA, we believed it would be more suitable if dev team could test it (I mentioned this PR in a post on team-core)

PululuK:
> @PululuK After concerting with the QA, we believed it would be more suitable if dev team could test it (I mentioned this PR in a post on team-core)


thanks @ngodefroy  I will check with core-team then!

classes/ObjectModel.php
PululuK: Hi @jolelievre  
Thanks for your recommendations... it is done !
 Can you take a look please?

Progi1984 (dismissed):

classes/ObjectModel.php
Progi1984: PR from @PululuK #21616

Progi1984:
@PululuK Could you rebase this PR and remove the "Merge commit", please ?

PululuK:
> @PululuK Could you rebase this PR and remove the "Merge commit", please ?

Done ! @Progi1984 sorry for the delay 😔

classes/ObjectModel.php
```diff
@@ -1872,19 +1872,32 @@ public function deleteImage($force_delete = false)
      * Checks if an object exists in database.
      *
      * @param int $id_entity
-     * @param string $table
+     * @param string $table Deprecated since 1.7.8.x
```
Progi1984, line 1893: Could use `Tools::displayParameterAsDeprecated` if $table is not null.

PululuK: Done ! Thanks

classes/ObjectModel.php
Progi1984: ping @jolelievre

PululuK:
Hello !
News about ? 
I'm thinking of adding tests here. What do you think ?

jolelievre (dismissed):
Thanks @PululuK 
Sorry for the late review 😅

Sinepel (dismissed):

PierreRambaud (dismissed):

khouloudbelguith:
Hi @PululuK,

It is ok :heavy_check_mark: 
1. I edited the file `/PrestaShop/controllers/admin/AdminImportController.php`
2. I added those lines:
```
dump(Product::existsInDatabase(1, 'product'));
dump(Product::existsInDatabase((int) $product->id, 'product'));
dump(Order::existsInDatabase(1, 'orders'));
exit();
```
3. I tried to import products

**Without PR**
image

**With PR**
image

Thanks!

kpodemski:
Hey @PululuK. Unfortunately, we can't merge it. You need to rebase this PR, can you do it? 👍

Progi1984:
@PululuK Could you rebase your PR and remove your merge commit, please ?

Progi1984:
@PululuK Ping for rebase.

PululuK:
> @PululuK Ping for rebase.

Hi @Progi1984, can you please do it ?

PierreRambaud:
> > @PululuK Ping for rebase.
> 
> Hi @Progi1984, can you please do it ?

I did it

Progi1984 (approved):

kpodemski:
@PierreRambaud 

:D I pushed my rebase to https://github.com/kpodemski/PrestaShop/commits/patch-15 by mistake and then got rejected from pushing to PululuK repo 😂 you were faster :D

kpodemski (approved):

kpodemski:
Waiting for CI 🟢

matks:
Close and reopen to reset the CI

jolelievre (approved):

PierreRambaud:
Thank you @PululuK

PululuK:
> > > @PululuK Ping for rebase.
> > 
> > 
> > Hi @Progi1984, can you please do it ?
> 
> I did it

Thank you so much @PierreRambaud

PR #21243, opened 2020-10-01
Simplified the Administration form

JevgenijVisockij:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Simplifies administration form
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Part of #16482
| How to test?  | Simplifying Advanced Parameters -> Administration form. Everything must look the same with couple exceptions: blue help box replaced with help text under input, required fields are marked as required now. Also form should throw errors now in case you input value that doesn't make sense. For example 'aaa' for Lifetime of front office cookies.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21243)

PR #21244, opened 2020-10-01
Simplified carrier preferences

JevgenijVisockij:
| Questions     | Answers
| Branch?       | develop
| Description?  | Simplfiies carrier preferences form
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Simplifying Improve -> Carrier -> Preferences form. Everything must look the same.

Aside for form simplification this PR does 2 things: Makes it so Symfony handles validation via constraint functionality and adds postProcess to LegacyController to make sure it actually works. Some of the params may will have red * next to it as they were required before but now it's actually displayed.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21243)

### DIFF
--- a/controllers/admin/AdminLegacyLayoutController.php
+++ b/controllers/admin/AdminLegacyLayoutController.php
@@ -106,7 +106,6 @@ class AdminLegacyLayoutControllerCore extends AdminController
      */
     public function initProcess()
     {
-
     }
 
     /**
--- a/controllers/admin/AdminLegacyLayoutController.php
+++ b/controllers/admin/AdminLegacyLayoutController.php
@@ -101,6 +101,15 @@ class AdminLegacyLayoutControllerCore extends AdminController
     }
 
     /**
+     * This helps avoiding handling legacy processes when in Symfony Controllers.
+     * Otherwise when using POST action to render form you sometimes get an exception.
+     */
+    public function initProcess()
+    {
+
+    }
+
+    /**
      * @param bool $isNewTheme
      */
     public function setMedia($isNewTheme = false)
--- a/src/PrestaShopBundle/Controller/Admin/Improve/Shipping/PreferencesController.php
+++ b/src/PrestaShopBundle/Controller/Admin/Improve/Shipping/PreferencesController.php
@@ -29,7 +29,6 @@ namespace PrestaShopBundle\Controller\Admin\Improve\Shipping;
 use PrestaShop\PrestaShop\Core\Form\FormHandlerInterface;
 use PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController;
 use PrestaShopBundle\Security\Annotation\AdminSecurity;
-use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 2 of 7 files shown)


JevgenijVisockij:
Similarly to https://github.com/PrestaShop/PrestaShop/pull/21243 this form has 2 fields (Handling charges and Free shipping starts at) which can only really be float type but you can put anything in and form will throw a success message( tough wrong values won't be saved).

Interestingly enough if you put wrong value into Free shipping field, it will throw an error message. So I wonder if I should make other 2 fields throw errors as well if value inputted is wrong?

JevgenijVisockij:
Travis failed because 

"E: Failed to fetch https://deb.nodesource.com/node_10.x/pool/main/n/nodejs/nodejs_10.22.1-1nodesource1_amd64.deb  Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds"

So probably will be enough to just run it again.

atomiix (dismissed):

matks (dismissed):

PierreRambaud (dismissed):

controllers/admin/AdminLegacyLayoutController.php
```diff
@@ -100,6 +100,14 @@ public function __construct(
         $this->useRegularH1Structure = $useRegularH1Structure;
     }
 
+    /**
+     * This helps avoiding handling legacy processes when in Symfony Controllers.
+     * Otherwise when using POST action to render form you sometimes get an exception.
+     */
+    public function initProcess()
```
matks, line 107: Hoho this one could introduce dangerous side-effects because it will disable this for _every migrated page_ (ping @atomiix who works on the Legacy Layout migration).

matks: OK, I checked the code of `AdminController::initProcess()` and it seems nothing in there is required for migrated pages. So this should not create issues (in theory 😅 )

matks (dismissed):

controllers/admin/AdminLegacyLayoutController.php
JevgenijVisockij: Yeah this place is my biggest concern since it's such a fundamental function. But so far couldn't figure out anything that could go wrong.

atomiix (approved):

matks (approved):

khouloudbelguith:
Hi @JevgenijVisockij,

It is ok.
But, there is only one difference.
**With PR**
image

**Without PR**
image

Thanks to fix it.

Thank you!

JevgenijVisockij:
> Hi @JevgenijVisockij,
> 
> It is ok.
> But, there is only one difference.
> **With PR**
> image
> 
> **Without PR**
> image
> 
> Thanks to fix it.
> 
> Thank you!

Hey I forgot to update description before, (did it now). Those 3 fields will appear as required now, they always were required, but before form simplification, it wasn't visible. Now it is.

khouloudbelguith:
Hi @JevgenijVisockij,

It is ok :heavy_check_mark: 

Thanks!

Progi1984:
Thanks @JevgenijVisockij & @khouloudbelguith

Issue #21281, opened 2020-10-03
[WIP] List/Grid/View color field improvements


PululuK:
- [x] #21280
- [x] #21283
- [x] #21284

khouloudbelguith:
Hi @PululuK,

The **Expected behavior**
You should not be able to submit the form without providing a valid value for the "color" field.
As reported here: https://github.com/PrestaShop/PrestaShop/issues/20441, so can we close this ticket?

Thanks!

khouloudbelguith:
Hi @PululuK,

Can I assign you this issue? 

Thanks!

PR #21282, opened 2020-10-03
The status name is not visible in the listing when the colors field is empty

### LINKED ISSUE #21280: [BO][LEGACY][LIST] The status name is not visible in the listing when the colors field is empty.
## Is your feature request related to a problem?

The status name is not visible in the listing when the colors field is empty.

## Describe the solution you'd like

Define a default color when the colors field is empty.

| Before      | After |
| ----------- | ----------- |
| 94993029-64d35b00-058e-11eb-8940-c4c120b5a35d (1) | image|

## Alternatives you've considered

The color field must be required in `OrderState` and `OrderReturnState` objects definition :

```php
'color' => [
 'type' => self::TYPE_STRING, 
 'validate' => 'isColor',
 'required' => true // Here
],
```
⚠️  This may introduce BC. Ex.: _imagine modules that create statuses without the color field_

### 🎥 [Click to see screen record](https://drive.google.com/file/d/1tcLhaUrs91xl4ylIsGH4G1gomwfD1up_/view)

PululuK:
| Questions     | Answers
| Branch?       | develop
| Description?  | The status name is not visible in the listing when the colors field is empty.
| Type?         | refacto
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/21280 part of https://github.com/PrestaShop/PrestaShop/issues/21281
| How to test?  | See https://github.com/PrestaShop/PrestaShop/issues/21280

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21282)

### DIFF
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -2251,13 +2251,13 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
             return $this->color;
         }
 
-        return !empty($this->color) ? $this->color : self::DEFAULT_COLOR;
+        return !empty($this->color) ? $this->color : static::DEFAULT_COLOR;
     }
 
     public function setDefautlColor()
     {
         if ($this->hasColorField() && $this->use_default_color) {
-            $this->{$this->color_field_name} = $this->{$this->color_field_name} ?: self::DEFAULT_COLOR;
+            $this->{$this->color_field_name} = $this->{$this->color_field_name} ?: static::DEFAULT_COLOR;
         }
     }
 
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -52,6 +52,8 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
     const HAS_ONE = 1;
     const HAS_MANY = 2;
 
+    const DEFAULT_COLOR = '#ffffff';
+
     /** @var int Object ID */
     public $id;
 
@@ -144,6 +146,10 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
 
     /** @var PrestaShopBundle\Translation\Translator */
     protected $translator;
+
+    /** @var string */
+    protected $use_default_color = false;
+    protected $color_field_name = 'color';
 
     /**
      * @var array Contains object definition
@@ -553,6 +559,11 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
             }
         }
 
+        // Set default color field value when this is empty
+        if (property_exists($this, $this->color_field_name) && empty($this->{$this->color_field_name})) {
+            $this->setDefautlColor();
+        }
+
         // Database insertion
         if (Shop::checkIdShopDefault($this->def['table'])) {
             $this->id_shop_default = (in_array(Configuration::get('PS_SHOP_DEFAULT'), $id_shop_list) == true) ? Configuration::get('PS_SHOP_DEFAULT') : min($id_shop_list);
@@ -724,6 +735,11 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
             if (isset($this->update_fields) && is_array($this->update_fields) && count($this->update_fields)) {
                 $this->update_fields['date_add'] = true;
             }
+        }
+
+        // Set default color field value when this is empty
+        if (property_exists($this, $this->color_field_name) && empty($this->{$this->color_field_name})) {
+            $this->setDefautlColor();
         }
 
         $id_shop_list = Shop::getContextListShopID();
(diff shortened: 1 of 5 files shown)


PululuK:
This PR is required for https://github.com/PrestaShop/PrestaShop/issues/21281

matks:
Hi @PululuK I am sorry but when reviewing this PR I'm wondering ... isn't it a bit complex for such a simple usecase 😅 ?

If I am not wrong, in this PR
- you add 2 configuration settings into ObjectModel and some logic functions related
- you use these configuration settings only for OrderState and OrderReturnState (so the changes in ObjectModel bring no value to every other ObjectModel inside PrestaShop codebase)
- you use these new features into HelperList and a controller

If this feature was going to be reused a lot for multiple ObjectModels it would make sense. But very few listings actually use such colors... I'm afraid in this PR we're bringing something that can be used for a lot of usecases but will maybe never be used outside of this usecase.

So maybe we should follow [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle here and do something really simpler for the listing, and wait to see if we see more usecases where such a solution is needed ? I'm thinking that a very simple rule inside the Controller would be enough. Also it makes more sense to me to write this "default fallback color logic" into the Controller or the Smarty template as it fits inside MVC principles. Having color logic into the Model seems like we're mixing the M and the V 😅 (or the C).

PululuK:
Hi @matks  nope !
This feature can be used in all controllers and objects.
I agree that I was not clear enough in my PR. I will provide the documentation, concrete use cases and a module to demonstrate it 😋

matks:
> Hi @matks nope !
> This feature can be used in all controllers and objects.
> I agree that I was not clear enough in my PR. I will provide the documentation, concrete use cases and a module to demonstrate it 😋

Yes, it can be used in all controllers or ObjectModels. But will it be used ? Are there a lot of usecases where this will be useful and consequently used ?

PululuK:
> Yes, it can be used in all controllers or ObjectModels. But will it be used ? Are there a lot of usecases where this will be useful and consequently used ?


## 1 - Use colors to differentiate carriers

image


## 2 - Use color to differentiate logs types

image


Etc etc etc


@matks In any case, I will provide more concrete cases and an example module!

classes/ObjectModel.php
```diff
@@ -145,6 +147,10 @@ abstract class ObjectModelCore implements \PrestaShop\PrestaShop\Core\Foundation
     /** @var PrestaShopBundle\Translation\Translator */
     protected $translator;
 
+    /** @var string */
+    protected $use_default_color = false;
+    protected $color_field_name = 'color';
```
Progi1984, line 152:
```suggestion
    /** @var bool */
    protected $use_default_color = false;
    
    /** @var string */
    protected $color_field_name = 'color';
```

classes/ObjectModel.php
```diff
@@ -553,6 +559,11 @@ public function add($auto_date = true, $null_values = false)
             }
         }
 
+        // Set default color field value when this is empty
+        if (property_exists($this, $this->color_field_name) && empty($this->{$this->color_field_name})) {
+            $this->setDefautlColor();
```
Progi1984, line 564:
```suggestion
            $this->setDefaultColor();
```

classes/ObjectModel.php
```diff
@@ -726,6 +737,11 @@ public function update($null_values = false)
...
```
Progi1984, line 742:
```suggestion
            $this->setDefaultColor();
```

classes/ObjectModel.php
```diff
@@ -2210,4 +2226,43 @@ public function addFieldsToUpdate(array $fields): void
             $this->update_fields[$fieldName] = $value;
         }
     }
+
+    public function hasColorField()
+    {
+        $def = $this->def;
+        $hasDefinition = !empty($def) && is_array($def) && array_key_exists('fields', $def);
```
Progi1984, line 2233:
```suggestion
        $hasDefinition = is_array($def) && array_key_exists('fields', $def);
```

classes/ObjectModel.php
```diff
@@ -2210,4 +2226,43 @@ public function addFieldsToUpdate(array $fields): void
             $this->update_fields[$fieldName] = $value;
         }
     }
+
+    public function hasColorField()
```
Progi1984, line 2230:
```suggestion
    /**
     * @return bool
     */
    public function hasColorField(): bool
```

classes/ObjectModel.php
```diff
@@ -2210,4 +2226,43 @@ public function addFieldsToUpdate(array $fields): void
...
+        if (!$hasDefinition || empty($this->color_field_name)) {
+            return false;
+        }
+
+        $field = $this->color_field_name;
+
+        return array_key_exists($field, $def['fields']) && $def['fields'][$field]['validate'] == 'isColor';
+    }
+
+    public function getColor()
```
Progi1984, line 2244:
```suggestion
    /**
     * @return string|null
     */
    public function getColor(): ?string
```

classes/ObjectModel.php
```diff
@@ -2210,4 +2226,43 @@ public function addFieldsToUpdate(array $fields): void
...
+        }
+
+        if (!$this->use_default_color) {
+            return $this->color;
+        }
+
+        return $this->color ?: static::DEFAULT_COLOR;
+    }
+
+    public function setDefautlColor()
```
Progi1984, line 2257:
```suggestion
    /**
     * @return void
     */
    public function setDefaultColor(): void
```

classes/ObjectModel.php
```diff
@@ -2210,4 +2226,43 @@ public function addFieldsToUpdate(array $fields): void
...
+    {
+        if ($this->hasColorField() && $this->use_default_color) {
+            $this->{$this->color_field_name} = $this->{$this->color_field_name} ?: static::DEFAULT_COLOR;
+        }
+    }
+
+    public function getColorFieldName()
```
Progi1984, line 2264:
```suggestion
    /**
     * @return string
     */
    public function getColorFieldName(): string
```

Progi1984 (commented):
Some comments & A rebase

kpodemski:
As I understand that it's maybe simpler for developer to add bg color now, I'd still go with callback instead of this, sorry.  

Using colors like this is **very** rare, and it's super easy to do with `callback`.

Maybe we could ask more module developers about it on Slack 🤔

cc @PrestaEdit @prestamodule @202-ecommerce

matks:
Hi everyone,

It looks like this PR is not moving anymore. Is it stale? Does someone wants to continue it?

PierreRambaud:
Seems this PR is not maintained anymore.
I'm closing it, if you think you will have time to rework on it, feel free to reopen, we will be glad to review it and test it :)

PR #21289, opened 2020-10-04
Refactor admin notifications.js

davidglezz:
| Questions     | Answers
| Branch?       | develop
| Description?  | The code before this patch had unused variables, repeating code, and functions with hundreds of lines ...  Initially I was only going to correct certain more important aspects, but seeing so much repeated code I have ended up refactoring it to make it more readable and maintainable.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | BO notifications work as expected

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21289)

### DIFF
--- a/admin-dev/themes/default/template/header.tpl
+++ b/admin-dev/themes/default/template/header.tpl
@@ -288,19 +288,19 @@
                   {$active = "active"}
                   {if $show_new_orders}
                     <li class="nav-item {$active}">
-                      <a class="nav-link" data-toggle="tab" data-type="order" href="#orders-notifications" role="tab" id="orders-tab">{l s='Latest orders' d='Admin.Navigation.Header'}<span id="orders_notif_value"></span></a>
+                      <a class="nav-link" data-toggle="tab" data-type="order" href="#orders-notifications" role="tab" id="orders-tab">{l s='Latest orders' d='Admin.Navigation.Header'}<span id="orders_notif_value" class="notif-counter"></span></a>
                     </li>
                     {$active = ""}
                   {/if}
                   {if $show_new_customers}
                     <li class="nav-item {$active}">
-                      <a class="nav-link" data-toggle="tab" data-type="customer" href="#customers-notifications" role="tab" id="customers-tab">{l s='New customers' d='Admin.Navigation.Header'}<span id="customers_notif_value"></span></a>
+                      <a class="nav-link" data-toggle="tab" data-type="customer" href="#customers-notifications" role="tab" id="customers-tab">{l s='New customers' d='Admin.Navigation.Header'}<span id="customers_notif_value" class="notif-counter"></span></a>
                     </li>
                     {$active = ""}
                   {/if}
                   {if $show_new_messages}
                     <li class="nav-item {$active}">
-                      <a class="nav-link" data-toggle="tab" data-type="customer_message" href="#messages-notifications" role="tab" id="messages-tab">{l s='Messages' d='Admin.Global'}<span id="customer_messages_notif_value"></span></a>
+                      <a class="nav-link" data-toggle="tab" data-type="customer_message" href="#messages-notifications" role="tab" id="messages-tab">{l s='Messages' d='Admin.Global'}<span id="customer_messages_notif_value" class="notif-counter"></span></a>
                     </li>
                     {$active = ""}
                   {/if}
--- a/js/admin/notifications.js
+++ b/js/admin/notifications.js
@@ -24,7 +24,7 @@
  */
 $(document).ready(function () {
   if (youEditFieldFor) {
-    $('.translatable span.hint').append(`<br /><span class="red">${youEditFieldFor}</span>`)
+    $('.translatable span.hint').append(`<br /><span class="red">${youEditFieldFor}</span>`);
   }
 
   $('.notification.dropdown-toggle').on('click', function () {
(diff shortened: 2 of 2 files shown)


admin-dev/themes/default/template/header.tpl
```diff
@@ -288,19 +288,19 @@
                   {$active = "active"}
                   {if $show_new_orders}
                     <li class="nav-item {$active}">
-                      <a class="nav-link" data-toggle="tab" data-type="order" href="#orders-notifications" role="tab" id="orders-tab">{l s='Latest orders' d='Admin.Navigation.Header'}<span id="orders_notif_value"></span></a>
+                      <a class="nav-link" data-toggle="tab" data-type="order" href="#orders-notifications" role="tab" id="orders-tab">{l s='Latest orders' d='Admin.Navigation.Header'}<span class="notif-counter"></span></a>
```
matks: Hi @davidglezz I agree with you that classes are better than IDs however I'm afraid we cannot remove these identifiers.
If a module uses selectors to target these identifiers, this will break this module behavior, so this would be a BC break.

We are not against BC breaks but they must serve a specific purpose (like fixing a bug). Introducing a BC break that might be destructive for some modules relying on it does not look like a purpose high enough.

However if you can provide valid reasons for removing these identifiers 🤔 maybe they can be removed. But without valid reasons, we have to keep them at least until next major version of PrestaShop 😞

NeOMakinG: Agree with matks, if these changes are trivial, we could accept it and tag it as a BC break, as well as document it inside the next release, but if not, then it's impossible on a minor release :/

NeOMakinG: By the way, you can still add classes and use them inside you'r JS, while you can keep Ids only for retrocompatibility

davidglezz: Agree with both.
I have kept the ids so as not a BC break.
The ids were only used by this js, but could be used by a third-party module.

If BC will be break, I think that is preferable a own PR only for it.

js/admin/notifications.js
```diff
@@ -22,151 +22,120 @@
...
-	if (youEditFieldFor)
-		hints.html(hints.html() + '<br /><span class="red">' + youEditFieldFor + '</span>');
-
-	var html = "";
-	var nb_notifs = 0;
-	var wrapper_id = "";
-	var type = new Array();
+$(document).ready(function () {
+  if (youEditFieldFor) {
+    $('.translatable span.hint').append('<br /><span class="red">' + youEditFieldFor + '</span>')
```
Progi1984:
```suggestion
    $('.translatable span.hint').append(`<br /><span class="red">${youEditFieldFor}</span>`)
```

js/admin/notifications.js
```diff
@@ -22,151 +22,120 @@
...
+  var html = "<a class='notif' href='" + baseAdminDir + "index.php?tab=AdminOrders&token=" + token_admin_orders;
+  html += "&vieworder&id_order=" + value.id_order + "'>";
+  html += "#" + value.id_order + " - ";
+  html += from_msg + "&nbsp;<strong>" + value.customer_name + "</strong>";
+  html += " (" + value.iso_code + ")";
+  html += "<strong class='pull-right'>" + value.total_paid + "</strong>";
+  if (value.carrier !== "") {
+    html += " - " + value.carrier;
+  }
+  return html + "</a>";
```
Progi1984:
```suggestion
  return `<a class='notif' href='${baseAdminDir}index.php?tab=AdminOrders&token=${token_admin_orders}`
      + `&vieworder&id_order=${value.id_order}'>`
      + `#${value.id_order} - ${from_msg}"&nbsp;`
      + `<strong>${value.customer_name}</strong>`
      + ` (${value.iso_code})`
      + `<strong class='pull-right'>${value.total_paid}</strong>`;
      + (value.carrier !== "" ? ` - ${value.carrier}` : "")
      + "</a>";
```

Progi1984: Could you apply this templating in all above functions ?

Progi1984 (commented):
Some comments

js/admin/notifications.js
davidglezz: I prefer to use template-strings but I think this file is not transpiled, it cannot be used if we want to maintain compatibility with some old browsers.

https://caniuse.com/?search=template%20strings

davidglezz: Where can I check the browser versions compatibility table to see in the development of the back office and the front for each Prestashop version?

IE 11 is no longer necessary right?

What about android 4.4 and earlier?

js/admin/notifications.js
davidglezz: If I could use template strings, could I remove the +?

davidglezz:
I just realized that it is slightly different if the page is legacy or modern.
image
image

js/admin/notifications.js
```diff
@@ -22,151 +22,120 @@
...
+    cache: false,
+    dataType: 'json',
+    success: function (json) {
+      setTimeout(getPush, 120000);
+      if (!json) {
+        return;
+      }
+      // Set moment language
+      moment.lang(full_language_code);
+
```
Progi1984: Remove it, it's unuseful.

Progi1984 (commented):
A last comment. And I approve. Good job @davidglezz

Progi1984 (approved):

matthieu-rolland (approved):

PierreRambaud (approved):

SD1982:
LGTM the notifications seems to work correctly as before 
Thanks @davidglezz !!

PR #21301, opened 2020-10-05
Add commands for Specific Price priorities handling

### LINKED ISSUE #19674: [Pricing tab] - Priority management 4/4
As a merchant, I want to be able to manage the prices of my products.

**Priorities** - Option selector 
2 options:
* Default (Displays the configuration settings from the Product Setting page)
* Specific to this product

When Specific to this product is selected
THEN the user can select the order of the condition and displays 4 dropdowns with the 4 options: Shop, Currency, Country, Group. 

By default, “Group” is selected for the first dropdown, “Currency” for the second, “Country” for the third, and “Shop” for the last.

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | introduce Domain/SpecificPrice/PriorityList Value object. Add command SetSpecificPricePriorityForProduct & SetGlobalSpecificPricePriority commands and handlers.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #19674

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21301)

### DIFF
--- a/src/Core/Domain/SpecificPrice/ValueObject/PriorityList.php
+++ b/src/Core/Domain/SpecificPrice/ValueObject/PriorityList.php
@@ -78,7 +78,7 @@ class PriorityList
     {
         $checkedPriorities = [];
         foreach ($priorities as $priority) {
-            if (!in_array($priority, static::AVAILABLE_PRIORITIES)) {
+            if (!in_array($priority, static::AVAILABLE_PRIORITIES, true)) {
                 throw new SpecificPriceConstraintException(
                     sprintf('Invalid priority value "%s"', $priority),
                     SpecificPriceConstraintException::INVALID_PRIORITY
--- a/src/Core/Domain/SpecificPrice/ValueObject/PriorityList.php
+++ b/src/Core/Domain/SpecificPrice/ValueObject/PriorityList.php
@@ -1,0 +1,94 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\SpecificPrice\ValueObject;
+
+use PrestaShop\PrestaShop\Core\Domain\SpecificPrice\Exception\SpecificPriceConstraintException;
+
+/**
+ * Holds valid specific price priority values
+ */
+class PriorityList
+{
+    const PRIORITY_COUNTRY = 'id_country';
+    const PRIORITY_CURRENCY = 'id_currency';
+    const PRIORITY_GROUP = 'id_group';
+    const PRIORITY_SHOP = 'id_shop';
+
+    const AVAILABLE_PRIORITIES = [
+        'country' => self::PRIORITY_COUNTRY,
+        'currency' => self::PRIORITY_CURRENCY,
+        'group' => self::PRIORITY_GROUP,
+        'shop' => self::PRIORITY_SHOP,
+    ];
+
+    /**
+     * @var string[]
+     */
+    private $priorities;
+
+    /**
+     * @param $priorities
+     */
+    public function __construct(array $priorities)
+    {
+        $this->assertPriorities($priorities);
+        $this->priorities = $priorities;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getPriorities(): array
+    {
+        return $this->priorities;
+    }
+
+    /**
+     * @param string[] $priorities
+     *
+     * @throws SpecificPriceConstraintException
+     */
+    public function assertPriorities(array $priorities): void
+    {
+        $checkedPriorities = [];
+        foreach ($priorities as $priority) {
+            if (!in_array($priority, static::AVAILABLE_PRIORITIES)) {
+                throw new SpecificPriceConstraintException('Invalid priority value "%s"', $priority);
+            }
+
+            if (in_array($priority, $checkedPriorities)) {
+                throw new SpecificPriceConstraintException(
+                    'Invalid priorities. Priorities cannot duplicate.'
+                );
+            }
+
+            $checkedPriorities[] = $priority;
+        }
+    }
+}
--- a/src/Core/Domain/SpecificPrice/ValueObject/PriorityList.php
+++ b/src/Core/Domain/SpecificPrice/ValueObject/PriorityList.php
@@ -53,7 +53,7 @@ class PriorityList
     private $priorities;
 
     /**
-     * @param $priorities
+     * @param string[] $priorities
      */
     public function __construct(array $priorities)
     {
--- a/src/Core/Domain/SpecificPrice/ValueObject/PriorityList.php
+++ b/src/Core/Domain/SpecificPrice/ValueObject/PriorityList.php
@@ -88,7 +88,7 @@ class PriorityList
             if (in_array($priority, $checkedPriorities)) {
                 throw new SpecificPriceConstraintException(
                     'Invalid priorities. Priorities cannot duplicate.',
-                    SpecificPriceConstraintException::INVALID_PRIORITY
+                    SpecificPriceConstraintException::DUPLICATE_PRIORITY
                 );
             }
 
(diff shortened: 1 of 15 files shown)


src/Core/Domain/SpecificPrice/ValueObject/PriorityList.php
```diff
@@ -0,0 +1,98 @@
...
+                throw new SpecificPriceConstraintException(
+                    sprintf('Invalid priority value "%s"', $priority),
+                    SpecificPriceConstraintException::INVALID_PRIORITY
+                );
+            }
+
+            if (in_array($priority, $checkedPriorities)) {
+                throw new SpecificPriceConstraintException(
+                    'Invalid priorities. Priorities cannot duplicate.',
+                    SpecificPriceConstraintException::INVALID_PRIORITY
```
jolelievre:
```suggestion
                    SpecificPriceConstraintException::DUPLICATE_PRIORITY
```

jolelievre: I wouldn't even have thought of checking duplicate inputs, nice catch But it deserves a dedicated error code then

tests/Integration/Behaviour/Features/Context/Domain/Product/SpecificPricePrioritiesFeatureContext.php
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param string $productReference
+     * @param TableNode $prioritiesTable
+     */
+    public function setPrioritiesForSingleProduct(string $productReference, TableNode $prioritiesTable): void
+    {
+        $priorities = PrimitiveUtils::castStringArrayIntoArray($prioritiesTable->getRow(0)[0]);
+
+        try {
+            SpecificPrice::flushCache();
```
jolelievre: You need to flush cache because of `SpecificPrice::getPriority` ?
Maybe it should be performed after the command succeed, or in the `assertProductPriorities` method then Just to make sure the handler behaves correctly even with the cache is existing

tests/Integration/Behaviour/Features/Context/Domain/Product/SpecificPricePrioritiesFeatureContext.php
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * @Then product :productReference should have following specific price priorities:
+     *
+     * @param string $productReference
+     * @param TableNode $prioritiesTable
+     */
+    public function assertProductPriorities(string $productReference, TableNode $prioritiesTable): void
+    {
+        $expectedPriorities = PrimitiveUtils::castStringArrayIntoArray($prioritiesTable->getRow(0)[0]);
```
jolelievre: The formatting and access of behat values i a bit weird, especially the way it's written with a formatted array inside a table..

Maybe it would be more readable this way (to keep the one-line format):

```
    When I set following specific price priorities for product "product1":
      | id_country | id_currency | id_group | id_shop |
    Then product "product1" should have following specific price priorities:
      | id_country | id_currency | id_group | id_shop |
```

Which should be accessible with `$prioritiesTable->getRow(0)` I think 🤔  (or close)

tests/Integration/Behaviour/Features/Scenario/Product/specific_price_priorities.feature
```diff
@@ -0,0 +1,77 @@
...
+  Scenario: I set specific price priorities to all products
+    Given I add product "product2" with following information:
+      | name       | en-US:golden wrist watch |
+      | is_virtual | false                    |
+    And I add product "product3" with following information:
+      | name       | en-US:silver wrist watch |
+      | is_virtual | false                    |
+    And product "product2" type should be standard
+    And product "product3" type should be standard
+    When I set following specific price priorities for all products:
```
jolelievre, line 33: Can you check the initial priorities before updating them please? (preferably with different values) To make sure the command actually modified the priorities

tests/Integration/Behaviour/Features/Scenario/Product/specific_price_priorities.feature
```diff
@@ -0,0 +1,77 @@
...
+    And product "product3" should have following specific price priorities:
+      | [id_country, id_currency, id_group, id_shop] |
+
+  Scenario: Specific price priorities that was set to a single product
+    overrides the previously set global priorities (overrides priorities for this product only)
+    Given I set following specific price priorities for all products:
+      | [id_shop, id_currency, id_group, id_country] |
+    And product "product1" should have following specific price priorities:
+      | [id_shop, id_currency, id_group, id_country] |
+    Then product "product2" should have following specific price priorities:
```
jolelievre:
```suggestion
    And product "product2" should have following specific price priorities:
```

tests/Integration/Behaviour/Features/Scenario/Product/specific_price_priorities.feature
```diff
@@ -0,0 +1,77 @@
...
+      | [id_country, id_currency, id_group, id_shop] |
+    And product "product3" should have following specific price priorities:
+      | [id_country, id_currency, id_group, id_shop] |
+
+  Scenario: Specific price priorities that was set to a single product
+    overrides the previously set global priorities (overrides priorities for this product only)
+    Given I set following specific price priorities for all products:
+      | [id_shop, id_currency, id_group, id_country] |
+    And product "product1" should have following specific price priorities:
```
jolelievre:
```suggestion
    Then product "product1" should have following specific price priorities:
```

tests/Integration/Behaviour/Features/Scenario/Product/specific_price_priorities.feature
```diff
@@ -0,0 +1,77 @@
...
+      | [id_shop, id_currency, id_group, id_country] |
+    And product "product3" should have following specific price priorities:
+      | [id_shop, id_currency, id_group, id_country] |
+    When I set following specific price priorities for product "product1":
+      | [id_group, id_currency, id_country, id_shop] |
+    Then product "product2" should have following specific price priorities:
+      | [id_shop, id_currency, id_group, id_country] |
+    And product "product3" should have following specific price priorities:
+      | [id_shop, id_currency, id_group, id_country] |
+    But product "product1" should have following specific price priorities:
```
jolelievre, line 62: Oh I didn't know we could use `But` !! That's nice!

tests/Integration/Behaviour/Features/Scenario/Product/specific_price_priorities.feature
```diff
@@ -0,0 +1,77 @@
...
+
+  Scenario: Specific price priorities that was set to all products
+    overrides the previously set priorities for a single product (overrides priorities for all products)
+    Given I set following specific price priorities for product "product1":
+      | [id_shop, id_currency, id_group, id_country] |
+    And I set following specific price priorities for product "product2":
+      | [id_currency, id_shop, id_group, id_country] |
+    And I set following specific price priorities for product "product3":
+      | [id_country, id_shop, id_group, id_currency] |
+    And product "product1" should have following specific price priorities:
```
jolelievre:
```suggestion
    Then product "product1" should have following specific price priorities:
```

tests/Unit/Core/Domain/SpecificPrice/ValueObject/PriorityListTest.php
```diff
@@ -0,0 +1,89 @@
...
+use PrestaShop\PrestaShop\Core\Domain\SpecificPrice\ValueObject\PriorityList;
+
+class PriorityListTest extends TestCase
+{
+    /**
+     * @dataProvider getValidPriorities
+     *
+     * @param string[] $priorities
+     */
+    public function test_creates_object_with_valid_priorities(array $priorities): void
```
jolelievre: please use camel case for test function names
```suggestion
    public function testCreatesObjectWithValidPriorities(array $priorities): void
```

src/Core/Domain/SpecificPrice/ValueObject/PriorityList.php
zuk3975: Didn't think it would bring lots of value, in the end it is still just invalid Priority list. But :ok:

tests/Integration/Behaviour/Features/Context/Domain/Product/SpecificPricePrioritiesFeatureContext.php
zuk3975: yep that should be better

tests/Integration/Behaviour/Features/Scenario/Product/specific_price_priorities.feature
zuk3975: `And` just continues to state what was`Given`. I would only use `Then` after `When` to clearly state the main action of the scenario. Example of more human language would be:
 
"this was given **and** that was given **and** that was given...",
"When I do this **and** I do that **and** that...", 
"Then product1 looks like this **and** product2 like that **and** product3 like that..."

At least I see it that way :thinking: 
Anyway this line was modified to first assert all products priorities before taking any action, so it should not be relevant anymore

tests/Unit/Core/Domain/SpecificPrice/ValueObject/PriorityListTest.php
zuk3975: I've seen it being used in some place for increased readability,
but apparently It wasn't prestashop. I've confused it with some other project, sorry :face_with_head_bandage:

tests/Integration/Behaviour/Features/Scenario/Product/specific_price_priorities.feature
jolelievre: Yep my bad I didn't realize it was part of the initial assertions

jolelievre: Oh but I think the problem is that you use `Given` for an action statement (set the global priorities) But since it's an action `When` would be more appropriate (even if it's not the main part of your scenario)

I'd reserve `Given` for assertions like `Given this country is in database` or `Given the global properties are currently...` (to check the default value)

jolelievre: I didn't read your last modifications before commenting here, indeed now that it's all initial assertions the use of `Given` makes more sense

jolelievre (approved):

matks (approved):

matks:
Thank you @zuk3975

PR #21326, opened 2020-10-06
Set order currency to  added cart rule

### LINKED ISSUE #21323: BO - Order page - Multicurrencies - Incorrect discount value & amount
#### Describe the bug

In the BO > Order detail page, when adding a discount amount with a different currency than the default one, an incorrect value is displayed and an incorrect amount calculated
image

#### Expected behavior

PS1768
image

#### Steps to Reproduce

Steps to reproduce the behavior: The default currency is Euro

1. Go to BO > International > Upload United states pack
2. Change the currency "US Dollar" **Exchange rate = 10**, **Decimals =2**
3. Create a FO > Create a new order with currency `$` selected
4. Go to the BO => Orders > Order details > add a voucher 10 $
5. see the discount is incorrect 

#### Additional information

* PrestaShop version: 1770rc1
* PHP version: 7.2

sowbiba:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Fix order calculations when adding a cart rule with secondary currency
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #21323
| How to test?  | Go to BO > International > Currencies and activate USD<br>Change the currency "US Dollar" Exchange rate = 10, Decimals =2<br>Create a FO > Create a new order with currency $ selected<br>Go to the BO => Orders > Order details > add a voucher 10 $<br>The discount and the order amounts must be correct<br><br><br>**EDIT :** I think now you can do all thinkable tests about order in secondary currency :<br> - cart rules : amount, percentage, free shipping, directly on order page or pre-saved discounts, applicable to any order, a specific product ...<br> - specific prices : defined on product page, defined when adding a product on order page

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21326)

### DIFF
--- a/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
@@ -30,8 +30,10 @@ use Cart;
 use CartRule;
 use Configuration;
 use Currency;
+use Customer;
 use Order;
 use OrderInvoice;
+use PrestaShop\PrestaShop\Adapter\ContextStateManager;
 use PrestaShop\PrestaShop\Adapter\Order\AbstractOrderHandler;
 use PrestaShop\PrestaShop\Adapter\Order\OrderAmountUpdater;
 use PrestaShop\PrestaShop\Core\Domain\CartRule\Exception\InvalidCartRuleDiscountValueException;
@@ -52,13 +54,18 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
      * @var OrderAmountUpdater
      */
     private $orderAmountUpdater;
+    /**
+     * @var ContextStateManager
+     */
+    private $contextStateManager;
 
     /**
      * @param OrderAmountUpdater $orderAmountUpdater
      */
-    public function __construct(OrderAmountUpdater $orderAmountUpdater)
+    public function __construct(OrderAmountUpdater $orderAmountUpdater, ContextStateManager $contextStateManager)
     {
         $this->orderAmountUpdater = $orderAmountUpdater;
+        $this->contextStateManager = $contextStateManager;
     }
 
     /**
@@ -68,6 +75,10 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
     {
         $this->assertPercentCartRule($command);
         $order = $this->getOrder($command->getOrderId());
+
+        $this->contextStateManager
+            ->setCurrency(new Currency($order->id_currency))
+            ->setCustomer(new Customer($order->id_customer));
 
         $computingPrecision = new ComputingPrecision();
         $currency = new Currency((int) $order->id_currency);
--- a/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
@@ -95,6 +95,7 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
         $cartRuleObj->quantity_per_user = 1;
         $cartRuleObj->active = 0;
         $cartRuleObj->highlight = 0;
+        $cartRuleObj->reduction_currency = (int) $order->id_currency;
 
         if ($command->getCartRuleType() === OrderDiscountType::DISCOUNT_PERCENT) {
             $cartRuleObj->reduction_percent = (float) (string) $command->getDiscountValue();
--- a/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
@@ -61,6 +61,7 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
 
     /**
      * @param OrderAmountUpdater $orderAmountUpdater
+     * @param ContextStateManager $contextStateManager
      */
     public function __construct(OrderAmountUpdater $orderAmountUpdater, ContextStateManager $contextStateManager)
     {
@@ -80,6 +81,26 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
             ->setCurrency(new Currency($order->id_currency))
             ->setCustomer(new Customer($order->id_customer));
 
+        try {
+            $this->addCartRuleAndUpdateOrder($command, $order);
+        } finally {
+            $this->contextStateManager->restoreContext();
+        }
+    }
+
+    /**
+     * @param AddCartRuleToOrderCommand $command
+     * @param Order $order
+     *
+     * @return void
+     *
+     * @throws InvalidCartRuleDiscountValueException
+     * @throws OrderException
+     * @throws PrestaShopException
+     * @throws \PrestaShopDatabaseException
+     */
+    private function addCartRuleAndUpdateOrder(AddCartRuleToOrderCommand $command, Order $order): void
+    {
         $computingPrecision = new ComputingPrecision();
         $currency = new Currency((int) $order->id_currency);
         $precision = $computingPrecision->getPrecision($currency->precision);
--- a/src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
@@ -29,6 +29,7 @@ namespace PrestaShop\PrestaShop\Adapter\Order\CommandHandler;
 use Cart;
 use Currency;
 use Customer;
+use Order;
 use OrderCartRule;
 use PrestaShop\PrestaShop\Adapter\ContextStateManager;
 use PrestaShop\PrestaShop\Adapter\Order\AbstractOrderHandler;
@@ -54,6 +55,7 @@ final class DeleteCartRuleFromOrderHandler extends AbstractOrderHandler implemen
 
     /**
      * @param OrderAmountUpdater $orderAmountUpdater
+     * @param ContextStateManager $contextStateManager
      */
     public function __construct(OrderAmountUpdater $orderAmountUpdater, ContextStateManager $contextStateManager)
     {
@@ -81,10 +83,14 @@ final class DeleteCartRuleFromOrderHandler extends AbstractOrderHandler implemen
             ->setCurrency(new Currency($order->id_currency))
             ->setCustomer(new Customer($order->id_customer));
 
-        // Delete Order Cart Rule and update Order
-        $orderCartRule->softDelete();
-        $cart->removeCartRule($orderCartRule->id_cart_rule);
+        try {
+            // Delete Order Cart Rule and update Order
+            $orderCartRule->softDelete();
+            $cart->removeCartRule($orderCartRule->id_cart_rule);
 
-        $this->orderAmountUpdater->update($order, $cart, $orderCartRule->id_order_invoice);
+            $this->orderAmountUpdater->update($order, $cart, $orderCartRule->id_order_invoice);
+        } finally {
+            $this->contextStateManager->restoreContext();
+        }
     }
 }
--- a/src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
+++ b/src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
@@ -27,7 +27,10 @@
 namespace PrestaShop\PrestaShop\Adapter\Order\CommandHandler;
 
 use Cart;
+use Currency;
+use Customer;
 use OrderCartRule;
+use PrestaShop\PrestaShop\Adapter\ContextStateManager;
 use PrestaShop\PrestaShop\Adapter\Order\AbstractOrderHandler;
 use PrestaShop\PrestaShop\Adapter\Order\OrderAmountUpdater;
 use PrestaShop\PrestaShop\Core\Domain\Order\Command\DeleteCartRuleFromOrderCommand;
@@ -44,13 +47,18 @@ final class DeleteCartRuleFromOrderHandler extends AbstractOrderHandler implemen
      * @var OrderAmountUpdater
      */
     private $orderAmountUpdater;
+    /**
+     * @var ContextStateManager
+     */
+    private $contextStateManager;
 
     /**
      * @param OrderAmountUpdater $orderAmountUpdater
      */
-    public function __construct(OrderAmountUpdater $orderAmountUpdater)
+    public function __construct(OrderAmountUpdater $orderAmountUpdater, ContextStateManager $contextStateManager)
     {
         $this->orderAmountUpdater = $orderAmountUpdater;
+        $this->contextStateManager = $contextStateManager;
     }
 
     /**
@@ -68,6 +76,10 @@ final class DeleteCartRuleFromOrderHandler extends AbstractOrderHandler implemen
         if (!Validate::isLoadedObject($cart) || $order->id_cart != $cart->id) {
             throw new OrderException('Invalid cart provided.');
         }
+
+        $this->contextStateManager
+            ->setCurrency(new Currency($order->id_currency))
+            ->setCustomer(new Customer($order->id_customer));
 
         // Delete Order Cart Rule and update Order
         $orderCartRule->softDelete();
--- a/src/Adapter/Order/OrderAmountUpdater.php
+++ b/src/Adapter/Order/OrderAmountUpdater.php
@@ -232,21 +232,6 @@ class OrderAmountUpdater
             $unitPriceTaxExcl = (float) $cartProduct['price_with_reduction_without_tax'];
             $unitPriceTaxIncl = (float) $cartProduct['price_with_reduction'];
 
-            $defaultCurrency = Currency::getDefaultCurrency();
-            if (false !== $defaultCurrency && (int) $defaultCurrency->id !== (int) $order->id_currency) {
-                $orderCurrency = new Currency((int) $order->id_currency);
-                $unitPriceTaxIncl = Tools::convertPriceFull(
-                  $unitPriceTaxIncl,
-                  $defaultCurrency,
-                  $orderCurrency
-                );
-                $unitPriceTaxExcl = Tools::convertPriceFull(
-                    $unitPriceTaxExcl,
-                    $defaultCurrency,
-                    $orderCurrency
-                );
-            }
-
             $orderDetail->product_price = (float) $cartProduct['price'];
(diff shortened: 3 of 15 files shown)


jolelievre (changes requested):
I'm fine with the modification and I trust it is enough to fix the bug.
However we can't push this without additional Behat tests that validate this scenario (and potentially test a whole order edition with another currency)

tests/Integration/Behaviour/Features/Context/CommonFeatureContext.php
```diff
@@ -90,6 +90,20 @@ public static function rebootKernelPrepareFeature()
...
     }
 
+    /**
+     * This hook can be used to flag a feature for kernel reboot, this is useful
+     * to force recreation of services (e.g: when you add some currencies in the
+     * database, you may need to reset the CLDR related services to use the new ones)
+     *
+     * @BeforeScenario  @reboot-kernel-before-scenario
+     */
+    public static function rebootKernelPrepareScenario()
```
jolelievre: I guess the `reboot-kernel-before-feature` was not enough? In previous feature as long as the currencies were created in the background it worked

tests/Integration/Behaviour/Features/Scenario/Order/order_currency.feature
```diff
@@ -0,0 +1,115 @@
...
+      | total_products           | 313.00 |
+      | total_products_wt        | 331.78 |
+      | total_discounts_tax_excl | 0.00   |
+      | total_discounts_tax_incl | 0.00   |
+      | total_paid_tax_excl      | 383.00 |
+      | total_paid_tax_incl      | 405.98 |
+      | total_paid               | 405.98 |
+      | total_paid_real          | 0.0    |
+      | total_shipping_tax_excl  | 70.00  |
+      | total_shipping_tax_incl  | 74.20  |
```
jolelievre: Nice scenarios!

Missing EOL here however
And could you add some additional tests please:
- add product
- edit product
- free shipping discount
- generate invoice and check the details (with a discount)
- change Carrier to check shipping is correctly computed

I know it is not the scope of the issue but since we don't have tests in other currencies it would be safer to ensure it works correctly rather than finding out later

Most of these tests already exist and just need to be copied here with adapted values

matks:
```
--- Failed scenarios:
    Features/Scenario/Order/order_currency.feature:57
    Features/Scenario/Order/order_currency.feature:78
    Features/Scenario/Order/order_currency.feature:99
```

jolelievre (dismissed):
Nice work The tests cover pretty much everything 😉

Progi1984 (dismissed):

atomiix (dismissed):

khouloudbelguith:
Hi @sowbiba,

I found an issue testing this PR:

Steps to reproduce the issue:

1. Go to BO > International > Upload United states pack
2. Change the currency "US Dollar" Exchange rate = 10, Decimals =2
3. Create a FO > Create a new order with currency $ selected
4. Go to the BO => Orders > Order details > add a voucher 10 $
5. See error the "Base price Tax included" and the "Total Tax included" are incorrect. (they are converted to Euro)

image

Thanks to check and feedback.

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -202,7 +202,22 @@ private function updateOrderDetails(Order $order, Cart $cart, int $computingPrec
             $cartProduct = $this->getProductFromCart($cartProducts, (int) $orderDetail->product_id, (int) $orderDetail->product_attribute_id);
 
             $unitPriceTaxExcl = (float) $cartProduct['price_with_reduction_without_tax'];
-            $unitPriceTaxIncl = (float) $cartProduct['price_without_reduction'];
+            $unitPriceTaxIncl = (float) $cartProduct['price_with_reduction'];
```
sowbiba, line 234: when the product has reduction, wrong price was taken
The orderDetail was incorrect after that

jolelievre: Can you add some checks in behat tests then please? Because there are other pending PRs that modify this function so we must avoid any future regression

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
```diff
@@ -725,16 +724,6 @@ private function getOrderDiscounts(Order $order): OrderDiscountsForViewing
 
         foreach ($discounts as $discount) {
             $discountAmount = $isTaxIncluded ? $discount['value'] : $discount['value_tax_excl'];
-
-            $cartRule = new CartRule((int) $discount['id_cart_rule']);
```
sowbiba, line 729: We dont need this because the stored discount is already converted to order currency

src/Adapter/Order/AbstractOrderHandler.php
```diff
@@ -225,24 +225,26 @@ protected function updateSpecificPrice(
...
-        $specificPrice->id_product_attribute = $combination ? $combination->id : 0;
-        $specificPrice->price = (float) (string) $priceTaxExcluded;
-        $specificPrice->from_quantity = SpecificPrice::ORDER_DEFAULT_FROM_QUANTITY;
-        $specificPrice->reduction = 0;
-        $specificPrice->reduction_type = 'amount';
-        $specificPrice->reduction_tax = !$priceTaxIncluded->equals($priceTaxExcluded);
-        $specificPrice->from = SpecificPrice::ORDER_DEFAULT_DATE;
-        $specificPrice->to = SpecificPrice::ORDER_DEFAULT_DATE;
-        $specificPrice->add();
+        if (!$priceTaxExcluded->equals(new Number($product->price))) {
```
jolelievre: The check is, hypothetically, already performed before via `if ($productOriginalPrice->equals($priceTaxExcluded))`

This part is quite tricky but this SpecificPrice is supposed to be fetched with specific criterias like specific cart and customer to ensure it is an override set for this specific Order, it's not the same as the catalog SpecificPrices which are used as an alternative to CartRule

So if a discount is applied on the product via a SpecificPrice it should be reflected via the `getProductRegularPrice` function which also has `$userreduc = true` only not specific to this order

src/Adapter/Order/OrderAmountUpdater.php
jolelievre: Oh I think I understand this now, the price_with_reduction is actually the one which applies SpecificPrice so in this case it all makes sense indeed Can you add a comment for this so we don't forget newt time someone needs to modify or read this code?

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
jolelievre: It is true for discount that were manually added to the order via the BO because they are created with the matching currency However a CartRule can be define as a generic shop cart rule in which case it will probably we configured with the default shop currency which is not necessarily the Order's one

tests/Integration/Behaviour/Features/Scenario/Order/order_currency.feature
```diff
@@ -0,0 +1,219 @@
...
+      | total_discounts_tax_excl | 70.00  |
+      | total_discounts_tax_incl | 74.20  |
+      | total_paid_tax_excl      | 238.00 |
+      | total_paid_tax_incl      | 252.28 |
+      | total_paid               | 252.28 |
+      | total_paid_real          | 0.00   |
+      | total_shipping_tax_excl  | 70.00   |
+      | total_shipping_tax_incl  | 74.20   |
+
+  Scenario: Add cart rule of type 'amount' to an order with secondary currency and a product with specific price
```
jolelievre, line 112: Wow this one is a tricky scenario 😛

tests/Integration/Behaviour/Features/Scenario/Order/order_currency.feature
```diff
@@ -0,0 +1,219 @@
...
+      | total_discounts_tax_excl | 0.0    |
+      | total_discounts_tax_incl | 0.0    |
+      | total_paid_tax_excl      | 308.00 |
+      | total_paid_tax_incl      | 326.48 |
+      | total_paid               | 326.48 |
+      | total_paid_real          | 0.0    |
+      | total_shipping_tax_excl  | 70.00  |
+      | total_shipping_tax_incl  | 74.20  |
+
+  Scenario: Add cart rule of type 'amount' to an order with secondary currency
```
jolelievre, line 56: You should add tests on OrderDetail using `@Then product :productName in order :orderReference has following details:` (`checkProductDetailsWithReference`) to make sure the OrderDetail's values are consistent with the Order totals

jolelievre (changes requested):
There are a few modifications that worry me

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
sowbiba: I just tested the scenario. when the cartRule is applied on FO or when the product is added through the BO, the cartRule discount is converted when it is applied

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -202,7 +202,8 @@ private function updateOrderDetails(Order $order, Cart $cart, int $computingPrec
             $cartProduct = $this->getProductFromCart($cartProducts, (int) $orderDetail->product_id, (int) $orderDetail->product_attribute_id);
 
             $unitPriceTaxExcl = (float) $cartProduct['price_with_reduction_without_tax'];
-            $unitPriceTaxIncl = (float) $cartProduct['price_without_reduction'];
+            // this is the price with specific_price applied
```
jolelievre, line 233: top 👍

src/Adapter/Order/QueryHandler/GetOrderForViewingHandler.php
jolelievre: mmm, so this means a generic CartRule is copied for each cart created from the FO?

jolelievre: Oh I understand, thats because the initial call is `$order->getCartRules()` which actually get OrderCartRules list, these one are indeed created for each order creation which contextual values

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -69,6 +76,10 @@ public function handle(AddCartRuleToOrderCommand $command): void
         $this->assertPercentCartRule($command);
         $order = $this->getOrder($command->getOrderId());
 
+        $this->contextStateManager
```
jolelievre, line 80: You need to `restoreContext` in this handler using a `try/finally` that encapsulates everything after this call (maybe use a function to avoid the diff to become unreadable just because of the indentation

src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
```diff
@@ -69,6 +77,10 @@ public function handle(DeleteCartRuleFromOrderCommand $command)
             throw new OrderException('Invalid cart provided.');
         }
 
+        $this->contextStateManager
```
jolelievre, line 82: Same here you need a `try/finally` block and call `$this->contextStateManager->restoreContext();`, and move all the following code in a function to make review easier (or else the whole file is modified for indentation)

tests/Integration/Behaviour/Features/Scenario/Order/order_currency.feature
```diff
@@ -0,0 +1,235 @@
...
+      | from_quantity  | 1          |
+      | reduction      | 0.25       |
+      | reduction_type | percentage |
+      | reduction_tax  | 1          |
+    And there is a cart rule named "CartRuleAmountOnSelectedProduct" that applies an amount discount of 1.0 with priority 1, quantity of 100 and quantity per user 100
+    And cart rule "CartRuleAmountOnSelectedProduct" has no discount code
+    And cart rule "CartRuleAmountOnSelectedProduct" is restricted to product "Test Product With Discount and SpecificPrice"
+    When I add products to order "bo_order1" with new invoice and the following products details:
+      | name          | Test Product With Discount and SpecificPrice |
+      | amount        | 1                                     |
```
jolelievre: Can you please add 2 products instead? Just to check that the total amounts are correctly computed in OrderDetail

tests/Integration/Behaviour/Features/Scenario/Order/order_currency.feature
```diff
@@ -0,0 +1,235 @@
...
+      | total_products           | 313.00 |
+      | total_products_wt        | 331.78 |
+      | total_discounts_tax_excl | 0.00   |
+      | total_discounts_tax_incl | 0.00   |
+      | total_paid_tax_excl      | 383.00 |
+      | total_paid_tax_incl      | 405.98 |
+      | total_paid               | 405.98 |
+      | total_paid_real          | 0.0    |
+      | total_shipping_tax_excl  | 70.00  |
+      | total_shipping_tax_incl  | 74.20  |
```
jolelievre, line 177: Please add OrderDetail checks here as well

tests/Integration/Behaviour/Features/Scenario/Order/order_currency.feature
```diff
@@ -0,0 +1,235 @@
...
+      | price         | 12.00                   |
+    Then order "bo_order1" should contain 3 products "Mug The best is yet to come"
+    And product "Mug The best is yet to come" in order "bo_order1" has following details:
+      | product_quantity            | 3     |
+      | product_price               | 12.00 |
+      | unit_price_tax_incl         | 12.72 |
+      | unit_price_tax_excl         | 12.00 |
+      | total_price_tax_incl        | 38.16 |
+      | total_price_tax_excl        | 36.00 |
+    And product "Mug The best is yet to come" in order "bo_order1" should have specific price 12.00
```
jolelievre, line 198: And in this case add tests for Order totals

jolelievre (changes requested):
Only missing the restoreContext in two handlers

And some last behat assertions for safety

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -69,6 +77,30 @@ public function handle(AddCartRuleToOrderCommand $command): void
...
+            ->setCurrency(new Currency($order->id_currency))
+            ->setCustomer(new Customer($order->id_customer));
+
+        try {
+            $this->addCartRuleAndUpdateOrder($command, $order);
+        } catch (Exception $exception) {
```
jolelievre: Catching exception to throw it again is not required And it also misses a case when no exception the context is not cleared Use try/finally instead the finally block is called wether an exception is caught or not, but it doesn't block the exception

src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
```diff
@@ -69,6 +79,30 @@ public function handle(DeleteCartRuleFromOrderCommand $command)
...
+            ->setCurrency(new Currency($order->id_currency))
+            ->setCustomer(new Customer($order->id_customer));
+
+        try {
```
jolelievre, line 86: Use try/finally

jolelievre (dismissed):
Nice work @sowbiba

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -52,13 +54,18 @@ final class AddCartRuleToOrderHandler extends AbstractOrderHandler implements Ad
      * @var OrderAmountUpdater
      */
     private $orderAmountUpdater;
+    /**
+     * @var ContextStateManager
+     */
+    private $contextStateManager;
 
     /**
      * @param OrderAmountUpdater $orderAmountUpdater
```
Progi1984, line 63:
```suggestion
     * @param OrderAmountUpdater $orderAmountUpdater
     * @param ContextStateManager $contextStateManager
```

src/Adapter/Order/CommandHandler/AddCartRuleToOrderHandler.php
```diff
@@ -69,6 +76,28 @@ public function handle(AddCartRuleToOrderCommand $command): void
...
+        } finally {
+            $this->contextStateManager->restoreContext();
+        }
+    }
+
+    /**
+     * @param AddCartRuleToOrderCommand $command
+     * @param Order $order
+     *
+     * @throws InvalidCartRuleDiscountValueException
```
Progi1984:
```suggestion
     * @param Order $order
     * @return void
     *
     * @throws InvalidCartRuleDiscountValueException
```

src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
```diff
@@ -44,13 +48,18 @@ final class DeleteCartRuleFromOrderHandler extends AbstractOrderHandler implemen
...
      */
     private $orderAmountUpdater;
+    /**
+     * @var ContextStateManager
+     */
+    private $contextStateManager;
 
     /**
      * @param OrderAmountUpdater $orderAmountUpdater
      */
```
Progi1984:
```suggestion
    /**
     * @param OrderAmountUpdater $orderAmountUpdater
     * @param ContextStateManager $contextStateManager
     */
```

Progi1984 (commented):
Some feedback

Progi1984 (dismissed):

atomiix (dismissed):

jolelievre (dismissed):

jolelievre (approved):

Progi1984 (approved):

ngodefroy:
Hello @sowbiba 

What I tested:

Go to BO > International > Currencies and activate USD
Change the currency "US Dollar" Exchange rate = 10, Decimals =2
Create a FO > Create a new order with currency $ selected
Go to the BO => Orders > Order details > add a voucher 10 $
The discount and the order amounts must be correct

Then:

- Add product
- Edit product
- Free ship discount
- Discount vouchers w and without tax
- change carrier shipping
- Added extra shipping fee for this specific product

**TEST OK** for me

I wait for double validation from @khouloudbelguith and we can merge this PR.

Thank you

khouloudbelguith:
Hi @sowbiba,

It is ok in my case also :heavy_check_mark: 
https://drive.google.com/file/d/1YSXBsB7NTqKEWS81szvuPrTtMpg41Nox/view?usp=sharing
Thanks, @ngodefroy!

Thank you!

atomiix:
Thanks @sowbiba, @ngodefroy & @khouloudbelguith !

sowbiba:
Thank you all for the work ;)

PR #21329, opened 2020-10-06
Do not escape already escaped module description

### LINKED ISSUE #21298: BO - Payment Methods - Special characters are not displayed correctly
#### Describe the bug

Special characters are not displayed correctly using 1.7.7.x.
Ok using develop branch

#### Steps to Reproduce

1. Go to BO -> payment
2. Click on payment methods
3. Scroll down to the bottom of the page

**Screenshots**

Capture d’écran 2020-10-05 à 11 42 47

#### Additional information

* PrestaShop version: 1.7.7.x
* PHP version: 7.1

atomiix:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Because we tried to escape content already escaped in module suggestions brought by the module ps_mbo, special characters were not well displayed in the BO. This PR fixes it.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #21298
| How to test?  | Please see #21298

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21329)

### DIFF
--- a/src/PrestaShopBundle/Resources/views/Admin/Module/Includes/tab-module-line.html.twig
+++ b/src/PrestaShopBundle/Resources/views/Admin/Module/Includes/tab-module-line.html.twig
@@ -55,7 +55,7 @@
       </div>
       <p class="module_description">
         {% if module.attributes.description is defined and module.attributes.description is not empty %}
-          {{ module.attributes.description }}
+          {{ module.attributes.description|raw }}
         {% endif %}
       </p>
       {% if (module.attributes.message is defined and module.attributes.name is not empty) and (module.attributes.type is not defined or module.attributes.type != 'addonsMustHave' or module.attributes.type != 'addonsNative') %}


PierreRambaud (changes requested):
I don't think it's the right way to do it. It leads to possible XSS.
The problem doesn't come from the Core and even not from the MBO module: 
Modules don't have the same way to handle characters :sweat: 
image
image

atomiix:
@PierreRambaud I think it depends on whether or not the data is fetched from [the addons API](https://api.addons.prestashop.com/request/index.php?method=listing&action=native&iso_lang=fr&iso_code=fr&version=1.7.7&channel=stable&format=json). From what I saw, the description part comes from the addons API and is already escaped.

PierreRambaud:
> @PierreRambaud I think it depends on whether or not the data is fetched from [the addons API](https://api.addons.prestashop.com/request/index.php?method=listing&action=native&iso_lang=fr&iso_code=fr&version=1.7.7&channel=stable&format=json). From what I saw, the description part comes from the addons API and is already escaped.

Yes, I don't understand why they did this but the fix must be done in the MBO module, we need to keep templates safe.
Because it depends on where the data is fetched, some are not escaped whereas others are.

PR #21331, opened 2020-10-07
Enable Gift and Recycled Packaging switches

### LINKED ISSUE #21300: BO - Missing Gift and Recycled Paper in Create Order migrated page
#### Describe the bug

On Create Order migrated page, when "Gift Wrapping" and "Recycled packaging" options are ON, these options are not displayed, so it's not possible to add them to the orders.

It was displayed in 176.

Edit : When reusing, in BO > Create order, an order made from FO with recycled packaging, Gift, and Gift message, theses options are not displayed due to the bug, BUT the Gift price is added in total at the end of the order. When order is created, the two options are displayed, and so is the gift message.
So there is no "calculations" missing from the page, just the display of these 3 options (two checkboxes and one text field).

#### Expected behavior

We should have the following options displayed in 177 migrated page : 

* Checkbox for "Recycled packaging"
* Checkbox for "Gift" (which had the "Gift-wrapping price" included in the total at the end of the order)
* Gift Message

#### Steps to Reproduce

1. Go to BO > Shop Parameters > Order Settings > Gift Options > Set Gift Wrapping and Recycled Packaging to YES
2. Go to BO > Orders > Add new order > Select a customer
3. Add a product, see Shipping block : There is no Gift of recycled packaging options displayed, there should be.

**Screenshots**

On 1.7.6.8 : 
create order no gift recycled options

On 1.7.7 RC1 :
create order no gift recycled options 177

#### Additional information

* PrestaShop version: 1770rc1build1
* PHP version: N/A

matks:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Convert 'Set free shipping' capability of Create an Order BO page into 'Update Cart delivery settings' capability, that handles 3 switches. The switches 'is a gift' ans 'use recycled packaging' are only enabled when the right configuration settings are turned on. See details below
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes https://github.com/PrestaShop/PrestaShop/issues/21300
| How to test?  | See ticket

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21331)

# Advanced description

I expanded the 'set free shipping' switch logic to all 3 switches, gathered under the concept of 'update delivery settings'. I could not use 'update delivery options', it's already how the 'change carrier' dropdown is named.

So we have
- 3 switches hardcoded in HTML although the 2 new switches are only displayed when the right configuration settings are turned on
- Javascript listening to any of these 3 input change, sending an ajax request with all the input data
- On Controller side, we retrieve all the data but if configuration settings are disabled we ignore 'recycled packaging' and 'is a gift' . I did that to avoid having `if ... exists` everywhere in the JS and Controller logic
- SetFreeShippingCommand becomes UpdateCartDeliverySettingsCommand with 2 optional parameters
- handler is able to handle the 3 data state changes
- all the view data is being sent as a JSON response to hydrate the template

This is not flawless. I could have done better but since this is blocking build 2 RC1 I don't want to refactor deeper to avoid introducing too many changes. The parts I am not proud of are:
- the naming : delivery settings is very vague (and ... it's not even related to delivery, it's related to packaging !)
- instead of hiding the unused HTML markup when configuration is disabled I should not load it all but that would put `if ...`everywhere in the code
- some weird configuration checks at controller level, that would not be needed if the item above was done better


admin-dev/themes/new-theme/js/pages/order/create/cart-editor.js
```diff
@@ -68,12 +68,17 @@ export default class CartEditor {
...
    *
    * @param {Number} cartId
-   * @param {Boolean} value
    */
-  setFreeShipping(cartId, value) {
-    $.post(this.router.generate('admin_carts_set_free_shipping', {cartId}), {
-      freeShipping: value,
-    }).then(cartInfo => EventEmitter.emit(eventMap.cartFreeShippingSet, cartInfo))
+  updateDeliveryOptions(cartId) {
+    const freeShippingEnabled = $(createOrderMap.freeShippingSwitch)[1].checked;
```
jolelievre, line 73: `[1]` because there is another checkbox somewhere else? Or does the index start at 1?

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -65,35 +66,16 @@ public function __construct(TranslatorInterface $translator, ConfigurationInterf
...
-
-        if ($command->allowFreeShipping()) {
-            if (null === $freeShippingCartRule) {
-                $freeShippingCartRule = $this->createCartRule($cart, $backOfficeOrderCode);
-            }
-            $cart->addCartRule((int) $freeShippingCartRule->id);
-
-            return;
+        $this->handleFreeShippingOption($cart, $command);
+        if ($command->isAGift() !== null) {
```
jolelievre: Just a suggestion, but you can check `if ($command->isAGift() !== null) { return; }` inside the `handleGiftOption` function if you want your main `handle` function to be a bit cleaner
Same for `handleRecycledWrappingOption`

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -139,4 +121,84 @@ private function createCartRule(Cart $cart, string $backOfficeOrderCode): CartRu
...
+     * @param Cart $cart
+     * @param UpdateCartDeliverySettingsCommand $command
+     *
+     * @throws CartException
+     * @throws PrestaShopException
+     */
+    private function handleGiftOption(Cart $cart, UpdateCartDeliverySettingsCommand $command): void
+    {
+        $cart->gift = $command->isAGift();
+        $cart->save();
```
jolelievre: Not sure this `$cart->save();` is useful since you also perform a `$cart->update()` after
One is a doublon

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -139,4 +121,84 @@ private function createCartRule(Cart $cart, string $backOfficeOrderCode): CartRu
...
+     * @param Cart $cart
+     * @param UpdateCartDeliverySettingsCommand $command
+     *
+     * @throws CartException
+     * @throws PrestaShopException
+     */
+    private function handleRecycledWrappingOption(Cart $cart, UpdateCartDeliverySettingsCommand $command): void
+    {
+        $cart->recyclable = $command->useRecycledPackaging();
+        $cart->save();
```
jolelievre: Same here, you can probably remove this one

src/Core/Domain/Cart/Command/UpdateCartDeliverySettingsCommand.php
```diff
@@ -43,14 +40,28 @@ class SetFreeShippingToCartCommand
...
+    private $useRecycledPackaging;
+
     /**
      * @param int $cartId
      * @param bool $allowFreeShipping
+     * @param bool $isAGift
+     * @param bool $useRecycledPackaging
      */
-    public function __construct(int $cartId, bool $allowFreeShipping)
+    public function __construct(int $cartId, bool $allowFreeShipping, ?bool $isAGift, ?bool $useRecycledPackaging)
```
jolelievre:
```suggestion
    public function __construct(int $cartId, bool $allowFreeShipping, ?bool $isAGift = null, ?bool $useRecycledPackaging = null)
```

Maybe they can be optional? (even if in the controller you always set them which is a good choice, other devs won't have to for `null` parameter if they don't use it)

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/shipping.html.twig
```diff
@@ -28,7 +28,10 @@
   </h3>
   <div class="card-body">
 
-    <form class="js-shipping-form d-none">
+    <form class="js-shipping-form d-none"
+          data-recycled-packaging-enabled="{{ recycledPackagingEnabled }}"
```
jolelievre: I didn't see any use of these data attributes? Did I miss something?

admin-dev/themes/new-theme/js/pages/order/create/cart-editor.js
jolelievre: Oh ok I got it this is because of how `ps-switch` is built (quite nasty 😅 )

src/Core/Domain/Cart/Command/UpdateCartDeliverySettingsCommand.php
jolelievre: Oh and seems like that's the reason why travis tests fail, the constructor expects 4 parameters instead of the 2 provided

matks: Yeah I rushed this part 😞 obviously Travis punishes me

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/shipping.html.twig
matks: Indeed. In my 1st implementation I hid the inputs using JS but then I hid them using Twig. These are leftovers.

matks:
@jolelievre Thank you :) the feedback was very valuable I updated the PR

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -139,4 +117,89 @@ private function createCartRule(Cart $cart, string $backOfficeOrderCode): CartRu
...
+    /**
+     * @param Cart $cart
+     * @param UpdateCartDeliverySettingsCommand $command
+     *
+     * @throws CartException
+     * @throws PrestaShopException
+     */
+    private function handleGiftOption(Cart $cart, UpdateCartDeliverySettingsCommand $command): void
+    {
+        if ($command->isAGift() !== null) {
```
jolelievre: Ideally you should check if the configuration is enabled here as well (for recycle as well), since the configuration service is already injected it's not too much to add either

tests/Integration/Behaviour/Features/Context/Domain/CartFeatureContext.php
```diff
@@ -339,7 +339,7 @@ public function checkCartCarrier(string $cartReference, string $carrierReference
     public function setFreeShippingToCart(string $cartReference)
     {
         $this->getCommandBus()->handle(
-            new SetFreeShippingToCartCommand(
+            new UpdateCartDeliverySettingsCommand(
```
jolelievre, line 342: Could you add a behat test to ensure the command correctly handles these two new fields?

admin-dev/themes/new-theme/js/pages/order/create/cart-editor.js
```diff
@@ -68,12 +68,17 @@ export default class CartEditor {
...
+    const isAGiftEnabled = $(createOrderMap.isAGiftSwitch)[1].checked;
```
PierreRambaud:
```suggestion
    const isAGiftEnabled = $(createOrderMap.isAGiftSwitch).val() === '1';
```

admin-dev/themes/new-theme/js/pages/order/create/cart-editor.js
```diff
@@ -68,12 +68,17 @@ export default class CartEditor {
...
+    const useRecycledPackagingEnabled = $(createOrderMap.recycledPackagingSwitch)[1].checked;
```
PierreRambaud:
```suggestion
    const useRecycledPackagingEnabled = $(createOrderMap.recycledPackagingSwitch).val() === '1';
```

admin-dev/themes/new-theme/js/pages/order/create/create-order-map.js
```diff
@@ -129,6 +129,8 @@ export default {
   deliveryOptionSelect: '#delivery-option-select',
   totalShippingField: '.js-total-shipping-tax-inc',
   freeShippingSwitch: '.js-free-shipping-switch',
+  recycledPackagingSwitch: '.js-recycled-packaging-switch',
+  isAGiftSwitch: '.js-is-gift-switch',
```
PierreRambaud, line 134:
```suggestion
  isAGiftSwitch: '.js-is-gift-switch:checked',
```

admin-dev/themes/new-theme/js/pages/order/create/create-order-map.js
```diff
@@ -129,6 +129,8 @@ export default {
   deliveryOptionSelect: '#delivery-option-select',
   totalShippingField: '.js-total-shipping-tax-inc',
   freeShippingSwitch: '.js-free-shipping-switch',
+  recycledPackagingSwitch: '.js-recycled-packaging-switch',
```
PierreRambaud, line 132:
```suggestion
  recycledPackagingSwitch: '.js-recycled-packaging-switch:checked',
```

admin-dev/themes/new-theme/js/pages/order/create/shipping-renderer.js
```diff
@@ -84,6 +86,36 @@ export default class ShippingRenderer {
     });
   }
 
+  /**
+   * @param useRecycledPackaging
+   *
+   * @private
+   */
+  _renderRecycledPackagingSwitch(useRecycledPackaging) {
```
PierreRambaud:
```suggestion
  renderRecycledPackagingSwitch(useRecycledPackaging) {
```

admin-dev/themes/new-theme/js/pages/order/create/shipping-renderer.js
```diff
@@ -84,6 +86,36 @@ export default class ShippingRenderer {
...
+      }
+    });
+  }
+
+  /**
+   * @param isAGift
+   *
+   * @private
+   */
+  _renderGiftSwitch(isAGift) {
```
PierreRambaud:
```suggestion
  renderGiftSwitch(isAGift) {
```

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
matks: I was hesitating about it, fifty/fifty because I'm not sure it's the domain responsibility to discard part of the Command content because configuration says it's incorrect. Or maybe Command should be rejected because configuration is wrong 🤔

admin-dev/themes/new-theme/js/pages/order/create/shipping-renderer.js
matks: On 177 and in this file, we use `_` everywhere (on develop it's fixed as eslint screams at us if we do that), can I keep it here to stay coherent with the rest of the file ?

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -139,4 +117,89 @@ private function createCartRule(Cart $cart, string $backOfficeOrderCode): CartRu
...
+
+        if (null === $freeShippingCartRule) {
+            return;
+        }
+
+        $cart->removeCartRule((int) $freeShippingCartRule->id);
+
+        try {
+            if (false === $freeShippingCartRule->delete()) {
+                throw new CannotDeleteCartRuleException(sprintf('Failed deleting cart rule #%s', $freeShippingCartRule->id));
```
PierreRambaud, line 189:
```suggestion
                throw new CannotDeleteCartRuleException(sprintf('Failed deleting cart rule #%d', $freeShippingCartRule->id));
```

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -139,4 +117,89 @@ private function createCartRule(Cart $cart, string $backOfficeOrderCode): CartRu
...
+            }
+        } catch (PrestaShopException $e) {
+            throw new CartRuleException(sprintf('An error occurred when trying to delete cart rule #%s', $freeShippingCartRule->id));
```
PierreRambaud, line 192:
```suggestion
            throw new CartRuleException(sprintf('An error occurred when trying to delete cart rule #%d', $freeShippingCartRule->id));
```

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -139,4 +117,89 @@ private function createCartRule(Cart $cart, string $backOfficeOrderCode): CartRu
...
+        }
+
+        $cart->gift = $command->isAGift();
+
+        try {
+            if (false === $cart->update()) {
+                throw new CartException('Failed to update cart gift option');
+            }
+        } catch (PrestaShopException $e) {
+            throw new CartException(sprintf('An error occurred while trying to update gift option for cart with id "%s"', $cart->id));
```
PierreRambaud:
```suggestion
            throw new CartException(sprintf('An error occurred while trying to update gift option for cart with id "%d"', $cart->id));
```

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -139,4 +117,89 @@ private function createCartRule(Cart $cart, string $backOfficeOrderCode): CartRu
...
+            return;
+        }
+        $cart->recyclable = $command->useRecycledPackaging();
+
+        try {
+            if (false === $cart->update()) {
+                throw new CartException('Failed to update cart recycle wrapping option');
+            }
+        } catch (PrestaShopException $e) {
+            throw new CartException(sprintf('An error occurred while trying to update recycle wrapping option for cart with id "%s"', $cart->id));
```
PierreRambaud:
```suggestion
            throw new CartException(sprintf('An error occurred while trying to update recycle wrapping option for cart with id "%d"', $cart->id));
```

src/Core/Domain/Cart/Command/UpdateCartDeliverySettingsCommand.php
```diff
@@ -43,14 +40,32 @@ class SetFreeShippingToCartCommand
...
 
+    /**
+     * @var bool
+     */
+    private $isAGift;
+
+    /**
+     * @var bool
+     */
+    private $useRecycledPackaging;
```
PierreRambaud, line 51: I don't see the reason to use private if the class is not final.
Use protected, otherwise, no one will be able to extend it, and if the extension is forbidden, use final :)

src/Core/Domain/Cart/Command/UpdateCartDeliverySettingsCommand.php
```diff
@@ -43,14 +40,32 @@ class SetFreeShippingToCartCommand
...
+     * @param bool $isAGift
+     * @param bool $useRecycledPackaging
      */
-    public function __construct(int $cartId, bool $allowFreeShipping)
+    public function __construct(
+        int $cartId,
+        bool $allowFreeShipping,
+        ?bool $isAGift = null,
+        ?bool $useRecycledPackaging = null)
     {
```
PierreRambaud:
```suggestion
        ?bool $useRecycledPackaging = null
    ) {
```

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/shipping.html.twig
```diff
@@ -65,6 +65,38 @@
...
         </div>
       </div>
+      <div class="form-group row {% if not recycledPackagingEnabled %}d-none{% endif %}">
+        <div class="col-3">
+        <span class="float-right">
+          {{ 'Recycled packaging'|trans({}, 'Admin.Orderscustomers.Feature') }}
+        </span>
+        </div>
+        <div class="col-4">
+          <span class="ps-switch">
```
PierreRambaud, line 75: There is already a ps-switch widget available :thinking:  (bootstrap_4_layout.html.twig)

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
PierreRambaud: I agree with @matks It's not a domain requirement. It should handle what have in its context :thinking:

admin-dev/themes/new-theme/js/pages/order/create/shipping-renderer.js
PierreRambaud: It will be easier for the merge into the develop branch. We already have files with mixed style in the 1.7.7.x.

src/Core/Domain/Cart/Command/UpdateCartDeliverySettingsCommand.php
matks: To avoid making it part of the public API 🤔 I want us to be able to modify our `private` variables in the future.

I remember we discussed this with Pablo on another topic

matks: This is a Command. I'm not against starting to make all our Command properties `protected` but I think it should be a global change (maybe directed by an ADR ?) rather than just these properties.

And this PR aims to fix a bug blocking 1.7.7.0 RC1 build 2 😅 I dont think it's the right moment to start discussing such changes.

Can we move the global discussion "should all our Command properties be `protected`" for 1.7.8 😉 ?

admin-dev/themes/new-theme/js/pages/order/create/create-order-map.js
matks: In `CreateOrderPage::_listenForCartEdit` we use this selector to listen for changes. If I follow your change, it won't work fully 🤔 . Clicking on the "checked" part will trigger the change but not the unchecked part, unlike current implementation.

Or I need 2 selectors.

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/shipping.html.twig
matks: Yes it is the Form Theme 🤔 

I am not aware it can be reused inside Twig template, outside of a `form_widget` statement. Can you point me to the right documentation about that ? I tried three Symfony documentation pages but all of them mention template inheritance or Form Theme usage, neither fit this usecase.

matks:
@jolelievre @PierreRambaud For each of your feedback, I have either adressed the issue or provided a question/answer 😉 . Only remaining todo item is the Behat test that I am starting now. Feel free to review the PR in the meantime 😉

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
jolelievre: Well the configuration is part of the domain, it enables/disables feature, it changes behaviour sometimes
The handler is supposed to be the "guardian" of the domain, in this case if the feature is disabled it should either ignore it or throw an exception because it's not possible to have a Gift order if the GIft feature is disabled

But if you simply apply it without caring about the configuration it means anyone could use the command however he wants (even directly via the API) regardless of the configuration

It's like the `Return product` configuration is disabled, but I can still use the API to force some refunds without any checking

admin-dev/themes/new-theme/js/pages/order/create/cart-editor.js
```diff
@@ -68,12 +68,19 @@ export default class CartEditor {
...
+    const giftMessage = $(createOrderMap.giftMessageField).val();
+
+    $.post(this.router.generate('admin_carts_set_delivery_settings', {cartId}), {
+      freeShipping: freeShippingEnabled,
+      isAGift: isAGiftEnabled,
+      useRecycledPackaging: useRecycledPackagingEnabled,
+      giftMessage : giftMessage,
```
Progi1984:
```suggestion
      giftMessage,
```

src/Core/Domain/Cart/Command/UpdateCartDeliverySettingsCommand.php
```diff
@@ -43,14 +40,40 @@ class SetFreeShippingToCartCommand
...
+     * @var string
+     */
+    private $giftMessage;
+
     /**
      * @param int $cartId
      * @param bool $allowFreeShipping
+     * @param bool $isAGift
+     * @param bool $useRecycledPackaging
+     * @param string $giftMessage
```
Progi1984:
```suggestion
     * @param bool|null $isAGift
     * @param bool|null $useRecycledPackaging
     * @param string|null $giftMessage
```

tests/Integration/Behaviour/Features/Context/Domain/CartFeatureContext.php
```diff
@@ -339,13 +339,34 @@ public function checkCartCarrier(string $cartReference, string $carrierReference
...
                 SharedStorage::getStorage()->get($cartReference),
                 true
             )
         );
     }
 
+    /**
+     * @When I declare cart :cartReference is a gift with message :message
```
Progi1984, line 350:
```suggestion
     * @When I declare that the cart :cartReference is a gift with the message :message
```

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
```diff
@@ -722,6 +722,22 @@ public function createdOrderShouldHaveFreeShipping(string $reference)
         }
     }
 
+    /**
+     * @Then order :reference should be a gift with message :message
```
Progi1984, line 748:
```suggestion
     * @Then the order :reference should be a gift with the message :message
```

Progi1984 (commented):
Some comments

matks:
It seems I have an undefined Travis step 🤔 
```
486 scenarios (485 passed, 1 undefined)
7707 steps (7705 passed, 1 undefined, 1 skipped)
10m45.46s (202.23Mb)
Script @php -d date.timezone=UTC ./vendor/bin/behat -c tests/Integration/Behaviour/behat.yml --no-snippets --strict handling the integration-behaviour-tests event returned with error code 1
```

I missed it because I did not run Travis with `strict` in localhost

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
matks: I'm not sure this is relevant 🤔 

First, we have multiple other Behat steps where we do not use `the`: https://github.com/PrestaShop/PrestaShop/blob/develop/tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php#L377
So I suggest sticking to this style to keep the scenarii way of being written consistent 😉 

Second, I think the sentence "Order reference 17636 should be a gift with message 'I love you'" is syntaxically right. I dont think there is a need to add `the` before every word. If there was no reference, so "which order are we talking about" would be a valid question, `the` would be useful to identify the order.

For example following sentence is OK: "Core Developer Progi1984 should be very efficient with tool Prestonbot". You don't need to turn it into "The Core Developer Progi1984 should be very efficient with the tool Prestonbot" 😉

admin-dev/themes/new-theme/js/pages/order/create/create-order-map.js
PierreRambaud: In this case you need two selectors :)

src/Core/Domain/Cart/Command/UpdateCartDeliverySettingsCommand.php
PierreRambaud: It's not only the ability to change the private variables, but it's also a usage problem.
I agree to have an issue about this problem :)

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
matks: Please have a [octogone](https://www.jewanda.com/2019/03/chronique-mais-cest-quoi-loctogone-de-booba-et-kaaris-en-fait/) and then tell me who is the winner, I'll apply his feedback

src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/shipping.html.twig
PierreRambaud: If it's an element in the form, it should be in the form widget no? :thinking:

matks: This part of the page is not a form 😞 . The "Create an Order" page is a hybrid page where the for is slowly being built in ajax following user inputs. First you select a customer, which loads the "add product" or "reuse a cart/order" blocks, then you create a cart which unlocks 'shipping', 'voucher' blocks ... until finally 'summary' block is displayed.

So the part of the code in this PR is HTML parts being hidden at first, then hydrated with cart data by JS and loaded when user reaches this part of the workflow.

Now that we look at it 🤔 we could have done better ( @zuk3975 was talking about it yesterday) , we have been influenced by how the legacy page was built and we somehow followed the same structure.

To make it short: this part is not a regular form, which is why it's not rendered by a `form_widget`.

matks: In this PR I followed how the initial block was built (see https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Resources/views/Admin/Sell/Order/Order/Blocks/Create/shipping.html.twig#L59)

admin-dev/themes/new-theme/js/pages/order/create/create-order-map.js
matks: Fixes using 2 selectors

Progi1984 (dismissed):

tests/Integration/Behaviour/Features/Context/Domain/OrderFeatureContext.php
jolelievre: I agree but we try to improve them bit by bit, it's not blocking but each time @Progi1984 was arguing I prefered not to discuss it as it takes more time and energy than making the change 😛

admin-dev/themes/new-theme/js/pages/order/create/create-order-page.js
```diff
@@ -205,7 +205,19 @@ export default class CreateOrderPage {
...
+
+    this.$container.on('change', createOrderMap.recycledPackagingSwitch, e =>
+      this.cartEditor.updateDeliveryOptions(this.cartId),
+    );
+
+    this.$container.on('change', createOrderMap.isAGiftSwitch, e =>
+      this.cartEditor.updateDeliveryOptions(this.cartId),
+    );
+
+    this.$container.on('blur', createOrderMap.giftMessageField, e =>
```
jolelievre, line 219: So you submit the message when you get out of focus of the text field right?
Only case where it could be a problem is if the merchant clicks on the submit button right after editing its message Maybe the ajax request won't be finished before leaving the page 🤷‍♂️

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -65,36 +67,18 @@ public function __construct(TranslatorInterface $translator, ConfigurationInterf
...
-                $freeShippingCartRule = $this->createCartRule($cart, $backOfficeOrderCode);
-            }
-            $cart->addCartRule((int) $freeShippingCartRule->id);
-
-            return;
-        }
-
-        if (null === $freeShippingCartRule) {
-            return;
+        if (!\Validate::isMessage($command->getGiftMessage())) {
```
jolelievre: can you add `use Validate;` please?

jolelievre: and this should probably be handled in `handleGiftMessageOption` or it could fail if `getGiftMessage` is null

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -139,4 +123,112 @@ private function createCartRule(Cart $cart, string $backOfficeOrderCode): CartRu
...
+    private function handleGiftOption(Cart $cart, UpdateCartDeliverySettingsCommand $command): void
+    {
+        if ($command->isAGift() === null) {
+            return;
+        }
+
+        $cart->gift = $command->isAGift();
+
+        try {
+            if (false === $cart->update()) {
```
jolelievre: There is a small issue, you can perform up to three calls to `$cart->update()` if three options are provided
The update should probably be handled in the `handle` method after those sub calls, either by checking if one of the three options is not null or by returning a boolean in these functions when an update is required

jolelievre (commented):
A few small things to fix and it's perfect

admin-dev/themes/new-theme/js/pages/order/create/create-order-page.js
matks: I'm listening for suggestions 😉 however keep in mind nobody noticed this was missing since 9 months in 177 Create an Order page so this extra level of being careful is maybe not needed for a feature so unused 😅

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
matks: I dont want to do this later because I dont want to start updating the cart then halfway I find an issue and I stop the process.

I want all the validation to happen at the very first, then if all the data is valid I proceed to cart update.

I will fix this by only performing validation if gift message is not null

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/21331/commits/2d6d2a7b4cfda0023ea2cef41026d1bf9c55ed68

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
matks: Fixed in https://github.com/PrestaShop/PrestaShop/pull/21331/commits/2d6d2a7b4cfda0023ea2cef41026d1bf9c55ed68

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -125,110 +150,99 @@ private function createCartRule(Cart $cart, string $backOfficeOrderCode): CartRu
...
+            return true;
         }
 
         if (null === $freeShippingCartRule) {
-            return;
+            // Step 2
+            return false;
         }
 
+        // Step 3
```
jolelievre: Be careful, you confused this one with the three others The three others required update the cart
But this one performed a delete on the cart rule, so it should be handled independently without managing the return as it doesn't require the cart update

jolelievre (changes requested):
Small mistake

admin-dev/themes/new-theme/js/pages/order/create/create-order-page.js
jolelievre: Oh yeah I totally agree, and didn't expect you to improve this 🙂
I just noticed the potential issue, so if someone opens a bug we will be able to find it quickly 😛

src/Adapter/Cart/CommandHandler/UpdateCartDeliverySettingsHandler.php
```diff
@@ -139,4 +147,97 @@ private function createCartRule(Cart $cart, string $backOfficeOrderCode): CartRu
...
+            return;
+        }
+
+        if (null === $freeShippingCartRule) {
+            return;
+        }
+
+        $cart->removeCartRule((int) $freeShippingCartRule->id);
+
+        return;
```
jolelievre:
```suggestion
        try {
            if (false === $freeShippingCartRule->delete()) {
                throw new CannotDeleteCartRuleException(sprintf('Failed deleting cart rule #%s', $freeShippingCartRule->id));
            }
        } catch (PrestaShopException $e) {
            throw new CartRuleException(sprintf('An error occurred when trying to delete cart rule #%s', $freeShippingCartRule->id));
        }
```

you removed a part of the function

jolelievre (changes requested):

jolelievre (dismissed):
Recycled packaging is not behat tested but it's fine 😜

jolelievre (commented):
> Reviewed 5 of 21 files at r1, 7 of 12 files at r3, 3 of 7 files at r5, 1 of 2 files at r6, 1 of 1 files at r7, 8 of 8 files at r9, 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21331)* status: all files reviewed, 30 unresolved discussions (waiting on @jolelievre, @matks, @PierreRambaud, @Progi1984, and @zuk3975)

Progi1984 (dismissed):

src/Core/Domain/Cart/Command/UpdateCartDeliverySettingsCommand.php
matks: See https://github.com/PrestaShop/PrestaShop/issues/21383

jolelievre (approved):

PierreRambaud (approved):

Robin-Fischer-PS:
Thanks @matks for the PR, it's QA :heavy_check_mark: !

pr21331 proof 2

matks:
Thank you for review and QA

PR #21344, opened 2020-10-07
Move product search controller that was not in the right place

jolelievre:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Move product search controller that was not in the right place So the AdminSecurity of this controller had to be mixed between Order and Product permissions which is a bad practice
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | ~
| How to test?  | Check that product search works fine in Order create and Order view (already tested by @jolelievre )

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21344)


PierreRambaud (approved):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -1578,6 +1581,45 @@ public function displayCustomizationImageAction(int $orderId, int $type, string
...
+            $currencyIsoCode = $currencyId !== null
+                ? Currency::getIsoCodeById((int) $currencyId)
+                : Currency::getIsoCodeById($defaultCurrencyId);
+            $orderId = null;
+            if ($request->query->has('order_id')) {
+                $orderId = (int) $request->query->get('order_id');
+            }
+
+            /** @var FoundProduct[] $foundProducts */
+            $foundProducts = $this->getQueryBus()->handle(new SearchProducts($searchPhrase, 10, $currencyIsoCode, $orderId));
```
Progi1984, line 1610: For 10 a constant ?

Progi1984 (approved):

SD1982:
LGTM Thanks @jolelievre !!

Progi1984:
Thanks @jolelievre & @SD1982

src/PrestaShopBundle/Resources/config/routing/admin/sell/orders/_orders.yml
```diff
@@ -12,7 +12,7 @@ _carts:
 
 _orders:
   resource: "orders.yml"
-  prefix: /orders/
```
matks, line 15: @jolelievre This change does not seem to be needed for this PR 🤔 ? Also it's a big change, can you explain ?

PR #10414, opened 2018-09-13
Link modules routes, position column and extension, GridPositionUpdater

jolelievre:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Manage _legacy_link route parameter to allow matching modules routes, add PositionColumn in Grid Component, add PositionExtension javascript, add GridPositionUpdater service which allows to update row positions
| Type?         | new feature
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #9746
| How to test?  | Please indicate how to best verify that this PR is correct.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/10414)

PR #21352, opened 2020-10-07
WIP: GridPositionMultiStoreUpdater implementation for CMS Pages

### LINKED ISSUE #18965: BO - Pages CMS - Position column should be displayed only in a specific store context when the multistore is enabled

We should have the same behavior on all BO grid in multistore when there is a position column

In Design > Pages, the position column should be displayed only in a specific store context when the multistore is enabled (same behavior as on Categories page) and I should be able to move pages only in a specific store context

kpodemski:
| Questions     | Answers
| Branch?       | develop
| Description?  | Possibility to sort entities in relation to shop table
| Type?         | improvement
| Category?     | CO
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | Fixes #18965
| How to test?  | Play with CMS pages drag&drop sort, with and without categories, with & without multi-store

## Some thoughts

1. Is it good to almost copy `DoctrinePositionUpdateHandler `? I didn't want to introduce to much changes to core for this, we have an easy way (IFs), and a bit more clear solution introduced with this PR **but** it is something to think about.
2. Implementation in `CategoryGrid` is cleaner but I think that it would be better for developers to have it (sorting capability for modern controllers) implemented in some way directly in core architecture of `Grid`... btw. sorting on Category pages in context of two different stores doesn't work 🤷
3. As there's a decision to use `final`, I wasn't able to extend base `GridPositionUpdater`, I think that at this point we cannot implement any strategy class to handle this because it would introduce unacceptable breaking change?
4. I changed interface, and at this point I wonder if we should start with splitting between multi-store/single here, as this is breaking change

## TODO

- [ ] architectural decisions
- [x] sorting fully working in BO for single store context
- [x] sorting fully working in BO for single store context and different categories
- [x] refacto `CMS::updatePosition`
- [x] refacto `CMS::cleanPositions`
- [x] refacto `CMS::getLastPosition`
- [x] refacto `CMS::getCMSPages`
- [ ] check for backward compatibility issues
- [ ] document changes in `CMS` class
- [ ] check for tests implementation, eg. for `PositionUpdater`
- [ ] E2E

## DEPRECATED METHODS

- `CMS::listCms` - no `block_cms` in PrestaShop anymore, not possible to sort pages if they're are in few stores
- `CMS::getLinks` same as above, hard to get real CMS position, data between this and proper `CMS::getPages` are different

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21352)

### DIFF
--- a/install-dev/data/db_structure.sql
+++ b/install-dev/data/db_structure.sql
@@ -393,6 +393,7 @@ CREATE TABLE `PREFIX_category_product` (
 CREATE TABLE `PREFIX_cms` (
   `id_cms` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `id_cms_category` int(10) unsigned NOT NULL,
+  `position` int(10) unsigned NOT NULL DEFAULT '0',
   `active` tinyint(1) unsigned NOT NULL DEFAULT '0',
   `indexation` tinyint(1) unsigned NOT NULL DEFAULT '1',
   PRIMARY KEY (`id_cms`)
--- a/install-dev/data/db_structure.sql
+++ b/install-dev/data/db_structure.sql
@@ -393,7 +393,6 @@ CREATE TABLE `PREFIX_category_product` (
 CREATE TABLE `PREFIX_cms` (
   `id_cms` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `id_cms_category` int(10) unsigned NOT NULL,
-  `position` int(10) unsigned NOT NULL DEFAULT '0',
   `active` tinyint(1) unsigned NOT NULL DEFAULT '0',
   `indexation` tinyint(1) unsigned NOT NULL DEFAULT '1',
   PRIMARY KEY (`id_cms`)
@@ -2329,6 +2328,7 @@ CREATE TABLE `PREFIX_address_format` (
 CREATE TABLE `PREFIX_cms_shop` (
   `id_cms` INT(11) UNSIGNED NOT NULL,
   `id_shop` INT(11) UNSIGNED NOT NULL,
+  `position` int(10) unsigned NOT NULL DEFAULT '0',
   PRIMARY KEY (`id_cms`, `id_shop`),
   KEY `id_shop` (`id_shop`)
 ) ENGINE=ENGINE_TYPE DEFAULT CHARSET=utf8mb4 COLLATION;
--- a/src/Core/Grid/Position/GridPositionMultiStoreUpdater.php
+++ b/src/Core/Grid/Position/GridPositionMultiStoreUpdater.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+namespace PrestaShop\PrestaShop\Core\Grid\Position;
+
+use PrestaShop\PrestaShop\Core\Grid\Position\UpdateHandler\PositionUpdateHandlerInterface;
+
+/**
+ * Class GridPositionMultiStoreUpdater, this class is responsible for updating the position of items
+ * of a grid using the information from a PositionUpdateInterface object but with respecting multi-store.
+ */
+final class GridPositionMultiStoreUpdater implements GridPositionUpdaterInterface
+{
+    /**
+     * @var PositionUpdateHandlerInterface
+     */
+    private $updateHandler;
+
+    /**
+     * @param PositionUpdateHandlerInterface $updateHandler
+     */
+    public function __construct(PositionUpdateHandlerInterface $updateHandler)
+    {
+        $this->updateHandler = $updateHandler;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function update(PositionUpdateInterface $positionUpdate)
+    {
+        $newPositions = $this->getNewPositions($positionUpdate);
+        $this->sortByPositionValue($newPositions);
+        $this->updateHandler->updatePositions($positionUpdate->getPositionDefinition(), $newPositions);
+    }
+
+    /**
+     * @param PositionUpdateInterface $positionUpdate
+     *
+     * @return array
+     */
+    private function getNewPositions(PositionUpdateInterface $positionUpdate)
+    {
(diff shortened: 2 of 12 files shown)


install-dev/data/db_structure.sql
```diff
@@ -393,7 +393,6 @@ CREATE TABLE `PREFIX_category_product` (
 CREATE TABLE `PREFIX_cms` (
   `id_cms` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `id_cms_category` int(10) unsigned NOT NULL,
-  `position` int(10) unsigned NOT NULL DEFAULT '0',
```
PierreRambaud: We can't delete SQL field like that, it's a huge BC, if people are using it, everything will fall in error :(

src/Core/Grid/Position/GridPositionMultiStoreUpdater.php
```diff
@@ -0,0 +1,84 @@
...
+
+        return $positions;
+    }
+
+    /**
+     * @param array $positions
+     */
+    private function sortByPositionValue(&$positions)
+    {
+        asort($positions);
```
PierreRambaud, line 83: Is there a reason to create a method to call a PHP function alone :thinking: 
Look like a little bit overkill :sweat_smile:

src/Core/Grid/Position/PositionDefinition.php
```diff
@@ -101,4 +109,15 @@ public function getParentIdField()
...
         return $this->parentIdField;
     }
+
+    /**
+     * If shop id is available we try to update position
+     * in relation to multi-store table
+     *
+     * @return int|null
+     */
+    public function getShopId()
```
PierreRambaud:
```suggestion
    public function getShopId(): ?int
```

src/Core/Grid/Position/PositionDefinitionInterface.php
```diff
@@ -61,4 +61,12 @@ public function getPositionField();
...
      */
     public function getParentIdField();
+
+    /**
+     * Id of the shop, if not null we are trying to update position
+     * in relation to multi-store context
+     *
+     * @return int|null
+     */
+    public function getShopId();
```
PierreRambaud:
```suggestion
    public function getShopId(): ?int;
```

src/Core/Grid/Position/UpdateHandler/DoctrinePositionMultiStoreUpdateHandler.php
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * @param Connection $connection
+     * @param string $dbPrefix
+     * @param int $shopId
+     */
+    public function __construct(
+        Connection $connection,
+        $dbPrefix,
+        $shopId
```
PierreRambaud:
```suggestion
        string $dbPrefix,
        int $shopId
```

src/Core/Grid/Position/UpdateHandler/DoctrinePositionMultiStoreUpdateHandler.php
```diff
@@ -0,0 +1,147 @@
...
+    ) {
+        $this->connection = $connection;
+        $this->dbPrefix = $dbPrefix;
+        $this->shopId = $shopId;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCurrentPositions(PositionDefinitionInterface $positionDefinition, $parentId = null)
```
PierreRambaud, line 75:
```suggestion
    public function getCurrentPositions(PositionDefinitionInterface $positionDefinition, int $parentId = null)
```

src/Core/Grid/Position/GridPositionMultiStoreUpdater.php
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
```
sowbiba, line 27:
```suggestion
declare(strict_types=1);

```

src/Core/Grid/Position/GridPositionMultiStoreUpdater.php
```diff
@@ -0,0 +1,84 @@
...
+        $this->sortByPositionValue($newPositions);
+        $this->updateHandler->updatePositions($positionUpdate->getPositionDefinition(), $newPositions);
+    }
+
+    /**
+     * @param PositionUpdateInterface $positionUpdate
+     *
+     * @return array
+     */
+    private function getNewPositions(PositionUpdateInterface $positionUpdate)
```
sowbiba:
```suggestion
    private function getNewPositions(PositionUpdateInterface $positionUpdate): array
```

src/Core/Grid/Position/UpdateHandler/DoctrinePositionMultiStoreUpdateHandler.php
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
```
sowbiba, line 27:
```suggestion
declare(strict_types=1);

```

sowbiba (commented):
Some comments

src/Core/Grid/Position/GridPositionMultiStoreUpdater.php
kpodemski: Well, here you have a reason:

image

https://github.com/PrestaShop/PrestaShop/pull/10414

Maybe it's a bit more readable, I left it as in original file 😅 I don't know why it's not handled inside `getNewPositions`, if we want to get new positions anyway and in handler we don't really use position but iterator value 🤷‍♂️

kpodemski:
@PrestaShop/prestashop-maintainers 

Hey y'all, I need:
- info if we can keep it as a GridPositionMultiStoreUpdater or we should switch to CQRS **or** leave it as is now and just add proper commands to use in different places?
- Grid tests are in `tests-legacy`, shall I add my tests there or move new set of tests for this implementation to `/tests`?
- can I create E2E tests or I should leave it to QA team?

Julievrz:
Is there some wording in this PR? I can see the label but I wasn't able to find any wording inside the code... :thinking:

kpodemski:
@Julievrz it seems that PrestonBot got lost because of some rebase 🤔

Julievrz:
> @Julievrz it seems that PrestonBot got lost because of some rebase 🤔

Oh ok so if there is no wording I cancel the label. Thanks 😄

src/Core/Grid/Position/GridPositionMultiStoreUpdater.php
matthieu-rolland: That's one of the "clean code" precepts, a well-named function is better than having to write a comment explaining what is happening.

It could be considered overkill, but I kinda like it too...

matthieu-rolland (dismissed):

src/Core/Grid/Position/GridPositionMultiStoreUpdater.php
PierreRambaud: It's overkill, you're not creating method each time you have to use `empty($array)` with name `checkIfArrayIsEmpty`.
I'm sad that a senior developer like him doesn't know the sort function which is a very old PHP4 function :(

matthieu-rolland: it's in between...  but you make a good point, the name of the variable `$newPositions` is self-explanatory, we could remove the comment and not use a function, either way, it is fine to me

matthieu-rolland: (this is the epitome of knit-picking though :D )

PR #21362, opened 2020-10-08
Possibility to check if we're in product quick view mode

### LINKED ISSUE #21359: Regression related to quick view and hookDisplayProductAdditionalInfo
#### Describe the bug

Because of this PR:
https://github.com/PrestaShop/PrestaShop/pull/15297

Module developers are unable to determine if product is refreshed in quick view context or not.

#### Expected behavior

We should be able to check wether module is displayed in quick view or not, previously it was possible.

#### Steps to Reproduce

Steps to reproduce the "good" behavior from 1.7.6:

1. Make sure you have `ps_sharebuttons` module installed and enabled
2. Edit `/themes/classic/modules/ps_sharebuttons/views/templates/hook/ps_sharebuttons.tpl`
3. Put at the beginning of the block: `<pre>{$smarty.post|@print_r}</pre>`
4. Go to homepage
5. Enable quick view on some product with attributes
6. See that there's an array:

```
[action] => quickview
[id_product] => 1
[id_product_attribute] => 1
```

7. change attribute, see that we still have an access to `[action] => quickview`

Of course this is wrong that `id_product_attribute` is not updated etc. and at the end #15297 is helping with that **but**, let's check this on 1.7.7 now:

1. Make sure you have `ps_sharebuttons` module installed and enabled
2. Edit `/themes/classic/modules/ps_sharebuttons/views/templates/hook/ps_sharebuttons.tpl`
3. Put at the beginning of the block: `<pre>{$smarty.post|@print_r}</pre>`
4. Go to homepage
5. Enable quick view on some product with attributes
6. See that there's an array:

```
 [action] => quickview
 [id_product] => 1
 [id_product_attribute] => 1
```

7. **Now the important part**, change attribute, new array:

```
[ajax] => 1
[action] => refresh
[quantity_wanted] => 1
```

As you can see `action` is different than in 1.7.6, because of this I can't as a module developer check from `.tpl` if module is in quick view or not, because of this some of modules I work with doesn't work.

PR to follow.

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: N/A

kpodemski:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Follow #21359 
| Type?         | bug fix 
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #21359 
| How to test?  | Edit ps_sharebutton module and try to check if you're in quick view mode using code below
| Sponsor company | impSolutions

#### In module .tpl (worth checking of we're 

```
{if $page.page_name == 'product' && Context::getContext()->controller->isQuickView()}
  we're in quick view mode for {$product.id_product} and attribute {$product.id_product_attribute}
{/if}
```

### In module .php files

```
if ($this->context->controller instanceof ProductController && $this->context->controller->isQuickView()) {
echo 'yep';
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21362)

### DIFF
--- a/controllers/front/ProductController.php
+++ b/controllers/front/ProductController.php
@@ -52,6 +52,9 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
     protected $quantity_discounts;
     protected $adminNotifications = [];
 
+    /**
+     * @var bool
+     */
     protected $isQuickView = false;
 
     public function canonicalRedirection($canonical_url = '')
--- a/controllers/front/ProductController.php
+++ b/controllers/front/ProductController.php
@@ -456,7 +456,7 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
             'product_has_combinations' => !empty($this->combinations),
             'id_product_attribute' => $product['id_product_attribute'],
             'product_title' => $product['title'],
-            'is_quick_view' => $isQuickView
+            'is_quick_view' => $isQuickView,
         ]));
     }
 
@@ -1325,8 +1325,8 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
     }
 
     /**
-     * Return information whether we are or not in quick view mode
-     * 
+     * Return information whether we are or not in quick view mode.
+     *
      * @return bool
      */
     public function isQuickView()
@@ -1335,9 +1335,7 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
     }
 
     /**
-     * Set if product is in quick view mode or not
-     * 
-     * @return void
+     * Set quick view mode.
      */
     public function setQuickViewMode($enabled = true)
     {
--- a/controllers/front/ProductController.php
+++ b/controllers/front/ProductController.php
@@ -52,6 +52,8 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
     protected $quantity_discounts;
     protected $adminNotifications = [];
 
+    protected $isQuickView = false;
+
     public function canonicalRedirection($canonical_url = '')
     {
         if (Validate::isLoadedObject($this->product)) {
@@ -394,6 +396,9 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
         $productForTemplate = $this->getTemplateVarProduct();
         ob_end_clean();
         header('Content-Type: application/json');
+
+        $this->setQuickViewMode();
+
         $this->ajaxRender(Tools::jsonEncode([
             'quickview_html' => $this->render(
                 'catalog/_partials/quickview',
@@ -410,6 +415,11 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
         $product = $this->getTemplateVarProduct();
         $minimalProductQuantity = $this->getProductMinimalQuantity($product);
         $isPreview = ('1' === Tools::getValue('preview'));
+        $isQuickView = ('1' === Tools::getValue('quickview'));
+
+        if ($isQuickView) {
+            $this->setQuickViewMode();
+        }
 
         ob_end_clean();
         header('Content-Type: application/json');
@@ -446,6 +456,7 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
             'product_has_combinations' => !empty($this->combinations),
             'id_product_attribute' => $product['id_product_attribute'],
             'product_title' => $product['title'],
+            'is_quick_view' => $isQuickView
         ]));
     }
 
@@ -1311,5 +1322,25 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
         }
 
         return false;
+    }
+
+    /**
+     * Return information whether we are or not in quick view mode
+     * 
+     * @return bool
+     */
+    public function isQuickView()
+    {
+        return $this->isQuickView;
+    }
+
+    /**
+     * Set if product is in quick view mode or not
+     * 
+     * @return void
+     */
+    public function setQuickViewMode($enabled = true)
+    {
+        $this->isQuickView = $enabled;
     }
 }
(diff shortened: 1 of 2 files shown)


themes/_core/js/product.js
```diff
@@ -154,6 +154,7 @@ function updateProduct(event, eventType, updateUrl) {
       url: updateUrl + ((updateUrl.indexOf('?') === -1) ? '?' : '&') + formSerialized + preview,
       method: 'POST',
       data: {
+        quickview: parseInt($('.modal.quickview.in').length),
```
matks: Suggestion: if we're going to introduce a `quickview` boolean both in PHP and in JS, maybe we can make it 100% unambiguous ?

This statement basically says "OK, if selector `'.modal.quickview.in'` does exist, then I know that I am in a quickview" so it's trying to find out whether or not it's a quickview using the HTML markup.

What about making it 100% clear ? I'm suggesting to put somewhere in the HTML markup a data attribute `is-quickview`. So the code would become

```suggestion
        quickview: $('my-div-with-my-informations').data('is-quickview')
```

matks (dismissed):

controllers/front/ProductController.php
```diff
@@ -1312,4 +1323,22 @@ protected function isValidCombination($productAttributeId, $productId)
 
         return false;
     }
+
+    /**
+     * Return information whether we are or not in quick view mode.
+     *
+     * @return bool
+     */
+    public function isQuickView()
```
PierreRambaud:
```suggestion
    public function isQuickView(): bool
```

controllers/front/ProductController.php
```diff
@@ -1312,4 +1323,22 @@ protected function isValidCombination($productAttributeId, $productId)
...
+    {
+        return $this->isQuickView;
+    }
+
+    /**
+     * Set quick view mode.
+     */
+    public function setQuickViewMode($enabled = true)
```
PierreRambaud:
```suggestion
    public function setQuickViewMode(bool $enabled = true)
```

controllers/front/ProductController.php
```diff
@@ -1312,4 +1323,22 @@ protected function isValidCombination($productAttributeId, $productId)
...
+     *
+     * @return bool
+     */
+    public function isQuickView()
+    {
+        return $this->isQuickView;
+    }
+
+    /**
+     * Set quick view mode.
```
PierreRambaud, line 1341:
```suggestion
     * Set quick view mode.
     *
     * @param bool $enabled
```

themes/_core/js/product.js
PierreRambaud:
```suggestion
        quickview: $('.modal.quickview.in').length,
```
parseInt is useless here

kpodemski: Hey, once again thanks for the suggestion, but as I said on Slack, I don't think that we would ever had different element with this kind of class in our shop, and thanks to checking it like this theme developers don't need to do anything to make them compatible with my solution

kpodemski:
Thanks @PierreRambaud

Sinepel (dismissed):

matks (dismissed):

controllers/front/ProductController.php
```diff
@@ -52,6 +52,8 @@ class ProductControllerCore extends ProductPresentingFrontControllerCore
     protected $quantity_discounts;
     protected $adminNotifications = [];
 
+    protected $isQuickView = false;
```
PierreRambaud, line 58:
```suggestion
    /**
     * @var bool
     */
    protected $isQuickView = false;
```

themes/_core/js/product.js
```diff
@@ -154,6 +154,7 @@ function updateProduct(event, eventType, updateUrl) {
       url: updateUrl + ((updateUrl.indexOf('?') === -1) ? '?' : '&') + formSerialized + preview,
       method: 'POST',
       data: {
+        quickview: $('.modal.quickview.in').length,
```
PierreRambaud, line 157: Has @matks said earlier, is there a quick way to put the `is_quick_view` smarty variable into the template to retrieve the data instead of hoping the theme is using the same modal? :)

kpodemski: Probably, but as it is regression and because of related PR changes my module doesn't work (and probably some other modules), I thought that it would be better to do it without it, because this would require theme authors to make proper changes too, and it's way harder to update theme than a module 🤷🏼‍♂️

Also, it would not be smart to use this set of classes in place different than this specific modal and what's more important some code in JS is related to .quickview class too :)

PierreRambaud: Oh, reviewing too much pull request didn't notice the 1.7.7.x branch :sweat_smile: 
Let's improve this for the develop branch :)

PierreRambaud (approved):

atomiix (approved):

khouloudbelguith:
Hi @kpodemski,

After a fresh install of this PR, The **action** is still refresh
I attached a screen record
https://drive.google.com/file/d/1_wa8MdEw-JbmtzqOWV5cI92TJ5rFXIoY/view?usp=sharing
image

Thanks to check it.

kpodemski:
Hello @khouloudbelguith 

Sorry, I wasn't precise enough, the point is to have quickview=1 and it's OK. We can't have same array before and after, module developers needs to change their code from action=quickview to new one for checking if we are, or not, in quickview mode.

Point of this PR was to have this possibility at all, because without it, we couldn't check for this state from module tpl or php after changing attributes.

QA is ok if:
- you can open quick view
- quickview = 1 is in modal
- you don't have quickview=1 on regular product page

If you need more informations let me know.

khouloudbelguith:
Hi @kpodemski,

Thanks for the feedback.
It is ok :heavy_check_mark: 
https://www.awesomescreenshot.com/video/1332722?key=3b55dcbb4e75458dc096bdd10640c15e

Thanks!

Progi1984:
Thanks @kpodemski & @khouloudbelguith

PR #21377, opened 2020-10-09
New hooks for ProductListingFrontController

zapalm:
| Questions     | Answers
| Branch?       | develop
| Description?  | Added new hooks for ProductListingFrontController: actionProductSearchProviderRunQueryBefore and actionProductSearchProviderRunQueryAfter. These hooks are needed so that you can change and SQL-query and a result of its execution. This will allow to developer not to override core classes, and also not to write hard code in some modules.
| Type?         | new feature
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | no
| How to test?  | You can check in any way you would normally to check a new action hook. (PS: You should be able to check with the QA module!)

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21377)


classes/controller/ProductListingFrontController.php
```diff
@@ -308,6 +312,11 @@ protected function getProductSearchVariables()
             $query
         );
 
+        Hook::exec('actionProductSearchProviderRunQueryAfter', [
+            'query'  => $query,
```
PierreRambaud:
```suggestion
            'query' => $query,
```

classes/controller/ProductListingFrontController.php
```diff
@@ -300,6 +300,10 @@ protected function getProductSearchVariables()
 
         $query->setEncodedFacets($encodedFacets);
 
+        Hook::exec('actionProductSearchProviderRunQueryBefore', [
+            'query' => $query,
```
PierreRambaud, line 304:
```suggestion
            'query' => $query,
            'provider' => $provider,
            'context' => $context,
```

PierreRambaud (dismissed):

classes/controller/ProductListingFrontController.php
zapalm: @PierreRambaud Do you have confidence that this is needed? If you are not sure, then this is over engineering.

classes/controller/ProductListingFrontController.php
zapalm: @PierreRambaud My version is easier to read, because rows are aligned to columns. May you tell me why you think your variant of alignment is better? As far as I know, PrestaShop has no an agreement on how to align such lines.

NeOMakinG (dismissed):

khouloudbelguith:
Hi @zapalm,

After a fresh install, I couldn't find the new hooks
image
I think because the new hooks are not defined in the 
`install-dev/data/xml/hook.xml`

Thanks to check and feedback.

zapalm:
@khouloudbelguith Yes indeed. Just now I found a guide ;)
https://devdocs.prestashop.com/1.7/development/components/hook/register-new-hook/
Thank you.

PierreRambaud (approved):

atomiix (approved):

khouloudbelguith:
Hi @zapalm,

It is ok :heavy_check_mark: 
**BO**
image
**FO**
image

Thanks!!

Progi1984:
Thanks @zapalm & @khouloudbelguith

PR #21266, opened 2020-10-02
Fix wrong combination image shown when clicking to zoom

atomiix:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 1.7.7.x
| Description?  | Since #19437 has been merged, the cover picture of products with combinations was used by default when zooming instead of the one clicked.
| Type?         | bug fix
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #21236
| How to test?  | Please see #21236

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21266)

PR #21390, opened 2020-10-10
Fix getting proper cover for image

### LINKED ISSUE #21389: Issue with getting proper cover image, invalid cache key
#### Describe the bug

Ok, first of all, this is really tricky error and it's kinda hard to prepare entire scenario for test.

I'm asking @jolelievre to take a look at this Issue as He was working on implementation of getting custom cover 😎 

When we have a module where we need to get list of products, and we want to make sure that cover image is displayed in relation to chosen product attribute, **and** same product is in our cart, or any other module **before** ours, our cover cannot be used.

#### Expected behavior

We need to be able to alter cover image, individually for every module/mechanism using `Product::getProductProperties`, without racing condition 👎 

#### Steps to Reproduce

Entire test scenario is related to `blockwishlist` module where I wanted to add proper cover image displayed, for chosen combination. You need to have module installed from this branch: 
https://github.com/PrestaShop/blockwishlist/tree/new-module

This is how module looks by default:

image

I wanted to make it a bit more clear for people, to look like this:
image

It's better to see chosen color, right? I've changed blockwishlist module, this is my file:
https://gist.github.com/kpodemski/7567a9ac574ff2c4a484f176e691a1ac

This is part of the code where I'm forcing `Product::getProductProperties` to display my cover instead of default one.
https://gist.github.com/kpodemski/7567a9ac574ff2c4a484f176e691a1ac#file-wishlistproductsearchprovider-php-L177

This is code where it's happening:
https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Product.php#L5569

Logic is simple, if we don't provide `cover_image_id`, PrestaShop will handle this and display cover selected by merchant (without taking care of combination images etc.)

Now, this is how wishlist product list is looking after my modification:
image

This is same product but with different combinations, now, lets add this product to cart, refresh... and...

image

Complete mess, why? Because by default cover is black, and cached version from `Product::getProductProperties` is getting only first needed result from the page, `Cart` is first, so we're looking at black cover instead of our selected one.

PR to follow 😎 

**Screenshots**

#### Additional information

* PrestaShop version: 1.7.7
* PHP version: N/A

kpodemski:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Follow detailed description in #21389 
| Type?         | bug fix
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #21389 .
| How to test?  | Follow detailed instructions from #21389 
| Sponsor company | impSolutions

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21390)

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -5054,7 +5054,10 @@ class ProductCore extends ObjectModel
                 $row['cover_image_id'] = $cover['id_image'];
             }
         }
-        $cache_key .= '-cover' . (int) $row['cover_image_id'];
+
+        if (isset($row['cover_image_id'])) {
+            $cache_key .= '-cover' . (int) $row['cover_image_id'];
+        }
 
         Hook::exec('actionGetProductPropertiesAfter', [
             'id_lang' => $id_lang,
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -5054,6 +5054,7 @@ class ProductCore extends ObjectModel
                 $row['cover_image_id'] = $cover['id_image'];
             }
         }
+        $cache_key .= '-cover' . (int) $row['cover_image_id'];
 
         Hook::exec('actionGetProductPropertiesAfter', [
             'id_lang' => $id_lang,
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -4862,6 +4862,17 @@ class ProductCore extends ObjectModel
             $cache_key .= '-pack' . $row['id_product_pack'];
         }
 
+        if (!isset($row['cover_image_id'])) {
+            $cover = static::getCover($row['id_product']);
+            if (isset($cover['id_image'])) {
+                $row['cover_image_id'] = $cover['id_image'];
+            }
+        }
+
+        if (isset($row['cover_image_id'])) {
+            $cache_key .= '-cover' . (int) $row['cover_image_id'];
+        }
+
         if (isset(self::$productPropertiesCache[$cache_key])) {
             return array_merge($row, self::$productPropertiesCache[$cache_key]);
         }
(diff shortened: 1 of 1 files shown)


PierreRambaud (changes requested):
By putting this here, you're never able to reach line 4865 :thinking:

kpodemski:
@PierreRambaud you're right, we'll miss the cache and that's it, I'll fix it in a moment ;)

kpodemski:
@PierreRambaud uhh, ok, fixed, sorry, stupid mistake, same behavior but w/o missing cache

atomiix (approved):

jolelievre (approved):
WAoooow, this one was tricky 😅

So including the cover-id into the cacheKey makes total sense, I was just worried that this piece of code was moved a bit too "early" int eh code because of potential hooks called in between But the reis no hook so it seems fine, all good for me!

However I will warn against testing thoroughly that this change doesn't break the expected behaviour that was fixed in this PR https://github.com/PrestaShop/PrestaShop/pull/19437

And one last thing I'm worried about, it means that if a module changes the `cover_image_id` field in the hook `actionGetProductPropertiesAfter` it won't match the one used in the cacheKey.. So maybe actually it's the expected behaviour because given these parameters and this cache key the module would always override the same way anyway so caching the override makes sense.. But that's a maybe I'm not 100% sure here 😅

Anyway it was not handled at all before so it's still better with this modification which is legitimate

matthieu-rolland (approved):

kpodemski:
@jolelievre 

> And one last thing I'm worried about, it means that if a module changes the cover_image_id field in the hook actionGetProductPropertiesAfter it won't match the one used in the cacheKey.. So maybe actually it's the expected behaviour because given these parameters and this cache key the module would always override the same way anyway so caching the override makes sense.. But that's a maybe I'm not 100% sure here 😅

hmmmm, in that case it should be ok to use `Before`, I think that it would be good to even recommend this way here:
https://devdocs.prestashop.com/1.7/faq/product/#product-cover

PierreRambaud (approved):

kpodemski:
Hey @khouloudbelguith 

You can use my branch with already implemented change:
https://github.com/kpodemski/blockwishlist/tree/cover-for-combinations

Make sure to:
`composer install`
`yarn` + `yarn run build` (or npm of course)

Let me know if you need any help

khouloudbelguith:
Hi @kpodemski,

Thanks for your feedback.
It is ok :heavy_check_mark: 

Check the [module](https://github.com/kpodemski/blockwishlist/tree/cover-for-combinations) => ok
image
Check this [PR](https://github.com/PrestaShop/PrestaShop/pull/19437) => ok (cover image is dispayed in the list page)
image
Check this [PR](https://github.com/PrestaShop/PrestaShop/pull/21266) => ok ( when zooming only selected images are displayed)
image


Thanks!

Progi1984:
Thanks @kpodemski & @khouloudbelguith

PR #21397, opened 2020-10-12
Introduce ProductTagUpdater & RemoveAllAssociatedProductTags & RemoveAllAssociatedAttachments command

### LINKED ISSUE #14777: [EPIC] Product page - Options tab 9/9

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Part of product page migration. Refactor Product tags handler to use TagRepository & ProductTagUpdater. Also implement RemoveAllAssociatedProductTags & RemoveAllAssociatedProductAttachments handlers to follow our new convention. Adjust behats. None of these classes have been released yet, so there is no BC break.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21397)

### DIFF
--- a/src/Adapter/Product/CommandHandler/RemoveAllAssociatedProductAttachmentsHandler.php
+++ b/src/Adapter/Product/CommandHandler/RemoveAllAssociatedProductAttachmentsHandler.php
@@ -28,19 +28,34 @@ declare(strict_types=1);
 
 namespace PrestaShop\PrestaShop\Adapter\Product\CommandHandler;
 
+use PrestaShop\PrestaShop\Adapter\Product\Update\ProductAttachmentUpdater;
 use PrestaShop\PrestaShop\Core\Domain\Product\Command\RemoveAllAssociatedProductAttachmentsCommand;
 use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\RemoveAllAssociatedProductAttachmentsHandlerInterface;
 
 /**
- * Removes all product-attachment associations for given product using legacy object model
+ * Removes all product-attachment associations for given product
  */
 class RemoveAllAssociatedProductAttachmentsHandler implements RemoveAllAssociatedProductAttachmentsHandlerInterface
 {
+    /**
+     * @var ProductAttachmentUpdater
+     */
+    private $productAttachmentUpdater;
+
+    /**
+     * @param ProductAttachmentUpdater $productAttachmentUpdater
+     */
+    public function __construct(
+        ProductAttachmentUpdater $productAttachmentUpdater
+    ) {
+        $this->productAttachmentUpdater = $productAttachmentUpdater;
+    }
+
     /**
      * {@inheritdoc}
      */
     public function handle(RemoveAllAssociatedProductAttachmentsCommand $command): void
     {
-        // TODO: Implement handle() method.
+        $this->productAttachmentUpdater->setAttachments($command->getProductId(), []);
     }
 }
--- a/src/Adapter/Product/CommandHandler/RemoveAllAssociatedProductAttachmentsHandler.php
+++ b/src/Adapter/Product/CommandHandler/RemoveAllAssociatedProductAttachmentsHandler.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Product\CommandHandler;
+
+use PrestaShop\PrestaShop\Core\Domain\Product\Command\RemoveAllAssociatedProductAttachmentsCommand;
+use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\RemoveAllAssociatedProductAttachmentsHandlerInterface;
+
+/**
+ * Removes all product-attachment associations for given product using legacy object model
+ */
+class RemoveAllAssociatedProductAttachmentsHandler implements RemoveAllAssociatedProductAttachmentsHandlerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(RemoveAllAssociatedProductAttachmentsCommand $command): void
+    {
+        // TODO: Implement handle() method.
+    }
+}
--- a/src/Adapter/Product/CommandHandler/RemoveAllAssociatedProductAttachmentsHandler.php
+++ b/src/Adapter/Product/CommandHandler/RemoveAllAssociatedProductAttachmentsHandler.php
@@ -35,7 +35,7 @@ use PrestaShop\PrestaShop\Core\Domain\Product\CommandHandler\RemoveAllAssociated
 /**
  * Removes all product-attachment associations for given product
  */
-class RemoveAllAssociatedProductAttachmentsHandler implements RemoveAllAssociatedProductAttachmentsHandlerInterface
+final class RemoveAllAssociatedProductAttachmentsHandler implements RemoveAllAssociatedProductAttachmentsHandlerInterface
 {
     /**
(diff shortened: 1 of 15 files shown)


src/Adapter/Product/Update/ProductTagUpdater.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Removes previous tags & sets new list of tags for a product.
+     *
+     * @param Product $product
+     * @param LocalizedTags[] $localizedTagsList
+     *
+     * @throws CannotUpdateProductException
+     * @throws CoreException
+     */
+    public function set(Product $product, array $localizedTagsList): void
```
matks: I am sorry I find it very misleading this function name 😢 . Setter functions are a standard, used to modify an object property. Everybody understands at first glance what `$product->setName()` means and does. But a `$productTagUpdater->set()` 🤔 I am very puzzled by it.

matks (commented):
PR is OK for me excluding one item

matks (dismissed):

src/Adapter/Product/CommandHandler/RemoveAllAssociatedProductAttachmentsHandler.php
```diff
@@ -0,0 +1,61 @@
...
+    public function __construct(
+        ProductAttachmentUpdater $productAttachmentUpdater
+    ) {
+        $this->productAttachmentUpdater = $productAttachmentUpdater;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(RemoveAllAssociatedProductAttachmentsCommand $command): void
```
jolelievre, line 57: IS it normal that the PR mixes attachment and tags? Maybe update the title of the PR to reflect this?

src/Adapter/Product/Repository/TagRepository.php
```diff
@@ -0,0 +1,113 @@
...
+use PrestaShop\PrestaShop\Core\Exception\CoreException;
+use PrestaShopException;
+use Tag;
+
+/**
+ * Accesses product Tag data source
+ */
+class TagRepository
+{
+    public function addTags(ProductId $productId, LanguageId $languageId, LocalizedTags $localizedTags): void
```
jolelievre:
```suggestion
    public function addTagsByLanguage(ProductId $productId, LanguageId $languageId, LocalizedTags $localizedTags): void
```

Or `addTagsInLang` since it only handles one Language at a time the function should reflect this behaviour, because at first I expected it to handle all languages in a loop. Also why not using only one parameter `LocalizedTags` since it also contains the `languageId` so it increases the number of parameter for no reason.

Unless we want to avoid using `LocalizedTags` in the repository to keep simpler input parameters which does make sense, but in this case the fourth parameter could be an array, the the updater is to only one to handle this `LocalizedTags` object and would call: `$this->tagRepository->addTags($productId, $localizedTags->getLanguageId(), $localizedTags-> getTags());`

src/Adapter/Product/Repository/TagRepository.php
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    /**
+     * @param ProductId $productId
+     *
+     * @throws CannotUpdateProductException
+     * @throws CoreException
+     */
+    public function deleteTags(ProductId $productId): void
```
jolelievre:
```suggestion
    public function deleteAllTags(ProductId $productId): void
```

src/Adapter/Product/Update/ProductTagUpdater.php
```diff
@@ -0,0 +1,90 @@
...
+        $productId = new ProductId((int) $product->id);
+
+        // delete all tags for product if array is empty
+        if (empty($localizedTagsList)) {
+            $this->tagRepository->deleteTags($productId);
+
+            return;
+        }
+
+        foreach ($localizedTagsList as $localizedTags) {
```
jolelievre, line 77: Maybe we should add a `PHPDoc` for `$localizedTags` type I know the array type is specified in the function doc, but it would avoid looking back to search the type and improve readability

src/Core/Domain/Product/Command/SetProductTagsCommand.php
```diff
@@ -45,7 +46,7 @@ class UpdateProductTagsCommand
     /**
      * @var LocalizedTags[]
      */
-    private $localizedTagsList;
+    private $localizedTagsList = [];
```
jolelievre: I'm a bit confused in `SetAssociatedProductAttachmentsCommand` you remove the default value, but in this one you add one 😅  So what's the purpose and/or the convention in this kind of cases?

I may be wrong but since empty value is forbidden for both and array input is mandatory in the constructor, seems like the initial value will be erased anyway

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateAttachmentFeatureContext.php
```diff
@@ -103,7 +105,14 @@ public function setAssociatedProductAttachments(string $productReference, string
             $attachmentIds[] = $this->getSharedStorage()->get($attachmentReference);
         }
 
-        $this->setProductAttachments($this->getSharedStorage()->get($productReference), $attachmentIds);
+        try {
+            $this->getCommandBus()->handle(new SetAssociatedProductAttachmentsCommand(
+                $this->getSharedStorage()->get($productReference),
+                $attachmentIds
+            ));
+        } catch (DomainException $e) {
```
jolelievre: The exception is too wide here, either use `ProductException` to limit it to the Product domain, allowing unexpected other exceptions to be thrown Because we usually only test lastException when we expect it, but never when nothing should raise, so quite often these caught exceptions just go unnoticed (until something worse bugs down the road 😅 )

But I recommended like in another PR, if there is no test related to this caught exception, we might as well not catch anything

jolelievre: This makes me think, maybe we should add a safeguard on lastException If the next behat check doesn't handle to check it, it probably means that we didn't expect this and we could throw a RuntimeException like `Some exception has been caught that did not seem to be expected since no assertions is related to it` What do you think?

zuk3975: It happens already. If some exception is thrown and is not handled by next scenario it will fail and tell something like "probably it was caused by {blabla}Exception"

src/Adapter/Product/CommandHandler/RemoveAllAssociatedProductAttachmentsHandler.php
zuk3975: The Pr wasn't big and RemoveAll*Handler shares same concept, so to not forget I just put it in one PR :smile:

src/Adapter/Product/Update/ProductTagUpdater.php
zuk3975: Im not a fan of many phpDocs, it will eventually be forgotten, outdated & misleading. I think it is enough that LocalizedTags has a typehint and it is an object, so its not too hard to see what it contains. (you don't even need to scroll up&down to see the typehint :smile: )

src/Core/Domain/Product/Command/SetProductTagsCommand.php
zuk3975: you are right, it is just a mistake, I think I put it before it had a dedicated command.
It is not needed anymore as this command cannot contain empty array anymore

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateAttachmentFeatureContext.php
zuk3975: Anyway I removed try-catch here as no exceptions are asserted in scenarios

src/Adapter/Product/Update/ProductTagUpdater.php
zuk3975: IMO its enough :man_shrugging: 
If I add it even PHPstorm underlines it suggesting that its too much as it already inherits the typehint from method docblock.

tests/Integration/Behaviour/Features/Context/Domain/Product/UpdateAttachmentFeatureContext.php
jolelievre: We have this safeguard indeed but only AfterScenario, the idea would to raise an error much sooner, like in the very next step To avoid having 4-5 steps that run in an unexpected state and start bugging, and sometimes you don't realize right away that the error happened a few steps earlier

It would also force us to correctly set the expected exception everywhere they are wanted And as you say we should improve the display as the gigantic message is not really easy to understand, I even believe sometimes it kind of "loops" into the stacktrace and display the same thing several times

Anyway it's a parallel improvement 🙂

jolelievre (approved):

matks (approved):

matks:
Thank you @zuk3975

Issue #10824, opened 2018-10-03
Product with numeric references not imported as accessories in CSV Import


jgtrescazes:
When using numeric references for products, while importing from CSV file, if in accessories column we put a list of numeric references, none of them is imported.
The problem is in **classes/Product.php,** method **getExistingIdsFromIdsOrRefs**

```
        foreach ((is_array($ids_or_refs) ? $ids_or_refs : array($ids_or_refs)) as $id_or_ref) {
            if (is_numeric($id_or_ref)) {
                $ids[] = (int)$id_or_ref;
            } elseif (is_string($id_or_ref)) {
                $refs[] = '\''.pSQL($id_or_ref).'\'';
            }
        }
```
Because of this condition when reference is a numeric value, the method use it as an ID but in our case it is not one, so none of the reference are found in database.
This conditions should rely on the "Use reference as key" import parameter or use both conditions for all values whatever their type is, like bellow:
```
        foreach ((is_array($ids_or_refs) ? $ids_or_refs : array($ids_or_refs)) as $id_or_ref) {
            $ids[] = (int)$id_or_ref;
            $refs[] = '\''.pSQL($id_or_ref).'\'';
        }
```
**To reproduce :** 
I'm using 1.7.4.2 but code is the same in the develop branch.
Create an import file with numeric references, and fill the accessories column, launch import
Expected behavior : product accessories are set
Result of the test : no accessories are set

**Severity :**
This is not critical because this method is easy to override but I think it is an issue because numeric references are valid and should be supported everywhere in the core.

khouloudbelguith:
Hi @jgtrescazes,

Could you please provide me your CSV file to test with it.
Thanks!

jgtrescazes:
Hi @khouloudbelguith 
I'm not allowed to give you the real file but you can reproduce the issue with this file : [reproduce-issue-10824.zip](https://github.com/PrestaShop/PrestaShop/files/2441675/reproduce-issue-10824.zip) (csv in zip)
**Import config:**
Separator=";"
Separator multiple values=","
**Mapping:**
Col1=Reference
Col2=Name
Col3=Accessories

After importing this none of the accessories are imported.
If you try with the fix I made in my previous comment, accessories are imported

khouloudbelguith:
Hi @jgtrescazes,

Thanks for this clarifications.
I manage to reproduce the issue described above.
In the Col3=Accessories: if we put a reference of the products as values => Accessories are not imported.
[...]

PR #21406, opened 2020-10-12
Fix/get existing ids from ids or refs

jonasburneika:
fix for https://github.com/PrestaShop/PrestaShop/issues/10824 issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21406)

### DIFF
--- a/classes/Language.php
+++ b/classes/Language.php
@@ -388,7 +388,7 @@ class LanguageCore extends ObjectModel
 
             $module_theme_files = (file_exists($tPath_from . 'modules/') ? scandir($tPath_from . 'modules/', SCANDIR_SORT_NONE) : array());
             foreach ($module_theme_files as $module) {
-                if ($module !== '.' && $module != '..' && $module !== '.svn' && file_exists($tPath_from . 'modules/' . $module . '/translations/' . (string) $iso_from . '.php')) {
+                if ($module !== '.' && $module !== 'index.php' && $module != '..' && $module !== '.svn' && file_exists($tPath_from . 'modules/' . $module . '/translations/' . (string) $iso_from . '.php')) {
                     $files_theme[$tPath_from . 'modules/' . $module . '/translations/' . (string) $iso_from . '.php'] = ($copy ? $tPath_to . 'modules/' . $module . '/translations/' . (string) $iso_to . '.php' : ++$number);
                 }
             }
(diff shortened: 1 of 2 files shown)


matks (changes requested):
Hi @jonasburneika thanks for the PR but there is an issue with it.

Even if the parameter `match_ref` is used in the Import page, we cannot fetch it like this from `Tools::getValue()` directly inside Product class. That would break the [MVC principle](https://devdocs.prestashop.com/1.7/modules/concepts/controllers/). This class belongs to the Model, the parameter `match_ref` seems to belong in the Controller.

IMO the right way to implement this behavior would be
- add a new optional argument to `getExistingIdsFromIdsOrRefs()` to modify this behavior
- make sure to pass the right value for this argument inside the Controller

See example below

matks:
Example:
```
public static function getExistingIdsFromIdsOrRefs($ids_or_refs, $forceUsingReferences = false)
```

The idea is that getExistingIdsFromIdsOrRefs can be used in many usecases where `Tools::getValue()` does not work. In a CLI context for example. So this function must be passed the right arguments through the code layer instead of getting the value directly from the request

matks:
Also please fill in the PR template https://github.com/PrestaShop/PrestaShop/blob/develop/.github/PULL_REQUEST_TEMPLATE.md 🙏 that would help us to review this PR

Julievrz:
Hi 😃 I cannot see any wording in this PR, is it normal?

atomiix:
> Hi 😃 I cannot see any wording in this PR, is it normal?

@Julievrz you're right, I don't know exactly what happened there but there's no wording in this PR. I deleted PrestonBot's comment and the *waiting for wording* label.

@jonasburneika you'll need to resolve the conflicting file. Also, if you could [squash you commits](https://www.internalpointers.com/post/squash-commits-into-one-git), it would be nice :)

matks:
Hi, [following our review guidelines](https://devdocs.prestashop.com/1.7/project/maintainers-guide/reviewing-pull-requests/) "Pull Requests may be closed after 30 days of inactivity following a request for modifications".

Since this PR did not get any activity since November 2020, I close it.

Please note this does not mean it's a final "won't merge" 😉  ; if you come back we can reopen it and update it

Thank you for the contribution

PR #21409, opened 2020-10-13
Delete unused facets tpl file

### LINKED ISSUE #14808: facets.tpl in themes/classic
Hello,

this tpl is not necessary ?

NeOMakinG:
| Questions     | Answers
| Branch?       | develop
| Description?  | Facets.tpl looks not used on classic theme
| Type?         | improvement
| Category?     | FO
| BC breaks?    | yes
| Deprecations? | no
| Fixed ticket? | Fixes #14808.
| How to test?  | Check on catalog pages if it change something, it should not as this file is not used

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21409)


PululuK:
> Facets.tpl looks not used on classic theme

Yes ! but ... @NeOMakinG  We need to be sure that no child themes use it 🤔 ?

Consider this case : **My theme child** 

```twig
  {extends file='parent:catalog/_partials/facets.tpl'}
```

Progi1984 (approved):

NeOMakinG:
> > Facets.tpl looks not used on classic theme
> 
> Yes ! but ... @NeOMakinG We need to be sure that no child themes use it 🤔 ?
> 
> Consider this case : **My theme child**
> 
> ```
>   {extends file='parent:catalog/_partials/facets.tpl'}
> ```

Damn, you should be right, it would be a BC

What do you think about just tagging it as a BC ? If developers are aware of this change, they will be able to create this file inside their theme instead of relying on the core file, as we don't use it at all

PululuK:
> Damn, you should be right, it would be a BC
> 
> What do you think about just tagging it as a BC ? If developers are aware of this change, they will be able to create this file inside their theme instead of relying on the core file, as we don't use it at all


Yep ! Add BC label please ... thanks @NeOMakinG

PierreRambaud:
I'm closing this pull request, this file is used.

PululuK:
thanks @PierreRambaud

Issue #19522, opened 2020-06-02
[Basic settings tab] - Related product 12/14


MatShir:
As a merchant I want to be able to complete the basic information of my products.

- [ ]  **Add a related product**: Search field to search a product. You can search by product name or product reference. When you start typing, if there are results, they are displayed and highlighted in the dropdown when using a keyboard and hover with mouse \([issue 14052](https://github.com/PrestaShop/PrestaShop/issues/14052)\).             
You can click on it to select it. It should not be possible to associate the same product with itself or to associate the same product several times \([issue 9960](https://github.com/PrestaShop/PrestaShop/issues/9960)\)        
You can have many associated products.

- [ ]  **Delete a related product**: The trash button is to delete all related products. When you click on it, you have a modal for cancel or confirm the action.       
If you want to delete only one associated product, you can click on the cross. When you click on it, you have a modal for cancel or confirm the action.

PR #21417, opened 2020-10-13
Introduce Command/Query handlers for related products

zuk3975:
| Questions     | Answers
| Branch?       | develop
| Description?  | Part of product page migration. Implemented following command handlers: `SetRelatedProducts, GetRelatedProducts, RemoveAllRelatedProducts`. Add some basic behat scenarios to support its behavior.
| Type?         | refacto
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | part of #19522

As discussed privately with @matks, It might be relevant to rename database table of related products in future. Currently the table is called "accessory". **EDIT**: issue created in this regard https://github.com/PrestaShop/PrestaShop/issues/21419

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21417)

### DIFF
--- a/src/Adapter/Product/QueryHandler/GetRelatedProductsHandler.php
+++ b/src/Adapter/Product/QueryHandler/GetRelatedProductsHandler.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * Copyright since 2007 PrestaShop SA and Contributors
+ * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/OSL-3.0
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://devdocs.prestashop.com/ for more information.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/OSL-3.0 Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Adapter\Product\QueryHandler;
+
+use PrestaShop\PrestaShop\Adapter\Product\Repository\ProductRepository;
+use PrestaShop\PrestaShop\Core\Domain\Language\ValueObject\LanguageId;
+use PrestaShop\PrestaShop\Core\Domain\Product\Query\GetRelatedProducts;
+use PrestaShop\PrestaShop\Core\Domain\Product\QueryHandler\GetRelatedProductsHandlerInterface;
+use PrestaShop\PrestaShop\Core\Domain\Product\QueryResult\RelatedProduct;
+
+final class GetRelatedProductsHandler implements GetRelatedProductsHandlerInterface
+{
+    /**
+     * @var ProductRepository
+     */
+    private $productRepository;
+
+    /**
+     * @var int
+     */
+    private $contextLangId;
+
+    /**
+     * @param ProductRepository $productRepository
+     * @param int $contextLangId
+     */
+    public function __construct(
+        ProductRepository $productRepository,
+        int $contextLangId
+    ) {
+        $this->productRepository = $productRepository;
+        $this->contextLangId = $contextLangId;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handle(GetRelatedProducts $query): array
+    {
+        $results = $this->productRepository->getRelatedProducts($query->getProductId(), new LanguageId($this->contextLangId));
+
+        $relatedProducts = [];
+
+        foreach ($results as $result) {
+            $relatedProducts[] = new RelatedProduct(
+                (int) $result['id_product'],
+                $result['name'],
+                $result['reference']
+            );
+        }
+
+        return $relatedProducts;
+    }
+}
--- a/src/Adapter/Product/QueryHandler/GetRelatedProductsHandler.php
+++ b/src/Adapter/Product/QueryHandler/GetRelatedProductsHandler.php
@@ -29,7 +29,6 @@ declare(strict_types=1);
 namespace PrestaShop\PrestaShop\Adapter\Product\QueryHandler;
 
 use PrestaShop\PrestaShop\Adapter\Product\Repository\ProductRepository;
-use PrestaShop\PrestaShop\Core\Domain\Language\ValueObject\LanguageId;
 use PrestaShop\PrestaShop\Core\Domain\Product\Query\GetRelatedProducts;
 use PrestaShop\PrestaShop\Core\Domain\Product\QueryHandler\GetRelatedProductsHandlerInterface;
 use PrestaShop\PrestaShop\Core\Domain\Product\QueryResult\RelatedProduct;
@@ -42,20 +41,12 @@ final class GetRelatedProductsHandler implements GetRelatedProductsHandlerInterf
     private $productRepository;
 
     /**
-     * @var int
-     */
-    private $contextLangId;
-
-    /**
      * @param ProductRepository $productRepository
-     * @param int $contextLangId
      */
     public function __construct(
-        ProductRepository $productRepository,
-        int $contextLangId
+        ProductRepository $productRepository
     ) {
         $this->productRepository = $productRepository;
-        $this->contextLangId = $contextLangId;
     }
 
     /**
@@ -63,7 +54,7 @@ final class GetRelatedProductsHandler implements GetRelatedProductsHandlerInterf
      */
     public function handle(GetRelatedProducts $query): array
     {
-        $results = $this->productRepository->getRelatedProducts($query->getProductId(), new LanguageId($this->contextLangId));
+        $results = $this->productRepository->getRelatedProducts($query->getProductId(), $query->getLanguageId());
 
         $relatedProducts = [];
 
--- a/src/Adapter/Product/Repository/ProductRepository.php
+++ b/src/Adapter/Product/Repository/ProductRepository.php
@@ -114,7 +114,11 @@ class ProductRepository extends AbstractObjectModelRepository
      * @param ProductId $productId
      * @param LanguageId $languageId
      *
-     * @return array<string, mixed>
+     * @return array<string, string>
+     *                       e.g [
+     *                       ['id_product' => '1', 'name' => 'Product name', 'reference' => 'demo15'],
+     *                       ['id_product' => '2', 'name' => 'Product name2', 'reference' => 'demo16'],
+     *                       ]
      *
      * @throws CoreException
      */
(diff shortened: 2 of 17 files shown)


matks:
See https://github.com/PrestaShop/PrestaShop/issues/21419

src/Adapter/Product/QueryHandler/GetPackedProductsHandler.php
```diff
@@ -36,7 +36,7 @@
 /**
  * Handles GetPackedProducts query using legacy object model
  */
-class GetPackedProductsHandler implements GetPackedProductsHandlerInterface
+final class GetPackedProductsHandler implements GetPackedProductsHandlerInterface
```
jolelievre, line 39: I wonder what is the criteria for us to decide if a handler is final or not? 🤷‍♂️

src/Adapter/Product/QueryHandler/GetRelatedProductsHandler.php
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * @param ProductRepository $productRepository
+     * @param int $contextLangId
+     */
+    public function __construct(
+        ProductRepository $productRepository,
+        int $contextLangId
+    ) {
+        $this->productRepository = $productRepository;
+        $this->contextLangId = $contextLangId;
```
jolelievre: I wonder if we should inject this language in the handler Or handle it via the command instead

It would be more flexible then I guess, besides when we create the API that uses this command, the controller can easily fetch the context language, so it won't force us to post the language ID via ajax

However it allows to easily adapt this command to more use cases, whereas changing the context language is no easy task

src/Adapter/Product/Repository/ProductRepository.php
```diff
@@ -64,6 +84,57 @@ public function assertProductExists(ProductId $productId): void
         $this->assertObjectModelExists($productId->getValue(), 'product', ProductNotFoundException::class);
     }
 
+    /**
+     * @param ProductId[] $productIds
+     *
+     * @throws ProductNotFoundException
+     */
+    public function assertProductsExists(array $productIds): void
```
jolelievre: I wonder if this could be implemented in the Abstract class in some ways I guess many repositories might need this methods to check if products exists as well
Although this could be done simply by looping through the array and using the existing `assertExists`

jolelievre: Also maybe we could find a less confusing function name, it's only different by a letter `s` so it can easily been mixed up using auto completion (and the pending s is actually a conjugation mistake)
Maybe we could use something like `assertProductListExists` `assertProductCollectionExists` or `assertAllProductsExist`

src/Adapter/Product/Repository/ProductRepository.php
```diff
@@ -64,6 +84,57 @@ public function assertProductExists(ProductId $productId): void
...
+        if (!$results || (int) $results['product_count'] !== count($ids)) {
+            throw new ProductNotFoundException('Some of products does not exist');
+        }
+    }
+
+    /**
+     * @param ProductId $productId
+     * @param LanguageId $languageId
+     *
+     * @return array<string, mixed>
```
jolelievre: The return format is not very clear Maybe we should add a comment for this, if I'm not mistaken it returns something like this:
```php
[
    ['id_product' => 42, 'name' => 'Product name', 'reference' => 'demo15'],
]
```

src/Adapter/Product/Update/RelatedProductsUpdater.php
```diff
@@ -0,0 +1,83 @@
...
+        ProductRepository $productRepository
+    ) {
+        $this->productRepository = $productRepository;
+    }
+
+    /**
+     * @param Product $product
+     * @param ProductId[] $relatedProductIds
+     */
+    public function set(Product $product, array $relatedProductIds): void
```
jolelievre:
```suggestion
    public function setRelatedProducts(ProductId $productId, array $relatedProductIds): void
```

Why not rely on the productId here, like it was done in `ProductSupplierUpdater::setProductSuppliers` The updater already has a dependency to the repository anyway But it would make one less in the handler

src/Adapter/Product/Update/RelatedProductsUpdater.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        try {
+            if (empty($relatedProductIds)) {
+                $product->deleteAccessories();
```
jolelievre: This is a doublon, you re-call `$product->deleteAccessories();` a few lines below, this could be simplified by doing:
- first check products exist
- then always perform deleteAccessories
- finally `if (!empty($ids)) { $product->changeAccessories($ids); }`

src/Core/Domain/Product/QueryResult/RelatedProduct.php
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace PrestaShop\PrestaShop\Core\Domain\Product\QueryResult;
+
+/**
+ * Transfers related product data
+ */
+class RelatedProduct
```
jolelievre, line 34: Quick question, does it mean this will always be enough for this query? So additional elements will either need a second query that returns details based on productId? Or would we create a new dedicated query altogether?

Or is this object potentially going to grow in the future?

tests/Integration/Behaviour/Features/Context/Domain/Product/RelatedProductsFeatureContext.php
```diff
@@ -0,0 +1,121 @@
...
+    public function assertProductHasNoRelatedProducts(string $productReference): void
+    {
+        $productId = $this->getSharedStorage()->get($productReference);
+        try {
+            $relatedProducts = $this->getQueryBus()->handle(new GetRelatedProducts($productId));
+            Assert::assertEmpty(
+                $relatedProducts,
+                sprintf('Product %s expected to have no related products', $productReference)
+            );
+        } catch (DomainException $e) {
```
jolelievre: You shouldn't catch such a wide exception If something unexpected happens because some code was modified it could hide the problem You should only catch the exception you expect to test
And if you have no error assertions to make after using this command it even should be left with a try/catch

It's better to let it throw the exception and crash so we can see right away what is the problem

tests/Integration/Behaviour/Features/Context/Domain/Product/RelatedProductsFeatureContext.php
```diff
@@ -0,0 +1,121 @@
...
+     * @When I remove all related products from product :productReference
+     *
+     * @param string $productReference
+     */
+    public function removeAllRelatedProducts(string $productReference)
+    {
+        $productId = $this->getSharedStorage()->get($productReference);
+        try {
+            $this->getCommandBus()->handle(new RemoveAllRelatedProductsCommand($productId));
+        } catch (DomainException $e) {
```
jolelievre: Same here, exception is too wide And if you don't expect to test the caught exception don't event catch it

src/Adapter/Product/QueryHandler/GetPackedProductsHandler.php
zuk3975: I remember only one clear criteria - if a class implements interface then it can be final.

src/Core/Domain/Product/QueryResult/RelatedProduct.php
zuk3975: I guess the aim is to allow this class grow if it needs to.
If we get another case when we need a different RelatedProduct, we implement a different query with a different result. Then I suppose the only problem will be the naming.

src/Adapter/Product/Update/RelatedProductsUpdater.php
zuk3975: the setRelatedProducts method naming is kinda redundant as its clearly expressed in class name, but ok, it will be clearer.

src/Adapter/Product/Update/RelatedProductsUpdater.php
zuk3975: Thats the point. When array is empty I don't need to check existence, I just delete all related products and don't go further.
Btw I just thought I might need to implement database transactions here in case changeAccessories() goes wrong after deletion. wdyt ? :thinking:

jolelievre: Ideally a transaction would be nice indeed

src/Adapter/Product/QueryHandler/GetPackedProductsHandler.php
jolelievre: Ok, but then most of our classes won't be extendable any more Not a problem for handlers I think 🤔
But for others it might be a bit harsh

src/Adapter/Product/Repository/ProductRepository.php
```diff
@@ -64,6 +84,61 @@ public function assertProductExists(ProductId $productId): void
...
+        $qb = $this->connection->createQueryBuilder();
+        $qb->select('COUNT(id_product) as product_count')
+            ->from($this->dbPrefix . 'product')
+            ->where('id_product IN (:productIds)')
+            ->setParameter('productIds', $ids, Connection::PARAM_INT_ARRAY);
+
+        $results = $qb->execute()->fetch();
+
+        if (!$results || (int) $results['product_count'] !== count($ids)) {
+            throw new ProductNotFoundException('Some of products does not exist');
```
jolelievre:
```suggestion
            throw new ProductNotFoundException(
                    sprintf(
                        'Some of these products do not exist: %s',
                        implode(',', $ids)
                    )
                );
```

src/Adapter/Product/Repository/ProductRepository.php
```diff
@@ -64,6 +84,61 @@ public function assertProductExists(ProductId $productId): void
...
+        }
+    }
+
+    /**
+     * @param ProductId $productId
+     * @param LanguageId $languageId
+     *
+     * @return array<string, string>
+     *                       e.g [
+     *                       ['id_product' => '1', 'name' => 'Product name', 'reference' => 'demo15'],
```
jolelievre, line 124: I wonder if it would be worth it to cast `id_product` If we used doctrine I guess it would automatically cast it

jolelievre (dismissed):

src/Adapter/Product/Repository/ProductRepository.php
jolelievre: Just a comment to improve this exception message, but it's not blocking

src/Adapter/Product/Update/RelatedProductsUpdater.php
jolelievre: Yeah I agree with the redundant name, I thought so as well at first but since we used this for ProductSupplier I guess it's ok 🤷‍♂️

src/Adapter/Product/Repository/ProductRepository.php
zuk3975: didn't put it to save additional loop.
In case we don't care about the loop, I suppose we could better make some DTO for this instead of array.

src/Adapter/Product/QueryHandler/GetPackedProductsHandler.php
matks: I think our Handlers are built with decoration in mind, not extension 👍  so it's good

matks: @PierreRambaud close to the "private VS protected" properties debate, we might have a "final or not" debate about our classes 🤔  wdyt ?

src/Adapter/Product/Repository/ProductRepository.php
```diff
@@ -64,6 +84,61 @@ public function assertProductExists(ProductId $productId): void
...
+     *
+     * @throws CoreException
+     */
+    public function getRelatedProducts(ProductId $productId, LanguageId $languageId): array
+    {
+        $this->assertProductExists($productId);
+        $productIdValue = $productId->getValue();
+
+        try {
+            $accessories = Product::getAccessoriesLight($languageId->getValue(), $productIdValue);
```
matks, line 136: the famous "accessories" 😄

matks (approved):

jolelievre (approved):

matks:
Thank you @zuk3975

PR #21425, opened 2020-10-13
Add autocomplete attributes on FO login form

### LINKED ISSUE #14506: Browser console suggestion for the password form
**Describe the bug**
What do you think about this suggestion in console? Should we need to implement it in the code?

**To Reproduce**
1. Go to 'Sign in/Sign Up' page
2. Open browser consolve
3. See suggestion

**Screenshots**
Untitled

**Additionnal information**
PrestaShop version: PrestaShop 1.7.6 RC

NeOMakinG:
| Questions     | Answers
| Branch?       | develop
| Description?  | Autocomplete is enabled by default on many browsers, but it can provide warnings or fill wrong information, this is better to force them.
| Type?         | improvement
| Category?     | FO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #14506.
| How to test?  | Go on login form, log for a first time if you want, then see if chrome provide you an autocomplete (if you saved you'r mail/password inside it (you may already refused that years ago, this means that you won't have an autocomplete), also a lot of browsers have this by default, but it display a warning if you doesn't have it, it may change nothing exept kicking the warning

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21425)

### DIFF
--- a/classes/form/FormField.php
+++ b/classes/form/FormField.php
@@ -187,7 +187,7 @@ class FormFieldCore
         return $this->constraints;
     }
 
-    public function setAutocompleteAttribute($autocomplete): FormFieldCore
+    public function setAutocompleteAttribute(string $autocomplete): FormFieldCore
     {
         $this->autocomplete = $autocomplete;
 
--- a/classes/form/FormField.php
+++ b/classes/form/FormField.php
@@ -34,6 +34,7 @@ class FormFieldCore
     private $maxLength = null;
     private $errors = [];
     private $constraints = [];
+    private $autocomplete = '';
 
     public function toArray()
     {
@@ -46,6 +47,7 @@ class FormFieldCore
             'availableValues' => $this->getAvailableValues(),
             'maxLength' => $this->getMaxLength(),
             'errors' => $this->getErrors(),
+            'autocomplete' => $this->getAutocomplete(),
         ];
     }
 
@@ -180,5 +182,17 @@ class FormFieldCore
     public function getConstraints()
     {
         return $this->constraints;
+    }
+
+    public function setAutocomplete($autocomplete)
+    {
+        $this->autocomplete = $autocomplete;
+
+        return $this;
+    }
+
+    public function getAutocomplete()
+    {
+        return $this->autocomplete;
     }
 }
--- a/classes/form/FormField.php
+++ b/classes/form/FormField.php
@@ -34,6 +34,9 @@ class FormFieldCore
     private $maxLength = null;
     private $errors = [];
     private $constraints = [];
+    /**
+     * @var string
+     */
     private $autocomplete = '';
 
     public function toArray()
--- a/classes/form/FormField.php
+++ b/classes/form/FormField.php
@@ -187,6 +187,11 @@ class FormFieldCore
         return $this->constraints;
     }
 
+    /**
+     * @param string $autocomplete
+     *
+     * @return FormFieldCore
+     */
     public function setAutocompleteAttribute(string $autocomplete): FormFieldCore
     {
         $this->autocomplete = $autocomplete;
@@ -194,6 +199,9 @@ class FormFieldCore
         return $this;
     }
 
+    /**
+     * @return string
+     */
     public function getAutocompleteAttribute(): string
     {
         return $this->autocomplete;
--- a/classes/form/FormField.php
+++ b/classes/form/FormField.php
@@ -47,7 +47,7 @@ class FormFieldCore
             'availableValues' => $this->getAvailableValues(),
             'maxLength' => $this->getMaxLength(),
             'errors' => $this->getErrors(),
-            'autocomplete' => $this->getAutocomplete(),
+            'autocomplete' => $this->getAutocompleteAttribute(),
         ];
     }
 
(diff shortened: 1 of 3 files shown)


classes/form/FormField.php
```diff
@@ -181,4 +183,16 @@ public function getConstraints()
     {
         return $this->constraints;
     }
+
+    public function setAutocomplete($autocomplete)
```
matks:
```suggestion
    public function setAutocompleteAttribute($autocomplete)
```

Suggestion: be as accurate as possible in function name to make sure people understand what it is. With 'autocomplete' alone someone could think it's, for example, a feature that performs ajax queries to auto-complete what you type 😄

matks: Also @Progi1984 will tell you to type-hint this if I am not wrong 😛

classes/form/FormField.php
```diff
@@ -181,4 +183,16 @@ public function getConstraints()
...
+    {
+        $this->autocomplete = $autocomplete;
+
+        return $this;
+    }
+
+    public function getAutocomplete()
```
matks: @Progi1984 will tell you to type-hint the return type, I'm sure 😄 
```suggestion
    public function getAutocompleteAttribute(): string
```

themes/classic/templates/_partials/form-fields.tpl
```diff
@@ -138,6 +138,7 @@
               name="{$field.name}"
               title="{l s='At least 5 characters long' d='Shop.Forms.Help'}"
               type="password"
+              {if $field.autocomplete}autocomplete="{$field.autocomplete}"{/if}
```
matks, line 141: Should'nt we use Smarty `escape` function ?

And I know this is not the scope of this PR but maybe we should (in another PR) apply the `escape` to EVERY other value in this TPL 😅 

Wdyt @PierreRambaud ?

matks (commented):
Hi @NeOMakinG nice PHP PR 😄 

PR is mainly good, I provided some suggestions to improve the details but there are no major issues in it

classes/form/FormField.php
NeOMakinG: What about every other fields which are not typed at all ?

I'm not in favor of typing only on method while every siblings methods are not typed at all, that should be a complete refacto don't you think?

themes/classic/templates/_partials/form-fields.tpl
NeOMakinG: Waiting for @PierreRambaud advice on this

classes/form/FormField.php
matthieu-rolland: I understand the will for homogeneity, but on a practical aspect, what would be the true downside of type-hinting the return type of new methods we add in legacy classes ?

NeOMakinG:
@matks Did type methods I added, but should I type methods already defined before this PR ? I think this could be a BC

classes/form/FormField.php
```diff
@@ -34,6 +34,7 @@ class FormFieldCore
     private $maxLength = null;
     private $errors = [];
     private $constraints = [];
+    private $autocomplete = '';
```
Progi1984, line 40:
```suggestion
    /**
     * @var string
     */
    private $autocomplete = '';
```

classes/form/FormField.php
```diff
@@ -181,4 +183,16 @@ public function getConstraints()
     {
         return $this->constraints;
     }
+
+    public function setAutocompleteAttribute($autocomplete): FormFieldCore
```
Progi1984:
```suggestion
    /**
     * @param string $autocomplete
     *
     * @return FormFieldCore
     */
    public function setAutocompleteAttribute(string $autocomplete): FormFieldCore
```

Progi1984 (commented):
Small improvemnts

classes/form/FormField.php
```diff
@@ -181,4 +186,16 @@ public function getConstraints()
...
     }
+
+    public function setAutocompleteAttribute(string $autocomplete): FormFieldCore
+    {
+        $this->autocomplete = $autocomplete;
+
+        return $this;
+    }
+
+    public function getAutocompleteAttribute(): string
```
Progi1984:
```suggestion
    /**
     * @param string $autocomplete
     *
     * @return FormFieldCore
     */
    public function setAutocompleteAttribute(string $autocomplete): FormFieldCore
    {
        $this->autocomplete = $autocomplete;

        return $this;
    }

    /**
     * @return string
     */
    public function getAutocompleteAttribute(): string
```

Progi1984 (approved):

themes/classic/templates/_partials/form-fields.tpl
PierreRambaud: AFAIK, smarty already automatically escape html
`$smarty->escape_html = true;`
https://www.smarty.net/docs/en/variable.escape.html.tpl

PierreRambaud (approved):

khouloudbelguith:
Hi @NeOMakinG,

It is ok :heavy_check_mark: 
https://drive.google.com/file/d/1gL6FydS2jVxbVKkNjxuSEVdjaYbwUkqm/view?usp=sharing
Without PR:
image

Thanks!

PR #21429, opened 2020-10-13
Fix symfony cache clear

### LINKED ISSUE #21413: Cache clear issue after optimization
#### Describe the bug

The way PrestaShop clears Symfony cache has been modified by this PR https://github.com/PrestaShop/PrestaShop/pull/20942

Although it is more performant, it also doesn't clear everything we need because the container is not rebuilt.
Among the things that I fear might be involved:
- Symfony routing updated
- Doctrine entities management
- Module's services updated

At least the routing has been proved to be failing with the module `ps_linklist` the other two remain to be confirmed, `productcomments` which uses both seemed to be fine but it's hard to tell if it's because the container was fully rebuilt after its installation.

The `Cache clear` feature is also impacted since it relies on the same method `SymfonyCacheClearer::clear`

#### Expected behavior

The Symfony cache should be completely cleared to ensure the container is correctly rebuilt with up to date information. However we must keep in mind that these two issues are still correctly handled:
https://github.com/PrestaShop/PrestaShop/issues/20852
https://github.com/PrestaShop/PrestaShop/issues/11105

#### Steps to Reproduce

Steps to reproduce the behavior (bug for routing at least):

1. Go to Module Manager
2. Uninstall `ps_linklist`
3. Remove the cache manually: `rm -fR ./var/cache/*`
4. Go to Module Catalog
5. Refresh the page so that the menu is updated (this should be handled by MBO)
6. Click on Design > LinkWidget
7. See the error from screenshot, the url should be managed by Symfony router via the `_legacy_link` property but since the Router info were not updated it tries to access the legacy controller which does not exist any more

**Screenshots**

Capture d’écran 2020-10-13 à 12 31 01

#### Potential solution

We need to revert the command used back to `cache:clear` however it cause some issues because the memory limit was reached So one possibility is to dynamically force the memory_limit when we perform this command

#### Additional information

* PrestaShop version: 1.7.8
* PHP version: N/A

jolelievre:
| Questions     | Answers
| Branch?       | develop
| Description?  | Remove pools:prune command which does not clear the container enough Launch the command only on process shutdown to avoid error during process and memory overloading In previous fix the Clear cache command was not executed correctly, the input arguments were not correctly set (use `'--no-warmup' => true,` not `'--no-warmup',`)
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #21413
| How to test?  | See issue to reproduce issue regarding `ps_linklist` routing problem Also try to uninstall, clear cache and reinstall `productcomments` to check Doctrine is correctly initialized

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21429)

### DIFF
--- a/src/Adapter/Cache/Clearer/SymfonyCacheClearer.php
+++ b/src/Adapter/Cache/Clearer/SymfonyCacheClearer.php
@@ -55,15 +55,18 @@ final class SymfonyCacheClearer implements CacheClearerInterface
             return;
         }
 
-        $application = new Application($kernel);
-        $application->setAutoExit(false);
+        register_shutdown_function(function () use ($kernel) {
+            $application = new Application($kernel);
+            $application->setAutoExit(false);
 
-        $input = new ArrayInput([
-            'command' => 'cache:pool:prune',
-        ]);
+            $input = new ArrayInput([
+                'command' => 'cache:clear',
+                '--no-warmup',
+            ]);
 
-        $output = new NullOutput();
-        $application->run($input, $output);
-        Hook::exec('actionClearSf2Cache');
+            $output = new NullOutput();
+            $application->run($input, $output);
+            Hook::exec('actionClearSf2Cache');
+        });
     }
 }
--- a/src/Adapter/Cache/Clearer/SymfonyCacheClearer.php
+++ b/src/Adapter/Cache/Clearer/SymfonyCacheClearer.php
@@ -56,6 +56,14 @@ final class SymfonyCacheClearer implements CacheClearerInterface
         }
 
         register_shutdown_function(function () use ($kernel) {
+            // The cache may have been removed by Tools::clearSf2Cache, it happens during install
+            // process, in which case we don't run the cache:clear command because it is not only
+            // useless it will simply fail as the container caches classes have been removed
+            $cacheDir = _PS_ROOT_DIR_ . '/var/cache/' . _PS_ENV_ . '/';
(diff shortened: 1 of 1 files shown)


jolelievre:
Note: there was only one thing I didn't manage to handle perfectly When you manually remove the cache folder before re-installing the module The install process happens correctly however the Configure button is not visible

However this only happens when you remove the cache manually, when you uninstall the module from the BO the cache is correctly cleared (no info regarding ps_linklist routing can be found in the cache folder), and when you reinstall it the cache is correctly rebuilt with appropriate routing infos And the Configure button is correctly displayed

Since removing the cache manually is not supposed to be done, and since it's not blocking anyway I think this "glitch" is acceptable

matthieu-rolland (approved):

atomiix (approved):

ngodefroy:
Hello @jolelievre 

What I tested:

Go to Module Manager
Uninstall ps_linklist
Remove the cache manually: rm -fR ./var/cache/*
Go to Module Catalog
Refresh the page so that the menu is updated (this should be handled by MBO)
Click on Design > Linkwidget is deleted

Same test without the removal of cache manually

uninstalled Product comment module
Cleared cache (manually & not)
Reinstalled it
Changed configuration into module
Saved

**TEST OK**

Thank you

Progi1984:
Thanks @jolelievre & @ngodefroy

PR #21435, opened 2020-10-14
Add hooks for the AdminLoginController, AdminController, FrontController and Controller

### LINKED ISSUE #21379: Hook displayBackOfficeTop not being called anymore on Login BO Page in 1.7.7
#### Describe the bug

I can no longer read post parameters from backofficetop hook on login page on 1.7.7. Please see below example:

#### Expected behavior

PS 1.6.1 - 1.7.6:
```
public function hookDisplayBackOfficeTop($params)
{
    if (Tools::isSubmit('submitLogin')) {
        exit;
    }
}
```
Go to login page -> Submit -> Die as expected

PS 1.7.7:
```
public function hookDisplayBackOfficeTop($params)
{
    if (Tools::isSubmit('submitLogin')) {
        exit;
    }
}
```
Go to login page -> Submit -> Logged in (not expected)

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Add some hooks for better flexibility.
| Type?         | improvement 
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #21379 
| How to test?  | Hooks are ok :sweat_smile: .

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21435)

### DIFF
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -255,6 +255,13 @@ class FrontControllerCore extends Controller
      */
     public function init()
     {
+        Hook::exec(
+            'actionFrontControllerInitBefore',
+            [
+                'controller' => $this,
+            ]
+        );
+
         /*
          * Globals are DEPRECATED as of version 1.5.0.1
          * Use the Context object to access objects instead.
@@ -465,7 +472,12 @@ class FrontControllerCore extends Controller
         $this->context->cart = $cart;
         $this->context->currency = $currency;
 
-        Hook::exec('actionFrontControllerAfterInit');
+        Hook::exec(
+            'actionFrontControllerInitAfter',
+            [
+                'controller' => $this,
+            ]
+        );
     }
 
     /**
--- a/install-dev/data/xml/hook.xml
+++ b/install-dev/data/xml/hook.xml
@@ -776,10 +776,71 @@
       <title>Manage elements added to the "prestashop" javascript object</title>
       <description>This hook allows you to customize the "prestashop" javascript object that is included in all front office pages</description>
     </hook>
-    <hook id="actionFrontControllerAfterInit">
-      <name>actionFrontControllerAfterInit</name>
+    <hook id="actionFrontControllerInitAfter">
+      <name>actionFrontControllerInitAfter</name>
       <title>Perform actions after front office controller initialization</title>
       <description>This hook is launched after the initialization of all front office controllers</description>
+    </hook>
+    <hook id="actionFrontControllerInitBefore">
+      <name>actionFrontControllerInitBefore</name>
+      <title>Perform actions before front office controller initialization</title>
+      <description>This hook is launched before the initialization of all front office controllers</description>
+    </hook>
+    <hook id="actionAdminControllerInitAfter">
+      <name>actionAdminControllerInitAfter</name>
+      <title>Perform actions after admin controller initialization</title>
+      <description>This hook is launched after the initialization of all admin controllers</description>
+    </hook>
+    <hook id="actionAdminControllerInitBefore">
+      <name>actionAdminControllerInitBefore</name>
+      <title>Perform actions before admin controller initialization</title>
+      <description>This hook is launched before the initialization of all admin controllers</description>
+    </hook>
+    <hook id="actionControllerInitAfter">
+      <name>actionControllerInitAfter</name>
+      <title>Perform actions after controller initialization</title>
+      <description>This hook is launched after the initialization of all controllers</description>
+    </hook>
+    <hook id="actionControllerInitBefore">
+      <name>actionControllerInitBefore</name>
+      <title>Perform actions before controller initialization</title>
+      <description>This hook is launched before the initialization of all controllers</description>
+    </hook>
+
+    <hook id="actionAdminLoginControllerBefore">
+      <name>actionAdminLoginControllerBefore</name>
+      <title>Perform actions before admin login controller initialization</title>
+      <description>This hook is launched before the initialization of the login controller</description>
+    </hook>
+    <hook id="actionAdminLoginControllerLoginBefore">
+      <name>actionAdminLoginControllerLoginBefore</name>
+      <title>Perform actions before admin login controller login action initialization</title>
+      <description>This hook is launched before the initialization of the login action in login controller</description>
+    </hook>
+    <hook id="actionAdminLoginControllerLoginAfter">
+      <name>actionAdminLoginControllerLoginAfter</name>
+      <title>Perform actions after admin login controller login action initialization</title>
+      <description>This hook is launched after the initialization of the login action in login controller</description>
+    </hook>
+    <hook id="actionAdminLoginControllerForgotBefore">
+      <name>actionAdminLoginControllerForgotBefore</name>
+      <title>Perform actions before admin login controller forgot action initialization</title>
+      <description>This hook is launched before the initialization of the forgot action in login controller</description>
+    </hook>
+    <hook id="actionAdminLoginControllerForgotAfter">
+      <name>actionAdminLoginControllerForgotAfter</name>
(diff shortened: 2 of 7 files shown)


install-dev/upgrade/sql/1.7.7.0.sql
```diff
@@ -541,12 +541,26 @@ INSERT IGNORE INTO `PREFIX_hook` (`id_hook`, `name`, `title`, `description`, `po
...
+  (NULL, 'actionAdminLoginControllerLoginAfter', 'Perform actions after admin login controller login action initialization', 'This hook is launched after the initialization of the login action in login controller'),
+  (NULL, 'actionAdminLoginControllerForgotBefore', 'Perform actions before admin login controller forgot action initialization', 'This hook is launched before the initialization of the forgot action in login controller'),
+  (NULL, 'actionAdminLoginControllerForgotAfter', 'Perform actions after admin login controller forgot action initialization', 'This hook is launched after the initialization of the forgot action in login controller'),
+  (NULL, 'actionAdminLoginControllerResetBefore', 'Perform actions before admin login controller reset action initialization', 'This hook is launched before the initialization of the reset action in login controller'),
+  (NULL, 'actionAdminLoginControllerResetBefore', 'Perform actions after admin login controller reset action initialization', 'This hook is launched after the initialization of the reset action in login controller')
```
Progi1984:
```suggestion
  (NULL, 'actionAdminLoginControllerResetAfter', 'Perform actions after admin login controller reset action initialization', 'This hook is launched after the initialization of the reset action in login controller')
```

classes/controller/FrontController.php
```diff
@@ -465,7 +472,12 @@ public function init()
         $this->context->cart = $cart;
         $this->context->currency = $currency;
 
-        Hook::exec('actionFrontControllerAfterInit');
+        Hook::exec(
+            'actionFrontControllerInitAfter',
+            [
+                'controller' => $this,
```
kpodemski, line 478: @PierreRambaud don't we have an access to `$this->context->controller` here? is there any point to provide `controller` parameter?

kpodemski: @PierreRambaud same in few different places where for example in `Controller` constructor we have `$this->context->controller = $this`

jolelievre: Always relying on the context is not a good practice, this is our main struggle in fixing so many bugs because it forces us to override the context, or mock the context, or init a missing context, ...

Imho hooks, like Symfony events, should provide the parameters related to their scope so that they can be handled independently and not relying on the context (or the least possible)

kpodemski: @jolelievre thanks, interesting point, I'd be happy to read more about possible issues when it comes to `$this->context->controller` vs `$params['controller']` in application which is already initialized and where `Context` should be our center of truth 👍

Progi1984 (dismissed):

classes/controller/FrontController.php
PierreRambaud: We should only trust what we have in parameters. The Context can be easily modified by everything.

Progi1984 (dismissed):

classes/controller/FrontController.php
jolelievre: Context has been the center of truth for years indeed, but the result is that everybody is used to access and, more dangerous, modify it (even inside the core code, it's not just in modules)

The problem is that all of the legacy code relies on the context in every single piece of code, meaning we can't reuse the legacy classes without having perfectly initialized or duplicated the Context Which makes it very hard to reuse, and even harder to test

So of course you will get the same thing in `$this->context->controller` and in `$params['controller']` but since the goal in the future is to remove the Context it's better to improve the rest of the code piece by piece so that we progressively stop relying on it

Progi1984:
@PierreRambaud Please fix your commits : a Merge Commit has been added.

classes/controller/FrontController.php
kpodemski: @PierreRambaud yeah but like... entire PrestaShop and its modules are relying on `Context`, what else do you recommend for checking current `Shop`, `Currency`, `Language`? :) We have `Context` injected everywhere in Symfony services 👀 

We literally don't have any hooks where you give parameters explicitly, even entire `Cart` mechanism is based on `Context`, not only in legacy part of code.

I'm just asking for anyone visiting this PR and for myself how can we approach this issue, what's the better way of getting all informations about current `state` of PrestaShop app, currency etc. We can't have immutable state, we need to be able to change `Context` in some scenarios :) 

Also I don't really see a way to override `$this->context->controller` in this place but that doesn't matter 👍 

I'd be happy to discuss this on Slack or something because you and @jolelievre give a pretty controversial advice ;)

PierreRambaud:
@Progi1984 Merge commit has been removed.

classes/controller/FrontController.php
PierreRambaud: Yes, it should be discussed on slack or in another issue. :sweat_smile:

Progi1984 (approved):

classes/controller/FrontController.php
jolelievre: You're right we rely on it everywhere and it's become the perfect shortcut, I'm not saying we can stop using it today but most of the object you mentioned have alternatives (some are already possible, some not, it's just to explain the ideal way):

- Currency should be fetched based on the Cart or Order you are editing
  - if it's to display something on front it should be fetched from the Customer's preferences
- Language should be fetched based on the employee or customer in the session
- Shop is trickier ^^ Sometimes it indeed depends on the context, sometimes (like when you edit an order) you shouldn't use the Context shop at all, but the one that was associated to the Order on creation

The thing is most of the time we don't even think of what we really need (should I use the shop assigned to the cart/order or the one from the context) and the context value is always used by default But it's not always the appropriate one Which is why the singleton approach has many flows

So it might seem controversial, sometimes you have no choice but to use the Context, but when given the possibility to do otherwise (especially when parameters adapted to the current action are provided) it should be favored versus using the Context

Now we clearly lack some proper alternatives Especially in Symfony services initialization where we still heavily rely on LegacyContext to inject language, shop or other stuffs However we always **try** not to inject Context directly, but rather the info we need specifically (like default language, or current shop context)

The road is long but we can still try to improve it ^^

jolelievre (approved):

sowbiba (approved):

khouloudbelguith:
Hi @PierreRambaud,

It is ok :heavy_check_mark: 

for example:
BO
image
Check the file on server
image

Thanks!

install-dev/data/xml/hook.xml
```diff
@@ -776,11 +776,72 @@
...
+      <name>actionAdminLoginControllerForgotAfter</name>
+      <title>Perform actions after admin login controller forgot action initialization</title>
+      <description>This hook is launched after the initialization of the forgot action in login controller</description>
+    </hook>
+    <hook id="actionAdminLoginControllerResetBefore">
+      <name>actionAdminLoginControllerResetBefore</name>
+      <title>Perform actions before admin login controller reset action initialization</title>
+      <description>This hook is launched before the initialization of the reset action in login controller</description>
+    </hook>
+    <hook id="actionAdminLoginControllerResetBefore">
```
ghost, line 840: There are two times <<  actionAdminLoginControllerResetBefore >>

classes/controller/FrontController.php
```diff
@@ -465,7 +472,12 @@ public function init()
         $this->context->cart = $cart;
         $this->context->currency = $currency;
 
-        Hook::exec('actionFrontControllerAfterInit');
```
jf-viguier, line 468: @PierreRambaud The actionFrontControllerAfterInit hook suppression is a BC break and should be documented I guess

PierreRambaud: Yes, the hook has been renamed.

PR #21447, opened 2020-10-14
Synchronize OrderDetail with Cart when removing a voucher

### LINKED ISSUE #21438: BO - Order - An exception displayed when removing a discount (gift)
#### Describe the bug

In the BO > Order detail page > when we try to delete a discount ( amount of a gift) => an exception is displayed

#### Expected behavior

Option 1:
Delete the gifted product after deleting the voucher
Option 2 ISO:
Keep the product

The option 1 is preferred but depending on the complexity the option 2 could be chosen.
 
#### Steps to Reproduce

1. Go to BO > Catalog > Discounts page 
2. Create a cart rule "free gift "demo_6 60*90"
image

3. Go to FO > Add demo_6 "40*60" to cart
4. Apply the discount
5. Continue checkout
6. don't close the Order_confirmation page
6. in the BO > Order detail page > delete the discount => exception
image
7. go back to the FO > Order confirmation page and reload the page => see error
image

#### Additional information

* PrestaShop version: 177x
* PHP version: 7.2

PierreRambaud:
| Questions     | Answers
| Branch?       | 1.7.7.x
| Description?  | Sync the OrderDetail with the Cart when removing a voucher.
| Type?         | bug fix
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #21438

Huge thanks to @jolelievre 

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21447)

### Scenarios to test

Then there are a few scenarios:

Add from the cart
- add product A in the cart then checkout, the Order must contain Product A and Product B
   - remove the Product A from the order, the cart rule is removed and Product B is removed
   - remove the cart rule from the order, Product B is removed from the order along with other discounts
- add Product B and Product A in the cart then checkout, the Order must contain the Product A and 2 Product B
   - remove the Product A from the order, the cart rule is removed and Product B has one product left
   - remove the cart rule from the order, Product B has only one left and product A is still here

Add from the order: create a cart with any product, ckecout and go the the Order view (this can be done as one scenario)
- add Product A in the Order, the Product B is added along with the other discounts
- remove Product B, the product disappears but the discount is still present with left discount (shipping + 1e)
- add the Product B again, the discount is updated and the gift is still free
- remove discount, the Product B is removed along with other discounts, Product A is still there


src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
```diff
@@ -70,9 +84,63 @@ public function handle(DeleteCartRuleFromOrderCommand $command)
...
+            }
+            $giftOrderDetails[] = $orderDetail;
+        }
+
+        if (empty($giftOrderDetails)) {
+            return null;
+        }
+
+        $giftOrderDetailId = null;
+        // We try to find a row with at least 2 quantities
```
sowbiba: I like that :D

Progi1984 (dismissed):

src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
```diff
@@ -85,12 +100,74 @@ public function handle(DeleteCartRuleFromOrderCommand $command)
...
         try {
             // Delete Order Cart Rule and update Order
+            $cartRule = new CartRule($orderCartRule->id_cart_rule);
             $orderCartRule->softDelete();
             $cart->removeCartRule($orderCartRule->id_cart_rule);
 
-            $this->orderAmountUpdater->update($order, $cart, $orderCartRule->id_order_invoice);
+            // If cart rule was a gift product we must update an OrderDetail manually
+            $giftOrderDetail = $this->getGiftOrderDetail($order, $cartRule);
+            if (null !== $giftOrderDetail) {
```
eternoendless: IMO this should be
```suggestion
            if ($giftOrderDetail instanceof OrderDetail) {
```
especially since you use one of its properties just after

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -292,6 +294,13 @@ private function getProductFromCart(array $cartProducts, int $productId, int $pr
 
             return $productMatch && $combinationMatch ? $item : null;
         });
+
+        // This shouldn't happen, if it does something was not done before updating the Order (removing an OrderDetail maybe)
+        if (null === $cartProduct) {
+            throw new OrderException('Could not find the product in cart, meaning Order and Cart are not synced');
```
sowbiba: no translation needed for this ?

src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
```diff
@@ -85,12 +100,74 @@ public function handle(DeleteCartRuleFromOrderCommand $command)
...
+                 * Note: we are lucky the stock updates happens smoothly, it was not re-injected by removing the CartRule
+                 * because Cart doesn't update the stock but it will while we update the OrderDetail Lucky
+                 * for us it's the same difference of 1 so it's all accurate in the end
+                 */
+                // This calls the OrderAmountUpdater internally so no need to perform both calls
+                $this->orderProductQuantityUpdater->update(
+                    $order,
+                    $giftOrderDetail,
+                    $newQuantity,
+                    $giftOrderDetail->id_order_invoice != 0 ? new OrderInvoice($giftOrderDetail->id_order_invoice) : null,
```
eternoendless: Unless `id_order_invoice` can be a string
```suggestion
                    $giftOrderDetail->id_order_invoice !== 0 ? new OrderInvoice($giftOrderDetail->id_order_invoice) : null,
```
Or else
```suggestion
                    $giftOrderDetail->id_order_invoice == 0 ? null : new OrderInvoice((int) $giftOrderDetail->id_order_invoice),
```

atomiix (dismissed):

Progi1984 (dismissed):

src/Adapter/Order/OrderAmountUpdater.php
jolelievre: Not in error messages This one is not destined to be displayed on front

matthieu-rolland (commented):
LGTM for me once the small feedbacks already given have been addressed

src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
jolelievre: Since the function return `?OrderDetail` it can return either an OrderDetail or null
So it's equivalent, it's like checking `$result == true` or `$result != false` 😛

But I'll change it anyway

src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
jolelievre: Yes in legacy objects most of the time int are string I added casting and strict equality

Progi1984 (dismissed):

matthieu-rolland (dismissed):

src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
```diff
@@ -85,12 +100,74 @@ public function handle(DeleteCartRuleFromOrderCommand $command)
...
+        foreach ($order->getOrderDetailList() as $orderDetail) {
+            if ((int) $orderDetail['product_id'] !== $productId || (int) $orderDetail['product_attribute_id'] !== $productAttributeId) {
+                continue;
+            }
+            $giftOrderDetails[] = $orderDetail;
+        }
+
+        if (empty($giftOrderDetails)) {
+            return null;
+        }
```
eternoendless: I don't understand why you collect all if you only want the first one whose product_quantity is > 1.
You could do it here and get rid of the rest of the method:

```suggestion
        $fallbackOrderDetailId = null;
        $giftOrderDetailId = null;
        foreach ($order->getOrderDetailList() as $orderDetail) {
            if ((int) $orderDetail['product_id'] !== $productId || (int) $orderDetail['product_attribute_id'] !== $productAttributeId) {
                continue;
            }
            
            // We try to find a row with at least 2 items
            if ($orderDetail['product_quantity'] > 1) {
                $giftOrderDetailId = $orderDetail['id_order_detail'];
                return new OrderDetail($giftOrderDetailId);
            }

			// keep the first one for fallback
            $fallbackOrderDetailId ??= $orderDetail['id_order_detail'];
        }

        return (null === $fallbackOrderDetailId) ? null : new OrderDetail($fallbackOrderDetailId);
```

This may need some adjustments if `id_order_detail` can be null.

src/Adapter/Order/OrderAmountUpdater.php
eternoendless:
```suggestion
            throw new OrderException('Could not find the product in cart, meaning Order and Cart are out of sync');
```

src/Adapter/Order/OrderProductQuantityUpdater.php
```diff
@@ -101,7 +102,8 @@ public function update(
         Order $order,
         OrderDetail $orderDetail,
         int $newQuantity,
-        ?OrderInvoice $orderInvoice
+        ?OrderInvoice $orderInvoice,
+        $updateCart = true
```
eternoendless:
```suggestion
        bool $updateCart = true
```

src/Adapter/Order/Refund/OrderProductRemover.php
```diff
@@ -66,8 +66,9 @@ public function __construct(LoggerInterface $logger, TranslatorInterface $transl
     /**
      * @param Order $order
      * @param OrderDetail $orderDetail
+     * @param bool $updateCart Used when you don't want to update the cart (CartRule removal for example)
      */
-    public function deleteProductFromOrder(Order $order, OrderDetail $orderDetail)
+    public function deleteProductFromOrder(Order $order, OrderDetail $orderDetail, bool $updateCart = true)
```
eternoendless:
```suggestion
    public function deleteProductFromOrder(Order $order, OrderDetail $orderDetail, bool $updateCart = true): void
```

eternoendless (dismissed):

src/Adapter/Order/CommandHandler/DeleteCartRuleFromOrderHandler.php
jolelievre: Yep, that works too indeed!

jolelievre: `id_order_detail` can not be null, the array is a result from a query that SELECT * from order_detail
If this table is inconsistent there are some much bigger issues that shouldn't be related to this handler ^^

sowbiba (dismissed):

khouloudbelguith:
Hi @PierreRambaud,

I found an exception following these steps:
1. Create a cart rule 
1.1 Information: without code
1.2 Conditions: Product selection demo_6
image
1.3 Actions: Free shipping + discount amount = 1euro tax included + Free gift (demo_11)
image
2. Go to FO > Add the demo_6  to cart ( discount applied)
3. Continue checkout (choose My carrier as the shipping method)
4. Go to BO > Order detail page
5. Try to remove “demo_6” from the order
6. Nothing happens => see an error in the console
image
7. Reload the page: product well removed
8. Delete the voucher => see the error
image

Thanks!

sowbiba (commented):
LGTM
I would add some comment (in the phpdoc) of each newly created method to briefly explain what it does
And add scenarii to test in the PR description

sowbiba (dismissed):

tests/Integration/Behaviour/Features/Scenario/Order/order_gift_cart_rule.feature
```diff
@@ -0,0 +1,932 @@
...
+      | total_shipping_tax_incl  | 7.42   |
+    And product "Test Product With Auto Gift" in order "bo_order1" has following details:
+      | product_quantity            | 1     |
+      | product_price               | 12.00 |
+      | unit_price_tax_incl         | 12.72 |
+      | unit_price_tax_excl         | 12.00 |
+      | total_price_tax_incl        | 12.72 |
+      | total_price_tax_excl        | 12.00 |
+    And order "bo_order1" should contain 0 product "Test Product Gifted"
+    And the available stock for product "Test Product Gifted" should be 100
```
sowbiba, line 722: great

sowbiba (dismissed):

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -202,60 +215,155 @@ private function assertOrderWasNotShipped(Order $order)
...
 
     /**
-     * This function extracts the newly added product from the cart and reformat the data in order to create a
-     * dedicated OrderDetail with appropriate amounts
+     * This function extracts the newly added product from the cart it detects the newly added products,
+     * for the one added via the command it reformats its quantity so that the created OrderDetail has
+     * the appropriate amount It can also detect other added products (like gift related to CartRule)
      *
      * @param Cart $cart
+     * @param array $oldProducts
```
Progi1984, line 308:
```suggestion
     * @param array $oldProducts
     * @param array $newProducts
```

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
```diff
@@ -202,60 +215,155 @@ private function assertOrderWasNotShipped(Order $order)
...
+            }
+
+            $deltaQuantity = (int) $newProduct['cart_quantity'] - (int) $affectedProduct['cart_quantity'];
+            if (0 === $deltaQuantity) {
+                continue;
+            }
+
+            $affectedOrderDetail = null;
+            foreach ($orderDetails as $orderDetailData) {
+                if ((int) $orderDetailData['product_id'] === (int) $affectedProduct['id_product']
```
matks, line 279: It might be checked in another location in the code but ... could this become a negative integer 😬 ?

Progi1984 (dismissed):

src/Adapter/Order/OrderAmountUpdater.php
```diff
@@ -292,6 +292,13 @@ private function getProductFromCart(array $cartProducts, int $productId, int $pr
 
             return $productMatch && $combinationMatch ? $item : null;
         });
+
+        // This shouldn't happen, if it does something was not done before updating the Order (removing an OrderDetail maybe)
```
matks, line 296: Missing one word in the comment 🤔 I think

themes/classic/templates/checkout/_partials/order-confirmation-table.tpl
```diff
@@ -39,7 +39,7 @@
         <div class="order-line row">
           <div class="col-sm-2 col-xs-3">
             <span class="image">
-              {if $product.cover}
+              {if !empty($product.cover)}
```
matks, line 42: This change seems out of the scope of this PR 🤔 ?

matks (dismissed):

Progi1984 (dismissed):

src/Adapter/Order/OrderAmountUpdater.php
jolelievre: No, but maybe it's unclear indeed
```suggestion
        // This shouldn't happen, if it this happens it means something was not correctly done before calling OrderAmountUpdater::update (removing an OrderDetail maybe) and the cart and order are out of synce
```

themes/classic/templates/checkout/_partials/order-confirmation-table.tpl
jolelievre: This was added by Pierre at the time there was still a bug, it caused the front to crash I believe
Since the bugs in BO have been fixed it shouldn't happen any more but it's an additional safety

sowbiba (dismissed):

matks (dismissed):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -662,6 +662,11 @@ private function handleOutOfStockProduct(OrderForViewing $orderForViewing)
      */
     public function addProductAction(int $orderId, Request $request): Response
     {
+        /** @var OrderForViewing $orderForViewing */
+        $orderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId));
+
+        $previousProducts = $orderForViewing->getProducts()->getProducts();
```
kpodemski, line 668: @jolelievre `getProducts()->getProducts()`? 😱

Progi1984 (approved):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
PierreRambaud: Don't ask why, I already complained about it :p

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
```diff
@@ -700,23 +705,31 @@ public function addProductAction(int $orderId, Request $request): Response
         /** @var OrderForViewing $orderForViewing */
         $orderForViewing = $this->getQueryBus()->handle(new GetOrderForViewing($orderId));
 
-        $products = $orderForViewing->getProducts()->getProducts();
-        $lastProduct = $products[array_key_last($products)];
+        $updatedProducts = $orderForViewing->getProducts()->getProducts();
+        $newProducts = [];
+        for ($i = count($previousProducts); $i < count($updatedProducts); ++$i) {
```
kpodemski, line 710: @jolelievre am I understanding correctly that you want to return difference between current order products and those added?

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
kpodemski: @PierreRambaud gosh :( can't we... do an alias for this...? good luck with getting back to this part of code in few months :/

kpodemski: Ok, nvm, I see a structure now 😿

src/Adapter/Order/CommandHandler/AddProductToOrderHandler.php
kpodemski: @matks product ids are always unsigned 🤔

matks: Bad line 😅 I wanted to mention `$newUpdatedQuantity = (int) $affectedOrderDetail->product_quantity + $deltaQuantity;`

matks (approved):

atomiix (approved):

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: Not exactly, I want to return the last created products So the added product along with a potential gift added via cart rule

src/PrestaShopBundle/Controller/Admin/Sell/Order/OrderController.php
jolelievre: I agree with you the structure doesn't make much sense 🤔
The extra layer `OrderProductsForViewing` seems quite useless to me.. Sadly I'm afraid it's a bit late to change it Although a small refacto PR could probably be proposed and be validate with Behat tests only

But I'm afrais I already know the answer if we propose this now 😅

jolelievre: I guess the idea is that OrderForViewing is split into several pieces: `OrderProductsForViewing` `OrderShippingForViewing` and so on While it makes sens for classes that hold both a collection of sub entities and additional info (total count, recycled, ...) For products it does seem redundant 🤷‍♂️
But I see the logic now, at the very least the first getter should have been name differently 😛

ngodefroy:
Hello @PierreRambaud 

After @khouloudbelguith and I tested this PR, I didn't found any other issues except the one Khouloud found and that will be fixed on other PR.
Therefore, TEST OK 👍

khouloudbelguith:
Thanks, @ngodefroy!
The list of issues:
https://github.com/PrestaShop/PrestaShop/issues/21500
https://github.com/PrestaShop/PrestaShop/issues/21501
https://github.com/PrestaShop/PrestaShop/issues/21504
https://github.com/PrestaShop/PrestaShop/issues/21506

Thanks, @jolelievre @MatShir!

Progi1984:
Thanks @jolelievre & @khouloudbelguith & @ngodefroy

PR #21450, opened 2020-10-14
WIP: Check that hooks registered by a module have a related listener implemented

### LINKED ISSUE #19230: Notify when a module registers to a hook but implements no hook listener
Following #15702

When a module registers to a hook using `$this->registerHook()`, no check is performed to make sure that said module implements a hook method to listen to those hooks. This could happen for example when there's a typing mistake in the hook name.

In debug mode, throw an exception like this when that happens:

> Hook with the name %hookName% has been registered by %moduleClassName%, but the corresponding method %functionName% has not been defined in the Module class.

matthieu-rolland:
| Questions     | Answers
| Branch?       | develop
| Description?  | When installing a module, we want to check that all registered hooks have corresponding listeners implemented in the module, otherwise throw an exception;
| Type?         | improvement
| Category?     | CO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19230
| How to test?  | Install a module that registers an existing hook without implementing a listener for it. In debug mode.

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21450)

### DIFF
--- a/classes/Hook.php
+++ b/classes/Hook.php
@@ -583,13 +583,12 @@ class HookCore extends ObjectModel
             }
 
             // check that hook listener is implemented by the module
-            $expectedHookListenerMethod = sprintf('hook%s', ucfirst($hook_name));
-            if (_PS_MODE_DEV_ && !method_exists($module_instance, $expectedHookListenerMethod)) {
+            if (_PS_MODE_DEV_ && !Hook::isHookCallableOn($module_instance, $hook_name)) {
                 $missingHookListenerErrorMsg = sprintf(
                     'Hook with the name %s has been registered by %s, but the corresponding method %s has not been defined in the Module class.',
                     $hook_name,
                     get_class($module_instance),
-                    $expectedHookListenerMethod
+                    sprintf('hook%s', ucfirst($hook_name))
                 );
                 throw new PrestaShopException($missingHookListenerErrorMsg);
             }
--- a/classes/Hook.php
+++ b/classes/Hook.php
@@ -583,7 +583,7 @@ class HookCore extends ObjectModel
             }
 
             // check that hook listener is implemented by the module
-            if (_PS_MODE_DEV_ && !Hook::isHookCallableOn($module_instance, $hook_name)) {
+            if (_PS_MODE_DEV_ && !Hook::isHookCallableOn($module_instance, $hook_name) && !($module_instance instanceof WidgetInterface)) {
                 $missingHookListenerErrorMsg = sprintf(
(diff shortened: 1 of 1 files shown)


classes/Hook.php
```diff
@@ -582,6 +582,18 @@ public static function registerHook($module_instance, $hook_name, $shop_list = n
                 return false;
             }
 
+            // check that hook listener is implemented by the module
+            $expectedHookListenerMethod = sprintf('hook%s', ucfirst($hook_name));
```
matks: I think - I'm not sure - that hook callbacks are case-insensitive so we might need to check both `sprintf('hook%s', ucfirst($hook_name))` and `sprintf('hook%s', lcfirst($hook_name))` 🤔  ?

kpodemski: They are ineed case-insensitive, you can have `hookdisplayprodctadditionalinfo` 👿 and it still does work

@matthieu-rolland maybe you could use this method:
https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Hook.php#L386 ?

matthieu-rolland: Indeed, looks like the methods `getAllKnownNames` and `getHookAliasesFor` take care of the different possible hook names.

Fixed 👍

kpodemski:
@PrestaShop/prestashop-maintainers 

`1) LegacyTests\Integration\classes\module\ModuleGetPossibleHooksListTest::testGetRightListForModule
PrestaShopException: Hook with the name displayHome has been registered by bankwire, but the corresponding method hookDisplayHome has not been defined in the Module class.`

https://github.com/PrestaShop/PrestaShop/blob/develop/tests-legacy/Integration/classes/module/ModuleGetPossibleHooksListTest.php#L45

Shall we skip/delete this test? Changing legacy test to `ps_wirepayment` is an option? Not sure how we approach legacy tests

atomiix (dismissed):

PierreRambaud (changes requested):
Make sure you can install PrestaShop :wink:

matthieu-rolland:
> Make sure you can install PrestaShop 😉

yep, this PR can't be merged as long as the modules I fixed are not released...

I put the PR in WIP

matks:
@matthieu-rolland maybe we can put a list of the modules waiting for a fix in the github issue, so we can monitor the progress? 😉

PierreRambaud:
Hi @matthieu-rolland ,

I'm not sure the work is still in progress. Feel free to reopen it if you think you're able to work again on it.

Thanks!

PR #21468, opened 2020-10-15
Add event and object assign to symfony routes JS side

### LINKED ISSUE #19473: Extend JS Router to add routing from modules
## Describe the bug:

As described in symfony [documentation](https://symfony.com/doc/4.0/routing/generate_url_javascript.html), there is a solution to generate routing in javascript like so:
`<script>
var url = Routing.generate('blog_show', {
    'slug': 'my-blog-post'
});
</script>`

Is it possible to do that with PrestaShop?

## Expected solution:

This JS routing is already used in the core pages, as explained in this comment https://github.com/PrestaShop/PrestaShop/issues/19473#issuecomment-636381199 

It should be usable in a module for core routes (doc needed) however it is not possible to add custom route from modules.

## Possible solution 

Since rewriting the source from the core (`admin-dev/themes/new-theme/js/fos_js_routes.json`) would be complicated and a hassle, a more flexible solution would be to allow the modules to extend the router with their own routes.

Using an `event` system (or something similar), modules should be able to add their routes in the Router js class: https://github.com/PrestaShop/PrestaShop/blob/develop/admin-dev/themes/new-theme/js/components/router.js

From the module point of view the idea would be to generate its own json file (or any other format for that matter), and be able to do something like this (very rough idea):

```js
import {EventEmitter} from '@components/event-emitter';
import routes from `my_module/my_routes.json`;

EventEmitter.on('router.routes', (event) => {
    event.addRoutes(routes);
});
```

And in the router something like this:

```js
...
import {EventEmitter} from './event-emitter';
import RoutesEvent from `@components/events/routes-events`
...

export default class Router {
  constructor() {
    const routesEvent = new RoutesEvent();
    EventEmitter.emit('router.routes', routesEvent);
    Routing.setData(routesEvent.getRoutes());
    Routing.setBaseUrl($(document).find('body').data('base-url'));

    return this;
  }

  ...
}
```

NeOMakinG:
| Questions     | Answers
| Branch?       | develop
| Description?  | Modules wants to be able to erase or add new routes into the route object
| Type?         | improvement
| Category?     | BO
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes #19473.
| How to test?  | Should be QA tested by a dev, you should add an EventEmitter.on('beforeRoutes' and use the method returned with routes parameters to override, then console.log routes inside the `admin-dev/themes/new-theme/js/components/router.js` constructor at the end

Please don't close the issue at the merge of this PR as it's a quickwin but it doesn't resolve the whole issue

---
This change is [Reviewable](https://reviewable.io/reviews/prestashop/prestashop/21468)

### DIFF
--- a/admin-dev/themes/new-theme/js/components/router.js
+++ b/admin-dev/themes/new-theme/js/components/router.js
@@ -25,7 +25,6 @@
 
 import Routing from 'fos-routing';
 import routes from '@js/fos_js_routes.json';
-import {EventEmitter} from '@components/event-emitter';
 
 const {$} = window;
 
@@ -44,11 +43,9 @@ const {$} = window;
  */
 export default class Router {
   constructor() {
-    const addRoutes = (customRoutes) => {
-      Object.assign(routes.routes, customRoutes);
-    };
-
-    EventEmitter.emit('beforeRoutes', addRoutes);
+    if (window.prestashop && window.prestashop.customRoutes) {
+      Object.assign(routes.routes, window.prestashop.customRoutes);
+    }
 
     Routing.setData(routes);
     Routing.setBaseUrl(
--- a/admin-dev/themes/new-theme/js/components/router.js
+++ b/admin-dev/themes/new-theme/js/components/router.js
@@ -25,6 +25,7 @@
 
 import Routing from 'fos-routing';
 import routes from '@js/fos_js_routes.json';
+import {EventEmitter} from '@components/event-emitter';
 
 const {$} = window;
 
@@ -43,8 +44,18 @@ const {$} = window;
  */
 export default class Router {
   constructor() {
+    EventEmitter.on('updateRoutes', (customRoutes) => {
+      Object.assign(routes.routes, customRoutes);
+    });
+
+    EventEmitter.emit('beforeRoutes');
+
     Routing.setData(routes);
-    Routing.setBaseUrl($(document).find('body').data('base-url'));
+    Routing.setBaseUrl(
+      $(document)
+        .find('body')
+        .data('base-url'),
+    );
 
     return this;
   }
@@ -58,7 +69,11 @@ export default class Router {
    * @returns {String}
    */
   generate(route, params = {}) {
-    const tokenizedParams = Object.assign(params, {_token: $(document).find('body').data('token')});
+    const tokenizedParams = Object.assign(params, {
+      _token: $(document)
+        .find('body')
+        .data('token'),
+    });
 
     return Routing.generate(route, tokenizedParams);
   }
--- a/admin-dev/themes/new-theme/js/components/router.js
+++ b/admin-dev/themes/new-theme/js/components/router.js
@@ -44,11 +44,11 @@ const {$} = window;
  */
 export default class Router {
   constructor() {
-    EventEmitter.on('updateRoutes', (customRoutes) => {
+    const addRoutes = (customRoutes) => {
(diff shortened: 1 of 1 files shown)


atomiix (dismissed):

matthieu-rolland (dismissed):

NeOMakinG:
@PrestaShop/prestashop-maintainers could someone test it ? QA won't be able I think

PierreRambaud (dismissed):

Progi1984 (dismissed):

admin-dev/themes/new-theme/js/components/router.js
```diff
@@ -43,8 +44,18 @@ const {$} = window;
  */
 export default class Router {
   constructor() {
+    EventEmitter.on('updateRoutes', (customRoutes) => {
```
jolelievre: Question, why having a `beforeRoutes` even to to submit, and then the module must dispatch its own `updateRoutes` event to add its custom routes?

Its it easier to dispatch an event from the Router like `EventEmitter.emit('initRoutes', routeEvent);` which modules can register to, they would update the `routeEvent` to add their routes and then the Router can use them

It would avoid having two different events for one feature, and modules wouldn't need to emit anything they just listen to an event It feels more natural for modules to register to an event to extend a behaviour than dispatching their own

jolelievre: To make it easier the `routeEvent` would have methods like `routeEvent.addRoutes()` to make additions more easily

NeOMakinG: You'r right, gonna update it

atomiix (dismissed):

PierreRambaud (approved):

SD1982:
@PrestaShop/prestashop-core-developers We let you do the QA for this one :+1:

admin-dev/themes/new-theme/js/components/router.js
```diff
@@ -43,8 +44,18 @@ const {$} = window;
  */
 export default class Router {
   constructor() {
+    const addRoutes = (customRoutes) => {
+      Object.assign(routes.routes, customRoutes);
+    };
+
+    EventEmitter.emit('beforeRoutes', addRoutes);
```
PierreRambaud: After tests, we can't accept this change because the event is emitted before being able to plug on it :sweat_smile:

PierreRambaud (changes requested):
Need another solution. Let's try with window.prestashop.routes variable or something similar :)

Progi1984 (approved):

atomiix (approved):

admin-dev/themes/new-theme/js/components/router.js
```diff
@@ -43,8 +43,16 @@ const {$} = window;
  */
 export default class Router {
   constructor() {
+    if (window.prestashop && window.prestashop.customRoutes) {
```
PierreRambaud, line 46: Could be interesting to add a part in the devdoc with all window.prestashop properties and how to use and extend them.

NeOMakinG: https://devdocs.prestashop.com/1.7/development/components/global-components/

But it needs to be updated

matthieu-rolland (approved):

PierreRambaud (approved):

PierreRambaud:
Thank you bro

